PR #3571, opened 2017-10-17
Updated several dependencies

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | none
| Related issues/PRs | none
| License | MIT
| Documentation PR | none

#### What's in this PR?

This PR updates following dependencies:

* symfony/swiftmailer-bundle (not 3.1.1 because of https://github.com/symfony/swiftmailer-bundle/issues/198)
* doctrine/data-fixtures
* twig/twig

#### To Do

- [x] Tests


composer.json
```diff
@@ -57,7 +57,6 @@
         "symfony/phpunit-bridge": "^2.8.7 || ^3.0",
         "sensio/framework-extra-bundle": "~3.0",
         "symfony/monolog-bundle": "^2.8.7 || ^3.0",
-        "twig/twig": "~1.11",
```
danrot: Since we are directly using twig, we still should list it in here, instead of relying on another package to require it.

wachterjohannes: twig was required twice (in require and require-dev).

PR #3572, opened 2017-10-17
Added breadcrumb component

3rfan:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

A breadcrumb to indicate the current page location.

#### Why?

We need a breadcrumb for the media view

#### Example Usage

~~~js
const handleClick1 = () => {
    alert(`You clicked on crumb 1`);
};
const handleClick2 = () => {
    alert(`You clicked on crumb 2`);
};

<Breadcrumb>
    <Breadcrumb.Crumb onClick={handleClick1}>Root Directory</Breadcrumb.Crumb>
    <Breadcrumb.Crumb onClick={handleClick2}>...</Breadcrumb.Crumb>
    <Breadcrumb.Crumb>Current Directory</Breadcrumb.Crumb>
</Breadcrumb>
~~~

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Breadcrumb/Breadcrumb.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Breadcrumb/Breadcrumb.js
@@ -1,0 +1,35 @@
+// @flow
+import React from 'react';
+import type {ChildrenArray, Element} from 'react';
+import Icon from '../Icon';
+import Crumb from './Crumb';
+import breadcrumbStyles from './breadcrumb.scss';
+
+const ARROW_RIGHT = 'chevron-right';
+
+type Props = {
+    children: ChildrenArray<Element<typeof Crumb>>,
+};
+
+export default class Breadcrumb extends React.PureComponent<Props> {
+    static Crumb = Crumb;
+
+    render() {
+        const {
+            children,
+        } = this.props;
+
+        return (
+            <ul className={breadcrumbStyles.breadcrumb}>
+                {React.Children.map(children, (crumb, index) => (
+                    <li>
+                        {crumb}
+                        {(index !== React.Children.count(children) - 1) &&
+                            <Icon name={ARROW_RIGHT} className={breadcrumbStyles.arrow} />
+                        }
+                    </li>
+                ))}
+            </ul>
+        );
+    }
+}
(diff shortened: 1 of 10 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/components/Breadcrumb/Breadcrumb.js
```diff
@@ -0,0 +1,35 @@
...
+import breadcrumbStyles from './breadcrumb.scss';
+
+const ARROW_RIGHT = 'chevron-right';
+
+type Props = {
+    children: ChildrenArray<Element<typeof Crumb>>,
+};
+
+export default class Breadcrumb extends React.PureComponent<Props> {
+    static Crumb = Crumb;
```
danrot: Shouldn't that be `BreadCrumbs` and `BreadCrumb`? /cc @chirimoya

src/Sulu/Bundle/AdminBundle/Resources/js/components/Breadcrumb/Breadcrumb.js
```diff
@@ -0,0 +1,35 @@
...
+        const {
+            children,
+        } = this.props;
+
+        return (
+            <ul className={breadcrumbStyles.breadcrumb}>
+                {React.Children.map(children, (crumb, index) => (
+                    <li>
+                        {crumb}
+                        {(index !== React.Children.count(children) - 1) &&
```
danrot: Maybe move `React.Children.count` outside the loop? I don't think that this calculation has to be done in every iteration.

src/Sulu/Bundle/AdminBundle/Resources/js/components/Breadcrumb/README.md
```diff
@@ -0,0 +1,16 @@
...
+const handleClick1 = () => {
+    alert(`You clicked on crumb 1`);
+};
+const handleClick2 = (value) => {
+    alert(`You clicked on crumb ${value}`);
+};
+
+<Breadcrumb>
+    <Breadcrumb.Crumb onClick={handleClick1}>Root Directory</Breadcrumb.Crumb>
+    <Breadcrumb.Crumb value={2} onClick={handleClick2}>...</Breadcrumb.Crumb>
```
danrot: I think it is debatable if we should have the possibility of different `onClick` handlers for each breadcrumb. Shouldn't there be only one on the container component, which is always called with the passed value?

danrot: Would also be more consistent with the behavior of the `Tabs`, wouldn't it? And I don't think that having completely different click handlers for each breadcrumb is something we want to support anyway (?) /cc @chirimoya

src/Sulu/Bundle/AdminBundle/Resources/js/components/Breadcrumb/Breadcrumb.js
danrot: `BreadCrumb` and `BreadCrumb.Item` would probably be also ok for me.

PR #3585, opened 2017-10-23
Fixed exception of blame for none sulu users

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

It avoids to throw an exception in the blame subscriber when the current user is not a sulu user.

#### Why?

If the website user is not a sulu user it will throw `RuntimeException: Expected user object to be an instance of (Sulu) UserInterface` currently. It should just ignore this user and don't set changer or creator for this entity.

#### BC Breaks/Deprecations

BlameSubscriber will not longer throw an exception when symfony returns a none sulu user.

#### To Do

- [x] Testcase

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 CHANGELOG for Sulu
 ==================
 
+* 1.5.8 (unreleased)
+    * HOTFIX      #3584 [SuluPresistanceBUndle] Fixed exception of blame for none sulu users
+
 * 1.5.7 (2017-10-12)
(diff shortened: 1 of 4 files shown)


alexander-schranz:
@danrot the `$blameQueue` and `handleUserBlame` seems to be unused should I remove this? Or are this private function/attribute called over some reflection hacks?

danrot:
@alexander-schranz I hope it's not accessed via reflections... I would suggest to try removing it, and see if everything still works.

PR #3591, opened 2017-10-25
Fixed RouteProvider for get articles in different formats

### LINKED ISSUE #3579: RouteBundle: Umlauts in route-path
| Q | A
| --- | ---
| Bug? | yes
| New Feature? | no
| Sulu Version | 1.6.*
| Browser Version | ---

#### Actual Behavior

When creating a route with `ß` or `ä` you cannot change this letter to `s` or `a` because mysql matches this characters. But for the routing process, there is no match.

#### Expected Behavior

This is a MySQL issue - but we should find a way to avoid it.

#### Steps to Reproduce

As described:

1. Create an article (for example) with a `ß` in the route
2. Change the route to `s` and see that the form save it correctly.
3. But in the database there will stay the `ß`

#### Possible Solutions

Use `rawurlencode` to encode the path before save it in the database.

ducho:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #3579 
| Related issues/PRs | #3579 
| License | MIT

#### What's in this PR?

Fixed issue for getting articles in different formats. The solution was to strip format from end of the url and then will be found by RouteRepository. The next was necessary to correct RouteCollection for later usage in Symfony UrlMacher.

#### Why?

The problem was that RouteRepository was not able to find path in Route entity with different format as html. Response was No route found.

#### Example Usage

https://domain/locale/blog/some-seo-url-name.json|rss|xml

### DIFF
--- a/src/Sulu/Bundle/RouteBundle/Routing/RouteProvider.php
+++ b/src/Sulu/Bundle/RouteBundle/Routing/RouteProvider.php
@@ -215,7 +215,7 @@ class RouteProvider implements RouteProviderInterface
         /** @var RequestAttributes $attributes */
         $attributes = $request->get('_sulu');
 
-        if($attributes && $route->getPath() !== $attributes->getAttribute('requestUri')) {
+        if ($attributes && $route->getPath() !== $attributes->getAttribute('requestUri')) {
             $routePath = $attributes->getAttribute('requestUri');
         }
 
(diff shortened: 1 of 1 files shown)


trickreich:
@wachterjohannes Have you checked this?

wachterjohannes:
@trickreich @ducho there are a few things that are missing (tests, changelog, ...).

wachterjohannes:
was replaced by https://github.com/sulu/sulu/pull/3652

PR #3592, opened 2017-10-25
Basic preview implementation

### LINKED ISSUE #4075: Live preview not always updated
| Q | A
| --- | ---
| Bug? | yes
| New Feature? | no
| Sulu Version | 1.6.21, Sulu 2.0-pre-alpha1

#### Actual Behavior

Live preview not always updated.

#### Expected Behavior

Live preview should be updated always when changing things.

#### Steps to Reproduce

- [x] **Add or remove single internal link**

When a single internal link is changed live preview is not updated

- [x] **Images not updated (maybe only when render with twig extension)**

1. Use `get_image` from [MassiveTwigExtension](https://github.com/massiveart/web-twig/blob/master/docs/image.md) to render an image the image is not updated.

- [x] **Add new block**

1. When add a new block the live preview is not updated

- [x] **Add text to not yet render field**

2. If the field is not yet render the preview seems not be updated

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #4075
| Related issues/PRs | https://github.com/sulu/sulu-minimal/pull/111
| License | MIT
| Documentation PR | none

#### What's in this PR?

This PR added a basic implementation of preview.

#### Deprecations

**sulu_preview.object_provider**

Before:
```xml
<tag name="sulu_preview.object_provider" class="Sulu\Bundle\ArticleBundle\Document\ArticleDocument"/>
```

After:
```xml
<tag name="sulu_preview.object_provider" provider-key="articles"/>
```

#### Open Questions

- [x] Should `id="content"` and `{% block content %}`be required? 

#### To Do

- [x] alias => provider
- [x] cache everything around `#content` and always return the whole html
- [x] Webprofiler in preview
- [x] webspace route id unique
- [x] replace `href="*"` with `href="#"`
- [x] reload and update should not scroll
- [x] PHPUnit tests
- [x] Jest tests
- [x] Flow (iframe handling)
- [x] lint:js (findDOMNode)
- [x] Toolbar
- [x] Documentation
- [x] PR for sulu-minimal


src/Sulu/Bundle/AdminBundle/Resources/js/views/List/List.js
```diff
@@ -42,7 +43,7 @@ class List extends React.PureComponent<ViewProps> {
             observableOptions.locale = this.locale;
         }
 
-        this.datagridStore = new DatagridStore(resourceKey, observableOptions);
+        this.datagridStore = new DatagridStore(resourceKey, observableOptions, {webspace: webspace});
```
wachterjohannes: is this a good idea? maybe we could introduce a `apiOptions` prop?

alexander-schranz:
does this fix common Preview bugs: see #4075?

src/Sulu/Bundle/PreviewBundle/Preview/Renderer/PreviewRenderer.php
```diff
@@ -158,6 +158,8 @@ public function render($object, $id, $webspaceKey, $locale, $partial = false, $t
         }
 
         $attributes = $this->routeDefaultsProvider->getByEntity(get_class($object), $id, $locale, $object);
+        $attributes['preview'] = true;
```
wachterjohannes: @alexander-schranz i think this is the bug in master which should fix the preview and partial flag in the webiste controller

alexander-schranz: I'm not sure about object and partial if they are needed? 
https://github.com/sulu/sulu/pull/4146/files

wachterjohannes: partial is needed because of the `WebsiteController` to render only the content (without navigation, ...)

the object will be set by the `RouteDefaultsProvider` - so not needed to set afterwards

src/Sulu/Bundle/PreviewBundle/Preview/Renderer/PreviewRenderer.php
```diff
@@ -236,7 +236,7 @@ private function createServerAttributes(PortalInformation $portalInformation, Re
     {
         // get server parameters
         $server = [];
-        $host = $this->defaultHost;
+        $host = null;
```
danrot: Why is it set at all if it is just set to `null`?

src/Sulu/Bundle/PreviewBundle/Resources/config/routing.xml
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+
+<routes xmlns="http://symfony.com/schema/routing"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://symfony.com/schema/routing http://symfony.com/schema/routing/routing-1.0.xsd">
+
+    <route id="sulu_preview.start" path="/start">
+        <default key="_controller">sulu_preview.preview_controller:startAction</default>
+        <default key="_fos_rest_zone">true</default>
```
danrot: What was our decision here? Maybe it is really worth thinking about setting the zone in the config to `/admin`?

src/Sulu/Bundle/PreviewBundle/Resources/js/index.js
```diff
@@ -0,0 +1,17 @@
+// @flow
+import {bundleReady} from 'sulu-admin-bundle/services';
+import initializer from 'sulu-admin-bundle/services/Initializer';
+import {sidebarViewRegistry} from 'sulu-admin-bundle/containers/Sidebar';
+import Preview, {previewConfigStore} from './views/Preview';
+
+initializer.addUpdateConfigHook('sulu_preview', (config: Object) => {
+    previewConfigStore.setConfig(config);
```
danrot: I am not totally sure, but I think until now a `store` is something communicating with the server, i.e. sending some requests. I think this is not the case for this class, as it gets the config passed. We have the concept of registries, which contain instances of some other classes, but that also does not match here... Maybe just call it `previewConfig`? Also to be discussed.

src/Sulu/Bundle/PreviewBundle/Resources/js/package.json
```diff
@@ -0,0 +1,21 @@
...
+    "description": "Adds support for preview to Sulu",
+    "license": "MIT",
+    "repository": "https://github.com/sulu/sulu.git",
+    "dependencies": {
+        "debounce": "^1.0.2",
+        "mobx-react": "^5.0.0",
+        "query-string": "^6.0.0",
+        "react": "^15.6.0",
+        "react-dom": "^16.2.0",
+        "sulu-admin-bundle": "file:../../../AdminBundle/Resources/js"
```
danrot: We have changed the handling of dependencies: stuff like `mobx`, `mobx-react`, `react`, `react-dom` (do you even really need that one in here?) are now `peerDependencies`.

src/Sulu/Bundle/PreviewBundle/Resources/js/views/Preview/Preview.js
```diff
@@ -0,0 +1,265 @@
+// @flow
```
danrot: Is it a good idea to put that into the `view` folder? `Views` are normally components that have the `ViewProps` type, and are rendered into the main part of the application. Maybe we should introduce another folder for sidebar-views? or just leave it as a container? To be discussed.

src/Sulu/Bundle/PreviewBundle/Resources/js/views/Preview/Preview.js
```diff
@@ -0,0 +1,265 @@
...
+import {action, autorun, observable, toJS} from 'mobx';
+import {observer} from 'mobx-react';
+import debounce from 'debounce';
+import Router from 'sulu-admin-bundle/services/Router';
+import {sidebarStore} from 'sulu-admin-bundle/containers/Sidebar';
+import {FormStore} from 'sulu-admin-bundle/containers/Form';
+import ResourceStore from 'sulu-admin-bundle/stores/ResourceStore';
+import Requester from 'sulu-admin-bundle/services/Requester';
+import Toolbar from 'sulu-admin-bundle/components/Toolbar';
+import Loader from 'sulu-admin-bundle/components/Loader';
```
danrot: Usually we only import using only one folder deep, i.e. `sulu-admin-bundle/components` or `sulu-admin-bundle/containers`, since we consider this the "public" interface of our bundles.

src/Sulu/Bundle/PreviewBundle/Resources/js/views/Preview/Preview.js
```diff
@@ -0,0 +1,265 @@
...
+import previewStyles from './preview.scss';
+import previewConfigStore from './stores/PreviewConfigStore';
+
+type Props = {
+    resourceStore: ResourceStore,
```
danrot: As described above, could be probably removed.

src/Sulu/Bundle/PreviewBundle/Resources/js/views/Preview/Preview.js
```diff
@@ -0,0 +1,265 @@
...
+    formStore: FormStore,
+    router: Router,
+};
+
+@observer
+export default class Preview extends React.Component<Props> {
+    @observable iframeRef: ?Object;
```
danrot: Isn't there a better type than `Object`? I think `ElementRef` can be used for refs.

src/Sulu/Bundle/PreviewBundle/Resources/js/views/Preview/Preview.js
```diff
@@ -0,0 +1,265 @@
...
+            resourceStore,
+            formStore,
+            router: {
+                attributes: {
+                    locale,
+                },
+            },
+        } = this.props;
+
+        const route = previewConfigStore.generateRoute('start', {
```
danrot: Maybe this should be a `preview` service? `generateRoute` does not sound like a function that should be in a config object.

danrot: The the `preview` service could have a `start` method taking these parameters, and all the `Requester` instances could probably be gone from this component.

src/Sulu/Bundle/PreviewBundle/Resources/js/views/Preview/Preview.js
```diff
@@ -0,0 +1,265 @@
...
+            provider: resourceStore.resourceKey,
+            locale: locale,
+            id: resourceStore.id,
+        });
+        if (!route) {
+            return;
+        }
+
+        Requester.get(route).then((response) => {
+            this.setToken(response.token);
```
danrot: This state could also go into the `preview` service :-)

src/Sulu/Bundle/PreviewBundle/Resources/js/views/Preview/Preview.js
```diff
@@ -0,0 +1,265 @@
...
+        });
+
+        this.typeDisposer = autorun(() => {
+            if (formStore.loading || !this.iframeRef) {
+                return;
+            }
+
+            // $FlowFixMe
```
danrot: What is this for?

src/Sulu/Bundle/PreviewBundle/Resources/js/views/Preview/Preview.js
```diff
@@ -0,0 +1,265 @@
...
+            this.updateContext(formStore.type.get());
```
danrot: `formStore.type` should not need a `get` call?

src/Sulu/Bundle/PreviewBundle/Resources/js/views/Preview/Preview.js
```diff
@@ -0,0 +1,265 @@
...
+                return;
+            }
+
+            document.open();
+            document.write(response.content);
+            document.close();
+        });
+    }, previewConfigStore.debounceDelay);
+
+    updateContext = debounce((type) => {
```
danrot: Does the `debounce` make sense here? That's nothing that happens so often, or is it?

src/Sulu/Bundle/PreviewBundle/Resources/js/views/Preview/Preview.js
```diff
@@ -0,0 +1,265 @@
...
+
+        Requester.post(route, {context: {template: type}}).then((response) => {
+            const document = this.getPreviewDocument();
+            if (!document) {
+                return;
+            }
+
+            document.open();
+            document.write(response.content);
+            document.close();
```
danrot: That's not the first time I read that code :wink: I think it would really make sense to extract that in some way.

src/Sulu/Bundle/PreviewBundle/Resources/js/views/Preview/Preview.js
```diff
@@ -0,0 +1,265 @@
...
+        const route = previewConfigStore.generateRoute('stop', {token: this.token});
+        if (!route) {
+            return;
+        }
+
+        Requester.get(route);
+    }
+
+    getPreviewDocument = (): ?Document => {
+        // eslint-disable-next-line
```
danrot: You can also add the name of the route after `eslint-disable-next-line`, then the other checks still apply

src/Sulu/Bundle/PreviewBundle/Resources/js/views/Preview/Preview.js
```diff
@@ -0,0 +1,265 @@
...
+    handleToggleSidebarClick = () => {
+        if (sidebarStore.size === 'medium') {
+            return sidebarStore.setSize('large');
+        }
+
+        sidebarStore.setSize('medium');
+    };
+
+    handleRefreshClick = () => {
+        const document = this.getPreviewDocument();
```
danrot: Couldn't that be cached, instead of being read from the DOM every time?

src/Sulu/Bundle/PreviewBundle/Resources/js/views/Preview/README.md
```diff
@@ -0,0 +1,22 @@
...
+        router,
+        resourceStore,
+    } = this.props;
+
+    return {
+        view: 'sulu_preview.preview',
+        sizes: ['medium', 'large'],
+        props: {
+            router: router,
+            resourceStore: resourceStore,
```
danrot: Should be updated

src/Sulu/Bundle/PreviewBundle/Resources/js/views/Preview/stores/PreviewConfigStore.js
```diff
@@ -0,0 +1,46 @@
...
+import queryString from 'query-string';
+
+export type PreviewRouteName = 'start' | 'render' | 'update' | 'update-context' | 'stop';
+export type PreviewMode = 'auto' | 'on_request' | 'off';
+
+type PreviewConfigStoreConfig = {
+    routes: { [PreviewRouteName]: string },
+    debounceDelay: number,
+    mode: PreviewMode,
+};
```
danrot: We always put types into a separate `types.js` file, with the only exception of Props for react components.

src/Sulu/Bundle/PreviewBundle/Resources/js/views/Preview/stores/PreviewConfigStore.js
```diff
@@ -0,0 +1,46 @@
...
+
+class PreviewConfigStore {
```
danrot: Maybe we should make this the `PreviewStore`, and let it also make all the requests, which are currently done in the component?

src/Sulu/Bundle/PreviewBundle/Preview/Renderer/PreviewRenderer.php
wachterjohannes: the default was only used for websocket

src/Sulu/Bundle/PreviewBundle/Resources/config/routing.xml
wachterjohannes: @chirimoya

src/Sulu/Bundle/PreviewBundle/Resources/js/index.js
wachterjohannes: @chirimoya

src/Sulu/Bundle/PreviewBundle/Resources/js/views/Preview/Preview.js
wachterjohannes: @chirimoya

src/Sulu/Bundle/PreviewBundle/Resources/js/views/Preview/Preview.js
wachterjohannes: rerendering will break the reference then

src/Sulu/Bundle/PreviewBundle/Resources/js/views/Preview/stores/PreviewConfigStore.js
wachterjohannes: and add exact type `{| ... |}`

src/Sulu/Bundle/AdminBundle/Controller/AdminController.php
```diff
@@ -224,6 +238,17 @@ public function configAction(): Response
...
                 'contact' => $contact,
             ],
+            'sulu_preview' => [
+                'routes' => [
+                    'start' => $this->urlGenerator->generate('sulu_preview.start'),
+                    'render' => $this->urlGenerator->generate('sulu_preview.render'),
+                    'update' => $this->urlGenerator->generate('sulu_preview.update'),
+                    'update-context' => $this->urlGenerator->generate('sulu_preview.update-context'),
+                    'stop' => $this->urlGenerator->generate('sulu_preview.stop'),
+                ],
```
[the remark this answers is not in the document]
wachterjohannes: @chirimoya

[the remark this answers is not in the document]
chirimoya: config request is fine

src/Sulu/Bundle/AdminBundle/Resources/js/views/Form/tests/Form.test.js
```diff
@@ -286,11 +287,89 @@ test('Should add items defined in ToolbarActions to Toolbar', () => {
...
+    const withSidebar = require('../../../containers/Sidebar/withSidebar');
+    const Form = require('../Form').default;
+    const sidebarFunction = findWithHighOrderComponentFunction(withSidebar, Form);
+    const ResourceStore = require('../../../stores/ResourceStore').default;
+    const resourceStore = new ResourceStore('snippet', 1);
+
+    const route = {
+        options: {
+            toolbarActions: [],
+            preview: true,
```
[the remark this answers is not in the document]
chirimoya: preview is dependent of some state of the form data

src/Sulu/Bundle/PreviewBundle/Resources/config/routing.xml
chirimoya: `/admin`

src/Sulu/Bundle/PreviewBundle/Resources/js/index.js
chirimoya: `previewStore` only

src/Sulu/Bundle/PreviewBundle/Resources/js/views/Preview/Preview.js
chirimoya: `container`

chirimoya: `SidebarViewRegistry` => `SidebarRegistry`

src/Sulu/Bundle/AdminBundle/Controller/AdminController.php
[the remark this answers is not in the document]
wachterjohannes: we leave it like this

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Sidebar/registries/SidebarViewRegistry.js
```diff
@@ -3,6 +3,7 @@ import type {SidebarView} from '../types';
 
 class SidebarViewRegistry {
     views: {[string]: SidebarView};
+    disabled = [];
```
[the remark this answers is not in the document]
wachterjohannes: as we dont call it views now - i will leave it as it is now

src/Sulu/Bundle/AdminBundle/Resources/js/utils/TestHelper/TestHelper.js
```diff
@@ -1,7 +1,7 @@
 // @flow
 import type {Component} from 'react';
 
-function findWithToolbarFunction(withToolbar: any, Component: Class<Component<*, *>>): Function {
+function findWithHighOrderComponentFunction(withToolbar: any, Component: Class<Component<*, *>>): Function {
```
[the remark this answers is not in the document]
wachterjohannes: i will call it `highOrderComponent`

src/Sulu/Bundle/AdminBundle/Resources/js/views/Form/tests/Form.test.js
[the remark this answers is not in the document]
wachterjohannes: i will add here an expression

src/Sulu/Bundle/PreviewBundle/Preview/Renderer/PreviewRenderer.php
wachterjohannes: the default was only used for websocket

src/Sulu/Bundle/PreviewBundle/Resources/js/views/Preview/stores/PreviewConfigStore.js
wachterjohannes: this will be removed as soon as i refactor the store

wachterjohannes:
@danrot this is ready to review again

src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/stores/PreviewStore.js
```diff
@@ -0,0 +1,89 @@
...
+    const query = queryString.stringify(options);
+    if (query.length === 0) {
+        return PreviewStore.routes[name];
+    }
+
+    return PreviewStore.routes[name] + '?' + query;
+};
+
+export default class PreviewStore {
+    static routes: { [PreviewRouteName]: string } = {};
```
danrot: We don't do padding in the types if they are written in a single line. Looks like we only have an eslint rule for ordinary object and not flow types :confused:

src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/tests/Preview.test.js
```diff
@@ -0,0 +1,140 @@
...
+    const router = new Router({});
+
+    expect(render(<Preview formStore={formStore} router={router} />)).toMatchSnapshot();
+});
+
+test('React and update preview when data is changed', () => {
+    const resourceStore = new ResourceStore('pages', 1, {title: 'Test'});
+    const formStore = new FormStore(resourceStore);
+
+    // $FlowFixMe
```
danrot: You can save these comments if you build the data and loading stuff a bit different, as e.g. in https://github.com/sulu/sulu/blob/develop/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/tests/FormInspector.test.js#L7. Then you also don't pass the arguments from the `ResourceStore` for nothing :wink:

src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/tests/Preview.test.js
```diff
@@ -0,0 +1,140 @@
...
+    formStore.data = observable.map({title: 'Test'});
+    // $FlowFixMe
+    formStore.loading = false;
+    // $FlowFixMe
+    formStore.type = observable.box('default');
+
+    // $FlowFixMe
+    formStore.type = observable.box('default');
```
danrot: Any reason you are setting the type twice?

src/Sulu/Bundle/AdminBundle/Resources/js/views/Form/Form.js
```diff
@@ -199,7 +207,10 @@ class Form extends React.Component<Props> {
                 if (editRoute) {
                     router.navigate(
                         editRoute,
-                        {id: resourceStore.id, locale: resourceStore.locale, ...editRouteParameters}
+                        {
+                            id: resourceStore.id,
+                            locale: resourceStore.locale, ...editRouteParameters,
```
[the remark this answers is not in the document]
danrot: Looks like it would be this line: https://eslint.org/docs/rules/object-curly-newline

danrot:
The reload button does not work correctly everytime... It feels like it is a timing issue... E.g. when I set a breakpoint or write a `console.log` statement it works more often then when there is not... Is it possible that react sometimes updates the document before the `reload` method is called, and it is therefore not working?

danrot:
One more thing :see_no_evil: The preview doesn't show up in the SEO and excerpt tab, I guess that's because we are using the wrong `resourceKey`... And even if it would work, wouldn't the SEO title then update the title from the details tab in the preview?

If we don't find a good doable solution tomorrow, I think we should only activate the preview on the first tab at the moment, and undo your latest change.

PR #3597, opened 2017-10-31
 Only display analytics in the actual environment

trickreich:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes --
| Related issues/PRs | --
| License | MIT
| Documentation PR | --

#### What's in this PR?

 Only display analytics in the actual environment.

#### Why?

All analytics are displayed in the list. Even the wrong ones.

### DIFF
--- a/src/Sulu/Bundle/WebsiteBundle/Analytics/AnalyticsManagerInterface.php
+++ b/src/Sulu/Bundle/WebsiteBundle/Analytics/AnalyticsManagerInterface.php
@@ -19,13 +19,13 @@ use Sulu\Bundle\WebsiteBundle\Entity\Analytics;
 interface AnalyticsManagerInterface
 {
     /**
-     * Returns all analytics for given webspace.
+     * Returns all analytics for given webspace and environment.
      *
-     * @param string $webspaceKey
+     * @param string $environment
      *
      * @return Analytics[]
      */
-    public function findAll($webspaceKey);
+    public function findAll($webspaceKey, $environment);
 
     /**
      * Returns key by id.
--- a/src/Sulu/Bundle/WebsiteBundle/Analytics/AnalyticsManagerInterface.php
+++ b/src/Sulu/Bundle/WebsiteBundle/Analytics/AnalyticsManagerInterface.php
@@ -21,11 +21,11 @@ interface AnalyticsManagerInterface
     /**
      * Returns all analytics for given webspace and environment.
      *
-     * @param string $environment
+     * @param string $webspaceKey
      *
      * @return Analytics[]
      */
-    public function findAll($webspaceKey, $environment);
+    public function findAll($webspaceKey);
 
     /**
      * Returns key by id.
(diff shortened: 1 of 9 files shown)


src/Sulu/Bundle/WebsiteBundle/Tests/Functional/Analytics/AnalyticsManagerTest.php
```diff
@@ -154,14 +165,15 @@ public function dataProvider()
...
+                'prod',
             ],
         ];
     }
 
     /**
      * @dataProvider dataProvider
      */
-    public function testCreate($webspaceKey, array $data)
+    public function testCreate($webspaceKey, array $data, $env)
```
alexander-schranz: replace `$env` with `$environment`

src/Sulu/Bundle/WebsiteBundle/Analytics/AnalyticsManagerInterface.php
```diff
@@ -19,13 +19,13 @@
...
-     * Returns all analytics for given webspace.
+     * Returns all analytics for given webspace and environment.
      *
-     * @param string $webspaceKey
+     * @param string $environment
      *
      * @return Analytics[]
      */
-    public function findAll($webspaceKey);
+    public function findAll($webspaceKey, $environment);
```
alexander-schranz: would inject the `$environment` into the AnalyticsManager to avoid this BC Break in a Hotfix.

src/Sulu/Bundle/WebsiteBundle/Entity/AnalyticsRepository.php
```diff
@@ -19,22 +19,25 @@
...
+     * Returns list of analytics filterd by webspace key and environment.
      *
-     * @param $webspaceKey
+     * @param string $webspaceKey
+     * @param string $environment
      *
      * @return Analytics[]
      */
-    public function findByWebspaceKey($webspaceKey)
+    public function findByWebspaceKeyAndEnvironment($webspaceKey, $environment)
```
alexander-schranz: undocumented BC Break. Add a line to the `UPGRADE.md` or avoid the bcbreak:

To avoid the bc break we could keep the old name and make the $environment optional 

```php
public function findByWebspaceKey($webspaceKey, $environment = null)
          // ...

          if ($environment) {
              $queryBuilder->andWhere('a.allDomains = TRUE OR domains.environment = :environment');
              $queryBuilder->setParameter('environment', $environment);
          } else {
              @trigger_error('Calling findByWebspace without environment is deprecated and will maybe return a false result');
          }

          // ...
}
```

/cc what you think @danrot

danrot: Why not using setter injection to set the environment? As a developer using this class I ideally do not have to worry about the environment.

alexander-schranz: @danrot sounds like a good idea:

@trickreich you can call the setter in the service definition:

```
        <service id="sulu_website.analytics.repository"
                 class="Sulu\Bundle\WebsiteBundle\Entity\AnalyticsRepository">
            <factory service="doctrine.orm.entity_manager" method="getRepository"/>

            <argument type="string">SuluWebsiteBundle:Analytics</argument>
            
            <call method="setEnvironment">
                <argument>%kernel.environment%</argument>
            </call>
        </service>
```

alexander-schranz:
waiting for the tests. styleci will be fixed in #3601

PR #3598, opened 2017-11-02
Fix misleading security warning message

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Change the security warning message to reload the browser instead of system.

#### Why?

From a slack discussion I was asked by a user which system he now need to reload after he changed the permissions. So this message seems to be misleading the user.

### DIFF
--- a/src/Sulu/Bundle/SecurityBundle/Resources/translations/sulu/backend.nl.xlf
+++ b/src/Sulu/Bundle/SecurityBundle/Resources/translations/sulu/backend.nl.xlf
@@ -124,7 +124,7 @@
             </trans-unit>
             <trans-unit id="1cd67d24ab0a432414bc0685517c721d" resname="security.warning">
                 <source>security.warning</source>
-                <target>Om de gewijzigde veiligheidsinstellingen te zien moet het browser opnieuw geladen worden.</target>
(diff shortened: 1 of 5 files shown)


src/Sulu/Bundle/SecurityBundle/Resources/translations/sulu/backend.nl.xlf
```diff
@@ -124,7 +124,7 @@
             </trans-unit>
             <trans-unit id="1cd67d24ab0a432414bc0685517c721d" resname="security.warning">
                 <source>security.warning</source>
-                <target>Om de gewijzigde veiligheidsinstellingen te zien moet het systeem opnieuw geladen worden.</target>
+                <target>Om de gewijzigde veiligheidsinstellingen te zien moet het browser opnieuw geladen worden.</target>
```
silvano-a: i would change `het browser` to `de browser`

alexander-schranz: Thank you! Did change it.

danrot:
Since I have seen in the slack channel that the french translation seems to be OK as well I merge this PR.

PR #3600, opened 2017-11-03
Restore routes instead of navigate when clicking back button

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR adds a history function to the router. It therefore keeps an object, which keeps track of the previous routes to which have been navigated. Then there is the `restore` function on the router, which allows to restore these old attributes and queries, which are merged with the additional passed parameters.

#### Why?

Because this way we can implement the back button correctly. All the parameters like for pagination will be taken from the old storage.

#### Example Usage

~~~javascript
router.restore('page', {uuid: '123'}, {locale: 'de'});
~~~

#### To Do

- [x] Document the restore function
- [ ] Fix bug in media overview, where not identical types (7 !== '7') cause too many histories to be written
- [x] Defaults should be written on `navigate`, because otherwise `navigate` is executed twice (once without and once with the defaults)
- [x] Introduce `navigateWithoutCreatingHistory`

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/services/Router/Router.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/services/Router/Router.js
@@ -87,8 +87,6 @@ export default class Router {
     }
 
     @action navigate(name: string, attributes: Object = {}, createHistory: boolean = true) {
-        const route = routeRegistry.get(name);
-
         if (!this.isRouteChanging(name, attributes)) {
             return;
         }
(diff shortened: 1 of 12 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/services/Router/tests/Router.test.js
```diff
@@ -158,30 +158,6 @@ test('Update observable attribute on route change', () => {
     expect(history.location.pathname).toBe('/list/de');
 });
 
-test('Navigate to route without default attribute when default observable is set', () => {
```
danrot: Due to refactoring the place where the default attributes are set this test wasn't working anymore... And when thinking about it, this setup doesn't really make a lot of sense anyway. A default value for a binding like this only makes sense if the parameter is optional IMO.

src/Sulu/Bundle/AdminBundle/Resources/js/services/Router/Router.js
```diff
@@ -150,6 +166,28 @@ export default class Router {
...
+        this.attributesHistory[this.route.name].push(toJS(this.attributes));
+    }
+
+    isRouteChanging(name: string, attributes: Object) {
+        const route = routeRegistry.get(name);
+
+        return !(
+            this.route
+            && this.route.name === route.name
+            && equal(this.attributes, attributes)
```
danrot: The `MediaOverview` still makes problems... The thing is that the ID changes its datatype from number to string when navigating with a number, which causes two history entries to be written. We've already fixed that for observable by using `==` instead of `===`. But here we are currently using the `fast-deep-equal` package, which asserts by type. So we have two options here:

* Remove the `fast-deep-equal` package and write the function to compare the attributes on our own.
* Always try to parse a number with each attribute.

danrot: We have decided to go for the second option, and always try to parse the attributes as numbers.

PR #3607, opened 2017-11-08
Implement default values for attributes

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | yes
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR will implement the possibility to add default values for attributes. This can be done on the serverside with the new `addAttributeDefault` function of the route.

It also combines the `attributes` and `query` from the router to just `query`.

#### Why?

Because we want to be able to set a default locale for every section, which means that navigating always works for these routes.

And we combined `attributes` and `query`, because this way the caller doesn't need to know where to place the arguments.

#### BC Breaks/Deprecations

The `navigate` function has no `query` parameter anymore and the 

#### To Do

- [ ] Create documentation
- [x] Convert this change: https://github.com/sulu/sulu/pull/3565/files#diff-27a1dff25d6e65d97354f899a6bd59beR109
- [x] Implement default values

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/services/Router/Router.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/services/Router/Router.js
@@ -100,7 +100,12 @@ export default class Router {
         this.attributes = attributes;
 
         for (const [key, observableValue] of this.bindings.entries()) {
-            observableValue.set(this.attributes[key] || this.bindingDefaults.get(key));
+            const value = this.attributes[key] || this.bindingDefaults.get(key);
+
+            // Type unsafe comparison to not trigger a new navigation when only data type changes
+            if (value != observableValue.get()) {
+                observableValue.set(value);
+            }
         }
     }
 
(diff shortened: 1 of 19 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/services/Router/Router.js
```diff
@@ -113,21 +114,34 @@ export default class Router {
...
+        for (const [key, observableValue] of this.bindings.entries()) {
             const value = observableValue.get();
-            if (value == this.queryBindDefaults.get(key)) {
-                searchParameters.delete(key);
-                continue;
+            attributes[key] = value;
+        }
+
+        const attributeDefaults = this.route.attributeDefaults;
+        Object.keys(attributeDefaults).forEach((key) => {
```
chirimoya: add comment

danrot: Done

PR #3609, opened 2017-11-08
added grid component

3rfan:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This PR adds a grid component to the UI library.

#### Why?

To display the form fields properly we need a grid system.

#### Example Usage

~~~js
<Grid>
    <Grid.Section size={12}>
        <Grid.Item size={4}>
            <div style={boxStyles}>
                <Input />
            </div>
        </Grid.Item>
        <Grid.Item size={4}>
            <div style={boxStyles}>
                <Input />
            </div>
        </Grid.Item>
    </Grid.Section>
    <Grid.Section size={12}>
        <Grid.Item size={4}>
            <div style={boxStyles}>
                <Input />
            </div>
        </Grid.Item>
        <Grid.Item size={4}>
            <div style={boxStyles}>
                <Input />
            </div>
        </Grid.Item>
        <Grid.Item size={4}>
            <div style={boxStyles}>
                <Input />
            </div>
        </Grid.Item>
    </Grid.Section>
</Grid>
~~~

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Grid/README.md
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Grid/README.md
@@ -1,7 +1,9 @@
 The grid system consists out of 3 components: `Grid`, `Section` and `Item`. The `Grid` component is the overall
 container and should always be used if you want to create a grid. The `Item` subcomponent can mostly be treated like a
 column in an usual grid system. If the sum of the item sizes exceeds 12 the next item will start in a new line.
-A `Section` serves as a container for `Item` components. 
+A `Section` serves as a container for `Item` components. It has the same `props` as the `Item` but is different in terms
+of style. The `Section` has no padding on the right side like the `Item` does but instead has a margin on the bottom
+side.
 
 ```javascript
 const boxStyles = {
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Grid/README.md
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Grid/README.md
@@ -1,0 +1,104 @@
+The grid system consists out of 3 components: `Grid`, `Section` and `Item`. The `Grid` component is the overall
+container and should always be used if you want to create a grid. The `Item` subcomponent can mostly be treated like a
+column in an usual grid system. If the sum of the item sizes exceeds 12 the next item will start in a new line.
+A `Section` serves as a container for `Item` components. 
+
+```javascript
+const boxStyles = {
+    width: `100%`,
+    height: 30,
+    margin: 2,
+};
+
+<Grid>
+    <Grid.Section size={4}>
+        <Grid.Item size={12}>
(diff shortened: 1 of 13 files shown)


3rfan:
@danrot

src/Sulu/Bundle/AdminBundle/Resources/js/components/Grid/README.md
```diff
@@ -0,0 +1,104 @@
+The grid system consists out of 3 components: `Grid`, `Section` and `Item`. The `Grid` component is the overall
+container and should always be used if you want to create a grid. The `Item` subcomponent can mostly be treated like a
+column in an usual grid system. If the sum of the item sizes exceeds 12 the next item will start in a new line.
+A `Section` serves as a container for `Item` components.
```
danrot: Maybe also mention what the difference to an `Item` is (I think they have a different margin?)

src/Sulu/Bundle/AdminBundle/Resources/js/components/Grid/types.js
```diff
@@ -0,0 +1,11 @@
+// @flow
+import type {Element} from 'react';
+
+export type Size = 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12;
+
+export type BaseItemProps = {
+    size: Size,
+    spaceBefore: Size,
+    spaceAfter: Size,
+    children: Element<*>,
```
danrot: Isn't this different for both elements? I'd say the section is only allowed to have other `Item` and `Section` elements, whereby the statement here is only true for the item component.

src/Sulu/Bundle/AdminBundle/Resources/js/components/Grid/tests/Grid.test.js
```diff
@@ -0,0 +1,55 @@
+/* eslint-disable flowtype/require-valid-file-annotation */
```
danrot: Oh, and by the way, we've decided to use flow also in tests, to avoid mocking classes in a wrong way. You can have a look at an example at https://github.com/sulu/sulu/pull/3607/files#diff-37fa534bf41281dea98a5fb1c89c0e49R1. Please update this test after the other PR is merged (there is some setup for this).

danrot: I've just tried that on my own, and the flow checks are already failing here, because the `Grid.Item` has to have children. That already raises the question if a `Grid.Item` is allowed to be empty. If no we should also adjust the tests here, if yes we have to make the children attribute optional in the types.

I think I would say that we accept anything in a `Grid.Item` it should also be allowed to pass nothing, so I would go for adjusting the type.

And in order for that to work correctly you have to do a rebase.

PR #3615, opened 2017-11-13
ResourceLocator component

trickreich:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes --
| Related issues/PRs | --
| License | MIT
| Documentation PR | --

#### What's in this PR?

Sulu 2.0 ResourceLocator component.

#### Example Usage

~~~javascript
initialState = {value: '/parent/child'};
onChange = (newValue) => {
	setState({value: newValue});
};

<div>
    <div style={{paddingBottom: '50px'}}>Current value: {state.value}</div>
    <ResourceLocator onChange={onChange} value={state.value} mode="leaf"/>
</div>
~~~

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Input/README.md
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Input/README.md
@@ -1,13 +1,21 @@
 The input component can be used to get input from the user in the same way as with the native browser input.
 
-```
-const onChange = (value) => {console.log(value)};
-<Input onChange={onChange} />
+```javascript
+initialState = {value: ''};
+onChange = (newValue) => {
+	setState({value: newValue});
+};
+
+<Input value={state.value} onChange={onChange} />
 ```
 
 Beneath attributes known from the native input, it provides properties to style the the input.
 
-```
-const onChange = (value) => {/* do something */};
-<Input icon="key" type="password" placeholder="Password" onChange={onChange} />
+```javascript
+initialState = {value: ''};
+onChange = (newValue) => {
+	setState({value: newValue});
+};
+
+<Input icon="key" type="password" placeholder="Password" value={state.value} onChange={onChange} />
 ```
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Input/README.md
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Input/README.md
@@ -2,7 +2,7 @@ The input component can be used to get input from the user in the same way as wi
 
 ```javascript
 initialState = {value: ''};
-onChange = (newValue) => {
+const onChange = (newValue) => {
 	setState({value: newValue});
 };
 
@@ -13,7 +13,7 @@ Beneath attributes known from the native input, it provides properties to style 
 
 ```javascript
 initialState = {value: ''};
-onChange = (newValue) => {
+const onChange = (newValue) => {
 	setState({value: newValue});
 };
 
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/ResourceLocator/README.md
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/ResourceLocator/README.md
@@ -1,19 +1,20 @@
 The ResourceLocator component can be used to get a URL from user input in two modes.
 
-* `full`: User can change everything after the initial slash.
-* `leaf`: User just can change the part after the last slash.
+In the `full` mode the user can change the entire URL except for the leading slash:
 
 ```javascript
 initialState = {value: '/parent'};
 const onChange = (newValue) => {
 	setState({value: newValue});
 };
-``
+
 <div>
     <div style={{paddingBottom: '50px'}}>Current value: {state.value}</div>
     <ResourceLocator onChange={onChange} value={state.value} mode="full"/>
 </div>
 ```
+
+In the `leaf` mode the user is only capable of editing the part after the last slash:
 
 ```javascript
 initialState = {value: '/parent/child'};
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/ResourceLocator/README.md
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/ResourceLocator/README.md
@@ -1,0 +1,25 @@
+The input component can be used to get input from the user in the same way as with the native browser input.
+
+```javascript
+initialState = {value: '/parent'};
+onChange = (newValue) => {
+	setState({value: newValue});
+};
+
+<div>
+    <div style={{paddingBottom: '50px'}}>Current value: {state.value}</div>
+    <ResourceLocator onChange={onChange} value={state.value} mode="full"/>
+</div>
+```
+
+```javascript
+initialState = {value: '/parent/child'};
+onChange = (newValue) => {
+	setState({value: newValue});
+};
+
+<div>
+    <div style={{paddingBottom: '50px'}}>Current value: {state.value}</div>
+    <ResourceLocator onChange={onChange} value={state.value} mode="leaf"/>
+</div>
+```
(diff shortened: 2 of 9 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/components/ResourceLocator/README.md
```diff
@@ -0,0 +1,25 @@
+The input component can be used to get input from the user in the same way as with the native browser input.
```
danrot: The input component?

src/Sulu/Bundle/AdminBundle/Resources/js/components/Input/README.md
```diff
@@ -1,13 +1,21 @@
 The input component can be used to get input from the user in the same way as with the native browser input.
 
-```
-const onChange = (value) => {console.log(value)};
-<Input onChange={onChange} />
+```javascript
+initialState = {value: ''};
+onChange = (newValue) => {
```
danrot: Please use const here, otherwise it is a global variable

danrot: Please use const here, otherwise it is a global variable

src/Sulu/Bundle/AdminBundle/Resources/js/components/Input/README.md
```diff
@@ -1,13 +1,21 @@
...
 ```
 
 Beneath attributes known from the native input, it provides properties to style the the input.
 
-```
-const onChange = (value) => {/* do something */};
-<Input icon="key" type="password" placeholder="Password" onChange={onChange} />
+```javascript
+initialState = {value: ''};
+onChange = (newValue) => {
```
danrot: Please use const here, otherwise it is a global variable

src/Sulu/Bundle/AdminBundle/Resources/js/components/ResourceLocator/README.md
danrot: The input component?

src/Sulu/Bundle/AdminBundle/Resources/js/components/ResourceLocator/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+```javascript
+initialState = {value: '/parent'};
+onChange = (newValue) => {
```
danrot: Please use const here, otherwise it is a global variable

src/Sulu/Bundle/AdminBundle/Resources/js/components/ResourceLocator/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+<div>
+    <div style={{paddingBottom: '50px'}}>Current value: {state.value}</div>
+    <ResourceLocator onChange={onChange} value={state.value} mode="full"/>
+</div>
+```
+
+```javascript
+initialState = {value: '/parent/child'};
+onChange = (newValue) => {
```
danrot: Please use const here, otherwise it is a global variable

src/Sulu/Bundle/AdminBundle/Resources/js/components/ResourceLocator/README.md
danrot: A little bit more explanation here would be good 😃

src/Sulu/Bundle/AdminBundle/Resources/js/components/ResourceLocator/ResourceLocator.js
```diff
@@ -0,0 +1,55 @@
...
+        const {value, mode} = this.props;
+        let parts;
+
+        switch (mode) {
+            case 'full':
+                this.fixed = '/';
+                this.changeable = value.substring(1);
+                break;
+            case 'leaf':
+                parts = value.split('/');
```
danrot: Instead of outside the `switch` you could use `const` and instantiate the `parts` variable here

src/Sulu/Bundle/AdminBundle/Resources/js/components/ResourceLocator/resourceLocator.scss
```diff
@@ -0,0 +1,15 @@
+@import '../../containers/Application/colors.scss';
+
+$textColor: $black;
+
+.resource-locator-container {
```
danrot: I would omit the `container` here

src/Sulu/Bundle/AdminBundle/Resources/js/components/ResourceLocator/ResourceLocator.js
```diff
@@ -0,0 +1,55 @@
...
+        const {onChange} = this.props;
+
+        onChange(this.fixed + value);
+    };
+
+    render() {
+        return (
+            <div className={resourceLocatorStyles.resourceLocatorContainer}>
+                <span className={resourceLocatorStyles.fixed}>{this.fixed}</span>
+                <Input onChange={this.handleChange} type="string" value={this.changeable} />
```
danrot: `string` is actually not a valid type here... Because `Input` just passes this type to the `input` type. And it seems as if the `Input` prop types are not specific enough here. In this special case I would omit the `type` (because `text` is the default). Maybe we should also adapt the types of `Input`?

src/Sulu/Bundle/AdminBundle/Resources/js/components/ResourceLocator/ResourceLocator.js
```diff
@@ -0,0 +1,55 @@
...
+            case 'leaf':
+                parts = value.split('/');
+                this.changeable = parts.pop();
+                this.fixed = parts.join('/') + '/';
+                break;
+        }
+    };
+
+    componentWillReceiveProps = (nextProps: Props) => {
+        this.changeable = nextProps.value.substring(this.fixed.length);
```
danrot: I think this is not tested, right?

src/Sulu/Bundle/AdminBundle/Resources/js/components/ResourceLocator/ResourceLocator.js
danrot: Ok, so this seems to fail eslint because of this rule: https://github.com/lukeapage/eslint-plugin-switch-case/blob/master/docs/rules/no-case-curly.md

What's our stand on this? /cc @chirimoya

danrot: Actually it seems to be this rule: https://eslint.org/docs/rules/no-case-declarations We've decided to deactivate this rule and declare this variable in the `case` statement.

src/Sulu/Bundle/AdminBundle/Resources/js/components/ResourceLocator/ResourceLocator.js
trickreich: It is with the full test

src/Sulu/Bundle/AdminBundle/Resources/js/components/ResourceLocator/README.md
```diff
@@ -0,0 +1,28 @@
+The ResourceLocator component can be used to get a URL from user input in two modes.
+
+* `full`: User can change everything after the initial slash.
+* `leaf`: User just can change the part after the last slash.
```
danrot: I wouldn't use a bullet list here... It might not be completely clear which rendered example below belongs to what. I would write it somehow like this instead:

```
In the `full` mode the user can change the entire URL except for the leading slash:

// First example

In the `leaf` mode the user is only capable of editing the part after the last slash:

// second example
```

src/Sulu/Bundle/AdminBundle/Resources/js/components/ResourceLocator/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+```javascript
+initialState = {value: '/parent'};
+const onChange = (newValue) => {
+	setState({value: newValue});
+};
+``
```
danrot: This is probably an error, right?

danrot:
image
There is also too much padding between the parent and the input field, should only be 10px according to the design

PR #3626, opened 2017-11-16
Editable time field in author-selection overlay (#3602)

### LINKED ISSUE #3602: Time of authored date should be editable
| Q | A
| --- | ---
| Bug? | no
| New Feature? | no
| Sulu Version | 1.6.6
| Browser Version | All

#### Actual Behavior

Currently only the date part of the authored field is editable not the time itself.

#### Expected Behavior

The whole datetime with time should be editable.

#### Steps to Reproduce

Edit the authored date in the settings tab of a content page. Only date is possible to be changed not the time.

#### Possible Solutions

Add a time field to the overlay where you can change the authored date.

Edvinas9:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #3602 
| License | MIT

#### What's in this PR?

author-selection overlay now now has a time field in addition to date field


Edvinas9:
Ideally, `authored` field type in`BasePageDocumentType` should be `Symfony\Component\Form\Extension\Core\Type\DateTimeType`, but for some reason it doesn't accept neither date string nor \DateTime object. For this reason I've made it's type `TextType` so it saves date string in the format that CMS sends in PUT API request.

All the dates in the database are in such format: `2017-04-04T14:54:00+0000`
author-selection generates date string in the following format (which is a little different but formatted according to the same ISO standard): `2017-04-15T03:02:00.000Z`

**Need to come up with a proper solution to format date that comes from author-selection overlay before merging.**

Edvinas9:
Date string format fixed by replacing `Date.toISOString();` to `Globalize.format(Date, "yyyy'-'MM'-'dd'T'HH':'mm':'sszz'00'");`

src/Sulu/Bundle/ContentBundle/Form/Type/BasePageDocumentType.php
```diff
@@ -62,8 +61,7 @@ public function buildForm(FormBuilderInterface $builder, array $options)
         $builder->add('shadowLocale', TextType::class); // TODO: Should be choice of available shadow locales
         $builder->add(
             'authored',
-            DateType::class,
-            ['widget' => 'single_text', 'model_timezone' => 'UTC', 'view_timezone' => 'UTC']
```
wachterjohannes: @Edvinas9 why is this needed? the problem with that is - that the type of the value in the database will change with this line. If this is needed - how could we ensure that the type in the database is DateTime

src/Sulu/Bundle/ContentBundle/Resources/config/serializer/Document.BasePageDocument.xml
```diff
@@ -23,9 +23,7 @@
         <property name="creator" type="integer" groups="defaultPage,preview"/>
         <property name="changer" type="integer" groups="defaultPage,preview"/>
         <property name="created" type="DateTime" groups="defaultPage,preview"/>
-        <property name="authored" groups="defaultPage,preview">
-            <type><![CDATA[DateTime<'Y-m-d'>]]></type>
```
wachterjohannes: when the type in the database is DateTime we have to include this (maybe with an additional time in the result).

src/Sulu/Bundle/ContentBundle/Resources/public/js/components/content/settings/author-selection/main.js
```diff
@@ -12,13 +12,14 @@ define(['jquery', 'text!/admin/api/contacts/fields'], function($, fieldsResponse
     'use strict';
 
     var constants = {
-            instanceName: 'author-selection'
```
wachterjohannes: some indentations are changed the wrong way in this file.

src/Sulu/Bundle/ContentBundle/Resources/public/js/components/content/settings/author-selection/main.js
```diff
@@ -68,51 +83,67 @@ define(['jquery', 'text!/admin/api/contacts/fields'], function($, fieldsResponse
...
+
+            var authored = this.options.data.authored ? this.options.data.authored : null;
+            var authoredTime = authored
+                ? Globalize.format(new Date(authored), Globalize.culture().calendar.patterns.t)
+                : null;
+
             this.html($(this.templates.skeleton({
                 translations: this.translations,
                 nullableAuthor: this.options.nullableAuthor,
-                authored: this.options.data.authored ? this.options.data.authored : null
```
wachterjohannes: the authored field will be empty with this change.

src/Sulu/Bundle/ContentBundle/Resources/config/serializer/Document.BasePageDocument.xml
Edvinas9: I've simply added `type="DateTime"` which seems to serialize DateTime to a suitable format (`2017-05-01T06:14:00+0000`).

Is anything wrong with that?

src/Sulu/Bundle/ContentBundle/Resources/public/js/components/content/settings/author-selection/main.js
Edvinas9: This is not required anymore since the field value will be set by form mapper (`this.sandbox.form.setData()`).
However, form mapper fails to set initial value for authoredTime field so I've left that one intact.

src/Sulu/Bundle/ContentBundle/Form/Type/BasePageDocumentType.php
Edvinas9: I've simply added type="DateTime" which seems to serialise DateTime to a suitable format (`2017-05-01T06:14:00+0000`).

Is anything wrong with that?

Edvinas9: I was struggling to get `DateTimeType` to accept datetime string values (for example: `2017-05-01T06:14:00+0000`) so temporary put TextType in place just because I had to get this working ASAP for a client's project. I've created a PR early just to get some feedback and/or suggestions on this.

However, I gave it another go and seems like `'widget' => 'single_text'` option works just fine. Are you happy with that?

wachterjohannes:
@Edvinas9 there are a lot of failures which happens because of the string in the authored field.

alexander-schranz:
@wachterjohannes @Edvinas9 when the `TextType` is really required we should use a DataTransformer: https://symfony.com/doc/current/form/data_transformers.html to convert the string to DateTime.

Edvinas9:
TextType has been already changed to DateTimeType and none of the tests fail anymore. I can see that AppVeyor build is failing but I presume that's a new integration you guys are still working on and is not configured properly yet (`fatal: Invalid path '/cygdrive/c/projects/sulu/C:\projects\sulu': No such file or directory`)

alexander-schranz:
@Edvinas9 👍  AppVeyor can currently be ignored. seems to be an error in appveyor itself.

wachterjohannes:
@Edvinas9 sorry my github window seems to frozen :)

Edvinas9:
Entry in CHANGELOG.md has been just added

PR #3569, opened 2017-10-16
Updated "friendsofsymfony/rest-bundle" and "jms/serializer-bundle"

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | none
| Related issues/PRs | none
| License | MIT
| Documentation PR | none

#### What's in this PR?

This PR updated `friendsofsymfony/rest-bundle` to version `^2.2`.

#### TODO

- [x] Tests for new compiler-passes

PR #3628, opened 2017-11-18
removed Symfony deprecations

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | https://github.com/sulu/sulu/pull/3569
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR removes the usage of deprecated Symfony functions. However, there are still deprecations left, either from ourselves or the FOSRestBundle, which will hopefully be gone in #3569 

#### Why?

Because they will be removed in Symfony 4, and we will stay compatible.


composer.json
```diff
@@ -32,7 +32,7 @@
         "stof/doctrine-extensions-bundle": "^1.2.2",
         "gedmo/doctrine-extensions": "~2.4",
         "symfony-cmf/routing-bundle": "^1.2 || ^2.0",
-        "symfony/symfony": "^2.8.7 || ^3.0",
+        "symfony/symfony": "^2.8.13 || ^3.0",
```
danrot: This is required because the `isMethodCachable` method was added in 2.8.13

tests/app/config/config.yml
```diff
@@ -5,3 +5,7 @@ sulu_core:
                  internal:
                      path: "%kernel.root_dir%/../../src/Sulu/Bundle/CoreBundle/Content/templates"
                      type: "page"
+
+services:
+    sulu_document_manager_test.property_encoder:
```
danrot: @chirimoya That's how we have to redefine private services now, in order to be able to test them using `getContainer()->get()`. The question is `sulu_document_manager_test` prefix ok for this.

danrot:
@chirimoya Otherwise this would be ready to review

danrot:
Yeah, everything that is left now are our own deprecations, and not from Symfony 3.3 (I am not talking about Symfony 3.4 yet)

PR #3631, opened 2017-11-20
Fix jsx no-bind rule from eslint

danrot:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

Fixed a few inline functions passed to JSX components.

#### Why?

The updated JSX rule in eslint seems to be better now, and recognizes more errors.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/ColumnList/ToolbarDropdownList.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/ColumnList/ToolbarDropdownList.js
@@ -1,6 +1,7 @@
 // @flow
 import React from 'react';
 import classNames from 'classnames';
+import ToolbarDropdownListOption from './ToolbarDropdownListOption';
 import type {ToolbarDropdownOptionConfig} from './types';
 import toolbarDropdownStyles from './toolbarDropdown.scss';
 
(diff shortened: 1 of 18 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/containers/Application/Application.js
```diff
@@ -97,6 +97,7 @@ export default class Application extends React.PureComponent<Props> {
                             return (
                                 <li
                                     key={index}
+                                    // eslint-disable-next-line react/jsx-no-bind
```
danrot: I have decided here to go for disabling the rule, since this is part of the navigation, which will be rewritten anyway

src/Sulu/Bundle/AdminBundle/Resources/js/components/ColumnList/ToolbarDropdownList.js
```diff
@@ -20,20 +21,12 @@ export default class ToolbarDropdownList extends React.Component<Props> {
...
-            const handleClick = () => {
-                dropdownOptionConfig.onClick(this.props.columnIndex);
-            };
+            const {onClick, label} = dropdownOptionConfig;
 
             return (
-                <li
-                    className={toolbarDropdownStyles.option}
-                    key={key}
-                    value={this.props.columnIndex}
```
danrot: @trickreich Had to change this file, because the newer version of the no bind rule seems to recognize the inline `handleClick` function here. I realized that there is this `value` on `li`, this shouldn't have any impact, right?

PR #3549, opened 2017-10-04
build locale-less form for contact

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | parts of #3548 
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR implements a form for the contacts, because contacts are not translated. So this complements #3489.

#### Why?

Because we need locale-less forms anyway, and there are some obstacles for implementing them. So we are doing this quite early.

#### To Do

- [x] Create contact form
- [x] Navigation should also include locale if required for that part of the application

PR #3635, opened 2017-11-20
build form using metadata

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR makes it possible to load the form using the metadata from the `ResourceMetadataStore` instead of being hardcoded in the `Form` view.

#### Why?

Because we want to load different forms.

#### To Do

- [x] Update documentation
- [x] Fix skipped test

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/MetadataStore.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/MetadataStore.js
@@ -1,0 +1,16 @@
+// @flow
+import resourceMetadataStore from '../../../stores/ResourceMetadataStore';
+import type {Schema} from '../../../stores/ResourceStore/types';
+
+class MetadataStore {
+    getFields(resourceKey: string): Schema {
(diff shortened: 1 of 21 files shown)


danrot:
The contact form can already be called with this PR, and it kinda works. The problem is that the `ResourceStore` cannot handle entities without locale yet. That should be fixed in #3549, which is a work in progress...

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/MetadataStore.js
```diff
@@ -0,0 +1,16 @@
+// @flow
+import resourceMetadataStore from '../../../stores/ResourceMetadataStore';
+import type {Schema} from '../../../stores/ResourceStore/types';
+
+class MetadataStore {
+    getFields(resourceKey: string): Schema {
```
chirimoya: `getSchema` ???

PR #3637, opened 2017-11-21
Add excluded ids parameter to media api

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Add excluded ids parameter to media api.

#### Why?

To show only unselected items in the media selection.

#### Example Usage

`/api/media?locale=en&excluded-ids=1,2,3`

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/Controller/MediaController.php
+++ b/src/Sulu/Bundle/MediaBundle/Controller/MediaController.php
@@ -118,8 +118,9 @@ class MediaController extends AbstractMediaController implements
         $locale = $this->getRequestParameter($request, 'locale', true);
         $fieldDescriptors = $this->getFieldDescriptors($locale, false);
         $ids = array_filter(explode(',', $request->get('ids')));
+        $excluded = array_filter(explode(',', $request->get('excluded')));
         $types = array_filter(explode(',', $request->get('types')));
-        $listBuilder = $this->getListBuilder($request, $fieldDescriptors, $ids, $types);
+        $listBuilder = $this->getListBuilder($request, $fieldDescriptors, $ids, $excluded, $types);
         $listResponse = $listBuilder->execute();
         $count = $listBuilder->count();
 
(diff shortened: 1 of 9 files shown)


src/Sulu/Component/Rest/ListBuilder/Expression/Doctrine/DoctrineNotInExpression.php
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+namespace Sulu\Component\Rest\ListBuilder\Expression\Doctrine;
+
+use Doctrine\ORM\QueryBuilder;
+
+/**
+ * Represents a NOT IN expression for doctrine - needs a field and an array of values.
+ */
+class DoctrineNotInExpression extends DoctrineInExpression
```
danrot: Wouldn't it be nicer to have a simple `DoctrineNotExpression`, which can also be used with other statements than `IN`?

alexander-schranz: @done in the ListBuilder we still need to have a `notIn` function

src/Sulu/Bundle/MediaBundle/Controller/MediaController.php
```diff
@@ -118,8 +118,9 @@ public function cgetAction(Request $request)
         $locale = $this->getRequestParameter($request, 'locale', true);
         $fieldDescriptors = $this->getFieldDescriptors($locale, false);
         $ids = array_filter(explode(',', $request->get('ids')));
+        $excludedIds = array_filter(explode(',', $request->get('excluded-ids')));
```
danrot: I have just realized that we are inconsistent in the naming here... Sometimes we use a dash and sometimes we use camelCase for these parameters... And in the action for loading the content of the smart content it is just called `excluded`... Maybe go with that, in order to avoid that issue 🙈

danrot:
I'll merge as soon as the build succeeded

PR #3639, opened 2017-11-21
Fix structure internal flag and add webspace validate command

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT
| Documentation | https://github.com/sulu/sulu-docs/pull/364

#### What's in this PR?

Fix structure internal flag as it is readed as boolean not string.

Also add better error message when structure metadata could not be found.

Also a command to validate the webspaces was added. It checks if the xml exists, view exists and controller exists.

#### Why?

Currently the excerpt.xml and other internal structure the internal flag is not correctly read and is false instead of true.

Currently error message was `Calling getProperties() on null`.

There was currently no way to validate the webspace configuration is ok.

#### Example Usage

Exception: Change default template of webspace to a none exist and run sulu:build dev.
Command: `bin/adminconsole sulu:content:validate:webspaces`:

Output:

bildschirmfoto 2018-02-23 um 20 01 06

#### BC Breaks/Deprecations

#### To Do

### DIFF
--- a/src/Sulu/Component/Content/Metadata/Loader/XmlLoader.php
+++ b/src/Sulu/Component/Content/Metadata/Loader/XmlLoader.php
@@ -36,7 +36,7 @@ class XmlLoader extends XmlLegacyLoader
         $structure->name = $data['key'];
         $structure->cacheLifetime = $data['cacheLifetime'];
         $structure->controller = $data['controller'];
-        $structure->internal = 'true' === $data['internal'];
+        $structure->internal = $data['internal'];
         $structure->view = $data['view'];
         $structure->tags = $data['tags'];
         $structure->parameters = $data['params'];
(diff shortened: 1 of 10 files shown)


src/Sulu/Component/Content/Metadata/Loader/XmlLoader.php
```diff
@@ -36,7 +36,7 @@ public function load($resource, $type = 'page')
         $structure->name = $data['key'];
         $structure->cacheLifetime = $data['cacheLifetime'];
         $structure->controller = $data['controller'];
-        $structure->internal = 'true' === $data['internal'];
+        $structure->internal = $data['internal'];
```
alexander-schranz: this is a boolean and seems always set to false.

alexander-schranz:
Backported the PR to 1.5 as I found a bug when building this feature (Internal Flag for Structure Metadata is never set to true)

PR #3640, opened 2017-11-22
Merge options of routes with parent options

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | #3549 
| License | MIT

#### What's in this PR?

This PR merges the options of the routes from the `Admin` with the options from the parent route. 

#### Why?

This will enable e.g. #3549 to remove the `locales` from the `sulu_snippet.form.detail` route, because it will be taken from `sulu_snippet.form`.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Admin/Routing/Route.php
+++ b/src/Sulu/Bundle/AdminBundle/Admin/Routing/Route.php
@@ -53,9 +53,20 @@ class Route
         $this->view = $view;
     }
 
+    public function getName(): string
+    {
+        return $this->name;
+    }
+
     public function addOption(string $key, $value): self
     {
         $this->options[$key] = $value;
+
+        return $this;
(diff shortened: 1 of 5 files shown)


src/Sulu/Bundle/AdminBundle/Admin/Routing/Route.php
```diff
@@ -53,13 +53,24 @@ public function __construct(string $name, string $path, string $view)
...
+    }
+
     public function addOption(string $key, $value): self
     {
         $this->options[$key] = $value;
 
         return $this;
     }
 
+    public function mergeRoute(self $route): self {
```
chirimoya: ` public function mergeOptions(array $options): self` ???

PR #3627, opened 2017-11-18
Added symfony components to dependencies and removed symfony/symfony

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | none
| Related issues/PRs | none
| License | MIT
| Documentation PR | none

#### What's in this PR?

This PR removes the dependency to "symfony/symfony" which will be the first step to "symfony/flex".

Basic PR foir recipe https://github.com/symfony/recipes-contrib/pull/142

#### To Do

- [x] Find direct dependencies
- [x] Bring the tests to green
- [x] Find indirect dependencies

#SymfonyConHackday2017

PR #3645, opened 2017-11-23
Use prestissimo to run composer update and parallel

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Use https://github.com/hirak/prestissimo to speedup composer.

#### Why?

To speed up ci script.

### DIFF
--- a/bin/travis/before_script_php.sh
+++ b/bin/travis/before_script_php.sh
@@ -14,5 +14,9 @@ fi
 echo "memory_limit=2048M" >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
 phpenv config-rm xdebug.ini
 composer self-update
+echo "Install prestissimo"
+composer global require "hirak/prestissimo:^0.3" --prefer-dist --no-progress --no-suggest --optimize-autoloader --classmap-authoritative  --no-interaction
+echo "Install dependencies ..."
 composer update $COMPOSER_FLAGS
 if [[ $SYMFONY__PHPCR__TRANSPORT = jackrabbit ]]; then composer require jackalope/jackalope-jackrabbit:~1.2 ; fi
+


alexander-schranz:
Did also test prestissimo on our gitlab ci runner. It did improve the CI script in a sulu project by 1min (before 1min 30sek for code check, with prestissimo 33sek). /cc @sulu/core-team

wachterjohannes:
as it is no "hard" dependency and it could be removed easily! yes we can merge this

bin/travis/before_script_php.sh
```diff
@@ -14,5 +14,9 @@ fi
 echo "memory_limit=2048M" >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
 phpenv config-rm xdebug.ini
 composer self-update
+echo "Install prestissimo"
+composer global require "hirak/prestissimo:^0.3" --prefer-dist --no-progress --no-suggest --optimize-autoloader --classmap-authoritative  --no-interaction
+echo "Install dependencies ..."
 composer update $COMPOSER_FLAGS
 if [[ $SYMFONY__PHPCR__TRANSPORT = jackrabbit ]]; then composer require jackalope/jackalope-jackrabbit:~1.2 ; fi
+
```
danrot: This new line is too much

bin/travis/before_script_php.sh
```diff
@@ -14,5 +14,9 @@ fi
 echo "memory_limit=2048M" >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
 phpenv config-rm xdebug.ini
 composer self-update
+echo "Install prestissimo"
+composer global require "hirak/prestissimo:^0.3" --prefer-dist --no-progress --no-suggest --optimize-autoloader --classmap-authoritative  --no-interaction
+echo "Install dependencies ..."
```
danrot: I wouldn't add `echo`s here... The log is already cluttered enough, and the command tells pretty good what's happening

danrot:
But what's happening in the [current build log](https://travis-ci.org/sulu/sulu/jobs/306312518#L503)?

wachterjohannes:
@danrot there is a mishandling of --prefer-lowest (also in develop) fixed inhttps://github.com/sulu/sulu/pull/3627

alexander-schranz:
@danrot PR updated

danrot:
It looks to me as if it does not improve the performance on travis as much as you described it here. From the numbers I can see that it even performs worse.

If you check [this build without prestissimo](https://travis-ci.org/sulu/sulu/jobs/315826666#L530) you see that the `before_script_php.sh` took 30.17s. I've compared that to the test run in this build, and it was over 80s. I thought that's quite big difference, an reran the test with prestissimo, but it still takes 31.07s, and therefore is slightly slower than the version without it.

I've also checked a few more PRs, and while it is sometimes a little bit slower without prestissimo, it's also faster sometimes. I have also seen one build running this part in [27s](https://travis-ci.org/sulu/sulu/jobs/315425195#L484).

Therefore it looks to me like it is not really improving performance, so I wouldn't include another dependency.

wachterjohannes:
i think the performance depends more on "travis" than our script. but I would go with this and try to validate it in a week for example! if it improves it only a bit I would give it a try.

danrot:
I have just double-checked: It's still slower than travis, and I've also talked to @alexander-schranz. He also thinks that it is not necessary to merge this, if it doesn't bring any improvements. Therefore I am closing this PR.

We also found that it might be faster on gitlab, because they've included prestissimo in the image they are running, which is not really possible in travis. So that might cause the difference on these two systems.

PR #3647, opened 2017-11-23
Fixed sulu_content_load throwing exception when reference was deleted

### LINKED ISSUE #3050: fixed uncatchable exception when use sulu_content_load twig extension
| Q | A
| --- | ---
| Bug? | yes
| New Feature? | no

What's in this PR?

Please check https://github.com/sulu/sulu/pull/2969, its kind a like the same issue but now its for sulu_content_load

When a page is deleted but the reference is still in a snippet, we are getting a exception. It would be nice the get null instead so we can check in Twig

Why?

You are not able to catch exceptions in twig.

Example Usage

{{  sulu_content_load(block.button_link).content.url }}

rogamoore:
Fixed sulu_content_load throwing exception when reference was deleted

| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #3050 
| License | MIT

#### What's in this PR?

This fixes #3050. When sulu_content_load is used with a non-existing guid, it just returns null instead of throwing an exception (which is not catchable and breaks the page)

#### Why?

You are not able to catch exceptions in twig. This breaks the page when a referenced node is deleted.

#### Example Usage

~~~twig
{% set content = sulu_content_load(null) %}
{# content is now null #}

{% set content = sulu_content_load('not-existing-guid') %}
{# content is now null #}

~~~

#### ---
Would be nice if this could be also released in the 1.4.x branch as we have existing installations and no resources to upgrade them to 1.6.x at the moment.

### TODO

- [ ] Tests need a helping hand

### DIFF
--- a/src/Sulu/Bundle/WebsiteBundle/Twig/Content/ContentTwigExtension.php
+++ b/src/Sulu/Bundle/WebsiteBundle/Twig/Content/ContentTwigExtension.php
@@ -14,6 +14,7 @@ namespace Sulu\Bundle\WebsiteBundle\Twig\Content;
 use Sulu\Bundle\WebsiteBundle\Resolver\StructureResolverInterface;
 use Sulu\Bundle\WebsiteBundle\Twig\Exception\ParentNotFoundException;
 use Sulu\Component\Content\Mapper\ContentMapperInterface;
+use Sulu\Component\DocumentManager\Exception\DocumentNotFoundException;
 use Sulu\Component\PHPCR\SessionManager\SessionManagerInterface;
 use Sulu\Component\Webspace\Analyzer\RequestAnalyzerInterface;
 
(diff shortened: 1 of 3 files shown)


src/Sulu/Bundle/WebsiteBundle/Twig/Content/ContentTwigExtension.php
```diff
@@ -73,13 +74,17 @@ public function getFunctions()
...
+            $contentStructure = $this->contentMapper->load(
+                $uuid,
+                $this->requestAnalyzer->getWebspace()->getKey(),
+                $this->requestAnalyzer->getCurrentLocalization()->getLocale()
+            );
 
-        return $this->structureResolver->resolve($contentStructure);
+            return $this->structureResolver->resolve($contentStructure);
+        } catch (DocumentNotFoundException $e) {
+            return;
```
wachterjohannes: we should avoid suppressing the exception (at least completely). @danrot @chirimoya what do you think? should we only log the exception or should we rethrow it in dev environment? The same question we should ask for all the twig extension

src/Sulu/Bundle/WebsiteBundle/Twig/Content/ContentTwigExtension.php
```diff
@@ -73,13 +74,17 @@ public function getFunctions()
      */
     public function load($uuid)
     {
-        $contentStructure = $this->contentMapper->load(
```
wachterjohannes: use a early return to catch the null uuid.

```php
if (!$uuid) {
    return;
}
```

rogamoore:
I somehow messed up things while trying to rebase it, so I decided to just create a new and clean PR based on master :)
https://github.com/sulu/sulu/pull/3649

PR #3648, opened 2017-11-24
Datagrid Adapter Switch

trickreich:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes --
| Related issues/PRs | --
| License | MIT
| Documentation PR | --

#### What's in this PR?

Implemented button to change the adapter in datagrid.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Datagrid/AdapterSwitchItem.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Datagrid/AdapterSwitchItem.js
@@ -24,7 +24,7 @@ export default class AdapterSwitchItem extends React.PureComponent<Props> {
         } = this.props;
 
         return (
-            <li key={adapter} onClick={this.handleClick}>
+            <li onClick={this.handleClick}>
                 {active ? (
                     <i>{adapter}</i>
                 ) : adapter}
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Datagrid/AdapterSwitchItem.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Datagrid/AdapterSwitchItem.js
@@ -1,0 +1,34 @@
+// @flow
+import React from 'react';
+
+type Props = {
+    adapter: string,
+    active: boolean,
+    onClick: (adapter: string) => void,
+};
+
+export default class AdapterSwitchItem extends React.PureComponent<Props> {
+    handleClick = () => {
+        const {
+            adapter,
+            onClick,
+        } = this.props;
+
+        onClick(adapter);
+    };
+
+    render() {
+        const {
+            active,
+            adapter,
+        } = this.props;
+
+        return (
+            <li key={adapter} onClick={this.handleClick}>
+                {active ? (
+                    <i>{adapter}</i>
+                ) : adapter}
+            </li>
+        );
+    }
+}
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Datagrid/Datagrid.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Datagrid/Datagrid.js
@@ -94,7 +94,9 @@ export default class Datagrid extends React.PureComponent<Props> {
                 />
                 {this.props.store.loading && this.props.store.loadingStrategy !== 'infiniteScroll'
                     ? <Loader />
-                    : <PaginationDecorator
+                    :
+                    // TODO: Try to remove this Decorator
+                    <PaginationDecorator
                         type={Adapter.getLoadingStrategy()}
                         total={pageCount}
                         current={page}
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Datagrid/Datagrid.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Datagrid/Datagrid.js
@@ -33,6 +33,7 @@ export default class Datagrid extends React.PureComponent<Props> {
         if (!equal(this.props.adapters, nextProps.adapters)) {
             this.validateAdapters();
         }
+        nextProps.store.init(this.currentAdapter.getLoadingStrategy());
     }
 
     validateAdapters() {
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Datagrid/Datagrid.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Datagrid/Datagrid.js
@@ -1,6 +1,6 @@
 // @flow
 import {observer} from 'mobx-react';
-import {observable, action} from 'mobx';
+import {observable, action, computed} from 'mobx';
 import React from 'react';
 import equal from 'fast-deep-equal';
 import Loader from '../../components/Loader';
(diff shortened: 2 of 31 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/containers/Datagrid/AdapterSwitchItem.js
```diff
@@ -0,0 +1,34 @@
...
+    };
+
+    render() {
+        const {
+            active,
+            adapter,
+        } = this.props;
+
+        return (
+            <li key={adapter} onClick={this.handleClick}>
```
danrot: `key` doesn't make any sense here. It is only needed in loops, not in the implementation of the element being created in a loop.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Datagrid/AdapterSwitchItem.js
```diff
@@ -0,0 +1,34 @@
...
+                {active ? (
+                    <i>{adapter}</i>
```
danrot: Why an `i` tag? I would make that at least an `a` tag, so that it is something clickable. Actually I think I would make a simple select here. That's not a lot of work, and would already look a lot nicer. I know that it will be rebuilt anyway, but the `develop` branch is already tested by a few developers out there, therefore I would try to not forget about this stuff completely 🙈

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Datagrid/Datagrid.js
```diff
@@ -1,53 +1,64 @@
...
+import AdapterSwitch from './AdapterSwitch';
 
 type Props = {
     onItemClick?: (itemId: string | number) => void,
     store: DatagridStore,
-    views: Array<string>,
+    adapters: Array<string>,
 };
 
+type CurrentAdapter = {
```
danrot: Isn't just the key enough to keep track of the current adapter? With the key you could get the right adapter from the `datagridAdapterRegistry` anyway.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Datagrid/adapters/TableAdapter.js
```diff
@@ -2,14 +2,29 @@
...
-export default class TableAdapter extends React.Component<DatagridAdapterProps> {
+export default class TableAdapter extends AbstractAdapter {
+    static getLoadingStrategy: () => string = () => { return 'pagination'; };
+    static getStorageStrategy: () => string = () => { return 'flat'; };
+
     static defaultProps = {
         data: [],
     };
 
+    renderCells(item: {[*]: *}, schemaKeys: Array<string>) {
```
danrot: I think just writing `item: {}` is a little bit easier 😃

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Datagrid/adapters/TableAdapter.js
```diff
@@ -2,14 +2,29 @@
...
+        return schemaKeys.map((schemaKey) => {
+            if (typeof item[schemaKey] === 'object') {
```
danrot: Maybe add a todo comment here, saying that we have to find a proper solution for this.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Datagrid/Datagrid.js
```diff
@@ -63,23 +74,32 @@ export default class Datagrid extends React.PureComponent<Props> {
...
         return (
             <div>
+                <AdapterSwitch
+                    adapters={adapters}
+                    currentAdapter={this.currentAdapter.key}
+                    onAdapterChange={this.handleAdapterChange}
+                />
                 {this.props.store.loading && !this.props.store.appendRequestData
                     ? <Loader />
                     : <PaginationDecorator
```
danrot: Maybe we should also add a TODO here, saying that we want to get rid of this decorator. Not that we forget it in the end 😃

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Datagrid/stores/DatagridStore.js
```diff
@@ -9,30 +9,51 @@ export default class DatagridStore {
...
+    @observable appendRequestData: boolean = false;
     disposer: () => void;
     resourceKey: string;
     options: Object;
     observableOptions: ObservableOptions;
-    appendRequestData: boolean;
     localeInterceptDisposer: () => void;
-    dataObserveDisposer: () => void;
+
+    static getAppendRequestData: (loadingStrategy: string) => boolean = (loadingStrategy: string) => {
```
danrot: Why don't we rename the `appendRequestData` to `loadingStrategy`? Then we wouldn't need this method at all, and also setting a new `loadingStrategy` becomes trivial.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Datagrid/stores/DatagridStore.js
```diff
@@ -9,30 +9,51 @@ export default class DatagridStore {
...
-        options: Object = {},
-        appendRequestData: boolean = false
+        options: Object = {}
     ) {
         this.resourceKey = resourceKey;
         this.observableOptions = observableOptions;
         this.options = options;
+    }
+
+    @action init = (loadingStrategy: string) => {
```
danrot: Instead of this additional constructor we could still make the `autorun` in the constructor, and build something like a guard in the `sendRequest` method, couldn't we? Would then look something like this:

```javascript
sendRequest = () => {
    if (!this.loadingStrategy) {
        return;
    }
}
```

The only thing we would have to take care of, is that this single `if` includes all strategies that are mandatory in order to do a request.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Datagrid/stores/DatagridStore.js
```diff
@@ -133,7 +157,9 @@ export default class DatagridStore {
     }
 
     destroy() {
-        this.disposer();
+        if (this.disposer) {
```
danrot: This probably also wouldn't be necessary if we let the `autorun` in the `constructor`

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Datagrid/tests/Datagrid.test.js
```diff
@@ -99,13 +113,15 @@ test('Selecting and deselecting items should update store', () => {
 });
 
 test('Selecting and unselecting all items on current page should update store', () => {
+    datagridAdapterRegistry.get.mockReturnValue(TableAdapter);
+    datagridAdapterRegistry.get.mockReturnValue(TableAdapter);
```
danrot: Why is that done twice?

trickreich:
@danrot Should be OK now

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Datagrid/tests/Datagrid.test.js
```diff
@@ -3,11 +3,14 @@ import {mount, shallow} from 'enzyme';
...
 import DatagridStore from '../stores/DatagridStore';
-import TableAdapter from '../adapters/TableAdapter';
 import datagridAdapterRegistry from '../registries/DatagridAdapterRegistry';
+import AbstractAdapter from '../adapters/AbstractAdapter';
+import TableAdapter from '../adapters/TableAdapter';
 
 jest.mock('../stores/DatagridStore', () => jest.fn(function() {
     this.setPage = jest.fn();
+    this.init = jest.fn();
+    this.appendRequestData = false;
```
danrot: This should be removed now, right?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Datagrid/tests/Datagrid.test.js
```diff
@@ -37,38 +40,48 @@ jest.mock('../../../services/Translator', () => ({
...
+    }
+}
+
 beforeEach(() => {
     datagridAdapterRegistry.has.mockReturnValue(true);
-    datagridAdapterRegistry.get.mockReturnValue({
-        Adapter: TableAdapter,
-        paginationType: 'default',
-    });
+    datagridAdapterRegistry.get.mockReturnValue(TestAdapter);
```
danrot: I would return a simple `jest.fn()` instead of a real implementation. That's less code, and has the advantage that calls can be asserted. Unless you tried and it was not working for some reason?

danrot:
@chirimoya LGTM now

PR #3649, opened 2017-11-24
Fixed sulu_content_load throwing exception when reference was deleted

### LINKED ISSUE #3050: fixed uncatchable exception when use sulu_content_load twig extension

rogamoore:
Fixed sulu_content_load throwing exception when reference was deleted

| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #3050 
| License | MIT

#### What's in this PR?

This fixes #3050. When sulu_content_load is used with a non-existing guid, it just returns null instead of throwing an exception (which is not catchable and breaks the page)

#### Why?

You are not able to catch exceptions in twig. This breaks the page when a referenced node is deleted.

#### Example Usage

~~~twig
{% set content = sulu_content_load(null) %}
{# content is now null #}

{% set content = sulu_content_load('not-existing-guid') %}
{# content is now null #}

~~~

#### ---
Would be nice if this could be also released in the 1.4.x branch as we have existing installations and no resources to upgrade them to 1.6.x at the moment.

### TODO

- [x] Tests need a helping hand

### DIFF
--- a/src/Sulu/Bundle/WebsiteBundle/Tests/Unit/Sulu/Bundle/WebsiteBundle/Twig/ContentTwigExtensionTest.php
+++ b/src/Sulu/Bundle/WebsiteBundle/Tests/Unit/Sulu/Bundle/WebsiteBundle/Twig/ContentTwigExtensionTest.php
@@ -169,6 +169,30 @@ class ContentTwigExtensionTest extends \PHPUnit_Framework_TestCase
         $this->assertEquals(['title' => []], $result['view']);
     }
 
+    public function testLoadNull()
+    {
+        $extension = new ContentTwigExtension(
+            $this->contentMapper->reveal(),
+            $this->structureResolver,
+            $this->sessionManager->reveal(),
+            $this->requestAnalyzer->reveal()
+        );
+
+        $this->assertNull($extension->load(null));
+    }
+
+    public function testLoadNotExistingDocument()
+    {
+        $extension = new ContentTwigExtension(
+            $this->contentMapper->reveal(),
+            $this->structureResolver,
+            $this->sessionManager->reveal(),
+            $this->requestAnalyzer->reveal()
+        );
+
+        $this->assertNull($extension->load('999-999-999'));
+    }
+
     public function testLoadParent()
     {
         $this
--- a/src/Sulu/Bundle/WebsiteBundle/Tests/Unit/Sulu/Bundle/WebsiteBundle/Twig/ContentTwigExtensionTest.php
+++ b/src/Sulu/Bundle/WebsiteBundle/Tests/Unit/Sulu/Bundle/WebsiteBundle/Twig/ContentTwigExtensionTest.php
@@ -13,6 +13,8 @@ namespace Sulu\Bundle\WebsiteBundle\Twig;
 
 use PHPCR\NodeInterface;
 use PHPCR\SessionInterface;
+use Prophecy\Argument;
+use Psr\Log\LoggerInterface;
 use Sulu\Bundle\WebsiteBundle\Resolver\StructureResolver;
 use Sulu\Bundle\WebsiteBundle\Resolver\StructureResolverInterface;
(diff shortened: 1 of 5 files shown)


src/Sulu/Bundle/WebsiteBundle/Twig/Content/ContentTwigExtension.php
```diff
@@ -73,13 +74,21 @@ public function getFunctions()
...
+            $contentStructure = $this->contentMapper->load(
+                $uuid,
+                $this->requestAnalyzer->getWebspace()->getKey(),
+                $this->requestAnalyzer->getCurrentLocalization()->getLocale()
+            );
 
-        return $this->structureResolver->resolve($contentStructure);
+            return $this->structureResolver->resolve($contentStructure);
+        } catch (DocumentNotFoundException $e) {
+            return;
```
danrot: We've talked about that today, and we thought it might make sense to throw some kind of `TwigException` here, or however that is called. The ones that twig throws in develop mode, and catches in production (because twig is running in strict mode). In the best case Twig would even write the error to the log. If all these assumptions are true, I would go for this way 😃

wachterjohannes: could you add some logging features here? like:

```php
$this->logger->error((string) $e);
```

src/Sulu/Bundle/WebsiteBundle/Tests/Unit/Sulu/Bundle/WebsiteBundle/Twig/ContentTwigExtensionTest.php
```diff
@@ -169,6 +169,30 @@ public function testLoad()
         $this->assertEquals(['title' => []], $result['view']);
     }
 
+    public function testLoadNull()
+    {
+        $extension = new ContentTwigExtension(
```
wachterjohannes: add following line here:

```php
         $this
             ->contentMapper
             ->load(Argument::cetera())
             ->shouldNotBeCalled();
```

src/Sulu/Bundle/WebsiteBundle/Tests/Unit/Sulu/Bundle/WebsiteBundle/Twig/ContentTwigExtensionTest.php
```diff
@@ -169,6 +169,30 @@ public function testLoad()
...
+            $this->sessionManager->reveal(),
+            $this->requestAnalyzer->reveal()
+        );
+
+        $this->assertNull($extension->load(null));
+    }
+
+    public function testLoadNotExistingDocument()
+    {
+        $extension = new ContentTwigExtension(
```
wachterjohannes: add following line before here:

```php
         $this
             ->contentMapper
             ->load(Argument::cetera())
             ->willThrow($this->prophesize(DocumentNotFoundException::class)->reveal());
```

src/Sulu/Bundle/WebsiteBundle/Twig/Content/ContentTwigExtension.php
rogamoore: Hm, I'm not sure which exception you mean. :) Currently it is already being catched by Twig and converted to a Twig_Runtime exception. On production this will show a HTTP 500 Error.

image

rogamoore:
@wachterjohannes Done! Thanks for the support :)

Issue #3655, opened 2017-11-28
Acceptance Testing for React UI


danrot:
| Q | A
| --- | ---
| Bug? | no
| New Feature? | yes
| Sulu Version | ---
| Browser Version | ---

#### Actual Behavior

At the moment we do not have any acceptance tests.

#### Expected Behavior

We should have a few acceptance tests for the most important user journeys.

The most important things we should test are:
- [ ] Changing collection in MediaOverview
- [ ] Every List in our system
- [ ] Moving from every List to the corresponding Form

#### Possible Solutions

There is a [nice blog article](https://www.viget.com/articles/acceptance-testing-react-apps-with-jest-and-nightmare) describing a setup with Nightmare and Jest. But there are also other solutions out there like [testcafe](https://devexpress.github.io/testcafe/), which would also work with different browsers.

PR #3656, opened 2017-11-28
add kernel for tests

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes parts of #3655
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR adds the necessary code to run a sulu installation with this repository only. Before it was not possible to run sulu without sulu-standard or sulu-minimal.

#### Why?

This would be very handy if we start writing end-to-end tests. They've used to be in sulu-standard, but the problem with that is, that they are not run if a PR goes into sulu/sulu.

#### To Do

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@
 /tests/app/data
 /tests/app/logs
 /tests/Resources/cache
+/tests/web
 /vendor
 
 # Composer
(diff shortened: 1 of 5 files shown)


danrot:
Replaced by #3878, which was already merged.

PR #3662, opened 2017-11-30
Added ImageFocusPoint component

3rfan:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

An `ImageFocusPoint` component to select and highlight important parts of an image which then can be considered for cropping.

#### Example Usage

~~~js
initialState = {
    value: {
        x: 1,
        y: 1,
    }
};

handleChange = (value) => {
    setState({
        value
    });
};

<div>
    <ImageFocusPoint
        value={state.value}
        onChange={handleChange}
        image="https://source.unsplash.com/random/600x400"
    />
    <div>Selection: {JSON.stringify(state.value)}</div>
</div>
~~~

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/Resources/js/components/ImageFocusPoint/FocusPoint.js
+++ b/src/Sulu/Bundle/MediaBundle/Resources/js/components/ImageFocusPoint/FocusPoint.js
@@ -1,0 +1,73 @@
+// @flow
+import React from 'react';
+import classNames from 'classnames';
+import {Icon} from 'sulu-admin-bundle/components';
+import type {Point} from './types';
+import focusPointStyles from './focusPoint.scss';
+
+const ARROW_UP_ICON = 'arrow-up';
+
+type Props = {
+    size: number,
+    value: Point,
+    active: boolean,
+    onClick?: (value: Point) => void,
+    showArrow: boolean,
+    arrowDirection: number,
+};
+
+export default class FocusPoint extends React.PureComponent<Props> {
+    static defaultProps = {
+        active: false,
+        showArrow: true,
+        arrowDirection: 0,
+    };
+
+    handleClick = () => {
+        const {
+            value,
+            onClick,
+        } = this.props;
+
+        if (onClick) {
+            onClick(value);
+        }
+    };
+
+    render() {
+        const {
+            size,
+            active,
+            showArrow,
+            arrowDirection,
+        } = this.props;
+        const buttonStyle = {
+            width: `${size}%`,
+            height: `${size}%`,
+        };
+        const iconStyle = {
+            transform: `rotate(${arrowDirection}deg)`,
+        };
+        const focusPointClass = classNames(
+            focusPointStyles.focusPoint,
+            {
+                [focusPointStyles.active]: active,
+            }
+        );
+
+        return (
+            <button
+                style={buttonStyle}
+                onClick={this.handleClick}
+                disabled={active}
+                className={focusPointClass}
+            >
+                {showArrow && !active &&
+                    <div style={iconStyle}>
+                        <Icon name={ARROW_UP_ICON} />
+                    </div>
+                }
+            </button>
+        );
+    }
+}
--- a/src/Sulu/Bundle/MediaBundle/Resources/js/components/ImageFocusPoint/ImageFocusPointCell.js
+++ b/src/Sulu/Bundle/MediaBundle/Resources/js/components/ImageFocusPoint/ImageFocusPointCell.js
@@ -3,7 +3,7 @@ import React from 'react';
 import classNames from 'classnames';
 import {Icon} from 'sulu-admin-bundle/components';
 import type {Point} from './types';
-import focusPointStyles from './focusPoint.scss';
+import imageFocusPointCellStyles from './imageFocusPointCell.scss';
 
 const ARROW_UP_ICON = 'arrow-up';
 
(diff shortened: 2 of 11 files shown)


src/Sulu/Bundle/MediaBundle/Resources/js/components/ImageFocusPoint/FocusPoint.js
```diff
@@ -0,0 +1,73 @@
...
+
+const ARROW_UP_ICON = 'arrow-up';
+
+type Props = {
+    size: number,
+    value: Point,
+    active: boolean,
+    onClick?: (value: Point) => void,
+    showArrow: boolean,
+    arrowDirection: number,
```
danrot: Why don't you use a union type with strings here? Similar to what you did for the `Grid` component? I guess that would be easier to read when you are writing code.

src/Sulu/Bundle/MediaBundle/Resources/js/components/ImageFocusPoint/FocusPoint.js
```diff
@@ -0,0 +1,73 @@
...
+import React from 'react';
+import classNames from 'classnames';
+import {Icon} from 'sulu-admin-bundle/components';
+import type {Point} from './types';
+import focusPointStyles from './focusPoint.scss';
+
+const ARROW_UP_ICON = 'arrow-up';
+
+type Props = {
+    size: number,
```
danrot: Do we need different sizes?

src/Sulu/Bundle/MediaBundle/Resources/js/components/ImageFocusPoint/FocusPoint.js
danrot: Then we could maybe also merge the two previous attributes?

src/Sulu/Bundle/MediaBundle/Resources/js/components/ImageFocusPoint/FocusPoint.js
```diff
@@ -0,0 +1,73 @@
...
+    value: Point,
+    active: boolean,
+    onClick?: (value: Point) => void,
+    showArrow: boolean,
+    arrowDirection: number,
+};
+
+export default class FocusPoint extends React.PureComponent<Props> {
```
danrot: I think that this name is a little bit confusing... Why not calling it something like `ImageFocusPointCell`? Optional I would still extract something like a `SelectionGrid` 😉

src/Sulu/Bundle/MediaBundle/Resources/js/components/ImageFocusPoint/FocusPoint.js
3rfan: No but you can't set that from outside anyway. I did it this way because I don't want define `FOCUS_POINT_MATRIX_SIZE` multiple times.

src/Sulu/Bundle/MediaBundle/Resources/js/components/ImageFocusPoint/FocusPoint.js
3rfan: It's not that easy to create a generic grid which can be used for this component. I wouldn't do that now when there is no need for that.

3rfan:
@danrot lit update
Alt Text

src/Sulu/Bundle/MediaBundle/Resources/js/components/ImageFocusPoint/ImageFocusPointCell.js
```diff
@@ -0,0 +1,68 @@
...
+import imageFocusPointCellStyles from './imageFocusPointCell.scss';
+
+const ARROW_UP_ICON = 'arrow-up';
+
+type Props = {
+    size: number,
+    value: Point,
+    active: boolean,
+    onClick?: (value: Point) => void,
+    arrowDirection?: '-90deg' | '-45deg' | '0deg' | '45deg' | '90deg' | '125deg' | '180deg' | '225deg',
```
danrot: Sorry, I wasn't clear enough about that... I would have been going for something like `up-left`, `up`, `up-right`, `left`, ... You get the idea. Don't you think that would be nicer? If I read deg I always have to think about what does negative mean, where does it start and so on 😕

src/Sulu/Bundle/MediaBundle/Resources/js/components/ImageFocusPoint/ImageFocusPointCell.js
```diff
@@ -0,0 +1,68 @@
...
+};
+
+export default class ImageFocusPointCell extends React.PureComponent<Props> {
+    static defaultProps = {
+        active: false,
+        showArrow: true,
```
danrot: Should be gone now.

danrot: Seems like there is already a rule for that, at least if you are using prop-types: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/default-props-match-prop-types.md Maybe this also works with flow? Can you try?

3rfan: Tried that and it didn't do anything. Maybe it just works with `React.PropTypes`

3rfan:
@danrot update
Alt Text

PR #3428, opened 2017-07-03
Cache Refactoring

trickreich:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #3439, fixes #2635, fixes #3454
| Related issues/PRs | #2635, https://github.com/sulu/sulu-minimal/pull/53
| License | MIT
| Documentation PR | https://github.com/sulu/sulu-docs/pull/348

#### What's in this PR?

Set cache life time headers without listener

PR #3664, opened 2017-11-30
fix regressions in Symfony 3.4

danrot:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR fixes the regressions in Symfony 3.4, and should make the tests running again. However, this does not include the deprecations.

Please remember to merge back to develop!

#### Why?

Because we want the tests to succeed 😃

### DIFF
--- a/src/Sulu/Bundle/HttpCacheBundle/Resources/config/event-subscribers.xml
+++ b/src/Sulu/Bundle/HttpCacheBundle/Resources/config/event-subscribers.xml
@@ -22,12 +22,12 @@
             <tag name="sulu_document_manager.event_subscriber"/>
         </service>
 
-        <service id="sulu_http_cache.event_subscriber.flush" class="%sulu_http_cache.event_subscriber.flush_subscriber.class%">
+        <service id="sulu_http_cache.event_subscriber.flush" class="%sulu_http_cache.event_subscriber.flush_subscriber.class%" public="true">
             <argument type="service" id="sulu_http_cache.handler"/>
             <tag name="kernel.event_subscriber"/>
         </service>
 
-        <service id="sulu_http_cache.event_subscriber.update_response" class="%sulu_http_cache.event_subscriber.update_response_subscriber.class%">
+        <service id="sulu_http_cache.event_subscriber.update_response" class="%sulu_http_cache.event_subscriber.update_response_subscriber.class%" public="true">
             <argument type="service" id="sulu_http_cache.handler"/>
             <tag name="kernel.event_subscriber"/>
         </service>
--- a/src/Sulu/Bundle/PreviewBundle/Command/CacheClearCommand.php
+++ b/src/Sulu/Bundle/PreviewBundle/Command/CacheClearCommand.php
@@ -22,6 +22,8 @@ use Symfony\Component\HttpKernel\KernelInterface;
 
 class CacheClearCommand extends BaseCacheClearCommand
 {
+    protected static $defaultName = 'cache:clear';
+
     /**
      * {@inheritdoc}
      */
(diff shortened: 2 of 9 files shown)


src/Sulu/Bundle/HttpCacheBundle/Resources/config/event-subscribers.xml
```diff
@@ -22,12 +22,12 @@
             <tag name="sulu_document_manager.event_subscriber"/>
         </service>
 
-        <service id="sulu_http_cache.event_subscriber.flush" class="%sulu_http_cache.event_subscriber.flush_subscriber.class%">
+        <service id="sulu_http_cache.event_subscriber.flush" class="%sulu_http_cache.event_subscriber.flush_subscriber.class%" public="true">
```
danrot: Since there is still PR #3428 about the cache refactoring open (*cough* @chirimoya *cough*), I've decided to go for just making these services public for the tests now.

src/Sulu/Bundle/PreviewBundle/Command/CacheClearCommand.php
```diff
@@ -22,6 +22,8 @@
 
 class CacheClearCommand extends BaseCacheClearCommand
 {
+    protected static $defaultName = 'cache:clear';
```
danrot: The names are now initialized lazy, that's why they have to be defined static (so that they are available at runtime).

src/Sulu/Bundle/SnippetBundle/Tests/Unit/Snippet/DefaultSnippetManagerTest.php
```diff
@@ -63,10 +63,10 @@ public function saveDataProvider()
...
             ['sulu_io', 'de', 'test', 'test_homepage', '123-123-123', true, false],
             ['sulu_io', 'de', 'test', 'test_homepage', '123-123-123', false, false],
-            ['sulu_io', 'de', 'test', 'testCase', '123-123-123'],
-            ['sulu_io', 'de', 'test', 'testCase', '123-123-123', false],
-            ['sulu_io', 'de', 'test', 'testCase', '123-123-123', true, false],
-            ['sulu_io', 'de', 'test', 'testCase', '123-123-123', false, false],
+            ['sulu_io', 'de', 'test', 'testcase', '123-123-123'],
+            ['sulu_io', 'de', 'test', 'testcase', '123-123-123', false],
+            ['sulu_io', 'de', 'test', 'testcase', '123-123-123', true, false],
+            ['sulu_io', 'de', 'test', 'testcase', '123-123-123', false, false],
```
danrot: @alexander-schranz You've been looking into that issue with the parameter bag once, right? Is this the right way to fix it?

src/Sulu/Bundle/TranslateBundle/Tests/Functional/Translate/ExportTest.php
```diff
@@ -114,7 +115,13 @@ public function testExportWithFrontEnd()
...
         $expectedHash = file_get_contents(self::$fixturePath . '/shouldbes/sulu-test.frontend.en.xlf');
         $actualHash = file_get_contents(self::$fixturePath . '/sulu-test.frontend.en.xlf');
+        $actualHash = $this->removeTranslationIds($actualHash);
 
         $this->assertSame($expectedHash, $actualHash);
     }
+
+    private function removeTranslationIds($hash)
+    {
+        return preg_replace('/ id="(\w+)" /', ' ', $hash);
```
danrot: This is necessary, because the `id` changed in Symfony 3.4, therefore we can't check the content of the entire file anymore, and be compatible to both versions. That's why I just remove the `id` here (since it doesn't really matter anyway).

src/Sulu/Bundle/SnippetBundle/Tests/Unit/Snippet/DefaultSnippetManagerTest.php
alexander-schranz: In the constructor of the [DefaultSnippetManager]( https://github.com/sulu/sulu/blob/1.6.8/src/Sulu/Bundle/SnippetBundle/Snippet/DefaultSnippetManager.php#L62) the `array_change_key_case` need to be removed the test case need to have `testCase` as before.

alexander-schranz:
@danrot the fix for appveyor is currently not in master branch can you add it? https://github.com/sulu/sulu/commit/efb56c262f1c5fc5ab2e258ed5ea374cf2c8c4b9#diff-180360612c6b8c4ed830919bbb4dd459

PR #3667, opened 2017-12-01
Remove PaginationDecorator from Datagrid

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | yes
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This removes the `PaginationDecorator` component from the datagrid.

#### Why?

Because it is not necessary, if both pagination options implement the same interface.

#### BC Breaks/Deprecations

- `PaginationDecorator` is gone
- the `onLoad` method from the `InfiniteScroller` is now called `onChange`

#### To Do

- [x] Remove PaginationDecorator
- [x] Find a solution for Translations
- [x] Move loader to pagination (except for initial loader)


src/Sulu/Bundle/AdminBundle/Resources/js/containers/Datagrid/Datagrid.js
```diff
@@ -85,6 +87,9 @@ export default class Datagrid extends React.PureComponent<Props> {
         const page = store.getPage();
         const pageCount = store.pageCount;
         const Adapter = this.currentAdapter;
+        const PaginationAdapter: ComponentType<PaginationProps> = Adapter.getLoadingStrategy() === 'infiniteScroll'
```
danrot: All I am doing here is to say that the `PaginationAdapter` should be a React component with the `PaginationProps`. But the components themselves do not use the `PaginationProps` themselves, they just happen to have the same properties (that's basically duck typing).

I thought I am using it this way, because then the `InfiniteScroller` and the `Pagination` (both components), do not have to be dependent on the `Datagrid` (a container).

@chirimoya WDYT? Is that a good idea, or should we solve that in a different way? The only other thing keeping our restrictions I could imagine is to introduce the `PaginationProps` somewhere else, but I don't find a place for that at the moment... Maybe there should be a `types.js` file in the root of the AdminBundle's JavaScript?

danrot:
@chirimoya Most of the files have just been touched, because the Translator was moved.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Datagrid/Datagrid.js
chirimoya: go with the types.js

PR #3672, opened 2017-12-04
Added MultiUpload

3rfan:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #issuenum
| Related issues/PRs | #issuenum
| License | MIT
| Documentation PR | sulu/sulu-docs#prnum

#### What's in this PR?

A container-component to upload multiple files with drag`n`drop

#### Why?

To upload new files to Sulu

#### Example Usage

~~~js
<MultiMediaDropzone
    locale={locale}
    collectionId={collectionStore.id}
    onUpload={this.handleUpload}
>
    <ChildComponent />
</MultiMediaDropzone>
~~~

#### To Do

- [x] Add more tests
- [x] Add multiupload to the `MediaSelectionOverlay`

PR #3673, opened 2017-12-05
Added form to MediaDetail view

3rfan:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This PR adds a form to the `MediaDetail` view which you can use to edit information about the media.

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/Resources/js/components/SingleMediaDropzone/singleMediaDropzone.scss
+++ b/src/Sulu/Bundle/MediaBundle/Resources/js/components/SingleMediaDropzone/singleMediaDropzone.scss
@@ -4,7 +4,7 @@ $dropzoneSize: 400px;
 $uploadIconColor: $scorpion;
 $uploadInfoTextColor: $scorpion;
 $uploadIndicatorBackgroundColor: rgba($alabaster, .8);
-$mediaContainerBreakpoint: 520px; /* dropzoneSize + 2 * $zviewPadding */
+$mediaContainerBreakpoint: 520px; /* $dropzoneSize + 2 * $viewPadding because calc doesn't work for media-queries */
 
 .media-container {
     position: relative;
@@ -12,7 +12,7 @@ $mediaContainerBreakpoint: 520px; /* dropzoneSize + 2 * $zviewPadding */
     justify-content: center;
     align-items: center;
     width: $dropzoneSize;
-    height: $dropzoneSize;
+    height: auto;
     background-image: url('./checkerBackground.gif');
     background-size: 10px;
     cursor: pointer;
@@ -25,7 +25,7 @@ $mediaContainerBreakpoint: 520px; /* dropzoneSize + 2 * $zviewPadding */
         }
     }
 
-    &::before{
+    &::after {
         content: '';
         display: block;
         padding-top: 100%;
@@ -33,7 +33,6 @@ $mediaContainerBreakpoint: 520px; /* dropzoneSize + 2 * $zviewPadding */
 
     @media (max-width: $mediaContainerBreakpoint) {
         width: 100%;
-        height: auto;
     }
 }
 
--- a/src/Sulu/Bundle/MediaBundle/Resources/js/components/SingleMediaDropzone/singleMediaDropzone.scss
+++ b/src/Sulu/Bundle/MediaBundle/Resources/js/components/SingleMediaDropzone/singleMediaDropzone.scss
@@ -4,6 +4,7 @@ $dropzoneSize: 400px;
 $uploadIconColor: $scorpion;
 $uploadInfoTextColor: $scorpion;
 $uploadIndicatorBackgroundColor: rgba($alabaster, .8);
+$mediaContainerBreakpoint: 520px; /* dropzoneSize + 2 * $zviewPadding */
 
 .media-container {
     position: relative;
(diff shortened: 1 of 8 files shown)


3rfan:
@danrot 
Alt Text

danrot:
Except for the failing build it looks good 😉  Might be solved with a rebase, but wait till #3672 is merged.

src/Sulu/Bundle/MediaBundle/Resources/js/views/MediaDetail/mediaDetail.scss
```diff
@@ -2,4 +2,26 @@
 
 .media-detail {
     padding: $viewPadding;
+    display: flex;
+
+    @media (max-width: 720px) {
```
danrot: Have you decided the breakpoint on your own? `720px` seem pretty small: 
image

src/Sulu/Bundle/MediaBundle/Resources/js/views/MediaDetail/mediaDetail.scss
```diff
@@ -2,4 +2,26 @@
...
+.form-container {
+    flex: 1 1;
+}
+
+.preview-container {
+    margin-right: $viewPadding;
+    flex: 0 0 400px;
+
+    @media (max-width: 1024px) {
+        margin-right: 30px;
```
danrot: Did the designer say there is another margin level?

src/Sulu/Bundle/MediaBundle/Resources/js/views/MediaDetail/mediaDetail.scss
```diff
@@ -2,4 +2,26 @@
...
+    }
+
+    @media (max-width: 720px) {
```
danrot: If we use the same width more often a variable would probably be a good idea.

src/Sulu/Bundle/MediaBundle/Resources/js/views/MediaDetail/mediaDetail.scss
```diff
@@ -2,4 +2,26 @@
...
+        margin-bottom: 30px;
```
danrot: Can't we always set that? Shouldn't hurt, or does it?

danrot: I guess we should also center the picture if we switch to a single column

danrot:
Talked to Tim, following adjustments should be made:

- The margin between the image and the form should always be the same
- The form goes under the image when the screen is less than 900px
- The image and positioning is fine, but when the screen gets smaller than the image size, the image should be as wide as the form

src/Sulu/Bundle/MediaBundle/Resources/js/components/SingleMediaDropzone/singleMediaDropzone.scss
```diff
@@ -4,6 +4,7 @@ $dropzoneSize: 400px;
 $uploadIconColor: $scorpion;
 $uploadInfoTextColor: $scorpion;
 $uploadIndicatorBackgroundColor: rgba($alabaster, .8);
+$mediaContainerBreakpoint: 520px; /* dropzoneSize + 2 * $zviewPadding */
```
danrot: Why don't you calculate the value as described in the comment using css?

danrot:
image
image

The first picture is at 521px and the second at 520px. I think the transparent background should stay, shouldn't it?

src/Sulu/Bundle/MediaBundle/Resources/js/components/SingleMediaDropzone/singleMediaDropzone.scss
3rfan: Because you can't use `calc` inside a `media` query. https://stackoverflow.com/questions/23667208/calc-not-working-within-media-queries?answertab=votes#tab-top

src/Sulu/Bundle/MediaBundle/Resources/js/components/SingleMediaDropzone/singleMediaDropzone.scss
```diff
@@ -23,6 +24,17 @@ $uploadIndicatorBackgroundColor: rgba($alabaster, .8);
             transform: scale(1);
         }
     }
+
+    &::before{
```
danrot: build is failing because of that missing space.

src/Sulu/Bundle/MediaBundle/Resources/js/components/SingleMediaDropzone/singleMediaDropzone.scss
```diff
@@ -23,6 +24,17 @@ $uploadIndicatorBackgroundColor: rgba($alabaster, .8);
...
+        content: '';
+        display: block;
+        padding-top: 100%;
+    }
+
+    @media (max-width: $mediaContainerBreakpoint) {
+        width: 100%;
+        height: auto;
```
danrot: I think we could set this on the outer selector, instead of setting the width and height, couldn't we? Then we could minimize the code.

src/Sulu/Bundle/MediaBundle/Resources/js/components/SingleMediaDropzone/singleMediaDropzone.scss
danrot: Can you maybe put a code comment here? Otherwise it is just confusing

danrot:
@chirimoya This can be reviewed now. There is still a bug in IE11 and small screen, but I assume that is not a real issue (or do we expect people using IE11 on smartphones? 🙈)

PR #3674, opened 2017-12-05
Fixing mobx flow types

danrot:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR fixes the mobx flow types.

#### Why?

MobX seems to have updated their flow types, so we are having some issues now.

#### To Do

- [x] Check eslint

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -30,7 +30,7 @@ install:
   # install SSL and php
   - ps: Set-Service wuauserv -StartupType Manual
   - cinst -y OpenSSL.Light
-  - cinst -y php
+  - cinst -y php -version 7.1.12
 
   # configure PHP and enable extensions.
   - cd c:\tools\php71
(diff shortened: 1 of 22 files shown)


danrot:
I have to disable the `import/named` command for interface imports. There seems to be a bug about that in the eslint rule: https://github.com/benmosher/eslint-plugin-import/issues/927

appveyor.yml
```diff
@@ -30,7 +30,7 @@ install:
   # install SSL and php
   - ps: Set-Service wuauserv -StartupType Manual
   - cinst -y OpenSSL.Light
-  - cinst -y php
+  - cinst -y php -version 7.1.12
```
danrot: That makes appveyor work again for now. Problem is that PHP was updated to 7.2, and the folder in which php is located changes (from C:\tools\php71 to C:\tools\php72). The other problem is, that Sulu doesn't seem to be compatible to PHP 7.2 🙈

PR #3678, opened 2017-12-06
[RFC] Use doctrine iterator with a generator for list builder results

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | yes
| Deprecations? | no
| License | MIT

#### What's in this PR?

Change the list builder to return a generator instead of the whole array result.

#### Why?

That the listbuilder does not load all entities into the memory in a future step for example a streamed response could be generated for big CSV exports.

Inspired by Doctrine recommendation: http://doctrine-orm.readthedocs.io/projects/doctrine-orm/en/latest/reference/batch-processing.html#iterating-large-results-for-data-processing

#### Example Usage

~~~php
$generator = $listBuilder->execute();

// ... do more
~~~

#### BC Breaks/Deprecations

The biggest BC Break is that a `\Generator` is not countable a `count` on `\Generator` always returns 1. In some cases I used a count on the listBuilder result to avoid the COUNT QUERY in case of there exist only a first page or the result was the last page as you see in the following code.

```php
$list = $listBuilder->execute();

// The following code is not longer possible as count will always return 1
$count = count($list);
$offset = ($listBuilder->getCurrentPage() - 1) * $listBuilder->getLimit();
$total = $offset + $count;

if ($count >= $limit) {
     // This if avoids a count request on the first page if there exist no more pages ($total = count)
     // Also it avoids a count request for the last page as it can be calculated with the offset ($total = $offset + $count)
     $total = $listBuilder->count();
}
```

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -14,6 +14,10 @@ Follow upgrade path of following libraries:
 
 The api endpoint for `/admin/api/nodes/filter` was removed and replaced by `/admin/api/items`.
 
+### ListBuilder
+
+The `ListBuilderInterface::execute` function will not longer return the whole result array instead for performance improvements it returns a ˚\Generator` this means a `count()` on the result is not longer possible. If you need to count the result items you need todo this in a loop on the generator object for the total result call the `$listBuilder->count()` function.
+
 ## 1.6.9
 
 ### CacheBuilder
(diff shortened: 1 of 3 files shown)


alexander-schranz:
Seems the iterator not working always if the entity joins a `*-to-many` relation. Maybe we could make the iterator optional or we check if an iterator is possible or not.

alexander-schranz:
checking for `*-to-many` relation would be an overhead and sometimes return an array and sometimes a generator would be inconsistent so I will close this PR.

PR #3682, opened 2017-12-07
Reintroduce cache builder for bc promise

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Reintroduce cache builder. And skip the cache clear for symfony >= 3.4.0

#### Why?

It was removed in https://github.com/sulu/sulu/pull/3671/files but it should just be deprecated and only be removed in develop branch. For bc promise we introduce this service. To avoid crashes with e.g. old article bundle versions.

### DIFF
--- a/src/Sulu/Bundle/CoreBundle/Build/CacheBuilder.php
+++ b/src/Sulu/Bundle/CoreBundle/Build/CacheBuilder.php
@@ -1,0 +1,67 @@
+<?php
+
+/*
+ * This file is part of Sulu.
+ *
+ * (c) MASSIVE ART WebServices GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\CoreBundle\Build;
+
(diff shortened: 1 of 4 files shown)


src/Sulu/Bundle/CoreBundle/Build/CacheBuilder.php
```diff
@@ -0,0 +1,66 @@
...
+    public function build()
+    {
+        @trigger_error(
+            'CacheBuilder is deprecated since version 1.6.10 and will be removed in 2.0 use the default symfony command cache:clear instead.',
+            E_USER_DEPRECATED
+        );
+
+        if (version_compare(Kernel::VERSION, '3.4.0') >= 0) {
+            @trigger_error(
+                'Skip clearing the cache with sulu:build as it is not longer supported with Symfony >=3.4 use cache:clear command instead.',
```
danrot: Is there no way we can really log something to the terminal? I think we really have to make people aware, that this step is not doing anything.

PR #3698, opened 2017-12-21
SEO description length changed from 155 to 320

trickreich:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes --
| Related issues/PRs | --
| License | MIT
| Documentation PR | --

#### What's in this PR?

SEO description length changed from 155 to 320

#### Why?

Google changed the max length.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@ CHANGELOG for Sulu
 ==================
 
 * dev-master
+    * ENHANCEMENT #XXXX [ContentBundle]           SEO description length changed from 155 to 320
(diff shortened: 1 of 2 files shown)


danrot (approved):

sanderquirynen:
While Google allows 320 characters length, Moz (and other sources) [recommend a max length of ~155-160 characters](https://moz.com/learn/seo/meta-description), because that's what they get truncated to in the search results.

And since the character limit is a soft limit, I think it might be better to change it back to 155 or 160.

PR #3703, opened 2017-12-30
Make Sulu compatible with PHP 7.2

danrot:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR makes sure the `Webspace::segments`  variable is an array from the very beginning.

#### Why?

Because starting with PHP 7.2 the ` count`  method will fail on this variable.

See https://wiki.php.net/rfc/counting_non_countables

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ matrix:
     - php: 5.5
       env:
         - COMPOSER_FLAGS="--prefer-lowest --prefer-dist --no-interaction"
-    - php: 7.0
+    - php: 7.2
       env:
         - COMPOSER_FLAGS="--prefer-dist --no-interaction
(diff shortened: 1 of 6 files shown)


danrot:
Looks like we can't test against php 7.2 before we update to [PHPUnit 6.5](https://github.com/sebastianbergmann/phpunit/issues/2887). This also includes changing all tests to using real namespaces in phpunit...

ichdertobi:
The changes were merged only into sulu:release/1.5. Would it be possible that the changes could be merged to sulu:release/1.6, too?

danrot:
We'll make a release soon, then you also have these changes in all new tags.

PR #3706, opened 2018-01-08
Add debug 404 image for image converter errors

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Return a svg with `ERROR CODE` when image format could not be generated or original file not found. And also log an error into the log.

#### Why?

To show devs some feedback when an image could not been generated.

#### Example Usage

Open an image which format does not exist. Should return an image like this:

bildschirmfoto 2018-02-16 um 12 45 58

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/Controller/MediaStreamController.php
+++ b/src/Sulu/Bundle/MediaBundle/Controller/MediaStreamController.php
@@ -45,7 +45,7 @@ class MediaStreamController extends Controller
     /**
      * @param Request $request
      *
-     * @return \Symfony\Component\HttpFoundation\Response
+     * @return Response
      */
     public function getImageAction(Request $request)
     {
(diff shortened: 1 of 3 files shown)


alexander-schranz:
@chirimoya @danrot what you think about this? The HTTP Statuscode is still 404 so it is logged correctly.

alexander-schranz:
what we do with this one @chirimoya 

A. Keep it this way
B. Move 404 image into twig file for easing overwrite it (BC Break in constructor as twig is needed)
C. Change nothing (decline PR) (atleast the logger should be added)

alexander-schranz:
@chirimoya @danrot what you think about this solution. Only output the error image in case of debug is enabled?

danrot:
I think I would keep it minimal, and just add a log message, which is currently also missing, right?

alexander-schranz:
@danrot no log error message is written.

mamazu:
I think this feature would still be great. Especially if you import the database from the live system but not the images. At least you don't get empty alt tags everywhere.

mamazu:
Replaced and rebased by #8000

PR #3714, opened 2018-01-16
Load metadata for Form and Datagrid asynchronous from server

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | yes
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR allows to dynamically load the metadata from the server. Only the data which is required from the server is loaded. The data is also cached for later use without loading it again from the server.

#### Why?

Because in large installations it might take a while to load all the data at once.

#### To Do


danrot:
I've seen that stores have no documentation yet. Wanted to add it, but then I'd have to add documentation for all stores. Might do that in a separate PR.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/FormStore.js
```diff
@@ -0,0 +1,66 @@
+// @flow
+import {action, computed, observable} from 'mobx';
+import type {IObservableValue} from 'mobx'; // eslint-disable-line import/named
+import ResourceStore from '../../../stores/ResourceStore';
+import type {Schema} from '../types';
+import metadataStore from './MetadataStore';
+
+function addObjectProperty(schema: Schema, key, object) {
```
chirimoya: `function addSchemaProperties(data, key, schema: Schema) {`

PR #3724, opened 2018-01-27
Support utf8mb4

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Related issues/PRs | https://github.com/jackalope/jackalope-doctrine-dbal/pull/353
| License | MIT

#### What's in this PR?

Decrease all indexed fields length to 191.

#### Why?

To support utf8mb4 the max index length on default innodb is 191. https://dev.mysql.com/doc/refman/5.6/en/charset-unicode-conversion.html

Testing this issue on different version of mysql different changes need to be done:
For example travis uses 5.6.33 there you only need to decrease indexes which are unique.
On my local installation 5.6.21 there you also need to decrease all indexed fields e.g. `FileVersion:mimeType`. So I did go the way to support also the older version of mysql with utf8mb4.

#### Example Usage

~~~php
doctrine:
    dbal:
        charset: utf8mb4
        default_table_options:
            charset: utf8mb4
            collate: utf8mb4_unicode_ci
~~~

#### BC Breaks/Deprecations

Some field length decreased.

#### To Do

- [x] Write UPGRADE file.
- [x] Needs https://github.com/jackalope/jackalope-doctrine-dbal/pull/353 (released in 1.3.2)
- [x] Should this be a ~1.6 hotfix~ or 2.0?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,8 @@ env:
     - SYMFONY__PHPCR__TRANSPORT=doctrinedbal
     - TEST_FLAGS=""
     - SYMFONY__DATABASE__DRIVER=pdo_mysql
+    - SYMFONY__DATABASE__CHARSET=utf8mb4
+    - SYMFONY__DATABASE__COLLATE=utf8mb4_unicode_ci
 
 matrix:
   include:
@@ -31,6 +33,8 @@ matrix:
         - SYMFONY__DATABASE__USER=postgres
         - SYMFONY__DATABASE__PASSWORD=postgres
         - SYMFONY__PHPCR__TRANSPORT=jackrabbit
+        - SYMFONY__DATABASE__CHARSET=UTF8
+        - SYMFONY__DATABASE__COLLATE=
         # restart jackrabbit after each suite see: https://github.com/sulu-io/sulu/issues/2137
         - TEST_FLAGS="--jackrabbit-restart"
 
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -14,6 +14,32 @@ Follow upgrade path of following libraries:
 
 The api endpoint for `/admin/api/nodes/filter` was removed and replaced by `/admin/api/items`.
 
+### Database changes
+
+To support utf8mb4 we needed to change some database entities which you also need to migrate when not using utf8mb4:
+Run the following SQL to migrate to the new schema:
+
+```sql
+ALTER TABLE me_format_options CHANGE format_key format_key VARCHAR(191) NOT NULL;
+ALTER TABLE me_collections CHANGE collection_key collection_key VARCHAR(191) DEFAULT NULL;
+ALTER TABLE me_collection_types CHANGE collection_type_key collection_type_key VARCHAR(191) DEFAULT NULL;
+ALTER TABLE me_media_types CHANGE name name VARCHAR(191) NOT NULL;
+ALTER TABLE me_file_versions CHANGE mimeType mimeType VARCHAR(191) DEFAULT NULL;
+ALTER TABLE se_users CHANGE email email VARCHAR(191) DEFAULT NULL;
+ALTER TABLE se_role_settings CHANGE settingKey settingKey VARCHAR(191) NOT NULL;
+ALTER TABLE se_permissions CHANGE context context VARCHAR(191) NOT NULL;
+ALTER TABLE se_access_controls CHANGE entityClass entityClass VARCHAR(191) NOT NULL;
+ALTER TABLE ca_categories CHANGE category_key category_key VARCHAR(191) DEFAULT NULL;
+ALTER TABLE ca_keywords CHANGE keyword keyword VARCHAR(191) NOT NULL;
+ALTER TABLE ta_tags CHANGE name name VARCHAR(191) NOT NULL;
+ALTER TABLE we_domains CHANGE url url VARCHAR(191) NOT NULL;
+ALTER TABLE we_analytics CHANGE webspace_key webspace_key VARCHAR(191) NOT NULL;
+ALTER TABLE ro_routes CHANGE path path VARCHAR(191) NOT NULL, CHANGE entity_class entity_class VARCHAR(191) NOT NULL, CHANGE entity_id entity_id VARCHAR(191) NOT NULL;
+ALTER TABLE me_collection_meta CHANGE title title VARCHAR(191) NOT NULL;
+ALTER TABLE me_file_version_meta CHANGE title title VARCHAR(191) NOT NULL;
+ALTER TABLE me_file_versions CHANGE name name VARCHAR(191) NOT NULL;
+```
+
 ## 1.6.11
 
 ### SEO Description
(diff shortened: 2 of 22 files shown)


src/Sulu/Bundle/MediaBundle/Resources/config/doctrine/CollectionMeta.orm.xml
```diff
@@ -4,7 +4,6 @@
                   xsi:schemaLocation="http://doctrine-project.org/schemas/orm/doctrine-mapping http://doctrine-project.org/schemas/orm/doctrine-mapping.xsd">
     <entity name="Sulu\Bundle\MediaBundle\Entity\CollectionMeta" table="me_collection_meta">
         <indexes>
-            <index columns="title"/>
```
alexander-schranz: not sure which fields we should set to 191 or remove from being indexed. /cc @chirimoya @danrot

src/Sulu/Bundle/MediaBundle/Resources/config/doctrine/FileVersion.orm.xml
```diff
@@ -8,7 +8,6 @@
             <index columns="mimeType"/>
```
alexander-schranz: this need to be set also to 191 did set it to 255 to test travis with utf8mb4 but travis seems to be not failing with utf8mb4. Seems like travis mysql runs on innodb barracuda driver and not the default storage of mysql.

src/Sulu/Bundle/MediaBundle/Resources/config/doctrine/FileVersion.orm.xml
```diff
@@ -8,7 +8,6 @@
...
             <index columns="size"/>
             <index columns="version"/>
-            <index columns="name"/>
```
alexander-schranz: I think the name don't need an index because the database can't use an Index with '%value%' ? /cc @sulu/core-team

src/Sulu/Bundle/MediaBundle/Resources/config/doctrine/FileVersionMeta.orm.xml
```diff
@@ -6,7 +6,6 @@
             table="me_file_version_meta"
             repository-class="Sulu\Bundle\MediaBundle\Entity\FileVersionMetaRepository">
         <indexes>
-            <index columns="title"/>
```
alexander-schranz: I think the title don't need an index because the database can't use an Index with '%value%' ? /cc @sulu/core-team

alexander-schranz:
Testing this issue on different version of mysql different changes need to be done:
For example travis uses 5.6.33 there you only need to decrease indexes which are unique.
On my local installation 5.6.21 there you also need to decrease all indexed fields e.g. `FileVersion:mimeType`. So I did go the way to support also the older version of mysql with utf8mb4.

.travis.yml
```diff
@@ -13,6 +13,8 @@ env:
     - SYMFONY__PHPCR__TRANSPORT=doctrinedbal
     - TEST_FLAGS=""
     - SYMFONY__DATABASE__DRIVER=pdo_mysql
+    - SYMFONY__DATABASE__CHARSET=utf8mb4
+    - SYMFONY__DATABASE__COLLATE=utf8mb4_unicode_ci
```
alexander-schranz: added that travis is correctly failing when somebody increase the field: example: https://travis-ci.org/sulu/sulu/jobs/335311387

src/Sulu/Bundle/MediaBundle/Resources/config/doctrine/CollectionMeta.orm.xml
danrot: Are you completely sure that the title is never queried with `value%`? Because that would work with an index. I think I would leave it there just to be save.

src/Sulu/Bundle/MediaBundle/Resources/config/doctrine/FileVersion.orm.xml
danrot: Same as above, I am not compeletely sure if we never query for `value%`... And thinking of it, if you just remove that index it might make applications written by other developers a lot slower.

src/Sulu/Bundle/MediaBundle/Resources/config/doctrine/FileVersionMeta.orm.xml
danrot: Same as above

src/Sulu/Bundle/SecurityBundle/Resources/config/doctrine/AccessControl.orm.xml
```diff
@@ -14,7 +14,7 @@
 
         <field name="permissions" type="smallint" column="permissions"/>
         <field name="entityId" type="integer" column="entityId"/>
-        <field name="entityClass" type="string" column="entityClass"/>
+        <field name="entityClass" type="string" column="entityClass" length="191"/>
```
danrot: What should we do if the `entityClass` is longer than this?

alexander-schranz: mysql would throw an exception (`Data too long for column 'entityClass'`). The current longest classname in my current project in the ro_routes table was 50 characters (`Sulu\Bundle\ArticleBundle\Document\ArticleDocument`).

src/Sulu/Bundle/MediaBundle/Resources/config/doctrine/CollectionMeta.orm.xml
alexander-schranz: Currently in the repository always with `%` prefix. And I think the field descriptors search do the same?
But we can let the index and set the field length 191 characters.

alexander-schranz:
@danrot as discussed I did add the indexes on the name and title fields again (needed for sorting in core). But did short the title and name to 191 as needed when they are indexed.

Branch is also rebased on develop.

UPGRADE.md
```diff
@@ -14,6 +14,32 @@ Follow upgrade path of following libraries:
...
+ALTER TABLE se_users CHANGE email email VARCHAR(191) DEFAULT NULL;
+ALTER TABLE se_role_settings CHANGE settingKey settingKey VARCHAR(191) NOT NULL;
+ALTER TABLE se_permissions CHANGE context context VARCHAR(191) NOT NULL;
+ALTER TABLE se_access_controls CHANGE entityClass entityClass VARCHAR(191) NOT NULL;
+ALTER TABLE ca_categories CHANGE category_key category_key VARCHAR(191) DEFAULT NULL;
+ALTER TABLE ca_keywords CHANGE keyword keyword VARCHAR(191) NOT NULL;
+ALTER TABLE ta_tags CHANGE name name VARCHAR(191) NOT NULL;
+ALTER TABLE we_domains CHANGE url url VARCHAR(191) NOT NULL;
+ALTER TABLE we_analytics CHANGE webspace_key webspace_key VARCHAR(191) NOT NULL;
+ALTER TABLE ro_routes CHANGE path path VARCHAR(191) NOT NULL, CHANGE entity_class entity_class VARCHAR(191) NOT NULL, CHANGE entity_id entity_id VARCHAR(191) NOT NULL;
```
danrot: I think you've forgot the new fields we have discussed here :-)

alexander-schranz: added

PR #3752, opened 2018-02-06
Overwrite 'doctrine:phpcr:workspace:import' set default to throw

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | none
| Related issues/PRs | https://github.com/phpcr/phpcr-utils
| License | MIT
| Documentation PR | none

#### What's in this PR?

This PR overwrites the command `doctrine:phpcr:workspace:import` and set the default of `uuid-behavior` to `throw`.

#### Why?

This disables the regeneration of `uuid` while import from xml.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -64,7 +64,8 @@
         "symfony/symfony": "2.8.10",
         "symfony/phpunit-bridge": "2.8.10",
         "symfony/monolog-bundle": "2.8.10",
-        "jackalope/jackalope": "1.2.8 || 1.3.0 || 1.3.1"
+        "jackalope/jackalope": "1.2.8 || 1.3.0 || 1.3.1",
+        "phpcr/phpcr-utils": "1.2.1 - 1.2.10 || 1.3.1 || 1.3.2"
     },
     "replace": {
         "sulu/media-bundle": "self.version",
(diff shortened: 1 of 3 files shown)


wachterjohannes:
@danrot @chirimoya the phpcr-utils are only compatible with php `^5.6`. what should we do?

wachterjohannes:
`phpcr/phpcr-utils` is released in 1.2 and 1.3 release

wachterjohannes:
@danrot this is ready to review

composer.json
```diff
@@ -64,7 +64,8 @@
         "symfony/symfony": "2.8.10",
         "symfony/phpunit-bridge": "2.8.10",
         "symfony/monolog-bundle": "2.8.10",
-        "jackalope/jackalope": "1.2.8 || 1.3.0 || 1.3.1"
+        "jackalope/jackalope": "1.2.8 || 1.3.0 || 1.3.1",
+        "phpcr/phpcr-utils": "1.2.1 - 1.2.10 || 1.3.1 || 1.3.2"
```
danrot: Why do we have such a strange version constraint here? Do you think 1.3.3 will already break again? Same for `jackalope/jackalope`, although that was already like that before.

danrot: Ah, sry, didn't see that this is the conflict section :see_no_evil:

PR #3757, opened 2018-02-08
Correct extension stripping

### LINKED ISSUE #3755: Using json by default (without url extension) results in 404 for articles
| Q | A
| --- | ---
| Bug? | yes
| New Feature? | no
| Sulu Version | 1.6.14 

#### Actual Behavior

When using json as request format by default, without using .json extensions, the RouteProvider cuts off the last character of the url.

#### Expected Behavior

The RouteProvider should check if the extension exists, or only remove the actual extension.

#### Steps to Reproduce
When you set the request format in your project:

``` php
$request->setRequestFormat('json');
```
And visit a URL that should be routed by the RouteProvider, the page is not found.

jaapromijn:
The removal of the extension of formats should only take place when
the format is present as extension in the url. (#3755)

| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #3755
| Related issues/PRs | #3755
| License | MIT

#### Why?

When using json as request format by default, without using .json extensions, the RouteProvider cut off the last character of the url, resulting in a 404.
In this PR, the stripping of the extension is only attempted if the format extension is present.

### DIFF
--- a/src/Sulu/Bundle/RouteBundle/Tests/Unit/Routing/RouteProviderTest.php
+++ b/src/Sulu/Bundle/RouteBundle/Tests/Unit/Routing/RouteProviderTest.php
@@ -354,4 +354,33 @@ class RouteProviderTest extends \PHPUnit_Framework_TestCase
 
         $this->assertCount(0, $collection);
     }
+
+    public function testGetRouteCollectionForRequestWithoutFormatExtension()
+    {
+        $request = $this->prophesize(Request::class);
+        $request->getPathInfo()->willReturn('/de/test');
+        $request->getLocale()->willReturn('de');
+        $request->getRequestFormat()->willReturn('json');
+
+        $routeEntity = $this->prophesize(RouteInterface::class);
+        $routeEntity->getEntityClass()->willReturn('Example');
+        $routeEntity->getEntityId()->willReturn('1');
+        $routeEntity->getId()->willReturn(1);
+        $routeEntity->getPath()->willReturn('/test');
+        $routeEntity->isHistory()->willReturn(false);
+        $routeEntity->getLocale()->willReturn('de');
+
+        $this->routeRepository->findByPath('/test', 'de')->willReturn($routeEntity->reveal());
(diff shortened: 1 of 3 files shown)


src/Sulu/Bundle/RouteBundle/Tests/Unit/Routing/RouteProviderTest.php
```diff
@@ -354,4 +354,33 @@ public function testGetRouteCollectionForRequestEndingDot()
 
         $this->assertCount(0, $collection);
     }
+
+    public function testGetRouteCollectionForRequestWithoutFormatExtension()
+    {
+        $request = $this->prophesize(Request::class);
+        $request->getPathInfo()->willReturn('/de/test');
```
wachterjohannes: i think the reuqest will return `/de/test.json` in a real-world example? didn't it?

jaapromijn: The use case is actually requesting a JSON page without using the `.json` extension. 
The requested url for example is:
```
https://www.example.com/en/articles/foo-bar
```
And in my own request analyser the request format is set to json:
```
 $request->setRequestFormat('json');
```
So the path info should not return a path with a `.json` extension in this case? What you describe is already [covered in another test](https://github.com/jaapromijn/sulu/blob/146c7d27b5ddb9923088cbfda7a248d0a5fc2de2/src/Sulu/Bundle/RouteBundle/Tests/Unit/Routing/RouteProviderTest.php#L153), I think.

wachterjohannes: correct :) haven't recognized this - this PR is good! thanks for contribution

jaapromijn: 👍

PR #3758, opened 2018-02-08
Implement creating new entities using Forms

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR will allow to add new entites using the standard list and form react components.

#### Why?

Because that's one of our most basic functionility.

#### To Do

### DIFF
--- a/src/Sulu/Bundle/SnippetBundle/Admin/SnippetAdmin.php
+++ b/src/Sulu/Bundle/SnippetBundle/Admin/SnippetAdmin.php
@@ -110,6 +110,7 @@ class SnippetAdmin extends Admin
             (new Route('sulu_snippet.add_form.detail', '/details', 'sulu_admin.form'))
                 ->addOption('tabTitle', 'sulu_snippet.details')
                 ->addOption('backRoute', 'sulu_snippet.list')
+                ->addOption('editRoute', 'sulu_snippet.edit_form.detail')
                 ->setParent('sulu_snippet.add_form'),
             (new Route('sulu_snippet.edit_form', '/snippets/:locale/:id', 'sulu_admin.resource_tabs'))
                 ->addOption('resourceKey', 'snippets')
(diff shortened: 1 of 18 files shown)


src/Sulu/Bundle/SnippetBundle/Admin/SnippetAdmin.php
```diff
@@ -100,22 +100,30 @@ function(Localization $localization) {
...
+                ->addOption('addRoute', 'sulu_snippet.add_form.detail')
+                ->addOption('editRoute', 'sulu_snippet.edit_form.detail')
                 ->addOption('locales', $snippetLocales)
                 ->addAttributeDefault('locale', $snippetLocales[0]),
-            (new Route('sulu_snippet.form', '/snippets/:locale/:id', 'sulu_admin.resource_tabs'))
+            (new Route('sulu_snippet.add_form', '/snippets/:locale/add', 'sulu_admin.resource_tabs'))
                 ->addOption('resourceKey', 'snippets')
                 ->addOption('locales', $snippetLocales),
-            (new Route('sulu_snippet.form.detail', '/details', 'sulu_admin.form'))
+            (new Route('sulu_snippet.add_form.detail', '/details', 'sulu_admin.form'))
```
danrot: The `sulu_snippet.add_form.detail` and `sulu_snippet.edit_form.detail` are almost identical. The only difference is the parent route. Maybe we should think about a solution for that. I came up with two different solutions:

1. Create a `withParent` method on the route. This works similar like for PSR-7 requests, the method would return a cloned route with this parent set. Then we could do something like this:

```php
$formDetailRoute = new Route(...);

return [
    $formDetailRoute->withParent('sulu_snippet.add_form'),
    $formDetailRoute->withParent('sulu_snippet.edit_form'),
];
```

2. Allow a route to have multiple parents

chirimoya: not yet

PR #3762, opened 2018-02-09
Implement Select FieldType

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

The select field for forms.

#### Why?

Because we need that for the contact section.

#### TODO

- [ ] Pass default value up to form


danrot:
@chirimoya Two things:

1. How do we handle the documentation of these fields? Or do we say they are already documented enough in our [Sulu docs](http://docs.sulu.io/en/latest/reference/content-types/single_select.html)?
2. I think I'll have a look at the default value todo after this has been merged in #3758, in the edit form there is no way to have it not selected, because there is always a value from the server coming.

PR #3763, opened 2018-02-09
Fixed bug when position of PopOver wasn't correctly calculated

trickreich:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes --
| Related issues/PRs | --
| License | MIT
| Documentation PR | --

#### What's in this PR?

Fixed bug when position of PopOver wasn't correctly calculated.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Popover/PopoverPositioner.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Popover/PopoverPositioner.js
@@ -55,8 +55,9 @@ export default class PopoverPositioner {
 
         let crop = PopoverPositioner.cropVerticalDimensions(dimensions, windowHeight);
 
-        // If after making sure, the popover does not overflow the top and the bottom border of the screen, the
-        // popover succeeds the minimum height, no more steps have to be taken.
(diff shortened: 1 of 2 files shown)


danrot:
I think especially for tricky behavior like this there should be a test...

src/Sulu/Bundle/AdminBundle/Resources/js/components/Popover/PopoverPositioner.js
```diff
@@ -55,9 +55,10 @@ export default class PopoverPositioner {
 
         let crop = PopoverPositioner.cropVerticalDimensions(dimensions, windowHeight);
 
-        // If after making sure, the popover does not overflow the top and the bottom border of the screen, the
-        // popover succeeds the minimum height, no more steps have to be taken.
-        if (crop.dimensions.height >= MIN_HEIGHT) {
+        // If after making sure, the popover does not overflow the top and the bottom border of the screen,
+        // the popover succeeds the minimum height, no more steps have to be taken.
+        // Only return here if "alignOnVerticalAnchorEdges" edges is false.
```
danrot: I really appreciate that you see this code is complex, but this additional comment does not help at all, so I would tend to remove it...

PR #3764, opened 2018-02-09
Allow dynamic order of elements in webspace xmls

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes sulu/sulu-docs#360
| License | MIT

#### What's in this PR?

It allow dynamic order of items in the webspace, template and format xmls.

#### Why?

For beginners a fix order can be a pitfall. 

#### Example Usage

See sulu/sulu-docs#360 issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 CHANGELOG for Sulu
 ==================
 
+* dev-release/1.5
+    * ENHANCEMENT #3764 [Component]             Allow dynamic order of elements in webspace xml
+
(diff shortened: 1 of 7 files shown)


danrot:
For me that change is fine, I think it makes writing these XMLs a lot easier, and I don't see any value in demanding a certain order.

But just to be sure, I have requested reviews from the @sulu/core-team, if anybody has any serious concerns about this.

trickreich:
is totally ok for me..

PR #3772, opened 2018-02-14
WebspaceOverview: Load webspaces from server

trickreich:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes --
| Related issues/PRs | --
| License | MIT
| Documentation PR | --

#### What's in this PR?

`WebspaceOverview`: Load webspaces from server.
Used is the `WebspaceStore`.

### DIFF
--- a/src/Sulu/Bundle/ContentBundle/Resources/js/stores/WebspaceStore/types.js
+++ b/src/Sulu/Bundle/ContentBundle/Resources/js/stores/WebspaceStore/types.js
@@ -1,0 +1,44 @@
+// @flow
+export type Webspace = {
+    name: string,
+    key: string,
+    localizations: Array<Localization>,
+    urls: Array<Url>,
+    allLocalizations: Array<LocalizationItem>,
+    portalInformation: Array<PortalInformation>,
+    children: Array<Localization>,
+};
+
+export type Localization = {
+    locale: string,
+    language: string,
+    country: string,
+    shadow: string,
+    default: boolean,
+    xDefault: boolean,
+    children: Array<Localization>,
+};
+
+export type Url = {
+    url: string,
+    language: string,
+    country: string,
+    segment: string,
+    redirect: string,
+    main: boolean,
+    analyticsKey: string,
+    environment: string,
+};
+
+export type LocalizationItem = {
+    localization: string,
+    name: string,
+};
+
+export type PortalInformation = {
+    webspaceKey: string,
+    portalKey: string,
+    locale: string,
+    url: string,
+    main: boolean,
+};
--- a/src/Sulu/Bundle/ContentBundle/Resources/js/stores/WebspaceStore/types.js
+++ b/src/Sulu/Bundle/ContentBundle/Resources/js/stores/WebspaceStore/types.js
@@ -6,7 +6,6 @@ export type Webspace = {
     urls: Array<Url>,
     allLocalizations: Array<LocalizationItem>,
     portalInformation: Array<PortalInformation>,
-    children: Array<Localization>,
 };
 
 export type Localization = {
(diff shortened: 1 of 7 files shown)


src/Sulu/Bundle/ContentBundle/Resources/js/stores/WebspaceStore/types.js
```diff
@@ -0,0 +1,44 @@
+// @flow
+export type Webspace = {
+    name: string,
+    key: string,
+    localizations: Array<Localization>,
+    urls: Array<Url>,
+    allLocalizations: Array<LocalizationItem>,
+    portalInformation: Array<PortalInformation>,
+    children: Array<Localization>,
```
danrot: Doesn't that go into the `Localization`? I think the webspace itself doesn't have any children...

src/Sulu/Bundle/ContentBundle/Resources/js/views/WebspaceOverview/WebspaceOverview.js
```diff
@@ -14,11 +17,45 @@ class WebspaceOverview extends React.Component<ViewProps> {
...
+        this.setDefaultLocaleForWebspace();
     };
 
+    @action setDefaultLocaleForWebspace = () => {
+        if (!this.selectedWebspace) {
+            return;
+        }
+
+        const localization = this.selectedWebspace.localizations.find((localizations) => {
+            if (localizations.children) {
```
danrot: `localization`, because it is a single instance, isn't it?

src/Sulu/Bundle/ContentBundle/Resources/js/views/WebspaceOverview/WebspaceOverview.js
```diff
@@ -14,11 +17,45 @@ class WebspaceOverview extends React.Component<ViewProps> {
...
+                localizations.children.find((localizations) => {
```
danrot: Does this also work more than two levels deep? I think this should be called recursive to work properly...

src/Sulu/Bundle/ContentBundle/Resources/js/views/WebspaceOverview/WebspaceOverview.js
```diff
@@ -47,44 +88,54 @@ class WebspaceOverview extends React.Component<ViewProps> {
...
     }
 }
 
 export default withToolbar(WebspaceOverview, function() {
-    // TODO: Load this data dynamically from server!
+    if (!this.selectedWebspace) {
+        return {};
+    }
+
+    const options = this.selectedWebspace.allLocalizations.map(function (localization) {
```
danrot: Any reason you are not using the `() => ...` shortcut as above here?

src/Sulu/Bundle/ContentBundle/Resources/js/views/WebspaceOverview/WebspaceOverview.js
```diff
@@ -47,44 +88,54 @@ class WebspaceOverview extends React.Component<ViewProps> {
...
     const locale = {
         value: this.locale.get(),
         onChange: action((locale) => {
             this.locale.set(locale);
         }),
-        options: [{
-            value: 'en',
-            label: 'en',
-        }],
+        options: options,
```
danrot: just `options,` would be enough in this line

src/Sulu/Bundle/ContentBundle/Resources/js/views/WebspaceOverview/WebspaceOverview.js
danrot: Especially because you could reduce it to something like this:

```javascript
(localization) => ({
    value: localization.localization,
    label: localization.name,
})
```

PR #3773, opened 2018-02-15
Add minOccurs and maxOccurs to Blocks

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR adds the possibility to have a min and max restriction on the blocks. It will also not be possible anymore to click add or remove if the max resp. min restriction is already met.

#### Why?

Because we want to be able to restrict the content manager in that regard.

#### To Do

- [x] Already show `minOccurs` items in the beginning

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/BlockCollection.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/BlockCollection.js
@@ -11,6 +11,8 @@ import blockCollectionStyles from './blockCollection.scss';
 import type {BlockEntry, RenderBlockContentCallback} from './types';
 
 type Props = {
+    maxOccurs?: number,
+    minOccurs?: number,
     onChange: (value: Array<BlockEntry>) => void,
     renderBlockContent: RenderBlockContentCallback,
     types?: {[key: string]: string},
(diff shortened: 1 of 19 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/BlockCollection.js
```diff
@@ -112,18 +141,20 @@ export default class BlockCollection extends React.Component<Props> {
...
                     types={types}
                     useDragHandle={true}
                     value={identifiedValues}
                 />
-                <Button skin="secondary" onClick={this.handleAddBlock}>
-                    <Icon name="su-plus" className={blockCollectionStyles.addButtonIcon} />
-                    {translate('sulu_admin.add_block')}
-                </Button>
+                {!this.hasMaximumReached() &&
+                    <Button skin="secondary" onClick={this.handleAddBlock}>
```
chirimoya: Button should only be disabled if max occurs is reached.

danrot:
@chirimoya I didn't find a clear design for the disabled button, so I have done the closest design IMO :see_no_evil:

danrot:
And personally I think that it is confusing to hide the remove icon but show the button only disabled.

PR #3784, opened 2018-02-20
Added change method to stores

trickreich:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Added `change` method on stores.

#### Why?

We need to distinguish between initializing and changing.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/FormStore.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/FormStore.js
@@ -117,6 +117,10 @@ export default class FormStore {
         this.resourceStore.set(name, value);
     }
 
+    change(name: string, value: mixed) {
+        this.resourceStore.change(name, value);
+    }
+
     @computed get locale(): ?IObservableValue<string> {
         return this.resourceStore.locale;
     }
@@ -129,6 +133,6 @@ export default class FormStore {
         }
 
         this.type = type;
-        this.resourceStore.set(TYPE, type);
+        this.set(TYPE, type);
     }
 }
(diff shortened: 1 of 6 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/FormStore.js
```diff
@@ -129,6 +133,6 @@ export default class FormStore {
         }
 
         this.type = type;
-        this.resourceStore.set(TYPE, type);
+        this.set(TYPE, type);
```
danrot: This means changing the template also doesn't trigger the save button, right? I don't think that this is the desired behavior... This should be only like that initially. Maybe also differ between `setType` and `changeType`?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/FormStore.js
```diff
@@ -117,18 +117,31 @@ export default class FormStore {
...
+        this.saveType(type);
+        this.set(TYPE, type);
+    }
+
+    changeType(type: string) {
+        this.saveType(type);
+        this.change(TYPE, type);
+    }
+
+    @action saveType(type: string) {
```
danrot: `saveType` is a bit irritating, because it sounds like it is saved on the server, although it isn't... Can't we come up with a better name? Otherwise I could imagine having the check below for the type in a `hasTypes` function and reuse that in the `setType` and `changeType` function, although I can see that this is duplicating some code :confused:

danrot: Or what about `setInternalType`?

trickreich:
@danrot Is solved

PR #3794, opened 2018-02-22
Replace Country entity with Symfony Intl Regionbundle

### LINKED ISSUE #758: Get rid of country table and use symfony regionbundle
RegionBundle from Symfony Component have all we need for countries. We should just store the countryShort ('at') in the contact and use e.g.: \Symfony\Component\Intl\Intl::getRegionBundle()->getCountryNames() to get Countries it also returns translated names.

### LINKED ISSUE #1932: Country Bouvet Island and Heard Island and McDonald Islands is not a valid ISO Country
Bouvet Island and Heard Island and McDonald Islands does not exist and have no ISO specification?
Its not part of the Symfony:Intl country list.

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | yes
| Fixed issues | fixes #758, fixes #1932
| Deprecations? | no
| License | MIT

#### What's in this PR?

Replace the country entity and table with the Symfony Intl Regionbundle.

#### Why?

The Symfony Intl Regionbundle provide an easy way to get translated countries over the intl php extension.

#### Example Usage

~~~php
$address->getCountryName('de');
~~~

#### BC Breaks/Deprecations

Country entity is not longer available and need to be migrated to a country string field.


src/Sulu/Bundle/ContactBundle/Entity/Address.php
```diff
@@ -341,11 +342,11 @@ public function getAddressType()
     /**
      * Set country.
      *
-     * @param \Sulu\Bundle\ContactBundle\Entity\Country $country
+     * @param string $country
      *
      * @return Address
      */
-    public function setCountry(\Sulu\Bundle\ContactBundle\Entity\Country $country = null)
+    public function setCountry($country)
```
danrot: You know that we could use here (and in a lot other places probably) scalar type hints now?

src/Sulu/Bundle/ContactBundle/Entity/Address.php
```diff
@@ -355,13 +356,29 @@ public function setCountry(\Sulu\Bundle\ContactBundle\Entity\Country $country =
...
+     *
+     * @return string|null
+     */
+    public function getCountryName($displayLocale = null)
+    {
+        if (!$this->country) {
+            return null;
+        }
+
+        return Intl::getRegionBundle()->getCountryName($this->country, $displayLocale);
```
danrot: Why do we have a `CountryRepository`, and only use it sometimes? I am totally ok with the line here, but then I am wondering why we are not accessing it this way everywhere.

src/Sulu/Bundle/ContactBundle/Entity/CountryRepository.php
```diff
@@ -0,0 +1,48 @@
...
+class CountryRepository
+{
+    public function findAll(string $displayLocale = null)
+    {
+        $countries = [];
+
+        foreach (Intl::getRegionBundle()->getCountryNames($displayLocale) as $code => $countryName) {
+            $countries[$code] = [
+                'code' => $code,
+                'name' => $countryName,
```
danrot: I am not a big fan of passing untyped associative arrays around :confused: Do we need this repository at all? Another way to go about this is to simply use the `Intl` stuff directly. Otherwise I would also see value in introducing some kind of `Country` model object (not an entity!), which is not saved in the DB.

src/Sulu/Bundle/ContactBundle/Entity/CountryRepository.php
```diff
@@ -0,0 +1,48 @@
...
+        $countryName = Intl::getRegionBundle()->getCountryName($code, $displayLocale);
+
+        if (!$countryName) {
+            return null;
+        }
+
+        return [
+            'code' => $code,
+            'name' => $countryName,
+        ];
```
danrot: Same as above

src/Sulu/Bundle/ContactBundle/Entity/Address.php
alexander-schranz: I did keep the CountryRepository to keep the old UI running. I think in this case it would be better to deprecate the CountryRepository and remove it with 2.0?

src/Sulu/Bundle/ContactBundle/Entity/CountryRepository.php
alexander-schranz: As said above the repository was more to keep the old UI running. Will deprecate it if its ok?

alexander-schranz:
@danrot rebased.

alexander-schranz:
@danrot rebased 😄

PR #3795, opened 2018-02-22
Load page content in Single internal link content type

### LINKED ISSUE #545: `single_page_selection` content type returns only uuid

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | yes
| Deprecations? | yes
| Fixed tickets | fixes #545
| License | MIT
| Documentation PR | https://github.com/sulu/sulu-docs/pull/363

#### What's in this PR?

The single internal link content type return the page content instead of a uuid.
Also the content mapper load function does not longer need the webspaceKey as document manager does not need a webspace key.

#### Why?

Currently it does just a uuid is returned instead of page content. The single internal link now will return the whole page so it avoid to call `sulu_content_load`. 

#### BC Breaks/Deprecations

Removed the webspaceKey from Contentmapper::load function as not longer needed to load a page.

#### To Do

### DIFF
--- a/src/Sulu/Bundle/ContentBundle/Content/Types/SingleInternalLink.php
+++ b/src/Sulu/Bundle/ContentBundle/Content/Types/SingleInternalLink.php
@@ -13,9 +13,12 @@ namespace Sulu\Bundle\ContentBundle\Content\Types;
 
 use PHPCR\NodeInterface;
 use Sulu\Bundle\WebsiteBundle\ReferenceStore\ReferenceStoreInterface;
+use Sulu\Bundle\WebsiteBundle\Resolver\StructureResolverInterface;
 use Sulu\Component\Content\Compat\PropertyInterface;
+use Sulu\Component\Content\Mapper\ContentMapperInterface;
 use Sulu\Component\Content\PreResolvableContentTypeInterface;
 use Sulu\Component\Content\SimpleContentType;
+use Sulu\Component\DocumentManager\Exception\DocumentNotFoundException;
 
 /**
  * ContentType for SingleInternalLink.
@@ -23,9 +26,14 @@ use Sulu\Component\Content\SimpleContentType;
 class SingleInternalLink extends SimpleContentType implements PreResolvableContentTypeInterface
 {
     /**
-     * @var string
+     * @var ContentMapperInterface
      */
-    private $template;
+    private $contentMapper;
+
+    /**
+     * @var StructureResolverInterface
+     */
+    private $structureResolver;
 
     /**
      * @var ReferenceStoreInterface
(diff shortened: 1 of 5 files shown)


wachterjohannes:
as discussed in this issue https://github.com/sulu/sulu/issues/545 you should extract the view data of resolved structure in the `getViewData` and the content in `getContentData` whats missing then are all of the "root" properties of the resolved structure.

src/Sulu/Bundle/ContentBundle/Content/Types/SingleInternalLink.php
```diff
@@ -92,6 +111,34 @@ public function getTemplate()
...
+    public function getContentData(PropertyInterface $property)
+    {
+        $data = $this->loadStructure($property->getValue(), $property->getStructure()->getLanguageCode());
+
+        if (isset($data['content'])) {
+            $content = $data['content'];
+            unset($data['view']);
+            unset($data['content']);
+
+            return array_merge($data, $content);
```
alexander-schranz: did merge the root data so `uuid`, `created`, `authored`, ... are also available in the content type.

wachterjohannes: @sulu/core-team what do you think? in my opinion we should merge this data into view and not into content

alexander-schranz: @wachterjohannes in the navigation this data is also available at the same level as the content.

wachterjohannes: that's right - but in the content variable, there are only `properties` defined in XML ... UUIDs and other configurations (see smart_content) are in the view variable

alexander-schranz: hmm ok but seems not be the same as for `internal_links` there it looks like the same as for the navigation.

danrot: I am not sure what the result of this statement is, but if there is a `single_internal_link` property on the root level of a template I'd expect to access the `title` of the link in the following way in the twig template:

``` jinja
{content.single_internal_link.title}
```

And the view should only hold the `uuid` of the selected page, since the data that is written in PHPCR should be exposed there.

Resp. it really looks weird to have the `view` data in here and in the `view` twig variable (that's what's happening here, right?)

alexander-schranz: @danrot content does not include the view data. The current data in the template twig file would look like this.

```yml
content:
     title: ...
     single_internal_link:
          # root data
          uuid: ...
          published: ...
          created: ...
          # content
          title: Test
          url: /test
          images: 
                - ...
          ...
view:
     title: ...
     single_internal_link:
          # view
          title: []
          url: []
          images:
              displayOption: left
          ...
```

so in twig the title is accessable as `{{ content.single_internal_link.title }}`

danrot: @sulu/core-team I think this is the only blocker here, right? I'd say we should decide in a team review what should be happening here, and then merge it. For me the resulting structure as @alexander-schranz has sketched it in the previous comment look good, if everybody is ok with that we could merge this after a rebase.

alexander-schranz: /cc @wachterjohannes @chirimoya

alexander-schranz:
This will currently not be implemented. We should maybe implement alternative twig extension instead see #4576

PR #3801, opened 2018-02-27
fix superfluous updates on binded observable on the router

danrot:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR fixes a bug, which caused some wrong update on render. The problem was that the `unbind` calls in `componentWillUnmount` were called after the `bind` calls of `componentWillMount`. That made the observables be updated on the `bind` calls and already called all reactions.

#### Why?

Because in quite a lot situation reactions were run, where they shouldn't. E.g. switching from the Snippet List to e.g. Contacts led to two requests, whereby one was sent to the "old" snippet endpoint with a locale of `undefined`. That should definitely not happen, and is fixed in this PR.

#### BC Breaks/Deprecations

The `unbind` method of the `Router` was removed, because all bindings are unbinded on the `navigate` call (which might be another BC break).

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/services/Router/README.md
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/services/Router/README.md
@@ -114,3 +114,6 @@ value.set('anything'); // will navigate to /page?value=anything
 
 router.unbind('value', value); // unbind to avoid leaking listeners
 ```
+
+Mind that the bindings will be cleared on every `navigate` call. This is necessary to avoid superfluous updates because
+of wrong update on observers.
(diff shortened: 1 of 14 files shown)


chirimoya:
@danrot rebase

danrot:
Is rebased. Do you remember what exactly was not working before the other PR? :see_no_evil:

chirimoya:
@danrot the `MediaOverview` ;-)

PR #3802, opened 2018-02-27
Fix xml loader internal flag

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Fix xml loader internal flag. 

#### Why?

Was false resolved and the new content type check was needed to be added.

#### BC Breaks/Deprecations

#### To Do

### DIFF
--- a/src/Sulu/Component/Content/Metadata/Loader/XmlLoader.php
+++ b/src/Sulu/Component/Content/Metadata/Loader/XmlLoader.php
@@ -52,7 +52,7 @@ class XmlLoader extends XmlLegacyLoader
         $structure->name = $data['key'];
         $structure->cacheLifetime = $data['cacheLifetime'];
         $structure->controller = $data['controller'];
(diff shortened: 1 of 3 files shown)


src/Sulu/Component/Content/Metadata/Loader/XmlLoader.php
```diff
@@ -88,7 +88,7 @@ private function createProperty($propertyName, $propertyData)
                 throw new \InvalidArgumentException(sprintf(
                     'Content type with alias "%s" has not been registered. Known content types are: "%s"',
                     $propertyData['type'],
-                    implode('", "', array_keys($this->contentTypeManager->getAll()))
+                    implode('", "', array_keys($this->contentTypeManager->getAll() ?: []))
```
alexander-schranz: @danrot did change this to have a better exception message in the test.

PR #3807, opened 2018-02-27
Remove recreation of MediaStores in MediaOverview and MediaSelection

### LINKED ISSUE #3743: Change folder in the MediaSelectionOverlay does not change sub folder if a media is selected
| Q | A
| --- | ---
| Bug? | yes
| New Feature? | no
| Sulu Version | d80dd7a222495eae8a25d96d7ed9c76b1fa22503
| Browser Version | all browsers

#### Actual Behavior

When you open the `MediaSelection` Overlay in a form and select some media, you can't change the collection afterwards. It will update the overlay except for the datagrid containing all the collections. This makes it impossible to select some sub collections. 

#### Expected Behavior

The collection should also be update, even when some media are already selected.

#### Steps to Reproduce

1. Go to a snippet with a `media_selection` field
2. Click on the add button of the `media_selection`
3. Select one of the appearing media from the lower datagrid
4. Click on a collection in the upper datagrid
5. See that everything changed accordingly, except for the upper datagrid, which still shows the sub collections of the previous active collection

#### Possible Solutions

The `MediaSelectionOverlay` has a [`createStores` function], which recreates the stores everytime the current collection changes. To remember the selection when changing the collection there is [an observer](https://github.com/sulu/sulu/blob/2224b374fa252b50b4295046db7da4ecdee2cb6f/src/Sulu/Bundle/MediaBundle/Resources/js/containers/MediaSelection/MediaSelectionOverlay.js#L181), which additionally keeps track of them.

Instead of this solution the current collection should act more like the `page` observable in the `Datagrid`, i.e. filtering the results of the datagrids. That should get rid of this weird behavior as well.

danrot:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #3743
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR removes the permanent recreation of the stores in the `MediaOverview` and `MediaSelection` components.

#### Why?

Because we have fixed our routing issue in #3801, but the fix does break the weird behavior with the recreation of these components.

#### ToDo

- [ ] How to show loader for collection datagrid? It changed a bit.
- [x] Fix reset to first page for `MediaDatagridStore` when collection is changed

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Controller/AdminController.php
+++ b/src/Sulu/Bundle/AdminBundle/Controller/AdminController.php
@@ -472,7 +472,8 @@ EOL
             "label": "Description",
             "type": "text_area"
         }
-    }
+    },
+    "schema": {}
 }
 EOL
             );
@@ -507,7 +508,8 @@ EOL
                 }
             }
         }
-    }
+    },
+    "schema": {}
 }
 EOL
             );
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Datagrid/stores/DatagridStore.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Datagrid/stores/DatagridStore.js
@@ -99,10 +99,9 @@ export default class DatagridStore {
         }
 
         const observableOptions = {};
-        observableOptions.page = this.observableOptions.page.get();
 
-        if (this.observableOptions.locale) {
-            observableOptions.locale = this.observableOptions.locale.get();
+        for (const key in this.observableOptions) {
+            observableOptions[key] = this.observableOptions[key].get();
         }
 
         this.setDataLoading(true);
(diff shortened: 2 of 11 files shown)


danrot:
The behavior of the loaders changed a bit... It was not really nice before that, and it is still not nice, don't really know what to do about it :see_no_evil:

danrot:
I actually thought I have already fixed the `11` issue and I also can't reproduce it anymore. It should actually be fixed by [replacing the string with a number at initialization](https://github.com/sulu/sulu/pull/3807/files#diff-c1244ad74ef56debe784c1b1884cbbd6R32).

And the other thing was also not working previously... The page always is and was reset. The problem here is that our mechanism fails a bit... E.g. if you go to page 2 on the root, go a level deeper and stay on page 1, and go another level deeper the last saved page is 1. If you click on "All Media" afterwards you would be on page 1 instead of the expected 2 if we use the `restore` functionality... I think this needs a bit more time to be fixed.

chirimoya:
@danrot please add an issue for the wrong pagination behaviour while moving through the collection tree

danrot:
Created #3818

PR #3810, opened 2018-02-28
Fixed rename and publish exception if page has link to a child

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | none
| Related issues/PRs | https://github.com/sulu/sulu-document-manager/pull/117  https://github.com/sulu/sulu-document-manager/pull/118 https://github.com/sulu/sulu-document-manager/pull/119
| License | MIT
| Documentation PR | none

#### What's in this PR?

This PR add a flush to node-controller between `persist` and `publish` call.

#### Why?

This is needed because of following scenario:

1. Imagine following content tree:

* Page 1
  * Page 1.1

2. Page one has an internal link (content-type) to page 1.1
3. Rename page 1 and click save and publish

The publish fails because the renaming of the node won't be synced to the repository and the following exception will be thrown.

```
Path not found (parent node /cmf/example/contents/page-1 moved in current session): /cmf/example/contents/page-1/page-1-1
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,7 @@
         "imagine/imagine": "~0.6.1 || ~0.7.0",
         "massive/search-bundle": "0.16.*",
         "sensio/framework-extra-bundle": "^3.0",
-        "sulu/document-manager": "~0.10.0",
+        "sulu/document-manager": "dev-master as 0.10.1",
         "stof/doctrine-extensions-bundle": "^1.2.2",
         "gedmo/doctrine-extensions": "~2.4",
         "symfony-cmf/routing-bundle": "^1.2 || ^2.0",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ CHANGELOG for Sulu
 
 * dev-master
     * BUGFIX      #3806 [All]                     Fix compatibility on lowest and fix appveyor
+    * HOTFIX      #3810 [ContentBundle]           Fixed rename and publish exception if page has link to a child
     * BUGFIX      #3805 [ContentBundle]           Fix spacing between rows and section in content template generation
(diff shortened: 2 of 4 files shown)


danrot:
Are you sure this has no uninteded side effects? I've got a bad feeling about this... Isn't there a cleaner way to fix that?

wachterjohannes:
the unintendet side effects happens without that ... this "simulates" the standard multi-request approach (first save draft and the publish it)

danrot:
Well, I've got two issues with that:

1. Just because of this line we are doing 12 requests more
2. It might lead to strange UX. When only the publish fails the changes are save partly, although an error has been displayed.

Apart from that it does not solve the actual problem. If somebody is using the document manager on his own, he'll still run into that problem. I think a better way to solve this would be to delay the actual renaming from `persist` to `flush`. Then the current code would still work, the performance would not get worse, and the "obvious" thing would also work in case somebody uses that on his own and if we have another feature relying on that it would also not fail.

wachterjohannes:
@danrot this seems to work - i have added another PR for document-manager

composer.json
```diff
@@ -28,7 +28,7 @@
         "imagine/imagine": "~0.6.1 || ~0.7.0",
         "massive/search-bundle": "0.16.*",
         "sensio/framework-extra-bundle": "^3.0",
-        "sulu/document-manager": "~0.10.0",
+        "sulu/document-manager": "dev-hotfix/auto-rename-path-node-name as 0.10.1",
```
wachterjohannes: will remove this when the tests are green

danrot:
Is this really not working with Jackrabbit, or are only the tests failing for some reason?

PR #3812, opened 2018-03-01
Fix polyfill requirements (based on Symfony 3.3)

wimvds:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no?
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

The symfony/polyfill dependencies are fixed, so we can integrate Sulu with Sylius 1.1 again.

#### Why?

We were unable to combine Sylius & Sulu in projects due to dependency hell because of the symfony/polyfill requirement.

https://github.com/symfony/polyfill#usage : "When using composer you should not require the symfony/polyfill package, but the standalone ones." 

I just copied over the standalone Symfony (3.3) polyfill package requirements and it solved our dependency issues.

I don't think this will cause BC breaks ...


composer.json
```diff
@@ -59,7 +59,11 @@
         "symfony/http-kernel": "^3.0",
         "symfony/intl": "^3.0",
         "symfony/options-resolver": "^3.0",
-        "symfony/polyfill": "^1.6",
+        "symfony/polyfill-apcu": "~1.1",
+        "symfony/polyfill-intl-icu": "~1.0",
+        "symfony/polyfill-mbstring": "~1.0",
+        "symfony/polyfill-php56": "~1.0",
```
alexander-schranz: is this needed when we support only 7.1? /cc @wachterjohannes

composer.json
```diff
@@ -59,7 +59,11 @@
...
+        "symfony/polyfill-php70": "~1.0",
```
alexander-schranz: same as above

composer.json
wachterjohannes: i think the only thing which is realy needed is `symfony/polyfill-mbstring` and `symfony/polyfill-apcu` (but only in `sulu-minimal` and `sulu-standard`

wachterjohannes: @wimvds what would you like to achieve with that? had you any problem without that?

alexander-schranz:
at current state I think we only need: `"symfony/polyfill-mbstring"`, `"symfony/polyfill-xml"` /cc @sulu/core-team

composer.json
wimvds: No, I just copied the default polyfills from a new Symfony 3.3 project. I don't need them & I can drop them if you want.

wimvds:
```symfony/polyfill-xml``` seems to be deprecated, should I add ```symfony/polyfill-php72``` instead (cfr. https://github.com/symfony/polyfill-xml)?

composer.json
alexander-schranz: @wimvds ok you can drop them and only add mbstring and xml as polyfill.

alexander-schranz:
@wimvds sounds good `symfony/polyfill-php72` and `symfony/polyfill-mbstring`.

wimvds:
@alexander-schranz the polyfills that are not needed have been removed.

alexander-schranz:
Thank you @wimvds! We will wait for the Travis CI.

PR #3816, opened 2018-03-01
Use circle ci to test if grunt build was run

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Use circle ci to test if grunt build was run for the bundles.

#### Why?

So no PR is merged where the build is missing.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,14 @@
+version: 2
+
+jobs:
+  build:
+    docker:
+      - image: circleci/php:7.1-node-browsers
+    steps:
+      - checkout
+      # Node
+      - run: sudo apt-get install ruby ruby-all-dev
+      - run: sudo gem update --system
+      - run: sudo gem install compass
+      - run: bin/build-bundles.sh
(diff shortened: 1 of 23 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,26 @@
+version: 2
+
+jobs:
+  build:
+    docker:
+      - image: circleci/php:7.1-node-browsers
+    steps:
+      - checkout
+# Composer (currently not needed)
```
danrot: Please remove if we don't need that.

bin/ci/validate-build-bundles.sh
```diff
@@ -0,0 +1,8 @@
+#! /bin/bash
+
+changes=$(git status -s | grep dist -c)
+expected=12
+exitCode=$(expr $changes - $expected)
+echo $exitCode
+exit $exitCode
+
```
danrot: No superfluous new line

PR #3825, opened 2018-03-06
Field Input

trickreich:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes --
| Related issues/PRs | --
| License | MIT
| Documentation PR | --

#### What's in this PR?

Added field `Input`.

#### Why?

`FieldTypeProps` should not be used in the component itself.


src/Sulu/Bundle/AdminBundle/Resources/js/components/Input/Input.js
```diff
@@ -4,24 +4,27 @@ import type {ElementRef} from 'react';
...
     icon?: string,
     type: string,
     loading?: boolean,
     placeholder?: string,
     inputRef?: (ref: ?ElementRef<'label'>) => void,
-    onFocus?: () => void,
+    valid: boolean,
+    value: ?string,
+    onBlur?: () => void,
+    onChange: (value?: string) => void,
```
danrot: That already had some quite strange errors earlier... I think we should write `value: ?string` instead.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/tests/fields/Input.test.js
```diff
@@ -0,0 +1,30 @@
...
+            onChange={jest.fn()}
+            onFinish={jest.fn()}
+            value={'xyz'}
+            error={error}
+        />
+    );
+
+    expect(inputInvalid.find(InputComponent).prop('valid')).toBe(false);
+
+    const inputValid = shallow(
```
danrot: I would put everything starting from here into a separate test. Smaller tests are easier to maintain when refactoring, and there is no dependency on the first part of the test.

danrot:
The `Input` component used in the `ResourceLocator` have not been correctly updated.

src/Sulu/Bundle/AdminBundle/Resources/js/components/Input/Input.js
```diff
@@ -35,16 +38,16 @@ export default class Input extends React.PureComponent<Props> {
     };
 
     handleBlur = () => {
-        const {onFinish} = this.props;
+        const {onBlur} = this.props;
 
-        if (onFinish) {
-            onFinish();
+        if (onBlur) {
```
danrot: @chirimoya WDYT about this? We need the `Input` component in a non-form environment, therefore we decided to remove the `FieldTypeProps` from this component and introduce a separate field (this is a good idea IMO). What's bothering me a little bit is the renaming of `onFinish` to `onBlur` here. For me it makes totally sense, because the `onBlur` is what you are also used to in HTML. But where do we stop then? Should the `AutoComplete` and `ResourceLocator` component also have an `onBlur` instead of an `onFinish` then?

chirimoya: +1 `onBlur` for all components which are close to an `Input` implementation

danrot: @trickreich Talked to @chirimoya, I'd say we change it also to `onBlur` in the `ResourceLocator` and `AutoComplete` in this PR.

At some point we can also try to do that for all the other components currently implementing the `FieldTypeProps` type (after a short check I've found `FieldBlocks` and `MediaSelection`) and move the `FieldTypeProps` to the form, but I think that is not necessary for this PR.

PR #3829, opened 2018-03-08
Toolbar: Disable dropdown when all options are disabled

trickreich:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes --
| Related issues/PRs | --
| License | MIT
| Documentation PR | --

#### What's in this PR?

Iterate over options to check if all options are disabled.
When yes, disable also the button itself.

#### Why?

When no active options is available, button is still clickable at the moment.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Toolbar/tests/Dropdown.test.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Toolbar/tests/Dropdown.test.js
@@ -96,8 +96,13 @@ test('Click on disabled option will not fire onClick', () => {
         options: [
             {
                 label: 'An option',
-                onClick: clickSpy,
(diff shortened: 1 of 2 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/components/Toolbar/tests/Dropdown.test.js
```diff
@@ -96,16 +96,48 @@ test('Click on disabled option will not fire onClick', () => {
         options: [
             {
                 label: 'An option',
-                onClick: clickSpy,
+                onClick: jest.fn(),
```
danrot: Shouldn't the `clickSpy` be here? Otherwise the check does not work correctly...

PR #3835, opened 2018-03-09
Added bundle name to config response

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | none
| Related issues/PRs | none
| License | MIT
| Documentation PR | none

#### What's in this PR?

This PR adds the name of the bundle as sections to admin config.

#### Why?

This will be needed to add other bundle configuration.


src/Sulu/Bundle/AdminBundle/Resources/js/index.js
```diff
@@ -65,3 +68,5 @@ Promise.all([
 ]).then(startApplication);
 
 bundleReady();
+
+export {configPromise};
```
danrot: I think I would call `bundleReady` after the export... Just to make sure everything has been called before the `bundleReady` call.

wachterjohannes:
@danrot should be ready

PR #3845, opened 2018-03-15
Improving styleguide build performance

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | https://github.com/styleguidist/react-styleguidist/issues/765
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR tries to remove some performance overhead by moving the `glob` for all the styleguide components to be exectued only once.

#### Why?

Because the `watch` task doesn't work when there is a `vendor` directory in sulu/sulu.

#### To Do

- [ ] Make the `styleguide` task work even when the `vendor` folder is installed.


danrot:
For some reason the watch task is now also working with a huge `vendor` folder without this patch, so I am going to close it :-)

PR #3850, opened 2018-03-16
Allow user to be null for security expressions for debug

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

It check if the tokenstorage can provide a user and don't throw an error then.

#### Why?

Some commands like debug:container, debug:router didn't work on specific routes or services.

Example:

```
bin/adminconsole debug:router sulu_admin_v2.resources
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@ CHANGELOG for Sulu
 ==================
 
 * dev-release/1.5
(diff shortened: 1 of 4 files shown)


alexander-schranz:
@danrot tests for SecurityConfig added.

PR #3852, opened 2018-03-16
Add category to admin v2 as resource.

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Add category to admin navigation and as resource

#### Why?

We need categories for v2.


danrot:
@alexander-schranz Sry to say, but most of these changes are not valid anymore... E.g. the endpoint is now delivered by the server instead of hardcoded.

danrot:
Will be replaced by #4002, and I move over what I need from here step by step.

PR #3855, opened 2018-03-19
Add FieldType for internal links

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #---
| Related issues/PRs | #---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR will add a new field type for the `internal_links` content type.

#### Why?

Because this field type has also to work in the new UI.

#### To Do

- [x] Reopen overlay fails because of active item (what should be open when reopening in general?)
- [x] How to handle media selection overlay in regard of width?
- [x] Fix scrolling in overlay when x-axis can also be scrolled
- [ ] Error message for handling linking page to itself
- [x] Issue for different Homepage title (webspace name vs "Homepage")
- [x] Add translations
- [x] Adapt overlays as in design
- [x] Implement selectable in all adapters


src/Sulu/Bundle/AdminBundle/Resources/js/index.js
```diff
@@ -44,6 +44,7 @@ function registerFieldTypes() {
 
     // TODO move to correct bundle or even allow to register somehow via the config request
```
danrot: I think adding the assignments via the server config is also something we should do in the near future.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Assignment/datagridOverlay.scss
```diff
@@ -1,3 +1,16 @@
...
 .datagrid {
     padding: 30px 60px 60px;
 }
+
+.adapter-container-column_list {
+    height: calc(100vh - $overlayHeaderHeight - $overlayFooterHeight - $overlayMarginBottom);
+    overflow: hidden;
+}
+
+.adapter-column_list {
```
chirimoya: `.adapter.column_list`

PR #3857, opened 2018-03-20
Fix errors in styleguide

danrot:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #---
| Related issues/PRs | #---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

There were some errors in the styleguide, which had to be fixed.

#### Why?

Because it doesn't look professional if there are loads of errors in the logs :-)


danrot:
The only thing still showing warnings are the translations. That's something else we should probably fix in the future. In the file where we will register the translations then, we could also set MobX to strict mode.

PR #3876, opened 2018-03-28
Load data with template

### LINKED ISSUE #3853: Field value will be overwritten when changing the template
| Q | A
| --- | ---
| Bug? | yes
| New Feature? | no
| Sulu Version | 1.6.15
| Browser Version | irelevant

#### Actual Behavior

When the attribute multilingual="false" of a property is set and you change the template of a language version the field value wont be loaded into the template and the save process will overwrite the value with an empty one, because it is shared over all languages.

#### Expected Behavior

The value f.e. image should be loaded when changing the template. 

#### Steps to Reproduce

Create a page with a template and create some translations. Change one language to a template which includes a property with the attribute multilingual="false", input some value and save the page. Go to another language change the templates to the one from before, the non multilingual content wont be loaded, save the page and the content is gone.

#### Possible Solutions

Load the value when changing the template or stop overwriting when the template changes and the field value is empty

trickreich:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #3853 
| Related issues/PRs | --
| License | MIT
| Documentation PR | --

#### What's in this PR?

Reload data from server when changing the template in page edit.


src/Sulu/Component/Content/Document/Subscriber/StructureSubscriber.php
```diff
@@ -186,6 +187,11 @@ public function handleHydrate(AbstractMappingEvent $event)
             $structureType = $this->getDefaultStructureType($document);
         }
 
+        $template = $event->getOption('template');
```
wachterjohannes: in document-manager i would call it `structureType` (in controller it is ok)

src/Sulu/Bundle/ContentBundle/Tests/Functional/Controller/NodeControllerTest.php
```diff
@@ -276,6 +276,55 @@ public function testGet()
...
+        $this->documentManager->flush();
+
+        $client = $this->createAuthenticatedClient();
+
+        $client->request('GET', '/api/nodes/' . $document->getUuid() . '?language=en');
+        $response = json_decode($client->getResponse()->getContent(), true);
+
+        $this->assertArrayNotHasKey('article', $response);
+
+        $client->request('GET', '/api/nodes/' . $document->getUuid() . '?language=en&template=default');
```
wachterjohannes: why no using another template? `default` is the template which is used before to create the page

PR #3879, opened 2018-03-29
Disable items by ids in datagrid

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | #3855
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR allows to disable certain items in the datagrid by passing the ids to disable in the `disabledIds` prop of the Datagrid.

#### Why?

Because in #3855 we introduced the internal links field type, but you can still select the page you are currently editing to itself, which the backend does not allow.

#### Example Usage

~~~jsx
<Datagrid disabledIds={[1, 4]} />
~~~

#### To Do

- [ ] Add disabledIds as property in the Assignment field type to avoid linking a page to itself


danrot:
Only the last two commit 2610ef8 and bc27792 are really in this PR, the rest comes from #3855.

PR #3881, opened 2018-03-29
Adjust FieldTypeProps to allow inspecting the form

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This PR will change the interface for FieldTypes.

#### Why?

Because some fields will need the possibility to access data from the form. This includes the following field types:

- ResourceLocator (Access other field by a tag)
- Position in the contact form (Access other field by path)
- Internal Link (resourceKey and id of the form to disable item in overlay)

This functionality should be added to a newly created `FormInspector`.

#### BC Breaks/Deprecations

The `FieldTypeProps` changed, which will break existing field types.

#### To Do

- [x] Implement `FormInspector` which only allows read access to `FormStore`.
- [x] Change existing field types


danrot:
Currently failing because of this regression: https://github.com/doctrine/dbal/issues/3082

danrot:
@chirimoya Found a completely different solution, which also fixes the issue :-)

PR #3893, opened 2018-04-05
Datagrid fields

trickreich:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes --
| Related issues/PRs | --
| License | MIT
| Documentation PR | --

#### What's in this PR?

Datagrid display fields correctly.

#### Why?

Otherwise it's not that cool..

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/ResourceMetadata/Datagrid/Field.php
+++ b/src/Sulu/Bundle/AdminBundle/ResourceMetadata/Datagrid/Field.php
@@ -28,6 +28,26 @@ class Field
      */
     protected $type;
 
+    /**
+     * @var bool
+     */
+    protected $disabled;
+
+    /**
+     * @var bool
+     */
+    protected $default;
+
+    /**
+     * @var bool
+     */
+    protected $sortable;
+
+    /**
+     * @var bool
+     */
+    protected $editable;
+
     public function __construct(string $name)
     {
         $this->name = $name;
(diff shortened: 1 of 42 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/containers/Datagrid/types.js
```diff
@@ -60,3 +66,7 @@ export type TreeItem = {
     data: DataItem,
     children: Array<TreeItem>,
 };
+
+export interface FieldTransformer {
+    transform(value: *): *,
```
danrot: Are there cases where this does not return a string?

danrot: Or to be more precise, that must be something we can use in JSX, right? So maybe use the `Node` type from react?

src/Sulu/Bundle/AdminBundle/Resources/js/index.js
```diff
@@ -63,8 +85,7 @@ function startApplication() {
     render(<Application router={router} />, applicationElement);
 }
 
-const translationPromise = Requester.get('/admin/v2/translations?locale=en')
-    .then((response) => setTranslations(response));
+const translationPromise = Requester.get('/admin/v2/translations?locale=en');
 
 const configPromise = Requester.get('/admin/v2/config').then((response) => {
```
danrot: To avoid more strange behavior as before I would say we do everything in this callback in the one for `Promise.all`.

src/Sulu/Component/Rest/ListBuilder/FieldDescriptorInterface.php
```diff
@@ -76,6 +76,11 @@ public function getSortable();
      */
     public function getEditable();
 
+    /**
+     * @return string
```
danrot: Could be added as a return type instead, right?

src/Sulu/Bundle/AdminBundle/ResourceMetadata/Datagrid/Field.php
```diff
@@ -28,6 +28,21 @@ class Field
...
 
+    /**
+     * @var string
+     */
+    protected $visibility;
+
+    /**
+     * @var bool
+     */
+    protected $sortable;
```
danrot: I think I would also remove that as long as it is not implemented... And btw, implementing it wouldn't be too much work I guess, right? Maybe that's also something we can get done :smiley:

danrot:
Changes look good to me, can you rebase, so that I can test once more?

PR #3894, opened 2018-04-06
Move assignment configuration to AdminController

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR moves the configuration of the assignment field types to the `AdminController`.

#### Why?

Because this would be a great extension point for external bundles. If an own entity will be defined a simple configuration in the backend would be enough to also have an assignment field type without any javascript.


.circleci/config.yml
```diff
@@ -7,15 +7,7 @@ jobs:
...
       # Node
-      - restore_cache:
-          keys:
-            - node-v1-{{ checksum "package.json" }}
-            - node-v1-
       - run: npm install
-      - save_cache:
-          key: node-v1-{{ checksum "package.json" }}
-          paths:
-            - node_modules
```
danrot: That change is something we probably should discuss. It seems like some dependency was updated (couldn't quickly find out which one), which breaks the tests now, because the ordering of attributes in snapshots change. However, this error didn't happen on circle, only on my locale machine. When we remove that cache it also happens on circle.

The difference for the install seems to be 25 seconds (can be compared [here](https://circleci.com/gh/sulu/sulu/870) and [here](https://circleci.com/gh/sulu/sulu/871))

src/Sulu/Bundle/AdminBundle/Resources/js/index.js
```diff
@@ -35,36 +35,18 @@ datagridAdapterRegistry.add('column_list', ColumnListAdapter);
...
-        adapter: 'column_list',
-        displayProperties: [
-            'title',
-        ],
-        icon: 'su-document',
-        label: translate('sulu_content.assignment_label'),
-        resourceKey: 'pages',
-        overlayTitle: translate('sulu_content.assignment_overlay_title'),
-    });
+    const assignmentConfigs = fieldTypesConfig['assignment'];
```
chirimoya: check if `fieldTypesConfig` has assignments

chirimoya:
else ACK

Issue #424, opened 2014-11-04
Load data for select FieldType using a Symfony service


sulu-io-user:
<a href="https://github.com/wachterjohannes"><img height="50" width="50" src="https://avatars.githubusercontent.com/u/1464615?v=2" /></a>
By @wachterjohannes _from_ https://github.com/sulu-cmf/SuluContentBundle/issues/222

Configurable over:
- service with interface (called from twig with webspace and locale)
- nesting params (property params)

danrot:
Already implemented checkbox and radio groups, which can simply be filled by data in the xml template. Still missing:
* selects
* load data from services

tniebergall:
Are there any updates for this content type?

wachterjohannes:
@tniebergall TBH not really :) there is a part implementation for checkbox and radios but for select there is currently no solution

danrot:
This has been implemented in the mean time :slightly_smiling_face:

PR #3898, opened 2018-04-10
Add user tab to contact

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

Add user tab to the contact form

#### Why?

Because we need a way to update the user for a contact.

#### To Do

- [x] `PropertiesXMLParser` should not fail if content type does not exist for field type
- [x] ~~~ContentType for PasswordConfirmation (?)~~~
- [x] ~~~Use anonymous class instead of ResourceMetadata, to make it cleaner to only support certain interfaces (also see https://markbakeruk.net/2016/05/12/anonymous-class-factory-the-results-are-in/, probably only create an issue)~~~
- [x] ~~~Add additional hook for validation of field types (it should not be possible to save with non-matching passwords)~~~ Created #3943
- [x] Remove password from user API
- [x] Password confirmation should only validate if both fields are filled
- [x] Switching between tabs reloads contact everytime
- [x] Load correct user based on API response from contact
- [x] Open user tab when no user exists yet
- [x] API should not return list for when loading via contact id
- [ ] Creating new user requires dropdown for system language (should still be loaded from service, see #424)
- [x] Fix entering of username in new user form
- [x] Add documentation (`idQueryParameter` in view and resourceStore, ...)


src/Sulu/Bundle/SecurityBundle/Admin/SecurityAdmin.php
```diff
@@ -110,6 +110,7 @@ public function getRoutes(): array
                 ->addOption('tabTitle', 'sulu_security.permissions')
                 ->addOption('backRoute', 'sulu_contact.contacts_datagrid')
                 ->addOption('resourceKey', 'users')
+                ->addOption('idQueryParameter', 'contactId')
```
danrot: I also don't know if it will be too many parameters here, if we also add another parameter for getting the parameter from the other direction (e.g. when loading the excerpt from an entity).

src/Sulu/Bundle/SecurityBundle/Controller/UserController.php
```diff
@@ -313,7 +313,18 @@ public function cgetAction(Request $request)
...
                 )->findUserByContact($contactId);
-                $view = $this->view($user ?? new \stdClass(), 200);
+
+                if (!$user) {
+                    $contact = $this->getDoctrine()->getRepository(
+                        $this->container->getParameter('sulu.model.contact.class')
+                    )->findById($contactId);
+
+                    $user = [
+                        'contact' => $contact,
```
danrot: That part is also a bit hard to solve... The thing is that we need to add a contact id to the response, in the following format, so that a part of the request has the following format:

```javascript
{
    contact: {
        id: 1
    },
    username: 'bla'
}
```

Since that is not easily possible in our generic form I have decided to also return that part in this response even if there is no user. Therefore I have to create an array faking this object, which makes it so weird.

In the [old UI it was simply reusing the already loaded contact for that](https://github.com/sulu/sulu/blob/abf10c8/src/Sulu/Bundle/SecurityBundle/Resources/public/js/components/users/main.js#L55). That's a viable solution, which would probably need us to make a custom view.

danrot: I've had an idea last night about this... The real problem is that we can't really predict what the static key `contact` should be in the JSON we must send, at least not without introducing another parameter which will make the component even harder to use. What if we make a slight change to the API instead? The problem is especially on the `POST` request, and we could probably also model it something like this:

```
POST /admin/api/users?contactId=5
```

Then we could say if the `Form` react component gets an `idQueryParameter` passed we simply add this to the request, and the backend can work with that if required.

@chirimoya WDYT?

PR #3901, opened 2018-04-11
Preview voter

silvano-a:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no (?)
| Deprecations? | no
| License | MIT

#### What's in this PR?

This PR adds a voter for the `PreviewKernel`, which allows all requests coming from the PreviewKernel

#### Why?

The preview uses no URI so the access control in `app/config/website/security.yml` wont work as expected. This could lead to access denied exceptions while the editor should be able to see the page he/she is trying to create / edit

### DIFF
--- a/src/Sulu/Bundle/PreviewBundle/Resources/config/config_preview.yml
+++ b/src/Sulu/Bundle/PreviewBundle/Resources/config/config_preview.yml
@@ -6,7 +6,7 @@ framework:
 
 services:
     sulu.preview.voter:
-        class: 'Sulu\Bundle\PreviewBundle\Security\PreviewVoter'
+        class: "Sulu\Bundle\PreviewBundle\Security\PreviewVoter"
         public: false
         tags:
-            - { name: security.voter }
+            - { name: "security.voter" }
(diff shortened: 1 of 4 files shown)


src/Sulu/Bundle/PreviewBundle/Resources/config/services.xml
```diff
@@ -60,6 +60,10 @@
 
             <tag name="sulu.websocket.message.handler" dispatcher="admin" alias="sulu_preview.preview"/>
         </service>
+
+        <service id="sulu_preview.preview.voter" class="Sulu\Bundle\PreviewBundle\Security\PreviewVoter" public="false">
```
wachterjohannes: this seems the wrong place :) you could simple register it in `preview_config`:

```
services:
    sulu_preview.preview.voter:
        class: Sulu\Bundle\PreviewBundle\Security\PreviewVoter
        public: false
        tag:
          - {name: "security.voter"}
```

then this service is only active in preview window

src/Sulu/Bundle/PreviewBundle/Resources/config/config_preview.yml
```diff
@@ -1,5 +1,15 @@
+parameters:
+    sulu.preview.voter_class: 'Sulu\Bundle\PreviewBundle\Security\PreviewVoter'
```
wachterjohannes: we does not use this parameter anymore (symfony has kicked out this parameters from best-practice)

wachterjohannes:
@silvano-a thanks for contribution :) this will be included in the next release 1.6.17

PR #3913, opened 2018-04-17
WIP: Fix cascading in media entities for removing multiple medias

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Related issues/PRs | https://github.com/sulu/SuluFormBundle/pull/158
| License | MIT

#### What's in this PR?

It add the cascade all to the media entities correctly.

Cascade-all does:
  - cascade-persist
  - cascade-remove
  - cascade-merge
  - cascade-detach

and this is mostly needed for all child elements of the media entity.

#### Why?

For example:

```php
$mediaManager->delete(1);
$mediaManager->delete(2);
```

Throws an error as only the Media entity is removed from memory and not the File and the File has still a relation to the Media. Else it throws 

`A new entity was found through the relationship 'Sulu\Bundle\MediaBundle\Entity\File#media' that was not configured to cascade persist operations for entity: . To solve this issue: Either explicitly call EntityManager#persist() on this unknown entity or configure cascade persist  this association in the mapping for example @ManyToOne(..,cascade={"persist"})`

#### Example Usage

```php
$mediaManager->delete(1);
$mediaManager->delete(2);
```

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/Resources/config/doctrine/BaseCollection.orm.xml
+++ b/src/Sulu/Bundle/MediaBundle/Resources/config/doctrine/BaseCollection.orm.xml
@@ -8,6 +8,7 @@
         <id name="id" type="integer" column="id">
             <generator strategy="AUTO"/>
         </id>
+
         <field name="style" type="string" column="style" length="255" nullable="true"/>
         <field name="lft" type="integer" column="lft">
             <gedmo:tree-left/>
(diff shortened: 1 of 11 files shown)


alexander-schranz:
As referenced in https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/reference/working-with-objects.html#removing-entities.

Cascade-Remove will create a DELETE request for every subentity this make it very slow for big entities like media.

alexander-schranz:
At current state it seems like more a doctrine issues as the following code fails:

```
        $this->em->remove($mediaEntity);
        $this->em->flush(); // this flush works
        $this->em->flush(); // this flush fails:
```

PR #3924, opened 2018-04-24
Set minimum stability of composer to dev

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Set minimum stability of composer to dev.

#### Why?

To get earlier problems with new versions of requirements.

#### Example Usage

~~~bash
composer update
~~~


alexander-schranz:
Following issues currently appear:

```
Interface 'FOS\HttpCache\SymfonyCache\CacheInvalidation' not found in /home/travis/build/sulu/sulu/src/Sulu/Bundle/HttpCacheBundle/Cache/AbstractHttpCache.php on line 27
```

```
Sulu\Bundle\SecurityBundle\Tests\Functional\Controller\ProfileControllerTest::testChangeLanguageAction
Error: Call to a member function getUser() on null
```

In:

 - AudienceTargetingBundle
  - HttpCacheBundle
  - SecurityBundle
  - WebsiteBundle

danrot:
What exactly do we gain by this? You say there are problems now, but we don't see them in the build. If somebody installs that on their own, and they forget the ´--prefer-stable` flag that means it will fail, which is not a very nice DX...

alexander-schranz:
Want to mention this PR again as we had in the last time many problems with breaking changes in symfony which we would get informed earlier if we test it on the newest develop version. /cc @sulu/core-team

PR #3927, opened 2018-04-25
Add upgrade for HttpCache class

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Add upgrade for httpcache class.

#### Why?

This class is common used for the `HEADER_REVERSE_PROXY_TTL` constant.

Before:

```php
HttpCache::HEADER_REVERSE_PROXY_TTL
```

after

```php
AbstractHttpCache::HEADER_REVERSE_PROXY_TTL
```

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -38,6 +38,10 @@ Renamed following Methods:
 
 * Sulu\Bundle\MediaBundle\Entity\CollectionRepositoryInterface::count => Sulu\Bundle\MediaBundle\Entity\CollectionRepositoryInterface::countCollections
 
+Rename following classes:
+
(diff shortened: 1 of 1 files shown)


alexander-schranz:
@sulu/core-team for me it feels a little bit strange that this constant is on `AbstractHttpCache` and not on a interface.

trickreich:
Why?

danrot:
@alexander-schranz Usually I am on you with this one, but we don't have an interface in our repository for that. And just adding it for a constant also feels weird to me.

danrot:
Merged anyway, since the change in this PR is true regardless of where we put that constant.

PR #3936, opened 2018-04-26
Added inactive state for checkbox

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | none
| Related issues/PRs | none
| License | MIT
| Documentation PR | none

#### What's in this PR?

This PR add a field-transformer for boolean values and an inactive state to the checkbox and switch component.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Checkbox/Checkbox.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Checkbox/Checkbox.js
@@ -8,6 +8,7 @@ import checkboxStyles from './checkbox.scss';
 type Props = SwitchProps & {
     skin: 'dark' | 'light',
     className?: string,
+    active: boolean,
     onChange?: (checked: boolean, value?: string | number) => void,
 };
 
@@ -16,6 +17,7 @@ const CHECKED_ICON = 'su-check';
 export default class Checkbox extends React.PureComponent<Props> {
     static defaultProps = {
         skin: 'dark',
+        active: true,
     };
 
     render() {
@@ -27,6 +29,7 @@ export default class Checkbox extends React.PureComponent<Props> {
             onChange,
             children,
             className,
+            active,
         } = this.props;
         const checkboxClass = classNames(
             checkboxStyles.checkbox,
@@ -42,6 +45,7 @@ export default class Checkbox extends React.PureComponent<Props> {
                 name={name}
                 icon={checked ? CHECKED_ICON : undefined}
                 onChange={onChange}
+                active={active}
             >
                 {children}
             </Switch>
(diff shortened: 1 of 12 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/components/Switch/Switch.js
```diff
@@ -32,10 +33,14 @@ export default class Switch extends React.PureComponent<Props> {
...
             className,
+            inactive,
         } = this.props;
         const switchClass = classNames(
             switchStyles.switch,
-            className
+            className,
+            {
+                [switchStyles.inactive]: inactive,
+            }
```
alexander-schranz: if inactive I would set `disabled` or `readonly` flag on the input.

src/Sulu/Bundle/AdminBundle/Resources/js/components/Checkbox/Checkbox.js
```diff
@@ -8,6 +8,7 @@ import checkboxStyles from './checkbox.scss';
 type Props = SwitchProps & {
     skin: 'dark' | 'light',
     className?: string,
+    inactive?: boolean,
```
danrot: How about `active` as a mandatory parameter, and setting it to `true` by default? Positive names are easier to understand IMO.

src/Sulu/Bundle/AdminBundle/Resources/js/components/Checkbox/Checkbox.js
```diff
@@ -27,10 +28,14 @@ export default class Checkbox extends React.PureComponent<Props> {
...
             children,
             className,
+            inactive,
         } = this.props;
         const checkboxClass = classNames(
             checkboxStyles.checkbox,
-            checkboxStyles[skin],
+            {
+                [checkboxStyles.inactive]: inactive,
+                [checkboxStyles[skin]]: !inactive,
```
danrot: Any reason not to set the skin when the button is not active? Maybe there will even be different inactive styles for each skin at some point.

src/Sulu/Bundle/AdminBundle/Resources/js/components/Switch/Switch.js
```diff
@@ -10,6 +10,7 @@ type Props = SwitchProps & {
     icon?: string,
     type: string,
     onChange?: (checked: boolean, value?: string | number) => void,
+    inactive?: boolean,
```
danrot: Would make the same change as in the `Checkbox` here

src/Sulu/Bundle/AdminBundle/Resources/js/components/Checkbox/Checkbox.js
wachterjohannes: currently the same css attributes are set differently. so there is currently no difference - but we can set both classes.

wachterjohannes:
@danrot @alexander-schranz should be good now

PR #3942, opened 2018-04-30
Refactored account-model

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | none
| Related issues/PRs | none
| License | MIT
| Documentation PR | none

#### What's in this PR?

This PR refactores the account-model to allow overwriting it.

#### To Do

- [x] Tests

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -126,9 +126,8 @@ Renamed following Methods:
 Rename following classes/interfaces:
 
 * Sulu\Component\HttpCache\HttpCache => Sulu\Bundle\HttpCacheBundle\Cache\AbstractHttpCache
-* Sulu\Component\Contact\Model\ContactInterface => Sulu\Bundle\ContactBundle\Model\ContactInterface
-* Sulu\Component\Contact\Model\ContactRepositoryInterface => Sulu\Bundle\ContactBundle\Model\ContactRepositoryInterface
-* Sulu\Bundle\ContactBundle\Entity\AccountInterface => Sulu\Bundle\ContactBundle\Model\AccountInterface
+* Sulu\Component\Contact\Model\ContactInterface => Sulu\Bundle\ContactBundle\Entity\ContactInterface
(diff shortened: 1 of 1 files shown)


UPGRADE.md
```diff
@@ -114,13 +114,21 @@ Removed following Classes:
...
+
 Renamed following Methods:
 
 * Sulu\Bundle\MediaBundle\Entity\CollectionRepositoryInterface::count => Sulu\Bundle\MediaBundle\Entity\CollectionRepositoryInterface::countCollections
 
-Rename following classes:
+Rename following classes/interfaces:
 
 * Sulu\Component\HttpCache\HttpCache => Sulu\Bundle\HttpCacheBundle\Cache\AbstractHttpCache
+* Sulu\Component\Contact\Model\ContactInterface => Sulu\Bundle\ContactBundle\Model\ContactInterface
```
danrot: Although I generally like the `Model` namespace better than `Entity` I would still go for `Entity` because it would be more consistent (or do we already have the `Model` namespace in any bundle?)

src/Sulu/Bundle/ContactBundle/Entity/AccountRepository.php
```diff
@@ -416,4 +366,16 @@ protected function appendJoins(QueryBuilder $queryBuilder, $alias, $locale)
             ->leftJoin($alias . '.categories', 'categories')
             ->leftJoin('categories.translations', 'translations');
     }
+
+    /**
+     * Create a new instance of a model.
+     *
+     * @return mixed
```
danrot: Since you've already removed the phpdoc on all other methods I would also remove it here.

UPGRADE.md
danrot: Didn't you adjust the upgrade file?

PR #3951, opened 2018-05-07
add yoda condition rule to eslint

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR adds a rule to always make use of yoda conditions. 

#### Why?

Yoda conditions make it impossible to use the assignment (`=`) operator instead of a comparison (`==` or `===`). However, the [page for the rule itself](https://eslint.org/docs/rules/yoda) explains that it might be less readable, and that eslint would detect assignments anyway (tested it, and it really does). But since we decided to do this as well in PHP it might be nice to be consistent here.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/AutoComplete/AutoComplete.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/AutoComplete/AutoComplete.js
@@ -124,7 +124,7 @@ export default class AutoComplete extends React.Component<Props> {
         } = this.props;
         const {inputValue} = this;
         const suggestions = this.createSuggestions(children);
-        const showSuggestionList = (!!inputValue && 0 < inputValue.length) && 0 < React.Children.count(children);
+        const showSuggestionList = (!!inputValue && inputValue.length > 0) && React.Children.count(children) > 0;
 
         return (
             <div className={autoCompleteStyles.autoComplete}>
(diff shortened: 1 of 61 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/components/AutoComplete/AutoComplete.js
```diff
@@ -124,7 +124,7 @@ export default class AutoComplete extends React.Component<Props> {
         } = this.props;
         const {inputValue} = this;
         const suggestions = this.createSuggestions(children);
-        const showSuggestionList = (!!inputValue && inputValue.length > 0) && React.Children.count(children) > 0;
+        const showSuggestionList = (!!inputValue && 0 < inputValue.length) && 0 < React.Children.count(children);
```
danrot: Especially the `<` and `>` operators feel a bit unusual this way.

danrot:
@chirimoya What's your take on this? I would say we should add a rule to be consistent, and have a slight preference to not use yoda conditions, because ESLint catches the assignments in an if statement (the reason that would speak for yoda conditions) and in some cases it is really less readable to me. Also, it seems we have done it the other way until now (if I apply the rule the other way there are only the two changes you suggested from #3934.

PR #3953, opened 2018-05-08
Fix sulu route options with dash

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

By default symfony converts dashes to underline.

#### Why?

Example:

```yaml
sulu_route:
    mappings:
        Sulu\Bundle\ArticleBundle\Document\ArticleDocument:
            generator: template
            options:
                recipe-gallery: "/galerie/{object.getTitle()}"
```

Currently `recipe-gallery` will be converted to `recipe_gallery` which does not match with the template key.

http://symfony.com/doc/current/components/config/definition.html#normalization

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 CHANGELOG for Sulu
 ==================
 
+* unreleased
+    * HOTFIX      #3953 [RouteBundle]           Fix route options with dash
+
 * 1.5.14 (2018-05-03)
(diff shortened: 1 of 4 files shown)


danrot:
Isn't that something that should have a test?

danrot:
And don't you think it makes more sense to have that be tested by @wachterjohannes? Maybe he can think about some unintended side effect I don't know.

alexander-schranz:
@danrot test added.

PR #3954, opened 2018-05-08
Color picker component

trickreich:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes --
| Related issues/PRs | --
| License | MIT
| Documentation PR | --

#### What's in this PR?

Adds component and field type `ColorPicker`.

#### Why?

We need such a component! ;)

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
         "flow-bin": "^0.67.1",
         "glob": "^7.1.2",
         "identity-obj-proxy": "^3.0.0",
-        "jest": "^22.4.0",
+        "jest": "^23.0.0-charlie.1",
         "mobx": "^4.0.0",
         "moment-timezone": "^0.5.14",
         "null-loader": "^0.1.1",
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
         "flow-bin": "^0.67.1",
         "glob": "^7.1.2",
         "identity-obj-proxy": "^3.0.0",
-        "jest": "^23.0.0-charlie.1",
+        "jest": "^22.4.0",
         "jest-canvas-mock": "^1.0.2",
         "mobx": "^4.0.0",
         "moment-timezone": "^0.5.14",
--- a/package.json
+++ b/package.json
@@ -43,6 +43,7 @@
         "glob": "^7.1.2",
         "identity-obj-proxy": "^3.0.0",
         "jest": "^23.0.0-charlie.1",
+        "jest-canvas-mock": "^1.0.2",
         "mobx": "^4.0.0",
         "moment-timezone": "^0.5.14",
         "null-loader": "^0.1.1",
@@ -66,6 +67,9 @@
         "moduleNameMapper": {
             "\\.(scss|css)$": "identity-obj-proxy"
         },
+        "setupFiles": [
+            "jest-canvas-mock"
+        ],
         "setupTestFrameworkScriptFile": "./testSetup.config.js",
         "snapshotSerializers": [
             "enzyme-to-json/serializer"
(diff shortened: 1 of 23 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/components/ColorPicker/ColorPicker.js
```diff
@@ -0,0 +1,163 @@
...
+@observer
+export default class ColorPicker extends React.Component<Props> {
+    static defaultProps = {
+        valid: true,
+    };
+
+    @observable value: ?string;
+    @observable showError: boolean = false;
+    @observable popoverOpen: boolean = false;
+    @observable popoverAnchorElement: ?ElementRef<*>;
```
danrot: This code is repeated quite often, right? Maybe we can refactor that into a separate `PopoverInput` component or some better name. Doesn't necessarily have to be in this PR, but if we see this pattern more often it might be worth a thought.

src/Sulu/Bundle/AdminBundle/Resources/js/components/ColorPicker/ColorPicker.js
```diff
@@ -0,0 +1,163 @@
...
+        }
+
+        const {onBlur} = this.props;
+
+        if (onBlur) {
+            onBlur();
+        }
+    };
+
+    handleChange = (value: Object) => {
```
danrot: We can't use the `isValidValue` function here, because this object is a representation from the `react-color` dependency, right?

src/Sulu/Bundle/AdminBundle/Resources/js/components/ColorPicker/README.md
```diff
@@ -0,0 +1,13 @@
+The email component can be used to get a valid email from the user.
```
danrot: `email`? :wink:

danrot: Maybe we should also mention in these components what happens when the icon is clicked.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/tests/fields/ColorPicker.test.js
```diff
@@ -0,0 +1,32 @@
...
+test('Pass props correctly to component', () => {
+    const field = shallow(
+        <ColorPicker
+            onChange={jest.fn()}
+            onFinish={jest.fn()}
+            value={'xyz'}
+        />
+    );
+
+    expect(field.find(ColorPickerComponent).prop('valid')).toBe(true);
```
danrot: I think we should also check the other props.

src/Sulu/Bundle/AdminBundle/Resources/js/components/ColorPicker/ColorPicker.js
trickreich: correct

src/Sulu/Bundle/AdminBundle/Resources/js/components/ColorPicker/ColorPicker.js
trickreich: It's not really always exact the same..

package.json
```diff
@@ -42,7 +42,8 @@
         "flow-bin": "^0.67.1",
         "glob": "^7.1.2",
         "identity-obj-proxy": "^3.0.0",
-        "jest": "^22.4.0",
+        "jest": "^23.0.0-charlie.1",
```
danrot: I would feel more comfortable if we would not use beta dependencies.

trickreich:
@danrot Is changed

PR #3957, opened 2018-05-14
Added function to get value for property name

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | none
| Related issues/PRs | none
| License | MIT
| Documentation PR | none

#### What's in this PR?

This PR adds a function `getValueByName` to build dependent field-types.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/tests/FormInspector.test.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/tests/FormInspector.test.js
@@ -7,6 +7,9 @@ import FormStore from '../stores/FormStore';
 jest.mock('../../../stores/ResourceStore', () => jest.fn(function(resourceKey, id, options) {
     this.resourceKey = resourceKey;
     this.id = id;
+    this.data = {
+        test: 'value',
+    };
 
     if (options) {
         this.locale = options.locale;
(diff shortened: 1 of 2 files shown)


danrot:
Build is failing

danrot:
Build is still failing. You can ignore the warnings from eslint, but the errors should be fixed.

alexander-schranz:
@danrot @wachterjohannes how does this behave inside blocks? 

There will be 2 use cases I think could happen:

1. I want a value from the same block by name
2. I want a value outside from the block

danrot:
I have also already thought about that, and I think at some point it would make sense to have a `getValueByPath` method where you could put a JSON Pointer or something like that. However, I think having an easy method to access the first level would make sense as well.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/tests/FormInspector.test.js
```diff
@@ -17,6 +17,9 @@ jest.mock('../stores/FormStore', () => jest.fn(function(resourceStore) {
     this.resourceKey = resourceStore.resourceKey;
     this.id = resourceStore.id;
     this.locale = resourceStore.locale;
+    this.data = {
```
danrot: Can you please move that to the test where that belongs? I am sure there will be more representations of data we have to test against in the future.

PR #3967, opened 2018-05-17
Fix loading of minified js files for production

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Related issues/PRs | PR for 1.6 bundles: https://github.com/sulu/sulu/pull/3968
| License | MIT

#### What's in this PR?

Fix loading of minified js files for production and avoid error output for grunt-cli.

#### Why?

Some bundles did load the dev JS files instead of the dist files because of false order of the grunt tasks.


danrot:
There are 404 errors from the `SuluSearchBundle` and from the `SuluLocationBundle`.

danrot:
The `leaflet.css` file from the `SuluLocationBundle` is still not found...

PR #3956, opened 2018-05-11
add media upload to contact and account

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR adds a media upload field type to the contact and account form.

#### Why?

Because we need that for the avatar of the contact and the logo of accounts.

#### To Do

- [x] Create a documentation PR
- [x] Add field type
- [x] Make width of `SingleMediaUpload` dynamic
- [x] Delete image from account/contact
- [x] Button for downloading the image
- [x] Make round representation for contacts
- [x] Make background full height if only mime type icon is shown
- [x] Change the API to not load/receive the entire media object (?)
- [x] Make consistent use of either `infotext` or `info_text`.
- [x] Show some kind of default icons if no image is assigned
- [x] Only show buttons if an image is currently assigned
- [x] Load into correct collection ID if new media is uploaded
- [x] `<property label="false" />` to disable label in form (#3965)
- [x] Ask if image should be deleted
- [x] Make buttons look nice (create a new one)
- [x] Reuse button and add an `icon` property to it instead of a separate `SingleMediaUpload/Button` (?)

Outsource to issues:

- [ ] Allow to have sections in sections
- [ ] 40px padding on sections
- [ ] Add mime type check for accounts and contacts

PR #3970, opened 2018-05-17
read properties param value using expression language

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR introduces the possibility to use the expression language in the template `param` xml tag using the `expr` attribute.

#### Why?

Because we need to find some params using services, e.g. when deciding to which collection images from the `SingleMediaUpload` should be uploaded.

#### Example Usage

see #3956


chirimoya:
ACK

PR #3975, opened 2018-05-18
Make grid responsive

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR makes the grid responsive.

#### Why?

Because on a very narrow view port it should be displayed as a single column.

#### To Do

- [x] Fix username next to email in permission tab of contact
- [x] Fix Password Confirmation field with screen width < 520px

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/tests/__snapshots__/FieldBlocks.test.js.snap
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/tests/__snapshots__/FieldBlocks.test.js.snap
@@ -10,15 +10,15 @@ exports[`Render block with schema 1`] = `
           <div class=\\"icons\\"><span class=\\"su-trash-alt clickable\\" aria-label=\\"su-trash-alt\\" role=\\"button\\" tabindex=\\"0\\"></span><span class=\\"su-times clickable\\" aria-label=\\"su-times\\" role=\\"button\\" tabindex=\\"0\\"></span></div>
         </header>
         <article>
-          <div class=\\"grid\\">
-            <div style=\\"width: 100%; margin-left: 0%; margin-right: 0%;\\" class=\\"item gridItem\\">
+          <div class=\\"grid grid\\">
+            <div class=\\"item gridItem size size-12 space-before-0 space-after-0\\">
               <div class=\\"field\\">
                 <label class=\\"label\\">Text 1</label>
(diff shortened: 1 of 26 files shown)


danrot:
I really would prefer to have some visual regression testing for the grid before we merge that :-/

chirimoya:
@danrot so far we really don't have that many forms which we need to keep an eye on. I'm +1 for merging this soonish and not really afraid of, but I'm not sure about the max-with. Maybe we should use 1140px as max-with. At least for contact & accounts forms. For the mobile version the image size is also too small. Let's discuss on monday.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/tests/__snapshots__/FieldBlocks.test.js.snap
```diff
@@ -10,15 +10,15 @@ exports[`Render block with schema 1`] = `
           <div class=\\"icons\\"><span class=\\"su-trash-alt clickable\\" aria-label=\\"su-trash-alt\\" role=\\"button\\" tabindex=\\"0\\"></span><span class=\\"su-times clickable\\" aria-label=\\"su-times\\" role=\\"button\\" tabindex=\\"0\\"></span></div>
         </header>
         <article>
-          <div class=\\"grid\\">
-            <div style=\\"width: 100%; margin-left: 0%; margin-right: 0%;\\" class=\\"item gridItem\\">
+          <div class=\\"grid grid\\">
```
chirimoya: why we have 2 times the same class?

chirimoya:
and the tests are failing ;)

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/tests/__snapshots__/FieldBlocks.test.js.snap
danrot: It is not really the same class, one is from the `Grid` component and one from the `FieldBlocks` component.

danrot:
I've actually just found a visual regression in one of the delivered standard templates in sulu-minimal :confused:

chirimoya:
min: 420px
break-point: 520px
max-with: 1140px

PR #3980, opened 2018-05-19
Fix kernel.project_dir parameter for PreviewKernel

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Fix kernel.project_dir parameter for PreviewKernel.

#### Why?

Currently the preview kernel use the false project_dir it should use the WebsiteKernel project_dir which is currently not being used.

### DIFF
--- a/src/Sulu/Bundle/PreviewBundle/Preview/Renderer/PreviewKernel.php
+++ b/src/Sulu/Bundle/PreviewBundle/Preview/Renderer/PreviewKernel.php
@@ -21,6 +21,11 @@ class PreviewKernel extends \WebsiteKernel
     const CONTEXT_PREVIEW = 'preview';
 
     /**
+     * @var string
+     */
+    private $projectDir;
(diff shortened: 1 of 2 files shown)


src/Sulu/Bundle/PreviewBundle/Preview/Renderer/PreviewKernel.php
```diff
@@ -49,6 +54,26 @@ public function getRootDir()
         return $this->rootDir;
     }
 
+    /**
+     * {@inheritdoc}
+     */
+    public function getProjectDir()
+    {
+        if (null === $this->projectDir) {
+            $r = new \ReflectionClass(\WebsiteKernel::class);
```
danrot: No abbreviations.

PR #3932, opened 2018-04-26
Added routes for account add/edit forms

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | none
| Related issues/PRs | none
| License | MIT
| Documentation PR | none

#### What's in this PR?

This PR adds the route definitions for account add and edit form.

PR #3989, opened 2018-05-23
add account selection to contact

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR introduces the `single_account_selection` field type and adds it to the `Contact` form.

#### Why?

Because a contact must be assignable to an account.

#### To Do

- [x] Think about how to split `SingleItemSelection` and `AutoComplete` functionality
    - Use the same field type for both and make a more complicated configuration
    - Make two separate field types (single_account_autocomplete and single_account_selection)
    - Decide on a single field type for all selections (single_account_selection is AutoComplete and single_media_selection is SingleItemSelection)
        - Would also mean to rename the `SingleSelection` field type component to `AutoComplete`
- [x] Extract `ResourceRequester` from `AutoComplete` container into separate store

Outsource to Issues / separate PR:

- [ ] Rename `Assignment` to `Selection` (both field type and container component)
- [x] ~~Rename `checkbox` field type to `toggler` and only show toggler~~ -> #3932
- [ ] Do not send a separate request in the `SingleMediaUpload` component to be consistent with the `SingleSelection`


src/Sulu/Bundle/AdminBundle/Resources/js/services/Initializer/Initializer.js
```diff
@@ -80,13 +81,15 @@ function registerFieldTypes() {
     fieldRegistry.add('text_line', Input);
     fieldRegistry.add('text_area', TextArea);
     fieldRegistry.add('time', Time);
+
+    registerFieldTypesFromConfiguration(fieldTypeOptions['assignment'], Assignment);
```
chirimoya: `registerFieldTypesWithOptions`

PR #3997, opened 2018-05-28
Add notes to account and contact

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR adds a note field to both accounts and contacts.

#### Why?

Because we need this field and it was there before.

#### BC Breaks/Deprecations

The `notes` field is deprecated now, and is replaced by a simple text field. This means that there is only one note per account/contact now.

#### To Do

- [ ] Write migration
- [ ] Remove deprecated property and everything related to it

### DIFF
--- a/src/Sulu/Bundle/ContactBundle/Resources/config/forms/Account.xml
+++ b/src/Sulu/Bundle/ContactBundle/Resources/config/forms/Account.xml
@@ -38,7 +38,7 @@
             <property name="parent" type="single_account_selection" colspan="6">
                 <meta>
                     <title lang="en">Parent company</title>
-                    <title lang="de">Mutterfirma</title>
+                    <title lang="de">Dachgesellschaft</title>
                 </meta>
             </property>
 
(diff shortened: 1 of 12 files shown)


src/Sulu/Bundle/ContactBundle/Resources/config/forms/Account.xml
```diff
@@ -27,6 +27,34 @@
...
+                <meta>
+                    <title lang="en">Corporation</title>
+                    <title lang="de">Gesellschaft</title>
+                </meta>
+            </property>
+
+            <property name="parent" type="single_account_selection" colspan="6">
+                <meta>
+                    <title lang="en">Parent company</title>
+                    <title lang="de">Mutterfirma</title>
```
chirimoya: Dachgesellschaft

src/Sulu/Bundle/ContactBundle/Resources/config/forms/Contact.xml
```diff
@@ -72,6 +72,13 @@
                     <title lang="de">Anrede</title>
```
chirimoya: Grußformel

PR #3998, opened 2018-05-29
Fixed user-blame

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | none
| Related issues/PRs | none
| License | MIT
| Documentation PR | none

#### What's in this PR?

This PR fixes the user-blame subscriber which should set the changer (and creator).

#### Why?

This seems not work (no idea since when - master seems to work maybe since doctrine update).

### DIFF
--- a/src/Sulu/Component/Persistence/EventSubscriber/ORM/UserBlameSubscriber.php
+++ b/src/Sulu/Component/Persistence/EventSubscriber/ORM/UserBlameSubscriber.php
@@ -153,25 +153,21 @@ class UserBlameSubscriber implements EventSubscriber
             $creatorChangeset = isset($changeset[self::CREATOR_FIELD]) ? $changeset[self::CREATOR_FIELD] : null;
             $changerChangeset = isset($changeset[self::CHANGER_FIELD]) ? $changeset[self::CHANGER_FIELD] : null;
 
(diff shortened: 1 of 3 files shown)


danrot:
The creator is still set on update when it was null before.

src/Sulu/Component/Persistence/EventSubscriber/ORM/UserBlameSubscriber.php
```diff
@@ -122,56 +122,45 @@ public function onFlush(OnFlushEventArgs $event)
...
-        $user = null;
+        $user = $this->getUser($token);
+
+        // if no sulu user, do nothing
+        if (!$user instanceof UserInterface) {
+            return;
+        }
+
+        $this->handleUserBlame($event, $user, true);
+        $this->handleUserBlame($event, $user, false);
```
danrot: I don't really see why we have to call that method twice. When I look at the code it looks to me like the changer is also set in the first call.

wachterjohannes: there could be insertions and updates in the same session

PR #4002, opened 2018-05-29
Add datagrid for Categories

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR will add a datagrid for the categories and adapt the categories to work correctly.

#### Why?

We need a list for the categories.

#### BC Breaks/Deprecations

The `CategoryController` will have a different API when finished.

#### To Do

- [x] Opening already loaded but not currently active category will reload children - Should categories be reloaded every time? Would be closer to how the column list behaves. -> always reload
- [x] `onItemActivation` and `onItemDeactivation` as mandatory properties? -> make mandatory
- [x] Error if all parent and child are deleted at once (parent will be deleted first, cascades delete to child, and child delete request will fail with 404) -> ignore 404 on delete
- [x] Add datagrid view
- [x] Change category API
- [x] Allow to sort categories
- [x] Add add button
- [x] Fix select all checkbox
- [x] Deactivate search (not completely clear how the search for categories should work yet)
- [x] Fix changing language to reload complete tree
- [x] Fix wrong open states in `TreeTable` (Maybe create two separate StructureStrategies - `Tree` and `Column` - instead of a single one and only pass visible data using the data attribute, would also move some logic outside of the Adapters)
- [x] Fix wrong position of `ColumnList` toolbar when after locale change less columns are available
- [x] Disallow sorting in page `TreeTable`
- [x] Category is wrongly described as expanded when grand parent category is opened and closed
- [x] Arrow icon is kept after all children of a category are deleted
- [x] make all the props from `DatagridAdapterProps` mandatory


danrot:
@chirimoya Instead of what we've discussed today with a `getVisibleItems` function in the `StructureStrategy` I thought it might make more sense to introduce a new one, called `ColumnStrategy`. That would allow us to move some logic from the [`ColumnList`](https://github.com/sulu/sulu/blob/develop/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Datagrid/adapters/ColumnListAdapter.js#L51) and the [`TreeTable`](https://github.com/sulu/sulu/blob/develop/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Datagrid/adapters/TreeListAdapter.js#L47) adapters to a place it better suits IMO, and would solve one aspect of the select all problem we currently have in the `TreeTable`, because we would only have the visible items retrieved by the `DatagridStore` anyway.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Datagrid/structureStrategies/ColumnStructureStrategy.js
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    @computed get data(): Array<Array<Object>> {
+        return Array.from(this.rawData.values());
+    }
+
+    @action getData(parent: ?string | number) {
+        const parents = this.activeItems;
+        const parentIndex = this.data.findIndex((column) => column.findIndex((item) => item.id === parent) !== -1);
+        parents.filter((parent, index) => index > parentIndex).forEach((parent) => this.rawData.delete(parent));
```
danrot: I don't think it is a good idea to delete this data in a getter...

.travis.yml
```diff
@@ -61,7 +61,7 @@ before_script:
   - composer update $COMPOSER_FLAGS
 
 script:
-  - time ./bin/runtests -i -a $TEST_FLAGS
+  - time ./bin/runtests -i -a $TEST_FLAGS --flags="--testdox"
```
alexander-schranz: :+1:

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Datagrid/structureStrategies/ColumnStructureStrategy.js
```diff
@@ -0,0 +1,64 @@
+// @flow
+import {action, computed, observable} from 'mobx';
+import type {StructureStrategyInterface} from '../types';
+
+export default class ColumnStructureStrategy implements StructureStrategyInterface {
+    @observable rawData: Map<?string | number, Array<Object>> = new Map();
+
+    @computed get visibleData(): Array<Object> {
```
chirimoya: visibleItems

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Datagrid/structureStrategies/ColumnStructureStrategy.js
```diff
@@ -0,0 +1,64 @@
...
+        return Array.from(this.rawData.values());
+    }
+
+    constructor() {
+        this.rawData.set(undefined, []);
+    }
+
+    activate(id: ?string | number) {
+        const parents = this.activeItems;
+        const parentIndex = this.data.findIndex((column) => column.findIndex((item) => item.id === id) !== -1);
```
chirimoya: columnIndex

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Datagrid/structureStrategies/ColumnStructureStrategy.js
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    @action getData(parent: ?string | number) {
+        return this.rawData.get(parent);
+    }
+
+    remove(identifier: string | number) {
+        for (const column of this.rawData.values()) {
+            for (const index of column.keys()) {
+                if (column[index].id === identifier) {
```
chirimoya: add TODO

PR #4009, opened 2018-06-05
Bugfix: Datagrid search reload

trickreich:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes --
| Related issues/PRs | --
| License | MIT
| Documentation PR | --

#### What's in this PR?

Set value of datagrid search correctly when it's changing from above (for example router/store).

#### Why?

Current search wasn't visible in the input when opening direct link.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Datagrid/Datagrid.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Datagrid/Datagrid.js
@@ -132,7 +132,7 @@ export default class Datagrid extends React.Component<Props> {
             <Fragment>
                 <div className={datagridStyles.toolbar}>
                     {searchable &&
-                        <Search onSearch={this.handleSearch} />
+                        <Search onSearch={this.handleSearch} value={store.searchTerm ? store.searchTerm.get() : null} />
                     }
                     <AdapterSwitch
                         adapters={adapters}
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Datagrid/Datagrid.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Datagrid/Datagrid.js
@@ -132,7 +132,7 @@ export default class Datagrid extends React.Component<Props> {
             <Fragment>
                 <div className={datagridStyles.toolbar}>
                     {searchable &&
-                        <Search onSearch={this.handleSearch} value={store.searchTerm ? store.searchTerm.get() : null} />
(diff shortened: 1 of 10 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/containers/Datagrid/tests/Search.test.js
```diff
@@ -11,15 +11,35 @@ jest.mock('../../../utils/Translator', () => ({
...
-        <Search onSearch={jest.fn()} />
+        <Search onSearch={jest.fn()} value={null} />
     );
 
     expect(search).toMatchSnapshot();
 });
 
+test('The component should render not collapsed when value is given', () => {
+    const search = shallow(
+        <Search onSearch={jest.fn()} value={'search-string'} />
```
danrot: Until now we have written `value="search-string"`. Also valid for the other tests here.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Datagrid/Search.js
```diff
@@ -22,6 +23,24 @@ export default class Search extends React.Component<Props> {
         this.value = value;
     }
 
+    updateState(value: ?string) {
```
danrot: `updateState` is a very generic name... Can't we come up with something more specific? Maybe `updateValue`?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Datagrid/Datagrid.js
```diff
@@ -132,7 +132,7 @@ export default class Datagrid extends React.Component<Props> {
             <Fragment>
                 <div className={datagridStyles.toolbar}>
                     {searchable &&
-                        <Search onSearch={this.handleSearch} />
+                        <Search onSearch={this.handleSearch} value={store.searchTerm ? store.searchTerm.get() : null} />
```
danrot: Is this check really necessary? From what I see `searchTerm` is always defined. Only the return value of `store.searchTerm.get()` could be undefined, but I think that would also be an acceptable option here.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Datagrid/Search.js
trickreich: It's not just an update of the value here..

PR #4013, opened 2018-06-07
Interpret .php as html request format to show correct error page

### LINKED ISSUE #4011: Requesing a URL ending with .php does not use correct error pages
| Q | A
| --- | ---
| Bug? | no
| New Feature? | no
| Sulu Version | 1.6.19

#### Actual Behavior

Hello, for an old website we have url like `/index.php?RubID=3&SRubID=3&PageID=50`. Sulu is used for the new website. When we follow old referenced links, we have 404 page (apache ?) but not sulu 404 custom page.

Same thing for sulu rocks `https://sulu.rocks/index.php`

@danrot said on slack : "Actually I think the `.php` suffix is the problem... We use the suffix to know which format we want to display (json, xml or html), and there is no php format, so that's why we have this strange error I guess"

#### Expected Behavior

Redirect to sulu error custom page.

#### Steps to Reproduce

Go to `https://sulu.rocks/index.php`

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #4011
| License | MIT

#### What's in this PR?

Interpret .php as html request format to show correct error page.

#### Why?

Else php will render the symfony standard error page.

#### Example Usage

`/test.php`

### DIFF
--- a/src/Sulu/Component/Webspace/Analyzer/Attributes/PortalInformationRequestProcessor.php
+++ b/src/Sulu/Component/Webspace/Analyzer/Attributes/PortalInformationRequestProcessor.php
@@ -114,6 +114,10 @@ class PortalInformationRequestProcessor implements RequestProcessorInterface
             strlen($portalInformation->getUrl())
         );
 
+        if ('php' === $formatResult) {
+            $formatResult = 'html';
+        }
+
         return [$resourceLocator, $formatResult];
     }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@ CHANGELOG for Sulu
 ==================
 
 * dev-release/1.5
+    * BUGFIX      #4013 [Components]            Fix 404 error page for .php requests
     * BUGFIX      #4018 [SnippetBundle]         Fix conflict when saving snippet in new language
     * BUGFiX      #3995 [TestBundle]            Fix tests for latest Symfony version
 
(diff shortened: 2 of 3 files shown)


src/Sulu/Component/Webspace/Analyzer/Attributes/PortalInformationRequestProcessor.php
```diff
@@ -114,6 +114,10 @@ private function getResourceLocatorFromRequest(PortalInformation $portalInformat
             strlen($portalInformation->getUrl())
         );
 
+        if ('php' === $formatResult) {
```
danrot: Shouldn't we fallback to `html` in general? We have exactly the same problem when people are typing something like `/index.asdf`.

alexander-schranz: @danrot 
@wachterjohannes or @chirimoya said there where usecases where the ending was something like `.newsletter` so this would make it a little difficult without a bc break doing that.

alexander-schranz: @danrot maybe instead of overwriting the request format. we should always render the when symfony exception controller return html: https://github.com/sulu/sulu/blob/1fee9e07b9c16e808c87974ef6a92c1da6dd4ba8/src/Sulu/Bundle/WebsiteBundle/Controller/ExceptionController.php#L92?

alexander-schranz: @danrot maybe we should just fix something in the exception controller instead of overwritting the request format behaviour? /cc @sulu/core-team

danrot: @alexander-schranz Your latest suggestion what we have just discussed with @chirimoya :-) I think we should do it this way. This keeps the old behaviour and fixes the messed up error message for unknown formats.

PR #4022, opened 2018-06-13
Force lowest symfony version to be 3.4

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Force lowest symfony version to be 3.4.

#### Why?

No incompatible symfony components should be installed.


composer.json
```diff
@@ -95,13 +94,10 @@
         "pulse00/ffmpeg-bundle": "^0.6",
         "symfony/browser-kit": "^3.4",
         "symfony/css-selector": "^3.4",
-        "symfony/debug": "^3.4",
         "symfony/debug-bundle": "^3.4",
         "symfony/dom-crawler": "^3.4",
         "symfony/monolog-bundle": "^3.1",
-        "symfony/monolog-bridge": "^3.4",
         "symfony/phpunit-bridge": "^3.4.8",
-        "symfony/process": "^3.4",
```
danrot: This seems to be used in the [`bin/runtests` script](https://github.com/sulu/sulu/blob/develop/bin/runtests#L21), so it should probably stay.

composer.json
```diff
@@ -95,13 +94,10 @@
         "pulse00/ffmpeg-bundle": "^0.6",
         "symfony/browser-kit": "^3.4",
         "symfony/css-selector": "^3.4",
-        "symfony/debug": "^3.4",
```
danrot: This is used in the [front controllers](https://github.com/sulu/sulu/blob/develop/web/app_dev.php#L17).

alexander-schranz: fixed

composer.json
alexander-schranz: fixed

PR #4026, opened 2018-06-15
Fix accidentally overwrites of unselected author

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Fix accidentally overwrites of unselected author.

#### Why?

Currently its not possible to unset an author as everytime the creator will be set instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 CHANGELOG for Sulu
 ==================
 
+* dev-release/1.6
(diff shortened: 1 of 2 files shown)


alexander-schranz:
@danrot before I invest fixing the tests. how would you fix this issue?

danrot:
Good question, had a look at it once for quite some short time, and I didn't really have an idea :see_no_evil: Does your approach work?

PR #4027, opened 2018-06-18
Fix correct url in preview

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Fix preview host replacement for render calls.

#### Why?

Subrequest (e.g. twig render) are created with Request::create which will replace currently `SERVER_PROTOCOL`, `HTTP_HOST`, and [many other headers](https://github.com/symfony/symfony/blob/v3.4.11/src/Symfony/Component/HttpFoundation/Request.php#L344-L357) which are currently not set in our Preview Request. If we forward them correctly in the preview they will not be overwritten by defaults.

#### Example Usage

1. Open page in backend
2. Use render function in template to render a sulu:link on a webspace with {host}
3. without this fix localhost would be used instead of the current domain

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ CHANGELOG for Sulu
 * dev-master
     * BUGFIX      #4042 [AudienceTargetingBundle] Add symfony 3.4.12 as conflict to fix caching tests
     * HOTFIX      #4019 [Component]               Fix handling of authored date on safari
+    * HOTFIX      #4027 [PreviewBundle]           Fix correct url in preview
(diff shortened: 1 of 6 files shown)


alexander-schranz:
ping @wachterjohannes

danrot:
It doesn't seem to work anymore with the `SuluThemeBundle`... However, I am not totally sure if that is my fault... Can you double check?

danrot:
Tested it on the `sulu-standard` edition, and it seems to work. But the tests are failing...

PR #4028, opened 2018-06-18
Log errors in image generation

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Related issues/PRs | extract parts from #3706
| License | MIT

#### What's in this PR?

Log errors in image calculation.

#### Why?

That errors are logged why a image could not be calculated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@ CHANGELOG for Sulu
 ==================
 
 * dev-release/1.5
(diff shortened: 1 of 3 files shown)


danrot:
@alexander-schranz This needs a rebase, and what's the easiest way to force such an error for testing?

alexander-schranz:
@danrot rebased. common error are:
 - original file does not longer exist
 - image type is not supported (e.g. png library not installed)

PR #4034, opened 2018-06-20
Fix naming of category tables and fields

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Fix naming of category tables and fields.

#### Why?

Currently the naming of the fields are auto generated and one table has interface in it and the other table has a false field name.

#### BC Breaks/Deprecations

Manual database migration is needed as schema update will not work here.


danrot:
@alexander-schranz Can you maybe rebase, so that I can see that the tests are actually succeeding?

PR #4037, opened 2018-06-21
Cleanup ContentTypeInterface for 2.0

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Remove unneeded functions from the ContentTypeInterface.

#### Why?

A content type does not longer need a template it need a field type be configured on this tag.

#### BC Breaks/Deprecations

getTemplate is not longer needed to be implement for Content Types.

#### To Do

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -2,6 +2,16 @@
 
 ## dev-develop
 
+### Content Types
+
+The following functions where removed from the `ContentTypeInterface`:
+
+ - getTemplate
+ - getDefaultParams
+ - getDefaultValue
+
+When you have a custom content type you need to register a field type on the same name.
+
 ### Test Setup changed
 
 If you use the SuluTestBundle to test your custom sulu bundles you maybe need to change in your test config.yml
--- a/src/Sulu/Bundle/ContentBundle/Resources/views/Template/macros/single.html.twig
+++ b/src/Sulu/Bundle/ContentBundle/Resources/views/Template/macros/single.html.twig
@@ -4,7 +4,15 @@
         {{ property.getTitle(userLocale) }}
     </label>
 
-    {% include type.template %}
+    {% if type.template|default(false) %}
+        {% include type.template %}
+    {% else %}
+        <div>
+            <span class="badge badge-red">Template for content type "{{ type.name }}" was not defined.</span>
+        </div>
+
+        <div class="block-content" data-sort-mode-id="{{ id|raw }}"></div>
+    {% endif %}
 
     {% if property.getInfoText(userLocale) %}
         <div class="input-description">{{ property.getInfoText(userLocale) }}</div>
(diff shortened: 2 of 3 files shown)


UPGRADE.md
```diff
@@ -2,6 +2,16 @@
 
 ## dev-develop
 
+### Content Types
+
+The following functions where removed from the `ContentTypeInterface`:
```
danrot: `were`

UPGRADE.md
```diff
@@ -2,6 +2,16 @@
...
+
+ - getTemplate
+ - getDefaultParams
+ - getDefaultValue
+
+When you have a custom content type you need to register a field type on the same name.
```
danrot: Somehow explain that this has to happen in JS now?

src/Sulu/Bundle/ContentBundle/Resources/views/Template/macros/single.html.twig
```diff
@@ -4,7 +4,15 @@
         {{ property.getTitle(userLocale) }}
     </label>
 
-    {% include type.template %}
+    {% if type.template|default(false) %}
```
danrot: Actually this entire file can be removed quite soon :-) Any reason you've built this if into it?

src/Sulu/Component/Content/ContentTypeInterface.php
```diff
@@ -95,31 +95,6 @@ public function remove(
...
-     * @return array
-     */
-    public function getDefaultParams(PropertyInterface $property = null);
-
-    /**
-     * returns default value of content type.
-     *
-     * @return mixed
-     */
-    public function getDefaultValue();
```
danrot: What exactly is this doing? Are we sure that we don't need default values in the future? :thinking:

src/Sulu/Component/Content/ContentTypeInterface.php
```diff
@@ -95,31 +95,6 @@ public function remove(
...
-    public function getTemplate();
-
-    /**
-     * Returns default parameters.
-     *
-     * @param PropertyInterface|null $property
-     *
-     * @return array
-     */
-    public function getDefaultParams(PropertyInterface $property = null);
```
danrot: Were this parameters only for the twig templates?

src/Sulu/Component/Content/ContentTypeInterface.php
alexander-schranz: Think a field type can have a default value but not a content type. A content type should just say how to save the given value to PHPCR and  how to load view and content data. /cc @sulu/core-team

src/Sulu/Bundle/ContentBundle/Resources/views/Template/macros/single.html.twig
alexander-schranz: only for the old UI.

UPGRADE.md
alexander-schranz: would link here to the documentation ->`Create custom field type` but this is currently not documented.

danrot:
Replaced by #4045

PR #4067, opened 2018-07-17
avoid returning the same id multiple times in filter query of smart content

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This PR avoids returning the same ID multiple times in the ID query of the smart content.

#### Why?

If a media has assigned Category A and B and a SmartContent using the `media` data provider has an or filter on Category A and B, then the ID query from the `DataProviderRepositoryTrait` returns the same media ID twice. That only matters if there is also a limit (either in the smart content overlay or using the `max_per_page` param in the template XML), because then less items will be returned in the final query.

### DIFF
--- a/src/Sulu/Component/SmartContent/Orm/DataProviderRepositoryTrait.php
+++ b/src/Sulu/Component/SmartContent/Orm/DataProviderRepositoryTrait.php
@@ -60,8 +60,8 @@ trait DataProviderRepositoryTrait
 
         $queryBuilder = $this->createQueryBuilder('c')
             ->select('c.id')
-            ->orderBy('c.id', 'ASC')
-            ->groupBy('c.id');
+            ->distinct()
+            ->orderBy('c.id', 'ASC');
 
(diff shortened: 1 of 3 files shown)


src/Sulu/Component/SmartContent/Orm/DataProviderRepositoryTrait.php
```diff
@@ -60,6 +60,7 @@ private function findByFiltersIds($filters, $page, $pageSize, $limit, $locale, $
 
         $queryBuilder = $this->createQueryBuilder('c')
             ->select('c.id')
+            ->distinct()
```
danrot: @chirimoya This was the easiest solution we could find. Another solution would be `groupBy('id')`. I have read that this might be faster in some situations, but I've used an `EXPLAIN` statement to check the execution plan, and they were identical for both solutions. So I decided to go with `distinct`, because it explains better what is actually happening here.

PR #4069, opened 2018-07-18
WIP: Remove sulu.rlp tag from test templates

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Remove sulu.rlp tag from test templates.

#### Why?

Sulu.rlp tag is deprecated and should not longer be needed.

##### TODO

 - [ ] UPGRADE
 - [ ] Fix services which require them

bildschirmfoto 2018-07-18 um 16 36 13

### DIFF
--- a/app/Resources/templates/pages/default.xml
+++ b/app/Resources/templates/pages/default.xml
@@ -32,8 +32,6 @@
                 <title lang="en">Resourcelocator</title>
                 <title lang="de">Adresse</title>
             </meta>
-
-            <tag name="sulu.rlp"/>
         </property>
 
         <property name="article" type="text_editor">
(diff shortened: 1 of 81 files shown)


alexander-schranz:
Not sure if we can really remove the `sulu.rlp` tag for the backend. There are tests where the `sulu.rlp` tag can be used on a `text_line` content type. /cc @sulu/core-team

danrot:
Since with the new react implementation it is not possible anymore to have the URL auto generated in a `text_line` field I would say we can still safely remove it. But that probably needs a few more changes, and I think it doesn't have a high priority currently.

alexander-schranz:
@danrot think we need to implement a `getPropertyByType` instead e.g. for the MaintainResourceLocatorCommand

PR #4070, opened 2018-07-18
Improve UPGRADING.md

reyostallenberg:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

Some hints on how to upgrade an existing sulu-minimal project are missing

#### Why?

I had problems upgrading an existing sulu-minimal project.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -248,8 +248,8 @@ DROP TABLE category_translation_media_interface;
 Some resources in app/config/admin/routing.yml need to be removed:
 - sulu_tag
 - sulu_contact
+- sulu_content
 - sulu_category
-- sulu_custom_urls_api 
 
 ### Security changes
 The admin security needs to be changed (app/config/admin/security.yml).
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -134,7 +134,13 @@ command:
 composer require pulse00/ffmpeg-bundle:^0.6
 ```
 
+Otherwise remove `new Dubture\FFmpegBundle\DubtureFFmpegBundle(),` from the list in app/AbstractKernel.php
+
 ### Deprecations
+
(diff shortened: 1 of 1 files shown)


UPGRADE.md
```diff
@@ -237,6 +243,62 @@ INSERT INTO ca_category_translation_medias (idCategoryTranslations, idMedia) SEL
 DROP TABLE category_translation_media_interface;
 ```
 
+### Routing changes
+
+Some resources in app/config/admin/routing.yml need to be removed:
+- sulu_tag
+- sulu_contact
+- sulu_category
+- sulu_custom_urls_api
```
alexander-schranz: Think this should be `sulu_content`.

I'm wondering if there are not more as I think there should now be only apis. /cc @danrot

reyostallenberg: I'm sorry, it should be the list as in https://github.com/sulu/sulu-minimal/compare/master...develop (I think)

danrot: Not really, since there are still a few non-API routes. I have only removed the ones which were generating the templates for our forms. There are still endpoints not being an API e.g. for generating images.

alexander-schranz: @danrot ok most of them seems to be api's like cache:clear or location, ... 

but think the following we could maybe remove:

 - sulu_media ( are just the form templates )
 - sulu_media_website ( was only here for the old preview I think? )
 - sulu_search ( was only here for the old preview I think? )

but thats now not related to this PRs.

@reyostallenberg think the only missing thing we had is the PHP 7.1 requirement which should also be added.

PR #4072, opened 2018-07-19
Change: always show content type in collapsed block

choonge:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | 
| Related issues/PRs | 
| License | MIT
| Documentation PR | 

#### What's in this PR?

Change: always show content type in collapsed block

#### Why?

Previously, the content type of a collapsed block would only be shown if
there was more than 1 type defined for the given block. However, for
editors that makes it impossible to see what kind of content is in that
spot. This change will make it so the content type is always shown.

### DIFF
--- a/src/Sulu/Bundle/ContentBundle/Resources/views/Template/content-types/block.html.twig
+++ b/src/Sulu/Bundle/ContentBundle/Resources/views/Template/content-types/block.html.twig
@@ -33,9 +33,7 @@
                         <input type="hidden" data-mapper-property="type" value="{{ type.name }}"/>
 
                         <div class="collapsed-container empty">
-                            {% if property.types|length > 1 %}
-                                <div class="type">{{ type.getTitle(userLocale) }}</div>
-                            {% endif %}
+                            <div class="type">{{ type.getTitle(userLocale) }}</div>
                             <div class="title"></div>
                             <div class="image"></div>
                             <div class="text hidden"></div>


danrot:
We have discussed that in the team, and this change does not really make sense for us. The type of the block is also only shown in the expanded block if there is more than one type. If there is just one block type, then neither the dropdown nor this information makes sense. In our opinion it would be confusing for the content manager, to see some string in the collapsed block, which is not visible at all in the expanded one. I am sorry, but therefore I am going to close this PR.

However, actually the content of the block should also be shown in the collapsed block, as in the following screenshot:
image

Isn't that the case for you?

choonge:
Thanks for taking this into consideration.

Indeed the content of the block does show in collapsed form, and that is often good enough for very basic blocks like some bodytext, or an image. However, we have several blocks which are a kind of 'widget', where the only inputfields are some translations. E.g. the following:

screen shot 2018-08-10 at 08 41 34

In this case the content in the collapsed block says nothing about what kind of block it is. And we would rather not add unused 'title' fields just to show in the admin, if we don't use those values in the front-end.

Long story short: I think not showing the type works in most cases, especially if you don't have many kinds of blocks. But when there are many block types and pages get complex, at least in our case, our editors have a hard time seeing which block they need to click in the admin, to adjust something on the front-end.

danrot:
Hm, okay, but what I don't get is how does adding the same type to every block instance help the content managers? All of them would probably be of type `Header left side` in your above example, wouldn't it?

choonge:
It's actually a 'chart' block that goes into that position. But interestingly, perhaps we did define too many sections. Perhaps we can find a way to merge some sections together, so they can contain multiple blocks. Then the block types will be shown.

danrot:
Well, then the content manager can also mix the blocks, depends on your use case if you want to have that. But if this is about charts, why isn't the section title, which is now "Header left side", "Charts"? Then it should also be clear that all subsequent blocks are of that "type", without repeating that information in every block.

choonge:
We interpreted `section` as being an area within a template. Since the header has a few different areas, we have a few different sections like `Header left side`, `Header right side` and `Header bottom`. This particular section can contain only a block of the type Chart, but other sections may contain multiple blocks of different types.

danrot:
Sorry, section title was the wrong word... You know that it is possible to give a [separate title for the blocks like in our sulu-demo](https://github.com/sulu/sulu-demo/blob/ae9b55e/app/Resources/templates/pages/services.xml#L30)?

choonge:
Thanks, we already do that indeed. But the problem we have, is that titel is never shown anywhere unless you have multiple possible blocks in that spot. If that spot can contain only 1 block type, you never see the title of that block.

Where you would normally see the block title is just empty (see arrow in screenshot).

sulu

danrot:
My reasoning above also matches your red arrow :-) And in our opinion it is better to name what is called `Header left side` in your screenshot to something that the content manager knows what he is editing there.

choonge:
Aaah I see. Thanks, I think we will do that indeed :)

PR #4079, opened 2018-07-23
Introduce SEO tab for pages

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR will add a SEO tab to the page forms. It will be built in a way to easily add it to forms of other resources as well.

#### Why?

Because this tab should be added to multiple different forms.

#### To Do


src/Sulu/Bundle/ContentBundle/Controller/SeoController.php
```diff
@@ -46,6 +48,12 @@ protected function getRepository()
         return $this->get('sulu_content.node_repository');
     }
 
+    public function cgetAction()
```
danrot: Don't see another way currently :confused:

src/Sulu/Bundle/ContentBundle/DependencyInjection/SuluContentExtension.php
```diff
@@ -46,6 +46,12 @@ public function prepend(ContainerBuilder $container)
                             ],
                         ],
                     ],
+                    'resources' => [
+                        'pages_seo' => [
+                            'form' => ['@SuluContentBundle/Resources/config/forms/PageSeo.xml'],
+                            'endpoint' => 'get_page-seos',
```
danrot: That's also weird... But I if I change the resource name to `page_seo`, then the URL will also be `/admin/api/page_seo`, which is suboptimal...

chirimoya:
@danrot SEO tab back button is missing

danrot:
@chirimoya Back button should be there now

PR #4082, opened 2018-07-25
Make ColumnList, Table and Navigation more keyboard friendly

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR adds a few `role="button"` attributes to make navigating without the mouse easier, since these elements are now recognized as button by external tools as well.

#### Why?

The Vimium Plugin in Chrome didn't recognize some of the clickable areas as clickable. There is still more to do, but what is fixed here are the most common action.

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,6 +87,7 @@
         "clearMocks": true,
         "transformIgnorePatterns": [
             "node_modules/(?!(sulu-(.*)-bundle|@ckeditor)/)"
-        ]
+        ],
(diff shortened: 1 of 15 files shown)


chirimoya:
@danrot rebase

PR #4085, opened 2018-07-31
Fix missing image in MediaDetail view

danrot:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | #4079 
| License | MIT
| Documentation PR | ---

#### What's in this PR?

The changes in #4079 broke the Media Detail view, which is fixed in this PR.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/Form.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/Form.js
@@ -1,5 +1,5 @@
 // @flow
-import {action, computed, observable} from 'mobx';
+import {action, autorun, computed, observable} from 'mobx';
 import {observer} from 'mobx-react';
 import React, {Fragment} from 'react';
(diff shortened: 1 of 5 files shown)


danrot:
I have realized that the `GhostPageOverlay` isn't working anymore for the same reason... And that the component within the view is not rerendered anymore on a locale change, which also makes problems for both features...

Going back to the old way before #4079 is also no option, because then requests will sometimes be duplicated... Will have a closer look at this tomorrow.

danrot:
I found out, that the stuff passed to the `MediaUploadStore` are observables, so doing this only once the `Form` is rendered using `when` works, and doing more would probably be just wasted execution time.

However, things are different in the `GhostDialog` case, because there an action has to be triggered. Therefore `autorun` seems to be a good choice. I've implemented it like that, and now both features work like I would imagine them to work.

PR #4098, opened 2018-08-07
Use ToolbarActions to define Toolbar in Form

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR will allow to add different toolbar items using an option on the routes.

#### Why?

Because in some Forms you might want to have different toolbar items in the Toolbar. Even in the core not every form has the same items in the Toolbar.

#### To Do

- [x] Add "Publish" option to save dropdown
- [x] Use `publishable_save` and `delete` in other page tabs as well (currently not possible because controllers can't handle publish and delete requests)
- [x] Remove delete button from contact/user
- [x] Save button gets active during tab change
- [x] Preselected categories make save button active
- [x] Parent resource store does not get updated when `publishedState` changes in SEO or Excerpt tab


src/Sulu/Bundle/AdminBundle/Resources/js/containers/Toolbar/types.js
```diff
@@ -28,3 +29,8 @@ export type ToolbarConfig = {
     locale?: Select,
     showSuccess?: IObservableValue<boolean>,
 };
+
+export interface ToolbarAction {
+    getElement(): Node,
```
chirimoya: `getNode` ???

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Toolbar/types.js
```diff
@@ -28,3 +29,8 @@ export type ToolbarConfig = {
...
+    getToolbarItemConfig(): ToolbarItem,
```
chirimoya: `ToolbarItemConfig`

chirimoya:
@danrot please add media delete

src/Sulu/Bundle/ContentBundle/Admin/ContentAdmin.php
```diff
@@ -98,8 +98,7 @@ public function getRoutes(): array
         ];
 
         $formToolbarActionsWithoutType = [
-            // TODO use publishable_save and delete when APIs support that
-            'sulu_admin.save',
+            'sulu_admin.publishable_save',
```
chirimoya: sulu_admin.save_with_publishing

src/Sulu/Bundle/AdminBundle/Resources/js/views/Form/Form.js
```diff
@@ -128,12 +157,34 @@ class Form extends React.Component<Props> {
...
-                    router.navigate(editRoute, {id: resourceStore.id, locale: resourceStore.locale});
+                    router.navigate(
+                        editRoute,
+                        {id: resourceStore.id, locale: resourceStore.locale, ...editRouteParameters}
+                    );
+                }
+
+                if (this.hasOwnResourceStore) {
+                    // Reload parent ResourceStore, since its data might have changed due to changes in this Form
+                    resourceStore.load();
```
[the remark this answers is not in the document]
danrot: We discussed this and came to the conclusion that we always reload the Content tab when we are reentering it.

src/Sulu/Bundle/AdminBundle/Resources/js/views/Form/Form.js
```diff
@@ -92,6 +93,13 @@ class Form extends React.Component<Props> {
                 : new ResourceStore(resourceKey, id, {locale}, formStoreOptions);
         } else {
             this.resourceStore = resourceStore;
+
+            if (Object.keys(this.resourceStore.data).length > 0) {
+                // data should be reloaded if ResourceTabs ResourceStore is used and user comes back from another tab
+                // the above check assumes that loading the data from the backend takes longer than calling this method
+                // the very unlikely worst case scenario if this assumption is not met, is that the data is loaded twice
+                this.resourceStore.load();
```
danrot: @chirimoya I have changed this now as we discussed yesterday. It works, with the only drawback mentioned in the comments above. This assumption is even met on my machine, when running Sulu locally, so if there is even more network latency that should be no problem. I've put the comment here, so that we remember this if we should be experiencing double requests.

And this way of handling it also solves the problem you've mentioned yesterday, the data is reset when you come back to the tab in the same way as if you are switching to another tab.

PR #4101, opened 2018-08-08
Add tab for page settings

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR adds the settings tab for pages.

#### Why?

Because some crucial functionality of the content management part is located in there.

#### BC Breaks/Deprecations

None yet.

#### To Do

- [ ] Do not display navigation select on certain conditions
- [x] Hide field completely if empty (done in #4155)
- [x] Do not validate if field is not returned at all resp. find a different solution (fields are not really required...)
- [x] disable id even if `resoureKey`s do not match but share same endpoint in `SingleSelection` field type
- [x] Do not show const errors


danrot:
DateTime for authored field is implemented, but I am not sure about the better solution for avoiding the double requests when reloading the page tab. Either we clone the `ResourceStore` if the endpoint is the same but the `resourceKey` is different, or we reuse the same one. The problem is the we also have to change the `resourceKey` of the `ResourceStore`, for which reason I would prefer to clone the store, so that the original one still works as expected (not that I would know of any side effects, but it would feel safer this way).

chirimoya:
- [ ] if you change the page type to internal or external link, the Link Title and Link should be mandatory
- [ ] it is not possible to remove the author or the authored date
- [ ] single selection with data grid overlay should use radios instead of checkboxes

src/Sulu/Bundle/ContentBundle/Resources/js/containers/Form/fields/PageSettingsNavigationSelect.js
```diff
@@ -0,0 +1,53 @@
...
+import {action, observable} from 'mobx';
+import {observer} from 'mobx-react';
+import {MultiSelect} from 'sulu-admin-bundle/components';
+import type {FieldTypeProps} from 'sulu-admin-bundle/types';
+import {translate} from 'sulu-admin-bundle/utils';
+import webspaceStore from '../../../stores/WebspaceStore';
+import type {Webspace} from '../../../stores/WebspaceStore/types';
+
+@observer
+export default class PageSettingsNavigationSelect extends React.Component<FieldTypeProps<Array<string | number>>> {
```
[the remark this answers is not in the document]
chirimoya: PageSettingsNavigationSelect

src/Sulu/Bundle/MediaBundle/Resources/js/views/MediaDetail/tests/MediaDetail.test.js
```diff
@@ -162,6 +162,7 @@ test('Should call update method of MediaUploadStore if a file was dropped', (don
     const metadataStore = require('sulu-admin-bundle/containers/Form/stores/MetadataStore');
     const resourceStore = new ResourceStore('test', testId, {locale: observable.box()});
     resourceStore.set('id', testId);
+    resourceStore.loading = false;
```
chirimoya: remove

danrot: Still seems to be needed :thinking:

Issue #3624, opened 2017-11-15
Normalize content type names for 2.0


alexander-schranz:
| Q | A
| --- | ---
| Bug? | no
| New Feature? | no
| Sulu Version | for 2.0

#### Actual Behavior

We have some multi selection content types like `snippet`, `form_select`, `media_selection`, `category_list`. Also the internal_links content type should be renamed to something which match more its case `page_selection` or `content_selection`.

#### Expected Behavior

This content types should be normalized to e.g.: `<type>_selection` or `<type>_list`

Here a list of all content types and which should be maybe renamed

| Current Name                | Expected Name                                   | Implemented
|----------------------|----------------------------------------------|------------------
| text_line            |                    | [x]
| text_area            |                   | [x]
| text_editor          |                  | [x]
| resource_locator     |             | [x]
| block                |                        | [x]
| **`contact`**              | contact_account_selection (#4566)                 | [ ] (ui missing)
|               | contact_selection                | [ ]
|               | account_selection                 | [ ]
| **`media_selection`**      |               | [x]
| **`category_list`**        | category_selection (#4335)  | [x]
| **`snippet`**              | snippet_selection (#4565)                 | [x]
| smart_content        |       |  [x]
| **`teaser_selection`** (#4528)    |              |  [x]
| **`internal_links`**       | page_selection (#4500)             |  [x]
| **`single_internal_link`** | single_page_selection (#4500)       |  [x] 
| phone                |                      | [x]
| password             |                    |  [x]
| url                  |                        |  [x]
| email                |                       | [x]
| date                 |                        |  [x]
| time                 |                        |  [x]
| color                |                       |  [x]
| **`checkbox`**             | ~~toggler? (#3992)~~                   |  [x]
| **`multiple_select`**     |  select (#4260)        |  [x]
| **`single_select`**        | single_select           |  [x]
| **`tag_list`**             | tag_selection (#4334)  |  [x]
| location (#4743)             |   |  [x]
| **route**                |   resource_locator (#4594)  |  [x]
| **`article_selection`**     | sulu/SuluArticleBundle#377  |  [x]
| page_tree_route      |   |  [ ]
| **`form_select`**          | form_selection  |  [ ]

Other components which exists currently as component but not as content type are:

| Name                                               | Implemented
|-------------------------------|------------------------
| **`single_contact_selection`**          |  [x] (#4162) 
| **`single_media_upload`**            |   [ ] (content type not needed I think)
[...]

PR #4102, opened 2018-08-09
Introduce AutoComplete holding multiple values

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR adds a `AutoComplete` field which can hold multiple values, as it is e.g. required for tag assignment.

#### Why?

Because we need a tag autocomplete in the Excerpt tab.

#### To Do

- [x] Move `AutoCompletePopover` to own component (?)
- [x] Rename `AutoComplete` to `SingleAutoComplete`
- [x] Introcude `MultiAutoComplete`
- [x] Add loader icon to `MultiAutoComplete`
- [x] Extract `AutoCompleteStore` to some generic place
- [x] check width of auto complete popover
- [x] Do not assign same IDs again
- [x] Add loader when initially loading


src/Sulu/Bundle/AdminBundle/Resources/js/components/MultiAutoComplete/README.md
```diff
@@ -0,0 +1,72 @@
+This component is the counterpart of the [`AutoComplete`](#autocomplete) component, but it
```
chirimoya: SingleAutoComplete

src/Sulu/Bundle/ContentBundle/Resources/config/forms/PageExcerpt.xml
```diff
@@ -26,6 +26,12 @@
             <title lang="en">Categories</title>
         </meta>
     </property>
+    <property name="tags" type="tag_list">
```
chirimoya: `tag_selection`

danrot: Renaming the content type results in a lot of test fails, so I would prefer to do that in a separate PR. There is already the #3624 tagged on the 2.0 release, so that we don't forget about it.

PR #4109, opened 2018-08-15
Remove validation-state from rendered link

reflexxion:
The LinkTag adds an attribute named `validation-state` if the target is not published.
After publishing the target page and not editing the link again the resulting link contains the invalid attribute.

| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

The LinkTag adds an attribute named `validation-state` if the target is not published.
After publishing the target page and not editing the link again the resulting link contains the invalid attribute.

#### Why?

Invalid attribute is rendered

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 CHANGELOG for Sulu
 ==================
 
-* dev-develop
+* dev-master
     * BUGFIX      #4109 [ContentBundle]         Remove validation-state from rendered link
 
 * 1.6.21 (2018-07-18)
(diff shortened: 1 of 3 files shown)


danrot:
Code looks good to me, but can you add a test to the [`LinkTagTest`](https://github.com/sulu/sulu/blob/cb84c75/src/Sulu/Bundle/ContentBundle/Tests/Unit/Markup/LinkTagTest.php#L54)? Not that we are reintroducing that error :smiley: 

@wachterjohannes Are there any more properties we should add to this list?

reflexxion:
Hi @danrot,
I added it to the test.

wachterjohannes:
@reflexxion thanks for this contribution

PR #4112, opened 2018-08-16
Added exception messages to user provider for enabled and locked

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Added exception messages to user provider for enabled and locked.

#### Why?

Easier to implement a better output for the community bundle if a user is not yet enabled.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,7 +55,7 @@ before_script:
         java -jar downloads/jackrabbit-standalone-$JACKRABBIT_VERSION.jar > /dev/null &
     fi
   # the content tests are intensive and there are memory leaks, this is more pronounced with the Jackalope DBAL PHPCR implementation.
-  - echo "memory_limit=2048M" >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
+  - echo "memory_limit=-1" >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
   - phpenv config-rm xdebug.ini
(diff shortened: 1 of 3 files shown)


danrot:
Why is it easier to build a better output this way? I think you shouldn't check for the string anyway, right? And there is also `getMessageKey` method in the Symfony exceptions which you could use.

Resp. could you link me to the place in the `SuluCommunityBundle` where this should help?

danrot:
An `instanceof` check also doesn't work? Or is there a `previous` exception? Checking on a string is very fragile, since nobody will think of that case when changing the string...

alexander-schranz:
@danrot the exception is written in the session store by symfony and read from the session store. as the session store serialize the data the previous exception is lost: https://github.com/symfony/symfony/blob/c31c40d83e55fcc804ff37a04a74c6a00ba71e8a/src/Symfony/Component/Security/Core/Exception/AuthenticationException.php#L41-L61

As additional to the message we could add a exception code? currently 1101 (RoleNameAlreadyExistsException) and 1009 (UserNotInSystemException) is used in the security bundle.

alexander-schranz:
ping @danrot

danrot:
Seems like the tests keep failing for php 5.5... can you try to raise the memory limit on composer install?

alexander-schranz:
@danrot memory limit increased, rebased and changelog added.

PR #4120, opened 2018-08-23
[POC] Added parameter to webspace-structure-format to set the format for ch…

wachterjohannes:
POC to overwrite checking format for webspace structures.


danrot:
@wachterjohannes Does it make sense to keep this open?

alexander-schranz:
Close because the WebspaceStructureProvider is not longer used and was replaced by excluded template on the webspace xml.

PR #4121, opened 2018-08-23
Set a timeout when purging caches

gisostallenberg:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?
Set a timeout when purging caches

#### Why?
The default behavior is to wait indefinitely (see [guzzle documentation](http://docs.guzzlephp.org/en/stable/request-options.html#connect-timeout)), which
may result in a very long request once the caches need to be cleared, but a long responding host is set in the config (see [example configuration](https://github.com/sulu/sulu-minimal/blob/7887ff338657fa5148c56153030d835d0da02c49/app/Resources/webspaces/example.com.xml#L43))

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@ CHANGELOG for Sulu
 ==================
 
 * dev-master
+    * BUGFIX      #4121 [HttpCache]             Set a timeout when purging caches
     * BUGFIX      #4109 [ContentBundle]         Remove validation-state from rendered link
 
 * 1.6.21 (2018-07-18)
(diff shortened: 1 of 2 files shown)


danrot:
@chirimoya In theory this is OK for me, but I don't really have experience with what a sensible value for the timeout is.

@gisostallenberg When does the request fail now? After the max execution time of PHP? And in the current master branch you still have to add a line in the `CHANGELOG`.

gisostallenberg:
@danrot, @chirimoya I will change the value to 5 seconds. The reasoning is that Sulu should take less than 3 seconds to spin up, but in some cases this might be to rigid. I'll also provide a `CHANGELOG` line.

gisostallenberg:
@danrot, @chirimoya I updated the PR as promised.

danrot:
@gisostallenberg thanks a lot!

PR #4124, opened 2018-08-29
Added http cache provider to sulu kernel

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Added http cache provider interface to sulu kernel.

#### Why?

This is required when optimize for single server setup. Same changes in sulu-minimal would also be needed but atleast we could now add the HttpCacheAware.

See also:
 - https://github.com/FriendsOfSymfony/FOSHttpCache/blob/master/doc/proxy-clients.rst#kerneldispatcher-for-single-server-installations
 - https://foshttpcachebundle.readthedocs.io/en/latest/features/symfony-http-cache.html#optimization-for-single-server-installations


alexander-schranz:
@sulu/core-team it seems this requires at least http cache at version 2.3. another way would be that we do this only in sulu-minimal and requires there at least http cache 2.3 so its also possible to remove this if somebody don't want it?

alexander-schranz:
Think we should do it by default. but yeah its better just do it in sulu-minimal where it can be removed by the people setting up another cache /cc @chirimoya

alexander-schranz:
closed in favor of sulu/sulu-minimal#122

Issue #3798, opened 2018-02-26
Fix default values for FieldTypes


danrot:
| Q | A
| --- | ---
| Bug? | yes
| New Feature? | no
| Sulu Version | 2a875c9234c4c224aa829b92c13b628cf1044ba6
| Browser Version | ---

#### Actual Behavior

We have some fields that need support for default values. Currently this concerns the `SingleSelect` and `ResourceLocator` FieldType. To support default values they [call `onChange`](https://github.com/sulu/sulu/blob/develop/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/SingleSelect.js#L19) at some point in their lifecycle. The problem is that this will set the `ResourceStore` to dirty, and immediately shows the enabled Save button

#### Expected Behavior

Even when a default was added the save button should not be enabled.

#### Steps to Reproduce

1. Go to the contact list
2. Press the add button
3. The save button will be enabled before anything was entered

#### Possible Solutions

Introduce some kind of static function, which can return the default value, and use this values when initializing the `ResourceStore`, so that it can take care of the right `dirty` state itself afterwards.

danrot:
This problem occurs in the following components:

- [ ] BlockCollection
- [ ] Select
- [ ] ResourceLocator

danrot:
The same problem occured on #4129 with the field type for URLs. There the problem is even worse, because when introducing the default value in the known way it somehow ends up in a endless loop.

danrot:
A solution was implemented in #4578, which would work for all fields, and we would not need to touch the fields currently not working.

PR #4129, opened 2018-09-05
Add a URL field type

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | #3798 
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR adds the `url` field type.

#### To Do

- [x] Implement `defaults` from XML template
- [x] Validation
- [ ] Change protocols to not use slash (?)

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Block/tests/__snapshots__/Block.test.js.snap
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Block/tests/__snapshots__/Block.test.js.snap
@@ -105,7 +105,7 @@ exports[`Render an expanded block with a multiple types 1`] = `
           class="select"
         >
           <button
-            class="displayValue"
+            class="displayValue default"
             type="button"
           >
             <div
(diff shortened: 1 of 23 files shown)


danrot:
I think implementing the `defaults` option from [http://docs.sulu.io/en/latest/reference/content-types/url.html] has to wait, because apart from showing the form as dirty immediately it also runs into an endless loop for some reason. I think the reason for that is that the `onChange` call in the constructor has not been done when rendering the component for the first time.

danrot:
The content of this PR should also be considered when fixing #3798

danrot:
@chirimoya The regex seems to get pretty slow, when long domain names are entered. Event the browser on my machine started hanging because of that... Either we say that such long domain names are unrealistic anyway, or we use a simpler (and probably wronger) regex to circumvent this issue.

danrot:
@chirimoya Isn't the error clear? You have not defined any schemes as defined in http://docs.sulu.io/en/latest/reference/content-types/url.html. If you want we can have some default values there, instead of throwing an error. Apart from that the default value in the xml definition is also not really nice to handle :-/

danrot:
@chirimoya I have added some defaults, and fixed two other strange behaviours I have encountered.

PR #4134, opened 2018-09-10
Added input field to pagination.

luca-rath:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | no
| Related issues/PRs | no
| License | MIT

#### What's in this PR?

Added input field to pagination.

#### Why?

To navigate manually to a specific page.

#### To Do

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Pagination/Pagination.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Pagination/Pagination.js
@@ -32,7 +32,7 @@ export default class Pagination extends React.Component<Props> {
     @action componentDidMount() {
         const {currentPage} = this.props;
 
-        if(!currentPage) {
+        if (!currentPage) {
             return;
         }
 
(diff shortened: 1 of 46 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/components/Pagination/Pagination.js
```diff
@@ -68,8 +112,33 @@ export default class Pagination extends React.PureComponent<Props> {
...
+        const page = parseInt(value);
+
+        if (!isNaN(page)) {
+            this.currentInputValue = page;
+        }
+    };
+
+    @action handleInputBlur = () => {
+        const {currentPage, onPageChange} = this.props;
+        const page = this.getPage();
```
danrot: I think this is the only line where we need to find the page in the way the `getPage` function does it, right? I think `getPage` is not a good name, because it is not clear what page it returns (could also be the current value of the input). Why don't we inline the function in here? Then we wouldn't have to find a name, because I can't think of a good one :see_no_evil:

PR #4140, opened 2018-09-11
Fixed missing http cache parameters under dev and test env

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

The parameters `sulu_http_cache.cache.max_age`and `sulu_http_cache.cache.shared_max_age` are currently not available in dev and test environment because of the `shouldCache` method.

#### Why?

The parameters should always be available.

#### Example Usage

Mostly I use the following snippet in my custom static controllers:

~~~php
$response = new Response();
$response->setPublic();
$response->setMaxAge($this->getParameter('sulu_http_cache.cache.max_age'));
$response->setSharedMaxAge($this->getParameter('sulu_http_cache.cache.shared_max_age'));

$response->headers->set(
    AbstractHttpCache::HEADER_REVERSE_PROXY_TTL,
    1209600
);
~~~

#### BC Breaks/Deprecations

This PR fixed an accidently BC Break in 2.0 and make the variables again available under dev and test.

### DIFF
--- a/src/Sulu/Bundle/HttpCacheBundle/DependencyInjection/SuluHttpCacheExtension.php
+++ b/src/Sulu/Bundle/HttpCacheBundle/DependencyInjection/SuluHttpCacheExtension.php
@@ -80,6 +80,9 @@ class SuluHttpCacheExtension extends Extension implements PrependExtensionInterf
         $configuration = $this->getConfiguration($configs, $container);
         $config = $this->processConfiguration($configuration, $configs);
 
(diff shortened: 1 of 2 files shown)


danrot:
Would it be hard to write a test for that? It is quite a subtle change, and I could imagine that this regression would be reintroduced in the future.

alexander-schranz:
@danrot tests added

PR #4143, opened 2018-09-12
Add limit to Datagrid pagination

luca-rath:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This PR adds an limit select to the pagination component.

#### Why?

The user should be able to change the amount of visible items per page.

#### Example Usage

~~~jsx
<Pagination {...otherProps}
    currentLimit={20} 
    onLimitChange={(limit) => {
        alert("The new limit is " + limit);
    }}
/>
~~~


danrot:
That border is definitely too thick:
screenshot from 2018-09-10 09-08-54

src/Sulu/Bundle/AdminBundle/Resources/translations/admin.de.json
```diff
@@ -7,6 +7,7 @@
     "sulu_admin.order": "Sortieren",
     "sulu_admin.of": "von",
     "sulu_admin.page": "Seite",
+    "sulu_admin.per_page": "Einträge pro Seite",
```
danrot: In the old UI it was `Elemente pro Seite` I would stick with that.

src/Sulu/Bundle/MediaBundle/Resources/js/views/MediaOverview/MediaOverview.js
```diff
@@ -33,21 +49,36 @@ class MediaOverview extends React.Component<ViewProps> {
...
         this.createCollectionDatagridStore();
         this.createMediaDatagridStore();
 
         router.bind('search', this.mediaDatagridStore.searchTerm);
+        router.bind('collectionLimit', this.collectionDatagridStore.limit, 10);
+        router.bind('mediaLimit', this.mediaDatagridStore.limit, 10);
+
+        this.collectionLimitDisposer = autorun(
+            () => userStore.setPersistentSetting(getLimitSettingKey(COLLECTIONS_RESOURCE_KEY),
+                this.collectionDatagridStore.limit.get())
```
danrot: Also if the function arguments don't fit in a line, they should get their own:

```javascript
() => userStore.setPersistentSetting(
    getLimitSettingKey(),
    this.collectionDatagridStore.limit.get()
);
```

I also thought we have already linted that, can you check if there is a rule?

src/Sulu/Bundle/MediaBundle/Resources/js/views/MediaOverview/MediaOverview.js
```diff
@@ -174,7 +204,7 @@ export default withToolbar(MediaOverview, function() {
                 onClick: () => {
                     this.clearDatagrids();
                     router.restore(
-                        COLLECTION_ROUTE,
+                        USER_SETTING_PREFIX,
```
danrot: I would keep the `MEDIA_ROUTE` and `USER_SETTING_PREFIX` constants separate. it feels strange to route to a prefix...

src/Sulu/Bundle/AdminBundle/Resources/js/views/Datagrid/Datagrid.js
```diff
@@ -92,11 +99,13 @@ class Datagrid extends React.Component<ViewProps> {
         router.bind('sortColumn', this.datagridStore.sortColumn);
         router.bind('sortOrder', this.datagridStore.sortOrder);
         router.bind('search', this.datagridStore.searchTerm);
+        router.bind('limit', this.datagridStore.limit);
```
danrot: Looks like we have to add the default value here as well... otherwise the initial value gets overriden from the router (you can check that by e.g. opening the snippet form and seeing that nothing is selected in the SingleSelect if you open it).

PR #4144, opened 2018-09-12
Added a check so that properties will only be added to MultipleTransl…

kfrohwein:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

A check that a new translatable property of a document is only added if it was not added before.

#### Why?

603 documents create 1.2 million calls to new Property() and that eats up s lot performance.

Sadly I am not 100% sure if there is any reason you created so many properties without any check. Maybe the namespace could change? I don't know and I can't really test it. But what I do is benchmark with blackfire. And after the check MultipleTranslatedProperties is not even to be seen in the blackfire tree and the same code runs 4 seconds faster.

I don't believe I should use ->getName() because it seems too much to do the same thing but by exception handling.

If it's a stupid idea sorry please delete :)

#### Example Usage

https://blackfire.io/profiles/1eb58276-921d-4e28-b548-f596dbea2267/graph
https://blackfire.io/profiles/98115b82-3381-4edc-a472-0249628a15a7/graph

### DIFF
--- a/src/Sulu/Component/Content/Mapper/Translation/MultipleTranslatedProperties.php
+++ b/src/Sulu/Component/Content/Mapper/Translation/MultipleTranslatedProperties.php
@@ -49,8 +49,10 @@ class MultipleTranslatedProperties
         $this->languageNamespace = $languageNamespace;
         $this->properties = [];
         foreach ($names as $name) {
-            $propertyName = (!empty($namespace) ? $namespace . '-' : '') . $name;
-            $this->properties[$name] = new Property($propertyName, [], 'none', false, true);
+            if (empty($this->properties[$name])) {
+                $propertyName            = (!empty($namespace) ? $namespace.'-' : '').$name;
+                $this->properties[$name] = new Property($propertyName, [], 'none', false, true);
+            }
         }
     }
 
--- a/src/Sulu/Component/Content/Mapper/Translation/MultipleTranslatedProperties.php
+++ b/src/Sulu/Component/Content/Mapper/Translation/MultipleTranslatedProperties.php
@@ -50,7 +50,7 @@ class MultipleTranslatedProperties
         $this->properties = [];
         foreach ($names as $name) {
             if (empty($this->properties[$name])) {
-                $propertyName            = (!empty($namespace) ? $namespace.'-' : '').$name;
+                $propertyName = (!empty($namespace) ? $namespace . '-' : '') . $name;
                 $this->properties[$name] = new Property($propertyName, [], 'none', false, true);
             }
         }
--- a/src/Sulu/Component/Content/Mapper/Translation/MultipleTranslatedProperties.php
+++ b/src/Sulu/Component/Content/Mapper/Translation/MultipleTranslatedProperties.php
@@ -49,10 +49,8 @@ class MultipleTranslatedProperties
         $this->languageNamespace = $languageNamespace;
         $this->properties = [];
         foreach ($names as $name) {
-            if (empty($this->properties[$name])) {
-                $propertyName = (!empty($namespace) ? $namespace . '-' : '') . $name;
-                $this->properties[$name] = new Property($propertyName, [], 'none', false, true);
-            }
+            $propertyName = (!empty($namespace) ? $namespace . '-' : '') . $name;
+            $this->properties[$name] = new Property($propertyName, [], 'none', false, true);
         }
     }
 
--- a/src/Sulu/Bundle/ContentBundle/Content/Structure/ExcerptStructureExtension.php
+++ b/src/Sulu/Bundle/ContentBundle/Content/Structure/ExcerptStructureExtension.php
@@ -247,6 +247,9 @@ class ExcerptStructureExtension extends AbstractExtension implements ExportExten
      */
     private function initProperties($locale)
     {
+        // Reset the properties before new initialization.
+        $this->properties = [];
+
         /** @var PropertyInterface $property */
         foreach ($this->getExcerptStructure($locale)->getProperties() as $property) {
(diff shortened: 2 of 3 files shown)


kfrohwein:
Just to note I did import like 50 times using this patch and it is really faster. I did not seem to have any issues with the imported nodes.

alexander-schranz:
@kfrohwein can you rebase this for the master branch and set the target branch to master?
@danrot can you have a look at it.

src/Sulu/Component/Content/Mapper/Translation/MultipleTranslatedProperties.php
```diff
@@ -49,8 +49,10 @@ public function __construct(
         $this->languageNamespace = $languageNamespace;
         $this->properties = [];
         foreach ($names as $name) {
-            $propertyName = (!empty($namespace) ? $namespace . '-' : '') . $name;
-            $this->properties[$name] = new Property($propertyName, [], 'none', false, true);
+            if (empty($this->properties[$name])) {
```
danrot: Any idea why this might be called multiple times? :confused:

kfrohwein:
> @kfrohwein can you rebase this for the master branch and set the target branch to master?
I tried but it fails a three way merge on the travis config file... Sorry I do currently not know how to do that safely...

kfrohwein:
I did debug this a bit. A simple save of one node creates 49 calls. 

7 excerpt properties are parsed 5 times. After 3 runs they are doubled as
`\Sulu\Bundle\ContentBundle\Content\Structure\ExcerptStructureExtension::initProperties `
runs twice and adds all values again. And this again runs twice probably because of the 2 languages we use. Sorry don't have time to really debug this. Anyways I'd say add the check because it seems to be done by circular events and could happen any time again.

kfrohwein:
Now I messed it up :cry:

src/Sulu/Component/Content/Mapper/Translation/MultipleTranslatedProperties.php
danrot: I've had a deeper look at this, and it feels like you are fixing this in the wrong place. I don't know when exactly you are entering this (for me it was when saving a page int he backend), but when I was having a look at this, it was not being executed multiple times for each name... But from what I have seen, the `$names` array being iterated here contained the value of `created` twice. That's the only name you would skip with this implementation. However, it would make more sense to find out why there are double entries in the `$names` array.

danrot: More context: When I was saving the page, the `ContentQueryBuilder` was called, and the `created` property is listed twice [here](https://github.com/sulu/sulu/blob/develop/src/Sulu/Component/Content/Query/ContentQueryBuilder.php#L56). Does your performance also improve when you if you remove the doubled entry? Or do you have a completely different entry point to this constructor?

danrot: Also, when the language is switched, it looks to me like the [`ExcerptStructureExtension`](https://github.com/sulu/sulu/blob/develop/src/Sulu/Bundle/ContentBundle/Content/Structure/ExcerptStructureExtension.php#L177) adds the properties without clearing it before. It feels like these are the issues we should address. And if we address all of them, we should have the same perfomance improvement as if we do it this way.

kfrohwein: As wrote above those lines are not the problem but more like a late cleanup. Still in my opinion there should be a check because it could happen all the time. If you don't want a check that just ignores the duplication maybe we should here raise an exception?

Otherwise you would never ever know that this happens and only an import of many nodes will really run into that problem.

danrot: Raising an exception actually sounds like a good idea. Then it should also be quite easy to fix all these issues now, shouldn't it? The exception will exactly tell you where you need to fix stuff. Otherwise the duplicated properties might make other problems as well :confused:

kfrohwein:
Finally I have time to really fix it. Sorry it took so long.

http://php.net/manual/en/function.array-merge.php
Merges the elements of one or more arrays together so that the values of one are appended to the end of the previous one. It returns the resulting array.

Sadly now the defaults where added all the time.

So I added
http://php.net/manual/de/function.array-unique.php
Takes an input array and returns a new array without duplicate values.

And that fixes the problem :)

kfrohwein:
There was a second occasion when the multiplication of properties did happen.

`    /**
     * Initiates structure and properties.
     *
     * @param string $locale
     */
    private function initProperties($locale)
    {
        // Reset the properties before new initialization.
        $this->properties = [];

        /** @var PropertyInterface $property */
        foreach ($this->getExcerptStructure($locale)->getProperties() as $property) {
            $this->properties[] = $property->getName();
        }
    }`

In my opinion I can't and should not try to change the whole node creation event system. So I decided that it makes sense that this method resets any existing properties. It will be called multiple times if there are multiple languages anyways.

danrot:
@kfrohwein Is it possible that you've started your branch form an old version? Can you rebase it? I've actually thought that the fail in this one Travis Build was already fixed by us. Just wanna be sure that everything is working on this version as well, since this is a change at the very core :slightly_smiling_face:

danrot:
And there is still the double `created` entry in the `ContentQueryBuilder` class, which is bothering me :see_no_evil: Can you just remove that line as well, if you are already on it? :slightly_smiling_face:

alexander-schranz:
This was fixed in #4304

PR #4146, opened 2018-09-13
Fix preview renderer preview attribute and xml http request

### LINKED ISSUE #4135: var $preview is empty not true on preview mode
| Q | A
| --- | ---
| Bug? | yes
| New Feature? | no
| Sulu Version | 1.6.21

#### Actual Behavior

On preview mode if I do 
```
{% set preview = app.request.attributes.get('preview')|default('') %}
```
```
console.log({{preview}})
```
it's return nothing while it should return `true`

#### Expected Behavior

true should be the value

#### Steps to Reproduce

on a newer install with `DefaultController indexAction`
```php
public function indexAction(StructureInterface $structure, $preview = false, $partial = false)
    {
        
        $response = $this->renderStructure(
            $structure,
            [],
            $preview,
            $partial
        );
        
        return $response;
    }
```
and twig template
```twig
{% extends "ClientWebsiteBundle:views:master.html.twig" %}
{% set preview = app.request.attributes.get('preview')|default('') %}
{% block content %}
<script>
    console.log({{preview}});
</script>
{% endblock %}
```

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #4135
| License | MIT

#### What's in this PR?

Remove the server variable which symfony use the indicate xmlhttprequest and set preview and the other attirbutes correctly.

#### Why?

XMLHttpRequest was true which shouldn't and preview was not longer set.

### DIFF
--- a/src/Sulu/Bundle/PreviewBundle/Preview/Renderer/PreviewRenderer.php
+++ b/src/Sulu/Bundle/PreviewBundle/Preview/Renderer/PreviewRenderer.php
@@ -158,6 +158,9 @@ class PreviewRenderer implements PreviewRendererInterface
         }
 
         $attributes = $this->routeDefaultsProvider->getByEntity(get_class($object), $id, $locale, $object);
+        $attributes['object'] = $object;
+        $attributes['preview'] = true;
+        $attributes['partial'] = $partial;
 
         // get server parameters
         $server = $this->createServerAttributes($portalInformation, $currentRequest);
@@ -261,6 +264,7 @@ class PreviewRenderer implements PreviewRendererInterface
         $server['SERVER_PORT'] = $port;
         $server['HTTP_HOST'] = $httpHost;
         $server['REQUEST_URI'] = $prefixPath . '/_sulu_preview';
+        unset($server['X-Requested-With']);
 
         return $server;
     }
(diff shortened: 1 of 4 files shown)


src/Sulu/Bundle/PreviewBundle/Preview/Renderer/PreviewRenderer.php
```diff
@@ -158,6 +158,9 @@ public function render($object, $id, $webspaceKey, $locale, $partial = false, $t
         }
 
         $attributes = $this->routeDefaultsProvider->getByEntity(get_class($object), $id, $locale, $object);
+        $attributes['object'] = $object;
```
wachterjohannes: will be set by the `RouteDefaultsProvider`

wachterjohannes: can be removed

src/Sulu/Bundle/PreviewBundle/Preview/Renderer/PreviewRenderer.php
```diff
@@ -158,6 +158,9 @@ public function render($object, $id, $webspaceKey, $locale, $partial = false, $t
...
+        $attributes['preview'] = true;
+        $attributes['partial'] = $partial;
```
wachterjohannes: will be used by the `WebsiteController` has to be here

alexander-schranz:
@wachterjohannes test added

alexander-schranz:
@wachterjohannes merge conflicts fixed

PR #4147, opened 2018-09-17
Added root kernel path to be available as document data fixtures paths

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Added root kernel path to be available as document data fixtures paths.

#### Why?

As the src folder is not longer a bundle we need to add `src/DataFixtures/Document` as data fixture folder. 

#### Example Usage

~~~php
<?php

// src/DataFixtures/Document/PageFixture.php

namespace App\DataFixtures\Document;

use Sulu\Bundle\DocumentManagerBundle\DataFixtures\DocumentFixtureInterface;
use Sulu\Component\DocumentManager\DocumentManager;

class PageFixture implements DocumentFixtureInterface
{
    public function getOrder()
    {
        return 10;
    }

    public function load(DocumentManager $documentManager)
    {
        // ...
    }
}
~~~

### DIFF
--- a/src/Sulu/Bundle/DocumentManagerBundle/Command/FixturesLoadCommand.php
+++ b/src/Sulu/Bundle/DocumentManagerBundle/Command/FixturesLoadCommand.php
@@ -19,6 +19,7 @@ use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Question\ConfirmationQuestion;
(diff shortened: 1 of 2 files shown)


alexander-schranz:
ping @danrot

danrot:
Didn't work when I was trying to test it, but I guess that's because this PR is not rebased.

alexander-schranz:
@danrot is rebased

danrot:
Tests are failing quite strange, doesn't feel like it is caused by this test... Will merge it anyway, and hope that it fixes itself :see_no_evil:

PR #4149, opened 2018-09-18
Role add/edit view

trickreich:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes --
| Related issues/PRs | --
| License | MIT
| Documentation PR | --

#### What's in this PR?

* Matrix Component
* Role add/edit view

#### Why?

The user needs to configure the permissions.


src/Sulu/Bundle/SecurityBundle/Resources/js/containers/Permissions/Permissions.js
```diff
@@ -0,0 +1,224 @@
...
+    static webspacePlaceholder = '#webspace#';
+
+    systemDisposer: () => void;
+
+    @observable securityContextGroups: SecurityContextGroups;
+
+    @action componentDidMount() {
+        this.systemDisposer = autorun(
+            () => {
+                securityContextsStore.loadSecurityContextGroups(this.system).then(action((securityContextGroups) => {
```
danrot: Until now we've named the stores in singular (e.g. `WebspaceStore`, `ResourceMetadataStore`, ...)

src/Sulu/Bundle/SecurityBundle/Resources/js/containers/Permissions/Permissions.js
```diff
@@ -0,0 +1,224 @@
...
+
+    @computed get system(): string {
+        return this.props.system;
+    }
+
+    @computed get webspaceContextPermissionPrefix(): string {
+        if (this.webspaceSecurityContextGroupKey) {
+            const securityContextGroup = this.securityContextGroups[this.webspaceSecurityContextGroupKey];
+            for (const securityContextKey of Object.keys(securityContextGroup)) {
+                if (securityContextKey.indexOf(Permissions.webspacePlaceholder) !== -1) {
```
danrot: I think most of the `array.indexOf(value) === -1` calls in this file would be a bit easier to read if they use `array.includes(value)`, since the intention is clearer then (at some points it is hard to distinguish if you really need the index or it is just about finding if subset is there). This also works for strings.

src/Sulu/Bundle/SecurityBundle/Resources/js/containers/Permissions/tests/Permissions.test.js
```diff
@@ -0,0 +1,821 @@
...
+            value={value}
+        />
+    );
+
+    return promise.then(() => {
+        expect(securityContextsStore.loadSecurityContextGroups).toBeCalledWith('Sulu');
+        permissions.update();
+
+        // Currently we have to load each child separately, because of a bug in enzyme.
+        // Fixme: https://github.com/airbnb/enzyme/issues/1213
```
danrot: Would stay with `TODO`s, as we've always done until now, e.g. in https://github.com/sulu/sulu/blob/develop/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Datagrid/tests/Datagrid.test.js#L301

src/Sulu/Bundle/SecurityBundle/Resources/js/containers/Permissions/tests/Permissions.test.js
```diff
@@ -0,0 +1,821 @@
...
```
danrot: Same as above

src/Sulu/Bundle/SecurityBundle/Resources/js/containers/Permissions/types.js
```diff
@@ -0,0 +1,7 @@
+// @flow
+export type ContextPermission = {
+    context: string,
+    id: ?number,
+    permissions: {[string]: boolean},
+};
+
```
danrot: No double new line. Seems like the behavior of eslint changed here in some version, so we also need to set the `maxEOF` option to `0` in the [`no-multiple-empty-lines`](https://eslint.org/docs/rules/no-multiple-empty-lines) rule.

src/Sulu/Bundle/SecurityBundle/Resources/js/stores/SecurityContextsStore/SecurityContextsStore.js
```diff
@@ -0,0 +1,23 @@
+// @flow
+import {Config, Requester} from 'sulu-admin-bundle/services';
+import type {SecurityContextGroups, Systems} from './types';
+
+class SecurityContextsStore {
```
danrot: As mentioned above, `SecurityContextStore` would be more consistent with the current naming scheme.

trickreich:
@danrot Clean up is finished.

PR #4150, opened 2018-09-19
Added .env file generation

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT
| Related PRs | https://github.com/sulu/sulu-minimal/pull/98

#### What's in this PR?

As we currently don't support symfony flex we need manually create the .env file.

#### Why?

Else composer create-project will fail with an error that the consoles don't work.

#### Example Usage

~~~json
"post-create-project-cmd": [
    "Sulu\\Bundle\\CoreBundle\\Composer\\ScriptHandler::copyEnvDistFile"
]
~~~

### DIFF
--- a/src/Sulu/Bundle/CoreBundle/Composer/ScriptHandler.php
+++ b/src/Sulu/Bundle/CoreBundle/Composer/ScriptHandler.php
@@ -11,12 +11,34 @@
 
 namespace Sulu\Bundle\CoreBundle\Composer;
 
+use Composer\Script\Event;
+
 /**
  * A handler for general tasks executed with composer scripts.
  */
 class ScriptHandler
 {
     const GIT_IGNORE_FILE = '.gitignore';
+
+    const ENV_FILE = '.env';
+
+    const ENV_DIST_FILE = '.env.dist';
+
+    /**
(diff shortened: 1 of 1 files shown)


src/Sulu/Bundle/CoreBundle/Composer/ScriptHandler.php
```diff
@@ -17,6 +17,21 @@
...
+     * Copy the .env.dist file to .env file if not exists and generate a app secret.
+     */
+    public static function copyEnvDistFile()
+    {
+        if (file_exists(static::ENV_FILE)) {
+            return;
+        }
+
+        $envFile = file_get_contents(static::ENV_FILE . '.dist');
+        $envFile = str_replace('APP_SECRET=', 'APP_SECRET=' . bin2hex(random_bytes(16)), $envFile);
```
alexander-schranz: inspired by the flex generate(secret)  function: https://github.com/symfony/flex/blob/5e6b06e0f452a895526873fa1a4531d2b209a098/src/Configurator/EnvConfigurator.php#L156

PR #4154, opened 2018-09-19
Fix loading of environment specific configurations

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Remove the `/**/` from the current environment glob loaders to avoid problems with the normal glob function.

#### Why?

Currently no configurations under the environment folder e.g. `/config/packages/dev/web_profiler.yaml` is loaded.

#### Example Usage

Create a environment specific configuration in `/config/packages/<environment>/<yourfile>.yaml`

### DIFF
--- a/src/Sulu/Component/HttpKernel/SuluKernel.php
+++ b/src/Sulu/Component/HttpKernel/SuluKernel.php
@@ -83,7 +83,7 @@ abstract class SuluKernel extends Kernel
         $confDir = $this->getProjectDir() . '/config';
 
         $this->load($loader, $confDir . '/{packages}/*');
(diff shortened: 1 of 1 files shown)


danrot:
Does that really work for you? If I test it, the wdt requests are failing, and after a few requests the toolbar is shown like this:
screenshot from 2018-09-20 12-15-52

And on an unrelated side note: The `.env` file doesn't seem to be copied when `composer create-project` is executed.

danrot:
And if I click on the link in the toolbar, I see the following error:
screenshot from 2018-09-20 12-17-53

PR #4159, opened 2018-09-20
Introduce fieldTypeDefaultProps and restructure TestHelper

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | extracted from #4101
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR introduces the `fieldTypeDefaultProps`, which introduces the most sensible defaults possible. They can be used in all tests for field types to avoid setting the same defaults again and again.

#### Why?

To avoid having to touch every test when a new prop is added to the field types.


src/Sulu/Bundle/AdminBundle/Resources/js/utils/TestHelper/findWithHighOrderFunction.js
```diff
@@ -0,0 +1,22 @@
+// @flow
+import type {Component} from 'react';
+
+export default function findWithHighOrderFunction(
```
danrot: @wachterjohannes In #4101 I have introduced another variable in this namespace, so I change the structure a bit... I have extracted that part, and also renamed the file, as you have done in #3592. I hope @chirimoya merges this soon, then we have less conflicts later :-) Just FYI.

PR #4161, opened 2018-09-21
Bugfix in select boxes in firefox

luca-rath:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Fixed a bug that only occurred in firefox in select boxes. The text wasn't vertically centered.


danrot:
It's better now, but still not completely the same height. In Firefox the select field has now a height of 30.5 px and e.g. the text-line 30 px. So when you put them next to each other, you still see a difference:
screenshot from 2018-09-24 10-44-31

I also realized that the border of the select is darker then the one from the input field. Can you also change the border color of the select field to match the input field?

PR #4168, opened 2018-09-26
Make toolbar responsive

luca-rath:
…space. added horizontal scrolling inside toolbar.

| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Add responsiveness to toolbars. The toolbar items will hide their text if two little space is available. If even less space is available, the toolbar items will be scrollable horizontally.

#### Why?

The toolbar elements are broken up on smaller devices really ugly.


luca-rath:
@danrot

src/Sulu/Bundle/AdminBundle/Resources/js/components/Toolbar/Items.js
```diff
@@ -10,11 +12,60 @@ type Props = {
...
+    componentDidMount() {
+        this.setExpandedWidth();
+        window.addEventListener('resize', this.setExpandedWidth);
+    }
+
+    componentDidUpdate() {
+        this.setExpandedWidth();
+    }
+
+    @action setExpandedWidth = () => {
```
danrot: If we agree on below comment about a computed property we could make this a `setDimensions` function, which is called when something is resized, and `this.setShowText` wouldn't have to be called in here.

danrot:
An idea for our `Sidebar` problem would be to add a `sidebar` prop (naming not ideal) to the `Toolbar`, and when it changes because of the sidebar we know that we have to recalculate if the text should be shown or not.

src/Sulu/Bundle/AdminBundle/Resources/js/components/Toolbar/Items.js
```diff
@@ -10,11 +12,60 @@ type Props = {
...
+export default class Items extends React.Component<Props> {
+    @observable expandedWidth = 0;
+    @observable showText = true;
+
     static defaultProps = {
         skin: 'light',
     };
 
+    parentRef: { current: null | HTMLDivElement } = React.createRef();
+    childRef: { current: null | HTMLUListElement } = React.createRef();
```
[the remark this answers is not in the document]
luca-rath: I was just trying because i wanted to know why it was failing back then so i came up with this. In this case, i knew that i only need to know the 'current' property of the ref, and this could be a HTMLElement or null. It works without any problems.

src/Sulu/Bundle/AdminBundle/Resources/js/components/Toolbar/Items.js
```diff
@@ -12,85 +13,92 @@ type Props = {
...
+const DEBOUNCE_TIME = 200;
+
 @observer
 export default class Items extends React.Component<Props> {
-    @observable expandedWidth = 0;
-    @observable showText = true;
+    @observable
+    expandedWidth: number = 0;
+    @observable
+    parentWidth: number = 0;
```
danrot: The annotation always goes into the same line as the variable

src/Sulu/Bundle/AdminBundle/Resources/js/components/Toolbar/Items.js
```diff
@@ -12,85 +13,92 @@ type Props = {
...
+            return;
         }
 
-        this.setShowText();
-    };
+        resizeObserver.observe(this.parentRef);
+    }
+
+    @action
+    setDimensions = () => {
```
danrot: annotation in the same line

src/Sulu/Bundle/AdminBundle/Resources/js/components/Toolbar/Items.js
```diff
@@ -12,85 +13,92 @@ type Props = {
...
+        const {parentRef} = this;
+        const {childRef} = this;
```
danrot: Can be combined to `const {childRef, parentRef} = this`.

src/Sulu/Bundle/AdminBundle/Resources/js/components/Toolbar/Items.js
```diff
@@ -12,85 +13,92 @@ type Props = {
...
         }
 
-        const parentWidth = this.parentRef.current.offsetWidth;
+        this.parentWidth = parentRef.offsetWidth;
+    };
 
-        if (parentWidth >= this.expandedWidth) {
-            this.showText = true;
+    @computed
+    get showText(): boolean {
```
danrot: Annotation in same line

src/Sulu/Bundle/AdminBundle/Resources/js/components/Toolbar/Items.js
```diff
@@ -12,85 +13,92 @@ type Props = {
...
+        if (this.parentWidth >= this.expandedWidth) {
+            return true;
         } else {
-            this.showText = false;
+            return false;
```
danrot: Could be simplified to `return this.parentWidth >= this.expandedWith;`

src/Sulu/Bundle/AdminBundle/Resources/js/components/Toolbar/Items.js
```diff
@@ -12,85 +13,92 @@ type Props = {
...
         }
-    };
-
-    componentWillUnmount() {
-        window.removeEventListener('resize', this.setExpandedWidth);
```
danrot: Isn't there any deregisteration for the `ResizeObserver` necessary?

src/Sulu/Bundle/AdminBundle/Resources/js/components/Toolbar/Items.js
```diff
@@ -12,85 +13,92 @@ type Props = {
...
-                                    skin: skin,
-                                })}
-                            </li>
-                        );
-                    })}
+            <div className={itemsStyles.itemsContainer} ref={this.setParentRef}>
+                <ul className={itemsClass} ref={this.setChildRef}>
+                    {children &&
+                        React.Children.map(children, (item, index) => {
+                            return (
```
danrot: The return is actually not necessary if you write `(item, index) => (/* JSX */)`.

PR #4169, opened 2018-09-27
Allow symfony 3.4 compatible versions

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Related issues/PRs | #4148, https://github.com/sulu/sulu-minimal/pull/117
| License | MIT

#### What's in this PR?

Allow symfony 3.4 compatible versions and remove unnecessary requirements.

#### Why?

We should support as many versions of a library as possible to make the update to sulu 2.0 easier.


composer.json
```diff
@@ -89,11 +88,11 @@
         "jackalope/jackalope-doctrine-dbal": "^1.2.5",
         "matthiasnoback/symfony-config-test": "^3.0 || ^4.0",
         "matthiasnoback/symfony-dependency-injection-test": "^2.0 || ^3.0",
-        "phpcr/phpcr-shell": "^1.1.1",
+        "phpcr/phpcr-shell": "^1.0",
         "phpdocumentor/reflection-docblock": "~3.1",
         "phpspec/prophecy": "^1.7",
         "phpunit/phpunit": "^6.5",
-        "php-ffmpeg/php-ffmpeg": "^0.13",
+        "php-ffmpeg/php-ffmpeg": ">0.5, <0.14",
```
wachterjohannes: i am no sure if we should make our life´s harder than it could be 🙈

alexander-schranz: it will avoid conflicts if somebody using the ffmpeg bundle

danrot: I don't really get why we have to care about the bundle here. If the bundle gets installed, it should have the correct requirements for itself, shouldn't it?

alexander-schranz: @danrot the requirements of the bundle conflicts with the requirements we set currently (0.13)  as the bundle only have ( 0.5, 0.6 ) in its [composer.json](https://github.com/pulse00/ffmpeg-bundle/blob/master/composer.json#L15). as we could also support 0.5 as there where no relevant changes for us we could support more versions of ffmpeg as currently set in the composer.json.

alexander-schranz: as discussed with @danrot I currently remove the compatibility to ffmpeg bundle

PR #4173, opened 2018-09-27
Improve phpdoc types  for phpstan and use AuditableTrait

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Improve phpdoc types for phpstan of Documents and Entities and use AuditableTrait where possible.

#### Why?

False return types make phpstan errors.


danrot:
@alexander-schranz I've got the feeling quite some tests are failing because of these changes :confused:

PR #4178, opened 2018-09-29
Make navigation pinnable

luca-rath:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This PR adds a pin button to the navigation. When the navigation is pinned it doesn't get closed automatically when navigation to another route or clicking the backdrop.

#### Why?

On big screens, it's very annoying to always have to reopen the navigation after navigating, although it would actually not be necessary.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Navigation/Navigation.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Navigation/Navigation.js
@@ -4,6 +4,8 @@ import type {ChildrenArray, Element} from 'react';
 import {observer} from 'mobx-react';
 import {action, observable} from 'mobx';
 import Icon from '../Icon';
+import Button from '../Button';
+import ButtonGroup from '../ButtonGroup';
 import Item from './Item';
 import navigationStyles from './navigation.scss';
 
@@ -13,6 +15,7 @@ type Props = {
     username: string,
     userImage: ?string,
     onLogoutClick: () => void,
+    onPinClick?: () => void,
     onProfileClick: () => void,
     suluVersion: string,
     suluVersionLink: string,
@@ -93,6 +96,14 @@ export default class Navigation extends React.Component<Props> {
         );
     }
 
+    handlePinClick = () => {
+        const {onPinClick} = this.props;
+
+        if (onPinClick) {
+            onPinClick();
+        }
+    };
+
     renderAppVersion() {
         const {
             title,
@@ -119,6 +130,7 @@ export default class Navigation extends React.Component<Props> {
             onProfileClick,
             suluVersion,
             suluVersionLink,
+            onPinClick,
         } = this.props;
 
         return (
@@ -141,10 +153,21 @@ export default class Navigation extends React.Component<Props> {
                 <div className={navigationStyles.items}>
                     {this.cloneChildren()}
                 </div>
-                <div className={navigationStyles.versions}>
-                    {this.renderAppVersion()}
-                    <div>
-                        Sulu (<a href={suluVersionLink} rel="noopener noreferrer" target="_blank">{suluVersion}</a>)
+                <div className={navigationStyles.footer}>
+                    {onPinClick &&
+                        <div className={navigationStyles.pinContainer}>
+                            <ButtonGroup>
+                                <Button className={navigationStyles.pin} onClick={this.handlePinClick}>
+                                    <Icon name="fa-thumb-tack" />
+                                </Button>
+                            </ButtonGroup>
+                        </div>
+                    }
+                    <div className={navigationStyles.versions}>
+                        {this.renderAppVersion()}
+                        <div>
+                            Sulu (<a href={suluVersionLink} rel="noopener noreferrer" target="_blank">{suluVersion}</a>)
+                        </div>
                     </div>
                 </div>
             </div>
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Application/Application.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Application/Application.js
@@ -1,7 +1,7 @@
 // @flow
 import './global.scss';
 import {observer} from 'mobx-react';
-import {action, observable} from 'mobx';
+import {action, observable, autorun} from 'mobx';
 import classNames from 'classnames';
 import React, {Fragment} from 'react';
 import Navigation from '../Navigation';
@@ -15,6 +15,13 @@ import {Backdrop} from '../../components';
 import Login from '../Login';
 import applicationStyles from './application.scss';
 
+const USER_SETTING_PREFIX = 'sulu_admin.application';
+const USER_SETTING_NAVIGATION_PINNED = 'navigation_pinned';
+
+function getNavigationPinnedSettingKey() {
+    return USER_SETTING_PREFIX + '.' + USER_SETTING_NAVIGATION_PINNED;
+}
+
 type Props = {
     router: Router,
 };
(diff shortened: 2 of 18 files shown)


luca-rath:
@danrot

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Application/Application.js
```diff
@@ -15,22 +15,68 @@ import {Backdrop} from '../../components';
...
+    componentWillUnmount() {
+        this.navigationPinnedDisposer();
+    }
+
     handleNavigationButtonClick = () => {
         this.toggleNavigation();
     };
 
+    handleNavigationPin = () => {
+        this.pinNavigation();
```
danrot: Is it a good idea to call this function without an argument? What if you pass a `pinned` parameter to the `handleNavigationPin` callback, and then call the correct `pinNavigation` or `unpinNavigation` method in here?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Application/application.scss
```diff
@@ -22,6 +22,12 @@
     width: 100vw;
     min-height: 500px;
     overflow: auto;
+    will-change: width;
+    transition: width $navigationMoveAnimationDuration;
+
+    &.with-pinned-navigation {
+        width: calc(100vw - 260px);
```
danrot: Can't we get the `260px` value somehow from a Navigation scss variable?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Toolbar/types.js
```diff
@@ -14,6 +14,7 @@ export type ToolbarProps = {
     storeKey?: string,
     onNavigationButtonClick?: () => void,
     navigationOpen?: boolean,
+    navigationPinned?: boolean,
```
danrot: I think from a `Toolbar` point of view, it doesn't matter if the navigation is pinned or not, just if the navigation should be shown or not. And until now we have defined that by checking if something has been passed to the `onNavigationButtonClick` prop. So instead of introducing this boolean, I would not pass a callback to the `onNavigationButtonClick` prop if the navigation is pinned.

src/Sulu/Bundle/AdminBundle/Resources/js/components/Navigation/Navigation.js
```diff
@@ -13,6 +15,7 @@ type Props = {
     username: string,
     userImage: ?string,
     onLogoutClick: () => void,
+    onPinClick?: () => void,
```
danrot: Maybe `onPinToggle` and pass the value?

src/Sulu/Bundle/AdminBundle/Resources/js/components/Navigation/Navigation.js
```diff
@@ -141,10 +153,21 @@ export default class Navigation extends React.Component<Props> {
...
                 </div>
-                <div className={navigationStyles.versions}>
-                    {this.renderAppVersion()}
-                    <div>
-                        Sulu (<a href={suluVersionLink} rel="noopener noreferrer" target="_blank">{suluVersion}</a>)
+                <div className={navigationStyles.footer}>
+                    {onPinClick &&
+                        <div className={navigationStyles.pinContainer}>
+                            <ButtonGroup>
+                                <Button className={navigationStyles.pin} onClick={this.handlePinClick}>
```
danrot: I would have imagined this button to stay active - i.e. same style as on hovered - when the navigation is pinned, so that it is easy recognizable if the navigation is pinned or not.

danrot: And is it possible that the button in the design is quite a bit smaller?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Toolbar/Toolbar.js
```diff
@@ -110,15 +111,16 @@ export default class Toolbar extends React.Component<*> {
...
                         type="success"
                         visible={this.toolbarStore.showSuccess}
                     />
                 }
                 <ToolbarComponent.Controls grow={true}>
                     {!!onNavigationButtonClick &&
                     <ToolbarComponent.Button
-                        icon={navigationOpen ? 'su-times' : 'su-bars'}
+                        disabled={navigationPinned}
+                        icon={navigationPinned ? 'su-sulu' : (navigationOpen ? 'su-times' : 'su-bars')}
```
danrot: Did you discuss that the Sulu Icon is shown if the navigation is pinned with somebody? Can understand that it looks a bit boring without the blue button, but on the other side it also takes quite some place from the toolbar, which might be better used in different ways (e.g. showing text for the other buttons instead of collapsing).

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Application/Application.js
```diff
@@ -15,22 +15,68 @@ import {Backdrop} from '../../components';
...
+
+    navigationPinnedDisposer: () => void;
+
+    constructor() {
+        super();
+
+        this.pinNavigation(!!userStore.getPersistentSetting(getNavigationPinnedSettingKey()));
+
+        this.navigationPinnedDisposer = autorun(
+            () => userStore.setPersistentSetting(getNavigationPinnedSettingKey(), this.navigationPinned ? 1 : 0)
```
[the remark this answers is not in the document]
luca-rath: The UserStore only allows String or Number values. @trickreich said that he was extending the UserStore at that time and i wanted to wait until he is finished.

src/Sulu/Bundle/AdminBundle/Resources/js/components/Navigation/Navigation.js
```diff
@@ -141,10 +153,21 @@ export default class Navigation extends React.Component<Props> {
...
                     {this.cloneChildren()}
                 </div>
-                <div className={navigationStyles.versions}>
-                    {this.renderAppVersion()}
-                    <div>
-                        Sulu (<a href={suluVersionLink} rel="noopener noreferrer" target="_blank">{suluVersion}</a>)
+                <div className={navigationStyles.footer}>
+                    {onPinClick &&
+                        <div className={navigationStyles.pinContainer}>
+                            <ButtonGroup>
```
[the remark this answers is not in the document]
luca-rath: In the styleguide is an example of the button group with just one button and the styling of the needed pin button matches more with a single button in a button group than with a normal button with skin="icon". Why else is there an example with only a single button in the button group in the styleguide?

src/Sulu/Bundle/AdminBundle/Resources/js/components/Button/Button.js
```diff
@@ -58,14 +57,20 @@ export default class Button extends React.PureComponent<Props> {
...
+        }
+
         const buttonClass = classNames(
             buttonStyles.button,
             buttonStyles[skin],
             {
                 [buttonStyles.loading]: loading,
                 [buttonStyles.active]: active,
-                [activeClassName]: active,
             },
```
danrot: Instead of the above if you could make use of the `classNames` functionality here:
```javascript
classNames({
    [activeClassName]: active && activeClassName,
});
```

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Application/Application.js
```diff
@@ -23,69 +22,71 @@ type Props = {
...
-    componentWillUnmount() {
-        this.navigationPinnedDisposer();
+        this.navigationPinned = !this.navigationPinned;
     }
 
     handleNavigationButtonClick = () => {
-        if (!this.navigationPinned) {
-            this.toggleNavigation();
+        if (this.navigationPinned) {
+            log.warn('Clicking navigation button is not allowed while navigation is pinned!');
```
danrot: Is that necessary here? The same warning would be logged when `toggleNavigation` below is called, because it is written a few lines above as well, wouldn't it?

luca-rath: This warning tells the developer that he displays the navigation button when it shouldn't be shown. The other warning tells the developer that e.g. some event tries to toggle the navigation. But i can remove it if you want.

src/Sulu/Bundle/AdminBundle/Resources/js/components/Button/Button.js
luca-rath: Flow tells me that i cannot assign computed property using undefined. This was the reason why i set the prop to required and the default value to an empty string the first time.

luca-rath: Found another way to solve it.

luca-rath:
@danrot

PR #4175, opened 2018-09-28
Refactored the translator to not use the request-analyzer

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #issuenum
| Related issues/PRs | #issuenum
| License | MIT
| Documentation PR | sulu/sulu-docs#prnum

#### What's in this PR?

This PR removes the dependency of translator to request_analyzer.

#### Why?

This solves one of the problems that the `cache:warmup` uses the entity-manager and crashes.

PR #4180, opened 2018-10-02
fix datetime formatting for Symfony regression

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | symfony/symfony#28372
| License | MIT
| Documentation PR | ---

#### What's in this PR?

The `DateTimeType` of the Symfony Form component doesn't accept any timezones anymore. Our datetime field doesn't add a timezone anyway, but the server should not return the timezone as well, because if the field is not edited, we don't update the value, and the validation will fail as well.

We might also have to backport this change to 1.x, because Symfony 3.4.16 should have the same bug, but doesn't work currently because of symfony/symfony#28658.

#### Why?

The tests were failing, and if a datetime field was in the form, it couldn't be sent to the server without an error without editing it first.

#### BC Breaks/Deprecations

The datetimes in a REST API do not contain the timezone anymore.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -2,6 +2,10 @@
 
 ## dev-develop
 
+### DateTime Serialization
+
+DateTimes in a REST response do not contain the timezone anymore.
+
 ### Websocket
 
 The websocket-bundle and component was removed without replacement.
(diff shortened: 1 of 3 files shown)


danrot:
If you check and merge that tomorrow morning I can merge #4175 :wink:

PR #4194, opened 2018-10-10
Make labels clickable for form fields

luca-rath:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

Adds an 'id' property to inline form fields and the according 'for' property to the label.

#### Why?

To make the label clickable.


src/Sulu/Bundle/AdminBundle/Resources/js/components/Form/Field.js
```diff
@@ -11,6 +11,7 @@ type Props = {|
     children: Node,
     description?: string,
     error?: string,
+    inputId?: string,
```
danrot: Why don't we just call it id as well? Maybe it is used for other fields than inputs one day.

src/Sulu/Bundle/AdminBundle/Resources/js/components/Form/Field.js
```diff
@@ -41,7 +42,14 @@ export default class Field extends React.Component<Props> {
                 spaceAfter={spaceAfter}
             >
                 <div className={fieldClass}>
-                    {label && <label className={fieldStyles.label}>{label}{required && ' *'}</label>}
+                    {label &&
+                    <label
+                        className={fieldStyles.label}
+                        htmlFor={inputId || undefined}
```
danrot: Can't you just put `inputId`? Would also return `undefined` if it is not given.

src/Sulu/Bundle/AdminBundle/Resources/js/components/Input/Input.js
```diff
@@ -136,6 +137,7 @@ export default class Input<T: ?string | ?number> extends React.PureComponent<Inp
 
                     <input
                         className={inputClass}
+                        id={id || undefined}
```
danrot: Same as above, `|| undefined` should not be necessary.

src/Sulu/Bundle/AdminBundle/Resources/js/components/MultiAutoComplete/MultiAutoComplete.js
```diff
@@ -180,6 +182,7 @@ export default class MultiAutoComplete extends React.Component<Props> {
                         ))}
                         <input
                             className={inputClass}
+                            id={id || undefined}
```
danrot: Same as above

src/Sulu/Bundle/AdminBundle/Resources/js/components/TextArea/TextArea.js
```diff
@@ -51,6 +53,7 @@ export default class TextArea extends React.PureComponent<Props> {
             <Fragment>
                 <textarea
                     className={textareaClass}
+                    id={id || undefined}
```
danrot: Same as above

src/Sulu/Bundle/AdminBundle/Resources/js/components/Url/Url.js
```diff
@@ -171,6 +172,7 @@ export default class Url extends React.Component<Props> {
                     </SingleSelect>
                 </div>
                 <input
+                    id={id || undefined}
```
danrot: Same as above

PR #4207, opened 2018-10-15
Hotfix - redirect pages with trailing slashes were missing query params

### LINKED ISSUE #4103: 301 Redirect urls with trailing slash and query parameters
| Q | A
| --- | ---
| Bug? | yes
| New Feature? | no
| Sulu Version | 1.6.21
| Browser Version | Chrome Version 68.0.3440.106

#### Actual Behavior

Urls ending with a slash are redirected to urls without a slash. For example the url https://sulu.rocks/en/contact/ is redirected with 301 to https://sulu.rocks/en/contact. 

When the url that ends with a slash has query parameters so https://sulu.rocks/en/contact/?source=google it is also redirected with a 301 to https://sulu.rocks/en/contact without the query parameters.

#### Expected Behavior

I expect urls ending with a slash to be redirected to urls without a slash but the query parameters should me maintained. 

For example https://sulu.rocks/en/contact/?source=google should be redirected with a 301 to https://sulu.rocks/en/contact?source=google

#### Steps to Reproduce

See: https://sulu.rocks/en/contact/?source=google

debreczeniandras:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #4103 
| Related issues/PRs | n/a
| License | MIT
| Documentation PR | n/a

#### What's in this PR?

Fixes  redirects that contain query parameters.

#### To Do

- [x] Fix tests

### DIFF
--- a/src/Sulu/Bundle/WebsiteBundle/Tests/Unit/Sulu/Bundle/WebsiteBundle/Routing/ContentRouteProviderTest.php
+++ b/src/Sulu/Bundle/WebsiteBundle/Tests/Unit/Sulu/Bundle/WebsiteBundle/Routing/ContentRouteProviderTest.php
@@ -30,6 +30,7 @@ use Sulu\Component\DocumentManager\Behavior\Mapping\UuidBehavior;
 use Sulu\Component\DocumentManager\DocumentManagerInterface;
 use Sulu\Component\DocumentManager\Metadata;
 use Sulu\Component\Localization\Localization;
+use Sulu\Component\Webspace\Analyzer\Attributes\RequestAttributes;
 use Sulu\Component\Webspace\Analyzer\RequestAnalyzer;
 use Sulu\Component\Webspace\Analyzer\RequestAnalyzerInterface;
 use Sulu\Component\Webspace\Portal;
@@ -727,7 +728,7 @@ class ContentRouteProviderTest extends \PHPUnit_Framework_TestCase
 
     public function testGetCollectionTrailingSlashWithQueryParams()
     {
-        $attributes = $this->prophesize('Sulu\Component\Webspace\Analyzer\Attributes\RequestAttributes');
+        $attributes = $this->prophesize(RequestAttributes::class);
         $portal = new Portal();
         $portal->setKey('portal');
         $webspace = new Webspace();
--- a/src/Sulu/Bundle/WebsiteBundle/Tests/Unit/Sulu/Bundle/WebsiteBundle/Routing/ContentRouteProviderTest.php
+++ b/src/Sulu/Bundle/WebsiteBundle/Tests/Unit/Sulu/Bundle/WebsiteBundle/Routing/ContentRouteProviderTest.php
@@ -724,7 +724,7 @@ class ContentRouteProviderTest extends \PHPUnit_Framework_TestCase
         $this->assertEquals('SuluWebsiteBundle:Redirect:redirect', $route->getDefaults()['_controller']);
         $this->assertEquals('de/qwertz', $route->getDefaults()['url']);
     }
-    
+
     public function testGetCollectionTrailingSlashWithQueryParams()
     {
         $attributes = $this->prophesize('Sulu\Component\Webspace\Analyzer\Attributes\RequestAttributes');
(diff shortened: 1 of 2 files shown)


debreczeniandras:
I was trying to write tests for this action, but it turns out there are already some traces there:

https://github.com/sulu/sulu/blob/b03edb75657d4141e2b294d06465896ac3c301aa/src/Sulu/Bundle/WebsiteBundle/Tests/Unit/Sulu/Bundle/WebsiteBundle/Controller/RedirectControllerTest.php#L76

but for the redirect data set the `RedirectController::redirectWebspaceAction` is used.
Upon changing that to the _correct_ redirectAction the tests obviously fail: this redirect action takes no interest in resourceLocatorPrefix and portalUrls.

Even if I use something like this for the redirectAction: 

``` php 
 $url = $this->resolveRedirectUrl(
            $request->get('redirect'),
            $request->getUri(),
            $request->get('_sulu')->getAttribute('resourceLocatorPrefix')
        );
```
It still fails, because there are some test cases where this action would also need the portalUrl.


Please advise:
* should I write new tests only for checking the query strings?
* should I rewrite the redirectAction based on redirectWebspaceAction (and use the _sulu `RequestAttributes` parameters for the `resolveRedirectUrl()` method)?

debreczeniandras:
As a long-term solution my suggestion would be to extract  `RedirectController::resolveRedirectUrl()`, since
* is completely static, 
* it is redundant: an almost identical method can be found here: in src/Sulu/Bundle/WebsiteBundle/EventListener/RedirectExceptionSubscriber.php.

and then use the request _sulu attribute as reference for the redirect.

As it is used currently in the WebsiteBundle it could

src/Sulu/Bundle/WebsiteBundle/Tests/Unit/Sulu/Bundle/WebsiteBundle/Routing/ContentRouteProviderTest.php
```diff
@@ -680,6 +680,53 @@ public function testGetCollectionTrailingSlash()
...
+
+        $this->resourceLocatorStrategy->loadByResourceLocator('/foo', 'webspace', 'de')->willReturn('some-uuid');
+
+        $redirectTargetDocument = $this->prophesize(ResourceSegmentBehavior::class);
+        $redirectTargetDocument->getResourceSegment()->willReturn('/foo');
+
+        $document = $this->prophesize(TitleBehavior::class)
+                         ->willImplement(RedirectTypeBehavior::class)
+                         ->willImplement(StructureBehavior::class)
+                         ->willImplement(UuidBehavior::class);
```
danrot: We don't align items like that, because that means we have change the alignment also when the variable name above changes. So please indent the above three lines only with 4 spaces.

debreczeniandras: Fixed. Sorry about that. 
Needless to say, PHPStorm does that automatically. 

Unfortunately, I couldn't find yet a way to import the .styleci.yml into my IDE. Perhaps if you could also post a sample Idea XML or JSCS code style file to the contribution guidelines, then we wouldn't run into issues like that.

debreczeniandras:
How does the backporting work?
Should I open a new PR against the 1.5 branch?

src/Sulu/Bundle/WebsiteBundle/Tests/Unit/Sulu/Bundle/WebsiteBundle/Routing/ContentRouteProviderTest.php
danrot: You are completely right. I would also be in favor of using PHP-CS-Fixer directly, instead of relying on StyleCI, then everybody could have the checks locally as well :confused: /cc @alexander-schranz That's something that could be done on hacktoberfest, right? :smile:

danrot:
@debreczeniandras There are different ways: Either you [rebase](https://git-scm.com/book/en/v1/Git-Branching-Rebasing), or you create a new branch and copy over the changes manually. The easiest way is probably to create a new PR.

debreczeniandras:
@danrot Now rebased.

src/Sulu/Bundle/WebsiteBundle/Tests/Unit/Sulu/Bundle/WebsiteBundle/Routing/ContentRouteProviderTest.php
```diff
@@ -725,6 +725,50 @@ public function testGetCollectionTrailingSlash()
         $this->assertEquals('de/qwertz', $route->getDefaults()['url']);
     }
 
+    public function testGetCollectionTrailingSlashWithQueryParams()
+    {
+        $attributes = $this->prophesize('Sulu\Component\Webspace\Analyzer\Attributes\RequestAttributes');
```
debreczeniandras: I'm using FQN here on purpose to avoid adding a `use` operator on the top of the page, and thus easing the merge.

alexander-schranz: always use  the `::class` and import the class with `use` it helps later if you need to refractor code or find usages of a class by IDEs.

debreczeniandras: Ok, done. I thought to impose as little changes as possible.

danrot:
Thanks a lot! :smiley:

PR #4215, opened 2018-10-18
Remove unneeded sensio extra framework bundle

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Remove unneeded sensio extra framework bundle.

#### Why?

Sulu is not related on the sensio extra framework bundle and sensio extra framework bundle is also not. Most used features like route annotations are part of symfony core now also thats way its mostly not longer used in symfony 4 projects and not part of symfony standard installation.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -46,7 +46,6 @@
         "pagerfanta/pagerfanta": "^1.0.4",
         "piwik/device-detector": "^3.7",
         "ramsey/uuid": "^3.1",
-        "sensio/framework-extra-bundle": "^5.0",
         "stof/doctrine-extensions-bundle": "^1.2.2",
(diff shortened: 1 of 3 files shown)


danrot:
Looks like the tests are still failing... Do we maybe have to raise the FOSRestBundle version, until they have correctly added this as a dependency?

danrot:
Searched for occurences of `FrameworkExtraBundle`, and found another one, that should probably be fixed (it's wrongly used here): https://github.com/sulu/sulu/blob/develop/src/Sulu/Bundle/SecurityBundle/Tests/Unit/EventListener/SuluSecurityListenerTest.php#L17

PR #4225, opened 2018-10-24
Add table to ckeditor

luca-rath:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This PR adds functionality and styling for tables in the ckeditor component.


package.json
```diff
@@ -21,6 +21,8 @@
         "sulu-security-bundle": "file:src/Sulu/Bundle/SecurityBundle/Resources/js"
     },
     "devDependencies": {
+        "@ckeditor/ckeditor5-dev-utils": "^11.0.1",
+        "@ckeditor/ckeditor5-theme-lark": "^11.1.0",
```
danrot: These dependencies should probably go as usual dependencies into the `AdminBundle`. This way it does not work if you only install the packages in `sulu-minimal`.

src/Sulu/Bundle/AdminBundle/Resources/js/components/CKEditor5/ckeditor5.css
```diff
@@ -0,0 +1,176 @@
...
+    --persianRed: #cf3939;
+    --pickledBluewood: #293650;
+    --roman: #d96161;
+    --scorpion: #585858;
+    --shakespeare: #52b6ca;
+    --silver: #ccc;
+    --silverChalice: #a2a2a2;
+    --stormGray: #70768b;
+    --white: #fff;
+    --wildSand: #f5f5f5;
```
danrot: That's IMO a no-go... Can't we use the colors we have defined in `colors.scss` instead?

src/Sulu/Bundle/AdminBundle/Resources/js/components/CKEditor5/ckeditor5.css
```diff
@@ -0,0 +1,176 @@
...
+
+    /* Overrides the default font size in the theme. */
+    --ck-font-size-base: 12px;
+    --ck-icon-size: 12px;
+
+    /* Helper variables to avoid duplication in the colors. */
+    --ck-custom-background: var(--silver);
+    --ck-custom-foreground: var(--dustyGray);
+    --ck-custom-border: var(--ck-custom-background);
+    --ck-custom-white: var(--white);
```
danrot: Have we defined this variables on our own?

src/Sulu/Bundle/AdminBundle/Resources/js/components/CKEditor5/ckeditor5.css
```diff
@@ -0,0 +1,176 @@
...
+    --ck-color-button-action-background: var(--shakespeare);
+    --ck-color-button-action-hover-background: var(--ck-color-button-action-background);
+    --ck-color-button-action-active-background: var(--ck-color-button-action-background);
+    --ck-color-button-action-active-shadow: none;
+    --ck-color-button-action-disabled-background: var(--ck-color-button-action-background);
+    --ck-color-button-action-text: var(--ck-custom-white);
+
+    /* Button Save and Cancel */
+    --ck-color-button-save: hsl(120, 100%, 46%);
+    --ck-color-button-cancel: hsl(15, 100%, 56%);
```
danrot: Do we need these buttons at all? :thinking:

src/Sulu/Bundle/AdminBundle/Resources/js/components/CKEditor5/ckeditor5.css
```diff
@@ -0,0 +1,176 @@
...
+
+    /* -- Overrides the default colors used by the ckeditor5-widget package. -------------------- */
+
+    --ck-color-widget-blurred-border: var(--nobel);
+    --ck-color-widget-hover-border: var(--nobel);
+    --ck-color-widget-editable-focus-background: var(--ck-custom-white);
+
+    /* -- Overrides the default colors used by the ckeditor5-link package. ---------------------- */
+
+    --ck-color-link-default: hsl(190, 100%, 75%);
```
danrot: For what reasons do we use `hsl` here? Are these the default values? Do we have to write the default values? If not I would omit them. Having a lot of code that is not required can lead to very confusing situations...

src/Sulu/Bundle/AdminBundle/Resources/js/components/Grid/grid.scss
```diff
@@ -2,7 +2,7 @@ $itemMarginRight: 20px;
 
 .grid {
     margin-right: -$itemMarginRight;
-    overflow: hidden;
+    overflow: visible;
```
danrot: There was actually a reason for it to be `hidden`... Why did you change it?

danrot (changes requested):

src/Sulu/Bundle/AdminBundle/Resources/js/components/Grid/grid.scss
luca-rath: if the text editor is the last item in the grid, that dropdown would be cut off. why should it be hidden?
screenshot

danrot: Hm, strange, I thought this was necessary to have not too much space at the end of a form... But it seems like this problem was solved in some other way in the mean time...

src/Sulu/Bundle/AdminBundle/Resources/js/components/CKEditor5/ckeditor5.scss
```diff
@@ -1,60 +1,79 @@
...
+
+        /* shadows */
+        --ck-color-shadow-inner: none;
+        --ck-color-shadow-drop: none;
+        --ck-color-button-default-active-shadow: none;
+        --ck-color-button-on-active-shadow: none;
+
+        /* toolbar */
+        --ck-ui-component-min-height: 20px;
+        --ck-color-toolbar-background: $silver;
```
danrot: I would still create variables like you see at the top of the old version of this file. We have decided to do this in order to let people override these styles later. This also applies to all other color references in this file.

webpack.config.js
```diff
@@ -6,7 +6,7 @@ const webpack = require('webpack');
 const CleanObsoleteChunksPlugin = require('webpack-clean-obsolete-chunks');
 const ManifestPlugin = require('webpack-manifest-plugin');
 const MiniCssExtractPlugin = require('mini-css-extract-plugin');
-
```
danrot: Please keep this new line. This way it is easier to distinguish between imports and other code.

danrot (changes requested):

danrot:
The failing tests should also be fixed with a rebase.

webpack.config.js
luca-rath: i thought, that this would be fixed automatically when merging, because there are no conflicts. but i'll do the rebase.

luca-rath:
i thought, that this would be fixed automatically when merging, because there are no conflicts. but i'll do the rebase.

danrot:
@luca-rath It probably would, but this way I don't know if your changes make the tests fail :wink:

luca-rath:
@danrot ok. now it's rebased.

danrot (approved):

PR #4226, opened 2018-10-24
Added namespaces to phpunit test for support symfony phpunit bridge

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | yes
| Deprecations? | no
| License | MIT

#### What's in this PR?

Added namespaces to phpunit test for support symfony phpunit bridge.

#### Why?

Cleanup and support phpunit bridge features like parallel testsuites in future.

#### Example Usage

~~~php
bin/simple-phpunit src/Sulu/Bundle
~~~

#### BC Breaks/Deprecations

`KERNEL_DIR` need to be replaced with `KERNEL_CLASS` environment variable in your test setup.

#### To Do


composer.json
```diff
@@ -146,11 +146,7 @@
     "autoload": {
         "psr-0": {
             "Sulu\\": "src/"
-        },
-        "exclude-from-classmap": [
-            "src/Sulu/Component/*/Tests/",
-            "src/Sulu/Bundle/*/Tests/"
```
danrot: Are you sure we should remove that? Doesn't that slow down everything in production?

src/Sulu/Bundle/ContentBundle/Tests/Functional/Controller/NodeControllerTest.php
```diff
@@ -2089,7 +2089,7 @@ public function testInternalLinkAutoName()
     public function testRenamePageWithLinkedChild()
     {
         $client = $this->createAuthenticatedClient();
-        $this->importer->import(__DIR__ . '/../../app/Resources/exports/tree.xml');
+        $this->importer->import(__DIR__ . '/../../files/exports/tree.xml');
```
danrot: Maybe fixtures instead of files? Or is that a bad idea? :thinking:

src/Sulu/Bundle/ContentBundle/phpunit.xml.dist
```diff
@@ -18,8 +18,7 @@
     </filter>
 
     <php>
-        <server name="KERNEL_DIR" value="Tests/app"/>
-        <var name="APP_DB" value="mysql"/>
```
danrot: Is that the default value? Or why can we remove that everywhere?

src/Sulu/Bundle/TestBundle/Resources/app/config/sulu.yml
```diff
@@ -13,25 +13,6 @@ monolog:
...
-             paths:
-                 pages:
-                     path: '%kernel.root_dir%/Resources/pages'
-                     type: page
-                 homepage:
-                     path: '%kernel.root_dir%/Resources/pages'
-                     type: home
-                 templates:
-                     path: '%kernel.root_dir%/Resources/snippets'
-                     type: snippet
```
danrot: Where is that gone? Is that prepended now?

src/Sulu/Bundle/TestBundle/Testing/KernelTestCase.php
```diff
@@ -76,7 +76,9 @@ protected function getKernel(array $options = [])
             'sulu_context' => 'admin',
         ], $this->getKernelConfiguration(), $options);
 
-        $kernel = new \AppKernel(
+        $kernelClass = getenv('KERNEL_CLASS') ?: '\AppKernel';
```
danrot: Is the `KERNEL_CLASS` also what Symfony is doing now? Resp. is that required for the symfony bridge?

src/Sulu/Bundle/TestBundle/Testing/SuluTestCase.php
```diff
@@ -26,7 +26,7 @@
 use Symfony\Component\Security\Core\Tests\Authentication\Token\TestUser;
 
 /**
- * Base test case for functional tests in Sulu.
+ * @deprecated use Sulu\Bundle\TestBundle\TestCase\SuluTestCase instead
```
danrot: Shouldn't we get rid of that class now, if we want to move it? Resp. why do we keep the `KernelTestCase` in this directory, but move the other one?

src/Sulu/Component/Webspace/Manager/WebspaceManager.php
```diff
@@ -319,7 +319,9 @@ public function getWebspaceCollection()
                 );
             }
 
-            require_once $cache->getPath();
+            if (!class_exists($class)) {
```
danrot: Why is that suddenly necessary? Feels a bit like hiding a bug...

src/Sulu/Bundle/TestBundle/TestCase/SuluTestCase.php
```diff
@@ -0,0 +1,86 @@
+<?php
```
danrot: Add this class later :-)

src/Sulu/Bundle/ContentBundle/phpunit.xml.dist
alexander-schranz: @danrot `APP_DB` and could not find anything about this var

src/Sulu/Bundle/TestBundle/Resources/app/config/sulu.yml
alexander-schranz: it now use the default configuration `%kernel.root_dir%/config/templates/snippets` thats why the folder was also moved

src/Sulu/Bundle/TestBundle/Testing/KernelTestCase.php
alexander-schranz: its required when we update to the WebTestCase which extends the [KernelTestCase](https://github.com/symfony/symfony/blob/master/src/Symfony/Bundle/FrameworkBundle/Test/KernelTestCase.php#L44)

src/Sulu/Bundle/TestBundle/Testing/SuluTestCase.php
alexander-schranz: this was too early didn't want to deprecate this class now

src/Sulu/Bundle/TestBundle/TestCase/SuluTestCase.php
alexander-schranz: ups 🙈

src/Sulu/Bundle/ContentBundle/Tests/Functional/Controller/NodeControllerTest.php
alexander-schranz: 👍

src/Sulu/Component/Webspace/Manager/WebspaceManager.php
alexander-schranz: when I did debug it. It seems like the $container from the `TestCase` called require_once and the container from `$client` did call require_once as they have different WebspaceManager instances. It did end up with: `Cannot declare class Test, because the name is already in use`

src/Sulu/Bundle/ContentBundle/Tests/Functional/Controller/NodeControllerTest.php
alexander-schranz: there exist a Tests/Fixtures folder but I think this classes are can be removed as they really seems to be unused and extend from a strange not existing class: https://github.com/sulu/sulu/tree/2.0.0-alpha4/src/Sulu/Bundle/ContentBundle/Tests/Fixtures

danrot: Can you remove them as well, if they are really unused? :see_no_evil:

danrot:
@alexander-schranz Anyway we can extract some parts of this PR, so that we can merge them separately? E.g. fixing all the wrong namespaces? It's quite a big PR which stays open for quite some time otherwise, and will cause frustration when rebasing etc.

alexander-schranz:
@danrot could found the error there was a mix of environments in the test case. sometimes it did you `dev` and sometimes it did you `test`. Forced always using the test environment for tests.

danrot:
@alexander-schranz Awesome! Any idea why it is so much faster now? :see_no_evil:

alexander-schranz:
@danrot it does not need to build the container twice (dev and test env), now its always test env.

src/Sulu/Component/phpunit.xml.dist
```diff
@@ -1,9 +1,12 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<phpunit bootstrap="./vendor/autoload.php" colors="true">
+<phpunit bootstrap="./../../../vendor/autoload.php" colors="true">
```
danrot: This file feels a bit lost here... Why did you have to move it?

src/Sulu/Bundle/ContentBundle/Tests/Functional/Controller/NodeControllerTest.php
alexander-schranz: this are removed

src/Sulu/Component/phpunit.xml.dist
alexander-schranz: currently not needed did move it back

src/Sulu/Bundle/AdminBundle/Tests/Functional/AppKernel.php
```diff
@@ -9,6 +9,8 @@
  * with this source code in the file LICENSE.
  */
 
+namespace Sulu\Bundle\AdminBundle\Tests\Functional;
```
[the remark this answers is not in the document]
alexander-schranz: `Functional` is like `src` folder here so I don't need to change many configurations. But for me its ok to move it outside it should not be too much work.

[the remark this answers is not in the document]
alexander-schranz: /cc @chirimoya whats your preferred Tests folder structure?

 - config
 - Unit
 - Functional
 - AppKernel.php

or

 - Unit
 - Functional
   - config
   - AppKernel.php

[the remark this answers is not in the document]
chirimoya: - Application
    - config
    - Kernel.php
- Functional
- Unit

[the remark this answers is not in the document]
alexander-schranz: @danrot is changed and rebased.

src/Sulu/Bundle/WebsiteBundle/Tests/Functional/Entity/DomainRepositoryTest.php
```diff
@@ -1,5 +1,16 @@
 <?php
 
+/*
+ * This file is part of Sulu.
+ *
+ * (c) MASSIVE ART WebServices GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
```
danrot: The comment would be sufficient once :wink:

alexander-schranz: ups there where 2 namespaces. is fixed!

PR #4229, opened 2018-10-25
Added support for list-builder configuration in project config directory

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Related issues/PRs | #4147
| License | MIT

#### What's in this PR?

Added support for list-builder configuration in project config directory. 

#### Why?

As the application itself is not longer a bundle we should allow to store `list-builder` configuration in project `config` directory to allow create custom entities in a project.

#### Example Usage

Create `config/list-builder/Test.xml`

~~~xml
<class xmlns="http://schemas.sulu.io/class/general"
       xmlns:orm="http://schemas.sulu.io/class/doctrine"
       xmlns:list="http://schemas.sulu.io/class/list">
    <properties>
        <property name="id" visibility="always" list:translation="app.id" searchability="yes">
            <orm:field-name>id</orm:field-name>
            <orm:entity-name>App\Entity\Test</orm:entity-name>
        </property>
    </properties>
</class>
~~~

try to load the fielddescriptors:

```php
function (FieldDescriptorFactoryInterface $fieldDescriptorsFactory) {
    $fieldDescriptors = $fieldDescriptorsFactory->getFieldDescriptorForClass(Location::class);
}
```

config folder structure will look like this in a project:

Project Config Folder

Additional folders in config could be `doctrine` or `serializer`?

### DIFF
--- a/src/Sulu/Component/Rest/ListBuilder/Metadata/Provider/ChainProvider.php
+++ b/src/Sulu/Component/Rest/ListBuilder/Metadata/Provider/ChainProvider.php
@@ -37,7 +37,13 @@ class ChainProvider implements ProviderInterface
         $classMetadata = new ClassMetadata($className);
 
         foreach ($this->chain as $provider) {
-            $classMetadata->merge($provider->getMetadataForClass($className));
+            $metaData = $provider->getMetadataForClass($className);
+
+            if (!$metaData) {
+                throw new \RuntimeException(sprintf('Could not find metadata for class "%s"', $className));
+            }
+
+            $classMetadata->merge($metaData);
         }
 
         return $classMetadata;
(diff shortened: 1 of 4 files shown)


wachterjohannes:
as long as we keep the templates/webspaces folder directly inside of the config folder i would also place the other folders beside them

src/Sulu/Component/Rest/ListBuilder/Metadata/Provider/ChainProvider.php
```diff
@@ -37,7 +37,13 @@ public function getMetadataForClass($className)
         $classMetadata = new ClassMetadata($className);
 
         foreach ($this->chain as $provider) {
-            $classMetadata->merge($provider->getMetadataForClass($className));
+            $metaData = $provider->getMetadataForClass($className);
```
danrot: `$metadata`

alexander-schranz: fixed

PR #4230, opened 2018-10-25
Allow to move media collections

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR adds the possibility to move collections in the UI.

#### Why?

Because it was possible in the old UI as well.

#### To Do

- [x] Move to top folder (without parent)
- [x] Remove `clearData` from `DatagridStore` in favour of `clear`


chirimoya:
- [ ] reloading root collection doesn't work
- [ ] after moving a collection, the collection tree in the overlay is outdated
- [ ] please keep the collection hover edit functionality for now until we have a better solution

src/Sulu/Bundle/AdminBundle/Resources/js/containers/DatagridOverlay/DatagridOverlay.js
```diff
@@ -54,7 +56,12 @@ export default class DatagridOverlay extends React.Component<Props> {
     }
 
     componentDidUpdate(prevProps: Props) {
-        const {clearSelectionOnClose, open} = this.props;
+        const {clearSelectionOnClose, open, reloadOnOpen} = this.props;
+
+        if (!this.datagridStore.loading && reloadOnOpen && prevProps.open === false && open === true) {
+            this.datagridStore.reload();
```
danrot: For me it is still not logical that this also clears selections, and only loads the first level :see_no_evil:

danrot:
For further reference: This PR introduces a `reloadOnOpen` flag on the `DatagridOverlay`, which defines whether or not the data in the overlay should be reloaded when it is reopened.

We have settled for the following guideline:

- The overlay should not reload in forms resp. views containing a single record
- The overlay should reload in lists, because there are too many operations (e.g. deleting, moving and copying) available, which have an impact on the content of the overlay.

PR #4239, opened 2018-10-30
implemented handling of undefined-value in single-select component

nnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This PR adjusts the single-select component to not show any value, if nothing is selected. Furthermore, this PR implements the support for options with the value undefined. An option with the value undefined can be used to reset the current selection.

#### Why?

We need to display select-components that do not have a value initially. In these cases it is misleading, if the component contains the value of the first option.

#### To Do

- [x] do not display value of first option if nothing is selected
- [x] allow option with undefined value-property 
- [x] adjust and create tests

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Option.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Option.js
@@ -13,9 +13,9 @@ type Props = {
     selected: boolean,
     disabled: boolean,
     focus: boolean,
-    value: string | number,
+    value?: string | number,
     children: string,
-    onClick?: (value: string | number) => void,
+    onClick?: (value?: string | number) => void,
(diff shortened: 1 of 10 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Option.js
```diff
@@ -29,7 +29,7 @@ export default class Option extends React.PureComponent<Props> {
         focus: false,
         selected: false,
         selectedVisualization: 'icon',
-        value: '',
+        value: undefined,
```
danrot: This would be the default, and doesn't have to be separately defined.

nnatter: fixed 🙂

PR #4240, opened 2018-10-30
fixed travis timeout for php 5.5 test case

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Debug php 5.5 on travis

#### Why?

Currently composer install fails on php 5.5

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ matrix:
   include:
     - php: 5.5
       env:
-        - COMPOSER_FLAGS="--prefer-lowest --prefer-dist --no-interaction"
+        - COMPOSER_FLAGS="--prefer-lowest --prefer-dist --no-interaction -vvv"
     - php: 7.1
       env:
         - COMPOSER_FLAGS="--prefer-dist --no-interaction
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -18,7 +18,7 @@ matrix:
   include:
     - php: 5.5
       env:
-        - COMPOSER_FLAGS="--prefer-lowest --prefer-dist --no-interaction"
+        - COMPOSER_FLAGS="--prefer-lowest --prefer-dist --no-interaction -vvv"
```
alexander-schranz: this seems the only way that travis end up with 10m without output error 🙈

wachterjohannes: dont like it 🙊 @danrot what do you think?

alexander-schranz: @wachterjohannes I also don't like it but its the only way composer does a output and travis does not say there is no 10m output

danrot:
This should go into the 1.5 branch.

alexander-schranz:
@danrot done

Issue #3248, opened 2017-03-03
Content Type should handle disabled, readonly, required like html standard


alexander-schranz:
| Q | A
| --- | ---
| Bug? | no
| New Feature? | yes

### Actual Behavior

Currently only simple content types are handling require correctly. Disabled current don't work with the form mapper and readonly is mostly not included.

### Expected Behavior

 - `required`: need to input something or choose at least one in a multi selection
 - `readonly`: The user can only see this input but not manipulate it
 - `disabled`: The different to readonly in html is that its not send to the server when the form is submitted. So the form mapper should only write disabled fields not read it.

For the input it should also check the other attributes work correctly e.g. `autofocus`: https://www.w3schools.com/tags/tag_input.asp

### Steps to Reproduce

Disabled currently works like readonly because the form mapper will still read the values.

### Possible Solutions

Every component should have the ability to be:

 - [x] Disabled
 - [ ] Readonly
 - [x] Required

danrot:
I am not sure if `readonly` will ever be implemented, but `disabled` and `required` works now for almost all fields. Therefore going to close this issue.

PR #4247, opened 2018-11-05
implemented disabled property for simple field-type components

nnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Related issues/PRs | #3248
| License | MIT

#### What's in this PR?

This PR implements a disabled property for a majority of the field-type components. Field-type components are the building blocks of forms and can be used inside of XML templates.

#### Why?

Our XML form templates already allow to set a disabledCondition on a property. To make use of this feature, we need to signify to the user that a field-type component is disabled.


src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/Checkbox.js
```diff
@@ -28,6 +29,12 @@ export default class Checkbox extends React.Component<FieldTypeProps<boolean>> {
             return <Toggler checked={!!value} onChange={this.handleChange}>{label}</Toggler>;
         }
 
-        return <CheckboxComponent checked={!!value} onChange={this.handleChange}>{label}</CheckboxComponent>;
+        return (
+            <CheckboxComponent
+                active={!disabled}
```
danrot: It's not very optimal that we called it everywhere else `disabled` but only on this component `active` :see_no_evil: Do you think it is a lot of work to change that? Should be quite easy because of flow, shouldn't it? (Same goes for the `Toggler` and `Switch` components).

src/Sulu/Bundle/AdminBundle/Resources/js/components/Url/Url.js
```diff
@@ -176,6 +178,7 @@ export default class Url extends React.Component<Props> {
                     </SingleSelect>
                 </div>
                 <input
```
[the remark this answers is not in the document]
danrot: That is because there is a single select right next to it. The standard `Input` component only accepts an Icon there, and we decided to implement an own component instead of bloating the existing one.

src/Sulu/Bundle/AdminBundle/Resources/js/components/Input/Input.js
```diff
@@ -89,6 +91,7 @@ export default class Input<T: ?string | ?number> extends React.PureComponent<Inp
             inputStyles[alignment],
             {
                 [inputStyles.error]: !valid,
+                [inputStyles.disabled]: disabled,
```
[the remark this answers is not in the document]
nnatter: i think so. we need a explicit class on the `component-element` (parent of the `input-element`) to set the background color of the component that is visible behind the icon.

screenshot: https://imgur.com/DkItSIK

src/Sulu/Bundle/AdminBundle/Resources/js/components/Url/Url.js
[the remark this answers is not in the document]
nnatter: actually i meant replacing the `input-element` that starts on line 178 with our `input-component`. i was thinking about this because we need to duplicate the styling of the `input-element` right now. (see https://github.com/sulu/sulu/pull/4247/files#diff-deaa8eaf7282fca4f86e7db8a68d3aedL26)

src/Sulu/Bundle/AdminBundle/Resources/js/components/Checkbox/Checkbox.js
```diff
@@ -5,18 +5,18 @@ import Switch from '../Switch';
 import type {SwitchProps} from '../Switch';
 import checkboxStyles from './checkbox.scss';
 
-type Props = SwitchProps & {
+type Props = {|
```
nnatter: i refactored the component to use an `exact object type` here. is there a reason for not using `exact object types` for component props?

src/Sulu/Bundle/AdminBundle/Resources/js/components/DatePicker/tests/DatePicker.test.js
```diff
@@ -13,52 +14,125 @@ beforeEach(() => {
 
 test('DatePicker should render', () => {
     const onChange = jest.fn();
-    expect(mount(<DatePicker onChange={onChange} value={null} />)).toMatchSnapshot();
+    const datePicker = mount(<DatePicker onChange={onChange} value={null} />);
+
+    expect(datePicker.render()).toMatchSnapshot();
+    expect(pretty(document.body ? document.body.innerHTML : '')).toMatchSnapshot();
+
+    datePicker.unmount();
```
nnatter: to keep tests/snapshots independent, we need to unmount the component (or use another way to clear the dom) after each test to remove the popover from the document body. unfortunatelly, calling `component.unmount()` at the end of every test is quite error prone, because it can be forgotten easily.

do we have a similar situation with another component? have we already implemented a less error prone solution there?

src/Sulu/Bundle/AdminBundle/Resources/js/components/Checkbox/Checkbox.js
danrot: Didn't know that the spread operator works in flow types, if it works this way, and all the type checks are correctly working this is even better than the current solution IMO.

The problem was that the `&` operator used before didn't do well with exact objects (which somehow also makes sense).

The other thing is that we didn't know about exact objects at the very beginning :slightly_smiling_face:  So if you encounter a component where no exact objects are used, and it would be easily possible, feel free to adapt it.

src/Sulu/Bundle/AdminBundle/Resources/js/components/DatePicker/tests/DatePicker.test.js
danrot: We have a similar problem e.g. with the `DatagridStore`, which should have its `destroy` method called, because otherwise too many `autorun`s are lying around... Don't know about any better solution at the moment :-/

src/Sulu/Bundle/AdminBundle/Resources/js/components/Input/Input.js
```diff
@@ -15,6 +15,7 @@ export default class Input<T: ?string | ?number> extends React.PureComponent<Inp
     static defaultProps = {
         alignment: 'left',
         collapsed: false,
+        disabled: false,
```
danrot: Why don't we set the `disabled` prop like this on every other component as well? Then we wouldn't have to make the `disabled` prop optional, and can be sure that it is always `false` or `true`, and never `undefined`, which would probably make it easier to reason about stuff.

src/Sulu/Bundle/AdminBundle/Resources/js/components/DatePicker/tests/__snapshots__/DatePicker.test.js.snap
```diff
@@ -1,1109 +1,1270 @@
...
+        type="text"
+        value=""
+      />
+    </label>
+  </div>
+</div>
+`;
+
+exports[`DatePicker should not open overlay on icon-click when disabled 2`] = `
+"<div>
```
danrot: Shouldn't that be empty now? Looks like the click opened the overlay. Also don't know if I would use snapshot testing here, since checking for the `open` attribute of the `Overlay` component would be enough.

src/Sulu/Bundle/AdminBundle/Resources/js/components/DatePicker/tests/__snapshots__/DatePicker.test.js.snap
```diff
@@ -1,1109 +1,1270 @@
...
+      <div class=\\"rdt\\">
+        <div></div>
+        <div class=\\"rdtPicker\\">
+          <div class=\\"rdtDays\\">
+            <table>
+              <thead>
+                <tr>
+                  <th class=\\"rdtPrev\\"><span>‹</span></th>
+                  <th class=\\"rdtSwitch\\" colspan=\\"5\\" data-value=\\"3\\">April 2017</th>
+                  <th class=\\"rdtNext\\"><span>
```
danrot: Is that snapshot necessary? This test should only check if the `error` class is correctly appended to the previous snapshot. Also feels like this one is from somewhere before, because in this test the overlay is not even opened...

danrot: It should not be possible that this overlay is opened, when the `disabled` flag is `true`, or should it?

danrot: Probably also not necessary to check the overlay in this test.

src/Sulu/Bundle/AdminBundle/Resources/js/components/SingleSelect/tests/SingleSelect.test.js
```diff
@@ -1,13 +1,13 @@
 /* eslint-disable flowtype/require-valid-file-annotation */
```
danrot: Have you tried to activate flow here as well?

src/Sulu/Bundle/AdminBundle/Resources/js/components/Toggler/Toggler.js
```diff
@@ -4,9 +4,10 @@ import Switch from '../Switch';
 import type {SwitchProps} from '../Switch';
 import togglerStyles from './toggler.scss';
 
-type Props = SwitchProps & {
+type Props = {|
+    ...SwitchProps,
     onChange?: (checked: boolean, value?: string | number) => void,
```
danrot: Couldn't that also be part of `SwitchProps`? Seems to be the same everywhere.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/Url.js
```diff
@@ -93,6 +94,7 @@ export default class Url extends React.Component<FieldTypeProps<?string>> {
 
         return (
             <UrlComponent
+                disabled={!!disabled}
```
danrot: Possible that you still have written no tests for the fields :wink:

src/Sulu/Bundle/AdminBundle/Resources/js/components/DatePicker/tests/DatePicker.test.js
danrot: But maybe it would work to get rid of `document.body` in these tests. I think if you query the `Portal` with enzyme, you can get whatever is rendered in there, without being dependant on `document.body`... Would at least be worth a try.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/Checkbox.js
```diff
@@ -25,9 +26,21 @@ export default class Checkbox extends React.Component<FieldTypeProps<boolean>> {
         } = this.props;
 
         if (type === 'toggler') {
-            return <Toggler checked={!!value} onChange={this.handleChange}>{label}</Toggler>;
+            return (
+                <Toggler
+                    checked={!!value}
+                    disabled={!!disabled}
+                    onChange={this.handleChange}
+                >{label}</Toggler>
```
danrot: That looks weird to me... if we have such a situation `{label}` usually goes into an own line...

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/Checkbox.js
```diff
@@ -25,9 +26,21 @@ export default class Checkbox extends React.Component<FieldTypeProps<boolean>> {
...
+            );
         }
 
-        return <CheckboxComponent checked={!!value} onChange={this.handleChange}>{label}</CheckboxComponent>;
+        return (
+            <CheckboxComponent
+                checked={!!value}
+                disabled={!!disabled}
+                onChange={this.handleChange}
+            >{label}</CheckboxComponent>
```
danrot: Same as above

PR #4254, opened 2018-11-07
implemented disabled property for sulu field-type components

nnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Related issues/PRs | #3248 
| License | MIT

#### What's in this PR?

This PR implements a disabled property for all components that are used by field-type components.. Field-type components are the building blocks of forms and can be used inside of XML templates.

#### Why?

Our XML form templates already allow to set a disabledCondition on a property. To make use of this feature, we need to signify to the user that a field-type component is disabled.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/SortableBlocks.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/SortableBlocks.js
@@ -7,8 +7,8 @@ import SortableBlock from './SortableBlock';
 import sortableBlockListStyles from './sortableBlockList.scss';
 import type {BlockEntry, RenderBlockContentCallback} from './types';
 
-type Props = {
-    disabled?: boolean,
+type Props = {|
+    disabled: boolean,
     blockTypes: Array<string>,
     expandedBlocks: Array<boolean>,
     onExpand: (index: number) => void,
@@ -18,10 +18,14 @@ type Props = {
     renderBlockContent: RenderBlockContentCallback,
     types?: {[key: string]: string},
     value: Array<BlockEntry>,
-};
+|};
 
 @observer
 class SortableBlocks extends React.Component<Props> {
+    static defaultProps = {
+        disabled: false,
+    };
+
     handleExpand = (index: number) => {
         const {onExpand} = this.props;
         onExpand(index);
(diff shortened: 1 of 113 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/SortableBlocks.js
```diff
@@ -2,11 +2,13 @@
 import React from 'react';
```
nnatter: i noticed that this file is named `SortableBlocks.js` and the respective css file is named `sortableBlockList.scss`. should we rename this file to `SortableBlockList.js`?

src/Sulu/Bundle/ContentBundle/Resources/config/forms/PageExcerpt.xml
```diff
@@ -18,12 +18,12 @@
                 <title>sulu_admin.description</title>
             </meta>
         </property>
-        <property name="categories" type="category_list">
+        <property name="categories" type="category_list" disabledCondition="title == 'test'">
```
danrot: That should probably not be here, right? :slightly_smiling_face:

src/Sulu/Bundle/ContentBundle/Resources/config/forms/PageExcerpt.xml
```diff
@@ -18,12 +18,12 @@
...
             <meta>
                 <title>sulu_category.categories</title>
             </meta>
         </property>
-        <property name="tags" type="tag_list">
+        <property name="tags" type="tag_list" disabledCondition="title == 'test'">
```
danrot: Same as above

src/Sulu/Bundle/SecurityBundle/Resources/js/containers/RoleAssignments/RoleAssignment.js
```diff
@@ -5,14 +5,19 @@ import {MultiSelect} from 'sulu-admin-bundle/components';
 import type {Localization} from 'sulu-admin-bundle/stores';
 import roleAssignmentStyle from './roleAssignment.scss';
```
danrot: Maybe also add an opacity here for consistency?

src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/SortableBlocks.js
danrot: I would say that makes sense, yeah :slightly_smiling_face:

PR #4255, opened 2018-11-07
Validate composer json

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Validate composer json

#### Why?

To avoid problems in the composer.json

#### Example Usage

Errors in the composer.json will be validated by the travis ci when create a PR.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -95,7 +95,6 @@
         "phpunit/phpunit": "^6.5",
         "php-ffmpeg/php-ffmpeg": "^0.13",
         "symfony/browser-kit": "^3.4 || ^4.0",
-        "symfony/css-selector": "^3.4 || ^4.0",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -95,7 +95,6 @@
         "phpunit/phpunit": "^6.5",
         "php-ffmpeg/php-ffmpeg": "^0.13",
         "symfony/browser-kit": "^3.4 || ^4.0",
-        "symfony/css-selector": "^3.4 || ^4.0",
```
alexander-schranz: > symfony/css-selector is required both in require and require-dev, this can lead to unexpected behavior

PR #4259, opened 2018-11-08
Fixed property param default value

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Remove the accidentally added default value of params value and fix single select to work without value as documented in the [sulu docs single_select](http://docs.sulu.io/en/latest/reference/content-types/single_select.html)

#### Why?

To fix that a value is `"string"` and avoid a not needed bc break as value is not mandatory in the xml scheme.

#### Example Usage

See http://docs.sulu.io/en/latest/reference/content-types/single_select.html

### DIFF
--- a/src/Sulu/Component/Content/Metadata/Parser/PropertiesXmlParser.php
+++ b/src/Sulu/Component/Content/Metadata/Parser/PropertiesXmlParser.php
@@ -345,7 +345,12 @@ class PropertiesXmlParser
                 $result['value'] = $this->loadParams('x:param', $xpath, $node);
                 break;
             default:
-                $result['value'] = $this->getValueFromXPath('@value', $xpath, $node, 'string');
+                $result['value'] = $this->getValueFromXPath('@value', $xpath, $node);
+
+                if (null === $result['value']) {
+                    $result['value'] = $result['name'];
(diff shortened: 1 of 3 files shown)


src/Sulu/Component/Content/Metadata/Parser/PropertiesXmlParser.php
```diff
@@ -345,7 +345,12 @@ private function loadParam(\DOMXPath $xpath, \DOMNode $node)
                 $result['value'] = $this->loadParams('x:param', $xpath, $node);
                 break;
             default:
-                $result['value'] = $this->getValueFromXPath('@value', $xpath, $node, 'string');
+                $result['value'] = $this->getValueFromXPath('@value', $xpath, $node);
+
+                if (null === $result['value']) {
```
alexander-schranz: will fix this in SingleSelect by using always name

danrot: So this PR is not needed anymore? Can it be closed then?

PR #4262, opened 2018-11-09
refactored select components to use exact object types

nnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This PR refactors the already implemented select components to use props with exact object types.

#### Why?

Exact object types provide a better type safety and allow for better static code checking.

#### Discussion Points

The implementation of the `Option` component allows for not setting the `value` prop. This is desired behaviour that allows to implement valueless-options that can be used to reset the value of a single-selection.

Unfortunately, the `SingleSelect` component and `MultiSelect` component are incompatible with these valueless-options as they do not expect undefined values in their `handleSelect()` methods. These inconsistencies were not detected by flow until now, because we were not using exact object types.

How should we handle these cases?

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/MultiSelect/MultiSelect.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/MultiSelect/MultiSelect.js
@@ -1,5 +1,6 @@
 // @flow
 import React from 'react';
+import type {Element} from 'react';
 import {observer} from 'mobx-react';
 import MultiSelectComponent from '../../components/MultiSelect';
 import ResourceListStore from '../../stores/ResourceListStore';
@@ -26,8 +27,6 @@ export default class MultiSelect<T: string | number> extends React.Component<Pro
         values: [],
     };
 
-    handleChange: (Array<T>) => void;
-
     resourceListStore: ResourceListStore;
 
     constructor(props: Props<T>) {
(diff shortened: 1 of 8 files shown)


danrot:
Well, `handleSelect` can receive a value of `undefined` now, right? Do we have another choice than accepting it in the type system?

danrot:
@nnatter ping :slightly_smiling_face:

src/Sulu/Bundle/AdminBundle/Resources/js/containers/MultiSelect/MultiSelect.js
```diff
@@ -78,11 +77,11 @@ export default class MultiSelect<T: string | number> extends React.Component<Pro
                 onChange={this.handleChange}
                 values={values}
             >
-                {this.resourceListStore.data.map((object, index) => (
+                {this.resourceListStore.data.map((object, index) => ((
                     <MultiSelectComponent.Option key={index} value={object[idProperty]}>
                         {object[displayProperty]}
                     </MultiSelectComponent.Option>
-                ))}
+                ): Element<typeof MultiSelectComponent.Option>))}
```
danrot: Why is that necessary? Is that a way to circumvent a flow bug?

nnatter: yes, flow infers Element<any> as return type of this arrow function and throws an error.

nnatter:
Unfortunately, missing data in the `props` of an `Option` component is still not detected by flow. This bug occurs only if the `Option` component has a static `defaultProps` variable.

I just checked the current develop branch and noticed that this error also occurs there.

PR #4271, opened 2018-11-13
Fixed hash serializer for inline serialization

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Replace deprecated `addData` function with `setData` function to avoid a serialization error when serialize [inline](https://jmsyst.com/libs/serializer/master/reference/annotations#inline). ~~Replace deprecated [GenericSerializationVisitor](https://github.com/schmittjoh/serializer/blob/1.13.0/src/JMS/Serializer/GenericSerializationVisitor.php#L11)~~

#### Why?

To fix hash serializer for inline serialization. If you have subentities which are also auditable and you try to inline them it will error with `_hash` exist instead. 

#### Example Usage

~~~php
    /**
     * @Serializer\Expose
     * @Serializer\Inline
     *
     * @ORM\OneToOne(targetEntity="App\Model\Location\LocationDimensionRestaurant", mappedBy="locationDimension", cascade={"persist", "remove"})
     */
    private $restaurant;

    /**
     * @Serializer\Expose
     * @Serializer\Inline
     *
     * @ORM\OneToOne(targetEntity="App\Model\Location\LocationDimensionBaker", mappedBy="locationDimension", cascade={"persist", "remove"})
     */
    private $baker;
~~~

### DIFF
--- a/src/Sulu/Component/Hash/Serializer/Subscriber/HashSerializeEventSubscriber.php
+++ b/src/Sulu/Component/Hash/Serializer/Subscriber/HashSerializeEventSubscriber.php
@@ -13,7 +13,6 @@ namespace Sulu\Component\Hash\Serializer\Subscriber;
 
 use JMS\Serializer\EventDispatcher\EventSubscriberInterface;
 use JMS\Serializer\EventDispatcher\ObjectEvent;
-use JMS\Serializer\EventDispatcher\PreSerializeEvent;
 use JMS\Serializer\JsonSerializationVisitor;
 use Sulu\Component\Content\Document\Behavior\LocalizedAuditableBehavior;
 use Sulu\Component\Hash\HasherInterface;
--- a/src/Sulu/Component/Hash/Serializer/Subscriber/HashSerializeEventSubscriber.php
+++ b/src/Sulu/Component/Hash/Serializer/Subscriber/HashSerializeEventSubscriber.php
@@ -13,7 +13,8 @@ namespace Sulu\Component\Hash\Serializer\Subscriber;
 
 use JMS\Serializer\EventDispatcher\EventSubscriberInterface;
 use JMS\Serializer\EventDispatcher\ObjectEvent;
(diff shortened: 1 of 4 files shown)


src/Sulu/Component/Hash/Serializer/Subscriber/HashSerializeEventSubscriber.php
```diff
@@ -13,7 +13,7 @@
 
 use JMS\Serializer\EventDispatcher\EventSubscriberInterface;
 use JMS\Serializer\EventDispatcher\ObjectEvent;
-use JMS\Serializer\GenericSerializationVisitor;
+use JMS\Serializer\JsonSerializationVisitor;
```
danrot: Is that change required? This means this subscriber is only working for JSON serialization...

src/Sulu/Component/Serializer/ArraySerializationVisitor.php
```diff
@@ -11,12 +11,12 @@
 
 namespace Sulu\Component\Serializer;
 
-use JMS\Serializer\GenericSerializationVisitor;
+use JMS\Serializer\JsonSerializationVisitor;
```
danrot: Same as above.

src/Sulu/Component/Hash/Serializer/Subscriber/HashSerializeEventSubscriber.php
alexander-schranz: currently not required but required when we will upgrade to 2.0

src/Sulu/Component/Serializer/ArraySerializationVisitor.php
alexander-schranz: did undo the change and still use the GenericSerializationVisitor until we upgrade

PR #4272, opened 2018-11-14
update flow to version 0.86

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets |---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR updates flow to the latest version.

#### Why?

Because it found a few more bug, which were not tackled before.


src/Sulu/Bundle/AdminBundle/Resources/js/containers/Datagrid/fieldTransformers/BytesFieldTransformer.js
```diff
@@ -5,7 +5,7 @@ import type {FieldTransformer} from '../types';
 export default class BytesFieldTransformer implements FieldTransformer {
     transform(value: *): Node {
         if (value === undefined) {
-            return undefined;
+            return null;
```
danrot: @trickreich That change should be ok, right? If not we would have to update the return value of the `transform` function. But `null` seems to be the more appropriate value in React, since a React component also can't return `undefined`. Asking for all the other transformers as well.

trickreich: Should be ok

PR #4274, opened 2018-11-14
added .editorconfig file to improve code style consistency

nnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This PR adds a `.editorconfig` file into the root directory of the project.

#### Why?

EditorConfig helps developers define and maintain consistent coding styles between different editors and IDEs. See https://editorconfig.org/

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -10,3 +10,7 @@ end_of_line = lf
 charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
+
+# markdown uses two trailing spaces for explicit line breaks
+[*.md]
+trim_trailing_whitespace = false
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,12 @@
+# EditorConfig helps developers define and maintain consistent coding styles between different editors and IDEs
+# editorconfig.org
+
+root = true
+
+[*]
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -0,0 +1,12 @@
...
+# editorconfig.org
+
+root = true
+
+[*]
+indent_style = space
+indent_size = 4
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
```
alexander-schranz: should not be used on markdown files

alexander-schranz: maybe oriented here on [symfony editorconfig](https://github.com/symfony/symfony/blob/master/.editorconfig)

nnatter: added an exception for `.md` files. but i think you could also use a backslash instead of two spaces to insert a line break in markdown 🙂

danrot:
Wouldn't it be a good idea to include our max line limit of 120 chars as well? Saw in the [symfony editorconfig](https://github.com/symfony/symfony/blob/master/.editorconfig#L16) that this is possible.

PR #4276, opened 2018-11-15
Add RouteBuilders

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Documentation PR | ---

#### What's in this PR?

This PR adds the `Builder` pattern to create routes.

#### Why?

Because it is more comfortable to work with.

#### To Do

- [x] Redirect `ResourceTabs` to first tab, if nothing is selected
- [x] How to handle inheritance? Remove and therefore duplicate a few configurations? Keep it and mix logic between different RouteBuilder?


src/Sulu/Bundle/AdminBundle/Admin/Routing/DatagridRouteBuilder.php
```diff
@@ -0,0 +1,141 @@
...
+
+    /**
+     * @var Route
+     */
+    private $route;
+
+    public function __construct(string $name, string $path)
+    {
+        // TODO If some of the Builder functions is called after the getRoute function, it will affect the outputted route.
+        // Maybe create the route later, and make it immutable?
```
danrot: That's a question that has to be answered for all builders

src/Sulu/Bundle/AdminBundle/Admin/Routing/FormRouteBuilder.php
```diff
@@ -0,0 +1,126 @@
...
+
+    public function getRoute(): Route
+    {
+        if (!$this->route->getOption('resourceKey')) {
+            throw new \DomainException(
+                'A route for a Form view needs a "resourceKey" option.'
+                . ' You have likely forgotten to call the "setResourceKey" method.'
+            );
+        }
+
```
danrot: Would love to check here other stuff as well, e.g. if the locale is set, but it is not that easy, because of the route inheritance we have, which is resolved at a later stage.

In our current setup e.g. the `:locale` placeholder in the path is set on the `ResourceTabs` route, and the `Form` route inherits that part of the path, but the builder doesn't know about it yet. So it is not possible to check it in this place, except we remove that inheritance stuff (for both options and path).

That would mean we have to write the full path in each route, even if the first part of the parent route could be used, and we would have to duplicate some options across routes. However, this duplication could be fixed by using variables.

I am really thinking about this options, since it would make the system less complex and magical, and therefore easier to understand.

src/Sulu/Bundle/AdminBundle/Admin/Routing/Route.php
```diff
@@ -100,13 +100,28 @@ public function getPath(): string
...
     }
 
+    public function getView(): string
+    {
+        return $this->view;
+    }
+
     public function addOption(string $key, $value): self
     {
+        // TODO rename to setOption, since old values will be simply overridden?
```
danrot: I think we really should do this change, although it is quite a big BC break.

src/Sulu/Bundle/AdminBundle/Admin/Routing/DatagridRouteBuilder.php
```diff
@@ -0,0 +1,141 @@
...
+            );
+        }
+
+        if (!$this->route->getOption('locales') && false !== strpos($this->route->getPath(), ':locale')) {
+            throw new \DomainException(
+                'A route for a datagrid cannot have a ":locale" placeholder in its URL if no "locales" have been set.'
+            );
+        }
+
+        return $this->route;
```
chirimoya: clone ?

src/Sulu/Bundle/AdminBundle/Admin/Routing/Route.php
chirimoya: ok

src/Sulu/Bundle/CategoryBundle/Admin/CategoryAdmin.php
```diff
@@ -79,34 +88,41 @@ function(Localization $localization) {
...
+                ->addDatagridAdapters(['tree_table'])
+                ->addLocales($locales)
+                ->setDefaultLocale($locales[0])
+                ->setAddRoute(static::ADD_FORM_ROUTE)
+                ->setEditRoute(static::EDIT_FORM_ROUTE)
+                ->enableSearching()
+                ->enableMoving()
+                ->getRoute(),
+            $this->routeBuilderFactory->createResourceTabRouteBuilder(static::ADD_FORM_ROUTE, '/categories/:locale/add')
+                ->setResourceKey('categories')
```
chirimoya: add `setBackRoute`

src/Sulu/Bundle/AdminBundle/Admin/Routing/ResourceTabRouteBuilderInterface.php
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+namespace Sulu\Bundle\AdminBundle\Admin\Routing;
+
+interface ResourceTabRouteBuilderInterface
+{
+    public function setResourceKey(string $resourceKey): self;
+
+    /**
+     * @param array $locales
```
chirimoya: add type or remove

UPGRADE.md
```diff
@@ -2,6 +2,14 @@
 
 ## dev-develop
 
+### RouteBuilder
+
+Instead of creating all the Routes in the `Admin::getRoutes` method completely on your own there is now the possibility
+to use the `RouteBuilderFactory`. It offers methods to create Form, Datagrid and ResourceTabs routes in a typed way,
+which means auto completion in the IDE should work as well.
+
+Apart from that the `Route::addOption` was renamed to `Route::setOption`, since it overrides existing option values.
```
chirimoya: add addAttributeDefault change

src/Sulu/Bundle/CategoryBundle/Admin/CategoryAdmin.php
```diff
@@ -79,36 +88,43 @@ function(Localization $localization) {
...
+                ->addLocales($locales)
+                ->setBackRoute(static::DATAGRID_ROUTE)
+                ->getRoute(),
+            $this->routeBuilderFactory->createFormRouteBuilder('sulu_category.add_form.detail', '/details')
+                ->setResourceKey('categories')
+                ->setFormKey('categories')
```
chirimoya: category.details

PR #4277, opened 2018-11-15
extracted media-selection field-type from media-selection container

nnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | hopefully no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This PR splits up the `MediaSelection` component into a field-type component and a container-component.

#### Why?

This allows to use the `MediaSelection` component outside of a form. Also, most other input-components are split into a container-component and a field-type too.

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/Content/Types/MediaSelectionContentType.php
+++ b/src/Sulu/Bundle/MediaBundle/Content/Types/MediaSelectionContentType.php
@@ -90,7 +90,7 @@ class MediaSelectionContentType extends ComplexContentType implements ContentTyp
         $languageCode,
         $segmentKey
     ) {
-        $data = json_decode($node->getPropertyValueWithDefault($property->getName(), '{}'), true);
+        $data = json_decode($node->getPropertyValueWithDefault($property->getName(), '{"ids": []}'), true);
 
         $property->setValue($data);
     }
--- a/src/Sulu/Bundle/MediaBundle/Resources/js/containers/Form/fields/MediaSelection.js
+++ b/src/Sulu/Bundle/MediaBundle/Resources/js/containers/Form/fields/MediaSelection.js
@@ -1,0 +1,37 @@
+// @flow
+import React from 'react';
+import {observer} from 'mobx-react';
+import type {FieldTypeProps} from 'sulu-admin-bundle/types';
+import type {Value} from '../../MediaSelection/types';
+import MediaSelectionComponent from '../../MediaSelection';
+
+@observer
+export default class MediaSelection extends React.Component<FieldTypeProps<Value>> {
+    handleChange = (selectedIds: Array<number>) => {
+        const {onChange, onFinish} = this.props;
+
+        onChange({
+            ids: selectedIds,
+        });
+        onFinish();
+    };
+
+    render() {
+        const {formInspector, disabled, value} = this.props;
+
+        if (!formInspector || !formInspector.locale) {
+            throw new Error('The media selection needs a locale to work properly');
+        }
+
+        const {locale} = formInspector;
+
+        return (
+            <MediaSelectionComponent
+                disabled={!!disabled}
+                locale={locale}
+                onChange={this.handleChange}
+                value={value && value.ids ? value.ids : []}
+            />
+        );
+    }
+}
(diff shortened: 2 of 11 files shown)


src/Sulu/Bundle/MediaBundle/Resources/js/containers/Form/fields/MediaSelection.js
```diff
@@ -0,0 +1,37 @@
...
+        }
+
+        const {locale} = formInspector;
+
+        return (
+            <MediaSelectionComponent
+                disabled={!!disabled}
+                locale={locale}
+                onChange={this.handleChange}
+                value={value && value.ids ? value.ids : []}
```
danrot: Not sure if I would pass only the ids here... Wouldn't it be better to pass the entire object? Then it would also be clear where the `displayPosition` goes, once it will be implemented...

src/Sulu/Bundle/MediaBundle/Resources/js/containers/Form/tests/fields/MediaSelection.test.js
```diff
@@ -0,0 +1,47 @@
...
+            disabled={true}
+            formInspector={formInspector}
+            value={{ids: [55, 66, 77]}}
+        />
+    );
+
+    expect(mediaSelection.find(MediaSelectionComponent).props().disabled).toEqual(true);
+    expect(mediaSelection.find(MediaSelectionComponent).props().locale.get()).toEqual('en');
+    expect(mediaSelection.find(MediaSelectionComponent).props().value).toEqual([55, 66, 77]);
+});
```
danrot: Maybe add another test showing that an error is thrown if no locale is passed and if the `onChange` and `onFinish` callback are called correctly.

src/Sulu/Bundle/MediaBundle/Resources/js/index.js
```diff
@@ -7,7 +7,7 @@ import MediaOverview from './views/MediaOverview';
 import MediaDetail from './views/MediaDetail';
 import MediaHistory from './views/MediaHistory';
 import MediaFormats from './views/MediaFormats';
-import MediaSelection from './containers/MediaSelection';
+import MediaSelection from './containers/Form/fields/MediaSelection';
```
danrot: I would import it together with the `SingleMediaUpload` above.

src/Sulu/Bundle/MediaBundle/Resources/js/stores/MediaSelectionStore/MediaSelectionStore.js
```diff
@@ -1,9 +1,9 @@
 // @flow
 import {action, computed, observable} from 'mobx';
 import type {IObservableValue} from 'mobx'; // eslint-disable-line
-import {arrayMove} from 'sulu-admin-bundle/components';
-import {ResourceRequester} from 'sulu-admin-bundle/services';
-import type {MediaItem} from '../types';
+import {arrayMove} from 'sulu-admin-bundle/components/index';
+import {ResourceRequester} from 'sulu-admin-bundle/services/index';
```
danrot: The two `index` above should not be necessary. Resp. why is it?

src/Sulu/Bundle/MediaBundle/Resources/js/containers/Form/fields/MediaSelection.js
```diff
@@ -30,7 +28,7 @@ export default class MediaSelection extends React.Component<FieldTypeProps<Value
                 disabled={!!disabled}
                 locale={locale}
                 onChange={this.handleChange}
-                value={value && value.ids ? value.ids : []}
+                value={value ? value : undefined}
```
danrot: Can't we just write `value` then?

src/Sulu/Bundle/MediaBundle/Content/Types/MediaSelectionContentType.php
```diff
@@ -90,7 +90,7 @@ public function read(
         $languageCode,
         $segmentKey
     ) {
-        $data = json_decode($node->getPropertyValueWithDefault($property->getName(), '{}'), true);
+        $data = json_decode($node->getPropertyValueWithDefault($property->getName(), '{"ids": []}'), true);
```
danrot: @wachterjohannes Can you have a quick look at this commit? The rest of the PR is fine, but this was written by me. It is necessary, because if we leave the default value as it is, then the serializer will write `[]` instead of `{}`, and this will break our administration interface.

wachterjohannes: as long as this is the value if you select no media in the Admin its ok for me :)

PR #4253, opened 2018-11-07
Added phpstan fixes

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Added phpstan fixes.

#### Why?

Phpstan fixes.

PR #4279, opened 2018-11-15
Fix errors and introduce PHPStan

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | #4253 
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR fixes the few remaining PHP errors, and introduce PHPStan in our CI build.

#### Why?

To get a higher code quality and avoid stupid but hard to see mistakes.

#### ToDo

- [ ] Fix `PropertyMetadata` parent constructor call
- [ ] Fix `SuluTestCase` class
- [ ] `CacheClearCommand` of `PreviewBundle` is not called anymore and contains errors


danrot:
There are still a few issues:

- [ ] A few private services are used in our tests
- [ ] The `PropertyMetadata` doesn't work if the parent is called for `VirtualPropertyMetadata`, because the `jms/metadata` package uses reflection and these properties do not exist on the entity.
- [ ] Our own [CacheClearCommand](https://github.com/sulu/sulu/blob/c58eb8bdc2a6bd472da2ec1fa00609349ac62ab5/src/Sulu/Bundle/PreviewBundle/Command/CacheClearCommand.php) throws some errors. Not sure if it is actually working, because it doesn't seem to be called anymore.

danrot:
Our own `CacheClearCommand` doesn't work anymore as we have implemented it in Symfony 3.x... We just changed the container in this command, and that worked, since the commands always got their dependencies from the container. In later Symfony versions proper dependency injection is used, and therefore it would even use the services from the wrong container, even if we change the container in the command.

A solution would be to start another process in the command, but then we would have to introduce a `previewconsole` as well...

Another idea I had was to maybe introduce a `--context` option in the `bin/console` command and remove the `PreviewKernel` in favour of another context on the existing kernel. However, this would be a bigger task I guess.

Question is if we should just remove the current command, since it is not working anymore, and we have to find another solution anyway, where we most probably can't use any of the existing code. /cc @chirimoya

PR #4281, opened 2018-11-16
implemented single media selection field-type

nnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Related issues/PRs | builds upon  #4277 
| License | MIT

#### What's in this PR?

This PR implements a SingleMediaSelection Component along with a respective FieldType and ContentType.

#### Why?

Because we want to provide a Component/FieldType/ContentType that allows to select a single media entity.


src/Sulu/Bundle/MediaBundle/Resources/js/containers/SingleMediaSelection/README.md
```diff
@@ -0,0 +1 @@
+The `SingleMediaSelection` is a content-type based on the `SingleItemSelection` component. By using the `SingleMediaSelection` you can assign a single media inside a form.
```
danrot: Same as above, would write component instead of content-type.

src/Sulu/Bundle/MediaBundle/Resources/js/containers/SingleMediaSelection/SingleMediaSelection.js
```diff
@@ -0,0 +1,139 @@
...
+    locale: IObservableValue<string>,
+    onChange: (selectedIds: Value) => void,
+    value: Value,
+|}
+
+@observer
+export default class SingleMediaSelection extends React.Component<Props> {
+    static defaultProps = {
+        disabled: false,
+        value: { id: undefined },
```
danrot: We actually do not add padding to the object literals... I thought we had a eslint rule activate for that, but I seem to be wrong... Maybe you can add [this rule](https://eslint.org/docs/rules/object-curly-spacing) in a separate PR?

src/Sulu/Bundle/MediaBundle/Resources/js/containers/SingleMediaSelection/SingleMediaSelection.js
```diff
@@ -0,0 +1,139 @@
...
+                    disabled={disabled}
+                    emptyText={translate('sulu_media.select_media_singular')}
+                    leftButton={{
+                        icon: 'su-image',
+                        onClick: this.handleOverlayOpen,
+                    }}
+                    onRemove={this.singleMediaSelectionStore.selectedMedia ? this.handleRemove : undefined}
+                >
+                    {selectedMedia &&
+                    <MediaSelectionItem
```
danrot: Usually we are adding padding to this component, so that you see it belongs to the above `&&`.

src/Sulu/Bundle/MediaBundle/Resources/js/containers/SingleMediaSelection/tests/SingleMediaSelection.test.js
```diff
@@ -0,0 +1,142 @@
...
+        id: 22,
+        title: 'test media',
+        mimeType: 'image/jpeg',
+        thumbnails: {
+            'sulu-25x25': '/images/25x25/awesome.png',
+        },
+    }));
+});
+
+test('onChange-property should be called handler if selection store changes', () => {
```
danrot: `Should call onChange handler if ...` - your version reads a bit strange :wink:

src/Sulu/Bundle/MediaBundle/Resources/js/stores/SingleMediaSelectionStore/SingleMediaSelectionStore.js
```diff
@@ -0,0 +1,59 @@
+// @flow
+import type {IObservableValue} from 'mobx';
+import {action, computed, observable} from 'mobx';
+import {ResourceRequester} from 'sulu-admin-bundle/services';
+import type {MediaItem} from '../MultiMediaSelectionStore/types';
```
danrot: That's a bit a strange dependency... Wouldn't it be a better idea to use the `Media` type from the `MediaBundle/Resources/js/types.js` file in both places?

danrot:
@nnatter Build is failing

danrot:
There are only two styling issues left:
- The `SingleMediaSelection` should use a smaller image format, I would suggest 20x20
- In both `MediaSelection` components there is too much margin on the left side of the image

src/Sulu/Bundle/MediaBundle/Resources/js/components/MediaSelectionItem/MediaSelectionItem.js
```diff
@@ -3,28 +3,47 @@ import React from 'react';
...
 import mediaSelectionItemStyle from './mediaSelectionItem.scss';
 
-type Props = {
+type Props = {|
     mimeType: string,
     children: string,
     thumbnail?: string,
-};
+    thumbnailSize: number,
+    thumbnailMargin: number,
```
danrot: Usually we don't pass numeric style values like this... E.g. in the `Overlay` component we have a `size` prop, which can takes values like `small` or `large`. That avoids inline styles and wouldn't allow for mixed styles all over the place.

In general I am wondering if it makes sense to extract that component... Doesn't seem to extract a lot of code.

src/Sulu/Bundle/MediaBundle/Resources/js/components/MediaSelectionItem/mediaSelectionItem.scss
```diff
@@ -6,12 +6,8 @@ $thumbnailSize: 25px;
     align-items: center;
 }
 
-.thumbnail {
-    margin: 0 20px 0 10px;
-    width: $thumbnailSize;
-    height: $thumbnailSize;
```
danrot: This variable is still declared above, and probably not needed anymore?

src/Sulu/Bundle/MediaBundle/Resources/js/containers/MultiMediaSelection/MultiMediaSelection.js
```diff
@@ -18,8 +19,6 @@ type Props = {|
     value: Value,
 |}
 
-const THUMBNAIL_SIZE = 'sulu-25x25';
```
danrot: Since this value is used multiple times, it would make sense to keep this constant, wouldn't it?

src/Sulu/Bundle/MediaBundle/Resources/js/containers/SingleMediaSelection/SingleMediaSelection.js
```diff
@@ -17,8 +18,6 @@ type Props = {|
     value: Value,
 |}
 
-const THUMBNAIL_SIZE = 'sulu-25x25';
```
danrot: Same as above, would probably make sense to keep this constant, right?

PR #4282, opened 2018-11-20
Put branch alias in extra section

gisostallenberg:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?
The docs of composer state the branch-alias settings should be placed under extra.
See https://getcomposer.org/doc/articles/aliases.md#branch-alias

#### Why?
This makes it possible to require a development version without having issues with other packages requirements

#### Example Usage

~~~bash
composer require sulu/sulu:2.0.x-dev
~~~

#### BC Breaks/Deprecations
none

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -160,8 +160,11 @@
         "preferred-install": {
             "*": "dist"
         },
-        "sort-packages": true,
+        "sort-packages": true
+    },
+    "extra": {
         "branch-alias": {
+            "dev-master": "1.7-dev",
             "dev-develop": "2.0-dev"
         }
     }
--- a/composer.json
+++ b/composer.json
@@ -164,7 +164,7 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "1.7-dev",
+            "dev-master": "1.6-dev",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -160,8 +160,11 @@
         "preferred-install": {
             "*": "dist"
         },
-        "sort-packages": true,
+        "sort-packages": true
+    },
+    "extra": {
         "branch-alias": {
+            "dev-master": "1.7-dev",
```
danrot: We are not planning to release 1.7... Don't know too much about these aliases, but would it work to write `1.6-dev` here? :thinking:

alexander-schranz: `1.6-dev` should be fine here but I'm not sure if it need to be on the master branch

gisostallenberg: In my opinion the aliases should be something that comes 'next'. I mean master is ahead of current 1.6 right? When following semver it should be 1.7. Because you will stay at 1.6 I'll change it to 1.6-dev for you.

PR #4289, opened 2018-11-23
Remove deprecation message from getcolspan on section

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Remove deprecation message from getcolspan on section.

#### Why?

Sections can have colspan in sulu 2.0.

### DIFF
--- a/src/Sulu/Component/Content/Metadata/PropertyMetadata.php
+++ b/src/Sulu/Component/Content/Metadata/PropertyMetadata.php
@@ -162,21 +162,11 @@ class PropertyMetadata extends ItemMetadata
 
     public function getColSpan(): ?int
     {
-        @trigger_error(
-            sprintf('Do not use getter "%s" from "%s"', 'getColSpan', __CLASS__),
-            E_USER_DEPRECATED
-        );
-
         return $this->colSpan;
     }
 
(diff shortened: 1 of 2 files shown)


alexander-schranz:
@danrot ok now i'm little bit confused about this deprecations as this function are still used and are spaming the log in my case when editing pages. How we can ride of them or should the functions stay?

danrot:
IIRC we decided to rename it to `getSize`, and this method should be used instead.

danrot:
@alexander-schranz Maybe we can fix the deprecation message to tell that `getSize` and `setSize` should be used instead? Or we could remove it all together, or doesn't that work yet?

alexander-schranz:
@danrot I can have a look at it maybe later, but I'm not sure if size is good choice for a value which only represent the width of the element which value is neither pixels or percents of the parent.
Colspan did represent it for me better what it does as size.

/cc @chirimoya

danrot:
`colspan` indicates to me that there should also be a `rowspan`, which is not the case... But I am not really sure if that was the real reason for this change.

danrot:
Replaced by #4459

PR #4290, opened 2018-11-27
 adjusted selection components to not fire change-handler while selection-store is loading

nnatter:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This PR adjusts the selection components to not fire the given change-handler while their selection-store is loading.

#### Why?

The current implementation calls the given change handler two times when a selection component is created with an initial value. This is because the selection-store needs to load the data asynchronously and the change handler is fired automatically when the data is loaded.

The behaviour described above can be reproduced with the `contact-selection` on the *Page Settings* tab.


danrot:
Is it possible that this is only the case in the `SingleSelection`, and the other files you've touched do not require this fix?

PR #4291, opened 2018-11-27
 added and implemented loading prop for single-item-selection component

nnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This PR introduces a `loading` prop for the `SingleItemSelection` component. If the `loading` prop is set to true, a loading icon ist displayed on the right side of the component.

Furthermore, the PR changes the `SingleSelection` component and the `SingleMediaSelection` component to use the introduced prop.

#### Why?

Because we want to signal to the user that the selection is loading. With the current implementation, it looks like the input field is empty while the component is loading. The current behaviour can be seen at the `contact-selection` on the **Page Settings** tab.


danrot:
Please also add tests for the `SingleSelection` and `SingleMediaSelection`.

PR #4293, opened 2018-11-27
load full item on select in single-selection

nnatter:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This PR adjusts the `SingleSelection`  component to load the whole item with a separate request after it was selected in the overlay.

#### Why?

Because the item returned from the overlay may not contain all properties that are used as `displayProperties`.


danrot:
Tests are failing :slightly_smiling_face:

PR #4287, opened 2018-11-23
Rename build folder to fix problems with php internal server

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Realted PRs | https://github.com/sulu/sulu-minimal/pull/136
| License | MIT

#### What's in this PR?

Rename the `admin/build` folder to `admin-build`

#### Why?

The admin/ folder conflicts with the `/admin` route when using the php internal webserver and so it should be avoiding creating a admin folder there.

#### Example Usage

~~~php
php -S 127.0.0.1:8000 -t public
~~~

#### BC Breaks/Deprecations

Output file changed.

#### To Do

- [ ] Add breaking changes to UPGRADE.md

PR #4296, opened 2018-11-27
Added easier way to overwrite public directory and base path

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Related PRs? | https://github.com/sulu/sulu/pull/4287, https://github.com/sulu/sulu-minimal/pull/141
| License | MIT

#### What's in this PR?

The basepath where to build the js should be easy overwriteable over env variables.
And the public-dir is read from the composer.json configuration see https://symfony.com/doc/current/configuration/override_dir_structure.html#override-the-public-directory.

#### Why?

In some cases you want to change the basepath where the js is build to.

#### Example Usage

```js
const webpackConfig = require('./vendor/sulu/sulu/webpack.config.js');

module.exports = (env, argv) => {
    if (!env) {
        env = {};
    }

    env.base_path = 'admin-build';

    return webpackConfig(env, argv);
};
```

and changing public dir in symfony:

```json
{
    ...
    "extra": {
        "public-dir": "web",
        "symfony-bin-dir": "bin"
    }
}
```

See https://github.com/sulu/sulu-docs/pull/410/files

#### BC Breaks/Deprecations

None

#### To Do

- [x] Create a documentation PR https://github.com/sulu/sulu-docs/pull/410

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,9 +9,8 @@
         "test": "jest",
         "styleguide": "styleguidist server",
         "styleguide:build": "styleguidist build",
-        "clean": "rimraf public/build/admin",
-        "build": "npm run clean && webpack -p",
-        "watch": "npm run clean && webpack -d -w"
+        "build": "webpack -p",
+        "watch": "webpack -d -w"
     },
     "dependencies": {
         "sulu-admin-bundle": "file:src/Sulu/Bundle/AdminBundle/Resources/js",
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -2,6 +2,7 @@
 /* eslint-disable import/no-nodejs-modules*/
 const path = require('path');
 const glob = require('glob');
+const rimraf = require('rimraf');
 const webpack = require('webpack');
 const CleanObsoleteChunksPlugin = require('webpack-clean-obsolete-chunks');
 const ManifestPlugin = require('webpack-manifest-plugin');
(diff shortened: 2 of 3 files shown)


webpack.config.js
```diff
@@ -22,25 +22,19 @@ entries.unshift('core-js/fn/symbol');
...
         new webpack.DefinePlugin({
             BUNDLE_ENTRIES_COUNT: entriesCount,
         }),
-        new MiniCssExtractPlugin({
-            filename: basePath + '/[name].[chunkhash].css',
-        }),
-        new ManifestPlugin({
-            fileName: basePath + '/manifest.json',
-        }),
+        new MiniCssExtractPlugin(),
```
danrot: This way the `chunkhash` is missing, and therefore the cache busting would not work.

webpack.config.js
```diff
@@ -22,25 +22,19 @@ entries.unshift('core-js/fn/symbol');
...
 entries.unshift('url-search-params-polyfill');
 
-const basePath = 'build/admin';
-
 module.exports = (env, argv) => ({ // eslint-disable-line no-undef
     entry: entries,
     output: {
-        path: path.resolve('public'),
-        filename: basePath + '/[name].[chunkhash].js',
+        path: path.resolve('public/build/admin'),
```
danrot: This would indeed be easier, but the reason it wasn't done this way, is that the `ManifestPlugin` now outputs the wrong paths (the `/build/admin` part is missing), and the files are not found if you reload in the browser.

In addition to that a lot of entries are not part of the `manifest.json` anymore, which is also not a good sign...

danrot: Or maybe even pass it using one of the other two variables? `env` probably doesn't work, since I believe it is just `development` or `production`, but `argv` sounds like it is about arguments you can pass inside :slightly_smiling_face:

webpack.config.js
```diff
@@ -22,103 +23,123 @@ entries.unshift('core-js/fn/symbol');
...
+        rimraf.sync(path.resolve(publicDir, basePath));
+    } else {
+        console.warn([ // eslint-disable-line no-console
+            'WARN: Using absolute path for "public-dir" detected',
+            '      This can end up in accidentally remove of files outside your project dir.',
+            '      Removing old build files was skipped!',
+        ].join('\n'));
+    }
+
+    return {
```
alexander-schranz: the `underneath` lines didn't changed. they were just indented

webpack.config.js
```diff
@@ -22,103 +23,123 @@ entries.unshift('core-js/fn/symbol');
 entries.unshift('whatwg-fetch');
 entries.unshift('url-search-params-polyfill');
```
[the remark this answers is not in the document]
danrot: Any reason you have not done this?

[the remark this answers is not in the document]
alexander-schranz: only time 😆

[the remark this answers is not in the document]
alexander-schranz: @danrot now fixed

package.json
```diff
@@ -9,9 +9,8 @@
         "test": "jest",
         "styleguide": "styleguidist server",
         "styleguide:build": "styleguidist build",
-        "clean": "rimraf public/build/admin",
-        "build": "npm run clean && webpack -p",
-        "watch": "npm run clean && webpack -d -w"
+        "build": "webpack -p",
+        "watch": "webpack -d -w"
```
danrot: This change would also have to go into `sulu-minimal`, right? Otherwise the build is cleaned twice.

webpack.config.js
```diff
@@ -2,123 +2,146 @@
...
 
-entries.unshift('core-js/fn/array/includes');
-entries.unshift('core-js/fn/array/find-index');
-entries.unshift('core-js/fn/array/fill');
-entries.unshift('core-js/fn/array/from');
-entries.unshift('core-js/fn/promise');
-entries.unshift('core-js/fn/symbol');
-entries.unshift('whatwg-fetch');
-entries.unshift('url-search-params-polyfill');
+    const composerConfig = require(path.resolve('composer.json')); // eslint-disable-line import/no-dynamic-require
```
danrot: Isn't that the wrong `composer.json` in case it is run in sulu-minimal? :thinking:

alexander-schranz: no else path.resolve('public')  before would also be false.

package.json
alexander-schranz: done: https://github.com/sulu/sulu-minimal/pull/141

webpack.config.js
```diff
@@ -22,103 +23,123 @@ entries.unshift('core-js/fn/symbol');
...
+    if (composerConfig.extra && composerConfig.extra['public-dir']) {
+        publicDir = composerConfig.extra['public-dir'];
+    }
+
+    // Remove old build files
+    if (!publicDir.startsWith('/')) {
+        rimraf.sync(path.resolve(publicDir, basePath));
+    } else {
+        console.warn([ // eslint-disable-line no-console
```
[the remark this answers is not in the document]
alexander-schranz: done

PR #4299, opened 2018-11-28
Decouple resourceKey from forms

### LINKED ISSUE #2137: Jackrabbit requires restarts inbetween test suites.
| Q | A |
| --- | --- |
| Bug? | no |
#### Actual Behavior

The CI test run restarts jackrabbit in between each test suite.

If jackrabbit is NOT restarted then, after a certain number of tests, we get errors similar to:

```
44) Sulu\Component\Content\Tests\Functional\Compat\Rlp\ResourceLocatorTest::testLoadFromProperty
PHPCR\NoSuchWorkspaceException: couldn't connect to host "http://localhost:8080/server/test/jcr:root/cmf.0.json"
```

(see [failing travis build](https://travis-ci.org/sulu-io/sulu/jobs/116571995))
#### Expected Behavior

Jackrabbit should not need to be restarted, it should not break.
#### Steps to Reproduce

Run the tests on Travis without the restart flag.
#### Possible Solutions

Currently we implement a hack whereby we automatically restart Jackrabbit between tests. However this is not a scalable and it is likely we will reencounter the issue.

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #2137
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR will decouple the resourceKey from the forms, in order to allow to have multiple different forms for the same `resourceKey`.

#### Why?

This is e.g. required to have the Page Settings tab, and reuse the same data from the Details tab. It already works now, but the resourceStore is cloned and therefore another instance, which complicates things. In addition to that it currently makes problems with another pending feature called tab conditions. Because the `ResourceTabs` do not have the same `resourceStore` as the `Form` component for the page settings, it is not possible to use the data in the `resourceStore` to evaluate if a tab should be shown or not.

#### BC Breaks/Deprecations

This will change the way the forms are configured.

#### To Do

- [x] Move metadata classes for forms from `ResourceMetadata` to `FormMetadata`
- [ ] Remove `FormResourceMetadataProvider` and `StructureResourceMetadataProvider`
- [ ] Remove `ResourceMetadataMapper`, `ResourceMetadataPool` and so on
- [x] Test returning actual content from `metadataAction`
- [x] ~~Also cache data in `FormMetadataProvider` to avoid unserializing data more often?~~
- [x] Remove `isSameEndpoint` function on `ResourceStore` and `ResourceMetadataStore`
- [x] Add `config/forms` as folder to discover form xml definitions (introduce in separate sulu-minimal PR)
- [x] Allow to extend forms using the same key and merge data
- [x] Introduce types in forms
- [x] Add Unit Test for `FormMetadataProvider`
- [x] Make `FormMetadataProvider` a `kernel.cache_warmer` (doesn't work currently, because system collections are currently not available at build time -> maybe also evaluate expression language later than at build time?)
- [x] Type `expression` on `param` instead of separate `expression` attribute?
- [x] Remove `Type` metadata in favour of `TypedForm`
- [x] Rename `page` to `pages` in FormMetadata (or make all formKeys singular?)
- [x] Reintroduce deleted `ResourceMetadataMapperTest` tests somewhere
- [x] Introduce `formKey` to load the form
- [x] Don't clone `resourceStore` if the same `resourceKey` is used among different tabs
- [x] ~~Reuse title from form XML instead of `tabTitle` on route options~~ (not easily possible because tabTitle has to be on route since non-forms can be child routes of the ResourceTab as well)
- [x] Test if preview requests in page are still sent
- [x] Add PR to make sulu-minimal compatible


src/Sulu/Bundle/AdminBundle/Metadata/Form/FormMetadataProvider.php
```diff
@@ -0,0 +1,386 @@
...
+        }
+
+        $structuresMetadataByTypes = [];
+        foreach ($this->structureMetadataFactory->getStructureTypes() as $structureType) {
+            foreach ($this->structureMetadataFactory->getStructures($structureType) as $structureMetadata) {
+                if ($structureMetadata->isInternal()) {
+                    continue;
+                }
+
+                $structuresMetadataByTypes[$structureType][] = $structureMetadata;
```
danrot: The `structureType` contains the values `home`, `page` and `snippet` from the DocumentManager in the default installation. 
There are two problems with that:

- The values are written in singular, whereby the keys in the frontend have been plural so far. I could use some Inflector service, or we decide for one or the other. Whereby I would probably feel less comfortable with changing the names in the DocumentManager...
- `page` and `home` are basically the same, and we either have to skip one of them here (question is how) or just ignore the fact that both are generated. Shouldn't be a too big problem in practice, because the metadata for `home` will never be called anyway.

src/Sulu/Bundle/AdminBundle/Metadata/Form/FormMetadataProvider.php
```diff
@@ -0,0 +1,386 @@
...
+
+            return new Property($propertyMetadata->getName(), $propertyMetadata->isRequired());
+        }, $propertyMetadata));
+
+        return new Schema($schemaProperties);
+    }
+
+    private function getConfigCache(string $key, string $locale): ConfigCache
+    {
+        return new ConfigCache(sprintf('%s%s%s.%s', $this->cacheDir, DIRECTORY_SEPARATOR, $key, $locale), $this->debug);
```
danrot: This and the usage of `Finder` makes it hard to test this class...

src/Sulu/Bundle/AdminBundle/Metadata/Form/FormMetadataProvider.php
chirimoya: - skip home
- use inflektor instead of renaming type

src/Sulu/Component/Util/SuluNodeHelper.php
```diff
@@ -107,11 +107,11 @@ public function getStructureTypeForNode(NodeInterface $node)
     {
         $mixinTypes = $node->getPropertyValueWithDefault('jcr:mixinTypes', []);
 
-        if (in_array('sulu:' . Structure::TYPE_PAGE, $mixinTypes)) {
```
chirimoya: revert

tests/Resources/DataFixtures/Webspace/invalid/massiveart.xml
```diff
@@ -9,7 +9,7 @@
     <theme>massiveart</theme>
 
     <default-templates>
-        <default-template type="page">default</default-template>
+        <default-template type="pages">default</default-template>
```
chirimoya: revert

chirimoya:
-  blocks are currently not working
-  cleanup XML format

PR #4300, opened 2018-11-28
implemented content-locale property in user store

nnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This PR adds a dummy `contentLocale` property to the `UserStore`. The `contentLocale` defines the locale in which localized content should be displayed.

The implementation in this PR returns a `contentLocale` fallback value all the time. In the future, the `contentLocale` should be updated and persisted when the user changes the displayed locale of a localized content in the UI.

#### Why?

Because we need a standardized way to access the locale in which localized content should be displayed.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/stores/UserStore/UserStore.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/stores/UserStore/UserStore.js
@@ -3,7 +3,10 @@ import 'core-js/library/fn/promise';
 import {action, computed, observable} from 'mobx';
 import {Config, Requester} from '../../services';
 import initializer from '../../services/Initializer';
+import localizationStore from '../LocalizationStore';
 import type {Contact, User} from './types';
+
+const CONTENT_LOCALE_SETTING_KEY = 'sulu_admin.user.content_locale';
 
 class UserStore {
     @observable persistentSettings: Map<string, string> = new Map();
@@ -28,6 +31,20 @@ class UserStore {
 
     @computed get systemLocale() {
         return this.user ? this.user.locale : Config.fallbackLocale;
+    }
+
+    @computed get contentLocale() {
+        const userLocale = this.getPersistentSetting(CONTENT_LOCALE_SETTING_KEY);
+
+        if (!userLocale) {
+            localizationStore.loadLocalizations().then((localizations) => {
+                const defaultLocalizations = localizations.filter((localization) => localization.default);
+                const fallbackLocalization = defaultLocalizations.length ? defaultLocalizations[0] : localizations[0];
+                this.setPersistentSetting(CONTENT_LOCALE_SETTING_KEY, fallbackLocalization.locale);
+            });
+        }
+
+        return userLocale ? userLocale : Config.fallbackLocale;
     }
 
     @action setLoggedIn(loggedIn: boolean) {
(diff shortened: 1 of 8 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/stores/UserStore/UserStore.js
```diff
@@ -26,10 +29,24 @@ class UserStore {
...
         return this.user ? this.user.locale : Config.fallbackLocale;
     }
 
+    @computed get contentLocale() {
+        const userLocale = this.getPersistentSetting(CONTENT_LOCALE_SETTING_KEY);
+
+        if (!userLocale) {
+            localizationStore.loadLocalizations().then((localizations) => {
+                const defaultLocalizations = localizations.filter((localization) => localization.default);
+                const fallbackLocalization = defaultLocalizations.length ? defaultLocalizations[0] : localizations[0];
```
danrot: Why don't you just return this value? Would be enough for now, and we are not sure yet if the locale will be read from the user settings afterwards.

src/Sulu/Bundle/MediaBundle/Resources/js/containers/Form/fields/MediaSelection.js
```diff
@@ -16,12 +18,9 @@ export default class MediaSelection extends React.Component<FieldTypeProps<Value
...
         const {formInspector, disabled, value} = this.props;
-
-        if (!formInspector || !formInspector.locale) {
-            throw new Error('The media selection needs a locale to work properly');
-        }
-
-        const {locale} = formInspector;
+        const locale = formInspector && formInspector.locale
+            ? formInspector.locale
+            : computed(() => userStore.contentLocale);
```
danrot: Why the `computed`? Same question for all the other components

src/Sulu/Bundle/MediaBundle/Resources/js/containers/Form/fields/MediaSelection.js
```diff
@@ -16,12 +18,9 @@ export default class MediaSelection extends React.Component<FieldTypeProps<Value
 
     render() {
         const {formInspector, disabled, value} = this.props;
-
-        if (!formInspector || !formInspector.locale) {
-            throw new Error('The media selection needs a locale to work properly');
-        }
-
-        const {locale} = formInspector;
+        const locale = formInspector && formInspector.locale
```
danrot: Is the check for `formInspector` really necessary? I think the `formInspector` should always be there, so only `formInspector.locale` should be enough...

src/Sulu/Bundle/MediaBundle/Resources/js/containers/Form/fields/MediaSelection.js
nnatter: the media-bundle components expect an observable/boxed string as locale. this is because they need to update the language of the displayed medias when the locale of the form is changed.

the `computed` function returns a boxed observable based on the return value of the given function. furthermore, the returned observable will update itself when an observable that is used in the function (in this case the `contentLocale` of the `userStore`) changes.

nnatter: Unfortunatelly, the object returned by the `computed` function does not provide an `intercept` method and therefore does not work inside a `DatagridStore` 🙁 Therefore i changed the components to use `observable.box`.

PR #4301, opened 2018-11-28
implemented value prop for action component 

nnatter:
…en onClick callback

| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This PR adds a `value` prop to the `Action` component. The given `value` is passed to the given `onClick` callback on click.

#### Why?

Because it might be useful to pass a value to the given `onClick`handler.


danrot:
Code looks good! But I would add a separate exampled to the `README` of that component - then people can easily discover that it is there, and I can easily test it now :wink:

PR #4306, opened 2018-12-01
Add warning message when try to install npm dependencies using yarn

alexander-schranz:
# ~~Fix building sulu with yarn~~

| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Related PRs | https://github.com/sulu/sulu-minimal/pull/154
| License | MIT

#### What's in this PR?

~~Fix building sulu with yarn.~~

Add warning message when try to install npm dependencies using yarn 

#### Why?

~~If somebody want to use yarn to build sulu he should not end up with an error.~~

We will currently not supporting yarn because of yarn is not supporting symlinks:

#### Example Usage

~~~php
yarn install
~~~

> You must use npm to install, yarn is not supported

#### TODO

 - [x] Create sulu-minimal PR where adding preinstall.js

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,9 @@ jobs:
       - image: circleci/node:10.9
     steps:
       - checkout
-      # Node
+      - run:
+            name: Check error for yarn install
+            command: tests/js/check-yarn-warning.sh
       - run: npm install
       - run: npm run lint:js
       - run: npm run lint:scss
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,7 +1,6 @@
 /* eslint-disable flowtype/require-valid-file-annotation */
 /* eslint-disable import/no-nodejs-modules*/
 const path = require('path');
-const glob = require('glob');
 const webpack = require('webpack');
 const CleanObsoleteChunksPlugin = require('webpack-clean-obsolete-chunks');
 const CleanWebpackPlugin = require('clean-webpack-plugin');
@@ -19,9 +18,15 @@ module.exports = (env, argv) => { // eslint-disable-line no-undef
         publicDir = composerConfig.extra['public-dir'];
     }
 
-    const entries = glob.sync(
-        path.resolve(__dirname, 'src/Sulu/Bundle/*/Resources/js/index.js') // eslint-disable-line no-undef
-    );
+    const entries = [];
+
+    entries.unshift('sulu-admin-bundle');
+    entries.unshift('sulu-contact-bundle');
+    entries.unshift('sulu-media-bundle');
+    entries.unshift('sulu-page-bundle');
+    entries.unshift('sulu-preview-bundle');
+    entries.unshift('sulu-security-bundle');
+
     const entriesCount = entries.length;
 
     entries.unshift('core-js/fn/array/includes');
@@ -57,7 +62,7 @@ module.exports = (env, argv) => { // eslint-disable-line no-undef
             rules: [
                 {
                     test: /\.js$/,
-                    exclude: /node_modules/,
+                    exclude: /node_modules\/(?!(sulu-(.*)-bundle|@ckeditor|lodash-es)\/)/,
                     use: 'babel-loader',
                 },
                 {
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@ composer.lock
 # npm
 package-lock.json
 /node_modules
+yarn.lock
 
 # phpunit
 .phpunit
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -2,6 +2,16 @@
 
 ## dev-develop
 
+### Add sulu preinstall script to your package.json
+
+Sulu will check if the dependencies are correctly install in a preinstall script
+add this to your project `package.json`:
+
+```diff
+     "scripts": {
++        "preinstall": "node vendor/sulu/sulu/preinstall.js",
+```
+
 ### Type information of contacts
 
 The contacts have different sub entities, which are assigned in combination with a type (e.g. phone numbers, fax number,
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
     "license": "MIT",
     "repository": "https://github.com/sulu/sulu.git",
     "scripts": {
+        "preinstall": "node preinstall.js",
         "lint:js": "eslint .",
         "lint:scss": "stylelint src/Sulu/Bundle/*/Resources/js/**/*.scss",
         "flow": "flow",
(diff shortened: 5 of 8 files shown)


webpack.config.js
```diff
@@ -1,16 +1,20 @@
...
-const entries = glob.sync(
-    path.resolve(__dirname, 'src/Sulu/Bundle/*/Resources/js/index.js') // eslint-disable-line no-undef
-);
+const entries = [];
+
+entries.unshift('sulu-admin-bundle');
+entries.unshift('sulu-content-bundle');
+entries.unshift('sulu-media-bundle');
+entries.unshift('sulu-preview-bundle');
+entries.unshift('sulu-security-bundle');
```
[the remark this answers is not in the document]
danrot: I like the change because these files are contained within our control anyway. The glob would only make sense if files from 3rd party bundles would be included as well. Maybe we should even change the `glob` to do so? Already thought about renaming `index.js` to `sulu-index.js`, so that it could be added to any bundle and it would be automatically discovered, making it easier to use.

But i don't see how this solves problems when moving files to `assets/admin`...

webpack.config.js
```diff
@@ -47,7 +51,7 @@ module.exports = (env, argv) => ({ // eslint-disable-line no-undef
         rules: [
             {
                 test: /\.js$/,
-                exclude: /node_modules/,
+                exclude: /node_modules\/(?!(sulu-(.*)-bundle|@ckeditor|lodash-es)\/)/,
```
danrot: Are these folders now correctly symlinked by both npm and yarn? Meaning that changes apply immediately when you are editing JS files from this repository?

config/routes/sulu_admin.yaml
```diff
@@ -75,6 +75,10 @@ sulu_route_api:
     resource: "@SuluRouteBundle/Resources/config/routing_api.xml"
     prefix: /admin/api
 
+sulu_preview:
+    resource: "@SuluPreviewBundle/Resources/config/routing.xml"
+    prefix: /admin/preview
```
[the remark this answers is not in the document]
alexander-schranz: yes only to run webserver directlry from sulu/sulu

.circleci/config.yml
```diff
@@ -1,15 +1,29 @@
...
+      - checkout
+      - run: yarn install
+      - run: yarn build
+
+workflows:
+  version: 2
+  test_and_yarn:
+    jobs:
+      - test
+      - yarn
```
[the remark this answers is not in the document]
alexander-schranz: parallel as the job has no requirement to another job

webpack.config.js
alexander-schranz: > But i don't see how this solves problems when moving files to assets/admin...

To the `glob` now NPM now that its a js file from external library defined in package.json and not a from the current project. So it resolves its requirements correctly thats why it solves problems moving files to assets/admin. `glob` should not be used to load things outside your project directory. and yes you are correctly we should remove glob from the requirements.

> Are these folders now correctly symlinked by both npm and yarn?  Meaning that changes apply immediately when you are editing JS files from this repository?

Yarn does never symlink this is just to let yarn work for the project where you don't change sulu/sulu code. Need to test if the watch task still works correctly for the core developer.

webpack.config.js
```diff
@@ -1,16 +1,20 @@
 /* eslint-disable flowtype/require-valid-file-annotation */
 /* eslint-disable import/no-nodejs-modules*/
 const path = require('path');
-const glob = require('glob');
```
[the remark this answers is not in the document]
alexander-schranz: glob is still used in the styleguide config so it can't be removed

alexander-schranz:
@danrot PR is rebased

webpack.config.js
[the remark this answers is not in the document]
alexander-schranz: couldn't find any problems with cache as npm symlinks and yarn seems to copy the folder always.

alexander-schranz:
@danrot couldn't find any problems with cache as npm symlinks and yarn copy the folder always.

danrot:
Okay, so maybe they changed the behavior. But you still have to execute `yarn` to get the changes you've done in your own filesystem, right? And maybe even clear the `node_modules` folder before that?

alexander-schranz:
@danrot @chirimoya as discussed we currently will not support yarn because of problems running install always. So I refractored this PR for showing a warning message when a user try to use yarn to install dependencies. It still contains some dependency fixes so webpack know that the bundles are own libraries and has there own dependencies.

webpack.config.js
```diff
@@ -57,7 +61,7 @@ module.exports = (env, argv) => { // eslint-disable-line no-undef
             rules: [
                 {
                     test: /\.js$/,
-                    exclude: /node_modules/,
+                    exclude: /node_modules\/(?!(sulu-(.*)-bundle|@ckeditor|lodash-es)\/)/,
```
danrot: Why is that suddenly necessary?

webpack.config.js
```diff
@@ -19,9 +18,14 @@ module.exports = (env, argv) => { // eslint-disable-line no-undef
...
-    const entries = glob.sync(
-        path.resolve(__dirname, 'src/Sulu/Bundle/*/Resources/js/index.js') // eslint-disable-line no-undef
-    );
+    const entries = [];
+
+    entries.unshift('sulu-admin-bundle');
+    entries.unshift('sulu-page-bundle');
+    entries.unshift('sulu-media-bundle');
+    entries.unshift('sulu-preview-bundle');
+    entries.unshift('sulu-security-bundle');
```
danrot: I am still not sure if I would remove `glob` here... Actually I even thought about expanding the `glob`, so that even 3rd party bundles would not have to touch the webpack config, making it easier for developers to start...

webpack.config.js
alexander-schranz: yes it is, its the same regex as used in transformIgnorePatterns.

webpack.config.js
alexander-schranz: The glob is the evil about the problems moving to assets/admin and have to define all dependencies also in the project because NPM think this not third party libraries and don't find its dependencies correctly.

danrot: Okay, that makes sense... However, you would still have to add `sulu-contact-bundle` now, since it was recently added :-)

webpack.config.js
danrot: I asked why it is suddenly necessary... Because it seems like it was not necessary before (the build was working without these lines... Is it because we are including the files by their package name instead of their path now?

webpack.config.js
alexander-schranz: @danrot added

webpack.config.js
alexander-schranz: yes because of that impports, the other (ckeditor, lodash) I added as the regex in the package.json did that also.

PR #4310, opened 2018-12-04
Evaluate expressions in params on the fly

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | #4299
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR will make the evaluation of the expressions in forms/structures everytime, instead of once at build time, and cache it afterwards.

#### Why?

This way it is possible to e.g. access the database and load values from there. Apart from that it will fix the bug in #4299.

#### To Do


danrot:
Doesn't make sense to do without #4299, so it is included there.

PR #4312, opened 2018-12-05
Floating problem in forms

luca-rath:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no

#### What's in this PR?

The main problem is, that if a form field is higher than it's next sibling, the layout can break. (Example below)
This PR sets a fixed height of 30px to the checkbox component. But this doesn't fix the problem at all. Using flex instead of float would probably solve this problem, but this needs further discussion.

#### Why?

Imagine the following template:
```xml
<section name="test">
    <meta>
        <title>Test</title>
    </meta>

    <properties>
        <property name="text1" type="text_line" colspan="6">
            <meta>
                <title>Text 1</title>
            </meta>
        </property>

        <property name="check1" type="checkbox" colspan="6">
            <meta>
                <title>Checkbox 1</title>
            </meta>
        </property>

        <property name="check2" type="checkbox" colspan="6">
            <meta>
                <title>Checkbox 2</title>
            </meta>
        </property>
    </properties>
</section>
```

The result you would expect:
image

But this is what you get:
image

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Switch/switch.scss
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Switch/switch.scss
@@ -4,6 +4,7 @@ $marginBottom: 10px;
     display: flex;
     align-items: center;
     cursor: pointer;
+    height: 30px;
 
     & > div {
         font-size: 14px;


danrot:
Thank you! As we have already discussed, there are also use cases where you don't want it to behave like a flexbox, that's why we chose not to use it. The thing is that there also some components that might be much higher than 30px, and then you might want there to be multiple checkbox next to it, not only one, since that would be a waste of space. So I don't think we are going to fix that in a general manner, but we have to take care that stuff is working quite nice together.

PR #4313, opened 2018-12-06
Fixed php 7.3 compatibility

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | **`Yes`**
| Deprecations? | no
| Related PRs | #3814, #3627
| License | MIT

#### What's in this PR?

Fixed PHP 7.3 compatibility on the #SymfonyConHackday2018, some BC breaks needed to done to upgrade doctrine.

This can't be fixed for 1.5 because doctrine 7.3 compatibility needs atleast symfony 3 support.

#### Why?

Test sulu on newest version.


appveyor.yml
```diff
@@ -8,10 +8,12 @@ environment:
     # see http://windows.php.net/downloads/releases/ for available versions
     - PHP_VERSION: 5.6.33
       SYMFONY__DATABASE__PASSWORD: Password12!
+      SYMFONY_DEPRECATIONS_HELPER: weak
     - PHP_VERSION: 7.1.14
       JACKRABBIT_VERSION: 2.12.0
       SYMFONY__PHPCR__TRANSPORT: jackrabbit
       SYMFONY__DATABASE__PASSWORD: Password12!
+      SYMFONY_DEPRECATIONS_HELPER: weak
```
danrot: Once is probably enough, isn't it? :slightly_smiling_face:

composer.json
```diff
@@ -39,7 +39,6 @@
         "pulse00/ffmpeg-bundle": "^0.6",
         "oro/doctrine-extensions": "^1.0",
         "goodby/csv": "^1.3",
-        "doctrine/data-fixtures": "1.1.*",
```
danrot: Don't we need that anymore?

alexander-schranz: @danrot comes with the doctrine/doctrine-fixtures-bundle and needs a higher versions. It seems we did set it to 1.1.* because 1.2.* did drop the php 5.5. support there where we where still no compatible to 7.0

PR #4317, opened 2018-12-10
Fixed getProperties on null

marcostastny:
Fixes a bug, where sulu backend would break, if you mess around with snippets during developement.

| Q | A
| --- | ---
| Bug fix? | Yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Contains a fix for sulu backend snippet handling.

#### Why?

Fixes a bug, where sulu backend would break, if you have added a snippet of a certain type and then remove this type (e.g.: For developement purposes)

### DIFF
--- a/src/Sulu/Component/Content/Document/Structure/ManagedStructure.php
+++ b/src/Sulu/Component/Content/Document/Structure/ManagedStructure.php
@@ -179,6 +179,9 @@ class ManagedStructure extends Structure
     {
         $this->init();
         $values = [];
+        if (is_null($this->structureMetadata)) {
+            return $values;
+        }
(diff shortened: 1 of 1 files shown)


danrot:
This somehow breaks the tests... And usually we always add tests if we fix a bug :-) And I can't really follow what the exact problem is. Maybe we could have a talk about that the next time I am in your office?

nnatter:
This error usually happens if you are trying to load a page or a snippet of a type that does not exist in the application. For example, this can happen when removing a template from a project without migrating the data. 

That said, I will close this because of inactivity 🙂

PR #4265, opened 2018-11-10
Added google-cloud storage for media

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #1137
| Related issues/PRs | requires #4264 and requires https://github.com/guzzle/psr7/issues/200
| License | MIT
| Documentation PR | https://github.com/sulu/sulu-docs/pull/412

#### What's in this PR?

This PR add google-cloud as storage for medias.

#### Why?

This will be the first external adapter to store medias.

#### BC Breaks/Deprecations

The storage configurations for `sulu_media`

#### To Do

- [x] Add breaking changes to UPGRADE.md
- [x] Add suggestions
- [x] Public permission for original url
- [x] Tests
- [x] Add documentation PRs (https://github.com/sulu/sulu-docs/pull/412)
- [x] 1.5.0 for psr7 https://github.com/guzzle/psr7/issues/200

PR #4321, opened 2018-12-15
Add s3 media storage

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | none
| Related issues/PRs | built ontop #4265 
| License | MIT
| Documentation PR | https://github.com/sulu/sulu-docs/pull/423

#### What's in this PR?

This PR add s3 as external storage

#### Why?

This can be used to upload media to aws or any other aws-s3 compatible storage.


wachterjohannes:
@danrot could you take a look at the configuration? the question is do we add storage providers (configuration) for each aws s3 compatible provider or do we add some additional configuration for the aws provider? /cc @alexander-schranz @chirimoya 

When we have decided that i will create the docs-pr.

src/Sulu/Bundle/MediaBundle/DependencyInjection/Configuration.php
```diff
@@ -204,6 +205,38 @@ private function addStorageSection(ArrayNodeDefinition $node)
...
+            $storages[] = 'minio';
+            $storagesNode
+                ->arrayNode('minio')
+                    ->addDefaultsIfNotSet()
+                    ->children()
+                        ->scalarNode('key')->isRequired()->end()
+                        ->scalarNode('secret')->isRequired()->end()
+                        ->scalarNode('endpoint')->isRequired()->end()
+                        ->scalarNode('bucket_name')->isRequired()->end()
+                        ->scalarNode('region')->defaultValue('us-east-1')->end()
```
alexander-schranz: why here a default?
and would be better if the order of the parameters are the same as above

src/Sulu/Bundle/MediaBundle/DependencyInjection/Configuration.php
```diff
@@ -204,6 +205,38 @@ private function addStorageSection(ArrayNodeDefinition $node)
...
+            $storagesNode
+                ->arrayNode('aws_s3')
+                    ->addDefaultsIfNotSet()
+                    ->children()
+                        ->scalarNode('key')->isRequired()->end()
+                        ->scalarNode('secret')->isRequired()->end()
+                        ->scalarNode('region')->isRequired()->end()
+                        ->scalarNode('bucket_name')->isRequired()->end()
+                        ->scalarNode('version')->defaultValue('latest')->end()
+                        ->scalarNode('endpoint')->defaultValue('latest')->end()
```
alexander-schranz: latest endpoint? looks false fore me?

alexander-schranz:
ping @wachterjohannes

src/Sulu/Bundle/MediaBundle/DependencyInjection/Configuration.php
wachterjohannes: this is the default value in minio configuration but it can be set individually.

src/Sulu/Bundle/MediaBundle/DependencyInjection/Configuration.php
wachterjohannes: that was wrong 🙈

src/Sulu/Bundle/MediaBundle/Media/Storage/S3Storage.php
```diff
@@ -0,0 +1,57 @@
...
+     * @var string
+     */
+    private $bucketName;
+
+    public function __construct(FilesystemInterface $filesystem, int $segments)
+    {
+        parent::__construct($filesystem, $segments);
+
+        if (!$filesystem instanceof Filesystem || !$filesystem->getAdapter() instanceof AwsS3Adapter) {
+            throw new \RuntimeException();
```
danrot: Wouldn't an exception text be nice? Would at the same time be a documentation for the one who reads the code. Is that something that is not supposed to happen? Also that could go into the message.

wachterjohannes:
@danrot fixed and added documentation

mareksuscak:
Was wondering if it's planned to support S3 media storage (this feature) in Sulu 1.6? Seems like develop branch contains Sulu v2 changes, correct?

alexander-schranz:
@mareksuscak sorry because of BC Promise its not possible to backport this changes to 1.6. But there is a project from @pawel-brzezinski which implements something like that. Keep in mind that the bundle overwrites sulu services and needed to duplicated some code to support this type of storage so you maybe don't get all fixes/changes when doing a sulu update.

In some cloud hosting services you also don't need this kind of implementation in php you can just tell that a specific folder in the operation system is a S3 or cloud storage so no changes are needed. But this depends on your hosting.

mareksuscak:
Thanks for your response. Yes we are using that 3rd party extension and one of our former colleagues customized it. The reason why we were looking to use a standard module is ease of maintenance and an easier way to migrate to new versions as you stated.

PR #4323, opened 2018-12-17
Add apiOptions-option to form-view

nnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This PR add an `apiOptions` option to the `Form` view that allows to set attributes which are then passed to the `FormStore`. The attributes are passed to the `FormStore` analog to the attributes specified in the `routerAttributesToFormStore` option.

#### Why?

Because it might be useful to define some data in the route definition that is sent to the server when the `Form` is saved.

#### Example Usage

~~~php
$route = new Route('sulu_category.add_form', '/categories/:locale/add', 'sulu_admin.resource_tabs');
$route->addOption('resourceKey', 'categories')
$route->addOption('toolbarActions', $formToolbarActions)
$route->addOption('apiOptions', ['apiKey' => 'value-that-is-sent-as-query-param'])
~~~

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/views/Form/Form.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/views/Form/Form.js
@@ -67,6 +67,7 @@ class Form extends React.Component<Props> {
                 options: {
                     idQueryParameter,
                     resourceKey,
+                    apiOptions,
                     routerAttributesToFormStore,
                 },
             },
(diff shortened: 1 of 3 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/views/Form/Form.js
```diff
@@ -67,6 +67,7 @@ class Form extends React.Component<Props> {
                 options: {
                     idQueryParameter,
                     resourceKey,
+                    apiOptions,
                     routerAttributesToFormStore,
```
nnatter: might be more consistent to rename this option to something like `routerAttributesToApiOptions`?

danrot:
I think we've agreed on a PR title mechanism last week, updated it for you :wink:

src/Sulu/Bundle/AdminBundle/Resources/js/views/Form/Form.js
danrot: I like the `ToFormStore` suffix, because there is also `routerAttributesToEditRoute`, so the naming is consistent there. Question is if we maybe shouldn't try to have the same naming as in the Datagrid view, and adapt the term `apiOptions` to better match the existing options? But I think it would also be fine for me the way it is now.

PR #4333, opened 2018-12-20
Fix checkbox height

luca-rath:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Related issues/PRs | #4312 

#### What's in this PR?

Removing 30px height of checkbox component and add a min-height of 30px for all field types.
Also added an invisible character to the after pseudo element of labels. The reason is, that labels had a height of 0, when they had no content.


luca-rath:
@danrot

PR #4336, opened 2018-12-20
Cleanup tag bundle dependencies

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Cleanup tag bundle dependencies.

#### Why?

Less dependencies = less circular reference problems in tag manager.

#### BC Breaks/Deprecations

See UPGRADE.md

#### To Do

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -2,6 +2,18 @@
 
 ## dev-develop
 
+### TagBundle services changed
+
+Following Services changed:
+
+ - `sulu_tag.tag_repository` removed use `sulu.repository.tag`
+
+Following Function removed:
+
+ - `TagManagerInterface::getFieldDescriptors` and `TagManagerInterface::getFieldDescriptor` removed use `FieldDescriptorFactory::getFieldDescriptorForClass` instead
+
+The `TagManager` doesn't longer depend on FieldDescriptorFactory, TagEntityName and UserRepository so its constructor changed.
+
 ### Change tag_list to tag_selection
 
 The tag_list content and field type was changed to `tag_selection`
--- a/src/Sulu/Bundle/TagBundle/Resources/config/services.xml
+++ b/src/Sulu/Bundle/TagBundle/Resources/config/services.xml
@@ -2,9 +2,6 @@
 <container xmlns="http://symfony.com/schema/dic/services"
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
     xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
-    <parameters>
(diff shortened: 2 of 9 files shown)


src/Sulu/Bundle/TagBundle/Resources/config/services.xml
```diff
@@ -3,43 +3,30 @@
...
             <tag name="sulu.admin"/>
             <tag name="sulu.context" context="admin"/>
             <argument type="service" id="sulu_security.security_checker"/>
         </service>
-        <service id="sulu_tag.tag_repository" class="%sulu_tag.tag_repository.class%">
-            <factory method="getRepository" service="doctrine.orm.entity_manager"/>
-            <argument>%sulu_tag.entity.tag%</argument>
-        </service>
-        <service id="sulu_tag.tag_manager" class="%sulu_tag.tag_manager.class%" public="true">
+        <service id="sulu_tag.tag_manager" class="Sulu\Bundle\TagBundle\Tag\TagManager" public="true">
```
danrot: Does it need to be public? From my understanding we were only doing this for the services that absolutely need it.

danrot:
Oh, and the tests seem to be failing.

src/Sulu/Bundle/TagBundle/Resources/config/services.xml
alexander-schranz: Was public before (little tricky in this diff as I removed the repository).
Currently it needs to be public as it is used in `TagController` and `SmartContentItemController`

UPGRADE.md
```diff
@@ -2,6 +2,17 @@
...
+
+Following Services changed:
+
+ - `sulu_tag.tag_repository` removed use `sulu.repository.tag`
+
+Following Function removed:
+
+ - `TagManagerInterface::getFieldDescriptors` and `TagManagerInterface::getFieldDescriptor` removed use `FieldDescriptorFactory::getFieldDescriptorForClass` instead
+
+The `TagManager` doesn't longer depend on FieldDescriptorFactory, TagEntityName and UserRepository so its constructor changed.
```
danrot: Use backticks for the latter 3 classes as well.

src/Sulu/Bundle/TagBundle/Tag/TagManager.php
```diff
@@ -172,16 +119,13 @@ public function save($data, $userId, $id = null)
                 $tag = $this->tagRepository->createNew();
             }
 
-            $user = $this->userRepository->findUserById($userId);
-
             // update data
             $tag->setName($name);
-            $tag->setChanger($user);
```
danrot: Just a small reminder for me: Test if this still works.

PR #4338, opened 2018-12-21
Fix accidently remove of build directory by IDE

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Change rimraf call be called before plugins by moving it to the plugin configuration.

#### Why?

Some IDE's e.g. PhpStorm parse the webpack config and calls rimraf which will remove all files in the build folder.


webpack.config.js
```diff
@@ -54,6 +51,10 @@ module.exports = (env, argv) => { // eslint-disable-line no-undef
         },
         devtool: argv.mode === 'development' ? 'eval-source-map' : 'source-map',
         plugins: [
+            () => {
```
danrot: I am not sure if plugins are written this way resp. if this has other unwanted side effects... Maybe we should use something like the [`WebpackShellPlugin`](https://github.com/1337programming/webpack-shell-plugin)? Or maybe there is an even better solution, since I can't believe we are the only people with this problem :confused:

danrot:
This does not clean the build folder for me anymore... I get the following message when executing `npm run build`:

```
clean-webpack-plugin: /home/daniel/Development/sulu/sulu-minimal/public/build/admin is outside of the project root. Skipping...
```

alexander-schranz:
@danrot seems like I need to activate allowExternal in our case

danrot:
Doesn't that mean that the wrong folder could be deleted? That's something we should definitely avoid... Can't we somehow change the root directory instead, so that the plugin thinks it is in the correct folder?

danrot:
Resp. wouldn't the `publicDir` variable already existing be a good match to set as root directory?

alexander-schranz:
@danrot as discussed I give now the rootDirectory as variable form the project directory: https://github.com/sulu/sulu-minimal/pull/151

danrot:
@alexander-schranz Not really related to this PR, but would still be a nice fit: I think we should add a `npm run` task to circleci, because at the moment we don't really know if the build is succeeding :see_no_evil:

PR #4339, opened 2018-12-21
Store value of MediaSelection field-type to ResourceStore in correct order 

nnatter:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This PR adjusts the `MultiMediaSelection` component and the `MultiMediaSelectionOverlay` component to sort copies of arrays instead of the original arrays to preserve the user-sorting.

Furthermore, this PR replaces the `autorun` of the `MultiMediaSelection` that is used to track changes in the `MultiMediaSelectionOverlay` with a `reaction` that allows for finer grained state-dependencies. This prevents a superfluous call of the change-handler of the `MultiMediaSelection`.

#### Why?

The `MultiMediaSelection` component allows the user to sort the selected medias in the frontend.  In the current implementation, changing the order of the frontend fires the `componentDidUpdate` method of the `MultiMediaSelection` component, a rerender of the `MultiMediaSelection` component and therefore a rerender of the `MultiMediaSelectionOverlay` component.

Furthermore, the `componentDidUpdate` method of the `MultiMediaSelection` component and the rerender of the `MultiMediaSelectionOverlay` component call sort on the `selectedMedia` array of the `MultiMediaSelectionStore`. 

The change of this array then fires the `change-autorun` of the `MultiMediaSelection` component which sets the alphabetically sorted value (instead of the user-sorted value) to the `MediaSelection` field type. Therefore, it was not possible to change the order of the selected medias in the frontend.

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/Resources/js/containers/MultiMediaSelection/MultiMediaSelection.js
+++ b/src/Sulu/Bundle/MediaBundle/Resources/js/containers/MultiMediaSelection/MultiMediaSelection.js
@@ -1,6 +1,6 @@
 // @flow
 import React, {Fragment} from 'react';
-import {action, autorun, toJS, observable, untracked} from 'mobx';
+import {action, reaction, toJS, observable} from 'mobx';
 import {observer} from 'mobx-react';
 import equals from 'fast-deep-equal';
 import {MultiItemSelection} from 'sulu-admin-bundle/components';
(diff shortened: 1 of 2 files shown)


nnatter:
this might be an issue on other components that are sortable via the frontend too

PR #4342, opened 2018-12-30
Make logger service optional

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | yes constructors changed
| Deprecations? | no
| License | MIT

#### What's in this PR?

Make logger service optional and use `PSR\Log\NullLogger` when not available.

#### Why?

Currently sulu container does not build without monolog bundle installed. So in many cases like in bundles which does not need monolog need to require monolog for there tests:

e.g.:

 - [Community Bundle](https://github.com/sulu/SuluCommunityBundle/blob/8c67eb0fc4970064b36c4a3b827c17bd7fce4187/composer.json#L35)
 -  [SuluFormBundle](https://github.com/sulu/SuluFormBundle/blob/7bc6e35d6d3619f3d0de9249ed3bbbb52f367122/composer.json#L15)
 - [SuluArticleBundle](https://github.com/sulu/SuluArticleBundle/blob/5f6b4e90821f71da9179aeb7ca642d88fba1bce1/composer.json#L18)

#### To Do

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -2,6 +2,20 @@
 
 ## dev-develop
 
+### Logger is optional
+
+As the logger is know optional in the following services the constructor changed and logger is injected
+over the `Psr/Log/LoggerAwareTrait::setLogger`:
+
+ - `sulu.content.type.internal_links`
+ - `sulu_content.compat.structure.legacy_property_factory`
+ - `sulu_content.node_repository`
+ - `sulu_media.search.subscriber.media`
+ - `sulu_media.format_manager`
+ - `sulu_media.storage.local`
+ - `sulu_snippet.import.snippet`
+ - `sulu_website.twig.content`
+
 ### TagBundle services changed
 
 Following Services changed:
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -4,8 +4,7 @@
 
 ### Logger is optional
 
-As the logger is know optional in the following services the constructor changed and logger is injected
-over the `Psr/Log/LoggerAwareTrait::setLogger`:
+As the logger is now optional in the following services the constructor changed:
 
  - `sulu.content.type.internal_links`
  - `sulu_content.compat.structure.legacy_property_factory`
--- a/src/Sulu/Bundle/ContentBundle/Repository/NodeRepository.php
+++ b/src/Sulu/Bundle/ContentBundle/Repository/NodeRepository.php
@@ -12,9 +12,6 @@
 namespace Sulu\Bundle\ContentBundle\Repository;
 
 use PHPCR\RepositoryException;
-use Psr\Log\LoggerAwareTrait;
-use Psr\Log\LoggerInterface;
-use Psr\Log\NullLogger;
 use Sulu\Bundle\AdminBundle\UserManager\UserManagerInterface;
 use Sulu\Bundle\ContentBundle\Content\InternalLinksContainer;
 use Sulu\Component\Content\Compat\StructureInterface;
@@ -86,11 +83,6 @@ class NodeRepository implements NodeRepositoryInterface
      */
     private $tokenStorage;
 
-    /**
-     * @var LoggerInterface
-     */
-    private $logger;
-
     public function __construct(
         ContentMapperInterface $mapper,
         SessionManagerInterface $sessionManager,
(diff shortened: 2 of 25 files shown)


UPGRADE.md
```diff
@@ -2,6 +2,19 @@
 
 ## dev-develop
 
+### Logger is optional
+
+As the logger is know optional in the following services the constructor changed and logger is injected
```
danrot: `now optional`

src/Sulu/Bundle/ContentBundle/Repository/NodeRepository.php
```diff
@@ -108,7 +105,7 @@ public function __construct(
         $this->queryExecutor = $queryExecutor;
         $this->accessControlManager = $accessControlManager;
         $this->tokenStorage = $tokenStorage;
-        $this->logger = $logger;
```
danrot: Why do we use now setter injection instead? I'd prefer constructor injection, and write this line as follows:

```php
$this->logger = $logger ?? new NullLogger();
```

With your current implementation we are always creating a `NullLogger` just to be replaced afterwards.

src/Sulu/Bundle/ContentBundle/Resources/config/export.xml
```diff
@@ -34,8 +34,10 @@
             <argument type="service" id="sulu.content.structure_manager" />
             <argument type="service" id="sulu_content.extension.manager" />
             <argument type="service" id="sulu_content.import.manager" />
-            <argument type="service" id="logger" />
             <argument type="service" id="sulu_content.import.webspace.xliff12" />
+            <call method="setLogger">
+                <argument type="service" id="logger" on-invalid="ignore"/>
```
danrot: This `on-invalid` could also be added to the constructor argument, I guess.

src/Sulu/Bundle/DocumentManagerBundle/Resources/config/core.xml
```diff
@@ -7,7 +7,7 @@
         <!-- Core -->
         <service id="sulu_document_manager.event_dispatcher.debug" class="Sulu\Component\DocumentManager\EventDispatcher\DebugEventDispatcher" public="false">
             <argument type="service" id="debug.stopwatch" />
-            <argument type="service" id="logger" />
+            <argument type="service" id="logger" on-invalid="null" />
```
danrot: This also doesn't seem to be consistent.

src/Sulu/Component/DocumentManager/EventDispatcher/DebugEventDispatcher.php
```diff
@@ -37,10 +38,10 @@ class DebugEventDispatcher extends EventDispatcher
      */
     public function __construct(
         Stopwatch $stopwatch,
-        LoggerInterface $logger
+        LoggerInterface $logger = null
     ) {
         $this->stopwatch = $stopwatch;
-        $this->logger = $logger;
+        $this->logger = $logger ?: new NullLogger();
```
danrot: That's the version I'd prefer :slightly_smiling_face:

src/Sulu/Bundle/DocumentManagerBundle/Resources/config/core.xml
alexander-schranz: @danrot the debug event dispatcher is basically for logging so it would be strange not having it in the constructor there.

src/Sulu/Bundle/ContentBundle/Repository/NodeRepository.php
alexander-schranz: @danrot I'm basically a friend of the `LoggerAwareInterface` and not having the logger in the constructor because I did see some strange constructor like this in the past:

```
__construct(
    SomeDependency $someDependency,
    LoggerInterface $logger = null,
    $parameter,
    $moreParameter
)
```

which basically will tell you in the IDE / code analyser optional parameters can not be in the middle of function, but if we say logger is always the last param of the constructor I'm fine with it.

danrot: Then I would say we make it the last argument :slightly_smiling_face:

UPGRADE.md
```diff
@@ -2,6 +2,20 @@
 
 ## dev-develop
 
+### Logger is optional
+
+As the logger is now optional in the following services the constructor changed and logger is injected
+over the `Psr/Log/LoggerAwareTrait::setLogger`:
```
danrot: That's not true anymore, right?

alexander-schranz: fixed

src/Sulu/Bundle/ContentBundle/Repository/NodeRepository.php
```diff
@@ -108,7 +109,7 @@ public function __construct(
...
+        $this->logger = $logger ?: new NullLogger();
```
danrot: Doesn't seem to be used at all? :thinking:

PR #4348, opened 2019-01-07
Fix cache invalidation and add tag invalidation

trickreich:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes --
| Related issues/PRs | --
| License | MIT
| Documentation PR | --

#### What's in this PR?

Add tag invalidation for Symfony http cache.
Also fixes not working invalidation at all.

#### Why?

Tag invalidation is nice! ;)

### DIFF
--- a/src/Sulu/Bundle/HttpCacheBundle/Cache/SuluHttpCache.php
+++ b/src/Sulu/Bundle/HttpCacheBundle/Cache/SuluHttpCache.php
@@ -12,7 +12,6 @@
 namespace Sulu\Bundle\HttpCacheBundle\Cache;
 
 use FOS\HttpCache\SymfonyCache\CacheInvalidation;
-use FOS\HttpCache\SymfonyCache\CleanupCacheTagsListener;
 use FOS\HttpCache\SymfonyCache\CustomTtlListener;
 use FOS\HttpCache\SymfonyCache\DebugListener;
 use FOS\HttpCache\SymfonyCache\EventDispatchingHttpCache;
--- a/src/Sulu/Bundle/HttpCacheBundle/Cache/SuluHttpCache.php
+++ b/src/Sulu/Bundle/HttpCacheBundle/Cache/SuluHttpCache.php
@@ -17,6 +17,7 @@ use FOS\HttpCache\SymfonyCache\CustomTtlListener;
 use FOS\HttpCache\SymfonyCache\DebugListener;
 use FOS\HttpCache\SymfonyCache\EventDispatchingHttpCache;
 use FOS\HttpCache\SymfonyCache\PurgeListener;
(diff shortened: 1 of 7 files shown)


src/Sulu/Bundle/HttpCacheBundle/Cache/SuluHttpCache.php
```diff
@@ -55,4 +60,12 @@ public function fetch(Request $request, $catch = false)
     {
         return parent::fetch($request, $catch);
     }
+
+    protected function createStore()
+    {
+        return new Psr6Store([
+            'cache_directory' => $this->kernel->getWebsiteCacheDir(),
```
danrot: Is using the website kernel really necessary here? Due to sulu/sulu-minimal#150 the cache should be only active in the website context anyway. And apart from that, this would probably lead to strange behavior in case we start caching some stuff in the admin context as well.

alexander-schranz:
We should document the common cache dir in: https://github.com/sulu/sulu-docs/blob/develop/cookbook/override-dir-structure.rst

src/Sulu/Bundle/HttpCacheBundle/Cache/SuluHttpCache.php
```diff
@@ -34,16 +39,21 @@ class SuluHttpCache extends HttpCache implements CacheInvalidation
      * @param HttpKernelInterface $kernel
      * @param string $cacheDir
      */
-    public function __construct(HttpKernelInterface $kernel, $cacheDir = null)
+    public function __construct(SuluKernel $kernel, $cacheDir = null)
```
alexander-schranz: would keep here `HttpKernelInterface` and inject `%sulu.common_cache_dir%` so its possible to change the directory for this service only else you are not able to cahnge it only for this service.

src/Sulu/Bundle/HttpCacheBundle/Cache/SuluHttpCache.php
```diff
@@ -55,4 +65,16 @@ public function fetch(Request $request, $catch = false)
...
     }
+
+    protected function createStore()
+    {
+        if (!$this->kernel instanceof SuluKernel) {
+            throw new \RuntimeException('Unexpected kernel instance given');
+        }
+
+        return new Psr6Store([
+            'cache_directory' => $this->kernel->getCommonCacheDir(),
```
alexander-schranz: as said above would inject it instead of directly get it from common cache dir.

alexander-schranz:
We should update the docs here: https://github.com/sulu/sulu-docs/blob/develop/bundles/http_cache.rst

PR #4349, opened 2019-01-07
Fix compatibility to symfony 3.4.21, 4.1.10 and 4.2.2

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Fix form submit of unstructured data.

#### Why?

Symfony seems since the latest release https://github.com/symfony/symfony/issues/29809 only allow scalar values be submit in the TextType.

> [structure.stagedData] This value is not valid. ({"{{ value }}":"NULL"})

I not really happy with the solution but currently seems the only working one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 CHANGELOG for Sulu
 ==================
 
+* dev-develop
+    * BUGFIX      #4349 [ContentBundle]         Fix compatibility to symfony 3.4.21, 4.1.10 and 4.2.2
+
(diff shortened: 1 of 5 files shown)


alexander-schranz:
@wachterjohannes sulu article bundle will also crash currently: https://github.com/sulu/SuluArticleBundle/blob/develop/Document/Form/ArticleDocumentType.php#L37

PR #4357, opened 2019-01-10
Added dimension properties to video media

Leobaillard:
Updated MediaBundle's MediaManager to get the video dimensions when ffprobe is available.

| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | N/A
| Related issues/PRs | <not-created>
| License | MIT

#### What's in this PR?

The MediaManager in the MediaBundle allows videos to be uploaded and processed with ffmpeg and ffprobe but the dimension informations are not saved. This PR adds, when available, the width and height of the video files to the media file properties.

#### Why?

It is useful to have the video dimensions available in Twig. It could for instance be used when we want to have a cover video while keeping the video's aspect ratio.

#### Example Usage

~~~twig
{# Provided you have a media_selection field called 'video' #}
<video width="{{ content.video[0].properties.width }}" height="{{ content.video[0].properties.height }}">
    <source src={{ content.video[0].url }}" type="{{ content.video[0].mimeType }}">
</video>
~~~

#### BC Breaks/Deprecations

N/A

#### To Do

- [x] Add changes to CHANGELOG.md

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/Media/Manager/MediaManager.php
+++ b/src/Sulu/Bundle/MediaBundle/Media/Manager/MediaManager.php
@@ -307,12 +307,24 @@ class MediaManager implements MediaManagerInterface
         $properties = [];
 
         try {
-            // if the file is a video we add the duration
+            // if the file is a video we add the properties related to it
             if (fnmatch('video/*', $mimeType)) {
+                // Duration
                 $properties['duration'] = $this->ffprobe->format($uploadedFile->getPathname())->get('duration');
+
+                // Dimensions
+                try {
+                    $dimensions = $this->ffprobe->streams( $uploadedFile->getPathname() )->videos()->first()->getDimensions();
+                    $properties['width'] = $dimensions->getWidth();
+                    $properties['height'] = $dimensions->getHeight();
+                } catch (InvalidArgumentException $e) {
+                    // Exception is thrown if the video stream could not be obtained
+                } catch (RuntimeException $e) {
+                    // Exception is thrown if the dimension could not be extracted
+                }
             }
         } catch (ExecutableNotFoundException $e) {
-            // Exception is thrown if ffmpeg is not installed -> duration is not set
+            // Exception is thrown if ffmpeg is not installed -> video properties are not set
         }
 
         return $properties;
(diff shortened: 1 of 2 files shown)


alexander-schranz:
Looks good for me.

/cc @danrot

src/Sulu/Bundle/MediaBundle/Media/Manager/MediaManager.php
```diff
@@ -307,12 +307,24 @@ private function getProperties(UploadedFile $uploadedFile)
...
                 $properties['duration'] = $this->ffprobe->format($uploadedFile->getPathname())->get('duration');
+
+                // Dimensions
+                try {
+                    $dimensions = $this->ffprobe->streams($uploadedFile->getPathname())->videos()->first()->getDimensions();
+                    $properties['width'] = $dimensions->getWidth();
+                    $properties['height'] = $dimensions->getHeight();
+                } catch (InvalidArgumentException $e) {
+                    // Exception is thrown if the video stream could not be obtained
+                } catch (RuntimeException $e) {
```
danrot: Isn't there a more specific exception that could be caught? This might hide all different kind of exceptions...

Leobaillard: Unfortunately, I don't think so. Looking at the ffmpeg lib, it seems that `RuntimeException` is what the different methods used to get the video's dimensions throw:
- https://github.com/PHP-FFMpeg/PHP-FFMpeg/blob/d1a281f30ffbee3110a58b1d41644ed0782f505b/src/FFMpeg/FFProbe.php#L238
- https://github.com/PHP-FFMpeg/PHP-FFMpeg/blob/d1a281f30ffbee3110a58b1d41644ed0782f505b/src/FFMpeg/FFProbe.php#L263
- https://github.com/PHP-FFMpeg/PHP-FFMpeg/blob/d1a281f30ffbee3110a58b1d41644ed0782f505b/src/FFMpeg/FFProbe/DataMapping/Stream.php#L46

We could also catch the `LogicException` from `FFProbe::getDimensions` but it should not be triggered as we only query video streams. In any case, this would not suffice to be more precise on exception catching.

If you think this is too broad, I can try formatting the code another way. I'm open to suggestions. Thanks for your review!

danrot: Well, the only thing we could do is check the message of the exception, and that's also not nice... I guess we let it that way in this case.

PR #4365, opened 2019-01-12
Add optional title to composer.json

rogoit:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?
Here is the start of a feature to read out a title from the composer.json file and show this on the backend

PHP file changes are working but it is still not shown in admin navigation overview.
I need a review to go back to work.

### DIFF
--- a/src/Sulu/Component/Util/SuluVersionPass.php
+++ b/src/Sulu/Component/Util/SuluVersionPass.php
@@ -67,7 +67,7 @@ class SuluVersionPass implements CompilerPassInterface
      */
     private function getAppVersion($dir)
     {
-        $version = '_._._';
+        $version = null;
 
         /** @var SplFileInfo $composerFile */
         $composerFile = new SplFileInfo($dir . '/composer.json', '', '');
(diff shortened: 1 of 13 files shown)


src/Sulu/Component/Util/SuluVersionPass.php
```diff
@@ -59,15 +60,41 @@ private function getSuluVersion($dir)
...
+    /**
+     * Get title from composer file.
+     *
+     * @param string $dir
+     *
+     * @return string|null
+     */
+    private function getTitle($dir)
+    {
+        $key = 'title';
```
alexander-schranz: Title does not exist in composer this will throw the following error:

> The property title is not defined and the definition does not allow additional properties

Only description exists else it need to be in `extra` part of the composer.json 

/cc @danrot

danrot:
Or we could just use the already existing `sulu_admin.name` parameter. That something you can change in a config file as follow:

```yaml
sulu_admin:
    name: "Something"
```

You can use the `sulu_admin.name` parameter directly in the service configuration for the `AdminController`.

danrot:
@rogoit Since you have not continued to work on that, I am going to close this one.

As @alexander-schranz already mentioned, the `title` attribute does not exist in `composer.json`, so we should use our own `sulu_admin.name` configuration instead. I guess most of the work you've done here is obsolete because of that.

I hope that's ok for you. In case you are interested to work on the other approach (using our configuration) feel free to create another PR :slightly_smiling_face:

And nevertheless, thank you for your work!

PR #4368, opened 2019-01-13
Show error message if media upload fails for SingleMediaDropzone

### LINKED ISSUE #4297: Media: Upload file size exceed: No error message and nothing happens
| Q | A
| --- | ---
| Bug? | yes
| New Feature? | no
| Sulu Version | 2.0-dev 930731896827d00fdc21d884327dc84256868c94
| Browser Version | Chrome 70

#### Actual Behavior

When uploading an image that is over the upload file size (be it PHP or nginx POST size or anything else like that), the upload fails but the user isn't notified about it at all.

#### Expected Behavior

Inform the user about the issue (maybe even include the current limit if possible)

#### Steps to Reproduce

Set upload size (e.g. nginx POST size) to 1 MB and upload an image with 2 MB

comixtrip:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #4297 
| Related issues/PRs | -
| License | MIT
| Documentation PR | No

#### What's in this PR?

This PR contains React/SCSS code to check the server response for uploaded media elements. If an error occurs when using media upload in the SingleDropzone component, the Dropzone shows a translatable error message. A translation for HTTP status code 413 and a general error message for failing XHR requests is included for DE and EN.

#### To Do

- [ ] Refactor implementation so it can be used also for MultiMediaDropzone
- [x] Write e2e tests
- [ ] Currently, only HTTP status code 413 and general XHR errors are handled properly, because each status code needs a specific translation at the moment. I'll try to find a more generic way (In combination with translated error messages) soon.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 CHANGELOG for Sulu
 ==================
+* 2.0.0-alpha4
+    * BUGFIX      #4297 [MediaBundle]           Fix unhandled server errors on media upload for SingleMediaDropzone
 
 * 1.6.24 (2019-01-09)
     * BUGFIX      #4349 [ContentBundle]         Fix compatibility to symfony 3.4.21, 4.1.10 and 4.2.2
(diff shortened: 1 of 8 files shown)


niklasnatter:
Thanks for your work! I have adjusted the code to the current codebase in #6008

Issue #4371, opened 2019-01-16
User contact permissions crashes when too many webspace and roles are defined


cdellamonica:
| Q | A
| --- | ---
| Bug? | yes
| New Feature? | no
| Sulu Version | 1.6
| Browser Version | Any

#### Actual Behavior

Our project has 50+ webspaces and as much roles so our users can manage only their webspace's content.

In the user admin panel, if I want to change the roles of a user, the request to /admin/api/users generates a 70Mb Json in my dev environnement and crashes in production.

#### Expected Behavior

The page loads quickly and I can manage my users.

#### Steps to Reproduce

Create 50+ webspaces and roles, edit a user's permissions.

#### Possible Solutions

Load less information to list the roles and permissions.

danrot:
Are you saying that the form for the roles is doing fine, while only the contact form is not working correctly?
Because then the number of webspace probably wouldn't mattter, but only the number of roles. Could you double check?

How does the request fail? Is it maybe the same error as in https://github.com/sulu/sulu/issues/2477?

cdellamonica:
Hello Daniel,

I think both problems are related. Webspaces increase the issue because each webspace create 4 new permissions ( webspace, default-snippets, analytics, custom-urls ). So I have currently 931 records in the se_permissions table. The admin/api/users Json returns Roles * Permissions * Language records, which becomes quite a lot in my case !
The request fails as it times out, because it seems like the backend does a SQL request for each Role/Permission/Language 

I think the contact form does not need the detail of all the permissions to work as it's not displaying them. It just needs the roles list. That would greatly decrease the complexity.
 
I hope my comment makes sense :)

danrot:
Ah, I see, I wasn't aware that the `permission` field is rendered as well... Removing that should be a quite easy fix which already should improve the response. I won't have time to fix it today, but I'll ask others from the team, and otherwise I'll fix it next week.

cdellamonica:
awesome ! Thanks :)

danrot:
@cdellamonica Can you copy over the two lines from [this file from my PR](https://github.com/sulu/sulu/pull/4376/files#diff-87867051cbf0ddc1f92a5fa5ab4c3532) and tell if that makes your problem go away?

cdellamonica:
@danrot, This fix works, although I think you should apply the "fullRole" group to the $settings attribute too.
Thanks !

danrot:
[...]

PR #4376, opened 2019-01-21
Improve performance of User API

danrot:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | #4371 
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR removes the `permissions` field under roles for the user request.

#### Why?

Because the response can get too big, which makes the application crash (i.e. the permissions tab from the contact).

#### BC Breaks/Deprecations

Somebody might have used this field when calling the API, now he would have to do another request to get this data.

#### To Do


danrot:
@chirimoya: @trickreich already complained that this is a BC break, which is correct. We could add a query parameter to include this additional serialization group or not, but I think it is not worth the hassle, since probably nobody used this field in their applications.

PR #4383, opened 2019-01-23
Add support for newest imagine version

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Upgrade imagine library version to allow ^1.0.

#### Why?

To support webp in future release.

#### To Do

- [x] Test Animated gif

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,7 +33,7 @@
         "goodby/csv": "^1.3",
         "guzzlehttp/guzzle": "^6.2",
         "guzzlehttp/psr7": "^1.5",
-        "imagine/imagine": "~0.6.1 || ~0.7.0",
+        "imagine/imagine": "~0.6.1 || ~0.7.0 || ^1.0",
         "jms/serializer": "^1.13",
         "jms/serializer-bundle": "^2.2",
         "layershifter/tld-extract": "^1.1",


composer.json
```diff
@@ -33,7 +33,7 @@
         "goodby/csv": "^1.3",
         "guzzlehttp/guzzle": "^6.2",
         "guzzlehttp/psr7": "^1.5",
-        "imagine/imagine": "~0.6.1 || ~0.7.0",
+        "imagine/imagine": "~0.6.1 || ~0.7.0 || ^1.0",
```
danrot: Since we are doing a major release anyway, wouldn't it make sense to remove the unstable versions here? Or is there a reason you are keeping them?

alexander-schranz: @danrot think that 0.6 and 0.7 is still more used as 1.0, because there where no releases for imagine a long time. So I would keep that as long as we can't to avoid conflicts if some other dependency require old imagine version.

danrot: I don't really get that... Can't we decide on our own what we want to use? It is just that we almost have no unstable dependencies (starting with a 0) anymore, and I would feel better if get rid of the few ones we have...

PR #4385, opened 2019-01-23
WIP: Fix usage of symfony cache with shared max age 0

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Fix usage of symfony cache with shared max age 0.

#### Why?

Symfony will not cache if shared max age is set and 0 because of $response->getMaxAge() in $response->isCacheable() will return at first s-max-age.

#### Example Usage

~~~php
sulu_http_cache:
    handlers:
        public:
            max_age: 100 
            shared_max_age: 0
~~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@ CHANGELOG for Sulu
 ==================
 
 * release/1.5
+    * BUGFIX      #4385 [SuluHttpCacheBundle]   Fix usage of symfony cache with s-max-age 0
     * ENHANCEMENT #4367 [WebsiteBundle]         Remove false deprecation of WebsiteController::renderStructure
     * BUGFIX      #4376 [SecurityBundle]        Exclude role permissions in user API to improve performance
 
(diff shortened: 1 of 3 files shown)


danrot:
I would move that logic into the `HttpCache` component from Symfony, because I would like to have the `s-maxage` header set to whatever is set in the config for external browsers/proxies. Therefore the [`validate` method of the `HttpCache` component](https://github.com/symfony/symfony/blob/572864b/src/Symfony/Component/HttpKernel/HttpCache/HttpCache.php#L345) could be overridden, which will store the response in case the default implementation didn't. That means we would have to introduce another check in the `validate` method.

alexander-schranz:
@danrot did have a look at it but seems the validate method is not the only case where isCacheable is called or where logic is based on the getMaxAge function of the Response.

alexander-schranz:
Maybe obsolute after the changes in: https://github.com/symfony/symfony/pull/26532

danrot:
What does "maybe" mean? :slightly_smiling_face: Can we close this PR?

PR #4398, opened 2019-01-29
Fix auto-wiring for context based services

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | none
| Related issues/PRs | none
| License | MIT
| Documentation PR | none

#### What's in this PR?

This PR increases the priority of the context compiler pass.

#### Why?

This will avoid the collecting and auto-wiring of tagged services.


alexander-schranz:
@wachterjohannes but still a bug as we used many times field descriptors in website context and they seems not longer be usable there.

alexander-schranz:
@wachterjohannes would change the PR title to something like: _`Fix autowiring for context based services`_ . We should keep this in mind if this could be needed in 1.x versions of sulu.

PR #4419, opened 2019-02-07
Add preview for collapsed form blocks

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR adds a preview for blocks which are collapsed in the Form. This works by introducing another Registry (`BlockPreviewTransformerRegistry`), which maps the field type to a transformer, which is called with the current value of the property. All properties with the `sulu.block_preview` will be taken into account.

#### Why?

Because it makes it easier for the content manager to get an idea what is in the block without expanding it.

#### To Do

- [x] Add fallback mechanism for blocks without a single `sulu.block_preview`

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/index.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/index.js
@@ -35,11 +35,13 @@ import {
 } from './containers/Datagrid';
 import FieldBlocks, {
     blockPreviewTransformerRegistry,
+    DateTimeBlockPreviewTransformer,
     SelectBlockPreviewTransformer,
     SingleSelectBlockPreviewTransformer,
     SmartContentBlockPreviewTransformer,
     StringBlockPreviewTransformer,
     StripHtmlBlockPreviewTransformer,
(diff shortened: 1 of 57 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/index.js
```diff
@@ -129,22 +149,22 @@ function registerFieldTypes(fieldTypeOptions) {
     fieldRegistry.add('block', FieldBlocks);
     fieldRegistry.add('changelog_line', ChangelogLine);
     fieldRegistry.add('checkbox', Checkbox);
-    fieldRegistry.add('color', ColorPicker);
+    fieldRegistry.add(FIELD_TYPE_COLOR, ColorPicker);
     fieldRegistry.add('date', DatePicker, {dateFormat: true, timeFormat: false});
```
chirimoya: Add date and datetime transformer.

PR #4426, opened 2019-02-11
Use reaction instead of autorun to observe changes in stores of selection containers

nnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This PR changes our selection containers to use a `reaction` instead of a `autorun` to observe changes in the store of the container.

#### Why?

Because the listener that calls the `onChange` callback of the container should only be executed if the value of the store of the container changes. In contrast to `autorun`, `reaction` allows us to specify precisely which data should be observed.

Right now, when using `autorun`, our `onChange` handler is unnecessarily called multiple times in various places. For example, this can be observed when changing the value of a `SingleSelection` inside a `Form` (ie. the author selection on the page-settings tag).

Furthermore, the unnecessary calls to our `onChange` handler lead to a strange bug that discards the ordering of the medias in our `MultiMediaSelection` container.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/Selection.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/Selection.js
@@ -1,6 +1,6 @@
 // @flow
 import React from 'react';
-import {autorun, computed, observable, toJS} from 'mobx';
+import {computed, observable, toJS, reaction} from 'mobx';
 import type {IObservableValue} from 'mobx';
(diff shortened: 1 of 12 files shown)


danrot:
@nnatter I've encountered the exact same problems for the categories and tags in the "Excerpt & Taxonomies" tab. Do you think you can fix it in there as well, before we forget about the solution again? :see_no_evil:

nnatter:
@danrot done 🙂

PR #4427, opened 2019-02-11
Save persistent settings on server

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR stores the settings which should be stored in a persistent way (set via the `setPersistentSettings` function on the `UserStore`) on the server. This will happen using a `debounce`, so that at least a few settings are combined before a request is sent.

#### Why?

Because it is important that the system remembers some of the settings of the user.


src/Sulu/Bundle/AdminBundle/Resources/js/stores/UserStore/UserStore.js
```diff
@@ -1,13 +1,17 @@
 // @flow
 import 'core-js/library/fn/promise';
 import {action, computed, observable} from 'mobx';
+import debounce from 'debounce';
 import {Config, Requester} from '../../services';
 import initializer from '../../services/Initializer';
 import localizationStore from '../LocalizationStore';
 import type {Contact, User} from './types';
 
+const UPDATE_PERSISTENT_SETTINGS_DELAY = 5000;
```
chirimoya: 2500

PR #4433, opened 2019-02-12
Fix fresh User object comparison to the deserialized User object

JorisH:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

It fixes the fresh User object comparison to the deserialized User object to make sure that they represent the same user.

#### Why?

Once the user is logged in, the entire User object is serialized into the session. On the next request, the User object is deserialized. Then, the value of the id property is used to re-query for a fresh User object from the database. Finally, the fresh User object is compared to the deserialized User object to make sure that they represent the same user. This is not working now because the username, password and salt are not serialized.

[Understanding serialize and how a user is saved in the session](https://symfony.com/doc/2.8/security/entity_provider.html#understanding-serialize-and-how-a-user-is-saved-in-the-session)

### DIFF
--- a/src/Sulu/Bundle/SecurityBundle/Entity/BaseUser.php
+++ b/src/Sulu/Bundle/SecurityBundle/Entity/BaseUser.php
@@ -43,7 +43,6 @@ abstract class BaseUser extends ApiEntity implements UserInterface, Serializable
 
     /**
      * @var string
-     * @Expose
      */
     protected $password;
 
--- a/src/Sulu/Bundle/SecurityBundle/Entity/BaseUser.php
+++ b/src/Sulu/Bundle/SecurityBundle/Entity/BaseUser.php
@@ -19,6 +19,7 @@ use JMS\Serializer\Annotation\VirtualProperty;
 use Serializable;
 use Sulu\Bundle\CoreBundle\Entity\ApiEntity;
 use Sulu\Component\Security\Authentication\UserInterface;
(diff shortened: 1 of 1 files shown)


src/Sulu/Bundle/SecurityBundle/Entity/BaseUser.php
```diff
@@ -41,6 +43,7 @@
 
     /**
      * @var string
+     * @Expose
```
alexander-schranz: Thank you for your Pull Request can you explain why this needed? The password should not be exposed to the Api as this would be a security issue.

alexander-schranz:
Just to link it here this change was also done in the Symfony 4 upgrade: https://github.com/sulu/sulu/pull/4148/commits/d68e5c5b39a0670df202b1a92311bbfacb613f40#diff-e5daccee35ca5d60e8d2b5521cb82fa3

alexander-schranz:
@wachterjohannes can you also have a look at it!
@JorisH can you add a entry to the CHANGELOG.md ?

wachterjohannes:
@alexander-schranz @JorisH i have added the changelog in https://github.com/sulu/sulu/commit/e75ddc67a426f8bfbabb016d7f5801bb1641ba14

PR #4442, opened 2019-02-15
Add resource store properties to list store options

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT
| Documentation PR | -

#### What's in this PR?

Add resource store properties to list store route options.

#### Why?

In some cases you want to use resourceStore properties as list store options to example filter by a specific value.

#### Example Usage

```php
$this->routeBuilderFactory->createListRouteBuilder(...)
    ->addResourceStorePropertiesToListStore([
        'resourceStoreProperty' => 'listStoreProperty',
    ]);
```

This pull request also change the `addRouterAttributesToListStore` from:

```php
    ->addRouterAttributesToListStore([
        'listStoreProperty' => 'routeAttribute',
    ]);
```

to

```php
    ->addRouterAttributesToListStore([
        'routeAttribute' => 'listStoreProperty',
    ]);
```

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/views/List/List.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/views/List/List.js
@@ -17,7 +17,7 @@ import listStyles from './list.scss';
 const USER_SETTINGS_KEY = 'list';
 
 type Props = ViewProps & {
-    resourceStore: ResourceStore,
+    resourceStore?: ResourceStore,
 };
 
 @observer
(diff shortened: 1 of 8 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/views/List/List.js
```diff
@@ -102,18 +110,32 @@ class List extends React.Component<ViewProps> {
...
+        resourceStore: ?ResourceStore
     ) {
         const listStoreOptions = apiOptions ? apiOptions : {};
 
         routerAttributesToListStore = toJS(routerAttributesToListStore);
         Object.keys(routerAttributesToListStore).forEach((key) => {
-            const attributeName = routerAttributesToListStore[key];
-            const listOptionKey = isNaN(key) ? key : routerAttributesToListStore[key];
+            const listOptionKey = routerAttributesToListStore[key];
+            const attributeName = isNaN(key) ? key : routerAttributesToListStore[key];
```
danrot: I think this change is worth being mentioned in the `UPGRADE.md` file. Also, I would say that if you do this change for the `List` view, it should also be done for the `Form` view, otherwise we are inconsistent, which is really confusing.

alexander-schranz: done and rebased

PR #4445, opened 2019-02-15
Fix debug document event dispatcher and media commands

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Fix debug document event dispatcher and media commands.

#### Why?

Some services seems to be private now and can not longer be accessed from container.

#### Example Usage

Run the commands.

#### BC Breaks/Deprecations

#### To Do


UPGRADE.md
```diff
@@ -2,6 +2,12 @@
 
 ## dev-develop
 
+### Commands changed
+
+The sulu specific commands where changed instead depending on the deprecated `ContainerAwareCommand` they are replaced
```
danrot: The Sulu commands were refactored to use the `Command` class instead of the `ContainerAwareCommand` class, because some services are private and cannot be used anymore without using dependency injection. If you have overridden a command before, you have to specify the constructor arguments correctly now.

PR #4446, opened 2019-02-18
Replaced the ToolbarDropdown Popoverlist with the ArrowMenu in the webspaces view

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Replaced the ToolbarDropdown Popoverlist with the ArrowMenu  in the webspaces view.
ArrowMenu Item and Action can be disabled now.
ArrowMenu.Action automatically closes the ArrowMenu after clicking on an action.

#### Why?

The old ToolbarDropdown Popoverlist didn't match to the rest of the design.


src/Sulu/Bundle/AdminBundle/Resources/js/components/ArrowMenu/Item.js
```diff
@@ -25,17 +26,19 @@ export default class Item extends React.PureComponent<Props> {
...
         const {
             children,
             active,
             icon,
+            disabled,
         } = this.props;
 
         const itemClass = classNames(
-            itemStyles.item,
+            (disabled ? itemStyles.itemDisabled : itemStyles.item),
```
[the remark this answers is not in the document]
danrot: Also, the rendered JSX below still uses a `div`, which is bad for accessibility. Could you change it to an `a` tag?

src/Sulu/Bundle/AdminBundle/Resources/js/components/ColumnList/ToolbarDropdown.js
```diff
@@ -1,65 +1,61 @@
...
-                                    onClick={this.handlePopoverClose}
-                                    options={options}
-                                />
-                            </div>
-                        )
-                    }
-                </Popover>
+                <ArrowMenu anchorElement={this.renderButton()} onClose={this.handleMenuClose} open={this.open}>
+                    <ArrowMenu.Section>
+                        {this.generateChildren()}
```
danrot: I prefer to have this inline, so that I don't have to jump around a lot of functions when reading this code. Especially in cases like this, where this would be quite easy. Instead of this line you can write something like this:

```javascript
{options.map(({label}) => (<ArrowMenu.Action ...>{label)</ArrowMenu.Action>))}
```

src/Sulu/Bundle/AdminBundle/Resources/js/components/ColumnList/tests/Toolbar.test.js
```diff
@@ -69,10 +68,9 @@ test('Should close dropdown when item is clicked', () => {
 
     const toolbar = mount(<Toolbar toolbarItems={toolbarItems} />);
 
-    expect(toolbar.find(ToolbarDropdownListOption)).toHaveLength(0);
+    expect(toolbar.find('ToolbarDropdown').find('Action')).toHaveLength(0);
     toolbar.find(ToolbarDropdown).find('a').simulate('click');
-    expect(toolbar.find(ToolbarDropdownListOption)).toHaveLength(2);
-
```
danrot: Would also leave this empty line, to indicate that there are two different cases in this test.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Datagrid/Datagrid.js
```diff
@@ -408,7 +407,9 @@ export default class Datagrid extends React.Component<Props> {
                     open={this.adapterOptionsOpen}
                 >
                     <ArrowMenu.Section>
-                        <ArrowMenu.Action onClick={this.handleColumnOptionsOpen}>
+                        <ArrowMenu.Action
+                            onClick={this.handleColumnOptionsOpen}
```
danrot: We decided that if it fits into one line, we don't break the line.

UPGRADE.md
```diff
@@ -1,6 +1,6 @@
 # Upgrade
 
-## dev-develop
+## 2.0.0-alpha5
```
danrot: That line feels odd, because this change should actually not be there... I've got a bad feeling, because somebody messed with the repository last week :see_no_evil: Can you try to rebase?

src/Sulu/Bundle/AdminBundle/Resources/js/components/ArrowMenu/Item.js
```diff
@@ -25,31 +26,34 @@ export default class Item extends React.PureComponent<Props> {
...
         const itemClass = classNames(
             itemStyles.item,
             {
                 [itemStyles.active]: active,
+                [itemStyles.disabled]: disabled,
             }
         );
 
         return (
-            <div className={itemClass} onClick={this.handleButtonClick}>
```
danrot: I have just realized two things about this. First, it will break the webspace select (there won't be a single line for every webspace anymore) and that the `Action` component is already using a `button` instead of an `a` tag... Can you make a button out of this and style it accordingly, so that multiple webspaces in the dropdown are also shown correctly? And sorry about that, should have checked the other component first :see_no_evil:

danrot:
Otherwise it looks good :+1: Just remember to pull the changes I've pushed into this branch... Normally I don't do that, but I thought that the colorpicker snapshot is the only thing that prevents merging :see_no_evil:

src/Sulu/Bundle/AdminBundle/Resources/js/components/ArrowMenu/item.scss
```diff
@@ -2,24 +2,35 @@
...
+    text-align: start;
+    width: 100%;
 
     &:last-child {
         margin-bottom: 0;
     }
 
     &.active,
-    &:hover {
+    &:hover:not([disabled]) {
```
danrot: What do the squared brackets mean? Have never seen that, and a google search didn't yield any results... Are you sure that should not be `&:hover:not(:disabled)`?

PR #4452, opened 2019-02-19
Add FormOverlayList view

nnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This PR adds a `FormOverlayList` view that is registered with the key `sulu_admin.form_overlay_list`. The view builds upon the default `List` view but allows to specify a `formKey` option instead of a `addRoute` and `editRoute` option.

In contrast to the default `List` view, the `FormOverlayList` view does not redirect to a different route when an item is added or edited, but opens an `Overlay` that displays a `Form` that allows to add/edit the item on the same page.

#### Why?

In some cases it might be useful to display a list of items inside of a `ResourceTabs` view (an example would be displaying all child entities of a specific parent entity). In these cases, it is not intuitive to redirect the user to a different route/view when adding/editing a child entity.


src/Sulu/Bundle/AdminBundle/Admin/Routing/FormOverlayListRouteBuilder.php
```diff
@@ -0,0 +1,222 @@
...
+
+        return $this;
+    }
+
+    public function disableMoving(): FormOverlayListRouteBuilderInterface
+    {
+        $this->route->setOption('movable', false);
+
+        return $this;
+    }
```
danrot: It seems like we've resolved a merge conflict in the wrong way :see_no_evil: This setting still exists in the `ListRouteBuilder`, although it is not used, because it has been removed in favour of the `toolbarActions`... Can you remove these functions in this and the other router builder?

src/Sulu/Bundle/AdminBundle/Resources/js/components/Form/grid.scss
```diff
@@ -9,7 +9,7 @@
 .grid-section {
     margin-bottom: -$fieldMarginBottom;
 
-    & .grid-section:last-child {
+    &:last-child {
```
danrot: What's the reasoning behind this change? Always afraid of this file, because we don't have visual regression tests until now :see_no_evil:

src/Sulu/Bundle/AdminBundle/Resources/js/views/FormOverlayList/ErrorSnackbar.js
```diff
@@ -0,0 +1,45 @@
...
+        return (
+            <div className={snackbarClass}>
+                <Icon className={errorSnackbarStyles.icon} name={ICON} />
+                <div className={errorSnackbarStyles.text}>
+                    <strong>{translate('sulu_admin.error')}</strong>
+                    {onCloseClick &&
+                    <button className={errorSnackbarStyles.closeButton} onClick={onCloseClick}>
+                        {translate('sulu_admin.close')}
+                        <Icon className={errorSnackbarStyles.closeButtonIcon} name="su-times" />
+                    </button>
```
danrot: The previous 4 lines should be indented... Wondering why eslint is not complaining :thinking:

src/Sulu/Bundle/AdminBundle/Resources/js/views/FormOverlayList/ErrorSnackbar.js
```diff
@@ -0,0 +1,45 @@
...
+import errorSnackbarStyles from './errorSnackbar.scss';
+
+type Props = {|
+    onCloseClick?: () => void,
+    visible: boolean,
+|};
+
+const ICON = 'su-exclamation-triangle';
+
+export default class ErrorSnackbar extends React.Component<Props> {
```
danrot: Isn't this class almost identical to the `Snackbar` component in the toolbar? Maybe we can extract and reuse it? What's the reason for not doing that?

src/Sulu/Bundle/AdminBundle/Resources/js/views/FormOverlayList/FormOverlayList.js
```diff
@@ -0,0 +1,199 @@
...
+import ResourceFormStore from '../../containers/Form/stores/ResourceFormStore';
+import formOverlayListStyles from './formOverlayList.scss';
+import ErrorSnackbar from './ErrorSnackbar';
+
+@observer
+export default class FormOverlayList extends React.Component<ViewProps> {
+    static getDerivedRouteAttributes = List.getDerivedRouteAttributes;
+
+    listRef: ?List;
+    formRef: ?Form;
```
danrot: Aren't there some `ElementRef` types missing? :thinking:

src/Sulu/Bundle/AdminBundle/Resources/js/views/FormOverlayList/FormOverlayList.js
```diff
@@ -0,0 +1,199 @@
...
+            this.formRef.submit();
+        }
+    };
+
+    handleFormOverlayClose = () => {
+        this.destroyFormOverlay();
+    };
+
+    handleFormSubmit = () => {
+        if (this.formStore) {
```
danrot: I would rewrite that part. I would throw an error at the very beginning if the formStore is not defined, telling that the form store is missing, and that this is likely a bug. Have similar errors in other places.

Because it should not happen that the formstore is not defined, right? Could lead to confusing situations otherwise...

src/Sulu/Bundle/AdminBundle/Resources/js/views/FormOverlayList/FormOverlayList.js
```diff
@@ -0,0 +1,199 @@
...
+            },
+        } = this.props;
+
+        if (this.formStore) {
+            this.formStore.destroy();
+        }
+
+        const observableOptions = {};
+        if (this.listRef && this.listRef.locale && this.listRef.locale.get()) {
+            observableOptions['locale'] = this.listRef.locale;
```
danrot: Why don't you use `observableOptions.locale` as anywhere else?

src/Sulu/Bundle/AdminBundle/Resources/js/views/FormOverlayList/FormOverlayList.js
```diff
@@ -0,0 +1,199 @@
...
+
+        return (
+            <Fragment>
+                <List
+                    {...this.props}
+                    onItemAdd={formKey && this.handleItemAdd}
+                    onItemClick={formKey && this.handleItemClick}
+                    ref={this.setListRef}
+                />
+                {this.renderFormOverlay()}
```
danrot: I don't really like these additional `render` methods :see_no_evil: I think the code should be easy enough to be just inlined here (which I would say is the case here) or if the code gets more complex I would build a separate react component anyway.

src/Sulu/Bundle/AdminBundle/Resources/js/views/FormOverlayList/README.md
```diff
@@ -0,0 +1,3 @@
+The form-overlay-list is registered with the key `sulu_admin.form_overlay_list`. 
+It shows a List container with a toolbar. The actions that are displayed in the toolbar can be configured using the route's options.
```
danrot: We usually also list the available options in a table at the end of the README. It's hard to see them somewhere otherwise.

src/Sulu/Bundle/AdminBundle/Resources/js/views/FormOverlayList/formOverlayList.scss
```diff
@@ -0,0 +1,7 @@
+@import '../../components/Overlay/variables.scss';
+@import '../../components/ColumnList/variables.scss';
+
+.form {
+    padding: 30px;
```
danrot: In all other overlays we have used the `$viewPadding` variable for that, haven't we?

src/Sulu/Bundle/AdminBundle/Resources/js/views/FormOverlayList/tests/FormOverlayList.test.js
```diff
@@ -0,0 +1,276 @@
...
+    (resourceStore) => {
+        return {
+            id: resourceStore.id,
+        };
+    }
+));
+
+let router;
+beforeEach(() => {
+    router = {
```
danrot: This way you are coupling all of these tests with each other... I would prefer to have the router defined whatever is needed in each single test. We have similar setups like this in our PHP tests, and they are not very easy to maintain, because if you touch this variable a lot of tests will start to break :confused:

src/Sulu/Bundle/AdminBundle/Resources/js/views/FormOverlayList/tests/FormOverlayList.test.js
```diff
@@ -0,0 +1,276 @@
...
+
+test('View should render with closed overlay', () => {
+    const formOverlayList = mount(<FormOverlayList router={router} />);
+
+    expect(formOverlayList.render()).toMatchSnapshot();
+});
+
+test('View should render with opened overlay', () => {
+    const formOverlayList = mount(<FormOverlayList router={router} />);
+    formOverlayList.instance().createFormOverlay(undefined, 'test-list-key');
```
danrot: I think you should open the overlay using some click event. This way it is not said that it really works...

src/Sulu/Bundle/AdminBundle/Resources/js/views/FormOverlayList/tests/FormOverlayList.test.js
```diff
@@ -0,0 +1,276 @@
...
+    const list = formOverlayList.find(List);
+
+    expect(list.props()).toEqual(expect.objectContaining(formOverlayList.props()));
+    expect(list.props().onItemAdd).toBeDefined();
+    expect(list.props().onItemClick).toBeDefined();
+});
+
+test('Should construct ResourceStore and ResourceFormStore with correct parameters on item-add callback', () => {
+    const formOverlayList = mount(<FormOverlayList router={router} />);
+    formOverlayList.find(List).props().onItemAdd();
```
danrot: Why not using some click as well? Can you check the other tests as well? This way the `onItemAdd` prop might not be correctly wired, and nothing works at all :confused:

src/Sulu/Bundle/AdminBundle/Resources/js/views/List/List.js
```diff
@@ -168,17 +173,26 @@ class List extends React.Component<ViewProps> {
...
-        } = router;
+    handleItemAdd = (parentId: string | number) => {
+        const {onItemAdd, router} = this.props;
+
+        if (onItemAdd) {
+            onItemAdd(parentId);
+            return;
+        }
+
+        router.navigate(router.route.options.addRoute, {locale: this.locale.get(), parentId: parentId});
```
danrot: I would still prefer to destruct these values, because it is easier to add more stuff later on then. And `parentId: parentId` is redundant and can be replaced by just `parentId`.

src/Sulu/Bundle/AdminBundle/Resources/js/views/List/List.js
```diff
@@ -168,17 +173,26 @@ class List extends React.Component<ViewProps> {
...
+    handleItemClick = (itemId: string | number) => {
+        const {onItemClick, router} = this.props;
+
+        if (onItemClick) {
+            onItemClick(itemId);
+            return;
+        }
 
-        router.navigate(addRoute, {locale: this.locale.get(), parentId: rowId});
+        router.navigate(router.route.options.editRoute, {id: itemId, locale: this.locale.get()});
```
danrot: Same as above

src/Sulu/Bundle/AdminBundle/Resources/js/views/List/toolbarActions/AddToolbarAction.js
```diff
@@ -8,7 +8,7 @@ export default class AddToolbarAction extends AbstractToolbarAction {
         return {
             icon: 'su-plus-circle',
             label: translate('sulu_admin.add'),
-            onClick: action(this.list.addItem),
+            onClick: action(this.list.handleItemAdd),
```
danrot: That's actually the reason it is not called `handleItemAdd`. It is an external API as well, in which case the `handle` part makes no sense...

src/Sulu/Bundle/AdminBundle/Resources/js/components/Form/grid.scss
nnatter: if the `section` is the last child of something, we do not need to "eat" the `margin-top` of the next `item` 🙂

src/Sulu/Bundle/AdminBundle/Resources/js/views/FormOverlayList/ErrorSnackbar.js
nnatter: i actually copied that class and removed everything but the error case 😉 .
i am open to extracting et into a separate component - where would we store that one?

src/Sulu/Bundle/AdminBundle/Resources/js/views/FormOverlayList/README.md
nnatter: i noticed that the available options in the other `README` files are kind of outdated. for example, the `README` of the `List` view contains only two options. therefore assumed that we dont update that anymore 🙈 . do we?

src/Sulu/Bundle/AdminBundle/Resources/js/views/List/toolbarActions/AddToolbarAction.js
nnatter: kind of feeling stupid now, but what is the reason? 🙈

src/Sulu/Bundle/AdminBundle/Resources/js/views/FormOverlayList/tests/FormOverlayList.test.js
nnatter: the thing here is that there is no element mounted that we could click. the `FormOverlayList` view does not render any clickable element itself but only sets the callbacks to the embedded `List` view.

This embedded `List` view, which renders the edit-button and sets the add-button to the toolbar, is mocked (see line 17) because we do not want to mock all dependencies of the `List` view in this test-suite in my opinion.

src/Sulu/Bundle/AdminBundle/Resources/js/views/FormOverlayList/tests/FormOverlayList.test.js
nnatter: As explained above, there is no clickable element mounted because the `List` view is mocked.
But in general, applying a click event to the `List` view would basically test the internals of the `List` component, wouldnt it?

In my opinion, testing if given callbacks are fired when respective elements are clicked should happen in the test-suite of the respective components. Otherwise we will implicitly test these implementations in many places and many tests will fail when the wiring inside an unrelated component breaks.

src/Sulu/Bundle/AdminBundle/Resources/js/views/FormOverlayList/README.md
danrot: Yeah, just because we might have forgotten that in other places, I would say that's not an excuse to stop doing it at all.

src/Sulu/Bundle/AdminBundle/Resources/js/views/List/toolbarActions/AddToolbarAction.js
danrot: The `addItem` method of the `List` view is part of its public API, meaning that anybody who gets access to its ref can call this method. And if you think about other objects/classes, you also wouldn't prefix every method with `handle`, or would you?

alexander-schranz:
@danrot with the consultation from @nnatter I did rebase the branch and add the new added `resourceStorePropertiesToListStore` List Property from #4442 also to this FormOverlayList.

src/Sulu/Bundle/AdminBundle/Resources/js/views/List/toolbarActions/AddToolbarAction.js
nnatter: alright, that makes sense. reverted this change :)

src/Sulu/Bundle/AdminBundle/Resources/js/views/FormOverlayList/README.md
nnatter: added the documentation for the `list` view and the `form-overlay-list` view

src/Sulu/Bundle/AdminBundle/Resources/js/views/FormOverlayList/FormOverlayList.js
```diff
@@ -0,0 +1,188 @@
...
+    handleFormSubmit = () => {
+        if (!this.formStore) {
+            throw new Error('The FormStore has not been initialized! This should not happen and is likely a bug.');
+        }
+
+        this.formStore.save()
+            .then(() => {
+                this.destroyFormOverlay();
+                if (this.listRef) {
+                    this.listRef.listStore.sendRequest();
```
danrot: Shouldn't we at least introduce a a `listRef.reload` method? Then it would be more similar to `formRef.submit` and less dependent on the `ListStore`.

danrot: Of course I would like to not have to use a ref here, but I don't know a better solution at the moment....

src/Sulu/Bundle/AdminBundle/Resources/js/views/FormOverlayList/tests/FormOverlayList.test.js
```diff
@@ -0,0 +1,384 @@
...
+    const router: Router = ({
+        route: {
+            options: {
+                addOverlayTitle: 'app.add_overlay_title',
+            },
+        },
+    }: any);
+
+    const formOverlayList = mount(<FormOverlayList route={route} router={router} />);
+    formOverlayList.instance().createFormOverlay(undefined);
```
danrot: Why can't we at least do something like this:

```javscript
formOverlayList.find('List').prop('onItemAdd')();
```

Then we would at least be sure that the function is correctly attached.

src/Sulu/Bundle/AdminBundle/Resources/js/views/FormOverlayList/tests/FormOverlayList.test.js
```diff
@@ -0,0 +1,384 @@
...
+test('Should submit Form container when Overlay is confirmed', () => {
+    const route: Route = ({}: any);
+    const router: Router = ({
+        route: {
+            options: {},
+        },
+    }: any);
+
+    const formOverlayList = mount(<FormOverlayList route={route} router={router} />);
+    formOverlayList.instance().createFormOverlay(undefined);
```
danrot: Same as above

src/Sulu/Bundle/AdminBundle/Resources/js/views/FormOverlayList/tests/FormOverlayList.test.js
```diff
@@ -0,0 +1,384 @@
...
+test('Should destroy ResourceFormStore without saving when Overlay is closed', () => {
+    const route: Route = ({}: any);
+    const router: Router = ({
+        route: {
+            options: {},
+        },
+    }: any);
+
+    const formOverlayList = mount(<FormOverlayList route={route} router={router} />);
+    formOverlayList.instance().createFormOverlay(undefined);
```
danrot: Same as above

src/Sulu/Bundle/AdminBundle/Resources/js/views/FormOverlayList/tests/FormOverlayList.test.js
```diff
@@ -0,0 +1,384 @@
...
+test('Should save ResoureFormStore, close overlay and reload List view on submit of Form', () => {
+    const route: Route = ({}: any);
+    const router: Router = ({
+        route: {
+            options: {},
+        },
+    }: any);
+
+    const formOverlayList = mount(<FormOverlayList route={route} router={router} />);
+    formOverlayList.instance().createFormOverlay(undefined);
```
danrot: Same as above

src/Sulu/Bundle/AdminBundle/Resources/js/views/FormOverlayList/tests/FormOverlayList.test.js
```diff
@@ -0,0 +1,384 @@
...
+test('Should display Snackbar if an error happens during saving of ResourceFormStore', () => {
+    const route: Route = ({}: any);
+    const router: Router = ({
+        route: {
+            options: {},
+        },
+    }: any);
+
+    const formOverlayList = mount(<FormOverlayList route={route} router={router} />);
+    formOverlayList.instance().createFormOverlay(undefined);
```
danrot: Same as above

src/Sulu/Bundle/AdminBundle/Resources/js/views/FormOverlayList/tests/FormOverlayList.test.js
```diff
@@ -0,0 +1,384 @@
...
+test('Should hide Snackbar when closeClick callback of Snackbar is fired', () => {
+    const route: Route = ({}: any);
+    const router: Router = ({
+        route: {
+            options: {},
+        },
+    }: any);
+
+    const formOverlayList = mount(<FormOverlayList route={route} router={router} />);
+    formOverlayList.instance().createFormOverlay(undefined);
```
danrot: Same as above

src/Sulu/Bundle/AdminBundle/Resources/js/views/FormOverlayList/tests/FormOverlayList.test.js
```diff
@@ -0,0 +1,384 @@
...
+test('Should destroy ResourceFormStore when component is unmounted', () => {
+    const route: Route = ({}: any);
+    const router: Router = ({
+        route: {
+            options: {},
+        },
+    }: any);
+
+    const formOverlayList = mount(<FormOverlayList route={route} router={router} />);
+    formOverlayList.instance().createFormOverlay(undefined);
```
danrot: Same as above

src/Sulu/Bundle/AdminBundle/Resources/js/views/FormOverlayList/FormOverlayList.js
nnatter: refactored the reload method of the `ListStore` and added a reload method to the `List` view

src/Sulu/Bundle/AdminBundle/Resources/js/views/FormOverlayList/FormOverlayList.js
```diff
@@ -0,0 +1,188 @@
...
+                },
+            },
+        } = this.props;
+
+        if (this.formStore) {
+            this.formStore.destroy();
+        }
+
+        const observableOptions = {};
+        if (this.listRef && this.listRef.locale && this.listRef.locale.get()) {
```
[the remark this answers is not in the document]
nnatter: done 🙂

src/Sulu/Bundle/AdminBundle/Resources/js/containers/ListOverlay/ListOverlay.js
```diff
@@ -59,7 +59,10 @@ export default class ListOverlay extends React.Component<Props> {
         const {clearSelectionOnClose, open, reloadOnOpen} = this.props;
 
         if (!this.listStore.loading && reloadOnOpen && prevProps.open === false && open === true) {
-            this.listStore.reload();
+            runInAction(() => {
```
danrot: Why don't we add an `@action` decorator to the enclosing function?

src/Sulu/Bundle/MediaBundle/Resources/js/containers/MediaCollection/MediaCollection.js
```diff
@@ -45,7 +45,12 @@ export default class MediaCollection extends React.Component<Props> {
     handleUpload = (media: Array<Object>) => {
         const {mediaListStore} = this.props;
 
-        mediaListStore.reload();
+        // reset list to first page and reload in single action to avoid multiple requests
+        runInAction(() => {
```
danrot: Same as above

PR #4457, opened 2019-02-26
Allow sub resources in ResourceRequester

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR enables the ResourceRequester to understand sub resources.

#### Why?

Because we need that in certain situations. We could work around that with many rest endpoints, but it is nice to specify more beautiful APIs.

#### BC Breaks/Deprecations

The interface of the `ResourceReuqester` changed. Also the configuration for the routes in the backend changed. 

#### To Do

- [x] Remove `AddResourceMetadataProviderPass`
- [x] Remove `ResourceMetadataPool`
- [x] Rename `endpoint` in config and `ResourceEndpointRegistry` to `routes`
- [x] Rename `ResourceEndpointRegistry` to `ResourceRouteRegistry`
- [x] Rename `resourceMetadataEndpoints` to `resources` in config request
- [ ] Remove `buildQueryString`?
- [x] Check why requests are sent twice when opening a form
- [x] Check viewRegistry failure after second login in previous state


src/Sulu/Bundle/PageBundle/Controller/AbstractExtensionController.php
```diff
@@ -56,13 +59,13 @@ public function cgetAction()
         throw new NotFoundHttpException();
     }
 
-    public function getAction(Request $request, $uuid)
```
chirimoya: ???

webpack.config.js
```diff
@@ -59,6 +59,14 @@ module.exports = (env, argv) => { // eslint-disable-line no-undef
             }),
             new CleanObsoleteChunksPlugin(),
         ],
+        resolve: {
+            alias: {
+                'fos-jsrouting': path.resolve(
```
chirimoya: 🙈

src/Sulu/Bundle/PageBundle/Controller/AbstractExtensionController.php
danrot: The name of this variable is also the name of the route parameter. If I would leave it to `uuid`, then I would have to handle both version (`uuid` and `id`) in the javascript code.

webpack.config.js
danrot: Works perfectly :stuck_out_tongue: Although I would still prefer to have a NPM package provided by them...

danrot: There seems to be an issue for that already. Keeping an eye on it: https://github.com/FriendsOfSymfony/FOSJsRoutingBundle/issues/342

PR #4458, opened 2019-02-26
Use php-cs-fixer instead of StyleCI

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

Use the standard [`php-cs-fixer` tool](https://cs.symfony.com/) instead of StyleCI.

#### Why?

It allows to lint the files also locally.


danrot:
@sulu/core-team Can you check if these changes are ok for you? They were all introduced by the new php-cs-fixer configuration, which can be configured in more detail, instead of simply disabling an entire set of rules.

The main changes are:

- closure functions now always have no space between the `function` keyword and the first parantheses
- phpdocs are forced not to be aligned anymore
- the `final` keyword is removed from `private` functions, I guess because private functions can't be overridden anyway, and this information is redundant

src/Sulu/Component/Snippet/Export/SnippetExport.php
```diff
@@ -90,7 +90,7 @@ public function getExportData()
         $progress = new ProgressBar($this->output, count($snippets));
         $progress->start();
 
-        /**
+        /*
```
alexander-schranz: remove comment

.php_cs
```diff
@@ -1,21 +0,0 @@
-<?php
-
-$header = <<<EOF
-This file is part of the Sulu.
-
-(c) MASSIVE ART WebServices GmbH
-
-This source file is subject to the MIT license that is bundled
-with this source code in the file LICENSE.
-EOF;
```
danrot: Add this header to the new configuration.

PR #4459, opened 2019-02-27
Use colspan instead of size in grid

trickreich:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | --
| Related issues/PRs | --
| License | MIT
| Documentation PR | --

#### What's in this PR?

Use `colspan` instead of `size` in grid in the frontend part.

#### Why?

No bc break.

### DIFF
--- a/src/Sulu/Component/Content/Metadata/Parser/PropertiesXmlParser.php
+++ b/src/Sulu/Component/Content/Metadata/Parser/PropertiesXmlParser.php
@@ -153,7 +153,7 @@ class PropertiesXmlParser
                 'default-type',
                 'minOccurs',
                 'maxOccurs',
-                'colSpan',
+                'colspan',
                 'cssClass',
                 'disabledCondition',
                 'visibleCondition',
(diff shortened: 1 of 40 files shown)


src/Sulu/Component/Content/Metadata/Parser/PropertiesXmlParser.php
```diff
@@ -154,7 +153,7 @@ private function loadBlock(\DOMXPath $xpath, \DOMNode $node, &$tags)
                 'default-type',
                 'minOccurs',
                 'maxOccurs',
-                'colspan',
+                'colSpan',
```
danrot: Shouldn't that be left to `colspan`? Isn't reading the value from the XML, where we've decided to keep it this way?

trickreich: I think you are right here. Should be ok now.

danrot:
Looks like we have to adapt the default templates provided by sulu-minimal... This is how they currently look like because of the still existing highlight section:

screenshot from 2019-03-01 09-37-20

Can you do that as well?

And the colspan is not working anymore for blocks.

danrot:
Forget the comment about the colspan on blocks... You've pushed to the wrong repository, that's why I got the old code...

PR #4460, opened 2019-02-28
Fix bug in Requester for nested objects

trickreich:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes --
| Related issues/PRs | --
| License | MIT
| Documentation PR | --

#### What's in this PR?

Fix bug in Requester for nested objects.

#### Why?

Because `undefined` is not converted to `null` currently, therefore the property will not be included in the payload.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/services/Requester/Requester.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/services/Requester/Requester.js
@@ -76,7 +76,7 @@ function transformRequestObject(data: Object): Object {
     }, {});
 }
 
-function transformRequestArray(data: Array<Object>) {
+function transformRequestArray(data) {
     return data.map((value) => {
         if (Array.isArray(value) || isObservableArray(value)) {
             return transformRequestArray(value);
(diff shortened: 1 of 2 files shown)


danrot:
The `media_selection` field type does not work correctly anymore... The list is always empty after saving.

trickreich:
@danrot Should be working again.

src/Sulu/Bundle/AdminBundle/Resources/js/services/Requester/Requester.js
```diff
@@ -12,25 +13,84 @@ const defaultOptions = {
...
+        if (value instanceof Object) {
+            return transformRequestObject(value);
+        }
+
+        return value;
+    });
+}
+
+function transformRequestData(originalData: Object | Array<Object>): Object | Array<Object> {
+    const data = toJS(originalData);
```
danrot: Can you add a test with some observable data for that? That's something that might be easily overseen in the future...

src/Sulu/Bundle/AdminBundle/Resources/js/services/Requester/tests/Requester.test.js
```diff
@@ -1,5 +1,6 @@
 // @flow
 import Requester from '../Requester';
+import {observable} from "mobx";
```
danrot: eslint will complain about the quotes here

src/Sulu/Bundle/AdminBundle/Resources/js/services/Requester/Requester.js
trickreich: is added

PR #4468, opened 2019-03-04
Implement export toolbar-action for contact list

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

What's in this PR?

This PR implements an export toolbar action which is used in the contact list.


src/Sulu/Bundle/AdminBundle/Resources/js/views/List/toolbarActions/ExportToolbarAction.js
```diff
@@ -0,0 +1,125 @@
...
+                key="sulu_admin.export"
+                onClose={this.handleClose}
+                onConfirm={this.handleConfirm}
+                open={this.showOverlay}
+                size="small"
+                title={translate('sulu_admin.export_overlay_title')}
+            >
+                <div className={exportToolbarActionStyles.overlay}>
+                    <Form>
+                        <Form.Section size={6}>
```
nnatter: i think `size` is called `colSpan` now 🙂

src/Sulu/Bundle/AdminBundle/Resources/js/views/List/tests/List.test.js
```diff
@@ -910,3 +916,118 @@ test('Should move items after move overlay was confirmed', () => {
...
+    const list = mount(<List router={router} />);
+    const listStore = list.instance().listStore;
+    listStore.selectionIds.push(1, 4, 6);
+
+    list.update();
+
+    getExportItem().onClick();
+    list.update();
+
+    expect(list.find('Overlay').find({confirmText: 'Export'})).toMatchSnapshot();
```
danrot: Isn't there a `render` call missing? Could be after the `find` method.

src/Sulu/Bundle/AdminBundle/Resources/js/views/List/toolbarActions/ExportToolbarAction.js
```diff
@@ -0,0 +1,125 @@
+// @flow
+import {action, observable} from 'mobx';
+import React from 'react';
+import {translate} from '../../../utils/Translator';
+import Overlay from '../../../components/Overlay';
+import Form from '../../../components/Form';
+import SingleSelect from '../../../components/SingleSelect';
+import resourceEndpointRegistry from '../../../services/ResourceRequester/registries/ResourceRouteRegistry';
```
danrot: That's called `resourceRouteRegistry` now, please also rename the variable accordingly.

src/Sulu/Bundle/AdminBundle/Resources/js/views/List/toolbarActions/ExportToolbarAction.js
```diff
@@ -0,0 +1,125 @@
...
+                    <Form>
+                        <Form.Section colSpan={6}>
+                            <Form.Field
+                                description={translate('sulu_admin.delimiter_description')}
+                                label={translate('sulu_admin.delimiter')}
+                            >
+                                <SingleSelect onChange={this.handleDelimiterChanged} value={this.delimiter}>
+                                    <SingleSelect.Option value=";">;</SingleSelect.Option>
+                                    <SingleSelect.Option value=",">,</SingleSelect.Option>
+                                    <SingleSelect.Option value={'\\t'}>\t</SingleSelect.Option>
```
danrot: I don't think that `\t` is understandable for the people using this dialog. What does e.g. Excel write in this dialog? Should we maybe use "Tab" as a written word? Would also mean it has to be translated.

src/Sulu/Bundle/AdminBundle/Resources/js/views/List/toolbarActions/ExportToolbarAction.js
```diff
@@ -0,0 +1,125 @@
...
+                                </SingleSelect>
+                            </Form.Field>
+                            <Form.Field
+                                description={translate('sulu_admin.enclosure_description')}
+                                label={translate('sulu_admin.enclosure')}
+                            >
+                                <SingleSelect onChange={this.handleEnclosureChanged} value={this.enclosure}>
+                                    <SingleSelect.Option value={'"'}>&quot;</SingleSelect.Option>
+                                    <SingleSelect.Option value=" ">&nbsp;</SingleSelect.Option>
```
danrot: That behaves weird, doesn't it? It doesn't look like this option can be selected... Maybe we should write "nothing" or something similar?

src/Sulu/Bundle/AdminBundle/Resources/js/views/List/toolbarActions/ExportToolbarAction.js
```diff
@@ -0,0 +1,125 @@
...
+                                <SingleSelect onChange={this.handleNewLineChanged} value={this.newLine}>
+                                    <SingleSelect.Option value={'\\n'}>\n</SingleSelect.Option>
+                                    <SingleSelect.Option value={'\\r\\n'}>\r\n</SingleSelect.Option>
+                                    <SingleSelect.Option value={'\\r'}>\r</SingleSelect.Option>
+                                </SingleSelect>
+                            </Form.Field>
+                        </Form.Section>
+                    </Form>
+                </div>
+
```
danrot: No new line

src/Sulu/Bundle/AdminBundle/Resources/js/views/List/toolbarActions/ExportToolbarAction.js
```diff
@@ -0,0 +1,125 @@
...
+    @action handleConfirm = () => {
+        window.location.assign(resourceEndpointRegistry.getListUrl(this.listStore.resourceKey) + '.csv' +
+            buildQueryString(
+                {
+                    flat: true,
+                    delimiter: this.delimiter,
+                    escape: this.escape,
+                    enclosure: this.enclosure,
+                    newLine: this.newLine,
+                }
```
danrot: All of this probably can be passed to the second argument of `getListUrl` now, since it has been refactored.

src/Sulu/Bundle/AdminBundle/Resources/translations/admin.en.json
```diff
@@ -105,5 +105,17 @@
...
+    "sulu_admin.delimiter": "Delimiter",
+    "sulu_admin.enclosure": "Enclosure",
+    "sulu_admin.escape": "Escape",
+    "sulu_admin.new_line": "New line",
+    "sulu_admin.delimiter_description": "Character for column separation.",
+    "sulu_admin.enclosure_description": "Enclose multiple word entries.",
+    "sulu_admin.escape_description": "Add-on for special characters.",
+    "sulu_admin.new_line_description": "Character for line brake."
+
+
```
danrot: No new lines here

src/Sulu/Bundle/ContactBundle/Admin/ContactAdmin.php
```diff
@@ -97,6 +97,7 @@ public function getRoutes(): array
         $listToolbarActions = [
             'sulu_admin.add',
             'sulu_admin.delete',
+            'sulu_admin.export',
```
danrot: I think this button makes sense in most lists... Can you add it everywhere, where it would be working? What comes to mind is accounts, tags, categories, roles, snippets, ...

src/Sulu/Bundle/AdminBundle/Resources/js/views/List/toolbarActions/ExportToolbarAction.js
Prokyonn: Should we then also rename the new line characters? If yes, how should we call them, carriage return and line feed?

src/Sulu/Bundle/AdminBundle/Resources/js/views/List/toolbarActions/ExportToolbarAction.js
```diff
@@ -49,7 +50,9 @@ export default class ExportToolbarAction extends AbstractToolbarAction {
                             >
                                 <SingleSelect onChange={this.handleEnclosureChanged} value={this.enclosure}>
                                     <SingleSelect.Option value={'"'}>&quot;</SingleSelect.Option>
-                                    <SingleSelect.Option value=" ">&nbsp;</SingleSelect.Option>
+                                    <SingleSelect.Option value="space">
+                                        {translate('sulu_admin.enclosure_space')}
```
danrot: Is that really a space? I thought if you choose this option it doesn't add any character, not even a space?

Prokyonn: Yes you are right. I misinterpreted it because the attribute was `value=" "` so it wraps the words inside two spaces. If we don't want any character at all we have to use `value=""`.

src/Sulu/Bundle/AdminBundle/Resources/js/views/List/toolbarActions/ExportToolbarAction.js
```diff
@@ -0,0 +1,126 @@
...
+    @action handleDelimiterChanged = (value: string) => {
+        this.delimiter = value;
+    };
+    @action handleEnclosureChanged = (value: string) => {
+        this.enclosure = value;
+    };
+    @action handleEscapeChanged = (value: string) => {
+        this.escape = value;
+    };
+    @action handleNewLineChanged = (value: string) => {
```
danrot: All of these function should have a line "padding" between them :slightly_smiling_face:

PR #4469, opened 2019-03-05
Sulu core configuration: Remove required attribute from paths

trickreich:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes --
| Related issues/PRs | --
| License | MIT
| Documentation PR | --

#### What's in this PR?

Remove `isRequired` from `sulu_core.content.structure.paths`.

#### Why?

Because it causes problems when trying to define for example only the `default_type`:

```
sulu_core:
    content:
        structure:
            default_type:
                article: "default"
```


trickreich:
@alexander-schranz @wachterjohannes As discussed I would remove this.

PR #4470, opened 2019-03-05
Fix sorting in SmartContent for custom simple entities

HadGS:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Add retype sortBy parameter to array, because from public part it came only as scalar value.
Add prefix for entity alias when adding sortBy columns to queryBuilder.

#### Why?

There was no way to sort custom simple entities.

#### ToDo

- [x] - Fix MediaBundle tests, which is broken by this.

### DIFF
--- a/src/Sulu/Component/SmartContent/Orm/DataProviderRepositoryTrait.php
+++ b/src/Sulu/Component/SmartContent/Orm/DataProviderRepositoryTrait.php
@@ -81,7 +81,11 @@ trait DataProviderRepositoryTrait
             $this->appendSortBy($sortBy, $sortMethod, $queryBuilder, 'c', $locale);
 
             foreach ($sortBy as $sortColumn) {
-                $queryBuilder->addSelect('c.'.$sortColumn);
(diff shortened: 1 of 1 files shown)


wachterjohannes:
@HadGS will you try to fix the tests? in general the code looks good for me - at least the logic behind it

danrot:
Fixed in a different way in #4646

HadGS:
@danrot merged is half of this bugfixes. At the end of changes there is adding of table alias to query. Because currently sorting entites works only by subentites ('subentityAlias.sortColumn') But when you try to sort by column of resulting entity, it does not work. Without entity alias, doctrine compliant, there must be an alias ('alias.column') but when you add alias, it fails because in one method entity alias is 'c', in other is an 'eClass'.

danrot:
@HadGS Sorry, I can't really follow you. Can you give a simple example of what is currently failing?

PR #4471, opened 2019-03-06
Fix CKEditor preview for content that is not wrapped in an HTML tag

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | needs: https://github.com/massiveart/husky/pull/782
| License | MIT
| Documentation PR | -

#### What's in this PR?

Fix CKEditor preview for content that is not wrapped in an HTML tag.
This includes new built version of husky with the fix done by @YetiCGN in https://github.com/massiveart/husky/pull/782

#### Why?

If its not wrapped in html the preview fails.

#### Example Usage

Save block with texteditor in source mode?


YetiCGN:
Verified that this fix makes `text_editor` fields without any HTML code work again. Also tested with tags within the text but not wrapped in an outer tag. Preview will work again and upon clicking into the textarea, CKEditor will open and process the text. The result is then wrapped in a `<p>` tag upon save.

Good to go!

YetiCGN:
> Example usage
> Save block with texteditor in source mode?

Another example would be switching a property type from `text_area` to `text_editor` with existing content, which was the origin of this issue for us. But testing is easier if you edit the content in source mode.

wachterjohannes:
@YetiCGN thanks for your fix :) i am also happy with it

danrot:
@wachterjohannes The husky PR should also be merged, shouldn't it? Otherwise this fix is lost with the next build...

PR #4473, opened 2019-03-07
Fix smart-content to allow null data

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | none
| Related issues/PRs | none
| License | MIT
| Documentation PR | none

#### What's in this PR?

This PR fix the preview for null values of tags/categories.

#### Why?

When you selct no tag/category the value is now null.

### DIFF
--- a/src/Sulu/Component/SmartContent/ContentType.php
+++ b/src/Sulu/Component/SmartContent/ContentType.php
@@ -248,7 +248,7 @@ class ContentType extends ComplexContentType implements ContentTypeExportInterfa
         if (!array_key_exists('tags', $filters) || null === $filters['tags']) {
             $filters['tags'] = [];
         }
-        if (!array_key_exists('categories', $filters)|| null === $filters['categories']) {
+        if (!array_key_exists('categories', $filters) || null === $filters['categories']) {
             $filters['categories'] = [];
         }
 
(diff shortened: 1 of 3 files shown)


danrot:
How can I test this bugfix?

src/Sulu/Component/SmartContent/ContentType.php
```diff
@@ -245,10 +245,10 @@ public function getContentData(PropertyInterface $property)
         $filters['excluded'] = [$property->getStructure()->getUuid()];
 
         // default value of tags/category is an empty array
-        if (!array_key_exists('tags', $filters)) {
+        if (!array_key_exists('tags', $filters) || null === $filters['tags']) {
```
alexander-schranz: @wachterjohannes why not use `!isset($filters['tags'])`  php.net/manual/en/function.isset.php

danrot:
@alexander-schranz Sorry, didn't see that comment before I merged :confused: But it does not really make a difference, does it?

alexander-schranz:
@danrot github is a little bit unsync today. But yeah in case of the logic its the same as isset.

PR #4476, opened 2019-03-07
Add resourceStorePropertiesToFormStore option to FormOverlayList

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

This PR is build on top of @nnatter #4452 and my #4477 overlaySize feature and extend the FormOverlayList to a option to provide resourceStoreProperties to the formStore which will be used to load the resource.

#### Why?

Sometimes you need to forward listStore properties also to the formStore. In my case the list is filtered by a specific dimensionId and only the dimension which is currently shown should be edit in the FormOverlayList so this property need to be forwared from the listStore to the formStore. 

EDIT: as mentioned by danrot we will forward the property directly from the resourceStore so you can have different values between listStore and formStore. 

#### Example Usage

~~~php
            ->resourceStorePropertiesToFormStore([
                'resourceStoreProperty' => 'formStoreProperty',
            ])
~~~

#### To Do

 - ~~[ ] Create a documentation PR~~
 - ~~[ ] Add breaking changes to UPGRADE.md~~
 - [x] Tests

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Admin/Routing/FormOverlayListRouteBuilder.php
+++ b/src/Sulu/Bundle/AdminBundle/Admin/Routing/FormOverlayListRouteBuilder.php
@@ -167,11 +167,11 @@ class FormOverlayListRouteBuilder implements FormOverlayListRouteBuilderInterfac
         return $this;
     }
 
-    public function addListStorePropertiesToFormStore(array $listStorePropertiesToFormStore): FormOverlayListRouteBuilderInterface
+    public function addResourceStorePropertiesToFormStore(array $resourceStorePropertiesToFormStore): FormOverlayListRouteBuilderInterface
     {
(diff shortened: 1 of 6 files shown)


danrot:
I've got a question regarding your concrete example. Where does the dimension come from? From what I see the options passed to the `ListStore` are all also coming from some options passed via the routes. So why don't you introduce that certain function (e.g. `addResourceStorePropertiesToFormStore`) instead of going the circumstance over the `ListStore`?

Feels a little bit error-prone and complicated to me. If the `ListStore` options change for some reason it will also affect the Form, which might still need the option just being removed. Also it seems to be more complicated to configure, because the arguments of two different methods of the builder have to match in some way.

PR #4483, opened 2019-03-12
Save permissions to live node

Quehnie:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Changed the SecuritySubscriber.php to also save the permissions to the live node.

#### Why?

In our project we use the sulu backend security to control the view permission of users in the front-end

#### Example Usage

Check the permissions in a CustomWebsiteController and redirect the user to a 401 Page or show the content.

#### To Do

### DIFF
--- a/src/Sulu/Component/Content/Document/Subscriber/SecuritySubscriber.php
+++ b/src/Sulu/Component/Content/Document/Subscriber/SecuritySubscriber.php
@@ -15,6 +15,7 @@ use PHPCR\PropertyInterface;
 use Sulu\Component\Content\Document\Behavior\SecurityBehavior;
 use Sulu\Component\DocumentManager\Event\HydrateEvent;
 use Sulu\Component\DocumentManager\Event\PersistEvent;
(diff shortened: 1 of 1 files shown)


wachterjohannes:
@danrot what do you think? could be usefull also for other projects

danrot:
Well, the initial idea was to have separate permissions for the website, because usually you would want to use a different system for that (the dropdown under the title in a role form). In the current permission tab you can only select roles from the Sulu system, so if we just copy them over, the use case with the other system might get a bit problematic... Or at least the entire thing gets more complicated. That was actually also the reason we didn't copy them over yet.

Maybe for a better understanding you should also be aware that if we are solving this problem that way, that every user for the website could also login in Sulu (since the user has a role with a Sulu system), which is not desirable.

@chirimoya What's your take on this?

danrot:
I have talked with @chirimoya and he agrees with what I stated above. Therefore I am going to close this issue.

PR #4486, opened 2019-03-18
Add options constructor-parameter to SearchStore

nnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no

#### What's in this PR?

This PR adds a `options` parameter to the constructor of the `SearchStore` class. The properties of the given `options` object are included as query parameters in the requests sent by the `SearchStore`. This behaviour is analogous to the `options` parameter of the `ListStore` class and the `ResourceFormStore` class.

Furthermore, this PR add a `options` prop to the `SingleAutoComplete` container that allows to specify options that are passed to the `SearchStore` that is constructed by the `SingleAutoComplete` container. This behaviour is analogous to the `options` prop of the `SingleListOverlay` and the `MultiListOverlay`.

#### Why?

Because it might be useful to set some options that will be sent to the server. A simple usecase would be limiting the results of the `SearchStore` to media with a specific type by setting the following hypothetical `options` object `{type: 'image'}`.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/MultiAutoComplete/MultiAutoComplete.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/MultiAutoComplete/MultiAutoComplete.js
@@ -17,6 +17,7 @@ type Props = {|
     idProperty: string,
     locale?: ?IObservableValue<string>,
     onChange: (value: Array<string | number>) => void,
+    options: Object,
     resourceKey: string,
     searchProperties: Array<string>,
     value: ?Array<string | number>,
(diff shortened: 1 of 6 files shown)


nnatter:
@danrot should i add a `locale` parameter too? 
Also, would you be open to add an optional `options`/`apiOptions` prop to the `SingleAutoCompletion` container? I guess we will need something like that in the future anyway

danrot:
Would it be a lot of work to add the exact same thing for the `MultiAutoComplete` container component? If no I would prefer to add it immediately, in order to avoid diverging these two container components too much.

nnatter:
@danrot done 🙂

PR #4489, opened 2019-03-19
Added single-account-selection content type

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Added new content type for single account selection

#### To Do

### DIFF
--- a/src/Sulu/Bundle/ContactBundle/Tests/Unit/SingleAccountSelectionTest.php
+++ b/src/Sulu/Bundle/ContactBundle/Tests/Unit/SingleAccountSelectionTest.php
@@ -1,0 +1,190 @@
+<?php
+
+/*
+ * This file is part of Sulu.
+ *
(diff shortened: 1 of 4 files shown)


src/Sulu/Bundle/ContactBundle/Tests/Unit/SingleAccountSelectionTest.php
```diff
@@ -0,0 +1,190 @@
...
+/*
+ * This file is part of Sulu.
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\ContactBundle\Util;
```
alexander-schranz: false namespace

PR #4491, opened 2019-03-20
Set ghost-locale to first available-locale if locale not exists in webspace

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | none
| Related issues/PRs | none
| License | MIT
| Documentation PR | none

#### What's in this PR?

This PR fixes the ghost locale handling when you query for pages in a non-existing locale for the given webspace.

#### Why?

When you have 2 webspace with different locales:

* test1: de
* test2: en

And create a new custom-url for `test1` the content tree don't show any titles (user locale is en).

This PR uses the first available locale to display the titles

#### To Do

- [x] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@ CHANGELOG for Sulu
 ==================
 
 * release/1.5
+    * HOTFIX      #4491 [ContentBundle]         Set ghost-locale to first available-locale if locale not exists in webspace
     * BUGFIX      #4433 [SecurityBundle]       Fix fresh User object comparison to the deserialized User object
     
 * 1.5.21 (2019-02-28)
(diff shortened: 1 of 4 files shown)


composer.json
```diff
@@ -66,7 +66,8 @@
         "symfony/phpunit-bridge": "2.8.10",
         "symfony/monolog-bundle": "2.8.10",
         "jackalope/jackalope": "1.2.8 || 1.3.0 || 1.3.1",
-        "phpcr/phpcr-utils": "1.2.0 - 1.2.10 || 1.3.0 - 1.3.2"
+        "phpcr/phpcr-utils": "1.2.0 - 1.2.10 || 1.3.0 - 1.3.2",
+        "twig/twig": ">=2.7"
```
wachterjohannes: in 2.7 `\Twig_Runtime_Error` was deprecated and therfor the tests fail because of a deprecation notice. /cc @danrot @alexander-schranz what do you think - conflict or weak deprecation handler?

alexander-schranz: set deprecation helper to weak

PR #4499, opened 2019-03-25
Check whether media has thumbnails

HadGS:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

Check if media has thumbnails set in media selections.

#### Why?

When non picture media (e.g. Word document) is uploaded to single or multi media selection, whole admin interface broke down, because thumbnails property is undefined.

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/Resources/js/containers/MultiMediaSelection/MultiMediaSelection.js
+++ b/src/Sulu/Bundle/MediaBundle/Resources/js/containers/MultiMediaSelection/MultiMediaSelection.js
@@ -134,7 +134,7 @@ export default class MultiMediaSelection extends React.Component<Props> {
                                 key={media.id}
                             >
                                 <div className={multiMediaSelectionStyle.mediaItem}>
-                                    {media.thumbnails[THUMBNAIL_SIZE]
+                                    {typeof media.thumbnails !== 'undefined' && media.thumbnails[THUMBNAIL_SIZE]
                                         ? <img
                                             alt={media.title}
                                             className={multiMediaSelectionStyle.thumbnailImage}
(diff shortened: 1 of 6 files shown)


src/Sulu/Bundle/MediaBundle/Resources/js/containers/MultiMediaSelection/MultiMediaSelection.js
```diff
@@ -134,7 +134,7 @@ export default class MultiMediaSelection extends React.Component<Props> {
                                 key={media.id}
                             >
                                 <div className={multiMediaSelectionStyle.mediaItem}>
-                                    {media.thumbnails[THUMBNAIL_SIZE]
+                                    {typeof media.thumbnails !== 'undefined' && media.thumbnails[THUMBNAIL_SIZE]
```
danrot: I think instead of the `typeof` check I would just go for `media.thumbnails && media.thumbnails[THUMBNAIL_SIZE]`, in order to be more consistent. Or didn't that work for some reason?

src/Sulu/Bundle/MediaBundle/Resources/js/containers/SingleMediaSelection/SingleMediaSelection.js
```diff
@@ -118,7 +118,7 @@ export default class SingleMediaSelection extends React.Component<Props> {
                 >
                     {media &&
                         <div className={singleMediaSelectionStyle.mediaItem}>
-                            {media.thumbnails[THUMBNAIL_SIZE]
+                            {typeof media.thumbnails !== 'undefined' && media.thumbnails[THUMBNAIL_SIZE]
```
danrot: Same as above

HadGS:
Thanks for reply @danrot 
I will change the conditions as you requested. 
But for the tests... I tried to add them for about 20 minutes but every change i made in test.js files was not reflected in runtests program. Even 100% wrong expectation was "correct". I don't know, what i done wrong.
And for travis build, i started from Friday's develop version...

danrot:
The `bin/runtests` command only runs the PHP tests, not the JS ones. In order to run the JS tests, you have to execute `npm install` in the `sulu/sulu` repository and afterwards execute `npm run test`. Do you want to try that on your own, or should I add the tests?

HadGS:
Oh, that explains a lot! Please, add the JS test.

danrot:
@HadGS Added the tests, just in case you are curious :slightly_smiling_face: Waiting for the builds, then I am going to merge it. Thank you in advance!

PR #4504, opened 2019-03-28
[WebsiteBundle] dispatch an event after cache clear

### LINKED ISSUE #4503: [WebsiteBundle] dispatch an event after cache clear
| Q | A
| --- | ---
| Bug? | no
| New Feature? | yes
| Sulu Version | 1.6.25

#### Actual Behavior

When cache clear is called from the backoffice, a `DELETE /admin/website/cache`  request issued which clears the http cache.

#### Expected Behavior

I would like to have an event dispatched at this state, so I can hook an own cache clear listener that performs more tasks related to cache clearing. 
For example purging all caches from a CDN. 

#### Background

The production site is behind Cloudflare in our case. A normal cache clear removing the HTTP Cache is not enough, so we need to call the purge api.

#### Possible Solutions

Dispatch a dedicated cache clear event in this method:

https://github.com/sulu/sulu/blob/ae060d1b1403f789653cff5c2478da5154a6b04a/src/Sulu/Bundle/WebsiteBundle/Controller/CacheController.php#L29-L38

debreczeniandras:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #4503 
| License | MIT
| Documentation PR | comes after PR accepted

#### What's in this PR?

Allows a developer to hook into the cache clearing process.

* added an events class that could be extended in the future
* added a cache event that is being passed to the cache events - holding the current request
* dispatch the post clear cache event in the CacheController

#### Example Usage

~~~php
class CdnCachePurge implements EventSubscriberInterface
{
    public static function getSubscribedEvents()
    {
        return [
            \Sulu\Bundle\WebsiteBundle\Cache\Events::POST_CLEAR => 'postClear',
        ];
    }
  
    public function postClear(\Sulu\Bundle\WebsiteBundle\Cache\CacheEvent $event)
    {
           $request = $event->getRequest();
           // remove caches
    }
}
~~~

#### To Do


src/Sulu/Bundle/WebsiteBundle/Controller/CacheController.php
```diff
@@ -24,16 +27,21 @@ class CacheController extends Controller
...
-    public function clearAction()
+    public function clearAction(Request $request)
     {
         if (!$this->checkLivePermissionForAllWebspaces()) {
             return new JsonResponse(null, 403);
         }
 
         $this->get('sulu_website.http_cache.clearer')->clear();
 
+        $cacheEvent = new CacheEvent($request);
```
wachterjohannes: why do you pass the request here? do you need it for your usecase? and i think we should raise the event in the service. /cc @danrot @chirimoya

debreczeniandras: I don't necessarily need the request here (yet). When needed I could get request-related info from inside the eventsubscriber, so I can remove it.

The CacheClearer::clear() is also called from other controllers (e.g. `AnalyticsController`), that's why I wanted to be more explicit and only raise the event in the CacheController, when the user actually clicks on Settings/Clear cache. 

(As a side-note: the `CacheClearer service is getting a little bloated with dependencies.)

However, if you still think it better to dispatch the event from within the Service, I'll move it there, and then I will remove the `Events` and `CacheEvent` classes as obsolete.

src/Sulu/Bundle/WebsiteBundle/Cache/Events.php
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * Events related to clearing caches.
+ */
+final class Events
+{
+    /**
+     * Will be raised after http cache has been cleared.
+     */
+    const POST_CLEAR = 'sulu_website.cache.post_clear';
```
danrot: I get that the `post` in the name would make more clear when it happens, but I think we have never used a notion like this in our code yet. Therefore I would prefer to rename it to `sulu_website.cache_clear`, in order to be consistent. Consistency makes the product easier to use :slightly_smiling_face:

src/Sulu/Bundle/WebsiteBundle/Controller/CacheController.php
danrot: I would also argue that putting the event dispatching into the service is a good idea. If you want to execute something when the cache is cleared, you probably also want to do it in the `AnalyticsController`. Or at least you would expect it to be ran always.

debreczeniandras: True. Changed.

src/Sulu/Bundle/WebsiteBundle/Cache/Events.php
debreczeniandras: Modified it, and defined it as a class constant in the `CacheClearer` service.
Do you prefer to keep such constants in the related interface instead?

danrot: I think until now we've always had a separate `Events` class, as you've had here... Can you again build it that way?

debreczeniandras: I re-added the `Events` class.

src/Sulu/Bundle/WebsiteBundle/Cache/Events.php
```diff
@@ -0,0 +1,23 @@
...
+/*
+ * This file is part of Sulu.
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\WebsiteBundle\Cache;
```
alexander-schranz: Based on the Symfony Best Practice a Bundle should contain all Events in a Events Class in the Root of its Bundle.

debreczeniandras: I moved the Events class.

danrot:
@wachterjohannes @alexander-schranz What is the state of this? Looks good to me now, if it is also ok for you, can somebody of you tests and merge it?

debreczeniandras:
Added entry to changelog.

Do I need to create a documentation PR? 
If yes, where exactly?

danrot:
@debreczeniandras Sorry for the long delay :see_no_evil: We have not documented all of our events somewhere, so I think we can merge it as it is. Tests were failing, so I've restarted them, otherwise it looks good. If they succeed now I'll merge this PR.

danrot:
@debreczeniandras Thanks for the contribution! :tada:

PR #4516, opened 2019-04-15
Fix SuluKernel without GLOB_BRACE support

### LINKED ISSUE #4513: GLOB_BRACE breaks Sulu on Alpine Linux
| Q | A
| --- | ---
| Bug? | yes
| New Feature? | no
| Sulu Version | 2.0.0-alpha5
| Browser Version | n.a.

#### Actual Behavior

Sulu crashes on systems which don't support GLOB_BRACE.

#### Expected Behavior

Sulu works on systems that don't support GLOB_BRACE.

#### Steps to Reproduce

Install Sulu 2.0 on Alpine Linux.

#### Possible Solutions

In the SuluKernel use the Glob polyfill provided by the zendframework/zend-stdlib package.

tvwijgerden:
Added a fallback to GlobResource if GLOB_BRACE does not exist

| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #4513 
| Related issues/PRs | none
| License | MIT
| Documentation PR | none

#### What's in this PR?

Fix for GLOB_BRACE not working on systems that don't support GLOB_BRACE.

### DIFF
--- a/src/Sulu/Component/HttpKernel/SuluKernel.php
+++ b/src/Sulu/Component/HttpKernel/SuluKernel.php
@@ -14,6 +14,7 @@ namespace Sulu\Component\HttpKernel;
 use Symfony\Bundle\FrameworkBundle\Kernel\MicroKernelTrait;
 use Symfony\Component\Config\Loader\LoaderInterface;
 use Symfony\Component\Config\Resource\FileResource;
+use Symfony\Component\Config\Resource\GlobResource;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\HttpKernel\Kernel;
 use Symfony\Component\Routing\RouteCollectionBuilder;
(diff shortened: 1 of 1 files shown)


tvwijgerden:
@alexander-schranz the circleci failed because of an out of memory issue. I don't think that this has anything to do with the changes.

src/Sulu/Component/HttpKernel/SuluKernel.php
```diff
@@ -131,6 +132,17 @@ protected function import(RouteCollectionBuilder $routes, $glob)
         }
     }
 
+    private function glob($confDir, $pattern)
+    {
+        if (\defined('GLOB_BRACE')) {
```
alexander-schranz: We should always use the GlobResource class of symfony. It will use glob function if available and will fallback to a custom implementation itself.

/cc @danrot

tvwijgerden: Removed it

PR #4522, opened 2019-04-18
Scroll only tables in the List view

### LINKED ISSUE #5162: List inside tab view with many columns has strange scroll behaviour
| Q | A
| --- | ---
| Sulu Version | 2.0.6
| Browser Version | -

# Actual Behavior

A list inside a tab view has strange scroll behaviours.

tab-list

Tabs are scrolling away.

# Expected Behavior

The tabs should not scroll away also I think all controls should stay where they are and the scrolling should only be inside the list. Currently we did this because the design wanted that the table goes to the edge not sure if this was a good choice. 

Also think if you have multiple tables inside a view only inside the table the scrolling should happen and not the whole view should scroll.

# Steps to Reproduce

Create a tab with a many columns and many data in it.

# Possible Solutions

Set table as overflow-x: auto; and remove the strange `-5px` margin from it and when needed add it to the parent container.

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #5162
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR makes only the Tables scroll in the list view, and not the heading and navigation buttongs.

#### Why?

Because it feels weird otherwise.

#### To Do

- [ ] Remove horizontal scroll bar?

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Table/Table.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Table/Table.js
@@ -1,7 +1,7 @@
 // @flow
-import {observer} from 'mobx-react';
-import React from 'react';
+import React, {Fragment} from 'react';
 import type {ChildrenArray, Element} from 'react';
+import {observer} from 'mobx-react';
 import classNames from 'classnames';
 import Icon from '../Icon';
 import Header from './Header';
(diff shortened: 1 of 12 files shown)


danrot:
There are two problems remaining:

1. There is a scrollbar which does not really look nice, if we want the entire table to scroll and being cut on the left side as well:
Screenshot from 2019-04-18 14-09-16

2. The `selection` field type e.g. for categories has a max-width and scrolling enabled. This breaks with the current solution:
Screenshot from 2019-04-18 14-08-31

The first "problem" can easily be solved, since [scrollbars can be hidden](https://www.geeksforgeeks.org/hide-scroll-bar-but-while-still-being-able-to-scroll-using-css/), but it might be hard to scroll for users with a mouse without horizontally scrolling (It only works by clicking on the table to focus it, and press the arrow buttons afterwards).

It looks like the second problem is not that easy to fix, because you can't only scroll in one direction and not in the other. I might find a solution after some research, but that might again break the other use case in the lists.

The easiest solution would be to only scroll the table itself without let it run out of space on the right to indicate that there is more data (similar to what we did in the old UI).

@chirimoya What do you think?

danrot:
It would be worth a try to scroll the entire main area instead of only the table, and make the other elements sticky in some way.

danrot:
Making elements sticky using `position: sticky` seems to only work for vertical scrolling, not for horizontal :confused:

PR #4526, opened 2019-04-19
Fix accidental reload when tab is changed from dirty form

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR introduces a priority on the `UpdateRouteHook`s.

#### Why?

Because this gives us more control on when to cancel the navigation, and let other hooks not be executed at all. This is required, because the `ResourceStore` was reloaded before the confirmation dialog was confirmed when the tab is changed in a dirty form. This already resulted in data loss, and a second change of the tab is not caught anymore.

#### To Do

- [x] Check flow errors

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/services/Router/Router.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/services/Router/Router.js
@@ -48,8 +48,6 @@ export default class Router {
     }
 
     @computed get sortedUpdateRouteHooks() {
-        // TODO Remove type casts with any when flow bug is fixed
-        // https://github.com/facebook/flow/issues/7652
         return Object.keys(this.updateRouteHooks)
             .sort((a, b) => ((b: any): number) - ((a: any): number))
             .reduce((sortedUpdateRouteHooks, priority) => {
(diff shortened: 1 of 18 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/services/Router/Router.js
```diff
@@ -40,24 +40,44 @@ export default class Router {
...
                 event.returnValue = true;
             }
         });
     }
 
-    addUpdateRouteHook(hook: UpdateRouteHook) {
-        this.updateRouteHooks.push(hook);
+    @computed get sortedUpdateRouteHooks() {
+        // TODO Remove type casts with any when flow bug is fixed
+        // https://github.com/facebook/flow/issues/7652
```
chirimoya: remove

src/Sulu/Bundle/AdminBundle/Resources/js/views/Form/Form.js
```diff
@@ -198,7 +203,10 @@ class Form extends React.Component<Props> {
 
     componentWillUnmount() {
         const {router} = this.props;
-        router.removeUpdateRouteHook(this.checkFormStoreDirtyStateBeforeNavigation);
+        router.removeUpdateRouteHook(
```
chirimoya: not 100% happy ;-)

PR #4528, opened 2019-04-24
Add teaser selection

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR adds the already known `teaser_selection` field type.

#### Why?

Because it was already there in the old UI.

#### To Do

- [x] Implement `displayOption`/`presentation`
- [x] Rename `content` type in teaser items to `pages`
- [x] Add media to teaser form
- [x] Check why items cannot be preselected -> `NodeController` does not accept multiple `selectedIds` (fixed in a different way, and not required currently)
- [x] Write migration for renaming from `content` to `pages` and `displayOption` to `presentation` in PHPCR
- [x] Make use of teaser id for `key`, `id` and so on
- [x] Implement `disabled` flag for `TeaserSelection` field
- [x] Show edited state when some values from the page request are passed
- [ ] Assigning startpage does not work
- [x] Check why there is a pause after clicking the edit button of a teaser item -> `TextEditor` start is not very efficient, especially if multiple text editors are open -> somehow connected to the `buttonEnabled` observables in the plugins
- [x] Remove teaser should also consider type, not only id
- [x] Fix adding of already existing teaser items
- [x] `preSelectedItems` don't seem to be updated
- [x] Render placeholder for image if nothing is selected
- [x] CSS adjustments (Edited in same line, Paddings 12px from above)


src/Sulu/Bundle/PageBundle/Tests/Unit/Teaser/Provider/TeaserProviderPoolTest.php
```diff
@@ -72,8 +72,8 @@ public function testHasProviderNotFound()
     public function testGetConfiguration()
     {
         $configuration = [
-            'content' => new TeaserConfiguration('sulu_test.content', 'content@sulutest'),
-            'media' => new TeaserConfiguration('sulu_test.media', 'media@sulutest'),
+            'content' => new TeaserConfiguration('content', 'pages', 'column_list', ['title'], 'Choose'),
```
chirimoya: Pages

PR #4535, opened 2019-04-28
Update js dependencies

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

Update js dependencies.

#### Why?

Security reasons we should keep the packages up to date.

#### BC Breaks/Deprecations

See UPGRADE.md

#### To Do


UPGRADE.md
```diff
@@ -2,6 +2,16 @@
...
 ## dev-develop
 
+### Upgrade js dependencies
+
+**This change only affects you if you have used a 2.0.0 alpha release before**
+
+The following js dependencies in your `package.json` need to be updated:
+
+```bash
+npm install --save-dev raw-loader@^2.0 clean-webpack-plugin@^2.0 css-loader@^1.0 file-loader@^3.0 whatwg-fetch@^3.0 url-search-params-polyfill@^6.0 mini-css-extract-plugin@^0.6 autoprefixer@^9.5 postcss-calc@^7.0 postcss-import@^12.0 postcss-nested@^4.0 postcss-simple-vars@^5.0
```
danrot: Until now we've said that `sulu-minimal` should be checked to see if something changed... I am slowly worried that we blow up our `UPGRADE.md` in a way that nobody reads it anymore :confused:

alexander-schranz: In my opinion all changes which are required should be added to the UPGRADE.md. In the skeleton there are just sometimes changes which are optional which are not required.

UPGRADE.md
```diff
@@ -2,6 +2,8 @@
 
 ## dev-develop
 
+When upgrading also have a look at the changes in the [sulu skeleton](https://github.com/sulu/sulu-minimal/compare/2.0.0-alpha5...develop).
```
alexander-schranz: @danrot what you think about this? so we don't spam the  UPGRADE.md but its clearly linked and documented that changes from the skeleton should be added and we add the diff link to every release.

danrot: That's indeed better :+1: But shouldn't it be `2.0.0-alpha5...2.0.0-alpha4` then?

package.json
```diff
@@ -34,9 +34,9 @@
         "babel-plugin-transform-object-rest-spread": "^6.23.0",
         "babel-preset-env": "^1.6.1",
         "babel-preset-react": "^6.24.1",
-        "clean-webpack-plugin": "^1.0.0",
+        "clean-webpack-plugin": "^2.0.1",
         "core-js": "^2.5.1",
-        "css-loader": "^0.28.4",
+        "css-loader": "^1.0.1",
```
danrot: Doesn't version 2 work?

PR #4540, opened 2019-04-29
Fix copying of teaser selection in WebspaceCopyCommand

### LINKED ISSUE #4539: WebspaceCopyCommand does not copy teaser selection properties
| Q | A
| --- | ---
| Bug? | yes
| New Feature? | no
| Sulu Version | 1.6

#### Actual Behavior

Executing the webspace copy command does not copy 'teaser selection' content types.

#### Expected Behavior

It should copy it, just like the 'block' types.

#### Steps to Reproduce

Create two webspaces, then create a page type with a teaser_selection content type. Execute the webspace copy command, and observe that the teaser selection content type is not copied.

#### Possible Solutions

This case is not handled here:

https://github.com/sulu/sulu/blob/b8bdb0185b1e2f1e81fd560cab1b6ee012bfce36/src/Sulu/Bundle/ContentBundle/Command/WebspaceCopyCommand.php#L417

debreczeniandras:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #4539
| License | MIT

#### What's in this PR?

Process teaser_selection content type.

#### To Do

- [x] Write tests upon request

### DIFF
--- a/src/Sulu/Bundle/ContentBundle/Command/WebspaceCopyCommand.php
+++ b/src/Sulu/Bundle/ContentBundle/Command/WebspaceCopyCommand.php
@@ -557,8 +557,7 @@ class WebspaceCopyCommand extends ContainerAwareCommand
         $localeSource,
         $localeDestination
     ) {
-        if (!array_key_exists($property->getName(),
-                $structureArray) || !$structureArray[$property->getName()]['items']) {
+        if (!isset($structureArray[$property->getName()]['items'])) {
             return;
         }
 
--- a/src/Sulu/Bundle/ContentBundle/Command/WebspaceCopyCommand.php
+++ b/src/Sulu/Bundle/ContentBundle/Command/WebspaceCopyCommand.php
@@ -563,6 +563,10 @@ class WebspaceCopyCommand extends ContainerAwareCommand
         }
 
         foreach ($structureArray[$property->getName()]['items'] as $key => $teaserItem) {
+            if ('content' !== $teaserItem['type']) {
+                continue;
+            }
+
             $targetDocumentDestination = $this->getTargetDocumentDestination(
                 $teaserItem['id'],
                 $localeSource,
--- a/src/Sulu/Bundle/ContentBundle/Command/WebspaceCopyCommand.php
+++ b/src/Sulu/Bundle/ContentBundle/Command/WebspaceCopyCommand.php
@@ -455,6 +455,14 @@ class WebspaceCopyCommand extends ContainerAwareCommand
                     $localeDestination
                 );
                 break;
+            case 'teaser_selection':
+                $this->updateTeaserSelection(
+                    $structureArray,
+                    $property,
+                    $localeSource,
+                    $localeDestination
+                );
+                break;
         }
     }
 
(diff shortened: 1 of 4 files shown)


src/Sulu/Bundle/ContentBundle/Command/WebspaceCopyCommand.php
```diff
@@ -535,6 +543,40 @@ protected function updateSmartContentStructure(
...
+        $localeSource,
+        $localeDestination
+    ) {
+        if (!array_key_exists($property->getName(),
+                $structureArray) || !$structureArray[$property->getName()]['items']) {
+            return;
+        }
+
+        foreach ($structureArray[$property->getName()]['items'] as $key => $teaserItem) {
+            $targetDocumentDestination = $this->getTargetDocumentDestination(
```
alexander-schranz: is here not a check missing todo this only for `type` page?

alexander-schranz: Not sure where the type is saved but it shoud something like:
foreach (...) {
if ('page' !== $teaserItem['type']) {
     continue;
}

debreczeniandras: I couldn't yet find it, but why would you want to limit which type of teasers should be processed? 

The `getTargetDocumentDestination()` method should be able to resolve any (even custom) Teaser types based on the uuid. 
No?

alexander-schranz: It should skip other types like `article` or any other custom entity.

My example is false the type for pages is `content` so it should be something like:

```
foreach (...) {
     if ('content' !== $teaserItem['type']) {
         continue;
     }

     // ...
}
```

debreczeniandras: @alexander-schranz it's now breaking from the loop in case it's not 'content'.

alexander-schranz: looks correct for me. should work for pages (content) still correctly or?

debreczeniandras: yes, it does work. already tested it.

src/Sulu/Bundle/ContentBundle/Command/WebspaceCopyCommand.php
```diff
@@ -535,6 +543,44 @@ protected function updateSmartContentStructure(
...
+     * @param string $localeSource
+     * @param string $localeDestination
+     */
+    protected function updateTeaserSelection(
+        array &$structureArray,
+        PropertyMetadata $property,
+        $localeSource,
+        $localeDestination
+    ) {
+        if (!array_key_exists($property->getName(),
```
alexander-schranz: @debreczeniandras sorry for a change request again. Did test it on my current webspace and we need to change also this lines to:

```
        if (!isset($structureArray[$property->getName()]['items'])) {
            return;
        }
```

as in some cases the property exist but could be null and isset would be enough for this case and error proof.

alexander-schranz:
@debreczeniandras and can you update your branch to the current master branch and add a line to the `CHANGELOG.md`: 

```
    * BUGFIX      #4540  [ContentBundle]          Fix copying of teaser selection in WebspaceCopyCommand
```

src/Sulu/Bundle/ContentBundle/Command/WebspaceCopyCommand.php
debreczeniandras: @alexander-schranz no problem, it's now changed.

alexander-schranz: @debreczeniandras  can you add the CHANGELOG.md entry also .

danrot:
Shouldn't we also add a test for this? :thinking:

alexander-schranz:
Thank you @debreczeniandras 
It is released in version 1.6.27

PR #4541, opened 2019-04-29
Fix types for getDerivedRouteAttributes function

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | facebook/flow#6318
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR fixes the typing for the `getDerivedRouteAttributes` function of a `View`.

#### Why?

Because this allows to encounter errors in this function even earlier.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -11,5 +11,4 @@ bin/
 config/
 public/
 templates/
-tests/
 var/
--- a/.flowconfig
+++ b/.flowconfig
@@ -19,5 +19,4 @@ bin/
 config/
 public/
 templates/
-tests/
 var/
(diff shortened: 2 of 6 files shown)


.eslintignore
```diff
@@ -11,5 +11,4 @@ bin/
 config/
 public/
 templates/
-tests/
```
danrot: @chirimoya Seems like this did not only ignore the top level `tests` directory, also explains why it was faster :see_no_evil:

PR #4549, opened 2019-05-06
Fix app ready by remove bundleReady and startAdmin manually

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | yes
| Deprecations? | no
| Fixed tickets |  fixes https://github.com/sulu/sulu-minimal/pull/173, fixes https://github.com/sulu/sulu/pull/4425
| Related issues/PRs | https://github.com/sulu/sulu-minimal/pull/174
| License | MIT

#### What's in this PR?

Remove bundleReady and startApplication manually once.

#### Why?

 - fix problems with the BUNDLE_ENTRIES_COUNT (regenerator-runtime error)
 - allows to define the polyfills he wanted to use
 - allows to remove a core bundle
 - avoids that somebody need to call bundleReady in there bundles/application

#### Example Usage

~~~php
npm install
npm run build
~~~

#### BC Breaks/Deprecations

`bundleReady` and `bundleReadyPromise` was removed.

#### To Do


.gitattributes
```diff
@@ -22,3 +22,4 @@ UPGRADE.md export-ignore
 /public export-ignore
 /tests export-ignore
 /templates export-ignore
+/assets export-ignore
```
danrot: See https://github.com/sulu/sulu-minimal/pull/174/files#r284544230
I think we should not ignore in this folder, so that it can be reused in sulu-minimal.

danrot: We have decided to duplicate the code for now, because it allows more fine-grained control on what should be included in the build (e.g. if somebody only supports modern browser, they can omit the polyfills this way).

PR #4578, opened 2019-05-23
Add displayOption to MediaSelection

### LINKED ISSUE #3798: Fix default values for FieldTypes

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #3798 
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR adds the possibility to choose the display option for an image.

#### Why?

Because this was already possible in the old UI.

#### BC Breaks/Deprecations

Previously some display options were shown by default. This PR changes that behavior. Instead of showing some on default, if the display options are not set, the feature is disabled. This way we hope to avoid having displayOptions not having any effect on the website.

#### To Do

- [x] Show icons next to the `ArrowMenu.Action`
- [x] Allow to set default value
- [x] Add this feature to the `SingleMediaSelection` as well
- [x] ~~Change options to be snake case?~~


src/Sulu/Bundle/AdminBundle/Resources/js/components/Icon/fonts/selection.json
```diff
@@ -0,0 +1 @@
... 162.5c-11.314 13.577-31.492 15.411-45.069 4.097s-15.411-31.492-4.097-45.069l142.928-171.514z","M832 800c17.673 0 32 14.327 32 32s-14.327 32-32 32h-224c-17.673 0-32-14.327-32-32 0-66.060 32.033-111.44 106.377-179.589 2.531-2.32 19.067-17.381 23.717-21.666 8.651-7.971 15.412-14.394 21.501-20.483 28.158-28.158 40.405-48.366 40.405-66.261 0-42.667-16-64-32-64-15.262 0-23.018 0.517-31.35 2.6-21.602 5.401-32.65 20.131-32.65 61.4 0 17.673-14.327 32-32 32s-32-14.327-32-32c0-70.138 30.525-110.838 81.128-123.489 14.94-3.735 26.579-4.511 46.872-4.511 56 0 96 53.333 96 128 0 39.349-20.027 72.393-59.15 111.516-6.8 6.8-14.15 13.782-23.388 22.295-4.815 4.437-21.436 19.576-23.839 21.778-45.435 41.648-70.403 71.044-80.396 100.411h186.773z"],"attrs":[{},{}],"isMulticolor":false,"isMulticolor2":false,"colorPermutations":{"110211511511511":[{},{}],"1151151151115792551":[{},{}]},"tags":["subscript"],"grid":0},"attrs":[{},{}],"properties":{"order":1306,"id":32,"name":"subscript","prevSize":32,"code":59746},"setIdx":0,"setId":1,"iconIdx":42},{"icon":{"paths":["M187.077 544c-14.954 0-27.077-14.327-27.077-32s12.123-32 27.077-32h649.846c14.954 0 27.077 14.327 27.077 32s-12.123 32-27.077 32h-649.846z","M450
```
alexander-schranz: 👍

PR #4580, opened 2019-05-27
Add documents tab to contact and account form

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | yes
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR adds the documents tab to the account and contact form.

#### Why?

Because this tab already existed in the old UI.

#### To Do

- [x] ~~Add negative margin to list if resource tab title is given~~ Refactor to render title in views instead
    - [x] Form view
    - [x] List view
    - [x] MediaFormat view
    - [x] MediaHistory view
- [x] ~~Introduce separate `listKey` for the documents list?~~
- [x] Save the user settings separately from other lists (use `listKey` or `resourceKey` to make that distinction? Currently it seems to be mixed)
- [x] Fix double request when pressing back button in documents tab
- [x] Fix selected fields when pressing back button in documents tab (also check if bindings work after cancelling navigate)


src/Sulu/Bundle/ContactBundle/Admin/ContactAdmin.php
```diff
@@ -134,6 +139,16 @@ public function getRoutes(): array
                 ->addToolbarActions($formToolbarActions)
                 ->setParent(static::CONTACT_EDIT_FORM_ROUTE)
                 ->getRoute(),
+            $this->routeBuilderFactory->createListRouteBuilder('sulu_contact.contact_documents', '/documents')
```
danrot: `sulu_contact.contact_documents_list`?

src/Sulu/Bundle/ContactBundle/Admin/ContactAdmin.php
```diff
@@ -167,6 +182,16 @@ public function getRoutes(): array
                 ->addToolbarActions($formToolbarActions)
                 ->setParent(static::ACCOUNT_EDIT_FORM_ROUTE)
                 ->getRoute(),
+            $this->routeBuilderFactory->createListRouteBuilder('sulu_contact.account_documents', '/documents')
```
danrot: `sulu_contact.account_documents_list`?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/ViewRenderer/ViewRenderer.js
```diff
@@ -4,22 +4,35 @@ import type {Element} from 'react';
...
 
     componentDidMount() {
+        const {router} = this.props;
+
         this.clearSidebarConfig();
+
+        router.addUpdateRouteHook((newRoute, newAttributes) => {
+            const {attributes: oldAttributes, route: oldRoute} = router;
+            if (this.getKey(newRoute, newAttributes) !== this.getKey(oldRoute, oldAttributes)) {
+                router.clearBindings();
```
danrot: Previously was forgotten to clear the bindings on the router on the `restore` operation, but it was correctly done on the `navigate` and `redirect` operations.

At this place it is done for all three of them, so that an error like this should not occur anymore. The other reason to move it here is another bug I've encountered: If one of the `updateRouteHooks` registered cancel the navigation, the bindings were already cleared, although the navigation did not happen at all, which should also be fixed with this implementation. (However, this does not happen in the core currently, but can be tested by adding a `updateRouteHook` returning false anywhere.

chirimoya:
@danrot please rebase

src/Sulu/Bundle/AdminBundle/Resources/js/containers/ViewRenderer/ViewRenderer.js
chirimoya: 👍

PR #4581, opened 2019-05-28
Changed function call to create findOrCreate

kleinkoerkamp:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | 
| Related issues/PRs | 
| License | MIT

#### What's in this PR?

Creation of tags from Account and Contact working again.

#### Why?

Now error occurs when you add tag from Contact or Account


danrot:
Looks like this has never worked before, but it is possible to with the new UI :slightly_smiling_face: 

Could you also add tests, so that we don't lose this functionality again by accident? Feel free to ask for assistance if you need any.

danrot:
@kleinkoerkamp Thank you! Fixed a few more styling issue on my own, will merge as soon as the tests have passed.

danrot:
@kleinkoerkamp By the way, I've found a related error: This still does not work when a new contact is created. However, I am still going to merge this PR, since it keeps the user from saving. In the add form the save succeeds, but the tags are not attached to the contact.

PR #4587, opened 2019-06-04
Add ToolbarAction to add contacts in accounts people list

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR adds a new action for the peoples tab in the account form to add more contacts to the account.

#### Why?

Because this was already possible in the old UI.

#### To Do

- [x] Fix adding contacts that have already been added

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/index.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/index.js
@@ -19,6 +19,7 @@ import Form, {CardCollection, fieldRegistry, FormInspector, ResourceFormStore, R
 import type {SchemaOption} from './Form/types';
 import ResourceLocatorHistory from './ResourceLocatorHistory';
 import ResourceMultiSelect from './ResourceMultiSelect';
+import ResourceSingleSelect from './ResourceSingleSelect';
 import MultiAutoComplete from './MultiAutoComplete';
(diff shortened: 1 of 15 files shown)


danrot:
Regarding the open to do "Fix adding contacts that have already been added" I see multiple options.

Either we show some kind of error message if somebody tries to add a contact that already existed, which does not feel very nice to me. I think it would be better if the auto complete does not show the contacts that have already been assigned. However, this does not seem to be easily possible, because (at least until now) we've always used the list endpoints (the one with the `flat=true` flag) and this endpoint does not support returning multiple accounts. The only account it returns and allows to filter for is the main account, but we would need to check if any of the linked accounts is the one we are trying to add...

So I came up with three different options:

1. Show an error message instead of filtering the auto complete
2. Extend the ListBuilder to be able to work with joins returning multiple rows (not sure if we want that, since it is not really possible to show that in our `List` component)
3. Do not use the flat API response and extend the other one with the `excludedAccountId` parameter resp. create a completely new one.

danrot:
Talked with @chirimoya, we'll implement Option 3 with a separate endpoint for this.

Issue #1482, opened 2015-08-14
Removing some permissions from user groups causes errors on other interfaces


kriswillis:
# Reproducing

If you were to remove all categories permissions from a group, users within that group then get errors within other interfaces (that use categories) that they do have permission for. If you were to navigate to edit a person, an alert with "Operation forbidden. The available permissions are not sufficient for this operation" is displayed, and the category selection has an infinite loading animation (due to the XHR failing with a 401).

Similarly, this also happens when removing "Organizations" permissions from a user; when they they try to edit a user an error alert is shown.

Again, removing permissions for "Tags" has a similar effect.
# What should happen

When a user doesn't have any permissions on an entity, controls for this entity should not be displayed. For example, having removed all permissions on "Categories", when navigating to edit a Person, the Category selection should not be visible, if they only have "view" permissions, the selection should be displayed read-only, etc.

danrot:
Disabling elements in the interface the user is not allowed to use is on our list, but I am not sure if we want e.g. the categories to be hidden then completely...

danrot:
This also occurs when the user has no permission to view roles and he tries to update the security for collections:

![image](https://cloud.githubusercontent.com/assets/405874/15185844/fd6cc7b6-179a-11e6-82f5-76960a8a0654.png)

And the same goes for the permission tab in pages, ...:
![image](https://cloud.githubusercontent.com/assets/405874/15185857/0fd991fe-179b-11e6-88ed-bfaa5d6a2584.png)

danrot:
We have mitigated this issue, by using the `EDIT` permission to decide if a navigation item should show up. So you can keep the `VIEW` permission selected on every entity that is needed by the user to successfully work with the system.

Hiding fields based on the permissions raises new questions and making stuff more complicated as it currently is, so we are not fixing it this way.

PR #4589, opened 2019-06-04
Set user permissions on new content page

### LINKED ISSUE #3705: Save and publish option is not deactivated when the user has no publish permisssion
| Q | A
| --- | ---
| Bug? | yes
| New Feature? | no
| Sulu Version | ^1.6
| Browser Version | Chromium 62.0.3202.94

#### Actual Behavior

When you create a role without publishing permissions. The CMS still shows the **Save and publish** option. Using that option will result in permission notices.
image

image

#### Expected Behavior

I would expect that the **Save and publish** option is disabled (like Publish).

#### Steps to Reproduce

- Create a new role: https://sulu.rocks/admin/#settings/roles/new where the Global **live** permission is deactivated.
- Create a CMS user for this role.
- Login in the CMS with this user.
- Add a new content page https://sulu.rocks/admin/#content/contents/sulu_demo/de/add/content
- The **save and publish** button will be active and using it will result in notices.

dsuurlant:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #3705 
| Related issues/PRs | #1482
| License | MIT
| Documentation PR | -

#### What's in this PR?

The fix loads the webspace's default content when no content id is present, but makes sure to only take the permissions from that response so we don't end up with conflicting data.

#### Why?

This fixes missing user permissions when creating a new content page, which causes the Save menu to display incorrect options.

### DIFF
--- a/src/Sulu/Bundle/ContentBundle/Resources/public/js/components/content/main.js
+++ b/src/Sulu/Bundle/ContentBundle/Resources/public/js/components/content/main.js
@@ -170,8 +170,20 @@ define([
                     }
                 );
             } else {
-                this.data = this.content.toJSON();
-                promise.resolve();
+                var baseContent = new Content();
+                baseContent.fullFetch(
+                  this.options.webspace,
+                  this.options.language,
+                  true,
+                  template,
+                  {
+                      success: function(content) {
+                          this.data = this.content.toJSON();
+                          this.data._permissions = content._permissions;
+                          promise.resolve();
+                      }.bind(this)
+                  }
+                );
             }
 
             return promise;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@ CHANGELOG for Sulu
 ==================
 
 * release/1.5
+    * BUGFIX      #4589 [ContentBundle]         Set user permissions on new content page
     * BUGFIX      #4572 [WebsiteBundle]         Fix false 406 status code for controller InvalidArgumentExceptions
     * BUGFIX      #4571 [WebsiteBundle]         Fix language switcher for static routes
     * BUGFIX      #4579 [ContentBundle]         Fix controller reference validation with FQCN


danrot:
I don't really get what error you are trying to fix here. Can you elaborate a bit further? Maybe even with a step-by-step reproduction?

dsuurlant:
We uncovered this error when implementing more fine-tuned user roles for the CMS end users.

Let's say User A has the role with permission to create new articles and save them as drafts, but not to publish them.
User B has the role with permission to also publish articles. 

When User A creates a new article, the menu shows the wrong options:
```
(greyed out) Save as Draft
(greyed out) Save and Publish
(available) Publish Now
```

However, the menu *should* show:
```
(available) Save As Draft
(greyed out) Save and Publish
(greyed out) Publish Now
```

After digging through the frontend code, I figured out that the user permissions on the content editor are retrieved through an API call. Only,  the new article just exists in the frontend and isn't saved to the database yet, so that API call returns empty.

Which means:
`this.data = this.content.toJSON();` leaves the user permissions empty, and the menu options wrong for the logged in user's actual permission.

This PR fixes this by retrieving the default user permissions for the webspace if we are editing a new article:

```
if (this.options.id !== undefined) {
 // retrieve content including user permissions for existing article
} else {
 // retrieve content from the webspace and use it only to set the user permissions
}
```

src/Sulu/Bundle/ContentBundle/Resources/public/js/components/content/main.js
```diff
@@ -170,8 +170,20 @@ define([
                     }
                 );
             } else {
-                this.data = this.content.toJSON();
-                promise.resolve();
+                var baseContent = new Content();
+                baseContent.fullFetch(
```
danrot: So we have to send a request when creating a new page? Where does this request go? I think that's a little bit much, especially since the permissions for the current user can be taken from `config.get('sulu_security.contexts')`.

dsuurlant: Thanks for the tip! I'll change this to use `sulu_security.contexts` instead.

alexander-schranz:
I will close this issue because inactivity. Feel still free to open a new PR if it is still relevant, to rediscuss the things.
Still thanks for all time which was spend here into this PR.

PR #4593, opened 2019-06-05
Add loadExcerpt to snippet twig extension

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes 
| Related issues/PRs | #4573
| License | MIT

#### What's in this PR?

Add new parameter to load the excerpt over the snippet twig extension.

#### Why?

Should also be possible from twig not only from content type.

#### Example Usage

~~~php
{% set snippet = sulu_snippet_load(uuid, null, true) %}
~~~


alexander-schranz:
moved to #4592

PR #4594, opened 2019-06-06
Add separate FieldType for route paths

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR adds the `route` FieldType, which is very similar to the `resource_locator` field type. However, the `resource_locator` FieldType is only working for pages. Both make use of the abstract `ResourceLocator` FieldType, which had to be adapted in order to work for both cases.

#### Why?

Because the `route` FieldType is used in the SuluArticleBundle, which has been updated to Sulu 2.0 in sulu/SuluArticleBundle#377.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -2,6 +2,48 @@
 
 ## dev-develop
 
+### RouteBundle API
+
+The API for routes under `/admin/api/routes` has changed a bit, in order to avoid having PHP class names in the
+frontent. Instead it uses the already known `resourceKey` now.
+
+Therefore the API has changed, instead of `entityClass` it now takes the `resourceKey`, and the `entityId` attribute has
+changed to just `id`.
+
+```
+# before
+/admin/api/routes?history=true&entityClass=Sulu\Bundle\ArticleBundle\Document\ArticleDocument&entityId=c88e4b89-7e2b-4161-b5d0-c33993535140&locale=en
+
+# after
+/admin/api/routes?history=true&resourceKey=articles&id=101b58ef-d422-4122-98f9-a4009bd74bd1&locale=en
+```
+
+This `resourceKey` now has to be configured in the `sulu_route` configuration as well:
+
+```yaml
+# before
+sulu_route:
+    mappings:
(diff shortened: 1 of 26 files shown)


UPGRADE.md
```diff
@@ -2,6 +2,48 @@
...
+        Sulu\Bundle\ArticleBundle\Document\ArticleDocument:
+            generator: "schema"
+            options:
+                route_schema: "/articles/{object.getTitle()}"
+
+# after
+sulu_route:
+    mappings:
+        Sulu\Bundle\ArticleBundle\Document\ArticleDocument:
+            resource_key: "articles"
```
danrot: That's maybe something we should also discuss with @chirimoya before merging it.

alexander-schranz: I personally would, because of problems in the past with overwriting and extending entities, prefer always referencing by resource_key (so also have the resource_key in the database) and get ride completely of the entityClass in the route bundle.

danrot: Nice idea, but that would be quite a breaking change, for which we can't even provide a migration by ourselves (because we don't know the other custom entities). We just have to decide if it is worth the breaking change.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/tests/fields/ResourceLocator.test.js
```diff
@@ -72,33 +115,135 @@ test('Throw an exception if a non-valid mode is passed', () => {
...
+                }}
                 formInspector={formInspector}
                 schemaOptions={schemaOptions}
             />
         )
     ).toThrow(/"leaf" or "full"/);
 });
 
-test('Throw an exception if a no generationUrl is passed', () => {
+test('Throw an exception if no hisotryResourceKey is given', () => {
```
wachterjohannes: `historyResourceKey `

PR #4598, opened 2019-06-11
Add missing UPGRADE information for upgrading database to UTF-8

danrot:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

There were some missing information and edge case handling when upgrading from Sulu 1.6.


UPGRADE.md
```diff
@@ -1503,6 +1503,56 @@ ALTER TABLE ca_category_translation_medias ADD CONSTRAINT FK_39FC41BA17CA14DA FO
 ALTER TABLE ca_category_translation_medias ADD CONSTRAINT FK_39FC41BA7DE8E211 FOREIGN KEY (idMedia) REFERENCES me_media (id) ON DELETE CASCADE;
 ```
 
+In addition that also the PHPCR tables have to be changed to utf8mb4 in case jackalope-doctrine-dbal is used:
+
+```sql
+ALTER TABLE `phpcr_binarydata` CHARACTER SET = utf8mb4;
```
alexander-schranz: The above queries don't switch tables to utf8mb4 but to beneath does so after doing this you have mixed tables utf8 in the sulu tables and utf8mb4 in the phpcr nodes table so this is not correctk to set the character set here but not for the sulu tables.

danrot: The thing is that if you are updating your schema when using the new `sulu-minimal`, then the all the tables generated by the ORM will be updated, but the jackalope tables will be untouched. That's also a strange behavior... I think we have to discuss this further.

chirimoya:
ping @alexander-schranz @danrot

alexander-schranz:
@chirimoya as said this is not needed and also its false to do this.

The problem here was the database was created and running as utf8 but configured was in the doctrine bundle utf8mb4 and then phpcr expect to be utf8mb4 else will crash.

So if you want to change the database from utf8 to utf8mb4 you would need to export it and import it to a utf8mb4 database.

alexander-schranz:
If we want to document how to switch from utf8 to utf8mb4 the following queries are needed:

```sql
     ALTER TABLE phpcr_namespaces CHANGE prefix prefix VARCHAR(191) NOT NULL;
     ALTER TABLE phpcr_workspaces CHANGE name name VARCHAR(191) NOT NULL;
     ALTER TABLE phpcr_nodes CHANGE path path VARCHAR(191) NOT NULL, CHANGE parent parent VARCHAR(191) NOT NULL, CHANGE local_name local_name VARCHAR(191) NOT NULL, CHANGE namespace namespace VARCHAR(191) NOT NULL, CHANGE workspace_name workspace_name VARCHAR(191) NOT NULL, CHANGE identifier identifier VARCHAR(191) NOT NULL, CHANGE type type VARCHAR(191) NOT NULL;
     ALTER TABLE phpcr_internal_index_types CHANGE type type VARCHAR(191) NOT NULL;
     ALTER TABLE phpcr_binarydata CHANGE property_name property_name VARCHAR(191) NOT NULL, CHANGE workspace_name workspace_name VARCHAR(191) NOT NULL;
     ALTER TABLE phpcr_nodes_references CHANGE source_property_name source_property_name VARCHAR(191) NOT NULL;
     ALTER TABLE phpcr_nodes_weakreferences CHANGE source_property_name source_property_name VARCHAR(191) NOT NULL;
     ALTER TABLE phpcr_type_nodes CHANGE name name VARCHAR(191) NOT NULL;
     ALTER TABLE phpcr_type_props CHANGE name name VARCHAR(191) NOT NULL;
```

And then export and then import as utf8mb4 of the whole mysql database is needed. This SQL is not needed when not using utf8mb4 or not using mysql.

UPGRADE.md
danrot: When thinking about it, I would say we should add these statements for the above tables as well... If the developer would not switch to utf8mb4, then the above lines would also not really be needed...

alexander-schranz: They are needed because our schema is defined so. Else it will not be a validate schema. 
The different between sulu tables and phpcr is that sulu in general uses shorter characters and phpcr used only shorted characters when using utf8mb4.

danrot: I know this difference, but according to this [great article about converting to `utf8mb4`](https://mathiasbynens.be/notes/mysql-utf8mb4) using just `utf8` also opens us up for some security vulnerabilities. So I would suggest we add the charset for all the tables to the `UPGRADE.md` as well.

danrot:
This would be my current suggestion, also considering what I've added in sulu/sulu-docs#465. As explaing in [the above comment](https://github.com/sulu/sulu/pull/4598#discussion_r344060666) I would suggest we always make the entire move to `utf8mb4` by default, because `utf8` also seems to cause security vulnerabilities.

alexander-schranz:
@danrot Ok got it now :) looks good for me 👍

danrot:
@alexander-schranz Can you take care of merging it? Or do you think somebody else should have a look as well?

UPGRADE.md
```diff
@@ -2474,6 +2474,84 @@ To support utf8mb4 we needed to change some database entities which you also nee
 Run the following SQL to migrate to the new schema:
 
 ```sql
+ALTER TABLE ca_categories CONVERT TO CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
```
alexander-schranz: @danrot depending on the mysql version this could throw an error todo this before shorten the indexed fields.

danrot: Are you sure? The [guide I based this upon](https://mathiasbynens.be/notes/mysql-utf8mb4) seems pretty complete, and is also explaining it this way. Would be even more afraid that it might not work the other way around :see_no_evil: Have you ever seen this failing?

alexander-schranz: It depends on the mysql version. `5.6.21` for example will throw an error in this case.

UPGRADE.md
```diff
@@ -2506,6 +2584,63 @@ ALTER TABLE ca_category_translation_medias ADD CONSTRAINT FK_39FC41BA17CA14DA FO
...
+ALTER TABLE `phpcr_type_nodes` CHANGE `name` `name` VARCHAR(191)  CHARACTER SET utf8mb4  NOT NULL  DEFAULT '';
+ALTER TABLE `phpcr_type_nodes` CHANGE `supertypes` `supertypes` VARCHAR(255)  CHARACTER SET utf8mb4  NOT NULL  DEFAULT '';
+ALTER TABLE `phpcr_type_nodes` CHANGE `primary_item` `primary_item` VARCHAR(255)  CHARACTER SET utf8mb4  NULL  DEFAULT NULL;
+
+ALTER TABLE `phpcr_type_props` CHANGE `name` `name` VARCHAR(191)  CHARACTER SET utf8mb4  NOT NULL  DEFAULT '';
+ALTER TABLE `phpcr_type_props` CHANGE `default_value` `default_value` VARCHAR(255)  CHARACTER SET utf8mb4  NULL  DEFAULT NULL;
+
+ALTER TABLE `phpcr_workspaces` CHANGE `name` `name` VARCHAR(191)  CHARACTER SET utf8mb4  NOT NULL  DEFAULT '';
+```
+
```
alexander-schranz: @danrot the database default character set and collate is missing here or?

```
ALTER DATABASE <database_name> CHARACTER SET = utf8mb4 COLLATE = utf8mb4_unicode_ci;
```

PR #4603, opened 2019-06-13
Using alias so that every project can override template

kleinkoerkamp:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

With this change in every project the admin base template is overridable and can be extended with project custom options.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Controller/AdminController.php
+++ b/src/Sulu/Bundle/AdminBundle/Controller/AdminController.php
@@ -233,7 +233,7 @@ class AdminController
         ];
 
         return $this->engine->renderResponse(
-            'SuluAdminBundle:Admin:main.html.twig',
+            '@SuluAdmin\Admin\main.html.twig',
             [
                 'translations' => $this->translations,
                 'fallback_locale' => $this->fallbackLocale,


alexander-schranz:
@kleinkoerkamp thank you for your pull request, for me its ok to make it overrideable, but I'm interested in what you want to override in [this template](https://github.com/sulu/sulu/blob/2.0.0-alpha6/src/Sulu/Bundle/AdminBundle/Resources/views/Admin/main.html.twig) the entry point for your custom js should be here:

https://github.com/sulu/sulu-minimal/blob/11bc9d23a9a571f733d1b69d8c0fe85d1a84e577/assets/admin/index.js#L23

And if you want to extend the config there will be a function in the admin soon which will be called: https://github.com/sulu/sulu/pull/4597

If we make the main.html.twig overrideable we should also make some twig blocks so he can override only specific parts of that template so he will not have problems when upgrading to a newer version.

kleinkoerkamp:
@alexander-schranz correct i found that out later as well. This is more something i ran into to make it more extendable.

danrot:
Change looks good, although I would still like to test it (you never know :wink:). However, can you rebase this brach before that? One of the tests failed, and I am wondering if you just branched of at a point where our tests were broken...

danrot:
@kleinkoerkamp Tested it locally, and it seems to be fine. Thanks for your contribution! :tada:

PR #4606, opened 2019-06-14
Avoid linebreak with long template titles in toolbar dropdown

danrot:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This title avoid line-breaks in the toolbar buttons.

#### Why?

Because if you had a long title for a template, then the dropdown in the toolbar of the Page form looked strange.

#### To Do

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Toolbar/button.scss
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Toolbar/button.scss
@@ -54,6 +54,7 @@ $primaryBackgroundColor: $shakespeare;
     background-color: transparent;
     font-size: 12px;
     cursor: pointer;
+    white-space: nowrap;
 
     &:disabled {
         cursor: default;


danrot:
@chirimoya This way the text in the button never breaks, but if we want to shorten the text we have to give the button some max-width. If we don't do that, then the button will stretch as long as it needs to be, but the text of the buttons is hidden earlier.

src/Sulu/Bundle/AdminBundle/Resources/js/components/Toolbar/button.scss
```diff
@@ -54,6 +54,7 @@ $primaryBackgroundColor: $shakespeare;
     background-color: transparent;
     font-size: 12px;
     cursor: pointer;
+    white-space: nowrap;
```
danrot: Here we would have to also set a `max-width`, `overflow: hidden` and `text-overflow: ellipsis`, if we want the text to be shortened at some point.

PR #4617, opened 2019-07-03
Avoid parallel generation of images

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes -
| Related issues/PRs | -
| License | MIT

#### What's in this PR?

Avoid parallel generation of images by using the symfony lock component.

#### Why?

Image generation could end up in high memory usage if a image is currently requested parallel by different clients a server should just generate it once and not multiple times to avoid high memory usage.

#### To Do

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/Controller/MediaStreamController.php
+++ b/src/Sulu/Bundle/MediaBundle/Controller/MediaStreamController.php
@@ -21,9 +21,14 @@ use Sulu\Bundle\MediaBundle\Media\Manager\MediaManagerInterface;
 use Sulu\Bundle\MediaBundle\Media\Storage\StorageInterface;
 use Symfony\Bundle\FrameworkBundle\Controller\Controller;
 use Symfony\Component\HttpFoundation\BinaryFileResponse;
+use Symfony\Component\HttpFoundation\RedirectResponse;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpFoundation\ResponseHeaderBag;
+use Symfony\Component\Lock\Factory;
+use Symfony\Component\Lock\Lock;
+use Symfony\Component\Lock\Store\FlockStore;
+use Symfony\Component\Lock\Store\SemaphoreStore;
 
 class MediaStreamController extends Controller
 {
@@ -57,6 +62,20 @@ class MediaStreamController extends Controller
             $url = $request->getPathInfo();
 
             list($id, $format) = $this->getCacheManager()->getMediaProperties($url);
+
+            $lockingKey = sprintf('media-cache_%s_%s', $id, $format);
+
+            $lock = $this->createLock($lockingKey);
+
+            if ($lock && !$lock->acquire()) {
+                // wait for releasing of lock
+                if ($lock->acquire(true)) {
+                    // when getting the lock release it as soon as possible
+                    $lock->release();
+                }
+
+                return new RedirectResponse($request->getUri());
+            }
 
             return $this->getCacheManager()->returnImage($id, $format);
         } catch (ImageProxyException $e) {
@@ -273,5 +292,21 @@ class MediaStreamController extends Controller
         }
 
         return $this->storage;
+    }
+
+    private function createLock(string $key): ?Lock
+    {
+        if (!class_exists(Factory::class)) {
+            // for locking the symfony/lock is required if not installed the media stream is not locked
+            return null;
+        }
+
+        if (SemaphoreStore::isSupported()) {
+            $store = new SemaphoreStore();
+        } else {
+            $store = new FlockStore();
+        }
+
+        return (new Factory($store))->createLock($key);
     }
 }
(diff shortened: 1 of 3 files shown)


chirimoya (approved):

src/Sulu/Bundle/MediaBundle/DependencyInjection/Configuration.php
```diff
@@ -96,6 +96,7 @@ public function getConfigTreeBuilder()
                             'Expires' => '+1 month',
                             'Pragma' => 'public',
                             'Cache-Control' => 'public',
+                            'X-SULU-MEDIA-CACHE' => 'MISS',
```
danrot: What is this header doing? Doesn't that always render the same header?

src/Sulu/Bundle/MediaBundle/Controller/MediaStreamController.php
```diff
@@ -274,4 +293,20 @@ protected function getStorage()
...
+            return null;
+        }
+
+        if (SemaphoreStore::isSupported()) {
+            $store = new SemaphoreStore();
+        } else {
+            $store = new FlockStore();
+        }
+
+        return (new Factory($store))->createLock($key);
```
danrot: If I read about the lock component in [this blog](https://symfony.com/blog/new-in-symfony-3-3-lock-component) I see that there should also be a service. Why aren't we using the service but creating the factory class and so on ourselves?

danrot: And this implementation would not allow to e.g. use the `RedisStore`, which would maybe be necessary if we have a distributed environment.

src/Sulu/Bundle/MediaBundle/Controller/MediaStreamController.php
```diff
@@ -58,6 +63,20 @@ public function getImageAction(Request $request)
 
             list($id, $format) = $this->getCacheManager()->getMediaProperties($url);
 
+            $lockingKey = sprintf('media-cache_%s_%s', $id, $format);
+
+            $lock = $this->createLock($lockingKey);
+
+            if ($lock && !$lock->acquire()) {
+                // wait for releasing of lock
+                if ($lock->acquire(true)) {
```
danrot: Just reading the documentation, so I am not sure about this, but wouldn't this suffice:

```php
if ($lock && !$lock->acquire(true)) {
    $lock->release();
    return new RedirectResponse($request->getUri());
}
```

Resp. why isn't that enough?

danrot (changes requested):

src/Sulu/Bundle/MediaBundle/DependencyInjection/Configuration.php
alexander-schranz: Its only set in PHP so when the image is generated, nginx will not set this header.

src/Sulu/Bundle/MediaBundle/Controller/MediaStreamController.php
alexander-schranz: See no possibility to use a `key` based locking over the service just a general lock?

Thats by design so as thumbnails by default are generated per server and should not block other servers.

danrot: From what I see in executing the `debug:container` command is that there is a `lock.factory` service, which is a class of type `Symfony\Component\Lock\Factory`. There you should be able to pass a key, shouldn't you?

There seems to be no documentation about this yet, but there is some explanation in [this PR](https://github.com/symfony/symfony/pull/22113).

src/Sulu/Bundle/MediaBundle/DependencyInjection/Configuration.php
danrot: But why is it needed?

danrot (commented):
Are there also any news on the other comments?

src/Sulu/Bundle/MediaBundle/DependencyInjection/Configuration.php
alexander-schranz: To see if the nginx correctly return the images and not always php is called and images are regenerated.

src/Sulu/Bundle/MediaBundle/Controller/MediaStreamController.php
alexander-schranz:
```php
$lock->acquire(true)
```

will waid until the other process release it and then return true so instead of doing a redirect it will regenerated the image which was generated before.

Thats why first we do a none blocking request for the lock if thats false we know somebody else is generating the image next we do a block request just to wait until that process did successfully generated the image and so release and do a redirect then.

src/Sulu/Bundle/MediaBundle/DependencyInjection/Configuration.php
danrot: Okay, totally makes sense :+1:

src/Sulu/Bundle/MediaBundle/Controller/MediaStreamController.php
danrot: Okay, now I understand it :+1:

src/Sulu/Bundle/MediaBundle/Controller/MediaStreamController.php
danrot: Any reason not to use the `lock.factory` service?

alexander-schranz:
Will currently close it, we should have a deeper look at it also in case of multi server setup with a decentral image storage on the filesystem.

PR #4627, opened 2019-07-12
Fix Routing when no prefix is Provided

### LINKED ISSUE #601: 301 redirect for requests with trailing slashes
Adding a trailing slash to a request to an existing page (e.g. calling http://www.myrobotcenter.com/de/team/ instead of http://www.myrobotcenter.com/de/team) should not cause a 404 error but redirect (301) to the page without slash.

Attention: If the requested page does not exist, it should cause a 404 immediately (e.g. it should not redirect from example.com/page/ to example.com/page and then to the 404 page but directly to the 404).

marcostastny:
There exists a Bug where routing would not redirect you correctly if you don't use the {locale} parameter in Webspace.xml.

For Example:
https://www.massiveart.com/blog/
should redirect to
https://www.massiveart.com/blog

But it doesn't.

| Q | A
| --- | ---
| Bug fix? | Yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #601
| Related issues/PRs | #601
| License | MIT

#### What's in this PR?

Routing Bugfix.

#### Why?

As above, fixes #601.

#### To Do


alexander-schranz:
@marcostastny tests are failing and there should a new test for this use case be added.

alexander-schranz:
@marcostastny can you add a entry to the CHANGELOG.md?

marcostastny:
@alexander-schranz done

src/Sulu/Bundle/WebsiteBundle/Routing/ContentRouteProvider.php
```diff
@@ -148,7 +148,11 @@ public function getRouteCollectionForRequest(Request $request)
                 return $collection;
             }
 
-            if (preg_match('/\/$/', $resourceLocator) && $prefix) {
```
danrot: Any reason you didn't just change the check to the following lines:

```php
if (preg_match('/\/$/', $resourceLocator)) {
    // ...
}
```

This works for me as well, without an additional if.

danrot: Ok, it does not work for the homepage :see_no_evil: What about the following approach:

```php
$prefix = $attributes->getAttribute('resourceLocatorPrefix') ?? '';

// ...

if (preg_match('/\/$/', $resourceLocator) && $resourceLocator !== '/) {
    // ...
}
```

This would avoid the deeper nested if and the need to change a variable in the middle of the function, which could lead to bad surprises.

src/Sulu/Bundle/WebsiteBundle/Tests/Unit/Sulu/Bundle/WebsiteBundle/Routing/ContentRouteProviderTest.php
```diff
@@ -724,6 +724,46 @@ public function testGetCollectionTrailingSlashWithQueryParams()
...
+        $webspace->setTheme('theme');
+        $portal->setWebspace($webspace);
+        $attributes->getAttribute('portal', null)->willReturn($portal);
+
+        $localization = new Localization('de', 'at');
+        $attributes->getAttribute('localization', null)->willReturn($localization);
+
+        $attributes->getAttribute('matchType', null)->willReturn(RequestAnalyzer::MATCH_TYPE_FULL);
+        $attributes->getAttribute('resourceLocator', null)->willReturn('/qwertz/');
+        $attributes->getAttribute('resourceLocatorPrefix', null)->willReturn('');
```
danrot: Don't you have to return `null` here, in order to test your use case?

src/Sulu/Bundle/WebsiteBundle/Routing/ContentRouteProvider.php
alexander-schranz: @danrot this will not work for homepage with prefix e.g. `/de/`:

Most cases are tested and `if (preg_match('/\/$/', $resourceLocator) && $prefix) {` was in our testing the only if statement which does really work for all cases.

danrot: Ok, sounds reasonable.

src/Sulu/Bundle/WebsiteBundle/Routing/ContentRouteProvider.php
```diff
@@ -148,7 +148,11 @@ public function getRouteCollectionForRequest(Request $request)
...
+            if (preg_match('/\/$/', $resourceLocator) && ('/' !== $resourceLocator || $prefix)) {
+                if (is_null($prefix)) {
```
danrot: But I think you don't need that check at all, because null would be casted to an empty string anyway. Or was there a reason for that?

src/Sulu/Bundle/WebsiteBundle/Tests/Unit/Sulu/Bundle/WebsiteBundle/Routing/ContentRouteProviderTest.php
marcostastny: @danrot Changed

src/Sulu/Bundle/WebsiteBundle/Routing/ContentRouteProvider.php
marcostastny: @danrot fixed

PR #4635, opened 2019-07-19
Filter media by types in media selection overlay

fdyckhoff:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

For Sulu 2.0: Filter results in media selection by given types set in the template config.

- add a media selection to a template
- set param "types" to one or more types (e.g. "image,video")
- when selection a media object the items to be selected should only contain images or videos

#### Why?

Missing feature from 1.x. The types parameter ist currently ignored on media selection.

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/Resources/js/containers/MediaSelectionOverlay/MediaSelectionOverlay.js
+++ b/src/Sulu/Bundle/MediaBundle/Resources/js/containers/MediaSelectionOverlay/MediaSelectionOverlay.js
@@ -54,7 +54,8 @@ class MediaSelectionOverlay extends React.Component<Props> {
     static createMediaListStore(
         collectionId: IObservableValue<?string | number>,
         excludedIds: IObservableValue<?Array<number>>,
-        locale: IObservableValue<string>
+        locale: IObservableValue<string>,
+        types: IObservableValue<?Array<string>>,
     ) {
         const options = {};
 
@@ -69,6 +70,10 @@ class MediaSelectionOverlay extends React.Component<Props> {
             'subVersion',
             'thumbnails',
         ];
+
+        if(types.length > 0) {
+            options.types = types.join(',');
+        }
 
         return new ListStore(
             MEDIA_RESOURCE_KEY,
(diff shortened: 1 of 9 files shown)


src/Sulu/Bundle/MediaBundle/Resources/js/containers/MediaSelectionOverlay/MediaSelectionOverlay.js
```diff
@@ -54,7 +54,8 @@ class MediaSelectionOverlay extends React.Component<Props> {
     static createMediaListStore(
         collectionId: IObservableValue<?string | number>,
         excludedIds: IObservableValue<?Array<number>>,
-        locale: IObservableValue<string>
+        locale: IObservableValue<string>,
+        types: IObservableValue<?Array<string>>,
```
danrot: Looks to me like the `types` variable is not an `IObservableValue`, but a just an `?Array<string>` instead

src/Sulu/Bundle/MediaBundle/Resources/js/types.js
```diff
@@ -17,3 +17,9 @@ export type DisplayOption =
     | 'leftBottom'
     | 'bottom'
     | 'rightBottom';
+
+export type MediaType =
+    | 'audio'
+    | 'document'
+    | 'image'
+    | 'video'
```
danrot: Semicolon is missing.

danrot: @alexander-schranz Is it possible to somehow extend these types? Or is our system somehow limited to these 4? Because if they are not I would not introduce a type for them, but simply use strings.

src/Sulu/Bundle/MediaBundle/Resources/js/utils/MediaSelectionHelper/validateMediaType.js
```diff
@@ -0,0 +1,8 @@
+// @flow
+
+export default function validateMediaType(name: ?string | number): boolean %checks {
+    return name === 'audio'
+        || name === 'document'
+        || name === 'image'
+        || name === 'video';
```
danrot: That's connected to the comment above. If these types can be extended, we probably shouldn't check that here, because then this code would make the application throw an exception.

src/Sulu/Bundle/MediaBundle/Resources/js/types.js
alexander-schranz: Its possible to extend them so we should not check for this types here in the frontend.

fdyckhoff:
@danrot I'm done with my changes today. Please review again :)

danrot:
Thank you for the contribution :slightly_smiling_face:

PR #4636, opened 2019-07-22
Fix nested properties within block

### LINKED ISSUE #4628: Nested properties within block
| Q | A
| --- | ---
| Bug? | yes
| New Feature? | no
| Sulu Version | 2.0
| Browser Version | -

#### Actual Behavior

Currently it's not possible to map data from a nested JSON object to a form property within a block element by using the path, e.g.:
`<property name="options/max" type="number">` 

#### Expected Behavior

It should be possible to map nested properties inside block elements to the associated JSON object. This behavior is already implemented for basic form fields but within a block element it's not working.

#### Possible Solutions

At the moment, new values are set in the handleBlockChange handler (FieldBlocks.js) in the following way:
`newValues[index][name] = value;`

By setting the new values via the json-pointer in this handler the mapping of nested properties should work:
`jsonpointer.set(newValues[index], '/' + name, value);`

markusmoosbrugger:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #4628
| Related issues/PRs | 
| License | MIT
| Documentation PR | 

#### What's in this PR?

Allow mapping between form properties within a block element and nested JSON data.

#### Why?

Previously it was not possible to map data from a nested JSON object to a form property within a block element.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
@@ -1,4 +1,5 @@
 // @flow
+import jsonpointer from 'json-pointer';
 import React, {Fragment} from 'react';
 import {toJS} from 'mobx';
 import BlockCollection from '../../components/BlockCollection';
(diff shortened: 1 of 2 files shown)


danrot:
What's definitely missing here is a test :slightly_smiling_face:

PR #4641, opened 2019-07-23
Hide the EnableUserToolbarAction for enabled users

nnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This adjusts the return type of the `getToolbarItemConfig` method of the `AbstractFormToolbarAction` to be nullable. Furthermore it adjusts the logic for rendering toolbar actions to hide toolbar-actions that return `null` from their `getToolbarItemConfig` method.

In addition, this PR renames the `AbstractToolbarAction` class of the `Form` to `AbstractFormToolbarAction` and the `AbstractToolbarAction` class of the `List` to `AbstractListToolbarAction`.

Finally, the PR hides the `EnableUserToolbarAction` for already enabled users.

#### Why?

Because a user can be enabled only once. Therefore it is not useful to display the `EnableUserToolbarAction` for users that are already enabled.


src/Sulu/Bundle/AdminBundle/Resources/js/containers/Toolbar/types.js
```diff
@@ -49,5 +49,5 @@ export type ToolbarConfig = {
 
 export interface ToolbarAction {
```
nnatter: not sure if this interface is still in the right place. it is used only in the `AbstractListToolbarAction` and the `AbstractFormToolbarAction` but nowhere inside the `Toolbar` directory

src/Sulu/Bundle/AdminBundle/Resources/js/views/Form/index.js
```diff
@@ -11,7 +11,7 @@ export default Form;
 
 export {
     toolbarActionRegistry,
```
nnatter: should we rename the `ToolbarActionRegistry` classes too?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Toolbar/types.js
danrot: I think it still makes sense, because it makes sure that the toolbar actions for forms and list implement the same methods. Otherwise the two implementation might diverge.

src/Sulu/Bundle/AdminBundle/Resources/js/views/Form/index.js
danrot: If we are already at it, I would say so... Is the situation the same as in the `ToolbarAction` case? I.e. wouldn't it break existing code, if we do that?

src/Sulu/Bundle/AdminBundle/Resources/js/views/List/registries/ToolbarActionRegistry.js
```diff
@@ -1,8 +1,8 @@
 // @flow
-import AbstractToolbarAction from '../toolbarActions/AbstractToolbarAction';
+import AbstractListToolbarAction from '../toolbarActions/AbstractListToolbarAction';
 
 class ToolbarActionRegistry {
```
danrot: If we change the name as you suggested above, we should also rename this class.

src/Sulu/Bundle/AdminBundle/Resources/js/views/index.js
```diff
@@ -1,10 +1,10 @@
 // @flow
 import Form, {
-    AbstractToolbarAction as AbstractFormToolbarAction,
+    AbstractFormToolbarAction as AbstractFormToolbarAction,
```
danrot: We can get rid of the `as` then, right?

src/Sulu/Bundle/AdminBundle/Resources/js/views/index.js
```diff
@@ -1,10 +1,10 @@
...
     toolbarActionRegistry as formToolbarActionRegistry,
 } from './Form';
 import List, {
-    AbstractToolbarAction as AbstractListToolbarAction,
+    AbstractListToolbarAction as AbstractListToolbarAction,
```
danrot: Same as above

src/Sulu/Bundle/SecurityBundle/Resources/js/views/Form/tests/toolbarActions/EnableUserToolbarAction.test.js
```diff
@@ -130,7 +127,10 @@ test('Call ResourceRequester with correct parameters when button is clicked', ()
     // $FlowFixMe
     toolbarAction.resourceFormStore.resourceStore.locale = 'de';
 
-    toolbarAction.getToolbarItemConfig().onClick();
+    const toolbarItemConfig = toolbarAction.getToolbarItemConfig();
+    expect(toolbarItemConfig).not.toBeFalsy();
+    // $FlowFixMe
```
danrot: What I usually prefer instead the expect and the `$FlowFixMe` is something like this:

```javascript
if (!toolbarItemConfig) {
    throw new Error('The ToolbarItemConfig should not undefined or null');
}
```

Still shows a nice error message when running jest and flow does not complain at all :slightly_smiling_face:

src/Sulu/Bundle/SecurityBundle/Resources/js/views/Form/tests/toolbarActions/EnableUserToolbarAction.test.js
```diff
@@ -148,13 +148,23 @@ test('Return item config with loading button during request', () => {
...
-    expect(toolbarAction.getToolbarItemConfig().loading).toBeFalsy();
+    expect(toolbarAction.getToolbarItemConfig()).toEqual(expect.objectContaining({
+        loading: false,
+    }));
 
-    toolbarAction.getToolbarItemConfig().onClick();
-    expect(toolbarAction.getToolbarItemConfig().loading).toBeTruthy();
+    const toolbarItemConfig = toolbarAction.getToolbarItemConfig();
+    expect(toolbarItemConfig).not.toBeFalsy();
+    // $FlowFixMe
```
danrot: Same as above

src/Sulu/Bundle/SecurityBundle/Resources/js/views/Form/tests/toolbarActions/EnableUserToolbarAction.test.js
```diff
@@ -148,13 +148,23 @@ test('Return item config with loading button during request', () => {
...
+    toolbarItemConfig.onClick();
+
+    expect(toolbarAction.getToolbarItemConfig()).toEqual(expect.objectContaining({
+        loading: true,
+    }));
 
     return deleteDraftPromise.then(() => {
```
danrot: Not in this PR, but this promise still has a strange name :see_no_evil:

src/Sulu/Bundle/SecurityBundle/Resources/js/views/Form/tests/toolbarActions/EnableUserToolbarAction.test.js
```diff
@@ -167,7 +177,10 @@ test('Set new enabled value to ResourceFormStore and show success-snackbar on su
     toolbarAction.resourceFormStore.resourceStore.data.id = 1234;
     toolbarAction.resourceFormStore.resourceStore.data.enabled = false;
 
-    toolbarAction.getToolbarItemConfig().onClick();
+    const toolbarItemConfig = toolbarAction.getToolbarItemConfig();
+    expect(toolbarItemConfig).not.toBeFalsy();
+    // $FlowFixMe
```
danrot: Same as above.

src/Sulu/Bundle/SecurityBundle/Resources/js/views/Form/tests/toolbarActions/EnableUserToolbarAction.test.js
```diff
@@ -186,7 +199,10 @@ test('Push error to form view on failed request', (done) => {
 
     expect(toolbarAction.form.errors).toHaveLength(0);
 
-    toolbarAction.getToolbarItemConfig().onClick();
+    const toolbarItemConfig = toolbarAction.getToolbarItemConfig();
+    expect(toolbarItemConfig).not.toBeFalsy();
+    // $FlowFixMe
```
danrot: Same as above

src/Sulu/Bundle/SecurityBundle/Resources/js/views/Form/toolbarActions/EnableUserToolbarAction.js
```diff
@@ -3,20 +3,21 @@ import {action, observable} from 'mobx';
...
 import {translate} from 'sulu-admin-bundle/utils';
-import type {ButtonItemConfig} from 'sulu-admin-bundle/containers/Toolbar/types';
 
 export default class EnableUserToolbarAction extends AbstractFormToolbarAction {
     @observable loading: boolean = false;
 
-    getToolbarItemConfig(): ButtonItemConfig {
-        const userIsEnabled = this.resourceFormStore.data.enabled;
+    getToolbarItemConfig() {
+        if (this.resourceFormStore.data.enabled) {
```
danrot: I think we also shouldn't show this button when the user is about to be created, right? Usually users created via this form are automatically enabled, and even if they weren't the button does not work.

src/Sulu/Bundle/SecurityBundle/Resources/js/views/Form/toolbarActions/EnableUserToolbarAction.js
```diff
@@ -3,20 +3,21 @@ import {action, observable} from 'mobx';
...
+            return null;
+        }
 
         return {
             type: 'button',
             icon: 'su-enter',
             onClick: this.handleEnableUserButtonClick,
-            disabled: this.resourceFormStore.loading || !this.resourceFormStore.data.id || userIsEnabled,
-            label: translate(userIsEnabled ? 'sulu_security.user_enabled' : 'sulu_security.enable_user'),
```
danrot: Looks like we can also get rid of one of the translations, right?

src/Sulu/Bundle/AdminBundle/Resources/js/views/Form/index.js
nnatter: well it will not break existing code for everyone that imported the registry from the `index.js` file of the admin bundle. it will break existing projects that import the `toolbarActionRegistry` directly. will adjust the `UPGRADE.md` just to be safe.

src/Sulu/Bundle/SecurityBundle/Resources/js/views/Form/tests/toolbarActions/EnableUserToolbarAction.test.js
nnatter: already changed this in #4642 🙂

PR #4642, opened 2019-07-23
Add toolbar-action for locking users to permission-tab of contact form

nnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

Builds upon: #4641 

#### What's in this PR?

This PR adds a toggler toolbar-action that allows to lock and unlock the current user to the Permissions tab of the Contact form.

#### Why?

Because we do not want to manually modify the database every time we need to lock a user 😉

### DIFF
--- a/src/Sulu/Bundle/SecurityBundle/Resources/js/views/Form/tests/toolbarActions/LockUserToolbarAction.test.js
+++ b/src/Sulu/Bundle/SecurityBundle/Resources/js/views/Form/tests/toolbarActions/LockUserToolbarAction.test.js
@@ -146,8 +146,9 @@ test('Call ResourceRequester with correct parameters when user is not locked and
     toolbarAction.resourceFormStore.resourceStore.locale = 'de';
 
     const toolbarItemConfig = toolbarAction.getToolbarItemConfig();
-    expect(toolbarItemConfig).not.toBeFalsy();
-    // $FlowFixMe
+    if (!toolbarItemConfig) {
+        throw new Error('The ToolbarItemConfig should not be undefined or null');
+    }
     toolbarItemConfig.onClick();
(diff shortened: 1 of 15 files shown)


src/Sulu/Bundle/SecurityBundle/Resources/translations/admin.en.json
```diff
@@ -2,6 +2,8 @@
     "sulu_security.roles": "User roles",
     "sulu_security.user_name": "Username",
     "sulu_security.enable_user": "Enable User",
+    "sulu_security.lock_user": "Lock User",
+    "sulu_security.user_locked": "User locked",
```
danrot: How about just using this translation? I think it is always confusing, if a toggler changes the text, because it is not clear if it is the current state, or the new one when you click on it. If we just use "User locked" both times it is clearer to me that the user is no locked if the toggler is deactivated.

src/Sulu/Bundle/SecurityBundle/Resources/js/views/Form/tests/toolbarActions/LockUserToolbarAction.test.js
```diff
@@ -0,0 +1,256 @@
...
+            return this.resourceStore.locale;
+        }
+
+        get loading() {
+            return this.resourceStore.loading;
+        }
+    },
+}));
+
+jest.mock('sulu-admin-bundle/services/Router', () => jest.fn(function() {}));
```
danrot: Just a small note, you don't have to pass an empty function, just `jest.fn()` would do just fine :slightly_smiling_face:

PR #4646, opened 2019-07-26
Fix smart-content sorting

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT

#### What's in this PR?

Fix for the sorting function of the smart-content type. The `QueryBuilder` for smart-content anticipated a `sortBy` array, but from the website-view only a `string` was given. Thus the orderBy defaulted to `sulu:order`.

#### Why?

Previously the sorting function was only working in the admin view but not on the website.


HadGS:
Same issue, different solving approach is there since March... No merge yet... https://github.com/sulu/sulu/pull/4470

danrot:
@HadGS Sorry, that seemd to slip through... But you also didn't fix the build. And I like this approach better, because it fixes the cause of the problem (that the same thing was returned in different data types). So I am going to merge this one instead of yours. Thanks for your effort anyway!

PR #4651, opened 2019-07-29
Upgrade KernelTestCase class to Symfony KernelTestCase

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes 
| License | MIT

#### What's in this PR?

Replace custom KernelTestCase with Symfony FrameworkBundle Testcase and split Sulu Specific TestCase function into traits to allow to use them for example in KernelTestCase, WebTestCase or PantherTestCase.

#### Why?

The FrameworkBundle::KernelTestCase add some features e.g. allow accessing of private service in tests.

#### BC Breaks/Deprecations

See UPGRADE.md

#### To Do


src/Sulu/Bundle/TestBundle/Testing/SuluTestCase.php
```diff
@@ -11,311 +11,14 @@
...
-        return $initializerDump;
-    }
-
-    /**
-     * Create a webspace node with the given locales.
-     *
-     * @param string $name
-     * @param string[] $locales
-     */
-    protected function createHomeDocument($name, array $locales)
```
alexander-schranz: could not found any usage for this should we still implement it?

src/Sulu/Bundle/TestBundle/Testing/SuluTestCase.php
```diff
@@ -11,311 +11,14 @@
...
 
-abstract class SuluTestCase extends KernelTestCase
+abstract class SuluTestCase extends WebTestCase
 {
-    private static $workspaceInitialized = false;
-
-    /**
-     * @var PHPCRImporter
-     */
-    protected $importer;
```
alexander-schranz: Is used only in the NodeControllerTest so I did only implement it there

src/Sulu/Bundle/AdminBundle/Tests/Functional/Metadata/FormMetadata/FormMetadataProviderTest.php
```diff
@@ -15,9 +15,9 @@
 use Sulu\Bundle\AdminBundle\Metadata\FormMetadata\FormMetadata;
 use Sulu\Bundle\AdminBundle\Metadata\FormMetadata\FormMetadataProvider;
 use Sulu\Bundle\AdminBundle\Metadata\FormMetadata\TypedFormMetadata;
-use Sulu\Bundle\TestBundle\Testing\KernelTestCase;
+use Sulu\Bundle\TestBundle\Testing\SuluTestCase;
 
-class FormMetadataProviderTest extends KernelTestCase
+class FormMetadataProviderTest extends SuluTestCase
```
danrot: Why do we need the `SuluTestCase` here? Shouldn't the simpler `KernelTestCase` be sufficient? What is the difference between these two in general?

src/Sulu/Bundle/MediaBundle/Tests/Functional/Controller/CollectionControllerTest.php
```diff
@@ -1067,6 +1066,8 @@ public function testPutWithChildCollection()
             5
         );
 
+        $client = $this->createAuthenticatedClient();
```
danrot: Why do we have to move it here? I think I've understood you differently the last time.

src/Sulu/Bundle/MediaBundle/Tests/Functional/Controller/MediaControllerTest.php
```diff
@@ -1332,12 +1332,13 @@ public function testTagRemove()
     {
         $media = $this->createMedia('photo');
 
-        $client = $this->createAuthenticatedClient();
-
         // Check Tag Remove
-        $this->getEntityManager()->remove($this->tag1);
+        $this->getEntityManager()->remove(
+            $this->getEntityManager()->getReference(Tag::class, $this->tag1->getId())
```
danrot: Why is that necessary?

src/Sulu/Bundle/PageBundle/Tests/Functional/Controller/NodeControllerTest.php
```diff
@@ -28,38 +29,54 @@
...
-        $this->em = $this->getEntityManager();
-        $this->session = $this->getContainer()->get('sulu_document_manager.default_session');
-        $this->liveSession = $this->getContainer()->get('sulu_document_manager.live_session');
-        $this->documentManager = $this->getContainer()->get('sulu_document_manager.document_manager');
-
         $this->initOrm();
         $this->initPhpcr();
     }
 
+    protected static function bootKernel(array $options = [])
```
danrot: Why is that in combination with all the static values only necessary in this test? :thinking:

src/Sulu/Bundle/PreviewBundle/Tests/Functional/Preview/Renderer/PreviewRendererTest.php
```diff
@@ -12,9 +12,9 @@
 namespace Sulu\Bundle\PreviewBundle\Tests\Functional\Preview\Renderer;
 
 use Sulu\Bundle\PreviewBundle\Preview\Renderer\PreviewRenderer;
-use Sulu\Bundle\TestBundle\Testing\KernelTestCase;
+use Sulu\Bundle\TestBundle\Testing\SuluTestCase;
 
-class PreviewRendererTest extends KernelTestCase
+class PreviewRendererTest extends SuluTestCase
```
danrot: Why do we need the other test case here?

src/Sulu/Bundle/ResourceBundle/Entity/FilterRepository.php
```diff
@@ -47,12 +47,13 @@ public function findByUserAndContextAndLocale($locale, $context, $userId)
             $qb = $this->getFilterQuery($locale);
             $qb->leftJoin('filter.user', 'user')
             ->andWhere('filter.context = :context')
-            ->orWhere('(user.id = :userId')
-            ->orWhere('filter.private = false)')
+            ->andWhere('(user.id = :userId OR filter.private = false)')
```
danrot: Maybe we should have deleted that bundle before you struggled with it because of the tests :sweat_smile:

src/Sulu/Bundle/SnippetBundle/Tests/Functional/Controller/SnippetControllerTest.php
```diff
@@ -50,18 +50,19 @@ class SnippetControllerTest extends SuluTestCase
     public function setUp()
     {
         parent::setUp();
-        $this->documentManager = $this->getContainer()->get('sulu_document_manager.document_manager');
-        $this->phpcrSession = $this->getContainer()->get('sulu_test.doctrine_phpcr')->getConnection();
         $this->initPhpcr();
-        $this->loadFixtures();
         $this->client = $this->createAuthenticatedClient();
```
danrot: Why can we do it like this in this Test, but have to put all of this stuff on static variables in the `NodeControllerTest`?

src/Sulu/Bundle/SnippetBundle/Tests/Functional/Controller/SnippetControllerTest.php
```diff
@@ -50,18 +50,19 @@ class SnippetControllerTest extends SuluTestCase
...
+        $this->documentManager = $this->getContainer()->get('sulu_document_manager.document_manager');
+        $this->loadFixtures();
     }
 
     /**
      * @dataProvider provideGet
      */
     public function testGet($locale, $expected)
     {
+        $this->documentManager->clear();
```
danrot: Why do we need this line only for this test?

src/Sulu/Bundle/TestBundle/Testing/AssertHttpStatusCodeTrait.php
```diff
@@ -0,0 +1,55 @@
...
+     */
+    protected static function assertHttpStatusCode(int $code, Response $response, int $debugLength = 50): void
+    {
+        $httpCode = $response->getStatusCode();
+
+        $message = '';
+        if ($code !== $httpCode) {
+            $message = $response->getContent();
+
+            if ('null' !== ($json = json_encode(json_decode($message, true), JSON_PRETTY_PRINT))
```
danrot: Are you sure that's doing the same as the one before? Don't you want to move it over the exact same way it was before?

src/Sulu/Bundle/TestBundle/Testing/AssertHttpStatusCodeTrait.php
```diff
@@ -0,0 +1,55 @@
...
+                && $json) {
+                $message = explode(PHP_EOL, $json);
+            } else {
+                $message = explode(PHP_EOL, $message);
+            }
+
+            $message = implode(PHP_EOL, \array_slice($message, 0, $debugLength));
```
danrot: If I personally would change anything in this function, it would be to remove this `debugLength`... If something is going wrong on CI I am always forced to do a rerun, because I might not have enough information.

src/Sulu/Bundle/TestBundle/Testing/CreateClientTrait.php
```diff
@@ -0,0 +1,55 @@
...
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\TestBundle\Testing;
+
+use Sulu\Component\HttpKernel\SuluKernel;
+use Symfony\Bundle\FrameworkBundle\KernelBrowser;
+use Symfony\Component\BrowserKit\Client;
+
+trait CreateClientTrait
```
danrot: Why don't we need all the environment information anymore in these functions?

src/Sulu/Bundle/TestBundle/Testing/KernelTrait.php
```diff
@@ -0,0 +1,148 @@
...
+            $debug = $_ENV['APP_DEBUG'];
+        } elseif (isset($_SERVER['APP_DEBUG'])) {
+            $debug = $_SERVER['APP_DEBUG'];
+        } else {
+            $debug = true;
+        }
+
+        if (isset($options['sulu.context'])) {
+            $suluContext = $options['sulu.context'];
+        } elseif (isset($options['sulu_context'])) {
```
danrot: Should we really introduce new deprecations before the 2.0 release? Wouldn't it be a better idea to just not handle this case?

src/Sulu/Bundle/TestBundle/Testing/KernelTrait.php
```diff
@@ -0,0 +1,148 @@
...
+     * @var ContainerInterface
+     */
+    protected static $container;
+
+    protected static function getSuluContext(): string
+    {
+        return SuluKernel::CONTEXT_ADMIN;
+    }
+
+    protected static function createKernel(array $options = []): SuluKernel
```
danrot: Shouldn't we return a `SuluTestKernel` here, as it was done before?

src/Sulu/Bundle/TestBundle/Testing/KernelTrait.php
```diff
@@ -0,0 +1,148 @@
...
+            $suluContext = $options['sulu_context'];
+        } else {
+            $suluContext = static::getSuluContext();
+        }
+
+        return new static::$class($env, (bool) $debug, $suluContext);
+    }
+
+    /**
+     * @deprecated The "getContainer" function is deprecated use self::$container variable instead.
```
danrot: I think if this function should not be used, we should remove it right now.

src/Sulu/Bundle/TestBundle/Testing/KernelTrait.php
```diff
@@ -0,0 +1,148 @@
...
+
+    protected static function ensureKernelShutdown()
+    {
+        parent::ensureKernelShutdown();
+        // In symfony <4.1 container is not set to null on ensureKernelShutdown this keep compatibility
+        static::$container = null;
+    }
+
+    /**
+     * @deprecated The "getKernel" function is deprecated use self::$kernel variable or createKernel instead
```
danrot: Same as above

src/Sulu/Bundle/TestBundle/Testing/SuluTestCase.php
```diff
@@ -11,311 +11,14 @@
...
-    /**
-     * {@inheritdoc}
-     */
-    public static function setUpBeforeClass()
-    {
-        // enables garbage collector because symfony/phpunit-bridge disables it. see:
-        // see: https://github.com/symfony/symfony/pull/13398/files#diff-81bfee6017752d99d3119f4ddb1a09edR1
-        // see: https://github.com/symfony/symfony/pull/13398 (feature list)
-        if (!gc_enabled()) {
-            gc_enable();
```
danrot: Is that not required anymore? Or was I just missing it during the review?

src/Sulu/Bundle/TestBundle/Testing/SuluTestCase.php
danrot: Sounds reasonable to me. :+1:

src/Sulu/Bundle/TestBundle/Testing/SuluTestCase.php
```diff
@@ -11,311 +11,14 @@
...
-
-        // close the jackalope connections - can be removed when
-        // https://github.com/sulu/sulu/pull/2125 is merged.
-        // this has a negligble impact on memory usage in anycase.
-        foreach ($this->getContainer()->get('doctrine_phpcr')->getConnections() as $connection) {
-            try {
-                $connection->logout();
-            } catch (\Exception $e) {
-            }
-        }
```
danrot: I don't remember seeing similar code to this. Don't we have to do that anymore? Or have I missed it?

src/Sulu/Bundle/AdminBundle/Tests/Functional/Metadata/FormMetadata/FormMetadataProviderTest.php
alexander-schranz: you are right KernelTestCase is here enough

src/Sulu/Bundle/MediaBundle/Tests/Functional/Controller/CollectionControllerTest.php
alexander-schranz: `createAuthenticedClient` seems to destroy the old kernel and create a new one and createCollection uses the entityManager which was created before and seems to access the `kernel` service in some listener and as the kernel was destroyed before it end ups in: synthetic service `kernel` not found errror. To avoid this I create the collection before the client so the old kernel exist before. Basically its just a bad idea to bind services inside the setup method instead for symfony 4.3 it should `self::$container->get(..)` be used.

src/Sulu/Bundle/SnippetBundle/Tests/Functional/Controller/SnippetControllerTest.php
alexander-schranz: It seems the only test which access the same document in different languages but will move it to the loadFixtures method so if new test are added not another side effects happens.

src/Sulu/Bundle/TestBundle/Testing/AssertHttpStatusCodeTrait.php
alexander-schranz: Its an improved version we used in projects which handles also HTML correct with the debug length.

src/Sulu/Bundle/TestBundle/Testing/CreateClientTrait.php
alexander-schranz: they matches the default environment variables so not needed to have them also here

src/Sulu/Bundle/TestBundle/Testing/AssertHttpStatusCodeTrait.php
alexander-schranz: Here I'm not with you. I had the problem the other way around I was always struggleing when the debugLength was not keep in mention for HTML responses and did spam the whole CI/Terminal with unnecessary outputs.

src/Sulu/Bundle/TestBundle/Testing/KernelTrait.php
alexander-schranz: No in projects you don't use a custom Test Kernel you use the Project Kernel.

src/Sulu/Bundle/TestBundle/Testing/KernelTrait.php
alexander-schranz: It should not be used by others but we need it to test symfony 3.4 at current state.

src/Sulu/Bundle/TestBundle/Testing/SuluTestCase.php
alexander-schranz: symfoyn does not longer disable gc in its phpunit bridge

src/Sulu/Bundle/TestBundle/Testing/SuluTestCase.php
alexander-schranz: the linked pull request has been merged

src/Sulu/Bundle/TestBundle/Testing/KernelTrait.php
alexander-schranz: I renamed it some time ago to be consistence to the `RemoveForeignContextServicesPass::SULU_CONTEXT_TAG`. Should I just remove the deprecation or do you want to keep the old one?

alexander-schranz:
@danrot The traits are needed if you want e.g. create a `PantherTestCase` or other similiar once as you can't use there the SuluTestCase because you need to extend from PanterTestCase and atleast use our KernelTrait.

The traits are following the single responsability and should make more clear which additional functions we provide:

 - **`KernelTrait`**: Create a sulu compatible Kernel with correct Context (overwrite default symfony behaviour which does not work with sulu kernels)
 - **`ContainerCompatibilityTrait`**: provides the getContainer method which is used in mostly any test by us and implement a BC Layer between symfony 3.4 and 4.1 to get the correct container for test.
 - **`AssertHttpStatusCodeTrait`**: Function for checking HttpCode
 - **`CreateClientTrait`**: Provides different clients (authenticated and website clients)
 - **`PurgeDatabaseTrait`**: Helper function to purge the database
 - **`PhpCrInitTrait`**: Helper function to create phpcr schema 
 - **`TestUserTrait`**: Provide the getTestUser and getTestUserId functions which are used in some tests

src/Sulu/Bundle/TestBundle/Testing/KernelTrait.php
alexander-schranz: removed

src/Sulu/Bundle/MediaBundle/Tests/Functional/Controller/CollectionControllerTest.php
danrot: So just that I get this right: We should now always access the container using the static variable, as described in [the Symfony documentation](https://symfony.com/doc/current/testing.html#accessing-the-container), right? The new Symfony documentation also gets the [client via a static method call](https://symfony.com/doc/current/testing.html#your-first-functional-test), how is that connected? Should we maybe also do that, and then it would not matter where the call took place?

src/Sulu/Bundle/PageBundle/Tests/Functional/Controller/NodeControllerTest.php
danrot: I am missing an answer here :slightly_smiling_face:

src/Sulu/Bundle/TestBundle/Testing/KernelTrait.php
danrot: I've seen you removed it, that's fine :+1:

src/Sulu/Bundle/SnippetBundle/Tests/Functional/Controller/SnippetControllerTest.php
danrot: Also missing an explanation here, probably connected to the comment above.

src/Sulu/Bundle/TestBundle/Testing/ContainerCompatibilityTrait.php
```diff
@@ -0,0 +1,75 @@
...
+namespace Sulu\Bundle\TestBundle\Testing;
+
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+trait ContainerCompatibilityTrait
+{
+    /**
+     * @internal The "getContainer" function is deprecated use self::$container variable instead.
+     *
+     * This function exist to keep a compatibility between symfony versions.
```
danrot: So as soon as Symfony 3.4 is not supported anymore, we can delete this trait? Then I would mention that here.

And why is that the case? Does that mean we always have to use this function now?

Btw, what's the state of Symfony 5 support? Can we support 3.4 at the same time? Wasn't that the condition to keep the 3.4 support?

danrot:
Locally a test in the MediaBundle is failing for me:

```
1) Sulu\Bundle\MediaBundle\Tests\Functional\Controller\MediaControllerTest::test404ResponseHeader
Failed asserting that an object is empty.

/home/daniel/Development/sulu/sulu-minimal/vendor/sulu/sulu/src/Sulu/Bundle/MediaBundle/Tests/Functional/Controller/MediaControllerTest.php:385
```

Any ideas?

src/Sulu/Bundle/MediaBundle/Tests/Functional/Controller/CollectionControllerTest.php
alexander-schranz: Yeah we can use the static variable and functions like its documented in the symfony documentation now. The difference between `getContainer` and `self::$container` is that `getContainer` would boot the kernel if not yet booted.

src/Sulu/Bundle/PageBundle/Tests/Functional/Controller/NodeControllerTest.php
alexander-schranz: Seems one of the only tests binding services to class variables which access the `kernel` service and createClient (which destroy and create a new kernel) is called after in its tests.

src/Sulu/Bundle/SnippetBundle/Tests/Functional/Controller/SnippetControllerTest.php
alexander-schranz: because `$this->client = $this->createAuthenticatedClient();` is called here before binding the services to the class variables so the service has the new kernel in them and not a kernel which get destroyed.

src/Sulu/Bundle/TestBundle/Testing/ContainerCompatibilityTrait.php
alexander-schranz: 5.0 will definitly need more work as this. As there are some other changes in the browser kit and http kernel.

alexander-schranz:
@danrot tests are now fixed and I tested the MediaBundle and it works for me locally.

PR #4652, opened 2019-07-30
Detach file and file version in delete of media manager

markusmoosbrugger:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | 
| Related issues/PRs | 
| License | MIT
| Documentation PR |

#### What's in this PR?

This PR detachs the file and file version in the delete method of the media manager. 

#### Why?

Currently an error occurs in the form bundle if the delete method of the media manager is called, followed by another remove call to the entity manager with another entity. The error message says that _a new entity was found through the relationship Sulu\Bundle\MediaBundle\Entity\File#media that was not configured to cascade persist operations for entity_.

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/Media/Manager/MediaManager.php
+++ b/src/Sulu/Bundle/MediaBundle/Media/Manager/MediaManager.php
@@ -719,7 +719,9 @@ class MediaManager implements MediaManagerInterface
                     // this will trigger massive-search deindex
                     $this->em->remove($fileVersionMeta);
                 }
+                $this->em->detach($fileVersion);
             }
+            $this->em->detach($file);
         }
 
         $this->em->remove($mediaEntity);
--- a/src/Sulu/Bundle/MediaBundle/Tests/Unit/Media/Manager/MediaManagerTest.php
+++ b/src/Sulu/Bundle/MediaBundle/Tests/Unit/Media/Manager/MediaManagerTest.php
@@ -278,7 +278,8 @@ class MediaManagerTest extends TestCase
         )->shouldBeCalled();
 
         $this->storage->remove(['segment' => '01', 'fileName' => 'test.jpg'])->shouldBeCalled();
-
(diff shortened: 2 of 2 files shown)


src/Sulu/Bundle/MediaBundle/Media/Manager/MediaManager.php
```diff
@@ -719,7 +719,9 @@ public function delete($id, $checkSecurity = false)
                     // this will trigger massive-search deindex
                     $this->em->remove($fileVersionMeta);
                 }
+                $this->em->detach($fileVersion);
```
danrot: Isn't that something that should always happen, no matter how the medium is deleted? Because I think that is something that should go into the Doctrine configuration (the `.orm.xml` files).

src/Sulu/Bundle/MediaBundle/Media/Manager/MediaManager.php
```diff
@@ -719,7 +719,9 @@ public function delete($id, $checkSecurity = false)
...
             }
+            $this->em->detach($file);
```
danrot: Same as above

danrot:
Resp. why isn't there a delete cascade from the media to the file to the fileversion? Then this code in general would be a lot easier to handle, wouldn't it?

alexander-schranz:
@danrot a delete cascades are evil because they are not on the database level and doctrine does send a delete query for every entity instead of letting the database handle the `on-delete="cascade"`. Which we could test is if `<cascade-detach />` work in this case correctly.

danrot:
You can also register that in a way that the database and not doctrine is handling this. We have e.g. done this in the [Media entity for its relation to the collection](https://github.com/sulu/sulu/blob/8e8f74d/src/Sulu/Bundle/MediaBundle/Resources/config/doctrine/Media.orm.xml#L25).

PR #4661, opened 2019-08-01
Hide UI elements with missing permission

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR hides UI elements, if the user has not been granted these permissions.

#### Why?

Because it is a better UX, than receiving an error on every click.

#### To Do

- [x] Avoid infinite loaders when 403 are returned

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/ViewRenderer/ViewRenderer.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/ViewRenderer/ViewRenderer.js
@@ -2,7 +2,6 @@
 import React from 'react';
 import type {Element} from 'react';
 import {observer} from 'mobx-react';
-import {sidebarStore} from '../Sidebar';
 import Router, {getViewKeyFromRoute} from '../../services/Router';
 import type {Route} from '../../services/Router';
 import viewRegistry from './registries/ViewRegistry';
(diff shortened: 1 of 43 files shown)


danrot:
- Extract `getKey` from `ViewRenderer`
- Add `updateRouteHook` to `withToolbar` and call `toolbarDisposer` when view changes (check using previously extracted `getKey`)
- Check if the sidebar should be reset the same way (instead of putting that logic into the `ViewRenderer`)

src/Sulu/Bundle/AdminBundle/Resources/js/containers/ViewRenderer/ViewRenderer.js
```diff
@@ -35,31 +28,6 @@ class ViewRenderer extends React.Component<Props> {
         }, UPDATE_ROUTE_HOOK_PRIORITY);
     }
 
-    clearSidebarConfig() {
-        if (!this.hasSidebar()) {
-            sidebarStore.clearConfig();
```
danrot: @wachterjohannes Any idea why this was implemented so complicated? I moved this call now to the `componentWillUnmount` of the `withSidebar` component, and I don't see a real difference in its behavior.

PR #4671, opened 2019-08-08
Support multiple list metadata loaders

markusmoosbrugger:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | 
| Related issues/PRs | https://github.com/sulu/SuluFormBundle/pull/197
| License | MIT
| Documentation PR | 

#### What's in this PR?

This PR adds the support for multiple list metadata loaders. The different `ListMetadataLoader` classes can be tagged in the associated `services.xml` files as **sulu_admin.list_metadata_loader** in order to get injected automatically in the `ListMetadataProvider`. The different `ListMetadataLoader` classes must implement the interface `ListMetadataLoaderInterface`.

The `ListRouteBuilder` has now a method `addRouterAttributesToListMetadata` that allows to add metadata options to the metadata request.

This PR also adds the abstract class `AbstractMetadata` which defines a property **cacheable** that allows to define whether this metadata instance should be cached or not. When cacheable is set to false, the `AdminController` adds the header **Cache-Control: no-store** to the response.
 
By checking this cache-control header the JS class `MetadataStore` can then decide if the metadata information should be cached or not.

#### Why?

Enable to use a custom list metadata loader provided from the form bundle (see https://github.com/sulu/SuluFormBundle/pull/197). The functionality can also be useful for other bundles that want to provide their own list metadata loader.

#### Example Usage

Add a custom list loader that implements the `ListMetadataLoaderInterface`. Tag this loader in the `services.xml` file as **sulu_admin.list_metadata_loader**.

~~~xml
<service id="sulu_form.metadata.dynamic_list_metadata_loader" class="Sulu\Bundle\FormBundle\Metadata\DynamicListMetadataLoader">
            <argument type="service" id="sulu_form.dynamic.form_field_type_pool"/>
            <argument type="service" id="translator"/>
            <argument type="service" id="sulu_form.manager.form"/>
            <tag name="sulu_admin.list_metadata_loader"/>
</service>
~~~

#### BC Breaks/Deprecations

#### To Do

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Controller/AdminController.php
+++ b/src/Sulu/Bundle/AdminBundle/Controller/AdminController.php
@@ -275,12 +275,18 @@ class AdminController
         $user = $this->tokenStorage->getToken()->getUser();
 
         $metadataOptions = $request->query->all();
-
+        $metadata = $this->metadataProviderRegistry->getMetadataProvider($type)->getMetadata($key, $user->getLocale(), $metadataOptions);
         $view = View::create(
-            $this->metadataProviderRegistry->getMetadataProvider($type)->getMetadata($key, $user->getLocale(), $metadataOptions)
+            $metadata
         );
         $view->setFormat('json');
 
-        return $this->viewHandler->handle($view);
+        $response = $this->viewHandler->handle($view);
+
+        if (!$metadata->isCacheable()) {
+            $response->headers->addCacheControlDirective('no-store', !$metadata->isCacheable());
+        }
+
+        return $response;
     }
 }
--- a/src/Sulu/Bundle/AdminBundle/Controller/AdminController.php
+++ b/src/Sulu/Bundle/AdminBundle/Controller/AdminController.php
@@ -278,9 +278,7 @@ class AdminController
         $metadata = $this->metadataProviderRegistry->getMetadataProvider($type)
             ->getMetadata($key, $user->getLocale(), $metadataOptions);
 
-        $view = View::create(
-            $metadata
-        );
+        $view = View::create($metadata);
         $view->setFormat('json');
 
         $response = $this->viewHandler->handle($view);
--- a/src/Sulu/Bundle/AdminBundle/Controller/AdminController.php
+++ b/src/Sulu/Bundle/AdminBundle/Controller/AdminController.php
@@ -224,12 +224,6 @@ class AdminController
 
         $config = [
             'sulu_admin' => [
-                'endpoints' => [
-                    'metadata' => $this->urlGenerator->generate(
-                        'sulu_admin.metadata',
-                        ['type' => ':type', 'key' => ':key']
-                    ),
-                ],
                 'fieldTypeOptions' => $this->fieldTypeOptionRegistry->toArray(),
                 'internalLinkTypes' => $this->linkProviderPool->getConfiguration(),
                 'navigation' => $this->navigationRegistry->getNavigation()->getChildrenAsArray(),
@@ -276,12 +270,17 @@ class AdminController
         return new JsonResponse($translations);
     }
 
-    public function metadataAction(string $type, string $key): Response
+    public function metadataAction(string $type, string $key, Request $request): Response
     {
         $user = $this->tokenStorage->getToken()->getUser();
 
+        $metadataOptions = [];
+        if ($request->query->get('metadataOptions')) {
+            $metadataOptions = $request->query->get('metadataOptions');
+        }
+
         $view = View::create(
-            $this->metadataProviderRegistry->getMetadataProvider($type)->getMetadata($key, $user->getLocale())
+            $this->metadataProviderRegistry->getMetadataProvider($type)->getMetadata($key, $user->getLocale(), $metadataOptions)
         );
         $view->setFormat('json');
 
--- a/src/Sulu/Bundle/AdminBundle/Controller/AdminController.php
+++ b/src/Sulu/Bundle/AdminBundle/Controller/AdminController.php
@@ -274,10 +274,7 @@ class AdminController
     {
         $user = $this->tokenStorage->getToken()->getUser();
 
-        $metadataOptions = [];
-        if ($request->query->get('metadataOptions')) {
-            $metadataOptions = $request->query->get('metadataOptions');
-        }
+        $metadataOptions = $request->query->all();
 
         $view = View::create(
             $this->metadataProviderRegistry->getMetadataProvider($type)->getMetadata($key, $user->getLocale(), $metadataOptions)
--- a/src/Sulu/Bundle/AdminBundle/Controller/AdminController.php
+++ b/src/Sulu/Bundle/AdminBundle/Controller/AdminController.php
@@ -275,7 +275,9 @@ class AdminController
         $user = $this->tokenStorage->getToken()->getUser();
 
         $metadataOptions = $request->query->all();
-        $metadata = $this->metadataProviderRegistry->getMetadataProvider($type)->getMetadata($key, $user->getLocale(), $metadataOptions);
+        $metadata = $this->metadataProviderRegistry->getMetadataProvider($type)
+            ->getMetadata($key, $user->getLocale(), $metadataOptions);
+
         $view = View::create(
             $metadata
         );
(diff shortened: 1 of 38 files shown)


src/Sulu/Bundle/AdminBundle/Controller/AdminController.php
```diff
@@ -276,15 +270,23 @@ public function translationsAction(Request $request): Response
...
     }
 
-    public function metadataAction(string $type, string $key): Response
+    public function metadataAction(string $type, string $key, Request $request): Response
     {
         $user = $this->tokenStorage->getToken()->getUser();
 
+        $metadataOptions = $request->query->all();
+        $metadata = $this->metadataProviderRegistry->getMetadataProvider($type)->getMetadata($key, $user->getLocale(), $metadataOptions);
         $view = View::create(
```
danrot: Write that into a single line, now that it easily fits.

src/Sulu/Bundle/AdminBundle/Controller/AdminController.php
```diff
@@ -276,15 +270,23 @@ public function translationsAction(Request $request): Response
...
-            $this->metadataProviderRegistry->getMetadataProvider($type)->getMetadata($key, $user->getLocale())
+            $metadata
         );
         $view->setFormat('json');
 
-        return $this->viewHandler->handle($view);
+        $response = $this->viewHandler->handle($view);
+
+        if (!$metadata->isCacheable()) {
+            $response->headers->addCacheControlDirective('no-store', !$metadata->isCacheable());
```
danrot: Instead of reevaluating `!$metadata->isCacheable()` we could just pass true, since that will result in the same value always, right?

src/Sulu/Bundle/AdminBundle/Metadata/MetadataProviderInterface.php
```diff
@@ -13,5 +13,5 @@
 
 interface MetadataProviderInterface
 {
-    public function getMetadata(string $key, string $locale);
+    public function getMetadata(string $key, string $locale, array $metadataOptions);
```
danrot: Now that the return value of this function must at least implement the `isCacheable` method, shouldn't we introduce a `MetadataInterface`, which is being returned here? Then accessing this method in the `metadataAction` would also be typesafe.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/tests/List.test.js
```diff
@@ -15,69 +15,83 @@ import StringFieldTransformer from '../fieldTransformers/StringFieldTransformer'
...
+        observableOptions = {},
+        options = {},
+        metadataOptions = {}
+    ) {
+        this.resourceKey = resourceKey;
+        this.listKey = listKey;
+        this.options = options;
+        this.userSettingsKey = userSettingsKey;
+        this.observableOptions = observableOptions;
+        this.options = options;
```
danrot: We don't need that twice :slightly_smiling_face:

src/Sulu/Bundle/AdminBundle/Resources/js/stores/MetadataStore/MetadataStore.js
```diff
@@ -1,24 +1,62 @@
...
+            type: type,
+            key: key,
+            ...metadataOptions,
+        };
 
-    loadMetadata(type: string, key: string): Promise<Object> {
         if (!this.metadataPromises[type]) {
             this.metadataPromises[type] = {};
         }
+        const keyWithOptions = buildQueryString({
```
danrot: I think that should be something like `key + buildQueryString(metadataOptions)`. Otherwise we are probably running into problems if `metadataOptions` contains an option with the key `key`.

src/Sulu/Bundle/AdminBundle/Resources/js/stores/MetadataStore/MetadataStore.js
```diff
@@ -1,24 +1,62 @@
...
+
+        if (!this.metadataPromises[type][keyWithOptions]) {
+            const url = symfonyRouting.generate('sulu_admin.metadata', parameters);
+            const response = fetch(url, defaultOptions).then((response) => {
+                if (!response.ok) {
+                    this.metadataPromises[type][keyWithOptions] = undefined;
+                    return Promise.reject(response);
+                }
+
+                if (response.status === 204) {
```
danrot: When can that happen?

src/Sulu/Bundle/AdminBundle/Resources/js/stores/MetadataStore/MetadataStore.js
```diff
@@ -1,24 +1,62 @@
...
+                    return Promise.resolve({});
+                }
+
+                const cacheControl = response.headers.get('cache-control');
+                if (cacheControl && cacheControl.includes('no-store')) {
+                    this.metadataPromises[type][keyWithOptions] = undefined;
+                }
+
+                return response.json()
+                    .then((data) => {
```
danrot: Does this `then` do something? To me it looks like this would work exactly the same if we just remove it.

src/Sulu/Bundle/AdminBundle/Resources/js/stores/MetadataStore/MetadataStore.js
```diff
@@ -1,24 +1,62 @@
...
+            });
 
-        if (!this.metadataPromises[type][key]) {
-            this.metadataPromises[type][key] = Requester.get(
-                this.endpoint
-                    .replace(':type', type)
-                    .replace(':key', key)
-            );
+            this.metadataPromises[type][keyWithOptions] = response;
+            return response;
```
danrot: This line could just be removed, then the next `return` handles this the exact same way.

src/Sulu/Bundle/AdminBundle/Tests/Unit/Controller/AdminControllerTest.php
```diff
@@ -295,14 +295,14 @@ public function testMetadataAction()
         $this->user->getLocale()->willReturn('en');
 
         $metadataProvider = $this->prophesize(MetadataProviderInterface::class);
-        $metadataProvider->getMetadata('pages', 'en')->willReturn($form);
+        $metadataProvider->getMetadata('pages', 'en', [])->willReturn($form);
```
danrot: Shouldn't there be also a test where some data is passed instead of just an empty array?

src/Sulu/Bundle/AdminBundle/Resources/js/views/List/List.js
```diff
@@ -106,12 +107,18 @@ class List extends React.Component<Props> {
...
+        );
+
         this.listStore = new ListStore(
             resourceKey,
             listKey,
             userSettingsKey,
             observableOptions,
-            listStoreOptions
+            listStoreOptions,
+            metadataOptions
```
danrot: Is it possible that the passing of these options is not tested? Would be important IMO, because these options are not used in the core itself. That might result in breaking change for the form bundle without us realizing it...

src/Sulu/Bundle/AdminBundle/Controller/AdminController.php
danrot: That comment seemed to slipped through :wink:

src/Sulu/Bundle/AdminBundle/Resources/js/stores/MetadataStore/MetadataStore.js
danrot: That didn't work out, because it is not typesafe.

src/Sulu/Bundle/AdminBundle/Resources/js/views/List/tests/List.test.js
```diff
@@ -1106,6 +1107,35 @@ test('Should pass router attributes array from router to the ListStore', () => {
...
+            title: 'Sulu is awesome',
+        },
+        route: {
+            options: {
+                adapters: ['table'],
+                apiOptions: {},
+                listKey: 'test',
+                locales: ['en', 'de'],
+                resourceKey: 'test',
+                routerAttributesToListMetadata: ['locale', 'title', 'id'],
```
danrot: Maybe leave one out here, so that we are sure that not just all properties are copied.

PR #4676, opened 2019-08-14
Splitted Home and Page indexes

### LINKED ISSUE #4663: Sulu Massive Search config not working with ES 6
| Q | A
| --- | ---
| Bug? | yes
| New Feature? | no
| Sulu Version | 2.0.0-RC1

#### Actual Behavior

`bin/console massive:search:reindex` is resulting in 
`{"error":{"root_cause":[{"type":"illegal_argument_exception","reason":"Reje  
  cting mapping update to [test_sulu_massive_page-nl-i18n] a  
  s the final mapping would have more than 1 type: [Sulu_Bundle_PageBundle_Do  
  cument_HomeDocument, Sulu_Bundle_SnippetBundle_Document_SnippetDocument]"}]  
  ,"type":"illegal_argument_exception","reason":"Rejecting mapping update to   
  [test_sulu_massive_page-nl-i18n] as the final mapping woul  
  d have more than 1 type: [Sulu_Bundle_PageBundle_Document_HomeDocument, Sul  
  u_Bundle_SnippetBundle_Document_SnippetDocument]"},"status":400} `

#### Expected Behavior

Working reindex

#### Steps to Reproduce

* Elasticsearch 6.7
* Configure massive search to use Elasticsearch adapter
* Adding some pages, including homepage
* Run `bin/console massive:search:reindex`

#### Possible Solutions

-

kleinkoerkamp:
Splitted indexes because Elasticsearch 6 does not support 2 types in 1 index

| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #4663
| License | MIT

#### What's in this PR?

Splitted HomeDocument and PageDocument in 2 indexes. From Elasticsearch 6 it is not supported anymore to have 2 types in 1 index.


src/Sulu/Bundle/PageBundle/DependencyInjection/SuluPageExtension.php
```diff
@@ -215,7 +215,7 @@ public function prepend(ContainerBuilder $container)
                     'search' => [
                         'mapping' => [
                             PageDocument::class => ['index' => 'page', 'decorate_index' => true],
-                            HomeDocument::class => ['index' => 'page', 'decorate_index' => true],
+                            HomeDocument::class => ['index' => 'home', 'decorate_index' => true],
```
alexander-schranz: @kleinkoerkamp is it possible to index both into the same index using something like:

```php
'mapping' => [
    BasePageDocumet::class => ['index' => 'page', 'decorate_index' => true],
]
```

or do we need to change massivesearchbundle for this change?

kleinkoerkamp: @alexander-schranz problem is that during indexing metadata on line https://github.com/massiveart/MassiveSearchBundle/blob/develop/Search/SearchManager.php#L83 is from the real object AKA PageDocument or HomeDocument. Class in this metadata is used as index type.

alexander-schranz: Ok I see it here: https://github.com/wimvds/MassiveSearchBundle/blob/a442e94f4fa233b41646655ab133ab2354c1f0ac/Search/Adapter/ElasticSearchAdapter.php#L126-L127

@danrot @wachterjohannes @nnatter what do you think would the best to do here. In ES7 there will be no longer a type so I think we maybe should get ride of it completely or use the indexname as type if it is required for <= ES6 and indexing the classname into an own field /cc @chirimoya

danrot: On the elastic search documentation they advice to use use multiple indices instead of types now: https://www.elastic.co/guide/en/elasticsearch/reference/6.7/removal-of-types.html

So I would go with that. But TBH I am a little bit confused, because the [API of the MassiveSearchBundle](https://github.com/sulu/sulu/blob/d5d841b/src/Sulu/Bundle/SearchBundle/Controller/WebsiteSearchController.php#L100) uses the word `index` already. Isn't what's passed there really the index? Aren't we already storing stuff in separate indices?

alexander-schranz: Yeah all stuff is in seperate indices (media, contact, ...). But currently it seems page and home are in the same index (thats why the error happens) and are seperated by types (which are not longer supported in es6). For me it would make sense to have home and page in the same index as it should make no difference if its the homepage or a subpage.

danrot: @kleinkoerkamp Can we agree then that we should get rid of the index type, instead of having two separate indices? I don't think that this separation is of any value, since having two classes in the first place has just some technical reasons, and it doesn't make sense to split that form a User's point of view.

@wachterjohannes Do you think removing the index type has any impact that would make this hard?

danrot: Can anybody tell if it is possible to index both kind of pages into the same index?

danrot:
Also, to which branch should this change go? Is it breaking? Then master would be fine, otherwise I would say release/2.0.

src/Sulu/Bundle/PageBundle/DependencyInjection/SuluPageExtension.php
kleinkoerkamp: @danrot not that i know of or it should be a change in the massivesearch bundle. cc @wachterjohannes

alexander-schranz:
With the new [MassiveSearchBundle 2.1.0](https://github.com/massiveart/MassiveSearchBundle/releases/tag/2.1.0) so this error should not longer happen.

Its needed to provide the Elasticsearch Version:

```yaml
massive_search:
    adapter: elastic
    adapters:
        elastic:
            version: 6.8
            hosts: ['%env(ELASTICSEARCH_HOST)%']
```

PR #4678, opened 2019-08-19
Add mandatory condition to templates

### LINKED ISSUE #4609: Form mandatory validation triggers on disabled fields
| Q | A
| --- | ---
| Bug? | yes
| New Feature? | no
| Sulu Version | sulu-2.0

#### Actual Behavior

Form validation triggers on mandatory disabled fields, which prevents form from being saved.

#### Expected Behavior

Form validation doesn't insist on fill disabled and mandatory fields.

#### Steps to Reproduce

```
<property name="relative" type="checkbox">
</property>
<property name="url" type="url" mandatory="true" visibleCondition="relative != true" disabledCondition="relative == true">
</property>
<property name="relativeUrl" type="text_line" mandatory="true" visibleCondition="relative == true" disabledCondition="relative != true">
</property>
```

#### Possible Solutions

Either form validation should skips disabled fields, or mandatory field should be conditionable like disabledCondition (which will give more freedom to use forms).

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| Fixes issue | fixes #4609
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?
Disable the mandatory property, when the field is hidden or disabled.
Add a mandatory condition to xml templates.

#### Why?
If a mandatory field is hidden, the form fails to validate and the user doesn't get a proper error message.

#### Example Usage

~~~xml
<property name="article" type="text_line" mandatoryCondition="title == 'testTitle">
        <meta>
        </meta>
</property>
~~~

#### To Do

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/FormMetadata/FormMetadataMapper.php
+++ b/src/Sulu/Bundle/AdminBundle/FormMetadata/FormMetadataMapper.php
@@ -72,6 +72,7 @@ class FormMetadataMapper
         $section->setColSpan($property->getColSpan());
         $section->setDisabledCondition($property->getDisabledCondition());
         $section->setVisibleCondition($property->getVisibleCondition());
+        $section->setMandatoryCondition($property->getMandatoryCondition());
 
         foreach ($property->getChildren() as $component) {
             if ($component instanceof BlockMetadata) {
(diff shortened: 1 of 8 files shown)


danrot:
We have just discussed that this approach cannot really work, because the serverside is not able to evaluate the `mandatoryCondition` attribute... We'll have to find a different approach, and therefore I am going to close this PR.

PR #4680, opened 2019-08-19
Use permissions from security context if no object security has been set

danrot:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR uses the permissions from the security context for the permissions if no permission on the object has been set. It also fixes the bug causing wrong behavior when a role has been created after object permission has been set.

### DIFF
--- a/src/Sulu/Component/Security/Authorization/AccessControl/AccessControlManager.php
+++ b/src/Sulu/Component/Security/Authorization/AccessControl/AccessControlManager.php
@@ -89,7 +89,7 @@ class AccessControlManager implements AccessControlManagerInterface
         }
 
         $objectPermissions = $this->getUserObjectPermission($securityCondition, $user);
-        $checkPermissionType = empty($objectPermissions);
+        $checkPermissionType = $objectPermissions === null;
 
         $securityContextPermissions = $this->getUserSecurityContextPermissions(
             $securityCondition->getLocale(),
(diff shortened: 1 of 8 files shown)


src/Sulu/Component/Security/Authorization/AccessControl/AccessControlManager.php
```diff
@@ -163,7 +163,7 @@ private function getUserObjectPermission(SecurityCondition $securityCondition, U
     private function getUserObjectPermissionByArray($permissions, UserInterface $user)
     {
         if (empty($permissions)) {
-            return [];
+            return null;
```
danrot: Maybe this change is not needed, if we make sure that a return value of an empty array always means that no permissions are set. This means that if we make sure that [the ContentRepository](https://github.com/sulu/sulu/blob/e58a64c/src/Sulu/Component/Content/Repository/ContentRepository.php#L854) returns an array with all permission values set false instead of returning an empty array, then this could work. This would have to look similar as in the [SecuritySubscriber](https://github.com/sulu/sulu/blob/e58a64c35406e1d0b02fda64a34c81071af11d31/src/Sulu/Component/Content/Document/Subscriber/SecuritySubscriber.php#L96-L105).

https://github.com/sulu/sulu/blob/e58a64c35406e1d0b02fda64a34c81071af11d31/src/Sulu/Component/Content/Document/Subscriber/SecuritySubscriber.php#L96-L105

PR #4684, opened 2019-08-20
Fix permissions in page list for roles without any permission

danrot:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR makes the `ContentRepository` return an empty array only if no permissions at all are set on the corresponding node. In case some permissions are set, empty arrays or missing information for a role will be interpreted as having no permissions at all.

#### Why?

Because an empty array will be interpreted as having all permissions in our security system.

That this behavior is not correct can be tested in the following way:

1. Create a role without the security permission for a webspace
2. Create a page in that webspace
3. Remove all permissions from the role of step 1
4. See that the page has an edit icon in the page list, even for user having only the role from step 1
5. If you click the edit icon you get an error message telling you that you don't have the required permissions

With this PR the edit icon from step 4 should not be visible.

### DIFF
--- a/src/Sulu/Component/Content/Tests/Functional/Repository/ContentRepositoryTest.php
+++ b/src/Sulu/Component/Content/Tests/Functional/Repository/ContentRepositoryTest.php
@@ -763,6 +763,11 @@ class ContentRepositoryTest extends SuluTestCase
 
     public function testFindWithPermissionsNotGranted()
     {
+        $packageVersion = $this->getPackageVersion('jackalope/jackalope');
+        if (version_compare($packageVersion, '1.3.0') < 0) {
+            $this->markTestSkipped('The jackalope/jackalope version below 1.3.0 has a bug causing this test to fail');
+        }
+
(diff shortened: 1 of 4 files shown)


src/Sulu/Component/Content/Tests/Functional/Repository/ContentRepositoryTest.php
```diff
@@ -719,7 +720,84 @@ public function testFindPermissions()
...
+
+        $page = $this->createPage(
+            'test-1',
+            'de',
+            [],
+            null,
+            [
+                1 => ['edit' => false],
+                2 => ['view' => false, 'archive' => false],
+                3 => ['add' => false],
```
danrot: This results in an error with a jackalope version <1.3. Shall we raise the minimum dependency, or should we just skip this test if that version is installed? (would require reading the `composer.lock` file)

PR #4690, opened 2019-08-22
Make repositories lazy

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | none
| Related issues/PRs | built upon #4697
| License | MIT
| Documentation PR | none

#### What's in this PR?

This PR makes all repositories lazy services.

#### Why?

This is a part of the problem that doctrine tries to connect to the database on `cache:warmup`.


danrot:
Just tested it with a few `cache:warmup` calls, this change really seems to make at least that one a bit faster.

PR #4696, opened 2019-08-23
Add message when no shadow locale is available

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | none
| Related issues/PRs | none
| License | MIT
| Documentation PR | none

#### What's in this PR?

This PR add a message for shadow locale select if no locale is available.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Select.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Select.js
@@ -3,6 +3,7 @@ import React from 'react';
 import type {Element, ElementRef} from 'react';
 import {action, observable} from 'mobx';
 import {observer} from 'mobx-react';
+import {translate} from '../../utils/Translator';
 import Popover from '../Popover';
 import Menu from '../Menu';
 import Action from './Action';
@@ -145,7 +146,11 @@ class Select<T> extends React.Component<Props<T>> {
                                 menuRef={setPopoverElementRef}
                                 style={popoverStyle}
                             >
-                                {clonedChildren}
+                                {React.Children.count(clonedChildren) > 0 ? clonedChildren : (
+                                    <Option disabled={true} value={null}>
+                                        {translate('sulu_admin.no_options_available')}
+                                    </Option>
+                                )}
                             </Menu>
                         )
                     }
--- a/src/Sulu/Bundle/PageBundle/Resources/js/containers/Form/fields/PageSettingsShadowLocaleSelect.js
+++ b/src/Sulu/Bundle/PageBundle/Resources/js/containers/Form/fields/PageSettingsShadowLocaleSelect.js
@@ -2,7 +2,6 @@
 import React from 'react';
 import {toJS} from 'mobx';
 import {SingleSelect} from 'sulu-admin-bundle/components';
-import {translate} from 'sulu-admin-bundle/utils';
 import type {FieldTypeProps} from 'sulu-admin-bundle/types';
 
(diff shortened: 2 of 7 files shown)


src/Sulu/Bundle/PageBundle/Resources/js/containers/Form/fields/PageSettingsShadowLocaleSelect.js
```diff
@@ -21,25 +22,31 @@ export default class PageSettingsShadowLocaleSelect extends React.Component<Fiel
...
+                    return (
+                        <SingleSelect.Option
+                            key={contentLocale}
+                            value={contentLocale}
+                        >
+                            {contentLocale}
+                        </SingleSelect.Option>
+                    );
+                })}
+                {filteredContentLocales.length === 0 && (
```
danrot: Shouldn't that kind of code be in the `Select` component? I think that we might have the same situation in multiple places.

We could introduce a new prop called `noOptionsText` or something like that, but I think I would even prefer a generic text here (the label is already "Shadow Locale", then a text like "No options available" would cut it for me as well).

src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Select.js
```diff
@@ -117,6 +120,7 @@ class Select<T> extends React.Component<Props<T>> {
             disabled,
             displayValue,
             skin,
+            noOptionsText,
```
danrot: Here the alphabetic order is not correct.

src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Select.js
```diff
@@ -145,7 +149,11 @@ class Select<T> extends React.Component<Props<T>> {
                                 menuRef={setPopoverElementRef}
                                 style={popoverStyle}
                             >
-                                {clonedChildren}
+                                {clonedChildren.length > 0 ? clonedChildren : (
+                                    <Option disabled={true}>
+                                        {translate(noOptionsText)}
```
danrot: I was actually just suggesting this option if we want to leave the other text. I think if we do not use this option yet, I would not implement it.

And btw, in cases like this we decided to translate the text in the caller component, so that we have only to send the string in this component. Makes it more flexible when we might have some phrases that are not translatable.

src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/tests/Select.test.js
```diff
@@ -68,6 +68,22 @@ test('The component should render with an icon', () => {
...
+    const select = mount(
+        <Select
+            displayValue="My text"
+            icon="su-plus"
+            isOptionSelected={isOptionSelected}
+            onSelect={onSelect}
+        >
+        </Select>
+    );
+    expect(select.render()).toMatchSnapshot();
```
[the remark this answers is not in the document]
wachterjohannes: should i remove it?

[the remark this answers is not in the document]
danrot: I would say try to click the button and see what's then in the snapshot :slightly_smiling_face: The text you have defined should show up there somehow. Otherwise I would not do a snapshot test, and check using the `prop` function.

wachterjohannes:
@danrot fixed

PR #4699, opened 2019-08-23
Avoid bc break in twig templates by provide uuid and id properties for pages

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes -
| Related issues/PRs | 
| License | MIT
| Documentation PR | -

#### What's in this PR?

Add uuid property to contentmapper besides the id.

#### Why?

Avoid twig developers to change all twig template which access uuid so introduce a bc layer with uuid.

Introduced in: https://github.com/sulu/sulu/pull/4550/files#diff-2a74d79cea8c966c3ad6a835206c6c3bL751

#### Example Usage

 - smart content
 - navigation twig extension
 - breadcrumb return uuid
 - page return uuid

#### To Do


nnatter:
looks good to me, but would prefer if someone else would have another look at it. i do not have that much experience with the page api 🙂

PR #4701, opened 2019-08-23
WIP: Add route collection builder

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | yes
| Deprecations? | no
| Fixed tickets | fixes -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

We introduce a RouteCollection so we could manipulate in an AppAdmin.

#### Why?

For example to add toolbarAction to exist routes.

#### BC Breaks/Deprecations

See UPGRADE.md

#### To Do

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Admin/Admin.php
+++ b/src/Sulu/Bundle/AdminBundle/Admin/Admin.php
@@ -41,7 +41,7 @@ abstract class Admin implements RouteProviderInterface, NavigationProviderInterf
      *
      * @return Route[]
      */
-    public function getRoutes(): array
+    public function getRoutes(RouteCollection $routeCollection): array
     {
         return [];
     }
--- a/src/Sulu/Bundle/AdminBundle/Admin/RouteProviderInterface.php
+++ b/src/Sulu/Bundle/AdminBundle/Admin/RouteProviderInterface.php
@@ -18,5 +18,5 @@ interface RouteProviderInterface
     /**
      * @return Route[]
      */
-    public function getRoutes(): array;
+    public function getRoutes(RouteCollection $routeCollection): array;
 }
--- a/src/Sulu/Bundle/AdminBundle/Admin/RouteRegistry.php
+++ b/src/Sulu/Bundle/AdminBundle/Admin/RouteRegistry.php
@@ -59,16 +59,16 @@ class RouteRegistry
 
     private function loadRoutes(): void
     {
-        $routes = [];
+        $routeCollection = new RouteCollection();
         foreach ($this->adminPool->getAdmins() as $admin) {
             if (!$admin instanceof RouteProviderInterface) {
                 continue;
             }
 
-            $routes = array_merge($routes, $admin->getRoutes());
+            $routeCollection->addRoutes($admin->getRoutes($routeCollection));
         }
 
-        $this->validateRoutes($routes);
+        $this->validateRoutes($routeCollection->all());
 
         array_walk($routes, function(&$route, $index) {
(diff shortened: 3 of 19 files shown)


alexander-schranz:
@nnatter @wachterjohannes we are discussing how we could implement this in the best way. @danrot and I thought it would make best sense to handle this by introducing a custom RouteBuilder and the route collection store instead of `Route` classes only `RouteBuilder` this is a bc break as routes can not be longer created with new Route instead they need to create a new RouteBuilder the function will match the exist Route functions. What do you think?

danrot:
Btw, I now think that we should the routes to the `RouteCollection` instead of returning them... Manipulating existing routes this way, and add new by returning something from the function feels strange :confused:

danrot:
@sulu/core-team I have been thinking even more about my statement before... I think with the change we are about to introduce the function name `getRoutes` does not make any sense anymore. The `get` prefix suggests that we are only returning stuff, but not that existing routes might be changed.

Apart from that we are not registering `Routes` anymore, but `RouteBuilders` instead. But I am not sure if we should reflect that in the name.

So basically we have two naming decisions to make here:

- `RouteCollection`/`RouteBuilderCollection` for the class holding all the route builder instances
- `registerRoutes`/`registerRouteBuilders`/`initializeRoutes`/`initializeRouteBuilder`/`manipulateRoutes`/`manipulateRouteBuilder`

I think I would go with `RouteColllection`, because from a user's POV he is still handling routes, and the `RouteBuilder` is "just" an implementation detail.

The second decision is a bit harder IMO. I am not really happy with any of my suggestions, because I can't think of a catchy name for something that might add new routes or update existing ones. I also can't think of any pattern that handles that case. If any of you have a better suggestion they are welcome :slightly_smiling_face:

nnatter:
I would probably go with `RouteBuilderCollection` because it is just the class name and it feels kind of cleaner/more explicit, but I dont have a very strong opinion here. 

What do you think about `configureRoutes(RouteBuilderCollection $routeBuilders)`? I would omit the `Builders` in the method name because the user is thinking about routes here as you said.

danrot:
I like `configureRoutes` better to the other suggestions I've made. But I would either go `configureRoutes(RouteCollection $routes)` or `configureRouteBuilders(RouteBuilderCollection $routeBuilders)`. Mixing both of them feels very weird to me. Either we say the user is thinking about `Routes` or about `RouteBuilders`, but then I would follow this decision in both cases.

wachterjohannes:
i would use `configureRoutes(RouteCollection $routes)`

alexander-schranz:
I would go with:

 - `RouteProviderInterface::configureRouteBuilders(RouteBuilderCollection $collection)`
 - `RouteBuilderCollection::add(RouteBuilder $routeBuilder)`
 - `RouteBuilderCollection::get('name'): RouteBuilder`
 - `RouteBuilderCollection::all(): array`

but I'm also ok with skipping the `Builder` from the namespace.

danrot:
After a short discussion we will go with `RouteProviderInterface.:configureRoutes(RouteCollection $collection)`. And we realized that we should have the same concept for the `Navigation`, so we'll rename `NavigationProviderInterface::getNavigation` to `NavigationProviderInterface::configurationNavigationItems(NavigationItemCollection $navigationItemCollection)`.

This will also allow us to remove the `getNavigationItemRoot`, `getNavigationItemSettings` and similar methods.

danrot:
replaced by #4711

PR #4706, opened 2019-08-28
Prevent unnecessary config request if user is not logged in

nnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This PR adjusts `Initializer` javascript class to prevent unnecessary requests if no user is logged in.

#### Why?

To prevent unnecessary 401 responses that will reset cached Basic Authentication credentials.
Without this change, if the Sulu admin runs behind a webserver with enabled Basic Authentication, the user must provide Basic Authentication credentials **two times** before he is able to log in with his sulu credentials.

Right now, the following happens if the Sulu admin runs behind a webserver with enabled Basic Authentication:

- The browser will ask for Basic Authentication credentials on the initial `admin/` request and cache the given credentials
- The `Initializer` dispatches the `admin/config` request (to which the browser appends the cached credentials) which will lead to a 401 response because the user is not logged in into Sulu yet
- The 401 response triggers the browser to reset the cached Basic Authentication credentials
- The `Initializer` dispatches a `admin/api/localizations?flat=true` request. The browser will not attach any credentials and therefore will ask for Basic Authentication credentials **again**
- The user sees the loaded login screen

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/views/Admin/main.html.twig
+++ b/src/Sulu/Bundle/AdminBundle/Resources/views/Admin/main.html.twig
@@ -12,7 +12,7 @@
         <script type="text/javascript">
             {% autoescape false %}
             const SULU_CONFIG = Object.freeze({
-                initialLoginState: {{ app.user ? 'true' : 'false' }},
+                initialLoginState: {{ is_granted('IS_AUTHENTICATED_REMEMBERED') ? 'true' : 'false' }},
                 translations: {{ translations|json_encode }},
                 fallbackLocale: '{{ fallback_locale }}',
                 endpoints: {{ endpoints|json_encode }},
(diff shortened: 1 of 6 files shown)


src/Sulu/Bundle/AdminBundle/Resources/views/Admin/main.html.twig
```diff
@@ -12,6 +12,7 @@
         <script type="text/javascript">
             {% autoescape false %}
             const SULU_CONFIG = Object.freeze({
+                initialLoginState: {{ app.user ? 'true' : 'false' }},
```
alexander-schranz: maybe we should use `is_granted('IS_AUTHENTICATED_REMEMBERED')` here instead of checking if `app.user` variable exist as I remeber it could be in some cases be string with the name `.anom`.

danrot: :+1:

nnatter: done 🙂

danrot:
@nnatter Can you please rebase? I think merging #4707 makes your PR incompatible with the lastest changes.

PR #4704, opened 2019-08-27
Add support for IE11

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR tries to add support for IE 11.

#### Why?

Because would like to support that browser as well.

#### To Do

- [ ] Create a documentation PR
- [ ] Add breaking changes to UPGRADE.md

PR #4709, opened 2019-08-29
Update to latest JS packages

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | https://github.com/sulu/sulu-minimal/pull/197
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR updates all the JS dependencies to their latest version. The only exception is mobx, because if we update it, it will not be compatible with IE11 anymore. However, this PR still removes all the polyfills required by IE11. The reasoning behind this is, that CKEditor5 is not working with IE11 at the moment anyway.

We can add the polyfills when continue to work on #4704, but it doesn't make sense as long as CKEditor is blocking it.

#### Why?

Because if we update to latest packages now we have less to support later.


.browserslistrc
```diff
@@ -1,3 +1,3 @@
 last 2 versions
```
alexander-schranz: add to `last 2 versions and not dead`

and would go with `> 0` i have strange feeling:

```
last 2 versions and not dead
not ie < 12
not ie_mob < 12
```

this will throw away exotic not longer used browsers here the result differnent of `npx browserslist`:

Bildschirmfoto 2019-08-29 um 09 45 15

src/Sulu/Bundle/AdminBundle/Resources/js/containers/CKEditor5/README.md
```diff
@@ -22,7 +22,7 @@ const handleBlur = () => alert('Text editing finished!');
 The editor can be extended by adding more plugins and more configuration. That's what the `PluginRegistry` and
 `ConfigRegistry` are for:
 
-```javascript
+```javascript static
```
alexander-schranz: looks misplaced

danrot: It actually is not. This makes the code below not rendered in the styleguide. Otherwise we would have to add the ckeditor5-font package somehow, because we get errors.

alexander-schranz: ok good to know

PR #4716, opened 2019-09-04
Introduce collaboration messages

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR will show a warning if other people are editing the same form.

#### Why?

Because this could prevent people from overriding the content of other users.

#### BC Breaks/Deprecations

None yet.

#### To Do

- [x] Implement heartbeat with 20s and a timeout of 60s
- [x] ~~Add dependency rules using dependency-cruiser and deptrac~~
- [x] Merge `CollaborationBundle` with `AdminBundle`
- [x] Increase timeout values in Config
- [ ] Add lock mechanism to avoid multiple collaboration request change cache at same time

##### Architectural questions

Is the `CollaborationBundle` is a low level bundle, which should not depend on any other bundle and add these rules in dependency-cruiser and deptrac?

How should the `CollaborationBundle` be integrated in the `AdminBundle`?
1. **The `AdminBundle` simply uses the `CollaborationBundle`.** Would have the advantage that if somebody builds a custom form the logic in the `CollaborationBundle` could still be used (and the `CollaborationBundle` could offer a `CollaborationStore` or something like that). The collaboration store could also implement a heartbeat/polling updating the other collaborators all the time.
2. **The `CollaborationBundle` uses one of the hooks from the `AdminBundle`.** Question is how does the `CollaborationBundle` know when to add this functionality? Could e.g. use the `UpdateRouteHook` mechanism, and check for a `collaboration` option on the route (could result in naming collisions) or just do that for every view (do we really want to see a warning banner e.g. in the list view? 

We decided to go for option 1.


src/Sulu/Bundle/AdminBundle/Controller/CollaborationController.php
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    private function getResourceKey(Request $request): string
+    {
+        return $request->query->get('resourceKey');
+    }
+
+    private function getConnectionId(Request $request)
+    {
+        return crypt($request->getSession()->getId(), $this->secret);
```
chirimoya: use default value for secret?

src/Sulu/Bundle/AdminBundle/Entity/CollaborationRepository.php
```diff
@@ -0,0 +1,101 @@
...
+     */
+    public function delete(Collaboration $collaboration): array
+    {
+        $cacheItem = $this->cache->getItem($this->getCacheIdFromCollaboration($collaboration));
+        $value = array_filter($cacheItem->get() ?? [], function(Collaboration $cachedCollaboration) use ($collaboration) {
+            return $collaboration->getConnectionId() !== $cachedCollaboration->getConnectionId();
+        });
+        $cacheItem->set($value);
+
+        $this->cache->save($cacheItem);
```
chirimoya: remove cache entry, if $value is []

src/Sulu/Bundle/AdminBundle/Resources/translations/admin.de.json
```diff
@@ -141,5 +141,6 @@
     "sulu_admin.no_permissions": "Keine Berechtigung",
     "sulu_admin.no_options_available": "Keine Optionen verfügbar",
     "sulu_admin.form_contains_invalid_values": "Das Formular beinhaltet ungültige Werte",
-    "sulu_admin.form_save_server_error": "Beim Speichern des Formulars auf dem Server ist ein Fehler aufgetreten"
+    "sulu_admin.form_save_server_error": "Beim Speichern des Formulars auf dem Server ist ein Fehler aufgetreten",
```
chirimoya: remove Server

src/Sulu/Bundle/AdminBundle/Entity/CollaborationRepository.php
danrot: The `delete` function is not always called, therefore it is not even sure that everything is deleted, if we do that here. For this reason we leave it as it is.

PR #4717, opened 2019-09-04
Show errors with message

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR will add some messages to errors, and introduce a new warning snackbar as an alternative.

#### Why?

Because error message at least give some feeling what is going wrong, and for the collaboration feature we need a warning, to signal to the user that somebody else is already editing this item.

#### BC Breaks/Deprecations

None yet

#### To Do

- [x] Add warnings

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Snackbar/README.md
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Snackbar/README.md
@@ -1,4 +1,4 @@
-The `Snackbar` componetn is used to display information that is moving in from the top of the screen using an animation.
+The `Snackbar` component is used to display information that is moving in from the top of the screen using an animation.
 
 ```javascript
 <Snackbar message="Something went wrong" type="error" />
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Snackbar/README.md
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Snackbar/README.md
@@ -1,13 +1,13 @@
 The `Snackbar` componetn is used to display information that is moving in from the top of the screen using an animation.
 
 ```javascript
-<Snackbar type="error" />
+<Snackbar message="Something went wrong" type="error" />
 ```
 
 Besides errors it can also show warnings:
 
 ```javascript
-<Snackbar type="warning" />
+<Snackbar message="Something not so bad went wrong" type="warning" />
 ```
 
 There are two different callbacks that can be added. One is the `onCloseClick` callback, which makes a close button
@@ -16,7 +16,7 @@ appear calling this callback when it is clicked.
 ```javascript
 const closeClickHandler = () => alert('The snackbar should be closed now');
 
-<Snackbar onCloseClick={closeClickHandler} type="error" />
+<Snackbar message="Something went wrong" onCloseClick={closeClickHandler} type="error" />
 ```
 
 The second one is the `onClick` callback, which is called whenever the snackbar is clicked.
@@ -24,7 +24,7 @@ The second one is the `onClick` callback, which is called whenever the snackbar 
 ```javascript
 const clickHandler = () => alert('The snackbar was clicked');
 
-<Snackbar onClick={clickHandler} type="error" />
+<Snackbar message="Something went wrong" onClick={clickHandler} type="error" />
 ```
 
 The toolbar is also animated when it (dis)appears.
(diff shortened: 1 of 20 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/components/Snackbar/README.md
```diff
@@ -1,13 +1,13 @@
 The `Snackbar` componetn is used to display information that is moving in from the top of the screen using an animation.
```
nnatter: component

src/Sulu/Bundle/AdminBundle/Resources/js/components/Snackbar/snackbar.scss
```diff
@@ -8,15 +8,20 @@ $snackbarWarningColor: $blueZodiac;
...
 .snackbar {
+    position: relative;
+    z-index: 2;
     height: $snackbarHeight;
     padding: 0 20px;
     display: flex;
     align-items: center;
     will-change: margin-top;
-    transition: margin-top .15s;
+    transition: margin-top .15s, z-index 0s .15s;
```
nnatter: why do we need a z-index now? and why do we animate it? 🤔

nnatter: same for visibility

src/Sulu/Bundle/AdminBundle/Resources/js/components/Snackbar/Snackbar.js
```diff
@@ -17,11 +20,38 @@ const ICONS = {
...
 };
 
-export default class Snackbar extends React.Component<Props> {
+@observer
+class Snackbar extends React.Component<Props> {
     static defaultProps = {
         visible: true,
     };
 
+    @observable message: ?string;
```
nnatter: what is the reason for storing the message as class property instead of just using the prop value? 🤔

src/Sulu/Bundle/AdminBundle/Resources/translations/admin.de.json
```diff
@@ -135,5 +136,7 @@
     "sulu_admin.has_changed_warning_dialog_text": "Dieser Inhalt wurde zwischenzeitlich von einem anderen Benutzer geändert. Wenn der Inhalt gespeichert wird, gehen diese Daten verloren.",
     "sulu_admin.edit_profile": "Profil bearbeiten",
     "sulu_admin.no_permissions": "Keine Berechtigung",
-    "sulu_admin.no_options_available": "Keine Optionen verfügbar"
+    "sulu_admin.no_options_available": "Keine Optionen verfügbar",
```
nnatter: is this translation used anywhere?

src/Sulu/Bundle/AdminBundle/Resources/js/components/Snackbar/snackbar.scss
danrot: It's not really animated, the `z-index` and `visibility` are only delayed. The thing is that we have two snackbars (one for warnings, one for errors), and they are interferring with each other.

The `z-index` is to make sure that the currently visible one lies over the one not being visible (otherwise the warning snackbar will slide over the error snackbar if both were visible, and appear after it has blocked the sight to the error snackbar).

The visibility also makes sure that the snackbar is not visible if it was e.g. moved down by a snackbar before itself (since we are animating using `margin-top` that can happen if the error toolbar is visible, but the warning is not).

And these two values are not really animated, they are "only" delayed (that's the second time value in the property, not the first one, which is the animation duration). And that's because otherwise the snackbar will be made invisible before it started moving up again when being closed.

src/Sulu/Bundle/AdminBundle/Resources/js/components/Snackbar/Snackbar.js
danrot: Because if we don't do that, then the message will disappear before the snackbar is animated away.

src/Sulu/Bundle/AdminBundle/Resources/translations/admin.de.json
danrot: I just had to add a comma at the end :wink: It is used, but not in this PR.

nnatter: 🤦‍♂

PR #4720, opened 2019-09-06
Make AdapterSwitch in PageList clickable

danrot:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | probably #4710 
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR makes the HTML container for the `WebspaceSelect` take only as much width as it needs.

#### Why?

Because the container component lies over the `AdapterSwitch` of the `List`, making it impossible to switch to the tree view.

### DIFF
--- a/src/Sulu/Bundle/PageBundle/Resources/js/views/WebspaceTabs/webspaceTabs.scss
+++ b/src/Sulu/Bundle/PageBundle/Resources/js/views/WebspaceTabs/webspaceTabs.scss
@@ -5,4 +5,5 @@
     z-index: 2;
     display: inline-block;
     margin-bottom: $webspaceTabsWebspaceSelectMargin;
+    width: 0;
 }


src/Sulu/Bundle/PageBundle/Resources/js/views/WebspaceTabs/webspaceTabs.scss
```diff
@@ -5,4 +5,5 @@
     z-index: 2;
     display: inline-block;
     margin-bottom: $webspaceTabsWebspaceSelectMargin;
+    width: 0;
```
danrot: @alexander-schranz I have tested this in Chrome, Firefox and Edge, and it seems to work everywhere, although it feels a bit dirty. Do you think that will make any problems?

alexander-schranz: I would prefer setting a margin-right or a max-width but width: 0; works also in this case aslong as this container has overflow visible and the button inside of the container has a min-width or width set.

danrot: The thing is that if I just set a width of 300px we also have to remember to fix that here as well, in case the width of the button changes one day. With the width set to 0 it will just always work.

alexander-schranz: A variable could be used for that width but aslong as the component is never used next to another component and will stay as here alone in its container I see no problem with width: 0. As this component is very special and will I think only be used in a similiar usecase I will going to merge this one.

PR #4722, opened 2019-09-06
Fix changing of locale for media after save

danrot:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR removes the edit route from the route for the edit form.

#### Why?

Because this will cause a redirect after saving the form (to itself) and destroy the `ResourceStore` to avoid memory leaks. But when the form is redirected to itself no redirect happens, only the `ResourceStore` is destroyed, which means that the content of the new locale is not loaded.

#### To Do

- [ ] Add check in RouteBuilder?


src/Sulu/Bundle/MediaBundle/Admin/MediaAdmin.php
```diff
@@ -134,7 +134,6 @@ public function configureRoutes(RouteCollection $routeCollection): void
                     ->setResourceKey('media')
                     ->setFormKey('media_details')
                     ->setTabTitle('sulu_media.information_taxonomy')
-                    ->setEditRoute(static::EDIT_FORM_DETAILS_ROUTE)
```
danrot: This route was directing to itself here. Wondering if we should catch that in the `RouteBuilder` and throw an exception, to make clear that this does not work correctly.

alexander-schranz: Throwing a LogicException or something similiar sounds like a good idea for me to avoid bug like this.

PR #4727, opened 2019-09-11
Replace Hateoas Bundle with custom implementation

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | yes
| Deprecations? | no
| Fixed tickets | part of #4724
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

Remove `Hateoas` and `HateoasBundle` as requirement

#### Why?

As we are not using `Hateoas` _links and relation in the react app I see no advantage to keep `Hateoas`. Also most of our entites has not the annotions set so I think we should remove them from all or we need to add them to all, which will be much work if we many entities with relations.

I personally would like to keep our dependency list as small as possible.

#### BC Breaks/Deprecations

Yes all references to `Hateoas` and `_links` are removed.

#### To Do

- [x] Remove entity annotations
- [x] Removing Bundle and Library as requirement
- [x] Replace `Hateoas\Representation\CollectionRepresentation`
- [x] Replace `Hateoas\Representation\RouteAwareRepresentation`
- [x] Create new `ListRepresentation` and constructor arguments optional for pagination


alexander-schranz:
@sulu/core-team what you like to keep Hateoas?

src/Sulu/Bundle/AdminBundle/Serializer/Handler/CollectionRepresentationHandler.php
```diff
@@ -0,0 +1,78 @@
...
+                'direction' => GraphNavigator::DIRECTION_SERIALIZATION,
+                'format' => 'json',
+                'type' => CollectionRepresentation::class,
+                'method' => 'serializeForJson',
+            ],
+        ];
+    }
+
+    /**
+     * FIXME This should be refractored so the annoations are used and no manually mapping here is needed.
```
danrot: Doesn't feel hacky to me at all.

src/Sulu/Bundle/AdminBundle/Serializer/Handler/CollectionRepresentationHandler.php
```diff
@@ -0,0 +1,78 @@
...
+    public function serializeForJson(
+        JsonSerializationVisitor $visitor,
+        CollectionRepresentation $representation,
+        array $type,
+        Context $context
+    ) {
+        $data = [];
+
+        if ($representation instanceof PaginatedRepresentation) {
+            $data['total'] = $representation->getTotal();
```
alexander-schranz: @nnatter suggestion was to add a `toArray()` function for this data so own representation could be created.

e.g.: the ItemCollectionRepresentation for SmartContent need a datasource (not sure if that is used by the js)

PR #4730, opened 2019-09-13
Add ToolbarAction to create dropdowns

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR will add a new `ToolbarAction`, that allows to combine existing `ToolbarAction`s to a dropdown.

#### Why?

Because this provides a lot more flexibility, and we can stretch our "no-JS" workflow even further.

#### To Do

- [x] Do not show button if no dropdown item is available


src/Sulu/Bundle/AdminBundle/Resources/js/views/Form/toolbarActions/CopyLocaleToolbarAction.js
```diff
@@ -0,0 +1,135 @@
...
+                confirmText={translate('sulu_admin.ok')}
+                key="sulu_admin.copy_locale"
+                onCancel={this.handleClose}
+                onConfirm={this.handleConfirm}
+                open={this.showCopyLocaleDialog}
+                title={translate('sulu_admin.copy_locale')}
+            >
+                <div className={copyLocaleActionStyles.dialog}>
+                    <p>{translate('sulu_admin.choose_target_locale')}:</p>
+                    {locales.map((locale) => currentLocale.get() === locale
```
alexander-schranz: Nothing todo with this PR but `currentLocale.get()` looks for me like a pitfall for other developers should we not return a string instead of the observable as computed value?

src/Sulu/Bundle/AdminBundle/Resources/js/views/Form/toolbarActions/CopyLocaleToolbarAction.js
```diff
@@ -0,0 +1,135 @@
...
+    @action handleCheckboxChange = (checked: boolean, value?: string | number) => {
+        if (checked && typeof value === 'string' && !this.selectedLocales.includes(value)) {
+            this.selectedLocales.push(value);
+        } else {
+            this.selectedLocales.splice(this.selectedLocales.findIndex((locale) => locale === value), 1);
+        }
+    };
+
+    @action clearSelectedLocales = () => {
+        this.selectedLocales.splice(0, this.selectedLocales.length);
```
alexander-schranz: Would `this.selectedLocales = [];` not be simplier?

src/Sulu/Bundle/AdminBundle/Resources/js/views/Form/toolbarActions/DeleteDraftToolbarAction.js
```diff
@@ -0,0 +1,104 @@
...
+                title={translate('sulu_page.delete_draft_warning_title')}
+            >
+                {translate('sulu_page.delete_draft_warning_text')}
+            </Dialog>
+        );
+    }
+
+    getToolbarItemConfig() {
+        const {
+            display_condition: displayCondition,
```
alexander-schranz: do we serialize this to snake_case? 😲

src/Sulu/Bundle/AdminBundle/Resources/js/views/Form/toolbarActions/SetUnpublishedToolbarAction.js
```diff
@@ -0,0 +1,109 @@
...
+                title={translate('sulu_page.unpublish_warning_title')}
+            >
+                {translate('sulu_page.unpublish_warning_text')}
+            </Dialog>
+        );
+    }
+
+    getToolbarItemConfig() {
+        const {
+            display_condition: displayCondition,
```
alexander-schranz: same as above why do we use snake_case here?

src/Sulu/Bundle/PageBundle/Admin/PageAdmin.php
```diff
@@ -116,9 +125,20 @@ public function configureRoutes(RouteCollection $routeCollection): void
             'sulu_admin.delete' => [
                 'display_condition' => '(!_permissions || _permissions.delete) && url != "/"',
             ],
-            'sulu_page.edit' => [
-                'copy_locale_display_condition' => '(!_permissions || _permissions.edit)',
-                'publish_display_condition' => '(!_permissions || _permissions.live)',
+            'sulu_admin.dropdown' => [
+                'label' => $this->translator->trans('sulu_admin.edit', [], 'admin'),
```
alexander-schranz: Normally we give in this kind of configuration only translation keys and translate it in the frontend e.g.: https://github.com/sulu/sulu/blob/f1aabb71cb2590115b7ff4771b4a8f81574d9675/src/Sulu/Bundle/PageBundle/DependencyInjection/SuluPageExtension.php#L86

So would prefer todo this also here to set only the translation key and use the translator in the dropdown component.

danrot: Seems like we already are inconsistent here... AFAIK we said in the beginning that we translate as much as possible on the server from the `/admin/config` request. This would have advantage that we have less work todo on the client, and this happens on the server only once per session. If you check e.g. the `teaser` stuff being returned from it or the occurence of the `sulu_admin.toggler` you'll see that this has already been translated on the server. I think the field type options are the exception here (or did you find any other examples?), because the translator is not available in the Extensions...

If we revoke that decision and say we always translate on the client, then we also have to fix that in other places, e.g. the teasers and other options on the route.

alexander-schranz: @danrot Ok I would see it the other way around to translate it in JS so the PHP Developer has less todo 😄. On the RouteBuilder and Navigation they also need only to give translation keys so currently it seems in the Admin Class only the ToolbarActions need the translator.

/cc @sulu/core-team 

For `/admin/config` not sure but thats a request which a normal developer does not extend and when then add new configs where it is their choice to translate it in php or js.

danrot:
I just looked into the translation problem, and it is not possible to make it consistent everywhere.

In the `Extension` classses (where we define the `fieldTypeOptions`) we don't have a translator available, because it is too early in the process.

On the other side we e.g. have the `indexes` defined in the `IndexConfigurationProvider` classes, where we can't translate everything, because we want the untranslated names of webspaces to occur on the website (at least we can't move the translation to the frontend if we want to have everything cluttered with warnings). Same is true for the translation of the list `FieldDescriptor` and the properties of the form, because they can either be translated in the XML file directly or a translation key is put in the XML.

Personally I would still go for the approach to translate as much as possible on the server, because then it has be done less times, and it is more explicity (i.e. the developer won't wonder why something different than he put into the options is shown).

PR #4733, opened 2019-09-17
Fixed video thumbnail generation

thomasduenser:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no

#### What's in this PR?

Fixes the generation of video thumbnails.

#### Why?

The wrong binary was used for `ffprobe` and the thumbnail generation for the newest FFmpeg version didn't work.

Errors that happened:
- `Unable to save frame` 
- `FFmpeg cannot edit existing files in-place.`

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/Media/ImageConverter/MediaImageExtractor.php
+++ b/src/Sulu/Bundle/MediaBundle/Media/ImageConverter/MediaImageExtractor.php
@@ -134,11 +134,13 @@ class MediaImageExtractor implements MediaImageExtractorInterface
      */
     private function convertVideoToImage($resource)
     {
-        $temporaryFilePath = $this->createTemporaryFile($resource);
-        $this->videoThumbnail->generate($temporaryFilePath, '00:00:02:01', $temporaryFilePath);
+        $source = $this->createTemporaryFile($resource);
+        $destination = tempnam(sys_get_temp_dir(), 'media');
+        $this->videoThumbnail->generate($source, '00:00:02:01', $destination);
 
-        $extractedImage = file_get_contents($temporaryFilePath);
-        unlink($temporaryFilePath);
+        $extractedImage = file_get_contents($destination);
+        unlink($source);
+        unlink($destination);
 
         return $this->createTemporaryResource($extractedImage);
     }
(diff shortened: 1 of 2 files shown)


thomasduenser:
@alexander-schranz @wachterjohannes Please review.

alexander-schranz (approved):
Looks good for me. We should make in some case a functional test newest ffmpeg versions agains our implementation.

alexander-schranz:
@thomasduenser thank you!

bopoda:
Hi guys,

thanks for fixing that.

I was debugging what is wrong with thumbnail generation from video after moving to a new server. I suppose the issue is with `FFmpeg` as you wrote:
> The wrong binary was used for ffprobe and the thumbnail generation for the newest FFmpeg version didn't work.

The code from `Sulu 1.6` generates the command like:
```
'/usr/bin/ffmpeg' '-y' '-ss' '00:00:02.01' '-i' '/tmp/mediaBinaryVideo1' '-vframes' '1' '-f' 'image2' '/tmp/mediaBinaryVideo1'
```

So, the same file is used as input and output (variable `$temporaryFilePath`): https://github.com/sulu/sulu/blob/1.6/src/Sulu/Bundle/MediaBundle/Media/ImageConverter/MediaImageExtractor.php#L156

FFmpeg fails with an error on a new server:
image

As you already fixed that,
Could you tell the easiest way to apply the fix to an existing `Sulu 1.6` project? I'm not really aware of Sulu. 
1. Migration to version `2.x+` is a good advice, but not now because of lack of time.
2. Some manipulations with `FFmpeg`. Not sure if that is a good idea as the newer, the better

alexander-schranz:
@bopoda A quickfix would be downgrade ffmpeg or overwrite the  https://github.com/sulu/sulu/blob/1.6/src/Sulu/Bundle/MediaBundle/Media/ImageConverter/MediaImageExtractor.php by defining your own `sulu_media.image.media_extractor` service: https://github.com/sulu/sulu/blob/538479c602b073bc0863a272e2adf63ebcb12918/src/Sulu/Bundle/MediaBundle/Resources/config/services.xml#L89-L93 which creates a temporary file like in this PR.
Its also possible to provide a PR for 1.6 with a fix, we are happy to review it.

bopoda:
@alexander-schranz thank you for the superfast reply!

>Its also possible to provide a PR for 1.6 with a fix, we are happy to review it.

Also, I will think about it :)

PR #4735, opened 2019-09-17
Add class for ToolbarAction

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR introduces a `ToolbarAction` class, which solves two problems:

* They can be serialized, and in the serialization process we can translate some translation keys, so that it is not necessary to translate stuff in the `Admin` class
* The type can be handled separately (the `ToolbarAction` class has a type and options)

#### Why?

Because we don't want to have the `translator` in every `Admin` class and the current array does not work when we want to have different `ToolbarAction`s be shown multiple times, because the key is used as the type, and if we add a second key it would override the value of the first one.
#### BC Breaks/Deprecations

Instead of an array you have to pass an array of `ToolbarAction` classes to the `addToolbarActions` method of the `RouteBuilder`s.

#### To Do

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Admin/Routing/DropdownToolbarAction.php
+++ b/src/Sulu/Bundle/AdminBundle/Admin/Routing/DropdownToolbarAction.php
@@ -1,0 +1,27 @@
+<?php
+
+/*
+ * This file is part of Sulu.
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\AdminBundle\Admin\Routing;
+
+class DropdownToolbarAction extends ToolbarAction
+{
+    public function __construct(string $label, string $icon, array $actions)
+    {
+        parent::__construct(
+            'sulu_admin.dropdown',
+            [
+                'label' => $label,
+                'icon' => $icon,
+                'actions' => $actions,
+            ]
+        );
+    }
+}
--- a/src/Sulu/Bundle/AdminBundle/Admin/Routing/DropdownToolbarAction.php
+++ b/src/Sulu/Bundle/AdminBundle/Admin/Routing/DropdownToolbarAction.php
@@ -13,14 +13,17 @@ namespace Sulu\Bundle\AdminBundle\Admin\Routing;
 
 class DropdownToolbarAction extends ToolbarAction
 {
-    public function __construct(string $label, string $icon, array $actions)
+    /**
+     * @param ToolbarAction[] $toolbarActions
+     */
+    public function __construct(string $label, string $icon, array $toolbarActions)
     {
         parent::__construct(
             'sulu_admin.dropdown',
             [
                 'label' => $label,
                 'icon' => $icon,
-                'actions' => $actions,
+                'toolbarActions' => $toolbarActions,
             ]
         );
     }
--- a/src/Sulu/Bundle/AdminBundle/Admin/Routing/ToolbarAction.php
+++ b/src/Sulu/Bundle/AdminBundle/Admin/Routing/ToolbarAction.php
@@ -1,0 +1,35 @@
+<?php
+
+/*
+ * This file is part of Sulu.
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\AdminBundle\Admin\Routing;
+
+use JMS\Serializer\Annotation\Groups;
+
+class ToolbarAction
+{
+    /**
+     * @var string
+     * @Groups({"frontend"})
+     */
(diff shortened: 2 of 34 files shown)


src/Sulu/Bundle/AdminBundle/Admin/Routing/DropdownToolbarAction.php
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\AdminBundle\Admin\Routing;
+
+class DropdownToolbarAction extends ToolbarAction
+{
+    public function __construct(string $label, string $icon, array $actions)
```
alexander-schranz: Arrays should have a phpdoc typehint with accepted types for static type checking:

```
/** 
 * @param ToolbarAction[] $actions
 */
public function __construct(string $label, string $icon, array $actions)
```

alexander-schranz: I personally would also name the variable $toolbarActions to be as specific as possible but think we should then also rename the option parameter in the js component.

src/Sulu/Bundle/AdminBundle/Serializer/Subscriber/DropdownToolbarActionSubscriber.php
```diff
@@ -0,0 +1,57 @@
...
+
+namespace Sulu\Bundle\AdminBundle\Serializer\Subscriber;
+
+use JMS\Serializer\EventDispatcher\Events;
+use JMS\Serializer\EventDispatcher\EventSubscriberInterface;
+use JMS\Serializer\EventDispatcher\ObjectEvent;
+use Sulu\Bundle\AdminBundle\Admin\Routing\DropdownToolbarAction;
+use Symfony\Component\Translation\TranslatorInterface;
+
+class DropdownToolbarActionSubscriber implements EventSubscriberInterface
```
alexander-schranz: I'm thinking if we should start mark this kind of services as `@internal` so IDE will tell the devs not using this class.

src/Sulu/Bundle/AdminBundle/Admin/Routing/ToolbarAction.php
```diff
@@ -0,0 +1,40 @@
...
+    /**
+     * @var array
+     * @Groups({"frontend"})
+     */
+    private $options;
+
+    public function __construct(string $type, array $options = [])
+    {
+        $this->type = $type;
+        $this->options = $options;
```
alexander-schranz: 'type'  can not be used as option or? should we do something like this here or on another place

```
if (array_key_exists('type', $options)) {
    throw new \DomainException(...);
}
```

src/Sulu/Bundle/AdminBundle/Resources/config/services.xml
```diff
@@ -121,5 +121,19 @@
         <service id="sulu_admin.schema_handler" class="Sulu\Bundle\AdminBundle\Serializer\Handler\SchemaHandler">
             <tag name="jms_serializer.subscribing_handler"/>
         </service>
+
+        <service id="sulu_admin.dropdown_toolbar_action_handler"
```
alexander-schranz: would be more specific with the id e.g. `sulu_admin.serializer_subscriber.dropdown...`

src/Sulu/Bundle/AdminBundle/Admin/Routing/ToolbarAction.php
danrot: No, that was the entire point of this operation: The `type` we have introduced here is not part of the options, so you can have another `type` in the options.

src/Sulu/Bundle/AdminBundle/Serializer/Subscriber/DropdownToolbarActionSubscriber.php
danrot: I think then we should add the `@internal` annotation a lot more often, and I am questioning if now is the right time to decide something like this.

src/Sulu/Bundle/AdminBundle/Resources/config/services.xml
danrot: Renamed the service to be callled `subscriber` at the end. Was experimenting with handlers, but that didn't work out well...

I am against introducing sub "namespaces", as this was already a mess in the old codebase. Just having the bundle prefixed with a dot seems like the cleaner solution to me.

src/Sulu/Bundle/AdminBundle/Admin/Routing/ToolbarAction.php
alexander-schranz: 👍

PR #4740, opened 2019-09-19
Add ContactAccountSelection

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR adds the `ContactAccountSelection` field type.

#### Why?

Because it was already possible to assign contacts and accounts in one place in the old UI.

#### To Do

- [x] Add reordering
- [x] Add remove functionality
- [x] Also remove when deselected in `MultiListOverlay`


src/Sulu/Bundle/ContactBundle/Resources/js/containers/ContactAccountSelection/ContactAccountSelection.js
```diff
@@ -0,0 +1,161 @@
...
+import {observer} from 'mobx-react';
+import {action, computed, observable, toJS} from 'mobx';
+import equals from 'fast-deep-equal';
+import {MultiItemSelection} from 'sulu-admin-bundle/components';
+import {MultiListOverlay} from 'sulu-admin-bundle/containers';
+import {translate} from 'sulu-admin-bundle/utils';
+import ContactAccountSelectionStore from './stores/ContactAccountSelectionStore';
+import contactAccountSelectionStyles from './contactAccountSelection.scss';
+
+// TODO extract into separate file?
```
danrot: These prefixes are need in this file and in the `ContactAccountSelectionStore`, should we extract them into a separate `constants.js` file? Or are there any other ideas?

src/Sulu/Bundle/ContactBundle/Resources/config/routing_api.yml
```diff
@@ -1,8 +1,3 @@
-sulu_contact.customer:
```
nnatter: 🎉

src/Sulu/Bundle/ContactBundle/Resources/js/containers/ContactAccountSelection/ContactAccountSelection.js
```diff
@@ -0,0 +1,158 @@
...
+import equals from 'fast-deep-equal';
+import {MultiItemSelection} from 'sulu-admin-bundle/components';
+import {MultiListOverlay} from 'sulu-admin-bundle/containers';
+import {translate} from 'sulu-admin-bundle/utils';
+import ContactAccountSelectionStore from './stores/ContactAccountSelectionStore';
+import contactAccountSelectionStyles from './contactAccountSelection.scss';
+
+type Props = {|
+    disabled: boolean,
+    onChange: (value: Array<Object>) => void,
```
nnatter: `Object` is basically an alias for `any` now. do we have any existing types for contacts or accounts?

src/Sulu/Bundle/ContactBundle/Resources/js/containers/ContactAccountSelection/ContactAccountSelection.js
```diff
@@ -0,0 +1,158 @@
...
+        this.callChange();
+    };
+
+    render() {
+        const {disabled, value} = this.props;
+
+        return (
+            <Fragment>
+                <MultiItemSelection
+                    disabled={disabled || false}
```
nnatter: should not be needed but 🤷‍♂

src/Sulu/Bundle/ContactBundle/Resources/js/containers/ContactAccountSelection/ContactAccountSelection.js
```diff
@@ -0,0 +1,158 @@
...
+                />
+                <MultiListOverlay
+                    adapter="table"
+                    listKey="accounts"
+                    onClose={this.handleOverlayClose}
+                    onConfirm={this.handleAccountConfirm}
+                    open={this.openedOverlay === 'accounts'}
+                    preloadSelectedItems={false}
+                    preSelectedItems={this.store.accountItems}
+                    resourceKey="accounts"
```
nnatter: i think we should extract global constants for our resource keys sometime. would be helpful for custom js too. but we dont need to do this now

src/Sulu/Bundle/ContactBundle/Resources/js/containers/ContactAccountSelection/ContactAccountSelection.js
```diff
@@ -0,0 +1,158 @@
...
+|};
+
+@observer
+class ContactAccountSelection extends React.Component<Props> {
+    static defaultProps = {
+        disabled: false,
+        value: [],
+    };
+
+    @observable openedOverlay: ?string = undefined;
```
nnatter: would propose `openedOverlayType`

src/Sulu/Bundle/ContactBundle/Resources/js/containers/ContactAccountSelection/ContactAccountSelection.js
```diff
@@ -0,0 +1,158 @@
...
+    };
+
+    @action handleOverlayClose = () => {
+        this.openedOverlay = undefined;
+    };
+
+    @action handleConfirm(items: Array<Object>, prefix: string) {
+        const {onChange, value} = this.props;
+
+        const itemIds = items.map((contact) => prefix + contact.id);
```
nnatter: `(item) => prefix + item.id`

src/Sulu/Bundle/ContactBundle/Resources/js/containers/ContactAccountSelection/ContactAccountSelection.js
```diff
@@ -0,0 +1,158 @@
...
+
+        onChange([
+            ...value.filter((id) => !id.startsWith(prefix) || itemIds.includes(id)),
+            ...itemIds,
+        ].filter((item, index, items) => index == items.indexOf(item)));
```
nnatter: wouldnt it be easier to just use `...itemIds.filter((id) => !value.includes(id))` above?

nnatter:
need a rebase here 🙂

src/Sulu/Bundle/ContactBundle/Resources/js/containers/ContactAccountSelection/ContactAccountSelection.js
danrot: No, we don't have any types at the moment... I know that it is currently an alias for `any`, but changing that now might take some time... And the annotation can at least be read as some kind of documentation now, so the developer knows what's at least expected here...

src/Sulu/Bundle/ContactBundle/Resources/js/containers/ContactAccountSelection/ContactAccountSelection.js
danrot: IIRC it is needed, because the `MultiItemSelection` component only accepts a boolean, and `disabled` could also be `undefined`.

src/Sulu/Bundle/ContactBundle/Resources/js/containers/ContactAccountSelection/ContactAccountSelection.js
danrot: I felt that it is easier to understand if I do that in two separate steps... Apart from that I am not totally sure if that would work :thinking:

danrot: Tested it, seems to work, will change it.

src/Sulu/Bundle/ContactBundle/Resources/js/containers/ContactAccountSelection/ContactAccountSelection.js
nnatter: you defined the type of the `disabled` prop of this component as non nullable `boolean` too, but i dont think this will lead to any problems 😉

PR #4748, opened 2019-09-23
Improve test bundle for easier usage in projects

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | yes
| Deprecations? | no
| Fixed tickets | fixes -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

 - Fix TestUserProvider with none plaintext password encoder
 - Fallback to correct UserProvider if other username then `test` is used
 - TestVoter will only allow all permissions for user `test`
 - Remove `sulu_test....` services aliases
 - Improve Message for AssertHttpStatusCode for redirect assertions.
 - Make FFmpeg only error when configured but not installed
 - Make `massive_search.adapter` test adapter public https://github.com/massiveart/MassiveSearchBundle/pull/133

#### Why?

So it can be used easier in a project.

#### BC Breaks/Deprecations

Some test services are removed.

#### To Do


src/Sulu/Bundle/TestBundle/Testing/AssertHttpStatusCodeTrait.php
```diff
@@ -31,23 +32,32 @@ protected static function assertHttpStatusCode(int $code, Response $response, in
 
         $message = '';
         if ($code !== $httpCode) {
-            $message = $response->getContent();
-
-            if ('null' !== ($json = json_encode(json_decode($message, true), JSON_PRETTY_PRINT))
-                && $json) {
-                $message = explode(PHP_EOL, $json);
+            if ($response->isRedirect()) {
```
danrot: Why are you doing that? What if I am expecting a redirect?

danrot:
And the tests are failing.

src/Sulu/Bundle/TestBundle/Testing/AssertHttpStatusCodeTrait.php
alexander-schranz: If its the correct status code it will not get to this position because of the if above.
If its the false status code it will end currently with:

```bash
Unexpected redirect to "/sitemaps/pages-1.xml" with status "301".
Failed asserting that 301 is identical to 302.
```

Should be change the text to something like:

```bash
Unexpected "301" status code with redirect to "/sitemaps/pages-1.xml".
Failed asserting that 301 is identical to 302.
```

Or whats your suggestion?

alexander-schranz: And why. Currently the output of a failed assertion looks like this:

```
HTTP status code 301 is not expected 302, showing 50 lines of the response body: <!DOCTYPE html>
<html>
    <head>
        <meta charset="UTF-8" />
        <meta http-equiv="refresh" content="0;url=/sitemaps/pages-1.xml" />

        <title>Redirecting to /sitemaps/pages-1.xml</title>
    </head>
    <body>
        Redirecting to <a href="/sitemaps/pages-1.xml">/sitemaps/pages-1.xml</a>.
    </body>
</html>
Failed asserting that 301 is identical to 302.
```

The HTML content is for redirects not needed so I want to simplify it and make the `URL` more readable

alexander-schranz:
@danrot can be reviewed again.

wachterjohannes:
@alexander-schranz looks good for me - if @danrot has no doubts he can merge it

PR #4760, opened 2019-09-26
Add profile and security contexts rest routes to api endpoint

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | yes
| Deprecations? | no
| Fixed tickets | fixes -
| License | MIT

#### What's in this PR?

Add rest routes to api endpoint

#### Why?

Rest endpoints should be all under `/admin/api` and the current definition in the `routing.yml` website does not work for older symfony version without the name_prefix and break currently when the container is build thats why in the travis ci I added the container build for both contexts.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -71,7 +71,9 @@ before_script:
   - composer update -o $COMPOSER_FLAGS
 
 script:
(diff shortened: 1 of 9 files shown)


danrot:
@chirimoya and I actually have decided that these endpoints are not part of the REST API, because `/api/profile` is a stateful API which is based on the session.

For me it is OK to change, but would like to hear @chirimoya opinion first.

PR #4762, opened 2019-09-27
Extend WebsiteController from AbstractController instead of deprecated Controller class

nnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This PR adjusts our default `WebsiteController` to extend the new `Symfony\Bundle\FrameworkBundle\Controller\AbstractController` class instead of the deprecated `Symfony\Bundle\FrameworkBundle\Controller\Controller` class.

This means that we need to configure the controller as a service from now on.

#### Why?

Because the `Symfony\Bundle\FrameworkBundle\Controller\Controller` class is deprecated and will be removed in the future.

#### BC Breaks/Deprecations

* `WebsiteController` extends `Symfony\Bundle\FrameworkBundle\Controller\AbstractController` now

### DIFF
--- a/src/Sulu/Bundle/WebsiteBundle/Controller/WebsiteController.php
+++ b/src/Sulu/Bundle/WebsiteBundle/Controller/WebsiteController.php
@@ -12,18 +12,20 @@
 namespace Sulu\Bundle\WebsiteBundle\Controller;
 
 use Sulu\Bundle\HttpCacheBundle\CacheLifetime\CacheLifetimeEnhancer;
+use Sulu\Bundle\HttpCacheBundle\CacheLifetime\CacheLifetimeEnhancerInterface;
 use Sulu\Bundle\PreviewBundle\Preview\Preview;
+use Sulu\Bundle\WebsiteBundle\Resolver\ParameterResolverInterface;
 use Sulu\Component\Content\Compat\StructureInterface;
-use Symfony\Bundle\FrameworkBundle\Controller\Controller;
+use Sulu\Component\Webspace\Analyzer\RequestAnalyzerInterface;
+use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\Exception\HttpException;
-use Twig\Template;
 
 /**
  * Basic class to render Website from phpcr content.
  */
-abstract class WebsiteController extends Controller
+abstract class WebsiteController extends AbstractController
 {
     /**
      * Returns a rendered structure.
@@ -110,8 +112,7 @@ abstract class WebsiteController extends Controller
         $twig = $this->get('twig');
         $attributes = $twig->mergeGlobals($attributes);
 
-        /** @var Template $template */
-        $template = $twig->loadTemplate($template);
+        $template = $twig->load($template);
 
         $level = ob_get_level();
         ob_start();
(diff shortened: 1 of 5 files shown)


src/Sulu/Bundle/WebsiteBundle/Controller/WebsiteController.php
```diff
@@ -161,4 +162,17 @@ protected function getCacheTimeLifeEnhancer(): ?CacheLifetimeEnhancer
...
         return $this->get('sulu_http_cache.cache_lifetime.enhancer');
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public static function getSubscribedServices()
+    {
+        $subscribedServices = parent::getSubscribedServices();
+        // $subscribedServices['sulu_website.resolver.parameter'] = ParameterResolverInterface::class;
```
nnatter: this method does not affect anything right now.. i guess there is some configuration missing, but this is kind of over my head 🤔

src/Sulu/Bundle/WebsiteBundle/Controller/WebsiteController.php
```diff
@@ -161,4 +162,17 @@ protected function getCacheTimeLifeEnhancer(): ?CacheLifetimeEnhancer
 
         return $this->get('sulu_http_cache.cache_lifetime.enhancer');
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public static function getSubscribedServices()
```
nnatter: this method should configure the minimal container that is passed to the controller. unfortunately apparently the method does not affect anything right now and the controller still receives the complete container 😢

nnatter: the only way that i found to solve this is adding `autowire: true` to the service definition of the container..

nnatter:
there are still some controllers left that extend the deprecated `Symfony\Bundle\FrameworkBundle\Controller\Controller` class:
* ResettingController
* MediaStreamController
* RedirectController
* MediaRedirectController
* CacheController
* LanguageController
* SnippetAreaController

nnatter:
Furthermore, the `FOS\RestBundle\Controller\FOSRestController` is deprecated too. This means that we will need to adjust out `Sulu\Component\Rest\RestController` that is the base class for most of our REST controllers 😨

src/Sulu/Bundle/WebsiteBundle/Controller/WebsiteController.php
nnatter: see https://github.com/symfony/symfony/issues/33755

alexander-schranz: @nnatter I think we should still merge this. But we should write into the `UPGRADE.md` that the services which are accessed need to define by adding an example code to the upgrade file.

src/Sulu/Bundle/WebsiteBundle/Resources/config/services.xml
```diff
@@ -28,6 +28,15 @@
     </parameters>
 
     <services>
+        <service id="sulu_website.default_controller" class="Sulu\Bundle\WebsiteBundle\Controller\DefaultController">
+            <tag name="container.service_subscriber" />
+            <tag name="controller.service_arguments" />
```
alexander-schranz: Would add here `<tag name="sulu.context" context="website" />` as this controller is only available in website context.

PR #4766, opened 2019-09-30
Remove check in javascript if a request is fired

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | none
| Related issues/PRs | fixes https://github.com/sulu/SuluArticleBundle/issues/395
| License | MIT
| Documentation PR | none

#### What's in this PR?

This PR removes the check if something is selected from JS and use the check in PHP which is already implemented there.


danrot:
Somehow every smart content now sends two requests...

PR #4767, opened 2019-09-30
Refactor deprecated controllers to use constructor injection for its dependencies

nnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Related issues/PRs | Builds upon https://github.com/sulu/sulu/pull/4762
| License | MIT

#### What's in this PR?

This PR refactors multiple controllers to use constructor injection instead of extending the deprecated `Symfony\Bundle\FrameworkBundle\Controller\Controller` class.

#### Why?

Because the `Symfony\Bundle\FrameworkBundle\Controller\Controller` class is deprecated and constructor injection is the preferred method for injecting dependencies now.

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/Resources/config/services.xml
+++ b/src/Sulu/Bundle/MediaBundle/Resources/config/services.xml
@@ -27,6 +27,13 @@
     </parameters>
 
     <services>
+        <service id="sulu_media.media_redirect_controller" class="Sulu\Bundle\MediaBundle\Controller\MediaRedirectController">
+            <argument type="service" id="sulu_media.media_manager" />
+
+            <tag name="controller.service_arguments" />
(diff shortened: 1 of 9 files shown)


src/Sulu/Bundle/MediaBundle/Resources/config/services.xml
```diff
@@ -27,6 +27,13 @@
     </parameters>
 
     <services>
+        <service id="sulu_media.media_redirect_controller" class="Sulu\Bundle\MediaBundle\Controller\MediaRedirectController">
+            <argument type="service" id="sulu_media.media_manager" />
+
+            <tag name="controller.service_arguments" />
```
nnatter: should we use `public` instead of the `controller.service_arguments` tag?

PR #4770, opened 2019-09-30
Refactor deprecated REST controllers to use constructor injection

nnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This PR refactors all Rest controllers to use constructor injection to gather their dependencies instead of extending the deprecated `FOS\RestBundle\Controller\FOSRestController` class.

#### Why?

Because the `FOS\RestBundle\Controller\FOSRestController` class is deprecated and constructor injection is the preferred method for injecting dependencies now.


src/Sulu/Component/Rest/RestController.php
```diff
@@ -11,18 +11,36 @@
...
 use Sulu\Component\Rest\Exception\EntityNotFoundException;
 use Sulu\Component\Rest\Exception\RestException;
 use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface;
 
 /**
  * Abstract Controller for extracting some required rest functionality.
  */
-abstract class RestController extends FOSRestController
+abstract class RestController
```
nnatter: we should not change this class because most of our bundles extend this too. a better solution would be deprecating this class and add a new non-deprecated base class for the refactored controllers.

src/Sulu/Component/Rest/Tests/Unit/RestControllerTrait.php
```diff
@@ -0,0 +1,249 @@
+<?php
+
+namespace Sulu\Component\Rest\Tests\Unit;
```
nnatter: i think we should move this 🙈

src/Sulu/Bundle/SecurityBundle/Controller/UserController.php
```diff
@@ -11,43 +11,87 @@
...
-use Sulu\Component\Rest\RestController;
 use Sulu\Component\Rest\RestHelperInterface;
 use Sulu\Component\Security\SecuredControllerInterface;
 use Symfony\Component\HttpFoundation\Request;
 
 /**
  * Makes the users accessible through a rest api.
  */
-class UserController extends RestController implements ClassResourceInterface, SecuredControllerInterface
+class UserController extends AbstractRestController implements ClassResourceInterface, SecuredControllerInterface
```
danrot: Somehow I cannot create a user with a system language, I don't have the permissions for anymore :thinking:

src/Sulu/Component/Rest/AbstractRestController.php
```diff
@@ -0,0 +1,55 @@
...
+    {
+        if (!$this->tokenStorage) {
+            throw new \LogicException('The TokenStorage property was not set via the constructor".');
+        }
+
+        if (null === $token = $this->tokenStorage->getToken()) {
+            return null;
+        }
+
+        if (!\is_object($user = $token->getUser())) {
```
danrot: We usually don't do the assignment in an if... is that really necessary? Would move it outside and write just `is_object($user)` here.

PR #4777, opened 2019-10-01
Fix sitemap dumper

### LINKED ISSUE #4094: Dumping XML sitemap should dump for domain not for locales or webspaces
| Q | A
| --- | ---
| Bug? | yes
| New Feature? | no
| Sulu Version | 1.6.21
| Browser Version | Chrome 67.0.3396.99 (Official Build) (64-bit)

#### Actual Behavior

XML sitemap dumper generates `sitemap.xml` files for each webspace and each locale.

#### Expected Behavior

There should only be one sitemap: for the domain. Because as discussed with @alexander-schranz on Slack:

* sitemap.xml already contains links to all alternative languages, so there is no need for a separate `sitemap.xml` per locale
* there can be multiple webspaces on a single domain

@alexander-schranz has stated the only thing that would make sense is if it would output `var/sitemaps/https/domain.xml`

#### Steps to Reproduce

Given the following portal configuration:

```
    <portals>
        <portal>
            <name>mysite.nl</name>
            <key>mywebspace</key>

            <environments>
                <environment type="prod">
                    <urls>
                        <url>{host}/{localization}</url>
                        <url language="region">{host}</url>
                    </urls>
                </environment>
                <environment type="stage">
                    <urls>
                        <url>{host}/{localization}</url>
                        <url language="region">{host}</url>
                    </urls>
                </environment>
                <environment type="test">
                    <urls>
                        <url>{host}/{localization}</url>
                        <url language="region">{host}</url>
                    </urls>
                </environment>
                <environment type="dev">
                    <urls>
                        <url>{host}/{localization}</url>
                        <url language="region">{host}</url>
                    </urls>
                </environment>
            </environments>
        </portal>
    </portals>
```

Then running the command:

```
$ bin/websiteconsole sulu:website:dump-sitemap --https
```

Will output:

```
Start dumping "sitemap.xml" files:
 - mywebspace (nl_nl)
 - mywebspace (nl_be)
 - mywebspace (es_es)
 - mywebspace (fr_fr)
 - mywebspace (fr_lu)
 - mywebspace (fr_ch)
 - mywebspace (fr_be)
 - mywebspace (en_gb)
 - mywebspace (en_ie)
 - mywebspace (de_de)
 - mywebspace (de_at)
 - mywebspace (de_ch)
 - mywebspace (de_lu)
```

Resulting in these files:

```
var/sitemaps/https/mywebspace/nl_nl/mysite.nl/sitemap.xml
var/sitemaps/https/mywebspace/nl_be/mysite.nl/sitemap.xml
.. etc etc
```

#### Possible Solutions

* `DumpSitemapCommand` should dump only for the default localization, instead of iterating through all localizations
* `XmlSitemapDumper::getDumpPath()` should be adjusted to only dump to a domain.xml
* The call from `SitemapController` to `XmlSitemapDumper::getDumpPath()` also needs to be adjusted so it can find the cached file

### LINKED ISSUE #4327: SitemapProvider - Disable creating sitemap for particular portal
| Q | A
| --- | ---
| Bug? | no
| New Feature? | no
| Sulu Version | 1.6
| Browser Version | All

#### Actual Behavior

I use SitemapProvider to generate custom sitemap for particular portal.
Custom sitemap is created for all site portals. But when we have `<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9" xmlns:xhtml="http://www.w3.org/1999/xhtml"/>`, Google Search Console give us a syntax Xml error.

#### Expected Behavior

To not created custom sitemap if i don't need to avoid error in Google Search Console.

#### Steps to Reproduce

Implements SitemapProvider.

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | yes
| Deprecations? | no
| Fixed tickets | fixes #4094 fixes #4327
| Related issues/PRs | -
| License | MIT
| Documentation PR | https://github.com/sulu/sulu-docs/pull/456

#### What's in this PR?

Fix sitemap dumper by going on host and not on a single portal or language.

#### Why?

Currently the sitemap dumper does not work as it will dump only for once portal and one locale. For this several interfaces in the sitemap namespace need to be changed to only have 

#### To Do

- [x] Documentation


src/Sulu/Component/Webspace/Manager/WebspaceManager.php
```diff
@@ -98,6 +98,34 @@ public function findPortalInformationByUrl(string $url, ?string $environment = n
         return null;
     }
 
+    public function findPortalInformationsByHost(
+        string $host,
+        bool $includeSubdomains = false,
```
danrot: Is that parameter needed? I have never seen this function being called with a different value than `true`...

alexander-schranz:
@danrot I moved the service tag alias to a `getAlias` function on the SitemapProviderInterface, so its now autowireable. Docs are also updated: https://github.com/sulu/sulu-docs/pull/456

src/Sulu/Bundle/WebsiteBundle/Sitemap/Provider/PagesSitemapProvider.php
```diff
@@ -35,75 +38,136 @@ class PagesSitemapProvider implements SitemapProviderInterface
...
     }
 
+    private function generateSitemapUrl(
+        Content $contentPage,
+        PortalInformation $portalInformation,
+        string $host,
+        string $scheme
+    ) {
+        $changed = $contentPage['changed'];
+        if (is_string($changed)) {
```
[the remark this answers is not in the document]
alexander-schranz: should only be string here. Seems like new \DateTime is only coming from the UnitTests

UPGRADE.md
```diff
@@ -1,5 +1,65 @@
 # Upgrade
 
+## unreleased
+
+### Sitemap Provider changed
+
+As a sitemap is always domain specific and a domain can have multiple webspaces and portal
+the `SitemapProviderInterface` changed. To make it possible to provide also none Sulu
+routes the `SitemapUrl` constructor changed and now needs as third paramter a defaultLocale.
```
danrot: Use backticks for `defaultLocale` and you've got a typo in "parameter". I would also change the sentence to "constructor introduced a `defaultLocale` as third parameter"

UPGRADE.md
```diff
@@ -1,5 +1,65 @@
...
+public function getAlias(): string {
+    return 'myalias';
+}
+
+public function getMaxPage($scheme, $host) {
+     return 1;
+}
+```
+
+In this case also the `XmlSitemapDumper` and `XmlSitemapRender` does not longer need PortalInformations.
```
danrot: "In this case" does not really fit here... How about "The `XmlSitemapDumper` and `XmlSitemapRenderer` no longer need `PortalInformations`."

src/Sulu/Component/Webspace/Manager/WebspaceManager.php
danrot: Then I would rename the function, instead of introducing a basically unused parameter. How about `findPortalInformationsByHostIncludingSubdomains`? Would also make the calling code easier to understand, otherwise you always have to lookup what this boolean flag means...

PR #4787, opened 2019-10-04
Fix optional ffmpeg configuration

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Removes the php-ffmpeg config from the TestKernel and adds it to the MediaBundle test configuration.

#### Why?

Because the dependency is only used in the MediaBundle and should not be required in all the other bundles.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -92,7 +92,6 @@
         "league/flysystem-aws-s3-v3": "^1.0.1",
         "matthiasnoback/symfony-config-test": "^4.0",
         "matthiasnoback/symfony-dependency-injection-test": "^4.0",
-        "php-ffmpeg/php-ffmpeg": "^0.13 || ^0.14",
         "phpcr/phpcr-shell": "^1.0",
         "phpspec/prophecy": "^1.7",
         "phpstan/phpstan": "^0.11.5",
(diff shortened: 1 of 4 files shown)


alexander-schranz:
@Prokyonn this is an optional dependency and it is used to create thumbnails and get detail information out of video files. And its also used by the community #4357.

composer.json
```diff
@@ -92,7 +92,6 @@
         "league/flysystem-aws-s3-v3": "^1.0.1",
         "matthiasnoback/symfony-config-test": "^4.0",
         "matthiasnoback/symfony-dependency-injection-test": "^4.0",
-        "php-ffmpeg/php-ffmpeg": "^0.13 || ^0.14",
```
alexander-schranz: should stay here :)

src/Sulu/Bundle/TestBundle/Resources/app/config/sulu.yml
```diff
@@ -54,11 +54,6 @@ jms_serializer:
     metadata:
         debug: '%kernel.debug%'
 
-sulu_media:
```
alexander-schranz: should be moved to the tests Application of the MediaBundle

PR #4795, opened 2019-10-07
Fix account interface function types

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | yes
| Deprecations? | no
| Fixed tickets | fixes -
| Related issues/PRs | https://github.com/sulu/sulu/pull/3942
| License | MIT
| Documentation PR | -

#### What's in this PR?

Allow set null of some fields on the AccounterInterface and remove the deprecated notes function from the interface.

#### Why?

With the AccountInterface refractoring of https://github.com/sulu/sulu/pull/3942 its not longer possible to unset some fields on the account. Sadly this is a bc break, but think we need accept this bc break to fix this bug.

#### BC Breaks/Deprecations

AccountInterface changed.

#### To Do

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -2,6 +2,27 @@
 
 ## dev-release/2.0
 
+### Fix AccountInterface nullable setters/getter
+
+Setters and getter of nullable fields on the Account entity were fixed.
+For some function on the AccountInterface need to be changed to the following:
+
+```php
+// Before
+public function setExternalId(string $externalId): AccountInterface;
+public function setNumber(string $number): AccountInterface;
+public function setRegisterNumber(string $registerNumber): AccountInterface;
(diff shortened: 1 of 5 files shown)


UPGRADE.md
```diff
@@ -1,5 +1,28 @@
...
+Setters and getter of nullable fields on the Account entity were fixed.
+For some function on the AccountInterface need to be changed to the following:
+
+```php
+// Before
+public function setExternalId(string $externalId): AccountInterface;
+public function setNumber(string $number): AccountInterface;
+public function setRegisterNumber(string $registerNumber): AccountInterface;
+public function setPlaceOfJurisdiction(string $placeOfJurisdiction): AccountInterface;
+public function setNote(?string $note): AccountInterface;
```
danrot: Looks the same as below

alexander-schranz: fixed

PR #4797, opened 2019-10-08
Fix symfony session regression

danrot:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | symfony/symfony#33487
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR fixes some regressions that were introduced by Symfony.

#### Why?

Because the tests were failing.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,7 @@
         "stof/doctrine-extensions-bundle": "^1.2.2",
         "gedmo/doctrine-extensions": "~2.4",
         "symfony-cmf/routing-bundle": "^1.2 || ^2.0",
-        "symfony/symfony": "^2.8.11 || ^3.0",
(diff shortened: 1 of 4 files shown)


composer.json
```diff
@@ -54,6 +54,7 @@
         "matthiasnoback/symfony-dependency-injection-test": "^0.7.3",
         "phpunit/phpunit": "^4.8.31",
         "phpspec/prophecy": "^1.4",
+        "symfony/symfony": "^2.8.51 || ^3.0",
```
danrot: This seems to fix the issue we are currently having in our tests... Altough if I only put it in `require-dev` we have some symfony versions not being tested anymore... should we just raise the minimum version in the real `require` section?

PR #4801, opened 2019-10-09
Show history URL for custom urls

danrot:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR fixes the history button of the custom urls.

#### Why?

Because the history url was always shown as empty.
#### BC Breaks/Deprecations

Yes, and not totally sure if we should do them, or already start having multiple paths in our code. BC breaks are in the `ResourceLocatorHistory` JS component (using the `resourcelocator` field instead of `path`), and in the `PageResourcelocatorController` (returns a `resourcelocator` field instead of `path`).

#### To Do

- [ ] Clarify BC breaks

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/ResourceLocatorHistory/ResourceLocatorHistory.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/ResourceLocatorHistory/ResourceLocatorHistory.js
@@ -96,7 +96,7 @@ class ResourceLocatorHistory extends React.Component<Props> {
                                 <Table.Body>
                                     {historyRoutes.map((historyRoute) => (
                                         <Table.Row id={historyRoute.id} key={historyRoute.id}>
-                                            <Table.Cell>{historyRoute.path}</Table.Cell>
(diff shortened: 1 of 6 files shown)


danrot:
@sulu/core-team WDYT about the BC breaks I've introduced here? The problem was that there are different properties on the response from the custom url history and page resource locator (`path` vs `resourcelocator`) so I made it `resourcelocator` everywhere, because it was the better match IMO (the URL shown in the custom urls does not qualify as a path IMO).

The alternative would be to return both properties in the `PageResourcelocatorController` or have a fallback from `resourcelocator` to `path` in the `ResourceLocatorHistory` js component.

danrot:
One reason I would prefer to make it not configurable, is to make sure that the APIs stay even consistent in the future :slightly_smiling_face: If we make it configurable I am sure that it will get inconsistent sooner or later.

alexander-schranz:
@danrot sounds good for me. And the fix is working for me.

Issue #399, opened 2014-11-04
Users should not be able to delete themselves


sulu-io-user:
<a href="https://github.com/turbo-ele"><img height="50" width="50" src="https://avatars.githubusercontent.com/u/5152633?v=2" /></a>
By @turbo-ele _from_ https://github.com/sulu-cmf/SuluContactBundle/issues/43

User should not be able to delete himself / herself / itself

PR #4802, opened 2019-10-09
Add template helper for underscore templates

luca-rath:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Adds a templateHelper utility class

#### Why?

By default, underscore's template engine has access to all variables in global scope to evaluate a template. With this helper, all data, which can be used by the template engine will be encapsulated in a local scope.

#### Example Usage

~~~php
define(['suluadmin/utils/templateHelper'], function(templateHelper) {
    'use strict';

    var url = _.template('/admin/api/test' +
        '<% if (typeof data.id !== "undefined") { %>/<%= data.id %><% } %>');

    return {
        url: templateHelper.transformTemplateData(url)
    };
});
~~~


alexander-schranz:
@luca-rath here seems to be a build missing.

danrot:
@alexander-schranz Since @luca-rath and me came up with this idea, I would prefer somebody else doing the review.

alexander-schranz:
@danrot @luca-rath can this be closed as #399 is merged and released?

PR #4803, opened 2019-10-09
Fix SingleAccountSelection ContentType

danrot:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR fixes the type mismatch between the `single_account_selection` field type and the ´SingleAccountSelection` content type.

#### Why?

Because otherwise the `single_account_selection` field type throws an error when used on a page.

### DIFF
--- a/src/Sulu/Bundle/ContactBundle/Content/Types/SingleAccountSelection.php
+++ b/src/Sulu/Bundle/ContactBundle/Content/Types/SingleAccountSelection.php
@@ -11,14 +11,15 @@
 
 namespace Sulu\Bundle\ContactBundle\Content\Types;
 
+use PHPCR\NodeInterface;
 use Sulu\Bundle\ContactBundle\Api\Account;
 use Sulu\Bundle\ContactBundle\Contact\AccountManager;
 use Sulu\Bundle\WebsiteBundle\ReferenceStore\ReferenceStoreInterface;
 use Sulu\Component\Content\Compat\PropertyInterface;
+use Sulu\Component\Content\ComplexContentType;
 use Sulu\Component\Content\PreResolvableContentTypeInterface;
-use Sulu\Component\Content\SimpleContentType;
 
-class SingleAccountSelection extends SimpleContentType implements PreResolvableContentTypeInterface
+class SingleAccountSelection extends ComplexContentType implements PreResolvableContentTypeInterface
 {
     /**
      * @var AccountManager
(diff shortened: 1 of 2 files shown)


src/Sulu/Bundle/ContactBundle/Content/Types/SingleAccountSelection.php
```diff
@@ -36,19 +37,59 @@ public function __construct(
     ) {
         $this->accountManager = $accountManager;
         $this->accountReferenceStore = $accountReferenceStore;
+    }
+
+    public function read(NodeInterface $node, PropertyInterface $property, $webspaceKey, $languageCode, $segmentKey)
+    {
+        $account = null;
```
alexander-schranz: There should be more error checking if the account really exist. And read currently does always return the value so I would do this:

```
        $value = null;

        if ($node->hasProperty($property->getName())) {
            $account = $this->accountManager->getById(
                $node->getPropertyValue($property->getName()),
                $property->getStructure()->getLanguageCode()
            );

            if ($account) {
                $value = [
                    'id' => $account->getId(),
                    'name' => $account->getName(),
                ];
            }
        }

        $property->setValue($value);

        return $value;
```

src/Sulu/Bundle/ContactBundle/Content/Types/SingleAccountSelection.php
```diff
@@ -11,14 +11,15 @@
...
 use Sulu\Bundle\ContactBundle\Api\Account;
 use Sulu\Bundle\ContactBundle\Contact\AccountManager;
 use Sulu\Bundle\WebsiteBundle\ReferenceStore\ReferenceStoreInterface;
 use Sulu\Component\Content\Compat\PropertyInterface;
+use Sulu\Component\Content\ComplexContentType;
 use Sulu\Component\Content\PreResolvableContentTypeInterface;
-use Sulu\Component\Content\SimpleContentType;
 
-class SingleAccountSelection extends SimpleContentType implements PreResolvableContentTypeInterface
+class SingleAccountSelection extends ComplexContentType implements PreResolvableContentTypeInterface
```
alexander-schranz: Is this not the same as for the `SingleContactSelection` think there we need to change it also?

danrot: I tested both field types, and the `single_contact_selection` was working, so I didn't touch it. But IIRC for the contacts it's not an `AutoComplete`, so that's probably the difference.

alexander-schranz: @danrot ok that make sense.

src/Sulu/Bundle/ContactBundle/Content/Types/SingleAccountSelection.php
danrot: Not all `read` methods return a value, so I assumed that the value is not used by the caller at all... And if this is the case I think it is really confusing to return something...

And the other thing is that an exception is thrown if the account is not existing... But gonna fix it anyway.

PR #4804, opened 2019-10-09
Fix custom serialization handler for iterator classes

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | https://github.com/sulu/SuluArticleBundle/pull/396
| License | MIT

#### What's in this PR?

Fix custom serialization handler for iterator classes.

#### Why?

Currently this handlers don't work as expected. 

Implementation is similiar to:

https://github.com/schmittjoh/serializer/blob/4c1e4296734385af7718ca71ec0febb4815b4a87/src/Handler/IteratorHandler.php#L104-L110

#### TODO

 - [ ] Tests


danrot:
I don't exactly understand this change, but ok... Can somebody with an ArticleBundle installation test if that works? Looks to me like you have done this because of the ArticleBundle...

alexander-schranz:
@danrot seems like this PR is really obsolete as I didn't need to change the ChildrenCollection for the ArticleBundle. And the serialization in example the ExtensionContainer seems to work correctly in the Page Serialization with the navigator accept.

PR #4809, opened 2019-10-11
Show error when deleting parent account

danrot:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR introduces an error message when an account which is the parent of another account is trying to be deleted.

#### Why?

Because we don't want to delete the child accounts when a parent account is deleted. What should happen instead is not completely clear, and this way the user can decide how to resolve this conflict.

And it is the same behavior as in Sulu 1.6.

#### To Do

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,6 +1,6 @@
 # Upgrade
 
-## unreleased
+## dev-release/2.0
 
 ### Sitemap Provider changed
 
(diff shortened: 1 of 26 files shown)


danrot:
IMO there is currently only one issue: When the error message is shown, deletion cannot be continued, but the dialog still shows two buttons: "Ok" and "Cancel". I think one of both should be gone, but I am not completely sure which one. @chirimoya Do you have any idea?

src/Sulu/Bundle/ContactBundle/Entity/AccountRepository.php
```diff
@@ -330,6 +330,24 @@ public function findChildrenAndContacts(int $id): ?AccountInterface
...
+        try {
+            $qb = $this->createQueryBuilder('account')
+                ->where('account.parent = :accountId');
+
+            $query = $qb->getQuery();
+            $query->setParameter('accountId', $id);
+
+            return $query->getResult();
+        } catch (NoResultException $ex) {
+            return null;
```
nnatter: i think i would expect an empty array in this case

UPGRADE.md
```diff
@@ -1,5 +1,40 @@
 # Upgrade
 
+## dev-release/2.0
+
+### DeleteToolbarAction with conflict
+
+The `DeleteToolbarAction` asks for confirmation if e.g. a page that is being tried to deleted is linked on other pages.
```
nnatter: not sure if `being tried to deleted` is correct. but dont have a better suggestion and i guess people will understand it anyway 🙂

src/Sulu/Bundle/ContactBundle/Entity/AccountRepository.php
danrot: You are right, that's better :slightly_smiling_face: Done.

PR #4811, opened 2019-10-11
Add MetadataRequestParameters to FormView and FormOverlayListView

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Adds the possibility to add custom metadata to the form of the FormOverlay and the FormView.

#### Why?

Because we want the possibility to add custom metadata to views.

##### ViewBuilder
~~~php
        $formOverlayListViewBuilder->setResourceKey(Example::RESOURCE_KEY)
            ->setListKey(Example::RESOURCE_KEY)
            ->setFormKey('example_details')
            ->addListAdapters(['table'])
            ->addMetadataRequestParameters(['test-data' => $data]);
~~~

##### CustomMetadataLoader
~~~php
class CustomMetadataLoader implements FormMetadataLoaderInterface
{
    public function getMetadata(string $key, string $locale, array $metadataOptions): ?MetadataInterface
    {
       //do something with the metadata
       $testData = $metadataOptions['test-data'];
       //...
   }
}
~~~

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Admin/View/FormOverlayListViewBuilder.php
+++ b/src/Sulu/Bundle/AdminBundle/Admin/View/FormOverlayListViewBuilder.php
@@ -186,9 +186,9 @@ class FormOverlayListViewBuilder implements FormOverlayListViewBuilderInterface
         return $this;
     }
 
-    public function addFormMetadata(array $formMetadata): FormOverlayListViewBuilderInterface
+    public function addMetadataRequestParameters(array $formMetadata): FormOverlayListViewBuilderInterface
     {
-        $this->addFormMetadataToView($this->view, $formMetadata);
+        $this->addMetadataRequestParametersToView($this->view, $formMetadata);
 
         return $this;
     }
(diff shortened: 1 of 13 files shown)


src/Sulu/Bundle/AdminBundle/Admin/View/FormOverlayListViewBuilder.php
```diff
@@ -186,6 +186,13 @@ public function setOverlaySize(string $overlaySize): FormOverlayListViewBuilderI
         return $this;
     }
 
+    public function addFormMetadata(array $formMetadata): FormOverlayListViewBuilderInterface
```
danrot: @sulu/core-team WDYT about this naming? Currently we have only `addRouterAttributesToListMetadata` being similar. There is also `setRequestParameters` on the `FormViewBuilder`, which is similar in regard that the options being passed are hardcoded.

Should we leave the short version `addFormMetadata` here or go for something like `addParametersToFormMetadata`?

nnatter: this adds options that are appended to the metadata request right? that would be similar to `setRequestParameters`, but for another request. what do you think about `setMetadataRequestParameters`?

danrot:
@Prokyonn While we're at it, don't you think it would make sense to introduce that in the `Form` view as well? I am sure somebody would ask for something like that sooner or later, and then it is hard to explain why it is not there.

Issue #4618, opened 2019-07-04
Deleting a page breaks the "webspaces" view


ctmarc:
| Q | A
| --- | ---
| Bug? | yes
| New Feature? | no
| Sulu Version | sulu-minimal / dev-develop
| Browser Version | All Webbrowsers

#### Actual Behavior

When deleting a page from the webspace `mywebspace`, the `se_user_settings` table still shows the deleted PHPCR ID in `sulu_admin.list_store.pages.webspace_overview_mywebspace-prod.active` for the active user ID.

The Webspaces view will stop loading because of a 404 error on: `/admin/api/pages?exclude-ghosts=false&exclude-shadows=false&locale=en&webspace=mywebspace-prod&expandedIds=832b1249-6539-461d-8a90-be9ca0d53e37&fields=title%2Cpublished%2Cid&flat=true`

Replacing the value in the `se_user_settings` table or in the URL by a valid node ID fixes the problem.

#### Expected Behavior

The setting `sulu_admin.list_store.pages.webspace_overview_mywebspace-prod.active` should be updated in `se_user_settings` when a node is deleted from the sitemap.

#### Steps to Reproduce

- Login to Sulu Backend
- Go to "Webspaces" and delete any child page
- Reload the browser / the "Webspaces" view

danrot:
This has already been fixed in #4550.

PR #4814, opened 2019-10-14
Fix missing webspace when page is deleted in form

danrot:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR uses `restore` when going back to the list after deleting an item in the form.

#### Why?

That avoids an error when a webspace is missing after deleting a page, if the webspace has never been chosen manually. If this is done before the fix, then the webspace select is empty.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/views/Form/toolbarActions/DeleteToolbarAction.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/views/Form/toolbarActions/DeleteToolbarAction.js
@@ -68,7 +68,7 @@ export default class DeleteToolbarAction extends AbstractFormToolbarAction {
     navigateBack = () => {
         const {backView} = this.router.route.options;
         const {locale} = this.resourceFormStore;
-        this.router.navigate(backView, {locale: locale ? locale.get() : undefined});
+        this.router.restore(backView, {locale: locale ? locale.get() : undefined});
     };
 
     @action handleCancel = () => {
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/views/Form/toolbarActions/DeleteToolbarAction.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/views/Form/toolbarActions/DeleteToolbarAction.js
@@ -66,9 +66,33 @@ export default class DeleteToolbarAction extends AbstractFormToolbarAction {
     }
 
     navigateBack = () => {
-        const {backView} = this.router.route.options;
+        const {attributes, route} = this.router;
+        const {backView} = route.options;
         const {locale} = this.resourceFormStore;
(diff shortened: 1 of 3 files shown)


alexander-schranz:
@danrot does this fix #4618?

danrot:
@alexander-schranz No, but this other issue is already solved, so I closed it :slightly_smiling_face:

src/Sulu/Bundle/AdminBundle/Resources/js/views/Form/toolbarActions/DeleteToolbarAction.js
```diff
@@ -68,7 +68,7 @@ export default class DeleteToolbarAction extends AbstractFormToolbarAction {
     navigateBack = () => {
         const {backView} = this.router.route.options;
         const {locale} = this.resourceFormStore;
-        this.router.navigate(backView, {locale: locale ? locale.get() : undefined});
+        this.router.restore(backView, {locale: locale ? locale.get() : undefined});
```
alexander-schranz: @danrot if I reload on the detail form of the page. And then remove the page it will error because it can not restore it I think.

danrot: I see multiple solutions to this problem:

1. We use the [`getDerivedRouteAttributes` function of the `WebspaceTabs`](https://github.com/sulu/sulu/blob/edbf2d5/src/Sulu/Bundle/PageBundle/Resources/js/views/WebspaceTabs/WebspaceTabs.js#L27) to link to a different webspace it is missing. Problem is that we currently have the `WebspaceStore`, which only offers async functions, which is not supported in this function... So we would have to change to a `WebspaceRegistry` without an async workflow, which would be a BC break, and load all webspaces initially (something that we originally wanted to avoid, but I am not sure if that is worth all the hassle...)
2. We blindly add all available options here, but I am afraid of all the possible side effects caused by that.
3. We introduce one of our loved `routerAttributesToBackView` options on this `ToolbarAction` as well.

The first approach would have the nice side effect that other code also gets easier, and the list would always show something, but I think the use case you are describing here could potentially lead to a wrong webspace.

Therefore I would go for the third option. Maybe we should also ask for the opinion of @sulu/core-team, because that's a very generic approach, which might spread further to other `ToolbarAction`s.

danrot: After a short discussion with @alexander-schranz and @nnatter we've decided to go for option 3, because it is not too much work, and limits the possible side effects.

PR #4823, opened 2019-10-16
Use exclude_duplicates parameter of SmartContent

danrot:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR makes use of the already existing `exclude_duplicates` parameter.

#### Why?

Because the preview in the admin showed the correct entries including this option, but the field itself in the form did not.

#### Example Usage

~~~xml
        <property name="smart" type="smart_content">
            <params>
                <param name="exclude_duplicates" value="true"/>
            </params>
        </property>

        <property name="smart2" type="smart_content">
            <params>
                <param name="exclude_duplicates" value="true"/>
            </params>
        </property>

        <property name="smart3" type="smart_content">
            <params>
                <param name="exclude_duplicates" value="true"/>
            </params>
        </property>

        <property name="smart4" type="smart_content">
            <params>
                <param name="exclude_duplicates" value="true"/>
            </params>
        </property>
~~~

#### To Do

- [x] Check if working without `exclude_duplicates` as well
- [x] Check if mixed providers are working (e.g. some contact and some media smart contents)
- [x] Maybe replace `previousSmartContentStore` in `smart_content` field with a more general function in `smartContentStorePool`

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/SmartContent.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/SmartContent.js
@@ -1,10 +1,11 @@
 // @flow
 import React from 'react';
-import {autorun, toJS} from 'mobx';
+import {autorun, computed, toJS, when} from 'mobx';
 import equals from 'fast-deep-equal';
 import type {FieldTypeProps} from '../../../types';
 import SmartContentComponent, {smartContentConfigStore, SmartContentStore} from '../../SmartContent';
 import type {FilterCriteria} from '../../SmartContent/types';
+import smartContentStorePool from './smartContentStorePool';
 
 type Props = FieldTypeProps<?FilterCriteria>;
 
@@ -22,8 +23,22 @@ const filterCriteriaDefaults = {
     tags: undefined,
 };
 
-export default class SmartContent extends React.Component<Props> {
+class SmartContent extends React.Component<Props> {
     smartContentStore: SmartContentStore;
+    filterCriteriaChangeDisposer: () => void;
+
+    @computed get previousSmartContentStores() {
+        const previousSmartContentStores = [];
+        for (const smartContentStore of smartContentStorePool.stores) {
+            if (smartContentStore === this.smartContentStore) {
+                break;
+            }
+
+            previousSmartContentStores.push(smartContentStore);
+        }
+
+        return previousSmartContentStores;
+    }
 
     constructor(props: Props) {
         super(props);
@@ -31,6 +46,9 @@ export default class SmartContent extends React.Component<Props> {
         const {
             formInspector,
             schemaOptions: {
+                exclude_duplicates: {
+                    value: excludeDuplicates,
+                } = {},
                 provider: {
                     value: provider,
                 } = {value: 'pages'},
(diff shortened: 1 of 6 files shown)


danrot:
Looks like until now the `exclude_duplicates` functionality was only implemented for pages until now... The other providers ignore the excluded flag on the server side as well...

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/tests/fields/SmartContent.test.js
```diff
@@ -65,8 +79,76 @@ test('Should correctly initialize SmartContentStore', () => {
...
+
+    expect(smartContentStorePool.add).toBeCalledWith(smartContentStore, false);
     expect(smartContentConfigStore.getConfig).toBeCalledWith('media');
     expect(SmartContentStore).toBeCalledWith('media', value, undefined, 'collections', undefined);
+
+    smartContent.unmount();
+    expect(smartContentStorePool.remove).toBeCalledWith(smartContentStore);
+});
+
+test('Should correctly initialize SmartContentStore with a negative exclude_duplicates values', () => {
```
nnatter: `negative`?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/SmartContent.js
```diff
@@ -52,11 +65,27 @@ export default class SmartContent extends React.Component<Props> {
...
 
-        autorun(this.handleFilterCriteriaChange);
+        smartContentStorePool.add(this.smartContentStore, excludeDuplicates);
+
+        this.filterCriteriaChangeDisposer = autorun(this.handleFilterCriteriaChange);
+
+        if (!excludeDuplicates || this.previousSmartContentStores.length === 0) {
+            this.smartContentStore.start();
+        } else {
+            when(
```
nnatter: is this `when()` necessary here? i think the `updateRecursiveExcludedIds` method of the `smartContentSotrePool` will already handle this

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/SmartContent.js
```diff
@@ -52,11 +65,27 @@ export default class SmartContent extends React.Component<Props> {
...
+                () => this.previousSmartContentStores.every((store) => !store.itemsLoading),
+                (): void => {
+                    smartContentStorePool.updateExcludedIds();
```
nnatter: also this line will affect the `itemsLoading` property of the `previousSmartContentStores`, right?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/SmartContent.js
danrot: Yeah, it is necessary, because otherwise there will be too many requests. We have to wait until the previous smart content is completely loaded before loading the next one excluding duplicates, because otherwise the request will be done first without the excluded ids, and as soon as the previous stores are loaded it will be done again with the excluded ids. The first request would be useless.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/SmartContent.js
danrot: This only makes the previous smart content stores load something if one of the previous stores changed its content as well, which is not possible, unless the content manager changed something in the mean time on their own.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/SmartContent.js
nnatter: this is really hard to understand. i guess its even harder if you dont know about the intention.. can we at least add some comment(s) why this is necessary?

nnatter:
functionality looks good 👍

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/SmartContent.js
danrot: I'll add a comment

PR #4829, opened 2019-10-17
Resolve persistence-bundle entities in references option of doctrine schemas

nnatter:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This PR enhances the `ResolveTargetEntitiesPass` to pass a target-entity mapping to the `ReferencesOption` service.

#### Why?

To allow for using interfaces in a `references` option inside of a doctrine schema.

#### Example Usage

~~~xml
        <field name="userId" column="userId" type="integer" unique="true" nullable="false">
            <options>
                <option name="references">
                    <option name="entity">Sulu\Component\Security\Authentication\UserInterface</option>
                    <option name="field">id</option>
                    <option name="onDelete">CASCADE</option>
                    <option name="onUpdate">CASCADE</option>
                </option>
            </options>
        </field>
~~~

### DIFF
--- a/src/Sulu/Bundle/CoreBundle/Resources/config/doctrine.xml
+++ b/src/Sulu/Bundle/CoreBundle/Resources/config/doctrine.xml
@@ -5,6 +5,8 @@
     <services>
         <service id="sulu_core.doctrine.references" class="Sulu\Component\Doctrine\ReferencesOption">
             <argument type="service" id="doctrine"/>
+            <!-- the $targetEntityMapping argument is set by the ResolveTargetEntitiesPass -->
+            <argument key="$targetEntityMapping" type="collection" />
 
             <tag name="doctrine.event_subscriber"/>
         </service>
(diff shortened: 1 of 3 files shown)


src/Sulu/Bundle/PersistenceBundle/DependencyInjection/Compiler/ResolveTargetEntitiesPass.php
```diff
@@ -48,15 +48,20 @@ private function resolve(ContainerBuilder $container)
...
-                        $this->getClass($container, $model),
-                        [],
-                    ]
-                );
+                ->addMethodCall('addResolveTargetEntity', [$interface, $interfaceImplementation, []]);
         }
+
+        // update $targetEntityMapping argument of ReferencesOption service
+        // this is needed to allow for using interfaces when using a "references" option in a doctrine schema
+        $doctrineReference = $container->findDefinition('sulu_core.doctrine.references');
```
nnatter: i guess we should check if `sulu_core.doctrine.references` exists before we access it? 🤔

src/Sulu/Bundle/CoreBundle/Resources/config/doctrine.xml
```diff
@@ -5,6 +5,8 @@
     <services>
         <service id="sulu_core.doctrine.references" class="Sulu\Component\Doctrine\ReferencesOption">
             <argument type="service" id="doctrine"/>
+            <!-- the $targetEntityMapping argument is set by the ResolveTargetEntitiesPass -->
+            <argument key="$targetEntityMapping" type="collection" />
```
nnatter: i did not find any documentation about the `key` argument - not sure if it works with older symfony versions

PR #4831, opened 2019-10-17
Show proper error message when not available default type is set on a block

danrot:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR throws a proper exception when a block is defined in a XML template without an existing `default-type`.

#### Why?

Because when you are configuring this in a wrong way at the moment, then the XML template is interepreted without any error, but when a new block is added in the UI, it will show a strange error like this:

```
FieldBlocks.js:76 Uncaught TypeError: Cannot read property 'form' of undefined
```

This is very confusing, and does not really help.

#### Example Usage

Try to use the following template and add a new block in the UI.

~~~xml
<?xml version="1.0" ?>

    <key>template_with_invalid_block_default_type</key>

    <view>page.html.twig</view>
    <controller>SuluPageBundle:Default:index</controller>
    <cacheLifetime>2400</cacheLifetime>

    <properties>
        <property name="title" type="text_line" mandatory="true" colspan="6"/>

        <property name="url" type="text_line">
            <tag name="sulu.rlp" />
        </property>

        <block name="blocks" default-type="test" minOccurs="0">
            <meta>
            </meta>
            <types>
                <type name="editor">
                    <properties>
                        <property name="article" type="text_editor"/>
                    </properties>
                </type>
                <type name="images">
                    <properties>
                        <property name="images" type="media_selection" colspan="3"/>
                    </properties>
                </type>
            </types>
        </block>
    </properties>
</template>
~~~

### DIFF
--- a/src/Sulu/Bundle/PageBundle/Tests/Application/config/templates/pages/complex.xml
+++ b/src/Sulu/Bundle/PageBundle/Tests/Application/config/templates/pages/complex.xml
@@ -35,8 +35,8 @@
             <tag name="sulu.rlp"/>
         </property>
 
-        <block name="block1" 
-               default-type="editor"
+        <block name="block1"
+               default-type="default"
                minOccurs="2"
                maxOccurs="10"
                mandatory="true">
--- a/src/Sulu/Component/Content/Exception/InvalidBlockDefaultTypeException.php
+++ b/src/Sulu/Component/Content/Exception/InvalidBlockDefaultTypeException.php
@@ -1,0 +1,69 @@
+<?php
+
+/*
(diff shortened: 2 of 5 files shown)


danrot:
@alexander-schranz Thought about it as well... But that would be a rather big change :confused: But it does show the error on admin at the very first time. The problem is that a part of the cache is already written by then, and on the next request it just assumes that it was already generated...Don't see a quick way to fix this for now.

PR #4834, opened 2019-10-21
Add branch aliases for development

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes -
| Related issues/PRs | -
| License | MIT

#### What's in this PR?

Add branch aliases for development

#### Why?

Allow to use @dev for specific version.

#### Example Usage

~~~json
"sulu/sulu": "^2.0@dev"
~~~

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -103,5 +103,12 @@
         "psr-0": {
             "Sulu\\": "tests/"
         }
+    },
+    "extra": {
+        "branch-alias": {
+            "dev-release/1.6": "1.6.x-dev",
+            "dev-release/2.0": "2.0.x-dev",
+            "dev-master": "2.1.x-dev"
+        }
     }
 }


composer.json
```diff
@@ -103,5 +103,12 @@
         "psr-0": {
             "Sulu\\": "tests/"
         }
+    },
+    "extra": {
+        "branch-alias": {
+            "dev-release/1.6": "1.6.x-dev",
+            "dev-release/2.0": "2.0.x-dev",
```
alexander-schranz: We could skip 1.6 and 2.0 branch aliases if we remove the `release/` prefix from the branches see: https://getcomposer.org/doc/articles/aliases.md#branch-alias

nnatter: would prefer to change our branch names if that is what composer expects per default

danrot: In that case I would also prefer to change the branch names, but that operation is always a bit risky, because PRs might get closed, URLs might break (or does github create correct redirects for them?), ...

Apart from that we have to adjust these lines always when a new version is released, because at least the `master` branch has to be kept up do date... So I think I would currently not take the risk.

PR #4844, opened 2019-10-22
Prevent editor toolbar to overlap sulu header

adrianmarte:
| Q | A
| --- | ---
| Bug fix? | Yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

PR generally reduces the size of the editor so that the problem of overlapping (see "Why?" below) will only occur in rare cases. If it still happens, then the sticky-toolbar of ck editor will be positioned below the header inside the visible area of the editor.

#### Why?

CK Editors toolbar overlaps the SULU header when top of editor is below header after scrolling. This happens when a lot of lines are in content area.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/CKEditor5/ckeditor5.scss
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/CKEditor5/ckeditor5.scss
@@ -1,5 +1,8 @@
 @import '../../containers/Application/colors.scss';
+@import '../../containers/Application/variables.scss';
 @import '../../components/Toolbar/variables.scss';
+@import '../../components/Snackbar/snackbar.scss';
+@import '../../components/Tabs/variables.scss';
 
 $textEditorFontSize: 12px;
 $textEditorButtonMinSize: 20px;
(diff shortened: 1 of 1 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/containers/CKEditor5/ckeditor5.scss
```diff
@@ -79,6 +80,17 @@ $textEditorUnpublishedLinkColor: $gold;
...
+        /**
+         * Move the sticky-toolbar below the SULU header.
+         * Still causes issues when snackbar is visible or when users scrolls whole page to top while editor with
+         * sticky-toolbar is active/visible.
+         */
+        .ck.ck-sticky-panel {
+            .ck.ck-sticky-panel__content_sticky {
+                top: calc($toolbarHeight);
+            }
+        }
```
adrianmarte: @danrot This is perhaps not necessary anymore, as I reduced height of editor. If we should use this feature, we should take the snackbar into account (only if it's visible) and also scrolling. But it would need JS to work as far as I could see.
Should I invest further here?

danrot: That's not the only issue here, if the sticky header appears, and you scroll up again, the position does not change correctly anymore (the toolbar will stay right below the toolbar and have a huge gap to the text editor).

I think we could argue that if the height never exceeds the height of the screen without the toolbar anyway, we don't need this feature anyway... Could it be deactivated in some way? Maybe even just with a few lines of CSS? Would be my preferred solution, because it will result in less trouble.

@chirimoya WDYT?

danrot:
@adrianmarte Just found another small error :see_no_evil: Steps to reproduce:

1. Move down until the ckeditor toolbar is not visible anymore
2. Click into the ckeditor
3. Scroll back up to see toolbar

In that case there is quite a bit gap between the label and the toolbar, which should not be there.

PR #4847, opened 2019-10-23
Implement a command that downloads the current JS build

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | TODO

#### What's in this PR?

This PR adds a command that downloads the JS build in the correct version for the developer.

#### Why?

Because that should make the update easier.

#### Example Usage

~~~bash
bin/console sulu:admin:download-build
~~~

#### To Do

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Command/DownloadBuildCommand.php
+++ b/src/Sulu/Bundle/AdminBundle/Command/DownloadBuildCommand.php
@@ -1,0 +1,160 @@
+<?php
+
+/*
+ * This file is part of Sulu.
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\AdminBundle\Command;
+
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Filesystem\Filesystem;
+use Symfony\Component\Templating\Tests\Storage\FileStorageTest;
+use Symfony\Contracts\HttpClient\HttpClientInterface;
+
+class DownloadBuildCommand extends Command
+{
+    protected static $defaultName = 'sulu:admin:download-build';
+
+    /**
+     * @var HttpClientInterface
+     */
+    private $httpClient;
+
+    /**
+     * @var string
+     */
+    private $projectDir;
+
+    /**
+     * @var string
+     */
+    private $suluVersion;
+
+    /**
+     * @var string
+     */
+    private $remoteRepository;
+
+    /**
+     * @var string
+     */
+    private $remoteArchive;
+
+    const ASSETS_DIR = DIRECTORY_SEPARATOR . 'assets' . DIRECTORY_SEPARATOR . 'admin' . DIRECTORY_SEPARATOR;
+
+    const BUILD_DIR = DIRECTORY_SEPARATOR . 'public' . DIRECTORY_SEPARATOR . 'build' . DIRECTORY_SEPARATOR . 'admin';
+
+    const REPOSITORY_NAME = 'skeleton';
+
+    public function __construct(HttpClientInterface $httpClient, string $projectDir, string $suluVersion)
+    {
+        parent::__construct();
+
+        $this->httpClient = $httpClient;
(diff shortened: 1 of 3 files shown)


src/Sulu/Bundle/AdminBundle/Command/DownloadBuildCommand.php
```diff
@@ -0,0 +1,160 @@
...
+
+    const REPOSITORY_NAME = 'skeleton';
+
+    public function __construct(HttpClientInterface $httpClient, string $projectDir, string $suluVersion)
+    {
+        parent::__construct();
+
+        $this->httpClient = $httpClient;
+        $this->projectDir = $projectDir;
+        $this->suluVersion = $suluVersion;
```
alexander-schranz: I would check if suluVersion is a valid sulu version and else ask the developer which version he want to download.

src/Sulu/Bundle/AdminBundle/Command/DownloadBuildCommand.php
```diff
@@ -0,0 +1,160 @@
...
+            );
+        }
+
+        $tempDirectory = sys_get_temp_dir() . DIRECTORY_SEPARATOR . static::REPOSITORY_NAME . uniqid(rand(), true);
+        $tempFileZip = $tempDirectory . '.zip';
+
+        $output->writeln('<info>Download remote repository...</info>');
+        $response = $this->httpClient->request('GET', $this->remoteArchive, [
+            'on_progress' => function(int $downloaded, int $size) {
+                /* echo $downloaded . '/' . $size . PHP_EOL; */
```
alexander-schranz: ?

src/Sulu/Bundle/AdminBundle/Command/DownloadBuildCommand.php
danrot: I thought about just throwing an exception when it is not a numeric version, because if you are on a develop branch you have to build it on your own anyway (it's not guaranteed for dev branches that the build is up to date).

src/Sulu/Bundle/AdminBundle/Command/DownloadBuildCommand.php
danrot: Actually I wanted to show a progress bar, but that does not seem to work correctly :confused: Both values only show 0, which is not correct for sure...

src/Sulu/Bundle/AdminBundle/Command/DownloadBuildCommand.php
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    protected function configure()
+    {
+        $this->setDescription('Downloads the current admin application build from the sulu/skeleton repository.');
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $output->writeln('<info>Checking for changed files...</info>');
```
alexander-schranz: You can use SymfonyStyle class to simplify outputs: https://symfony.com/doc/current/console/style.html#basic-usage.

src/Sulu/Bundle/AdminBundle/Command/DownloadBuildCommand.php
```diff
@@ -0,0 +1,160 @@
...
+
+        if ($localIndexJsHash !== $remoteIndexJsHash
+            || $localPackageJsonHash !== $remotePackageJsonHash
+            || $localWebpackConfigJsHash !== $remoteWebpackConfigJsHash
+        ) {
+            throw new \Exception(
+                sprintf(
+                    'The files in the local "%s" folder do not match the ones in the remote repository "%s".' . PHP_EOL
+                    . 'Either the build has been modified, which means it has to be done manually with NPM or the files'
+                    . ' in your repository are outdated and have to be copied from the remote repository.',
```
alexander-schranz: Maybe add also something like "If you have additional bundles or custom admin code you need to build the admin.js yourself"

src/Sulu/Bundle/AdminBundle/Command/DownloadBuildCommand.php
alexander-schranz: @danrot throwing an exception is ok for me, should just contain a text like: "Looks like you are not a tagged version of sulu, you need to build the admin js yourself"

PR #4850, opened 2019-10-23
Fix page rule to allow history and redirect pages

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | none
| Related issues/PRs | none
| License | MIT
| Documentation PR | none

#### What's in this PR?

This PR contains a fix to ignore redirects for history and trailing slashes.

#### Why?

Histories and trailing slashes throw an exception.


danrot:
This only happens if there a page which leads to a redirect is loaded AND one of the configured target groups is using the `PageRule`.

PR #4862, opened 2019-10-31
Add media-adapter for vips

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | none
| Related issues/PRs | none
| License | MIT
| Documentation PR | https://github.com/sulu/sulu-docs/pull/463

#### What's in this PR?

This PR add adapter for [rokka-io/imagine-vips](https://github.com/rokka-io/imagine-vips/issues).

See Blog post about it at the [liip blog ](https://www.liip.ch/de/blog/libvips-adapter-for-php-imagine)

#### Why?

It is an alternative for imagick and gd.

#### Example Installation on mac

~~~bash
brew install pkg-config
brew install vips
pecl install vips
composer require rokka/imagine-vips
~~~

#### To Do


src/Sulu/Bundle/MediaBundle/Resources/config/services.xml
```diff
@@ -158,6 +158,7 @@
         <service id="sulu_media.adapter.gd" class="Imagine\Gd\Imagine" public="false"/>
         <service id="sulu_media.adapter.imagick" class="Imagine\Imagick\Imagine" public="false"/>
         <service id="sulu_media.adapter.gmagick" class="Imagine\Gmagick\Imagine" public="false"/>
+        <service id="sulu_media.adapter.vips" class="Imagine\Vips\Imagine" public="false"/>
```
alexander-schranz: service should only be registered when class is available

src/Sulu/Bundle/MediaBundle/Media/ImageConverter/ImagineImageConverter.php
```diff
@@ -191,7 +191,11 @@ public function convert(FileVersion $fileVersion, $formatKey, $imageFormat)
 
         // Set Interlacing to plane for smaller image size.
         if (1 == count($image->layers())) {
-            $image->interlace(ImageInterface::INTERLACE_PLANE);
+            try {
+                $image->interlace(ImageInterface::INTERLACE_PLANE);
+            } catch (\Exception $exception) {
```
danrot: Isn't there a better exception we can catch?

wachterjohannes:
@danrot @alexander-schranz fixed and ready to review again

src/Sulu/Bundle/MediaBundle/Resources/config/services.xml
alexander-schranz: @wachterjohannes this should be removed from this file

wachterjohannes: done

danrot:
I would say the `rokka/imagine-vips` package should at least be added as suggestion, shouldn't it?

src/Sulu/Bundle/MediaBundle/DependencyInjection/SuluMediaExtension.php
```diff
@@ -270,10 +273,15 @@ public function load(array $configs, ContainerBuilder $container)
         $loader->load('command.xml');
 
         if ('auto' === $config['adapter']) {
-            $container->setAlias(
-                'sulu_media.adapter',
-                'sulu_media.adapter.' . (class_exists('Imagick') ? 'imagick' : 'gd')
-            );
+            $adapter = 'gd';
+            if (class_exists('Imagine\Vips\Imagine')) {
```
danrot: Any reason not to use the `::class` constant`

src/Sulu/Bundle/MediaBundle/DependencyInjection/SuluMediaExtension.php
```diff
@@ -270,10 +273,15 @@ public function load(array $configs, ContainerBuilder $container)
...
+                $loader->load('services_imagine_vips.xml');
+                $adapter = 'vips';
+            } elseif (class_exists('Imagick')) {
```
danrot: Same as above

src/Sulu/Bundle/MediaBundle/Media/ImageConverter/ImagineImageConverter.php
```diff
@@ -191,7 +191,13 @@ public function convert(FileVersion $fileVersion, $formatKey, $imageFormat)
 
         // Set Interlacing to plane for smaller image size.
         if (1 == count($image->layers())) {
-            $image->interlace(ImageInterface::INTERLACE_PLANE);
+            try {
+                $image->interlace(ImageInterface::INTERLACE_PLANE);
+            } catch (\InvalidArgumentException $exception) {
+                // ignore unsupported Interface type
+            } catch (\RuntimeException $exception) {
```
danrot: Is there no better exception? Same question for the above `catch`

alexander-schranz: @wachterjohannes It should be a `Imagine\Exception\RuntimeException`

src/Sulu/Bundle/MediaBundle/Media/ImageConverter/ImagineImageConverter.php
```diff
@@ -191,7 +191,13 @@ public function convert(FileVersion $fileVersion, $formatKey, $imageFormat)
 
         // Set Interlacing to plane for smaller image size.
         if (1 == count($image->layers())) {
-            $image->interlace(ImageInterface::INTERLACE_PLANE);
+            try {
+                $image->interlace(ImageInterface::INTERLACE_PLANE);
+            } catch (\InvalidArgumentException $exception) {
+                // ignore unsupported Interface type
```
alexander-schranz: this seems have nothing todo with the [vips](https://github.com/rokka-io/imagine-vips/blob/3039f6a27d481c4a61363b68a910a8110973bcd1/lib/Imagine/Vips/Image.php#L505) adapter what case does this then happen should maybe be not part of this PR?

src/Sulu/Bundle/MediaBundle/Media/ImageConverter/ImagineImageConverter.php
wachterjohannes: @alexander-schranz thats not correct - the Vips Image throws a `\RuntimeException` because the function it is not implemented there.

https://github.com/rokka-io/imagine-vips/blob/0.9.0/lib/Imagine/Vips/Image.php#L514

what you have changed yesterday there 🙈 i will also catch the new exception here.

src/Sulu/Bundle/MediaBundle/Media/ImageConverter/ImagineImageConverter.php
wachterjohannes: not directly but thats a possible exception which is thrown in other adapters of imagine

danrot: Which does not automatically mean that we should simply catch and ignore. This way we won't even realize that an exception happened. Or is that in any way required in order to work?

wachterjohannes: yes - if you have installed the vips adapter (or any other adapter) which does not implement interlace sulu will crash :) we have to ignore it - as we also do it on other places in this class

danrot: But as @alexander-schranz already described `interlace` in that case throws a `RuntimeException` (which I also don't like, but that seems to be the way it is...) not an `InvalidArgumentException`, which could also have loads of other causes.

src/Sulu/Bundle/MediaBundle/Media/ImageConverter/ImagineImageConverter.php
danrot: Btw, I am even not sure if they did that on purpose, because sometime they throw the `Imagine\Exception\RuntimeException` as @alexander-schranz mentioned... I think that might be even a bug in their adapter.

src/Sulu/Bundle/MediaBundle/DependencyInjection/SuluMediaExtension.php
wachterjohannes: yes - phpstan will throw an error  in the CI/CD (https://travis-ci.org/sulu/sulu/jobs/634123652#L1272) and we cannot install the lib - because then we would have to install the extension `vips`

src/Sulu/Bundle/MediaBundle/Media/ImageConverter/ImagineImageConverter.php
alexander-schranz: For clearification in version < 1.0 imagine. Not Implemented or supported function did throw a `Imagine\Exception\RuntimeException`. In >=1.0 they did add a new exception the `Imagine\Exception\NotSupportedException` which extends from `Imagine\Exception\RuntimeException`. Sadly the vips adapter didn't implement that correctly so I created a PR not sure if they gonna release it soon.

src/Sulu/Bundle/MediaBundle/Media/ImageConverter/ImagineImageConverter.php
wachterjohannes: OK will remove the catch

src/Sulu/Bundle/MediaBundle/DependencyInjection/SuluMediaExtension.php
alexander-schranz: If its just phpstan I would add it to ignore errors.

src/Sulu/Bundle/MediaBundle/Media/ImageConverter/ImagineImageConverter.php
```diff
@@ -11,6 +11,7 @@
 
 namespace Sulu\Bundle\MediaBundle\Media\ImageConverter;
 
+use Imagine\Exception\NotSupportedException;
```
danrot: Is it safe to use this also with versions of the imagine library not having this exception? If not we would have to rise the minimum version of it in `composer.json`. On the other hand I think it is not necessary to catch this exception separately, if we already catch the `RuntimeException` (hopefully soon the one from Imagine if the imagine-vips adapter releases a new version), which the `NotSupportedException` is inheriting from.

src/Sulu/Bundle/MediaBundle/Media/ImageConverter/ImagineImageConverter.php
danrot: Let's wait for their answer in rokka-io/imagine-vips#7

src/Sulu/Bundle/MediaBundle/Media/ImageConverter/ImagineImageConverter.php
wachterjohannes: have removed the `NotSupportedException ` from the catch clause

src/Sulu/Bundle/MediaBundle/Media/ImageConverter/ImagineImageConverter.php
```diff
@@ -203,6 +203,8 @@ public function convert(FileVersion $fileVersion, $formatKey, $imageFormat)
             }
         } catch (RuntimeException $exception) {
             // ignore exceptions here (some imagine adapter does not implement this)
+        } catch (\RuntimeException $exception) {
```
danrot: Wasn't the reason we were waiting for the other merge, that we don't need that `catch` anymore?

danrot:
I can confirm that this does not break using the old imagine adapters. However, I have a hard time finding out how to install `vips` including the php extension on my system :see_no_evil: 

@alexander-schranz Would be great if you could test that, and merge if the tests succeed.

alexander-schranz:
@wachterjohannes at current state to support animated gifs we need change the following in the `modifyAllLayers` function:

```diff
diff --git a/src/Sulu/Bundle/MediaBundle/Media/ImageConverter/ImagineImageConverter.php b/src/Sulu/Bundle/MediaBundle/Media/ImageConverter/ImagineImageConverter.php
index ba29d16187..6cd2557cf4 100644
--- a/src/Sulu/Bundle/MediaBundle/Media/ImageConverter/ImagineImageConverter.php
+++ b/src/Sulu/Bundle/MediaBundle/Media/ImageConverter/ImagineImageConverter.php
@@ -16,6 +16,7 @@ use Imagine\Filter\Basic\Autorotate;
 use Imagine\Image\ImageInterface;
 use Imagine\Image\ImagineInterface;
 use Imagine\Image\Palette\RGB;
+use Imagine\Vips\Image as VipsImage;
 use Sulu\Bundle\MediaBundle\Entity\FileVersion;
 use Sulu\Bundle\MediaBundle\Entity\FormatOptions;
 use Sulu\Bundle\MediaBundle\Media\Exception\ImageProxyInvalidFormatOptionsException;
@@ -403,7 +404,10 @@ class ImagineImageConverter implements ImageConverterInterface

         if (count($layers)) {
             $countLayer = 0;
-            $image->layers()->coalesce();
+
+            if (!$image instanceof VipsImage) {
+                $image->layers()->coalesce();
+            }

             /** @var ImageInterface $temporaryImage */
             $temporaryImage = null;
@@ -411,7 +415,7 @@ class ImagineImageConverter implements ImageConverterInterface
                 ++$countLayer;
                 $layer = call_user_func($modifier, $layer);
                 if (1 === $countLayer) {
-                    $temporaryImage = $layer; // use first layer as main image
+                    $temporaryImage = clone $layer; // use first layer as main image
                 } else {
                     $temporaryImage->layers()->add($layer);
                 }

```

src/Sulu/Bundle/MediaBundle/Media/ImageConverter/ImagineImageConverter.php
```diff
@@ -401,15 +402,20 @@ private function modifyAllLayers(ImageInterface $image, callable $modifier)
 
         if (count($layers)) {
             $countLayer = 0;
-            $image->layers()->coalesce();
+
+            // coalesce currently not supported by the VipsAdapter:
+            // see: https://github.com/rokka-io/imagine-vips/issues/
```
danrot: Issue number would also be nice :sweat_smile:

src/Sulu/Bundle/MediaBundle/Media/ImageConverter/ImagineImageConverter.php
```diff
@@ -401,15 +402,20 @@ private function modifyAllLayers(ImageInterface $image, callable $modifier)
...
+            if (!$image instanceof VipsImage) {
```
danrot: @alexander-schranz Didn't we say that the `coalesce` function should be a noop in the `Vips` adapter, because it seems it is not necessary? If the maintainers agree and would make a release soonish, I would be more comfortable with waiting until this is fixed before merging this, instead of having this `if` with a reference to a concrete adapter...

wachterjohannes:
@alexander-schranz your fix breaks the behavior for GD see travis error https://travis-ci.org/wachterjohannes/sulu/jobs/634609962#L1005

alexander-schranz:
@wachterjohannes can nothing todo with that changes for the gif as that only applies to multi layer images and the test which is failing is a jpeg and that part is never reached by a jpg image.

src/Sulu/Bundle/MediaBundle/Media/ImageConverter/ImagineImageConverter.php
alexander-schranz: @wachterjohannes https://github.com/rokka-io/imagine-vips/issues/8

src/Sulu/Bundle/MediaBundle/Media/ImageConverter/ImagineImageConverter.php
```diff
@@ -399,17 +400,22 @@ private function modifyAllLayers(ImageInterface $image, callable $modifier)
             $layers = [];
         }
 
-        if (count($layers)) {
+        if (count($layers) > 1) {
```
alexander-schranz: Did debug and this will save us a lot of memory 40mb -> 8mb tested on the imagick adapter.

alexander-schranz: @danrot backported this to 1.6: https://github.com/sulu/sulu/pull/4968

composer.json
```diff
@@ -151,6 +152,7 @@
     },
     "suggest": {
         "contao/imagine-svg": "^0.2.2 - Rendering raw svg formats for svg uploads",
+        "rokka/imagine-vips": "^0.9.1 - Rendering formats by using the libvips",
```
danrot: Replace that with `^0.9.2`

src/Sulu/Bundle/MediaBundle/Media/ImageConverter/ImagineImageConverter.php
danrot: I would prefer to not have this change in here - then it is less likely that there'll be a conflict when I merge back, and I have less to test in this PR :-)

alexander-schranz: @danrot tests will fail then as the clone doesn't work on `GD` driver.

danrot: But doesn't that mean that it will fail if you upload an animated GIF if using GD in general? :thinking:

alexander-schranz: GD doesn't support animated gifs or other multi layer image formats [see](https://imagine.readthedocs.io/en/stable/usage/layers.html#disclaimer).

PR #4863, opened 2019-10-31
Add optional svg support

### LINKED ISSUE #2246: No preview images for SVG and EPS files
| Q | A |
| --- | --- |
| Bug? | no |
| New Feature? | yes |
| Sulu Version | 1.2.0-RC4 |
#### Actual Behavior

If a EPS or SVG file is uploaded no preview image are shown: 

image
#### Expected Behavior

A preview image should be shown. In addition vector graphics should probably have its own media type, since it doesn't make too much sense to calculate image formats (would somehow contradict the use case of vectory graphics), and therefore wouldn't be used with formats in twig. In that case it should not be possible to upload a new version with a bitmap graphic.
#### Steps to Reproduce
1. Navigation to media
2. Select an arbitrary collection
3. Upload a SVG or EPS image
#### Possible Solutions

Generating preview images probably doesn't make any sense in that case, since these images are vectorgraphics anyway. We should probably simply show them without any image processing.

However, it might also make sense to rasterize vector graphics, although it might be heavier to load then, depending on the image.

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #2246
 Related issues/PRs | none
| License | MIT
| Documentation PR | https://github.com/sulu/sulu-docs/pull/463

#### What's in this PR?

This PR add the optional dependency `contao/imagine-svg` to support svg.

#### Example Usage

~~~bash
composer require contao/imagine-svg
~~~

#### To Do


danrot:
Don't forget to close #2246 when this gets merged

danrot:
And I would argue to create a test for that, because nobody of us will realize when this stops working.

wachterjohannes:
@danrot @alexander-schranz fixed and ready to review again

src/Sulu/Bundle/MediaBundle/Tests/Functional/Controller/MediaStreamControllerTest.php
```diff
@@ -121,6 +121,31 @@ public function testGetImageAction()
         $this->assertSame('image/gif', $client->getResponse()->headers->get('Content-Type'));
     }
 
+    public function testGetImageActionSvg()
+    {
+        if (!class_exists(\Imagick::class)) {
```
alexander-schranz: Don't understand this skip. If imagick is not available it should use gd and in this case it should not matter as it should use the svg adapter.

src/Sulu/Bundle/MediaBundle/Media/ImageConverter/ImagineImageConverter.php
```diff
@@ -124,8 +124,14 @@ public function getSupportedOutputImageFormats(?string $mimeType): array
 
                 switch ($mimeType) {
                     case 'image/png':
+                        $preferredExtension = 'png';
+                        break;
                     case 'image/svg+xml':
+                    case 'image/svg':
                         $preferredExtension = 'png';
+                        if (class_exists('Contao\ImagineSvg\Imagine')) {
```
alexander-schranz: Would import it for better IDE refractoring support:

`use Contao\ImagineSvg\Imagine as SvgImagine;`

src/Sulu/Bundle/MediaBundle/Tests/Functional/Controller/MediaStreamControllerTest.php
wachterjohannes: had following error: https://travis-ci.org/sulu/sulu/jobs/633654526#L1012

i think gd is not able to to render svg as png (last check in this testcase).

will change that to only check the conversion when imagick is available

wachterjohannes:
@alexander-schranz fixed

src/Sulu/Bundle/MediaBundle/Media/ImageConverter/ImagineImageConverter.php
```diff
@@ -124,8 +125,14 @@ public function getSupportedOutputImageFormats(?string $mimeType): array
...
```
alexander-schranz: use `SvgImagine::class` or any problems with that?

wachterjohannes:
@alexander-schranz done

composer.json
```diff
@@ -98,6 +98,7 @@
         "twig/twig": "^1.41 || ^2.0"
     },
     "require-dev": {
+        "contao/imagine-svg": "^0.2.2",
```
danrot: I guess that's not just a dev requirement, or is it?

danrot: Resp. at least it should also be listed as suggestion, if we don't want to install it by default. But is there any reason you would not want to install it? Is that somehow super heavy to initialize or has another negative impact on performance?

src/Sulu/Bundle/MediaBundle/Media/ImageConverter/ImagineImageConverter.php
```diff
@@ -155,8 +162,13 @@ public function convert(FileVersion $fileVersion, $formatKey, $imageFormat)
     {
         $imageResource = $this->mediaImageExtractor->extract($this->storage->load($fileVersion->getStorageOptions()));
 
+        $imagine = $this->imagine;
+        if ('svg' === $imageFormat && class_exists(SvgImagine::class)) {
+            $imagine = new SvgImagine();
```
danrot: Wouldn't it be better for performance and testability if we inject that as a service?

src/Sulu/Bundle/MediaBundle/Media/ImageConverter/ImagineImageConverter.php
```diff
@@ -189,9 +201,13 @@ public function convert(FileVersion $fileVersion, $formatKey, $imageFormat)
...
 
-        // Set Interlacing to plane for smaller image size.
-        if (1 == count($image->layers())) {
-            $image->interlace(ImageInterface::INTERLACE_PLANE);
+        try {
+            // Set Interlacing to plane for smaller image size.
+            if (1 == count($image->layers())) {
+                $image->interlace(ImageInterface::INTERLACE_PLANE);
+            }
+        } catch (\Exception $exception) {
```
danrot: Isn't there a better exception to catch?

src/Sulu/Bundle/MediaBundle/Media/ImageConverter/ImagineImageConverter.php
```diff
@@ -382,7 +398,13 @@ private function getCropParameters(ImageInterface $image, $formatOptions, array
      */
     private function modifyAllLayers(ImageInterface $image, callable $modifier)
     {
-        if (count($image->layers())) {
+        try {
+            $layers = $image->layers();
+        } catch (\Exception $exception) {
```
danrot: Is there no better exception to catch?

src/Sulu/Bundle/MediaBundle/Tests/Functional/Controller/MediaStreamControllerTest.php
```diff
@@ -121,6 +121,27 @@ public function testGetImageAction()
...
+
+        $client->request('GET', $media->getFormats()['small-inset']);
+        $this->assertHttpStatusCode(200, $client->getResponse());
+        $this->assertSame('image/svg+xml', $client->getResponse()->headers->get('Content-Type'));
+
+        $client->request('GET', $media->getFormats()['small-inset.svg']);
+        $this->assertHttpStatusCode(200, $client->getResponse());
+        $this->assertSame('image/svg+xml', $client->getResponse()->headers->get('Content-Type'));
+
+        if (class_exists(\Imagick::class)) {
```
danrot: Wouldn't it make sense to move that into a separate test method, and skip it if this is not available?

src/Sulu/Bundle/MediaBundle/Media/ImageConverter/ImagineImageConverter.php
alexander-schranz: @wachterjohannes change to atleast [Imagine\Exception\RuntimeException](https://github.com/contao/imagine-svg/blob/b62cae7d217c7d11efd88b5c2f373f87814767f8/src/Image.php#L15)

src/Sulu/Bundle/MediaBundle/Tests/Functional/Controller/MediaStreamControllerTest.php
alexander-schranz: Also I'm not sure why to skip this as png should also work on gd but maye travis doesn't support png so maybe you should request `.jpg` instead of `.png` here as jpg seems to work above and as @danrot wrote I would create an own testcase for SvgAsJpg case.

composer.json
wachterjohannes: will add it to suggestions

src/Sulu/Bundle/MediaBundle/Media/ImageConverter/ImagineImageConverter.php
wachterjohannes: have changed it to `Imagine\Exception\RuntimeException `

src/Sulu/Bundle/MediaBundle/DependencyInjection/SuluMediaExtension.php
```diff
@@ -269,6 +270,10 @@ public function load(array $configs, ContainerBuilder $container)
         $loader->load('services.xml');
         $loader->load('command.xml');
 
+        if (SvgImagine::class) {
```
danrot: Are you missing a `class_exists` call or something like that around this? I get an error currently, saying that the `Contao\ImagineSvg\Imagine` class cannot be loaded, if the package is not installed...

mareksuscak:
Has this been released yet? It doesn't seem to be working in Sulu 2.0.6 after installing `contao/imagine-svg` package.

image

alexander-schranz:
@mareksuscak you need to install 2.1.0-RC1 if you want to use this feature.

PR #4878, opened 2019-11-08
Show better preview content block exception

### LINKED ISSUE #4877: Notice: Undefined offset: 2
| Q | A
| --- | ---
| Bug? | yes
| New Feature? | no
| Sulu Version | 2.0.1

#### Actual Behavior

How does Sulu behave at the moment? 
obraz

#### Expected Behavior

correct preview

#### Steps to Reproduce

Probably in data structure is something wrong, but I don't know how check it.

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #4877
| Related issues/PRs | -
| License | MIT

#### What's in this PR?

Show better preview content block exception

#### Why?

The current code end up in an error without throwing a exception why it did happened.

#### Usage?

Remove the `block content` the following exception should then be shown.

### DIFF
--- a/src/Sulu/Bundle/PreviewBundle/Preview/Renderer/PreviewKernel.php
+++ b/src/Sulu/Bundle/PreviewBundle/Preview/Renderer/PreviewKernel.php
@@ -13,6 +13,7 @@ namespace Sulu\Bundle\PreviewBundle\Preview\Renderer;
 
 use App\Kernel;
 use Symfony\Component\Config\Loader\LoaderInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
 
 /**
  * Extends website-kernel from sulu-installation and override configuration.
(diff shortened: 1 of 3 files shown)


danrot:
I am missing a test here :slightly_smiling_face:

nnatter:
@alexander-schranz ping 😇

src/Sulu/Bundle/PreviewBundle/Preview/Renderer/PreviewKernel.php
```diff
@@ -34,9 +35,13 @@ public function registerContainerConfiguration(LoaderInterface $loader)
         parent::registerContainerConfiguration($loader);
 
         if (in_array($this->getEnvironment(), ['dev', 'test'])) {
-            $loader->load(
-                implode(DIRECTORY_SEPARATOR, [__DIR__, '..', '..', 'Resources', 'config', 'config_preview_dev.yml'])
-            );
+            $loader->load(function(ContainerBuilder $container) use ($loader) {
+                if ($container->hasExtension('web_profiler')) {
```
danrot: I also don't really understand what this is doing... Why are we nesting the `load` calls? :thinking: And if we check for the existence of the `web_profiler`, couldn't we remove the environment check above? Then this would even work if somebody wants to activate the `web_profiler` in production for some reason.

alexander-schranz: the nested load is needed as its the only way to access the `$container` here correctly.
And yes you are right I did remove the environment check.

src/Sulu/Bundle/PreviewBundle/Preview/Renderer/PreviewKernel.php
```diff
@@ -33,11 +34,13 @@ public function registerContainerConfiguration(LoaderInterface $loader)
     {
         parent::registerContainerConfiguration($loader);
 
-        if (in_array($this->getEnvironment(), ['dev', 'test'])) {
-            $loader->load(
-                implode(DIRECTORY_SEPARATOR, [__DIR__, '..', '..', 'Resources', 'config', 'config_preview_dev.yml'])
-            );
-        }
+        $loader->load(function(ContainerBuilder $container) use ($loader) {
```
nnatter: `// disable web_profiler toolbar in preview if the web_profiler is enabled`

alexander-schranz: Comment added

PR #4889, opened 2019-11-13
Validate required fields in form blocks

### LINKED ISSUE #4629: Form mandatory validation should work within block
| Q | A
| --- | ---
| Bug? | yes
| New Feature? | no
| Sulu Version | 2.0
| Browser Version | -

#### Actual Behavior

Form mandatory validation doesn't work within a block element.

#### Expected Behavior

Form validation should also work within a block element. The validation does work if the mandatory property is located in a normal form outside the block.

#### Steps to Reproduce

Add a block which contains at least one mandatory property and save the form afterwards in the browser.

```
<block name="receivers" default-type="to">
    <types>
        <type name="to">
            <properties>
                <property name="email" type="email" colspan="6" mandatory="true"/>     
            </properties>
        </type>
    </types>
</block>
```

danrot:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #4629
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR considers the block content type when generating the JSON schema for a form.

#### Why?

Because this way the validation for required fields will also work in blocks.

#### Example Usage

~~~xml
        <block name="blocks" default-type="text" minOccurs="0">
            <meta>
                <title lang="de">Text</title>
                <title lang="en">Text</title>
            </meta>
            <types>
                <type name="text">
                    <meta>
                        <title lang="de">Text</title>
                        <title lang="en">Text</title>
                    </meta>
                    <properties>
                            <meta>
                                <title lang="de">Title</title>
                            </meta>
                        </property>
                        <property name="editor" type="text_editor" mandatory="true">
                            <meta>
                                <title lang="de">Article</title>
                            </meta>
                        </property>
                    </properties>
                </type>
                <type name="image">
                    <meta>
                        <title lang="de">Text</title>
                        <title lang="en">Text</title>
                    </meta>
                    <properties>
                            <meta>
                                <title lang="de">Title</title>
                            </meta>
                        </property>
                        <property name="image" type="media_selection" mandatory="true">
                            <meta>
                                <title lang="de">Image</title>
                                <title lang="en">Image</title>
                            </meta>
                        </property>
                    </properties>
                </type>
            </types>
        </block>
~~~

#### BC Breaks/Deprecations

To discuss

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/FormMetadata/FormMetadataMapper.php
+++ b/src/Sulu/Bundle/AdminBundle/FormMetadata/FormMetadataMapper.php
@@ -224,10 +224,6 @@ class FormMetadataMapper
                 );
             }
 
-            if (!$itemMetadata->isRequired()) {
-                return;
-            }
-
             return new PropertyMetadata($itemMetadata->getName(), $itemMetadata->isRequired());
         }, $itemsMetadata));
     }
--- a/src/Sulu/Bundle/AdminBundle/FormMetadata/FormMetadataMapper.php
+++ b/src/Sulu/Bundle/AdminBundle/FormMetadata/FormMetadataMapper.php
@@ -17,9 +17,12 @@ use Sulu\Bundle\AdminBundle\Metadata\FormMetadata\ItemMetadata;
 use Sulu\Bundle\AdminBundle\Metadata\FormMetadata\OptionMetadata;
 use Sulu\Bundle\AdminBundle\Metadata\FormMetadata\SectionMetadata;
 use Sulu\Bundle\AdminBundle\Metadata\FormMetadata\TagMetadata;
+use Sulu\Bundle\AdminBundle\Metadata\SchemaMetadata\ArrayMetadata;
+use Sulu\Bundle\AdminBundle\Metadata\SchemaMetadata\ConstMetadata;
 use Sulu\Bundle\AdminBundle\Metadata\SchemaMetadata\PropertyMetadata;
 use Sulu\Bundle\AdminBundle\Metadata\SchemaMetadata\SchemaMetadata;
 use Sulu\Component\Content\Metadata\BlockMetadata;
+use Sulu\Component\Content\Metadata\BlockMetadata as ContentBlockMetadata;
 use Sulu\Component\Content\Metadata\ItemMetadata as ContentItemMetadata;
 use Sulu\Component\Content\Metadata\PropertyMetadata as ContentPropertyMetadata;
 use Sulu\Component\Content\Metadata\SectionMetadata as ContentSectionMetadata;
@@ -201,6 +204,24 @@ class FormMetadataMapper
         return array_filter(array_map(function(ContentItemMetadata $itemMetadata) {
             if ($itemMetadata instanceof ContentSectionMetadata) {
                 return $this->mapSchemaProperties($itemMetadata->getChildren());
+            }
+
+            if ($itemMetadata instanceof ContentBlockMetadata) {
+                $blockTypeSchemas = [];
+                foreach ($itemMetadata->getComponents() as $blockType) {
+                    $blockTypeSchemas[] = new SchemaMetadata(
+                        array_merge(
+                            $this->mapSchemaProperties($blockType->getChildren()),
+                            ['type' => new ConstMetadata('type', true, $blockType->getName())]
+                        )
+                    );
+                }
+
+                return new ArrayMetadata(
+                    $itemMetadata->getName(),
+                    $itemMetadata->isRequired(),
+                    new SchemaMetadata([], $blockTypeSchemas)
+                );
             }
 
             if (!$itemMetadata->isRequired()) {
--- a/src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/ConstMetadata.php
+++ b/src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/ConstMetadata.php
@@ -1,0 +1,41 @@
+<?php
+
+/*
+ * This file is part of Sulu.
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\AdminBundle\Metadata\SchemaMetadata;
+
+class ConstMetadata extends PropertyMetadata
+{
+    /**
+     * @var string|number|null
+     */
+    private $value;
+
+    /**
+     * @param string|number|null $value
+     */
+    public function __construct(string $name, bool $mandatory, $value = null)
+    {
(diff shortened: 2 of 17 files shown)


src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/PropertyMetadata.php
```diff
@@ -11,49 +11,35 @@
 
 namespace Sulu\Bundle\AdminBundle\Metadata\SchemaMetadata;
 
-class PropertyMetadata
+class PropertyMetadata extends AbstractPropertyMetadata
```
danrot: @sulu/core-team If I would do that one again, I would probably name it `ConstMetadata`. Initially I thought having one `PropertyMetadata` will be enough, but then we might end up with a huge constructor when adding more stuff to validate. I've already had the `properties` from the `ArrayMetadata` added here as well, but I decided to split these two.

Question is if we should rename this class (I would suggest so), and if we introduce a BC layer (Creating a new class called `ConstMetadata` extending from this one, and deprected the `PropertyMetadata` class).

I would go for the deprecation, and if we do that, I would first get rid of all deprecations that currently have no alternative (if there is no alternative it should not be a deprecation) and configure our [ŧests to fail if we use some of our own deprecated code](https://symfony.com/doc/current/components/phpunit_bridge.html#internal-deprecations).

src/Sulu/Bundle/AdminBundle/FormMetadata/FormMetadataMapper.php
```diff
@@ -203,6 +206,24 @@ private function mapSchemaProperties(array $itemsMetadata): array
...
+                }
+
+                return new ArrayMetadata(
+                    $itemMetadata->getName(),
+                    $itemMetadata->isRequired(),
+                    new SchemaMetadata([], $blockTypeSchemas)
+                );
+            }
+
             if (!$itemMetadata->isRequired()) {
```
nnatter: does it still make sense to return here? 🤔

src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/ConstMetadata.php
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+namespace Sulu\Bundle\AdminBundle\Metadata\SchemaMetadata;
+
+class ConstMetadata extends PropertyMetadata
+{
+    /**
+     * @var string|number|null
+     */
+    private $value;
```
nnatter: does it make sense that this is nullable?

src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/PropertyMetadata.php
```diff
@@ -47,13 +41,6 @@ public function isMandatory(): bool
...
     {
-        if (null === $this->value) {
-            return null;
-        }
-
-        return [
-            'name' => $this->name,
-            'const' => $this->value,
-        ];
+        return null;
```
nnatter: okay i dont know how this works tbh. but does this still have an effect if we return null here?

src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/ConstMetadata.php
```diff
@@ -0,0 +1,41 @@
...
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\AdminBundle\Metadata\SchemaMetadata;
+
+class ConstMetadata extends PropertyMetadata
+{
+    /**
+     * @var string|number|null
```
alexander-schranz: number?

src/Sulu/Bundle/AdminBundle/FormMetadata/FormMetadataMapper.php
danrot: No, I think that is not necessary anymore :slightly_smiling_face:

src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/ConstMetadata.php
danrot: At first I thought you were right, but when thinking about it again I would leave it like this, but also return a schema if `null` is a returned. This way we could also say that a value has to be `null` (who knows if we need that one day?)

src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/PropertyMetadata.php
danrot: The thing with JSON Schema is that if a field is required or not is not returned by the field itself, but by the schema it belongs to. We make use of this in the [`SchemaMetadata`](https://github.com/sulu/sulu/blob/d5d841b/src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/SchemaMetadata.php#L47)

src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/ConstMetadata.php
danrot: What do you want to say by that? A field in a form can return about anything, so we should at least cover these two (I don't imagine right now that people are using this in combination with objects as well).

src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/ConstMetadata.php
danrot: Although what I've just described is not working because if I remove the null check in the `toJsonSchema` method of this class, then all the assigned properties say that `null` is their constant value, and the validation fails for a lot :see_no_evil: I would leave it like it is at the moment.

nnatter: yeah returning a schema in the case of `null` it makes sense 🙂

src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/PropertyMetadata.php
nnatter: 👍

src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/ConstMetadata.php
alexander-schranz: Number does not exist as type in PHP. https://docs.phpdoc.org/guides/types.html
PHP only know about int|float

PR #4890, opened 2019-11-13
Add method to handle routables without the routable interface

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | none
| Related issues/PRs | none
| License | MIT
| Documentation PR | none

#### What's in this PR?

This PR add a method to the `RouteManager` to handle routables without the `RoutableInterface`.

#### Why?

This will be used by the SuluContentBundle.

#### Example Usage

~~~php
        $this->routeManager->createOrUpdateByAttributes(
            $localizedContentDimension->getContentClass(),
            (string)$localizedContentDimension->getContentId(),
            $dimension->getLocale(),
            $routePath
        );
~~~

#### BC Breaks/Deprecations

New method in the `RouteManagerInterface` and `RouteRepositoryInterface`.

#### To Do

### DIFF
--- a/src/Sulu/Bundle/RouteBundle/Manager/RouteManager.php
+++ b/src/Sulu/Bundle/RouteBundle/Manager/RouteManager.php
@@ -37,11 +37,6 @@ class RouteManager implements RouteManagerInterface
      */
     private $routeRepository;
 
-    /**
-     * @param ChainRouteGeneratorInterface $chainRouteGenerator
-     * @param ConflictResolverInterface $conflictResolver
-     * @param RouteRepositoryInterface $routeRepository
-     */
     public function __construct(
         ChainRouteGeneratorInterface $chainRouteGenerator,
         ConflictResolverInterface $conflictResolver,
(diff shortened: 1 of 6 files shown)


src/Sulu/Bundle/RouteBundle/Manager/RouteManager.php
```diff
@@ -98,24 +87,60 @@ public function update(RoutableInterface $entity, $path = null, $resolveConflict
...
-        $historyRoute = $entity->getRoute()->setHistory(true)->setTarget($route);
-        $route->addHistory($historyRoute);
+        $route = $this->handleHistoryRoutes($entity->getRoute(), $route);
+
+        $entity->setRoute($route);
+
+        return $route;
+    }
+
+    public function createOrUpdateByAttributes(string $entityClass, string $id, string $locale, string $path): RouteInterface
```
alexander-schranz: What if we create a `RouteFactory::create` service and the above function use it also?

wachterjohannes: why a service? a method inside this class would also be enough i think :)

alexander-schranz: Single responsibility principe 😄

wachterjohannes:
@alexander-schranz this PR is ready

PR #4893, opened 2019-11-14
Add RouteableStructureInterface

### LINKED ISSUE #3452: CacheBehavior
| Q | A
| --- | ---
| Bug? | no
| New Feature? | yes
| Sulu Version | -- / develop
| Browser Version | --

#### Actual Behavior

CacheLifetimeEnhancer:
```
if (!$structure instanceof PageInterface) {
    return;
}
```

#### Expected Behavior

Maybe implement a CacheBehavior, this should be discussed.

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #3452
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

Add RouteableStructureInterface and fixes some phpstan issues.

#### Why?

Currently only a PageInterface set the CacheLifetime. And in ArticleBundle and ContentBundle we also use the Controller and View variables but they are in no interface to generalize this I moved this functions into an own interface.

#### To Do

### DIFF
--- a/src/Sulu/Component/Content/Compat/RoutableStructureInterface.php
+++ b/src/Sulu/Component/Content/Compat/RoutableStructureInterface.php
@@ -1,0 +1,36 @@
+<?php
+
+/*
+ * This file is part of Sulu.
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Component\Content\Compat;
+
+interface RoutableStructureInterface
+{
+    /**
+     * twig template of template definition.
+     *
+     * @return string
+     */
+    public function getView();
+
+    /**
+     * controller which renders the template definition.
+     *
+     * @return string
+     */
+    public function getController();
+
+    /**
+     * cacheLifeTime of template definition.
+     *
+     * @return array
+     */
+    public function getCacheLifeTime();
+}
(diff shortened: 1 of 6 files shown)


src/Sulu/Component/Content/Compat/RoutableStructureInterface.php
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Component\Content\Compat;
+
+interface RoutableStructureInterface
```
alexander-schranz: @danrot not sure about this name 🤔

danrot: Not sure about the name either, and I don't really understand why we need that... Is there a pressing reason for this? It's a change with a certain risk, and to me it feels like this part will change a lot in the future anyway :see_no_evil: 

If there is a pressing reason, why not moving that part to the `StructureInterface`? Don't structures always have these properties as well? Or is there something like a `SnippetStructure` as well?

alexander-schranz:
I will close this as this kind of interface was implemented in the `SuluContentBundle` calling `RoutableInterface`.  I think adding it for Compat or PHPCR is currently not needed.

PR #4895, opened 2019-11-14
Added positionName to ContactAccountSelection

### LINKED ISSUE #5358: Contact Account Selection only returns ID of position
# Problem description

The contact account Selection new to Sulu 2 only returns the ID of a Job Position. Currently there seems to be no Solution for getting the Job Position of a Contact within the Contact Account Selection. As a quick Solution you are able to load the whole contact with the sulu_resolve_contact Twig extension and then check the AccountContact relation. There is also the possibility to write a user defined twig extension which loads contact positions over the Repository.

# Proposed solution

Either enhance the Serialization of the contact to also include the position (preferably like in sulu 1.6 where you could use contact.position.position, as this would mean no rewriting of code is needed in update cases) or provide a twig extension which loads the position of contact by the given id.

martinlagler:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #5358
| License | MIT

#### What's in this PR?

Added getPositionName function to Contact API.

#### Why?

To get the Contact position name, so you don't have to resolve it by the id.

### DIFF
--- a/src/Sulu/Bundle/ContactBundle/Api/Contact.php
+++ b/src/Sulu/Bundle/ContactBundle/Api/Contact.php
@@ -246,6 +246,26 @@ class Contact extends ApiWrapper
     }
 
     /**
+     * Get position.
+     *
+     * @return string
+     *
+     * @VirtualProperty
+     * @SerializedName("positionName")
+     * @Groups({"contactPosition"})
+     */
+    public function getPositionName()
+    {
+        $position = $this->entity->getPosition();
+
+        if (!$position) {
+            return null;
+        }
+
+        return $position->getPosition();
+    }
+
+    /**
      * Set birthday.
      *
      * @param \DateTime $birthday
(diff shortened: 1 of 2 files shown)


src/Sulu/Bundle/ContactBundle/Api/Contact.php
```diff
@@ -245,6 +245,26 @@ public function getPosition()
...
     }
 
+    /**
+     * Get position.
+     *
+     * @return string
+     *
+     * @VirtualProperty
+     * @SerializedName("positionName")
+     * @Groups({"contactPosition"})
```
danrot: I think I would make that part of the `fullContact` group instead.

danrot:
And I think that's something we could consider a bug... Because of that I would even backport it to the 1.6 release of Sulu. That would mean that PR should go into the `release/1.6` branch.

src/Sulu/Bundle/ContactBundle/Api/Contact.php
alexander-schranz: @danrot then we should also join the position else it will get lazy loaded

danrot: That's right, thought that was already joined, but it seems not the `co_position` table, which finally holds the name...

danrot:
Replaced by #5380

PR #4897, opened 2019-11-15
Add the SymfonyTestsListener to all phpunit configuration files

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR makes sure that the deprecations are listened on every phpunit call.

#### Why?

Because this setting was not used on my local machine.

#### To Do

- [ ] Add settings to show internal deprecations

### DIFF
--- a/src/Sulu/Bundle/MarkupBundle/phpunit.xml.dist
+++ b/src/Sulu/Bundle/MarkupBundle/phpunit.xml.dist
@@ -20,4 +20,8 @@
     <php>
         <env name="APP_ENV" value="test"/>
     </php>
+
+    <listeners>
+        <listener class="Symfony\Bridge\PhpUnit\SymfonyTestsListener" />
+    </listeners>
 </phpunit>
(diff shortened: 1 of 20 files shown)


src/Sulu/Bundle/MarkupBundle/phpunit.xml.dist
```diff
@@ -20,4 +20,8 @@
     <php>
         <env name="APP_ENV" value="test"/>
```
alexander-schranz: missing the `SYMFONY_DEPRECATIONS_HELPER` definition here

danrot: Good catch! Added it :slightly_smiling_face:

PR #4904, opened 2019-11-23
Fix some php-stan level 2 errors

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Updates php-docs for phpstan level 2

#### Why?

Because we want a stricter static code analysis.

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -30,3 +30,4 @@ parameters:
     ignoreErrors:
         - '#Service "test_user_provider" is not registered in the container.#'
         - '#Method Symfony\\Contracts\\EventDispatcher\\EventDispatcherInterface::dispatch\(\) invoked with 2 parameters, 1 required.#'
+        - '#Call to an undefined method [a-zA-Z0-9\\_]+::[a-zA-Z0-9\\_]+\(\)#'
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -3,7 +3,7 @@ includes:
 parameters:
     paths:
         - ./
-    level: 1
+    level: 2
     symfony:
         container_xml_path: %rootDir%/../../../var/cache/admin/dev/adminApp_KernelDevDebugContainer.xml
     excludes_analyse:
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -3,7 +3,7 @@ includes:
 parameters:
     paths:
         - ./
-    level: 2
+    level: 1
     symfony:
         container_xml_path: %rootDir%/../../../var/cache/admin/dev/adminApp_KernelDevDebugContainer.xml
     excludes_analyse:
@@ -30,4 +30,3 @@ parameters:
     ignoreErrors:
         - '#Service "test_user_provider" is not registered in the container.#'
         - '#Method Symfony\\Contracts\\EventDispatcher\\EventDispatcherInterface::dispatch\(\) invoked with 2 parameters, 1 required.#'
-        - '#Call to an undefined method [a-zA-Z0-9\\_]+::[a-zA-Z0-9\\_]+\(\)#'
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -29,3 +29,4 @@ parameters:
         - %currentWorkingDirectory%/*/Tests/*
     ignoreErrors:
         - '#Service "test_user_provider" is not registered in the container.#'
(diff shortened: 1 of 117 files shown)


phpstan.neon
```diff
@@ -3,7 +3,7 @@ includes:
 parameters:
     paths:
         - ./
-    level: 1
+    level: 2
```
nnatter: can you reset this file? then we could merge the fixes that are already implemented 🙂

src/Sulu/Component/Content/Export/WebspaceExport.php
```diff
@@ -54,7 +55,8 @@ public function __construct(
         ExtensionManagerInterface $extensionManager,
         ExportManagerInterface $exportManager,
         array $formatFilePaths
-    ) {
+    )
+    {
```
alexander-schranz: @Prokyonn run pretty ci

phpstan.neon
```diff
@@ -29,3 +29,5 @@ parameters:
         - %currentWorkingDirectory%/*/Tests/*
     ignoreErrors:
         - '#Service "test_user_provider" is not registered in the container.#'
+        - '#Method Symfony\\Contracts\\EventDispatcher\\EventDispatcherInterface::dispatch\(\) invoked with 2 parameters, 1 required.#'
```
danrot: Do we still support Symfony versions that only work with two parameters? If not I would prefer to rewrite these calls, instead of excluding it from the check.

phpstan.neon
```diff
@@ -29,3 +29,5 @@ parameters:
...
+        - '#Call to an undefined method [a-zA-Z0-9\\_]+::[a-zA-Z0-9\\_]+\(\)#'
```
danrot: That does not look like a reasonable thing to do... Why do you want to allow every call to an undefined method?

phpstan.neon
alexander-schranz: @danrot sadly they can not be changed  as the event subscribers are subscribed to the optional name and if we changed them they will not longer working.

But phpStan will not error when upgrading to symfony 5.0 there the second parameter is not removed and so its still possible to throw events with names:

https://github.com/symfony/symfony/blob/5.0/src/Symfony/Contracts/EventDispatcher/EventDispatcherInterface.php#L30

I think they keeped the eventName as the [workflow component](https://symfony.com/doc/4.4/workflow.html#using-events) or other dynamic event systems need this type dynamic event names.

phpstan.neon
Prokyonn: I just ignored those errors to fix the easier ones first, this should not be merged like this.

src/Sulu/Component/Content/Compat/Structure/LegacyPropertyFactory.php
```diff
@@ -68,7 +68,8 @@ public function createTranslatedProperty($property, $locale, StructureInterface
     /**
      * Create a new property.
      *
-     * @param Item $item
+     * @param ItemMetadata $property
+     * @param null|StructureInterface $structure
```
danrot: Actually we were saying that we only add these params if they add some value. In that case all this information is already in the function signature itself, therefore I would just remove these two lines.

src/Sulu/Component/Webspace/NavigationContext.php
```diff
@@ -43,11 +43,11 @@ public function getKey()
     }
 
     /**
-     * @param $locale
+     * @param string $locale
      *
      * @return null|string
      */
-    public function getTitle($locale)
+    public function getTitle(string $locale)
```
danrot: This is also duplicated information... I would say remove the type in this line, to be absolutely sure that we don't make a BC break now.

src/Sulu/Component/Webspace/Portal.php
```diff
@@ -189,9 +189,9 @@ public function getXDefaultLocalization()
     /**
      * Adds an environment to this portal.
      *
-     * @param $environment Environment The environment to add
+     * @param Environment $environment Environment The environment to add
      */
-    public function addEnvironment($environment)
+    public function addEnvironment(Environment $environment)
```
danrot: Same as above, remove the type here to avoid BC breaks.

PR #4906, opened 2019-11-25
Provide a default user provider for the firewall and fix strange error under symonfy 4.4

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | https://github.com/sulu/skeleton/pull/26, symfony/symfony#34596
| License | MIT
| Documentation PR | -

#### What's in this PR?

Provide a default user provider for the firewall

#### Why?

Avoid following error message in test environment when using Symfony 4.4.

> Not configuring explicitly the provider for the "anonymous" listener on "admin" firewall is ambiguous as there is more than one registered provider.   

#### To Do

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,23 @@
 # Upgrade
 
+## dev-release/2.0
+
+### Admin Default User Provider for Test Environment
+
+To fix compatibility to Symfony 4.4 version set the default provider for the sulu admin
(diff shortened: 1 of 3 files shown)


alexander-schranz:
CI requires #4905

PR #4925, opened 2019-11-29
find and findChildren should return either a NavigationItem or null

tstauer:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

The NavigationItem features the methods find and findChildren.
This pull requests fixes the phpdoc and behaviour of these methods by returning Navigationitem or null.

#### Why?

In case of  find it is documented that it would return false if no navigation item is found.
However find returns void instead of false if no navigation item is found.

In case of findChildren the phpdoc states that it would return a NavigationItem or null.
However findChildren returns a NavigationItem or void instead.


alexander-schranz:
@tsauer thank you!

PR #4927, opened 2019-11-30
add used PHP extensions to composer.json

tstauer:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

This pull requests add the used PHP extensions to the composer.json file.

#### Why?

Composer should fail at install time instead of having runtime errors in Sulu.
One can compile a nice list of dependencies when migrating servers.

It's a very good practice to always add the used dependencies.

Citing https://getcomposer.org/doc/04-schema.md
> Note: It is important to list PHP extensions your project requires.

| Extension | Usage
| --- | ---
| ext-json | All over the place
| ext-dom | \Sulu\Bundle\MediaBundle\Media\FormatLoader\BaseXmlFormatLoader, \Sulu\Bundle\AdminBundle\FormMetadata\FormXmlLoader
| ext-fileinfo | \Sulu\Bundle\MediaBundle\Media\FormatManager\FormatManager
| ext-simplexml | \Sulu\Bundle\PageBundle\DependencyInjection\Compiler\WebspacesPass
| ext-zip | \Sulu\Bundle\AdminBundle\Command\DownloadBuildCommand

I added neihter mbstring nor iconv because only one of these extension is required and AFAIK there is no way to specify that only one of those is needed.

Also ext-openssl is optional. I suggest one should have a look at the TokenGenerator as it seems a bit dated.

#### Example Usage

~~~bash
php -d memory_limit=1G composer.phar install
~~~

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,6 +26,11 @@
     },
     "require": {
         "php": "^7.2",
+        "ext-SimpleXML": "*",
+        "ext-dom": "*",
+        "ext-fileinfo": "*",
+        "ext-json": "*",
+        "ext-zip": "*",
         "aferrandini/urlizer": "^1.0",
         "dantleech/phpcr-migrations-bundle": "^1.0",
         "doctrine/annotations": "^1.2",
--- a/composer.json
+++ b/composer.json
@@ -26,11 +26,9 @@
     },
     "require": {
         "php": "^7.2",
-        "ext-SimpleXML": "*",
         "ext-dom": "*",
-        "ext-fileinfo": "*",
         "ext-json": "*",
-        "ext-zip": "*",
+        "ext-simplexml": "*",
         "aferrandini/urlizer": "^1.0",
         "dantleech/phpcr-migrations-bundle": "^1.0",
         "doctrine/annotations": "^1.2",
@@ -154,6 +152,8 @@
         "sulu/translate-bundle": "self.version"
     },
     "suggest": {
+        "ext-fileinfo": "* - Needed to recognize file types of uploaded files",
+        "ext-zip": "* - Needed by the download commands for the admin build and its translations",
         "handcraftedinthealps/zendsearch": "To use the PHP based Zend Search library (based on Lucene)",
         "league/flysystem": "^1.0 - Needed for remote media-storages",
         "league/flysystem-aws-s3-v3": "^1.0.1 - Needed for AWS S3 compatible storages (aws s3 or minio)",


alexander-schranz:
Would not add ext to sulu/sulu this avoids creating/using polyfills for this. And as not every project using sulu need all functionality ext are optional and no requirement.

tstauer:
> Would not add ext to sulu/sulu this avoids creating/using polyfills for this. And as not every project using sulu need all functionality ext are optional and no requirement.

At the beginning I thought the same but then I changed my mind (hence the PR).

These reasons came into my mind:
* The tests should depend on all of the used extensions.
* It shows the packaging of the bundles could be improved.
* It's better to fail on install time than on runtime.

Feel free to disagree.

danrot:
@alexander-schranz Is there any polyfill for a complete extension? If I check e.g. [symfony/polyills](https://github.com/symfony/polyfill) it only adds certain functions, not entire extensions. However, I would not add `ext-zip` as an absolute requirement, because it would not be required if people are building the administration javascript on their own using NPM and so on.

danrot:
I just checked, there is even an [`ext-` dependency in symfony](https://github.com/symfony/symfony/blob/1b4ab81/composer.json#L20). Can we also list optional `ext-` dependencies? Then we could add `ext-zip` as an optional dependency.

alexander-schranz:
@danrot @tstauer Would move `ext-fileinfo` and `ext-zip` to `suggest` section the rest I think is currently needed to build to container or?
But think should be no problem as they are compiled normally with PHP package.

Maybe in future we could have a look if we could replace ext-dom with symfony/dom-crawler or something like that. And think ext-simplexml would also be gone when moving the webspace to the symfony config package.

danrot:
Is the media section working at all without `fileinfo`? If yes I would agree the it should be moved to the `suggest` section.

We should definitely remove them again in case these extensions are not needed anymore at some point. However, I could imagine the `ext-simplexml` is also used by the symfony config if you write XML configs... But would still be optional then I guess. However, in the moment it is an absolute requirement anyway, so we should list it.

composer.json
```diff
@@ -26,6 +26,11 @@
     },
     "require": {
         "php": "^7.2",
+        "ext-SimpleXML": "*",
```
danrot: Does this have to be written this way? Why not using `ext-simplexml`, like in all other cases?

danrot:
@alexander-schranz I've updated the PR as we've discussed, can you have another look and merge if it is ok?

PR #4935, opened 2019-12-09
Add method to add resource-store-properties to metadata-request

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | yes
| Fixed tickets | none
| Related issues/PRs | sulu/SuluFormBundle#211
| License | MIT
| Documentation PR | none

#### What's in this PR?

This PR add following methods:

* `ListViewBuilderInterface::addResourceStorePropertiesToListMetadata`
* `ListViewBuilderInterface::addRequestParameters`
* `FormViewBuilderInterface::addRequestParameters`
* `FormOverlayListViewBuilderInterface::addRequestParameters`
* `PreviewFormViewBuilderInterface::addRequestParameters`

#### Why?

See PR in sulu-form-bundle.

#### BC Breaks/Deprecations

New methods in interfaces.

#### To Do

- [x] JS tests
- [x] PHPUnit tests

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -5,6 +5,16 @@
 When upgrading also have a look at the changes in the
 [sulu skeleton](https://github.com/sulu/skeleton/compare/2.0.2...2.0.3).
 
+### New methods in ViewBuilderInterfaces
+
+Following methods has been added:
+
+* `ListViewBuilderInterface::addResourceStorePropertiesToListMetadata`
+* `ListViewBuilderInterface::addRequestParameters`
+* `FormViewBuilderInterface::addRequestParameters`
+* `FormOverlayListViewBuilderInterface::addRequestParameters`
+* `PreviewFormViewBuilderInterface::addRequestParameters`
+
 ### Symfony/templating requirement removed
 
 Sulu does not longer need the `symfony/templating` package which is deprecated.
--- a/src/Sulu/Bundle/AdminBundle/Admin/View/FormOverlayListViewBuilder.php
+++ b/src/Sulu/Bundle/AdminBundle/Admin/View/FormOverlayListViewBuilder.php
@@ -20,6 +20,7 @@ class FormOverlayListViewBuilder implements FormOverlayListViewBuilderInterface
         ListViewBuilderTrait::setEditViewToView insteadof FormViewBuilderTrait;
         ListViewBuilderTrait::addLocalesToView insteadof FormViewBuilderTrait;
         ListViewBuilderTrait::addToolbarActionsToView insteadof FormViewBuilderTrait;
+        ListViewBuilderTrait::addRequestParametersToView insteadof FormViewBuilderTrait;
     }
     use TabViewBuilderTrait;
 
@@ -51,8 +52,13 @@ class FormOverlayListViewBuilder implements FormOverlayListViewBuilderInterface
         return $this;
     }
 
+    /**
+     * @deprecated The usage of the "setRequestParameters" method in the FormOverlayListViewBuilder is deprecated. Please use "addRequestParameters" instead.
+     */
     public function setRequestParameters(array $requestParameters): FormOverlayListViewBuilderInterface
     {
+        @trigger_error('The usage of the "setRequestParameters" method in the FormOverlayListViewBuilder is deprecated. Please use "addRequestParameters" instead.', E_USER_DEPRECATED);
+
         $this->setRequestParametersToView($this->view, $requestParameters);
 
         return $this;
@@ -175,6 +181,13 @@ class FormOverlayListViewBuilder implements FormOverlayListViewBuilderInterface
         $oldResourceStorePropertiesToFormRequest = $this->view->getOption('resourceStorePropertiesToFormRequest');
         $newResourceStorePropertiesToFormRequest = $oldResourceStorePropertiesToFormRequest ? array_merge($oldResourceStorePropertiesToFormRequest, $resourceStorePropertiesToFormRequest) : $resourceStorePropertiesToFormRequest;
         $this->view->setOption('resourceStorePropertiesToFormRequest', $newResourceStorePropertiesToFormRequest);
+
+        return $this;
+    }
+
+    public function addRequestParameters(array $requestParameters): FormOverlayListViewBuilderInterface
+    {
+        $this->addRequestParametersToView($this->view, $requestParameters);
 
         return $this;
     }
--- a/src/Sulu/Bundle/AdminBundle/Admin/View/FormViewBuilder.php
+++ b/src/Sulu/Bundle/AdminBundle/Admin/View/FormViewBuilder.php
@@ -38,8 +38,13 @@ class FormViewBuilder implements FormViewBuilderInterface
         return $this;
     }
 
+    /**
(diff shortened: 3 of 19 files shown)


UPGRADE.md
```diff
@@ -5,6 +5,16 @@
 When upgrading also have a look at the changes in the
 [sulu skeleton](https://github.com/sulu/skeleton/compare/2.0.2...2.0.3).
 
+### New methods for Form/ListViewBuilder
```
danrot: Would write "New methods in ViewBuilderInterfaces", because otherwise it is not really a breaking change.

src/Sulu/Bundle/AdminBundle/Admin/View/FormOverlayListViewBuilder.php
```diff
@@ -51,13 +51,6 @@ public function setFormKey(string $formKey): FormOverlayListViewBuilderInterface
         return $this;
     }
 
-    public function setRequestParameters(array $requestParameters): FormOverlayListViewBuilderInterface
```
danrot: That's a pretty hard BC break IMO. I would keep the `setRequestParameters` function, and add it to the ViewBuilders where it is missing..

src/Sulu/Bundle/AdminBundle/Admin/View/ViewBuilderTrait.php
```diff
@@ -57,4 +57,12 @@ public function addRerenderAttribute(string $attribute): ViewBuilderInterface
 
         return $this;
     }
+
+    private function addRequestParametersToView(View $route, array $requestParameters): void
```
danrot: I don't think that this makes sense on the general `ViewBuilderTrait`. I would move it to the traits that actually need this function.

src/Sulu/Bundle/AdminBundle/Resources/js/views/List/List.js
```diff
@@ -124,7 +125,9 @@ class List extends React.Component<Props> {
 
         const metadataOptions = this.buildMetadataOptions(
             attributes,
-            routerAttributesToListMetadata
+            props.resourceStore,
```
danrot: I would try to keep the order similar to the function above, and move the `resourceStore` at the end.

src/Sulu/Bundle/AdminBundle/Admin/View/FormOverlayListViewBuilder.php
danrot: However, if we want to remove this function we could deprecate it. But I don't really see a reason to do so, because this probably won't make any huge problems in the future.

wachterjohannes: the function was in no interface - so i would remove it because we never have setters (always add methods)

src/Sulu/Bundle/AdminBundle/Admin/View/ViewBuilderTrait.php
wachterjohannes: cannot do that 🙈 i would have to add it to the `ListViewBuilderTrait` and the `FormViewBuilderTrait` - but then the `FormOverlayListViewBuilder` cannot implement both traits.

src/Sulu/Bundle/AdminBundle/Admin/View/FormOverlayListViewBuilder.php
danrot: You even had to change some code in our codebase, because of this change... And people had no other choice than using this function if they wanted to add some hardcoded values, so I think there is quite a high chance that people have already used.

Of course you could argue, that we only had setters until now, but IMO that still does not justify the total removal of this method. If we say we don't want to have that method in the future, I would just deprecate it, and mention the `addRequestParameters` method in the deprecation message.

But I think I would go even further, and just keep it as it is, and maybe even add it in the interface, because it is a very simple function, that probably won't add a lot of maintenance burden.

src/Sulu/Bundle/AdminBundle/Admin/View/ViewBuilderTrait.php
danrot: That's not true - actually we already do that in the [`FormOverlayListViewBuilder` using the `insteadof` keyword](https://github.com/sulu/sulu/blob/master/src/Sulu/Bundle/AdminBundle/Admin/View/FormOverlayListViewBuilder.php#L18-L22), because these two traits already share more method names.

That also brings me to another question: Do we want to support adding different `requestParameters` for the list and form requests?

src/Sulu/Bundle/AdminBundle/Admin/View/FormOverlayListViewBuilder.php
```diff
@@ -53,6 +54,8 @@ public function setFormKey(string $formKey): FormOverlayListViewBuilderInterface
 
     public function setRequestParameters(array $requestParameters): FormOverlayListViewBuilderInterface
```
danrot: I think I would also add the `@deprecated` annotation, because then it will also be shown in IDEs that this function should not be used.

src/Sulu/Bundle/AdminBundle/Admin/View/FormViewBuilder.php
```diff
@@ -40,6 +40,8 @@ public function setFormKey(string $formKey): FormViewBuilderInterface
 
     public function setRequestParameters(array $requestParameters): FormViewBuilderInterface
     {
+        @trigger_error('The usage of the "setRequestParameters" method in the FormOverlayListViewBuilder is deprecated. Please use "addRequestParameters" instead.', E_USER_DEPRECATED);
```
danrot: That is not the `FormOverlayListViewBuilder` :slightly_smiling_face:

src/Sulu/Bundle/AdminBundle/Admin/View/FormViewBuilder.php
```diff
@@ -40,6 +40,8 @@ public function setFormKey(string $formKey): FormViewBuilderInterface
 
     public function setRequestParameters(array $requestParameters): FormViewBuilderInterface
```
danrot: Add `@deprecated` annotation

src/Sulu/Bundle/AdminBundle/Admin/View/PreviewFormViewBuilder.php
```diff
@@ -45,13 +45,6 @@ public function setFormKey(string $formKey): PreviewFormViewBuilderInterface
         return $this;
     }
 
-    public function setRequestParameters(array $requestParameters): PreviewFormViewBuilderInterface
```
danrot: I would not remove this function here either.

src/Sulu/Bundle/AdminBundle/Admin/View/FormViewBuilder.php
```diff
@@ -38,8 +38,13 @@ public function setFormKey(string $formKey): FormViewBuilderInterface
         return $this;
     }
 
+    /**
+     * @deprecated The usage of the "setRequestParameters" method in the FormOverlayListViewBuilder is deprecated. Please use "addRequestParameters" instead.
```
danrot: Wrong class name :see_no_evil:

PR #4937, opened 2019-12-09
Consider installed commands when configuring supported mime types for image conversion

danrot:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR only generate thumbnail URLs if the required tools on the server are installed to do so.

#### Why?

Because otherwise e.g. the thumbnail used for video files in the list view just shows a broken image instead of something useful.

Also, it saves us a request for the generated image, which should lead to better performance.


src/Sulu/Bundle/MediaBundle/DependencyInjection/SuluMediaExtension.php
```diff
@@ -335,4 +341,23 @@ private function configureStorage(array $config, ContainerBuilder $container, Lo
...
         $container->setAlias('sulu_media.storage', 'sulu_media.storage.' . $storage)->setPublic(true);
     }
+
+    private function getSupportedMimeTypes($ghostScriptPath, $ffmpegBinary, $ffprobeBinary)
+    {
+        $mimeTypes = ['image/*'];
+
+        if ($ffmpegBinary
+            && $ffprobeBinary
+            && shell_exec('which ' . $ffmpegBinary)
```
alexander-schranz: `which` will not work on windows i guess

danrot: I looked it up, seems to exist on windows as well. The thing is that I don't want the actual command to be executed.

danrot:
This causes the following message multiple lines in appveyor:

```
The system cannot find the path specified.
```

I think that is because by default we use paths that are not valid windows paths (i.e. not starting with a letter representing the drive). I tried silencing them by redirecting to `/dev/null`, but even if I also redirect the errors using `which command > /dev/null 2>&1` the error stills shows :confused:

src/Sulu/Bundle/MediaBundle/DependencyInjection/SuluMediaExtension.php
```diff
@@ -335,4 +341,23 @@ private function configureStorage(array $config, ContainerBuilder $container, Lo
...
         $container->setAlias('sulu_media.storage', 'sulu_media.storage.' . $storage)->setPublic(true);
     }
+
+    private function getSupportedMimeTypes($ghostScriptPath, $ffmpegBinary, $ffprobeBinary)
+    {
+        $mimeTypes = ['image/*'];
+
+        if ($ffmpegBinary
+            && $ffprobeBinary
+            && shell_exec('which ' . $ffmpegBinary . ' > /dev/null')
```
alexander-schranz: For windows the commands to check if ghostscript or ffmpeg is available are the following:

```cmd
where.exe ffmpeg.exe /Q
where ffmpeg.exe /Q
where.exe "C:\ffmpeg\bin:ffmpeg.exe"
```

```cmd
where.exe gswin64.exe /Q
where.exe "C:\Program Files\gs\gs9.50\bin:gswin64.exe" /Q
```

Attention if search in a specific path the `path` is splitted with `:` not with `\`.

`\Q` means no output and we need the exit code if it is found > /dev/null is not supported by windows and will output the warning shown in the CI currently.

So the correct way on windows could be:

```php
exec('where.exe "C:\Program Files\gs\gs9.50\bin:gswina64.exe"', $output, $exitCode);
$ghostScriptAvailable = (0 === $exitCode);
```

danrot: The `-q` option shows an error in linux... But at least I know now where the error origins from... I think there is no other way than building an `if` to be able to build this in two different ways...

danrot:
@alexander-schranz I think I solved this issue now, but it is hard to tell if it is really working, and I don't want to setup multiple appveyor builds to test with and without ffmpeg/ghostscript... Can you maybe test on your windows machine, if everyhing else is ok for you?

danrot:
There is indeed an `ExecutableFinder` in Symfony. That would avoid having an `if` for the operating system in our code... I think I am going to try that afterwards.

src/Sulu/Bundle/MediaBundle/Tests/Unit/DependencyInjection/SuluMediaExtensionTest.php
```diff
@@ -9,7 +9,17 @@
  * with this source code in the file LICENSE.
  */
 
-namespace Sulu\Bundle\SnippetBundle\Tests\Unit\DependencyInjection;
+namespace Symfony\Component\Process;
```
alexander-schranz: Can this not throw some redeclare error?

Would go with something like the following instead:

```
    public function __construct(ExecutableFinder $executableFinder = null)
    {
        $this->executableFinder = $executableFinder ?: new ExecutableFinder();
    }
```

to mock the ExecutableFinder.

Symfony uses this to Unit Test there Messenger Connections.
https://github.com/symfony/symfony/blob/65a988880d719f0de2329b39eb2e7c1b56d7532f/src/Symfony/Component/Messenger/Transport/AmqpExt/Connection.php#L74

danrot: Well, it does not throw any errors in our test suite... The other way feels a bit weird, because the class is not really designed to be used with something else... But it is ok, I am going to change it.

src/Sulu/Bundle/MediaBundle/DependencyInjection/SuluMediaExtension.php
```diff
@@ -335,4 +352,28 @@ private function configureStorage(array $config, ContainerBuilder $container, Lo
...
+        if ($ghostScriptPath && $this->checkCommandAvailability($ghostScriptPath)) {
+            $mimeTypes[] = 'application/pdf';
+        }
+
+        return $mimeTypes;
+    }
+
+    private function checkCommandAvailability($command)
+    {
+        return null !== $this->executableFinder->find($command);
```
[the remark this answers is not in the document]
alexander-schranz: @danrot maybe we should go with something like because the executeableFinder search only in PATHs dirs we would need to at the given dir as extraDir if its not inside the PATH:

```
return null !== $this->executableFinder->find(basename($command), null, [dirname($command)]);
```

src/Sulu/Bundle/MediaBundle/DependencyInjection/SuluMediaExtension.php
```diff
@@ -335,4 +352,28 @@ private function configureStorage(array $config, ContainerBuilder $container, Lo
...
+        if ($ghostScriptPath && $this->checkCommandAvailability($ghostScriptPath)) {
+            $mimeTypes[] = 'application/pdf';
+        }
+
+        return $mimeTypes;
+    }
+
+    private function checkCommandAvailability($command)
+    {
+        return null !== $this->executableFinder->find($command) || is_executable($command);
```
danrot: @alexander-schranz WDYT about this solution? The one with `basename` and `dirname` you suggested has the problem that it can also find stuff, that does not really exists, e.g. `/some/stupid/not/existing/path/gs` would be found, because `gs` is in the PATH.

alexander-schranz: Sounds good for me but would silence the `is_executable` like its done in the executableFinder with `@is_executable` to ignore a warning about open_basedir restriction effect when open_basedir is configured.

e.g.:

```
 php -dopen_basedir=~/Documents -r "var_dump(is_executable('/tmp/world'));
```

vs

```
 php -dopen_basedir=~/Documents -r "var_dump(@is_executable('/tmp/world'));
```

danrot: Fixed

PR #4938, opened 2019-12-11
Add linting of container to ci

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

Add linting of container to test if services are defined correctly.

#### Why?

Invalid definition of services should be avoided with this linting commands.

Will need symfony 4.4.2 release because of the following bugfix symfony/symfony#34794

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,7 @@ matrix:
         - SYMFONY_DEPRECATIONS_HELPER="weak"
         - PHPSTAN=true
         - PECL_INSTALL=imagick
+        - LINT_CONTAINER=true
       services:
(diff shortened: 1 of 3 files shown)


danrot:
If we only run this in the build not using `--prefer-lowest` we should also be able to use that in a sooner release, shouldn't it?

alexander-schranz:
@danrot 4.4.2 is released now so this PR is ready.

PR #4939, opened 2019-12-11
Add PageTreeRoute

luca-rath:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Adds page tree route content type + field type.

#### Why?

Because it's needed by ArticleBundle and should be implemented in Sulu Core.

- [x] JS Component
- [x] JS Tests
- [x] ContentType
- [x] Save parent uuid on route
- [x] PageTreeUpdater


src/Sulu/Bundle/RouteBundle/Resources/js/index.js
```diff
@@ -12,3 +13,15 @@ fieldRegistry.add(
...
+
+fieldRegistry.add(
+    'page_tree_route',
+    PageTreeRoute,
+    {
+        historyResourceKey: 'routes',
+        modeResolver: () => {
+            return Promise.resolve('leaf');
+        },
+        options: {history: true},
```
danrot: We should probably think about which of these options make sense in that case. I think we could remove all of them, unless you are (or intent to) register these component with other options as well.

danrot (changes requested):
At least some of the changes I really request :see_no_evil: Especially for the PHP part @wachterjohannes should have another look.

src/Sulu/Bundle/RouteBundle/Document/Subscriber/PageTreeRouteSubscriber.php
```diff
@@ -0,0 +1,170 @@
...
+
+    /**
+     * Update route-paths of documents which are linked to the given page-document.
+     *
+     * @param AbstractMappingEvent $event
+     */
+    public function handlePublish(AbstractMappingEvent $event)
+    {
+        $document = $event->getDocument();
+        if (!$document instanceof PageDocument || !$this->hasChangedResourceSegment($document)) {
```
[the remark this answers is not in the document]
luca-rath: Maybe it's more readable using
```php
public function handlePublish(AbstractMappingEvent $event)
{
    $document = $event->getDocument();
    if ($document instanceof PageDocument && $this->hasChangedResourceSegment($document)) {
        $this->routeUpdater->update($document);
    }
}
```
In my opinion, this is exactly the behavior, which is necessary here.

src/Sulu/Bundle/RouteBundle/Document/Subscriber/PageTreeRouteSubscriber.php
```diff
@@ -0,0 +1,170 @@
...
+use Sulu\Component\DocumentManager\DocumentManagerInterface;
+use Sulu\Component\DocumentManager\Event\AbstractMappingEvent;
+use Sulu\Component\DocumentManager\Event\MoveEvent;
+use Sulu\Component\DocumentManager\Events;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+/**
+ * Handles relation between documents and pages.
+ */
+class PageTreeRouteSubscriber implements EventSubscriberInterface
```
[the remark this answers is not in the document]
luca-rath: It triggers the PageTreeUpdater, if the url of a page has been changed or a page has been moved to another parent.

src/Sulu/Bundle/RouteBundle/PageTree/PageTreeRepository.php
```diff
@@ -0,0 +1,217 @@
...
+use Sulu\Component\Route\Document\Behavior\RoutableBehavior;
+
+/**
+ * Update the route of documents synchronously.
+ */
+class PageTreeRepository implements PageTreeUpdaterInterface, PageTreeMoverInterface
+{
+    const ROUTE_PROPERTY = 'routePath';
+
+    const TAG_NAME = 'sulu_route.routePath';
```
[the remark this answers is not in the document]
luca-rath: In my opinion, the best place to define them is the `PageTreeRouteContentType`. But it would be a little bit confusing then, because the `PageTreeRouteContentType` doesn't even need these constants. What do you think?

src/Sulu/Bundle/RouteBundle/Resources/config/services.xml
```diff
@@ -8,6 +8,34 @@
             <tag name="sulu.content.type" alias="route"/>
         </service>
 
+        <service id="sulu_route.content_type.page_tree_route" class="Sulu\Bundle\RouteBundle\Content\Type\PageTreeRouteContentType">
```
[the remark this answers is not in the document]
luca-rath: I think yes. The `page_tree_route` works like normal, also if `page_tree_update.xml` is not loaded. The only difference is, that the child routes don't get updated, when the parent route changes.

src/Sulu/Bundle/RouteBundle/Resources/js/containers/Form/fields/PageTreeRoute.js
```diff
@@ -0,0 +1,177 @@
+// @flow
+import React from 'react';
+import type {FieldTypeProps} from 'sulu-admin-bundle/containers/Form/types';
+import {observer} from 'mobx-react';
+import {action, observable} from 'mobx';
+import type {IObservableValue} from 'mobx';
+import userStore from 'sulu-admin-bundle/stores/userStore';
+import {Grid} from 'sulu-admin-bundle/components';
+import SingleSelection from 'sulu-admin-bundle/containers/SingleSelection';
+import ResourceLocator from 'sulu-admin-bundle/components/ResourceLocator';
```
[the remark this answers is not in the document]
luca-rath: `components/index.js` doesn't export ResourceLocator yet. Fixing it.

src/Sulu/Bundle/RouteBundle/Content/Type/PageTreeRouteContentType.php
```diff
@@ -0,0 +1,256 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    public function read(NodeInterface $node, PropertyInterface $property, $webspaceKey, $languageCode, $segmentKey)
+    {
+        $propertyName = $property->getName();
+
+        /*
+         * TODO
```
danrot: Is that really a todo, or should we just remove that line? :see_no_evil: @wachterjohannes @alexander-schranz Is that comment ok for you? Does it explain everything well enough?

src/Sulu/Bundle/RouteBundle/Content/Type/PageTreeRouteContentType.php
```diff
@@ -0,0 +1,243 @@
...
+        $path = $this->resolveConflicts($path);
+        $suffix = '/' . $this->getSuffix($path, $page['path'] ?? '/');
+
+        $propertyName = $property->getName();
+        $node->setProperty($propertyName, $path);
+        $node->setProperty($propertyName . '-suffix', $suffix);
+
+        $pagePropertyName = $propertyName . '-page';
+        if ($node->hasProperty($pagePropertyName)) {
+            $node->getProperty($pagePropertyName)->remove();
```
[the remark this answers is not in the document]
danrot: Can you at least try to explain what's happening here? :thinking:

src/Sulu/Bundle/RouteBundle/DependencyInjection/PageTreeCompilerPass.php
```diff
@@ -0,0 +1,58 @@
...
+            foreach ($structureFactory->getStructures($structureType) as $metadata) {
+                if ($this->hasPageTree($metadata->getProperties())) {
+                    $pageTreeEnabled = true;
+
+                    break 2;
+                }
+            }
+        }
+
+        $container->setParameter('sulu_route.page_tree_enabled', $pageTreeEnabled);
```
danrot: If I read that code correctly we are only setting this to true if one of the pages is using a `page_tree_route`, right? Does that save any performance? Or how does that justify this code? Asking because it is kind of an anti-pattern to rely on services in a compiler pass, we are having troubles in other places because of that.

src/Sulu/Bundle/RouteBundle/Document/Subscriber/PageTreeRouteSubscriber.php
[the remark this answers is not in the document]
danrot: Ah, this is updating all the routes of mostly articles, when a page is updated, right?

src/Sulu/Bundle/RouteBundle/PageTree/PageTreeTrait.php
```diff
@@ -0,0 +1,77 @@
...
+use Sulu\Bundle\RouteBundle\Content\Type\PageTreeRouteContentType;
+use Sulu\Bundle\RouteBundle\Document\Subscriber\RoutableSubscriber;
+use Sulu\Component\Content\Metadata\PropertyMetadata;
+use Sulu\Component\Content\Metadata\StructureMetadata;
+use Sulu\Component\Route\Document\Behavior\RoutableBehavior;
+
+/**
+ * Encapsulates function to extract parent page from structure-metadata.
+ */
+trait PageTreeTrait
```
danrot: Where is that actually used? I can't find another occurrence of this text...

src/Sulu/Bundle/RouteBundle/Resources/js/containers/Form/fields/PageTreeRoute.js
```diff
@@ -0,0 +1,175 @@
...
+            fieldTypeOptions: {
+                modeResolver,
+            },
+        } = props;
+
+        if (!modeResolver) {
+            throw new Error('The "modeResolver" must be a function returning a promise with the desired mode');
+        }
+
+        modeResolver(props).then(action((mode) => this.mode = mode));
```
danrot: Looks like you have ignored my other comment :wink: This is some of the code that is duplicated from the `ResourceLocator` field type... Doesn't it make sense to reuse that field type here?

src/Sulu/Bundle/RouteBundle/Resources/js/containers/Form/fields/PageTreeRoute.js
```diff
@@ -0,0 +1,175 @@
...
+                resourceKey,
+            },
+            fieldTypeOptions: {
+                historyResourceKey,
+                options,
+            },
+        } = this.props;
+
+        return (
+            <>
```
danrot: Until now we have always imported `React.Fragment` and wrote `<Fragment>` here.

src/Sulu/Bundle/RouteBundle/Resources/js/containers/Form/fields/pageTreeRoute.scss
```diff
@@ -0,0 +1,13 @@
+.resource-locator-container {
```
danrot: If we would reuse the `ResourceLocator` field type it would also not be necessary to duplicate this code.

src/Sulu/Bundle/RouteBundle/Resources/js/index.js
danrot: WDYT about this?

danrot (changes requested):

src/Sulu/Bundle/RouteBundle/Resources/js/index.js
luca-rath: I don't know about future plans with the PageTreeRoute. I don't need these options. Just added them, because i saw them in ResourceLocator and from the point of extensibility I expected the same behavior here.

src/Sulu/Bundle/RouteBundle/Content/Type/PageTreeRouteContentType.php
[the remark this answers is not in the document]
luca-rath: I don't know the purpose of this code. The result of this code is to remove the whole property instead of just setting it to null, if the page has been set to null in the admin. Maybe @wachterjohannes can help here.

src/Sulu/Bundle/RouteBundle/Resources/js/index.js
danrot: These are my standpoints on each one of them:

- Since the `page_tree_route` always creates some `Route` object from the `RouteBundle`, I would argue that the `historyResourceKey` will never change, so that could be hardcoded.
- I don't know if the mode can be changed, depending on that it might also be possible to remove the `modeResolver`. I am sure @wachterjohannes can tell us more about it.
- The `options` are passed to the `ResourceLocatorHistory` (could probably have been named a little bit better...) and AFAIK the `routes` resource always needs this `history` flag. So this could also be hardcoded and not be made configurable.

src/Sulu/Bundle/RouteBundle/DependencyInjection/PageTreeCompilerPass.php
luca-rath: I think this is only used to enable/disable the MovePageTreeCommand

src/Sulu/Bundle/RouteBundle/PageTree/PageTreeTrait.php
luca-rath: It's only used by the article bundle so far, but it belongs to the RouteBundle in my opinion

luca-rath: ArticleIndexer and WebspaceSubscriber

wachterjohannes (changes requested):

src/Sulu/Bundle/RouteBundle/Resources/js/containers/Form/fields/PageTreeRoute.js
```diff
@@ -0,0 +1,164 @@
...
+                path,
+            },
+        });
+
+        onFinish();
+    };
+
+    handleSuffixChange = (value: ?string): void => {
+        this.handleChange({
+            ...this.props.value, suffix: value,
```
danrot: Both of them should go into separate lines.

danrot (changes requested):

wachterjohannes (approved):

PR #4961, opened 2020-01-03
Consolidate getMedia methods

### LINKED ISSUE #1297: SuluContactBundle Api Account
As all API functions to get the assets is called "getMedia" and the API to set the media is "setMedia" ... 
however in the contact bundle the names are "getMedias" and "addMedia". This is not really consistent

kfrohwein:
https://github.com/sulu/sulu/issues/1297

I renamed the methods and created some studs. I am not sure what is the best way or demanded way to write deprecation notes. All in all I hope it is fine that way :)

### DIFF
--- a/src/Sulu/Bundle/ContactBundle/Api/Account.php
+++ b/src/Sulu/Bundle/ContactBundle/Api/Account.php
@@ -1066,9 +1066,24 @@ class Account extends ApiWrapper
      *
      * @param MediaInterface $media
      *
+     * @deprecated Please use setMedia instead.
+     *
      * @return Account
+     *
      */
     public function addMedia(MediaInterface $media)
+    {
+        return $this->setMedia($media);
+    }
+
+    /**
+     * Set media.
+     *
+     * @param MediaInterface $media
+     *
+     * @return Account
+     */
+    public function setMedia(MediaInterface $media)
     {
         $this->entity->addMedia($media);
 
@@ -1088,12 +1103,24 @@ class Account extends ApiWrapper
     /**
      * Get medias.
      *
+     * @deprecated Please use getMedia() instead.
+     *
+     * @return Media[]
+     */
+    public function getMedias()
+    {
+        return $this->getMedia();
+    }
+
+    /**
+     * Get media.
+     *
      * @return Media[]
      * @VirtualProperty
      * @SerializedName("medias")
      * @Groups({"fullAccount"})
      */
-    public function getMedias()
+    public function getMedia()
     {
         $medias = [];
         if ($this->entity->getMedias()) {
--- a/src/Sulu/Bundle/ContactBundle/Api/Contact.php
+++ b/src/Sulu/Bundle/ContactBundle/Api/Contact.php
@@ -1036,7 +1036,7 @@ class Contact extends ApiWrapper
         if (!is_null($contactAddresses)) {
             /** @var ContactAddressEntity $contactAddress */
             foreach ($contactAddresses as $contactAddress) {
-                if ((bool) $contactAddress->getMain()) {
+                if ((bool)$contactAddress->getMain()) {
                     return $contactAddress->getAddress();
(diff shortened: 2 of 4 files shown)


alexander-schranz:
@kfrohwein `add` and `remove` is used for collection and so I would not replace that with a setMedia that is this definitly missleading and false.

The issue you linked about is about the `media` vs `medias` as the official plural of `media` is `media` the word `medias` is wrong. If we deprecate getMedias we should also create a @trigger_error deprecation.

And please use the Pull Request templates!

src/Sulu/Bundle/ContactBundle/Api/Account.php
```diff
@@ -1066,9 +1066,24 @@ public function getLogo()
      *
      * @param MediaInterface $media
      *
+     * @deprecated Please use setMedia instead.
+     *
      * @return Account
+     *
```
danrot: We don't make an additional new line at the end of the comment.

src/Sulu/Bundle/ContactBundle/Api/Account.php
```diff
@@ -1066,9 +1066,24 @@ public function getLogo()
...
+    }
+
+    /**
+     * Set media.
+     *
+     * @param MediaInterface $media
+     *
+     * @return Account
+     */
+    public function setMedia(MediaInterface $media)
```
danrot: As @alexander-schranz already mentioned, this is not doing what I would expect it to do. I would remove the above code again, since this is just about renaming the `getMedias` method.

src/Sulu/Bundle/ContactBundle/Api/Account.php
```diff
@@ -1088,12 +1103,24 @@ public function removeMedia(MediaInterface $media)
     /**
      * Get medias.
      *
+     * @deprecated Please use getMedia() instead.
```
danrot: We usually mention the full method name and the version since when it is deprecated. So this should be something like this:

```
@deprecated since 2.0, use Account::getMedia instead
```

src/Sulu/Bundle/ContactBundle/Api/Account.php
```diff
@@ -1066,9 +1066,24 @@ public function getLogo()
      *
      * @param MediaInterface $media
      *
+     * @deprecated Please use setMedia instead.
```
danrot: After fixing the below comment, this deprecation is not true anymore.

src/Sulu/Bundle/ContactBundle/Api/Account.php
```diff
@@ -1088,12 +1103,24 @@ public function removeMedia(MediaInterface $media)
     /**
      * Get medias.
      *
+     * @deprecated Please use getMedia() instead.
+     *
+     * @return Media[]
+     */
+    public function getMedias()
+    {
+        return $this->getMedia();
```
danrot: Sadly just the `deprecated` comment above is not enough to also trigger a deprecation in Symfony... In order for that to work you have to trigger an error in here:

```php
@trigger_error('The Account::getMedias method is deprecated, use Account::getMedia instead', E_USER_DEPRECATED);
```

src/Sulu/Bundle/ContactBundle/Api/Contact.php
```diff
@@ -1036,7 +1036,7 @@ public function getMainAddress()
         if (!is_null($contactAddresses)) {
             /** @var ContactAddressEntity $contactAddress */
             foreach ($contactAddresses as $contactAddress) {
-                if ((bool) $contactAddress->getMain()) {
+                if ((bool)$contactAddress->getMain()) {
```
danrot: This change causes the failure in CI, because we have configured it to have a space in that case.

nnatter:
Closing this because of inactivity. Thanks a lot for your effort anyway! 🙂

PR #4963, opened 2020-01-07
Default template based on the template of the parent page

### LINKED ISSUE #4922: Default template based on the template of the parent page
| Q | A
| --- | ---
| Bug? | no
| New Feature? | yes
| Sulu Version | 2.0.2
| Browser Version | any

#### Actual Behavior

When creating a new child-page, the template that's chosen by default almost never corresponds with the one you'll be using. 

#### Expected Behavior

Having the option to define in your parent template what the default template should be for your new child page. 

#### Steps to Reproduce

F.e. you have the artists page on the demo, if you add a page underneath it will most likely only be of type artist. 

#### Possible Solutions

In a template, have an option to specify the child template. That template will then be used when creating a child on that page.

cirykpopeye:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #4922
| Related issues/PRs | #4922
| License | MIT
| Documentation PR | sulu/sulu-docs#479

#### What's in this PR?

It allows the user to define the child template being used on a given parent template.

#### Why?

When we have a list of artist we'd have a Artist overview page and an artist detail page. Underneath an artist overview you would (most of the time) have a artist detail page. Having the option to set that as default is now made available.

#### Example Usage

**In your webspace:**
~~~xml
<default-templates>
        <default-template type="page">default</default-template>
        <default-template type="home">homepage</default-template>
        <default-template type="page" parent-template="artists">artist</default-template>
</default-templates>
~~~

Add a page with the Artists template, now add a child page of that page and the Artist template will be automatically selected.

#### To Do

### DIFF
--- a/src/Sulu/Bundle/PageBundle/Resources/js/views/Form/toolbarActions/TemplateToolbarAction.js
+++ b/src/Sulu/Bundle/PageBundle/Resources/js/views/Form/toolbarActions/TemplateToolbarAction.js
@@ -31,7 +31,7 @@ export default class TemplateToolbarAction extends AbstractFormToolbarAction {
         if (parentPageId && this.webspace && !this.parentPage) {
             ResourceRequester.get('pages', {
                 id: parentPageId,
-                language: this.router.attributes.locale
+                language: this.router.attributes.locale,
             }).then((response) => {
                 this.parentPage = response;
                 const parentTemplate = response.template;
@@ -41,13 +41,12 @@ export default class TemplateToolbarAction extends AbstractFormToolbarAction {
                         break;
                     }
                 }
-            })
+            });
         }
 
         if (!this.resourceFormStore.typesLoading && Object.keys(formTypes).length === 0) {
             throw new Error('The ToolbarAction for types only works with entities actually supporting types!');
         }
-
 
         return {
             type: 'select',
--- a/src/Sulu/Bundle/PageBundle/Resources/js/views/Form/toolbarActions/TemplateToolbarAction.js
+++ b/src/Sulu/Bundle/PageBundle/Resources/js/views/Form/toolbarActions/TemplateToolbarAction.js
@@ -8,7 +8,7 @@ import type {Webspace} from '../../../stores/webspaceStore/types';
 
 export default class TemplateToolbarAction extends AbstractFormToolbarAction {
     @observable webspace: ?Webspace = undefined;
-    parentPage;
+    parentPage: any;
 
     @computed get defaultTemplate(): ?string {
         if (!this.webspace) {
--- a/src/Sulu/Bundle/PageBundle/Resources/js/views/Form/toolbarActions/TemplateToolbarAction.js
+++ b/src/Sulu/Bundle/PageBundle/Resources/js/views/Form/toolbarActions/TemplateToolbarAction.js
@@ -4,6 +4,7 @@ import {AbstractFormToolbarAction} from 'sulu-admin-bundle/views';
 import type {ToolbarItemConfig} from 'sulu-admin-bundle/types';
 import webspaceStore from '../../../stores/webspaceStore';
 import type {Webspace} from '../../../stores/webspaceStore/types';
+import {ResourceRequester} from 'sulu-admin-bundle/services';
 
 export default class TemplateToolbarAction extends AbstractFormToolbarAction {
     @observable webspace: ?Webspace = undefined;
@@ -29,17 +30,17 @@ export default class TemplateToolbarAction extends AbstractFormToolbarAction {
         if (this.router.attributes.parentId && this.webspace) {
             ResourceRequester.get('pages', {
                 id: this.router.attributes.parentId,
-                language: this.router.attributes.locale
-            }).then(response => {
+                language: this.router.attributes.locale,
+            }).then((response) => {
                 const parentTemplate = response.template;
-                for (let key in this.webspace.defaultTemplates) {
-                    let defaultTemplate = this.webspace.defaultTemplates[key];
+                for (const key in this.webspace.defaultTemplates) {
+                    const defaultTemplate = this.webspace.defaultTemplates[key];
                     if (defaultTemplate.parentTemplate === parentTemplate) {
                         this.resourceFormStore.setType(key);
                         break;
                     }
                 }
-            })
+            });
         }
 
         if (!this.resourceFormStore.typesLoading && Object.keys(formTypes).length === 0) {
--- a/src/Sulu/Bundle/PageBundle/Resources/js/views/Form/toolbarActions/TemplateToolbarAction.js
+++ b/src/Sulu/Bundle/PageBundle/Resources/js/views/Form/toolbarActions/TemplateToolbarAction.js
@@ -2,9 +2,9 @@
 import {action, computed, observable} from 'mobx';
 import {AbstractFormToolbarAction} from 'sulu-admin-bundle/views';
 import type {ToolbarItemConfig} from 'sulu-admin-bundle/types';
+import {ResourceRequester} from 'sulu-admin-bundle/services';
 import webspaceStore from '../../../stores/webspaceStore';
 import type {Webspace} from '../../../stores/webspaceStore/types';
-import {ResourceRequester} from 'sulu-admin-bundle/services';
 
 export default class TemplateToolbarAction extends AbstractFormToolbarAction {
     @observable webspace: ?Webspace = undefined;
--- a/src/Sulu/Bundle/PageBundle/Resources/js/views/Form/toolbarActions/TemplateToolbarAction.js
+++ b/src/Sulu/Bundle/PageBundle/Resources/js/views/Form/toolbarActions/TemplateToolbarAction.js
@@ -35,11 +35,11 @@ export default class TemplateToolbarAction extends AbstractFormToolbarAction {
             }).then((response) => {
                 this.parentPage = response;
                 const parentTemplate = response.template;
-                for (const defaultTemplate of this.webspace.defaultTemplates['page']) {
-                    if (defaultTemplate.parentTemplate === parentTemplate) {
-                        this.resourceFormStore.setType(defaultTemplate.template);
-                        break;
-                    }
+                const defaultTemplate = this.webspace.defaultTemplates['page'].find((defaultTemplate) => {
+                    return defaultTemplate.parentTemplate === parentTemplate;
+                });
+                if (defaultTemplate) {
+                    this.resourceFormStore.setType(defaultTemplate.template);
                 }
             });
         }
--- a/src/Sulu/Bundle/PageBundle/Resources/js/views/Form/toolbarActions/TemplateToolbarAction.js
+++ b/src/Sulu/Bundle/PageBundle/Resources/js/views/Form/toolbarActions/TemplateToolbarAction.js
@@ -8,6 +8,7 @@ import type {Webspace} from '../../../stores/webspaceStore/types';
 
 export default class TemplateToolbarAction extends AbstractFormToolbarAction {
     @observable webspace: ?Webspace = undefined;
+    parentPage;
 
     @computed get defaultTemplate(): ?string {
         if (!this.webspace) {
--- a/src/Sulu/Bundle/PageBundle/Resources/js/views/Form/toolbarActions/TemplateToolbarAction.js
+++ b/src/Sulu/Bundle/PageBundle/Resources/js/views/Form/toolbarActions/TemplateToolbarAction.js
@@ -26,6 +26,22 @@ export default class TemplateToolbarAction extends AbstractFormToolbarAction {
             this.resourceFormStore.setType(this.defaultTemplate);
         }
 
+        if (this.router.attributes.parentId && this.webspace) {
+            ResourceRequester.get('pages', {
+                id: this.router.attributes.parentId,
+                language: this.router.attributes.locale
+            }).then(response => {
+                const parentTemplate = response.template;
+                for (let key in this.webspace.defaultTemplates) {
+                    let defaultTemplate = this.webspace.defaultTemplates[key];
+                    if (defaultTemplate.parentTemplate === parentTemplate) {
+                        this.resourceFormStore.setType(key);
+                        break;
+                    }
+                }
+            })
+        }
+
         if (!this.resourceFormStore.typesLoading && Object.keys(formTypes).length === 0) {
             throw new Error('The ToolbarAction for types only works with entities actually supporting types!');
         }
--- a/src/Sulu/Bundle/PageBundle/Resources/js/views/Form/toolbarActions/TemplateToolbarAction.js
+++ b/src/Sulu/Bundle/PageBundle/Resources/js/views/Form/toolbarActions/TemplateToolbarAction.js
@@ -27,25 +27,27 @@ export default class TemplateToolbarAction extends AbstractFormToolbarAction {
             this.resourceFormStore.setType(this.defaultTemplate);
         }
 
-        if (this.router.attributes.parentId && this.webspace) {
+        const parentPageId = this.router.attributes.parentId;
+        if (parentPageId && this.webspace && !this.parentPage) {
             ResourceRequester.get('pages', {
-                id: this.router.attributes.parentId,
-                language: this.router.attributes.locale,
+                id: parentPageId,
+                language: this.router.attributes.locale
             }).then((response) => {
+                this.parentPage = response;
                 const parentTemplate = response.template;
-                for (const key in this.webspace.defaultTemplates) {
-                    const defaultTemplate = this.webspace.defaultTemplates[key];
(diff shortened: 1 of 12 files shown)


cirykpopeye:
@danrot

src/Sulu/Component/Webspace/Loader/XmlFileLoader11.php
```diff
@@ -54,10 +54,14 @@ protected function generateDefaultTemplates(Webspace $webspace)
             /* @var \DOMNode $node */
             $template = $node->nodeValue;
             $type = $node->attributes->getNamedItem('type')->nodeValue;
+            $parentTemplateNode = $node->attributes->getNamedItem('parent-template');
```
danrot: Why did you handle that here in a different and more complicated way as opposed to `XmlFileLoader10`?

src/Sulu/Component/Webspace/Loader/schema/webspace/webspace-1.0.xsd
```diff
@@ -130,6 +130,7 @@
         <xs:simpleContent>
             <xs:extension base="xs:string">
                 <xs:attribute name="type" type="xs:string" use="required"/>
+                <xs:attribute name="parent-template" type="xs:string" use="optional"/>
```
danrot: I think in general we could agree to only add this new feature to `webspace-1.0.xsd` and `XmlFileLoader11.php`.

src/Sulu/Component/Webspace/Webspace.php
```diff
@@ -495,10 +495,14 @@ public function getTemplates()
      *
      * @param string $type
      * @param string $template
+     * @param string $parentTemplate
      */
-    public function addDefaultTemplate($type, $template)
+    public function addDefaultTemplate($type, $template, $parentTemplate)
     {
-        $this->defaultTemplates[$type] = $template;
+        $this->defaultTemplates[$template] = [
```
danrot: This might not work as expected... If you have the same template as default for multiple parent-templates, this way it will only work for one of them. I think we have to keep it the other way, and make this a multidimensional array, something like this:

```php
$this->defaultTemplates[$type][$parentTemplate] = $template;
```

This would only work if `null` is a valid array key, since `$parentTemplate` might be null... If that is not possible we would have to introduce a different structure. In general it would be favourable to introduce a proper object instead of an associative array.

src/Sulu/Bundle/PageBundle/Resources/js/views/Form/toolbarActions/TemplateToolbarAction.js
```diff
@@ -26,6 +27,22 @@ export default class TemplateToolbarAction extends AbstractFormToolbarAction {
             this.resourceFormStore.setType(this.defaultTemplate);
         }
 
+        if (this.router.attributes.parentId && this.webspace) {
+            ResourceRequester.get('pages', {
+                id: this.router.attributes.parentId,
```
danrot: Extract that into an own variable like e.g. `formTypes` above, and reuse it in the surrounding `if` as well.

src/Sulu/Bundle/PageBundle/Resources/js/views/Form/toolbarActions/TemplateToolbarAction.js
```diff
@@ -26,6 +27,22 @@ export default class TemplateToolbarAction extends AbstractFormToolbarAction {
             this.resourceFormStore.setType(this.defaultTemplate);
         }
 
+        if (this.router.attributes.parentId && this.webspace) {
+            ResourceRequester.get('pages', {
```
danrot: I didn't think of that, that is a little bit unfortunate :see_no_evil: But I also don't another way than loading the parent page, as this information is not really available anywhere else :thinking:

However, especially with that line of code I would include a check if an ID is set on the surrounding `if` as well... And we should make sure that this request ist not made multiple times, I am not completely sure how often this function is called during runtime...

src/Sulu/Component/Webspace/Loader/XmlFileLoader11.php
cirykpopeye: XmlFileLoader was the one I was testing with, they are now equal. (11 had more checks for failure).

src/Sulu/Component/Webspace/Loader/schema/webspace/webspace-1.0.xsd
cirykpopeye: So I should remove this from webspace-1.1.xsd?

src/Sulu/Component/Webspace/Webspace.php
cirykpopeye: I created a class for it called DefaultTemplate, with attributes type, template and parentTemplate. This way we can add more features to it later on.

`$this->defaultTemplates[$type] = new DefaultTemplate($type, $template, $parentTemplate);`

src/Sulu/Bundle/PageBundle/Resources/js/views/Form/toolbarActions/TemplateToolbarAction.js
cirykpopeye: Corrected

src/Sulu/Bundle/PageBundle/Resources/js/views/Form/toolbarActions/TemplateToolbarAction.js
cirykpopeye: I've set the parentPage as soon as the request is completed, the next time this function will be called there's a check if parentPage is already set or not.

src/Sulu/Component/Webspace/Loader/schema/webspace/webspace-1.0.xsd
danrot: Sorry, wrote the wrong comment :see_no_evil: We only add this feature to `webspace-1.1.xsd` and `XmlFileLoader11.php`, so it should be removed on the older `webspace-1.0.xsd` and `XmlFileLoader10.php` files.

src/Sulu/Bundle/PageBundle/Resources/js/views/Form/toolbarActions/TemplateToolbarAction.js
danrot: Did you check if that was really necessary first? I.e. if the request was sent multiple times in some circumstances?

src/Sulu/Bundle/PageBundle/Resources/js/views/Form/toolbarActions/TemplateToolbarAction.js
```diff
@@ -26,6 +27,23 @@ export default class TemplateToolbarAction extends AbstractFormToolbarAction {
...
 
+        const parentPageId = this.router.attributes.parentId;
+        if (parentPageId && this.webspace && !this.parentPage) {
+            ResourceRequester.get('pages', {
+                id: parentPageId,
+                language: this.router.attributes.locale,
+            }).then((response) => {
+                this.parentPage = response;
+                const parentTemplate = response.template;
+                for (const defaultTemplate of this.webspace.defaultTemplates['page']) {
```
danrot: Instead of looping over everything I would use the [`find` function of the array](https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Array/find), which results in less levels of nesting.

src/Sulu/Component/Webspace/DefaultTemplate.php
```diff
@@ -0,0 +1,100 @@
...
+     */
+    public function getType(): string
+    {
+        return $this->type;
+    }
+
+    /**
+     * @param string $type
+     */
+    public function setType(string $type): void
```
danrot: We usually don't implement setter, unless they are really necessary. All the properties in this class are not supposed to be changed after object's construction, therefore I would not have any setters at all.

src/Sulu/Component/Webspace/DefaultTemplate.php
```diff
@@ -0,0 +1,100 @@
...
+     * @param string|null $parentTemplate
+     */
+    public function setParentTemplate(?string $parentTemplate): void
+    {
+        $this->parentTemplate = $parentTemplate;
+    }
+
+    public function isValid()
+    {
+        return !empty($this->type) && !empty($this->template);
```
danrot: We have proper types on the constructor and if we do not have any setters there is no way these properties are empty. So why is this method necessary?

src/Sulu/Component/Webspace/Loader/schema/webspace/webspace-1.0.xsd
danrot: Sorry, mentioned the wrong file, we should only do this on `webspace-1.1.xsd`, and not in this one.

src/Sulu/Component/Webspace/Resources/skeleton/WebspaceCollectionClass.php.twig
```diff
@@ -32,62 +32,66 @@ class {{ cache_class }} extends {{ base_class }}
         // new webspace
         $webspace = new Webspace();
         $webspace->setKey('{{ webspace.key }}');
-        $webspace->setName('{{ webspace.name }}');
```
danrot: All these indentations were here by intention, because then the cache file is easier readable. So please revert these indentation changes.

src/Sulu/Component/Webspace/Webspace.php
```diff
@@ -495,10 +495,14 @@ public function getTemplates()
      *
      * @param string $type
      * @param string $template
+     * @param string|null $parentTemplate
      */
-    public function addDefaultTemplate($type, $template)
+    public function addDefaultTemplate($type, $template, $parentTemplate = null)
```
danrot: In e.g. `addLocalization` we are passing the `Localization` object instead of its properties. Would handle it the same way here.

danrot:
Still not totally sure about how to handle the loading of the parent page... Let me think about that a bit.

src/Sulu/Bundle/PageBundle/Resources/js/views/Form/toolbarActions/TemplateToolbarAction.js
cirykpopeye: It was sent multiple times, though only one time with a non-empty webspace...

src/Sulu/Bundle/PageBundle/Resources/js/views/Form/toolbarActions/TemplateToolbarAction.js
```diff
@@ -26,6 +27,23 @@ export default class TemplateToolbarAction extends AbstractFormToolbarAction {
...
 
+        const parentPageId = this.router.attributes.parentId;
+        if (parentPageId && this.webspace && !this.parentPage) {
+            ResourceRequester.get('pages', {
+                id: parentPageId,
+                language: this.router.attributes.locale,
+            }).then((response) => {
+                this.parentPage = response;
+                const parentTemplate = response.template;
+                const defaultTemplate = this.webspace.defaultTemplates['page'].find((defaultTemplate) => {
```
danrot: Use `this.webspace.defaultTemplates.page` in order to avoid different ways of accessing a field.

src/Sulu/Bundle/PageBundle/Resources/js/views/Form/toolbarActions/TemplateToolbarAction.js
```diff
@@ -26,6 +27,23 @@ export default class TemplateToolbarAction extends AbstractFormToolbarAction {
...
+                    return defaultTemplate.parentTemplate === parentTemplate;
+                });
+                if (defaultTemplate) {
```
danrot: Would add an empty line before this one.

src/Sulu/Component/Webspace/DefaultTemplate.php
```diff
@@ -0,0 +1,71 @@
...
+     * @var string|null
+     */
+    private $parentTemplate;
+
+    /**
+     * DefaultTemplate constructor.
+     *
+     * @param string $type
+     * @param string $template
+     * @param string|null $parentTemplate
```
danrot: We don't add the `@param` tags if they don't add any value (what you've written here is also already written in the type hints of the constructor). Sorry, could have seen that before :see_no_evil:

cirykpopeye: Ok, removed :)

src/Sulu/Bundle/PageBundle/Resources/js/views/Form/toolbarActions/TemplateToolbarAction.js
cirykpopeye: Ok

src/Sulu/Bundle/PageBundle/Resources/js/views/Form/toolbarActions/TemplateToolbarAction.js
cirykpopeye: Correct ;-)

src/Sulu/Component/Webspace/Webspace.php
cirykpopeye: The object is passed instead of its properties

src/Sulu/Component/Webspace/Resources/skeleton/WebspaceCollectionClass.php.twig
cirykpopeye: Ok, indentations are back, though I now have 2 for loops instead of just 1. Don't know if the indentations are correct?

src/Sulu/Component/Webspace/DefaultTemplate.php
cirykpopeye: Removed, it is no longer necessary.

src/Sulu/Component/Webspace/DefaultTemplate.php
cirykpopeye: All setters are removed

src/Sulu/Bundle/PageBundle/Resources/js/views/Form/toolbarActions/TemplateToolbarAction.js
cirykpopeye: Replace with the find function

src/Sulu/Component/Webspace/Loader/schema/webspace/webspace-1.0.xsd
cirykpopeye: Ok, parent template functionality removed from 1.0, added in 1.1

src/Sulu/Bundle/PageBundle/Resources/js/views/Form/toolbarActions/TemplateToolbarAction.js
danrot: Okay, we've put some thoughts into this, and this is probably not the way to go. Loading this data in here, means that we would have to send another request in another `ToolbarAction`, if we also need data from the parent there.

In order to avoid this request, we could do that directly in the [`handleItemAdd` function of our `PageList`](https://github.com/sulu/sulu/blob/f39900f/src/Sulu/Bundle/PageBundle/Resources/js/views/PageList/PageList.js#L152). We already have there everything we need: The `Webspace` object and the `id` of the parent page. Then we could add this to the object being passed to the `navigate` call as `defaultTemplate`, and just use `this.router.attributes.defaultTemplate` in the `TemplateToolbarAction`.

WDYT?

src/Sulu/Component/Webspace/DefaultTemplate.php
```diff
@@ -0,0 +1,64 @@
...
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Component\Webspace;
+
+/**
+ * Container for a default template definition.
```
danrot: That comment does not add any value, therefore I would remove it.

jordygroote:
@danrot We would love this feature, is there any chance this PR will be finished you think?

cirykpopeye:
> @danrot We would love this feature, is there any chance this PR will be finished you think?

I will free up some time for this asap to continue my work on this issue, but many things pushed it forward

danrot:
Just wanted to add that I have currently other priorities... But it would of course be awesome if @cirykpopeye could finish this :slightly_smiling_face:

nnatter:
Just remembered that the `webspaceStore` was refactored to allow for accessing the data synchronously in #5119. Hopefully that simplifies this feature a bit 🙂

danrot:
It would be easier, if he would not have already implemented it :see_no_evil: But yeah, we should probably use the newer approach here, since the other one is already deprecated.

cirykpopeye:
I will close this PR and create a new one.

PR #4971, opened 2020-01-10
Display add button in category list only if user has sufficient permissions

nnatter:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This PR hides the add button inside the tree-table list-adapter of the category list if the current user does not have the respective permission.

#### Why?

Because Sulu usually hides control elements for which the current user does not have sufficient permissions. Eg. toolbar actions are only displayed if the current user has the permissions to execute the action.

### DIFF
--- a/src/Sulu/Bundle/CategoryBundle/Admin/CategoryAdmin.php
+++ b/src/Sulu/Bundle/CategoryBundle/Admin/CategoryAdmin.php
@@ -105,8 +105,7 @@ class CategoryAdmin extends Admin
                 ->enableSearching()
                 ->addToolbarActions($listToolbarActions);
 
-            // set add view of list-view only if user has add permission
-            // this will hide the add button of the tree_table adapter if the user has no add permission
+            // hide add button of the tree_table adapter by not setting an add view if the user has no add permission
             if ($this->securityChecker->hasPermission(self::SECURITY_CONTEXT, PermissionTypes::ADD)) {
                 $listViewBuilder->setAddView(static::ADD_FORM_VIEW);
             }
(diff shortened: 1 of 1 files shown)


src/Sulu/Bundle/CategoryBundle/Admin/CategoryAdmin.php
```diff
@@ -93,20 +93,25 @@ public function configureViews(ViewCollection $viewCollection): void
...
+                ->setTitle('sulu_category.categories')
+                ->addListAdapters(['tree_table'])
+                ->addLocales($locales)
+                ->setDefaultLocale($locales[0])
+                ->setEditView(static::EDIT_FORM_VIEW)
+                ->enableSearching()
+                ->addToolbarActions($listToolbarActions);
+
+            // set add view of list-view only if user has add permission
+            // this will hide the add button of the tree_table adapter if the user has no add permission
```
danrot: Isn't that clear from the code below? I think I would have not added this comment, as it doesn't add any value to me, and duplicates stuff :see_no_evil:

danrot: I would at least remove the first sentence, if not both. But the second one is ok for me, if you insist on it :see_no_evil: I just got bitten by wrong comments too many times in the past, therefore I am very careful about adding them if they don't add value.

nnatter: you are right, the first line is quite obvious. i added the comment because we dont do this in any other `Admin` class. the reason for this is that we dont use the `tree_table` adapter anywhere else and therefore it is not needed..

PR #4972, opened 2020-01-10
Display icon in edit button of ColumnListAdapter based on permissions

nnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This PR adjust the ColumnListAdapter of the List container to display different icons in the edit button based on the item-specific permission of the current user.

#### Why?

Because it is kind of misleading to display a `pen` icon if the current user is not allowed to edit anything.


src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/adapters/ColumnListAdapter.js
```diff
@@ -113,23 +113,23 @@ class ColumnListAdapter extends AbstractAdapter {
         const {
             _permissions: {
                 view: viewPermission = true,
+                edit: editPermission = true,
             } = {},
         } = item;
 
         if (onItemClick) {
+            let itemIcon = 'su-eye';
```
danrot: Personally I still try to avoid `let` :see_no_evil: I think I would even have preferred a nested ternary. I think they read quite nice, if you nest them correctly (and I've just checked, eslint doesn't even allow to nest them in a different way):

```javascript
const itemIcon = isGhost
    ? 'su-plus-circle'
    : editPermission
        ? 'su-pen'
        : 'su-eye';
```

They are also [gaining popularity in JS](https://medium.com/javascript-scene/nested-ternaries-are-great-361bddd0f340). Just a small input, I am also fine if we leave it this way, but I would slightly prefer the other way.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/tests/adapters/ColumnListAdapter.test.js
```diff
@@ -111,6 +111,61 @@ test('Render different kind of data with edit button', () => {
...
+    const columnListAdapter = render(
+        <ColumnListAdapter
+            {...listAdapterDefaultProps}
+            activeItems={[]}
+            data={data}
+            onItemClick={jest.fn()}
+        />
+    );
+
+    expect(columnListAdapter).toMatchSnapshot();
```
danrot: I'd like to reduce the number of snapshot tests, since we've got way too many :see_no_evil: If you change something in a dumb component, that causes too many tests to fail, and people (including me) are too lazy to look at the snapshots in detail... I would suggest including these examples in the previous tests, which also includes a snapshot, or use clearer assertions here (then we also know what exactly the test is looking at).

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/adapters/ColumnListAdapter.js
nnatter: sounds good to me 🙂

PR #4973, opened 2020-01-10
Fix Update of List Selection Value from outside

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

Fix Update of List Selection Value from outside.

#### Why?

If you manipulate the selection value from outside e.g.: setting it directly on the resourceStore or in my case over a custom form the selection is not updated.

Also the selection is not updated if the response from the server would add because of businessLogic new selections.

#### Example Usage

Easiest way to test it is using the following inside a page template:

```xml
        <section name="hello-1">
            <properties>
                <property name="categories" type="category_selection"/>
            </properties>
        </section>

        <section name="hello-2">
            <properties>
                <property name="categories" type="category_selection"/>
            </properties>
        </section>
```

#### To Do

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/Selection.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/Selection.js
@@ -78,6 +78,24 @@ export default class Selection extends React.Component<Props> {
         }
     }
 
+    componentDidUpdate(prevProps: Props) {
+        this.updateListStoreSelection(prevProps);
+    }
+
+    updateListStoreSelection(prevProps: Props) {
+        if (prevProps.value === this.props.value) {
+            return;
+        }
+
+        const listStore = this.listStore;
+
+        if (!listStore) {
+            return;
+        }
+
+        listStore.updateSelectionIds(toJS(this.props.value) || []);
+    }
+
     componentWillUnmount() {
         if (this.changeListDisposer) {
             this.changeListDisposer();
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/stores/ListStore.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/stores/ListStore.js
@@ -526,6 +526,66 @@ export default class ListStore {
         });
     };
 
+    updateSelectionIds(newSelectionIds: Array<string | number>) {
+        const oldSelectionIds = this.selectionIds;
+
+        if (oldSelectionIds === newSelectionIds) {
+            return;
+        }
+
+        const removedValues = oldSelectionIds.filter((x) => !newSelectionIds.includes(x));
+        const addedValues = newSelectionIds.filter((x) => !oldSelectionIds.includes(x));
+        const notLoadedIds = [];
+
+        removedValues.forEach((oldId: string | number) => {
+            this.deselectById(oldId);
+        });
+
+        addedValues.forEach((newId: string | number) => {
+            const row = this.findById(newId);
+
+            if (!row) {
+                notLoadedIds.push(newId);
+
+                return;
+            }
+
+            this.select(row);
+        });
+
+        if (!notLoadedIds.length) {
+            return;
+        }
+
+        const observableOptions = {};
+
+        for (const key in this.observableOptions) {
+            observableOptions[key] = this.observableOptions[key].get();
+        }
+
+        this.setDataLoading(true);
+        const options = {...observableOptions, ...this.options};
+
+        options.selectedIds = newSelectionIds.join(',');
+
+        this.loadingStrategy.load(
+            this.resourceKey,
+            options
+        ).then(action(() => {
+            this.setDataLoading(false);
+
+            newSelectionIds
+                .map((selectionId) => this.findById(selectionId))
+                .forEach((selectionRow) => {
+                    if (!selectionRow) {
+                        return;
+                    }
+
+                    this.select(selectionRow);
+                });
+        }));
+    }
+
     @action setDataLoading(dataLoading: boolean) {
         this.dataLoading = dataLoading;
     }


src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/Selection.js
```diff
@@ -78,6 +79,42 @@ export default class Selection extends React.Component<Props> {
...
+        }
+
+        const arrayDiff = new ArrayDiff(selectionIds, newSelectionIds);
+
+        arrayDiff.removedValues.forEach((oldId: string | number) => {
+            listStore.deselectById(oldId);
+        });
+
+        arrayDiff.addedValues.forEach((newId: string | number) => {
+            listStore.select(listStore.findById(newId));
```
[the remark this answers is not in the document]
danrot: Resp. maybe it doesn't make sense to extract the entire part, but I think we need something like this...

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/Selection.js
```diff
@@ -78,6 +79,42 @@ export default class Selection extends React.Component<Props> {
         }
     }
 
+    componentDidUpdate(prevProps: Props) {
```
danrot: This function definitely needs a test, especially because this code probably has no effect in the core system :see_no_evil:

danrot (changes requested):

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/Selection.js
```diff
@@ -78,6 +78,24 @@ export default class Selection extends React.Component<Props> {
...
+        this.updateListStoreSelection(prevProps);
+    }
+
+    updateListStoreSelection(prevProps: Props) {
+        if (prevProps.value === this.props.value) {
```
danrot: Since you are using `this.props.value` below again, I would say we should destructure it.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/Selection.js
danrot: This still holds true :slightly_smiling_face:

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/stores/ListStore.js
```diff
@@ -526,6 +526,66 @@ export default class ListStore {
...
+        for (const key in this.observableOptions) {
+            observableOptions[key] = this.observableOptions[key].get();
+        }
+
+        this.setDataLoading(true);
+        const options = {...observableOptions, ...this.options};
+
+        options.selectedIds = newSelectionIds.join(',');
+
+        this.loadingStrategy.load(
```
danrot: That is also showing the newly selected items in the list immediately, right? Is that the desired behavior in every case? I think that would e.g. also add all selected items in a table view in an unexpected manner. Have you tried that?

danrot (changes requested):
This PR is still missing tests.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/stores/ListStore.js
alexander-schranz: @danrot thats the case we want, think when the developer doesn't want that he should not reuse the ResourceStore instead clone the ResourceStore or what do you think?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/Selection.js
```diff
@@ -78,6 +78,24 @@ export default class Selection extends React.Component<Props> {
...
+            return;
+        }
+
+        const listStore = this.listStore;
```
danrot: Any reason not to use destructuring instead?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/stores/ListStore.js
danrot: We are not even in the `ResourceStore` here, so I guess you wanted to say `ListStore`.

But I think I was mistaken anyway... The `loadingStrategy` really only loads the data, and does nothing else... So that code should be fine :+1:

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/stores/ListStore.js
```diff
@@ -526,6 +526,66 @@ export default class ListStore {
...
+            this.resourceKey,
+            options
```
danrot: I think instead of all the above code you could simply use `{...this.observableOptions, ...this.options}` here instead, as [it is already done in the `ListStore`](https://github.com/sulu/sulu/blob/0d416ba/src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/stores/ListStore.js#L545).

And I guess you also want to include the `fields` here, so that the correct columns are loaded, so the option would be something like the following:

```javascript
{...this.observableOptions, ...this.options, fields: this.field}
```

danrot (changes requested):
This probably also needs a rebase (changed quite some stuff here, just want to be sure) and definitely needs a test :slightly_smiling_face:

alexander-schranz:
@nnatter I know that this was some issue in a projects DirectoryBundle we did build some time ago. Where the `Overlay` did access the `FormResourceStore` and updated an value but it was not updated in the browser, do you know if we did fix this in another way?

niklasnatter:
No think this is still a problem, I can have a look at it when I find the time. But looks like nobody else noticed this, so I think this does not have high priority at the moment 🙂

Issue #3943, opened 2018-04-30
Enable self-validation of field types


danrot:
| Q | A
| --- | ---
| Bug? | no
| New Feature? | yes
| Sulu Version | 24fbb9e69bf2e644274534ffb17879a356f7b303
| Browser Version | ---

#### Actual Behavior

Some FieldTypes already validate themself, and show a red border in case their content is not valid. This is required because we want these FieldTypes to be usable outside of the `Form` context as well.

However, when a FieldType validates itself, it is currently not possible prevent the form from being saved.

#### Expected Behavior

When a FieldType validates itself and says that it is not valid, it should not be able to save the form.

#### Steps to Reproduce

1. Create a form, which contains e.g. the `email` field type (or use the permissions tab in the contact form which will be available after the merge of https://github.com/sulu/sulu/pull/3898)
2. Write some invalid value in the `email` field (e.g. omit the `@` in the email adress).
3. Click the save button
4. The form will be saved although there is an invalid field

#### Possible Solutions

I could imagine that the `onFinish` callback from the `FieldTypeProps` type should get an argument, which returns an instance of an error. Then the upper component could take care of stopping the save from happening.

nnatter:
I think it would be useful to think about #5426 too when working on this.

PR #4977, opened 2020-01-13
Fixes tag_selection for tags containing commas

### LINKED ISSUE #4976: TagSelection not work with Tag containing a comma
| Q | A
| --- | ---
| Bug? | yes
| New Feature? | no
| Sulu Version | 2.0.2
| Browser Version | -

#### Actual Behavior

TagSelection not work with Tag containing a comma.

#### Expected Behavior

Tags with comma should also work.

#### Steps to Reproduce

1. Log into Sulu
2. Go to Settings
3. Create a Tag with a a comma e.g. `Tag, Test`
4. Go to a Page
5. Add the new created Tag to the Tag List
6. Refresh (Tag is not longer shown)

luca-rath:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #4976
| License | MIT

#### What's in this PR?

This PR adds a fix for the `tag_selection` to display tags containing a comma.

#### Why?

Selected tags containing a comma don't show up in the `tag_selection`.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Input/types.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Input/types.js
@@ -26,6 +26,8 @@ export type InputProps<T: ?string | ?number> = {|
     onClearClick?: () => void,
     onIconClick?: () => void,
     onKeyPress?: (key: ?string, event: SyntheticKeyboardEvent<HTMLInputElement>) => void,
+    pattern?: string,
+    patternDescription?: string,
     placeholder?: string,
     segmentDelimiter?: string,
     skin?: 'default' | 'dark',
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/stores/MultiSelectionStore/MultiSelectionStore.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/stores/MultiSelectionStore/MultiSelectionStore.js
@@ -69,7 +69,6 @@ export default class MultiSelectionStore<T = string | number, U: {id: T} = Objec
             ...this.requestParameters,
             locale: this.locale ? this.locale.get() : undefined,
             [this.idFilterParameter]: itemIds.join(','),
-            [this.idFilterParameter + '-array']: JSON.stringify(itemIds),
             limit: undefined,
             page: 1,
         }).then(action((data) => {
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/stores/MultiSelectionStore/MultiSelectionStore.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/stores/MultiSelectionStore/MultiSelectionStore.js
@@ -69,6 +69,7 @@ export default class MultiSelectionStore<T = string | number, U: {id: T} = Objec
             ...this.requestParameters,
             locale: this.locale ? this.locale.get() : undefined,
             [this.idFilterParameter]: itemIds.join(','),
+            [this.idFilterParameter + '-array']: JSON.stringify(itemIds),
             limit: undefined,
             page: 1,
         }).then(action((data) => {
--- a/src/Sulu/Bundle/TagBundle/Controller/TagController.php
+++ b/src/Sulu/Bundle/TagBundle/Controller/TagController.php
@@ -17,7 +17,6 @@ use FOS\RestBundle\View\ViewHandlerInterface;
 use HandcraftedInTheAlps\RestRoutingBundle\Routing\ClassResourceInterface;
 use Sulu\Bundle\TagBundle\Admin\TagAdmin;
 use Sulu\Bundle\TagBundle\Controller\Exception\ConstraintViolationException;
-use Sulu\Bundle\TagBundle\Exception\InvalidTagNameException;
 use Sulu\Bundle\TagBundle\Tag\Exception\TagAlreadyExistsException;
 use Sulu\Bundle\TagBundle\Tag\Exception\TagNotFoundException;
 use Sulu\Bundle\TagBundle\Tag\TagManagerInterface;
--- a/src/Sulu/Bundle/TagBundle/Controller/TagController.php
+++ b/src/Sulu/Bundle/TagBundle/Controller/TagController.php
@@ -136,8 +136,7 @@ class TagController extends AbstractRestController implements ClassResourceInter
             $fieldDescriptors = $this->fieldDescriptorFactory->getFieldDescriptors('tags');
             $listBuilder = $this->listBuilderFactory->create($this->tagClass);
 
-            $names = \json_decode($request->get('names-array', '[]'));
-
+            $names = \array_filter(\explode(',', $request->get('names', '')));
             if (\count($names) > 0) {
                 $listBuilder->in($fieldDescriptors['name'], $names);
             }
--- a/src/Sulu/Bundle/TagBundle/Controller/TagController.php
+++ b/src/Sulu/Bundle/TagBundle/Controller/TagController.php
@@ -136,7 +136,8 @@ class TagController extends AbstractRestController implements ClassResourceInter
             $fieldDescriptors = $this->fieldDescriptorFactory->getFieldDescriptors('tags');
             $listBuilder = $this->listBuilderFactory->create($this->tagClass);
 
-            $names = \array_filter(\explode(',', $request->get('names', '')));
+            $names = \json_decode($request->get('names-array', '[]'));
+
             if (\count($names) > 0) {
                 $listBuilder->in($fieldDescriptors['name'], $names);
             }
--- a/src/Sulu/Bundle/TagBundle/Controller/TagController.php
+++ b/src/Sulu/Bundle/TagBundle/Controller/TagController.php
@@ -17,6 +17,7 @@ use FOS\RestBundle\View\ViewHandlerInterface;
 use HandcraftedInTheAlps\RestRoutingBundle\Routing\ClassResourceInterface;
 use Sulu\Bundle\TagBundle\Admin\TagAdmin;
 use Sulu\Bundle\TagBundle\Controller\Exception\ConstraintViolationException;
+use Sulu\Bundle\TagBundle\Exception\InvalidTagNameException;
 use Sulu\Bundle\TagBundle\Tag\Exception\TagAlreadyExistsException;
 use Sulu\Bundle\TagBundle\Tag\Exception\TagNotFoundException;
 use Sulu\Bundle\TagBundle\Tag\TagManagerInterface;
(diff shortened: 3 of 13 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/stores/MultiSelectionStore/MultiSelectionStore.js
```diff
@@ -51,7 +51,13 @@ export default class MultiSelectionStore<T = string | number, U: {id: T} = Objec
         this.setLoading(true);
         return ResourceRequester.getList(this.resourceKey, {
             locale: this.locale ? this.locale.get() : undefined,
-            [this.idFilterParameter]: itemIds.join(','),
+            [this.idFilterParameter]: itemIds.map((itemId) => {
+                if (typeof itemId !== 'string') {
+                    return itemId;
+                }
+
+                return itemId.replace(/[/,]/g, '/$&');
```
danrot: Shouldn't the comma be escaped by a backslash?

src/Sulu/Bundle/TagBundle/Controller/TagController.php
```diff
@@ -138,7 +138,8 @@ public function cgetAction(Request $request)
             $fieldDescriptors = $this->fieldDescriptorFactory->getFieldDescriptors('tags');
             $listBuilder = $this->listBuilderFactory->create($this->tagClass);
 
-            $names = array_filter(explode(',', $request->get('names', '')));
+            $names = $this::explodeTags($request->get('names', ''));
```
danrot: Why not `$this->explodeTags`?

src/Sulu/Bundle/TagBundle/Controller/TagController.php
```diff
@@ -281,7 +282,7 @@ public function deleteAction($id)
     public function postMergeAction(Request $request)
     {
         try {
-            $srcTagIds = explode(',', $request->get('src'));
+            $srcTagIds = $this::explodeTags($request->get('src', ''));
```
danrot: Why not `$this->explodeTags`?

src/Sulu/Bundle/AdminBundle/Resources/js/stores/MultiSelectionStore/MultiSelectionStore.js
danrot: And what exactly are you replacing it with?

src/Sulu/Bundle/TagBundle/Controller/TagController.php
```diff
@@ -356,4 +357,40 @@ protected function getData(Request $request)
     {
         return $request->request->all();
     }
+
+    /**
+     * Returns array of tags from escaped input string.
+     */
+    private function explodeTags(string $input): array
```
danrot: I have a very hard time understanding this function :confused: Isn't there an easier solution to this problem? Can't we escape the comma in a different way?

src/Sulu/Bundle/TagBundle/Controller/TagController.php
```diff
@@ -356,4 +357,40 @@ protected function getData(Request $request)
...
+    {
+        $tags = [];
+        $append = false;
+        foreach (explode(',', $input) as $tag) {
```
danrot: I think we should be able to escape the comma in a different way, since `explode` also excepts more than one character. Would probably make the code aftewards a lot easier.

danrot (changes requested):
More general question: Do we really want that? I think I would prefer to not allow commas in tags, since IMO they are confusing. And you've just fixed parts of this issue, because the `AutoCompleteList` for tags will still create a new tag if you enter a comma.

danrot:
Apart from my concerns if we really want to do this, this should probably go into the `release/2.0` branch, independent of how we handle this. Seems like you can create a tag with a comma but cannot properly use it aftewards, which I would consider a bug that needs to be fixed (either by disallowing this specific character or try to make it work somehow).

alexander-schranz:
I'm also not totally sure if we should implement it as it seems to add high complexibility to the code. Maybe we should go with throwing an exception in `setName` instead when somebody tries to create a Tag which contains a `,` so its not possible to create such tags from every case (CSV imports like we had or over the Sulu UI). @danrot what do you think about that?

danrot:
That sounds about right. I think it would also be nice if we could add a pattern to an input field, so that our validation fails if a comma is included in the text. Then the user would immediately know what is wrong.

src/Sulu/Bundle/TagBundle/Controller/TagController.php
luca-rath: The problem is, that the escape character could be escaped as well, otherwise i could explode by `/(?<!\\),/`, but since i cannot make a negative lookbehind, which checks that the comma is not preceded by an odd amount of backslashes, it's not possible with regex.

src/Sulu/Bundle/TagBundle/Controller/TagController.php
luca-rath: https://github.com/sulu/sulu/pull/4977#discussion_r366756243

src/Sulu/Bundle/TagBundle/Controller/TagController.php
luca-rath: This was a static function before ... i was probably a little bit sloppy here

src/Sulu/Bundle/AdminBundle/Resources/js/stores/MultiSelectionStore/MultiSelectionStore.js
luca-rath: It will be replaced by itself preceded by a slash

luca-rath: I agree, that the escape character should be a backslash normally, but it's a lot more difficult to handle, because if the request contains two backslashes, php treats it as one. If the request contains two slashes, php also treats it as two slashes. This is a lot more predictable.

luca-rath:
I've got an idea how to solve this issue. The `MultiSelectionStore` should send the `ids` in json-encoded form using a new key (like `names-array` for tags) to the `cgetAction` of it's corresponding controller ... Then we could adjust all our controllers to use the new key. Since we wouldn't remove the old key, this wouldn't even be a bc break and all custom controllers will continue to work. A benefit of this method compared to @alexander-schranz suggestion https://github.com/sulu/sulu/pull/4977#issuecomment-574354683 is, that existing tags containing a `,` will be shown in the `tag_selection` field type in the admin again.

One could say, that this fix is irrelevant for tags, because it doesn't really makes sense for tags to contain a comma, but this issue affects all custom entities with a string identifier of the admin users choice.

/cc @sulu/core-team

alexander-schranz:
I personally would not support `,` and add just a `\InvalidArgumentException` in `setName` of the Tags so this can not longer happen. If we really want to support tags with `,` we should go the default http way of sending arrays and provide a array query string e.g.: `?name[]=Tag1&name[]=Tag2`. I think having json in the query string is something we should not do.

The api would just have a way to  handle both `is_string` do an explode or `is_array` it can use it directly. No need to do any json decode or any other decode on the admin side which I would really avoid.

danrot:
I also would not allow commas in tags... I get what you are saying about custom entities, but I think this is a non-issue. The tags are already an edge case, since in all other cases a UUID or numeric ID are sent, which will never have a comma. Therefore I don't see why we should go down that road.

And if we really want to support something like this I would also prefer @alexander-schranz suggestion over creating a json string within the URL.

luca-rath:
@danrot does something like this look like a good solution to you? I guess we somehow have to extract the pattern validation to the `FormStore` to disable saving, if the pattern fails, but unfortunately I have no idea how 🙈 Can you give me a hint on how to achieve that?

nnatter:
@luca-rath I think your problem is related/similar to #3943. Would be nice to finally implement a solution for this IMO, but not sure if there are resources for that at the moment 🙂

src/Sulu/Bundle/AdminBundle/Resources/js/components/Input/types.js
```diff
@@ -26,6 +26,8 @@ export type InputProps<T: ?string | ?number> = {|
     onClearClick?: () => void,
     onIconClick?: () => void,
     onKeyPress?: (key: ?string, event: SyntheticKeyboardEvent<HTMLInputElement>) => void,
+    pattern?: string,
```
nnatter, line 29: like the idea 👍  would be nice to include other standard `<input>` attributes such as `maxlength` in the future too

luca-rath: The problem with those attributes is, that they will only be validated when the form is submitted. And because the form is never being submitted in sulu, the browser doesn't validate those fields 😬

nnatter: yes, i guess we will need to implement the handling of the attributes by ourselves. but it would be nice to provide the functionality to our users 🙂

luca-rath: 👍

danrot: First of all, isn't the `submit` event of the form emitted? I thought I have implemented it that way, exactly for reasons like this (also other code might listen to that event...).

Second, I am not sure if I would implement it like this... AFAIK browsers also show their own error based on these properties, which would then probablyl not match our design. I think the best way forward would be to finally allowing field types resp. content types to have an influence on the generated JSON schema. If the JSON schema is generated as required, the frontend would not even have to be touched.

This approach would also fix both issues #5426 and #3943. So we could also corretly check the mandatory field for e.g. the media selection field type. If that's OK for everybody I would be happy to discuss that in detail.

luca-rath: Can be, but the native input validation for the `pattern` was never triggered.

If everybody is fine with your suggestion, I would just pause working on this pr until a solution for self-validation is finished.

nnatter: I think I would not event pass the `pattern` to the `<input>`, but implement the handling and the respective message in our `Input` component. Using the already present JSON schema sounds good to me too, but I feel like it will be harder to output helpful error messages that way. Also I dont have an idea how to implement that in a pleasent way in terms of DX, but I have not thought about it in detail. Happy to talk about this too, it would be nice to get some progress in this area soon!

luca-rath:
Closing in favour of #5672

PR #4983, opened 2020-01-16
Fix conflict message when referenced snippet is being deleted

danrot:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR fixes the message when a referenced snippet is tried to be deleted.

#### Why?

Because now a JS error was thrown, which caused all other clicks to have no effect.

#### To Do

- [ ] Check `SnippetController::putAction` with referential integrity message

### DIFF
--- a/src/Sulu/Bundle/SnippetBundle/Controller/SnippetController.php
+++ b/src/Sulu/Bundle/SnippetBundle/Controller/SnippetController.php
@@ -269,7 +269,7 @@ class SnippetController implements SecuredControllerInterface, ClassResourceInte
         $references = $this->snippetRepository->getReferences($id);
 
         if (count($references) > 0) {
-            $force = $request->headers->get('SuluForceRemove', false);
+            $force = $request->query->get('force', false);
             if ($force) {
                 $this->contentMapper->delete($id, $webspaceKey, true);
             } else {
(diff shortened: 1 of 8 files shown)


src/Sulu/Bundle/SnippetBundle/Controller/SnippetController.php
```diff
@@ -117,7 +123,8 @@ public function __construct(
         FormFactory $formFactory,
         RequestHashChecker $requestHashChecker,
         ListRestHelper $listRestHelper,
-        MetadataFactoryInterface $metadataFactory
+        MetadataFactoryInterface $metadataFactory,
+        TranslatorInterface $translator
```
alexander-schranz: Should we document BC Breaks in constructors/services definitions @sulu/core-team?

alexander-schranz:
@danrot Would call the confirmation button something like `Delete` instead of Ok.

Bildschirmfoto 2020-01-20 um 10 31 06

@chirimoya what do you think?

src/Sulu/Bundle/SnippetBundle/Controller/SnippetController.php
danrot: That's the reason I would not do it. It is very unlikely that this is done, and if we start to document BC breaks on that level, then the UPGRADE.md file gets very bloated.

danrot:
@alexander-schranz The button was already there before, haven't touched it :see_no_evil: And if we rename it here, it should probably also be renamed in the previous overlay, where it currently also says "Ok".

PR #4984, opened 2020-01-16
Rename display_condition option of toolbar actions of Form view to visible_condition

nnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This PR renames the `display_condition` option of various toolbar actions of the Form view to `visible_condition`. The old `display_condition` was deprecated and will log a deprecation warning if it is used.

#### Why?

Because we already support a `visibleCondition` attribute in our Form XML files that allows to hide specific field-types and i would like to provide a consistent naming. 

#### Example Usage

~~~php
new ToolbarAction(
    'sulu_admin.copy_locale',
    [
         'visible_condition' => '(!_permissions || _permissions.edit)',
    ]
),
~~~

#### BC Breaks/Deprecations

The `display_condition` option of the following toolbar actions was deprecated:
* CopyLocaleToolbarAction
* DeleteDraftToolbarAction
* DeleteToolbarAction
* SetUnpublishedToolbarAction

The `publish_display_condition` and `save_display_condition` of the following toolbar action was deprecated:
* SaveWithPublishingToolbarAction

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/views/Form/toolbarActions/CopyLocaleToolbarAction.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/views/Form/toolbarActions/CopyLocaleToolbarAction.js
@@ -31,6 +31,7 @@ export default class CopyLocaleToolbarAction extends AbstractFormToolbarAction {
         } = options;
 
         if (displayCondition) {
+            // @deprecated
             log.warn(
                 'The "display_condition" option is deprecated since version 2.0 and will be removed. ' +
                 'Use the "visible_condition" option instead.'
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/views/Form/toolbarActions/CopyLocaleToolbarAction.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/views/Form/toolbarActions/CopyLocaleToolbarAction.js
@@ -2,17 +2,47 @@
 import React from 'react';
 import {action, observable} from 'mobx';
 import jexl from 'jexl';
+import log from 'loglevel';
 import Checkbox from '../../../components/Checkbox';
 import Dialog from '../../../components/Dialog';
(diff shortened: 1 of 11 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/views/Form/toolbarActions/CopyLocaleToolbarAction.js
```diff
@@ -2,18 +2,48 @@
...
+        locales: ?Array<string>,
+        options: {[key: string]: mixed}
+    ) {
+        const {
+            display_condition: displayCondition,
+            visible_condition: visibleCondition,
+        } = options;
+
+        if (displayCondition) {
+            log.warn(
```
danrot: We also have to think about how to find these deprecations later when we want to remove them... Searching for `log.warn` might be a little bit a hassle... Maybe we should use the `deprecated` annotation of JS Doc? Should be easy to search for afterwards :slightly_smiling_face:

This of course applies to all deprecations you have done in here.

nnatter: conducted a short google search and did not find anything better than the `@deprecated` JSDoc. but we are not deprecating a function here.. 

where would you add the `@deprecated` tag? as a comment in the conde?

danrot: That's the closest thing I can think of. If you add a message we know what to do, when we search for the `@deprecated` tag in the entire project before the 3.0 release.

nnatter: done 🙂

PR #4985, opened 2020-01-17
Update issue templates

danrot:
[Github now allows to have multiple issue templates for one repository.](https://help.github.com/en/github/building-a-strong-community/about-issue-and-pull-request-templates#issue-templates)

I think we should make use of this feature, in order to allow people create proper feature requests as well. What I especially like about this, is that the correct label is automatically added.

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -1,0 +1,15 @@
+---
+name: Feature request
+about: Suggest an idea for this project
+title: ''
+labels: Feature
+assignees: ''
+---
+
+#### Problem description
+
+A clear description of what the problem is.
+
+#### Proposed solution
+
+A clear description of what you want to be implemented.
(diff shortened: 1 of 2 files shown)


danrot:
@sulu/core-team WDYT? Would you suggest any changes? I kept the already existing issue template as bug report, and the feature one is an adapted version of what GitHub is suggesting.

danrot:
And I am very sorry that I created the PR on the wrong repository, but I used GitHub's internal tooling, and then it was created here :see_no_evil:

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,15 @@
+---
+name: Feature request
+about: Suggest an idea for this project
+title: ''
+labels: Feature
+assignees: ''
+---
+
+#### Problem description
```
danrot: I have used a level 4 heading here, because the same one is used in the current issue template. However, this might lead to a confusing structure, if the feature needs more explanation and more sub headings. I think I would go here for a level 1 heading, and I would do the same for the bug report. WDYT?

The only reason speaking against this is that it might take more space, but I think that is not a big issue. Putting some headlines here for a comparison:

# Heading 1
## Heading 2
### Heading 3
#### Heading 4
##### Heading 5
###### Heading 6

Personally I prefer clear structure over some visual aspects.

PR #4989, opened 2020-01-17
Add itemDisabledCondition option to List view

nnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This PR adds a `itemDisabledCondition` to the `List` view that allows to pass a `jexl` expression. This `jexl` expression is evaluated on each item of the list. If the expression evaluates to true for an item, the item is rendered in a disabled state.

Additionally, this PR adds a `setItemDisabledCondition` method to the `FormOverlayListViewBuilderInterface` and the `ListViewBuilderInterface` that allows to set the `itemDisabledCondition` via a view-builder instance.

The PR builds upon the existing `disabledIds` prop of the `ListAdapterProps` type. With this PR, the following list adapters mind the given `disabledIds`:
* ColumnListAdapter
* TableAdapter
* TreeTableAdapter

#### Why?

One usecase that is enabled by this feature is to disable rows for which the current user does not have sufficient permissions. When using the `TreeTableAdapter`, it is not possible to hide those rows because the user might have sufficient permissions for a child row.

#### Example Usage

~~~php
$this->viewBuilderFactory->createListViewBuilder('xyz-view', '/xyz')
     ->setItemDisabledCondition('(_permissions && !_permissions.view)')
~~~

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/adapters/TableAdapter.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/adapters/TableAdapter.js
@@ -47,12 +47,13 @@ class TableAdapter extends AbstractTableAdapter {
     };
 
     renderRows(): Array<Element<typeof Table.Row>> {
-        const {data, selections} = this.props;
+        const {data, selections, disabledIds} = this.props;
 
         return data.map((item) => {
             return (
                 <Table.Row
                     buttons={this.getButtons(item)}
+                    disabled={disabledIds.includes(item.id)}
                     id={item.id}
                     key={item.id}
                     selected={selections.includes(item.id)}
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/adapters/TableAdapter.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/adapters/TableAdapter.js
@@ -47,7 +47,7 @@ class TableAdapter extends AbstractTableAdapter {
     };
 
     renderRows(): Array<Element<typeof Table.Row>> {
-        const {data, selections, disabledIds} = this.props;
+        const {data, disabledIds, selections} = this.props;
 
         return data.map((item) => {
             return (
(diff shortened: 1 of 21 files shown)


danrot:
Since there was no feature like this in Sulu 1.6, I would consider this a completely new one. Therefore it should go into the `master` branch.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/adapters/TableAdapter.js
```diff
@@ -47,12 +47,13 @@ class TableAdapter extends AbstractTableAdapter {
     };
 
     renderRows(): Array<Element<typeof Table.Row>> {
-        const {data, selections} = this.props;
+        const {data, selections, disabledIds} = this.props;
```
danrot: Sort alphabetically.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/tests/adapters/TreeTableAdapter.test.js
```diff
@@ -474,6 +483,58 @@ test('Render correct buttons based on permissions when item permissions are prov
     expect(treeListAdapter.find('Row').at(3).find('ButtonCell').at(1).props().disabled).toEqual(false);
 });
 
+test('Render correct buttons based on permissions when item permissions are provided', () => {
```
danrot: You don't really check the buttons here, do you?

src/Sulu/Bundle/ContactBundle/Admin/ContactAdmin.php
```diff
@@ -128,6 +128,7 @@ public function configureViews(ViewCollection $viewCollection): void
                     ->setListKey('contacts')
                     ->setTitle('sulu_contact.people')
                     ->addListAdapters(['table'])
+                    ->setItemDisabledCondition('lastName == "Test"')
```
danrot: I think that should be gone before the merge, otherwise it will result in very confusing behavior :see_no_evil:

nnatter: 😂

nnatter:
in theory master would probably the correct target branch. it practice, it would be nice to get this into 2.0 because we need this for a project and the `master` does not get updates/bugfixes very often.. @wachterjohannes

PR #4998, opened 2020-01-21
Fix assigning more than 10 teasers when using elasticsearch

### LINKED ISSUE #4331: Teaser selection has fixed limit of 10
| Q | A
| --- | ---
| Bug? | yes
| New Feature? | no
| Sulu Version | <= 1.6.23
| Browser Version | not relevant

#### Actual Behavior

When selecting more than 10 pages in a teaser_selection block, only 10 pages are displayed. This is probably because the default limit of elasticsearch is 10

#### Expected Behavior

Expected is that selecting more than 10 pages every selected page is displayed

#### Steps to Reproduce

Create a teaser_selection block and add more than 10 pages

#### Possible Solutions

Set limit here:

https://github.com/sulu/sulu/blob/1.6.23/src/Sulu/Bundle/ContentBundle/Teaser/ContentTeaserProvider.php#L61

so this block would be something like:

```
$searchResult = $this->searchManager
            ->createSearch(implode(' OR ', $statements))
            ->setLimit(count($ids))
            ->indexes($this->getPageIndexes())
            ->locale($locale)
            ->execute();
```

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #4331 
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR fixes the bug that only 10 teaser items are returned if there are more than 10 assigned.

### DIFF
--- a/src/Sulu/Bundle/TestBundle/Testing/TestUserProvider.php
+++ b/src/Sulu/Bundle/TestBundle/Testing/TestUserProvider.php
@@ -12,7 +12,6 @@
 namespace Sulu\Bundle\TestBundle\Testing;
 
 use Doctrine\ORM\EntityManager;
-use Sulu\Bundle\SecurityBundle\Entity\User;
 use Sulu\Component\Contact\Model\ContactRepositoryInterface;
 use Sulu\Component\Security\Authentication\UserRepositoryInterface;
 use Symfony\Component\Security\Core\Exception\UnsupportedUserException;
@@ -141,7 +140,7 @@ class TestUserProvider implements UserProviderInterface
      */
     public function supportsClass($class)
     {
-        return $class instanceof UserInterface;
+        return is_subclass_of($class, UserInterface::class);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/Sulu/Bundle/TestBundle/Testing/TestUserProvider.php
```diff
@@ -141,7 +140,7 @@ public function refreshUser(UserInterface $user)
      */
     public function supportsClass($class)
     {
-        return $class instanceof UserInterface;
+        return is_subclass_of($class, UserInterface::class);
```
alexander-schranz: @danrot Looks like tests did call the false user provider before for supportsClass. This was fixed in symfony 3.4.37 and so I copied the supportsClass function from our UserProvider to the TestUserProvider. As `$class instanceof UserInterface` does nothing.

alexander-schranz: In 2.0 we wrap the origina UserProvider so the fix looks different there #4999

PR #5000, opened 2020-01-22
Do not select disabled items when select-all checkbox in List is clicked

nnatter:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Related issues/PRs | Builds upon #4992
| License | MIT

#### What's in this PR?

This PR adjusts the behaviour of the `List` container to only select **non-disabled** items if the`onAllSelectionChange` callback is fired by a `ListAdapter` (for example when the select-all checkbox of the `TableAdapter` is clicked). 

Furthermore, this PR fixes a bug that rendered the select-all checkbox of the `TableAdapter` with the wrong skin. Also, the select-all checkbox of the `TableAdapter` is now rendered as checked if all non-disabled rows are selected.

#### Why?

Because the goal of disabling items is that they cannot get selected.. 😉

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Table/Header.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Table/Header.js
@@ -96,7 +96,7 @@ export default class Header extends React.PureComponent<Props> {
                     <Checkbox
                         checked={allSelected}
                         onChange={this.handleAllSelectionChange}
-                        skin={skin}
+                        skin={skin === 'dark' ? 'light' : 'dark'}
                     />
                 </span>
                 {children}
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Table/tests/Table.test.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Table/tests/Table.test.js
@@ -332,6 +332,25 @@ test('Render the Table component in multiple selection mode', () => {
     )).toMatchSnapshot();
 });
 
+test('Render the Table component in multiple selection mode with select inside first cell', () => {
+    expect(render(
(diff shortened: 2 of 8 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/components/Table/tests/Table.test.js
```diff
@@ -371,6 +390,59 @@ test('Clicking a checkbox should call onRowSelectionChange with the selection st
     expect(onChangeSpy).toHaveBeenCalledWith(rowIdOne, true);
 });
 
+test('Select-all checkbox should be checked if every non-disabled line is selected', () => {
```
danrot: Any reason not to split this test into three different tests?

src/Sulu/Bundle/AdminBundle/Resources/js/components/Table/Header.js
```diff
@@ -96,7 +96,7 @@ export default class Header extends React.PureComponent<Props> {
                     <Checkbox
                         checked={allSelected}
                         onChange={this.handleAllSelectionChange}
-                        skin={skin}
+                        skin={skin === 'dark' ? 'light' : 'dark'}
```
danrot: Why is this line necessary? As far as I see this will not change anything :thinking:

nnatter: it sets the skin of the checkbox to `light` if the skin of the header is `dark` and vice versa. this only affects tables that render the checkbox into the first cell (eg. the category list).

src/Sulu/Bundle/AdminBundle/Resources/js/components/Table/tests/Table.test.js
nnatter: done 🙂

PR #5001, opened 2020-01-22
Fix types in CollectionManagerInterface and StructureInterface

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | possible
| Deprecations? | no
| Fixed tickets | fixes -
| Related issues/PRs | -
| License | MIT

#### What's in this PR?

Fix phpdoc types in CollectionManagerInterface and StructureInterface.

#### Why?

Because they are wrong, to improve static code analyzers.

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/Collection/Manager/CollectionManagerInterface.php
+++ b/src/Sulu/Bundle/MediaBundle/Collection/Manager/CollectionManagerInterface.php
@@ -97,7 +97,7 @@ interface CollectionManagerInterface
      *
      * @return Collection
      */
-    public function save(array $data, int $userId, bool $breadcrumb = false);
+    public function save(array $data, ?int $userId = null, bool $breadcrumb = false);
 
     /**
      * Deletes a collection with a given id.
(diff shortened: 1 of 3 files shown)


src/Sulu/Bundle/MediaBundle/Collection/Manager/CollectionManagerInterface.php
```diff
@@ -97,7 +97,7 @@ public function getCount();
      *
      * @return Collection
      */
-    public function save(array $data, int $userId, bool $breadcrumb = false);
+    public function save(array $data, ?int $userId, bool $breadcrumb = false);
```
alexander-schranz: Could be a little BC Break but think that someone uses the CollectionManagerInterface alone would never happen and the CollectionManager class the $userId is `?int`

src/Sulu/Bundle/MediaBundle/Collection/Manager/CollectionManagerInterface.php
```diff
@@ -97,7 +97,7 @@ public function getCount();
      *
      * @return Collection
      */
-    public function save(array $data, int $userId, bool $breadcrumb = false);
+    public function save(array $data, ?int $userId = null, bool $breadcrumb = false);
```
alexander-schranz: As adding `?int` also to the interface is a little bc break I also did add a default value for the userId

PR #5002, opened 2020-01-22
Add getPriority function to admin pool compiler pass

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | none
| Related issues/PRs | used in sulu/SuluFormBundle#225
| License | MIT
| Documentation PR | none

#### What's in this PR?

This PR introduces a `getPriority` function in the admin.

#### Why?

This allow to put the concrete admin at a specific place inside the admin chain.

It will be used in the form bundle to ensure that the check if a view is available will be called as the last admin.

#### Example Usage

~~~php
class AppAdmin extends Admin
{
    public static function getPriority(): int
    {
        return -1024;
    }
    ...
}
~~~

#### BC Breaks/Deprecations

No BC break all the admins existing admin automatically get the priority `0`.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/DependencyInjection/Compiler/AddAdminPass.php
+++ b/src/Sulu/Bundle/AdminBundle/DependencyInjection/Compiler/AddAdminPass.php
@@ -19,6 +19,7 @@ use Symfony\Component\DependencyInjection\ContainerBuilder;
  */
 class AddAdminPass implements CompilerPassInterface
 {
+    const ADMIN_POOL_DEFINITION_ID = 'sulu_admin.admin_pool';
     const ADMIN_TAG = 'sulu.admin';
 
     /**
@@ -26,7 +27,7 @@ class AddAdminPass implements CompilerPassInterface
      */
     public function process(ContainerBuilder $container)
     {
-        $pool = $container->getDefinition('sulu_admin.admin_pool');
+        $pool = $container->getDefinition(self::ADMIN_POOL_DEFINITION_ID);
 
         $adminServiceDefinitions = [];
         foreach ($container->findTaggedServiceIds(self::ADMIN_TAG) as $id => $tags) {
--- a/src/Sulu/Bundle/AdminBundle/DependencyInjection/Compiler/AddAdminPass.php
+++ b/src/Sulu/Bundle/AdminBundle/DependencyInjection/Compiler/AddAdminPass.php
@@ -28,11 +28,24 @@ class AddAdminPass implements CompilerPassInterface
     {
         $pool = $container->getDefinition('sulu_admin.admin_pool');
 
-        $taggedServices = $container->findTaggedServiceIds(self::ADMIN_TAG);
+        $adminServiceDefinitions = [];
+        foreach ($container->findTaggedServiceIds(self::ADMIN_TAG) as $id => $tags) {
+            $serviceDefinition = $container->getDefinition($id);
 
-        foreach ($taggedServices as $id => $attributes) {
-            $admin = $container->getDefinition($id);
-            $pool->addMethodCall('addAdmin', [$admin]);
+            $class = $container->getParameterBag()->resolveValue($serviceDefinition->getClass());
+
+            /** @var callable $callable */
+            $callable = [$class, 'getPriority'];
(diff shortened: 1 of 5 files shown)


src/Sulu/Bundle/AdminBundle/DependencyInjection/Compiler/AddAdminPass.php
```diff
@@ -28,11 +28,24 @@ public function process(ContainerBuilder $container)
...
 
-        $taggedServices = $container->findTaggedServiceIds(self::ADMIN_TAG);
+        $adminServiceDefinitions = [];
+        foreach ($container->findTaggedServiceIds(self::ADMIN_TAG) as $id => $tags) {
+            $serviceDefinition = $container->getDefinition($id);
 
-        foreach ($taggedServices as $id => $attributes) {
-            $admin = $container->getDefinition($id);
-            $pool->addMethodCall('addAdmin', [$admin]);
+            $class = $container->getParameterBag()->resolveValue($serviceDefinition->getClass());
```
wachterjohannes: neccessary to be sure that class parameters are resolved (example `%sulu_media.admin.class%`)

wachterjohannes:
@danrot ping

src/Sulu/Bundle/AdminBundle/DependencyInjection/Compiler/AddAdminPass.php
```diff
@@ -28,11 +28,24 @@ public function process(ContainerBuilder $container)
...
+
+            /** @var callable $callable */
+            $callable = [$class, 'getPriority'];
+            $priority = call_user_func($callable);
+
+            $adminServiceDefinitions[$priority][] = $serviceDefinition;
+        }
+
+        krsort($adminServiceDefinitions);
+        $adminServiceDefinitions = call_user_func_array('array_merge', $adminServiceDefinitions);
```
danrot: Wouldn't `array_merge(...$adminServiceDefinitions)` be a bit easer to read?

wachterjohannes:
@danrot fixed

danrot:
Actually you only fixed half of the comments :slightly_smiling_face: 

> And shouldn't we add a test for that? Otherwise we won't recognize that we are breaking the FormBundle, because internally we don't have any priorities.

wachterjohannes:
😂 but now everything should be fixed - also added the priority to SecurityAdmin to be sure that is runs after ContactAdmin

src/Sulu/Bundle/SecurityBundle/Admin/SecurityAdmin.php
```diff
@@ -38,6 +38,14 @@ class SecurityAdmin extends Admin
 
     const EDIT_FORM_VIEW = 'sulu_security.role_edit_form';
 
+    /**
+     * Should be called after ContactAdmin.
+     */
+    public static function getPriority(): int
+    {
+        return -10;
```
danrot: In other places we were using powers of 2 for similar values, because they easily allow to put other priorities in between them. And maybe we should make sure that there is enough room in between? So use something like `-1024`?

src/Sulu/Bundle/SecurityBundle/Admin/SecurityAdmin.php
```diff
@@ -180,7 +188,9 @@ public function configureViews(ViewCollection $viewCollection): void
             );
         }
 
-        if ($this->securityChecker->hasPermission(static::USER_SECURITY_CONTEXT, PermissionTypes::EDIT)) {
+        if ($viewCollection->has(ContactAdmin::CONTACT_EDIT_FORM_VIEW)
```
danrot: IIRC this is the missing link, which would make it possible to avoid [creating the `$contactEditFormView`](https://github.com/sulu/sulu/blob/6154460/src/Sulu/Bundle/ContactBundle/Admin/ContactAdmin.php#L95-L99) in a different way. Maybe we should get rid of that in the `ContactAdmin` now? Would also allow us to remove [this weird `else`](https://github.com/sulu/sulu/blob/6154460/src/Sulu/Bundle/ContactBundle/Admin/ContactAdmin.php#L176-L178).

PR #5003, opened 2020-01-22
Adjust style of disabled items in SingleSelection and MultiSelection component

nnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Related issues/PRs | Builds upon #4992
| License | MIT

#### What's in this PR?

This PR adjusts the `SingleSelection` component and the `MultiSelection` component to render items in a disabled state if they fulfill the given `itemDisabledCondition` or their id is contained in the given `disabledIds`. Furthermore it adds a `allow_deselect_for_disabled_items` option that allows to configure if the user should be able to deselect/remove disabled items from a selection.

#### Why?

Because the user should see if a selected item is disabled without opening the `ListOverlay`.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/MultiItemSelection/item.scss
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/MultiItemSelection/item.scss
@@ -1,18 +1,24 @@
 @import '../../containers/Application/colors.scss';
 
 $itemBackgroundColor: $white;
+$disabledItemBackgroundColor: $wildSand;
+
+$itemColor: $black;
+$disabledItemColor: $gray;
 
 .item {
     display: flex;
     justify-content: space-between;
+    color: $itemColor;
     background-color: $itemBackgroundColor;
 
     &.disabled {
-        opacity: .5;
-        pointer-events: none;
+        color: $disabledItemColor;
+        background-color: $disabledItemBackgroundColor;
 
-        .buttons {
-            display: none;
+        .drag-handle,
+        .content {
+            pointer-events: none;
         }
     }
 }
@@ -37,6 +43,10 @@ $itemBackgroundColor: $white;
     padding: 13px 10px 0 0;
     cursor: pointer;
     background-color: transparent;
+
+    &.hidden {
+        display: none;
+    }
 }
 
 .drag-handle {
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/MultiItemSelection/item.scss
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/MultiItemSelection/item.scss
@@ -43,10 +43,6 @@ $disabledItemColor: $gray;
     padding: 13px 10px 0 0;
     cursor: pointer;
     background-color: transparent;
-
-    &.hidden {
-        display: none;
-    }
 }
 
 .drag-handle {
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/MultiItemSelection/item.scss
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/MultiItemSelection/item.scss
@@ -6,6 +6,15 @@ $itemBackgroundColor: $white;
     display: flex;
     justify-content: space-between;
     background-color: $itemBackgroundColor;
+
+    &.disabled {
+        opacity: .5;
+        pointer-events: none;
+
+        .buttons {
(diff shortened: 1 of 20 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/components/MultiItemSelection/item.scss
```diff
@@ -28,6 +43,10 @@ $itemBackgroundColor: $white;
     padding: 13px 10px 0 0;
     cursor: pointer;
     background-color: transparent;
+
+    &.hidden {
+        display: none;
```
danrot: Why don't we simply not render the components instead?

src/Sulu/Bundle/AdminBundle/Resources/js/components/SingleItemSelection/SingleItemSelection.js
```diff
@@ -9,6 +9,7 @@ import Button from './Button';
 import type {Button as ButtonConfig} from './types';
 
 type Props = {|
+    allowRemoveWhileDisabled: boolean,
```
danrot: I think here this naming is a little bit confusing, therefore I would change it in all places.

At least the way I would imagine it, is that if the field itself is disabled, I cannot remove anything. But if only the item assigned to it is disabled, I would think I can still remove it.

Therfore I would name this option something like `allowRemoveForDisabledItems`.

src/Sulu/Bundle/AdminBundle/Resources/js/components/SingleItemSelection/singleItemSelection.scss
```diff
@@ -44,6 +44,10 @@ $singleItemSelectionDisabledColor: $gray;
         margin: 0 10px;
         cursor: pointer;
         background-color: transparent;
+
+        &.hidden {
+            display: none;
```
danrot: Same question as above, why don't we just not render this element?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/Selection.js
```diff
@@ -183,6 +186,7 @@ export default class Selection extends React.Component<Props> {
         return (
             <MultiSelectionComponent
                 adapter={adapter}
+                allowDeselectForDisabledItems={!!allowDeselectForDisabledItems}
```
danrot: Or maybe we should also go with this name instead of what I have suggested above?

src/Sulu/Bundle/AdminBundle/Resources/js/components/MultiItemSelection/item.scss
nnatter: not rendering the elements is okay for me too 🙂

src/Sulu/Bundle/AdminBundle/Resources/js/components/SingleItemSelection/SingleItemSelection.js
nnatter: This is not the field, but the container. The container itself does not know if it is just the item that is disabled right now

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/Selection.js
```diff
@@ -159,6 +159,9 @@ export default class Selection extends React.Component<Props> {
                 item_disabled_condition: {
                     value: itemDisabledCondition,
                 } = {},
+                allow_deselect_for_disabled_items: {
+                    value: allowDeselectForDisabledItems,
```
danrot: I think the default value here should be `true`, because by default the item should be able to be removed. It is strange that an item that is actually not even allowed to be there cannot be even removed :see_no_evil:

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/SingleSelection.js
```diff
@@ -99,6 +99,9 @@ export default class SingleSelection extends React.Component<Props>
                 item_disabled_condition: {
                     value: itemDisabledCondition,
                 } = {},
+                allow_deselect_for_disabled_items: {
+                    value: allowDeselectForDisabledItems,
```
danrot: Same as above

PR #5008, opened 2020-01-24
Fix jest flow types

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

Fix jest flow types.

#### Why?

Currently we get the IDE autocompleted for example `toBeObject` but as we dont use the jest extended matchers. It will fail. So I removed the extend libraries from the jest flow types for the Developer Experience.

#### To Do


danrot:
The thing with these types is that we actually don't maintain them ourselves... We use them from the [flow-typed repository](https://github.com/flow-typed/flow-typed/blob/master/definitions/npm/jest_v24.x.x/flow_v0.39.x-v0.103.x/jest_v24.x.x.js) meaning that the next time we update them, your changes will be gone... So the cleaner solution would be to ask in this repository if they would accept a PR to split this into multiple types. Gonna create an issue for that.

danrot:
Looks like the reason for this is that flow does not support extending existing types... In that case it seems if we want to do this, we only can by deleting the lines as you did. However, this needs a rebase. Can you do that? And I can't promise that these changes are not reverted someday in the future, when we update those types.

nnatter:
Would prefer to merge this. Right now, IntelliJ suggests matcherst that are not available and flow does not detect this as error. I know that this might be overridden in the future, but hopefully the reviewer gets curious if someone adds 400 lines to this file 🙂

danrot:
@nnatter That's fine for me, just waiting for @alexander-schranz to rebase :slightly_smiling_face:

alexander-schranz:
@danrot rebased

PR #5010, opened 2020-01-27
Fix fast pagination of List view

### LINKED ISSUE #4912: Page tree gets cleared after clicking on different pages before first request is finished
| Q | A
| --- | ---
| Bug? | yes
| New Feature? | no
| Sulu Version | 2.0.2

#### Actual Behavior

The webspace page tree gets cleared completely.

#### Expected Behavior

The page tree shows the contents of the last clicked page.

#### Steps to Reproduce

Go to webspaces, click on any child page of the homepage and then on a different page immediately after the first click.
If your second click is on a page with children, all children except the first one get displayed.

danrot:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #4912 
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR cancels the request from a list when a new request for the same list has been sent.

#### Why?

Because otherwise it can happen that sooner requests finish later than the ones starting later, causing the wrong page to display. E.g. if you navigate from page 1 to page 10 by clicking the next icon without waiting for the requests to finish, you could e.g. to be shown page 9, although the list shows page 10.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/stores/ListStore.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/stores/ListStore.js
@@ -491,7 +491,7 @@ export default class ListStore {
 
         log.info('List loads "' + this.resourceKey + '" data with the following options:', options);
 
-        if (this.pendingRequest instanceof RequestPromise) {
+        if (this.pendingRequest) {
             this.pendingRequest.abort();
         }
 
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/stores/ListStore.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/stores/ListStore.js
@@ -2,7 +2,7 @@
 import {action, autorun, computed, intercept, observable, untracked} from 'mobx';
 import type {IObservableValue, IValueWillChange} from 'mobx';
 import log from 'loglevel';
-import ResourceRequester from '../../../services/ResourceRequester';
+import ResourceRequester, {RequestPromise} from '../../../services/ResourceRequester';
 import type {
     LoadingStrategyInterface,
     ObservableOptions,
@@ -55,6 +55,7 @@ export default class ListStore {
     activeSettingDisposer: () => void;
     sendRequestDisposer: () => void;
     initialSelectionIds: ?Array<string | number>;
+    pendingRequest: ?RequestPromise<*>;
     metadataOptions: ?Object;
 
     static getActiveSetting(listKey: string, userSettingsKey: string): string {
@@ -490,11 +491,16 @@ export default class ListStore {
 
         log.info('List loads "' + this.resourceKey + '" data with the following options:', options);
 
-        this.loadingStrategy.load(
+        if (this.pendingRequest instanceof RequestPromise) {
+            this.pendingRequest.abort();
+        }
+
+        this.pendingRequest = this.loadingStrategy.load(
             this.resourceKey,
             options,
             options.expandedIds ? undefined : active
         ).then(action((response) => {
+            this.pendingRequest = undefined;
             this.pageCount = response.pages;
             this.setDataLoading(false);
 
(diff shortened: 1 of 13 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/stores/ListStore.js
```diff
@@ -511,6 +517,11 @@ export default class ListStore {
                 this.initialSelectionIds = undefined;
             }
         })).catch((response) => {
+            if (response.name === 'AbortError') {
```
nnatter: i am a bit hesitant to rely on the `name` here.. isnt there at least a constant somewhere that we could use?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/stores/ListStore.js
```diff
@@ -490,11 +491,16 @@ export default class ListStore {
 
         log.info('List loads "' + this.resourceKey + '" data with the following options:', options);
 
-        this.loadingStrategy.load(
+        if (this.pendingRequest instanceof RequestPromise) {
```
nnatter: wouldnt it be more intuitive to check if it is set/not undefined? 🙈

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/stores/ListStore.js
nnatter: also why dont you set `this.pendingRequest` to undefined in this case? :

nnatter:
looks good otherwise. but i need a rebase 🙂

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/stores/ListStore.js
danrot: Actually I thought that it makes more sense this way... I even thought about a `CancelablePromise` or something like that, but the issue is that only requests can be cancelled.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/stores/ListStore.js
danrot: At least the `name` attribute is safe to use, because it is standardized, as documented in [MDN](https://developer.mozilla.org/en-US/docs/Web/API/AbortController/AbortController) and [Google Developers](https://developers.google.com/web/updates/2017/09/abortable-fetch#reacting_to_an_aborted_fetch).

src/Sulu/Bundle/AdminBundle/Resources/js/services/Requester/RequestPromise.js
```diff
@@ -0,0 +1,28 @@
...
+
+    then(onFulfilled: ?(*) => Promise<*> | *, onRejected: ?(*) => Promise<*> | *): RequestPromise<*> {
+        const requestPromise: RequestPromise<*> = ((super.then(onFulfilled, onRejected): any): RequestPromise<*>);
+        requestPromise.setAbortController(this.abortController);
+
+        return requestPromise;
+    }
+
+    catch(onReject: ?(*) => Promise<*> | *): RequestPromise<*> {
+        // $FlowFixMe
```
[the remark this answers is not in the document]
danrot: Removed the `$FlowFixMe` and set the AbortController. Strictly speaking the latter would probably not be necessary, because you can't cancel something that has already failed... Added it anyway just to be sure, maybe there is some kind of edge case.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/stores/ListStore.js
danrot: Realized that because of some type changes I don't even need the `instanceof` anymore, therefore I removed it :slightly_smiling_face:

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/stores/ListStore.js
danrot: Regarding setting `this.pendingRequest`. We could do that, but I think it is safer not to. The only place we abort the request is right before we are assigning a new value to `this.pendingRequest` anyway. So assigning another value does not make too much sense, since it will be reassigned immediately. Additionally I am not sure if that could cause some kind of race condition, therefore I think it is safer not to do it.

PR #5011, opened 2020-01-28
Add request_parameters option to Selection field-type

nnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This PR adds two schema-options to the `Selection` field-type. The `request_parameters` option allows to set static parameters that are appended to the requests made by the field-type and therefore is similar to the `ListViewBuilderInterface::addRequestParameters` method. 

The `resource_store_properties_to_request` option allows to pass the name of properties. The value of these properties then is appended to the requests made by the field-type. Furthermore, if the value of one of the property changes, the field-type fires a new request. This allows to implement a dependent selects (eg a city select that only displays the cities of the currently selected country). This option is similar to the `ListViewBuilderInterface::addResourceStorePropertiesToListRequest` method.

#### Why?

These options make the `Selection` field-type more versatile. For example, the `request_parameters` option can be used to filter selectable items by sending a specific type to the server and only return matching items. 

#### Example Usage

~~~xml
<property name="topics" type="topic_selection">
    <meta>
        <title>sulu_directory.topics</title>
    </meta>
    <params>
        <param name="request_parameters" type="collection">
            <param name="includeAssignPermissions" value="true" />
        </param>
        <param name="resource_store_properties_to_request" type="collection">
            <param name="assignedDirectoryIds" value="directories" />
        </param>
    </params>
</property>
~~~

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/Selection.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/Selection.js
@@ -174,7 +174,7 @@ class Selection extends React.Component<Props> {
 
         resourceStorePropertiesToRequest.forEach((propertyToRequest) => {
             const {name: parameterName, value: propertyName} = propertyToRequest;
-            const propertyPath = propertyName || parameterName;
+            const propertyPath = typeof propertyName === 'string' ? propertyName : parameterName;
             requestOptions[parameterName] = formInspector.getValueByPath('/' + propertyPath);
         });
 
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/Selection.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/Selection.js
@@ -1,6 +1,6 @@
 // @flow
 import React from 'react';
-import {computed, observable, intercept, toJS, reaction, autorun} from 'mobx';
+import {computed, observable, intercept, toJS, reaction} from 'mobx';
 import type {IObservableValue} from 'mobx';
 import equal from 'fast-deep-equal';
 import {observer} from 'mobx-react';
@@ -12,6 +12,7 @@ import {translate} from '../../../utils/Translator';
 import MultiSelectionComponent from '../../MultiSelection';
 import userStore from '../../../stores/userStore';
 import type {FieldTypeProps} from '../../../types';
+import type {SchemaOption} from '../types';
 import selectionStyles from './selection.scss';
 
 type Props = FieldTypeProps<Array<string | number>>;
@@ -56,12 +57,19 @@ class Selection extends React.Component<Props> {
                 },
                 formInspector,
                 schemaOptions: {
+                    request_parameters: {
+                        value: requestParameters = [],
+                    } = {},
                     resource_store_properties_to_request: {
                         value: resourceStorePropertiesToRequest = [],
                     } = {},
                 },
                 value,
             } = this.props;
+
+            if (!Array.isArray(requestParameters)) {
+                throw new Error('The "request_parameters" schemaOption must be an array!');
+            }
 
             if (!Array.isArray(resourceStorePropertiesToRequest)) {
                 throw new Error('The "resource_store_properties_to_request" schemaOption must be an array!');
@@ -72,7 +80,7 @@ class Selection extends React.Component<Props> {
                 listKey || resourceKey,
                 USER_SETTINGS_KEY,
                 {locale: this.locale, page: observable.box()},
-                this.buildRequestOptions({}, resourceStorePropertiesToRequest, formInspector),
+                this.buildRequestOptions(requestParameters, resourceStorePropertiesToRequest, formInspector),
                 undefined,
                 value
             );
@@ -83,13 +91,18 @@ class Selection extends React.Component<Props> {
             );
 
             this.changeListOptionsDisposer = reaction(
-                () => this.buildRequestOptions({}, resourceStorePropertiesToRequest, formInspector),
+                () => this.buildRequestOptions(requestParameters, resourceStorePropertiesToRequest, formInspector),
                 (requestOptions) => {
+                    const listStore = this.listStore;
+                    if (!listStore) {
+                        throw new Error('The ListStore has not been initialized! This is likely a bug.');
+                    }
+
                     // reset liststore to reload whole tree instead of children of current active item
-                    this.listStore.reset();
+                    listStore.reset();
                     // set selected items as initialSelectionIds to expand them in case of a tree
-                    this.listStore.initialSelectionIds = this.listStore.selectionIds;
-                    this.listStore.options = {...this.listStore.options, ...requestOptions};
+                    listStore.initialSelectionIds = listStore.selectionIds;
+                    listStore.options = {...listStore.options, ...requestOptions};
                 }
             );
 
@@ -149,15 +162,20 @@ class Selection extends React.Component<Props> {
     }
 
     buildRequestOptions(
-        requestParameters: Object,
+        requestParameters: Array<SchemaOption>,
         resourceStorePropertiesToRequest: Array<SchemaOption>,
         formInspector: FormInspector
     ) {
-        const requestOptions = requestParameters ? requestParameters : {};
+        const requestOptions = {};
+
+        requestParameters.forEach((parameter) => {
+            requestOptions[String(parameter.name)] = parameter.value;
+        });
 
         resourceStorePropertiesToRequest.forEach((propertyToRequest) => {
             const {name: parameterName, value: propertyName} = propertyToRequest;
-            requestOptions[parameterName] = formInspector.getValueByPath('/' + propertyName || parameterName);
+            const propertyPath = String(propertyName || parameterName);
+            requestOptions[String(parameterName)] = formInspector.getValueByPath('/' + propertyPath);
         });
 
         return requestOptions;
@@ -197,6 +215,9 @@ class Selection extends React.Component<Props> {
                 },
             },
             schemaOptions: {
+                request_parameters: {
+                    value: requestParameters = [],
+                } = {},
                 resource_store_properties_to_request: {
                     value: resourceStorePropertiesToRequest = [],
                 } = {},
@@ -212,6 +233,10 @@ class Selection extends React.Component<Props> {
             },
             value,
         } = this.props;
+
+        if (!Array.isArray(requestParameters)) {
+            throw new Error('The "request_parameters" schemaOption must be an array!');
+        }
 
         if (!Array.isArray(resourceStorePropertiesToRequest)) {
             throw new Error('The "resource_store_properties_to_request" schemaOption must be an array!');
@@ -229,13 +254,13 @@ class Selection extends React.Component<Props> {
             throw new Error('The "allow_deselect_for_disabled_items" schema option must be a boolean if given!');
         }
 
-        const requestParameters = {};
-        if (types) {
-            requestParameters.types = types;
-        }
-
         if (!adapter) {
             throw new Error('The selection field needs a "adapter" option to work properly');
+        }
+
+        const options = this.buildRequestOptions(requestParameters, resourceStorePropertiesToRequest, formInspector);
+        if (types) {
+            options.types = types;
         }
 
         return (
@@ -251,7 +276,7 @@ class Selection extends React.Component<Props> {
                 listKey={listKey || resourceKey}
                 locale={this.locale}
                 onChange={this.handleMultiSelectionChange}
-                options={this.buildRequestOptions(requestParameters, resourceStorePropertiesToRequest, formInspector)}
+                options={options}
                 overlayTitle={translate(overlayTitle)}
                 resourceKey={resourceKey}
                 value={value || []}
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/Selection.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/Selection.js
@@ -4,7 +4,7 @@ import {computed, observable, intercept, toJS, reaction} from 'mobx';
 import type {IObservableValue} from 'mobx';
 import equal from 'fast-deep-equal';
 import {observer} from 'mobx-react';
-import {FormInspector} from '../../../containers';
+import FormInspector from '../FormInspector';
 import List from '../../../containers/List';
 import ListStore from '../../../containers/List/stores/ListStore';
 import MultiAutoComplete from '../../../containers/MultiAutoComplete';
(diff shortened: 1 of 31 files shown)


danrot:
That one is not yet finished, right? Needs at least a rebase and a description which is not just the PR template :wink:

nnatter:
no, not finished yet. unfortunately i do not know how i can mark this as a draft after the first save 😇

danrot:
Unfortunately you can't know, because [it is not possible](https://github.community/t5/How-to-use-Git-and-GitHub/Feature-Request-Switch-from-ready-to-draft-in-pull-requests/td-p/19107) :confused:

danrot:
@nnatter You ping me when this is done, right? :thinking:

nnatter:
@danrot finally ready 🙂

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/Selection.js
```diff
@@ -119,6 +161,26 @@ export default class Selection extends React.Component<Props> {
...
 
+    buildRequestOptions(
+        requestParameters: Array<SchemaOption>,
+        resourceStorePropertiesToRequest: Array<SchemaOption>,
+        formInspector: FormInspector
+    ) {
+        const requestOptions = {};
+
+        requestParameters.forEach((parameter) => {
+            requestOptions[String(parameter.name)] = parameter.value;
```
danrot: Why is the `String` call necessary here? Have never seen that before in our codebase :thinking:

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/Selection.js
```diff
@@ -119,6 +161,26 @@ export default class Selection extends React.Component<Props> {
...
+        });
+
+        resourceStorePropertiesToRequest.forEach((propertyToRequest) => {
+            const {name: parameterName, value: propertyName} = propertyToRequest;
+            const propertyPath = String(propertyName || parameterName);
```
danrot: Same as above

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/Selection.js
```diff
@@ -119,6 +161,26 @@ export default class Selection extends React.Component<Props> {
...
+            requestOptions[String(parameterName)] = formInspector.getValueByPath('/' + propertyPath);
```
danrot: Same as above

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/tests/fields/Selection.test.js
```diff
@@ -557,7 +625,71 @@ test('Should pass props correctly to list component', () => {
...
+
+    const locale = observable.box('en');
+
+    const formInspector = new FormInspector(
+        new ResourceFormStore(
+            new ResourceStore('pages', 1, {locale}),
+            'pages'
+        )
+    );
+    formInspector.getValueByPath.mockReturnValue('value-returned-by-form-inspector');
```
danrot: I think we should either implement the mock in a way that this value is only returned if `otherPropertyName` is passed as argument, or just check afterwards if the mock was actually called with that value.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/MultiListOverlay/MultiListOverlay.js
```diff
@@ -1,6 +1,6 @@
 // @flow
 import React from 'react';
-import {computed, observable} from 'mobx';
+import {computed, observable, reaction, comparer} from 'mobx';
```
danrot: Sort alphabetically.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/MultiListOverlay/MultiListOverlay.js
```diff
@@ -67,13 +68,27 @@ class MultiListOverlay extends React.Component<Props> {
...
+        this.changeOptionsDisposer = reaction(
+            () => this.props.options,
+            (options) => {
+                // reset liststore to reload whole tree instead of children of current active item
+                this.listStore.reset();
+                // set selected items as initialSelectionIds to expand them in case of a tree
+                this.listStore.initialSelectionIds = this.listStore.selectionIds;
+                this.listStore.options = {...this.listStore.options, ...options};
+            },
+            {equals: comparer.structural}
```
danrot: Why is that needed here, but not in the `Selection` field type?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/MultiListOverlay/MultiListOverlay.js
```diff
@@ -67,13 +68,27 @@ class MultiListOverlay extends React.Component<Props> {
             USER_SETTINGS_KEY,
             observableOptions,
             options,
+            undefined,
             preloadSelectedItems ? preSelectedItems.map((preSelectedItem) => preSelectedItem.id) : undefined
         );
+
+        this.changeOptionsDisposer = reaction(
+            () => this.props.options,
```
danrot: I think that this is running too often... You should be able to avoid that, by creating a computed getter for `options`.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/MultiSelection/MultiSelection.js
```diff
@@ -59,6 +61,15 @@ class MultiSelection extends React.Component<Props> {
                 }
             }
         );
+
+        this.changeOptionsDisposer = reaction(
+            () => this.props.options,
```
danrot: Same as above

src/Sulu/Bundle/AdminBundle/Resources/js/containers/MultiSelection/tests/MultiSelection.test.js
```diff
@@ -165,28 +188,85 @@ test('Pass itemDisabledCondition to MultiListOverlay', () => {
     expect(selection.find('MultiListOverlay').prop('itemDisabledCondition')).toEqual('status == "inactive"');
 });
 
-test('Show with passed values as items in right locale', () => {
+test('Construct MultiSelectionStore with correct parameters', () => {
     const locale = observable.box('en');
+    const options = {key: 'value-1'};
```
danrot: I think I would not extract that into an own variable... The `MultiSelection` component could falsely manipulate that object, and then the check below would still work.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/MultiSelection/tests/MultiSelection.test.js
```diff
@@ -165,28 +188,85 @@ test('Pass itemDisabledCondition to MultiListOverlay', () => {
...
+        />
+    );
 
-    expect(render(
+    expect(MultiSelectionStore).toBeCalledWith('snippets', [1, 2, 5], locale, 'ids', options);
+});
+
+test('Update requestParameters and reload items of MultiSelectionStore when options prop is changed', () => {
+    const locale = observable.box('en');
+    const oldOptions = {key: 'value-1'};
```
danrot: Same as above

src/Sulu/Bundle/AdminBundle/Resources/js/stores/MultiSelectionStore/MultiSelectionStore.js
```diff
@@ -10,19 +10,21 @@ export default class MultiSelectionStore<T = string | number, U: {id: T} = Objec
     resourceKey: string;
     locale: ?IObservableValue<string>;
     idFilterParameter: string;
+    requestParameters: {[string]: any};
```
danrot: Wouldn't `{[string]: mixed}` be better for type safety? Or is that too much work to get it working?

src/Sulu/Bundle/AdminBundle/Resources/js/stores/MultiSelectionStore/MultiSelectionStore.js
```diff
@@ -10,19 +10,21 @@ export default class MultiSelectionStore<T = string | number, U: {id: T} = Objec
...
 
     constructor(
         resourceKey: string,
         selectedItemIds: Array<T>,
         locale: ?IObservableValue<string>,
-        idFilterParameter: string = 'ids'
+        idFilterParameter: string = 'ids',
+        requestParameters: {[string]: any} = {}
```
danrot: Same as above

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/Selection.js
nnatter: because the `SchemaOption` type is defined as:
```
export type SchemaOption = {
    infoText?: string,
    name?: string | number,
    title?: string,
    value?: ?string | number | boolean | Array<SchemaOption>,
};
```

Therefore `parameter.name` could be `undefined` in theory and flow complains that i cannot use `undefined` as kay for the assignment. do you have a better idea how to solve this? should we throw an error?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/MultiListOverlay/MultiListOverlay.js
nnatter: using `reaction` actually has the same effect as using a `computed` getter inside a `autorun`. the side effect of the `reaction` is only executed if the return value of the first function changes. in case of the `autorun`, the `autorun` is only executed if the value of the `computed` getter is changed (but not if one of the observables used inside the getter is changed).

see: https://mobx.js.org/refguide/reaction.html

src/Sulu/Bundle/AdminBundle/Resources/js/containers/MultiListOverlay/MultiListOverlay.js
nnatter: okay so this will get kinda long 🙈 

in case of the `Selection` field type we are reacting to everything that is accessed in the `buildRequestOptions` method. The `buildRequestOptions` method accesses the observable properties of the resource store through the `formInspector.getValueByPath` method. this reaction will only run, if one of these observable properties is changed.

in case of the `MultiSelection` component (i am using that one for this example because it is higher up in the component tree, the explanation is the same for the `MultiListOverlay`), we are reacting to the value of `props.options`.  This `props.options` object is constructed in the render method of the `Selection` field type. Therefore, if the `Selection` field type is rerendered (eg. because it gets disabled), a new object will be constructed for `props.options`. 

Unfortunately, javascript evaluates `{key: 'value'} == {key: 'value'}` to `false` and therefore the new assignment of the `props.options` property will fire the reaction inside the `MultiSelection`, even if the scalar values inside the `props.options` did not change.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/MultiSelection/MultiSelection.js
nnatter: see answer above 🙂

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/Selection.js
danrot: I think the `SchemaOption` type is wrong... In a template XML you have to set a name, but the value can be left empty. Therefore I would say that we change the type here.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/MultiListOverlay/MultiListOverlay.js
danrot: I think that makes sense :slightly_smiling_face:

danrot:
@nnatter Seems like this needs a rebase, because I have already merged #5003

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/Selection.js
```diff
@@ -166,6 +234,14 @@ export default class Selection extends React.Component<Props> {
             value,
         } = this.props;
 
+        if (!Array.isArray(requestParameters)) {
+            throw new Error('The "request_parameters" schemaOption must be an array!');
+        }
+
+        if (!Array.isArray(resourceStorePropertiesToRequest)) {
```
danrot: Wouldn't it make sense to put this check into the `buildRequestOptions` function instead of doing it twice in two different places?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/Selection.js
```diff
@@ -68,6 +82,22 @@ export default class Selection extends React.Component<Props> {
                 this.handleListSelectionChange
             );
 
+            this.changeListOptionsDisposer = reaction(
+                () => this.buildRequestOptions(requestParameters, resourceStorePropertiesToRequest, formInspector),
+                (requestOptions) => {
+                    const listStore = this.listStore;
```
danrot: This reaction runs too often. I have added a `resource_store_properties_to_request` option to a text field, and on every single keystroke this code is ran and sends another request... This should actually be done only when a field calls its `onFinish` handler.

The `ResourceLocator` field has already a finished handler, which acts this way.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/Selection.js
```diff
@@ -40,12 +42,53 @@ class Selection extends React.Component<Props> {
...
+        }
+
+        this.requestOptions = this.buildRequestOptions(
+            requestParameters,
+            resourceStorePropertiesToRequest,
+            formInspector
+        );
+
+        // update requestOptions observable if one of the "resource_store_properties_to_request" properties is changed
+        const observedDataPaths = resourceStorePropertiesToRequest.map((property) => {
```
danrot: Why don't you move that code into the finish handler? This way it won't work if the `resourceSTorePropertiesToRequest` change.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/Selection.js
```diff
@@ -201,7 +202,7 @@ class Selection extends React.Component<Props> {
         resourceStorePropertiesToRequest.forEach((propertyToRequest) => {
             const {name: parameterName, value: propertyName} = propertyToRequest;
             const propertyPath = typeof propertyName === 'string' ? propertyName : parameterName;
-            requestOptions[parameterName] = formInspector.getValueByPath('/' + propertyPath);
+            requestOptions[parameterName] = toJS(formInspector.getValueByPath('/' + propertyPath));
```
danrot: Why was that necessary?

nnatter: because the value of the property might be an observable array and the `fast-deep-equal` package unfortunately does not compare the values inside of these 🙈 

for example, we need to do the same thing in the `MultiSelection` component: https://github.com/sulu/sulu/blob/master/src/Sulu/Bundle/AdminBundle/Resources/js/containers/MultiSelection/MultiSelection.js#L56

PR #5012, opened 2020-01-29
Update to jest 25 for better testing performance

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | #5008
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR updates jest to version 25.

#### Why?

Because there were hardly any BC breaks, and our tests run 10-20s faster then before.


src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Option.js
```diff
@@ -12,7 +11,6 @@ type Props<T> = {|
     anchorWidth: number,
     children: string,
     disabled: boolean,
-    focus: boolean,
```
danrot: I have checked for this property, and it wasn't used in the entire code base. Therefore I've decided to remove it.

alexander-schranz: Should we not focus always the selected element when opening the select dropdown?

danrot: That's not done via this attribute (just double-checked, there is no occurence of this property, and the selects are still working fine).

Issue #2300, opened 2016-04-15
Show only languages with permissions


sabinebaer:
| Q | A
| --- | ---
| Bug? | no
| New Feature? | yes
| Sulu Version | 1.2.0
| Browser Version | All

#### Actual Behavior
All available languages are shown in the language switcher in admin area, even if the current user has no permissions for this language.

![sulu-languages](https://cloud.githubusercontent.com/assets/5627878/14555561/a5dd45ae-02f3-11e6-8cdf-9b03a7592289.png)

#### Expected Behavior

Only languages for which the user has at least view permissions should be shown. Not sure about what should happen if the current user do not have permissions for any language.

danrot:
If the user doesn't have permission for any language, he should probably not even be able to navigate there.

tniebergall:
Hi are there any news about this issue?

danrot:
@tniebergall Hey, since we are already working on Sulu 2.0 I am not sure if we will fix that for the 1.x series. Or is that a real blocker for you?

tniebergall:
While we're in the development phase until february 2018 not, but after that it is a real blocker because we've 46 languages and most of the users only have access to 2-5 languages.

danrot:
I'd say we try to figure out how we solve this issue in Sulu 2.0, and afterwards we can see if we are able to backmerge that into the 1.6 branch :-)

danrot:
Also see comment at https://github.com/sulu/sulu/issues/5004#issuecomment-579628339

alexander-schranz:
Parts of this issue are fixed in #5016 we still show all languages in the dropdown but give the user Feedback if he had no permissions for that page:

![Language-no-permission-sulu](https://user-images.githubusercontent.com/1698337/73447041-339aef00-435e-11ea-8567-ecc585a2b9ca.png)

PR #5016, opened 2020-01-29
Show hint for missing permissions in forms

danrot:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes parts of #2300
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR shows a hint for missing permissions, when a form is loaded which is not granted access for the current user. This can easily happen when single languages are restricted for certain users.


nnatter:
looks like the API does not prevent to delete system collections but prevents to edit them. i think we should fix this too 🙂

PR #5018, opened 2020-01-29
Disallow add, delete and edit operations on SystemCollections

### LINKED ISSUE #5005: System Collections should not be able to be edit / removed
| Q | A
| --- | ---
| Sulu Version | 2.0.3
| Browser Version | All

# Actual Behavior

Currently in the UI the following Buttons are shown:

 - Move Collection
 - Edit Collection
 - Delete Collection
 - Add new Collection

all this should not be possible for a System Collection. Most of that will throw currently an Api error (unless delete which should really not be possible for system collections).

# Expected Behavior

The buttons should not be available for system collections and we should show system collection with another Icon maybe a lock icon.

# Steps to Reproduce

Try to add collections inside a system collection the api will error. 

# Possible Solutions

Disable or hide the buttons by the condition that the collection is from type system.

danrot:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #5005 
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR only shows the add, delete and edit operations on Collections not being a SystemCollection.

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/Controller/CollectionController.php
+++ b/src/Sulu/Bundle/MediaBundle/Controller/CollectionController.php
@@ -107,9 +107,6 @@ class CollectionController extends AbstractRestController implements ClassResour
     /**
      * Shows a single collection with the given id.
      *
-     * @param $id
-     * @param Request $request
-     *
      * @return \Symfony\Component\HttpFoundation\Response
      */
     public function getAction($id, Request $request)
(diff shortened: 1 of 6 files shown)


nnatter:
The API prevents to edit `system_collection` collections but not to remove them. I think we should adjust this to prevent to remove them too. 🙂

PR #5022, opened 2020-01-30
Add default sorting for PageDataProvider

### LINKED ISSUE #5017: Add order option in smart content sorting
# Problem description

I have a set of pages that are sorted in a specific way, now i need the top 3 items displayed on the homepage. I would like to use smart content for this but in the sort by dropdown you have no option that takes your sorting into account.

sort-by

# Proposed solution

It would be nice to have an option called order so that when selecting this it maintains the order you made. if i would then select order en limit my results to 3 items. i would get the top 3 items.

danrot:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #5017
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR allows to set the default sort order for the `PageDataProvider`. This is the admin order as given in our webspace page tree if the "Include sub folder" toggle is not set, and an alphabetic representation if the toggle is set.

### DIFF
--- a/src/Sulu/Component/Content/SmartContent/QueryBuilder.php
+++ b/src/Sulu/Component/Content/SmartContent/QueryBuilder.php
@@ -181,7 +181,12 @@ class QueryBuilder extends ContentQueryBuilder
         $sortBy = $this->getConfig('sortBy');
 
         if ($sortBy) {
-            $sql2Order[] = 'page.[i18n:' . $locale . '-' . $sortBy . '] ' . $sortOrder;
+            $order = 'page.[i18n:' . $locale . '-' . $sortBy . '] ';
+            if (!in_array($sortBy, ['published', 'created', 'changed', 'authored'])) {
+                $order = sprintf('lower(%s)', $order);
+            }
+
+            $sql2Order[] = $order . $sortOrder;
         } elseif (!$this->getConfig('includeSubFolders', false)) {
             $sql2Order[] = 'page.[sulu:order] ' . $sortOrder;
         }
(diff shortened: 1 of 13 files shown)


src/Sulu/Component/Content/SmartContent/QueryBuilder.php
```diff
@@ -178,17 +178,11 @@ protected function buildOrder($webspaceKey, $locale)
...
-            $order = 'page.[i18n:' . $locale . '-' . $sortBy . '] ';
-            if (!in_array($sortBy, ['published', 'created', 'changed', 'authored'])) {
-                $order = sprintf('lower(%s)', $order);
-            }
-
-            $sql2Order[] = $order . ' ' . $sortOrder;
-        } else {
+        if ($sortBy) {
+            $sql2Order[] = 'page.[i18n:' . $locale . '-' . $sortBy . '] ' . $sortOrder;
+        } elseif (!$this->getConfig('includeSubFolders', false)) {
```
danrot: @chirimoya This fix is kind of a compromise... Instead of Admin sorting there is now a default sorting, which behaves differently based on the `includeSubFolders` flag. If it is not set we sort it after the order of our page tree, and if it is set we don't sort it at all. Then PHPCR seems to return a less confusing order, although it is not respecting the `sulu:order` field.

chirimoya: 🙈 ... +1

PR #5024, opened 2020-01-31
Replace host right after loading webspaces to avoid repeating this logic

### LINKED ISSUE #5020: {host} is not correctly replaced in the preview
| Q | A
| --- | ---
| Sulu Version | 2.0.3
| Browser Version | Chrome 79.0.3945.130

# Actual Behavior

The {host} part is not correctly replaced in the preview.

# Expected Behavior

It should replace the current host

# Steps to Reproduce

Displaying the preview in /admin 🙂

danrot:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #5020
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR will replace the host immediately after the webspace collection is loaded from the cache. This has the advantage, that nobody else has to take care of the `{host}` placeholder.

#### Why?

Because it solves #5020, and allows to delete other code handling the exact same thing.

#### BC Breaks/Deprecations

The `{host}` placeholder is not available anymore. Code that relies on that is not necessary and anymore and can be deleted. In case somebody abused that placeholder for something else, that code might not work anymore.

#### To Do

- [x] Delete code not being necessary anymore

### DIFF
--- a/src/Sulu/Bundle/WebsiteBundle/Routing/PortalLoader.php
+++ b/src/Sulu/Bundle/WebsiteBundle/Routing/PortalLoader.php
@@ -47,7 +47,7 @@ class PortalLoader extends Loader
                     $importedRoute->getDefaults(),
                     array_merge(['prefix' => '.*', 'host' => '.+'], $importedRoute->getRequirements()),
                     $importedRoute->getOptions(),
-                    (!empty($importedRoute->getHost()) ? $importedRoute->getHost() : '{host}'),
+                    $importedRoute->getHost(),
                     $importedRoute->getSchemes(),
                     $importedRoute->getMethods(),
                     $condition . (!empty($importedCondition) ? ' and (' . $importedCondition . ')' : '')
--- a/src/Sulu/Component/Webspace/Manager/WebspaceManager.php
+++ b/src/Sulu/Component/Webspace/Manager/WebspaceManager.php
@@ -124,10 +124,6 @@ class WebspaceManager implements WebspaceManagerInterface
             function(PortalInformation $portalInformation) use ($host) {
                 $portalHost = $portalInformation->getHost();
 
-                if ($this->urlReplacer->hasHostReplacer($portalHost)) {
-                    $portalHost = $this->urlReplacer->replaceHost($portalHost, $host);
-                }
-
                 // add a slash to avoid problems with "example.co" and "example.com"
                 return false !== strpos($portalHost . '/', $host . '/');
             }
@@ -471,10 +467,6 @@ class WebspaceManager implements WebspaceManagerInterface
         if (false !== strpos($portalUrl, '/')) {
             // trim slash when resourceLocator is not domain root
             $resourceLocator = rtrim($resourceLocator, '/');
-        }
-
-        if ($domain && $this->urlReplacer->hasHostReplacer($portalUrl)) {
-            $portalUrl = $this->urlReplacer->replaceHost($portalUrl, $domain);
         }
 
         return rtrim(sprintf('%s://%s', $scheme, $portalUrl), '/') . $resourceLocator;
--- a/src/Sulu/Component/Webspace/Manager/WebspaceManager.php
+++ b/src/Sulu/Component/Webspace/Manager/WebspaceManager.php
@@ -21,6 +21,7 @@ use Sulu\Component\Webspace\Url\ReplacerInterface;
 use Sulu\Component\Webspace\Webspace;
 use Symfony\Component\Config\ConfigCache;
 use Symfony\Component\Config\Loader\LoaderInterface;
+use Symfony\Component\HttpFoundation\RequestStack;
 
 /**
  * This class is responsible for loading, reading and caching the portal configuration files.
@@ -33,11 +34,6 @@ class WebspaceManager implements WebspaceManagerInterface
     private $webspaceCollection;
 
     /**
-     * @var array
-     */
-    private $options;
-
-    /**
      * @var LoaderInterface
      */
     private $loader;
@@ -48,20 +44,39 @@ class WebspaceManager implements WebspaceManagerInterface
     private $urlReplacer;
 
     /**
+     * @var RequestStack
+     */
+    private $requestStack;
+
+    /**
+     * @var array
+     */
+    private $options;
+
+    /**
      * @var string
      */
     private $environment;
 
+    /**
+     * @var string
+     */
+    private $defaultHost;
+
     public function __construct(
         LoaderInterface $loader,
         ReplacerInterface $urlReplacer,
+        RequestStack $requestStack,
         array $options,
-        string $environment
+        string $environment,
+        string $defaultHost
     ) {
         $this->loader = $loader;
         $this->urlReplacer = $urlReplacer;
+        $this->requestStack = $requestStack;
         $this->setOptions($options);
         $this->environment = $environment;
+        $this->defaultHost = $defaultHost;
     }
 
     public function findWebspaceByKey(?string $key): ?Webspace
@@ -362,6 +377,19 @@ class WebspaceManager implements WebspaceManagerInterface
             require_once $cache->getPath();
 
             $this->webspaceCollection = new $class();
+
+            $currentRequest = $this->requestStack->getCurrentRequest();
+
+            $host = $currentRequest ? $currentRequest->getHost() : $this->defaultHost;
+            foreach ($this->getPortalInformations() as $portalInformation) {
+                $portalInformation->setUrl($this->urlReplacer->replaceHost($portalInformation->getUrl(), $host));
+                $portalInformation->setUrlExpression(
+                    $this->urlReplacer->replaceHost($portalInformation->getUrlExpression(), $host)
+                );
+                $portalInformation->setRedirect(
+                    $this->urlReplacer->replaceHost($portalInformation->getRedirect(), $host)
+                );
+            }
         }
 
         return $this->webspaceCollection;
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,13 @@
 # Upgrade
 
+## 2.0.4
+
+### Replace {host} placeholder right after loading webspaces
+
+Previously a `host` parameter had to be passed when a route of type `portal` was generated on a webspace containing a
+`{host}` placeholder in its URL. This is not necessary anymore, and if you still do it, it will cause a `host` query
+parameter to be added to the generated URL.
+
 ## 2.0.3
 
 When upgrading also have a look at the changes in the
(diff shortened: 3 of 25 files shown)


danrot:
@sulu/core-team Any objections? Any code I should check in particular? The only thing that came to my mind was the sitemap dumping, which seems to keep working until now (although I probably will touch that code once more anyway).

src/Sulu/Component/Webspace/Manager/WebspaceManager.php
```diff
@@ -362,6 +370,17 @@ public function getWebspaceCollection(): WebspaceCollection
             require_once $cache->getPath();
 
             $this->webspaceCollection = new $class();
+
+            $currentRequest = $this->requestStack->getCurrentRequest();
+
+            // TODO change default to whatever is used for sitemaps, so the code can also be removed there
+            $host = $currentRequest ? $currentRequest->getHost() : '{host}';
```
alexander-schranz: think it should be `getHttpHost()` so the port is also added or?

alexander-schranz: Can we if we have no request available fallback to the parameter which symfony uses to generate routes for CLI:

https://symfony.com/doc/current/routing.html#generating-urls-in-commands

alexander-schranz: If I remember correclty I used that also for the dump of the sitemap.

danrot:
@alexander-schranz Would that be ok for you like that? If yes I would adjust the code in other places as well (There is some replacement code lying around not needed anymore).

danrot:
There are still some `{host}` occurences in our code I have to double-check...

src/Sulu/Bundle/WebsiteBundle/Routing/PortalLoader.php
```diff
@@ -47,7 +47,7 @@ public function load($resource, $type = null)
                     $importedRoute->getDefaults(),
                     array_merge(['prefix' => '.*', 'host' => '.+'], $importedRoute->getRequirements()),
                     $importedRoute->getOptions(),
-                    (!empty($importedRoute->getHost()) ? $importedRoute->getHost() : '{host}'),
```
danrot: @wachterjohannes You have introduced that line in #2384, do you remember why? From what I can see `{host}` would have matched any route anyway, so it should be the same as if you put no host there at all. Or am I overlooking something?

alexander-schranz: I personally would prefer removing here `{host}` and set it to empty so it uses `ANY` instead of `{host}` for this routes:

Bildschirmfoto 2020-02-05 um 08 59 23

alexander-schranz: ping @wachterjohannes

danrot: That's what I have basically done here, right? :see_no_evil:

alexander-schranz: Oh sorry did look at the old code 🙈 :

Bad sadly I tested this and I think because of bc we can not remove the `host` as if we remove it it will add the one which generated routes adding the host as queryparameter e.g. `/search?host=test.com`

so I think we should go with setting the host to empty as default e.g.:

```php
            $defaults = $importedRoute->getDefaults();
            if (empty($importedRoute->getHost())) {
                $defaults['host'] = '';
            }

            $collection->add(
                $importedRouteName,
                new PortalRoute(
                    '{prefix}' . $importedRoute->getPath(),
                    $defaults,
                    array_merge(['prefix' => '.*', 'host' => '.+|^$'], $importedRoute->getRequirements()),
                    $importedRoute->getOptions(),
                    !empty($importedRoute->getHost()) ? $importedRoute->getHost() : '{host}',
                    $importedRoute->getSchemes(),
                    $importedRoute->getMethods(),
                    $condition . (!empty($importedCondition) ? ' and (' . $importedCondition . ')' : '')
                )
            );
```

What do you think?

alexander-schranz: @danrot ok that doesn't work as expected maybe you  have an idea I'm trying the following:

```yaml
# config/services.xml
parameters:
    router.request_context.host: 'example.org:8181'
```

```php
<?php

// src/Command/RouteTestCommand.php

namespace App\Command;

use Symfony\Component\Console\Command\Command;
use Symfony\Component\Console\Input\InputInterface;
use Symfony\Component\Console\Output\OutputInterface;
use Symfony\Component\Console\Style\SymfonyStyle;
use Symfony\Component\Routing\RouterInterface;

class RouteTestCommand extends Command
{
    /**
     * @var RouterInterface
     */
    private $router;

    /**
     * RouteTestCommand constructor.
     * @param RouterInterface $router
     */
    public function __construct(RouterInterface $router)
    {
        $this->router = $router;

        parent::__construct('app:test:route');
    }

    protected function execute(InputInterface $input, OutputInterface $output)
    {
        $routeOnlyPrefix = $this->router->generate(
            'sulu_search.website_search',
            [
                'prefix' => '/en',
            ],
            0
        );
        $expectedRouteOnlyPrefix = 'http://example.org:8181/en/search';

        $routeWithCustomHost = $this->router->generate(
            'sulu_search.website_search',
            [
                'prefix' => '/en',
                'host' => 'mytestdomain.com',
            ],
            0
        );
        $expectedRouteWithCustomHost = 'http://mytestdomain.com/en/search';

        $symfonyRoute = $this->router->generate('sulu_website.sitemap_index', [], 0);
        $expectedSymfonyRoute = 'http://example.org:8181/sitemap.xml';

        $data = [
            [
                'generatedRoute' => $routeOnlyPrefix,
                'expectedRoute' => $expectedRouteOnlyPrefix,
                'match' => $routeOnlyPrefix === $expectedRouteOnlyPrefix,
            ],
            [
                'generatedRoute' => $routeWithCustomHost,
                'expectedRoute' => $expectedRouteWithCustomHost,
                'match' => $routeWithCustomHost === $expectedRouteWithCustomHost,
            ],
            [
                'generatedRoute' => $symfonyRoute,
                'expectedRoute' => $expectedSymfonyRoute,
                'match' => $symfonyRoute === $expectedSymfonyRoute,
            ],
        ];

        $ui = new SymfonyStyle($input, $output);

        $ui->table(array_keys($data[0]), $data);
    }
}
```

Can not get all cases working as expected:

Bildschirmfoto 2020-02-05 um 11 05 18

Bildschirmfoto 2020-02-05 um 11 05 53

alexander-schranz: @danrot but If I think about it if you have use `{host}` in dev env and used as recommended the real domain in prod the same would had happened on prod. So maybe we should use your change and never set the `{host}` here.

src/Sulu/Bundle/WebsiteBundle/Routing/PortalLoader.php
```diff
@@ -47,7 +47,7 @@ public function load($resource, $type = null)
                     $importedRoute->getDefaults(),
                     array_merge(['prefix' => '.*', 'host' => '.+'], $importedRoute->getRequirements()),
```
alexander-schranz: as we not longer use `{host}` here we should remove also this requirement or?

alexander-schranz:
@danrot what do you think about if you also get the scheme correctly from the `router_context` this would need some changes in the interfaces e.g.:

```diff
diff --git a/src/Sulu/Bundle/CoreBundle/Resources/config/webspace.xml b/src/Sulu/Bundle/CoreBundle/Resources/config/webspace.xml
index d6e9f32cec..8209fb92d7 100644
--- a/src/Sulu/Bundle/CoreBundle/Resources/config/webspace.xml
+++ b/src/Sulu/Bundle/CoreBundle/Resources/config/webspace.xml
@@ -42,6 +42,7 @@
             </argument>
             <argument>%kernel.environment%</argument>
             <argument>%router.request_context.host%</argument>
+            <argument>%router.request_context.scheme%</argument>
 
             <tag name="sulu.localization_provider"/>
         </service>
diff --git a/src/Sulu/Component/Webspace/Manager/WebspaceManager.php b/src/Sulu/Component/Webspace/Manager/WebspaceManager.php
index 3132a24048..6a10b5eaed 100644
--- a/src/Sulu/Component/Webspace/Manager/WebspaceManager.php
+++ b/src/Sulu/Component/Webspace/Manager/WebspaceManager.php
@@ -63,13 +63,19 @@ class WebspaceManager implements WebspaceManagerInterface
      */
     private $defaultHost;
 
+    /**
+     * @var string
+     */
+    private $defaultScheme;
+
     public function __construct(
         LoaderInterface $loader,
         ReplacerInterface $urlReplacer,
         RequestStack $requestStack,
         array $options,
         string $environment,
-        string $defaultHost
+        string $defaultHost,
+        string $defaultScheme
     ) {
         $this->loader = $loader;
         $this->urlReplacer = $urlReplacer;
@@ -77,6 +83,7 @@ class WebspaceManager implements WebspaceManagerInterface
         $this->setOptions($options);
         $this->environment = $environment;
         $this->defaultHost = $defaultHost;
+        $this->defaultScheme = $defaultScheme;
     }
 
     public function findWebspaceByKey(?string $key): ?Webspace
@@ -187,7 +194,7 @@ class WebspaceManager implements WebspaceManagerInterface
         string $languageCode,
         ?string $webspaceKey = null,
         ?string $domain = null,
-        string $scheme = 'http'
+        ?string $scheme = null
     ): array {
         if (null === $environment) {
             $environment = $this->environment;
@@ -216,7 +223,7 @@ class WebspaceManager implements WebspaceManagerInterface
         string $languageCode,
         ?string $webspaceKey = null,
         ?string $domain = null,
-        string $scheme = 'http'
+        ?string $scheme = null
     ): ?string {
         if (null === $environment) {
             $environment = $this->environment;
@@ -462,8 +469,12 @@ class WebspaceManager implements WebspaceManagerInterface
      *
      * @return string
      */
-    private function createResourceLocatorUrl($scheme, $portalUrl, $resourceLocator, $domain = null)
+    private function createResourceLocatorUrl(?string $scheme, $portalUrl, $resourceLocator, $domain = null)
     {
+        if (!$scheme) {
+            $scheme = $this->defaultScheme;
+        }
+
         if (false !== strpos($portalUrl, '/')) {
             // trim slash when resourceLocator is not domain root
             $resourceLocator = rtrim($resourceLocator, '/');
diff --git a/src/Sulu/Component/Webspace/Manager/WebspaceManagerInterface.php b/src/Sulu/Component/Webspace/Manager/WebspaceManagerInterface.php
index a2543c230c..728a2cdfbf 100644
--- a/src/Sulu/Component/Webspace/Manager/WebspaceManagerInterface.php
+++ b/src/Sulu/Component/Webspace/Manager/WebspaceManagerInterface.php
@@ -87,7 +87,7 @@ interface WebspaceManagerInterface extends LocalizationProviderInterface
         string $languageCode,
         ?string $webspaceKey = null,
         ?string $domain = null,
-        string $scheme = 'http'
+        ?string $scheme = 'http'
     ): ?string;
 
     /**
```

Because currently you are forced to set the scheme when using `sulu_content_path` in the CLI.

danrot:
Like the idea with the `scheme`, but I would say that is not related to this PR, and we should solve that in a separate one.

danrot:
I closed #5020 for you :wink:

PR #5026, opened 2020-02-02
Add 'id' as a reserved property

### LINKED ISSUE #2147: Can't save page when block has property with name "type"
| Q | A |
| --- | --- |
| Bug? | yes |
| New Feature? | no |
| Sulu Version | 1.1.10 |
| Browser Version | Chrome |
#### Actual Behavior

It's not possible to save the page
#### Expected Behavior

It should be possible to save the page
#### Steps to Reproduce

Create a template with block-property and add a property with name "type":

```
<block name="block"
       default-type="default"
       minOccurs="1"
       maxOccurs="10">
    <meta>
        <title lang="de">Block</title>
        <title lang="en">Block</title>
    </meta>

    <types>
        <type name="default">
            <meta>
                <title lang="de">Block Typ Standard</title>
                <title lang="en">Block Type Default</title>
            </meta>

            <properties>
                <property name="type" type="text_line">
                    <meta>
                        <title lang="de">Typ</title>
                        <title lang="en">Type</title>
                    </meta>
                </property>
            </properties>
        </type>
    </types>
</block>
```

### LINKED ISSUE #4100: Reserved keywords in page templates
| Q | A
| --- | ---
| Bug? | yes
| New Feature? | no
| Sulu Version | 1.6.13

#### Actual Behavior

When using reserved keywords as property names in a page template, problems occur when trying to save pages.

#### Example

Using the reserved keyword `id` as a property name results in errors when trying to create a new page:

`Could not find document with ID or path "xyz"`
`RepositoryException: Not an absolute path 'xyz'`

#### Steps to Reproduce

- Create new page template in a webspace
- Create a property with a reserved keyword as name
```
<property name="id" type="text_line">
    <meta>
        <title lang="en">ID</title>
    </meta>
</property>
```
- Create a new page in the webspace
- Save or Publish the page

eekes:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #4100 and fixes #2147
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

Adds 'id' and 'type' to the array of reserved properties

#### Why?

Using 'id' or 'type' as a property name in a page template results in errors when trying to create/edit pages

### DIFF
--- a/src/Sulu/Component/Content/Metadata/Loader/XmlLegacyLoader.php
+++ b/src/Sulu/Component/Content/Metadata/Loader/XmlLegacyLoader.php
@@ -72,6 +72,8 @@ class XmlLegacyLoader implements LoaderInterface
         'shadow-base',
         'author',
         'authored',
+        'type',
+        'id',
     ];
 
     /**


danrot:
First of all, thank you for the contribution!

But would you mind backporting this to the `release/1.6` branch? I think the exact same fix is also valid there, and it will then be merged back to the other branches anyway.

Generally speaking, the `master` branch will be the next minor release (currently 2.1), and the `release/1.6` and `release/2.0` branch include everything that will make it into the next corresponding patch release.

danrot:
I just also discovered #2147, which is also related... I think that `type` is also something that we should add to this list.

eekes:
@danrot should be OK now

src/Sulu/Component/Content/Metadata/Loader/XmlLegacyLoader.php
```diff
@@ -72,6 +72,8 @@ class XmlLegacyLoader implements LoaderInterface
         'shadow-base',
         'author',
         'authored',
+        'type',
+        'id'
```
danrot: We also put a comma after the last array item :slightly_smiling_face: That will reduce that diff size if we add more items later.

danrot:
@eekes Thank you very much for the contribution!

PR #5032, opened 2020-02-04
Fixing TwigException parameter types

kfrohwein:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?
Currently TwigException uses
https://github.com/sulu/sulu/blob/ade35fff4cb8ced777c726dcb730c5c5f5720960/src/Sulu/Bundle/PreviewBundle/Preview/Exception/TwigException.php#L22
https://github.com/sulu/sulu/blob/ade35fff4cb8ced777c726dcb730c5c5f5720960/src/Sulu/Bundle/PreviewBundle/Preview/Exception/TwigException.php#L23

The parent uses:
https://github.com/sulu/sulu/blob/ade35fff4cb8ced777c726dcb730c5c5f5720960/src/Sulu/Bundle/PreviewBundle/Preview/Exception/PreviewRendererException.php#L44
https://github.com/sulu/sulu/blob/ade35fff4cb8ced777c726dcb730c5c5f5720960/src/Sulu/Bundle/PreviewBundle/Preview/Exception/PreviewRendererException.php#L45

And I matched those types.

#### Why?
Just wrong PHPDoc types and PHPStorm pointed that out.

#### BC Breaks/Deprecations

None

#### To Do

Nothing

### DIFF
--- a/src/Sulu/Bundle/PreviewBundle/Preview/Exception/TwigException.php
+++ b/src/Sulu/Bundle/PreviewBundle/Preview/Exception/TwigException.php
@@ -19,8 +19,8 @@ use Twig\Error\Error;
 class TwigException extends PreviewRendererException
 {
     /**
-     * @param int $object
-     * @param mixed $id
+     * @param mixed $object
+     * @param int|string $id
      * @param string $webspaceKey
      * @param string $locale
      */
--- a/src/Sulu/Bundle/PreviewBundle/Preview/Exception/PreviewRendererException.php
+++ b/src/Sulu/Bundle/PreviewBundle/Preview/Exception/PreviewRendererException.php
@@ -24,7 +24,7 @@ abstract class PreviewRendererException extends PreviewException
     private $object;
 
     /**
-     * @var string
+     * @var int|string
      */
     private $id;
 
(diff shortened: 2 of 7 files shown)


src/Sulu/Bundle/PreviewBundle/Preview/Exception/TwigException.php
```diff
@@ -19,8 +19,8 @@
 class TwigException extends PreviewRendererException
 {
     /**
-     * @param int $object
-     * @param mixed $id
+     * @param mixed $object
+     * @param string $id
```
alexander-schranz: I think the id is depending on the object behind and so can be `int|string`

kfrohwein: I always choose the wrong branch :sweat_smile: 

If I change that in the TwigException I should change the PreviewRendererException as well. I just made the PHPDoc match but didn't look at it what could be really inserted.

alexander-schranz: Yes think the `PreviewRendererException` also should accept id as string|int.

The branch need to be rebased that it can be merged to release/2.0

kfrohwein: The force was strong with me and I rebased successfully.

And I checked all exceptions that they now use the same int|string $id and mixed $object.

Issue #4358, opened 2019-01-10
SmartContent - numeric Tags not working


patrickpotthoff:
| Q | A
| --- | ---
| Bug? | yes
| New Feature? | no
| Sulu Version | 1.6.23
| Browser Version | All typical Browser

#### Actual Behavior

Simple Numeric tags wont get saved inside of the filter options.
Backend:
In the first step, the tag get saved and the correct items will show up. After changing the site and returning back to this site, the tag isnt set anymore.

Frontend:
The filter doesnt work correctly.

danrot:
I just tested it, and I think it is not removed, but the number is interpreted as ID, and tries to load the tag with that ID...

dduenker:
Maybe the documentation of smart content and tags could be updated to include this information. This behaviour is very confusing (still present in 2.6.9).

PR #5035, opened 2020-02-05
Allow user to filter lists

### LINKED ISSUE #4855: Sulu 2 admin filters
| Q | A
| --- | ---
| Bug? | no
| New Feature? | yes
| Sulu Version | 2.0

Support filters components in the admin datagrids.

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #4855
| Related issues/PRs | ---
| License | MIT

#### What's in this PR?

This PR adds a new feature to filter list for values in certain columns.

#### Why?

Because we need that in quite a few projects.

#### Example Usage

~~~xml

~~~

#### BC Breaks/Deprecations

- The `ListBuilderInterface` has  a new `filter` method

#### To Do

- [x] Check if filters also work if they are not part of the selected fields
- [ ] Check if sortFields really need to be iterated over in `findByIdsGivenCriteria` method, because it seems to be done another time in `assignSortFields`
- [ ] Check if `tagIds` e.g. in `contact.xml`, `webspaceKeys` in `target_groups.xml` or `tagIds` in `accounts.xml` list configuration is possible. It is quite hard, because the `ListBuilder` builds a `GROUP_CONCAT` based on that field. The `Selection` filter type however would still work for 1:n relationships, but this one is n:m...
- [ ] Should filter work on trees? (mainly pages and categories)
- [ ] How to handle PHPCR content? (pages, snippets, custom-urls, analytics)
- [x] Check if some filters should be removed because they don't make any sense
- [x] `MultiAutoComplete` should take `SelectionStore` as argument to avoid multiple requests in `SelectionFieldFilterType`

Functional todos:

- [x] Design of `Chip` component
- [x] Where to place filter chips
- [x] Positioning of buttons (search, column options and filter)
- [x] Use auto-complete instead of `MultiSelection`
- [ ] Search on enter in text field and dates (?)
- [x] Show value in chip
- [x] Don't show value in chip before input has blurred, OK button has been pressed or whatever
- [x] Dropdown for `media` (`type`, `mimeType`)?
- [x] Return promise in `getValueNode` function to allow resolving of IDs for `SelectionFieldFilterType`?
- [x] How to get old value (e.g. to merge object in `DateFieldFilterType` and `NumberFieldFilterType`
- [ ] Partial and exact match for `TextFilterType`
- [x] Add filter information to URL in order to remember what has been selected after back button click
- [x] Double requests when date filters are used
- [x] Search button should also be left if no filters are available
- [x] Excerpt returns tag ids as strings, causing a reaction to execute and trigger the save button immediately after opening the form
- [x] SmartContent does not filter correctly when tags have already been stored before (sends the tag itself instead of the id in the items request for the form) (related to #4358?)


lightglitch:
Its looking good. It's suppose so solve https://github.com/sulu/sulu/issues/4855?

danrot:
@lightglitch Thank you for reminding me of that issue, yeah, this will be solved by this :slightly_smiling_face:

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/fieldFilterTypes/SelectionFieldFilterType.js
```diff
@@ -0,0 +1,56 @@
...
+            throw new Error('The "SelectionFieldFilterType" needs some parameters to work!');
+        }
+
+        const {displayProperty, listAdapter, resourceKey} = parameters;
+
+        if (typeof displayProperty !== 'string') {
+            throw new Error('The "displayProperty" parameter must be a string!');
+        }
+
+        if (typeof listAdapter !== 'string') {
```
danrot: Not needed anymore.

chirimoya:
@danrot Nice Job! Personally, I would prefer one-liners and a real Button. Also depends on the decision regarding the label. 

- [ ] Store filter for user session
- [ ] Contact Birthday filter didn't work
- [ ] Contact Organization filter didn't work
- [ ] Media Type filter didn't work
- [ ] Missing filters for Snippets & Categories
- [ ] Date format changes... see screen shot

Bildschirmfoto 2020-02-27 um 15 21 07

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/FieldFilter.js
```diff
@@ -0,0 +1,148 @@
...
+
+        return (
+            <div className={fieldFilterStyles.fieldFilter}>
+                {Object.keys(fields).length > 0 &&
+                    <Fragment>
+                        <ArrowMenu
+                            anchorElement={
+                                <div className={fieldFilterStyles.filterButton}>
+                                    <Button
+                                        icon="fa-filter"
```
danrot: Should be "su-filter"

src/Sulu/Bundle/TagBundle/Resources/config/lists/tags.xml
```diff
@@ -35,7 +35,6 @@
             <entity-name>%sulu.model.tag.class%</entity-name>
         </property>
 
-        <!-- TODO: Check if editable is needed in 2.0 -->
```
nnatter: :joy

src/Sulu/Bundle/CoreBundle/Resources/config/rest.xml
```diff
@@ -21,10 +21,63 @@
...
+        <service
+            id="Sulu\Component\Rest\ListBuilder\Doctrine\DoctrineListBuilderFactoryInterface"
+            alias="sulu_core.doctrine_list_builder_factory"
+        />
+
+        <service
+            id="sulu_core.list_builder_filter_type_registry"
+            class="Sulu\Component\Rest\ListBuilder\Filter\FilterTypeRegistry"
+        >
+            <argument type="tagged" tag="sulu_core.list_builder_filter_type" index-by="alias" />
```
nnatter: would prefer if we would start using `default_index_method` with a static method on the class. this will make things easier for the developer because he can just autowiring in a project

src/Sulu/Component/Rest/ListBuilder/AbstractListBuilder.php
```diff
@@ -204,6 +215,19 @@ public function addSearchField(FieldDescriptorInterface $fieldDescriptor)
...
         $this->search = $search;
+
+        return $this;
+    }
+
+    public function filter($filters)
+    {
+        foreach ($filters as $fieldName => $options) {
+            $fieldDescriptor = $this->fieldDescriptors[$fieldName];
+            $this->filterTypeRegistry->getFilterType($fieldDescriptor->getMetadata()->getFilterType())
```
nnatter: whould we throw an explicit exception if the `getFilterType()` method returns null?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/tests/List.test.js
```diff
@@ -387,6 +397,46 @@ test('Render the adapter in disabled state', () => {
     ).toMatchSnapshot();
 });
 
+test('Render the adapter with filters', () => {
```
nnatter: `with filterable fields`?

src/Sulu/Component/Rest/ListBuilder/AbstractListBuilder.php
```diff
@@ -204,6 +215,19 @@ public function addSearchField(FieldDescriptorInterface $fieldDescriptor)
     public function search($search)
     {
         $this->search = $search;
+
+        return $this;
+    }
+
+    public function filter($filters)
```
nnatter: you are using `$filter` instead of `$filters` as plural in most other places

src/Sulu/Component/Rest/ListBuilder/ListBuilderInterface.php
```diff
@@ -108,6 +108,13 @@ public function addSearchField(FieldDescriptorInterface $fieldDescriptor);
      */
     public function search($search);
 
+    /**
+     * @param array $filters
+     *
+     * @return ListBuilderInterface
+     */
+    public function filter($filters);
```
nnatter: you are not using `$filters` as plural in other places

src/Sulu/Component/Rest/ListBuilder/ListRestHelper.php
```diff
@@ -168,4 +168,12 @@ public function getSearchFields()
 
         return (null != $searchFields) ? explode(',', $searchFields) : [];
     }
+
+    /**
+     * @return array
+     */
+    public function getFilters()
```
nnatter: you are not using `filters` as plural in other places

src/Sulu/Component/Rest/ListBuilder/ListRestHelperInterface.php
```diff
@@ -84,4 +84,9 @@ public function getSearchPattern();
      * @return array|null
      */
     public function getSearchFields();
+
+    /**
+     * @return array|null
+     */
+    public function getFilters();
```
nnatter: you are not using `filters` as plural in other places

src/Sulu/Component/Rest/ListBuilder/ListRestHelper.php
```diff
@@ -168,4 +168,12 @@ public function getSearchFields()
...
+    {
+        return $this->getRequest()->get('filter', []);
```
nnatter: event the parameter is called `filter` without a `s` in the end 🙂

src/Sulu/Component/Rest/ListBuilder/Filter/DateTimeFilterType.php
```diff
@@ -0,0 +1,32 @@
...
+use Sulu\Component\Rest\ListBuilder\ListBuilderInterface;
+
+class DateTimeFilterType implements FilterTypeInterface
+{
+    public function filter(
+        ListBuilderInterface $listBuilder,
+        FieldDescriptorInterface $fieldDescriptor,
+        $options
+    ): void {
+        if (!is_array($options) || !isset($options['from']) || !isset($options['to'])) {
```
nnatter: shouldnt it be possible to only set `from` or `to`?

src/Sulu/Component/Rest/ListBuilder/Filter/DropdownFilterType.php
```diff
@@ -0,0 +1,32 @@
...
+        FieldDescriptorInterface $fieldDescriptor,
+        $options
+    ): void {
+        if (!is_string($options)) {
+            throw new InvalidFilterTypeOptionsException(
+                'The DropdownFilterType requires its options to be comma-separated list of values'
+            );
+        }
+
+        $listBuilder->in($fieldDescriptor, explode(',', $options));
```
nnatter: i think this will break if `$options`is an empty string

src/Sulu/Component/Rest/ListBuilder/Filter/SelectionFilterType.php
```diff
@@ -0,0 +1,32 @@
...
+        FieldDescriptorInterface $fieldDescriptor,
+        $options
+    ): void {
+        if (!is_string($options)) {
+            throw new InvalidFilterTypeOptionsException(
+                'The SelectionFilterType requires its options to be comma-separated list of IDs'
+            );
+        }
+
+        $listBuilder->in($fieldDescriptor, explode(',', $options));
```
nnatter: i think this will break if $optionsis an empty string

src/Sulu/Bundle/AdminBundle/Resources/js/components/Chip/chip.scss
```diff
@@ -19,6 +19,10 @@ $chipDisabledBackgroundColor: $silver;
     .icon {
         margin-left: 10px;
     }
+
+    > * {
```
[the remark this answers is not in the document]
danrot: It is required because a `Loader` is passed to it right next to a text in the [`FieldFilterItem`](https://github.com/sulu/sulu/pull/5035/files#diff-1c4849a1f68379b37193d70c99797491R122). The `Loader` is a block element, so if I don't set this for it, it will cause the loader to appear in an own row.

IMO it is ok this way, because I think the Chip should never be stretched over multiple lines, and this statement helps with that.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/FieldFilterItem.js
```diff
@@ -0,0 +1,142 @@
...
+                this.setValueNodeLoading(true);
+                valueNodePromise.then(action((valueNode) => {
+                    this.setValueNodeLoading(false);
+                    this.setValueNode(valueNode);
+                }));
+            }
+        });
+    }
+
+    @computed get propValue() {
```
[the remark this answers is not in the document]
danrot: The value passed via props is the one that currently is applied as a filter. I just need the computed getter, because otherwise I was having some mobx updating issues. I think `propValue` tells pretty good what it is. The internal value is required because the value is not passed to the outside, until the user clicks the "OK" button in the `ArrowMenu`. I don't really like `unconfirmedValue`, because the value might be confirmed, if it is not changed yet. However I could live with `displayedValue` instead of `this.value`, but I can't think of a better name for `propValue` right now. Although I am still open for suggestions :slightly_smiling_face:

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/List.js
```diff
@@ -502,12 +518,17 @@ class List extends React.Component<Props> {
...
             <div className={listStyles.listContainer}>
                 {header}
-                {!store.schemaLoading && (searchable || this.currentAdapter.hasColumnOptions || adapters.length > 1) &&
+                {!schemaLoading && (searchable || adapters.length > 1 || filterable || hasColumnOptions) &&
                     <div className={toolbarClass}>
                         {searchable &&
                             <Search onSearch={this.handleSearch} value={store.searchTerm.get()} />
                         }
-                        {this.currentAdapter.hasColumnOptions &&
+                        <FieldFilter
```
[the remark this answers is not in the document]
danrot: It will, and that's on purpose, because the filters will take all the space between the search and column options button. If we remove it, if no filters are applied, the search button will move the right, causing an inconsistency in the UI.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/fieldFilterTypes/NumberFieldFilterType.js
```diff
@@ -0,0 +1,92 @@
...
+import numberFieldFilterTypeStyles from './numberFieldFilterType.scss';
+
+const operatorMapping = {
+    lt: '<',
+    eq: '=',
+    gt: '>',
+};
+
+function getOperatorFromValue(value: ?{[string]: ?number}) {
+    const valueKeys = value ? Object.keys(value) : {};
```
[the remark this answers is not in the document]
danrot: Good catch :slightly_smiling_face:

src/Sulu/Bundle/TagBundle/Resources/config/lists/tags.xml
danrot: All time classic :joy:

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/FieldFilter.js
```diff
@@ -0,0 +1,148 @@
...
+
+type Props = {|
+    fields: Schema,
+    onChange: ({[string]: mixed}) => void,
+    value: {[string]: mixed},
+|};
+
+@observer
+class FieldFilter extends React.Component<Props> {
+    @observable fieldFilterTypes: {[column: string]: AbstractFieldFilterTypes<*>} = {};
```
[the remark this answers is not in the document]
danrot: Seems like it :+1:

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/FieldFilter.js
```diff
@@ -0,0 +1,148 @@
...
+    handleFilterItemChange = (column: string, columnValue: mixed) => {
+        const {onChange, value} = this.props;
+        onChange({...value, [column]: columnValue});
+        this.closeFilterItem();
+    };
+
+    @action handleFilterItemDelete = (column: string) => {
+        const {onChange, value} = this.props;
+
+        onChange(
```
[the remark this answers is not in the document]
danrot: That's sick :see_no_evil: Works in jest as well, and I guess babel will compile it anyway :slightly_smiling_face:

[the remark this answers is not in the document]
danrot: Just tested it, also works in firefox. Can you test the anti-browser? I have the wrong operating system for it :stuck_out_tongue:

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/FieldFilter.js
```diff
@@ -0,0 +1,148 @@
...
+    onChange: ({[string]: mixed}) => void,
+    value: {[string]: mixed},
+|};
+
+@observer
+class FieldFilter extends React.Component<Props> {
+    @observable fieldFilterTypes: {[column: string]: AbstractFieldFilterTypes<*>} = {};
+    @observable filterOpen: boolean = false;
+    @observable filterChipOpen: ?string = undefined;
+    @observable value: {[string]: mixed} = {};
```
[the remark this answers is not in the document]
danrot: Seems like it, removed it.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/FieldFilter.js
```diff
@@ -0,0 +1,148 @@
...
+        );
+    };
+
+    render() {
+        const {fields} = this.props;
+
+        return (
+            <div className={fieldFilterStyles.fieldFilter}>
+                {Object.keys(fields).length > 0 &&
+                    <Fragment>
```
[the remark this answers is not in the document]
danrot: It probably wasn't at some point :see_no_evil: Removed it.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/FieldFilterItem.js
```diff
@@ -0,0 +1,142 @@
...
+
+    @action setValueNodeLoading(valueNodeLoading: boolean) {
+        this.valueNodeLoading = valueNodeLoading;
+    }
+
+    @action setValueNode(valueNode: ?Node) {
+        this.valueNode = valueNode;
+    }
+
+    @action handleFilterFieldChange = (value: mixed) => {
```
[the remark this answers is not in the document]
danrot: Probably because I am already working too long on this task :see_no_evil:

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/fieldFilterTypes/DropdownFieldFilterType.js
```diff
@@ -0,0 +1,51 @@
...
+            </MultiSelect>
+        );
+    }
+
+    getValueNode(values: ?Array<string>): ?Promise<string> {
+        if (!values) {
+            return null;
+        }
+
+        return Promise.resolve(values.map((value) => translate(this.options[value])).join(', '));
```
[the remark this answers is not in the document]
danrot: Above I am also mapping a different variable.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/fieldFilterTypes/NumberFieldFilterType.js
```diff
@@ -0,0 +1,92 @@
...
+
+    if (valueKeys.length > 1) {
+        throw new Error('The "NumberFilterFieldType" only accepts an array with exactly one key!');
+    }
+
+    return valueKeys[0];
```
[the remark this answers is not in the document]
danrot: If `valueKeys` is empty `undefined` is returned, as I would expect.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/fieldFilterTypes/SelectionFieldFilterType.js
```diff
@@ -0,0 +1,125 @@
+// @flow
+import React from 'react';
+import {autorun, computed, toJS, untracked, when} from 'mobx';
+import equals from 'fast-deep-equal';
+import MultiSelectionStore from '../../../stores/MultiSelectionStore';
+import MultiAutoComplete from '../../MultiAutoComplete';
+import AbstractFieldFilterType from './AbstractFieldFilterType';
+import selectionFieldFilterTypeStyles from './selectionFieldFilterType.scss';
+
+class SelectionFieldFilterType extends AbstractFieldFilterType<?Array<string | number>> {
```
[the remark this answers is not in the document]
danrot: Custom entity should not be a big deal, because you can pass the `resourceKey` in the list configuration.

Actually Thomas and I agreed that there is only one version, not having multiple ways of assigning stuff as in the `Selection` field type for forms, therefore I implemented it this way.

And even if we would build something similar, I would not use the `Selection` field type for it. We would have to translate all the properties in some way, also properties which are not type checked... Sounds to me like this could easily fail in some way.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/stores/ListStore.js
```diff
@@ -263,9 +322,26 @@ export default class ListStore {
         ListStore.setSchemaSetting(this.listKey, this.userSettingsKey, schemaSettings);
     };
 
+    @computed get filterableFields(): ?Schema {
```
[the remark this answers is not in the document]
danrot: That would make the types a lot more complex IMO... Do you really think it is worth it?

src/Sulu/Bundle/CoreBundle/Resources/config/rest.xml
danrot: Would do it, if I could get it working :see_no_evil: Should it work if I just omit the `index-by` definition here, and add a `public static function getDefaultIndexName` to the `FilterType`s?

src/Sulu/Component/Rest/ListBuilder/AbstractListBuilder.php
danrot: Yes, `getFilterType` will throw an exception

src/Sulu/Component/Rest/ListBuilder/Filter/DateTimeFilterType.php
danrot: Maybe, didn't implement it for now, in order to get this finished :see_no_evil:

src/Sulu/Component/Rest/ListBuilder/Filter/DropdownFilterType.php
danrot: It does not break, it just returns nothing :see_no_evil: Since somebody would to actually build send the request manually for this to happen, I would accept this for now.

src/Sulu/Component/Rest/ListBuilder/Filter/SelectionFilterType.php
danrot: It does not break, it just returns nothing see_no_evil Since somebody would to actually build send the request manually for this to happen, I would accept this for now.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/tests/List.test.js
danrot: Have overlooked that one, gonna fix it when doing the other missing changes, after you have commented again.

danrot: Actually I think the name is correct... it does not only render with some filterable fields, it renders with some filters really applied

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/FieldFilter.js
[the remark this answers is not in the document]
nnatter: works in safari too 🙂

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/FieldFilterItem.js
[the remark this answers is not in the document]
nnatter: 😂

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/stores/ListStore.js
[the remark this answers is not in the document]
nnatter: not sure 🙂 its okay for me this way for now

src/Sulu/Bundle/CoreBundle/Resources/config/rest.xml
nnatter: no, you need to define the `default_index_method` as attribute of this argument 🙂 see last secion of https://symfony.com/blog/new-in-symfony-4-3-indexed-and-tagged-service-collections

src/Sulu/Component/Rest/ListBuilder/Filter/DateTimeFilterType.php
nnatter: okay, shouldnt be hard to implement this later on

src/Sulu/Bundle/CoreBundle/Resources/config/rest.xml
danrot: If I am reading the article correctly, this attribute is optional, and only necessary if you want to use something different than `getDefaultIndexName`. However, even when I've tried to add this, the XSD schema complained about it, because this attribute does not seem to exist there :confused:

danrot: Although, maybe I was using underscores instead of dashes... gonna try it again.

nnatter: okay looks like you still ned a `index-by` attribute: https://symfony.com/doc/4.4/service_container/service_subscribers_locators.html#indexing-the-collection-of-services

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/FieldFilterItem.js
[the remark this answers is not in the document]
nnatter: we dio use this concept of an internal `this.value` and an external `this.props.value` in other places right?

you could also use a `reaction` instead of an `autorun`, then you would not need to computed getter.

src/Sulu/Bundle/CoreBundle/Resources/config/rest.xml
danrot: Okay, the problem was something different... This only works, if you keep the `index-by` attribute as well...

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/List.js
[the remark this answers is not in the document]
nnatter: okay it does not feel intuitive to me that the `List` container relies on the `FieldFilter` component to expand. especially because the `FieldFilter` component has no internal/own reason for doing this.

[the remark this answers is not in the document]
danrot: Well, internally the `FieldFilter` component behaves exactly the same, and does not take care of this. The component that needs this, is the `List` container, therefore I think it is find the way it currently is.

The alternative that comes to my mind is to have a more complex DOM setup, instead of just relying on flex-box. But I like that option less than the current one, since I prefer to rely on CSS as much as possible, in order to keep the markup simple :slightly_smiling_face:

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/FieldFilterItem.js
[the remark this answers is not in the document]
danrot: It's just that I like reactions a lot less than `autorun`s :see_no_evil: At first, they are not executed initially by default, and I have to repeat stuff, although the `autorun` is capable of finding that out on its own.

In addition to that I think that writing `this.props.value` or `this.propValue` makes not a big difference in understanding code, if it makes a difference at all. Then I would prefer to call this getter `value` and rename the current observable to `displayedValue`.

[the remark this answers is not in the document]
nnatter: we are not on the same page here, i like that a reaction explicitly specifies what it reacts to. the fact that the `autorun` is executed if any of the accessed observables is changed scares me 🙂 
imo it is really easy to introduce performance problems (that are hard to spot) by accessing an observable in a method that is coincidentally called in some `autorun` (maybe even in another class).

i would not call the getter `value` because it is inconsistent with the other components that use an internal value (eg `SmartContent`). so i think we will just leave it like this 🙂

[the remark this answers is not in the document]
danrot: The whole point of MobX is to avoid to explicitly specify what it has to react to, but to do it automatically instead. That's exactly why it is so great! If I want to specify every variable something reacts to, using MobX makes no sense IMO. If I change another variable that has an effect on this `autorun` I want the `autorun` to be executed in case it changes, since the result might be a different one. If I use a `reaction` instead, the possibility that I am introducing an outdated error, because I forgot a value to add to the `reaction`, is very high.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/List.js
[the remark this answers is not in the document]
nnatter: the `fieldFilter.scss` contains `flex-grow: 1;` even though there is no reason for this in the `FieldFilter.js`. the only reason the `flex-grow: 1;` is there is because the `List` container (which the `FieldFilter` should not know about) needs it. 

i dont think this is intuitive and would probably remove it when i am refactoring something there. furthermore, there is not test case that would detect this as removing the style would not change anything visible in our snapshot test.

but i understand that you dont want to make the dom more complex just because of this. the correct solution would probably to pass an additional class prop from the `List` to the `FieldFilter`.  but thats quite some work for a single style property so i can live with the current solution too, just dont like it 😉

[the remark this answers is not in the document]
danrot: That's also an opinion I don't share :see_no_evil: The `FieldFilter` is part of the `List`, and not used anywhere else, it is even in the same Folder. Since it is part of the list, it is reasonable that it can make some assumptions about it

There are many other CSS properties that require certain stuff to be set on its parent, otherwise they won't work as expected. And if we can't rely on that within components that belong together, we are making our lives much harder than they need to be.

Apart from that non of our tests currently check for CSS changes, no matter how we structure this. Just because I see a class in the snapshot, it does not mean that it is rendering correctly. That's why I would love to introduce some screenshot tests for critical parts of the application like the `Grid`.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/FieldFilterItem.js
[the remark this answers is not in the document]
nnatter: i absolutely agree with you that this is great about mobx, but unfortunately things dont work as intuitively as they should. the whole point why you need the `propValue` getter is because you **dont** want to run the `autorun` if the `this.props` observable changes (which you would need to access if you dont hide it behind a computed getter) 😂 

somebody that does not know about this quirks will see that the `propValue` getter is used only in a single plac, inline it and cause a performance issue 🙂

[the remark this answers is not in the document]
danrot: Yeah, that is a really weird behavior IMO... Still prefer that behavior over a `reaction`, as it will show bugs (if something changes that shouldn't change I'll have a chance to find it, which might otherwise cause bugs in even weirder places) and does not introduce new ones by requiring you to think about something, that MobX could actually already take care for you.

nnatter:
i am not able to remove a boolean filter that is set to false if it was set to true before 🤔

nnatter:
The following things should probably be addressed in another pr:
* selection filter looks broken if i select more than one entity (eg two different titles in contact list)
* dropdown popover of the dropdown filter covers the **OK** button when filtering meda types. i can only close it by clicking somewhere outside of the filter chip

nnatter:
Furthermore, i would propose the following usability improvements:
* close the filter chip with ESC
* confirm the filter chip with ENTER
* focus the text field automatically when i open a filter chip

nnatter:
Apart from these small quirks, the functionality feels great and integrates beautifully into the UI. I think our users will really like this feature! 🙂 

I would propose to merge this PR if the boolean filter bug is resolved and implement the other issues in another PR. Is this okay for you? @sulu/core-team ? would be nice to get this huge thing out of the way 🙂

danrot:
So to summarize:

- [ ] Removing a boolean filter does not work
- [ ] Selection filter looks broken if i select more than one entity (eg two different titles in contact list)
- [ ] Dropdown popover of the dropdown filter covers the OK button when filtering meda types. i can only close it by clicking somewhere outside of the filter chip
- [ ] close the filter chip with ESC
- [ ] confirm the filter chip with ENTER
- [ ] focus the text field automatically when i open a filter chip

danrot:
The dropdown issue is (of course) a z-index issue... Don't know how to quickly fix that... I would love to get rid of all z-indexes all together, they always cause trouble at some point... I am checking the other issues right now, maybe I can fix one or the other before merging this PR, shouldn't be too bad.

PR #5045, opened 2020-02-10
Remove unused colors from global color definitions

nnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This PR removes colors which are not used within the whole application rom the global `colors.scss` file. Furthermore, this PR removes the `$java` color which is nearly identical to the `$shakespeare` color. 

#### Why?

The  `$java` color and the `$shakespeare` color are both used on the Dashboard page right now (One for the search button, one for the navigation button). I dont think it is a good idea to use two colors that are nearly identical on the same page.


src/Sulu/Bundle/AdminBundle/Resources/js/components/Navigation/item.scss
```diff
@@ -16,7 +16,7 @@ $activeBackgroundColor: $ebonyClayDark;
     position: relative;
 
     &:hover {
-        color: $java;
+        color: $shakespeare;
```
danrot: That should actually be an own variable as well, similar to `$activeColor`. Should we maybe even reuse `$activeColor`?

PR #5048, opened 2020-02-11
Add Tree Transliteration Generator Decorator.

Warxcell:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets |
| Related issues/PRs |
| License | MIT
| Documentation PR | sulu/sulu-docs#489

#### What's in this PR?

Add the transliterator Tree Generator Decorator.

#### Why?

When cyrillic is used in Titles/names and we want to transliterate these somehow.

#### Example Usage

See docs PR.

#### BC Breaks/Deprecations

No

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -124,7 +124,8 @@
         "symfony/var-dumper": "^4.3",
         "symfony/web-profiler-bundle": "^4.3",
         "symfony/web-server-bundle": "^4.3",
-        "handcraftedinthealps/zendsearch": "^2.0"
+        "handcraftedinthealps/zendsearch": "^2.0",
+        "ext-intl": "*"
     },
     "conflict": {
         "doctrine/dbal": "2.7.0",
--- a/src/Sulu/Component/Content/Types/ResourceLocator/Strategy/TreeTransliteratorGeneratorDecorator.php
+++ b/src/Sulu/Component/Content/Types/ResourceLocator/Strategy/TreeTransliteratorGeneratorDecorator.php
@@ -1,0 +1,40 @@
+<?php
+
+/*
+ * This file is part of Sulu.
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Component\Content\Types\ResourceLocator\Strategy;
+
+/**
+ * Uses Intl component to transliterate generated titles.
+ */
+class TreeTransliteratorGeneratorDecorator implements ResourceLocatorGeneratorInterface
+{
+    /** @var ResourceLocatorGeneratorInterface */
+    private $originalGenerator;
+
+    /** @var \Transliterator */
+    private $transliterator;
+
+    public function __construct(ResourceLocatorGeneratorInterface $originalGenerator, \Transliterator $transliterator)
+    {
+        $this->originalGenerator = $originalGenerator;
+        $this->transliterator = $transliterator;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function generate($title, $parentPath = null)
+    {
+        $generated = $this->originalGenerator->generate($title, $parentPath);
+
+        return $this->transliterator->transliterate($generated);
+    }
+}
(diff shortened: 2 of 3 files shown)


src/Sulu/Component/Content/Types/ResourceLocator/Strategy/TreeTransliteratorGeneratorDecorator.php
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function generate($title, $parentPath = null)
+    {
+        $generated = $this->originalGenerator->generate($title, $parentPath);
+
+        return $this->transliterator->transliterate($generated);
```
danrot: Where do you configure the language for the `Transliterator`? I.e. what you have set to "Russian-Latin/BGN" in the test?

composer.json
```diff
@@ -124,7 +124,8 @@
         "symfony/var-dumper": "^4.3",
         "symfony/web-profiler-bundle": "^4.3",
         "symfony/web-server-bundle": "^4.3",
-        "handcraftedinthealps/zendsearch": "^2.0"
+        "handcraftedinthealps/zendsearch": "^2.0",
+        "ext-intl": "*"
```
danrot: Wondering if we should only make that a suggested dependency... Because this might break existing systems, and some of them won't even need it...

Warxcell: @danrot but the tests will break if no ext-intl is installed. That's why I included it.

src/Sulu/Component/Content/Types/ResourceLocator/Strategy/TreeTransliteratorGeneratorDecorator.php
Warxcell: @danrot you pass already configured Transliterator to Decorator. You can see the docs PR.

Warxcell:
@danrot I think it's easier to configure/override that behaviour with Decorator. Somebody will want to implement they own behaviour.

Warxcell:
@danrot and I think that should be removed: https://github.com/sulu/sulu/blob/release/2.0/src/Sulu/Component/PHPCR/PathCleanup.php#L77

so it will allow non-latin urls

composer.json
danrot: Then you could also add it as a `dev` requirement. And isn't there another transliterator we could use, without requiring this PHP extension? Might be worth a thought, since it would not be a BC break then.

danrot:
My point is, that this solution makes it very hard for people to use it. They have to know that they have to configure it this way, otherwise you'll still have the same problem.

The thing is that the `PathReplacer` already does something similar for german (using its `$replacers` variable). So instead of that, we could probably also use a transliterator for that (e.g. in german we also have characters called umlauts like `ä`, `ö` and `ü`, which need to be replaced somehow). And I would argue that a transliterator is necessary for any language except for english, so I would more likely to go for a way where this is activated by default.

But maybe the rest of the @sulu/core-team has a different opinion on that.

Warxcell:
@danrot but how will be activated by default? With which settings? Russian/BGN? German? Korean? So many options. I doubt it should be by default

danrot:
@Warxcell I would say using the currently selected content language would make sense.

composer.json
Warxcell: @danrot it is already in `require-dev`, not `require`. I didn't searched for other transliterators, but I think PHP extensions are good enough and all other will be unefficient.

Warxcell:
@danrot that could work, then we must have some kind of mapping language to Transliterator Settings. And ability to configure it.

danrot:
When thinking more about it, we actually have to do it this way... What if you manage e.g. chinese and bulgarian content? Then you can't set this configuration a single time, but you need to have different transliterators.

Regarding the mapping: Do we really have to define that ourselves, or isn't there a way, to get that from somewhere else? The `Transliterator` documentation from PHP is not very exhaustive :confused:

Warxcell:
@danrot will do research and refactor this.

danrot:
@Warxcell Can this be closed in favour of #5057?

Warxcell:
I guess so. If the final decision is to integrate it directly, instead of this method.

PR #5049, opened 2020-02-11
Fix media copyright and description to be deletable

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Fixes the copyright and description text, so that it is possible to remove the text again.

#### Why?

Because the copyright and description text could not be removed again after it was set once.

#### Steps to Reproduce
* Set copyright or description for a media
* save it
* try to delete the text again and save the media


alexander-schranz:
@Prokyonn change looks good for me. Can you add a test?

src/Sulu/Bundle/MediaBundle/Media/Manager/MediaManager.php
```diff
@@ -526,8 +526,8 @@ protected function setDataToMedia(Media $media, $data, $user)
             if ($value ||
                 ('tags' === $attribute && null !== $value) ||
                 ('size' === $attribute && null !== $value) ||
-                ('description' === $attribute && null !== $value) ||
-                ('copyright' === $attribute && null !== $value) ||
+                ('description' === $attribute) ||
+                ('copyright' === $attribute) ||
                 ('credits' === $attribute && null !== $value) ||
```
alexander-schranz: Think is the same for credits

src/Sulu/Bundle/MediaBundle/Tests/Functional/Controller/MediaControllerTest.php
```diff
@@ -1192,6 +1192,81 @@ public function testFileVersionDeleteActiveShould400()
         $this->assertHttpStatusCode(400, $client->getResponse());
     }
 
+    public function testPutRemovingCopyright()
+    {
+        $media = $this->createMedia('image', 'de');
+
+        $fileVersionMetadata = $media->getFiles()[0]->getFileVersions()[0]->getMeta()[0];
+        $fileVersionMetadata->setCopyright('Copyright by Test');
```
alexander-schranz: give it as parameter to the createMedia so you don't flush multiple times.

src/Sulu/Bundle/MediaBundle/Tests/Functional/Controller/MediaControllerTest.php
```diff
@@ -1192,6 +1192,81 @@ public function testFileVersionDeleteActiveShould400()
...
+        $response = json_decode($client->getResponse()->getContent(), true);
+
+        $this->assertSame(null, $response['copyright']);
+    }
+
+    public function testPutRemovingDescription()
+    {
+        $media = $this->createMedia('image', 'de');
+
+        $fileVersionMetadata = $media->getFiles()[0]->getFileVersions()[0]->getMeta()[0];
```
alexander-schranz: same as above add it as parametre to the createMedia method.

src/Sulu/Bundle/MediaBundle/Tests/Functional/Controller/MediaControllerTest.php
```diff
@@ -1192,6 +1192,81 @@ public function testFileVersionDeleteActiveShould400()
...
+                'description' => null,
+            ]
+        );
+
+        $response = json_decode($client->getResponse()->getContent(), true);
+
+        $this->assertSame(null, $response['description']);
+    }
+
+    public function testPutRemovingCredits()
```
alexander-schranz: Would create one test for all fields because functional tests cost performance or what do you think @danrot.

Prokyonn:
> Still missing the `credits` field or will this be an own PR?

Added the `credits` field

src/Sulu/Bundle/MediaBundle/Media/Manager/MediaManager.php
```diff
@@ -526,9 +526,9 @@ protected function setDataToMedia(Media $media, $data, $user)
             if ($value ||
                 ('tags' === $attribute && null !== $value) ||
                 ('size' === $attribute && null !== $value) ||
-                ('description' === $attribute && null !== $value) ||
-                ('copyright' === $attribute && null !== $value) ||
-                ('credits' === $attribute && null !== $value) ||
```
danrot: Why do we only remove it for these three attributes? What about all the others?

src/Sulu/Bundle/MediaBundle/Resources/translations/admin.de.json
```diff
@@ -32,6 +32,7 @@
     "sulu_media.mime_type": "MIME-Typ",
     "sulu_media.license": "Lizenz",
     "sulu_media.copyright": "Copyright Informationen",
+    "sulu_media.credits": "Credits",
```
danrot: That's not the german translation :wink:

Prokyonn: In Sulu 1.6 it was called `Bildnachweis`, but the copyright is in 1.6 also translated to `Urheberrechtsinformationen ` should we change this translation as well ?

danrot: To what should we change it? If you have no better idea I would just take the last value from 1.6.

Issue #5050, opened 2020-02-11
Cyrillic node title / name


Warxcell:
| Q | A
| --- | ---
| Sulu Version | 2.0.3
| Browser Version | Not related to browser

# Actual Behavior

When creating new page with Cyrillic name error occurs:
`InvalidArgumentException: Name can not be empty in /var/www/vendor/jackalope/jackalope/src/Jackalope/Node.php:1622`

# Expected Behavior

Save the new page.

# Steps to Reproduce

Create new page in https://sulu.rocks/ with cyrillic "Тест на страница с кирилица"

# Possible Solutions

I don't know what is that name used for, but one of the possible is to use Transliterator which will always transliterate everything to LATIN. The other possible solution is to remove [this code](https://github.com/sulu/sulu/blob/master/src/Sulu/Bundle/DocumentManagerBundle/Slugifier/Urlizer.php#L600-L604). 

Full stack:
[...]

PR #5057, opened 2020-02-12
Fix url and path generation for text with non-latin symbols

### LINKED ISSUE #3367: Use of non-latin characters for content title causes 500 server error
| Q | A
| --- | ---
| Bug? | yes
| New Feature? | no
| Sulu Version | 1.5+, dev-develop
| Browser Version | Chrome 58.0.3029.110, Avira Scout 17.5.3029.2702

#### Actual Behavior

Even without saving 500 error will be thrown for `post_node_resourcelocator_generate` route with the following backtrace:
```
[1] PHPCR\RepositoryException: Invalid path '/cmf/jamchernihiv/routes/uk/entertainment/'
    at n/a
        in /app/vendor/phpcr/phpcr-utils/src/PHPCR/Util/PathHelper.php line 329

    at PHPCR\Util\PathHelper::error('Invalid path \'/cmf/jamchernihiv/routes/uk/entertainment/\'', true)
        in /app/vendor/phpcr/phpcr-utils/src/PHPCR/Util/PathHelper.php line 55

    at PHPCR\Util\PathHelper::assertValidAbsolutePath('/cmf/jamchernihiv/routes/uk/entertainment/', false, true)
        in /app/vendor/phpcr/phpcr-utils/src/PHPCR/Util/PathHelper.php line 170

    at PHPCR\Util\PathHelper::normalizePath('/cmf/jamchernihiv/routes/uk/entertainment/')
        in /app/vendor/jackalope/jackalope/src/Jackalope/ObjectManager.php line 203

    at Jackalope\ObjectManager->getNodeByPath('/cmf/jamchernihiv/routes/uk/entertainment/')
        in /app/vendor/jackalope/jackalope/src/Jackalope/Session.php line 360

    at Jackalope\Session->nodeExists('/cmf/jamchernihiv/routes/uk/entertainment/')
        in /app/vendor/jackalope/jackalope/src/Jackalope/Node.php line 996

    at Jackalope\Node->hasNode('entertainment/')
        in /app/vendor/sulu/sulu/src/Sulu/Component/Content/Types/ResourceLocator/Mapper/PhpcrMapper.php line 389

    at Sulu\Component\Content\Types\ResourceLocator\Mapper\PhpcrMapper->isUnique(object(Node), '/entertainment/')
        in /app/vendor/sulu/sulu/src/Sulu/Component/Content/Types/ResourceLocator/Mapper/PhpcrMapper.php line 324

    at Sulu\Component\Content\Types\ResourceLocator\Mapper\PhpcrMapper->getUniquePath('/entertainment/', 'jamchernihiv', 'uk', null)
        in /app/vendor/sulu/sulu/src/Sulu/Component/Content/Types/ResourceLocator/Strategy/ResourceLocatorStrategy.php line 119

    at Sulu\Component\Content\Types\ResourceLocator\Strategy\ResourceLocatorStrategy->generate('Не-латиниця', 'b6ffab6a-2ddf-49f3-837f-7f006f49172b', 'jamchernihiv', 'uk', null)
        in /app/vendor/sulu/sulu/src/Sulu/Bundle/ContentBundle/Repository/ResourceLocatorRepository.php line 65

    at Sulu\Bundle\ContentBundle\Repository\ResourceLocatorRepository->generate(array('title' => 'Не-латиниця'), 'b6ffab6a-2ddf-49f3-837f-7f006f49172b', 'jamchernihiv', 'uk', 'default')
        in /app/vendor/sulu/sulu/src/Sulu/Bundle/ContentBundle/Controller/NodeResourcelocatorController.php line 45

    at Sulu\Bundle\ContentBundle\Controller\NodeResourcelocatorController->postGenerateAction(object(Request))
        in  line 

    at call_user_func_array(array(object(NodeResourcelocatorController), 'postGenerateAction'), array(object(Request)))
        in /app/var/cache/admin/dev/classes.php line 4228

    at Symfony\Component\HttpKernel\HttpKernel->handleRaw(object(Request), 1)
        in /app/var/cache/admin/dev/classes.php line 4183

    at Symfony\Component\HttpKernel\HttpKernel->handle(object(Request), 1, true)
        in /app/vendor/symfony/symfony/src/Symfony/Component/HttpKernel/Kernel.php line 168

    at Symfony\Component\HttpKernel\Kernel->handle(object(Request))
        in /app/web/admin.php line 53
```

#### Expected Behavior

Normally, at this step, resourcelocator should be generated automatically.

#### Steps to Reproduce

Just create new content with title set to non-latin characters (eg. `"Не-латиниця"`)

Warxcell:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no I hope
| Deprecations? | no
| Fixed tickets | fixes #3367 #
| Related issues/PRs | #5048 
| License | MIT
| Documentation PR |

#### What's in this PR?

Better handling of non-latin symbols for Node Name and ResourceLocator. They are now transliterated.

#### BC Breaks/Deprecations

There shoudn't be any BC breaks. Minor things can change like **world's** was transliterated to **worlds**, now it's **world-s**


alexander-schranz:
Because of BC I would not remove the replacers and as said in the [issue](https://github.com/sulu/sulu/issues/5050#issuecomment-584637355) I know projects directly using the `new PathCleanup();` so I would not change the constructor of that class just directly create the AsciiSlugger inside the PathCleanup construct, then still run the replacers and give then that string to the `AsciiSlugger`. But thats my personal opinion to keep the bc compatibility in this service /cc @sulu/core-team

Warxcell:
I don't think that these replaces are needed. They just bloat the source code. 
`new PathCleanup();` is problematic indeed. This is outside of core?
In that case we can create new class that will replace PathCleanup (PathCleanup will continue to exists as-is, just not to break other projects, but will not be used by Sulu) ?

src/Sulu/Bundle/CoreBundle/DataFixtures/replacers.xml
```diff
@@ -1,1214 +0,0 @@
...
-    </item>
-    <item>
-        <column name="locale">DE</column>
-        <column name="from"><![CDATA[&]]></column>
-        <column name="to">und</column>
-    </item>
-    <item>
-        <column name="locale">EN</column>
-        <column name="from"><![CDATA[&]]></column>
-        <column name="to">and</column>
```
danrot: @Warxcell Can you check how the transliterator behaves in this case? An `&` should be replaced by `and` translated to the given language.

src/Sulu/Component/PHPCR/PathCleanup.php
```diff
@@ -35,14 +35,9 @@ class PathCleanup implements PathCleanupInterface
...
-    /**
-     * PathCleanup constructor.
-     *
-     * @param array $replacers
-     */
-    public function __construct(array $replacers)
+    public function __construct(SluggerInterface $slugger)
     {
-        $this->replacers = $replacers;
+        $this->slugger = $slugger;
```
danrot: I think that's where @alexander-schranz mentioned his different approach. If we really know about code directly using this class, we should really handle this case, although I think it is very bad practice...

But I would handle it differently than alex suggested. I would make the parameter in this constructor nullable, and fallback to instantiate the `Slugger` ourself. But I would throw a [deprecation message](https://symfony.com/doc/current/components/phpunit_bridge.html#trigger-deprecation-notices) in that.

src/Sulu/Component/PHPCR/PathCleanup.php
```diff
@@ -55,44 +50,16 @@ public function __construct(array $replacers)
...
-        }
-
-        if (count($replacers) > 0) {
-            foreach ($replacers as $key => $value) {
-                $dirty = str_replace($key, $value, $dirty);
-            }
+        $parts = explode('/', $dirty);
+        $newParts = [];
+        foreach ($parts as $part) {
+            $part = str_replace('&', '-and-', $part);
```
danrot: I am sorry, but this is not gonna cut it. We can't just replace it with "and", but the "and" has to be translated.

English: "You & I" => "you-and-i"
German: "Du & Ich" => "du-und-ich"
Spanish: "Tu & yo" => "tu-y-yo"

Do you get the gist? That's something we might still need the `replacers.xml` file for, if the transliterator/urlizer or whoever is not taking care of that.

Warxcell: @danrot that's the case only for ResourceLocator, right? In Urlizer I don't see code that handles these.

danrot: Well, if we implement this I would say we also handle the node names this way. Don't see any reason to introduce two different ways here.

If the Urlizer is not handling this (have you tested that?) we probably have to keep the `replacers.xml` file with just these few replacements defined.

Warxcell: Yup, I'm sure. Just tested it. You & I becomes **you-and**. That's current behaviour, with replacers (btw replacers aren't used in Urlizer)

Warxcell: Yup, just tested it. **You & I** becomes **you-i**. Btw replaces are not used in Urlizer at all.

src/Sulu/Component/PHPCR/PathCleanup.php
alexander-schranz: Sounds good for me:

@Warxcell so this would look like something this to keep the backward compatibility:

```php
public function __construct(array $replacers, SluggerInterface $slugger = null)
{
    if (!$slugger) {
            @trigger_error('Initializing the PathCleanup without a slugger is deprecated since Sulu 2.1.', E_USER_DEPRECATED);
            $slugger = new ASSCISlugger();
    }

    $this->replayers = $replacers;
    $this->slugger = $slluger;
}
```

src/Sulu/Component/PHPCR/PathCleanup.php
alexander-schranz: As written I would because of BC not remove the replacers part as somebody could have used it to replace special things in its url generation.

So I would go with keeping this part:

https://github.com/sulu/sulu/blob/94fb1b2f5f6b7566c5f3765800567ed7bc0f0cfe/src/Sulu/Component/PHPCR/PathCleanup.php#L58-L71

and then call the slugger after that. We could maybe remove the replaces which are not longer necessary but would keep the logic of replacing.

Warxcell:
Refactored.

src/Sulu/Component/PHPCR/PathCleanup.php
```diff
@@ -69,30 +81,30 @@ public function cleanup($dirty, $languageCode = null)
...
-
-        $clean = strtolower($dirty);
-
-        // Inspired by ZOOLU
-        // delete problematic characters
-        $clean = str_replace('%2F', '/', urlencode(preg_replace('/([^A-za-z0-9\s\-_\/])/', '', $clean)));
-
         // replace multiple dash with one
-        $clean = preg_replace('/([-]+)/', '-', $clean);
+        $dirty = preg_replace('/([-]+)/', '-', $dirty);
```
nnatter: i think we should apply these filters after applying the slugger. otherwise we might sill get one of these things after applying the `implode`

src/Sulu/Component/PHPCR/Tests/Unit/PathCleanupTest.php
```diff
@@ -44,15 +45,34 @@ protected function setUp(): void
...
+     */
+    public function testCleanup($a, $b, $locale)
     {
-        $clean = $this->cleaner->cleanup('-/aSDf     asdf/äöü-/hello: world\'s', 'de');
+        $clean = $this->cleaner->cleanup($a, $locale);
+        $this->assertEquals($b, $clean);
+    }
 
-        $this->assertEquals('/asdf-asdf/aeoeue/hello-worlds', $clean);
+    public function cleanupProvider()
```
nnatter: can we add a testcase for `multiple dashes`, `dash before slash`, `dash after slash`, `dash at the beginning, `dash at the end` and `multiple slashes`? 😇

src/Sulu/Component/PHPCR/PathCleanup.php
```diff
@@ -69,30 +81,30 @@ public function cleanup($dirty, $languageCode = null)
...
-        $clean = preg_replace('/\/[-]+/', '/', $clean);
+        $dirty = preg_replace('/\/[-]+/', '/', $dirty);
 
         // delete dash at the beginning or end
-        $clean = preg_replace('/^([-])/', '', $clean);
-        $clean = preg_replace('/([-])$/', '', $clean);
+        $dirty = preg_replace('/^([-])/', '', $dirty);
+        $dirty = preg_replace('/([-])$/', '', $dirty);
+
+        $dirty = str_replace('//', '/', $dirty);
```
nnatter: can we use `preg_replace` here too? otherwise we will only replace two slashes with one but fail if there are three slashes 🙈

nnatter: should be something like `$dirty = preg_replace('/([\/]+)/', '-', $dirty);`

src/Sulu/Component/DocumentManager/Slugifier/PathCleanupSlugifier.php
```diff
@@ -0,0 +1,31 @@
...
+    private $pathCleanup;
+
+    public function __construct(PathCleanupInterface $pathCleanup)
+    {
+        $this->pathCleanup = $pathCleanup;
+    }
+
+    public function slugify($text)
+    {
+        return $this->pathCleanup->cleanup($text);
```
nnatter: looks like this is the reason for the `ContentRepositoryTest` errors. phpcr cannot handle `/` inside of paths. the previously used `Urlizer` class replaced these with `-`. so we need to replace `/` with `-` in this class too to keep things compatible.

src/Sulu/Component/PHPCR/PathCleanup.php
```diff
@@ -69,30 +81,30 @@ public function cleanup($dirty, $languageCode = null)
...
 
-        // remove double slashes
-        $clean = str_replace('//', '/', $clean);
+        $parts = explode('/', $dirty);
+        $newParts = [];
+        foreach ($parts as $part) {
+            $slug = $this->slugger->slug($part, '-', $languageCode);
```
nnatter: looks like this is the reason why the `PageControllerTest::testCopyWithShadow` fails. the previously used regex (`'/([^A-za-z0-9\s\-_\/])/'`) allowed `_` characters. Now they are replaced with `-`. But it would be okay to me to adjust the tests here, i dont think it is good practice to use `_` in urls anymore 🙂

src/Sulu/Component/DocumentManager/Slugifier/PathCleanupSlugifier.php
danrot: And I think our resource locator field in the form also behaves in an unexpected behavior if the editable part contains a slash.

src/Sulu/Component/PHPCR/PathCleanup.php
Warxcell: Will add tests for these

src/Sulu/Component/PHPCR/Tests/Unit/PathCleanupTest.php
Warxcell: Will add test for theses

src/Sulu/Component/PHPCR/PathCleanup.php
Warxcell: Not a good idea either. Because it will explode multiple dashes and create empty array that will implode later and introduce same multiple dashes.

nnatter: but if we apply the filters after the `implode` calls the multiple slashes will be removed, right? 🤔

src/Sulu/Component/DocumentManager/Slugifier/PathCleanupSlugifier.php
```diff
@@ -0,0 +1,50 @@
...
+        $this->pathCleanup = $pathCleanup;
+    }
+
+    public function slugify($text)
+    {
+        $text = $this->pathCleanup->cleanup($text);
+        $text = str_replace('/', '-', $text);
+
+        // Remove apostrophes which are not used as quotes around a string
+        $text = preg_replace('/(\\w)\'(\\w)/', '${1}${2}', $text);
```
nnatter: is there a specific reason for applying exactly these filters here? 🙂

Warxcell: I copied the old ones, hoping they will fix the tests. :) I don't have PDO installed on this machine and I cannot run functional tests. :(

alexander-schranz:
@Warxcell your changes are correct but we need to change the tests:

```diff
diff --git a/src/Sulu/Bundle/PageBundle/Tests/Functional/Controller/PageControllerTest.php b/src/Sulu/Bundle/PageBundle/Tests/Functional/Controller/PageControllerTest.php
index 7338e78f27..3a54f1f3ab 100644
--- a/src/Sulu/Bundle/PageBundle/Tests/Functional/Controller/PageControllerTest.php
+++ b/src/Sulu/Bundle/PageBundle/Tests/Functional/Controller/PageControllerTest.php
@@ -1418,7 +1418,7 @@ class PageControllerTest extends SuluTestCase
     {
         $document = $this->createPageDocument();
         $document->setTitle('test_en');
-        $document->setResourceSegment('/test_en');
+        $document->setResourceSegment('/test-en');
         $document->setStructureType('default');
         $document->getStructure()->bind([
             'tags' => [
@@ -1433,7 +1433,7 @@ class PageControllerTest extends SuluTestCase
 
         $document = $this->documentManager->find($document->getUuid(), 'de', ['load_ghost_content' => false]);
         $document->setTitle('test_de');
-        $document->setResourceSegment('/test_de');
+        $document->setResourceSegment('/test-de');
         $document->setStructureType('default');
         $document->getStructure()->bind([
             'tags' => [
@@ -1464,10 +1464,10 @@ class PageControllerTest extends SuluTestCase
         $uuid = json_decode($this->client->getResponse()->getContent(), true)['id'];
 
         $germanDocument = $this->documentManager->find($uuid, 'de');
-        $this->assertStringStartsWith('/test_de/test_de', $germanDocument->getResourceSegment());
+        $this->assertStringStartsWith('/test-de/test-de', $germanDocument->getResourceSegment());
 
         $englishDocument = $this->documentManager->find($uuid, 'en');
-        $this->assertStringStartsWith('/test_en/test_en', $englishDocument->getResourceSegment());
+        $this->assertStringStartsWith('/test-en/test-en', $englishDocument->getResourceSegment());
     }
 
     public function testCopyNonExistingSource()
diff --git a/src/Sulu/Bundle/PageBundle/Tests/Functional/Import/WebspaceImportTest.php b/src/Sulu/Bundle/PageBundle/Tests/Functional/Import/WebspaceImportTest.php
index d41252f4c6..33044904f2 100644
--- a/src/Sulu/Bundle/PageBundle/Tests/Functional/Import/WebspaceImportTest.php
+++ b/src/Sulu/Bundle/PageBundle/Tests/Functional/Import/WebspaceImportTest.php
@@ -207,8 +207,8 @@ class WebspaceImportTest extends SuluTestCase
         $this->assertEquals('/cmf/sulu_io/contents/child', $loadedDocuments[1]->getPath());
 
         // resource segment
-        $this->assertEquals('/parent-ru-new-shit', $loadedDocuments[0]->getResourceSegment());
-        $this->assertEquals('/child-ru-new-shit', $loadedDocuments[1]->getResourceSegment());
+        $this->assertEquals('/privet', $loadedDocuments[0]->getResourceSegment());
+        $this->assertEquals('/privet-privet-privet', $loadedDocuments[1]->getResourceSegment());
 
         // seo
         $this->assertEquals('', $loadedDocuments[0]->getExtensionsData()->toArray()['seo']['title']);
```

As discussed with @danrot @chirimoya underlines in url is ok to be replaced with dash and the WebspaceImport test need to be changed as we are now able to generate the correct title out of the russian letters.

If you also fix the merge conflict this PR is for me ready to be merged.

@Warxcell Good work 👍

composer.json
```diff
@@ -98,7 +98,8 @@
         "symfony/validator": "^4.3",
         "symfony/yaml": "^4.3",
         "toflar/psr6-symfony-http-cache-store": "^1.1 || ^2.0",
-        "twig/twig": "^1.41 || ^2.0"
+        "twig/twig": "^1.41 || ^2.0",
+        "symfony/string": "^5.0"
```
danrot: You've added this manually right? If you add it using composer it would be alphabetically sorted :slightly_smiling_face: Can you please change that?

src/Sulu/Bundle/CoreBundle/DataFixtures/replacers.xml
```diff
@@ -1211,4 +1211,10 @@
         <column name="from"><![CDATA[&]]></column>
         <column name="to">y</column>
     </item>
```
danrot: Shouldn't we remove all the replacers just for the letters (so basically everything except the replacers for `&`)? Isn't that handled by the `String` component now? Might lead to confusing situations otherwise.

src/Sulu/Bundle/DocumentManagerBundle/DependencyInjection/Configuration.php
```diff
@@ -72,6 +72,7 @@ public function getConfigTreeBuilder()
                 ->end()
                 ->scalarNode('slugifier')
                     ->defaultValue('Sulu\Bundle\DocumentManagerBundle\Slugifier\Urlizer::urlize')
+                    ->setDeprecated()
```
danrot: Is just the `defaultValue` deprecated, or the entire option? Is the `slugifier` option used at all now?

src/Sulu/Component/DocumentManager/Slugifier/PathCleanupSlugifier.php
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+namespace Sulu\Component\DocumentManager\Slugifier;
+
+use Sulu\Component\PHPCR\PathCleanupInterface;
+use Symfony\Cmf\Api\Slugifier\SlugifierInterface;
+
+class PathCleanupSlugifier implements SlugifierInterface
+{
+    /** @var PathCleanupInterface */
```
danrot: We format these comments usually this way:

```php
/**
 * @var PathCleanupInterface
 */
```

src/Sulu/Component/PHPCR/PathCleanup.php
```diff
@@ -11,8 +11,13 @@
 
 namespace Sulu\Component\PHPCR;
 
+use Symfony\Component\String\Slugger\AsciiSlugger;
+use Symfony\Component\String\Slugger\SluggerInterface;
+
 /**
  * cleans path strings.
+ *
+ * @internal this class is for internal use
```
danrot: I would just write `@internal`, since the additional comment is just an explanation of what internal means.

src/Sulu/Component/PHPCR/PathCleanup.php
```diff
@@ -23,6 +28,9 @@ class PathCleanup implements PathCleanupInterface
      */
     protected $replacers = [];
 
+    /** @var SluggerInterface */
```
danrot:
```php
/*
 * @var SluggerInterface
 */
```

src/Sulu/Bundle/CoreBundle/DataFixtures/replacers.xml
Warxcell: I left them as-is because of BC. If somebody parses and uses them manually.

src/Sulu/Bundle/DocumentManagerBundle/DependencyInjection/Configuration.php
Warxcell: The whole option. It's not used. Internally now it uses PathCleaner.

src/Sulu/Bundle/CoreBundle/DataFixtures/replacers.xml
danrot: Okay, sounds good :+1:

src/Sulu/Component/PHPCR/PathCleanup.php
Warxcell: Or better say if you use it - use it throught services provided by bundle?

src/Sulu/Bundle/DocumentManagerBundle/DependencyInjection/Configuration.php
danrot: I think the `setDeprecated` function takes a message as argument, right? Then I would add one to make it more clear:

```
The "sulu_document_manager.slugifier" configuration option is not used anymore.
```

src/Sulu/Component/PHPCR/PathCleanup.php
danrot: If we allow that the service/class is not internal IMO. Maybe that's the even better answer to this? Other people are already using this service, so there seems to be a need for it. Maybe it should not be internal at all?

Warxcell: Yup, I think it can be used, but not instantiate it directly. That was the main issue here. 
People should follow the interface, not class - but in this case we DIDN'T changed the class because people don't follow  the interface. (The interface does not contains anything about __construct)

Warxcell: If people follows the inteface, they should care only about `cleanup` and `validate` methods to be same - and this would not  be necessary

`        if (null === $slugger) {
            @trigger_error(
                'Initializing the PathCleanup without a slugger is deprecated since Sulu 2.1.',
                E_USER_DEPRECATED
            );
            $slugger = new AsciiSlugger();
        }`

src/Sulu/Bundle/DocumentManagerBundle/DependencyInjection/Configuration.php
```diff
@@ -72,7 +72,7 @@ public function getConfigTreeBuilder()
                 ->end()
                 ->scalarNode('slugifier')
                     ->defaultValue('Sulu\Bundle\DocumentManagerBundle\Slugifier\Urlizer::urlize')
-                    ->setDeprecated()
+                    ->setDeprecated('The "sulu_document_manager.slugifier" is deprecated since 2.1. To be removed 3.0.')
```
danrot: The "not used" part was very important to me, otherwise people will still wonder if this does not have any effect.

src/Sulu/Bundle/DocumentManagerBundle/DependencyInjection/Configuration.php
Warxcell: added better messages. As you said @internal says what it have to say, no need for explanatory. Here is the same. :)

src/Sulu/Component/PHPCR/PathCleanup.php
danrot: That's at least not my understanding of the `@internal` tag. I would assume that the entire class may not be used outside of the sulu/sulu package. Let me quote the [phpDocumentor documentation](https://docs.phpdoc.org/latest/references/phpdoc/tags/internal.html#description):

> The @internal tag can be used as counterpart of the @api tag, indicating that the associated Structural Elements are used purely for the internal workings of this piece of software.

@sulu/core-team Does anyone of you have a different opinion on that? If this annotation is not used consistently they are not very helpful.

Warxcell:
why is `ci/circleci ` failing? (think)

nnatter:
> why is `ci/circleci ` failing?

It looks like the service has problems since yesterday that dont have to do anything with this project. So you dont have to worry about that one 🙂

src/Sulu/Component/PHPCR/PathCleanup.php
nnatter: Thats my understanding of the `@internal` tag too, therefore i would not mark the `PathCleanup` as internal. 

I think it is an anti-pattern in general to directly instantiate something that is available as a service. So i dont think it is necessary to explicitly clarify this here. 🙂

danrot:
Slightly adjusted the deprecation message, especially because php-cs-fixer build was not running (might be related that we introduced that one after you created this PR).

Just waiting for the tests now and testing it manually, if everything works I am ready to merge :slightly_smiling_face:

alexander-schranz:
@Warxcell Good work 👍  Thank you!

danrot:
@Warxcell Yes, was a pleasure to work with you :+1: Thanks for the contribution!

PR #5065, opened 2020-02-18
Extract chip component for better reusability

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR is extracted from #5035 and contains a new `Chip` component.

#### Why?

Because the `Chip` component will also be used when filtering lists. Until now this component was only available to the `MultiAutoComplete`.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Chip/tests/Chip.test.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Chip/tests/Chip.test.js
@@ -11,6 +11,16 @@ test('Should render item without icon in disabled state', () => {
     expect(render(<Chip disabled={true} onDelete={jest.fn()} value={{}}>Name</Chip>)).toMatchSnapshot();
 });
 
+test('Should call onClick callback when the button is clicked', () => {
+    const clickSpy = jest.fn();
+    const value = {name: 'Test'};
+    const item = shallow(<Chip onClick={clickSpy} onDelete={jest.fn()} value={value}>Test</Chip>);
+
+    item.find('button').simulate('click');
+
+    expect(clickSpy).toBeCalledWith(value);
+});
+
 test('Should call onDelete callback when the times icon is clicked', () => {
     const deleteSpy = jest.fn();
     const value = {name: 'Test'};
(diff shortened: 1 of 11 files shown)


danrot:
Includes ee40522c5b7224697c320692188899db14eb0b3e and dfa3d095c6495e468b25379debf7c0496b52125d from #5035.

nnatter:
i think this pr contains unrelated changes right now

nnatter:
should we target the `release/2.0` branch here? otherwise we might end up with unnecessary merge conflicts in the future. this is not a new feature imo

danrot:
The unrelated changes will be gone as soon as I push the `master` branch :slightly_smiling_face: 

> should we target the `release/2.0` branch here? otherwise we might end up with unnecessary merge conflicts in the future. this is not a new feature imo

But it not a bugfix either... I think I would leave it as it is, because it might have potential bc breaks, and I don't think that this component will change a lot in 2.0...

src/Sulu/Bundle/AdminBundle/Resources/js/components/Chip/tests/Chip.test.js
```diff
@@ -0,0 +1,36 @@
+// @flow
+import React from 'react';
+import {render, shallow} from 'enzyme';
+import Chip from '../Chip';
+
+test('Should render item with children', () => {
+    expect(render(<Chip value={{}}>Name</Chip>)).toMatchSnapshot();
+});
+
+test('Should render item with delete icon', () => {
```
nnatter: should we add a test case to check if the item is rendered `clickable` if an `onClick` callback is giben?

src/Sulu/Bundle/AdminBundle/Resources/js/components/Chip/tests/Chip.test.js
```diff
@@ -0,0 +1,36 @@
...
+    expect(render(<Chip onDelete={jest.fn()} value={{}}>Name</Chip>)).toMatchSnapshot();
+});
+
+test('Should render item without icon in disabled state', () => {
```
nnatter: `without delete icon`

PR #5066, opened 2020-02-18
Add tags to metadata and add filter for metadata provider

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | none
| Related issues/PRs | none
| License | MIT
| Documentation PR | none

#### What's in this PR?

This PR add tags to form metadata and the ability to filter types by tags.

#### Why?

This is needed in the article bundle to allow article-type based forms.

#### Example Usage

~~~php
$this->viewBuilderFactory->createPreviewFormViewBuilder(static::EDIT_FORM_VIEW_DETAILS . '_' . $typeKey, '/details')
    ->setResourceKey('articles')
    ->addMetadataRequestParameters([
       'tags[sulu_article.type]' => $typeConfig['type'] ? true : false,
       'tags[sulu_article.type][type]' => $typeConfig['type'],
    ])
~~~

OR:

~~~php
$this->viewBuilderFactory->createPreviewFormViewBuilder(static::EDIT_FORM_VIEW_DETAILS . '_' . $typeKey, '/details')
    ->setResourceKey('articles')
    ->addMetadataRequestParameters([
       'tags[sulu_article.type][type]' => $typeConfig['type'],
    ])
~~~

#### To Do

- [x] Add tests

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/FormMetadata/FormXmlLoader.php
+++ b/src/Sulu/Bundle/AdminBundle/FormMetadata/FormXmlLoader.php
@@ -101,6 +101,7 @@ class FormXmlLoader extends AbstractLoader
     private function mapFormsMetadata(ExternalFormMetadata $formMetadata, string $locale): FormMetadata
     {
         $form = new FormMetadata();
+        $form->setTags($this->formMetadataMapper->mapTags($formMetadata->getTags()));
         $form->setItems($this->formMetadataMapper->mapChildren($formMetadata->getChildren(), $locale));
 
         $schema = $this->formMetadataMapper->mapSchema($formMetadata->getProperties());
--- a/src/Sulu/Bundle/AdminBundle/FormMetadata/FormXmlLoader.php
+++ b/src/Sulu/Bundle/AdminBundle/FormMetadata/FormXmlLoader.php
@@ -72,6 +72,7 @@ class FormXmlLoader extends AbstractLoader
         $form = new ExternalFormMetadata();
         $form->setResource($resource);
         $form->setKey($xpath->query('/x:form/x:key')->item(0)->nodeValue);
+        $form->setTags($this->loadStructureTags('/x:form/x:tag', $xpath));
 
         $propertiesNode = $xpath->query('/x:form/x:properties')->item(0);
         $properties = $this->propertiesXmlParser->load(
--- a/src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/FormMetadata.php
+++ b/src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/FormMetadata.php
@@ -118,15 +118,19 @@ class FormMetadata extends AbstractMetadata
         return $this->tags;
     }
 
-    public function getTag(string $name): ?TagMetadata
+    /**
+     * @return TagMetadata[]
+     */
+    public function getTagsByName(string $name): array
     {
+        $tags = [];
         foreach ($this->getTags() as $tag) {
             if ($tag->getName() === $name) {
-                return $tag;
+                $tags[] = $tag;
             }
         }
 
-        return null;
+        return $tags;
     }
 
     public function addTag(TagMetadata $tag): void
--- a/src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/FormMetadata.php
+++ b/src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/FormMetadata.php
@@ -44,6 +44,11 @@ class FormMetadata extends AbstractMetadata
      */
     private $key;
 
+    /**
+     * @var TagMetadata[]
+     */
+    protected $tags;
+
     public function setName(string $name)
     {
         $this->name = $name;
@@ -103,6 +108,38 @@ class FormMetadata extends AbstractMetadata
     public function getSchema(): SchemaMetadata
     {
         return $this->schema;
+    }
+
+    /**
+     * @return TagMetadata[]
+     */
+    public function getTags(): array
+    {
+        return $this->tags;
+    }
+
+    public function getTag(string $name): ?TagMetadata
+    {
+        foreach ($this->getTags() as $tag) {
+            if ($tag->getName() === $name) {
+                return $tag;
+            }
+        }
+
+        return null;
+    }
+
+    public function addTag(TagMetadata $tag): void
+    {
+        $this->tags[] = $tag;
+    }
+
+    /**
+     * @param TagMetadata[] $tags
+     */
+    public function setTags(array $tags): void
+    {
+        $this->tags = $tags;
     }
 
     public function merge(self $otherForm): FormMetadata
--- a/src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/FormMetadataProvider.php
+++ b/src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/FormMetadataProvider.php
@@ -58,9 +58,8 @@ class FormMetadataProvider implements MetadataProviderInterface
 
             if (array_key_exists('tags', $metadataOptions)) {
                 $tags = $metadataOptions['tags'];
-                $tagAttributes = $metadataOptions['tagAttributes'] ?? [];
-                foreach ($tags as $tagName => $tagValue) {
-                    $this->filterByTag($form, $tagName, $tagValue, $tagAttributes[$tagName] ?? 'value');
+                foreach ($tags as $tagName => $tagAttributes) {
+                    $this->filterByTag($form, $tagName, $tagAttributes ?: []);
                 }
             }
         }
@@ -68,16 +67,32 @@ class FormMetadataProvider implements MetadataProviderInterface
         return $form;
     }
 
-    private function filterByTag(TypedFormMetadata $form, string $tagName, string $tagValue, string $tagAttribute): void
+    private function filterByTag(TypedFormMetadata $form, string $tagName, array $tagAttributes): void
     {
         foreach ($form->getForms() as $formKey => $childForm) {
-            $tag = $childForm->getTag($tagName);
-            $actualTagValue = $tag ? $tag->getAttribute($tagAttribute) : '';
-
-            if ($actualTagValue !== $tagValue) {
+            if (!$this->matchFormAgainstTag($childForm, $tagName, $tagAttributes)) {
                 $form->removeForm($formKey);
             }
         }
+    }
+
+    private function matchFormAgainstTag(FormMetadata $form, string $tagName, array $tagAttributes): bool
+    {
+        $tags = $form->getTagsByName($tagName);
+        if (array_key_exists('exists', $tagAttributes)) {
+            $exists = filter_var($tagAttributes['exists'], FILTER_VALIDATE_BOOLEAN);
+            if (($exists && 0 === \count($tags)) || (!$exists && 0 < \count($tags))) {
+                return false;
+            }
+        }
+
+        foreach ($tags as $tag) {
+            if (!$tag->matchAttributes($tagAttributes)) {
+                return false;
+            }
+        }
+
+        return true;
     }
 
     /**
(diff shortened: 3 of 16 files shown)


src/Sulu/Bundle/AdminBundle/FormMetadata/FormXmlLoader.php
```diff
@@ -101,6 +101,7 @@ protected function parse($resource, \DOMXPath $xpath, $type): LocalizedFormMetad
     private function mapFormsMetadata(ExternalFormMetadata $formMetadata, string $locale): FormMetadata
     {
         $form = new FormMetadata();
+        $form->setTags($this->formMetadataMapper->mapTags($formMetadata->getTags()));
```
wachterjohannes: i cannot set it in the tests because the xml does not accept it - should i remove the mapping here?

src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/FormMetadataProvider.php
```diff
@@ -55,11 +55,31 @@ public function getMetadata(string $key, string $locale, array $metadataOptions
             foreach ($form->getForms() as $formType) {
                 $this->evaluateFormItemExpressions($formType->getItems());
             }
+
+            if (array_key_exists('tags', $metadataOptions)) {
+                $tags = $metadataOptions['tags'];
+                $tagAttributes = $metadataOptions['tagAttributes'] ?? [];
```
nnatter: this does not look intuitive to me.. shouldnt it be possible to filter for tags without a specific value?

nnatter:
do you see a reason for not using something like this?
```
    ->addMetadataRequestParameters([
        'tags[sulu_article.type][type]' => $typeConfig['type'] ?: '',
    ])
```

src/Sulu/Bundle/AdminBundle/FormMetadata/FormXmlLoader.php
danrot: Why not adding it to the other XML as well? :slightly_smiling_face:

src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/FormMetadata.php
```diff
@@ -105,6 +110,38 @@ public function getSchema(): SchemaMetadata
...
+    public function getTags(): array
+    {
+        return $this->tags;
+    }
+
+    public function getTag(string $name): ?TagMetadata
+    {
+        foreach ($this->getTags() as $tag) {
+            if ($tag->getName() === $name) {
+                return $tag;
```
danrot: Is that really correctly implemented this way? Asking because in Symfony you can also have multiple tags with the same type, wondering how we've implemented it...

src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/FormMetadataProvider.php
danrot: I think that might be a valid use case. And I have to agree that I have a hard time understanding this code :confused:

src/Sulu/Bundle/AdminBundle/Tests/Functional/Metadata/FormMetadata/FormMetadataProviderTest.php
```diff
@@ -50,4 +50,30 @@ public function testGetMetadataFromStructureLoader()
...
+
+        $typedForm = $this->formMetadataProvider->getMetadata('page', 'de', ['tags' => ['test' => 'test-value2']]);
+        $this->assertInstanceOf(TypedFormMetadata::class, $typedForm);
+        $this->assertCount(0, $typedForm->getForms());
+
+        $typedForm = $this->formMetadataProvider->getMetadata('page', 'de', ['tags' => ['test2' => 'test-value2']]);
+        $this->assertInstanceOf(TypedFormMetadata::class, $typedForm);
+        $this->assertCount(0, $typedForm->getForms());
+
+        $typedForm = $this->formMetadataProvider->getMetadata('page', 'de', ['tags' => ['test2' => 'test-value2'], 'tagAttributes' => ['test2' => 'test']]);
```
danrot: What does `['test2' => 'test']` mean in that case? I don't really understand what is being filtered here by just reading that line.

src/Sulu/Bundle/AdminBundle/FormMetadata/FormXmlLoader.php
wachterjohannes: will do it

nnatter:
The new syntax definitely looks better to me 🙂 One thing that still is not perfect is that `exists` could be an attribute of the tag. What do you think about this:

```php
->addMetadataRequestParameters([
    // tag must exist / must not exist
    'tags[sulu_article.type] => $typeConfig['type'] ? true : false,
    // tag must exist and attribute of tag must have the given value
    'tags[sulu_article.type][type]' => $typeConfig['type'],
])
```

This makes things a bit more complex on the php side that evaluates the `tags` query parameter, but it would make the url design more consistent/intuitive imo. What do you think?

alexander-schranz:
I personally like `tags[sulu_article.type][type]` I would had prefered that syntax also in the property name instead of ext/seo/title. But at current state its inconsistent if we know use html array syntax here and the json path syntax somewhere else so if we want to be consistent we should use `tags/sulu_article.type/type`.

alexander-schranz:
I personally would not implement filtering the template dropdown by type, as in the past we had request from content managers who wanted to switch a template to another type template and we had to go into the database todo that. So I see the types more as a group in the template dropdown and not filter the dropdown by the type.

wachterjohannes:
@alexander-schranz i cannot agree with your opinion - the type change in the article-bundle should be a conscious decision and for example an own button `Change type`. Although that the grouping in other parts of the system makes sense.

The problem is when we allow changing the type in the article bundle we have to reconsider the routing of the list / add forms - which contains currently (in the 1.x version) the type `articles/de/my-type`, `articles/en/my-type/aba31ccd-e403-415b-b8b9-3d8a4d9126c3/details`. When we then change the template we would have to update the URL which is a quite heavy task i think and is quite article-bundle specific.

@danrot what do you think?

danrot:
I think it depends on the use case, but in our discussion today it sounded like a lot is happening when the Article Type is changed, so I would also say it should be a more conscious decision.

Regarding the suggestion of @nnatter: That's something that we could implement, but I think that could also be in a separate step, and must not be necessarily implemented in this PR.

nnatter:
Changing the semantics of the `tags` parameter at a later time would be kind of a bc break, wouldnt it?

danrot:
@nnatter The way I see it, it does not...

```php
->addMetadataRequestParameters([
    // this is currently not working at all, so it would not be a BC break if we make it work at some point
    'tags[sulu_article.type]' => true,
    // this is currently working, and would not change in a future release
    'tags[sulu_article.type][type]' => $typeConfig['type'],
])
```

Or am I mixing up something?

nnatter:
The behaviour of the following would change as described in the comment:
```php
->addMetadataRequestParameters([
    // current semantic: the 'sulu_article.type' exists on the template
    // future semantic: the 'exists' attribute of the 'sulu_article.type' tag matches 'true'
    'tags[sulu_article.type][exists]' => true,
])
```

danrot:
Ah, haven't had a closer look at the PR now :see_no_evil: If we do it this way I would currently just remove the special handling of the `exists` attribute, resp. change the behavior to what you've suggested immediately. Shouldn't be too much work, should it? :thinking:

wachterjohannes:
@danrot if we agree in that i will change the configuration and remove the `exists` :)

src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/FormMetadataProvider.php
```diff
@@ -55,11 +55,57 @@ public function getMetadata(string $key, string $locale, array $metadataOptions
             foreach ($form->getForms() as $formType) {
                 $this->evaluateFormItemExpressions($formType->getItems());
             }
+
+            if (array_key_exists('tags', $metadataOptions)) {
+                $tags = $metadataOptions['tags'];
```
danrot: Are you only using that variable once, or am I missing something? :thinking:

src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/TagMetadata.php
```diff
@@ -42,4 +47,30 @@ public function setPriority(?int $priority): void
...
+    {
+        return $this->attributes;
+    }
+
+    public function getAttribute(string $name)
+    {
+        return $this->attributes[$name] ?? null;
+    }
+
+    public function matchAttributes(array $attributes): bool
```
danrot: How about naming this `hasAttributes`? Would be clearer to me what this function does then. If you e.g. think about `preg_match` it does a little bit more than just returning a boolean.

src/Sulu/Bundle/AdminBundle/Tests/Functional/Metadata/FormMetadata/FormMetadataProviderTest.php
```diff
@@ -50,4 +50,59 @@ public function testGetMetadataFromStructureLoader()
...
+            'de',
+            ['tags' => ['test2' => ['value' => 'test-value2']]]
+        );
+        $this->assertInstanceOf(TypedFormMetadata::class, $typedForm);
+        $this->assertCount(0, $typedForm->getForms());
+
+        $typedForm = $this->formMetadataProvider->getMetadata(
+            'page',
+            'de',
+            ['tags' => ['test2' => ['exists' => true, 'test' => 'test-value2']]]
```
danrot: Did that `exists` slip? Or does it have a purpose?

src/Sulu/Component/Content/Metadata/Loader/schema/form-1.0.xsd
```diff
@@ -16,11 +16,12 @@
     <xs:element name="form" type="formType"/>
 
     <xs:complexType name="formType">
-        <xs:all>
+        <xs:choice maxOccurs="unbounded">
```
danrot: What was that change for?

src/Sulu/Bundle/AdminBundle/Tests/Functional/Metadata/FormMetadata/FormMetadataProviderTest.php
wachterjohannes: no it slipped :)

src/Sulu/Component/Content/Metadata/Loader/schema/form-1.0.xsd
wachterjohannes: it is copied from `template-1.0.xsd` it allows to add the mulitple tags to the form

src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/TagMetadata.php
wachterjohannes: 👍

wachterjohannes:
@danrot fixed

src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/TagMetadata.php
```diff
@@ -42,4 +47,30 @@ public function setPriority(?int $priority): void
...
+
+    public function getAttribute(string $name)
+    {
+        return $this->attributes[$name] ?? null;
+    }
+
+    public function hasAttributes(array $attributes): bool
+    {
+        foreach ($this->attributes as $key => $value) {
+            if (($attributes[$key] ?? null) !== $value) {
```
danrot: Are you sure that's correct? It's not really easy to understand, and I have encountered an error. If I put something like this on the `PageAdmin`, it will also show pages that have this tag without the given attribute:

```php
->addMetadataRequestParameters([
    'tags[sulu_page.type][type]' => 'test',
])
```

This will also return template that have the `sulu_page.type` tag without any attribute.

PR #5067, opened 2020-02-18
Change structure of filter list metadata

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | #5035
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR changes the structure of the list metadata regarding the filters.

#### Why?

The filters were not functional until now, and a change was required anyway, because we need to allow parameters to be nested, similar to the collections in the XML for the page templates.

#### BC Breaks/Deprecations

Yes, the XSD for list metadata was changed.

The new files have to look something like that:

```diff
-        <identity-property name="accountId" visibility="never" filter-type="auto-complete" translation="sulu_contact.organization">
+        <identity-property name="accountId" visibility="never" translation="sulu_contact.organization">
             <field-name>account</field-name>
             <entity-name>SuluContactBundle:AccountContact</entity-name>
 
             <joins ref="accountContact"/>
 
-            <filter-type-parameters>
-                <parameter key="singleUrl"><![CDATA[/admin/api/accounts/{id}]]></parameter>
-                <parameter key="remoteUrl">
-                    <![CDATA[/admin/api/accounts?searchFields=name,number&fields=id,name&flat=true]]>
-                </parameter>
-                <parameter key="resultKey">accounts</parameter>
-                <parameter key="valueKey">name</parameter>
-            </filter-type-parameters>
+            <filter type="selection">
+                <param name="displayProperty" value="name" />
+                <param name="resourceKey" value="accounts" />
+            </filter>
         </identity-property>
```

#### To Do

- [x] Check if information is passed via the REST API


src/Sulu/Component/Rest/ListBuilder/Metadata/ListXmlLoader.php
```diff
@@ -129,8 +129,14 @@ private function loadPropertyMetadata(\DOMXPath $xpath, \DOMNode $propertyNode)
         $propertyMetadata->setSortable(
             XmlUtil::getBooleanValueFromXPath('@sortable', $xpath, $propertyNode, true)
         );
-        $propertyMetadata->setFilterType(XmlUtil::getValueFromXPath('@filter-type', $xpath, $propertyNode));
-        $propertyMetadata->setFilterTypeParameters($this->getFilterTypeParameters($xpath, $propertyNode));
```
danrot: Not sure if we should introduce a new `FilterTypeMetadata` class instead of having two prefixed properties on that `PropertyMetadata` class. But somehow it feels like not big enough of a deal to accept that BC break...

danrot:
Includes dbb9db43d26ab4834b2b387668f970c387c8a8f7 from #5035

src/Sulu/Bundle/TagBundle/Resources/config/lists/tags.xml
```diff
@@ -40,11 +40,12 @@
             name="name"
             visibility="always"
             searchability="yes"
-            filter-type="string"
             translation="sulu_tag.tag"
         >
             <field-name>name</field-name>
             <entity-name>%sulu.model.tag.class%</entity-name>
+
+            <filter type="text" />
```
nnatter: shouldnt the user use the search bar? i guess there is only a single tab with a specific name?

src/Sulu/Bundle/TagBundle/Resources/config/lists/tags.xml
```diff
@@ -60,6 +61,11 @@
 
                 <joins ref="changer"/>
             </field>
+
+            <filter type="selection">
```
nnatter: you did not add this filter to other lists. is there a reason for adding it here specifically?

src/Sulu/Bundle/TagBundle/Resources/config/lists/tags.xml
```diff
@@ -85,6 +93,11 @@
 
                 <joins ref="creator"/>
             </field>
+
+            <filter type="selection">
```
nnatter: same as above

src/Sulu/Component/Rest/ListBuilder/Metadata/ListXmlLoader.php
nnatter: depends if we plan to support multiple filter-types per property in the future. if yes, i would probably make this BC break now

nnatter: also if we expect that the `filter` node will get additional attributes in the future

src/Sulu/Component/Rest/ListBuilder/Metadata/Resources/schema/list-2.0.xsd
```diff
@@ -72,18 +72,20 @@
         <xs:attributeGroup ref="fieldAttributeGroup"/>
     </xs:complexType>
 
-    <xs:complexType name="parametersType">
+    <xs:complexType name="filterType">
         <xs:sequence>
-            <xs:element type="parameterType" name="parameter" minOccurs="1" maxOccurs="unbounded"/>
+            <xs:element type="paramType" name="param" minOccurs="0" maxOccurs="unbounded"/>
```
nnatter: inf the form xml files, we use a `<params>` tag that wraps the parameters? should we use that here too? 

maybe the filters will get a `<meta>` tag too in the future? eg for defining a filter name?

src/Sulu/Component/Rest/ListBuilder/Metadata/Resources/schema/list-2.0.xsd
```diff
@@ -72,18 +72,20 @@
...
-        <xs:simpleContent>
-            <xs:extension base="xs:string">
-                <xs:attribute name="key" type="xs:string"/>
-            </xs:extension>
-        </xs:simpleContent>
+    <xs:complexType name="paramType">
+        <xs:sequence>
+            <xs:element type="paramType" name="param" minOccurs="0" maxOccurs="unbounded"/>
+        </xs:sequence>
+        <xs:attribute name="name" type="xs:string"/>
```
nnatter: in the `properties-1.0.xsd` we declared this as required

src/Sulu/Component/Rest/ListBuilder/Metadata/Resources/schema/list-2.0.xsd
```diff
@@ -72,18 +72,20 @@
...
+        <xs:attribute name="type" type="xs:string"/>
```
nnatter: in the `properties-1.0.xsd` we restricted this to a set of strings and set a default value

src/Sulu/Component/Rest/Tests/Unit/ListBuilder/Metadata/ListXmlLoaderTest.php
```diff
@@ -224,10 +224,10 @@ public function testLoadMetadataFromFileParameters()
 
     public function testLoadMetadataFromFileNoInputType()
     {
-        $result = $this->listXmlLoader->load(__DIR__ . '/Resources/filter-type-no-input.xml');
+        $result = $this->listXmlLoader->load(__DIR__ . '/Resources/filter-type-collection.xml');
```
nnatter: `filter-type-collection-param.xml`?

src/Sulu/Component/Rest/Tests/Unit/ListBuilder/Metadata/ListXmlLoaderTest.php
```diff
@@ -224,10 +224,10 @@ public function testLoadMetadataFromFileParameters()
 
     public function testLoadMetadataFromFileNoInputType()
```
nnatter: not sure if this is still accurate 😉

UPGRADE.md
```diff
@@ -2,6 +2,52 @@
 
 ## dev-master
 
+### Configuration of list filters
+
+The old filter configuration on a list XML file was ignored until now. But the new list filtering functionality requires
+some changes in order to work. That also has some effects on the list XML files. Mind that this only concerns you if
+you had some `filter-type` attributes in your configurations, which would not have any effect at all until now.
```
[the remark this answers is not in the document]
danrot: I would argue that this is a [if conditional type two](https://www.ego4u.de/de/cram-up/grammar/conditional-sentences) :slightly_smiling_face: It is kind of unlikely that somebody used that, therefore "would" makes sense to me.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/types.js
```diff
@@ -12,6 +12,8 @@ export type ColumnItem = DataItem & {
 };
 
 export type SchemaEntry = {
+    filterType: ?string,
+    filterTypeParameters: ?{[string]: mixed},
```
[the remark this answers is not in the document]
danrot: I think you are right, but I think I rather change the value of the API instead of switching between array and object type. Will change the API.

src/Sulu/Component/Rest/ListBuilder/Metadata/Resources/schema/list-2.0.xsd
danrot: I already hat the `params` tag there :see_no_evil: But if the majority wants to have them I am fine with introducing it.

I don't see any reason `meta` tags should not be able to be added later, so I would not do that right now.

src/Sulu/Component/Rest/ListBuilder/Metadata/Resources/schema/list-2.0.xsd
danrot: And it has already caused problems. Sometimes just value would make more sense, especially in the context of a collection. Therefore I would not start with the same mistake again.

src/Sulu/Component/Rest/ListBuilder/Metadata/ListXmlLoader.php
danrot: In that case I would say we don't do it. Defining multiple filters would break our entire UI concept, so that probably won't happen before a major release anyway, resp. only by using some kind of BC layer.

src/Sulu/Bundle/TagBundle/Resources/config/lists/tags.xml
danrot: Because it seems that this is the only file where this property exists :slightly_smiling_face:

src/Sulu/Bundle/TagBundle/Resources/config/lists/tags.xml
danrot: Because it seems that this is the only file where this property exists :slightly_smiling_face:

src/Sulu/Component/Rest/ListBuilder/Metadata/Resources/schema/list-2.0.xsd
nnatter: yes the `meta` tag can be added later. i just figured that this might be the reason for the `params` tag in the form xmls. not sure if i like the `params` tag there, but i think we should keep things consistent? 🤔

src/Sulu/Component/Rest/ListBuilder/Metadata/Resources/schema/list-2.0.xsd
nnatter: did not know about that 🙂 in which case does it make sense to have just a value?

src/Sulu/Component/Rest/ListBuilder/Metadata/Resources/schema/list-2.0.xsd
danrot: Ok, consistency argument wins, will add it.

src/Sulu/Component/Rest/ListBuilder/Metadata/Resources/schema/list-2.0.xsd
danrot: e.g. something like this:

```xml
<param name="dropdown-options" type="collection">
    <param value="1" />
    <param value="2" />
    <!-- ... -->
</param>
```

src/Sulu/Component/Rest/ListBuilder/Metadata/ListXmlLoader.php
```diff
@@ -225,17 +231,29 @@ private function loadConcatenationPropertyMetadata(\DOMXPath $xpath, \DOMNode $p
...
-            $key = XmlUtil::getValueFromXPath('@key', $xpath, $parameterNode);
-            $parameters[$key] = $this->parameterBag->resolveValue(trim($parameterNode->nodeValue));
+        foreach ($xpath->query('x:param', $filterNode) as $paramNode) {
+            $name = XmlUtil::getValueFromXPath('@name', $xpath, $paramNode);
+            $type = XmlUtil::getValueFromXPath('@type', $xpath, $paramNode);
+
+            if ('collection' === $type) {
+                $parameters[$name] = $this->getFilterTypeParameters($xpath, $paramNode);
+            } else {
+                $parameters[$name] = $this->parameterBag->resolveValue(
```
nnatter: i this will behave kind of strange if the `name` attribute of a param is not set

src/Sulu/Component/Rest/ListBuilder/Metadata/Resources/schema/list-2.0.xsd
nnatter: okay i think we need to adjust the logic of the `ListXmlLoader` a bit then?

src/Sulu/Bundle/AudienceTargetingBundle/Resources/config/lists/target_groups.xml
```diff
@@ -3,37 +3,40 @@
...
+        <property name="id" translation="sulu_admin.id" type="string">
             <field-name>id</field-name>
             <entity-name>%sulu.model.target_group.class%</entity-name>
         </property>
-        <property name="title" filter-type="string" translation="sulu_admin.title" visibility="always">
+        <property name="title" translation="sulu_admin.title" visibility="always">
             <field-name>title</field-name>
             <entity-name>%sulu.model.target_group.class%</entity-name>
+
+            <filter type="text" />
```
nnatter: not sure if it makes sense to provide a filter for this

src/Sulu/Bundle/ContactBundle/Resources/config/lists/accounts.xml
```diff
@@ -112,42 +116,44 @@
...
             name="addressLine"
-            filter-type="string"
             translation="sulu_contact.address_line"
             sortable="false"
             glue=" "
         >
             <field property-ref="street"/>
             <field property-ref="streetNumber"/>
+
+            <filter type="text" />
```
nnatter: i think you removed the filter for this in the `contacts.xml`

src/Sulu/Bundle/SecurityBundle/Resources/config/lists/roles.xml
```diff
@@ -21,6 +21,8 @@
         >
             <field-name>name</field-name>
             <entity-name>%sulu.model.role.class%</entity-name>
+
+            <filter type="text" />
```
nnatter: not sure if it makes sense to provide a filter for this

Issue #5090, opened 2020-03-04
Blocked file types and max file size are never set


tvwijgerden:
| Q | A
| --- | ---
| Sulu Version | 1.6.x & 2.0.x
| Browser Version | -

# Actual Behavior

Setting the `sulu_media.format_manager.blocked_file_types` and `sulu_media.upload.max_filesize` configuration has no effect.

# Expected Behavior

Setting the `sulu_media.format_manager.blocked_file_types` configuration should block the configured file types.

Setting the`sulu_media.upload.max_filesize` configuration should limit the file size to the configured max file size

# Steps to Reproduce

Add `application/pdf` to the `blocked_file_types` configuration and upload a pdf file.

# Possible Solutions

The `FileValidator` class contains a `setBlockedMimeTypes` method and a `setMaxFileSize` method. These methods are never called, so max file size and blocked mime types are never set.

A possible solution would be calling these methods in the `SuluMediaExtension` and passing the correct configuration option for both methods. 

If you want to see this changed I can make a PR for it. If I want to make the changes for both version 1.6 and 2.0, should I make two PR's or how would I do that?

danrot:
@tvwijgerden Since this bug also appears in Sulu 1.6, you should make the change in the `release/1.6` branch. We will take care of merging the changes to 2.0 afterwards. In general you should always fix a bug in the oldest supported version (currently 1.6), in which the bug appears.

danrot:
Fixed in #5096

PR #5096, opened 2020-03-09
Fix FileValidator class not being configured

tvwijgerden:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | #5090
| Related issues/PRs | #5090
| License | MIT
| Documentation PR | -

#### What's in this PR?

Configures the `FileValidator` class by calling the setter methods and with the correct configuration.

Also changes the `max_filesize` configuration option to be a `string` instead of an `integer`. Because that's what the `FileValidator` class expects`


src/Sulu/Bundle/MediaBundle/DependencyInjection/Configuration.php
```diff
@@ -73,10 +73,7 @@ public function getConfigTreeBuilder()
             ->arrayNode('upload')
                 ->addDefaultsIfNotSet()
                 ->children()
-                    ->integerNode('max_filesize')
-                        ->defaultValue(256)
-                        ->min(0)
-                    ->end()
+                    ->scalarNode('max_filesize')->defaultValue('256MB')->end()
```
alexander-schranz: For BC we should not change this type and keep this file as before

src/Sulu/Bundle/MediaBundle/DependencyInjection/SuluMediaExtension.php
```diff
@@ -364,6 +365,15 @@ private function configureStorage(array $config, ContainerBuilder $container, Lo
         $container->setAlias('sulu_media.storage', 'sulu_media.storage.' . $storage)->setPublic(true);
     }
 
+    private function configureFileValidator(array $config, ContainerBuilder $container)
+    {
+        $definition = $container->getDefinition('sulu_media.file_validator');
+        $definition->addMethodCall('setMaxFileSize', [$config['upload']['max_filesize']]);
```
alexander-schranz: Add `MB` here

alexander-schranz:
@tvwijgerden as this is a bugfix for 2.0 your branch should be rebased to `release/2.0` branch.

danrot:
@alexander-schranz Couldn't this PR be set against `release/1.6`? I think the part we are seeing here should be the same for that release as well, shouldn't it?

src/Sulu/Bundle/MediaBundle/DependencyInjection/Configuration.php
```diff
@@ -105,7 +105,7 @@ public function getConfigTreeBuilder()
                         ->prototype('scalar')->end()
                     ->end()
                     ->arrayNode('blocked_file_types')
-                        ->prototype('scalar')->end()->defaultValue(['file/exe'])
+                        ->prototype('scalar')->end()->defaultValue([])
```
alexander-schranz: This was never validated so from BC I will remove the default value with file/exe as I know there exist sulu projects where they have downloadable executable which I don't want to break when they update.

src/Sulu/Bundle/MediaBundle/DependencyInjection/SuluMediaExtension.php
```diff
@@ -119,7 +119,6 @@ public function load(array $configs, ContainerBuilder $container)
         $container->setParameter('sulu_media.ghost_script.path', $config['ghost_script']['path']);
 
         // storage
-        $container->setParameter('sulu_media.media.max_file_size', '16MB');
```
alexander-schranz: strange parameter injected into the MediaManager but never used 🙈 set it to the same value as `sulu_media.upload.max_filesize` beneath

danrot: Any reason to keep it at all then? Do you want to keep it for BC reasons? If yes I would at least add some comment in here, so that we remember removing it at some point.

PR #5099, opened 2020-03-10
Fix switching of template with different block types

### LINKED ISSUE #5098: Switching template with different block types crash Admin UI
| Q | A
| --- | ---

# Actual Behavior

If you have a `blocks` property in 2 different templates and the template have different block types the admin ui will crash on template change because it can not find the old template type.

# Expected Behavior

Instead of crashin it should display the default type because the same also happens if a developer removes a type from the definition.

# Steps to Reproduce

<details>
<summary>Template A</summary>

```xml
<?xml version="1.0" ?>

    <key>case-studies</key>

    <meta>
        <title lang="de">Case studies</title>
        <title lang="en">Case studies</title>
    </meta>

    <properties>
            <meta>
                <title lang="de">Titel</title>
            </meta>

            <tag name="sulu.node.name"/>
        </property>

        <block name="blocks" default-type="case-studies">
            <meta>
                <title lang="de">Blöcke</title>
                <title lang="en">Blocks</title>
            </meta>

            <types>
                <type name="case-studies">
                    <properties>
                        <property name="title" type="text_line">
                            <meta>
                                <title lang="de">Titel</title>
                            </meta>
                        </property>
                    </properties>
                </type>
            </types>
        </block>
    </properties>
</template>
```

</details>

<details>
<summary>Template B</summary>

```xml
<?xml version="1.0" ?>

    <key>features</key>

    <meta>
        <title lang="en">Features</title>
        <title lang="de">Eigenschaften</title>
    </meta>

    <properties>
            <meta>
                <title lang="de">Titel</title>
            </meta>

            <tag name="sulu.node.name"/>
        </property>

        <block name="blocks" default-type="features">
            <meta>
                <title lang="de">Blöcke</title>
                <title lang="en">Blocks</title>
            </meta>

            <types>
                <type name="features">
                    <properties>
                        <property name="title" type="text_line">
                            <meta>
                                <title lang="de">Titel</title>
                            </meta>
                        </property>
                    </properties>
                </type>
            </types>
        </block>
    </properties>
</template>
```

</details>

1. Add block to template A
2. Switch template to B

# Possible Solutions

Check if type definition exist else fallback to the default block type.

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #5098 
| Related issues/PRs | -
| License | MIT

#### What's in this PR?

Fix switching of block types.

#### Why?

The admin UI should not crash if a block type was removed by the developer or switching to a template without the exist block type.


src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
```diff
@@ -81,22 +108,7 @@ export default class FieldBlocks extends React.Component<FieldTypeProps<Array<Bl
...
     // eslint-disable-next-line no-unused-vars
     renderCollapsedBlockContent = (value: Object, type: string, index: number) => {
-        if (!type) {
-            throw new Error(
-                'It is impossible that a collapsed block has no type. This should not happen and is likely a bug.'
-            );
-        }
-
-        const {formInspector, schemaPath} = this.props;
-        const blockSchemaTypes = formInspector.getSchemaEntryByPath(schemaPath).types;
```
alexander-schranz: @danrot do you know why here was used `formInspector.getSchemaEntryByPath(schemaPath).types` instead of `this.props.types`?

danrot:
I have just tested it with a more complicated template setup:

<details>
<summary>Template Case studies</summary>

```xml
                                                                                                                                                                                                           
    <key>case-studies</key>                                                                                                                                                                                
    <view>pages/default</view>                                                                                                                                                                             
    <controller>Sulu\Bundle\WebsiteBundle\Controller\DefaultController::indexAction</controller>                                                                                                           
                                                                                                                                                                                                           
    <meta>                                                                                                                                                                                                 
        <title lang="de">Case studies</title>                                                                                                                                                              
        <title lang="en">Case studies</title>
    </meta>

    <properties>
            <meta>
                <title lang="de">Titel</title>
            </meta>

            <tag name="sulu.node.name"/>
        </property>

        <property name="url" type="resource_locator" mandatory="true">
            <meta>
                <title lang="de">Adresse</title>
            </meta>

            <tag name="sulu.rlp"/>
        </property>

        <block name="blocks" default-type="case-studies">
            <meta>
                <title lang="de">Blöcke</title>
                <title lang="en">Blocks</title>
            </meta>

            <types>
                <type name="case-studies">
                    <properties>
                        <property name="title" type="text_line">
                            <meta>
                                <title lang="de">Case Study Titel</title>
                                <title lang="en">Case Study Title</title>
                            </meta>
                        </property>
                    </properties>
                </type>
                <type name="test2">
                    <properties>
                        <property name="title" type="text_line">
                            <meta>
                                <title lang="de">Test2 Titel</title>
                                <title lang="en">Test2 Title</title>
                            </meta>
                        </property>
                    </properties>
                </type>
            </types>
        </block>
    </properties>
</template>
```
</details>

<details>
<summary>Template Features</summary>

```xml
<?xml version="1.0" ?>                                                                                                                                                                                     
                                                                                                                                                                                                           
    <key>features</key>                                                                                                                                                                                    
    <view>pages/default</view>                                                                                                                                                                             
    <controller>Sulu\Bundle\WebsiteBundle\Controller\DefaultController::indexAction</controller>                                                                                                           
                                                                                                                                                                                                           
    <meta>                                                                                                                                                                                                 
        <title lang="en">Features</title>                                                                                                                                                                  
        <title lang="de">Eigenschaften</title>                                                                                                                                                             
    </meta>                                                                                                                                                                                                
                                                                                                                                                                                                           
    <properties>                                                                                                                                                                                           
            <meta>                                                                                                                                                                                         
                <title lang="de">Titel</title>                                                                                                                                                             
            </meta>                                                                                                                                                                                        
                                                                                                                                                                                                           
            <tag name="sulu.node.name"/>                                                                                                                                                                   
        </property>                                                                                                                                                                                        
                                                                                                                                                                                                           
        <property name="url" type="resource_locator" mandatory="true">                                                                                                                                     
            <meta>                                                                                                                                                                                         
                <title lang="de">Adresse</title>                                                                                                                                                           
            </meta>                                                                                                                                                                                        
                                                                                                                                                                                                           
            <tag name="sulu.rlp"/>                                                                                                                                                                         
        </property>                                                                                                                                                                                        
                                                                                                                                                                                                           
        <block name="blocks" default-type="features">                                                                                                                                                      
            <meta>                                                                                                                                                                                         
                <title lang="de">Blöcke</title>                                                                                                                                                            
                <title lang="en">Blocks</title>                                                                                                                                                            
            </meta>                                                                                                                                                                                        
                                                                                                                                                                                                           
            <types>                                                                                                                                                                                        
                <type name="features">                                                                                                                                                                     
                    <properties>                                                                                                                                                                           
                        <property name="title" type="text_line">                                                                                                                                           
                            <meta>                                                                                                                                                                         
                                <title lang="de">Feature Titel</title>                                                                                                                                     
                                <title lang="en">Feature Title</title>                                                                                                                                     
                            </meta>                                                                                                                                                                        
                        </property>                                                                                                                                                                        
                    </properties>                                                                                                                                                                          
                </type>                                                                                                                                                                                    
                <type name="test">                                                                                                                                                                         
                    <properties>                                                                                                                                                                           
                        <property name="title" type="text_line">                                                                                                                                           
                            <meta>                                                                                                                                                                         
                                <title lang="de">Test Titel</title>                                                                                                                                        
                                <title lang="en">Test Title</title>                                                                                                                                        
                            </meta>                                                                                                                                                                        
                        </property>                                                                                                                                                                        
                    </properties>                                                                                                                                                                          
                </type>                                                                                                                                                                                    
            </types>                                                                                                                                                                                       
        </block>                                                                                                                                                                                           
    </properties>                                                                                                                                                                                          
</template>
```
</details>

With these templates, whereby each of them consists of two different types, two problems occur with your solution when creating a page with the `case-studies` template, adding a block and switching to the `feature` template:

1. The type select will show "Please choose"
2. When trying to save the form without changing the select, the form validation will fail (because a type is selected behind the scenes, that does not exist for the block on the current template).

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
```diff
@@ -13,6 +13,39 @@ const MISSING_BLOCK_ERROR_MESSAGE = 'The "block" field type needs at least one t
 const BLOCK_PREVIEW_TAG = 'sulu.block_preview';
 
 export default class FieldBlocks extends React.Component<FieldTypeProps<Array<BlockEntry>>> {
+    componentDidUpdate(prevProps: FieldTypeProps<Array<BlockEntry>>) {
+        const {defaultType, types, value: oldValues, onChange} = this.props;
+        const {types: oldTypes} = prevProps;
+
+        if (!oldValues) {
+            return;
```
danrot: I would not use early returns in here. The problem is that this gets very complicated, if there is a second prop changing, that you want to handle.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
```diff
@@ -13,6 +13,46 @@ const MISSING_BLOCK_ERROR_MESSAGE = 'The "block" field type needs at least one t
...
+        }
+
+        const newValues = toJS(oldValues);
+        const usedTypes = newValues.map((block) => block.type);
+
+        // do nothing when used types where included in oldTypes
+        if (usedTypes.every((type) => Object.keys(oldTypes).includes(type))) {
+            return;
+        }
```
alexander-schranz: @danrot this could be omited by doing something like beneath:

```
        let hasChanged = false;

        newValues.forEach((blockValue, i) => {
            if (!types[blockValue.type]) {
                hasChanged = true;
                newValues[i].type = defaultType;
            }
        });

        if (hasChanged) {
            onChange(newValues)
        }
```

what do you like better?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
```diff
@@ -13,6 +13,46 @@ const MISSING_BLOCK_ERROR_MESSAGE = 'The "block" field type needs at least one t
...
+            throw new Error(MISSING_BLOCK_ERROR_MESSAGE);
+        }
+
+        // do nothing when types is same reference as oldTypes
+        if (types === oldTypes) {
+            return;
+        }
+
+        const newValues = toJS(oldValues);
+        const usedTypes = newValues.map((block) => block.type);
```
danrot: Distinguishing between `oldValues` and `newValues` is a little bit confusing here IMO. At this point both of them are exactly the same, right? I would rename the `oldValues` to just `value` (then it is clearer what they are) and get the `usedTypes` from the `value`. Then you could do the below checks exactly the same, and change the value by calling `onChange` somehow like this:

```javascript
onChange(value.map((blockValue) => ({
    ...blockValue,
    type: types[blockValue.type] ? types[blockValue.type].type : defaultType,
})));
```

That woud work around the thing that is IMO a naming issue.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
```diff
@@ -31,6 +71,38 @@ export default class FieldBlocks extends React.Component<FieldTypeProps<Array<Bl
...
+            );
+        }
+
+        if (!types[defaultType]) {
+            throw new Error(
+                'The default type should exist in block "' + schemaPath + '".'
+            );
+        }
+
+        return types[defaultType];
```
danrot: Do we still need the default type handling here? Is that because it will be rendered once before `componentDidUpdate` is called?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/tests/FieldBlocks.test.js
```diff
@@ -384,6 +384,239 @@ test('Render block with schema', () => {
...
+    const fieldBlocks = shallow(
+        <FieldBlocks
+            {...fieldTypeDefaultProps}
+            {...oldProps}
+            formInspector={formInspector}
+            onChange={changeSpy}
+        />
+    );
+
+    const newProps = {
```
danrot: Can't we just inline this to the `setProps` call? Would be more straight forward to read, and making the test easier to understand.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
alexander-schranz: @danrot yes we still need this here it will fail without it.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
danrot: Personally I prefer solutions without `let` :see_no_evil: And as written above I would turn that statement around, to avoid a early return in `componentDidUpdate`:

```javascript
if (!usedTypes.every((type) => Object.keys(oldTypes).includes(type))) {
    // Do all the required stuff when types are not available
}
```

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
```diff
@@ -13,6 +13,40 @@ const MISSING_BLOCK_ERROR_MESSAGE = 'The "block" field type needs at least one t
 const BLOCK_PREVIEW_TAG = 'sulu.block_preview';
 
 export default class FieldBlocks extends React.Component<FieldTypeProps<Array<BlockEntry>>> {
+    componentDidUpdate(prevProps: FieldTypeProps<Array<BlockEntry>>) {
+        const {defaultType, types, value: value, onChange} = this.props;
```
danrot: Could be written as just `value`, not `value: value`.

danrot: Maybe we could even include the [eslint rule for that](https://eslint.org/docs/rules/no-useless-rename)?

danrot: oh, and please sort it alphabetically :slightly_smiling_face:

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
```diff
@@ -13,6 +13,40 @@ const MISSING_BLOCK_ERROR_MESSAGE = 'The "block" field type needs at least one t
...
+        const {types: oldTypes} = prevProps;
+
+        if (!types || !oldTypes) {
+            throw new Error(MISSING_BLOCK_ERROR_MESSAGE);
+        }
+
+        const newValue = toJS(value);
+        let hasValueChanged = false;
```
danrot: Is there a reason you have done it now in a different way than I suggested in [my comment](https://github.com/sulu/sulu/pull/5099#discussion_r394349791)? I.e. did you have to use the `let` here now?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
```diff
@@ -13,6 +13,40 @@ const MISSING_BLOCK_ERROR_MESSAGE = 'The "block" field type needs at least one t
...
+        if (value && types !== oldTypes) {
+            if (!defaultType) {
+                throw new Error(
+                    'It is impossible that a block has no defaultType. This should not happen and is likely a bug.'
+                );
+            }
+
+            // set block to default type if type does not longer exist
+            // this could happen for example in a template switch
+            newValue.forEach((block, i) => {
```
danrot: This part is also different from what I've described in [another comment](https://github.com/sulu/sulu/pull/5099#discussion_r394340140).

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
```diff
@@ -13,6 +13,40 @@ const MISSING_BLOCK_ERROR_MESSAGE = 'The "block" field type needs at least one t
...
+                if (!types[block.type]) {
+                    hasValueChanged = true;
+                    newValue[i].type = defaultType;
+                }
+            });
+        }
+
+        // onChange should only be called when value was changed else it will end in a infinite loop
```
danrot: Why isn't that done in the `if` block above? Then it would only be called if the block types changed, and this entire logic does not have to be executed on every update.

Actually I would say it is a best practice to first check if props (or whatever you are listening to) as soon as possible. E.g. also the [react documentation](https://reactjs.org/docs/react-component.html#componentdidupdate) makes a check as soon as possible.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
alexander-schranz: First to make it easier extend so if somebody change in componentDidUpdate the value it just need to set that boolean flag to true to call onChange and to avoid a second loop over all blocks.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
alexander-schranz: same as above I think if we change the value in componentDidUpdate we should call onChange only once and thats onChange is not hear and why I keep the forEach to set the boolean flag

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
alexander-schranz: as above think `onChange` should only be called once in `componentDidUpdate` so if somebody later add something here to change the value he should insert that code before this call. If I do it directly in if I think if somebody adds here something `onChange` is maybe called twice instead of once.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
alexander-schranz: rule added and sorted alphabetically :)

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
danrot: Ok, that makes sense, especially since if we end up with two `onChange` calls in here one would override the other changes.

But couldn't we do this without this variable? I would just check if the old value is the same as the new value using `equals` as we do in other places. With this implementation the `hasValueChanged` flag might have been set although the `value` has not changed. If we keep all of that in a single variable, it is guaranteed to be correct.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
danrot: Ok, but as described above I would compare the old value with the new one, instead of introducing a new variable. Would also have the advantage that it would not call the `onChange` handler if a first block changes the value, and the second block reverts this changes for some reason.

I also think it is not necessary to use the `i` variable in here, because you should be able to just use `block.type = defaultType` below, without the need of accessing the array via the index.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
```diff
@@ -13,6 +14,41 @@ const MISSING_BLOCK_ERROR_MESSAGE = 'The "block" field type needs at least one t
...
+                throw new Error(
+                    'It is impossible that a block has no defaultType. This should not happen and is likely a bug.'
+                );
+            }
+
+            // set block to default type if type does not longer exist
+            // this could happen for example in a template switch
+            newValue = newValue.map((block) => {
+                if (!types[block.type]) {
+                    return {...block, type: defaultType}
```
alexander-schranz: @danrot need to copy the block else the equals check will fail as newValue blocks and oldValue blocks is the same object

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
alexander-schranz: In case of performance I don't wanted to clone the whole object on every componentDidUpdate run, also when nothing necessary changed so I did go with only cloning the block object which where changed inside the `.map` function hope that solution match your criterias 🙂.

PR #5100, opened 2020-03-10
Fix display column options in List container even when searching is disabled

nnatter:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This PR adjusts the `List` container to still render the column-options dropdown when searching is disabled for the `List` container.

#### Why?

Because it should still be possible to change the displayed columns even if the list is not searchable.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/List.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/List.js
@@ -502,7 +502,7 @@ class List extends React.Component<Props> {
         return (
             <div className={listStyles.listContainer}>
                 {header}
-                {!store.schemaLoading && (searchable || adapters.length > 1) &&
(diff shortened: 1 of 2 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/List.js
```diff
@@ -502,7 +502,7 @@ class List extends React.Component<Props> {
         return (
             <div className={listStyles.listContainer}>
                 {header}
-                {!store.schemaLoading && (searchable || adapters.length > 1) &&
+                {!store.schemaLoading && (searchable || this.currentAdapter.hasColumnOptions || adapters.length > 1) &&
```
alexander-schranz: missing a test for this

nnatter: 😢

PR #4774, opened 2019-10-01
Rename routes to views

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR renames everything route related to views.

#### Why?

Because we think it is easier to explain this way.

#### BC Breaks/Deprecations

Yes

PR #5101, opened 2020-03-11
Add sticky locale for user

### LINKED ISSUE #5080: The chosen language should persist if the backend view changes
# Problem description

If I enter the sulu.rocks backoffice and klick on "webspaces" the default language will be "en". If I change that to "de", click on a content, edit and save the content and use the "go back" button it will stay "de". But if I click on webspaces again or click for example on "snippets" the language will be reset to "en" without any further notification.

# Proposed solution

Keep the language for everything until the user actively changes it. Or at least give the content manager a noticeable warning that the default language is used.

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #5080
| Related issues/PRs | 
| License | MIT
| Documentation PR | -

#### What's in this PR?

~~Similiar to the symfony router I added a router context. So as defaultAttributes first the router context locale is used.~~ As we don't implement which locales are allowed for current route into the route path regular expression we need todo the check on the locales attribute ourself.

#### Why?

If the locale is changed and the user clicks on the menu item it should try to use the same locale as the user was before. 

#### To Do


src/Sulu/Bundle/AdminBundle/Admin/View/FormViewBuilderTrait.php
```diff
@@ -50,6 +50,10 @@ private function addLocalesToView(View $view, array $locales): void
         $oldLocales = $view->getOption('locales');
         $newLocales = $oldLocales ? array_merge($oldLocales, $locales) : $locales;
         $view->setOption('locales', $newLocales);
+
+        if (!$view->getAttributeDefault('locale') && isset($newLocales[0])) {
+            $view->setAttributeDefault('locale', $newLocales[0]);
```
danrot: Feels like a sensible default to me :+1:

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Application/Application.js
```diff
@@ -69,8 +69,8 @@ class Application extends React.Component<Props>{
 
     navigationPinnedDisposer: () => void;
 
-    constructor() {
-        super();
+    constructor(props) {
+        super(props);
```
danrot: :scream: Did that also make any actual problems?

src/Sulu/Bundle/AdminBundle/Resources/js/index.js
```diff
@@ -122,6 +122,21 @@ const FIELD_TYPE_TEXT_LINE = 'text_line';
...
+const router = new Router(createHashHistory());
+
+router.addUpdateRouteHook((newRoute, newAttributes) => {
+    if (newAttributes.locale) {
+        router.setContext('locale', newAttributes.locale);
+        userStore.updateContentLocale(newAttributes.locale);
+    }
+
+    if (newAttributes.webspace) {
+        router.setContext('webspace', newAttributes.webspace);
```
danrot: I don't think that this is necessary, because the webspace is already correctly remembered. Also, I would prefer not to couple the `AdminBundle` to the locale or webspace...

src/Sulu/Bundle/AdminBundle/Resources/js/stores/userStore/userStore.js
```diff
@@ -60,13 +61,26 @@ class UserStore {
...
-            this.contentLocale = fallbackLocalization ? fallbackLocalization.locale : this.contentLocale;
-        }));
+        const contentLocale = this.persistentSettings.get(CONTENT_LOCALE_SETTING_KEY);
+
+        if (contentLocale) {
+            this.contentLocale = contentLocale;
+        }
+
+        if (!this.contentLocale) {
+            // TODO this code should be adjusted/removed when a proper content-locale handling is implemented
```
danrot: Should we maybe remove the comment? What better approach of defining a `contentLocale` at the very beginning do we have than the code below?

src/Sulu/Bundle/PageBundle/Resources/js/views/PageList/PageList.js
```diff
@@ -27,7 +28,7 @@ type Props = ViewProps & {
 @observer
 class PageList extends React.Component<Props> {
     page: IObservableValue<number> = observable.box();
-    locale: IObservableValue<string> = observable.box();
+    locale: IObservableValue<string> = observable.box(userStore.contentLocale);
```
danrot: Is that necessary? The way I think about this, this will never apply, because the default value will be used.

src/Sulu/Bundle/AdminBundle/Resources/js/services/Router/Router.js
```diff
@@ -163,6 +164,10 @@ export default class Router {
         this.handleNavigation(name, attributes, this.redirect);
     };
 
+    @action setContext = (name: string, value: any): void => {
+        this.context[name] = value;
```
danrot: I don't think that I would add another feature to the router, that makes stuff even more complicated than it already is. Especially since there is already a `getDerivedRouteAttributes` feature in the router, that would also be suitable to solve this problem. This is how we e.g. [already successfully set the default webspace](https://github.com/sulu/sulu/blob/edbf2d5/src/Sulu/Bundle/PageBundle/Resources/js/views/WebspaceTabs/WebspaceTabs.js#L26-L32) or [other attributes in the `List` view](https://github.com/sulu/sulu/blob/edbf2d56fd7843ee6775ee47a5a0e394bb4fe708/src/Sulu/Bundle/AdminBundle/Resources/js/views/List/List.js#L39-L55).

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Application/Application.js
alexander-schranz: Didn't make any problems did try many things today and found out that this was missing here 😄

src/Sulu/Bundle/PageBundle/Resources/js/views/PageList/PageList.js
alexander-schranz: The webspace default value is only used when the current `this.locale` is not available in the current webspace.

Example I have an additional webspace:

```
example2
  nl (default)
  fr
```

Now I go to example2 switch to fr. When I now relogin (removing hash url) it was again `nl` instead of `fr` but if I use userStore.contentLocale as default value it loads correctly the fr locale. So the defaults set by the developer are now only fallbacks if never a contentLocale was set.

src/Sulu/Bundle/AdminBundle/Resources/js/services/Router/Router.js
alexander-schranz: @danrot the context I introduced because it did feel a little bit strange importing the `userStore` into the `Router.js` which is needed when you navigate to a route without a locale (e.g. Contact) `attributes.locale` will be empty and needed to fallback to user settings content locale `userStore.contentLocale` when navigate again to a route with a locale, but if its okay for you I can import the userStore into the Router.js.

src/Sulu/Bundle/AdminBundle/Resources/js/index.js
alexander-schranz: sounds reasonable 👍

src/Sulu/Bundle/AdminBundle/Resources/js/stores/userStore/userStore.js
alexander-schranz: I think we should use something which does not run `async` as currently this could end in a race condition when localizations are still loaded.

I think the best default value would be the first language which I have permission for, so the first available language of the first userRole.

Maybe we introduce a `getContentLocale` method on the User object itself as it could be interesting I think for PHP aswell or what do you think?

e.g.:

```diff
diff --git a/src/Sulu/Bundle/AdminBundle/Resources/js/stores/userStore/types.js b/src/Sulu/Bundle/AdminBundle/Resources/js/stores/userStore/types.js
index b2a3577bf0..0bff12df17 100644
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/stores/userStore/types.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/stores/userStore/types.js
@@ -3,6 +3,7 @@
 export type User = {
     id: number,
     locale: string,
+    contentLocale: string,
     settings: {[string]: string},
     username: string,
 };
diff --git a/src/Sulu/Bundle/AdminBundle/Resources/js/stores/userStore/userStore.js b/src/Sulu/Bundle/AdminBundle/Resources/js/stores/userStore/userStore.js
index 0491739074..e7f1549484 100644
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/stores/userStore/userStore.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/stores/userStore/userStore.js
@@ -61,20 +61,7 @@ class UserStore {
             this.persistentSettings.set(key, persistentSettings[key]);
         });
 
-        const contentLocale = this.persistentSettings.get(CONTENT_LOCALE_SETTING_KEY);
-
-        if (contentLocale) {
-            this.contentLocale = contentLocale;
-        }
-
-        if (!contentLocale) {
-            // load and use first (default) localization of first webspace as content-locale for the user
-            localizationStore.loadLocalizations().then(action((localizations) => {
-                const defaultLocalizations = localizations.filter((localization) => localization.default);
-                const fallbackLocalization = defaultLocalizations.length ? defaultLocalizations[0] : localizations[0];
-                this.contentLocale = fallbackLocalization ? fallbackLocalization.locale : this.contentLocale;
-            }));
-        }
+        this.contentLocale = user.contentLocale;
     }
 
     @action updateContentLocale(contentLocale: string) {
diff --git a/src/Sulu/Bundle/SecurityBundle/Entity/User.php b/src/Sulu/Bundle/SecurityBundle/Entity/User.php
index 9f3820e1bd..fa1ab3e956 100644
--- a/src/Sulu/Bundle/SecurityBundle/Entity/User.php
+++ b/src/Sulu/Bundle/SecurityBundle/Entity/User.php
@@ -32,6 +32,8 @@ use Symfony\Component\Security\Core\User\UserInterface as SymfonyUserInterface;
  */
 class User extends ApiEntity implements UserInterface, Serializable, EquatableInterface
 {
+    public const USER_SETTINGS_KEY_CONTENT_LOCALE = 'sulu_admin.content_locale';
+
     /**
      * @var int
      * @Expose
@@ -678,6 +680,29 @@ class User extends ApiEntity implements UserInterface, Serializable, EquatableIn
         return $userSettingValues;
     }
 
+    /**
+     * @VirtualProperty
+     * @Groups({"frontend"})
+     */
+    public function getContentLocale(): string
+    {
+        $contentLocale = $this->getSettings()[self::USER_SETTINGS_KEY_CONTENT_LOCALE] ?? null;
+
+        if ($contentLocale) {
+            return $contentLocale;
+        }
+
+        foreach ($this->userRoles as $userRole) {
+            $locales = $userRole->getLocales();
+
+            if (isset($locales[0])) {
+                return $locales[0];
+            }
+        }
+
+        return $this->getLocale();
+    }
+
     /**
      * Set contact.
      *
```

danrot: I don't see how the current solution would end up in a race condition... The worst thing that can happen is that initially it loads without language, and then a second request happens, although I have never seen that happening.

However, I think it might still make sense to put this logic on the server, so that eventual other clients don't have to implement that as well. But I think we should get rid of this bug ASAP, and not discuss this now.

src/Sulu/Bundle/PageBundle/Resources/js/views/PageList/PageList.js
danrot: Okay, it probably also doesn't hurt, I was just curious.

src/Sulu/Bundle/AdminBundle/Resources/js/services/Router/Router.js
danrot: I don't really see how these things are connected... The fallback could still be handled by the `getDerivedRouteAttributes` (as said, that's what they are built for). And you could still use an `UpdateRouteHook` to update the `contentLocale`, although I think I would not do it the way you have handled it here. I would update the content locale manually in the `List` and `Form` views/containers (not sure currently where this is done) instead of trying to find a solution working magically. If somebody creates their own route containing a `locale` parameter used differently, that could lead to very confusing behavior, which might be very hard to debug.

src/Sulu/Bundle/AdminBundle/Resources/js/services/ContentLocaleUpdater/updateRouterAttributesFromUserStoreContentLocale.js
```diff
@@ -0,0 +1,32 @@
...
+
+    // do nothing when the route does not require a locale
+    if (!keyNames.includes('locale')) {
+        return attributes;
+    }
+
+    const locales = toJS(route.options.locales);
+
+    // set content locale if route accept the current content locale
+    if (locales && locales.includes(userStore.contentLocale)) {
```
danrot: As I've already said I don't like that a lot of assumptions are made here (there has to be a `locale` and `locales` parameters), and it conflicts when the these parameters are used in other views in a different way, resulting in weird behavior.

src/Sulu/Bundle/AdminBundle/Resources/js/services/ContentLocaleUpdater/updateUserStoreContentLocaleFromRouterAttributes.js
```diff
@@ -0,0 +1,17 @@
...
+import type {UpdateRouteHook} from '../Router/types';
+import userStore from '../../stores/userStore/userStore';
+
+const updateUserStoreContentLocaleFromRouterAttributes: UpdateRouteHook = function(newRoute, newAttributes) {
+    if (!newAttributes) {
+        return true;
+    }
+
+    if (newAttributes.locale) {
+        userStore.updateContentLocale(newAttributes.locale);
```
danrot: I don't like this automatic mechanism for the same reasons as mentioned above.

src/Sulu/Bundle/AdminBundle/Resources/js/stores/userStore/userStore.js
```diff
@@ -60,13 +61,25 @@ class UserStore {
...
-        // load and use first (default) localization of first webspace as content-locale for the user
-        localizationStore.loadLocalizations().then(action((localizations) => {
-            const defaultLocalizations = localizations.filter((localization) => localization.default);
-            const fallbackLocalization = defaultLocalizations.length ? defaultLocalizations[0] : localizations[0];
-            this.contentLocale = fallbackLocalization ? fallbackLocalization.locale : this.contentLocale;
-        }));
+        const contentLocale = this.persistentSettings.get(CONTENT_LOCALE_SETTING_KEY);
+
+        if (contentLocale) {
+            this.contentLocale = contentLocale;
```
danrot: early return after this line instead of two `if`s?

alexander-schranz: done

alexander-schranz:
@danrot I don't think that we will get around this automatic mechanism as the whole feature request is about automatically use the contentLocale and update it when a view change it.
I really don't see an alternative way of implementing this and I think if somebody creates a new view he should not take care of this.

src/Sulu/Bundle/AdminBundle/Resources/js/stores/userStore/userStore.js
```diff
@@ -69,6 +77,11 @@ class UserStore {
         }));
     }
 
+    @action updateContentLocale(contentLocale: string) {
+        this.contentLocale = contentLocale;
```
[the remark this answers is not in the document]
alexander-schranz: `contentLocale` can currently not be a computed as it is set as fallback by the localizationStore asynchronly. If we refractor that in #5119 we could change it to a computed value.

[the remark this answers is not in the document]
danrot: Then I would prefer to wait until #5119 is merged... We wanted to release that together anyway, and if we don't do this in this PR, I am afraid that we forget about it :see_no_evil: 

Would that be ok for you?

src/Sulu/Bundle/AdminBundle/Resources/js/stores/userStore/updateRouterAttributesFromUserStoreContentLocale.js
```diff
@@ -0,0 +1,36 @@
...
+
+    // do nothing when the route does not require a locale
+    if (!keyNames.includes('locale')) {
+        return attributes;
+    }
+
+    const locales = toJS(route.options.locales);
+
+    // set content locale if route accept the current content locale
+    // or does not specify which locales it accepts e.g.: PageList
```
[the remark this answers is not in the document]
alexander-schranz: Then the expected behaviour does not longer work. We need todo this here else you will land again on the default webspace locale everytime.

src/Sulu/Bundle/AdminBundle/Resources/js/stores/userStore/updateRouterAttributesFromUserStoreContentLocale.js
```diff
@@ -0,0 +1,36 @@
...
+    route: Route,
+    attributes: AttributeMap
+) {
+    // do nothing when locale is explicit set
+    if (attributes.locale) {
+        return attributes;
+    }
+
+    const keys = [];
+    pathToRegexp(route.path, keys);
```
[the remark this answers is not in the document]
alexander-schranz: @danrot was a little bit hard to update all tests but now regexp and availableAttributes are on the route available.

src/Sulu/Bundle/AdminBundle/Resources/js/stores/userStore/userStore.js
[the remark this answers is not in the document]
alexander-schranz: Is ok for me, would still be good if you could review the last changes as soon as you have some time for it.

src/Sulu/Bundle/AdminBundle/Resources/js/stores/userStore/updateRouterAttributesFromUserStoreContentLocale.js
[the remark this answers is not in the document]
danrot: I don't think this is the case if you add setting the locale to the `getDerivedRouteAttributes` method of the `PageList` instead of doing it here... But I know that you won't like that. I'll just keep that comment open, and test this part extensively at the end, to see if something is not working as expected.

src/Sulu/Bundle/AdminBundle/Resources/js/services/Router/registries/routeRegistry.js
```diff
@@ -23,6 +24,15 @@ class RouteRegistry {
                 ...routeConfig,
                 children: [],
                 parent: undefined,
+                get availableAttributes() {
```
danrot: Have you also tested if these getters are really cached? Will test it later as well, just to be sure.

src/Sulu/Bundle/AdminBundle/Resources/js/services/Router/tests/Router.test.js
```diff
@@ -280,14 +289,15 @@ test('Update observable attribute on route change', () => {
...
-        list: {
+        list: createRoute({
             name: 'list',
             type: 'list',
             path: '/list',
             attributeDefaults: {
                 exclude: true,
             },
-        },
+            availableAttributes: [],
```
danrot: You should not have to set this, right?

src/Sulu/Bundle/AdminBundle/Resources/js/services/Router/tests/Router.test.js
```diff
@@ -430,12 +440,13 @@ test('Navigate to route by removing parameters', () => {
...
 test('Navigate to route changing only search parameters', () => {
     routeRegistry.getAll.mockReturnValue({
-        page: {
+        page: createRoute({
             name: 'page',
             type: 'form',
             path: '/pages/:uuid',
             attributeDefaults: {},
-        },
+            availableAttributes: ['uuid'],
```
danrot: That should be done by the `createRoute` function, shouldn't it?

src/Sulu/Bundle/AdminBundle/Resources/js/services/Router/tests/Router.test.js
```diff
@@ -452,12 +463,13 @@ test('Navigate to route changing only search parameters', () => {
...
 test('Navigate to route by adding search parameters', () => {
     routeRegistry.getAll.mockReturnValue({
-        page: {
+        page: createRoute({
             name: 'page',
             type: 'form',
             path: '/pages/:uuid',
             attributeDefaults: {},
-        },
+            availableAttributes: ['uuid'],
```
danrot: Same as above

src/Sulu/Bundle/AdminBundle/Resources/js/services/Router/registries/routeRegistry.js
danrot: I know I am a little bit late to recognize this, but it seems I didn't think about the impact of this enough... Especially when I have a look at the tests, that's probably a nightmare to maintain, because we have to duplicate some information...

Wouldn't you agree that instead of this, it would start to make sense to have a dedicated `Route` class? This class could then handle the available attributes and the regexp, and we could reuse it also in the tests. There could be a static `fromRouteConfig` factory method on the route, which could also be reused in all the tests.

And sorry again, I know that implementing it like this was my idea :grimacing:

src/Sulu/Bundle/AdminBundle/Resources/js/services/Router/registries/routeRegistry.js
```diff
@@ -48,7 +34,7 @@ class RouteRegistry {
         });
     }
 
-    get(name: string): Route {
+    get(name: string): RouteType {
```
danrot: Doesn't that return a `Route` now?

src/Sulu/Bundle/AdminBundle/Resources/js/services/Router/tests/registries/RouteRegistry.test.js
```diff
@@ -10,11 +10,11 @@ test('Clear routes from RouteRegistry', () => {
     routeRegistry.addCollection([
         {
             name: 'route',
-            view: 'view',
             path: '/route',
             options: {},
             attributeDefaults: {},
             rerenderAttributes: [],
+            type: 'view',
```
danrot: Why didn't we keep `view`? What we are describing here is always a `view`, and it would not be a BC break that way.

src/Sulu/Bundle/AdminBundle/Resources/js/services/Router/types.js
```diff
@@ -1,15 +1,15 @@
...
     parent?: string,
     path: string,
-    rerenderAttributes: Array<string>,
-    view: string,
+    rerenderAttributes?: Array<string>,
+    type: string,
 |};
 
-export type Route = {|
+export interface Route {
```
danrot: Do we really need that interface?

src/Sulu/Bundle/AdminBundle/Resources/js/services/Router/tests/registries/RouteRegistry.test.js
alexander-schranz: This is incorrect the `/admin/config` request does not return `view` anymore since we refractor it and the `ViewRender` does render the [.type](https://github.com/sulu/sulu/blob/85f3a8fa62440e14d277724e20ca8b2e362758c1/src/Sulu/Bundle/AdminBundle/Resources/js/containers/ViewRenderer/ViewRenderer.js#L32)  the `.view` is never accessed its false in the RouteConfig type decleration.

src/Sulu/Bundle/AdminBundle/Resources/js/services/Router/types.js
alexander-schranz: We could remove it and use the `Route` class everywhere instead.

src/Sulu/Bundle/AdminBundle/Resources/js/services/Router/tests/registries/RouteRegistry.test.js
alexander-schranz: @danrot the flow RouteConfig type (which is only used in the routeRegistry) was forgotten to be refractored when we changed this in `/admin/config` api that `routes` return `type` instead of `view`.

src/Sulu/Bundle/AdminBundle/Resources/js/services/Router/Route.js
```diff
@@ -0,0 +1,48 @@
...
+        if (config.rerenderAttributes) {
+            this.rerenderAttributes = config.rerenderAttributes;
+        }
+
+        if (config.type) {
+            this.type = config.type;
+        }
+    }
+
+    @computed get availableAttributes(): Regexp {
```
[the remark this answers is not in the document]
alexander-schranz: correct should be `Array<string>`

src/Sulu/Bundle/AdminBundle/Resources/js/services/Router/Route.js
```diff
@@ -0,0 +1,48 @@
...
+
+        if (config.options) {
+            this.options = config.options;
+        }
+
+        if (config.rerenderAttributes) {
+            this.rerenderAttributes = config.rerenderAttributes;
+        }
+
+        if (config.type) {
```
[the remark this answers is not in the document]
alexander-schranz: you are right did remove it

src/Sulu/Bundle/AdminBundle/Resources/js/services/Router/registries/routeRegistry.js
alexander-schranz: Removed the interface so all which did use the `type Route` before use now the `class Route`

src/Sulu/Bundle/AdminBundle/Resources/js/services/Router/README.md
```diff
@@ -6,7 +6,7 @@ routeRegistry.addCollection([
     {
         name: 'sulu_contact.list',
         path: '/contacts',
-        view: 'sulu_admin.list',
+        type: 'sulu_admin.list',
```
alexander-schranz: This was forgotten to be changed in RouteConfig in #4774

src/Sulu/Bundle/AdminBundle/Resources/js/containers/ViewRenderer/tests/updateRouterAttributesFromView.test.js
```diff
@@ -1,7 +1,7 @@
 // @flow
 import updateRouterAttributesFromView from '../updateRouterAttributesFromView';
 import viewRegistry from '../registries/viewRegistry';
-import type {Route} from '../../../services/Router';
+import Route from '../../../services/Router/Route';
```
danrot: Use only `services/Router` instead of the direct file to import from another package.

src/Sulu/Bundle/AdminBundle/Resources/js/views/Form/Form.js
```diff
@@ -10,10 +10,11 @@ import PublishIndicator from '../../components/PublishIndicator';
 import {default as FormContainer, ResourceFormStore} from '../../containers/Form';
 import {withToolbar} from '../../containers/Toolbar';
 import type {ViewProps} from '../../containers/ViewRenderer';
-import type {AttributeMap, Route, UpdateRouteMethod} from '../../services/Router/types';
+import type {AttributeMap, UpdateRouteMethod} from '../../services/Router/types';
 import ResourceStore from '../../stores/ResourceStore';
 import CollaborationStore from '../../stores/CollaborationStore';
 import {translate} from '../../utils/Translator';
+import Route from '../../services/Router/Route';
```
danrot: Use `services/Router`.

src/Sulu/Bundle/AdminBundle/Resources/js/views/FormOverlayList/tests/FormOverlayList.test.js
```diff
@@ -9,8 +9,8 @@ import ResourceStore from '../../../stores/ResourceStore';
 import ResourceFormStore from '../../../containers/Form/stores/ResourceFormStore';
 import Form from '../../../containers/Form';
 import Router from '../../../services/Router';
-import type {Route} from '../../../services/Router';
 import Snackbar from '../../../components/Snackbar';
+import Route from '../../../services/Router/Route';
```
danrot: Use `services/Router`

src/Sulu/Bundle/AdminBundle/Resources/js/views/List/List.js
```diff
@@ -8,9 +8,9 @@ import equals from 'fast-deep-equal';
 import {default as ListContainer, ListStore} from '../../containers/List';
 import {withToolbar} from '../../containers/Toolbar';
 import type {ViewProps} from '../../containers/ViewRenderer';
-import type {Route} from '../../services/Router/types';
 import {translate} from '../../utils/Translator';
 import ResourceStore from '../../stores/ResourceStore';
+import Route from '../../services/Router/Route';
```
danrot: Use `services/Router`

src/Sulu/Bundle/AdminBundle/Resources/js/views/ResourceTabs/ResourceTabs.js
```diff
@@ -4,10 +4,10 @@ import {computed, observable, toJS} from 'mobx';
 import {observer} from 'mobx-react';
 import jexl from 'jexl';
 import Loader from '../../components/Loader';
-import type {Route} from '../../services/Router';
 import Tabs from '../../views/Tabs';
 import type {ViewProps} from '../../containers/ViewRenderer';
 import ResourceStore from '../../stores/ResourceStore';
+import Route from '../../services/Router/Route';
```
danrot: use `services/Router`

src/Sulu/Bundle/AdminBundle/Resources/js/views/Tabs/Tabs.js
```diff
@@ -5,8 +5,8 @@ import {autorun, computed} from 'mobx';
 import {observer} from 'mobx-react';
 import TabsComponent from '../../components/Tabs';
 import type {ViewProps} from '../../containers/ViewRenderer';
-import type {Route} from '../../services/Router';
 import {translate} from '../../utils/Translator';
+import Route from '../../services/Router/Route';
```
danrot: Use `services/Router`

src/Sulu/Bundle/AdminBundle/Resources/js/views/Tabs/tests/Tabs.test.js
```diff
@@ -3,7 +3,7 @@ import React from 'react';
 import {mount, render} from 'enzyme';
 import Router from '../../../services/Router';
 import Tabs from '../Tabs';
-import type {Route} from '../../../services/Router';
+import Route from '../../../services/Router/Route';
```
danrot: Use `services/Router`

src/Sulu/Bundle/PageBundle/Resources/js/views/PageList/PageList.js
```diff
@@ -7,9 +7,10 @@ import {List, ListStore, withToolbar} from 'sulu-admin-bundle/containers';
 import userStore from 'sulu-admin-bundle/stores/userStore/userStore';
 import type {Localization} from 'sulu-admin-bundle/stores';
 import type {ViewProps} from 'sulu-admin-bundle/containers';
-import type {AttributeMap, Route} from 'sulu-admin-bundle/services';
+import type {AttributeMap} from 'sulu-admin-bundle/services';
 import {translate} from 'sulu-admin-bundle/utils';
 import {CacheClearToolbarAction} from 'sulu-website-bundle/containers';
+import Route from 'sulu-admin-bundle/services/Router/Route';
```
danrot: Use `sulu-admin-bundle/services`

src/Sulu/Bundle/PageBundle/Resources/js/views/WebspaceTabs/WebspaceTabs.js
```diff
@@ -5,9 +5,10 @@ import type {IObservableValue} from 'mobx';
 import {observer} from 'mobx-react';
 import {Loader} from 'sulu-admin-bundle/components';
 import type {ViewProps} from 'sulu-admin-bundle/containers';
-import type {AttributeMap, Route} from 'sulu-admin-bundle/services';
+import type {AttributeMap} from 'sulu-admin-bundle/services';
 import {userStore} from 'sulu-admin-bundle/stores';
 import {Tabs} from 'sulu-admin-bundle/views';
+import Route from 'sulu-admin-bundle/services/Router/Route';
```
danrot: use `sulu-admin-bundle/services`

src/Sulu/Bundle/PageBundle/Resources/js/views/WebspaceTabs/tests/WebspaceTabs.test.js
```diff
@@ -3,7 +3,7 @@ import React from 'react';
 import {mount, render} from 'enzyme';
 import {Router} from 'sulu-admin-bundle/services';
 import {userStore} from 'sulu-admin-bundle/stores';
-import type {Route} from 'sulu-admin-bundle/services/Router';
+import Route from 'sulu-admin-bundle/services/Router/Route';
```
danrot: use `sulu-admin-bundle/services`

src/Sulu/Bundle/AdminBundle/Resources/js/stores/userStore/tests/updateRouterAttributesFromUserStoreContentLocale.test.js
```diff
@@ -0,0 +1,92 @@
+// @flow
+import updateRouterAttributesFromUserStoreContentLocale from '../updateRouterAttributesFromUserStoreContentLocale';
+import Route from '../../../services/Router/Route';
+
+let mockedContentLocale = undefined;
+
+jest.mock('../../../stores/userStore/userStore', () => ({
+    get contentLocale() {
+        return mockedContentLocale;
```
danrot: Did the old way stop working? If it is just about flow, which rightfully complains that you can't assign something to a getter, we used the // $FlowFixMe command to get rid of these warnings, since they don't apply here.

src/Sulu/Bundle/PageBundle/Resources/js/containers/Form/tests/fields/TeaserSelection.test.js
```diff
@@ -20,7 +20,13 @@ jest.mock('sulu-admin-bundle/containers/Form/FormInspector', () => jest.fn(funct
     this.locale = formStore.locale;
 }));
 
-jest.mock('sulu-admin-bundle/stores/userStore', () => ({}));
+let mockedContentLocale = 'en';
+
+jest.mock('sulu-admin-bundle/stores/userStore', () => ({
+    get contentLocale(): string {
+        return mockedContentLocale;
```
danrot: Did the old way stop working? If it is just about flow, which rightfully complains that you can't assign something to a getter, we used the // $FlowFixMe command to get rid of these warnings, since they don't apply here.

alexander-schranz: yeah it was only flow. Did now go with $FlowFixME

src/Sulu/Bundle/PageBundle/Resources/js/views/WebspaceTabs/tests/WebspaceTabs.test.js
```diff
@@ -1,13 +1,13 @@
 // @flow
 import React from 'react';
 import {mount} from 'enzyme';
-import {Router} from 'sulu-admin-bundle/services';
+import {Router, Route} from 'sulu-admin-bundle/services';
 import {userStore} from 'sulu-admin-bundle/stores';
-import Route from 'sulu-admin-bundle/services/Router/Route';
+
```
danrot: No new line here

alexander-schranz:
@danrot this problem did also exist before und was not introduced by this pull request. The problem here is the PageList handles the locale instead of the Webspace Tabs.

Think we need to move the handling of the locale also to the Webspace Tabs else always a double routing happen first to the default locale which does maybe not exist and then to the locale which exists in the webspace.

danrot:
Ok, just realized that the problem is lying else where, but the change I mentioned above still makes that more likely to happen...

But the root cause is that when we change the webspace select, all attributes except for the `webspaceKey` are kept, including the locale... And the `WebspaceTabs` cannot handle the locale, because not every child view of it has locales.

But what we can do is to avoid the [`this.locale.set` calls](https://github.com/sulu/sulu/pull/5101/files#diff-75291f7966bfac8603b92fd1ff8ecc6aR57). We could try to do a `router.redirect` instead. Then it would still apply the routing twice, but the first one would be removed from the history, making the back button work as expected.

alexander-schranz:
@danrot did refractor it as you mention it to use a redirect instead of update the route now it should behave as expected.

src/Sulu/Bundle/PageBundle/Resources/js/views/PageList/PageList.js
```diff
@@ -53,20 +54,28 @@ class PageList extends React.Component<Props> {
...
             );
         }
 
+        if (locale === this.locale.get()) {
+            return;
+        }
+
+        router.redirect(router.route.name, {...router.attributes, locale: locale});
+
         this.locale.set(locale);
```
danrot: This should not be necessary anymore, because the redirect before will change this value already.

src/Sulu/Bundle/PageBundle/Resources/js/views/PageList/PageList.js
```diff
@@ -39,11 +39,12 @@ class PageList extends React.Component<Props> {
     static getDerivedRouteAttributes(route: Route, attributes: AttributeMap) {
         return {
             active: ListStore.getActiveSetting(PAGES_RESOURCE_KEY, getUserSettingsKeyForWebspace(attributes.webspace)),
+            locale: userStore.contentLocale,
         };
     }
 
     @action setDefaultLocaleForWebspace = () => {
-        const {webspace} = this.props;
+        const {webspace, router} = this.props;
```
danrot: Sort alphabetically

src/Sulu/Bundle/PageBundle/Resources/js/views/PageList/PageList.js
```diff
@@ -39,11 +39,12 @@ class PageList extends React.Component<Props> {
     static getDerivedRouteAttributes(route: Route, attributes: AttributeMap) {
         return {
             active: ListStore.getActiveSetting(PAGES_RESOURCE_KEY, getUserSettingsKeyForWebspace(attributes.webspace)),
+            locale: userStore.contentLocale,
```
danrot: Seems like I didn't state clearly enough what I meant: Since the problem I've found was not based by the missing assignment here, it would have been ok for me to still set this in the `updateRouteAttributesFromUserStoreContentLocale` function. You can leave it here or move it back to the other function, whatever you prefer now, since it makes no difference.

src/Sulu/Bundle/PageBundle/Resources/js/views/PageList/PageList.js
```diff
@@ -39,11 +39,12 @@ class PageList extends React.Component<Props> {
...
         };
     }
 
     @action setDefaultLocaleForWebspace = () => {
```
danrot: Since the function should not set anything anymore, I would rename it to something like `redirectToWebspaceLocale`. Other suggestions are welcome :slightly_smiling_face:

src/Sulu/Bundle/PageBundle/Resources/js/views/PageList/PageList.js
```diff
@@ -53,20 +54,28 @@ class PageList extends React.Component<Props> {
             return;
         }
 
-        if (webspace.allLocalizations.find((localization) => localization.localization === userStore.contentLocale)) {
-            this.locale.set(userStore.contentLocale);
+        let locale = null;
```
danrot: Would prefer to make this a `const`, to avoid having troubles by somebody reassigning this.

```javascript
const locale = webspace.allLocalizations.find((localization) => localization.localization === userStore.contentLocale)
    ? userStore.contentLocale
    : this.findDefaultLocale(webspace.localizations);
```

src/Sulu/Bundle/PageBundle/Resources/js/views/PageList/PageList.js
```diff
@@ -51,14 +54,28 @@ class PageList extends React.Component<Props> {
...
             throw new Error(
                 'Default locale in webspace "' + webspace.key + '" not found'
             );
         }
 
+        if (locale === this.locale.get()) {
+            return;
+        }
+
+        router.redirect(router.route.name, {...router.attributes, locale: locale});
```
danrot: `locale` instead of `locale: locale`. Have we only added this rule in the `master` branch yet? :thinking:

src/Sulu/Bundle/PageBundle/Resources/js/views/PageList/PageList.js
alexander-schranz: Changed!

src/Sulu/Bundle/PageBundle/Resources/js/views/PageList/PageList.js
alexander-schranz: The `no-useless-rename` rule does only work for import, export and destructering there would be a additonal rule: https://eslint.org/docs/rules/object-shorthand

alexander-schranz: Own PR created #5135

danrot:
Thank you! And sorry again for the confusion :see_no_evil:

alexander-schranz:
No problem, now its working and merged  🎉 

git-merge

PR #5105, opened 2020-03-12
Fix ResourceRouteRegistry to accept boxed observable inside of parameters

nnatter:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### Why?

For example, the `SetUnpublishedToolbarAction` passes the current locale as boxed observable. Without this fix, this will lead to the following error: `Uncaught RangeError: Maximum call stack size exceeded`


alexander-schranz:
@nnatter can we link the PR here which did cause this error?

alexander-schranz:
@nnatter can we link the PR here which did cause this error?

nnatter:
Should be this one: https://github.com/sulu/sulu/pull/5091

PR #4798, opened 2019-10-08
Add compatibility to symfony 5

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | yes
| Deprecations? | no
| Fixed tickets | fixes -
| Related issues/PRs | see below
| License | MIT
| Documentation PR | -

#### What's in this PR?

Symfony 5.0 compatibility.

#### Why?

We should be compatible with newest major version of Symfony.

#### Example Usage

Check if we are fully symfony 5 compatible with:

```bash
composer require symfony/symfony:^5.0 --dev
```

And run tests with:

```bash
bin/runtests -i -a
```

#### To Do

We need first update of dependencies this is are the related issues or PR for the symfony 5 compatibility:

- [x] toflar/psr6-symfony-http-cache-store https://github.com/Toflar/psr6-symfony-http-cache-store/pull/18
- [x] symfony/symfony
   - [x] https://github.com/symfony/symfony/pull/33916
   - [x] https://github.com/symfony/symfony/pull/33918
- [x] phpspec/phpspec (required by some dependencies)
- [x] phpcr/phpcr-utils https://github.com/phpcr/phpcr-utils/pull/187
- [x] phpcr/phpcr-api-tests https://github.com/phpcr/phpcr-api-tests/pull/192
- [x] jackalope/jackalope-fs https://github.com/jackalope/jackalope-fs/pull/26
- [x] alchemy/binary-driver (php-ffmpeg)
   - [x] https://github.com/alchemy-fr/BinaryDriver/pull/32
   - [x] https://github.com/alchemy-fr/BinaryDriver/pull/33
- [x] neutron/temporary-filesystem (php-ffmpeg) https://github.com/romainneutron/Temporary-Filesystem/pull/15
- [x] jackalope/jackalope https://github.com/jackalope/jackalope/pull/356
- [x] phpcr/phpcr-migrations https://github.com/phpcr/phpcr-migrations/pull/9
- [x] symfony-cmf/testing https://github.com/symfony-cmf/testing/pull/200 https://github.com/symfony-cmf/testing/pull/201
- [x] massiveart/search-bundle https://github.com/massiveart/MassiveSearchBundle/pull/134
  - [x] ~~behat/symfony2-extension (massive/search-bundle)~~
  - [x] ~~behat/mink (massive/search-bundle)~~ https://github.com/minkphp/Mink/pull/794
[...]

PR #5113, opened 2020-03-17
Replace deprecated TranslatorInterface with new one

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes -
| License | MIT
| Documentation PR | -

#### What's in this PR?

Replace the `Symfony\Component\Translation\TranslatorInterface` with `Symfony\Contracts\Translation\TranslatorInterface`.

Basically Symfony have 2 different Translator the first one is the **[Symfony\Component\Translation\DataCollectorTranslator](https://github.com/symfony/symfony/blob/v4.4.5/src/Symfony/Component/Translation/DataCollectorTranslator.php)** which is used under development to track untranslated strings. The second one is the **[Symfony\Component\Translation\Translator](https://github.com/symfony/symfony/blob/v4.4.5/src/Symfony/Component/Translation/Translator.php)** which is used under prod environment. Both implements the `Symfony\Contracts\Translation\TranslatorInterface`.

In **Symfony 4.4** both have the `LegacyTranslatorInterface` which implements the `LocaleAwareInterface` which is also used in Symfony 5 for the Translator. In the services where we need to switch the locale of the translator we need to check manually for `LocaleAwareInterface` as there is no common interface.

In **Symfony 4.3** both have the `LegacyTranslatorInterface` but there exist no `LocaleAwareInterface` so we need to check there for `LegacyTranslatorInterface`.

So to support 4.3 to 5.0 we need to check for the following interface when the service need to change the Locale of the Translator:

 - `Symfony\Contracts\Translation\TranslatorInterface`  
     needed in symfony **^4.3 || ^5.0**  for **trans** method
 - `Symfony\Contracts\Translation\LocaleAwareInterface`  
    needed in symfony **^4.4 || ^5.0** for **getLocale**, **setLocale** methods
 - `Symfony\Component\Translation\TranslatorInterface as LegacyTranslatorInterface`  
    needed in symfony **4.3.x**  for **getLocale**, **setLocale** methods

#### Why?

The Component TranslatorInterface is deprecated and should not longer be used.

#### BC Breaks/Deprecations

I don't think this is a BC break as constructor can be overriden without matching the types of the parent class and he get still the same instance of translator.

#### TODO

 - [x] Update UpdateRouteCommand with LocaleAwareInterfaceCheck
 - [x] Update UserLocaleListener with LocaleAwareInterfaceCheck


src/Sulu/Bundle/RouteBundle/Command/UpdateRouteCommand.php
```diff
@@ -46,6 +48,16 @@ public function __construct(
         EntityManagerInterface $entityManager,
         RouteManagerInterface $routeManager
     ) {
+        if (!$translator instanceof LocaleAwareInterface && !$translator instanceof LegacyTranslatorInterface) {
```
wachterjohannes: can you explain me this?

alexander-schranz: read description of pull request

PR #5117, opened 2020-03-17
Add sulu admin http client service

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

Add sulu admin http client service.

#### Why?

The `http_client` service is not available by default in Symfony 5 and the config to activate the `http_client` service does only exist since Symfony 4.4 so we have 2 possibilities.

**A.:** **Keep support for symfony 4.3** Create a custom http_client service like in this pull request. Disadvantage request by a custom http_client are not displayed in the symfony web profiler.

**B.:** **Drop support for `symfony/framework-bundle` 4.3** which basically means dropping whole symfony 4.3 support for sulu 2.1 then we could `prepend` the framework bundle configuration:

```yaml
framework:
     http_client:
           enabled: true
```

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/config/services.xml
+++ b/src/Sulu/Bundle/AdminBundle/Resources/config/services.xml
@@ -158,8 +158,12 @@
             <tag name="jms_serializer.event_subscriber"/>
         </service>
 
+        <service id="sulu_admin.http_client" class="Symfony\Contracts\HttpClient\HttpClientInterface">
+            <factory class="Symfony\Component\HttpClient\HttpClient" method="create"/>
+        </service>
+
         <service id="sulu_admin.download_build_command" class="Sulu\Bundle\AdminBundle\Command\DownloadBuildCommand">
-            <argument type="service" id="http_client" />
+            <argument type="service" id="sulu_admin.http_client" />
             <argument>%kernel.project_dir%</argument>
             <argument>%sulu.version%</argument>
             <tag name="console.command"/>
@@ -169,7 +173,7 @@
             id="sulu_admin.download_language_command"
             class="Sulu\Bundle\AdminBundle\Command\DownloadLanguageCommand"
         >
-            <argument type="service" id="http_client" />
+            <argument type="service" id="sulu_admin.http_client" />
             <argument type="service" id="filesystem" />
             <argument>%kernel.project_dir%</argument>
             <argument>%sulu_core.locales%</argument>


alexander-schranz:
@chirimoya and @danrot can you both give your response what the plan is for 2.1 if we can drop support for Symfony 4.3 or would we try to invest time to keep backward compatibility to symfony 4.3?

chirimoya:
@alexander-schranz I'm fine with dropping 4.3 support for the 2.1 release if it makes things easier to maintain /cc @danrot  @wachterjohannes

danrot:
Currently we only use the `http_client` for the two download commands you have adjusted here, right? Then I would say it is quite easy to support Symfony 4.3, therefore I would not drop the support (personally I would not drop it unless completely necessary to support Symfony 5).

chirimoya:
@danrot I would drop the support for 4.3. 4.3 is currently in security fixes only mode and will be unsupported soon. https://symfony.com/releases/4.3

danrot:
It is still supported another 3 months, so we would already force people to upgrade now. And personally I really wouldn't do that, because people might use other libraries that have issues with later versions of Symfony (you basically never know). I would even wait with that till the next major release if possible.

But these are just my 2c, if we do differently I can also live with it.

chirimoya:
@danrot upgrading from 4.3 to 4.4 shouldn't be an issue and is also recommended.

chirimoya:
@sulu/core-team what are your thoughts?

nnatter:
dont have a strong opinion here 😇 if it simplifies things on our side i am okay with dropping support for 4.3 as updating the project as recommended anyway.

danrot:
> @danrot upgrading from 4.3 to 4.4 shouldn't be an issue and is also recommended.

In theory yes, but you know that we even had issues with Symfony patch releases lately :wink:

chirimoya:
> > @danrot upgrading from 4.3 to 4.4 shouldn't be an issue and is also recommended.
> 
> In theory yes, but you know that we even had issues with Symfony patch releases lately 😉

@danrot exactly, and that's the reason why I want to support as less minor versions as possible 🙈

danrot:
> > > @danrot upgrading from 4.3 to 4.4 shouldn't be an issue and is also recommended.
> > 
> > 
> > In theory yes, but you know that we even had issues with Symfony patch releases lately wink
> 
> @danrot exactly, and that's the reason why I want to support as less minor versions as possible

For me it is exactly the other way round :see_no_evil: I would like not to force people to upgrade just because it is easier for us to support only later versions, because they also might run into issues when doing that.

alexander-schranz:
@sulu/core-team nevertheless if we drop 4.3 or not I would merge this here as it is.

chirimoya:
@sulu/core-team decided to merge, independent of future symfony 4.3 support

PR #5119, opened 2020-03-19
Use default template from webspace when new page is added

### LINKED ISSUE #5055: Default template config in webspace is ignored
| Q | A
| --- | ---
| Sulu Version | 2.0.4
| Chrome Version 79.0.3945.130 | Chrome Version 79.0.3945.130

# Actual Behavior
The default-templates in the webspace config is ignored when creating a new page.

# Expected Behavior

When configuring the default template for the `page` type, I expect this template to be selected automatically when creating a new page.

# Steps to Reproduce

Add the following config in your webspace file:
```xml
    <default-templates>
        <default-template type="page">default</default-template>
        <default-template type="home">homepage</default-template>
    </default-templates>
```
Create extra page templates in your project.
Create a new page in Sulu, a different template is selected then the default.

danrot:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #5055
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR will replace the request for webspaces and localizations, and retrieve them via the `config` request at the very beginning of the application. That is possible, because these entities cannot be changed by the user.

Then the calls to the `webspactStore` and `localizationStore` do not need to be asynchronous, making certain tasks easier. E.g. determining the default template for a webspace, which is the reason we are doing this now.

#### BC Breaks/Deprecations

The `WebspaceController` and `LocalizationController` are deprecated now.

#### To Do

- [x] Adding the `webspace` information to the `config` request
- [x] Rewrite the `webspaceStore` to use the data from the `config` request
- [x] Check how to solve previous `checkForPermissions` flag in `WebspaceController` for `webspaceStore`
- [x] Rewrite `webspacestore` calls in `WebspaceTabs` containers
- [x] Fix bug in `TemplateToolbarAction` by using new `webspaceStore` calls
- [x] Rewrite `webspacestore` calls in `PageTabs` containers
- [x] Rewrite `webspaceStore` calls in `loadResoureLocatorInputTypeByWebspace` util
- [x] Rewrite `webspacestore` calls in `PageSettingsNavigationSelect`
- [x] Rewrite `webspacestore` calls in `CustomUrlsLocaleSelect`
- [x] Rewrite `webspacestore` calls in `CustomUrlsDomainSelect`
- [x] Rewrite `webspacestore` calls in `Preview` container
- [x] Rewrite `webspacestore` calls in `AnalyticsDomainSelect`
- [x] Adding the `localization` information to the `config` request
- [x] Rewrite the `localizationStore` to use the data from the `config` request
- [x] Rewrite `localizationStore` calls in `RoleAssignment`
- [x] Rewrite `localizationStore` calls in `userStore`
- [x] Replace `ResourceMultiSelect` in `Permissions` container with a usual select using the new `webspaceStore`
- [ ] Mode of `ResourceLocator` field type would not need to be a promise anymore


src/Sulu/Bundle/PageBundle/Admin/PageAdmin.php
```diff
@@ -343,6 +343,7 @@ public function getConfig(): ?array
         return [
             'teaser' => $this->teaserProviderPool->getConfiguration(),
             'versioning' => $this->versioningEnabled,
+            'webspaces' => $this->webspaceManager->getWebspaceCollection()->getWebspaces(),
```
danrot: Here I am just outputting all the webspaces. The `WebspaceController` had a separate flag `checkForPermissions`, based on which was decided if only the webspaces for which the user has the `VIEW` permission or all of them should be returned.

Wondering how we should implement that now... Maybe add a `_permissions` object, similar to what we already have in the responses of the `PageController`.

Then the `webspaceStore` could offer two methods:
- `getAllWebspaces()` returning all the webspaces in the system
- `getGrantedWebspaces()` only returning webspaces for which the current user has the `VIEW` permission

@sulu/core-team WDYT?

alexander-schranz: The `_permissions` sounds good for me. With `getGrantedWebspaces` I'm not sure, as I think it mostly is forgotten to use that method so I would say that `getAllWebspaces` should by default return only the webspaces I have permissions for like the api did and add a optional bool flag or additional method to get also the ones without I have permissions for.

danrot: Actually my reasoning was that there is no `getWebspaces` method, so that it is clear if I get all of them, or just the ones I was granted for. I thought that should be pretty clear this way. Boolean flags in functions tend to explode, therefore I would like to have to separate methods for that (actually I am currently even favouring computed properties, but my previous statement is still the same).

alexander-schranz: @danrot sounds good for me 👍

danrot:
The field type options for the `ResourceLocator` contain a `modeResolver` field, which is a promise. That was only necessary, because the webspaces were loaded asynchronously. Since this is not the case anymore, we could change that as well, although that would be a BC break. I guess we would have to rename it to just `mode`, and therefore break e.g. the `ArticleBundle`. On the other handside the props are received by the `modeResolver`, and maybe some day somebody would like to make a request then to find out :man_shrugging:

src/Sulu/Bundle/CustomUrlBundle/Resources/js/containers/Form/fields/CustomUrlsDomainSelect.js
```diff
@@ -26,19 +15,15 @@ class CustomUrlsDomainSelect extends React.Component<FieldTypeProps<string>> {
...
+        const {disabled, formInspector, value} = this.props;
 
         return (
             <SingleSelect
                 disabled={!!disabled}
                 onChange={this.handleChange}
                 value={value}
             >
-                {this.webspace.customUrls.map(({url}) => (
+                {webspaceStore.getWebspace(formInspector.options.webspace).customUrls.map(({url}) => (
```
alexander-schranz: would it not make sense to create a `@computed` value for webspaceStore.getWebspace(formInspector.options.webspace).customUrls ?

src/Sulu/Bundle/CustomUrlBundle/Resources/js/containers/Form/fields/CustomUrlsLocaleSelect.js
```diff
@@ -26,19 +15,15 @@ class CustomUrlsLocaleSelect extends React.Component<FieldTypeProps<string>> {
...
+        const {disabled, formInspector, value} = this.props;
 
         return (
             <SingleSelect
                 disabled={!!disabled}
                 onChange={this.handleChange}
                 value={value}
             >
-                {this.webspace.allLocalizations.map(({localization}) => (
+                {webspaceStore.getWebspace(formInspector.options.webspace).allLocalizations.map(({localization}) => (
```
alexander-schranz: same here `webspaceStore.getWebspace(formInspector.options.webspace).allLocalizations` would this not make sense to be a computed value?

src/Sulu/Bundle/PageBundle/Resources/js/containers/Form/fields/PageSettingsNavigationSelect.js
```diff
@@ -41,7 +26,7 @@ class PageSettingsNavigationSelect extends React.Component<FieldTypeProps<Array<
                 onChange={this.handleChange}
                 values={value || []}
             >
-                {this.webspace.navigations.map(({key, title}) => (
+                {webspaceStore.getWebspace(formInspector.options.webspace).navigations.map(({key, title}) => (
```
alexander-schranz: same as above `{webspaceStore.getWebspace(formInspector.options.webspace).navigations` could be a `@computed` value

src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/Preview.js
```diff
@@ -241,25 +211,17 @@ class Preview extends React.Component<Props> {
...
-
-        this.previewWindow = window.open(previewStore.renderRoute);
+        this.previewWindow = window.open(this.previewStore.renderRoute);
         this.previewWindow.addEventListener('beforeunload', action(() => {
             this.previewWindow = undefined;
         }));
     };
 
     render() {
-        const {previewStore} = this;
```
alexander-schranz: Was there any specific reason to remove the destructuring or just because of consistence to the other used vars?

src/Sulu/Bundle/WebsiteBundle/Resources/js/containers/Form/fields/AnalyticsDomainSelect.js
```diff
@@ -26,19 +15,15 @@ class AnalyticsDomainSelect extends React.Component<FieldTypeProps<Array<string>
...
+        const {disabled, formInspector, value} = this.props;
 
         return (
             <MultiSelect
                 disabled={!!disabled}
                 onChange={this.handleChange}
                 values={value || []}
             >
-                {this.webspace.urls.map(({url}) => (
+                {webspaceStore.getWebspace(formInspector.options.webspace).urls.map(({url}) => (
```
alexander-schranz: same as for the others think would make sense to have `webspaceStore.getWebspace(formInspector.options.webspace).urls` as `@computed` value

src/Sulu/Component/Localization/Localization.php
```diff
@@ -57,41 +58,47 @@ public static function createFromString($locale, $format = self::UNDERSCORE)
      * The language of the localization.
      *
      * @var string
+     * @Groups({"frontend", "Default"})
```
alexander-schranz: Should we not also add the `ExclusionStrategy` we normally use on all entities?

src/Sulu/Bundle/AudienceTargetingBundle/Tests/Functional/Controller/AdminControllerTest.php
```diff
@@ -15,6 +15,13 @@
 
 class AdminControllerTest extends SuluTestCase
 {
+    public function setUp(): void
+    {
+        parent::setUp();
+        $this->initPhpcr();
+        $this->purgeDatabase();
```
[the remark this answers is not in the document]
danrot: Copied the code from somewhere... Seems like `purgeDatabase` doesn't delete anything PHPCR related, because without `initPhpcr` this test is not working. Can still turn it around to avoid confusion.

src/Sulu/Bundle/CustomUrlBundle/Resources/js/containers/Form/fields/CustomUrlsDomainSelect.js
danrot: Since we are not accessing it multiple times it should not make a big difference. Can still do it, would maybe also make that part easier to read.

src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/Preview.js
danrot: It was about consistency :slightly_smiling_face:

src/Sulu/Bundle/AudienceTargetingBundle/Tests/Functional/Controller/AdminControllerTest.php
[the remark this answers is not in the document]
alexander-schranz: Ok good to know would be great to turn it around to avoid confusion :)

src/Sulu/Component/Localization/Localization.php
danrot: I am afraid of doing that, because that might result in some kind of BC break.

src/Sulu/Bundle/CustomUrlBundle/Resources/js/containers/Form/fields/CustomUrlsDomainSelect.js
alexander-schranz: Is a computed recalculated in every render call? If not I would move them to @computed.

src/Sulu/Bundle/CustomUrlBundle/Resources/js/containers/Form/fields/CustomUrlsLocaleSelect.js
alexander-schranz: same as above would move to `@computed` if render to avoid this on every render call

src/Sulu/Bundle/CustomUrlBundle/Resources/js/containers/Form/fields/CustomUrlsDomainSelect.js
danrot: No, it is not. It probably still doesn't make a huge difference, but I already moved it anyway :slightly_smiling_face:

src/Sulu/Bundle/PageBundle/Resources/js/containers/Form/fields/PageSettingsNavigationSelect.js
alexander-schranz: same as above would move to `@computed` if render to avoid this on every render call

src/Sulu/Component/Localization/Localization.php
alexander-schranz: Ok for me.

PR #5126, opened 2020-03-22
[MediaBundle] Azure Blob Storage implementation

plozmun:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs |-
| License | MIT
| Documentation PR | sulu/sulu-docs#502

#### What's in this PR?

 Azure Blob Storage implementation

#### Why?

We have a proposal from a potential client that requires the use of azure. More and more companies are using the Microsoft cloud

#### Example Usage

See docs

#### To Do

- [X] Create a documentation PR

I suggest to refactor s3 Storage because in 2.0 access to client in s3 will be removed ( see https://github.com/thephpleague/flysystem-azure-blob-storage/pull/18 )

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/DependencyInjection/Configuration.php
+++ b/src/Sulu/Bundle/MediaBundle/DependencyInjection/Configuration.php
@@ -12,6 +12,7 @@
 namespace Sulu\Bundle\MediaBundle\DependencyInjection;
 
 use League\Flysystem\AwsS3v3\AwsS3Adapter;
+use League\Flysystem\AzureBlobStorage\AzureBlobStorageAdapter;
 use Superbalist\Flysystem\GoogleStorage\GoogleStorageAdapter;
 use Symfony\Component\Config\Definition\Builder\ArrayNodeDefinition;
 use Symfony\Component\Config\Definition\Builder\TreeBuilder;
@@ -22,6 +23,8 @@ class Configuration implements ConfigurationInterface
     const STORAGE_GOOGLE_CLOUD = 'google_cloud';
 
     const STORAGE_S3 = 's3';
+
+    const STORAGE_AZURE_BLOB = 'azure_blob';
 
     public function getConfigTreeBuilder()
     {
@@ -221,6 +224,21 @@ class Configuration implements ConfigurationInterface
                             ->scalarPrototype()->end()
                         ->end()
                     ->end()
+                ->end();
+        }
+
+        if (class_exists(AzureBlobStorageAdapter::class)) {
+            $storages[] = self::STORAGE_AZURE_BLOB;
+
+            $storagesNode
+                ->arrayNode(self::STORAGE_AZURE_BLOB)
+                ->addDefaultsIfNotSet()
+                ->children()
+                ->scalarNode('connection_string')->isRequired()->end()
+                ->scalarNode('container_name')->isRequired()->end()
+                ->scalarNode('path_prefix')->defaultNull()->end()
+                ->scalarNode('segments')->defaultValue(10)->end()
+                ->end()
                 ->end();
         }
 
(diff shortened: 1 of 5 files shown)


plozmun:
AppVeyor fails for execution time and travis doesn't seem to want to start 😅

src/Sulu/Bundle/MediaBundle/DependencyInjection/Configuration.php
```diff
@@ -224,6 +227,21 @@ private function addStorageSection(ArrayNodeDefinition $node)
                 ->end();
         }
 
+        if (class_exists(AzureBlobStorageAdapter::class)) {
+            $storages[] = self::STORAGE_AZURE_BLOB;
+
+            $storagesNode
+                ->arrayNode(self::STORAGE_AZURE_BLOB)
+                ->addDefaultsIfNotSet()
```
alexander-schranz: Can you indent this lines to match the tree like in the other configuraitons

plozmun: Done!

alexander-schranz: This line should also be indented. So L235-L241 should be +4 spaces. See S3 example.

plozmun: I didn't realize 😅 , I think it's already

alexander-schranz: Thank you!

src/Sulu/Bundle/MediaBundle/Media/Storage/AzureBlobStorage.php
```diff
@@ -0,0 +1,79 @@
...
+        $this->container = $container;
+        $this->adapter = $filesystem->getAdapter();
+    }
+
+    public function load(array $storageOptions)
+    {
+        $resource = parent::load($storageOptions);
+        $contents = stream_get_contents($resource);
+
+        $stream = fopen('php://memory', 'r+');
```
wachterjohannes: @plozmun @alexander-schranz why do have to copy it to the memory?

wachterjohannes:
@alexander-schranz @plozmun i have added a question to the code - anything else looks good

but i also have no azure account

src/Sulu/Bundle/MediaBundle/Media/Storage/AzureBlobStorage.php
plozmun: Azure Filesystem returns a not seekable object, not like Google Filesystem than returns a PSR7Message seekable, then this fails when it tries to generate thumbnail https://github.com/sulu/sulu/blob/master/src/Sulu/Bundle/MediaBundle/Media/ImageConverter/ImagineImageConverter.php#L152
due to this https://github.com/sulu/sulu/blob/master/src/Sulu/Bundle/MediaBundle/Media/ImageConverter/MediaImageExtractor.php#L52 mime_content_type doest't support not seekable objects

wachterjohannes: OK sounds logic to me - i would at least add a comment for this to be sure that we do not remove it intentionally.

plozmun: Done! Check the comment and tell me if it's enough

alexander-schranz: Can we add the comment before `$contents = stream_get_contents($resource);` on the method its mostly overlooked.

Would also add more information to it e.g.:

```php
// Azure Filesystem returns a not seekable resource
// Need convert to a seekable resource to allow get mimetype from it for MediaImageExtractor
$contents = stream_get_contents($resource);
...
```

alexander-schranz: @plozmun thank you!

plozmun:
Have you been able to test this? We have it in pre-production with a git patch and everything is correct.

alexander-schranz:
Not sure why travis is not triggered for this pull request I tested it locally and the tests did work, so I decided to merge it manually.

@plozmun Thank you for this contribution 👍

PR #5129, opened 2020-03-23
Add sticky ckeditor toolbar

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

Make the ckeditor toolbar sticky.

#### Why?

I personally think the max-height is annoying because there is never a nice UX when having 2 scrollbars in a view. I would make the text editor grow and make the toolbar sticky at current state.

sticky-ckeditor-before

#### New implementation

Form:

sticky-ckeditor

Overlay:

sticky-ckeditor-2

#### To Do

- [x] Find a way that sticky top 0 works
   - [x] e.g.: Get ride of the -60px margin-top on tabs container base padding top of the view https://github.com/sulu/sulu/pull/6676
- [x] Check if it works correctly in a Form inside a Overlay

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/FormOverlay/formOverlay.scss
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/FormOverlay/formOverlay.scss
@@ -1,6 +1,7 @@
 @import '../Application/variables.scss';
 
 .form {
-    /* this container should not have a overflow hidden as that would break position sticky for texteditor and blocks */
+    /* this container should not have a overflow hidden with position relative
+     * as that would break position sticky for texteditor and blocks */
     padding: $viewPaddingVertical $viewPaddingHorizontal;
 }
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/FormOverlay/formOverlay.scss
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/FormOverlay/formOverlay.scss
@@ -1,7 +1,6 @@
 @import '../Application/variables.scss';
 
 .form {
-    // this container should not have a overflow hidden as that would break position sticky for texteditor and blocks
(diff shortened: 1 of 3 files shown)


alexander-schranz:
Maybe we can after #6676 implemented this avoid duplicated scrollbars and having nice sticky ckeditor toolbar. This would be a great UI improvements specially for articles with a lot of texts inside the editor.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FormOverlay/formOverlay.scss
```diff
@@ -1,7 +1,6 @@
 @import '../Application/variables.scss';
 
 .form {
-    margin: $viewPaddingVertical $viewPaddingHorizontal;
-    overflow: hidden;
-    position: relative;
```
alexander-schranz, line 6: The position relative and overflow hidden was introduced with the snackbar component in https://github.com/sulu/sulu/pull/4717. But in the same pull request the snackbar was moved from content part of the overlay to the main part and so the change itself was not required and so it is obsolete. And as written in the comment the overflow hidden makes problems here in combination with position relative. So I removed it. 

Alternative overflow auto on and setting container on height 100% and making so this container the scroll container could be done. But at current state I don't see any usages of the relative here and so I just removed it.

luca-rath (changes requested):
Two things:

1. An old bug is back now

image

2. The table has a higher z-index than the toolbar

image

alexander-schranz:
Good that you are reviewing this and knowing all this ckeditor quirks 😄

luca-rath (approved):

PR #5130, opened 2020-03-23
WIP: Add support for sqlite

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes -
| Related issues/PRs | -
| License | MIT
| Documentation PR | sulu/sulu-docs#prnum TODO

#### What's in this PR?

Add support for sqlite.

#### Why?

Sqlite is used mostly for development and to speed up tests. 

#### Example Usage

~~~php
DATABASE_URL=sqlite:////tmp/sulu_test.db
~~~

#### BC Breaks/Deprecations

No BC Break the service is only registered when the connection dsn uses sqlite.

#### To Do

- [x] Fix GROUP_CONCAT
- [ ] Add `?foreign-keys=true` to doctrine/dbal
   - [ ] Move ForeignKeyActivationSubscriber to doctrine/dbal
   - [ ] Activate foreignKeys for sqlite platform when flag is set

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -9,7 +9,7 @@ environment:
     MAILER_URL: null://localhost
     SULU_ADMIN_EMAIL:
     PHPCR_TRANSPORT: doctrinedbal
-    DATABASE_URL: sqlite:///c:/sulu_test.db?mode=0666
+    DATABASE_URL: "sqlite://:memory:"
     DATABASE_CHARSET: utf8mb4
     DATABASE_COLLATE: utf8mb4_unicode_ci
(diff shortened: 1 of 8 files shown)


danrot:
Do we really want to use sqlite for tests? I think that the system should work with the databases we are actually running Sulu on production. If Sulu cannot run correctly (including foreign keys) using SQLite, I also don't see any point in using it in the tests.

alexander-schranz:
Closing this in favor of:  https://github.com/doctrine/dbal/issues/2833 think this issue should be solved on dbal level.

PR #5131, opened 2020-03-24
Focus elements in FieldFilterTypes where appropriate

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | #5035 
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR adds the possibility to retrieve the refs for input elements of the `MultiAutoComplete` and `DateTime` component. This is done to set the focus on these elements in there respective `FieldFilterType`s.

#### Why?

Because this results in a better usability, because less clicks are necessary to setup a filter.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/fieldFilterTypes/DateTimeFieldFilterType.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/fieldFilterTypes/DateTimeFieldFilterType.js
@@ -1,5 +1,6 @@
 // @flow
 import React from 'react';
+import type {ElementRef} from 'react';
 import DatePicker from '../../../components/DatePicker';
 import AbstractFieldFilterType from './AbstractFieldFilterType';
 import dateTimeFieldFilterTypeStyles from './dateTimeFieldFilterType.scss';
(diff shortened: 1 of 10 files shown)


nnatter:
Wouldnt this make sense for the `number` filter-type too? 🤔

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/fieldFilterTypes/DateTimeFieldFilterType.js
```diff
@@ -27,12 +28,22 @@ class DateTimeFieldFilterType extends AbstractFieldFilterType<?{from?: Date, to?
...
+    }
+
     getFormNode() {
         const {value} = this;
 
         return (
             <div className={dateTimeFieldFilterTypeStyles.dateTimeFieldFilterType}>
-                <DatePicker onChange={this.handleFromChange} value={value ? value.from : undefined} />
+                <DatePicker
+                    inputRef={this.setInputRef}
```
nnatter: would prefer to rename this to `setFromInputRef`

nnatter:
Found another strange bug while testing this:
When I add a `text` filter, set a text and confirm, then remove the text and confirm, i am not able to remove the filter anymore. I need to confirm a new text to be able to remove it again.

nnatter:
Otherwise this feels good! Although it is a pity that i cannot confirm the value of a filter with `enter` while an input is focused. Is there a reason for not supporting that?

danrot:
The `enter` keypress is somehow swallowed by the `input` field itself, which also makes sense in someway. I guess `Mousetrap` does not interfere with the browser behavior. We could implement this separately for each `fieldFilterType`, but then we would not only have to pass an `onChange` but also an `onConfirm` callback to the `FieldFilterType`. Would be ok for me to do that, but I would not do that in this PR :see_no_evil: 

The other bug you've found I added to my comment in #5035, so that I can also fix that in a separate PR.

PR #5133, opened 2020-03-24
Add select type for SelectionFieldFilterType

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | #5035, dependant on #5138
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR introduces a new `type` parameter for the `selection` filter. This allows to render the possible values as a simple select dropdown instead of an auto complete field.

#### Why?

Because sometimes this variant is more comfortable to use.

#### Example Usage

~~~xml
 <filter type="selection">                                                                
    <params>                          
        <param name="displayProperty" value="name" />                                                                                                                                          
        <param name="resourceKey" value="accounts" />                                                                                                                                          
        <param name="type" value="select" /><!-- Can also be set to "auto_complete", which is the default value -->                                                                                                                                          
    </params>                         
</filter>                             
~~~

#### To Do

- [ ] Create a documentation PR (Actually the list XML part does not have its own documentation yet)
- [x] Fix the z-index problem already mentioned in #5035, because this makes the dropdown ignore its changes when the "OK" button of the `FieldFilterItem` is clicked without closing the select dropdown first by clicking somewhere outsie the `FieldFilterItem`.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/fieldFilterTypes/SelectionFieldFilterType.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/fieldFilterTypes/SelectionFieldFilterType.js
@@ -1,17 +1,26 @@
 // @flow
 import React from 'react';
 import type {ElementRef} from 'react';
-import {autorun, computed, toJS, untracked, when} from 'mobx';
+import {action, autorun, computed, observable, toJS, untracked, when} from 'mobx';
 import equals from 'fast-deep-equal';
 import MultiSelectionStore from '../../../stores/MultiSelectionStore';
 import MultiAutoComplete from '../../MultiAutoComplete';
+import ResourceMultiSelect from '../../ResourceMultiSelect';
 import AbstractFieldFilterType from './AbstractFieldFilterType';
 import selectionFieldFilterTypeStyles from './selectionFieldFilterType.scss';
+
+const TYPE_AUTO_COMPLETE = 'auto_complete';
+const TYPE_SELECT = 'select';
 
 class SelectionFieldFilterType extends AbstractFieldFilterType<?Array<string | number>> {
     selectionStore: MultiSelectionStore<string | number>;
     selectionStoreDisposer: () => void;
+    @observable selectValue: Array<string | number> = [];
     valueDisposer: () => void;
+
+    @computed get type() {
+        return this.parameters && (this.parameters.type || TYPE_AUTO_COMPLETE);
+    }
 
     constructor(
         onChange: (value: ?Array<string | number>) => void,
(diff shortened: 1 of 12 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/fieldFilterTypes/SelectionFieldFilterType.js
```diff
@@ -44,11 +43,15 @@ class SelectionFieldFilterType extends AbstractFieldFilterType<?Array<string | n
...
         this.valueDisposer = autorun(() => {
-            const {value = []} = this;
+            const value = toJS(this.value || []);
 
-            if (!equals(toJS(value), untracked(() => this.selectionStore.ids))) {
+            if (!equals(value, untracked(() => this.selectionStore.ids))) {
                 this.selectionStore.loadItems(value);
             }
+
+            if (!equals(value, untracked(() => this.selectValue))) {
```
nnatter: this is not tested. but i am not sure if there was a test for the previous implementation

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/fieldFilterTypes/SelectionFieldFilterType.js
```diff
@@ -1,36 +1,35 @@
...
 import AbstractFieldFilterType from './AbstractFieldFilterType';
 import selectionFieldFilterTypeStyles from './selectionFieldFilterType.scss';
 
+const TYPE_AUTO_COMPLETE = 'auto_complete';
+const TYPE_SELECT = 'select';
+
 class SelectionFieldFilterType extends AbstractFieldFilterType<?Array<string | number>> {
     selectionStore: MultiSelectionStore<string | number>;
     selectionStoreDisposer: () => void;
+    @observable selectValue: Array<string | number> = [];
```
nnatter: edit: took some time to understand why this is needed. but dont see a more intuitive solution at the moment. maybe add a comment that we need to buffer this to not send `SelectionStore` requests while the eselect is opened?

danrot:
@nnatter Fixed the two issues in the last commit.

nnatter:
The size of the loader of the `ResourceMultiSelect` kinda feels a bit too big. The height of the chip popover is also changing when the loader is displayed. Setting the size to `30` like in the `ResourceSingleSelect` would solve this.

nnatter:
Also, the `ResourceMultiSelect` sends a request each time the chip popover is opened.. Is this expected? Otherwise this looks good to me 🙂

PR #5135, opened 2020-03-25
Add object-shorthand eslint rule

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

Add object-shorthand eslint rule and activate eslint cache by default. 

I personally not sure about the rule configuraiton but was requested by danrot to use shorthand syntax in #5101. 

#### Why?

For consistence.

#### BC Breaks/Deprecations

No bc breaks

#### To Do

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
     "repository": "https://github.com/sulu/sulu.git",
     "scripts": {
         "preinstall": "node preinstall.js",
-        "lint:js": "eslint .",
+        "lint:js": "eslint . --cache",
         "lint:scss": "stylelint src/Sulu/Bundle/*/Resources/js/**/*.scss",
         "flow": "flow",
         "test": "jest",
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/DatePicker/DatePicker.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/DatePicker/DatePicker.js
@@ -182,7 +182,7 @@ class DatePicker extends React.Component<Props> {
         const inputProps = {
             placeholder: placeholder ? placeholder : this.getFormat(),
             valid: valid && !this.showError,
-            disabled: disabled,
+            disabled,
             icon: fieldOptions.dateFormat ? 'su-calendar' : 'su-clock',
         };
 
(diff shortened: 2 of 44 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/components/DatePicker/DatePicker.js
```diff
@@ -182,7 +182,7 @@ class DatePicker extends React.Component<Props> {
         const inputProps = {
             placeholder: placeholder ? placeholder : this.getFormat(),
             valid: valid && !this.showError,
-            disabled: disabled,
+            disabled,
```
alexander-schranz: I personally find this harder to read should we go atleast with [consistent](https://eslint.org/docs/rules/object-shorthand) rule flag /cc @danrot @nnatter

package.json
```diff
@@ -4,7 +4,7 @@
     "repository": "https://github.com/sulu/sulu.git",
     "scripts": {
         "preinstall": "node preinstall.js",
-        "lint:js": "eslint .",
+        "lint:js": "eslint . --cache",
```
nnatter: 🎉 🎉 🎉 🎉 🎉 🎉 🎉

src/Sulu/Bundle/AdminBundle/Resources/js/components/DatePicker/DatePicker.js
nnatter: not a fan of `consistent` because there are two correct ways for some objects. would prefer `consistent-as-needed` then. but i think the readability is just a matter of getting used to it, so `always` is okay for me too 🙂

danrot: I am already used to this, think it is easier to read this way, and in many places of the code base it is already handled like this. I trip when I see the same word twice in a line :see_no_evil:

package.json
danrot: I hope that does not affect the CI server... Otherwise I would be against it :see_no_evil: But I think we don't cache this file between runs there, so it should be fine.

PR #5137, opened 2020-03-25
Fixed correct locale for article route generation

adriansieber:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Fixes the route generation in articles that have translated routes.

#### Why?

Wrong Route Translation when adding new Article: https://github.com/sulu/SuluArticleBundle/issues/432

### DIFF
--- a/src/Sulu/Bundle/RouteBundle/Generator/SymfonyExpressionTokenProvider.php
+++ b/src/Sulu/Bundle/RouteBundle/Generator/SymfonyExpressionTokenProvider.php
@@ -42,9 +42,6 @@ class SymfonyExpressionTokenProvider implements TokenProviderInterface
         try {
             if (method_exists($entity, 'getLocale')) {
                 $this->translator->setLocale($entity->getLocale());
-            } elseif(is_a($entity->getLocale, \Closure::class)) {
-                $getLocale = $entity->getLocale;
-                $this->translator->setLocale($getLocale());
             }
 
             $result = $this->expressionLanguage->evaluate($name, ['object' => $entity, 'translator' => $this->translator]);
--- a/src/Sulu/Bundle/RouteBundle/Generator/SymfonyExpressionTokenProvider.php
+++ b/src/Sulu/Bundle/RouteBundle/Generator/SymfonyExpressionTokenProvider.php
@@ -37,9 +37,9 @@ class SymfonyExpressionTokenProvider implements TokenProviderInterface
 
     public function provide($entity, $name)
     {
-        try {
-            $locale = $this->translator->getLocale();
+        $locale = $this->translator->getLocale();
 
+        try {
             if (method_exists($entity, 'getLocale')) {
                 $this->translator->setLocale($entity->getLocale());
             }
--- a/src/Sulu/Bundle/RouteBundle/Generator/SymfonyExpressionTokenProvider.php
+++ b/src/Sulu/Bundle/RouteBundle/Generator/SymfonyExpressionTokenProvider.php
@@ -50,6 +50,8 @@ class SymfonyExpressionTokenProvider implements TokenProviderInterface
             return $result;
         } catch (\Exception $e) {
             throw new CannotEvaluateTokenException($name, $entity, $e);
+        } finally {
+            $this->translator->setLocale($locale);
         }
     }
 }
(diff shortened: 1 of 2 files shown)


src/Sulu/Bundle/RouteBundle/Generator/SymfonyExpressionTokenProvider.php
```diff
@@ -38,7 +38,12 @@ public function __construct(TranslatorInterface $translator)
     public function provide($entity, $name)
     {
         try {
-            return $this->expressionLanguage->evaluate($name, ['object' => $entity, 'translator' => $this->translator]);
+            $locale = $this->translator->getLocale();
+            $this->translator->setLocale($entity->getLocale());
```
alexander-schranz: We can not be sure that `$entity->getLocale()` exist @wachterjohannes what do you think should we go with:

```
if (method_exists($entity, 'getLocale')) {

}
```

wachterjohannes: @alexander-schranz your right - in a "normal" usecase with routebundle it will have a `getLocale` but yes the method allows also other entities

@adriansieber will you add this condition?

danrot: Apart from that, don't you think setting the locale this way might lead to strange bugs? When somebody catches the `CannotEvaluateTokenException` below, the wrong state is kept. I don't know how this is used afterwards, but the `trans` function also accepts a `locale` attribute. Maybe we can decorate this service somehow, so that this `$locale` attribute is always passed?

alexander-schranz: @danrot think we should reset the translator locale in finally.

danrot: Yeah, at least we should that. Would still prefer not changing it in the first place :see_no_evil:

danrot:
@adriansieber It would be nice if you could run your tests locally before commiting. You are clogging our pipeplines this way :see_no_evil:

src/Sulu/Bundle/RouteBundle/Tests/Unit/Generator/SymfonyExpressionTokenProviderTest.php
```diff
@@ -21,31 +21,36 @@ class SymfonyExpressionTokenProviderTest extends TestCase
     public function testResolve()
     {
         $translator = $this->prophesize(TranslatorInterface::class);
+        $translator->getLocale()->willReturn('de');
+        $translator->setLocale('de')->shouldBeCalled();
```
wachterjohannes: @adriansieber theoretically the translator should also be called with `en` because the entity has the function `getLocale` and return `en` - can you check why this is not happening?

src/Sulu/Bundle/RouteBundle/Generator/SymfonyExpressionTokenProvider.php
```diff
@@ -42,6 +42,9 @@ public function provide($entity, $name)
         try {
             if (method_exists($entity, 'getLocale')) {
                 $this->translator->setLocale($entity->getLocale());
+            } elseif(is_a($entity->getLocale, \Closure::class)) {
```
wachterjohannes: ??? wou thats a hard thing 😂

src/Sulu/Bundle/RouteBundle/Tests/Unit/Generator/SymfonyExpressionTokenProviderTest.php
```diff
@@ -22,6 +22,7 @@ public function testResolve()
     {
         $translator = $this->prophesize(TranslatorInterface::class);
         $translator->getLocale()->willReturn('de');
+        $translator->setLocale('en')->shouldBeCalled();
         $translator->setLocale('de')->shouldBeCalled();
         $entity = new \stdClass();
```
wachterjohannes: maybe use this snippet here:

```
$entity = $this->prophesize(RoutableInterface::class);
$entity->getLocale()->wilLReturn('en');
```

Then you should be able to remove the `is_a` from the production code

wachterjohannes:
@alexander-schranz now it looks good for me - what do you think?

alexander-schranz:
@wachterjohannes code looks ok for me but think this fix should be backported to 1.6 or not?

wachterjohannes:
@alexander-schranz yes but i would merge this and cherry-pick it on the release/1.6 branch

wachterjohannes:
@alexander-schranz @adriansieber backorted in https://github.com/sulu/sulu/commit/861d59ebb5

PR #5138, opened 2020-03-25
Remove z-index values

danrot:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | #5035 
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR removes all `z-index` values, and sometimes has to reorder some HTML components to make this possible.

#### Why?

This PR fixes a problem that occured in #5035. The dropdown filter introduced there could only be closed if something outside of the `FieldFilterItem` (containing the `ArrowMenu`) was clicked. Now it is possible to also click on the `ArrowMenu` to close it.

#### To Do

- [x] Test in different browser


src/Sulu/Bundle/MediaBundle/Resources/js/containers/MultiMediaDropzone/dropzoneOverlay.scss
```diff
@@ -8,7 +8,6 @@ $dropAreaInfoColor: $white;
     position: fixed;
     top: 0;
     left: 0;
-    z-index: 1;
```
danrot: This does not seem to work anymore after changing the `Header`

danrot:
@nnatter Tested this already in chrome, firefox and safari, and didn't encounter any errors :slightly_smiling_face: The important thing to test is the dropdown error you have found in #5035, this should work now as well.

nnatter:
Did encounter two quirks while testing this, otherwise it is looking good:
* The `DropzoneOverlay` (at least i think that is what it is called) that appears when drag-and-dropping an image flickers like crazy
* Looks like the `ckeditor` uses z-index for parts of its interface. At least some parts cover the area of a dialog in my case:
Screenshot 2020-03-26 at 16 18 08

danrot:
@nnatter Both issues should be fixed now. And "is using z-index" becomes a new criteria when choosing a 3rd party package :see_no_evil:

nnatter:
works like a charm 🙂

PR #5145, opened 2020-03-26
Adjust ArrowMenu to size changes of its children

danrot:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | #5035 
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR changes the size of the `ArrowMenu` component when one of its children changes.

#### Why?

Because when implementing filters this use case occured when e.g. using the `MultiAutoComplete` for setting the organization filter.


src/Sulu/Bundle/AdminBundle/Resources/js/components/Popover/Popover.js
```diff
@@ -38,19 +38,26 @@ class Popover extends React.Component<Props> {
...
-    componentDidMount() {
+    mutationObserver: MutationObserver;
+
+    constructor(props: Props) {
+        super(props);
+
         window.addEventListener('blur', this.close);
         window.addEventListener('resize', this.close);
+        this.mutationObserver = new MutationObserver(() => {
+            this.setPopoverSize(0, 0);
```
nnatter: i guess this is necessary somehow? asking because the `updateDimensions()` method will call `setPopoverSize()` with some calculated parameters anyway

src/Sulu/Bundle/AdminBundle/Resources/js/components/MultiAutoComplete/multiAutoComplete.scss
```diff
@@ -29,9 +29,9 @@ $multiAutoCompleteDisabledBackgroundColor: $wildSand;
         display: flex;
         justify-content: center;
         align-items: center;
-        height: calc(100% - 10px);
+        align-self: stretch;
```
nnatter: one of these styling changes increases the height of the autocomplete input field when the loader is displayed for me. can be fixed by setting the size of the loader to `16px` in the following file: https://github.com/sulu/sulu/blob/master/src/Sulu/Bundle/AdminBundle/Resources/js/components/MultiAutoComplete/MultiAutoComplete.js#L186

src/Sulu/Bundle/AdminBundle/Resources/js/components/Popover/Popover.js
danrot: Yeah, the problem is that if we don't reset the popover size first, there will already be some width and height defined on that element, and so it will still not grow. Can also add a comment explaining that.

nnatter: okay, that would be nice! 🙂

PR #5148, opened 2020-03-27
Only set default template for pages when no template is currently selected

danrot:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | introduced in #5119
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR chooses the default template based on a value from the type schema request.

#### Why?

Because the `ResourceFormStore` was setting the first one as default, and completely ignoring the defaults defined in the webspace.


src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/StructureFormMetadataLoader.php
```diff
@@ -86,6 +86,8 @@ public function getMetadata(string $key, string $locale, array $metadataOptions
         if (isset($metadataOptions['webspace'])) {
             $webspace = $this->webspaceManager->findWebspaceByKey($metadataOptions['webspace']);
 
+            $form->setDefaultType($webspace->getDefaultTemplate($key));
```
alexander-schranz: Missing that for example the default type is correctly set by the `sulu.content.structure.default_types` parameter which is used for snippets and articles

```php
$form->setDefaultType($this->defaultTypes[$key] ?? null);
```

```yaml
sulu_core:
    content:
        structure:
            default_type:
                snippet: default
```

PR #5149, opened 2020-03-30
Make the datepicker elements clickable for the user

danrot:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | #5138 
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR makes sure the `Backdrop` of a `DatePicker` is behind the `DatePicker` it self. It does that by putting the `Backdrop` into the same `Portal` as the `DatePicker`, and renders it before the `DatePicker`.

Previously the functionality was dependant on the correct calls of the `createPortal` function. If the `createPortal` for the `Backdrop` was called after the one for the `DatePicker`, the order in the DOM was wrong, and therefore the user clicked on the invisible `Backdrop` instead of the `DatePicker`.

#### Why?

Because otherwise the user cannot click any date on the date picker.

#### BC Breaks/Deprecations

Some props might be removed from the `Backdrop` props, and its default behavior changes. But that's open for discussion

#### To Do

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Overlay/Overlay.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Overlay/Overlay.js
@@ -102,7 +102,6 @@ class Overlay extends React.Component<Props> {
             confirmDisabled,
             confirmLoading,
             confirmText,
-            onClose,
             onConfirm,
             title,
             size,
@@ -126,9 +125,9 @@ class Overlay extends React.Component<Props> {
 
         return (
             <Fragment>
-                <Backdrop onClick={onClose} open={visible} />
                 {visible &&
                     <Portal>
+                        <Backdrop local={true} />
                         <div
                             className={containerClass}
                             onTransitionEnd={this.handleTransitionEnd}
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Overlay/Overlay.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Overlay/Overlay.js
@@ -127,7 +127,7 @@ class Overlay extends React.Component<Props> {
             <Fragment>
                 {visible &&
                     <Portal>
-                        <Backdrop local={true} />
+                        <Backdrop />
                         <div
                             className={containerClass}
                             onTransitionEnd={this.handleTransitionEnd}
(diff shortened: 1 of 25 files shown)


danrot:
The first commit is the one actually fixing the bug. This was done by moving the `Backdrop` to another place.

The second commit makes exactly the same for the `Overlay` and `Dialog`, which I would recommend to do in order to avoid similar problems in the future.

Finally the third commit removes the options, that are not used any more after the second commit. This is clearly a BC break, but I would accept that. If we decide to do so, I would also have to upgrade the `Backdrop` documentation and add some instructions to `UPGRADE.md`.

@sulu/core-team What do you think? Would it be ok for you to apply all three commits, although they contain BC breaks? Have you used the `Backdrop` component directly in some projects?

nnatter:
@danrot I never used the `Backdrop` component inside of a project - always used the `Overlay` or `Dialog` component. The `local` prop confused my anyway, so I am okay with removing the props.

alexander-schranz:
@danrot I also did not use Backdrop directly and so never got in contact with this props. Think we could change this as I don't think someone did use it directly.

danrot:
Then I am happy to do so, because these properties can introduce problems :see_no_evil: Who of you would like to review that PR? :slightly_smiling_face:

src/Sulu/Bundle/AdminBundle/Resources/js/components/Overlay/Overlay.js
```diff
@@ -126,9 +125,9 @@ class Overlay extends React.Component<Props> {
 
         return (
             <Fragment>
-                <Backdrop onClick={onClose} open={visible} />
                 {visible &&
                     <Portal>
+                        <Backdrop />
```
nnatter: what is the intention behind removing the `onClick={onClose}` prop? 🤔

danrot: It didn't work anyway, because the `Overlay` component was putting another div on top of it. Also, the `Dialog` component is also not closable by clicking on the `Backdrop`.

nnatter: alright - goot to know 🙂

PR #5152, opened 2020-03-30
Fix SuluSecurityListener with InvokeableControllers

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes -
| Related issues/PRs | -
| License | MIT

#### What's in this PR?

Fix problems with callable classes as controller actions

#### Why?

The SuluSecurityController have problems if the given controller is a callable by itself with a `__invoke` function. 

#### Example Usage

[InvokeableController](https://symfony.com/doc/4.4/controller/service.html#invokable-controllers)

~~~php
use Symfony\Component\HttpFoundation\Response;
use Symfony\Component\Routing\Annotation\Route;

/**
 * @Route("/api/admin/my-resource", name="my")
 */
class MyAction {
       public function __invoke(Request $request, $id) {
             return new JsonResponse(['id' => 1, 'name' => 'Test']);
       }
}
~~~

### DIFF
--- a/src/Sulu/Bundle/SecurityBundle/EventListener/SuluSecurityListener.php
+++ b/src/Sulu/Bundle/SecurityBundle/EventListener/SuluSecurityListener.php
@@ -41,8 +41,18 @@ class SuluSecurityListener
      */
     public function onKernelController(FilterControllerEvent $event)
     {
-        $controllerDefinition = $event->getController();
-        $controller = $controllerDefinition[0];
+        $controller = $event->getController();
+        $action = '__invoke';
+
+        if (is_array($controller)) {
+            if (isset($controller[1])) {
+                $action = $controller[1];
+            }
+
+            if (isset($controller[0])) {
+                $controller = $controller[0];
+            }
+        }
 
         if (
             !$controller instanceof SecuredControllerInterface &&
@@ -61,7 +71,7 @@ class SuluSecurityListener
                 $permission = PermissionTypes::VIEW;
                 break;
             case 'POST':
-                if ('postAction' == $controllerDefinition[1]) { // means that the ClassResourceInterface has to be used
+                if (in_array($action, ['postAction', '__invoke'])) { // means that the ClassResourceInterface has to be used
                     $permission = PermissionTypes::ADD;
                 } else {
                     $permission = PermissionTypes::EDIT;
--- a/src/Sulu/Bundle/SecurityBundle/Tests/Unit/EventListener/SuluSecurityListenerTest.php
+++ b/src/Sulu/Bundle/SecurityBundle/Tests/Unit/EventListener/SuluSecurityListenerTest.php
@@ -170,6 +170,28 @@ class SuluSecurityListenerTest extends TestCase
             ->shouldHaveBeenCalled();
     }
 
+    /**
+     * @dataProvider provideInvokeMethodActionMapping
+     */
+    public function testCallableControllerPermission($method, $permission)
+    {
+        $request = $this->prophesize(Request::class);
+        $request->getMethod()->willReturn($method);
+        $request->get('id')->willReturn('1');
+
+        $controller = $this->prophesize(SecuredControllerInterface::class);
+        $controller->getSecurityContext()->willReturn('security.context');
+        $controller->getLocale(Argument::any())->willReturn('de');
+
+        $this->filterControllerEvent->getRequest()->willReturn($request->reveal());
+        $this->filterControllerEvent->getController()->willReturn($controller->reveal());
+
+        $this->securityListener->onKernelController($this->filterControllerEvent->reveal());
+
+        $this->securityChecker->checkPermission(Argument::any(), $permission, Argument::any())
+            ->shouldHaveBeenCalled();
+    }
+
     public function testLocale()
     {
         $request = $this->prophesize(Request::class);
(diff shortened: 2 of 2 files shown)


src/Sulu/Bundle/SecurityBundle/EventListener/SuluSecurityListener.php
```diff
@@ -61,7 +71,7 @@ public function onKernelController(FilterControllerEvent $event)
                 $permission = PermissionTypes::VIEW;
                 break;
             case 'POST':
-                if ('postAction' == $controllerDefinition[1]) { // means that the ClassResourceInterface has to be used
```
nnatter: are you sure that `__invoke` is always `PermissionTypes::ADD`? 🤔

src/Sulu/Bundle/SecurityBundle/EventListener/SuluSecurityListener.php
```diff
@@ -41,8 +41,18 @@ public function __construct(SecurityCheckerInterface $securityChecker)
      */
     public function onKernelController(FilterControllerEvent $event)
     {
-        $controllerDefinition = $event->getController();
-        $controller = $controllerDefinition[0];
+        $controller = $event->getController();
```
nnatter: looks like this never worked with a callable like `MyClass::myCallbackMethod`?
https://www.php.net/manual/de/language.types.callable.php

src/Sulu/Bundle/SecurityBundle/EventListener/SuluSecurityListener.php
```diff
@@ -61,7 +71,7 @@ public function onKernelController(FilterControllerEvent $event)
                 $permission = PermissionTypes::VIEW;
                 break;
             case 'POST':
-                if ('postAction' == $controllerDefinition[1]) { // means that the ClassResourceInterface has to be used
+                if (in_array($action, ['postAction', '__invoke'])) { // means that the ClassResourceInterface has to be used
```
alexander-schranz: @nnatter Can be discussed. I personally think if he want to EDIT he should use  `put` or `patch` action instead of a `post`, but at current state I would go with `ADD` as its more common that `post` is a, add and not a edit.  /cc @danrot what do you think

src/Sulu/Bundle/SecurityBundle/EventListener/SuluSecurityListener.php
alexander-schranz: @nnatter oh thats interesting 🙈

alexander-schranz: @nnatter any way to reproduce it? We definitly need a callable which has an instance of `$controller` here as we call `$controller->getLocale()` in this function.
Atleast the listener will not throw an error if a string is given, but you will not be able to use the `SecuredControllerInterface` on your controller then.

src/Sulu/Bundle/SecurityBundle/EventListener/SuluSecurityListener.php
danrot: I think it is a huge misconception that POST should only be used for create. If you have to add some business logic (as we do in some cases, e.g. when publishing a page) it is totally fine to use POST as well. You can take [Roy Fielding's (the creator of REST) word for it](https://roy.gbiv.com/untangled/2009/it-is-okay-to-use-post).

Regarding `__invoke`: I would say if a controller is defined like that, it gives no clue whatsoever what it is doing. Therefore I would not include any logic about this in here, and let the controller itself define the security if necessary. It's the only place where we really know.

But of course this code should not crash when something unexpected is delivered, so I would **only** fix that.

alexander-schranz: @danrot sure you can use POST for more as create but its more common and I would more connect POST to ADD permisson as to EDIT permission as this is more common.

danrot: I think I don't really understand the problem here: `postAction` requires an `ADD` permission here, so that should be fine.

I think what @nnatter meant (and I agree with him) is the `__invoke` action. As said, `__invoke` can be anyhting, and the function name does not imply anything about its usage. Therefore I would not do anything in this listener when it gets an action called `__invoke`, except for making sure that it does not break anything.

danrot: I think even the example code you've posted in the PR description would require only a `VIEW` permission, since it does not add anything, it just returns some values.

alexander-schranz: So your suggestion is to ignore `SecuredControllerInterface` and just do an early return in this case?

danrot: Hm, either that, or throw a clear error message saying that the `SecuredControllerInterface` only works with classes that also implement the `ClassResourceInterface`. BTW, what is happening currently? :thinking:

chirimoya: As I read this dicuseen, I think your ar not talking about the same thing. It's important to consider that's always a combination of HTTP method and method name. A GET and `__invoke` is a `READ`. Only for POST method we need to make assumptions and if somebody implements the `SecuredControllerInterface` it should work.

danrot: Ah, completely missed the `case 'POST'` part before it :see_no_evil: Now everything makes more sense. I was also mislead by the @nnatter's comment... `__invoke` is not always `POST` with this implementation, right? Then everything is fine :+1:

alexander-schranz: @danrot ok good :)

> BTW, what is happening currently?

Currently callable classes didn't work and did on `$controller[0]` throw a PHP Error trying to access object as array.

PR #5153, opened 2020-03-30
Adjust size of Loader in ResourceMultiSelect to prevent height changes

nnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### Why?

Because right now, the loader that is displayed during loading is higher than the `MultiSelectComponent` that is displayed after the data is loaded. This means that the height of the component changes unnecessarily after the data is loaded. This causes a "jump" in the ui as the position of all elements below the component is changed. 

Example: https://sulu.rocks/admin/#/contacts/2/permissions


danrot:
This PR should probably be set against `release/2.0`, shouldn't it?

nnatter:
Yes, you are right. 🙂

PR #5159, opened 2020-04-01
Update react-dropzone

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR updates the `react-dropzone` package. There were a few changes which needed to be addressed.

#### Why?

Because we want our packages to stay up to date :slightly_smiling_face:

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/Resources/js/components/SingleMediaDropzone/SingleMediaDropzone.js
+++ b/src/Sulu/Bundle/MediaBundle/Resources/js/components/SingleMediaDropzone/SingleMediaDropzone.js
@@ -1,6 +1,6 @@
 // @flow
 import React, {Fragment} from 'react';
-import {observer} from 'mobx-react';
+import {observer, Observer} from 'mobx-react';
 import {observable, action} from 'mobx';
 import classNames from 'classnames';
 import Dropzone from 'react-dropzone';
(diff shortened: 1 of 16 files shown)


src/Sulu/Bundle/MediaBundle/Resources/js/components/SingleMediaDropzone/SingleMediaDropzone.js
```diff
@@ -115,49 +115,62 @@ class SingleMediaDropzone extends React.Component<Props> {
...
-
-                {!uploading
-                    ? <div className={singleMediaDropzoneStyles.uploadIndicatorContainer}>
-                        <div className={singleMediaDropzoneStyles.uploadIndicator}>
-                            <div>
-                                <Icon className={singleMediaDropzoneStyles.uploadIcon} name={UPLOAD_ICON} />
-                                {uploadText &&
-                                    <div className={singleMediaDropzoneStyles.uploadInfoText}>{uploadText}</div>
+                {({getInputProps, getRootProps}) => (
+                    <Observer>
```
alexander-schranz, line 126: Interesting didn't know that something like this exists 😲

danrot: Yeah, and it is necessary in that case... The thing is that because the `react-dropzone` now only accepts a function, the stuff we add here is actually rendered by `react-dropzone`, and then the `SingleMediaDropzon` does not seem to observe it anymore. Therefore we need to add another anonymous `Observer` here.

Issue #3675, opened 2017-12-05
Add system roles to avoid problems with static checks


alexander-schranz:
| Q | A
| --- | ---
| Bug? | yes
| New Feature? | no
| Sulu Version | 1.6.9
| Releated Issues | https://github.com/sulu/SuluCommunityBundle/issues/63

#### Actual Behavior

Currently the role key which is generated by the role name (https://github.com/sulu/sulu/blob/1.6.9/src/Sulu/Bundle/SecurityBundle/Entity/BaseRole.php#L96) get useless for the developer if a content manager change the role name.

#### Expected Behavior

A role key should not change its key when the content manager change its name.

#### Steps to Reproduce

Create a sulu role. Add a check for it to the `security.yml` e.g.:

```yml
    access_control:
       - { path: ^/test, roles: ROLE_SULU_USER }
```

After renaming the role a user with the role is not longer able to access the route as the roles KEY is changed.

#### Possible Solutions

The role should get an additional `key` property which is generated when the role is created (also based on the given name). The only difference to the current behaviour is that the key is not changed when the name is changed by the content manager this make it possible to check for the specific role in the security.yml.

danrot:
Sounds good! The only thing I am worried about is that it might get hard to get these connections (because the key doesn't match the name after some time, or might even exchange names). But if @chirimoya is ok with that I don't see any reason we shouldn't go for that option.

alexander-schranz:
As discussed with @danrot @chirimoya we should add a property `parent`, `symfonyRole` or `systemRole` to link a sulu role to a symfony role which can be used in the firewall. So the role name used in the firewall is not longer based on the name the content manager can change instead based on the symfony role.

danrot:
When thinking about it, it might also make sense to have something similar to system collections or system categories (not implemented but discussed in #374) for roles. Otherwise we have the same problem as always, we need to put something in the configuration, that has a dependency on the database... And if somebody deletes that role, it might lead to unexpected behavior.

nnatter:
Should we close this because of #5166? @danrot

danrot:
Just had this issue open a few seconds ago as well :see_no_evil: I was not sure, because we've said it is only a temporary solution, actually we need something like system roles... I think I am going to rename that issue to better reflect that and leave it open.

nnatter:
Alright, then renaming sounds good 🙂

fnagel:
@alexander-schranz Shouldn't this be closed as #5166 has been merged?

[...]

PR #5166, opened 2020-04-02
Add key property to Role entity

nnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | yes
| Deprecations? | no
| Related issues/PRs | #3675
| License | MIT

#### What's in this PR?

This PR adds a `key` property to the `Role` entity. The `key` property is implemented similar to the `key` property of the `Category` entity. 

Furthermore, it adjust the logic to use the `key` property instead of the `name` property for generating the identifier of a role. The identifier of a role is used for restricting access of a user in the [Symfony security configuration](https://symfony.com/doc/current/security.html#securing-url-patterns-access-control).

#### Why?

At the moment, we use the `name` property for generating the identifier of a role. This means that changing the `name` of a role might break the Symfony security configuration of the project. See #3675 for more information.

I am aware that this is not a perfect solution, but it is a solution that improves the current situation and is implementable with relatively low effort and small BC breaks. We can still implement something like `System Roles` in the future.

#### BC Breaks/Deprecations

See `UPGRADE.MD`

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -2,6 +2,25 @@
 
 ## dev-master
 
+### Add key property to Role entity
+
+To allow for referencing `Role` entities by a human readable string, a `key` property was added to the `Role` entity.
+Furthermore, the `key` property is used for generating the identifier of a role instead of the `name` property now.
+The following statements update the database to include the new property and set the key of each existing Role to 
+keep backwards compatibility:
+
+```sql
+DROP INDEX UNIQ_13B749A05E237E06 ON se_roles;
+ALTER TABLE se_roles ADD `role_key` VARCHAR(60) DEFAULT NULL;
+UPDATE se_roles SET `role_key` = `name` WHERE `role_key` IS NULL;
+CREATE UNIQUE INDEX UNIQ_13B749A03EF22FDB ON se_roles (role_key);
+```
+
+### Remove getRole() method of Role entity
+
+The `getRole()` method of the `Role` class was removed as it is not part of the `RoleInterface`. 
+Use the `getIdentifier()` instead.
+
 ### Backdrop component
 
 Our `Backdrop` React component does not have the `local` and `open` props anymore. It is not supported anymore to render
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -5,9 +5,8 @@
 ### Add key property to Role entity
 
 To allow for referencing `Role` entities by a human readable string, a `key` property was added to the `Role` entity.
-Furthermore, the `key` property is used for generating the identifier of a role instead of the `name` property now.
-The following statements update the database to include the new property and set the key of each existing Role to 
-keep backwards compatibility:
+Furthermore, if the `key` is set, it is used for generating the identifier of a role instead of the `name` property.
+The following statements update the database to include the new property:
 
 ```sql
 DROP INDEX UNIQ_13B749A05E237E06 ON se_roles;
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -9,8 +9,7 @@ Furthermore, if the `key` is set, it is used for generating the identifier of a 
 The following statements update the database to include the new property:
 
 ```sql
-DROP INDEX UNIQ_13B749A05E237E06 ON se_roles;
-ALTER TABLE se_roles ADD `role_key` VARCHAR(60) DEFAULT NULL;
+ALTER TABLE se_roles ADD role_key VARCHAR(60) DEFAULT NULL;
 CREATE UNIQUE INDEX UNIQ_13B749A03EF22FDB ON se_roles (role_key);
 ```
 
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -12,14 +12,8 @@ keep backwards compatibility:
 ```sql
 DROP INDEX UNIQ_13B749A05E237E06 ON se_roles;
 ALTER TABLE se_roles ADD `role_key` VARCHAR(60) DEFAULT NULL;
-UPDATE se_roles SET `role_key` = `name` WHERE `role_key` IS NULL;
 CREATE UNIQUE INDEX UNIQ_13B749A03EF22FDB ON se_roles (role_key);
 ```
-
-### Remove getRole() method of Role entity
-
-The `getRole()` method of the `Role` class was removed as it is not part of the `RoleInterface`. 
-Use the `getIdentifier()` instead.
 
 ### Backdrop component
 
--- a/src/Sulu/Bundle/SecurityBundle/Controller/RoleController.php
+++ b/src/Sulu/Bundle/SecurityBundle/Controller/RoleController.php
@@ -16,7 +16,7 @@ use Doctrine\ORM\EntityManagerInterface;
 use FOS\RestBundle\Routing\ClassResourceInterface;
 use FOS\RestBundle\View\ViewHandlerInterface;
 use Sulu\Bundle\SecurityBundle\Entity\Permission;
-use Sulu\Bundle\SecurityBundle\Exception\RoleNameAlreadyExistsException;
+use Sulu\Bundle\SecurityBundle\Exception\RoleKeyAlreadyExistsException;
 use Sulu\Component\Rest\AbstractRestController;
 use Sulu\Component\Rest\Exception\EntityNotFoundException;
 use Sulu\Component\Rest\Exception\InvalidArgumentException;
@@ -188,6 +188,7 @@ class RoleController extends AbstractRestController implements ClassResourceInte
     public function postAction(Request $request)
     {
         $name = $request->get('name');
+        $key = $request->get('key');
         $system = $request->get('system');
 
         try {
@@ -195,12 +196,13 @@ class RoleController extends AbstractRestController implements ClassResourceInte
                 throw new InvalidArgumentException('Role', 'name');
             }
             if (null === $system) {
-                throw new InvalidArgumentException('Role', 'name');
+                throw new InvalidArgumentException('Role', 'system');
             }
 
             /** @var RoleInterface $role */
             $role = $this->roleRepository->createNew();
             $role->setName($name);
+            $role->setKey($key);
             $role->setSystem($system);
 
             $permissions = $request->get('permissions');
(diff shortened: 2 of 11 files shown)


UPGRADE.md
```diff
@@ -2,6 +2,25 @@
...
+keep backwards compatibility:
+
+```sql
+DROP INDEX UNIQ_13B749A05E237E06 ON se_roles;
+ALTER TABLE se_roles ADD `role_key` VARCHAR(60) DEFAULT NULL;
+UPDATE se_roles SET `role_key` = `name` WHERE `role_key` IS NULL;
+CREATE UNIQUE INDEX UNIQ_13B749A03EF22FDB ON se_roles (role_key);
+```
+
+### Remove getRole() method of Role entity
```
alexander-schranz, line 19: We should just deprecate the getRole. We should not make life heavier as it should be.

src/Sulu/Bundle/SecurityBundle/Entity/Role.php
```diff
@@ -93,14 +98,9 @@ public function getId()
...
 
-    public function getRole()
-    {
-        return 'ROLE_SULU_' . strtoupper($this->name);
-    }
-
     public function getIdentifier()
     {
-        return 'ROLE_SULU_' . strtoupper($this->getName());
+        return 'ROLE_SULU_' . strtoupper($this->getKey());
```
alexander-schranz, line 103: I would make the key nullable and fallback to the name makes the migration simplier and doesn't crash projects where developer did just `doctrine:schema:update` or where developer did build custom logic with ROLE_USER_{NAME}.

```suggestion
$key = $this->getKey();

if (!$key) {
    $key = $this->getName();
}

return 'ROLE_SULU_' . strtoupper($key);
```

src/Sulu/Bundle/SecurityBundle/Resources/config/doctrine/Role.orm.xml
```diff
@@ -8,7 +8,8 @@
             <generator strategy="AUTO"/>
         </id>
 
-        <field name="name" type="string" column="name" length="60" unique="true"/>
+        <field name="name" type="string" column="name" length="60"/>
+        <field name="key" type="string" length="60" column="role_key" nullable="true" unique="true"/>
```
alexander-schranz, line 12: would make the key nullable

src/Sulu/Bundle/SecurityBundle/Resources/config/forms/role_details.xml
```diff
@@ -15,7 +15,13 @@
...
         </property>
 
-        <property name="system" type="single_select" mandatory="true" colspan="3" spaceAfter="9">
+        <property name="key" type="text_line" colspan="9">
+            <meta>
+                <title>sulu_admin.key</title>
+            </meta>
+        </property>
+
+        <property name="system" type="single_select" mandatory="true" colspan="3">
```
alexander-schranz, line 24: would move system before key because key is optional

src/Sulu/Bundle/SecurityBundle/Entity/Role.php
```diff
@@ -127,6 +127,18 @@ public function getName()
...
     }
 
+    public function getKey()
+    {
+        return $this->key;
+    }
+
+    public function setKey($key)
+    {
+        $this->key = $key;
```
alexander-schranz, line 137: Can roles with spaces be configured in the firewall e.g. `ROLE_TEST USER`?

UPGRADE.md
```diff
@@ -2,6 +2,25 @@
...
+### Add key property to Role entity
+
+To allow for referencing `Role` entities by a human readable string, a `key` property was added to the `Role` entity.
+Furthermore, the `key` property is used for generating the identifier of a role instead of the `name` property now.
+The following statements update the database to include the new property and set the key of each existing Role to 
+keep backwards compatibility:
+
+```sql
+DROP INDEX UNIQ_13B749A05E237E06 ON se_roles;
+ALTER TABLE se_roles ADD `role_key` VARCHAR(60) DEFAULT NULL;
```
alexander-schranz, line 14: would make the key nullable

UPGRADE.md
```diff
@@ -2,6 +2,25 @@
...
+UPDATE se_roles SET `role_key` = `name` WHERE `role_key` IS NULL;
```
alexander-schranz, line 15: this would not be needed when making it nullable

src/Sulu/Bundle/SecurityBundle/Entity/Role.php
alexander-schranz: Also think its important to have the key nullable know as I think there are synchronisations with e.g Active Directories which will only update the role `NAME` and not the key. So I think its important to have the key nullable and fallback to the name in the getIdentifier method.

src/Sulu/Bundle/SecurityBundle/Entity/Role.php
```diff
@@ -93,14 +98,9 @@ public function getId()
         return $this->id;
     }
 
-    public function getRole()
-    {
-        return 'ROLE_SULU_' . strtoupper($this->name);
-    }
-
```
alexander-schranz, line 100:
```suggestion

    /**
     * @deprecated Deprecated since Sulu 2.1 will be removed in 3.0 use "getIdentifier" instead.
     */
    public function getRole()
    {
        @trigger_error(sprintf('The "%s" method is deprecated since Sulu 2.1, use "%s" instead.', __METHOD__, 'getIdentifier'), E_USER_DEPRECATED)
        return $this->getIdentifier();
    }
```

src/Sulu/Bundle/SecurityBundle/Entity/Role.php
nnatter: i think so, according to the docs a role only *must start with `ROLE_`*, otherwise its an arbitrary string:
https://symfony.com/doc/current/security.html#roles

src/Sulu/Bundle/SecurityBundle/Resources/config/doctrine/Role.orm.xml
nnatter: it is nullable 🙂

UPGRADE.md
nnatter: it is 🙂

UPGRADE.md
nnatter: it is not needed, but by doing this i dont need to fallback to the `name` in the `getIdentifier` method. but i am okay with implementing a fallback too

alexander-schranz: I would definitely remove this and do a fallback else it could end strange for this who implemented synchronisation scripts.

nnatter: alright 👍

src/Sulu/Bundle/SecurityBundle/Controller/RoleController.php
```diff
@@ -242,12 +244,14 @@ public function putAction(Request $request, $id)
         /** @var RoleInterface $role */
         $role = $this->roleRepository->findRoleById($id);
         $name = $request->get('name');
+        $key = $request->get('key');
```
alexander-schranz, line 247: We should always specify where to get the value: `$request->request->get('key')`.

src/Sulu/Bundle/SecurityBundle/Tests/Functional/Controller/RoleControllerTest.php
```diff
@@ -494,15 +503,16 @@ public function testPutNotExisting()
         $this->assertStringContainsString('11230', $response->message);
     }
 
-    public function testPutWithExistingName()
+    public function testPutWithExistingKey()
```
alexander-schranz, line 506: can we add a test where `'key' => null`

src/Sulu/Bundle/SecurityBundle/Exception/RoleKeyAlreadyExistsException.php
```diff
@@ -12,31 +12,31 @@
 namespace Sulu\Bundle\SecurityBundle\Exception;
 
 /**
- * Exception is thrown when a Role is created or updated with an already existing name.
+ * Exception is thrown when a Role is created or updated with an already existing key.
  */
-class RoleNameAlreadyExistsException extends \Exception
```
alexander-schranz, line 17: we should just deprecate this one

src/Sulu/Bundle/SecurityBundle/Resources/config/doctrine/Role.orm.xml
```diff
@@ -8,7 +8,8 @@
             <generator strategy="AUTO"/>
         </id>
 
-        <field name="name" type="string" column="name" length="60" unique="true"/>
+        <field name="name" type="string" column="name" length="60"/>
```
alexander-schranz, line 11: Would still keep the uniquiness

UPGRADE.md
```diff
@@ -2,6 +2,18 @@
...
 ## dev-master
 
+### Add key property to Role entity
+
+To allow for referencing `Role` entities by a human readable string, a `key` property was added to the `Role` entity.
+Furthermore, if the `key` is set, it is used for generating the identifier of a role instead of the `name` property.
+The following statements update the database to include the new property:
+
+```sql
+DROP INDEX UNIQ_13B749A05E237E06 ON se_roles;
```
alexander-schranz, line 12: we should keep the uniqueness of the name

src/Sulu/Bundle/SecurityBundle/Exception/RoleKeyAlreadyExistsException.php
alexander-schranz: if we keep uniquiness we should just keep or introduce a `RoleAlreadyExistsException`.

UPGRADE.md
nnatter: not sure about that 🤔 why would you keep it?

alexander-schranz: Role Name think also from UX be unique.

alexander-schranz: Also we still need to use the name and so should not change the behaviour is somebody rely on the RoleNameAlreadyExistsExcpetion

nnatter: well i will trust you here 🙂

alexander-schranz: I hope I don't disappoint you.

nnatter: me too 😉

Issue #5161, opened 2020-04-01
Allow to select webspace security system when editing role


nnatter:
| Q | A
| --- | ---
| Sulu Version | 2.0.6

# Actual Behavior

Sulu allows to configure a `Security System` inside of a [webspace configuration](https://github.com/sulu/sulu/blob/master/src/Sulu/Component/Webspace/Loader/schema/webspace/webspace-1.1.xsd#L16).
Although i can configure such a system, I cannot select it in the `System` dropdown when creating/editing a `User Role` via the Sulu admin interface.

# Expected Behavior

I would expect to be able to select the configured role in the `System` dropdown.

# Steps to Reproduce

1. Configure a new `Security System` inside of a webspace configuration with
```xml 
    <security>
        <system>Webspace System</system>
    </security>
```

2. Create a new Role via `Settings / User Roles / Add`

alexander-schranz:
duplicated to #4719 will close the other issue.

My Proposel would the `PageAdmin` should create the missing Security Contexts like the Community Bundle does it currently.

danrot:
Fixed in #5171

PR #5169, opened 2020-04-02
Use configured security system name in Admin classes instead of hardcoding it

nnatter:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | yes
| Deprecations? | no
| License | MIT

#### What's in this PR?

This PR adjust the `getSecurityContexts()` method of various (i hope i catched all of them) `Admin` classes to use the configured security system name instead of hardcoding `Sulu`.

#### Why?

Because the name of the security system used by the Sulu admin can be configured via `sulu_security.system`. If this is set to something other than `Sulu`, it is not possible to assign admin related permissions to a role of the configured system.

A positive side effect of this is that it makes it understandable what the key of the array returned by the `getSecurityContexts()` method represents.

#### BC Breaks/Deprecations

The constructor of various Admin classes was changed to inject the `sulu_security.system` parameter.

### DIFF
--- a/src/Sulu/Bundle/AudienceTargetingBundle/Admin/AudienceTargetingAdmin.php
+++ b/src/Sulu/Bundle/AudienceTargetingBundle/Admin/AudienceTargetingAdmin.php
@@ -50,14 +50,21 @@ class AudienceTargetingAdmin extends Admin
      */
     private $ruleCollection;
 
+    /**
+     * @var string
+     */
+    private $suluSecuritySystem;
+
     public function __construct(
         ViewBuilderFactoryInterface $viewBuilderFactory,
         RuleCollectionInterface $ruleCollection,
-        SecurityCheckerInterface $securityChecker
+        SecurityCheckerInterface $securityChecker,
+        string $suluSecuritySystem
     ) {
         $this->viewBuilderFactory = $viewBuilderFactory;
         $this->ruleCollection = $ruleCollection;
         $this->securityChecker = $securityChecker;
+        $this->suluSecuritySystem = $suluSecuritySystem;
     }
 
     public function configureNavigationItems(NavigationItemCollection $navigationItemCollection): void
@@ -142,7 +149,7 @@ class AudienceTargetingAdmin extends Admin
     public function getSecurityContexts()
     {
         return [
-            'Sulu' => [
+            $this->suluSecuritySystem => [
                 'Settings' => [
                     self::SECURITY_CONTEXT => [
                         PermissionTypes::VIEW,
--- a/src/Sulu/Bundle/AudienceTargetingBundle/Resources/config/services.xml
+++ b/src/Sulu/Bundle/AudienceTargetingBundle/Resources/config/services.xml
@@ -11,6 +11,7 @@
             <argument type="service" id="Sulu\Bundle\AdminBundle\Admin\View\ViewBuilderFactoryInterface"/>
             <argument type="service" id="sulu_audience_targeting.rules_collection"/>
             <argument type="service" id="sulu_security.security_checker"/>
+            <argument>%sulu_security.system%</argument>
 
             <tag name="sulu.admin"/>
             <tag name="sulu.context" context="admin"/>
--- a/src/Sulu/Bundle/CategoryBundle/Admin/CategoryAdmin.php
+++ b/src/Sulu/Bundle/CategoryBundle/Admin/CategoryAdmin.php
@@ -46,14 +46,21 @@ class CategoryAdmin extends Admin
      */
     private $localizationManager;
 
+    /**
+     * @var string
+     */
+    private $suluSecuritySystem;
+
     public function __construct(
         ViewBuilderFactoryInterface $viewBuilderFactory,
         SecurityCheckerInterface $securityChecker,
-        LocalizationManagerInterface $localizationManager
+        LocalizationManagerInterface $localizationManager,
+        string $suluSecuritySystem
     ) {
         $this->viewBuilderFactory = $viewBuilderFactory;
         $this->securityChecker = $securityChecker;
         $this->localizationManager = $localizationManager;
+        $this->suluSecuritySystem = $suluSecuritySystem;
     }
 
     public function configureNavigationItems(NavigationItemCollection $navigationItemCollection): void
@@ -165,7 +172,7 @@ class CategoryAdmin extends Admin
     public function getSecurityContexts()
     {
         return [
-            'Sulu' => [
+            $this->suluSecuritySystem => [
                 'Settings' => [
                     static::SECURITY_CONTEXT => [
                         PermissionTypes::VIEW,
--- a/src/Sulu/Bundle/CategoryBundle/Resources/config/services.xml
+++ b/src/Sulu/Bundle/CategoryBundle/Resources/config/services.xml
@@ -4,11 +4,13 @@
            xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
     <services>
         <service id="sulu_category.admin" class="Sulu\Bundle\CategoryBundle\Admin\CategoryAdmin">
-            <tag name="sulu.admin"/>
-            <tag name="sulu.context" context="admin"/>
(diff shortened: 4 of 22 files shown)


src/Sulu/Bundle/AudienceTargetingBundle/Admin/AudienceTargetingAdmin.php
```diff
@@ -50,14 +50,21 @@ class AudienceTargetingAdmin extends Admin
...
+     * @var string
+     */
+    private $suluSecuritySystem;
+
     public function __construct(
         ViewBuilderFactoryInterface $viewBuilderFactory,
         RuleCollectionInterface $ruleCollection,
-        SecurityCheckerInterface $securityChecker
+        SecurityCheckerInterface $securityChecker,
+        string $suluSecuritySystem
```
alexander-schranz, line 62: we need to set here a default value to avoid bc break `string $suluSecuritySystem = 'SULU'`

danrot: Since these classes are only instantiated resp. registered as a service by us, I would say that this is not necessary.

danrot:
I am also curious: Why did you have to change that value?

src/Sulu/Bundle/AudienceTargetingBundle/Admin/AudienceTargetingAdmin.php
alexander-schranz: If you did overwrite the service it will fail when adding a new required constructor parameter.

alexander-schranz: But I hope we can avoid injecting it and go with a constants instead :)

nnatter:
> I am also curious: Why did you have to change that value?

I did not need to change it. Stumbled upon the configuration/parameter when debugging #5161 and tried to change it 🙂 I think I would be okay with using a constant too. But I guess there was i reason why this was introduced in [1.2.1](https://github.com/sulu/sulu/blob/master/UPGRADE.md)?

src/Sulu/Bundle/AudienceTargetingBundle/Admin/AudienceTargetingAdmin.php
danrot: If we want to fix the bug that the security system is not read from the configuration correctly, I would definitely not set it as a default value here. That would mean that if somebody overwrote this service, the bug would still persist, without that person even realizing it.

Where would you put that constant? That's always the tricky part... It is also harder to override then, but it is questionable if we have to offer the option to change that at all.

That somehow contratics [@nnatter's comment below](https://github.com/sulu/sulu/pull/5169#issuecomment-607823063), but I think back then it was introduced in order to be able to set a different value for the system in the `WebsiteKernel`, so that the same `UserProvider` can be used to login a user in the frontend. I think that's not how we handle this anymore, right? When thinking about it, it would actually be less dependencies on our two kernel setup, if we get rid of that.

nnatter: If we agree that the `%sulu_security.system%` basically contains the security system of the admin, i would propose to put a `SULU_ADMIN_SECURITY_SYSTEM` on the `Sulu\Bundle\AdminBundle\Admin\Admin` class 🙂

alexander-schranz: Would use `SECURITY_SYSTEM` as prefix (so all are prefixed with the same if somebody adds a new one) and remove ADMIN from the constant as it does not say that its only for admin so I would go with: `public const SECURITY_SYSTEM_SULU = 'Sulu';`

danrot: For me the problem is that it is not clearly related to the `Admin`, but to a few otherwise unrelated places as well. So I am not sure where to put it, and I can't think of any class I would  like to have that constant. Therefore I would personally prefer to keep it as a configuration value, because this somehow hovers over all of them, and we are not forced to think about which class would be the best owner.

Also, what would we be doing with [that parameter](https://github.com/sulu/sulu/blob/9f6c619/src/Sulu/Bundle/SecurityBundle/DependencyInjection/SuluSecurityExtension.php#L37)? Simply deleting it? Keeping it, but still using the constant and making that parameter obsolete? Both options are also BC breaks.

And is it still possible to set it do a different system for the `WebsiteKernel` and use then a firewall with our `UserProvider`? If yes some people might be using it, this would be another argument for me to keep the config value.

nnatter: >  it does not say that its only for admin

> For me the problem is that it is not clearly related to the Admin, but to a few otherwise unrelated places as well.

in what other places is it used? do we recommend to use it there?

> Also, what would we be doing with that parameter? Simply deleting it? Keeping it, but still using the constant and making that parameter obsolete? Both options are also BC breaks.

I think i would be okay with deprecating it (an not using it anywhere on our side) as it does not work (at least in the admin context) anyway at the moment 🙂 

> And is it still possible to set it do a different system for the WebsiteKernel and use then a firewall with our UserProvider?

Yes i think this is still possible. If no security system is configured in the webspace config, the `UserProvider` will use the configured `suluSecuritySystem`.

I understand the arguments of both sides. I think that it should be easy to register a custom admin class too. Requiring a string that is not autowirable does definitely not help in that regard. 

Maybe we can use a service which is autowirable and provide the value via a getter? From the existing services in the `SecurityBundle`, the `UserManager` would probably be the most suitable but we could also introduce a new one?

danrot: The parameter has to be used e.g. in the [UserProvider](https://github.com/sulu/sulu/blob/8dfc342/src/Sulu/Bundle/SecurityBundle/User/UserProvider.php#L120). This is what makes the last thing I mentioned possible.

Simply deprecating it and living with the fact that the admin breaks if you set it to something else in the `admin` context would also be OK with me. But I actually thought the reason for this PR is to make this work :see_no_evil: 

But if we deprecate it, we have to provide an update path, what would that be? Use the `system` from the webspace configuration instead? Does that work in the same way?

src/Sulu/Bundle/AudienceTargetingBundle/Admin/AudienceTargetingAdmin.php
```diff
@@ -142,7 +149,7 @@ public function configureViews(ViewCollection $viewCollection): void
     public function getSecurityContexts()
     {
         return [
-            'Sulu' => [
+            $this->suluSecuritySystem => [
```
alexander-schranz, line 152: I think we are mixing here something. The parameter used by the security system is I think not the parameter which should be used here. So I think it is correct that this was hardcoded before.

danrot: I don't think we are mixing anything here. If you change the other parameter, and the user can't login aftewards, that's clearly a bug, so it is definitely not correct that it is currently hardcoded. I don't know why somebody would want to change that, but if they do it should continue to work. And this seems like the proper fix to me. And I would accept the BC break for that.

But as already said, I'd be also fine with deprecating that parameter, recommend to use the `system` tag in the webspace instead, live with the fact that it currently does not work, and completely remove it with the next major release.

alexander-schranz: If we go on the parameter its a behaviour change on website context. It will set the permission registered in every admin also to the new security system. But the permission defined in the Admin classes are only for the Sulu security system not the website context security system which the user did define.

danrot: Since the admin class should only be defined to be registered in the `admin` context, this should not be a problem. However, as already said, I think the easiest way forward is simply deprecating the configuration value and keep the behavior as it is. I don't think that many people changed that value in the admin configuration anyway.

danrot:
Replaced by #5199

PR #5170, opened 2020-04-02
Add separate filter type for countries

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | #5035
| License | MIT
| Documentation PR | ---

#### What's in this PR?

We have decided to add a separate filter type for countries, because this way we can reuse the country data that is already delivered via the `/admin/config` request to the frontend. We also decided to use multiple checkboxes instead of a dropdown in the `FieldFilterItem`, and included a search field, because there are quite some countries.

#### Why?

Because filtering for countries is probably a very popular use case.


danrot:
@chirimoya This is how it currently looks like, is that ok? You can also check out the branch, everything should work.

Screenshot from 2020-04-02 16-31-07

The other question is if/how we replace the dropdowns in the other field types. For the `DropdownFieldFilterType` it would be very easy to do it the same way as it is done here for the countries.

For the `SelctionFieldFilterType` stuff is a little bit more complicated... There we are currently using the `ResourceMultiSelect`. I see 3 different ways of solving this:

1. We could add a `skin` property to the `ResourceMultiSelect`, which decides if a `Select` or a `Checkbox`es are used. However, this might be problematic, because this might end up in the props being the union of `Select` and `Checkbox` (e.g. currently we have the `allSelectedText` on the `ResourceMultiSelect`, which would do nothing when checkboxes are used)
2. Introduce a completely new `ResourceCheckboxGroup` component, which does the same as the `ResourceMultiSelect` but only for checkboxes. We might end up duplicating some code, but it would result in a nicer interface for the component.
3. Extract the generic part from the `ResourceMultiSelect` into a separate component and reuse it in the `ResourceMultiSelect`. Less code duplication, but feels like a higher maintaining burden to me.
4. Implement everything that would come into the `ResourceCheckboxGroup` directly in the `SelectionFieldFilterType` (which would make this type more bloated additionally to the already existing `type` prop there, which changes its appearances compeltely)

@sulu/core-team Any opinions?

nnatter:
I think this looks nice 🙂 Would maybe add a bit more space between the search field and the items. But I guess this should be decided by a designer.

Regarding to the implementation - I would probably go with option 2. Not sure if it is worth the hassle to implement both "skins" into a single component. Do you see any advantages of that approach?

chirimoya:
@danrot looks good. Would add a search indicator to the input field.

danrot:
@nnatter The only advantage would be that there is less code duplication, but since it is not a lot I would also prefer Option 2.

src/Sulu/Bundle/AdminBundle/Resources/js/components/Checkbox/CheckboxGroup.js
```diff
@@ -0,0 +1,44 @@
...
+    };
+
+    render() {
+        const {className, disabled, values} = this.props;
+
+        return (
+            <div className={className}>
+                {React.Children.map(this.props.children, (child) => {
+                    return React.cloneElement(child, {
+                        checked: !!values && values.includes(child.props.value),
```
nnatter, line 36: does flow expect the `!!values` check here? because this should always be an array according to the props type

src/Sulu/Bundle/ContactBundle/Resources/js/containers/List/fieldFilterTypes/CountryFieldFilterType.js
```diff
@@ -0,0 +1,51 @@
+// @flow
+import React, {Fragment} from 'react';
+import {action, observable} from 'mobx';
+import {Checkbox, CheckboxGroup, Input} from 'sulu-admin-bundle/components';
+import {AbstractFieldFilterType} from 'sulu-admin-bundle/containers';
+import countryFieldFilterTypeStyles from './countryFieldFilterType.scss';
+
+class CountryFieldFilterType extends AbstractFieldFilterType<?Array<string>> {
+    static countries: {[key: string]: string} = {};
```
nnatter, line 9: do you see advantages in setting the countries to this static property instead of accessing the `Config` service? do we use the same strategy for other config values?

src/Sulu/Bundle/AdminBundle/Resources/js/components/Checkbox/CheckboxGroup.js
danrot: TBH I just copied this component from somewhere and adjusted it... Seems like this is really not necessary :slightly_smiling_face:

src/Sulu/Bundle/ContactBundle/Resources/js/containers/List/fieldFilterTypes/CountryFieldFilterType.js
danrot: Yes, that I don't have to use the `Config` service :see_no_evil: That results in less coupling, and makes it easier to test, because the `Config` service does not need to be mocked. Also, it is [what we are doing with other components as well](https://github.com/sulu/sulu/blob/a32bf97/src/Sulu/Bundle/ContactBundle/Resources/js/index.js#L40-L49).

nnatter: alright 🙂

PR #5174, opened 2020-04-03
Use checkboxes instead of MultiSelect for DropdownFilterType

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR uses checkboxes instead of a multi select for e.g. type filters like for medias.

#### Why?

Because they are more comfortable to use, and require one less click. We also avoid to `Popover`s being displayed at the same time.

#### To Do

- [ ] Rename `dropdown` filter type to something else?


danrot:
@sulu/core-team Should we rename the `dropdown` key for this filter? It somehow still is a dropdown, but every filter is a dropdown if we consider it like this. Should we maybe rename it to `select`? So that it is clear that something is chosen from a predefined set of values? Or does anybody have a better idea, or do you still feel comfortable with `dropdown`?

nnatter:
I think renaming it to `select` is good idea. That way, the naming is consistent with our field types 🙂

danrot:
Or what about `selection`? Our `selection` field type is also more generic, and means just "select something" and offers different ways to do so (if you think about a `select` in HTML you also end up as a `dropdown`...)

nnatter:
using that line of reasoning, the `auto_complete` is a `selection` too.. 🙂 that way there would a `selection` type and a `auto_complete` type in the `SelectionFilterType` right?

danrot:
Didn't think about the fact that there is already a `SelectionFieldFilterType` :see_no_evil: And the `selection` field type from the forms is also only a selection from entities, not some hardcoded values... I think I would still go with `select` then, seems at least less wrong than `dropdown` :see_no_evil:

nnatter:
@danrot looks like the styleguide build is failing 🤔

danrot:
It looks to me like this is not related to this PR, but to a new release of webpack's `css-loader`, because it also fails in other PRs... I'll wait till tomorrow morning, if they didn't realize it until then I am going to create an issue.

nnatter:
alright - will merge this one 🙂

PR #5175, opened 2020-04-03
Add swiftmailer listener plugin

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Adds a markup listener for `sulu-link` as a swiftmailer plugin

#### Why?

The body of e-mails sent with the swiftmailer are not parsed by the markup bundle and thus the `sulu-link` tag is not replaced with the correct `<a>` element.

### DIFF
--- a/src/Sulu/Bundle/MarkupBundle/DependencyInjection/CompilerPass/ParserCompilerPass.php
+++ b/src/Sulu/Bundle/MarkupBundle/DependencyInjection/CompilerPass/ParserCompilerPass.php
@@ -22,7 +22,7 @@ class ParserCompilerPass implements CompilerPassInterface
 {
     const SERVICE_IDS = [
         'sulu_markup.response_listener',
-        'sulu_markup.swift_mailer_listener'
+        'sulu_markup.swift_mailer_listener',
     ];
 
     const TAG_NAME = 'sulu_markup.parser';
--- a/src/Sulu/Bundle/MarkupBundle/DependencyInjection/CompilerPass/ParserCompilerPass.php
+++ b/src/Sulu/Bundle/MarkupBundle/DependencyInjection/CompilerPass/ParserCompilerPass.php
@@ -20,10 +20,7 @@ use Symfony\Component\DependencyInjection\Reference;
  */
 class ParserCompilerPass implements CompilerPassInterface
 {
-    const SERVICE_IDS = [
-        'sulu_markup.response_listener',
-        'sulu_markup.swift_mailer_listener',
-    ];
+    const SERVICE_ID = 'sulu_markup.response_listener';
 
     const TAG_NAME = 'sulu_markup.parser';
 
@@ -31,23 +28,21 @@ class ParserCompilerPass implements CompilerPassInterface
 
     public function process(ContainerBuilder $container)
     {
-        foreach (self::SERVICE_IDS as $serviceId) {
-            if (!$container->hasDefinition($serviceId)) {
-                continue;
-            }
-
-            $references = [];
-            foreach ($container->findTaggedServiceIds(self::TAG_NAME) as $id => $tags) {
-                foreach ($tags as $attributes) {
-                    $references[$attributes[self::TYPE_ATTRIBUTE]] = new Reference($id);
-                }
-            }
-
-            if (0 === count($references)) {
-                continue;
-            }
-
-            $container->getDefinition($serviceId)->replaceArgument(0, $references);
+        if (!$container->hasDefinition(self::SERVICE_ID)) {
+            return;
         }
+
+        $references = [];
+        foreach ($container->findTaggedServiceIds(self::TAG_NAME) as $id => $tags) {
+            foreach ($tags as $attributes) {
+                $references[$attributes[self::TYPE_ATTRIBUTE]] = new Reference($id);
+            }
+        }
+
+        if (0 === count($references)) {
+            return;
+        }
+
+        $container->getDefinition(self::SERVICE_ID)->replaceArgument(0, $references);
     }
 }
--- a/src/Sulu/Bundle/MarkupBundle/DependencyInjection/CompilerPass/ParserCompilerPass.php
+++ b/src/Sulu/Bundle/MarkupBundle/DependencyInjection/CompilerPass/ParserCompilerPass.php
@@ -20,7 +20,10 @@ use Symfony\Component\DependencyInjection\Reference;
  */
 class ParserCompilerPass implements CompilerPassInterface
 {
-    const SERVICE_ID = 'sulu_markup.response_listener';
+    const SERVICE_IDS = [
+        'sulu_markup.response_listener',
+        'sulu_markup.swift_mailer_listener'
+    ];
 
     const TAG_NAME = 'sulu_markup.parser';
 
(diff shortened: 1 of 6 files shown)


src/Sulu/Bundle/MarkupBundle/Listener/SwiftMailerListener.php
```diff
@@ -0,0 +1,66 @@
...
+        $message = $event->getMessage();
+
+        $body = $message->getBody();
+        $format = $message->getBodyContentType();
+
+        if (count($explodedFormat = explode('/', $format)) > 1) {
+            $format = $explodedFormat[1];
+        }
+
+        $locale = $this->requestStack->getCurrentRequest()->getLocale();
```
alexander-schranz, line 48: `getCurrentRequest` could be empty (e.g. emails send from cli) we need to fallback here to the default (%kernel.locale%).

src/Sulu/Bundle/MarkupBundle/DependencyInjection/CompilerPass/ParserCompilerPass.php
```diff
@@ -20,29 +20,34 @@
  */
 class ParserCompilerPass implements CompilerPassInterface
 {
-    const SERVICE_ID = 'sulu_markup.response_listener';
+    const SERVICE_IDS = [
+        'sulu_markup.response_listener',
+        'sulu_markup.swift_mailer_listener'
```
alexander-schranz, line 25: for new services we should use something like here: https://github.com/sulu/SuluHeadlessBundle/pull/20/files#diff-66e8662ac6150c33e219664495191cf6R105

so no changes in this compilerpass should be needed

src/Sulu/Bundle/MarkupBundle/Tests/Unit/Listener/SwiftMailerListenerTest.php
```diff
@@ -0,0 +1,82 @@
...
+
+        $this->listener = new SwiftMailerListener(['html' => $this->markupParser->reveal()], $this->requestStack->reveal());
+    }
+
+    public function testReplaceMarkup(): void
+    {
+        /** @var Request|ObjectProphecy $request */
+        $request = $this->prophesize(Request::class);
+        $request->getLocale()->willReturn('de');
+        $this->requestStack->getCurrentRequest()->willReturn($request->reveal());
```
alexander-schranz, line 70: add test where request is null

src/Sulu/Bundle/MarkupBundle/Markup/HtmlMarkupParser.php
```diff
@@ -60,12 +60,17 @@ public function validate($content, $locale)
...
+            $tags = $this->tagRegistry->getTag($tagMatchGroup->getTagName(), $this->getType(), $tagMatchGroup->getNamespace())
                 ->validateAll($tagMatchGroup->getTags(), $locale);
 
             $result = array_merge($result, $tags);
         }
 
         return $result;
     }
+
+    public function getType(): string
```
Prokyonn, line 72: @alexander-schranz I guess we cannot take this method into the interface because of bc right?

alexander-schranz: Would not add this method just index by the key of the tagged service. See https://symfony.com/blog/new-in-symfony-4-3-indexed-and-tagged-service-collections

src/Sulu/Bundle/MarkupBundle/Tests/Unit/Listener/SwiftMailerListenerTest.php
```diff
@@ -0,0 +1,107 @@
...
+        $this->simpleMessage->setBody('<html><a href="/test">Page-Title</a></html>')->shouldBeCalled();
+
+        $this->listener->beforeSendPerformed($this->event->reveal());
+    }
+
+    public function testReplaceMarkupWithoutRequest(): void
+    {
+        $this->requestStack->getCurrentRequest()->willReturn(null);
+
+        $this->simpleMessage->getBodyContentType()->wilLReturn('text/html');
```
alexander-schranz, line 97: `wilL`

src/Sulu/Bundle/MarkupBundle/Tests/Unit/Listener/SwiftMailerListenerTest.php
```diff
@@ -0,0 +1,107 @@
...
+
+        $this->markupParser->parse('<html><sulu-link href="123-123-123"/></html>', 'de')
+            ->willReturn('<html><a href="/test">Page-Title</a></html>')->shouldBeCalled();
+
+        $this->simpleMessage->setBody('<html><a href="/test">Page-Title</a></html>')->shouldBeCalled();
+
+        $this->listener->beforeSendPerformed($this->event->reveal());
+    }
+
+    public function testReplaceMarkupWithoutRequest(): void
```
alexander-schranz, line 93: we should also add a test with a unkown `contentType` e.g.: `text/plain`

src/Sulu/Bundle/MarkupBundle/Tests/Unit/Listener/SwiftMailerListenerTest.php
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    public function testReplaceMarkup(): void
+    {
+        /** @var Request|ObjectProphecy $request */
+        $request = $this->prophesize(Request::class);
+        $request->getLocale()->willReturn('de');
+        $this->requestStack->getCurrentRequest()->willReturn($request->reveal());
+
+        $this->simpleMessage->getBodyContentType()->wilLReturn('text/html');
```
alexander-schranz, line 82: `wilL`

src/Sulu/Bundle/MarkupBundle/Tests/Unit/Listener/SwiftMailerListenerTest.php
```diff
@@ -0,0 +1,125 @@
...
+        $request->getLocale()->willReturn('de');
+        $this->requestStack->getCurrentRequest()->willReturn($request->reveal());
+
+        $this->simpleMessage->getBodyContentType()->willReturn('text/plain');
+        $this->simpleMessage->getBody()->willReturn('<html><sulu-link href="123-123-123"/></html>');
+
+        $this->markupParser->parse('<html><sulu-link href="123-123-123"/></html>', 'de')
+            ->shouldNotBeCalled();
+
+        $this->simpleMessage->setBody('<html><a href="/test">Page-Title</a></html>')->shouldNotBeCalled();
```
alexander-schranz, line 121: why does it still get replaced its unkown response type?

alexander-schranz: Oh sorry did not see shouldNotBeCalled 🙈

alexander-schranz:
Code looks good for me need to test it in combination with form bundle.

alexander-schranz:
Works, the only sad thing is that the symfony profiler does not show the replaced link but think we can do there nothing.

PR #5176, opened 2020-04-03
Use new ResourceCheckboxGroup for SelectionFieldFilterType

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR introuces a `ResourceCheckboxGroup` component which displays checkboxes for all entries from a given resource and makes it available to select via checkboxes.

Then this component will be used in the `SelectionFieldFilterType`, which is e.g. used in the `position` field of the `contacts`.

#### Why?

Because it is more convenient and requires less clicks.


src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/fieldFilterTypes/SelectionFieldFilterType.js
```diff
@@ -123,10 +123,9 @@ class SelectionFieldFilterType extends AbstractFieldFilterType<?Array<string | n
                     />
                 }
                 {this.type === TYPE_SELECT &&
```
danrot, line 125: @nnatter I think this is where calling it select starts feeling weird :see_no_evil: Referring to [your comment](https://github.com/sulu/sulu/pull/5174#issuecomment-608378649) in #5174.

nnatter: there is a `[single_select` type](https://github.com/sulu/sulu/blob/master/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/SingleSelection.js#L66) in the `SingleSelection` too, so at least things would be consistent 🙂

danrot: I don't see how this would be consistent. The `single_select` type is actually a  `SingleSelect` component, which is not the case here.

nnatter: From the user perspective:

There is a `single_select` field-type and a `single_selection` field-type. 
If i use a `single_selection` field-type with the `single_select` parameter, it looks like the `single_select` field-type.

Now there is a new `select` filter-type and a `selection` filter-type.
If I want to display a `selection` filter-type that looks like the `select` filter type, it would be consistent to use `select` as parameter value.

danrot: The relationships between these two terms might be the same, but `select` still means something different in both cases, also from a user's perspective.

nnatter: I am not sure about that - to me, both of them are input elements that allow me to select options from a list of available options. What is the difference in your opinion?

danrot: Keep in mind that the developer has to set this value, and if they ready `select` they probably gonna think of a HTML `select` element instead of a list of checkboxes. But I don't feel like this discussion leads anywhere, and I don't have a better idea anyway... As already stated in [the other PR](https://github.com/sulu/sulu/pull/5174) I guess I am going with `select` anyway.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/ResourceCheckboxGroup/tests/ResourceCheckboxGroup.test.js
```diff
@@ -0,0 +1,304 @@
...
+
+    expect(resourceCheckboxGroup.find('Loader')).toHaveLength(1);
+});
+
+test('Pass requestParameters', () => {
+    // $FlowFixMe
+    ResourceListStore.mockImplementation(function() {
+        this.loading = false;
+        this.data = [
+            {
```
nnatter, line 105: dont think there is any data needed for this test. but shouldnt hurt anything

PR #5180, opened 2020-04-06
Use central cache directory for symfony filesystem cache pool

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes -
| Related issues/PRs | https://github.com/sulu/skeleton/pull/45
| License | MIT
| Documentation PR | -

#### What's in this PR?

Use central cache directory for symfony filesystem cache pool.

#### Why?

When change the cache adapter to something like like apc or redis which is more common in production it would behave the same as using the same directory. So there should be no difference when using the filesystem cache or something else and in future we want to go more to be single kernel.

### DIFF
--- a/src/Sulu/Bundle/CoreBundle/DependencyInjection/SuluCoreExtension.php
+++ b/src/Sulu/Bundle/CoreBundle/DependencyInjection/SuluCoreExtension.php
@@ -214,7 +214,7 @@ class SuluCoreExtension extends Extension implements PrependExtensionInterface
                 'framework',
                 [
                     'cache' => [
-                        'directory' => '%sulu.common_cache_dir%',
+                        'directory' => '%sulu.common_cache_dir%/pools',
                     ],
                 ]
             );
--- a/src/Sulu/Bundle/CoreBundle/DependencyInjection/SuluCoreExtension.php
+++ b/src/Sulu/Bundle/CoreBundle/DependencyInjection/SuluCoreExtension.php
@@ -208,6 +208,17 @@ class SuluCoreExtension extends Extension implements PrependExtensionInterface
                 ]
             );
         }
+
+        if ($container->hasExtension('framework')) {
+            $container->prependExtensionConfig(
+                'framework',
+                [
+                    'cache' => [
+                        'directory' => '%sulu.common_cache_dir%',
+                    ],
+                ]
+            );
+        }
     }
 
     public function load(array $configs, ContainerBuilder $container)


src/Sulu/Bundle/CoreBundle/DependencyInjection/SuluCoreExtension.php
```diff
@@ -208,6 +208,17 @@ public function prepend(ContainerBuilder $container)
                 ]
             );
         }
+
+        if ($container->hasExtension('framework')) {
+            $container->prependExtensionConfig(
+                'framework',
+                [
+                    'cache' => [
+                        'directory' => '%sulu.common_cache_dir%',
```
danrot, line 217: Is it possible that this is not the only place where we have to fix this? It is still creating some cache entries in the `admin` and `website` cache folders for me. Actually it seems to me like only the collaborations created by Sulu are stored in the `common` folder this way.

danrot: FYI: I tested by just running `bin/console server:run` in sulu/sulu.

alexander-schranz: Currently it seems only the `cache.app` which use the `cache.adapter.filesystem` this directory is used and set correctly here:

https://github.com/symfony/symfony/blob/bfe6b6fb41b677a35e7fbc89fcc7d9382be0c41b/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php#L1806-L1807

The cache directory for the `system` cache is not replaced as you see and is hardcoded in:

https://github.com/symfony/symfony/blob/bfe6b6fb41b677a35e7fbc89fcc7d9382be0c41b/src/Symfony/Bundle/FrameworkBundle/Resources/config/cache.xml#L49

Not sure if this is a bug on symfony side.

danrot:
What bug does this try to solve exactly? I think now we might still have a different behavior now when  used in combination with redis, right?

src/Sulu/Bundle/CoreBundle/DependencyInjection/SuluCoreExtension.php
alexander-schranz: I created a PR on symfony side lets see what the will do https://github.com/symfony/symfony/pull/36378. But it has no affect on us as we don't use the `cache.system`.

alexander-schranz:
> What bug does this try to solve exactly? I think now we might still have a different behavior now when used in combination with redis, right?

Currently if you did use filesystem cache, the files where written differently for contexts `var/cache/admin` and `var/cache/website`. But if you use redis cache or apcu the cache was central. With this configuration we make sure that the filesystem cache behave the same as when using the redis or apcu cache adapter.

danrot:
But for what exact use case is that problematic? Or is there currently no problem with that at all? The only feature that is affected by this is the collaboration feature, which only runs in the admin anyway.

alexander-schranz:
@danrot need it will be for PHPCR Node cache PR as linked above (https://github.com/sulu/skeleton/pull/45). So that the cache is invalided correctly also when using filesystem cache and not redis.

PR #5181, opened 2020-04-06
Remove pretty as a dependency

danrot:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR removes the `pretty` dependency for testing.
#### Why?

`pretty` was used when we used `Portal`s to display content outside of the current component. Old versions of `enzyme` couldn't handle that, so we access the `document` variable and made the HTML better readable by using `pretty`.

There was recently another upgrade of `pretty` which broke some of our tests. In addition to that `enzyme` can now access the portals anyway, so it is not absolutely necessary anymore. So I removed it in order to hopefully lower the number of times our snapshots are broken.

While I were at it, I removed some snapshot tests, when a very similar one already existed. E.g. if a snapshot for the component in its disabled and enabled state. Instead of doing two snapshots I used some assertion in the one for the disabled state.


src/Sulu/Bundle/AdminBundle/Resources/js/components/ColorPicker/tests/ColorPicker.test.js
```diff
@@ -1,66 +1,38 @@
...
-    const value = '#abc';
-    expect(render(<ColorPicker onChange={onChange} value={value} />)).toMatchSnapshot();
-});
-
-test('ColorPicker should render when disabled', () => {
-    const onChange = jest.fn();
-    const value = '#abc';
-    expect(render(<ColorPicker disabled={true} onChange={onChange} value={value} />)).toMatchSnapshot();
-});
+test('ColorPicker should disabled Input when disabled', () => {
```
nnatter, line 11: should disable

src/Sulu/Bundle/AdminBundle/Resources/js/components/ColorPicker/tests/ColorPicker.test.js
```diff
@@ -1,66 +1,38 @@
 // @flow
 import React from 'react';
-import {render, mount} from 'enzyme';
-import pretty from 'pretty';
+import {mount} from 'enzyme';
 import ColorPicker from '../ColorPicker';
 
-afterEach(() => {
```
nnatter, line 7: this is good to see 🙂

src/Sulu/Bundle/AdminBundle/Resources/js/components/Popover/tests/Popover.test.js
```diff
@@ -54,13 +50,16 @@ test('The popover should render in body when open', () => {
...
     );
-    expect(view.find('Backdrop')).toHaveLength(1);
-    expect(view.find('Backdrop').prop('open')).toEqual(true);
-    expect(pretty(body.innerHTML)).toMatchSnapshot();
+
+    popover.instance().popoverWidth = 20;
+    popover.instance().popoverHeight = 100;
+    popover.update();
+
+    expect(popover.instance().dimensions.scrollTop).toBe(4);
```
nnatter, line 58: should i understand this? 🙈

danrot: I don't understand it completely as well :see_no_evil: But this is a check if the positioning is working, which does not seem to reflected in the snapshot, so I included it here as well. Didn't want to delete too much :see_no_evil:

PR #5182, opened 2020-04-06
Remove Preview Kernel class

### LINKED ISSUE #4782: Previewkernel expects an App\Kernel
| Q | A
| --- | ---
| Bug? | no
| New Feature? | no
| Sulu Version | 2.0.0-RC3

#### Actual Behavior

The `PreviewKernel` extends the `App/Kernel` [here](https://github.com/sulu/sulu/blob/develop/src/Sulu/Bundle/PreviewBundle/Preview/Renderer/PreviewKernel.php#L14). This means you can't change the namespace of your `src/Kernel` without adding an alias for this class.

#### Expected Behavior

I expect I should be able to change the namespace of my project `src` dir without breaking the preview.

#### Steps to Reproduce

- Create a new project based on [sulu/skeleton](https://github.com/sulu/skeleton)
- Change the namespace of your `src` dir and in the `Kernel` file.
- Edit a page in Sulu and an error will be thrown in the preview.

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #4782
| License | MIT

#### What's in this PR?

Remove Preview Kernel class.

#### Why?

Currently its not possible to control how the PreviewKernel is created by the Application. By moving it to `createPreviewKernel` function on the SuluKernel the project can control how the PreviewKernel is created and with with parameters.

#### To Do


danrot:
TBH, I don't think that this is a good idea. Now the generic `SuluKernel` has to know a lot about how the `PreviewBundle` works. IMO it should be the other way round. I think achieving this by decorating the `WebsiteKernel` would be the better approach. The `PreviewKernel` could then pass many things to the `SuluKernel`, and only adjust what is necessary. This way the `SuluKernel` would need no knowledge at all, and the same structure would work if we need another kernel like the `PreviewKernel` at some point.

alexander-schranz:
I will close this PR as I think this is also part of the `One Kernel` setup issue. 

In the future it will be the task of the `SuluContentBundle` and the a new `ContentTypeResolverInterface` (maybe FieldTypeResolverInterface) that the controller will give a context e.g. `"stage": "draft"` into the Resolver that the the content type knows if it need to load draft or live pages and not that it depends on `sulu.context` kernel parameter. This will also solve that we would need an extra `PreviewKernel` class.

PR #5184, opened 2020-04-06
Normalize generated container class name

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes -
| Related issues/PRs | https://github.com/sulu/sulu/pull/5180, https://github.com/sulu/skeleton/pull/45, https://github.com/sulu/sulu/pull/5189
| License | MIT

#### What's in this PR?

Fix generated container class name.

#### Why?

In all contexts admin, website and preview the `getContainerClassName` should be the same as it is used by symfony as cache seed https://github.com/symfony/symfony/blob/v4.4.7/src/Symfony/Component/Cache/DependencyInjection/CachePoolPass.php#L56

### DIFF
--- a/src/Sulu/Bundle/PreviewBundle/Preview/Renderer/PreviewKernel.php
+++ b/src/Sulu/Bundle/PreviewBundle/Preview/Renderer/PreviewKernel.php
@@ -44,7 +44,7 @@ class PreviewKernel extends Kernel
     }
 
     /**
-     * The "getContainerClass" need to be normalized for preview and over contexts
+     * The "getContainerClass" need to be normalized for preview and other contexts
      * as it is used by the symfony cache component as prefix.
      *
      * @see SuluKernel::getContainerClass
--- a/src/Sulu/Bundle/PreviewBundle/Preview/Renderer/PreviewKernel.php
+++ b/src/Sulu/Bundle/PreviewBundle/Preview/Renderer/PreviewKernel.php
@@ -43,6 +43,18 @@ class PreviewKernel extends Kernel
         $loader->load(implode(DIRECTORY_SEPARATOR, [__DIR__, '..', '..', 'Resources', 'config', 'config_preview.yml']));
     }
 
+    /**
+     * The "getContainerClass" need to be normalized for preview and over contexts
+     * as it is used by the symfony cache component as prefix.
+     *
+     * @see SuluKernel::getContainerClass
+     */
+    protected function getContainerClass()
+    {
+        // use parent class to normalize the generated container class.
+        return $this->generateContainerClass(get_parent_class());
+    }
+
     public function getRootDir()
     {
         if (null === $this->rootDir) {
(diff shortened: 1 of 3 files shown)


src/Sulu/Component/HttpKernel/SuluKernel.php
```diff
@@ -85,6 +85,41 @@ protected function configureContainer(ContainerBuilder $container, LoaderInterfa
...
+    {
+        return $this->generateContainerClass(static::class);
+    }
+
+    /**
+     * @internal
+     *
+     * This is only for internal use. To get the container class use `getContainerClass` instead.
+     *
+     * This is a copy of the symfony 5.0 getContainerClass which does not include $this->name.
```
danrot, line 104: And where does this function include `$this->name`?

alexander-schranz: > which does not include $this->name

The function from symfony 4.4 did include it this->name https://github.com/symfony/symfony/blob/0889df1e88559968eb291c938d25b96e252f7bed/src/Symfony/Component/HttpKernel/Kernel.php#L478

Not sure how to make this clear 🙈 

The function from symfony 5.0 does not longer include that name and thats why its copied so we have the behaviour from symfony 5 also for 4.4.

danrot:
How can I test that? Should the nodes cache work with that change?

src/Sulu/Component/HttpKernel/SuluKernel.php
```diff
@@ -85,6 +85,41 @@ protected function configureContainer(ContainerBuilder $container, LoaderInterfa
...
+     */
+    protected function getContainerClass()
+    {
+        return $this->generateContainerClass(static::class);
+    }
+
+    /**
+     * @internal
+     *
+     * This is only for internal use. To get the container class use `getContainerClass` instead.
```
danrot, line 102: `This is only used to support Symfony 4.4 and 5 at the same time.` - How about something like that?

src/Sulu/Component/HttpKernel/SuluKernel.php
danrot: Ok, I've misread :see_no_evil:

alexander-schranz:
@danrot yes the node cache should then work and invalidate correcty.

See https://github.com/sulu/skeleton/pull/45/files for configuration.

src/Sulu/Bundle/PreviewBundle/Preview/Renderer/PreviewKernel.php
```diff
@@ -43,6 +43,18 @@ public function registerContainerConfiguration(LoaderInterface $loader)
         $loader->load(implode(DIRECTORY_SEPARATOR, [__DIR__, '..', '..', 'Resources', 'config', 'config_preview.yml']));
     }
 
+    /**
+     * The "getContainerClass" need to be normalized for preview and over contexts
```
danrot, line 47: Should `over` be `other`? Reads strange :thinking: Same in the other comments.

danrot:
Shouldn't that PR also include setting the node cache, so that I have something to test? Or have you done that in a separate PR?

danrot:
But shouldn't that be included in this PR as well? We've also got a production config in sulu/sulu, don't we?

PR #5191, opened 2020-04-07
Remove deprecated twig spaceless tag and fix sitemap index for multiple providers

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

Currently on some places we use deprecated `spaceless` and on others we used the newer `apply spaceless` (e.g.: https://github.com/sulu/sulu/blob/2.0.0/src/Sulu/Bundle/WebsiteBundle/Resources/views/Sitemap/sitemap.xml.twig#L1). I adjusted all to use the newer version and set the min version to twig to `2.9`.

Apply is availabe since [1.40](https://twig.symfony.com/doc/1.x/tags/apply.html) / [2.9](https://twig.symfony.com/doc/2.x/tags/apply.html)

When adding tests for sitemap index I found also out the if there are multiple providers and only one does return data for the current host still a sitemap index was rendered. This is also fixed in this pull request else the others tests would fail, because I needed to add an additional sitemap provider to test the sitemap index file.

#### Why?

Avoid deprecation messages in our twig files and an error on master branch where we support twig 3.0 where spaceless tag does not longer exist.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -99,7 +99,7 @@
         "symfony/validator": "^4.3",
         "symfony/yaml": "^4.3",
         "toflar/psr6-symfony-http-cache-store": "^1.1 || ^2.0",
-        "twig/twig": "^1.41 || ^2.0"
+        "twig/twig": "^1.41 || ^2.9"
     },
     "require-dev": {
         "google/cloud-storage": "^1.0",
--- a/composer.json
+++ b/composer.json
@@ -99,7 +99,7 @@
         "symfony/validator": "^4.3",
         "symfony/yaml": "^4.3",
         "toflar/psr6-symfony-http-cache-store": "^1.1 || ^2.0",
-        "twig/twig": "^1.41 || ^2.9"
+        "twig/twig": "^1.41 || ^2.10"
     },
     "require-dev": {
         "google/cloud-storage": "^1.0",
--- a/src/Sulu/Bundle/PageBundle/Resources/views/Export/Snippet/1.2.xliff.twig
+++ b/src/Sulu/Bundle/PageBundle/Resources/views/Export/Snippet/1.2.xliff.twig
@@ -1,4 +1,4 @@
-{% extends "SuluPageBundle:Export:Snippet/base.export.twig" %}
+{% extends "@SuluPage/Export/Snippet/base.export.twig" %}
 
 {% block main %}
 <?xml version="1.0" encoding="utf-8"?>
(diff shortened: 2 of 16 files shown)


composer.json
```diff
@@ -98,7 +98,7 @@
         "symfony/validator": "^4.3",
         "symfony/yaml": "^4.3",
         "toflar/psr6-symfony-http-cache-store": "^1.1 || ^2.0",
-        "twig/twig": "^1.41 || ^2.0"
+        "twig/twig": "^1.41 || ^2.10"
```
alexander-schranz, line 101: this is needed for `filter` (1.41 / 2.10) and apply (1.40 / 2.09)

It doesnt break do anything as its also min requirement of symfony twig bridge 4.3.0:

https://github.com/symfony/twig-bridge/blob/5ed4b37b36e37baeb36028e2e27725c571853444/composer.json#L20

src/Sulu/Bundle/WebsiteBundle/Tests/Application/config/webspaces/test_with_index.xml
```diff
@@ -0,0 +1,68 @@
...
+                <environment type="test">
+                    <urls>
+                        <url language="en">sulu.index</url>
+                    </urls>
+                </environment>
+            </environments>
+        </portal>
+    </portals>
+</webspace>
+
```
danrot, line 68: Too many new lines

src/Sulu/Bundle/WebsiteBundle/Controller/SitemapController.php
```diff
@@ -86,9 +87,24 @@ public function indexAction(Request $request)
...
             if (!$sitemap) {
-                $aliases = array_keys($this->sitemapProviderPool->getProviders());
+                $sitemapAlias = null;
 
-                return $this->sitemapPaginatedAction($request, reset($aliases), 1);
+                foreach ($this->sitemapProviderPool->getProviders() as $sitemapAlias => $provider) {
+                    if ($provider->getMaxPage($request->getScheme(), $request->getHost()) > 0) {
+                        $sitemapAlias = $provider->getAlias();
+
+                        break;
```
danrot, line 96: This code is really confusing :thinking: I think that's because of the `SitemapProvider`... Are you checking in above's `if` if there is data at all? Does `getMaxPage` only return `0` if there is no data? 

Do we assume that if `renderIndex` above returns a falsy value, that no index has to be created and we return simply the first sitemap we find? Because if there would be two sitemaps we wouldn't get to this code anyway? :thinking:

alexander-schranz: > Does getMaxPage only return 0 if there is no data?

Yes if a sitemap have no data for the current Domain it will return getMaxPage 0

> Do we assume that if renderIndex above returns a falsy value

yes this logic is strange `renderIndex` will return null if there is no need for an index file (only one provider does return has getMaxPage 0). Currently it did take with `reset()` the first provider but thats not always correct instead it should take the first provider which has data to provider thats why the loop was introduced.

src/Sulu/Bundle/PageBundle/Resources/views/Export/Snippet/1.2.xliff.twig
```diff
@@ -20,7 +20,7 @@
 
     {# view #}
     {%- block view -%}
-        {% spaceless %}
+        {% apply spaceless %}
```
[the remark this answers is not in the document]
alexander-schranz: Oh yeah looks like the export doesn't work if templating is not installed 😲

src/Sulu/Bundle/PageBundle/Resources/views/Export/Snippet/1.2.xliff.twig
```diff
@@ -8,36 +8,33 @@
 </xliff>
 {% endblock main %}
 
-{% if false %} {# avoid output of block directly #}
```
alexander-schranz, line 11: This did also throw an error but was also unnecessary as it would never be output directly as it extends from another template.

src/Sulu/Bundle/PageBundle/Resources/views/Export/Webspace/base.export.twig
```diff
@@ -6,12 +6,12 @@
     {# Iterate over the pages #}
     {%- block pages -%}
         {%- for document in documents -%}
-            {% spaceless %}
```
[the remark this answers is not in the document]
alexander-schranz: Fixed this by using the `getLoader` which we also used in the `WebsiteController`

PR #5196, opened 2020-04-08
Fix preview image in Media list

nnatter:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This PR fixes the Media list to display the preview image of a media, if one is set. 

This can be tested by the following steps:
1. Upload a `.xls` file (eg [this example file](https://file-examples.com/wp-content/uploads/2017/02/file_example_XLS_10.xls))
2. Edit the `.xls` file and set a preview image
3. Navigate back to the media list and see the preview image

#### Why?

Because we want to display the preview image, if the user sets one.
Also, this finally fixes the `MediaControllerTest::testCgetWithPreview` test that annoyed me for quite some time 🙂

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/Controller/MediaController.php
+++ b/src/Sulu/Bundle/MediaBundle/Controller/MediaController.php
@@ -204,28 +204,14 @@ class MediaController extends AbstractMediaController implements
         $listBuilder->setParameter('locale', $locale);
         $listResponse = $listBuilder->execute();
 
-        $previewImageIds = [];
-        foreach ($listResponse as $listItem) {
-            $previewImageId = $listItem['previewImageId'] ?? null;
-            if ($previewImageId) {
-                $previewImageIds[] = $previewImageId;
-            }
-        }
-        $previewImageFormats = $this->mediaManager->getFormatUrls($previewImageIds, $locale);
-
         for ($i = 0, $length = count($listResponse); $i < $length; ++$i) {
-            $previewImageId = $listResponse[$i]['previewImageId'] ?? null;
-            if ($previewImageId) {
-                $format = $previewImageFormats[$previewImageId];
-            } else {
-                $format = $this->formatManager->getFormats(
-                    $listResponse[$i]['id'],
-                    $listResponse[$i]['name'],
-                    $listResponse[$i]['version'],
-                    $listResponse[$i]['subVersion'],
-                    $listResponse[$i]['mimeType']
-                );
-            }
+            $format = $this->formatManager->getFormats(
+                $listResponse[$i]['previewImageId'] ?? $listResponse[$i]['id'],
+                $listResponse[$i]['previewImageName'] ?? $listResponse[$i]['name'],
+                $listResponse[$i]['previewImageVersion'] ?? $listResponse[$i]['version'],
+                $listResponse[$i]['previewImageSubVersion'] ?? $listResponse[$i]['subVersion'],
+                $listResponse[$i]['previewImageMimeType'] ?? $listResponse[$i]['mimeType']
+            );
 
             if (0 < count($format)) {
                 $listResponse[$i]['thumbnails'] = $format;
@@ -335,6 +321,10 @@ class MediaController extends AbstractMediaController implements
 
         // field which will be needed afterwards to generate route
         $listBuilder->addSelectField($fieldDescriptors['previewImageId']);
+        $listBuilder->addSelectField($fieldDescriptors['previewImageName']);
+        $listBuilder->addSelectField($fieldDescriptors['previewImageVersion']);
+        $listBuilder->addSelectField($fieldDescriptors['previewImageSubVersion']);
+        $listBuilder->addSelectField($fieldDescriptors['previewImageMimeType']);
         $listBuilder->addSelectField($fieldDescriptors['version']);
         $listBuilder->addSelectField($fieldDescriptors['subVersion']);
         $listBuilder->addSelectField($fieldDescriptors['name']);
--- a/src/Sulu/Bundle/MediaBundle/Controller/MediaController.php
+++ b/src/Sulu/Bundle/MediaBundle/Controller/MediaController.php
@@ -206,14 +206,16 @@ class MediaController extends AbstractMediaController implements
 
         $previewImageIds = [];
         foreach ($listResponse as $listItem) {
-            if ($listItem['previewImageId']) {
-                $previewImageIds[] = $listItem['previewImageId'];
+            $previewImageId = $listItem['previewImageId'] ?? null;
+            if ($previewImageId) {
+                $previewImageIds[] = $previewImageId;
             }
         }
         $previewImageFormats = $this->mediaManager->getFormatUrls($previewImageIds, $locale);
 
         for ($i = 0, $length = count($listResponse); $i < $length; ++$i) {
-            if ($previewImageId = $listResponse[$i]['previewImageId']) {
+            $previewImageId = $listResponse[$i]['previewImageId'] ?? null;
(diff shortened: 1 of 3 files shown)


src/Sulu/Bundle/MediaBundle/Controller/MediaController.php
```diff
@@ -203,16 +203,27 @@ public function cgetAction(Request $request)
         $listBuilder = $this->getListBuilder($request, $fieldDescriptors, $types);
         $listBuilder->setParameter('locale', $locale);
         $listResponse = $listBuilder->execute();
-        $count = $listBuilder->count();
+
+        $previewImageIds = [];
+        foreach ($listResponse as $listItem) {
+            if ($listItem['previewImageId']) {
```
alexander-schranz, line 209: We should check if `$listItem['previewImageId']` isset or fallback with `??` to `null`

src/Sulu/Bundle/MediaBundle/Controller/MediaController.php
```diff
@@ -203,16 +203,27 @@ public function cgetAction(Request $request)
...
 
         for ($i = 0, $length = count($listResponse); $i < $length; ++$i) {
-            $format = $this->formatManager->getFormats(
-                $listResponse[$i]['previewImageId'] ?? $listResponse[$i]['id'],
-                $listResponse[$i]['name'],
-                $listResponse[$i]['version'],
-                $listResponse[$i]['subVersion'],
-                $listResponse[$i]['mimeType']
-            );
+            if ($previewImageId = $listResponse[$i]['previewImageId']) {
```
alexander-schranz, line 216: same here before it was fallback to null and nothing did happens if previewImageId was not set

src/Sulu/Bundle/MediaBundle/Controller/MediaController.php
```diff
@@ -203,16 +203,29 @@ public function cgetAction(Request $request)
...
-        $count = $listBuilder->count();
+
+        $previewImageIds = [];
+        foreach ($listResponse as $listItem) {
+            $previewImageId = $listItem['previewImageId'] ?? null;
+            if ($previewImageId) {
+                $previewImageIds[] = $previewImageId;
+            }
+        }
+        $previewImageFormats = $this->mediaManager->getFormatUrls($previewImageIds, $locale);
```
danrot, line 214: This loads the entire preview image right? The thing is that the preview image is already displayed for me, isn't it for you? If it is already working I don't see why we should make all these db requests.

nnatter: yes, the `MediaManager` probably load the whole media for this. the alternative would be to load the `name`, `version`, `subversion` and `mimeType` of the preview image in the list response. but that would complicate the media list xml a lot.

alexander-schranz: @danrot it display the image because the imageAction only checks for the `id` but the image is then always recalculated on the server as the filename and folder does not match where it is stored.

Instead using the mediaManager we could join the columns of the previewImage we need:

```php
                    $listResponse[$i]['previewImageId'],
                    $listResponse[$i]['previewImageName'],
                    $listResponse[$i]['previewImageVersion'],
                    $listResponse[$i]['previewImageSubversion'],
                    $listResponse[$i]['previewImageMimeType']
```

danrot: Would IMO also make the code easier to read. Can you maybe also make a very short benchmark if it is really improving the performance compared to what is currently done in this PR?

nnatter: regarding to that the image is already working for you. looks like this depends on the system for `pdf` files. i adapted the pr description to use an `xls` file. does that work for you too?

danrot: Nope, does not seem to work for `xls` files

alexander-schranz: That make sense as the false mimeType was given (original media instead of preview media) and pdf is supported and xls is not to generate images.

nnatter: i adjusted the pr to use the `ListBuilder` for loading the required properties of the preview image. this makes the media list config more complicated, but we dont need any additional request.

did not do any benchmark because i am not sure how to do that in a useful manner on my machine. but if anybody wants to do that, he can reset the last commit to compare the two strategies 🙂

src/Sulu/Bundle/MediaBundle/Resources/config/lists/media.xml
```diff
@@ -8,6 +8,23 @@
...
+            <field-name>%sulu.model.media.class%.previewImage</field-name>
+        </join>
+
+        <join>
+            <entity-name>PreviewImage_SuluMediaBundle:File</entity-name>
+            <field-name>PreviewImage_%sulu.model.media.class%.files</field-name>
+        </join>
+
+        <join>
+            <entity-name>PreviewImage_SuluMediaBundle:FileVersion</entity-name>
```
alexander-schranz, line 23: would name this just `previewImageFileVersion`

src/Sulu/Bundle/MediaBundle/Resources/config/lists/media.xml
```diff
@@ -8,6 +8,23 @@
...
         </join>
     </joins>
+    <joins name="previewImageFileVersion">
+        <join>
+            <entity-name>PreviewImage_%sulu.model.media.class%</entity-name>
+            <field-name>%sulu.model.media.class%.previewImage</field-name>
+        </join>
+
+        <join>
+            <entity-name>PreviewImage_SuluMediaBundle:File</entity-name>
```
alexander-schranz, line 18: would name this just `previewImageFile`

src/Sulu/Bundle/MediaBundle/Resources/config/lists/media.xml
```diff
@@ -8,6 +8,23 @@
             <field-name>%sulu.model.media.class%.type</field-name>
         </join>
     </joins>
+    <joins name="previewImageFileVersion">
+        <join>
+            <entity-name>PreviewImage_%sulu.model.media.class%</entity-name>
```
alexander-schranz, line 13: would name this just `previewImage`

nnatter: not really euphoric about any of the two, so i dont care about that 😉 will rename it if @danrot is okay with `previewImage ` too

danrot: Hm, until now we always used names of classes here, but we never had to link the same entitiy itself as far as i can remember... I woul dhav ea tendency to keep the class name in there and probably would have suffixed the `PreviewImage` part. This way it would still be clear what exactly is loaded when looking at the queries. Might be obvious in this case anyway, but if we want to use the same conventions in other places it might not.

Also, until now all entity names have the class included in some way.

danrot: Oh, and I was wrong, found a similar use case: https://github.com/sulu/sulu/blob/0a08318/src/Sulu/Bundle/CategoryBundle/Resources/config/lists/categories.xml#L21

nnatter: adjusted the `entity-name` to use this format: `%sulu.model.media.class%PreviewImage`

PR #5197, opened 2020-04-08
Fix minimum size calculation for crops

### LINKED ISSUE #5155: Image Cropping: Minimal Size Area not draggable
| Q | A
| --- | ---
| Sulu Version | 2.0.6
| Browser Version | Chrome, Ubuntu, Version 80.0.3987.149 (Official Build) (64-bit)

# Actual Behavior

If you want to crop an image that is smaller then the minimum crop size you can only double click to maximize the cropping area in the center. Then maybe you can drag it to the top of the image but that is more like a flicker. You can't drag it where you want it.

# Expected Behavior

I say it's okay that you can't make the area smaller because it is already too small. Still it would be nice to drag the area as you like and avoid flickering. I guess the JS is using some too large numbers and messes up the calculation.

# Steps to Reproduce

Take a small image and set an image style like:

```
    <format key="hero-swiper-xl">
        <meta>
            <title lang="en">Hero Swiper XL</title>
            <title lang="de">Hero Swiper XL</title>
        </meta>

        <scale x="2560" y="853" mode="outbound"/>
    </format>
```

danrot:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #5155
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR changes the way the min size of the crop is calculated. Previously it was based on the ratio between the real and the displayed size, which is only correct for the values that are passed. If the reqeusted format is bigger than the full image the calculation was wrong, because it was downsizing too less.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/ImageRectangleSelection/ImageRectangleSelection.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/ImageRectangleSelection/ImageRectangleSelection.js
@@ -3,7 +3,7 @@ import {action, computed, observable} from 'mobx';
 import log from 'loglevel';
 import {observer} from 'mobx-react';
 import React from 'react';
-import RectangleSelection, {RoundingNormalizer} from '../RectangleSelection';
+import RectangleSelection from '../RectangleSelection';
 import type {SelectionData} from '../RectangleSelection';
 import withContainerSize from '../withContainerSize';
 import imageRectangleSelectionStyles from './imageRectangleSelection.scss';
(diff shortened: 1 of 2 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/components/ImageRectangleSelection/ImageRectangleSelection.js
```diff
@@ -80,22 +79,47 @@ class ImageRectangleSelection extends React.Component<Props> {
 
     handleRectangleSelectionChange = (data: ?SelectionData) => {
         const {onChange} = this.props;
-        onChange(data ? this.rounding.normalize(this.scaledDataToNatural(data)) : undefined);
```
danrot, line 83: @alexander-schranz This will fix a bug that makes this component fail to recognize the minimum size of pictures, but it also means that the result is not rounded anymore, e.g. decimal values are sent to the server. Do you think that this can cause problems?

danrot:
@alexander-schranz Can you also review the rest of the code? I think there is nobody other than me who knows that code in particular, and since it is about image rendering you are probably the best person to do so :-)

PR #5198, opened 2020-04-08
Update flow version to version 0.122

nnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This PR updates the `flow-bin` dependency to the version `0.122`. Furthermore, it adjusts the codebase to be compatible with the new version.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/AutoCompletePopover/tests/AutoCompletePopover.test.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/AutoCompletePopover/tests/AutoCompletePopover.test.js
@@ -159,7 +159,8 @@ test('Pressing up should select previous item', () => {
     const suggestionElement2 = {focus: jest.fn()};
 
     Object.defineProperty(document, 'activeElement', {
-        value: (suggestionElement2: any),
+        // $FlowFixMe
+        value: suggestionElement2,
     });
 
     autoCompletePopover.instance().suggestionsRef = {
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/AutoCompletePopover/tests/AutoCompletePopover.test.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/AutoCompletePopover/tests/AutoCompletePopover.test.js
@@ -158,9 +158,8 @@ test('Pressing up should select previous item', () => {
     const suggestionElement1 = {focus: jest.fn()};
     const suggestionElement2 = {focus: jest.fn()};
 
-    // $FlowFixMe
     Object.defineProperty(document, 'activeElement', {
-        value: suggestionElement2,
+        value: (suggestionElement2: any),
     });
 
     autoCompletePopover.instance().suggestionsRef = {
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/utils/TestHelper/defaultWebspace.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/utils/TestHelper/defaultWebspace.js
@@ -1,0 +1,19 @@
+// @flow
+
+const defaultWebspace = {
+    _permissions: {},
+    allLocalizations: [],
+    customUrls: [],
+    defaultTemplates: {},
+    key: 'webspace-key',
+    localizations: [],
+    name: 'webspace-name',
+    navigations: [],
+    portalInformation: [],
+    resourceLocatorStrategy: {
+        inputType: 'resource-locator-strategy-input-type',
+    },
+    urls: [],
+};
+
+export default defaultWebspace;
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/utils/TestHelper/defaultWebspace.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/utils/TestHelper/defaultWebspace.js
@@ -5,9 +5,7 @@ const defaultWebspace = {
     allLocalizations: [],
     customUrls: [],
     defaultTemplates: {},
-    key: 'webspace-key',
     localizations: [],
-    name: 'webspace-name',
     navigations: [],
     portalInformation: [],
     resourceLocatorStrategy: {
--- a/src/Sulu/Bundle/ContactBundle/Resources/js/components/ContactDetails/ContactDetails.js
+++ b/src/Sulu/Bundle/ContactBundle/Resources/js/components/ContactDetails/ContactDetails.js
@@ -160,9 +160,8 @@ class ContactDetails extends React.Component<Props> {
         const {value: {emails}} = this.props;
 
         if (emails.length === 0) {
-            // $FlowFixMe
             return [
-                this.getEmptyEntry('emails'),
+                (this.getEmptyEntry('emails'): any),
             ];
         }
 
@@ -173,9 +172,8 @@ class ContactDetails extends React.Component<Props> {
         const {value: {phones}} = this.props;
 
         if (phones.length === 0) {
-            // $FlowFixMe
             return [
-                this.getEmptyEntry('phones'),
+                (this.getEmptyEntry('phones'): any),
             ];
         }
 
--- a/src/Sulu/Bundle/ContactBundle/Resources/js/components/ContactDetails/ContactDetails.js
+++ b/src/Sulu/Bundle/ContactBundle/Resources/js/components/ContactDetails/ContactDetails.js
@@ -161,7 +161,7 @@ class ContactDetails extends React.Component<Props> {
 
         if (emails.length === 0) {
             return [
-                (this.getEmptyEntry('emails'): any),
+                ((this.getEmptyEntry('emails'): any): EmailType),
             ];
         }
 
@@ -173,7 +173,7 @@ class ContactDetails extends React.Component<Props> {
 
         if (phones.length === 0) {
             return [
-                (this.getEmptyEntry('phones'): any),
+                ((this.getEmptyEntry('phones'): any): PhoneType),
             ];
         }
 
--- a/src/Sulu/Bundle/ContactBundle/Resources/js/components/ContactDetails/ContactDetails.js
+++ b/src/Sulu/Bundle/ContactBundle/Resources/js/components/ContactDetails/ContactDetails.js
@@ -160,9 +160,8 @@ class ContactDetails extends React.Component<Props> {
         const {value: {emails}} = this.props;
 
         if (emails.length === 0) {
-            return [
-                ((this.getEmptyEntry('emails'): any): EmailType),
-            ];
+            // $FlowFixMe
+            return [this.getEmptyEntry('emails')];
         }
 
         return emails;
@@ -172,9 +171,8 @@ class ContactDetails extends React.Component<Props> {
         const {value: {phones}} = this.props;
 
         if (phones.length === 0) {
-            return [
-                ((this.getEmptyEntry('phones'): any): PhoneType),
-            ];
+            // $FlowFixMe
+            return [this.getEmptyEntry('phones')];
         }
 
         return phones;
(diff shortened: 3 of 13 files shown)


src/Sulu/Bundle/PageBundle/Resources/js/views/PageList/tests/PageList.test.js
```diff
@@ -90,8 +90,7 @@ beforeEach(() => {
 
 test('Render PageList', () => {
     const webspaceKey = observable.box('sulu');
-    // $FlowFixMe
-    const webspace = {};
+    const webspace = ({}: any);
```
nnatter, line 93: we cannot easily set a useful value here because the `redirectToWebspaceLocale()` method of the component will always trigger if the `webspace.localizations` are set

nnatter: also, it looks to me like the redirect logic of the component is not tested at the moment 🙈

src/Sulu/Bundle/AdminBundle/Resources/js/utils/TestHelper/defaultWebspace.js
```diff
@@ -0,0 +1,19 @@
+// @flow
+
+const defaultWebspace = {
+    _permissions: {},
+    allLocalizations: [],
+    customUrls: [],
+    defaultTemplates: {},
+    key: 'webspace-key',
```
danrot, line 8: In the other helpers delivering some default values I avoided to set values like this. I tried to only use `{}`, `[]` and `undefined`, whereby undefined is always the preferred choice. This way we don't endanger ourselves of the possibility that we are coupling the tests too hard on this file and break everything when somebody wants to change this `key`. It also makes it harder to understand the tests, because you don't easily see where this value is coming from.

I would say the same is true for the `name` property. I would just leave them out and set them in every test as required.

src/Sulu/Bundle/PageBundle/Resources/js/views/PageList/tests/PageList.test.js
```diff
@@ -90,8 +90,6 @@ beforeEach(() => {
 
 test('Render PageList', () => {
     const webspaceKey = observable.box('sulu');
-    // $FlowFixMe
```
danrot, line 93: Resp. does this not work anymore? Although I find the solution with the `defaultWebspace` cleaner anyway, I am just curious how that changed.

src/Sulu/Bundle/SecurityBundle/Resources/js/containers/Permissions/tests/Permissions.test.js
```diff
@@ -226,17 +227,19 @@ test('Render with empty webspace section', () => {
     const promise = Promise.resolve(securityContextGroups);
     securityContextStore.loadSecurityContextGroups.mockReturnValue(promise);
 
-    // $FlowFixMe
     webspaceStore.allWebspaces = [
         {
+            ...defaultWebspace,
```
danrot, line 232: That's how I would do it everywhere.

src/Sulu/Bundle/MediaBundle/Resources/js/containers/Form/fields/MediaVersionUpload.js
```diff
@@ -3,6 +3,7 @@ import React from 'react';
 import {ResourceStore} from 'sulu-admin-bundle/stores';
 import type {FieldTypeProps} from 'sulu-admin-bundle/types';
 import MediaVersionUploadComponent from '../../MediaVersionUpload';
+import ResourceFormStore from "sulu-admin-bundle/containers/Form/stores/ResourceFormStore";
```
danrot: import from `sulu-admin-bundle/containers`

src/Sulu/Bundle/ContactBundle/Resources/js/components/ContactDetails/ContactDetails.js
```diff
@@ -160,9 +160,8 @@ class ContactDetails extends React.Component<Props> {
         const {value: {emails}} = this.props;
 
         if (emails.length === 0) {
-            // $FlowFixMe
             return [
-                this.getEmptyEntry('emails'),
+                (this.getEmptyEntry('emails'): any),
```
danrot, line 164: I would not just cast it to `any`, since we are losing potential subsequent type checks. If we do something like this, I would cast to the type we want afterwards as well:

```javascript
((this...): any): EmailType)
```

src/Sulu/Bundle/ContactBundle/Resources/js/components/ContactDetails/ContactDetails.js
```diff
@@ -173,9 +172,8 @@ class ContactDetails extends React.Component<Props> {
         const {value: {phones}} = this.props;
 
         if (phones.length === 0) {
-            // $FlowFixMe
             return [
-                this.getEmptyEntry('phones'),
+                (this.getEmptyEntry('phones'): any),
```
danrot, line 176: I would not just cast it to `any`, since we are losing potential subsequent type checks. If we do something like this, I would cast to the type we want afterwards as well:

```javascript
((this...): any): PhoneType)
```

src/Sulu/Bundle/AdminBundle/Resources/js/components/AutoCompletePopover/tests/AutoCompletePopover.test.js
```diff
@@ -158,9 +158,8 @@ test('Pressing up should select previous item', () => {
     const suggestionElement1 = {focus: jest.fn()};
     const suggestionElement2 = {focus: jest.fn()};
 
-    // $FlowFixMe
     Object.defineProperty(document, 'activeElement', {
-        value: suggestionElement2,
+        value: (suggestionElement2: any),
```
danrot, line 162: I would personally make more use of `// $FlowFixMe` instead of casting everything to `any`. With `any` we are losing all typechecks that are happening afterwards, and I think with `$FlowFixMe` we are only fixing this error. Especially since it seems that they also want to support suppressing only certain error codes.

And it still works when moving the comment above this line directly.

src/Sulu/Bundle/MediaBundle/Resources/js/containers/Form/fields/MediaVersionUpload.js
```diff
@@ -2,8 +2,8 @@
 import React from 'react';
 import {ResourceStore} from 'sulu-admin-bundle/stores';
 import type {FieldTypeProps} from 'sulu-admin-bundle/types';
+import ResourceFormStore from 'sulu-admin-bundle/containers/Form/stores/ResourceFormStore';
```
danrot, line 5: Use `sulu-admin-bundle/stores` to import that file

src/Sulu/Bundle/PageBundle/Resources/js/views/PageList/tests/PageList.test.js
danrot: Can't we set some localizations and set the value of the `locale` observable in the component to one of the existing ones?

src/Sulu/Bundle/AdminBundle/Resources/js/components/AutoCompletePopover/tests/AutoCompletePopover.test.js
nnatter: you are right, will move the `$FlowFixMe`

src/Sulu/Bundle/MediaBundle/Resources/js/containers/Form/fields/MediaVersionUpload.js
nnatter: looks like `ResourceFormStore` is not exported in `sulu-admin-bundle/stores`. but i will use `sulu-admin-bundle/containers`

src/Sulu/Bundle/AdminBundle/Resources/js/utils/TestHelper/defaultWebspace.js
nnatter: i set the `key` and the `name` because they are not nullable in the `Webspace` type. but omiting them in the `defaultWebspace` and setting them in the test case is okay for me too 🙂

src/Sulu/Bundle/PageBundle/Resources/js/views/PageList/tests/PageList.test.js
nnatter: the `$FlowFixMe`comment now needs to be at the "primary location": https://medium.com/flow-type/making-flow-error-suppressions-more-specific-280aa4e3c95c

i guess in our case this would be the line above the `<PageList route ...` JSX inside of the `mount()` call. but have not tested this to be honest

src/Sulu/Bundle/PageBundle/Resources/js/views/PageList/tests/PageList.test.js
nnatter: unfortunately i dont think so. the redirect logic is executed in the constructor of the `PageList` component, therefore i cannot change any property of the component before it is executed: https://github.com/sulu/sulu/blob/master/src/Sulu/Bundle/PageBundle/Resources/js/views/PageList/PageList.js#L103

furthermore, it looks like i am not able to set the `locale` via the router, as it is read from the router only after the redirect logic is executed. not sure if this is correct anyway, but i am also hesitant to change such things in this pr 🙈 

how should we proceed here?

danrot: The router would have that already set, and this is considered in the `redirctToWebspace` function. So I would say the behavior is correct, although it might not be easy to test. So you would have to mock the `attributes` of the router I guess... But you are right, that's not part of this PR.

src/Sulu/Bundle/PageBundle/Resources/js/views/PageList/tests/PageList.test.js
```diff
@@ -114,11 +113,10 @@ test('Should show the locales from the webspace configuration for the toolbar',
 
     const webspaceKey = observable.box('sulu');
 
-    // $FlowFixMe
-    const webspace = {
+    const webspace = ({
         key: 'sulu',
         allLocalizations: [{localization: 'en', name: 'en'}, {localization: 'de', name: 'de'}],
-    };
+    }: any);
```
danrot, line 119: Why don't you use the `defaultWebspace` with the spread operator here, to avoid casting this to `any`? Also true for all other occurences in this file.

src/Sulu/Bundle/PageBundle/Resources/js/views/PageList/tests/PageList.test.js
danrot: Maybe that would be the better way to solve it instead of casting everything to any? :thinking:

src/Sulu/Bundle/PageBundle/Resources/js/views/PageList/tests/PageList.test.js
nnatter: because if the `webspace.localizations` property of the passed webspace is not undefined, the redirect logic of the `PageList` component will trigger 🙈

danrot: Then I would probably set do something like this:

```javascript
const webspace = {
    ...defaultWebspace,
    localizations: undefined,
};
```

And add `// $FlowFixMe` comments where necessary.

nnatter: did it that way 🙂

src/Sulu/Bundle/PageBundle/Resources/js/views/PageList/tests/PageList.test.js
nnatter: yes, you are right - changed it to not cast to `any`

PR #5199, opened 2020-04-09
Deprecate sulu_security.system configuration option

nnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This PR is another take on the problem I was trying to solve in #5169.
The PR deprecates the `sulu_security.system` and adds a `SULU_ADMIN_SECURITY_SYSTEM ` constant to make the array returned by the `getSecurityContexts()` method of the `Admin` classes more understandable.

#### Why?

Because setting the `sulu_security.system` in the admin context currently breaks permission related things in the admin. We cannot remove the option because somebody might use it in the website context.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Admin/Admin.php
+++ b/src/Sulu/Bundle/AdminBundle/Admin/Admin.php
@@ -21,6 +21,7 @@ use Sulu\Bundle\AdminBundle\Admin\View\ViewProviderInterface;
  */
 abstract class Admin implements ViewProviderInterface, NavigationProviderInterface
 {
+    const SULU_ADMIN_SECURITY_SYSTEM = 'Sulu';
     const SETTINGS_NAVIGATION_ITEM = 'sulu_admin.settings';
 
     public static function getPriority(): int
--- a/src/Sulu/Bundle/SecurityBundle/Build/UserBuilder.php
+++ b/src/Sulu/Bundle/SecurityBundle/Build/UserBuilder.php
@@ -11,6 +11,7 @@
 
 namespace Sulu\Bundle\SecurityBundle\Build;
 
+use Sulu\Bundle\AdminBundle\Admin\Admin;
 use Sulu\Bundle\CoreBundle\Build\SuluBuilder;
 
 /**
@@ -33,7 +34,7 @@ class UserBuilder extends SuluBuilder
         $user = 'admin';
         $password = 'admin';
         $roleName = 'User';
-        $system = 'Sulu';
+        $system = Admin::SULU_ADMIN_SECURITY_SYSTEM;
         $locale = 'en';
         $doctrine = $this->container->get('doctrine')->getManager();
         $userRep = $this->container->get('sulu.repository.user');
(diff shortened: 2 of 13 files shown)


src/Sulu/Bundle/AdminBundle/Admin/Admin.php
```diff
@@ -21,6 +21,7 @@
  */
 abstract class Admin implements ViewProviderInterface, NavigationProviderInterface
 {
+    const SULU_ADMIN_SECURITY_SYSTEM = 'Sulu';
```
nnatter, line 24: hope this is okay for you. would like to use a constant that describes the meaning of the value in the `Admin` classes. also,  the way i understood security systems, we should use a dedicated security system for admin related roles 🙂

danrot: I would prefer not to introduce a new constant, because it is only introduced to avoid a bigger change... However, if you insist on that I am ok with it.

src/Sulu/Bundle/SecurityBundle/Build/UserBuilder.php
```diff
@@ -33,7 +34,7 @@ public function build()
         $user = 'admin';
         $password = 'admin';
         $roleName = 'User';
-        $system = 'Sulu';
+        $system = Admin::SULU_ADMIN_SECURITY_SYSTEM;
```
danrot, line 37: That's where it gets weird IMO, also a reason I actually would prefer not to have this constant.

nnatter: why do you feel this is weird? the builder creates the default user for the admin, right?

danrot: To me it feels more like the security system itself should have control over how this is called, but it only makes it a little bit better... And that I guess this is only a temporary solution until we have a proper fix (which would even include getting rid of the array being returned from the `getSecurityContexts` function) it will continue to feel weird to me... That's why I would not introduce this variable at this point. But as said, if you insist on it I can live with it.

src/Sulu/Bundle/AdminBundle/Admin/Admin.php
nnatter: i think using a constant makes things more understandable. otherwise, it is really hard to understand what the `'Sulu'` string means that is used in the `Admin` classes..

what bigger change do you mean? what disadvantages do you see?

src/Sulu/Bundle/SecurityBundle/Build/UserBuilder.php
nnatter: alright. properly fixing this would be nice indeed - not sure if it is a good idea that a security system is just a string value in the first place.. but until then, i would prefer to use a constant instead of continue using tha string without context 😇

src/Sulu/Bundle/AdminBundle/Admin/Admin.php
danrot: Explained it in the comment below. I am just not sure if this is the right place and if it should be a constant at all, so this contradicts a bit with a mentality of "Duplication is better than the wrong abstraction". But as I already repeatedly said, if you insist on it I can live with it. It's probably not the worst thing we are doing.

nnatter: sorry - i deleted my comment after i saw the answer below 🙈 as @alexander-schranz has approved this, i will dare to merge this and we can hopefully forget about this topic again 😅

PR #5202, opened 2020-04-09
Add all available content types to default template

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR introduces all available content types to the `default` template.

#### Why?

Because currently testing stuff is cumbersome, if you have to e.g. create some blocks for testing. We've used to had such a template in sulu-standard, where it confused people when they were starting a new project, that was why we abandoned this template in sulu-minimal.

However, I think it makes sense to introduce such a template in here, because this template is only used when the webserver is started in this repository, what is not done for projects anyway, but only for developers working on Sulu.


templates/templates/pages/default.html.twig
```diff
@@ -1,9 +1,5 @@
 {% extends "base.html.twig" %}
 
 {% block content %}
-    <h1>{{ content.title }}</h1>
-
-    <div>
-        {{ content.article|raw }}
-    </div>
+    {{ dump() }}
```
danrot, line 4: Should probably be implemented properly.

danrot:
@sulu/core-team WDYT?

nnatter:
would maybe use an `example` template instead of the `default` template. otherwise it might be confusing why everything but this file is in sync with the `sulu/skeleton`. but otherwise i think this is a good idea!

danrot:
I have changed the `default` to the `example` template and made it the default for the webspace.

PR #5204, opened 2020-04-10
Fix clearing of preview cache

### LINKED ISSUE #4369: Cache clear for preview not working anymore
| Q | A
| --- | ---
| Bug? | yes
| New Feature? | no
| Sulu Version | 2.0.0-alpha4
| Browser Version | -

#### Actual Behavior

The Cache-Clear command does not clear the preview cache anymore.
Only the base cache clear command of Symfony is executed.

#### Expected Behavior

The preview cache should be cleared when the admin cache is refreshed

#### Steps to Reproduce

Install sulu-minimal project and execute `php bin/adminconsole cache:clear`

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #4369
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

Fix clearing of preview cache.

#### Why?

Currently the preview cache is not cleared when running `bin/console cache:clear` and so changes in twig files has no effect under `prod`. 

#### Example Usage

1. Change env to `.prod`
2. Open page with preview
3. Change `.html.twig`  preview
4. Run bin/console cache:clear
5. Preview should show changes

### DIFF
--- a/src/Sulu/Bundle/PreviewBundle/EventSubscriber/CacheCommandSubscriber.php
+++ b/src/Sulu/Bundle/PreviewBundle/EventSubscriber/CacheCommandSubscriber.php
@@ -1,0 +1,74 @@
+<?php
+
+/*
+ * This file is part of Sulu.
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\PreviewBundle\EventSubscriber;
+
+use Sulu\Bundle\PreviewBundle\Preview\Renderer\KernelFactoryInterface;
+use Symfony\Bundle\FrameworkBundle\Command\CacheClearCommand;
+use Symfony\Bundle\FrameworkBundle\Command\CacheWarmupCommand;
+use Symfony\Bundle\FrameworkBundle\Console\Application;
+use Symfony\Component\Console\ConsoleEvents;
+use Symfony\Component\Console\Event\ConsoleCommandEvent;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+class CacheCommandSubscriber implements EventSubscriberInterface
+{
+    /**
+     * @var KernelFactoryInterface
+     */
+    private $kernelFactory;
+
+    /**
+     * @var string
+     */
+    private $environment;
+
+    /**
+     * Needed for testing.
+     *
+     * @see Sulu\Bundle\PreviewBundle\Tests\Unit\EventSubscriber\CacheCommandSubscriberTest
+     *
+     * @var Application|null
+     */
+    private $application;
+
+    public function __construct(
+        KernelFactoryInterface $kernelFactory,
+        string $environment
+    ) {
+        $this->kernelFactory = $kernelFactory;
+        $this->environment = $environment;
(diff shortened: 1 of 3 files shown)


src/Sulu/Bundle/PreviewBundle/EventSubscriber/CacheCommandSubscriber.php
```diff
@@ -0,0 +1,65 @@
...
+    public function onCommand(ConsoleCommandEvent $event)
+    {
+        if (!in_array($event->getCommand()->getName(), [
+            CacheClearCommand::getDefaultName(),
+            CacheWarmupCommand::getDefaultName(),
+        ])) {
+            return;
+        }
+
+        $kernel = $this->kernelFactory->create($this->environment);
```
nnatter, line 60: would be easier to understand what this does if we call this `$previewKernel` 😇

alexander-schranz: sounds good. Renamed!

src/Sulu/Bundle/PreviewBundle/EventSubscriber/CacheCommandSubscriber.php
```diff
@@ -0,0 +1,74 @@
...
+        if (!in_array($event->getCommand()->getName(), [
+            CacheClearCommand::getDefaultName(),
+            CacheWarmupCommand::getDefaultName(),
+        ])) {
+            return;
+        }
+
+        $previewKernel = $this->kernelFactory->create($this->environment);
+
+        $application = $this->application ?: new Application($previewKernel);
```
danrot, line 71: Any reason not to do this in the constructor? Would avoid creating multiple `Application` instances in certain situations. I can imagine that you did that because you did not want to allow to set this variable in other cases, but I think I would allow to do so. Would also mean that we don't have to use reflections in the test. And maybe there are also other use cases for this we don't think of right now.

alexander-schranz: Basically the Application needs the PreviewKernel and so every command would initialize the Preview Container, which I would avoid doing in the constructor of a subscriber.

I still can add the application to the constructor to avoid the reflection but this part would stay the same.

danrot: Okay, that's true, didn't think of that. But I think I would still move it to the constructor. If you are just looking at that code it looks a bit confusing, because it feels like you could simply delete the `application` class variable.

Or maybe it would even be better to use setter injection here? The setter could be marked as internal, and it would be no BC break when we add more arguments to the constructor before that argument (which probably would have to be always the last, because it is optional).

danrot:
And that would also fix #4369, right?

src/Sulu/Bundle/PreviewBundle/EventSubscriber/CacheCommandSubscriber.php
alexander-schranz: Setter injection sounds good for me 👍

PR #5208, opened 2020-04-12
Handle non ajax logout request

lightglitch:
| Q | A
| --- | ---
| Bug fix? | yes
| BC breaks? | no

#### What's in this PR?
Handle logout for normal http requests

#### Why?
Sometime we may need to call the logout action from pages that don't use ajax to do it.

### DIFF
--- a/src/Sulu/Bundle/SecurityBundle/Resources/config/services.xml
+++ b/src/Sulu/Bundle/SecurityBundle/Resources/config/services.xml
@@ -94,7 +94,9 @@
             <argument type="service" id="session"/>
         </service>
 
-        <service id="sulu_security.logout_success_handler" class="Sulu\Bundle\SecurityBundle\Security\LogoutSuccessHandler" />
+        <service id="sulu_security.logout_success_handler" class="Sulu\Bundle\SecurityBundle\Security\LogoutSuccessHandler">
+            <argument type="service" id="router"/>
+        </service>
 
         <service id="sulu_security.mask_converter" class="%sulu_security.mask_converter.class%" public="true">
             <argument>%permissions%</argument>
--- a/src/Sulu/Bundle/SecurityBundle/Security/LogoutSuccessHandler.php
+++ b/src/Sulu/Bundle/SecurityBundle/Security/LogoutSuccessHandler.php
@@ -12,14 +12,32 @@
 namespace Sulu\Bundle\SecurityBundle\Security;
 
 use Symfony\Component\HttpFoundation\JsonResponse;
+use Symfony\Component\HttpFoundation\RedirectResponse;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Routing\RouterInterface;
 use Symfony\Component\Security\Http\Logout\LogoutSuccessHandlerInterface;
 
 class LogoutSuccessHandler implements LogoutSuccessHandlerInterface
 {
+    /**
+     * @var RouterInterface
+     */
+    private $router;
+
+    public function __construct(RouterInterface $router)
+    {
+        $this->router = $router;
+    }
+
     public function onLogoutSuccess(Request $request)
     {
-        return new JsonResponse(null, Response::HTTP_OK);
+        if ($request->isXmlHttpRequest()) {
+            $response = new JsonResponse(null, Response::HTTP_OK);
+        } else {
+            $response = new RedirectResponse($this->router->generate('sulu_admin'));
+        }
+
+        return $response;
     }
 }


src/Sulu/Bundle/SecurityBundle/Resources/config/services.xml
```diff
@@ -94,15 +94,16 @@
...
+            <argument type="service" id="router"/>
+        </service>
 
         <service id="sulu_security.mask_converter" class="%sulu_security.mask_converter.class%" public="true">
             <argument>%permissions%</argument>
         </service>
 
         <service id="sulu_security.salt_generator" class="%sulu_security.salt_generator.class%" public="true">
         </service>
-        <service id="%sulu_security.salt_generator.class%" alias="sulu_security.salt_generator"/>
```
danrot, line 105: Was that an accident? I am quite sure we can't simply remove that line.

src/Sulu/Bundle/SecurityBundle/Security/LogoutSuccessHandler.php
```diff
@@ -12,14 +12,29 @@
...
 use Symfony\Component\HttpFoundation\JsonResponse;
+use Symfony\Component\HttpFoundation\RedirectResponse;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Routing\RouterInterface;
 use Symfony\Component\Security\Http\Logout\LogoutSuccessHandlerInterface;
 
 class LogoutSuccessHandler implements LogoutSuccessHandlerInterface
 {
+    private $router;
```
danrot, line 23: Type that variable using a `@var` annotation:

```php
/**
 * @var RouterInterface
 */
private $router;
```

lightglitch:
@danrot  Both those changes were made when I run cs fixer on those files. Bu I will revert them.

danrot:
And can you also answer what's the use case for this?

src/Sulu/Bundle/SecurityBundle/Resources/config/services.xml
```diff
@@ -94,7 +94,9 @@
             <argument type="service" id="session"/>
         </service>
 
-        <service id="sulu_security.logout_success_handler" class="Sulu\Bundle\SecurityBundle\Security\LogoutSuccessHandler" />
+        <service id="sulu_security.logout_success_handler" class="Sulu\Bundle\SecurityBundle\Security\LogoutSuccessHandler" >
```
danrot, line 97: No space right before the closing bracket.

lightglitch:
We have an old symfony backoffice that while it's being migrated to sulu "admin framework" still needs to work, so we merge both projects. The authentication system was integrated and in that section of the backoffice they still need to logout without going to sulu interface and it's not using ajax.

danrot:
Ok, and you are reusing the same session to check for login in both applications?

lightglitch:
Yep, same session and also merged with sulu roles and permissions.

alexander-schranz:
Is there any possibility to fallback to the symfony DefaultLogoutSuccessHandler which would take the target configuration into count or any possibility to inject the target configuration?

```yaml
            logout:
                path: sulu_admin.logout
                target: target_route_name
```

lightglitch:
Yep, if LogoutSuccessHandler extends DefaultLogoutSuccessHandler.

danrot:
If we are doing this I would prefer to [decorate the service](https://symfony.com/doc/current/service_container/service_decoration.html). Then the configuration etc. is already correctly passed by Symfony itself.

lightglitch:
Well symfony only passes the target configuration if the handler is not defined:
```
            // add logout success handler
            if (isset($firewall['logout']['success_handler'])) {
                $logoutSuccessHandlerId = $firewall['logout']['success_handler'];
            } else {
                $logoutSuccessHandlerId = 'security.logout.success_handler.'.$id;
                $logoutSuccessHandler = $container->setDefinition($logoutSuccessHandlerId, new ChildDefinition('security.logout.success_handler'));
                $logoutSuccessHandler->replaceArgument(1, $firewall['logout']['target']);
            }
```
So we can't get target value for our listener.

nnatter:
If we use decoration for the `LogoutSuccessHandlerInterface`, shouldnt we do the same thing for the `AuthenticationSuccessHandlerInterface` and the `AuthenticationFailureHandlerInterface`? 

Both of them are currently implemented by the `AuthenticationHandler` and use a hardcoded default behaviour like implemented in this PR.

danrot:
> Well symfony only passes the target configuration if the handler is not defined:
> 
> ```
>             // add logout success handler
>             if (isset($firewall['logout']['success_handler'])) {
>                 $logoutSuccessHandlerId = $firewall['logout']['success_handler'];
>             } else {
>                 $logoutSuccessHandlerId = 'security.logout.success_handler.'.$id;
>                 $logoutSuccessHandler = $container->setDefinition($logoutSuccessHandlerId, new ChildDefinition('security.logout.success_handler'));
>                 $logoutSuccessHandler->replaceArgument(1, $firewall['logout']['target']);
>             }
> ```
> 
> So we can't get target value for our listener.

That's some weird code :confused: But it looks to me like this would still work, if you don't pass and `success_handler` via the config, but just decorate the existing service instead.

danrot:
> If we use decoration for the `LogoutSuccessHandlerInterface`, shouldnt we do the same thing for the `AuthenticationSuccessHandlerInterface` and the `AuthenticationFailureHandlerInterface`?
> 
> Both of them are currently implemented by the `AuthenticationHandler` and use a hardcoded default behaviour like implemented in this PR.

If we need the behavior of the other handler as well I would say ideally yes.

nnatter:
> If we need the behavior of the other handler as well I would say ideally yes.
Not sure if we need them. At leas until now, we did not 😅 Guess my point is that it would be okay for me to hardcode the behaviour in the `LogoutSuccessHandler` too, because we already did that in the `AuthenticationHandler`.

danrot:
Yeah, since the behavior is better than before this is ok for me as well :see_no_evil: If @alexander-schranz can live with that I am fine with merging that.

@lightglitch But at least you should rebase the PR once more, I have fixed the `php-cs-fixer` issue that breaks your build in another PR.

alexander-schranz:
@danrot is fine for me.

lightglitch:
@danrot done.

danrot:
@lightglitch Thank you for your contribution!

PR #5209, opened 2020-04-12
Don't use spaces on the role identifier

lightglitch:
| Q | A
| --- | ---
| Bug fix? | yes
| BC breaks? | yes

#### Why?

Since we are concatenating with "ROLE_SULU_" the rest of the role should be coherent.

### DIFF
--- a/src/Sulu/Bundle/SecurityBundle/Entity/Role.php
+++ b/src/Sulu/Bundle/SecurityBundle/Entity/Role.php
@@ -95,12 +95,12 @@ class Role implements RoleInterface
 
     public function getRole()
     {
-        return 'ROLE_SULU_' . strtoupper($this->name);
+        return 'ROLE_SULU_' . str_replace(" ", "_", strtoupper($this->getName()));
     }
 
     public function getIdentifier()
     {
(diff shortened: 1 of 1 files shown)


danrot:
The problem with this kind of logic is that it might lead to two different roles having the same key... 

The other thing is that @nnatter has already fixed this in a different way in #5166, therefore I think this is not neccessary anymore.

nnatter:
First of all, thanks for the contribution!

Unfortunately, in addition to the duplicated identifier problem, changing this behaviour is a big BC break. With this PR, we would break every project that uses a role with a space in its security configuration. 
(Despite of the `ROLE_` prefix, symfony roles are just strings that can contain spaces: https://symfony.com/doc/current/security.html#roles). So if we wanted to do something like this, we would need to do it in a minor release.

What is your main reason for this PR? did you experience any bugs?

lightglitch:
Thx, #5166 does fix my issue because I have some symfony configuration where I have to put roles and the name "ROLE_SULU_ADMIN TEST" was just stange for me. With key parameter we can define the value it's ok.

PR #5213, opened 2020-04-14
Run preview cache clear command only when App\Kernel exists

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes -
| Related issues/PRs | #4782
| License | MIT

#### What's in this PR?

Aslong the PreviewKernel extends from App\Kernel we need to check in this listener if `App\Kernel` exist and only then allow to clear also the preview cache.

This lines can be removed when https://github.com/sulu/sulu/issues/4782 is fixed.

#### Why?

Currently all bundles with use of `cache:warmup` to generate data for phpstan crash because they don't have a App\Kernel in it.

### DIFF
--- a/src/Sulu/Bundle/PreviewBundle/EventSubscriber/CacheCommandSubscriber.php
+++ b/src/Sulu/Bundle/PreviewBundle/EventSubscriber/CacheCommandSubscriber.php
@@ -62,6 +62,13 @@ class CacheCommandSubscriber implements EventSubscriberInterface
             return;
         }
 
+        // avoid to clear cache for preview if no \App\Kernel exists
+        // this can happens in test kernels of bundles
+        // this can be removed when https://github.com/sulu/sulu/issues/4782 is fixed
+        if (!class_exists(\App\Kernel::class)) {
+            return;
+        }
+
         $previewKernel = $this->kernelFactory->create($this->environment);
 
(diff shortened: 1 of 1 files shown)


danrot:
Do I get this right: This is required to silent phpstan? What is the exact error it throws? This check feels a bit weird to me at this place... Shouldn't this cause an error already when trying to extend `\App\Kernel` if it does not exist?

alexander-schranz:
@danrot no you misunderstand it. Its not phpstan throwing an error.

If you use the [phpstan symfony plugin](https://github.com/phpstan/phpstan-symfony) you need to run `bin/console cache:clear` that the `container.xml` is generated. But as most bundles doesn't have a `App\Kernel` it will throw an error when it nows try to also clear/warmup the Preview cache.

danrot:
Can you maybe give me an example of a bundle where I can test this?

alexander-schranz:
@danrot ArticleBundle or FormBundle it seems only fail `dump-autoload --optimize` was called:

```
composer create-project sulu/form-bundle:dev-master --keep-vcs
cd vendor/sulu/sulu/
git fetch origin
git checkout origin/release/2.0
cd ../../..
composer dump-autoload -o
Tests/Application/bin/adminconsole cache:clear
```

PR #5215, opened 2020-04-14
Remove requirement for doctrine cache bundle

### LINKED ISSUE #4873: Replace Doctrine Cache Bundle with Symfony Cache Component for Symfony 5 compatibility
| Q | A
| --- | ---
| Bug? | no
| New Feature? | no
| Sulu Version | for 2.0
| Browser Version | -
| Related PRs/Issues | #4798

#### Actual Behavior

Actually it seems we are using the doctrine/doctrine-cache-bundle this will not be adopted to the Symfony 5 (https://github.com/doctrine/DoctrineCacheBundle/issues/156). 

#### Expected Behavior

Sulu should not use doctrine/cache component instead we should migrate to symfony/cache component.

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #4873
| Related issues/PRs | #4798
| License | MIT
| Documentation PR | https://github.com/sulu/sulu-docs/pull/505

#### What's in this PR?

Remove requirement for doctrine cache bundle.

#### Why?

The doctrine cache bundle is deprecated and will not be supported in symfony 5, it is suggested to use the symfony cache component instead.

#### Example Usage

~~~yaml
sulu_preview:
    cache_adapter: "cache.app"
~~~

#### BC Breaks/Deprecations

It can still use the DoctrineCacheBundle for preview cache configuration but it is now recommended using the Symfony Cache configuration.

#### To Do


src/Sulu/Bundle/PreviewBundle/Preview/PreviewCache.php
```diff
@@ -0,0 +1,102 @@
...
+ */
+
+namespace Sulu\Bundle\PreviewBundle\Preview;
+
+use Doctrine\Common\Cache\Cache;
+use Psr\Cache\CacheItemInterface;
+use Psr\Cache\CacheItemPoolInterface;
+
+/**
+ * @internal BC Layer between doctrine and symfony cache
```
danrot, line 19: Maybe also add the `deprecated` annotation already?

src/Sulu/Bundle/PreviewBundle/Preview/PreviewCache.php
```diff
@@ -0,0 +1,102 @@
...
+
+        $this->cache = $cache;
+    }
+
+    public function delete(string $id): void
+    {
+        $this->cache->delete($id);
+    }
+
+    public function contains(string $id): bool
```
danrot, line 63: Would it be easily possible to use the naming of the new `Cache` component here? Then removing this layer will be easier later.

danrot: Or do you think this class will stay anyway? I've seen that some methods consist of more than just a single call.

alexander-schranz: I personally would keep it but stay as internal class.

alexander-schranz:
@danrot tests should now working again :)

src/Sulu/Bundle/PreviewBundle/DependencyInjection/SuluPreviewExtension.php
```diff
@@ -36,10 +36,12 @@ public function load(array $configs, ContainerBuilder $container)
 
     public function prepend(ContainerBuilder $container)
     {
-        if ($container->hasExtension('doctrine_cache')) {
-            $configs = $container->getExtensionConfig($this->getAlias());
-            $config = $this->processConfiguration(new Configuration(), $configs);
+        $configs = $container->getExtensionConfig($this->getAlias());
+        $config = $this->processConfiguration(new Configuration(), $configs);
 
+        $cacheAdapter = $config['cache_adapter'];
```
danrot, line 42: Doesn't that belong in the `else` branch? Doesn't seem like you are using it outside of it

alexander-schranz: You are right did directly use the `$config['cache_adapter']` to avoid creating the variable.

src/Sulu/Bundle/PreviewBundle/DependencyInjection/SuluPreviewExtension.php
```diff
@@ -36,10 +36,10 @@ public function load(array $configs, ContainerBuilder $container)
 
     public function prepend(ContainerBuilder $container)
     {
-        if ($container->hasExtension('doctrine_cache')) {
-            $configs = $container->getExtensionConfig($this->getAlias());
-            $config = $this->processConfiguration(new Configuration(), $configs);
+        $configs = $container->getExtensionConfig($this->getAlias());
+        $config = $this->processConfiguration(new Configuration(), $configs);
 
+        if ($config['cache']['type'] && $container->hasExtension('doctrine_cache')) {
```
danrot, line 42: Should we maybe throw an error when the `doctrine_cache` extension does not exist but the `cache.type` configuration is set? I just tried to reuse the old configuration, and tripped because of that.

PR #5218, opened 2020-04-15
Add routerAttributesToBlacklist to TabViewBuilder

luca-rath:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | yes
| Deprecations? | no
| License | MIT

#### What's in this PR?

Add method `addRouterAttributesToBlacklist` to `TabViewBuilderInterface`.

#### Why?

Because it's missing in `TabViewBuilder`

#### Example Usage

~~~php
(new TabViewBuilder(...))->addRouterAttributesToBlacklist(['sortColumn', 'sortOrder']);
~~~

#### BC Breaks/Deprecations

A method has been added to `TabViewBuilderInterface`

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Admin/View/ResourceTabViewBuilder.php
+++ b/src/Sulu/Bundle/AdminBundle/Admin/View/ResourceTabViewBuilder.php
@@ -55,19 +55,6 @@ class ResourceTabViewBuilder implements ResourceTabViewBuilderInterface
         return $this;
     }
 
-    public function addRouterAttributesToBlacklist(
-        array $routerAttributesToBlacklist
-    ): TabViewBuilderInterface {
-        $oldRouterAttributesToBlacklist = $this->view->getOption('routerAttributesToBlacklist');
-        $newRouterAttributesToBlacklist = $oldRouterAttributesToBlacklist
-            ? array_merge($oldRouterAttributesToBlacklist, $routerAttributesToBlacklist)
-            : $routerAttributesToBlacklist;
-
-        $this->view->setOption('routerAttributesToBlacklist', $newRouterAttributesToBlacklist);
-
-        return $this;
-    }
-
     public function setTitleProperty(string $titleProperty): ResourceTabViewBuilderInterface
     {
         $this->view->setOption('titleProperty', $titleProperty);
--- a/src/Sulu/Bundle/AdminBundle/Admin/View/ResourceTabViewBuilder.php
+++ b/src/Sulu/Bundle/AdminBundle/Admin/View/ResourceTabViewBuilder.php
@@ -56,7 +56,7 @@ class ResourceTabViewBuilder implements ResourceTabViewBuilderInterface
 
     public function addRouterAttributesToBlacklist(
         array $routerAttributesToBlacklist
-    ): ResourceTabViewBuilderInterface {
+    ): TabViewBuilderInterface {
         $oldRouterAttributesToBlacklist = $this->view->getOption('routerAttributesToBlacklist');
         $newRouterAttributesToBlacklist = $oldRouterAttributesToBlacklist
             ? array_merge($oldRouterAttributesToBlacklist, $routerAttributesToBlacklist)
--- a/src/Sulu/Bundle/AdminBundle/Admin/View/ResourceTabViewBuilder.php
+++ b/src/Sulu/Bundle/AdminBundle/Admin/View/ResourceTabViewBuilder.php
@@ -62,6 +62,13 @@ class ResourceTabViewBuilder implements ResourceTabViewBuilderInterface
         return $this;
     }
 
+    public function addRouterAttributesToBlacklist(array $routerAttributesToBlacklist): ResourceTabViewBuilderInterface
+    {
+        $this->addRouterAttributesToBlacklistToView($routerAttributesToBlacklist);
+
+        return $this;
+    }
+
     public function getView(): View
     {
         if (!$this->view->getOption('resourceKey')) {
--- a/src/Sulu/Bundle/AdminBundle/Admin/View/ResourceTabViewBuilder.php
+++ b/src/Sulu/Bundle/AdminBundle/Admin/View/ResourceTabViewBuilder.php
@@ -15,6 +15,7 @@ class ResourceTabViewBuilder implements ResourceTabViewBuilderInterface
 {
     use ViewBuilderTrait;
     use FormViewBuilderTrait;
+    use TabViewBuilderTrait;
 
     const TYPE = 'sulu_admin.resource_tabs';
 
--- a/src/Sulu/Bundle/AdminBundle/Admin/View/ResourceTabViewBuilderInterface.php
+++ b/src/Sulu/Bundle/AdminBundle/Admin/View/ResourceTabViewBuilderInterface.php
@@ -27,10 +27,10 @@ interface ResourceTabViewBuilderInterface extends ViewBuilderInterface
      */
     public function addRouterAttributesToBackView(array $routerAttributesToBackView): self;
 
-    public function setTitleProperty(string $titleProperty): self;
-
     /**
      * @param string[] $routerAttributesToBlacklist
      */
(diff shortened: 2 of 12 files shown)


src/Sulu/Bundle/AdminBundle/Admin/View/TabViewBuilder.php
```diff
@@ -26,4 +26,17 @@ public function getView(): View
     {
         return clone $this->view;
     }
+
+    public function addRouterAttributesToBlacklist(
```
luca-rath, line 30: Should we extract this method into a trait? Unfortunately, the `TabViewBuilderTrait` is already used by children of a tab view.

src/Sulu/Bundle/AdminBundle/Admin/View/ResourceTabViewBuilder.php
```diff
@@ -56,7 +56,7 @@ public function addRouterAttributesToBackView(
 
     public function addRouterAttributesToBlacklist(
         array $routerAttributesToBlacklist
-    ): ResourceTabViewBuilderInterface {
+    ): TabViewBuilderInterface {
```
danrot, line 59: That's wrong, isn't it? This function returns a `ResourceTabViewBuilderInterface`. This way it looks like only methods of the `TabViewBuilderInterface` can be called after calling this function, which is not true.

src/Sulu/Bundle/AdminBundle/Admin/View/ResourceTabViewBuilderInterface.php
```diff
@@ -11,7 +11,7 @@
 
 namespace Sulu\Bundle\AdminBundle\Admin\View;
 
-interface ResourceTabViewBuilderInterface extends ViewBuilderInterface
+interface ResourceTabViewBuilderInterface extends TabViewBuilderInterface
```
danrot, line 14: This part was already a bit messy, but there is a reason we didn't extend interface in other places (e.g. `PreviewFormViewBuilderInterface` does not extend the `FormViewBuilderInterface`). That is like that because  [return type covariance](https://www.php.net/manual/de/migration74.new-features.php#migration74.new-features.core.type-variance) has not been supported before PHP 7.4, and we are still supporting older versions.

This is also the only reason we are using traits here, they would not be necessary if we could already add the correct return type in the first place.

src/Sulu/Bundle/AdminBundle/Admin/View/TabViewBuilder.php
danrot: Still not a big fan of traits :see_no_evil: But as explained above, the current situation makes that necessary. So yeah, we should extract that into a trait. Renaming the other trait would be a BC break... Any suggestions for a new name? Maybe we could even use the same trait, since the methods are private anyway, but that feels a bit weird...

@sulu/core-team WDYT?

danrot (changes requested):
Othwerise the code looks good :+1:

src/Sulu/Bundle/AdminBundle/Admin/View/ResourceTabViewBuilder.php
luca-rath: Because of the return type covariance, as you mentioned below.

danrot: Right, and as mentioned below that is the reason we do not inherit from other interfaces and itnroduced traits in this code :slightly_smiling_face:

src/Sulu/Bundle/AdminBundle/Admin/View/TabViewBuilder.php
luca-rath: Could we maybe rename the current `TabViewBuilderTrait` to e.g. `ChildTabViewBuilderTrait`, create a new `TabViewBuilderTrait` with `addRouterAttributesToBlacklist` and use `ChildTabViewBuilderTrait` there. Then mark the methods from `ChildTabViewBuilderTrait` inside of `TabViewBuilderTrait` as deprecated. Would that solve the bc break issue?

luca-rath: Example:

~~~php
trait TabViewBuilderTrait
{
    use ChildTabViewBuilderTrait {
        setTabTitleToView as parentSetTabTitleToView;
        setTabConditionToView as parentSetTabConditionToView;
        setTabOrderToView as parentSetTabOrderToView;
        setTabPriorityToView as parentSetTabPriorityToView;
    }

    /**
     * @deprecated Will be removed in Sulu 3.0. Use ChildTabViewBuilderTrait instead.
     */
    private function setTabTitleToView(View $view, string $tabTitle): void
    {
        $this->parentSetTabTitleToView($view, $tabTitle);
    }

    /**
     * @deprecated Will be removed in Sulu 3.0. Use ChildTabViewBuilderTrait instead.
     */
    private function setTabConditionToView(View $view, string $tabCondition): void
    {
        $this->parentSetTabConditionToView($view, $tabCondition);
    }

    /**
     * @deprecated Will be removed in Sulu 3.0. Use ChildTabViewBuilderTrait instead.
     */
    private function setTabOrderToView(View $view, int $tabOrder): void
    {
        $this->parentSetTabOrderToView($view, $tabOrder);
    }

    /**
     * @deprecated Will be removed in Sulu 3.0. Use ChildTabViewBuilderTrait instead.
     */
    private function setTabPriorityToView(View $view, int $tabPriority): void
    {
        $this->parentSetTabPriorityToView($view, $tabPriority);
    }
    
    public function addRouterAttributesToBlacklist(
        array $routerAttributesToBlacklist
    ): TabViewBuilderInterface {
        $oldRouterAttributesToBlacklist = $this->view->getOption('routerAttributesToBlacklist');
        $newRouterAttributesToBlacklist = $oldRouterAttributesToBlacklist
            ? array_merge($oldRouterAttributesToBlacklist, $routerAttributesToBlacklist)
            : $routerAttributesToBlacklist;

        $this->view->setOption('routerAttributesToBlacklist', $newRouterAttributesToBlacklist);

        return $this;
    }
}

trait ChildTabViewBuilderTrait
{
    private function setTabTitleToView(View $view, string $tabTitle): void
    {
        $view->setOption('tabTitle', $tabTitle);
    }

    private function setTabConditionToView(View $view, string $tabCondition): void
    {
        $view->setOption('tabCondition', $tabCondition);
    }

    private function setTabOrderToView(View $view, int $tabOrder): void
    {
        $view->setOption('tabOrder', $tabOrder);
    }

    private function setTabPriorityToView(View $view, int $tabPriority): void
    {
        $view->setOption('tabPriority', $tabPriority);
    }
}
~~~

danrot: Would work for me. With a small addition: Every function being deprecated must trigger a deprecation error, otherwise Symfony does not recognize it as deprecation:

```php
@trigger_error('error message', E_USER_DEPRECATED);
```

luca-rath: Are you fine with the name `ChildTabViewBuilderTrait`? I'm not very happy with it, it was just the first thing that came to my mind.

danrot: When asking me like that I would maybe turn the words: `TabViewChildBuilderTrait`, so that the `Child` is at the end. Having the `TabView` at the end would probably mean that it is a `TabView` (at least we are following this convention in other places).

luca-rath: fixed it

danrot (approved):
LGTM, but we don't have a `TabView` in the core... Can somebody test it with the project you need that for?

luca-rath:
ping @sulu/core-team would somebody be willing to try this?

Prokyonn:
@danrot @luca-rath I just tested it in the project, it's working as expected

Issue #5156, opened 2020-03-31
Teasers don't have a Webspace Key


marcostastny:
# Problem description

Currently the Content Type does not send the Webspace Key of the Item. This is according to Documentation. Sadly therefore its not possible to simply call `{{ sulu_content_path(teaser.url, teaser.webspaceKey) }}` and therefore Teasers in different Webspaces will always link to the current Webspace.

# Proposed solution

As Described by Alex there are multiple Solutions for how to fix this Problem. You could generate the full URL already in the Teaser Provider, or add a field WebspaceKey which will be null on Articles and filled on pages. I would prefer the latter, because we could then programmatically check which webspace this is and add for example custom styling.

Additionally it also would be possible to enhance the documentation, by documenting this behaviour and adding the following Workaround as a Solution to this Problem.

# Workaround

There is a workaround for this problem, by simply using `sulu:link` instead of the `sulu_content_path` function.

`<sulu:link href="{{ teaser.id }}"></sulu:link>`

alexander-schranz:
Think the best solution would be to generate the whole url in the TeaserProvider itself and still provide the webspaceKey as teaser.attribute.

marcostastny:
We just ran into a problem with the `sulu:link` way will break with an Error, when using it to display articles. `Call to a member function getLocalization() on null at "/vendor/sulu/sulu/src/Sulu/Component/Content/Compat/LocalizationFinder.php:40"` Therefore something like the following will be needed:
```twig
{% set element = teaserItem.type == 'content' ? 'sulu:link href="' ~ teaserItem.id ~ '"' : 'a href="' ~ sulu_content_path(teaserItem.url) ~ '"' %}
{% set endElement = teaserItem.type == 'content' ? 'sulu:link' : 'a' %}

<{{ element|raw }}></{{ endElement }}>
```

danrot:
@alexander-schranz The attributes of a provider is just a simple array anyway, and could easily contain context-specific options, right? Then it would also be ok for me to add a `webspaceKey` there.

danrot:
In #5264 I've added the `webspaceKey` as teaser attribute. Can we consider this issue solve because of that?

alexander-schranz:
@danrot think thats currently the best solution without any bc break gonna close it.

PR #5219, opened 2020-04-16
Add deep links in referencing field types

### LINKED ISSUE #5143: Editing and creating entities in selection field type
# Problem description

Now in sulu you can add snippets or custom entities but if you want to change a value in them you have to look for them and edit in another window. 

# Proposed solution

Add a new button near trash button to edit the content and open a popup if the edit view exists ( or maybe a link that section ). The problem may be a circular reference that would have to be controlled at 1 level perhaps. 
Extra for this feature will be inline creation :)

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes parts of #5143 
| Related issues/PRs | ---
| License | MIT
| Documentation PR | sulu/sulu-docs#507

#### What's in this PR?

This PR adds a deep link functionality to the `selection` field types and the `smart_content`.

#### BC Breaks

- The response of the `PageController::cgetAction` changed when called with the `ids` parameter
- The `SnippetController::cgetAction` now correctly returns the loaded language as the `locale`

#### To Do

- [x] Create a documentation PR (especially new configuration in `field_type_options`)
- [ ] Does `PageList` really need `rerenderAttributes` or were we just lazy, and it could handle be in this component by an `autorun`?
- [ ] Check why `smart_content` and `teaser_selection` loads twice when switching languages (caused by the fact that `autorun`s are destroyed on `componentWillUnmount` and the new `autorun` from the new field is already created, already existed in `release/2.0` branch)
- [x] Clicking page after assigning to page selection throws error
- [x] page selection does not load ghost pages
- [x] snippet selection shows empty title for ghost pages
- [x] snippet selection uses wrong language when clicked on a ghost snippet
- [ ] Check if standalone form (without `ResourceTabs`) work as well. Feels like the form standalone is not usable at all
- [x] Check if it does not fail when no `view` is configured
- [x] Check if there are double requests when `resourceKey` changes
- [ ] Show ghost locale in selection field type
- [x] Check why creating new page asks for confirmation as if route changes (was already the case in current `master`, seems to be caused by `tag_selection`) (#5246)
- [ ] Check if linking from a page in one webspace to another webspace in a non-existing locale works (this is not possible at all currently, because you can only link pages form webspaces, which share the current language)
- [ ] Check #5156 when handling teasers (might be solved by it, because we might need the `webspaceKey` anyway)


src/Sulu/Bundle/PageBundle/Controller/PageController.php
```diff
@@ -484,16 +480,20 @@ public function cgetAction(Request $request)
 
         $contents = [];
 
-        if ($parent) {
-            $contents = $contentRepository->findByParentUuid($parent, $locale, $webspaceKey, $mapping, $user);
-        } elseif ($webspaceKey) {
-            $contents = $contentRepository->findByWebspaceRoot($locale, $webspaceKey, $mapping, $user);
-        }
+        if ($ids) {
+            $contents = $contentRepository->findByUuids($ids, $locale, $mapping, $user);
```
danrot, line 484: Instead of relying on the `nodeRepository` I am now relying on the `contentRepository` to load the pages. That way is ensured that the response looks the same, no matter if the `ids` parameter is passed or not (currently `admin/api/pages?flat=true` and `admin/api/pages?flat=true&ids=1,2,3` have a different structure).

Some of the fields are the same, and the only place I know this is used is the `page_selection`, which is still working after this change. I also have a similar problem for the snippets... The controller for snippets handles the localization different, it returns the an `originalLocale`, which is the one being requested and called `locale` e.g. for the pages`. Therefore it uses the `locale` property to return from which locale the data has been loaded, not which one was requested. I would like to make that consistent, because that is causing some bugs, although it is a BC break.

@sulu/core-team Do you see any problems with these changes?

alexander-schranz: From my side I always thought:

`locale` = the requested locale
`originalLocale` = from where the data is loaded

Think we can change this to be consistent is it behave for pages and snippets different currently this is a bugfix for me we should definitly do.

nnatter: I dont know a lot about the implementation of snippets so i cannot really estimates the implications of this. But the way you describe it, it sound quite confusing at the moment. So if it makes things more consistent, I like it 🙂

danrot: I just realized that there is a similar issue for the `single_selections`, which is not that easy to solve.

When a single page is requested from the API, it does not return the `locale` as part of the response. We also cannot simply add it, since it was possible until now to have a `locale` field in your template, which would then conflict, i.e. we have to get that in some different way. So we probably need to add a `router_attributes_to_view` configuration or something like that anyway, at least for the `single_page_selection`. Question is if we should handle that here in a similar way. Then it would not matter how the `locale` is returned from the API, although I would argue it would still make sense to do that in a more consistent way.

danrot: I just omitted the `locale` in the configurations for the selections now, and this seems to be working as expected, because the sticky user locale mechanism @alexander-schranz built in #5101 kicks in. So strictly speaking this change would not be totally necessary. Should we still keep it and strive for consistency, or should I revert this change?

alexander-schranz:
Code looks good from my side but still think not having `a href` links make this feature from my side basically not finished. Because it leaks in user experience not have the possibility to open such items in a new tab.

danrot:
Well, as already said, this is true for any link we currently have in our application. I also think it would be better (especially also for the `Navigation`), but that should definitely go into a separate PR, because it affects more, and it is questionable how this should be implemented.

PR #5221, opened 2020-04-16
Add Github actions badge to README.md

nnatter:
See: https://github.com/nnatter/sulu/tree/enhancement/readme-123/README.md

### DIFF
--- a/.github/workflows/test-application.yaml
+++ b/.github/workflows/test-application.yaml
@@ -1,6 +1,10 @@
 name: Test application
 
-on: pull_request
+on:
+    pull_request:
+    push:
+        branches:
+            - master
 
 jobs:
     download-language:
--- a/.github/workflows/test-application.yaml
+++ b/.github/workflows/test-application.yaml
@@ -5,6 +5,7 @@ on:
     push:
         branches:
             - master
+            - release/**
 
 jobs:
     download-language:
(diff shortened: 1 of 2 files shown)


.github/workflows/test-application.yaml
```diff
@@ -1,6 +1,10 @@
 name: Test application
 
-on: pull_request
+on:
+    pull_request:
+    push:
+        branches:
+            - master
```
alexander-schranz, line 7: Should we also run `release/2.0` here /cc @danrot

nnatter: I mainly did this to be able to display the current status of the master branch in the readme. but it might make sense to do this for release branches too, yes. dont have a opinion here 🙂

danrot: If we do it, I would also do it for both branches. Maybe even use `release/**` to do that for all release branches. But maybe check on your forked repository first how that behaves, i.e. if this is also executed for branches that do not have a github actions configuration.

nnatter: I added `release/**` to the configuration. But I am not sure how I am able to test the behaviour? 🙈

danrot: Well, I thought you have activated github actions as well in your repo... But we can also try to merge and see what happens.

PR #5228, opened 2020-04-17
Extend sulu event from new event class

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes -
| Related issues/PRs | part of https://github.com/sulu/sulu/pull/4798
| License | MIT
| Documentation PR | -

#### What's in this PR?

Extend sulu event from new event class. ~~This is no BC Break as the new class does extend the old one in symfony 4.3 and 4.4.~~

#### Why?

The Symfony Component Event was deprecated in 4.3 and moved the event contracts.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -2,6 +2,12 @@
 
 ## dev-master
 
+### Event classes changed
+
+The Sulu `Event` classes extend now from the new `Symfony\Contracts\EventDispatcher\Event`
+class instead of the deprecated `Symfony\Component\EventDispatcher\Event`.
+If you have code depending on the old class you need to update it.
+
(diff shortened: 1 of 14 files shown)


danrot:
@alexander-schranz Can you rebase that one? And just for the record, I think that is a minor BC break, because [the new class does not seem to extend the old one](https://github.com/symfony/symfony/blob/4.4/src/Symfony/Contracts/EventDispatcher/Event.php).

alexander-schranz:
@danrot oh you are correctly. Should we add something to the UPGRADE?

danrot:
I guess we should, shouldn't we?

Issue #1527, opened 2015-08-25
Homepage is never shown in navigation


danrot:
The homepage of the website is not returned as part of the navigation (doesn't matter if tree of flat), even if it has some navigation contexts.

wachterjohannes:
how should this be handled? In my opinion a homepage should be handled like the other pages.

But this will be a BC break - at least when you call `get_navigation` without context

/cc @danrot @chirimoya how should we handle this?

wachterjohannes:
it would be quite easy to fix - we only have to add a additional `or issamenode` condition here https://github.com/sulu-io/sulu/blob/develop/src/Sulu/Bundle/WebsiteBundle/Navigation/NavigationQueryBuilder.php#L45

danrot:
If I think about it again it doesn't make too much sense to have a navigation context on the homepage. It would basically mean that there is always only one item on the most top level, if you retrieve the navigation as a tree. It would basically only make sense for the flat representation. Maybe we should not display this dropdown for homepages at all, and if somebody wants to have the homepage as a navigation item, an internal link has to be created?

alexander-schranz:
when then the homepage should be on the same level as the content pages. but would agree not to display the dropdown on homepages. internal link seems for me a good solution for this who want the homepage in a navigation.

danrot:
@chirimoya That's actually also interesting for the new UI. Do you think the navigation dropdown should be shown in the settings tab of the homepage? Or should the return value of the navigation be fixed in someway.

PhilippSchreiber:
Just jumping in from a user perspective: 

If someone adds the home page to a nav context (like I did just 60min ago) it should appear there. If the user gets a tree with only the home page on the top level he would know what just happened. If the home page simply does not appear everyone will search for a bug and then after 60min debugging his own code base will do a google search and find this issue on github ;)

In my case I want to get a flat navigation and now it's a pain for me to add the home page (from a SEO perspective an internal link with a 301 redirect is a really bad solution!).

Please do not "protect" the users.

alexander-schranz:
@PhilippSchreiber sorry that you had to fight with this issue. Just want to mention it that when creating a Page from Type Internal Link add it to the Navigation there will be no 301 Redirect as it will directly link to the homepage so there will be no difference or SEO issue. Else your dev did maybe use the false attribute to create the navigation, please check that the navigation is build using the `url` attribute and not the path attribute.

PhilippSchreiber:
[...]

Issue #5029, opened 2020-02-03
Introduce screenshot tests for grid


danrot:
| Q | A
| --- | ---
| Sulu Version | 2.0
| Browser Version | ---

# Actual Behavior

We don't have any screenshot tests for our `Grid` component in Javascript.

# Possible Solutions

We should add some screenshot tests, because otherwise we are going to break this grid again and again. There is also already some stuff that should be changed, but I am afraid to do so, because I might break other scenarios. The scenarios I am currently aware of that need to be tested using snapshots are:

- Root level properties after a section
- A section after another section
- Two section with a combined `colspan` value of 12 or less should float next to each other
- There should be no margin after the last section

PR #5232, opened 2020-04-18
Remove navigation context from homepage

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes -
| Related issues/PRs | related to #1527
| License | MIT
| Documentation PR | -

#### What's in this PR?

Remove navigation context from homepage.

#### Why?

Currently the homepage is never shown so we should hide that option from the settings tab for the homepage.

### DIFF
--- a/src/Sulu/Bundle/PageBundle/Resources/config/forms/page_settings.xml
+++ b/src/Sulu/Bundle/PageBundle/Resources/config/forms/page_settings.xml
@@ -52,7 +52,7 @@
             name="navContexts"
             type="page_settings_navigation_select"
             colspan="6"
-            visibleCondition="shadowOn == false"
+            visibleCondition="shadowOn == false &amp;&amp; path != ''"
         >
             <meta>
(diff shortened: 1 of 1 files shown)


danrot:
Is it possible that you have not branched from the latest release/2.0 state? I thought the error you have in here has been already fixed.

alexander-schranz:
@danrot don't see it hide on the homepage does it work for you?

danrot:
Sorry, I was talking about the circle ci build, there is an error :see_no_evil:

danrot:
The margin on to the top element is not correct, but this is also the case currently when the page is switched to a shadow page... Something to be done when #5029 is implemented.

alexander-schranz:
@danrot yes the problem is that the first section always gets a margin top. The same happens for old projects which updates to 2.0 and still have the highlight section in there templates.

PR #5179, opened 2020-04-06
Fix sorting of smart-content for doctrine entities

danrot:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #5142 
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR fixes the loading of doctrine entities in smart contents with sorting.

PR #5234, opened 2020-04-19
Enhance DataProviderRepositoryTrait

luca-rath:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This enhancement fixes sorting of the `DataProviderRepositoryTrait`.

#### Why?

Because `sortBy` is not supposed to be array anymore.

### DIFF
--- a/src/Sulu/Component/SmartContent/Orm/DataProviderRepositoryTrait.php
+++ b/src/Sulu/Component/SmartContent/Orm/DataProviderRepositoryTrait.php
@@ -67,7 +67,14 @@ trait DataProviderRepositoryTrait
 
         if (array_key_exists('sortBy', $filters)) {
             $sortMethod = array_key_exists('sortMethod', $filters) ? $filters['sortMethod'] : 'asc';
-            $this->appendSortBy($filters['sortBy'], $sortMethod, $queryBuilder, 'c', $locale);
+            $sortColumn = $filters['sortBy'];
+
+            if ($sortColumn === explode('.', $sortColumn)[0]) {
+                $sortColumn = 'c.' . $sortColumn;
+            }
+
+            $queryBuilder->addSelect($sortColumn);
+            $this->appendSortBy($sortColumn, $sortMethod, $queryBuilder, 'c', $locale);
         }
 
         $parameter = array_merge($parameter, $this->append($queryBuilder, 'c', $locale, $options));
@@ -374,8 +381,7 @@ trait DataProviderRepositoryTrait
             $this->appendSortByJoins($queryBuilder, $alias, $locale);
         }
 
-        $queryBuilder->addSelect($alias . '.' . $sortBy);
-        $queryBuilder->orderBy($alias . '.' . $sortBy, $sortMethod);
+        $queryBuilder->orderBy($sortBy, $sortMethod);
     }
 
     /**
--- a/src/Sulu/Component/SmartContent/Orm/DataProviderRepositoryTrait.php
+++ b/src/Sulu/Component/SmartContent/Orm/DataProviderRepositoryTrait.php
@@ -58,21 +58,16 @@ trait DataProviderRepositoryTrait
     {
         $parameter = [];
 
-        $queryBuilder = $this->createQueryBuilder('c')
-            ->select('c.id')
-            ->distinct()
-            ->orderBy('c.id', 'ASC');
+        $queryBuilder = $this->createQueryBuilder('c');
+
+        $parameter = array_merge($parameter, $this->appendSelect($queryBuilder, 'c', $locale, $options));
 
         $tagRelation = $this->appendTagsRelation($queryBuilder, 'c');
         $categoryRelation = $this->appendCategoriesRelation($queryBuilder, 'c');
 
-        if (array_key_exists('sortBy', $filters) && is_array($filters['sortBy'])) {
+        if (array_key_exists('sortBy', $filters)) {
             $sortMethod = array_key_exists('sortMethod', $filters) ? $filters['sortMethod'] : 'asc';
             $this->appendSortBy($filters['sortBy'], $sortMethod, $queryBuilder, 'c', $locale);
-
-            foreach ($filters['sortBy'] as $sortColumn) {
-                $queryBuilder->addSelect($sortColumn);
-            }
         }
 
         $parameter = array_merge($parameter, $this->append($queryBuilder, 'c', $locale, $options));
@@ -284,6 +279,7 @@ trait DataProviderRepositoryTrait
     /**
      * Append additional condition to query builder for "findByFilters" function.
      *
+     * @param string $alias
      * @param string $locale
      * @param array $options
      *
@@ -292,6 +288,25 @@ trait DataProviderRepositoryTrait
     protected function append(QueryBuilder $queryBuilder, $alias, $locale, $options = [])
     {
         // empty implementation can be overwritten by repository
+        return [];
+    }
+
+    /**
+     * Append select to query builder for "findByFilters" function.
+     *
+     * @param string $alias
+     * @param string $locale
+     * @param array $options
+     *
+     * @return array parameters for query
+     */
+    protected function appendSelect(QueryBuilder $queryBuilder, $alias, $locale, $options = [])
+    {
+        $queryBuilder
+            ->select($alias . '.id')
+            ->distinct()
+            ->orderBy($alias . '.id', 'ASC');
+
         return [];
     }
 
@@ -348,12 +363,10 @@ trait DataProviderRepositoryTrait
     /**
      * Extension point to append order.
      *
-     * @param array $sortBy
+     * @param string $sortBy
      * @param string $sortMethod
      * @param string $alias
      * @param string $locale
-     *
-     * @return array parameters for query
      */
     protected function appendSortBy($sortBy, $sortMethod, QueryBuilder $queryBuilder, $alias, $locale)
     {
@@ -361,7 +374,8 @@ trait DataProviderRepositoryTrait
             $this->appendSortByJoins($queryBuilder, $alias, $locale);
         }
 
-        $queryBuilder->orderBy($sortBy, $sortMethod);
+        $queryBuilder->addSelect($alias . '.' . $sortBy);
+        $queryBuilder->orderBy($alias . '.' . $sortBy, $sortMethod);
     }
 
     /**
--- a/src/Sulu/Component/SmartContent/Orm/DataProviderRepositoryTrait.php
+++ b/src/Sulu/Component/SmartContent/Orm/DataProviderRepositoryTrait.php
@@ -56,9 +56,12 @@ trait DataProviderRepositoryTrait
      */
     private function findByFiltersIds($filters, $page, $pageSize, $limit, $locale, $options = [])
     {
-        $queryBuilder = $this->createQueryBuilder('c');
+        $parameter = [];
 
-        $parameter = $this->appendSelect($queryBuilder, 'c', $locale, $options);
+        $queryBuilder = $this->createQueryBuilder('c')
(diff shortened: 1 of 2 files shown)


src/Sulu/Component/SmartContent/Orm/DataProviderRepositoryTrait.php
```diff
@@ -58,21 +58,16 @@ private function findByFiltersIds($filters, $page, $pageSize, $limit, $locale, $
...
-            ->distinct()
-            ->orderBy('c.id', 'ASC');
+        $queryBuilder = $this->createQueryBuilder('c');
+
+        $parameter = array_merge($parameter, $this->appendSelect($queryBuilder, 'c', $locale, $options));
 
         $tagRelation = $this->appendTagsRelation($queryBuilder, 'c');
         $categoryRelation = $this->appendCategoriesRelation($queryBuilder, 'c');
 
-        if (array_key_exists('sortBy', $filters) && is_array($filters['sortBy'])) {
```
luca-rath, line 69: `sortBy` should never be an array, right?

luca-rath: This fix allows sorting during `findByFilterIds`

nnatter: i think it cannot be an array anmore. at least https://github.com/sulu/sulu/pull/5179/files#diff-3ba8d43dcc752376edcd2f5156c711bfL360 removed some functionality regarding to this too

src/Sulu/Component/SmartContent/Orm/DataProviderRepositoryTrait.php
```diff
@@ -58,21 +58,23 @@ private function findByFiltersIds($filters, $page, $pageSize, $limit, $locale, $
     {
         $parameter = [];
 
-        $queryBuilder = $this->createQueryBuilder('c')
-            ->select('c.id')
-            ->distinct()
-            ->orderBy('c.id', 'ASC');
+        $queryBuilder = $this->createQueryBuilder('c');
+
+        $parameter = array_merge($parameter, $this->appendSelect($queryBuilder, 'c', $locale, $options));
```
danrot, line 63: Parameter will always be an empty array here, right? Why don't we remove the above variable then, and initialize it here? Then we also would not need to use `array_merge`.

src/Sulu/Component/SmartContent/Orm/DataProviderRepositoryTrait.php
danrot: Right, it cannot be an array at all. Therefore you can remove the code handling sorting by multiple columns.

src/Sulu/Component/SmartContent/Orm/DataProviderRepositoryTrait.php
```diff
@@ -348,12 +370,10 @@ protected function appendDatasource($datasource, $includeSubFolders, QueryBuilde
...
      *
-     * @param array $sortBy
+     * @param string $sortBy
      * @param string $sortMethod
      * @param string $alias
      * @param string $locale
-     *
-     * @return array parameters for query
      */
     protected function appendSortBy($sortBy, $sortMethod, QueryBuilder $queryBuilder, $alias, $locale)
```
danrot, line 378: This function also contains some array specific code.

src/Sulu/Component/SmartContent/Orm/DataProviderRepositoryTrait.php
```diff
@@ -295,6 +298,25 @@ protected function append(QueryBuilder $queryBuilder, $alias, $locale, $options
...
+     * @param array $options
+     *
+     * @return array parameters for query
+     */
+    protected function appendSelect(QueryBuilder $queryBuilder, $alias, $locale, $options = [])
+    {
+        $queryBuilder
+            ->select($alias . '.id')
+            ->distinct()
+            ->orderBy($alias . '.id', 'ASC');
```
danrot, line 315: This does more than just append a select clause... Do we really need to extract that as a separate function?

danrot: Also, the `createQueryBuilder` function can already be used to something like this, if I am not mistaken. E.g. the [`MediaDataProvider`](https://github.com/sulu/sulu/blob/b48343f/src/Sulu/Bundle/MediaBundle/Entity/MediaDataProviderRepository.php#L205) already makes use of that.

danrot (changes requested):
And don't forget to add tests when we've agreed on everything :wink:

src/Sulu/Component/SmartContent/Orm/DataProviderRepositoryTrait.php
luca-rath: But the `->select()` overrides all previous selects 🤔 then we would have to change that to `->addSelect()`

luca-rath: The same for the `->orderBy()`

danrot: The only call I can find of this function is [here](https://github.com/sulu/sulu/pull/5234/files#diff-3ba8d43dcc752376edcd2f5156c711bfR63). So it only seems to be used for initialization. And as mentioned in my previous comment, there is already the `createQueryBuilder` function for that.

Or am I missing something? Is this function also called in other places?

luca-rath: At least, in `findByFiltersIds()`, your custom `->select()` or `->orderBy`, that has been set in your custom `DataProviderRepository` in `createQueryBuilder()` would always be overridden [here](https://github.com/sulu/sulu/blob/f5bff6a9cccfc6587b6e59e0a4464ec23b15facf/src/Sulu/Component/SmartContent/Orm/DataProviderRepositoryTrait.php#L62-L64)

luca-rath: But if we change `->select()` to `->addSelect()` and `->orderBy()` to `->addOrderBy()`, it wouldn't override previous selects or orderBys

src/Sulu/Component/SmartContent/Orm/DataProviderRepositoryTrait.php
luca-rath: I think you removed that in #5179

src/Sulu/Component/SmartContent/Orm/DataProviderRepositoryTrait.php
luca-rath: I just realized that i create the select depending on `$locale` and `$options`. So i can't use `createQueryBuilder` anyways.

src/Sulu/Component/SmartContent/Orm/DataProviderRepositoryTrait.php
```diff
@@ -56,23 +56,23 @@ public function findByFilters($filters, $page, $pageSize, $limit, $locale, $opti
...
 
-        if (array_key_exists('sortBy', $filters) && is_array($filters['sortBy'])) {
+        if (array_key_exists('sortBy', $filters)) {
             $sortMethod = array_key_exists('sortMethod', $filters) ? $filters['sortMethod'] : 'asc';
-            $this->appendSortBy($filters['sortBy'], $sortMethod, $queryBuilder, 'c', $locale);
+            $sortColumn = $filters['sortBy'];
 
-            foreach ($filters['sortBy'] as $sortColumn) {
-                $queryBuilder->addSelect($sortColumn);
+            if ($sortColumn === explode('.', $sortColumn)[0]) {
```
danrot, line 70: What are you checking for here?

src/Sulu/Component/SmartContent/Orm/DataProviderRepositoryTrait.php
danrot: Can you link or paste your example code? I simply don't want to have two very similar methods, because it makes things harder to understand. And `createQueryBuilder` is at least sometimes already used in a similar context.

luca-rath: https://github.com/sulu/SuluContentBundle/pull/131/files#diff-04d8a1150ad493bbb84cdc1b20be43fcR150

src/Sulu/Component/SmartContent/Orm/DataProviderRepositoryTrait.php
luca-rath: checking if `$sortColumn` contains a `.` ... just realized i could use `strpos` for that.

src/Sulu/Component/SmartContent/Orm/DataProviderRepositoryTrait.php
danrot: Is it just me, or is there really no usage of the `DataProviderRepositoryTrait`?

Actually I am asking because I'd like to see how it is called. Wouldn't it also work if we add `$options` as an argument to the `createQueryBuilder` method?

src/Sulu/Component/SmartContent/Orm/DataProviderRepositoryTrait.php
danrot: Actually I was asking why you were checking if it contains a `.` :slightly_smiling_face:

src/Sulu/Component/SmartContent/Orm/DataProviderRepositoryTrait.php
luca-rath: We decided to not use the `DataProviderRepositoryTrait` anymore, so we have full control over the code. Do you still want some of the changes of this pull request to be merged? Then I'm going to adjust it now as you want.

src/Sulu/Component/SmartContent/Orm/DataProviderRepositoryTrait.php
luca-rath: So I can prepend the `sortColumn` with the `alias`.

src/Sulu/Component/SmartContent/Orm/DataProviderRepositoryTrait.php
danrot: May I ask why? Do you need to further optimize your query than this trait would allow you to?

And I think most of the changes in here make sense. I would just integrate them in the `createQueryBuilder` method instead of creating a new one. I think that should be possible, shouldn't it?

luca-rath: `createQueryBuilder()` is used by both, `findByFilters()` and `findByFiltersIds()`.

`findByFilters()` selects the whole entity where `id` in `$ids` ordered by `id` `ASC`.

`findByFilterIds()` selects only the distinct `id` without where and also ordered by `id` `ASC`.

The only common code here is the `->orderBy($alias . '.id', 'ASC')`, so we can't extract that code into `createQueryBuilder()` IMO. Also this would be a quite bigger bc break, because it's likely that someone has overridden the `createQueryBuilder()` method.

danrot: And you want to do that only in the `findByFiltersIds` function? As said, one of the issues here IMO is that it does not only append thee select, but a few things more...

I've had a closer look, and there is already the `append` function in this trait offering the same method signature, and it seems to be called in this function already. Why can't we use that one for what you are trying to implement here?

luca-rath:
Closing because it's not needed anymore

PR #5236, opened 2020-04-19
Change service definition for DefaultController

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes -
| Related issues/PRs | part of #4798, https://github.com/symfony/symfony/issues/36495 
| License | MIT

#### What's in this PR?

Change service definition for DefaultController.

#### Why?

~~Not sure why but the alias does not longer work and the controller does not get the container correctly set. So at current state only switch alias and id did work supporting symfony 5.~~

As shown in the linked symfony issue we need to make the service and alias public that it works in symfony 5.

#### To Do

### DIFF
--- a/src/Sulu/Bundle/WebsiteBundle/DependencyInjection/SuluWebsiteExtension.php
+++ b/src/Sulu/Bundle/WebsiteBundle/DependencyInjection/SuluWebsiteExtension.php
@@ -11,6 +11,7 @@
 
 namespace Sulu\Bundle\WebsiteBundle\DependencyInjection;
 
+use Sulu\Bundle\WebsiteBundle\Controller\DefaultController;
 use Sulu\Bundle\WebsiteBundle\Sitemap\SitemapProviderInterface;
 use Sulu\Component\HttpKernel\SuluKernel;
 use Symfony\Component\Config\FileLocator;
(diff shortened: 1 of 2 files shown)


src/Sulu/Bundle/WebsiteBundle/DependencyInjection/SuluWebsiteExtension.php
```diff
@@ -120,6 +121,10 @@ public function load(array $configs, ContainerBuilder $container)
 
             // default local provider
             $container->setAlias('sulu_website.default_locale.provider', $config['default_locale']['provider_service_id']);
+
+            // add alias for default controller
+            $container->setAlias(DefaultController::class, 'sulu_website.default_controller')
```
alexander-schranz, line 126: aliases can not be tagged so we need to define it in php inside an if statement

danrot: What exactly are you saying here? Isn't it possible to set the `public="true"` attribute on the alias as it was before?

alexander-schranz: Defining the alias in the `xml` fails on symfony 5 in the `admin context` as the target service `sulu_website.default_controller` does not exist there.

alexander-schranz: and `aliases` can not be tagged with `<tag name="sulu.context" context="website" />`

PR #5243, opened 2020-04-20
Change Document Fixtures to dependency injection

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes -
| License | MIT
| Documentation PR | https://github.com/sulu/sulu-docs/pull/514

#### What's in this PR?

Avoid call of kernel root dir in fixtures loader. Change the fixtures loader to service injection.

#### Why?

The kernel.root_dir does not longer exist in symfony 5.

#### To Do

- [x] Create a documentation PR https://github.com/sulu/sulu-docs/pull/514
- [x] Add autowire alternative of fixture services like its now also done in doctrine


src/Sulu/Bundle/DocumentManagerBundle/Command/FixturesLoadCommand.php
```diff
@@ -113,7 +113,13 @@ function(BundleInterface $bundle) {
                 $this->kernel->getBundles()
             );
 
-            $directories[] = $this->kernel->getRootDir();
+            if (method_exists($this->kernel, 'getRootDir')) {
+                $directories[] = $this->kernel->getRootDir();
+            }
+
+            $directories[] = $this->kernel->getProjectDir();
+            $directories[] = $this->kernel->getProjectDir() . '/src';
```
alexander-schranz, line 121: I find this detection based on the directory should be deprecated. We should go for service injection like its also done for doctrine fixture bundle since [3.0](https://github.com/doctrine/DoctrineFixturesBundle/blob/master/UPGRADE.md#upgrade-to-30)

@sulu/core-team what do you think defining a `sulu_document_manager.fixture` tag and autowire the `DocumentFixtureInterface` to it so we can inject it here.

danrot: Sounds good to me. Would just call the tag `sulu.document_manager_fixture`, because prefixing everything with `sulu.` is more consistent IMO.

src/Sulu/Bundle/DocumentManagerBundle/DependencyInjection/Compiler/DocumentFixturePass.php
```diff
@@ -0,0 +1,43 @@
...
+{
+    const TAG_NAME = 'sulu.document_manager_fixture';
+
+    public function process(ContainerBuilder $container)
+    {
+        foreach ($container->findTaggedServiceIds(self::TAG_NAME) as $id => $tags) {
+            $definition = $container->getDefinition($id);
+
+            if (is_subclass_of($definition->getClass(), ContainerAwareInterface::class)) {
+                @trigger_error(
```
danrot, line 29: Do we still need that Compiler pass? Shouldn't that also be something we can safely delete now?

alexander-schranz: No this does not work with symfony. Symfony does not autowire the ContainerAwareInterface to make the update as smooth as possible I would keep it.

danrot: I was especially wondering, because the Fixture on our website implements that `ContainerAwareInterface`, and I still can't find this deprecation anywhere... Can you explain why that is? I assumed that this code is not called for service-container registered features because of that, which would make that class obsolete.

alexander-schranz: The `call` for `setContainer` is not set without the CompilerPass:

```bash
bin/console debug:container App\DataFixtures\Document\DocumentFixture --show-arguments
```

If you disable the compilerpass and clear the cache you will then get:

```bash
Loading fixtures
 - loading "App\DataFixtures\Document\DocumentFixture"
Symfony\Component\Debug\Exception\FatalThrowableError^ {#2035
  -originalClassName: "Error"
  #message: "Call to a member function get() on null"
  #code: 0
  #file: "./src/DataFixtures/Document/DocumentFixture.php"
```

> I still can't find this deprecation anywhere

Maybe the deprecation isn't logged correctly because its a compilerpass, I will add also a deprecation to the command itself.

alexander-schranz: @danrot deprecation message also added to the command

PR #5244, opened 2020-04-20
Avoid duplicated boot of test kernel

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes -
| License | MIT

#### What's in this PR?

Avoid duplicate boot of test kernel.

#### Why?

In Symfony its not longer allowed that TestKernel is booted multiple times. Currently in most cases our TestKernel is booted again in `createClient` so we need to move that at the begin in most setups.

#### To Do


src/Sulu/Bundle/AdminBundle/Tests/Functional/Controller/AdminControllerTest.php
```diff
@@ -53,7 +59,8 @@ public function testGetConfig()
 
     public function testGetNotExistingMetdata()
     {
-        $client = $this->createAuthenticatedClient();
+        $client = $this->client;
```
danrot, line 62: Instead of writing it like this I would prefer to use `$this->client` everywhere. But if it is too much work to change everywhere this works for me as well.

src/Sulu/Bundle/AudienceTargetingBundle/Tests/Functional/Controller/TargetGroupControllerTest.php
```diff
@@ -374,15 +379,17 @@ public function testSingleDelete()
 
         $this->getEntityManager()->flush();
 
-        $client = $this->createAuthenticatedClient();
-        $client->request('DELETE', self::BASE_URL . '/' . $targetGroup->getId());
+        $targetGroupId = $targetGroup->getId();
```
danrot, line 382: Is there any specific reason to extract that into a separate variable? In tests I usually try to avoid something like this, because if the `id` of the target group changes for some reason the testing logic might be flawed.

src/Sulu/Bundle/CategoryBundle/Tests/Functional/Controller/CategoryControllerTest.php
```diff
@@ -1436,26 +1439,28 @@ public function testDelete()
...
         $client->request(
             'DELETE',
-            '/api/categories/' . $category->getId()
+            '/api/categories/' . $categoryId
         );
 
         $this->assertHttpStatusCode(204, $client->getResponse());
 
-        $client = $this->createAuthenticatedClient();
+        $client = $this->client;
```
danrot, line 1452: This line would not be necessary, would it?

src/Sulu/Bundle/ContactBundle/Tests/Functional/Controller/ContactControllerTest.php
```diff
@@ -1877,20 +1873,22 @@ public function testDelete()
...
+        $contactId = $contact->getId();
+
+        $client = $this->client;
+        $client->request('DELETE', '/api/contacts/' . $contactId);
 
         $this->assertHttpStatusCode(204, $client->getResponse());
 
-        $client = $this->createTestClient();
-        $client->request('GET', '/api/contacts/' . $contact->getId());
+        $client = $this->client;
```
danrot, line 1883: Would also not be necessary. If we would always be using `$this->client` this error also would not happen :see_no_evil:

src/Sulu/Bundle/CoreBundle/Tests/Functional/Controller/LocalizationControllerTest.php
```diff
@@ -11,25 +11,14 @@
...
-    public function setUp(): void
-    {
-        $this->em = $this->getEntityManager();
-        $this->purgeDatabase();
-    }
-
     public function testCgetAction()
     {
         $client = $this->createAuthenticatedClient();
+        $this->purgeDatabase();
```
danrot, line 21: Shouldn't we keep that in the `setUp`, in case there'll be more tests added?

src/Sulu/Bundle/CustomUrlBundle/EventListener/CustomUrlSerializeEventSubscriber.php
```diff
@@ -85,12 +85,20 @@ public function onPostSerialize(ObjectEvent $event)
             $customUrlProperty
         );
 
-        $creatorFullName = $this->userManager->getFullNameByUserId($customUrl->getCreator());
+        $creatorFullName = null;
+        if ($creator = $customUrl->getCreator()) {
```
danrot, line 89: This is already part of #5242, right?

src/Sulu/Bundle/DocumentManagerBundle/Command/FixturesLoadCommand.php
```diff
@@ -113,7 +113,13 @@ function(BundleInterface $bundle) {
                 $this->kernel->getBundles()
             );
 
-            $directories[] = $this->kernel->getRootDir();
+            if (method_exists($this->kernel, 'getRootDir')) {
+                $directories[] = $this->kernel->getRootDir();
+            }
+
+            $directories[] = $this->kernel->getProjectDir();
+            $directories[] = $this->kernel->getProjectDir() . '/src';
```
danrot, line 121: This is also part of another PR, isn't it?

src/Sulu/Bundle/MediaBundle/Tests/Application/Kernel.php
```diff
@@ -20,6 +21,10 @@ public function registerContainerConfiguration(LoaderInterface $loader)
     {
         parent::registerContainerConfiguration($loader);
 
-        $loader->load(__DIR__ . '/config/config.yml');
+        if (SuluKernel::CONTEXT_WEBSITE === $this->getContext()) {
```
danrot, line 24: Why was that necessary?

src/Sulu/Bundle/MediaBundle/Tests/Functional/Controller/MediaControllerTest.php
```diff
@@ -341,11 +347,10 @@ protected function setUpCollection()
         $this->collection->setStyle(json_encode($style));
 
         // Create Collection Type
-        $this->collectionType = new CollectionType();
-        $this->collectionType->setName('Default Collection Type');
-        $this->collectionType->setDescription('Default Collection Type');
+        $collectionTypeFixtures = new LoadCollectionTypes();
```
danrot, line 350: How many collection types does this create? What if just one would be enough for this test? Then this test is much slower now, isn't it?

src/Sulu/Bundle/PageBundle/Tests/Functional/Controller/PageControllerTest.php
```diff
@@ -42,7 +43,6 @@ class PageControllerTest extends SuluTestCase
 
     public function setUp(): void
     {
-        parent::setUp();
```
danrot, line 45: Aren't we missing something because of that?

src/Sulu/Bundle/RouteBundle/Tests/Unit/Generator/SymfonyExpressionTokenProviderTest.php
```diff
@@ -55,6 +55,8 @@ public function testResolveNotExists()
     {
         $this->expectException(CannotEvaluateTokenException::class);
         $translator = $this->prophesize(Translator::class);
+        $translator->getLocale()->willReturn('en');
+        $translator->setLocale('en')->shouldBeCalled();
```
danrot, line 59: Why did that change?

src/Sulu/Bundle/SearchBundle/Tests/Functional/Controller/SearchControllerTest.php
```diff
@@ -43,9 +43,8 @@ class SearchControllerTest extends SuluTestCase
 
     public function setUp(): void
     {
-        parent::setUp();
```
danrot, line 46: Can we just remove that?

src/Sulu/Bundle/SnippetBundle/Tests/Functional/Controller/SnippetControllerTest.php
```diff
@@ -55,10 +55,9 @@ class SnippetControllerTest extends SuluTestCase
 
     public function setUp(): void
     {
-        parent::setUp();
```
danrot, line 58: Can we just remove that?

src/Sulu/Bundle/WebsiteBundle/Cache/CacheClearer.php
```diff
@@ -78,7 +79,7 @@ public function clear()
 
             $this->cacheManager->invalidateDomain($request->getHost());
 
-            $this->eventDispatcher->dispatch(Events::CACHE_CLEAR);
+            $this->eventDispatcher->dispatch(new CacheClearEvent(), Events::CACHE_CLEAR);
```
danrot, line 82: I thought they changed this call to only have on parameter? :thinking:

danrot: And how is this related to this PR anyway?

danrot:
Oh, and I am not sure if you've realized, but the tests are not working.

alexander-schranz:
> 1. I would avoid lines like $client = $this->client. I can imagine that you've done that because it is faster, but would removing them not be executing two regex?

Also because of git history and the code review it would make it really hard else to see what I also needed to change.

> 2. You have extracted quite some IDs from entities in the tests? Why is that? I tried to avoid storing values like that in tests, because it might hide some bugs (imagine the value is changes by the system under test in the mean time).

That it basically needed now as the Test itself share the same Kernel as the Request they share the same Doctrine Unit of Work and if you remove a entity from the database your stored `$entity` variable will not longer have its id `$entity->getId()` is null until its removed. As in most deletions we do after also a `GET` request on e.g. `/api/media/ . $media->getId()` will now return `/api/media/` because getId is null. So that we can really do that check we need to store that `id` in a seperate variable.

src/Sulu/Bundle/AudienceTargetingBundle/Tests/Functional/Controller/TargetGroupControllerTest.php
alexander-schranz: commented below. basically `$targetGroup->getId();` will return null after it was removed because they share now the same doctrine unit of work.

src/Sulu/Bundle/MediaBundle/Tests/Application/Kernel.php
alexander-schranz: To run MediaStreamControllerTest and MediaWebsiteControllerTest under the correct WebsiteKernel as they test website routes and not admin routes.

src/Sulu/Bundle/PageBundle/Tests/Functional/Controller/PageControllerTest.php
alexander-schranz: No there is no parent implimentation and as most TestCases don't call it I removed it also from this ones to be consistent.

src/Sulu/Bundle/RouteBundle/Tests/Unit/Generator/SymfonyExpressionTokenProviderTest.php
alexander-schranz: The translator does not longer accept `null` as locale and did fail. Related to this changes #5137

src/Sulu/Bundle/SearchBundle/Tests/Functional/Controller/SearchControllerTest.php
alexander-schranz: yes they is no parent implementation I removed it to be consistent

src/Sulu/Bundle/SnippetBundle/Tests/Functional/Controller/SnippetControllerTest.php
alexander-schranz: yes they is no parent implementation I removed it to be consistent

src/Sulu/Bundle/WebsiteBundle/Cache/CacheClearer.php
alexander-schranz: Oh sure I wanted to move this into an own PR 🙈 . 
No it are still 2 parameters the second one the string is optional and needed to be keeped for backward compatibility.

danrot:
1. I think I would not make code "weird" because we don't want to change the git history. And GitHub would probably do a good job on higlighting all the added `$this->` parts, so I think I would be able to handle the review.
2. That totally makes sense :+1:

src/Sulu/Bundle/WebsiteBundle/Cache/CacheClearer.php
alexander-schranz: Own PR created #5247

src/Sulu/Bundle/MediaBundle/Tests/Functional/Controller/MediaControllerTest.php
alexander-schranz: It create the 2 types (Default and System) some test failed now because the system collection was not there or had a false id 🤔

src/Sulu/Bundle/DocumentManagerBundle/Command/FixturesLoadCommand.php
alexander-schranz: Removed from this PR

src/Sulu/Bundle/CustomUrlBundle/EventListener/CustomUrlSerializeEventSubscriber.php
alexander-schranz: Removed from this PR

src/Sulu/Bundle/CoreBundle/Tests/Functional/Controller/LocalizationControllerTest.php
alexander-schranz: Did move it to the setup.

src/Sulu/Bundle/CategoryBundle/Tests/Functional/Controller/CategoryControllerTest.php
alexander-schranz: Removed all appearance of `$client = $this->client;`

danrot:
Still wondering if the file "src/Sulu/Bundle/MediaBundle/Tests/Unit/Media/Manager/test.txt" is necessary...

danrot:
And some tests are still failing.

alexander-schranz:
@danrot The `test.txt` is needed because we need to construct the UploadedFile for the test and Symfony will do in Symfony 5 a file_exists check. 

Fixed the tests but still would need #5242 that the CustomUrlBundle tests work.

alexander-schranz:
@danrot lowest tests are also fixed.

danrot:
Would have been nice if you would have done that in a separate commit :confused: But from what I've seen the fix seems to be some `clear` calls on the entity manager, is that right? Why was that necessary?

alexander-schranz:
@danrot its a seperate commit: https://github.com/sulu/sulu/pull/5244/commits/bec2236eeab274f098eeaf677053dea4d9599819

alexander-schranz:
The unit of work seems to bettoo buggy in the lower versions to be shared between the and think its also good to clear the entity manager else you can not be sure if its really loaded from database and not from the unit of work when requested a edit item.

PR #5247, opened 2020-04-21
Fix deprecated cache clear event

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes -
| Related issues/PRs | -
| License | MIT

#### What's in this PR?

Fix deprecated cache clear event.

#### Why?

Throwing an event without an object is deprecated, so we need to create a new CacheClearEvent class.

### DIFF
--- a/src/Sulu/Bundle/WebsiteBundle/Tests/Unit/Sulu/Bundle/WebsiteBundle/Cache/CacheClearerTest.php
+++ b/src/Sulu/Bundle/WebsiteBundle/Tests/Unit/Sulu/Bundle/WebsiteBundle/Cache/CacheClearerTest.php
@@ -1,0 +1,143 @@
+<?php
+
+/*
+ * This file is part of Sulu.
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\WebsiteBundle\Tests\Unit\Sulu\Bundle\WebsiteBundle\Cache;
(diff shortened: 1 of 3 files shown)


src/Sulu/Bundle/WebsiteBundle/Tests/Unit/Sulu/Bundle/WebsiteBundle/Cache/CacheClearerTest.php
```diff
@@ -0,0 +1,143 @@
...
+        $this->eventDispatcher->dispatch(
+            Argument::type(CacheClearEvent::class),
+            Events::CACHE_CLEAR
+        )->shouldBeCalled();
+
+        $cacheClearer = $this->createCacheClearer(true);
+        $cacheClearer->clear();
+    }
+
+    private function createCacheClearer(bool $withCacheManager = false): CacheClearer
```
danrot, line 126: Why don't we pass the `cacheManager` itself instead of a boolean? I think that `$this->createCacheClearer($cacheManager)` would be clearer than `$this->createCacheClearer(true)`. In that case I would also create the `CacheManager` prophecy only in the tests that require it. Then we have only the stuff that is shared among every test in the `setUp`.

alexander-schranz: Sounds good. Did change the test.

PR #5248, opened 2020-04-21
Invalidate System collections cache in warmUp

### LINKED ISSUE #5245: System media collections are not created when cache is not cleared manually
| Q | A
| --- | ---
| Sulu Version | 2.0.6
| Browser Version | Chrome last 

# Actual Behavior

Hi, i have a strange issue. In a fresh install, System media collections are not created, all media is empty. Then if i try to upload an image in a organization it fails becase it trying to upload the image in collection "6" that not exists. 
It's strange because I recently saw that they were created correctly when trying to upload an image 

# Expected Behavior

System collections must be created when install and maybe protected.

# Steps to Reproduce

Clear db, execute install and inspect media. Later try to add an image to a organization

plozmun:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes  #5245 
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

Invalidate media cache file on warmup.

I have changed test to use PHPUnit without `prophesize`. I think it's more clear . WDT?

#### BC Breaks/Deprecations

No BC Breaks

### DIFF
--- a/src/Sulu/Component/Media/SystemCollections/SystemCollectionManager.php
+++ b/src/Sulu/Component/Media/SystemCollections/SystemCollectionManager.php
@@ -76,6 +76,9 @@ class SystemCollectionManager implements SystemCollectionManagerInterface
 
     public function warmUp()
     {
+        if ($this->cache->isFresh()) {
+            $this->cache->invalidate();
+        }
         $this->getSystemCollections();
     }
 
--- a/src/Sulu/Component/Media/SystemCollections/SystemCollectionManager.php
+++ b/src/Sulu/Component/Media/SystemCollections/SystemCollectionManager.php
@@ -76,9 +76,7 @@ class SystemCollectionManager implements SystemCollectionManagerInterface
 
     public function warmUp()
     {
-        if ($this->cache->isFresh()) {
-            $this->cache->invalidate();
-        }
+        $this->cache->invalidate();
         $this->getSystemCollections();
     }
 
--- a/src/Sulu/Component/Media/Tests/Unit/SystemCollections/SystemCollectionManagerTest.php
+++ b/src/Sulu/Component/Media/Tests/Unit/SystemCollections/SystemCollectionManagerTest.php
@@ -345,17 +345,29 @@ class SystemCollectionManagerTest extends TestCase
         $collectionManager = $this->getCollectionManager($existingCollections, $notExistingCollections);
 
         $entityManager = $this->prophesize(EntityManagerInterface::class);
-        $cache = $this->prophesize(CacheInterface::class);
-        $cache->isFresh()->shouldBeCalled()->willReturn(false);
-        $cache->write($data)->shouldBeCalled();
-        $cache->read()->shouldBeCalled()->willReturn($data);
+        $cache = $this->createMock(CacheInterface::class);
+
+        $cache
+            ->expects($this->exactly(2))
+            ->method('isFresh')
+            ->willReturnOnConsecutiveCalls(false, false);
+
+        $cache
+            ->expects($this->once())
+            ->method('write')
+            ->with($data);
+
+        $cache
+            ->expects($this->once())
+            ->method('read')
+            ->willReturn($data);
 
         $manager = new SystemCollectionManager(
             $config,
             $collectionManager->reveal(),
             $entityManager->reveal(),
             $tokenStorage->reveal(),
-            $cache->reveal(),
+            $cache,
             'en'
         );
 
@@ -371,17 +383,33 @@ class SystemCollectionManagerTest extends TestCase
         $collectionManager = $this->getCollectionManager($existingCollections, $notExistingCollections, false);
 
         $entityManager = $this->prophesize(EntityManagerInterface::class);
-        $cache = $this->prophesize(CacheInterface::class);
-        $cache->isFresh()->shouldBeCalled()->willReturn(true);
-        $cache->write($data)->shouldNotBeCalled();
-        $cache->read()->shouldBeCalled()->willReturn($data);
+        $cache = $this->createMock(CacheInterface::class);
+
+        $cache
+            ->expects($this->exactly(2))
+            ->method('isFresh')
+            ->willReturnOnConsecutiveCalls(true, false);
+
+        $cache
+            ->expects($this->once())
+            ->method('invalidate');
+
+        $cache
+            ->expects($this->once())
+            ->method('write')
+            ->with($data);
+
+        $cache
+            ->expects($this->once())
+            ->method('read')
+            ->willReturn($data);
 
         $manager = new SystemCollectionManager(
             $config,
             $collectionManager->reveal(),
             $entityManager->reveal(),
             $tokenStorage->reveal(),
-            $cache->reveal(),
+            $cache,
             'en'
         );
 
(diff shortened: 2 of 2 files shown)


alexander-schranz:
I don't think this will fix #5245 as the cache is never warmup in sulu:build command. As said there. There should be an own SuluBuilder class which does clear this cache.

plozmun:
The SystemCollectionBuilder is who execute warmUp , here: https://github.com/sulu/sulu/blob/e816ec1651c11cffa627d9e27284d10367d3907f/src/Sulu/Component/Media/SystemCollections/SystemCollectionBuilder.php#L53

plozmun:
I have checked and it remove the file in sulu:build command :)

alexander-schranz:
Ok there seems to be an exist a builder. But the `if fresh` is a little bit strange think it should be always invalidated the cache on warmUp then.

src/Sulu/Component/Media/Tests/Unit/SystemCollections/SystemCollectionManagerTest.php
```diff
@@ -371,17 +383,33 @@ public function testWarmUpFresh($config, $existingCollections, $notExistingColle
         $collectionManager = $this->getCollectionManager($existingCollections, $notExistingCollections, false);
 
         $entityManager = $this->prophesize(EntityManagerInterface::class);
-        $cache = $this->prophesize(CacheInterface::class);
-        $cache->isFresh()->shouldBeCalled()->willReturn(true);
-        $cache->write($data)->shouldNotBeCalled();
-        $cache->read()->shouldBeCalled()->willReturn($data);
+        $cache = $this->createMock(CacheInterface::class);
```
alexander-schranz, line 386: please use `prophesize` to mock the the cache

src/Sulu/Component/Media/Tests/Unit/SystemCollections/SystemCollectionManagerTest.php
```diff
@@ -345,17 +345,29 @@ public function testWarmUpNotFresh($config, $existingCollections, $notExistingCo
         $collectionManager = $this->getCollectionManager($existingCollections, $notExistingCollections);
 
         $entityManager = $this->prophesize(EntityManagerInterface::class);
-        $cache = $this->prophesize(CacheInterface::class);
-        $cache->isFresh()->shouldBeCalled()->willReturn(false);
-        $cache->write($data)->shouldBeCalled();
-        $cache->read()->shouldBeCalled()->willReturn($data);
+        $cache = $this->createMock(CacheInterface::class);
```
alexander-schranz, line 348: use `prophesize` to mock the cache we should be consistent

plozmun:
> Ok there seems to be an exist a builder. But the `if fresh` is a little bit strange think it should be always invalidated the cache on warmUp then.

Is the behavior of DataCache , isFresh only check if file exists before remove otherwise it launch a IOException

src/Sulu/Component/Media/Tests/Unit/SystemCollections/SystemCollectionManagerTest.php
plozmun: Ok, I will change but why use prophesize when is not necessary ?

alexander-schranz: To be consistent in the code.

nnatter:
> Is the behavior of DataCache , isFresh only check if file exists before remove otherwise it launch a IOException

This is a strange behaviour, isnt it? 🙈 Maybe we should add the same check as in the `read()` method? At least I would add a comment why the `isFresh()` call is needed - I wondered about that too 🙂

plozmun:
> > Is the behavior of DataCache , isFresh only check if file exists before remove otherwise it launch a IOException
> 
> This is a strange behaviour, isnt it? 🙈 Maybe we should add the same check as in the `read()` method? At least I would add a comment why the `isFresh()` call is needed - I wondered about that too 🙂

Sorry, my bad, Exception is not launched. I will remove the isFresh check

src/Sulu/Component/Media/SystemCollections/SystemCollectionManager.php
```diff
@@ -76,6 +76,9 @@ public function __construct(
 
     public function warmUp()
     {
+        if ($this->cache->isFresh()) {
```
danrot, line 79: This reads very strange... If the cache is still fresh, there is actually no point of invalidating it, isn't it? But I have seen that the `isFresh` call just checks for the existence of the file... Since it reads so strange I would either simply always call `$this->cache->invalidate()`. If it causes any problems when the file is not existing, I would handle that in the `invalidate` method (just do nothing then). I think that would be a reasonable behavior, since the end result would be the same.

plozmun:
Hi! I have updated PR like suggestions. I left only one test because now the file is always deleted.
IMHO I think DataCache must be replaced by cache.app or removed completely and let the responsability to the CollectionManager

danrot:
Seems like travis thinks you are abusing our build :see_no_evil: 
Screenshot from 2020-04-22 19-51-26
 Tried to amend the latest commit, let's see if it works then.

danrot:
@plozmun Have you never logged in into travis ci? If not, can you try do so? According to [this article in their community that might help](https://travis-ci.community/t/some-pr-are-not-analyzed-with-message-abuse-detected/1191)

plozmun:
> @plozmun Have you never logged in into travis ci? If not, can you try do so? According to [this article in their community that might help](https://travis-ci.community/t/some-pr-are-not-analyzed-with-message-abuse-detected/1191)

Nope, I just logged in

plozmun:
Captura de pantalla 2020-04-22 a las 20 06 29
:scream: :scream: :scream:

alexander-schranz:
@plozmun can you try to contact them?

plozmun:
The warning disappeared when I resynced with github. I hope it doesn't happen again 🤞

alexander-schranz:
@plozmun can you try to repush to se if travis is now triggered correctly:

```
git commit --amend --date="now"
git push yourremote bugfix/warmup-cache-fresh --force
```

plozmun:
Done!

danrot:
It didn't seem to work and now travis is under maintenance... I'll just run the tests locally and merge it. Hopefully it does not break anything, but appveyor seems to work.

plozmun:
Now it works again, but i can's see the build in my dashboard. Is abuse detected again? If so, let me the URL o ID to write a email to support . Thanks

danrot:
Screenshot from 2020-04-23 08-57-26
Looks like it... I guess they won't change that opinion about you in that PR anymore :see_no_evil: Let's hope it works in the next one.

plozmun:
Ok, i will write to travis support because I'm afraid it will happen again. My last two PR in sulu repository had the same problem with travis :(

danrot:
Yeah, but now you are registered there, so maybe that really helps. But contacting them just to be sure might not be the worst idea :slightly_smiling_face:

danrot:
@plozmun Thank you for your contribution! Hopefully tests will also succeed on travis now :see_no_evil:

plozmun:
Thanks to you @danrot

PR #5251, opened 2020-04-21
Allow access to current request in RequestAnalyzer

therouv:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

It is not really a new feature or bug fix rather a small improvment for developers. This PR adds the method `getCurrentRequest` to the `Sulu\Component\Webspace\Analyzer\RequestAnalyzer` as a handy method to access the current request.

#### Why?

In my current project I often need access to the current request and the Sulu attributes of the request. I use the latter for convenience instead of writing the same logic over and over again. But this means that I will have to inject the `RequestStack` in my class as well. I personally like to have as few constructor arguments as possible. So instead of injecting both services, I propose to add the `getCurrentRequest` method to the `RequestAnalyzer` in order to make life a little bit easier for developers :-)

#### BC Breaks/Deprecations

None that I can think of.

### DIFF
--- a/src/Sulu/Component/Webspace/Analyzer/RequestAnalyzer.php
+++ b/src/Sulu/Component/Webspace/Analyzer/RequestAnalyzer.php
@@ -62,9 +62,14 @@ class RequestAnalyzer implements RequestAnalyzerInterface
         }
     }
 
+    public function getCurrentRequest()
+    {
+        return $this->requestStack->getCurrentRequest();
+    }
+
     public function getAttribute($name, $default = null)
     {
-        $request = $this->requestStack->getCurrentRequest();
+        $request = $this->getCurrentRequest();
 
         if (null === $request) {
             return $default;
--- a/src/Sulu/Component/Webspace/Analyzer/RequestAnalyzerInterface.php
+++ b/src/Sulu/Component/Webspace/Analyzer/RequestAnalyzerInterface.php
@@ -141,4 +141,11 @@ interface RequestAnalyzerInterface
      * @param mixed|null $default
      */
     public function getAttribute($name, $default = null);
+
+    /**
+     * Returns the current request.
+     *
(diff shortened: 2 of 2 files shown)


alexander-schranz:
@therouv we do not recommend longer to use the `requestAnalyzer` to access its data. We recommend to inject the `request_stack` service and access from there the current `request` all sulu attributes you will find under `$request->attributes->get('_sulu')`.

I personally would not extend the RequestAnalyzer service. I personally hope that we get ride of it in 3.0.

therouv:
@alexander-schranz Good to know. Is that documented somewhere? 🤔  
Should we rather add a `@deprecated` hint to the class instead?

danrot:
I've got the same opinion as @alexander-schranz. What I am envisioning is that you can access the webspace/portal information in a controller provided by a [`ArgumentResolver`](https://symfony.com/doc/current/controller/argument_value_resolver.html):

```php
class Controller {
    public function index(Portal $portal) {
    }
}
```

Actually that topic is on my mind since years :see_no_evil:

danrot:
Regarding deprecation: Since we are even using it internally, I think it might be a bit early to deprecate it, because there is no solid alternative there yet. @alexander-schranz Or would you say what you've suggested above is the solution? Because if that's the case I say we should really deprecate it, and get rid of all internal calls we have.

alexander-schranz:
@danrot it still needed to set the attributes and call the processors. Maybe more marking the service as `@internal` instead and deprecating the getter functions.

therouv:
Ok! An `@internal` notice would be useful or some information that the service should no longer be used. I will create a custom "RequestAnalyzer" and bundle all my required business logic in there so that I can simplify my code.

Closing this PR.

PR #5256, opened 2020-04-22
Throw meaningful error when value of MediaSelection field does not match expected format

### LINKED ISSUE #5253: Error if API data is not formatted correctly for content type "media_selection
| Q | A
| --- | ---
| Sulu Version | Sulu (2.0.4)
| Browser Version | Firefox 75.0 (64-Bit)

# Actual Behavior

If you create a Content Type "media_selection" with the name "documents" for a custom entity form in the Sulu Admin area, this leads to a Javascript Error as soon as you try to assign a media file.

Bildschirmfoto 2020-04-22 um 10 47 12

# Expected Behavior

Either you get an error message that "documents" is not allowed as name or the selection of media files works normally.

# Steps to Reproduce

`
        <property name="documents" type="media_selection">
            <meta>
                <title>Documents</title>
            </meta>

            <params>
                <param name="types" value="image,video"/>
                <param name="displayOptions" type="collection">
                    <param name="leftTop" value="true"/>
                    <param name="top" value="true"/>
                    <param name="rightTop" value="true"/>
                    <param name="left" value="true"/>
                    <param name="middle" value="false"/>
                    <param name="right" value="true"/>
                    <param name="leftBottom" value="true"/>
                    <param name="bottom" value="true"/>
                    <param name="rightBottom" value="true"/>
                </param>
                <param name="defaultDisplayOption" value="left"/>
            </params>
        </property>
    </properties>
`

# Possible Solutions
Hint: If you change the type to single_media_selection, everything works fine.

nnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #5253
| License | MIT

#### What's in this PR?

This PR adjusts the `MediaSelection` field to throw an error with a meaningful message if the given value does not match the expected format.

#### Why?

Because it is a frequent mistake to pass an array of ids to the field. This leads to a cryptic error message at the moment.

#### How to test?

Adjust the `read()` method of the `MediaSelectionContentType` to match this:
```
    public function read(
        NodeInterface $node,
        PropertyInterface $property,
        $webspaceKey,
        $languageCode,
        $segmentKey
    ) {
        $data = json_decode($node->getPropertyValueWithDefault($property->getName(), '{"ids": []}'), true);
        
        $property->setValue(isset($data['ids']) ? [] : null);
    }
```

Then open the `Create Page` form, assign a media, save the page and see a meaningful error in the console of your browser.

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/Resources/js/containers/Form/fields/MediaSelection.js
+++ b/src/Sulu/Bundle/MediaBundle/Resources/js/containers/Form/fields/MediaSelection.js
@@ -3,7 +3,7 @@ import React from 'react';
 import {observer} from 'mobx-react';
 import type {FieldTypeProps} from 'sulu-admin-bundle/types';
 import userStore from 'sulu-admin-bundle/stores/userStore';
-import {observable} from 'mobx';
+import {isArrayLike, observable} from 'mobx';
 import {
     convertDisplayOptionsFromParams,
     convertMediaTypesFromParams,
@@ -25,8 +25,11 @@ class MediaSelection extends React.Component<FieldTypeProps<Value>> {
             } = {},
         } = schemaOptions;
 
-        if (value !== undefined && value !== null && (typeof value !== 'object' || !Array.isArray(value.ids))) {
-            throw new Error('The "MediaSelection" field expects an object with an "ids" property as value.');
+        if (value !== undefined && value !== null && (typeof value !== 'object' || !isArrayLike(value.ids))) {
+            throw new Error(
+                'The "MediaSelection" field expects an object with an "ids" property and '
+                + 'an optional "displayOption" property as value.'
+            );
         }
 
         if (!defaultDisplayOption) {
--- a/src/Sulu/Bundle/MediaBundle/Resources/js/containers/Form/fields/MediaSelection.js
+++ b/src/Sulu/Bundle/MediaBundle/Resources/js/containers/Form/fields/MediaSelection.js
@@ -25,7 +25,7 @@ class MediaSelection extends React.Component<FieldTypeProps<Value>> {
             } = {},
         } = schemaOptions;
 
-        if (value !== undefined && (typeof value !== 'object' || !Array.isArray(value.ids))) {
+        if (value != null && (typeof value !== 'object' || !Array.isArray(value.ids))) {
             throw new Error('The "MediaSelection" field expects an object with an "ids" property as value.');
         }
 
--- a/src/Sulu/Bundle/MediaBundle/Resources/js/containers/Form/fields/MediaSelection.js
+++ b/src/Sulu/Bundle/MediaBundle/Resources/js/containers/Form/fields/MediaSelection.js
@@ -25,6 +25,10 @@ class MediaSelection extends React.Component<FieldTypeProps<Value>> {
             } = {},
         } = schemaOptions;
 
+        if (value !== undefined && (typeof value !== 'object' || !Array.isArray(value.ids))) {
+            throw new Error('The "MediaSelection" field expects an object with an "ids" property as value.');
+        }
+
         if (!defaultDisplayOption) {
             return;
         }
--- a/src/Sulu/Bundle/MediaBundle/Resources/js/containers/Form/fields/MediaSelection.js
+++ b/src/Sulu/Bundle/MediaBundle/Resources/js/containers/Form/fields/MediaSelection.js
@@ -25,7 +25,7 @@ class MediaSelection extends React.Component<FieldTypeProps<Value>> {
             } = {},
         } = schemaOptions;
 
-        if (value != null && (typeof value !== 'object' || !Array.isArray(value.ids))) {
+        if (value !== undefined && value !== null && (typeof value !== 'object' || !Array.isArray(value.ids))) {
             throw new Error('The "MediaSelection" field expects an object with an "ids" property as value.');
         }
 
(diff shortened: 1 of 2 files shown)


src/Sulu/Bundle/MediaBundle/Content/Types/MediaSelectionContentType.php
```diff
@@ -70,7 +70,7 @@ public function read(
     ) {
         $data = json_decode($node->getPropertyValueWithDefault($property->getName(), '{"ids": []}'), true);
 
-        $property->setValue(isset($data['ids']) ? $data : null);
+        $property->setValue(isset($data['ids']) ? [] : null);
```
alexander-schranz, line 73: ???

alexander-schranz: @nnatter Think you accidently pushed this

nnatter: 🙈

nnatter: yes, this is the code for testing the change - sorry 😇

src/Sulu/Bundle/MediaBundle/Resources/js/containers/Form/fields/MediaSelection.js
```diff
@@ -25,6 +25,10 @@ class MediaSelection extends React.Component<FieldTypeProps<Value>> {
             } = {},
         } = schemaOptions;
 
+        if (value != null && (typeof value !== 'object' || !Array.isArray(value.ids))) {
```
danrot, line 28: Which values are you trying to exclude here? Especially with the first non-typesafe comparison?

Would it maybe be easier to read if we turn that into a set of acceptable values?

```javascript
if (!(value === null || (typeof value === 'object' && Array.isArray(value.ids)) {}
```

I am assuming here the value must be `null` or an object containing an `ids` array.

Resp. your current solution would also work for me if the first check is type strict. But if we are using just `!=` it almost always feels like there is a trap :see_no_evil:

danrot: And isn't there a similar check missing in the `SingleMediaSelection` field type as well?

nnatter: i initially used `value !== undefined` instead of `value != null`, but then flow complains that `Property ids is missing in null or undefined`. This is because the `value` prop is of the type `?T` which can be `null` or `undefined` apparently 🙈

nnatter: flow will not complain when using `value !== null` or your suggestions, but the application will crash if the value is set to `undefined` (which it is when opening the add form). This is because undefined is not strictly equal to null and `typeof undefined` is `undefined`. 

So i am not sure how to write the condition correctly. Should I just check for `null` and for `undefined`? 😅

nnatter: Regarding to the `SingleMediaSelection` - this is not a big problem there because the `SingleMediaSelection` will not crash if the given value has no `id` property. It will just act like nothing is selected. But I can add a similar check there too if you think that would be important

danrot: > flow will not complain when using `value !== null` or your suggestion, but the application will crash if the value is set to `undefined` (which it is when opening the add form). This is because undefined is not strictly equal to null and `typeof undefined` is `"undefined"`.
> 
> So i am not sure how to write the condition correctly. Should I just check for `null` and for `undefined`?

I think checking for both would make the code easier to understand, don't you think so too? This way you have to know how the casting in JS works, which probably nobody knows by heart :see_no_evil: Resp. we could maybe also adjust the type passing that in? Can it be `null` or undefined? If it can't be `null`, we could also change the type from `?T` to `T | undefined`, right?

> Regarding to the `SingleMediaSelection` - this is not a big problem there because the `SingleMediaSelection` will not crash if the given value has no `id` property. It will just act like nothing is selected. But I can add a similar check there too if you think that would be important

Well, if it is working correctly there I don't see any point in doing so.

nnatter: > I think checking for both would make the code easier to understand, don't you think so too? This way you have to know how the casting in JS works, which probably nobody knows by heart 🙈

Thats okay for me 🙂 

> Resp. we could maybe also adjust the type passing that in? Can it be null or undefined? If it can't be null, we could also change the type from ?T to T | undefined, right?

I am a bit hesitant about changing this, because it is part of the `FieldTypeProps` type and therefore would be changed for each of our field types 🙈

danrot:
This is also failing with correct values from the API for me. I think that's because at the spot where you are doing the check an observable array is passed, which is not fulfilling the `Array.isArray` check.

nnatter:
> This is also failing with correct values from the API for me. I think that's because at the spot where you are doing the check an observable array is passed, which is not fulfilling the Array.isArray check.

Good catch! I changed the code to use [isArrayLike](https://mobx.js.org/refguide/api.html#isarraylike). Interestingly, we did not use this anywhere else as far as i saw. DoIs there another solution or didnt we need to do this anywhere else?

danrot:
I didn't know about `isArrayLike` yet :see_no_evil: Maybe it wasn't there when we started with MobX. What we did until now was something like `Array.isArray(value) || isObservableArray(value)`. But your solution definitely is more elegant :slightly_smiling_face:

PR #5257, opened 2020-04-23
Use interface of CategoryTranslation for CategoryTranslationMedia

thomasduenser:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

When overwriting the `CategoryTranslation` Entity, an error is thrown after the execution of `bin/adminconsole doctrine:schema:validate`.

```
[FAIL] The entity-class Sulu\Bundle\CategoryBundle\Entity\CategoryTranslationMedia mapping is invalid:
 * The association Sulu\Bundle\CategoryBundle\Entity\CategoryTranslationMedia#categoryTranslation refers to the inverse side field Sulu\Bundle\CategoryBundle\Entity\CategoryTranslation#medias which does not exist.
```

#### Why?

This PR fixes the `target-entity` of the `CategoryTranslationMedia.orm.xml` file to prevent this error.

### DIFF
--- a/src/Sulu/Bundle/CategoryBundle/Resources/config/doctrine/CategoryTranslationMedia.orm.xml
+++ b/src/Sulu/Bundle/CategoryBundle/Resources/config/doctrine/CategoryTranslationMedia.orm.xml
@@ -14,7 +14,7 @@
             </options>
         </field>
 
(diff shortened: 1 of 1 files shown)


nnatter:
Thanks for the PR! 🙂

PR #5260, opened 2020-04-27
Add deep links for smart content

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | #5219 
| License | MIT
| Documentation PR | sulu/sulu-docs#511

#### What's in this PR?

This PR adds deep links to the smart content. Therefore the configuration returned by different `DataProvider`s had to be extended, so that the smart content knows where the link should go.

#### Why?

Because it is a useful feature.

#### To Do

### DIFF
--- a/config/templates/pages/homepage.xml
+++ b/config/templates/pages/homepage.xml
@@ -42,5 +42,41 @@
                 <title lang="de">Seiten</title>
             </meta>
         </property>
+
+        <property name="snippet" type="smart_content">
+            <meta>
+                <title lang="en">Snippets</title>
+            </meta>
+            <params>
+                <param name="provider" value="snippets" />
+            </params>
+        </property>
+
+        <property name="media" type="smart_content">
+            <meta>
+                <title lang="en">Media</title>
+            </meta>
+            <params>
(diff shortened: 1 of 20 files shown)


alexander-schranz:
@danrot @wachterjohannes I wanted to test this with the article bundle https://github.com/sulu/SuluArticleBundle/pull/467. After getting a route not found errors, I found out that there it  does not work because every type has its own view:

https://github.com/sulu/SuluArticleBundle/blob/146bc852228d3c06020353f9eb292b10cb9796b6/Admin/ArticleAdmin.php#L229

We should find a way to avoid that, else I'm not sure how we can implement this into the ArticleBundle which would be really sad 😢  or do you see any other way?

danrot:
I think we should continue this discussion somewhere in the ArticleBundle... Can you create an issue or another comment in the PR you have already created? I think in here it will just be lost :see_no_evil:

alexander-schranz:
https://github.com/sulu/SuluArticleBundle/issues/468

PR #5263, opened 2020-04-27
Typo in InitializeCommand.php

kfrohwein:
Typo is "confiration"


danrot:
Thanks for the fix! :slightly_smiling_face:

PR #5266, opened 2020-04-28
Add sulu webspace based error controller

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | yes
| Deprecations? | no
| Fixed tickets | fixes -
| License | MIT
| Documentation PR | https://github.com/sulu/sulu-docs/pull/517

#### What's in this PR?

Add sulu webspace based error controller.

#### Why?

The exception controller is deprecated so we need to disable it to be compatible with symfony 5 and decorate the error controller of symfony to render based 

#### To Do

- [x] Create a documentation PR https://github.com/sulu/sulu-docs/pull/517


src/Sulu/Bundle/WebsiteBundle/Controller/ErrorController.php
```diff
@@ -0,0 +1,113 @@
...
+    ) {
+        $this->symfonyErrorController = $symfonyErrorController;
+        $this->templateAttributeResolver = $templateAttributeResolver;
+        $this->twig = $twig;
+        $this->debug = $debug;
+    }
+
+    public function __invoke(Request $request, \Throwable $exception): Response
+    {
+        // call base error controller on debug and when showException is true
```
danrot, line 59: I feel like this comment does not add any value.

src/Sulu/Bundle/WebsiteBundle/Controller/ExceptionController.php
```diff
@@ -60,6 +62,8 @@ public function __construct(
         Environment $twig,
         $debug
     ) {
+        @trigger_error(__CLASS__ . ' is deprecated since version sulu/sulu 2.0 and will be removed in 3.0. Use the ErrorController instead.', E_USER_DEPRECATED);
```
danrot, line 65: Isn't that fully in our control? I guess we don't have this registered for a route anymore, right? Then I would argue we could simply remove that controller, can't we?

src/Sulu/Bundle/WebsiteBundle/DependencyInjection/SuluWebsiteExtension.php
```diff
@@ -29,9 +30,11 @@ class SuluWebsiteExtension extends Extension implements PrependExtensionInterfac
 {
     public function prepend(ContainerBuilder $container)
     {
-        $container->prependExtensionConfig('twig', [
-            'exception_controller' => 'sulu_website.exception.controller:showAction',
-        ]);
+        if ($container->hasExtension('twig') && class_exists(ExceptionController::class)) {
+            $container->prependExtensionConfig('twig', [
+                'exception_controller' => null,
```
danrot, line 35: Does the new way also work with Symfony 4.4? Then this setting would not be necessary at all, would it? What's the point of just setting it to `null`?

src/Sulu/Bundle/WebsiteBundle/Tests/Unit/Sulu/Bundle/WebsiteBundle/Controller/ExceptionControllerTest.php
```diff
@@ -9,11 +9,10 @@
  * with this source code in the file LICENSE.
  */
 
-namespace Sulu\Bundle\WebsiteBundle\Tests\Functional\Controller;
+namespace Sulu\Bundle\WebsiteBundle\Controller;
```
danrot, line 12: Actually we usually set the `Tests\Functional` namespace, because the `Tests` namespace is not autoloaded in production.

src/Sulu/Bundle/WebsiteBundle/DependencyInjection/SuluWebsiteExtension.php
alexander-schranz: @danrot the new way does work for 4.4 but that it works we need to set this to null:

https://github.com/symfony/symfony/blob/cd66cd57a05fac7deaacf45c3b6ae5282e04d344/src/Symfony/Bundle/TwigBundle/DependencyInjection/Configuration.php#L38-L42

src/Sulu/Bundle/WebsiteBundle/Tests/Unit/Sulu/Bundle/WebsiteBundle/Controller/ExceptionControllerTest.php
alexander-schranz: yeah the whole website bundle uses false namespace and strange nested folder. Can refractor that.

src/Sulu/Bundle/WebsiteBundle/Controller/ExceptionController.php
alexander-schranz: Aslong as he did not overwrite it and extend from it and configure its own twig exception controller we could remove it.

danrot: Okay, lets keep it, it shouldn't hurt too much, I guess.

src/Sulu/Bundle/WebsiteBundle/DependencyInjection/SuluWebsiteExtension.php
danrot: Okay, that's weird :see_no_evil:

src/Sulu/Bundle/WebsiteBundle/Tests/Unit/Sulu/Bundle/WebsiteBundle/Controller/ExceptionControllerTest.php
danrot: I am not saying that you have to update everything, but here you are removing the `Tests` namespace, I think that should not be done, but kept as it was.

alexander-schranz: @danrot oh I moved this from Functional to the Unit test namespace as its a Unit Test and not a Functional Test. And as the Unit Test Namespace for the files in there is `namespace Sulu\Bundle\WebsiteBundle\Controller;` I changed it to be consistent.

alexander-schranz: @danrot Did refractor the website bundle test namespace.

PR #5267, opened 2020-04-28
Include all schema properties in userSchema of ListStore

nnatter:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This PR fixes the `userSchema` property of the `ListStore` to include all properties of the schema if there are existing `schemaSettings` for the current user. At the moment, the `userSchema` contains only schema properties that were present when the `schemaSettings` were saved.

#### Why?

Because the `userSchema` property is used for rendering the display options of the list. At the moment, if there are existing `schemaSettings` for a user, the user will not see new schema properties in his UI. This is a common problem during development.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/stores/ListStore.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/stores/ListStore.js
@@ -229,10 +229,6 @@ export default class ListStore {
         }
 
         const schemaSettings = ListStore.getSchemaSetting(this.listKey, this.userSettingsKey) || [];
-        if (!schemaSettings) {
-            return this.schema;
-        }
-
         const userSchema = {};
 
         for (const schemaSettingsEntry of schemaSettings) {
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/stores/ListStore.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/stores/ListStore.js
@@ -229,17 +229,27 @@ export default class ListStore {
         }
 
         const schemaSettings = ListStore.getSchemaSetting(this.listKey, this.userSettingsKey) || [];
+        if (!schemaSettings) {
+            return this.schema;
+        }
 
-        const userSchema = {...this.schema};
+        const userSchema = {};
+
         for (const schemaSettingsEntry of schemaSettings) {
-            if (!userSchema.hasOwnProperty(schemaSettingsEntry.schemaKey)) {
+            if (!this.schema.hasOwnProperty(schemaSettingsEntry.schemaKey)) {
                 continue;
             }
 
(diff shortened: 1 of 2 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/tests/stores/ListStore.test.js
```diff
@@ -266,6 +266,98 @@ test('The user store should be called correctly when changing the schema', () =>
     });
 });
 
+test('The userSchema should include schema properties that are not present in the schemaSetting of the user', () => {
+    const loadingStrategy = new LoadingStrategy();
+    const structureStrategy = new StructureStrategy();
+    const page = observable.box(1);
+    const locale = observable.box();
+    const additionalValue = observable.box(5);
```
danrot, line 274: The `locale` and `additionalValue` do nothing in that test, right? Then I would prefer to not have them in here, because I'd like to keep the tests minimal for whatever they are testing.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/stores/ListStore.js
```diff
@@ -228,20 +228,16 @@ export default class ListStore {
...
 
-        const schemaSettings = ListStore.getSchemaSetting(this.listKey, this.userSettingsKey);
-        const schema = this.schema;
-        if (!schemaSettings) {
-            return schema;
-        }
+        const schemaSettings = ListStore.getSchemaSetting(this.listKey, this.userSettingsKey) || [];
 
-        const userSchema = {};
+        const userSchema = {...this.schema};
```
danrot, line 233: This completely overrides any ordering done by the user, doesn't it? I think we should start with the `schemaSettings` variable from above, and only add what is missing at the end, shouldn't we?

And I just tested, with this change it is not possible to sort the columns at all.

nnatter: did not think about the order at all 🙈 will adjust it

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/stores/ListStore.js
```diff
@@ -228,24 +228,30 @@ export default class ListStore {
             return {};
         }
 
-        const schemaSettings = ListStore.getSchemaSetting(this.listKey, this.userSettingsKey);
-        const schema = this.schema;
+        const schemaSettings = ListStore.getSchemaSetting(this.listKey, this.userSettingsKey) || [];
```
danrot, line 231: Because of the default value used here the below `if` will never be true. We can either keep the default value and ditch the `if` or remove the default value, both options are fine for me.

PR #5270, opened 2020-05-01
Register deprecated exception controller when decorated service exist

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes -
| Related issues/PRs | #5266, part of #4798 
| License | MIT
| Documentation PR | -

#### What's in this PR?

Register deprecated exception controller when decorated service exist.

#### Why?

Else it will fail on symfony 5 when the parent services dont exist in container build.

#### To Do

### DIFF
--- a/src/Sulu/Bundle/WebsiteBundle/DependencyInjection/SuluWebsiteExtension.php
+++ b/src/Sulu/Bundle/WebsiteBundle/DependencyInjection/SuluWebsiteExtension.php
@@ -128,6 +128,10 @@ class SuluWebsiteExtension extends Extension implements PrependExtensionInterfac
             // add alias for default controller
             $container->setAlias(DefaultController::class, 'sulu_website.default_controller')
                 ->setPublic(true);
+
+            if ($container->hasExtension('twig') && class_exists(ExceptionController::class)) {
+                $loader->load('exception_controller.xml');
+            }
         }
     }
 }
--- a/src/Sulu/Bundle/WebsiteBundle/DependencyInjection/SuluWebsiteExtension.php
+++ b/src/Sulu/Bundle/WebsiteBundle/DependencyInjection/SuluWebsiteExtension.php
@@ -129,7 +129,7 @@ class SuluWebsiteExtension extends Extension implements PrependExtensionInterfac
             $container->setAlias(DefaultController::class, 'sulu_website.default_controller')
                 ->setPublic(true);
 
-            if ($container->hasExtension('twig') && class_exists(ExceptionController::class)) {
+            if (class_exists(ExceptionController::class)) {
                 $loader->load('exception_controller.xml');
             }
         }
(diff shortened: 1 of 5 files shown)


src/Sulu/Bundle/WebsiteBundle/Resources/config/exception_controller.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<container xmlns="http://symfony.com/schema/dic/services"
+           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+           xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+
+    <parameters>
+        <parameter key="sulu_website.exception.controller.class">Sulu\Bundle\WebsiteBundle\Controller\ExceptionController</parameter>
```
danrot, line 7: Didn't we stop using parameters like these?

src/Sulu/Bundle/WebsiteBundle/Resources/config/exception_controller.xml
```diff
@@ -0,0 +1,24 @@
...
+    </parameters>
+
+    <services>
+        <!-- exception controller -->
```
danrot, line 11: Doesn't add any value IMO.

src/Sulu/Bundle/WebsiteBundle/Resources/config/exception_controller.xml
alexander-schranz: I just moved it. It did exist before and don't wanted to remove it.

danrot: Ah, sorry, thought for a second we can do BC breaks now, but that's actually not true :sweat_smile:

src/Sulu/Bundle/WebsiteBundle/DependencyInjection/SuluWebsiteExtension.php
```diff
@@ -128,6 +128,10 @@ public function load(array $configs, ContainerBuilder $container)
             // add alias for default controller
             $container->setAlias(DefaultController::class, 'sulu_website.default_controller')
                 ->setPublic(true);
+
+            if ($container->hasExtension('twig') && class_exists(ExceptionController::class)) {
```
danrot, line 132: `hasExtension` returns false in my case... Might that be because we are relying on the order of the bundles here? Do we need this check at all? The `ExceptionController` is only available if twig is installed, isn't it? AFAIK the only edge case would be if somebody installed twig, but did not register it in the kernel, right?

PR #5273, opened 2020-05-05
Remove the path attribute from twig templates

### LINKED ISSUE #4128: Return full phpcr path instead of relative to webspace node
| Q | A
| --- | ---
| Bug? | no
| New Feature? | no
| Sulu Version | for 2.0

#### Actual Behavior

Actually the `path` in twig return the relative path to the webspace node. This mostly make the problem that the `path` is mistaken as `url`.

#### Expected Behavior

If we return the full phpcr path this will not longer be mistaken as url and the developer can better work with the phpcr path.

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #4128
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

Remove the path attribute from twig templates.

#### Why?

The `path` get mostly be confused by the `url` and so should be removed from the twig side. This effects the StructureResolver (page data) and the ContentMapper (smart content, page selection, navigation). 

#### Example Usage

~~~php
sulu_website:
    twig:
        attributes:
            path: false
~~~

#### To Do

### DIFF
--- a/src/Sulu/Bundle/WebsiteBundle/DependencyInjection/Configuration.php
+++ b/src/Sulu/Bundle/WebsiteBundle/DependencyInjection/Configuration.php
@@ -49,7 +49,7 @@ class Configuration implements ConfigurationInterface
                                 ->beforeNormalization()
                                     ->ifTrue(function($v) { return false !== $v; })
                                     ->then(function($v) {
-                                        @\trigger_error('Enable the urls parameter is deprecated since sulu/sulu 2.2.', \E_USER_DEPRECATED);
+                                        @\trigger_error('Enabling the "urls" parameter is deprecated since sulu/sulu 2.2.', \E_USER_DEPRECATED);
 
                                         return $v;
                                     })
@@ -60,7 +60,7 @@ class Configuration implements ConfigurationInterface
                                 ->beforeNormalization()
                                     ->ifTrue(function($v) { return false !== $v; })
                                     ->then(function($v) {
-                                        @trigger_error('Enable the path parameter is deprecated since sulu/sulu 2.1.', E_USER_DEPRECATED);
+                                        @\trigger_error('Enabling the "path" parameter is deprecated since sulu/sulu 2.3.', \E_USER_DEPRECATED);
 
                                         return $v;
                                     })
--- a/src/Sulu/Bundle/WebsiteBundle/DependencyInjection/Configuration.php
+++ b/src/Sulu/Bundle/WebsiteBundle/DependencyInjection/Configuration.php
@@ -55,6 +55,17 @@ class Configuration implements ConfigurationInterface
                                     })
                                 ->end()
                             ->end()
+                            ->booleanNode('path')
+                                ->defaultTrue()
+                                ->beforeNormalization()
+                                    ->ifTrue(function($v) { return false !== $v; })
+                                    ->then(function($v) {
+                                        @trigger_error('Enable the path parameter is deprecated since sulu/sulu 2.1.', E_USER_DEPRECATED);
+
+                                        return $v;
+                                    })
+                                ->end()
+                            ->end()
                         ->end()
                     ->end()
                     ->arrayNode('navigation')
--- a/src/Sulu/Bundle/WebsiteBundle/Navigation/NavigationMapper.php
+++ b/src/Sulu/Bundle/WebsiteBundle/Navigation/NavigationMapper.php
@@ -77,7 +77,7 @@ class NavigationMapper implements NavigationMapperInterface
         $this->enabledTwigAttributes = $enabledTwigAttributes;
 
         if ($enabledTwigAttributes['path']) {
-            @trigger_error('Enabling the "path" parameter is deprecated since sulu/sulu 2.3.', E_USER_DEPRECATED);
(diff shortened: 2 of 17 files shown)


src/Sulu/Bundle/WebsiteBundle/DependencyInjection/Configuration.php
```diff
@@ -30,6 +30,22 @@ public function getConfigTreeBuilder()
             ->arrayNode('analytics')
                 ->canBeDisabled()
             ->end()
+            ->arrayNode('attributes')
```
danrot, line 33: Are these attributes only for twig templates? Asking because there is already a `twig` node, to which it would better suite then IMO.

src/Sulu/Bundle/WebsiteBundle/Navigation/NavigationMapper.php
```diff
@@ -239,4 +229,17 @@ public function inNavigation(StructureInterface $content, $context = null)
         // do not show
         return false;
     }
+
+    private function normalizeResult(array $result)
+    {
+        foreach ($result as &$item) {
```
danrot, line 235: Since you are returning the new array and reassigning it above anyway, wouldn't it be less error-prone to return a new copy here, instead of altering the existing references?

src/Sulu/Bundle/WebsiteBundle/Resolver/StructureResolver.php
```diff
@@ -35,12 +36,24 @@ class StructureResolver implements StructureResolverInterface
...
+    /**
+     * @var bool
+     */
+    private $enabledPathAttribute = true;
+
     public function __construct(
         ContentTypeManagerInterface $contentTypeManager,
-        ExtensionManagerInterface $structureManager
+        ExtensionManagerInterface $structureManager,
+        bool $enabledPathAttribute = true
```
danrot, line 47: When reading that code, shouldn't we make that already an array like this:
```php
['path' => true]
```

What if we come up with more attributes we want to get rid of? That would result in quite a massive constructor.

src/Sulu/Component/Content/Compat/Structure/StructureBridge.php
```diff
@@ -449,6 +450,18 @@ public function getNodeName()
         return $this->getDocument()->getTitle();
     }
 
+    public function getNodePath(): ?string
+    {
+        if ($this->document instanceof RedirectTypeBehavior &&
+            RedirectType::INTERNAL == $this->document->getRedirectType() &&
+            null !== $this->document->getRedirectTarget()
+        ) {
+            return $this->getDocument()->getRedirectTarget()->getPath();
```
danrot, line 459: Currently the `path` attribute did not resolve the redirect, did it? If we are going to introduce that one as a replacement for the old `path`, shouldn't it behave the same?

src/Sulu/Component/Content/Mapper/ContentMapper.php
```diff
@@ -144,7 +149,8 @@ public function __construct(
         SessionManagerInterface $sessionManager,
         EventDispatcherInterface $eventDispatcher,
         ResourceLocatorStrategyPoolInterface $resourceLocatorStrategyPool,
-        NamespaceRegistry $namespaceRegistry
+        NamespaceRegistry $namespaceRegistry,
+        bool $enablePathAttribute = true
```
danrot, line 153: Same question as above, shouldn't we make that an array in case more attributes will be passed in the future?

src/Sulu/Component/Content/Compat/Structure/StructureBridge.php
danrot: And btw, in the [`ContentMapper`](https://github.com/sulu/sulu/pull/5273/files#diff-2a74d79cea8c966c3ad6a835206c6c3bR733) it also does not seem to be resolved, am I right?

src/Sulu/Component/Content/SmartContent/PageDataProvider.php
```diff
@@ -167,7 +167,19 @@ public function resolveDatasource($datasource, array $propertyParameter, array $
...
+            if (isset($result[0]['path'])) {
+                $path = '/' . ltrim($result[0]['path'], '/');
+            } elseif (isset($result[0]['nodePath'])) {
+                $pathParts = explode('/', $result[0]['nodePath']);
+                $pathParts = array_slice($pathParts, 4);
+                $path = '/' . implode('/', $pathParts);
+            }
+        }
+
+        return new DatasourceItem($result[0]['id'], $result[0]['title'], $path);
```
danrot, line 182: Didn't we make the mistake we were trying to avoid here on our own? Shouldn't we be passing the URL after all? :thinking: If you think about it, the `path` is nothing the user can handle, especially when he is not managing the UI in a different language than the default language...

danrot:
Generally I have seen that you were checking first for `path` and fallback to `nodePath` on a few occasions. Why is that? If we are going to introduce the `nodePath` (have alreayd questioned if we should do that in a separate comment) we could simply always use that value, can't we?

nnatter:
I adjusted this PR to match the implementation in https://github.com/sulu/sulu/pull/5361

nnatter:
Fixes #4128

PR #5275, opened 2020-05-06
Fix partial portal matching for url for specified language

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

Fix partial portal matching for url for specified language.

#### Why?

When have the following portal:

```xml
<url language="en" country="uk">uk.mydomain.com/{language}</url>
<url language="de" country="at">austria.mydomain.com/{language}</url>
```

The partial portal mapping does not work this means `/`is not redirected to `/en` and `/sitemap.xml` will not work.

In the project they had the following url `austria.mydomai.com/de` but as we can not just replace `de` i suggested to use `{language}` placeholder but that doesn't currently work so it need to replaced correctly also there and that the partial map work the logic need to be moved out of if else.

#### Example Usage

~~~xml
<url language="en" country="uk">uk.mydomain.com/{language}</url>
<url language="de" country="at">austria.mydomain.com/{language}</url>
~~~

#### To Do

### DIFF
--- a/src/Sulu/Component/Webspace/Manager/WebspaceCollectionBuilder.php
+++ b/src/Sulu/Component/Webspace/Manager/WebspaceCollectionBuilder.php
@@ -330,11 +330,17 @@ class WebspaceCollectionBuilder
             $country = $url->getCountry();
             $locale = $language . ($country ? '_' . $country : '');
 
+            $replacers = [
+                ReplacerInterface::REPLACER_LANGUAGE => $language,
+                ReplacerInterface::REPLACER_COUNTRY => $country,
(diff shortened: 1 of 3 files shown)


alexander-schranz:
ping @adriansieber can you test if this fix your issue with the sitemap when using `{language}` instead of `/de` in your webspace configurations.

C-burr:
@alexander-schranz this fix seems to work!

PR #5277, opened 2020-05-06
Allow to switch between different segments on the website

### LINKED ISSUE #5271: Allow to segment entire webspace into different URLs
# Problem description

It would be nice if the entire webspace could be segmented into two different "versions" of the website. That is useful especially for seasonal websites; many tourism websites for e.g. skiing areas have dedicated pages for winter and summer, because they offer different leisure activities based on the time of the year.

# Proposed solution

This feature is similar to the already existing AudienceTargeting feature, the main exception being that the current segment is not chosen by some rules, but solely determined by the URL. This means that the following features are required:

- [x] Recognize the segment from cookie
- [x] Allow to assign segments in the excerpt tab of pages
- [x] Smart Content must not return pages not being part of the segment passed via the URL
- [x] Allow to assign segments to page blocks
- [x] Only provide the blocks to the template that match the current segment
- [x] ~~Redirect URLs without segments to the default segment~~ (No redirect necessary because cookies are used)
- [x] Add a dropdown in the preview toolbar to watch the website in different segments
- [x] ~~Consider the segment when generating URLs for the sitemap~~ (Not necessary because URL stays the same, since cookies are used now)
- [x] ~~Correctly generate URLs including the segment (`sulu_content_path`, navigation, page selection)~~ (Since cookies are used the URLs are always the same)
- [x] Add parameter to navigation twig extensions for segments
- [ ] Cache pages with segments correclty using the Cookie

# Open questions

- [x] How to determine which segment is the default (for the skiing area case mentioned above that should probably change based on the time of the year)
-> see https://github.com/sulu/sulu/issues/5271#issuecomment-628474230, there is only the default in the XML, has to be changed manually
- [ ] Should there be a switch in the Smart Content configuration to tell if the segment should be considered?
- [x] What about blocks that have no segments assigned? Are they visible for all segments or for none of them?
-> Will be visible in all segments
- [x] ~~Should it also be possible to throw 404s when the passed segment in the URL does not exist?~~
-> segments are not part of the URLs for now
- [x] ~~Should we maybe include the segment in the URL only if the segment changes the content? Would avoid duplicate content issues, but requires us to remember the current segment in a different way, e.g. in a cookie. Alternatively we could always have the segment in the URL, but would have to be careful when setting canonical links. In that case it might also be problematic if the default segment changes every 6 months and therefore also the canonical tag changes.~~
-> We go even a step further, and never pass the segment in the URL. A cookie will be used instead. That has the adavantage that we don't have to worry about duplicate content at all.

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #5271
| Related issues/PRs | ---
| License | MIT

#### What's in this PR?

This PR will allow to specify different segments for each webspace, and assign pages and blocks to these segments. Then these segments can be switched on the frontend using cookies and certain pages will automatically set the cookie, e.g. if they only exist in one segment.

The use case for this is e.g. a winter/summer segmentation for tourism websites.

#### Example Usage

~~~php
// If you added new features, show examples of how to use them here
// (remove this section if not a new feature)

$foo = new Foo();

// Now we can do
$foo->doSomething();
~~~

#### BC Breaks/Deprecations

#### To Do

- [x] `StructureXmlLoader` should add default settings form to `block` field types
- [x] Segments should be translatable to show correct segment e.g. in settings tab, segment chooser on website, preview, ...
- [x] Rename `settingsForm` variables in `FieldBlocks` to `settingsFormKey`
- [x] Make `settings` name in blocks a reserved keyword (done in #5367)
- [x] If no snippets at all are created, the smart content using the `SnippetDataProvider` does not work (fixed in #5323)
- [x] Account and Contact smart content is not working (fixed in #5294)
- [x] ~~Should other `segment` occurences not needed any more (`Replacer`, `SessionManager`, `ContentMapper`, `PhpcrMapper`, `PortalInformation`) also be removed? Might lead to confusing situations in the future.~~ (Not for now)
- [x] Should be a new `release/2.1` branch be created before merging this?
- [x] ~~Should the cookie be webspace specific (set `path` for current webspace when setting cookie if existing (how to get URL without language?))~~ (hard to implement, won't do that for now, since it is an edge case)
- [x] How to avoid having a cache entry without cookie if there is a default cookie?
- [x] Introduce `SegmentStore` acting like `TargetGroupStore` (can also recognize if `getSegment` function was called, and allows to check if setting the `Vary` header is necessary) or at least set `Vary` header only if current webspace has segments
- [x] Check combination with Caching in AudienceTargeting
- [x] Pass current segment in twig template
- [x] Change segment if page does not exist in current segment
- [x] Adjust design of segments in block settings overlay
- [x] Add missing translations in block settings overlay for segments
- [x] Deprecate `urls` in twig template by introducing a new configuration option (Done in #5361)
- [x] ~~Filter for segment in teaser selection~~ (Neither selections nor teaser should consider segments, because the user should not assign the items if it should not show up)
- [x] Separate field type for segments to also display it for multiple webspace in articles
    - [x] Extract first commit of this PR into separate PR, because it is useful, but not necessary for this PR anymore (#5366)
    - [x] Remove `SegmentHelper` and its occurences in XML files
- [x] ~~Add segements to snippets~~ (Not required at the moment, therefore postponing until there is a need for it)


src/Sulu/Bundle/PageBundle/Admin/PageAdmin.php
```diff
@@ -271,6 +271,7 @@ public function configureViews(ViewCollection $viewCollection): void
                     ->setTabPriority(512)
                     ->addToolbarActions($formToolbarActionsWithoutType)
                     ->addRouterAttributesToFormRequest($routerAttributesToFormRequest)
+                    ->addRouterAttributesToFormMetadata($routerAttributesToFormMetadata)
```
danrot, line 274: Move to Excerpt tab

src/Sulu/Bundle/PageBundle/Document/BasePageDocument.php
```diff
@@ -509,4 +516,14 @@ public function setAuthor($author)
     {
         $this->author = $author;
     }
+
+    public function setSegment(?string $segment = null)
```
danrot, line 520: Neither these functions nor a separate behavior would be needed when moved to excerpt tab.

src/Sulu/Bundle/PageBundle/Form/Type/BasePageDocumentType.php
```diff
@@ -48,6 +48,7 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                 'allow_delete' => true,
             ]
         );
+        $builder->add('segment', TextType::class);
```
danrot, line 51: Would be handled in excerpt extension instead of here if moved to excerpt.

src/Sulu/Bundle/PageBundle/Resources/config/serializer/Document.BasePageDocument.xml
```diff
@@ -12,6 +12,7 @@
         <property name="changed" type="DateTime" groups="defaultPage,preview"/>
         <property name="resourceSegment" exclude="true"/>
         <property name="navigationContexts" type="array" serialized-name="navContexts" groups="defaultPage,smallPage"/>
+        <property name="segment" type="string" serialized-name="segment" groups="defaultPage,smallPage"/>
```
danrot, line 15: Would be serialized to the `ext` property if moved.

src/Sulu/Bundle/SnippetBundle/Form/SnippetType.php
```diff
@@ -39,5 +39,6 @@ public function buildForm(FormBuilderInterface $builder, array $options)
         $builder->add('resourceSegment', TextType::class, ['mapped' => false]);
         $builder->add('navigationContexts', TextType::class, ['mapped' => false]);
         $builder->add('shadowLocaleEnabled', TextType::class, ['mapped' => false]);
+        $builder->add('segment', TextType::class, ['mapped' => false]);
```
danrot, line 42: Would not be needed if not passed as root prop but as excerpt instead.

src/Sulu/Component/Content/Compat/DataNormalizer.php
```diff
@@ -61,6 +61,7 @@ public static function normalize(FormEvent $event)
             'redirectTarget' => self::getAndUnsetValue($data, 'internal_link'),
             'redirectExternal' => self::getAndUnsetValue($data, 'external'),
             'navigationContexts' => self::getAndUnsetValue($data, 'navContexts'),
+            'segment' => self::getAndUnsetValue($data, 'segment'),
```
danrot, line 64: No need for this change if move to excerpt.

src/Sulu/Component/Content/Compat/DataNormalizer.php
```diff
@@ -72,7 +73,7 @@ public static function normalize(FormEvent $event)
         ];
 
         foreach ($normalized as $key => $value) {
-            if (null === $value) {
+            if (null === $value && $key !== 'segment') {
```
danrot, line 76: No need for this change if move to excerpt.

src/Sulu/Component/Content/Document/Behavior/SegmentBehavior.php
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+namespace Sulu\Component\Content\Document\Behavior;
+
+/**
+ * The implementing document can have a single segment applied to it.
+ *
+ * The segment indicates in which segment the implementing document will appear in.
+ */
+interface SegmentBehavior
```
danrot, line 19: No need for this change if move to excerpt.

src/Sulu/Component/Content/Document/Subscriber/SegmentSubscriber.php
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+namespace Sulu\Component\Content\Document\Subscriber;
+
+use Sulu\Component\Content\Document\Behavior\SegmentBehavior;
+use Sulu\Component\DocumentManager\Event\MetadataLoadEvent;
+use Sulu\Component\DocumentManager\Events;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+class SegmentSubscriber implements EventSubscriberInterface
```
danrot, line 19: No need for this change if move to excerpt.

src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/Preview.js
```diff
@@ -294,6 +317,19 @@ class Preview extends React.Component<Props> {
                                     value={this.previewStore && this.previewStore.targetGroup}
                                 />
                             }
+                            {this.segments.length > 0 &&
```
danrot, line 320: That part is still something I am not completely happy with. The `Preview` already has to know about `webspaces`, `segments` and `targetGroups`, and it is not really extensible. I still think we should make use of the `ToolbarConfigStore` here, that one would be easier to extend than the current approach.

src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/Preview.js
```diff
@@ -45,6 +45,22 @@ class Preview extends React.Component<Props> {
...
+    @computed get webspaceKey() {
+        const {
+            router: {
+                attributes: {
+                    webspace,
+                },
+            },
+        } = this.props;
+
+        return webspace;
```
alexander-schranz, line 57: this is in case of the SuluArticleBundle undefined and does currently fail.

Beneath (Line 75) we did fallback to the first available webspace

>             this.webspaceKey || this.webspaceOptions[0].value,

src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/Preview.js
```diff
@@ -45,6 +45,22 @@ class Preview extends React.Component<Props> {
     typeDisposer: () => mixed;
     dataDisposer: () => mixed;
 
+    @computed get webspaceKey() {
+        const {
+            router: {
+                attributes: {
+                    webspace,
```
alexander-schranz, line 52: can we define a type for this `?string` destructering here, so flow would fail when try to call function which do not accept null or undefined

danrot: That's a good idea, will fix that. Will probably include some other changes as well, therefore going to do that in a separate PR in release/2.0.

src/Sulu/Bundle/WebsiteBundle/Tests/Unit/Resolver/RequestAnalyzerResolverTest.php
```diff
@@ -112,20 +116,23 @@ public function testResolve()
...
                     'webspaceName' => 'Sulu',
+                    'segmentKey' => 's',
                     'portalKey' => 'sulu_io_portal',
                     'portalName' => 'Sulu Portal',
                     'defaultLocale' => 'de',
                     'portalUrl' => 'sulu.io/de',
                     'resourceLocatorPrefix' => '/de',
                     'resourceLocator' => '/search',
                 ],
+                'segmentKey' => 's',
```
danrot, line 135: Not totally sure about this one... The `webspaceKey` is also available under the `request` key and on the top level in the twig template... But that's because the `webspaceKey` is returned by the `RequestAnalyzerResolver` and the `StructureResolver`. But the `StructureResolver` does not know about the segment, so I can't add it there as well...

Quehnie:
Error when used with ArticleBundle:

Variable "webspace" is not valid around position 54 for expression `service('sulu_page.segment_select_helper').getValues(webspace)`.

src/Sulu/Bundle/PageBundle/Resources/config/forms/page_excerpt.xml
```diff
@@ -36,6 +36,21 @@
...
+            visibleCondition="__webspace.segments.length &gt; 0"
+        >
+            <meta>
+                <title>sulu_admin.segments</title>
+            </meta>
+            <params>
+                <param
+                    name="values"
+                    type="expression"
+                    value="service('sulu_page.segment_select_helper').getValues(webspace)"
```
danrot, line 50: This line is the problem of [@Quehnie's comment in the ArticleBundle](https://github.com/sulu/sulu/pull/5277#issuecomment-648222022).

I was not aware that this exact form is used in the ArticleBundle. The problem here now is, that they actually need this feature also in the ArticleBundle, but articles don't have a clearly assigned webspace... How should we handle this?

Three possibilities come to mind:

1. Ditch the segments from the webspace and make that a global setting. Would probably help with the project they are currently working on, but might be problematic in other projects.
2. Use the default webspace from the article. This would still be problematic when the article should be shown in multiple webspaces, where each webspace contains segments.
3. Show a separate segment dropdown for each webspace in the excerpt tab of the articles.

danrot: Have discussed that with @chirimoya, and we will implement option 3 by adding a new field type, which automatically recognizes if a single webspace has been assigned or if the segments for all webspaces should be shown.

src/Sulu/Bundle/WebsiteBundle/Tests/Unit/Resolver/RequestAnalyzerResolverTest.php
wachterjohannes: is that evertime the same? if yes why not removing one of them - if not we need another name - like `currentSegmentKey` and `documentSegmentKey` or something like this

src/Sulu/Component/Webspace/Manager/WebspaceCollectionBuilder.php
```diff
@@ -323,14 +292,13 @@ private function buildUrls(
...
                 $environment,
-                $segments,
                 $replacers,
                 $urlAddress,
                 $portal->getLocalization($locale),
                 $url
             );
         } else {
-            // create all the urls for every localization/segment combination
+            // create all the urls for every localizationcombination
```
wachterjohannes, line 301: `localization combination `

src/Sulu/Component/Content/Tests/Unit/Block/BlockContentTypeTest.php
```diff
@@ -752,7 +765,7 @@ public function testGetContentDataWithEmptySettings()
                         'settings' => new \stdClass(),
                     ],
                 ],
-                'settings' => [],
+                'settings' => new \stdClass(),
```
wachterjohannes, line 768: why `stdClass`?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
```diff
@@ -97,15 +100,32 @@ class FieldBlocks extends React.Component<FieldTypeProps<Array<BlockEntry>>> {
...
+        const iconMappingReducerCreator = (prefixSchemaKey: string = '') => (iconsMapping, schemaKey) => {
+            const pointer = '/' + prefixSchemaKey + schemaKey;
+
+            if (!jsonpointer.has(settingsSchema, pointer)) {
+                return iconsMapping;
+            }
+
+            const schemaEntry = jsonpointer.get(settingsSchema, pointer);
+
+            if (!schemaEntry.tags && schemaEntry.items) {
```
nnatter, line 112: cant a `schemaEntry` have an unrelated tag but also further items?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/AbstractFormStore.js
```diff
@@ -190,6 +200,10 @@ export default class AbstractFormStore
         this.updateFieldPathEvaluations();
     }
 
+    setMultiple() {
```
nnatter, line 203: this feels a bit strange.. why not just call `updateFieldPathEvaluations()` from the child class?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/SchemaFormStoreDecorator.js
```diff
@@ -6,10 +6,14 @@ import metadataStore from './metadataStore';
 export default class SchemaFormStoreDecorator implements FormStoreInterface {
     @observable innerFormStore: ?FormStoreInterface;
 
-    constructor(initializer: (schema: RawSchema, jsonSchema: Object) => FormStoreInterface, formKey: string) {
+    constructor(
+        initializer: (schema: RawSchema, jsonSchema: Object) => FormStoreInterface,
+        formKey: string,
```
nnatter, line 11: shouldnt we add a `type` param now? later on we need to use another parameter order than the `metadataStore `

src/Sulu/Bundle/WebsiteBundle/Tests/Unit/Resolver/RequestAnalyzerResolverTest.php
danrot: I decided to put it in both places, because personally I would expect the `segmentKey` at the same place where I get the `webspaceKey`, since they are related. And since the `webspaceKey` is in two different places, I've also put the segment in two different places.

Prefixing `segmentKey` with `current` or `document` does not make a lot of sense to me, because none of the other twig variables have any of these prefixes.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/MemoryFormStore.js
```diff
@@ -41,6 +43,8 @@ export default class MemoryFormStore extends AbstractFormStore implements FormSt
 
     @action setMultiple(data: Object) {
         this.data = data;
+
+        super.setMultiple();
```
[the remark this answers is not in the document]
danrot: Because currently the parent class does not have any use for the data. And introducing it would make eslint fail, which would mean we would have to ignore that error. Instead I prefer to not pass that information, since the abstract super class can't set the data anyway.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js
```diff
@@ -143,6 +143,8 @@ export default class ResourceFormStore extends AbstractFormStore implements Form
 
     setMultiple(data: Object) {
         this.resourceStore.setMultiple(data);
+
+        super.setMultiple();
```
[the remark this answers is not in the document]
danrot: See comment above.

src/Sulu/Bundle/HttpCacheBundle/Resources/varnish/sulu.vcl
```diff
@@ -48,6 +48,10 @@ sub sulu_recv {
     if (req.http.Authorization) {
         return (pass);
     }
+
+    if (req.http.Cookie ~ "_ss") {
```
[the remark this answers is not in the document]
danrot: Not necessary, since the documentation includes exactly that file anyway: https://docs.sulu.io/en/2.0/cookbook/caching-with-varnish.html?highlight=varnish#varnish-configuration

src/Sulu/Bundle/PageBundle/Resources/config/forms/page_block_settings.xml
```diff
@@ -25,5 +25,31 @@
...
+                            </meta>
+                        </param>
+                        <param name="skin" value="heading" />
+                        <param name="type" value="toggler" />
+                    </params>
+
+                    <tag name="sulu.block_setting_icon" icon="su-focus" />
+                </property>
+
+                <property name="segment" type="segment_select" visibleCondition="segment_enabled == true" />
```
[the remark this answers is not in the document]
danrot: No, that's how it was intended in the design. This property belongs visually to the headings defined in the previous property.

src/Sulu/Bundle/PageBundle/Resources/js/containers/Form/fields/SegmentSelect.js
```diff
@@ -0,0 +1,43 @@
+// @flow
+import React from 'react';
+import {computed} from 'mobx';
+import {SingleSelect} from 'sulu-admin-bundle/components';
+import type {FieldTypeProps} from 'sulu-admin-bundle/types';
+import {translate} from 'sulu-admin-bundle/utils';
+import webspaceStore from '../../../stores/webspaceStore';
+
+export default class SegmentSelect extends React.Component<FieldTypeProps<string>> {
```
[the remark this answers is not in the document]
danrot: This has not been implementd like this on purpose. The thing is that the `page_excerpt` form makes use of this `SegmentSelect`, and this form is also used in the ArticleBundle. And in the case of the ArticleBundle it has to be possible to assign a segment per webspace. Since the same form is used in combination with different entities, this field has to implement this logic. If a webspace is set, it should only show a `SingleSelect`, if no webspace is set, it should show a `SingleSelect` for each available webspace.

This has just not been implemented in this PR, because it is so big already. Therefore this field type has to be changed when the ArticleBundle integrates Segments, but it should not be renamed then.

src/Sulu/Bundle/WebsiteBundle/Controller/SegmentController.php
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    public function switchAction(Request $request)
+    {
+        $webspace = $this->requestAnalyzer->getWebspace();
+        $segments = $webspace->getSegments();
+        $defaultSegment = $webspace->getDefaultSegment();
+
+        $url = $request->query->get('url');
+        $segmentKey = $request->query->get('segment');
```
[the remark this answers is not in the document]
danrot: We can, although it isn't that bad, because it will be reset to `null` if the segment does not exist in another place. Will do it anyway, just to be sure.

src/Sulu/Bundle/WebsiteBundle/Routing/ContentRouteProvider.php
```diff
@@ -170,6 +178,15 @@ public function getRouteCollectionForRequest(Request $request)
             } elseif (!$this->checkResourceLocator($resourceLocator, $prefix)) {
                 return $collection;
             } else {
+                if ($document instanceof ExtensionBehavior) {
+                    $documentSegmentKey = $document->getExtensionsData()['excerpt']['segment'];
+                    $segment = $this->requestAnalyzer->getSegment();
+
+                    if ($segment && $segment->getKey() !== $documentSegmentKey) {
+                        $this->requestAnalyzer->changeSegment($documentSegmentKey);
```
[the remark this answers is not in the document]
danrot: It probably has, but since the ArticleBundle will end up in a separate PR anyway, I would add that code in the other PR.

src/Sulu/Component/Content/Tests/Unit/Block/BlockContentTypeTest.php
danrot: Because an array with string keys is used here, which is serialized to a JSON object. And the problem with PHP is, that an empty array is serialized to a JSON array instead of object, leading to some weird inconsistencies.

src/Sulu/Bundle/PageBundle/Tests/Functional/Controller/PageControllerTest.php
```diff
@@ -1883,6 +1883,45 @@ public function testNavContexts()
         $this->assertFalse($response['hasSub']);
     }
 
+    public function testSegment()
```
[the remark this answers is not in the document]
danrot: Good catch! Will fix it, it should set a segment.

src/Sulu/Bundle/WebsiteBundle/Resolver/RequestAnalyzerResolver.php
```diff
@@ -41,17 +41,22 @@ public function resolve(RequestAnalyzerInterface $requestAnalyzer)
...
                 'webspaceName' => $requestAnalyzer->getWebspace()->getName(),
+                'segmentKey' => $segmentKey,
                 'portalKey' => $requestAnalyzer->getPortal()->getKey(),
                 'portalName' => $requestAnalyzer->getPortal()->getName(),
                 'defaultLocale' => $defaultLocale,
                 'portalUrl' => $requestAnalyzer->getPortalUrl(),
                 'resourceLocatorPrefix' => $requestAnalyzer->getResourceLocatorPrefix(),
                 'resourceLocator' => $requestAnalyzer->getResourceLocator(),
             ],
+            'segmentKey' => $segmentKey,
```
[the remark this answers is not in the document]
danrot: See https://github.com/sulu/sulu/pull/5277#discussion_r453568473 - but it is open for discussion.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
danrot: This code is there for `section`s, which never have a `tag`, but I get that this is a bit confusing... I am not sure anymore, but if I have another look I think the `items` property only exists on `sections` anyway, so I should be able to just check for its existence... E.g. something like:

```javascript
if (schemaEntry.items) {
    // ...
}
```

I'll try it, and see what happens.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/AbstractFormStore.js
danrot: I prefer this over calling `updateFieldPathEvaluations` in the child classes... There is already the `MemoryFormStore` and the `ResourceFormStore`, so we would have to do that at two different places. Which gets even worse, if we realize that there is more code we want to run when multiple properties are changed.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/SchemaFormStoreDecorator.js
danrot: Good catch, will add it.

danrot:
Stuff missing for the ArticleBundle:
- Implement SegmentSelect to also support multiple segments per webspace
- Call `RequestAnalyzerInterface::changeSegment` also in ArticleBundle

src/Sulu/Bundle/PageBundle/Resources/js/containers/Form/fields/SegmentSelect.js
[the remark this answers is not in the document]
nnatter: sounds reasonable 🙂

src/Sulu/Bundle/WebsiteBundle/Tests/Unit/Resolver/RequestAnalyzerResolverTest.php
nnatter: think i would only add it to the `request` array. two places for the same information feel strange to me. we cant fix it for the `webspaceKey` now, but we also dont need to repeat it 🙂

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
nnatter: makes sense 🙂

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/AbstractFormStore.js
nnatter: alright, dont have a strong opinion 🙂

danrot:
@nnatter Removed that variable once. Should be fine now, the only thing missing is the documentation :see_no_evil: But I think currently I won't make it, has to wait a bit, but hopefully before the release.

src/Sulu/Component/Webspace/Analyzer/RequestAnalyzer.php
```diff
@@ -97,6 +117,16 @@ public function getSegment()
         return $this->getAttribute('segment');
     }
 
+    public function changeSegment(string $segmentKey)
+    {
+        $request = $this->requestStack->getCurrentRequest();
```
nnatter, line 122: i think this is not used

src/Sulu/Component/Webspace/Analyzer/RequestAnalyzer.php
```diff
@@ -97,6 +117,16 @@ public function getSegment()
         return $this->getAttribute('segment');
     }
 
+    public function changeSegment(string $segmentKey)
```
nnatter, line 120: not sure if this should be nullable. is it possible to remove the segment?

src/Sulu/Bundle/WebsiteBundle/Routing/ContentRouteProvider.php
```diff
@@ -170,6 +178,15 @@ public function getRouteCollectionForRequest(Request $request)
             } elseif (!$this->checkResourceLocator($resourceLocator, $prefix)) {
                 return $collection;
             } else {
+                if ($document instanceof ExtensionBehavior) {
+                    $documentSegmentKey = $document->getExtensionsData()['excerpt']['segment'];
+                    $segment = $this->requestAnalyzer->getSegment();
+
+                    if ($segment && $segment->getKey() !== $documentSegmentKey) {
```
nnatter, line 185: either we need to check if the `$documentSegmentKey` is not null, or we need to adjust the `requestAnalyzer->changeSegment()` to accept null as parameter 🙂

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/SingleSelect.js
```diff
@@ -46,13 +46,13 @@ export default class SingleSelect extends React.Component<FieldTypeProps<string
 
         return (
             <SingleSelectComponent disabled={!!disabled} onChange={this.handleChange} value={value}>
-                {values.value.map(({name: value, title}) => {
-                    if (typeof value !== 'string' && typeof value !== 'number') {
+                {values.value.map(({name: value, title}, index) => {
+                    if (typeof value !== 'string' && typeof value !== 'number' && value !== undefined) {
                         throw new Error('The children of "values" must only contain values of type string or number!');
```
nnatter, line 51: looks like this is not correct anymore

src/Sulu/Bundle/PageBundle/Resources/js/containers/Form/fields/SegmentSelect.js
```diff
@@ -0,0 +1,43 @@
...
+    render() {
+        const {disabled, value} = this.props;
+
+        return (
+            <SingleSelect
+                disabled={!!disabled}
+                onChange={this.handleChange}
+                value={value}
+            >
+                <SingleSelect.Option>
```
nnatter, line 32: this is only displayed if i explicitly select it. after saving. it shows `please choose` again. i think the problem is that the `single_select` uses an empty string instead of `null` as default value

src/Sulu/Bundle/PageBundle/Resources/config/forms/page_block_settings.xml
nnatter, line 51: would be nice to hide the `none selected` option here 🙂

nnatter: but we can do that in the article pr

src/Sulu/Component/Webspace/Analyzer/RequestAnalyzer.php
danrot: It probably should not... If no segment is set, it should be the default segment.

src/Sulu/Bundle/WebsiteBundle/Routing/ContentRouteProvider.php
danrot: Is this the error you have talked about today? I can't really reproduce it, how have you managed that?

src/Sulu/Bundle/PageBundle/Resources/js/containers/Form/fields/SegmentSelect.js
danrot: Hm, that's weird... You are talking about [this line](https://github.com/sulu/sulu/blob/b48343f/src/Sulu/Bundle/PageBundle/Content/Types/SingleSelect.php#L25), right? We might be able to fix that in this `SegmentSelect` component, but that would feel more like a workaround... I tend to change the default value in the line mentioned before, because that would fix the problem in general. Or do you think that is too much of a BC break? :thinking:

src/Sulu/Bundle/PageBundle/Resources/config/forms/page_block_settings.xml
danrot: Hm, talked to @chirimoya, and we said that this is good enough... The thing is that the default segment should probably be selected by default, and it would need the webspace for that... But yeah, we can reevaluate in the PR for the article segments, and if it can be done easily we can add it.

src/Sulu/Bundle/PageBundle/Resources/js/containers/Form/fields/SegmentSelect.js
danrot: Have implemented it for now like that.

src/Sulu/Bundle/WebsiteBundle/Routing/ContentRouteProvider.php
danrot: No idea why it wasn't happening before :see_no_evil: Fixed it now in my last commit, the segment does not change if no segment is set on the loaded document.

src/Sulu/Bundle/PageBundle/Resources/js/containers/Form/fields/SegmentSelect.js
nnatter: no, changing the value is okay for me 🙂

PR #5278, opened 2020-05-06
Rename sulu_content_path in sulu_content_url

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

#### Why?

Since sulu [1.2](https://github.com/sulu/sulu/blob/master/UPGRADE.md#sulu_content_path) the `sulu_content_path` does return the full url. To be similar to [Symfony Url Twig Extension](https://symfony.com/doc/4.4/reference/twig_reference.html#url) which does also return the full path I would rename the function to _url instead of _path.

#### To Do

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,12 @@
 # Upgrade
 
+## unreleased
+
+### The `sulu_content_path` and `sulu_content_root_path` are deprecated
+
+The `sulu_content_path` and `sulu_content_root_path` are deprecated use instead the
+new `sulu_content_url` and `sulu_content_root_url` twig extension.
+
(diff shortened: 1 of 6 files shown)


danrot:
TBH, I am a bit reluctant to that change... Are we sure that we want to keep a function like this, and not use the internal Symfony `url` or `path` methods in some way? Changing that now just for cosmetic reasons seems a little bit tedious for developers using it, if we might change that behavior again later.

PR #5280, opened 2020-05-07
Allow to use SetUnpublishedToolbarAction for unlocalized entities

nnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This PR adjusts the `SetUnpublishedToolbarAction` to not throw an error if the resource store of the form does not have a locale.

#### Why?

Because right now, the `SetUnpublishedToolbarAction` cannot be used for unlocalized entities.


nnatter:
@danrot Do you know if there was a specific reason for the check I removed? 
As far as I understand, the `DeleteDraftToolbarAction` (which is very similar to the `SetUnpublishedToolbarAction`) does not contain such a check.

danrot:
My guess would be that this is a leftover from when we have implemented a separate toolbar action just for the pages...

PR #5282, opened 2020-05-08
Create MemoryFormStoreFactory with decorator handling schema loading

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | supports #5277, some discussion in https://github.com/sulu/sulu/issues/5271#issuecomment-625735771
| License | MIT

#### What's in this PR?

This PR adds a `MemoryFormStoreFactory`, which allows to use the `MemoryFormStore` without the hassle of loading the schema from the server.

#### Why?

Because this is quite cumbersome when being done manually (handling of async calls, knowledge required how this stuff is wired together, ...). It also allows to pass the store without already being initialized, which allows to have the entire loading handling be implemented in the `Form` component, and don't worry about that in the outer component at all.

#### Example Usage

~~~javascript
const memoryFormStore = memoryFormStoreFactory.createFormStoreFactory(formKey, data);
~~~

#### BC Breaks/Deprecations

None, everything should continue to work as it has already done, just with a new way of `MemoryFormStore` instantiation.

#### To Do

- [x] Add all `FormStoreInterface` properties  to `SchemaFormStore`


src/Sulu/Bundle/AdminBundle/Resources/js/containers/ProfileFormOverlay/ProfileFormOverlay.js
```diff
@@ -25,21 +24,17 @@ const RESOURCE_KEY = 'profile';
...
-            metadataStore.getJsonSchema(FORM_KEY),
-            ResourceRequester.get(RESOURCE_KEY),
-        ]).then(this.handleResponse);
+        ResourceRequester.get(RESOURCE_KEY).then(this.handleResponse);
     }
 
-    @action handleResponse = ([schema, jsonSchema, data]: [RawSchema, Object, Object]) => {
-        this.formStore = new MemoryFormStore(data, schema, jsonSchema);
+    @action handleResponse = (data: Object) => {
+        this.formStore = memoryFormStoreFactory.createFromFormKey(FORM_KEY, data);
```
danrot, line 37: @nnatter @alexander-schranz I think this here is the interesting part. No need to handle the schema asynchronously on your own, and still an easy way to create the store, while still being able to passing it to the `Form` component.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/SchemaFormStore.js
```diff
@@ -0,0 +1,69 @@
...
+import {computed} from 'mobx';
+import type {FormStoreInterface, RawSchema, Schema} from '../types';
+import metadataStore from './metadataStore';
+
+export default class SchemaFormStore implements FormStoreInterface {
+    innerFormStore: ?FormStoreInterface;
+
+    constructor(formKey: string, initializer: (schema: RawSchema, jsonSchema: Object) => FormStoreInterface) {
+        Promise.all([
+            metadataStore.getSchema(formKey),
```
danrot, line 11: Eventually we might want to pass `type` and `metadataOptions` here as well.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/SchemaFormStore.js
```diff
@@ -0,0 +1,69 @@
+// @flow
+import {computed} from 'mobx';
+import type {FormStoreInterface, RawSchema, Schema} from '../types';
+import metadataStore from './metadataStore';
+
+export default class SchemaFormStore implements FormStoreInterface {
+    innerFormStore: ?FormStoreInterface;
+
+    constructor(formKey: string, initializer: (schema: RawSchema, jsonSchema: Object) => FormStoreInterface) {
```
danrot, line 9: I think we should move the `initializer` first, because then it is easier to add more parameters like `formKey` at the end of the function signature.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/ProfileFormOverlay/ProfileFormOverlay.js
nnatter: I am kind of curious why this is implemented using a `MemoryFormStore` when the data is loaded from the server anyway 😅 
But beside of that, the solution looks good to me 🙂

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/SchemaFormStore.js
nnatter: Not sure if it is a good idea, but we could even make a `Promise<FormStoreInterface>` the only argument of the constructor and call it something like `DeferredFormStore`. Both ways, i feel like its hard to find a good name for this.

nnatter:
The proposal look good to me. I think it would be nice to already implement a basic `ResourceFormStoreFactory` to keep things consistent 🙂

src/Sulu/Bundle/AdminBundle/Resources/js/containers/ProfileFormOverlay/ProfileFormOverlay.js
danrot: @nnatter At first I thought the same, but then I remembered that this is loading the profile, which is not a standard REST interface, because it does not use an `id` parameter in the URL but data from the session to know which user to load.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/SchemaFormStore.js
danrot: Don't know, that feels somehow weird to me :see_no_evil: And I would have to create another `FormStore` implementation... Also, I don't think that there'll be huge BC issues here, because people should use the factory.

danrot:
@nnatter I only created one method in the `resourceFormStoreFactory`, because I was not sure how to create the one which would also create the `ResourceStore` for you (and I also only found two place where that would be applicable). I didn't want to invest too much time into this now, therefore I decided to not implement that for now.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/SchemaFormStore.js
```diff
@@ -0,0 +1,169 @@
+// @flow
+import {action, computed, observable} from 'mobx';
+import type {FormStoreInterface, RawSchema, Schema, SchemaEntry} from '../types';
+import metadataStore from './metadataStore';
+
+export default class SchemaFormStore implements FormStoreInterface {
```
nnatter, line 6: not sure about this name as it is very similar to `MemoryFormStore` and `ResourceFormStore` but the class is quite different from them. but i dont really have a better suggestion 🙈 is it common to mark decorator classes with a suffix or something?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/tests/stores/memoryFormStoreFactory.test.js
```diff
@@ -0,0 +1,34 @@
...
+    const schemaPromise = Promise.resolve(schema);
+    const jsonSchemaPromise = Promise.resolve(jsonSchema);
+
+    metadataStore.getSchema.mockReturnValue(schemaPromise);
+    metadataStore.getJsonSchema.mockReturnValue(jsonSchemaPromise);
+
+    const memoryFormStore = memoryFormStoreFactory.createFromFormKey('test');
+
+    expect(memoryFormStore).toBeInstanceOf(SchemaFormStore);
+    setTimeout(() => {
```
nnatter, line 29: did you use `setTimeout` instead of `return Promise.all([schemaPromise, jsonSchemaPromise]).then(() => {` on purpose here?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/SchemaFormStore.js
danrot: I think I have seen examples without a decorator suffix and with a decorator suffix. `SchemaFormStoreDecorator` would also be fine for me, especially because nobody will instantiate that class themselves anyway :slightly_smiling_face:

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/tests/stores/memoryFormStoreFactory.test.js
danrot: Yeah, `Promise.all` was not working, but I was playing around quite a bit, maybe it does work now without it... Will double-check it.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/SchemaFormStore.js
nnatter: would like `SchemaFormStoreDecorator` better because it hints that it is not another `FormStore` implementation 🙂

danrot:
@nnatter Fixed both issues.

config/packages/dev/sulu_admin.yaml
```diff
@@ -0,0 +1,3 @@
+sulu_admin:
```
nnatter, line 1: not sure if you commited this intentionally? 🙂

danrot: Right :see_no_evil: Have to fix that properly in a separate PR, so that I don't have to keep that file lying around in my file system :grimacing:

PR #5294, opened 2020-05-14
Fix sorting of smart content when used with doctrine

danrot:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR fixes a bug that occured when the `DataProviderRepositoryTrait` without any sorting. This can be reproduced by using the following property in an XML file:

```xml
<property name="accounts" type="smart_content">
    <meta>
        <title lang="en">Accounts</title>
    </meta>
    <params>
        <param name="provider" value="accounts"/>
    </params>
</property>
```

The following error should appear without this fix:

```
Notice: Trying to access array offset on value of type null
```

### DIFF
--- a/src/Sulu/Component/SmartContent/Orm/DataProviderRepositoryTrait.php
+++ b/src/Sulu/Component/SmartContent/Orm/DataProviderRepositoryTrait.php
@@ -31,7 +31,7 @@ trait DataProviderRepositoryTrait
         $this->appendJoins($queryBuilder, $alias, $locale);
 
         if (isset($filters['sortBy'])) {
-            $sortMethod = isset($filters['sortMethod']) ? $filters['sortMethod'] : 'asc';
+            $sortMethod = $filters['sortMethod'] ?? 'asc';
             $this->appendSortBy($filters['sortBy'], $sortMethod, $queryBuilder, $alias, $locale);
         }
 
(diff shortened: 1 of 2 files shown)


src/Sulu/Component/SmartContent/Orm/DataProviderRepositoryTrait.php
```diff
@@ -30,8 +30,8 @@ public function findByFilters($filters, $page, $pageSize, $limit, $locale, $opti
             ->orderBy($alias . '.id', 'ASC');
         $this->appendJoins($queryBuilder, $alias, $locale);
 
-        if (array_key_exists('sortBy', $filters)) {
-            $sortMethod = array_key_exists('sortMethod', $filters) ? $filters['sortMethod'] : 'asc';
+        if (isset($filters['sortBy'])) {
+            $sortMethod = isset($filters['sortMethod']) ? $filters['sortMethod'] : 'asc';
```
alexander-schranz, line 34: could be simplified using the null coalescing operator:

```suggestion
$sortMethod = $filters['sortMethod'] ?? 'asc';
```

As it works the same as isset e.g.: `php -r '$data = ["test" => null]; var_dump($data["test"] ?? "replace");'`

danrot:
Done :+1:

PR #5303, opened 2020-05-20
Activate cache based on environment configuration instead of in extenion

danrot:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | #5277, sulu/skeleton#55
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR removes a confusing behavior. Instead of activating resp. deactivating the cache in the `SuluHttpCacheExtension`, it is done now by properly configuring it in the configuration files.

#### Why?

Because it took me quite some time to get the tests running, and this was one of the reasons for that.


src/Sulu/Bundle/HttpCacheBundle/DependencyInjection/SuluHttpCacheExtension.php
```diff
@@ -43,23 +43,21 @@ public function prepend(ContainerBuilder $container)
             ];
         }
 
-        if ($this->shouldCache($container)) {
```
alexander-schranz, line 46: This is a BC Break there should be an additional disable or enable flag which is set by a `kernel.environment`  another default.

See for example FOSRestBundle:

https://github.com/FriendsOfSymfony/FOSRestBundle/blob/007cba22ba3addae80c42c47f61cca740fb11273/DependencyInjection/Configuration.php#L36-L39

https://github.com/FriendsOfSymfony/FOSRestBundle/blob/007cba22ba3addae80c42c47f61cca740fb11273/DependencyInjection/FOSRestExtension.php#L39-L42

so then this will go with:

```php
if ($config['enabled']) {
    // ...
}
```

danrot: Right, I forgot to mention the BC break... But the example you have been posted uses the `debug` variable, not the `environment`... Having the environment being handled in a different way is exactly what made this code so hard to understand...

So I would suggest we make this change in the `master` branch and add a text in the `UPGRADE.md`. If we include some environment specific code in the extension again, developers touching this code will wonder why the application behaves differently than how the configuration would suggest.

We can maybe also add a separate `enabled` flag, but I don't think that this is absolutely necessary. But if we add it, I would not prefill it in a magical way for the using developer by the `environment` variable.

alexander-schranz: I would keep the current behaviour as it is. It was just an example you can use the `kernel.environment`. I would not make a need for an UPGRADE here its typical behaviour have cache by default enabled only for stage and prod env.

danrot: The thing about the current behavior is that it is incredibly confusing. In addition to that it is currently impossible to activate caching for the `test` environment, but I would need that, in order to be able to add a test for the segment caching in #5277.

Having a separate `enabled` flag improves the situation a bit, since it at least would be possible to activate the caching in the `dev` and `test` environment, but it would not help with the confusion it causes. Even if I add the configuration as it is done now, it will not work in the `dev` and `test` environment, and it is not easy to find out why. In order to do so, you would have to dig into the `SuluHttpCacheExtension` code. It literally took me an entire afternoon in order to get this running, although I know the code base pretty well.

Therefore I would say that this justifies the BC break, especially because the only thing that might break is the `dev` environment, `prod` doesn't change its behavior. We can add it to the `UPGRADE.md`, and our [Upgrade Guide](https://docs.sulu.io/en/2.0/upgrades/upgrade-2.0.html) also says that you have to check the changes in the `sulu/skeleton` repository.

The alternative is to live with that "quirks" (at least it is one IMO, since e.g. our [doctrine](https://github.com/sulu/skeleton/blob/master/config/packages/prod/doctrine.yaml) and [routing](https://github.com/sulu/skeleton/blob/master/config/packages/prod/routing.yaml) configuration don't work this way), and change it in 3.0, and I am sure that we will trip another time when implementing a new caching feature before then.

danrot: I think I have adopted it in the way you have suggested, but I am not totally happy with it. In addition to the reasons I have already stated above, it makes configuring this bundle unnecessarily hard. E.g. a configuration like this is now possible, although it does not make a lot of sense:

```yaml
sulu_http_cache:
    enabled: false
    tags:
        enabled: true
```

But since it seems that there were also not a lof of issues with that previously, I hope it stays this way... Would still implement it differently, but I can live with that solution.

alexander-schranz: I totally understand why you want that magic, but from BC I think it will annoy many developers when cache is automatically activated then in dev and test environments when updating a project. But If I'm the only one who see it like that we can do this bc break. /cc @sulu/core-team

wachterjohannes: @danrot i dont have a strong oppinion - but i would also avoid the BC break and set the correct default value in the configuration

nnatter: i dont really like the `enabled` option, but i can live with it if we need it to prevent a bc break. 
can we somehow directly deprecate that option? eg by using a `disabled` option instead?

danrot: @chirimoya also mentioned that part of the code: https://github.com/sulu/skeleton/blob/723a99d/public/index.php#L54

So in the `dev` environment the response will still not be cached because of that. I also tested it with 10e5b940453300e09685c4d05688db6f9f40180f, and even the response headers to not change. So is it really worth the pain?

Regarding a `disabled` option: I think that even makes stuff more confusing. We could probably still use the `enabled` option to do that and deprecate just that one. Although the actual behavior I am trying to deprecate cannot be deprecated in that case (that different environments have different settings without any configuration).

alexander-schranz: @danrot ok if the cache is disabled in the `index.php` that sounds good enough for me. Can you based on 10e5b940453300e09685c4d05688db6f9f40180f add to the `UPGRADE.md` also that the `config/packages/stage/sulu_http_cache.yaml` should be created with the following content:

```yaml
imports:
    - { resource: '../prod/sulu_http_cache.yaml' }
```

And I would write it that its not something optional currently the UPGRADE.md text with:

> and if e.g. the `dev` environment should not use the HTTP cache, it has to be configured
appropriately

sounds like its optional change. I don't think all know how the http cache effects the DX and not sure how it effects tests, but at least if they fail they know they didn't configure something correctly. But I would more write it that changes are required.

Something like:

    The `sulu_http_cache` configuration behaved differently based on the configured environment. Since this behavior was causing some configuration to be ignored in the `dev` and `test` environment, it was very hard to understand. Therefore we removed this behavior, so the following change is required to disable the HTTP cache again on `dev` and `test` environment:
      
    ```bash
    git mv config/packages/sulu_http_cache.yaml config/packages/prod/sulu_http_cache.yaml
    ```
    
    ```bash
    # config/packages/stage/sulu_http_cache.yaml
    imports:
        - { resource: '../prod/sulu_http_cache.yaml' }
    ```

danrot: Reverted the changes and updated the UPGRADE.md

PR #5304, opened 2020-05-25
Add possibility to include search fields as indexed

### LINKED ISSUE #5290: MassiveSearch Document title should also be indexed for search boost
| Q | A
| --- | ---
| Sulu Version | 1.6.33

# Actual Behavior

As far as I can see, it is impossible to influence the search result order by making the title or url more important.

To work around this I wanted to give the title a boost directly in the search query. (something like  `__title:("test")^2`)

Unfortunately this is also not possible right now, as the __title is only stored but not indexed.

I have the following property definition
```
<property name="headline" type="text_line" mandatory="true">
    <meta>
        <title lang="de">Überschrift</title>
    </meta>
    <params>
    </params>
    <tag name="sulu.search.field" role="title" />
</property>
```

When I dump the result of a search (e.g. a `ZendSearch\Lucene\Document` object)

> "__title" => Field {
      +name: "__title"
      +value: "xxxxxxx"
      +isStored: true
      +isIndexed: 0
      +isTokenized: 0
      +isBinary: false
      +storeTermVector: false
      +boost: 1.0
      +encoding: "UTF-8"
    }

it shows, that the property is not indexed.

# Expected Behavior

The title search field should be indexed, same goes for the url in my opinion.
I want to be able to query for the title explicitly, which is not possible.

Alternatively having any kind of option of giving a search field a boost or a weight would also be a step in the right direction.

ampaze:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

The PR makes it possible to include fields as `indexed`, not just as part of the aggregate.

#### Why?

If a field is (additionally) indexed by itself, it can be used in a search query with higher priority.

#### Example Usage

The XML would look something like this:

```xml
<property name="search_keywords" type="text_line" mandatory="false">
    <meta>
        <title lang="de">Suchpriorisierung für folgende Schlüsselwörter</title>
    </meta>
    <params>
        <param name="search_keywords" value="true"/>
    </params>
    <tag name="sulu.search.field" index="indexed"/>
</property>
```

The query something like this

> search_keywords:("test")^2

For more
https://github.com/sulu/sulu/issues/5290

### DIFF
--- a/src/Sulu/Bundle/ContentBundle/Search/Metadata/StructureProvider.php
+++ b/src/Sulu/Bundle/ContentBundle/Search/Metadata/StructureProvider.php
@@ -159,7 +159,7 @@ class StructureProvider implements ProviderInterface
                                         '[' . $componentProperty->getName() . ']'
                                     ),
                                     'aggregate' => true,
-                                    'indexed' => false,
+                                    'indexed' => isset($tagAttributes['index']) && 'indexed' === $tagAttributes['index'],
                                 ]
                             );
                         }
@@ -386,7 +386,7 @@ EOT;
                     'type' => isset($tagAttributes['type']) ? $tagAttributes['type'] : 'string',
                     'field' => $this->getContentField($property),
                     'aggregate' => true,
-                    'indexed' => false,
+                    'indexed' => isset($tagAttributes['index']) && 'indexed' === $tagAttributes['index'],
                 ]
             );
         }


alexander-schranz:
As `indexed` seems to be a boolean flag I would go with:

```xml
<tag name="sulu.search.field" indexed="true"/>
```

/cc @sulu/core-team

ampaze:
I think having both `index` and `indexed` attributes makes it even more confusing as it is right now.

You could think of `index="aggregated"` as the default value.

alexander-schranz:
@ampaze ah know I understand the parameter did exist before. Thought it was introduced.

ampaze:
I just used what was already there ;)

Can this be merged?

danrot:
Would you say this solves #5290?

I have seen that [you have added a comment there](https://github.com/sulu/sulu/issues/5290#issuecomment-630915820), with an example of a configuration like this:

```xml
<tag name="sulu.search.field" weight="2"/>
```

Do you think that replaces that proposal? Or is that somehow a prerequesite? The XML proposal would be much easier to use, but then it is fixed, and can't really be changed in the query. I would even say that the interface of the `SearchManager` would have to look different in that case. So I would say having the solution as you've suggested in this PR would solve the other issue.

And the code looks good to me, but didn't test it yet... And we can only merge that if there are some tests :wink: 

Does this PR require https://github.com/massiveart/MassiveSearchBundle/pull/145 in order to work?

ampaze:
For me it kind of would solve #5290. 

While it does not require https://github.com/massiveart/MassiveSearchBundle/pull/145, it will not  add the title or url field into the index, as these are hardcoded to be only stored right now. 

The suggestion

```xml
<tag name="sulu.search.field" weight="2"/>
```

was just a thought, I don't think it is actually possible. You would still have to use a different search query, as the default query only searches in the aggregated field. 

So i will just stick with my custom search query, which I need this PR to be merged for ...

ampaze:
> And the code looks good to me, but didn't test it yet... And we can only merge that if there are some tests 😉

I wouldn't even know where to start with writung a test for something like that. 
Can you show me the current test for the **index** attribute?

danrot:
@ampaze You are right, there are no tests yet... While that it is not really good, I'll merge that after I've tested it.

danrot (approved):

danrot:
Thank you for your contribution!

almare:
Why do you choose the word weight? In the context of the search engines like solr or elasticsearch it's called boost. Do I misunderstand the use case here for weight or in other words is boost !== weight?

danrot:
@almare This has not been implemented now anyway. Instead you can write search queries like this, in order to influence the weight, if you index that attribute:

```
search_keywords:("test")^2
```

PR #5305, opened 2020-05-25
Add smart-content migration for sortBy

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Adds missing migration for `sortBy` property in snippets and pages.

### DIFF
--- a/src/Sulu/Bundle/PageBundle/Resources/phpcr-migrations/Version202005250917.php
+++ b/src/Sulu/Bundle/PageBundle/Resources/phpcr-migrations/Version202005250917.php
@@ -16,7 +16,7 @@ use PHPCR\SessionInterface;
 use Symfony\Component\DependencyInjection\ContainerAwareInterface;
 use Symfony\Component\DependencyInjection\ContainerAwareTrait;
 
-class Version202005191116 implements VersionInterface, ContainerAwareInterface
+class Version202005250917 implements VersionInterface, ContainerAwareInterface
 {
     use ContainerAwareTrait;
 
@@ -45,7 +45,7 @@ class Version202005191116 implements VersionInterface, ContainerAwareInterface
     {
         $queryManager = $session->getWorkspace()->getQueryManager();
 
-        $query = 'SELECT * FROM [nt:unstructured] WHERE ([jcr:mixinTypes] = "sulu:snippet")';
+        $query = 'SELECT * FROM [nt:unstructured] WHERE ([jcr:mixinTypes] = "sulu:snippet" or [jcr:mixinTypes] = "sulu:page")';
         $rows = $queryManager->createQuery($query, 'JCR-SQL2')->execute();
 
         foreach ($rows as $row) {
@@ -54,11 +54,10 @@ class Version202005191116 implements VersionInterface, ContainerAwareInterface
             foreach ($node->getProperties() as $property) {
                 if (\is_string($property->getValue())) {
                     $propertyValue = json_decode($property->getValue(), true);
-                    if (\is_array($propertyValue) && \array_key_exists('items', $propertyValue)) {
-                        foreach ($propertyValue['items'] as &$item) {
-                            if (isset($item['type']) && 'content' === $item['type']) {
-                                $item['type'] = 'pages';
-                            }
+                    if (\is_array($propertyValue) && \array_key_exists('presentAs', $propertyValue) &&
+                        \array_key_exists('limitResult', $propertyValue) && \array_key_exists('sortBy', $propertyValue)) {
+                        if (is_array($propertyValue['sortBy'])) {
+                            $propertyValue['sortBy'] = \count($propertyValue['sortBy']) > 0 ? $propertyValue['sortBy'][0] : null;
                         }
 
                         $property->setValue(json_encode($propertyValue));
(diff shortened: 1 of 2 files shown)


src/Sulu/Bundle/PageBundle/Resources/phpcr-migrations/Version202005250917.php
```diff
@@ -0,0 +1,95 @@
...
+        $rows = $queryManager->createQuery($query, 'JCR-SQL2')->execute();
+
+        foreach ($rows as $row) {
+            $node = $row->getNode();
+
+            foreach ($node->getProperties() as $property) {
+                if (\is_string($property->getValue())) {
+                    $propertyValue = json_decode($property->getValue(), true);
+                    if (\is_array($propertyValue) && \array_key_exists('presentAs', $propertyValue) &&
+                        \array_key_exists('limitResult', $propertyValue) && \array_key_exists('sortBy', $propertyValue)) {
```
danrot, line 58: I have just seen that you are adding the backslash in front of every native function call... I know that this is supposed to be faster, but I still haven't added it until now, because I don't like the look of it, but that's a personal opinion :see_no_evil: But I have also seen that this is not even consistent within this file (see the `is_array` call in the next line).

I think we should introduce a php-cs-fixer rule for this. There is the `native_constant_invocation` and the `native_function_invocation` rule, which would allow to force this on every native function.

But I would prefer the `global_namespace_import` rule, which would force us to import all of the functions we need in these files at the very top. Personally I like the cleaner looks, a look at the very top of the file would allow you to see what is required in this file and in addition the rule is not risky, whereby the previously two being mentioned are.

@sulu/core-team What's your take on this?

nnatter: dont have a strong opinion except that being consistent would be nice 🙂 `global_namespace_import` sounds good to me

alexander-schranz: At current state in bundles we are only using `\` for the functions which have performance improvement [then](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/d658dc0201bdd457d32fc412d810885f282ac064/src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php#L356-L398) over the `native_function_invocation`.

I personally not a big fan of importing global classes because everytime I review code I need to look at the top if it is a global `\RuntimeException` or a namespaced `RuntimeException` from the library, so I would prefer forcing to write `\RuntimeException` for classes and not importing global ones there. For functions its okay for me to import them as we don't use libraries which provide global functions.

danrot: > I personally not a big fan of importing global classes because everytime I review code I need to look at the top if it is a global `\RuntimeException` or a namespaced `RuntimeException` from the library, so I would prefer forcing to write `\RuntimeException` for classes and not importing global ones there. For functions its okay for me to import them as we don't use libraries which provide global functions.

I don't really like a rule which is being mixed, like functions are being imported, but classes are not. I think consistency is key here, everything else is hard to understand. I have never experienced the problem you have described here (I always assume that `RuntimeException` is the global class :see_no_evil:), and I have never had the case that this is important when I was reading the code. Additionally I think it is much nicer to see on the very top what is going to be used in this class. Has also the advantage that a simple search easily yields all occurences of the `RuntimeException` (just search for `use RuntimeException`), which would also work in environments outside of your editor (e.g. anywhere on GitHub).

Also, we are using e.g. `\InvalidArgumentException` for years, and I have never really liked it :see_no_evil: While this is very subjective, the fact that it is not used consistently makes it very hard to put any trust in rules like a missing starting `\` means that the method has been overridden.

alexander-schranz: > I have never had the case that this is important when I was reading the code. 

Depends on which libraries you use if you use libraries which don't use things like there own `RuntimeException` you never will get into this issue, but example `ffmpeg` and `imagine` have there own `RuntimeException`.  

> Has also the advantage that a simple search easily yields all occurences of the RuntimeException

Think there is really no difference to get the occurences in today IDEs, 

> Additionally I think it is much nicer to see on the very top what is going to be used in this class

Thats the problem as in review you then always need first open the file that you know which classes are important at the top. Not always the class was added as use in the same PR. 

>  a missing starting \ means that the method has been overridden.

Thats why I would force `\` for all so we can trust it that its global.

danrot: > > I have never had the case that this is important when I was reading the code.
> 
> Depends on which libraries you use if you use libraries which don't use things like there own `RuntimeException` you never will get into this issue, but example `ffmpeg` and `imagine` have there own `RuntimeException`.

Since that problem exists in any class, that is not such a big deal to me (again, this might be subjective).

> > Has also the advantage that a simple search easily yields all occurences of the RuntimeException
> 
> Think there is really no difference to get the occurences in today IDEs,

Personally I don't really like the thinking of "my IDE can do that", because code is quite often read outside of an IDE, especially during code review, or when looking something up in github.

> > Additionally I think it is much nicer to see on the very top what is going to be used in this class
> 
> Thats the problem as in review you then always need first open the file that you know which classes are important at the top. Not always the class was added as use in the same PR.

As said, can't remember I ran into this issue. Then I would have to wonder for almost any class I encounter in a code review if it is really the correct one, I just assume that it is the correct one, since it would horribly break if the wrong class is used (and to some extent PHPStan also covers that).

> > a missing starting \ means that the method has been overridden.
> 
> Thats why I would force `\` for all so we can trust it that its global.

As you might have already guessed, I am preferring importing over putting a `\` in front of some functions. In theory I am also fine with putting the `\` in front of every native function/constant, but I am a bit afraid of it, because this is a risky rule, and it might break some stuff. So it feels safer to me to use the `global_namespace_import`.

I think further discussion does not make any sense, I would put it to a vote in our slack channel, and just go for whatever the majority likes. Only if the chosen final solution does not case any problems when we are trying to do that, of course.

PR #5313, opened 2020-06-02
Fix filtering of snippets when choosing snippet areas

### LINKED ISSUE #5311: Default snippet selection overlay ignores snippet type
| Q | A
| --- | ---
| Sulu Version | 2.0.7

# Actual Behavior

When I want to select the default snippet for a specific type/area, the overlay contains all snippets regardless of their type. When I select a snippet of the wrong type, the administration interface crashes.

current behaviour

# Expected Behavior

The overlay should only contain snippets that are allowed for the type/area.

danrot:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #5311 
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR fixes the name of the query parameter used for filtering snippet areas.

### DIFF
--- a/src/Sulu/Bundle/TestBundle/Testing/TestUserProvider.php
+++ b/src/Sulu/Bundle/TestBundle/Testing/TestUserProvider.php
@@ -87,13 +87,15 @@ class TestUserProvider implements UserProviderInterface
             $this->entityManager->persist($contact);
 
             $user = $this->userRepository->createNew();
-            $this->setCredentials($user);
-            $user->setLocale('en');
             $user->setContact($contact);
             $this->entityManager->persist($user);
-        } else {
-            $this->setCredentials($user);
         }
+
+        $user->setUsername(self::TEST_USER_USERNAME);
+        $user->setSalt('');
+        $encoder = $this->userPasswordEncoderFactory->getEncoder($user);
+        $user->setPassword($encoder->encodePassword('test', $user->getSalt()));
+        $user->setLocale('en');
 
         $this->entityManager->flush();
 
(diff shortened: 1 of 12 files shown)


danrot:
@nnatter There was another flaky test, that only failed if the entire test suite was ran, not only the tests of the SnippetBundle :see_no_evil: Now it should be working, you can already have a look at it :slightly_smiling_face:

src/Sulu/Bundle/TestBundle/Testing/TestUserProvider.php
```diff
@@ -87,14 +87,16 @@ public function getUser()
...
             $user = $this->userRepository->createNew();
-            $this->setCredentials($user);
-            $user->setLocale('en');
             $user->setContact($contact);
             $this->entityManager->persist($user);
-        } else {
-            $this->setCredentials($user);
         }
 
+        $user->setUsername(self::TEST_USER_USERNAME);
```
nnatter, line 94: no sure if you commited this on purpose?

danrot: I did, this is the fix for [this error](https://travis-ci.org/github/sulu/sulu/jobs/694915779#L1757).

It especially is about the `setLocale` below. One of the tests of the SecurityBundle sets the locale to another language, and it will stay like that for all tests afterwards. For this reason the tests were failing if all tests are executed, but not if only the SnippetBundle tests were executed.

src/Sulu/Bundle/TestBundle/Testing/TestUserProvider.php
```diff
@@ -163,9 +165,5 @@ public function supportsClass($class)
      */
     private function setCredentials(UserInterface $user)
```
danrot, line 166: But then I should also delete this function completely :see_no_evil: Already on it!

PR #5325, opened 2020-06-04
Update branch aliases

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

Update branch aliases.

#### Why?

As long as we use the release prefixes in sulu for branches we need to update them also in the composer.json.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -107,7 +107,8 @@
         "branch-alias": {
             "dev-release/1.6": "1.6.x-dev",
             "dev-release/2.0": "2.0.x-dev",
-            "dev-master": "2.1.x-dev"
+            "dev-release/2.1": "2.1.x-dev",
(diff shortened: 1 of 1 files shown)


danrot:
Is that really required in the `release/1.6` branch? Wouldn't it be enough to add this to the `master` branch? I suppose the `master` branch is where packagist takes this information from.

alexander-schranz:
@danrot we currently did update it in all branches to be consistence and avoid confusing.

danrot:
Don't think that doing this only on the `master` branch would caus any confusion, but it also does not hurt, so I don't really care.

PR #5327, opened 2020-06-04
Add svg support by default

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

Add svg support by default.

#### Why?

I think we should provide the svg support by default, its just something which will be forget in many projects to be added. As the svg library doesn't have big [requirements](https://github.com/contao/imagine-svg/blob/master/composer.json) I think we should install it by default. Also there will be a stable 1.0 release soon: https://github.com/contao/imagine-svg/issues/22

/cc @sulu/core-team what do you think?

#### To Do

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,6 +31,7 @@
         "ext-simplexml": "*",
         "aferrandini/urlizer": "^1.0",
         "antishov/doctrine-extensions-bundle": "^1.2.2",
+        "contao/imagine-svg": "^0.2.2",
         "dantleech/phpcr-migrations-bundle": "^1.0",
         "doctrine/annotations": "^1.2",
         "doctrine/common": "^2.7.1",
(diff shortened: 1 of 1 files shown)


wachterjohannes:
@alexander-schranz i would wait for the 1.0 version but then its OK for me :)

alexander-schranz:
@wachterjohannes 1.0 will not be released soon. And it doesn't matter as we use only the Imagine Interface and not specific svg functions.

wachterjohannes:
@alexander-schranz if its ok for the rest it ok for me - but i would say we don't want to require pre versions if not neccessary.

nnatter:
sounds good to me 🙂

alexander-schranz:
@wachterjohannes  we also support all compatible versions of imagine from 0.6 - 1.2. As long as there are no breaking changes I would allow any version we can easily support without additional work on our side.

danrot:
I am on the same side as @wachterjohannes, I would wait for the stable release... And the problem is that you don't know if there will be any breaking changes in advance :confused:

alexander-schranz:
@danrot 1.0 is released.

alexander-schranz:
ping @danrot

PR #5328, opened 2020-06-05
fix url generation in link-prover with portals

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes none
| Related issues/PRs | none
| License | MIT
| Documentation PR | none

#### What's in this PR?

This PR add the domain to the link-provider for pages.

#### Why?

If you use multiple portals for the same webspace the portal will change when using links in the CKEditor.


src/Sulu/Bundle/PageBundle/Markup/Link/PageLinkProvider.php
```diff
@@ -107,18 +109,19 @@ function(Content $content) use ($locale, $scheme) {
      *
      * @param string $locale
      * @param string $scheme
+     * @param string|null $domain
      *
      * @return LinkItem
      */
-    protected function getLinkItem(Content $content, $locale, $scheme)
+    protected function getLinkItem(Content $content, $locale, $scheme, $domain)
```
alexander-schranz, line 116: As this is protected we should go with `$domain = null` if somebody used it.

alexander-schranz:
Also missing a TestCase for this bug.

wachterjohannes:
> I think this need to target `release/1.6` branch as this bug also appears there.

@alexander-schranz the problem is that there is no PageLinkProvider - i will create an own PR for 1.6

alexander-schranz:
@wachterjohannes there is the same PageLinkProvider in 1.6 here: https://github.com/sulu/sulu/blob/release/1.6/src/Sulu/Bundle/ContentBundle/Markup/Link/PageLinkProvider.php and if we merge it back to 2.0 this is correctly there.

CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 CHANGELOG for Sulu
 ==================
 
+* release/1.6
+    * BUGFIX      #5323  [SnippetBundle]           Fix /cmf/snippets node not exist in live workspace
```
alexander-schranz, line 5: something strange with the diff 🤔 can you repush and add a changelog entry here

PR #5340, opened 2020-06-09
Add refresh button for URL

### LINKED ISSUE #4221: Regenerate resource locator slug based on new page title
| Q | A
| --- | ---
| Bug? | no
| New Feature? | yes
| Sulu Version | 1.6x
| Browser Version | all

#### Actual Behavior

When you copy a page from one language to another and change the page's title afterwards, the resource locator/slug has to be changed/set manually. 

#### Expected Behavior

There should be a way to (re)generate the slug based on the new title to prevent users from trying to e.g. type in a slug with special characters or umlauts.

#### Steps to Reproduce

Add a new page, give it a title in one language (the slug ist generated automatically) and save/publish the page. Copy the page to another language/locale, change the page's title.

#### Possible Solutions

Add an event listener that (re)generates the page's slug whenever the title field is changed. Perhaps there could be a parameter in the config to enable/disable this feature.

cirykpopeye:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #4221
| Related issues/PRs | #4221
| License | MIT

#### What's in this PR?

A refresh button to regenerate the page url
Open to suggestions on the layout of things, but is a important feature for us, and probably many others.

#### Why?

If you want to change the URL atm, you have to manually type it in, keeping in mind of url regulations ( no spaces etc ). 
This button simplifies this by using the title to regenerate the url field.

#### To Do

- [ ] Add margin between buttons, atm there's no room between them, don't know how with Sulu UI?

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/ResourceLocator.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/ResourceLocator.js
@@ -7,6 +7,8 @@ import ResourceLocatorHistory from '../../../containers/ResourceLocatorHistory';
 import Requester from '../../../services/Requester';
 import type {FieldTypeProps} from '../../../types';
 import resourceLocatorStyles from './resourceLocator.scss';
+import {translate} from "../../../utils/Translator";
+import Button from "../../../components/Button";
 
 const PART_TAG = 'sulu.rlp.part';
 
@@ -15,9 +17,11 @@ const HOMEPAGE_RESOURCE_LOCATOR = '/';
 @observer
 class ResourceLocator extends React.Component<FieldTypeProps<?string>> {
     @observable mode: string;
+    @observable forceRefresh: boolean;
 
     constructor(props: FieldTypeProps<?string>) {
         super(props);
+        this.forceRefresh = false;
 
         const {dataPath, onChange, fieldTypeOptions, formInspector, value} = this.props;
         const {generationUrl, modeResolver} = fieldTypeOptions;
@@ -41,18 +45,25 @@ class ResourceLocator extends React.Component<FieldTypeProps<?string>> {
         }
 
         formInspector.addFinishFieldHandler((finishedFieldDataPath, finishedFieldSchemaPath) => {
-            if (value !== undefined) {
+            if (value !== undefined && !this.forceRefresh) {
                 return;
             }
 
-            if (formInspector.isFieldModified(dataPath)) {
+            if (formInspector.isFieldModified(dataPath) && !this.forceRefresh) {
                 return;
             }
+
 
             const {tags: finishedFieldTags} = formInspector.getSchemaEntryByPath(finishedFieldSchemaPath) || {};
-            if (!finishedFieldTags || !finishedFieldTags.some((tag) => tag.name === PART_TAG)) {
+            console.log(finishedFieldTags)
+            if (
+                (!finishedFieldTags || !finishedFieldTags.some((tag) => tag.name === PART_TAG)) &&
+                !this.forceRefresh
+            ) {
                 return;
             }
+
+            this.forceRefresh = false;
 
             const partEntries = formInspector.getPathsByTag(PART_TAG)
                 .map((path: string) => [path, formInspector.getValueByPath(path)])
@@ -70,6 +81,7 @@ class ResourceLocator extends React.Component<FieldTypeProps<?string>> {
                 return;
             }
 
+
             Requester.post(
                 generationUrl,
                 {
@@ -86,6 +98,13 @@ class ResourceLocator extends React.Component<FieldTypeProps<?string>> {
 
     handleBlur = () => {
         const {onFinish} = this.props;
+        onFinish();
+    };
+
+    handleButtonClick = () => {
+        const {onFinish} = this.props;
+        this.forceRefresh = true;
+        console.log(this);
         onFinish();
     };
 
@@ -135,6 +154,9 @@ class ResourceLocator extends React.Component<FieldTypeProps<?string>> {
                 </div>
                 {formInspector.id &&
                     <div className={resourceLocatorStyles.resourceLocatorHistory}>
+                        <Button icon="su-sync" onClick={this.handleButtonClick} skin="link">
+                            {translate('sulu_admin.refresh_url')}
+                        </Button>
                         <ResourceLocatorHistory
                             id={formInspector.id}
                             options={{
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/ResourceLocator.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/ResourceLocator.js
@@ -51,15 +51,12 @@ class ResourceLocator extends React.Component<FieldTypeProps<?string>> {
                 return;
             }
 
-
             const {tags: finishedFieldTags} = formInspector.getSchemaEntryByPath(finishedFieldSchemaPath) || {};
-            if (
-                !finishedFieldTags || !finishedFieldTags.some((tag) => tag.name === PART_TAG)
-            ) {
+            if (!finishedFieldTags || !finishedFieldTags.some((tag) => tag.name === PART_TAG)) {
                 return;
             }
 
-            this.regenerateUrl();
+            this.generateUrl();
         });
     }
 
@@ -68,7 +65,7 @@ class ResourceLocator extends React.Component<FieldTypeProps<?string>> {
         onFinish();
     };
 
-    regenerateUrl = () => {
+    generateUrl = () => {
         const {onChange, fieldTypeOptions, formInspector} = this.props;
         const {generationUrl} = fieldTypeOptions;
 
@@ -102,7 +99,7 @@ class ResourceLocator extends React.Component<FieldTypeProps<?string>> {
     };
 
     handleRegenerateButtonClick = () => {
-        this.regenerateUrl();
+        this.generateUrl();
     };
 
     render() {
@@ -150,10 +147,12 @@ class ResourceLocator extends React.Component<FieldTypeProps<?string>> {
                     />
                 </div>
                 {formInspector.id &&
-                    <div className={resourceLocatorStyles.resourceLocatorHistory}>
-                        <Button icon="su-sync" onClick={this.handleRegenerateButtonClick} skin="link">
-                            {translate('sulu_admin.refresh_url')}
-                        </Button>
+                    <div className={resourceLocatorStyles.resourceLocatorActions}>
+                        {formInspector.formStore.isFieldModified(dataPath) &&
+                            <Button icon="su-sync" onClick={this.handleRegenerateButtonClick} skin="link">
+                                {translate('sulu_admin.refresh_url')}
+                            </Button>
+                        }
                         <ResourceLocatorHistory
                             id={formInspector.id}
                             options={{
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/ResourceLocator.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/ResourceLocator.js
@@ -7,8 +7,8 @@ import ResourceLocatorHistory from '../../../containers/ResourceLocatorHistory';
 import Requester from '../../../services/Requester';
 import type {FieldTypeProps} from '../../../types';
 import resourceLocatorStyles from './resourceLocator.scss';
-import {translate} from "../../../utils/Translator";
-import Button from "../../../components/Button";
+import {translate} from '../../../utils/Translator';
+import Button from '../../../components/Button';
 
 const PART_TAG = 'sulu.rlp.part';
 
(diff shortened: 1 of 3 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/ResourceLocator.js
```diff
@@ -89,6 +101,13 @@ class ResourceLocator extends React.Component<FieldTypeProps<?string>> {
         onFinish();
     };
 
+    handleButtonClick = () => {
+        const {onFinish} = this.props;
+        this.forceRefresh = true;
```
danrot, line 106: Why the indirection by using a separate `forceRefresh` variable? I would just extract the part generating the URL from the `addFinishHandler` above into a separate function, and call this function here.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/ResourceLocator.js
```diff
@@ -89,6 +101,13 @@ class ResourceLocator extends React.Component<FieldTypeProps<?string>> {
         onFinish();
     };
 
+    handleButtonClick = () => {
```
danrot, line 104: Not sure about the naming yet - I would prefer "Regenerate URL" over "Refresh URL") - but depending on what we choose I would call this method `handleRegenerateButtonClick` or `handleRefreshButtonClick`.

danrot:
@cirykpopeye I am also going to talk about that with one of our designers, and will let you know then how that should be implemented.

danrot:
@cirykpopeye Just talked to our designer, since there are two buttons now, we would move them into a separate line. The buttons should have a margin of `20px` between each other, and always be visible.

But if their functionality does not make any sense, they should be disabled instead of hidden now. That means the buttons should be disabled in the following cases:

- "Show history" is disabled on the add form
- "Refresh URL" is disabled if the form has not changed yet

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/ResourceLocator.js
cirykpopeye: I now seperated this into a new function, but I transformed "generateUrl" in its own function to keep it clean.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/ResourceLocator.js
cirykpopeye: I renamed it to handleRegenerateButtonClick for now, since generate is used within Sulu related to the URL.

cirykpopeye:
> @cirykpopeye Just talked to our designer, since there are two buttons now, we would move them into a separate line. The buttons should have a margin of `20px` between each other, and always be visible.
> 
> But if their functionality does not make any sense, they should be disabled instead of hidden now. That means the buttons should be disabled in the following cases:
> 
> * "Show history" is disabled on the add form
> * "Refresh URL" is disabled if the form has not changed yet

Both buttons are currently not shown on the add form. How can I intercept wether there's change on the edit form? ( Should I disable the button, or hide the button?)

What's the Sulu way of adding those 20px?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/ResourceLocator.js
```diff
@@ -49,38 +51,15 @@ class ResourceLocator extends React.Component<FieldTypeProps<?string>> {
...
-                    if (path.startsWith('/')) {
-                        return [path.substr(1), value];
-                    }
-
-                    return [path, value];
-                });
-
-            if (partEntries.length === 0) {
+            const {tags: finishedFieldTags} = formInspector.getSchemaEntryByPath(finishedFieldSchemaPath) || {};
+            if (
```
danrot, line 56: Can you format it the same way it was before? Then it is easier to review for me :slightly_smiling_face:

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/ResourceLocator.js
```diff
@@ -89,6 +68,43 @@ class ResourceLocator extends React.Component<FieldTypeProps<?string>> {
         onFinish();
     };
 
+    regenerateUrl = () => {
```
danrot, line 71: Would call it just `generateUrl`, because when being called for the first time, it is not a regeneration.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/ResourceLocator.js
```diff
@@ -135,6 +151,9 @@ class ResourceLocator extends React.Component<FieldTypeProps<?string>> {
                 </div>
                 {formInspector.id &&
```
danrot, line 152: This causes the elements not to be visible on the add form. You should be able to use the same conditional in the `disabled` property of below buttons.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/ResourceLocator.js
```diff
@@ -135,6 +151,9 @@ class ResourceLocator extends React.Component<FieldTypeProps<?string>> {
...
                     <div className={resourceLocatorStyles.resourceLocatorHistory}>
```
danrot, line 153: This causes this stuff to be styled. There should be a `resourceLocator.scss` file in the same folder, which contains a class named like this property here. Should probably be renamed, because it is not only the `resourceLocatorHistory` anymore.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/ResourceLocator.js
cirykpopeye: Atm both buttons aren't shown in the add form, which i think is a good thing? 
Imo the refresh url button should be disabled when there's not change to the rpl tag fields
history should only be enabled when there's history, or would that be too much to fetch?

How can I detect change on a particular field? In the addFinishFieldHandler? I tried {formInspector.formStore.isFieldModified(dataPath)} but that doesn't seem to do the trick.

danrot: I would say the difference between the history and the regenerate button is that the history cannot have any values in the add form, therefore I think it makes sense to be hidden. But at the same time the regenerate button might be of value also in the add form.

If you edit the URL field directly, the form does not update its value anymore, even if one of the fields with the `sulu.rlp.part` tag is changed. So after having edited the URL directly it would make sense to have the regenerate button also available in the add form.

danrot:
> > @cirykpopeye Just talked to our designer, since there are two buttons now, we would move them into a separate line. The buttons should have a margin of `20px` between each other, and always be visible.
> > But if their functionality does not make any sense, they should be disabled instead of hidden now. That means the buttons should be disabled in the following cases:
> > 
> > * "Show history" is disabled on the add form
> > * "Refresh URL" is disabled if the form has not changed yet
> 
> Both buttons are currently not shown on the add form. How can I intercept wether there's change on the edit form? ( Should I disable the button, or hide the button?)

Using `formInspector.isFieldModified` should return if the field has already modified, as you can see in https://github.com/sulu/sulu/blob/master/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/ResourceLocator.js#L48

> What's the Sulu way of adding those 20px?

You have to make use of [ŧhat import](https://github.com/sulu/sulu/blob/master/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/ResourceLocator.js#L9). In Sulu you can import a CSS file, and the classes are then available via an object. Just search for the occurence of that import in this file, and you see how to use them :slightly_smiling_face:

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/ResourceLocator.js
```diff
@@ -19,7 +21,11 @@ class ResourceLocator extends React.Component<FieldTypeProps<?string>> {
     constructor(props: FieldTypeProps<?string>) {
         super(props);
 
-        const {dataPath, onChange, fieldTypeOptions, formInspector, value} = this.props;
+        this.state = {
```
danrot, line 24: We don't use react state, because it is async and causes some issues... Use a MobX `@observable` variable as for the `mode` above. Then MobX will take care of rerendering the component, when you assign a new value to that variable.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/ResourceLocator.js
```diff
@@ -114,7 +135,7 @@ class ResourceLocator extends React.Component<FieldTypeProps<?string>> {
             disabled,
             formInspector,
             onChange,
-            value,
+            value
```
danrot, line 138: We have these trailing commas on purpose, because they'll result in smaller diffs when more values are added in this object.

src/Sulu/Bundle/AdminBundle/Resources/translations/admin.en.json
```diff
@@ -149,5 +149,6 @@
     "sulu_admin.form_save_server_error": "There was an error when trying to save the form",
     "sulu_admin.form_used_by": "This form is currently used by",
     "sulu_admin.from": "from",
-    "sulu_admin.until": "until"
+    "sulu_admin.until": "until",
+    "sulu_admin.refresh_url": "Refresh URL"
```
danrot, line 153: Shouldn't that be `regenerate_url` and "Regenerate URL"?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/ResourceLocator.js
```diff
@@ -7,6 +7,8 @@ import ResourceLocatorHistory from '../../../containers/ResourceLocatorHistory';
 import Requester from '../../../services/Requester';
 import type {FieldTypeProps} from '../../../types';
 import resourceLocatorStyles from './resourceLocator.scss';
+import {translate} from "../../../utils/Translator";
+import Button from "../../../components/Button";
```
danrot, line 11: Regarding the two previous lines:
- We use single quote instead of double quotes
- The imports using `..` should be before imports with `.`

These are probably some of the reasons CI is currently failing.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/ResourceLocator.js
```diff
@@ -134,7 +155,12 @@ class ResourceLocator extends React.Component<FieldTypeProps<?string>> {
                     />
                 </div>
                 {formInspector.id &&
-                    <div className={resourceLocatorStyles.resourceLocatorHistory}>
+                    <div className={resourceLocatorStyles.resourceLocatorActions}>
+                        {this.state.tagValuesChanged &&
```
danrot, line 159: This will still hide the button, although it should be only disabled.

danrot (changes requested):

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/ResourceLocator.js
```diff
@@ -15,11 +17,12 @@ const HOMEPAGE_RESOURCE_LOCATOR = '/';
 @observer
 class ResourceLocator extends React.Component<FieldTypeProps<?string>> {
     @observable mode: string;
+    @observable tagValuesChanged: string;
```
danrot, line 20: That value is not a string.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/ResourceLocator.js
```diff
@@ -40,47 +43,25 @@ class ResourceLocator extends React.Component<FieldTypeProps<?string>> {
...
 
-        formInspector.addFinishFieldHandler((finishedFieldDataPath, finishedFieldSchemaPath) => {
-            if (value !== undefined) {
-                return;
-            }
-
-            if (formInspector.isFieldModified(dataPath)) {
-                return;
-            }
+        this.tagValuesChanged = false;
```
danrot, line 46: You can also assign a default value when defining this one at the top:

```javascript
@observable tagValuesChanged: boolean = false;
```

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/ResourceLocator.js
danrot: This issue still seems to be open. As explained in my previous comment, the regenerate button makes sense in the add form, while the history does not.

danrot (changes requested):
This probably also needs a rebase, and I can't imagine that the tests are still working, resp. there should be another one for the new feature. Do you think you can do that? Otherwise I can take some time and write the test.

danrot:
Closing in favour of #5551. Still thank you a lot for starting this :slightly_smiling_face:

PR #5346, opened 2020-06-15
Add request-parameters to selection-field-filter-type

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | none
| Related issues/PRs | none
| License | MIT
| Documentation PR | none

#### What's in this PR?

This PR adds `requestParameters` to the SelectionFieldFilterType to allow passing parameters to the api.

#### Why?

Is needed to set the limit of the APIs when loading the entities

#### Example Usage

~~~xml
<property name="eventId" visibility="never">
    <field-name>eventId</field-name>
    <entity-name>App\Entity\Show</entity-name>

    <filter type="selection">
        <params>
            <param name="displayProperty" value="title"/>
            <param name="resourceKey" value="events"/>
            <param name="type" value="select"/>
            <param name="requestParameters" type="collection">
                <param name="limit" value="99"/>
            </param>
        </params>
    </filter>
</property>
~~~

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/fieldFilterTypes/SelectionFieldFilterType.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/fieldFilterTypes/SelectionFieldFilterType.js
@@ -30,7 +30,7 @@ class SelectionFieldFilterType extends AbstractFieldFilterType<?Array<string | n
     ) {
         super(onChange, parameters, value);
 
-        this.selectionStore = new MultiSelectionStore(this.resourceKey, []);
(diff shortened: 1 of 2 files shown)


alexander-schranz:
I think we should fix this by generally by adding `limit=''` to get all or something like we did it for the roles dropdown.
https://github.com/sulu/sulu/pull/5707

alexander-schranz:
Or maybe #5707 did fix this issue 🤔 @wachterjohannes could you test it?

wachterjohannes:
@alexander-schranz i am not sure if we have used this somewhere - but yes the given PR seems to fix this! will close this PR but keep the branch if there is a project out there 😂

PR #5348, opened 2020-06-15
Show publish indicator on list

### LINKED ISSUE #5225: Page table does not display indicator for draft pages
| Q | A
| --- | ---
| Sulu Version | 2.0.6

# Actual Behavior

When I change the display mode of my page list to `tree_table`, I no longer see any indicators about the status of my pages. This makes it hard to see which pages are published at the moment.

current behaviour

# Expected Behavior

The `tree_table` adapter should display the same indicators as the `column_list` adapter.

# Steps to Reproduce

1. Create a new page
2. Save the page as draft only
3. Navigate to the page list
4. Switch the display mode of the list to `tree_table`

luca-rath:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #5225
| License | MIT

#### What's in this PR?

Show publish indicator on list

#### Why?

Because it's displayed in the `column_list` adapter, but not in the `table` and `tree_table` adapters.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/adapters/AbstractTableAdapter.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/adapters/AbstractTableAdapter.js
@@ -45,21 +45,25 @@ export default class AbstractTableAdapter extends AbstractAdapter {
                     indicators.push(
                         <GhostIndicator
                             className={abstractTableAdapterStyles.ghostIndicator}
+                            key="ghost"
                             locale={item.ghostLocale}
                         />
                     );
                 } else {
-                    const draft = item.publishedState === undefined ? false : !item.publishedState;
-                    const published = item.published === undefined ? false : !!item.published;
+                    if (item.publishedState !== undefined || item.published !== undefined) {
+                        const draft = !item.publishedState;
+                        const published = !!item.published;
 
-                    if (draft || !published) {
-                        indicators.push(
-                            <PublishIndicator
-                                className={abstractTableAdapterStyles.publishIndicator}
-                                draft={draft}
-                                published={published}
-                            />
-                        );
+                        if (draft || !published) {
+                            indicators.push(
+                                <PublishIndicator
+                                    className={abstractTableAdapterStyles.publishIndicator}
+                                    draft={draft}
+                                    key="publish"
+                                    published={published}
+                                />
+                            );
+                        }
                     }
                 }
             }
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/adapters/AbstractTableAdapter.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/adapters/AbstractTableAdapter.js
@@ -41,7 +41,9 @@ export default class AbstractTableAdapter extends AbstractAdapter {
 
             return (
                 <Table.Cell key={item.id + schemaKey}>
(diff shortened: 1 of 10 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/adapters/AbstractTableAdapter.js
```diff
@@ -40,6 +41,15 @@ export default class AbstractTableAdapter extends AbstractAdapter {
 
             return (
                 <Table.Cell key={item.id + schemaKey}>
+                    {index === 0 && !item.ghostLocale
```
danrot, line 44: I feel like we should make sure, that we are always referencing the `index` of 0, shouldn't we? Would imagine that somehow like this:

```javascript
<Table.Cell>
    {index === 0 &&
        item.ghostLocale
            ? <GhostIndicator ... />
            : (draft || ! published)
                ? <PublishIndicator />
                : null
    }
</Table.Cell>
```

(Didn't implement it all way through)

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/adapters/abstractTableAdapter.scss
```diff
@@ -1,3 +1,7 @@
+.publish-indicator {
```
danrot, line 1: Should that be exactly the same as in the `ghost-indicator` class? Why not using the selector `.publish-indicator, .ghost-indicator` then?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/tests/adapters/TableAdapter.test.js
```diff
@@ -68,6 +68,84 @@ test('Render data with schema', () => {
     expect(tableAdapter).toMatchSnapshot();
 });
 
+test('Render different kind of data with schema', () => {
```
danrot, line 71: Can't we integrate that test in the above one? I've been trying to reduce the amount of snapshot tests lately, because they also have downsides if there are too many (nobody checks snapshots anymore, it is not completely clear what is being tested here, ...)

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/tests/adapters/TreeTableAdapter.test.js
```diff
@@ -78,6 +78,114 @@ test('Render data with schema', () => {
     expect(treeListAdapter).toMatchSnapshot();
 });
 
+test('Render different kind of data with schema', () => {
```
danrot, line 81: Same as above, can't we integrate that in the previous test? Also the name `different kind of data` is not really clear...

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/tests/adapters/__snapshots__/TreeTableAdapter.test.js.snap
```diff
@@ -53,7 +53,11 @@ exports[`Render data with pencil button when onItemEdit callback is passed 1`] =
         >
           <div
             class="cellContent"
-          />
+          >
+            <div
+              class="publishIndicator publishIndicator"
```
danrot, line 58: That's odd... If this indicator is not shown in the real application, these DOM elements also don't exist... Can you find out what is wrong here? Maybe the test data you are using is not what is actually returned by the API?

PR #5352, opened 2020-06-16
Show icons in list

luca-rath:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Add functionality to show icons in list

#### Why?

SuluAutomationBundle needs to show icons for task status.

#### Example Usage

~~~xml
<property name="status">
    <field-name>status</field-name>
    <entity-name>Task\TaskBundle\Entity\TaskExecution</entity-name>

    <transformer type="icon">
        <params>
            <param name="mapping" type="collection">
                <param name="planned" value="su-clock"/>
                <param name="running" value="su-process"/>
                <param name="completed" type="collection">
                    <param name="icon" value="su-check-circle"/>
                    <param name="color" value="#6ac86b"/>
                </param>
                <param name="failed" type="collection">
                    <param name="icon" value="su-ban"/>
                    <param name="color" value="#cf3939"/>
                </param>
            </param>
        </params>
    </transformer>
</property>
~~~

The above configuration would map the states `planned`, `running`, `completed` and `failed` to the following icons.

task-list


luca-rath:
@sulu/core-team what do you thing about the configuration?

danrot:
The suggested configuration looks good to me :+1:

alexander-schranz:
Same from my side 👍 . I also see other usecases for the params. We basically had in every project we currently did build have a multiple custom translation fielddescriptors which could with params be implemented this way:

```xml
<property name="status" type="translation">
    <field-name>status</field-name>
    <entity-name>Task\TaskBundle\Entity\TaskExecution</entity-name>

    <params>
        <param name="prefix" value="app.task_prefix."/>
    </params>
</property>
```

Which would then use e.g. `app.task_prefix.planned` as translation key.

alexander-schranz:
@danrot We needed a customJS build which did register multiple TranslationKeyFieldTransformer Example from a project:

```js
listFieldTransformerRegistry.add('show_place', new TranslationKeyFieldTransformer('app.show_place.'));
listFieldTransformerRegistry.add('ticket_place', new TranslationKeyFieldTransformer('app.ticket_place.'));
listFieldTransformerRegistry.add('order_place', new TranslationKeyFieldTransformer('app.order_place.'));
listFieldTransformerRegistry.add('promotion_type', new TranslationKeyFieldTransformer('app.promotion_type.'));
```

@danrot I don't see that this would happen but I'm also fine if we create a map:

```xml
<property name="status" type="translation">
    <field-name>status</field-name>
    <entity-name>Task\TaskBundle\Entity\TaskExecution</entity-name>

    <param name="translations" type="collection">
        <param name="planned" value="app.task_status.planned"/>
    </param>
</property>
```

danrot:
It's not only about deleting, also about changing a translation, and trying to find out where it is used. I think a code base should be as searchable as possible to find occurrences. But generally I still like your idea, but I would go with the map.

danrot:
While reading the code I realized that the suggested format is a little bit strange. If you have a look at the filters, they are defined a little bit different:

```xml
        <identity-property name="accountId" visibility="never" translation="sulu_contact.organization">
            <field-name>account</field-name>
            <entity-name>SuluContactBundle:AccountContact</entity-name>

            <joins ref="accountContact"/>

            <filter type="selection">
                <params>
                    <param name="displayProperty" value="name" />
                    <param name="resourceKey" value="accounts" />
                </params>
            </filter>
        </identity-property>
```

The `params` for the filters are defined in the `filter` tag, but the `params` for the type are defined on the top level. It feels strange to me that they are handled in a different way. I get that we have the `type` attribute on the `property` tag, but it still does not feel that connected. Additionally, what if we need some more params, which might conflict with each other? Therefore I think something like the following would make sense:

```xml
<property name="status" type="icon">
    <field-name>status</field-name>
    <entity-name>Task\TaskBundle\Entity\TaskExecution</entity-name>

    <type>
        <params>
            <param name="icons" type="collection">
                <param name="planned" value="su-clock"/>
                <param name="running" value="su-process"/>
                <param name="completed" type="collection">
                    <param name="icon" value="su-check-circle"/>
                    <param name="color" value="#6ac86b"/>
                </param>
                <param name="failed" type="collection">
                    <param name="icon" value="su-ban"/>
                    <param name="color" value="#cf3939"/>
                </param>
            </param>
        </params>
    </type>
</property>
```

In order to be completely consistent we would have to remove the `type` attribute from the `property` tag and move it to the `type` tag, but that would be a BC break. But the solution I just mentioned would be at least a little bit more consistent.

@sulu/core-team WDYT?

src/Sulu/Component/Rest/ListBuilder/Metadata/ListXmlLoader.php
```diff
@@ -126,12 +126,23 @@ private function loadPropertyMetadata(\DOMXPath $xpath, \DOMNode $propertyNode)
         $propertyMetadata->setSortable(
             XmlUtil::getBooleanValueFromXPath('@sortable', $xpath, $propertyNode, true)
         );
+
+        $parameters = $xpath->query('x:params', $propertyNode);
+        if (\count($parameters) > 0) {
+            $propertyMetadata->setParameters(
+                $this->getParameters(
+                    $xpath,
+                    $parameters->item(0) // There can only be one filter node
```
danrot, line 135: Wrong comment, "There can onl ybe one params node", resp. we should wait on the decision about the suggestion I just posted.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/fieldTransformers/IconFieldTransformer.js
```diff
@@ -0,0 +1,72 @@
...
+        }
+
+        const iconConfig = icons[value];
+        if (!iconConfig) {
+            return value;
+        }
+
+        let icon = null;
+        let color = null;
+        switch (typeof iconConfig) {
```
danrot, line 33: Don't you think that code would be easier to read (and we could make more use of `const`) if we move that into two separate functions? Talking about something like that:

```javascript
if (typeof iconConfig === 'object) {
    return transformObjectConfig(iconConfig);
}

if (typeof iconConfig === 'string') {
    return transformStringConfig(iconConfig);
}

log.error('...');

return null;
```

danrot (changes requested):

alexander-schranz:
@danrot I don't think that a `type` is good, it adds nothing and would be confused with `block` -> `types` -> `type`, and then you would think you could configure multiple types.

danrot:
In general I don't really agree with your arguments. Words also have meaning in different contexts, especially generic ones like `type`. If that's a problem, we would make up a lot of words on our own, and there are probably a lot of cases like this. E.g. the `property` tag is used in list and form configuration, the term type is used for block type and form types, etc.

I also think that it adds value. Consider an example with a filter and a type configuration:

```xml
<property name="status" type="icon">
    <field-name>status</field-name>
    <entity-name>Task\TaskBundle\Entity\TaskExecution</entity-name>

    <filter type="selection">
        <params>
            <param name="displayProperty" value="name" />
            <param name="resourceKey" value="accounts" />
        </params>
    </filter>

    <params>
        <param name="icons" type="collection">
            <param name="planned" value="su-clock"/>
            <param name="running" value="su-process"/>
            <param name="completed" type="collection">
                <param name="icon" value="su-check-circle"/>
                <param name="color" value="#6ac86b"/>
            </param>
            <param name="failed" type="collection">
                <param name="icon" value="su-ban"/>
                <param name="color" value="#cf3939"/>
            </param>
        </param>
    </params>
</property>
```

(The example does not really make sense, I just merged the two above ones)

This is confusing for multiple reasons:

1. Why are there params specific to the filters and generic params?
1. What are the generic params used for? There is no obvious connection to the `type` attribute. That probably only makes sense to us right now, because we are introducing them together.
1. Since we also have not considered that use case when introducing the params for filters, we can't exclude the possibility that we need another type of parameters in the future.

However, in that specific use case I think we could make rename `type` to `transformer`, since it better describes what it is doing.

If I would be able to build that again, I would come up with one of two different options (using `transformer` instead of `type` already):

1. **Only have a single `params` node**

Would also mean that we should probably get rid of the `filter` node. The drawback is that the params are shared, and if both the filter and type required the same param we might run into troubles. Could be fixed by introducing a `transformer` and `filter` top level param, but that is not really intuitive, and would still have to be handled in the respective JavaScript components.

```xml
<property name="status" filter="selection" transformer="icon">
    <field-name>status</field-name>
    <entity-name>Task\TaskBundle\Entity\TaskExecution</entity-name>

    <params>
        <param name="filter" type="collection">
            <param name="displayProperty" value="name" />
            <param name="resourceKey" value="accounts" />
        </param>
        <param name="transformer" type="collection">
            <param name="planned" value="su-clock"/>
            <param name="running" value="su-process"/>
            <param name="completed" type="collection">
                <param name="icon" value="su-check-circle"/>
                <param name="color" value="#6ac86b"/>
            </param>
            <param name="failed" type="collection">
                <param name="icon" value="su-ban"/>
                <param name="color" value="#cf3939"/>
            </param>
        </param>
    </params>
</property>
```

2. **Have separate params as already suggested**

My favourite, no naming conflicts and clear isolation of the parameters, no questions where these parameters are used later. Consistency as an additional bonus.

```xml
<property name="status">
    <field-name>status</field-name>
    <entity-name>Task\TaskBundle\Entity\TaskExecution</entity-name>

    <filter type="selection">
        <params>
            <param name="displayProperty" value="name" />
            <param name="resourceKey" value="accounts" />
        </params>
    </filter>

    <transformer type="icon">
        <params>
            <param name="icons" type="collection">
                <param name="planned" value="su-clock"/>
                <param name="running" value="su-process"/>
                <param name="completed" type="collection">
                    <param name="icon" value="su-check-circle"/>
                    <param name="color" value="#6ac86b"/>
                </param>
                <param name="failed" type="collection">
                    <param name="icon" value="su-ban"/>
                    <param name="color" value="#cf3939"/>
                </param>
            </param>
        </params>
    </transformer>
</property>
```

I would also be fine with building the latter variant, and keeping the current `type` attribute just for BC reasons, and throw a deprecation when it is used. But the current suggestion combines the downsides of both variants, with no obvious advantages to me.

nnatter:
I partly agree with @danrot arguments and think the second suggestion looks good. But I am not sure if the advantages are big enough to change this now. Also I think it would be nice to keep things similar to the form configuration. So a simple `<params>` would be okay for me too.

danrot:
> I partly agree with @danrot arguments and think the second suggestion looks good. But I am not sure if the advantages are big enough to change this now. Also I think it would be nice to keep things similar to the form configuration. So a simple `<params>` would be okay for me too.

The generic `params` are introduced in this PR; they do not exist currently. So it is not about changing what is already here, that's a very important difference to me.

Another important difference to the form configuration is, that in the form configuration only one type of parameters exist, and it is clearer that they belong to the entire field.

So even if we say that we do not want to invest the time to deprecate the current `type` parameter (although I don't think that this is a lot of work), I would still introduce a separate a `transformer` tag without attributes around the new `params`.

danrot:
After a short call with @chirimoya @luca-rath @alexander-schranz and @nnatter we have decided to go with the following variant (basically only renamed the `icons` parameter to `mapping`).

```xml
<property name="status">
    <field-name>status</field-name>
    <entity-name>Task\TaskBundle\Entity\TaskExecution</entity-name>

    <filter type="selection">
        <params>
            <param name="displayProperty" value="name" />
            <param name="resourceKey" value="accounts" />
        </params>
    </filter>

    <transformer type="icon">
        <params>
            <param name="mapping" type="collection">
                <param name="planned" value="su-clock"/>
                <param name="running" value="su-process"/>
                <param name="completed" type="collection">
                    <param name="icon" value="su-check-circle"/>
                    <param name="color" value="#6ac86b"/>
                </param>
                <param name="failed" type="collection">
                    <param name="icon" value="su-ban"/>
                    <param name="color" value="#cf3939"/>
                </param>
            </param>
        </params>
    </transformer>
</property>
```

The `type` attribute on the `property` tag should continue to work for BC reasons, but should be deprecated.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/tests/adapters/TableAdapter.test.js
```diff
@@ -72,13 +74,81 @@ test('Render data with schema', () => {
...
+        <TableAdapter
+            {...listAdapterDefaultProps}
+            data={data}
+            page={1}
+            pageCount={1}
+            schema={schema}
+        />
+    );
+
+    expect(tableAdapter).toMatchSnapshot();
```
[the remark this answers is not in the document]
luca-rath: Not possible, because I need to mock `listFieldTransformerRegistry.get()` different in this test case.

luca-rath:
I will change `<property type="..."/>` to `<transformer type="..."/>` in an extra pr, if that's okay

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/tests/adapters/TableAdapter.test.js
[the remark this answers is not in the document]
danrot: Can you in that case avoid the snapshot test, and test explicitly what you are looking for? Hard to tell otherwise, and the real sense of the test might get lost in the future.

danrot:
> I will change `<property type="..."/>` to `<transformer type="..."/>` in an extra pr, if that's okay

Actually I would prefer doing that in the same PR. Otherwise I would have to review and test the "wrong" params for no good reason.

luca-rath:
> Actually I would prefer doing that in the same PR. Otherwise I would have to review and test the "wrong" params for no good reason.
> 
> Additionally the resulting code should not be a lot longer than the current one.

okay, just thought it would be easier to review

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/fieldTransformers/IconFieldTransformer.js
```diff
@@ -0,0 +1,79 @@
...
+        if (typeof iconConfig.icon !== 'string') {
+            log.error(`Transformer parameter "mapping/${value}/icon" needs to be of type string.`);
+
+            return null;
+        }
+
+        const icon = iconConfig.icon;
+
+        if (iconConfig.color) {
+            if (typeof iconConfig.color !== 'string') {
```
danrot, line 62: With the above in place we could write the check like this:

```javascript
if (color !== undefined && typeof color !== 'string') {
    log.error('...');

    return null;
}
```

Then the style could be created like this, and we could avoid creating the same jsx twice:

```
const style = {};

if (color) {
    style.color = color;
}

return <Icon style={style} />;
```

src/Sulu/Component/Rest/ListBuilder/Metadata/ListXmlLoader.php
```diff
@@ -126,12 +122,30 @@ private function loadPropertyMetadata(\DOMXPath $xpath, \DOMNode $propertyNode)
         $propertyMetadata->setSortable(
             XmlUtil::getBooleanValueFromXPath('@sortable', $xpath, $propertyNode, true)
         );
+
+        if (null !== $type = XmlUtil::getValueFromXPath('x:transformer/@type', $xpath, $propertyNode)) {
+            $propertyMetadata->setType($type);
+        } elseif (null !== $type = XmlUtil::getValueFromXPath('@type', $xpath, $propertyNode)) {
+            @\trigger_error('Attribute "type" of list property should not be used anymore! Use "<transformer type="..."/>" inside of property instead.');
```
danrot, line 129: Is there a test covering this case as well?

danrot (changes requested):

danrot (approved):
@luca-rath Please don't forget to add the transformer for the `type` attribute of the media list.

luca-rath:
> @luca-rath Please don't forget to add the transformer for the `type` attribute of the media list.

done in #5369

PR #5361, opened 2020-06-24
Add localizations variable to twig

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | #5277 
| License | MIT

#### What's in this PR?

This PR introduces a `localizations` twig parameter, which contains all information the localization and also contains the URL for the current page in that localization.

#### Why?

The new `localizations` parameter should be used instead of the deprecated `urls` parameter. This has been introduced because `urls` doesn't say what other kind of URLs this variable holds. A `Urls` suffix has been avoided, because this variable contains more information than just the URL.

The reason for this was #5277, which introduces segments offering another kind of alternative URL. Having `urls` and `segmentUrls` seemed confusing, and we also needed to pass more information about the segments, so we wanted to make that consistent.

#### Example Usage

~~~jinja
{# Before #}
<ul>
    {% for locale, url in urls %}
        <li>
            <a href="{{ sulu_content_path(url, request.webspaceKey, locale) }}">{{ locale }}</a>
        </li>
    {% endfor %}
</ul>

{# After #}
<ul>
    {% for localization in localizations %}
        <li>
            <a href="{{ localization.url }}">{{ localization.locale }}</a>
        </li>
    {% endfor %}
</ul>
~~~

#### BC Breaks/Deprecations

The `urls` twig variable has been deprecated.

#### To Do

- [ ] Create skeleton PR

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -2,6 +2,33 @@
 
 ## dev-master
 
+### Deprecated urls variable in twig
+
+The `urls` twig variable has been deprecated in favour of the `localizations` variable. So the code should be adapted
+as shown in the following snippet:
+
+```jinja
+{# Before #}
+<ul>
+    {% for locale, url in urls %}
+        <li>
+            <a href="{{ sulu_content_path(url, request.webspaceKey, locale) }}">{{ locale }}</a>
+        </li>
+    {% endfor %}
+</ul>
+
+{# After #}
+<ul>
+    {% for localization in localizations %}
+        <li>
+            <a href="{{ localization.url }}">{{ localization.locale }}</a>
+        </li>
+    {% endfor %}
+</ul>
+```
+
+### HTTP Cache environment handling
+
 The `sulu_http_cache` configuration behaved differently based on the configured environment. Since this behavior was
 causing some configuration to be ignored in the `dev` and `test` environment, it was very hard to understand. Therefore
 we removed this behavior. In order to imitate the old behavior, the `sulu_http_cache` has to be configured
--- a/templates/base.html.twig
+++ b/templates/base.html.twig
@@ -34,6 +34,16 @@
                     {% endfor %}
                 </ul>
             </nav>
+
+            <nav>
+                <ul>
+                    {% for localization in localizations %}
+                        <li>
+                            <a href="{{ localization.url }}">{{ localization.locale }}</a>
+                        </li>
+                    {% endfor %}
+                </ul>
+            </nav>
         {% endblock %}
     </header>
 
--- a/config/packages/sulu_website.yaml
+++ b/config/packages/sulu_website.yaml
@@ -1,0 +1,4 @@
+sulu_website:
+    twig:
+        attributes:
+            urls: false
--- a/config/webspaces/sulu-blog.xml
+++ b/config/webspaces/sulu-blog.xml
@@ -39,22 +39,22 @@
             <environments>
                 <environment type="prod">
                     <urls>
-                        <url language="en">{host}</url>
+                        <url language="en">{host}/blog</url>
                     </urls>
                 </environment>
                 <environment type="stage">
                     <urls>
-                        <url language="en">{host}</url>
+                        <url language="en">{host}/blog</url>
                     </urls>
                 </environment>
                 <environment type="test">
                     <urls>
-                        <url language="en">{host}</url>
+                        <url language="en">{host}/blog</url>
                     </urls>
                 </environment>
                 <environment type="dev">
                     <urls>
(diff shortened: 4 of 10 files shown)


UPGRADE.md
```diff
@@ -2,6 +2,33 @@
 
 ## dev-master
 
+### Deprecated urls variable in twig
+
+The `urls` twig variable has been deprecated in favour of the `localizations` variable. So the code should be adapted
+as shown in the following snippet:
+
+```jinja
```
alexander-schranz, line 10: `twig`

src/Sulu/Bundle/WebsiteBundle/Resources/views/Extension/seo.html.twig
```diff
@@ -28,8 +28,8 @@
 
 {% set seoCanonical = seo.canonicalUrl|default() %}
 
-{%- if not seoCanonical and shadowBaseLocale and urls[shadowBaseLocale]|default() %}
-    {% set seoCanonical = sulu_content_path(urls[shadowBaseLocale], null, shadowBaseLocale) %}
+{%- if not seoCanonical and shadowBaseLocale and localizations[shadowBaseLocale]|default() %}
```
alexander-schranz, line 31: think here is missing or urls missing:

```
%- if not seoCanonical and shadowBaseLocale and (localizations[shadowBaseLocale]|default() or urls[shadowBaseLocale]|default()) %}
```

alexander-schranz: I think I would do something like this on the top of this file so we can easily remove it in future and throw also a deprecated message in this twig file:

```twig
{% if localizations is not defined %}
    {% deprecated 'The "urls" variable is deprecated, set "localizations" instead in the "seo.html.twig".' %}
    {% set localizations = {} %}
    {% for locale, url in urls %}
        {% set localizations = localizations|merge({
            (locale): {
                locale: locale, 
                url: sulu_content_path(url, null, locale),
            }
        }) %}
    {% endfor %}
{% endif %}
```

src/Sulu/Bundle/WebsiteBundle/Resources/views/Extension/seo.html.twig
```diff
@@ -6,6 +6,19 @@
  # localizations array
 -#}
 
+{% if localizationsa is not defined %}
```
alexander-schranz, line 9: `localizationsa` should be `localizations` you can test it by adding `only` keyword behind the include of the in the base.html.twig e.g.:

```
{% include "@SuluWebsite/Extension/seo.html.twig" with {
            "seo": extension.seo|default([]),
            "content": content|default([]),
            "urls": localizations|default([]),
            "shadowBaseLocale": shadowBaseLocale|default(),
            "defaultLocale": app.request.locale
} only %}
```

templates/base.html.twig
```diff
@@ -34,6 +35,16 @@
                     {% endfor %}
                 </ul>
             </nav>
+
+            <nav>
```
alexander-schranz, line 39: In html roles like `nav`, `main`, ... should only exist once. If there are multiple navigations they need to be labeled with an `aria-label` or `aria-labeledby` tag else accessibility is not given.

templates/base.html.twig
```diff
@@ -9,9 +9,10 @@
         {% include "@SuluWebsite/Extension/seo.html.twig" with {
             "seo": extension.seo|default([]),
             "content": content|default([]),
-            "urls": urls|default([]),
+            "localizations": localizations|default([]),
             "shadowBaseLocale": shadowBaseLocale|default(),
-            "defaultLocale": app.request.locale
+            "defaultLocale": app.request.locale,
+            "urls": urls|default([])
```
alexander-schranz, line 15: Think this case can be removed as localizations is defined urls is not longer accessed when the variable is correct above?

templates/base.html.twig
danrot: I can change it, because they are right next to each other. But I think what you are saying is not true. [MDN says multiple `nav` elements can be on a single page](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nav). Only [`main` should be unique](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/main).

alexander-schranz: Sure from html perspective true but not from accessibility, accessibility tools will fail when having unlabeled multiple navs :)

PR #5362, opened 2020-06-24
Added user factory

JakobTolkemit:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | yes
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT

#### What's in this PR?

This adds a user factory. The `CreateUserCommand` uses this factory instead of creating the user itself.

#### Why?

It's easier to override/decorate the factory instead of the command. The factory can also be used in other services. 

#### Example Usage

You could integrate an Ldap login using this.

#### BC Breaks/Deprecations

`CreateUserCommand` has fewer constructor arguments:
```php
UserRepository $userRepository,
RoleRepositoryInterface $roleRepository,
UserFactoryInterface $userFactory,
array $locales
```

#### To Do

### DIFF
--- a/src/Sulu/Bundle/SecurityBundle/Command/CreateUserCommand.php
+++ b/src/Sulu/Bundle/SecurityBundle/Command/CreateUserCommand.php
@@ -11,34 +11,20 @@
 
 namespace Sulu\Bundle\SecurityBundle\Command;
 
-use Doctrine\ORM\EntityManagerInterface;
-use Sulu\Bundle\ContactBundle\Entity\ContactInterface;
-use Sulu\Bundle\ContactBundle\Entity\ContactRepositoryInterface;
 use Sulu\Bundle\SecurityBundle\Entity\UserRepository;
-use Sulu\Bundle\SecurityBundle\Entity\UserRole;
-use Sulu\Component\Localization\Localization;
-use Sulu\Component\Localization\Manager\LocalizationManagerInterface;
+use Sulu\Bundle\SecurityBundle\Factory\UserFactoryInterface;
 use Sulu\Component\Security\Authentication\RoleInterface;
 use Sulu\Component\Security\Authentication\RoleRepositoryInterface;
-use Sulu\Component\Security\Authentication\SaltGenerator;
-use Sulu\Component\Security\Authentication\UserInterface;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Question\ChoiceQuestion;
 use Symfony\Component\Console\Question\Question;
-use Symfony\Component\Security\Core\Encoder\EncoderFactoryInterface;
-use Symfony\Component\Security\Core\Encoder\PasswordEncoderInterface;
 
 class CreateUserCommand extends Command
 {
     protected static $defaultName = 'sulu:security:user:create';
-
-    /**
-     * @var EntityManagerInterface
-     */
-    private $entityManager;
 
     /**
      * @var UserRepository
@@ -51,24 +37,9 @@ class CreateUserCommand extends Command
     private $roleRepository;
 
     /**
-     * @var ContactRepositoryInterface
+     * @var UserFactoryInterface
      */
-    private $contactRepository;
-
-    /**
-     * @var LocalizationManagerInterface
-     */
-    private $localizationManager;
-
-    /**
-     * @var SaltGenerator
-     */
-    private $saltGenerator;
-
-    /**
-     * @var EncoderFactoryInterface
-     */
-    private $encoderFactory;
+    private $userFactory;
 
     /**
      * @var string[]
@@ -76,28 +47,20 @@ class CreateUserCommand extends Command
     private $locales;
 
     public function __construct(
-        EntityManagerInterface $entityManager,
         UserRepository $userRepository,
         RoleRepositoryInterface $roleRepository,
-        ContactRepositoryInterface $contactRepository,
-        LocalizationManagerInterface $localizationManager,
-        SaltGenerator $saltGenerator,
-        EncoderFactoryInterface $encoderFactory,
+        UserFactoryInterface $userFactory,
         array $locales
     ) {
         parent::__construct();
 
-        $this->entityManager = $entityManager;
         $this->userRepository = $userRepository;
         $this->roleRepository = $roleRepository;
-        $this->contactRepository = $contactRepository;
-        $this->localizationManager = $localizationManager;
-        $this->saltGenerator = $saltGenerator;
-        $this->encoderFactory = $encoderFactory;
         $this->locales = $locales;
+        $this->userFactory = $userFactory;
     }
 
-    protected function configure()
+    protected function configure(): void
     {
         $this->setDescription('Create a user.')
             ->setDefinition(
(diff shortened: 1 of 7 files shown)


src/Sulu/Bundle/SecurityBundle/Command/CreateUserCommand.php
```diff
@@ -143,7 +96,7 @@ protected function execute(InputInterface $input, OutputInterface $output)
             return 1;
         }
 
-        if (!\in_array($locale, $this->locales)) {
+        if (!\in_array($locale, $this->locales, true)) {
```
danrot, line 99: In order to avoid any uninteded side effects, I would prefer to keep changes to a minimum, so I would not change this in this PR.

src/Sulu/Bundle/SecurityBundle/Command/CreateUserCommand.php
```diff
@@ -216,7 +136,7 @@ protected function interact(InputInterface $input, OutputInterface $output)
         if (!$input->getArgument('username')) {
             $question = new Question('Please choose a username: ');
             $question->setValidator(
-                function($username) use ($userRepository) {
+                static function($username) use ($userRepository) {
```
danrot, line 139: I think instaed of making these functions `static`, I would omit the `use` and write `$this->userRepository`. Have not heard about static anonymous functions before, so that might confuse some people.

src/Sulu/Bundle/SecurityBundle/Factory/UserFactory.php
```diff
@@ -0,0 +1,152 @@
+<?php
+
+declare(strict_types=1);
```
danrot, line 3: We haven't used strict_types yet in the code base, so I would like to not start with that yet.

src/Sulu/Bundle/SecurityBundle/Factory/UserFactory.php
```diff
@@ -0,0 +1,152 @@
...
+use Sulu\Component\Localization\Manager\LocalizationManagerInterface;
+use Sulu\Component\Security\Authentication\RoleInterface;
+use Sulu\Component\Security\Authentication\RoleRepositoryInterface;
+use Sulu\Component\Security\Authentication\SaltGenerator;
+use Sulu\Component\Security\Authentication\UserInterface;
+use Sulu\Component\Security\Authentication\UserRepositoryInterface;
+use Symfony\Component\Security\Core\Encoder\EncoderFactoryInterface;
+use Webmozart\Assert\Assert;
+
+final class UserFactory implements UserFactoryInterface
```
danrot, line 29: Although I would be in favour of making more classes `final`, we have not used that a lot until now.

@sulu/core-team Should we start using `final` classes more often?

src/Sulu/Bundle/SecurityBundle/Factory/UserFactory.php
```diff
@@ -0,0 +1,152 @@
...
+{
+    /** @var EntityManagerInterface */
```
danrot, line 31: Until now we have written all these annotations in three lines:

```php
/**
 * @var EntityManagerInterface
 */
```

src/Sulu/Bundle/SecurityBundle/Factory/UserFactory.php
```diff
@@ -0,0 +1,152 @@
...
+    public function create(
+        string $username,
+        string $firstName,
+        string $lastName,
+        string $email,
+        string $locale,
+        string $password,
+        string $roleName = 'User'
+    ): UserInterface {
+        $existing = $this->userRepository->findOneBy(['username' => $username]);
```
danrot, line 79: `$existing` is a weird variable name IMO, because it does not say what kind of value it holds. Why not just call it `$user`?

src/Sulu/Bundle/SecurityBundle/Factory/UserFactory.php
```diff
@@ -0,0 +1,152 @@
...
+            return $existing;
+        }
+
+        $locales = \array_keys($this->localizationManager->getLocalizations());
+
+        return $this->createUser($firstName, $lastName, $email, $username, $password, $locale, $roleName, $locales);
+    }
+
+    /**
+     * Generates a random salt for the password.
```
danrot, line 90: We only add comments like this if they add value. This comment more or less repeats what the function name already says, therefore I would remove it.

src/Sulu/Bundle/SecurityBundle/Factory/UserFactory.php
```diff
@@ -0,0 +1,152 @@
...
+     */
+    private function generateSalt(): string
+    {
+        return $this->saltGenerator->getRandomSalt();
+    }
+
+    /**
+     * Encodes the given password, for the given password, with he given salt and returns the result.
```
danrot, line 98: Would also remove that entire comment, and add correct type hints to the function.

src/Sulu/Bundle/SecurityBundle/Factory/UserFactory.php
```diff
@@ -0,0 +1,152 @@
...
+        $user->setContact($contact);
+        $user->setUsername($username);
+        $user->setSalt($this->generateSalt());
+        $user->setPassword($this->encodePassword($user, $password, $user->getSalt()));
+        $user->setLocale($locale);
+        $user->setEmail($email);
+
+        /** @var RoleInterface|null $role */
+        $role = $this->roleRepository->findOneBy(['name' => $roleName]);
+        Assert::notNull($role);
```
danrot, line 139: `Webmozart/Assert` is not part of our `composer.json`, therefore it is not safe to use it here. Is it necessary at all? Doesn't `findOneBy` throw an exception if no row is returned at all?

src/Sulu/Bundle/SecurityBundle/Factory/UserFactoryInterface.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
+declare(strict_types=1);
```
danrot, line 3: Same as above, not used until now.

src/Sulu/Bundle/SecurityBundle/Factory/UserFactory.php
JakobTolkemit: No, unfortunately `findOneBy` does not throw an exception.

src/Sulu/Bundle/SecurityBundle/Factory/UserFactory.php
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    public function create(
+        string $username,
+        string $firstName,
+        string $lastName,
+        string $email,
+        string $locale,
+        string $password,
+        string $roleName = 'User'
```
danrot, line 89: Is that necessary? Seems a little bit weird, since the name of the role is only placed in the database, and people can change it in the UI. If they do, the default parameter would cause an error...

src/Sulu/Bundle/SecurityBundle/Factory/UserFactoryInterface.php
```diff
@@ -0,0 +1,27 @@
...
+interface UserFactoryInterface
+{
+    public function create(
+        string $username,
+        string $firstName,
+        string $lastName,
+        string $email,
+        string $locale,
+        string $password,
+        string $roleName = 'User'
```
danrot, line 25: Same as above, would avoid to set the default value to something that might break if we change a value in the database.

danrot (changes requested):

danrot:
The most important question remains...

@alexander-schranz Do you know of many use cases, that have overridden this command? Would you be OK with that change and just a mention in the `UPGRADE.md` file?

nnatter:
Closing this because of inactivity. Thanks a lot for the time you have invested in this!

PR #5367, opened 2020-06-25
Throw errors for reserved property names in blocks

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | #5277 
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR throws an exception when `type` or `settings` is used as a name of a property within a block.

#### Why?

Because this is currently possible, but leads to unintented side effects.


danrot:
It was actually an error with the new test cases... It also loaded the test XML files with the error in a case where it shouldn't. Moved the XML files with errors to a separate `invalid-forms` directory.

PR #5374, opened 2020-06-30
Add port to urls generated by the WebspaceManager

### LINKED ISSUE #5364: Links inside the text editor do not respect http port
| Q | A
| --- | ---
| Sulu Version | 2.1.0
| Browser Version | N/A

# Actual Behavior

I am running sulu via local php server on port 9090.
A link created inside a text field is rendered without the port. For example it renders `https://127.0.0.1/en/test`

# Expected Behavior

The link should be rendered as `https://127.0.0.1:9090/en/test`

# Steps to Reproduce

- Bootstrap a fresh sulu 2.1.0
- Create a new page
- Add a textfield and add a link inside the textfield (an internal link)

# Possible Solutions

Somewhere the port information is lost and needs to be added.

nnatter:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #5364
| License | MIT

#### What's in this PR?

This PR adjusts the `WebspaceManager` to add the port of the current request to generated urls, if the urls match the host of the current request. This fixes the url generation for the following services when serving a website via a non-default port:
* `ContentPathTwigExtension` (port was already added by the service itself before)
* `PageLinkProvider`
* `PagesSitemapProvider`
* `ContentRouteProvider`
* `SitemapTwigExtension`
* `RedirectEnhancer`
* `SeoEnhancer`

#### Why?

Because a lot of developers serve their application via a non-default (eg `8080`) port during development. At the moment, in such a setup, the links generated by the services listed above will not work because they omit the port.

### DIFF
--- a/src/Sulu/Bundle/WebsiteBundle/Tests/Unit/Sulu/Bundle/WebsiteBundle/Twig/ContentPathTwigExtensionTest.php
+++ b/src/Sulu/Bundle/WebsiteBundle/Tests/Unit/Sulu/Bundle/WebsiteBundle/Twig/ContentPathTwigExtensionTest.php
@@ -83,7 +83,6 @@ class ContentPathTwigExtensionTest extends TestCase
     public function testGetContentPath()
     {
         $this->requestAnalyzer->getAttribute('host')->willReturn('www.sulu.io');
-        $this->requestAnalyzer->getAttribute('port')->willReturn(80);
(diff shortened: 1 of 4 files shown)


nnatter:
Pipeline is failing because of an unrelated error 😢

PR #5378, opened 2020-06-30
Make sure ContentMapperSubscriber runs before the RemoveSubscriber

### LINKED ISSUE #5376: Event weights for removing PHPCR nodes can result in the wrong order
| Q | A
| --- | ---
| Sulu Version | 2.1.0

# Actual Behavior

The default installation runs fine. Still our installation for some reason gives a different order after a PHP krsort()

# Expected Behavior

Fixed order of events

# Steps to Reproduce

Sorry I can not really reproduce the behaviour still:

Vanilla Sulu does:

1. StructureRemoveSubscriber 550
2. CustomUrlSubscriber 550
3. ContentMapperSubscriber (PreRemove) 500
4. RemoveSubscriber 500

My Sulu intance does:

1. CustomUrlSubscriber 550
2. StructureRemoveSubscriber 550
3. RemoveSubscriber 500
4. ContentMapperSubscriber (PreRemove) 500

If the RemoveSubscriber is run before the ContentMapperSubscriber the PHPCR node gets status 4 which means it is deleted. The content mapper checks the webspace then getPath which will result in an PHPCR Exception that the node is deleted in jackalope/jackalope/src/Jackalope/Item.php:685.

# Possible Solutions

Reoder the events a bit. I provide a patch soon and we can discuss the weights to use.

kfrohwein:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #5376
| License | MIT

#### What's in this PR?

project/vendor/sulu/sulu/src/Sulu/Component/DocumentManager/Subscriber/Phpcr/RemoveSubscriber.php:46 is set to 500 as well. So I change this weight to 550 and it runs before the PHPCR node is set to "removed".

#### Why?

See #5376

### DIFF
--- a/src/Sulu/Component/Content/Document/Subscriber/Compat/ContentMapperSubscriber.php
+++ b/src/Sulu/Component/Content/Document/Subscriber/Compat/ContentMapperSubscriber.php
@@ -87,7 +87,7 @@ class ContentMapperSubscriber implements EventSubscriberInterface
     {
         return [
             Events::REMOVE => [
-                ['handlePreRemove', 500],
+                ['handlePreRemove', 550],
                 ['handlePostRemove', -100],
             ],
             Events::PERSIST => 'handlePersist',
--- a/src/Sulu/Component/Content/Document/Subscriber/Compat/ContentMapperSubscriber.php
+++ b/src/Sulu/Component/Content/Document/Subscriber/Compat/ContentMapperSubscriber.php
@@ -87,7 +87,7 @@ class ContentMapperSubscriber implements EventSubscriberInterface
     {
         return [
             Events::REMOVE => [
-                ['handlePreRemove', 550],
+                ['handlePreRemove', 510],
                 ['handlePostRemove', -100],
             ],
             Events::PERSIST => 'handlePersist',


src/Sulu/Component/Content/Document/Subscriber/Compat/ContentMapperSubscriber.php
```diff
@@ -87,7 +87,7 @@ public static function getSubscribedEvents()
     {
         return [
             Events::REMOVE => [
-                ['handlePreRemove', 500],
+                ['handlePreRemove', 550],
```
alexander-schranz, line 90: would not set it to the same as the other subscribers else we could run there into the same problem:

```
bin/console sulu:document:subscriber:debug remove
```

Would use in this case 510.

kfrohwein: Sure will do. In my local test it did not make any problems as those subscribes "do things" in their scope. And I wasn't sure if you want to stick to +-50 steps.

kfrohwein:
The Tests still fail because of the Audience targeting bundle. Doesn't seem related thought...

nnatter:
Thanks for the contribution @kfrohwein!
The failing tests are not related to this PR. We will hopefully fix that soon. But could you use `release/1.6` as target branch? As @alexander-schranz already said, it looks like the bug is already present there.
If not, I will create a new PR targeting `release/1.6` in the next days. Hope that is okay for you 🙂

nnatter:
Changed the base branch for you 🙂

alexander-schranz:
@kfrohwein Thank you!

kfrohwein:
Sorry I was sick last week and it seems I didn't see all you wrote  :sweat: 
Thank you!

PR #5380, opened 2020-06-30
Add position name to serialized contact

### LINKED ISSUE #5358: Contact Account Selection only returns ID of position

nnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #5358
| Related issues/PRs | replaces #4895
| License | MIT

#### What's in this PR?

This PR adds a `positionName` property to the serialized representation of the contact entity. This serializer representation is returned by the `ContactAccountSelection` and by the `contacts` smart-content data-provider.

The PR also adjusts the `ContactRepository` to join the respective tables to prevent the lazy-loading of entities during serialization. I tested the adjustments and did not notice any additional queries in the Symfony profiler.

#### Why?

See #5358

### DIFF
--- a/src/Sulu/Bundle/ContactBundle/Api/Contact.php
+++ b/src/Sulu/Bundle/ContactBundle/Api/Contact.php
@@ -227,7 +227,7 @@ class Contact extends ApiWrapper
     /**
      * Get position.
      *
-     * @return string
+     * @return int
      *
      * @VirtualProperty
      * @SerializedName("position")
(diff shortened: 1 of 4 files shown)


alexander-schranz:
Code looks good for me @danrot posted that he wanted to backport this to 1.6 release [here](https://github.com/sulu/sulu/pull/4895#issuecomment-564004213). So I will let him choose if he want still backport this or not.

danrot:
The issue description says that there was a different way of handling this in Sulu 1.6, but it was possible. So I would say now that this is a bug being introduced in Sulu 2.0, so fixing it there should be fine.

PR #5392, opened 2020-07-03
Add link to selected default-snippet in SnippetAreas view

nnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Related issues/PRs | related to #5143
| License | MIT

#### What's in this PR?

This PR adjusts the `SnippetAreas` view to provide a link to to the selected snippet. The `SnippetAreas` view is used to set default snippets for a webspace. 

#### Why?

Because the `Selection`, `SingleSelection`, `SmartContent` and `TeaserSelection` provide a similar link feature. Also, it is quite annoying to search for the name of the selected snippet in the snippet list if there are a lot of snippets 🙂

### DIFF
--- a/src/Sulu/Bundle/SnippetBundle/Resources/js/views/SnippetAreas/SnippetAreas.js
+++ b/src/Sulu/Bundle/SnippetBundle/Resources/js/views/SnippetAreas/SnippetAreas.js
@@ -35,6 +35,13 @@ class SnippetAreas extends React.Component<ViewProps> {
         this.cacheClearToolbarAction = new CacheClearToolbarAction();
     }
 
+    @action handleSnippetClick = (snippetUuid: string) => {
+        const {router, route} = this.props;
+        const {snippetEditView} = route.options;
+
+        router.navigate(snippetEditView, {id: snippetUuid});
+    };
+
     @action handleAddClick = (areaKey: string) => {
         this.openedAreaKey = areaKey;
     };
@@ -97,9 +104,14 @@ class SnippetAreas extends React.Component<ViewProps> {
                                     <Table.Cell>
                                         {defaultUuid
                                             ? <Fragment>
-                                                <div className={snippetAreasStyles.title}>
+                                                <Button
+                                                    className={snippetAreasStyles.titleButton}
+                                                    onClick={this.handleSnippetClick}
+                                                    skin="link"
+                                                    value={defaultUuid}
+                                                >
                                                     {defaultTitle}
-                                                </div>
+                                                </Button>
                                                 <Button
                                                     className={snippetAreasStyles.deleteButton}
                                                     icon="su-trash-alt"
--- a/src/Sulu/Bundle/SnippetBundle/Resources/js/views/SnippetAreas/SnippetAreas.js
+++ b/src/Sulu/Bundle/SnippetBundle/Resources/js/views/SnippetAreas/SnippetAreas.js
@@ -107,7 +107,7 @@ class SnippetAreas extends React.Component<ViewProps> {
                                                 <Button
                                                     className={snippetAreasStyles.titleButton}
                                                     onClick={this.handleSnippetClick}
-                                                    skin="link"
+                                                    skin="text"
                                                     value={defaultUuid}
                                                 >
                                                     {defaultTitle}
--- a/src/Sulu/Bundle/SnippetBundle/Resources/js/views/SnippetAreas/tests/SnippetAreas.test.js
+++ b/src/Sulu/Bundle/SnippetBundle/Resources/js/views/SnippetAreas/tests/SnippetAreas.test.js
@@ -1,15 +1,13 @@
 // @flow
 import React from 'react';
 import {shallow, mount, render} from 'enzyme';
-import Router from 'sulu-admin-bundle/services/Router';
-import Route from 'sulu-admin-bundle/services/Router/Route';
+import {Route, Router} from 'sulu-admin-bundle/services';
 import {findWithHighOrderFunction} from 'sulu-admin-bundle/utils/TestHelper';
 
 jest.mock('sulu-admin-bundle/containers', () => ({
     SingleListOverlay: jest.fn(() => null),
     withToolbar: jest.fn((Component) => Component),
 }));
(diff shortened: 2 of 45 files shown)


alexander-schranz:
For me this should target 2.1 branch as a bugfix, as it was just forgot there :)

alexander-schranz:
I did test it and it did work as expected but it would be great if @danrot could have a look at the code also as he is most familiar with the deep links.

src/Sulu/Bundle/SnippetBundle/Resources/js/views/SnippetAreas/tests/SnippetAreas.test.js
```diff
@@ -1,7 +1,8 @@
 // @flow
 import React from 'react';
 import {shallow, mount, render} from 'enzyme';
-import {Router} from 'sulu-admin-bundle/services';
+import Router from 'sulu-admin-bundle/services/Router';
+import Route from 'sulu-admin-bundle/services/Router/Route';
```
danrot, line 5: Shouldn't `import {Route, Router} from 'sulu-admin-bundle/services';` also work?

src/Sulu/Bundle/SnippetBundle/Resources/js/views/SnippetAreas/SnippetAreas.js
```diff
@@ -97,9 +104,14 @@ class SnippetAreas extends React.Component<ViewProps> {
                                     <Table.Cell>
                                         {defaultUuid
                                             ? <Fragment>
-                                                <div className={snippetAreasStyles.title}>
+                                                <Button
+                                                    className={snippetAreasStyles.titleButton}
+                                                    onClick={this.handleSnippetClick}
+                                                    skin="link"
```
danrot, line 110: I don't think that this results in the desired design, since the other deep links are also shown differently (i.e. without an underline). The other deep links are simply `div` elements with a `button` role, would keep it the same way here for consistency.

src/Sulu/Bundle/SnippetBundle/Resources/js/views/SnippetAreas/tests/SnippetAreas.test.js
```diff
@@ -27,10 +34,7 @@ test('Show loader when loading snippet areas', () => {
     const SnippetAreas = require('../SnippetAreas').default;
     const SnippetAreaStore = require('../stores/SnippetAreaStore');
 
-    const router = new Router({});
-    router.attributes = {
-        webspace: 'sulu',
-    };
```
danrot, line 33: Why are the `attributes` removed in all these tests? Haven't they been necessary before as well?

danrot (changes requested):

src/Sulu/Bundle/SnippetBundle/Resources/js/views/SnippetAreas/SnippetAreas.js
nnatter: the reason i implemented this using a `Button` is that the `react/jsx-no-bind` eslint rule forces me to extract this "simple `<div>` into a separate component. 

this is the second time in a short timespan that this rule complicates the implementation of a simple component. i would really like to enable the `allowArrowFunctions` of this it.

danrot: Personally I don't think that this makes it a lot more complicated, and extracting separate components is most of the time not such a bad idea. Also, as [described in the rule documentation](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md#lists-of-items), that should speed up rendering. I tend to believe that, unless I see a benchmark proofing the opposite. I think we already have talked about this, and this rule would indeed not work anymore with hooks, but since we are not using them yet I would keep this rule for now.

nnatter: do you think its a good idea to extract a component in this case?
regarding performance, i tend to *dont* believe that, unless I see a benchmark proofing the opposite. Also, see https://cdb.reacttraining.com/react-inline-functions-and-performance-bdff784f5578

danrot: TBH, although it is an unpopular opinion, I would do it, because [MobX advises to use small components](https://mobx.js.org/best/react-performance.html#use-many-small-components) (which is MobX specific, so that's for sure not mentioned in that article) and because the rule also avoids code like this:

```javascript
<div
    onClick={() =>{
        methodCall1();
        methodCall2();
        if (someValue == 5) {
            methodCall3();
        } else {
            methodCall4();
            methodCall5();
        }
    }}
/>
```

This code should IMO definitely be extracted into a separate function, and the rule also enforces this.

These two reasons alone would be enough for me to accept the minor inconvenience creating a separate component causes. But if we still decide to not do that anymore I would change that in the eslint config instead of disabling it using comments, because it has about the same effect (I do not want to think or argue at every PR if that is a valid edge case for disabling this rule).

danrot: After some thinking this discussion took maybe a bit an unnecessary turn... I didn't dislike the `Button` solution very much, the only real concern for me was that the styling did not match. But this could be solved by introducing another skin (maybe "text"?) that would display it in the way we want it to be displayed. This way we could keep the rule, and the issue you have mentioned would be solved in many cases. What do you think?

nnatter: Yes, a `text` skin for the `Button` component sounds okay to me. It will also prevent the issues with the `react/jsx-no-bind` for the most simple cases in the future, so thats nice 🙂

danrot (changes requested):
Not really sure why, but your PR also changed the text color of the primary button:

Screenshot from 2020-09-09 10-46-08

danrot (approved):

PR #5395, opened 2020-07-06
Fix API for saving target groups

nnatter:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This PR fixes the API for saving target groups by clearing the entity manager after deserializing the target group entity from the data passed to the API. **This is not a great solution and we should probably not merge this.**

This fixes the `TargetGroupControllerTest::testPutWithRemoveRoleAndWebspaces` test that is failing at the moment. I figured that the error happens when the `jms/serializer-bundle` package is updated from `3.6` to `3.7`. Somehow, after this update, the entity which is deserialized in the `TargetGroupController` is detected as `STATE_MANAGED` by doctrine. Because of this, the `$this->getEntityManager()->merge($targetGroup);` does not do anything.

### DIFF
--- a/src/Sulu/Bundle/AudienceTargetingBundle/Controller/TargetGroupController.php
+++ b/src/Sulu/Bundle/AudienceTargetingBundle/Controller/TargetGroupController.php
@@ -274,6 +274,7 @@ class TargetGroupController extends AbstractRestController implements ClassResou
             $this->targetGroupRepository->getClassName(),
             'json'
         );
+        $this->entityManager->clear();
 
         return $result;
     }


nnatter:
Repeating this for visibility: **This is not a great solution and we should probably not merge this.**

src/Sulu/Bundle/AudienceTargetingBundle/Controller/TargetGroupController.php
```diff
@@ -274,6 +274,7 @@ private function deserializeData($data)
             $this->targetGroupRepository->getClassName(),
             'json'
         );
+        $this->entityManager->clear();
```
alexander-schranz, line 277: I find this very strange that we are calling a setter on a serialize subscriber here:

https://github.com/sulu/sulu/blob/e3318df4ed1a13229dd86d8ff480fb3e56af3219/src/Sulu/Bundle/AudienceTargetingBundle/Serializer/Subscriber/TargetGroupSerializeSubscriber.php#L57

Instead we should add it to serialization not manipulate entities on serialization.

alexander-schranz: Oh sorry we are in deserialize here 🙈

nnatter:
Replaced by #5414

PR #5396, opened 2020-07-07
Accept zero as valid latitude and longitude in Location component

### LINKED ISSUE #5389: when adding a location on a map with a lat or long of 0 you can't save it
| Q | A
| --- | ---
| Sulu Version | 2.1.0
| Browser Version | latest Chrome

# Actual Behavior
You can't save a location on a map when the lat or long is 0.

# Expected Behavior
you should be able to add a location with 0 in one or both of these fields.

# Steps to Reproduce
type a '0' in the lat or location field

# Possible Solutions
the location lat 0 long 0 is also a locations so pick that as default and just allow a 0 in the fields.

nnatter:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #5389
| License | MIT

#### What's in this PR?

This PR adjusts the overlay of the `Location` component to accept 0 as valid latitude and valid longitude.

### DIFF
--- a/src/Sulu/Bundle/LocationBundle/Resources/js/containers/Location/LocationOverlay.js
+++ b/src/Sulu/Bundle/LocationBundle/Resources/js/containers/Location/LocationOverlay.js
@@ -179,7 +179,8 @@ class LocationOverlay extends React.Component<Props> {
         } = this.props;
 
         // enable confirm button if all marker properties are set or no property is set in case of a reset
-        const confirmEnabled = (this.markerLat && this.markerLong) || (!this.markerLat && !this.markerLong);
+        const confirmEnabled = (this.markerLat !== null && this.markerLong !== null)
+            || (this.markerLat === null && this.markerLong === null);
 
         return (
             <Overlay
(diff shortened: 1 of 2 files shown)


alexander-schranz:
@nnatter looks for me like this should target release/2.0 branch?

nnatter:
you are right 🙂

src/Sulu/Bundle/LocationBundle/Resources/js/containers/Location/LocationOverlay.js
```diff
@@ -179,7 +179,8 @@ class LocationOverlay extends React.Component<Props> {
         } = this.props;
 
         // enable confirm button if all marker properties are set or no property is set in case of a reset
-        const confirmEnabled = (this.markerLat && this.markerLong) || (!this.markerLat && !this.markerLong);
+        const confirmEnabled = (this.markerLat !== null && this.markerLong !== null)
```
alexander-schranz, line 182: This seems not be enough if I reopen the overlay the data is not there. So it seems like there is another check change needed where the data is stored.

nnatter: you are right - should hopefully be fixed now 🙂

alexander-schranz:
@nnatter Did work this way as it should 👍

PR #5402, opened 2020-07-09
Set max-height for Dialog component to prevent hidden controls

### LINKED ISSUE #5377: Category selection overlay not scrollable
| Q | A
| --- | ---
| Sulu Version | 2.1
| Browser Version | Chrome

# Actual Behavior

When many categories and opening the smart content filter for categories, and when you expand all of the subcategories suddenly the cross to exit the overlay is not available anymore (until your close all).

Schermafbeelding 2020-06-30 om 14 47 20

# Expected Behavior

Maybe a scroll behaviour that the list of categories is limited height

# Steps to Reproduce

Add many categories, open a smart content filter and browse through categories

# Possible Solutions

Scroll functionality with max-height (like in fragment?)

nnatter:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #5377
| License | MIT

#### What's in this PR?

This PR sets a `max-height` to the `Dialog` component. The height handling is implemented similar to the existing height handling of the `Overlay` component. 

#### Why?

Because the `Dialog` might overflow the screen and therefore hide the control buttons if there is a lot of content. If this happens, the user cannot close the dialog via the user interface anymore. See #5377

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Dialog/dialog.scss
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Dialog/dialog.scss
@@ -1,4 +1,5 @@
 @import '../../containers/Application/colors.scss';
+@import './variables.scss';
 
 $contentBackgroundColor: $concrete;
 $footerBackgroundColor: $white;
@@ -33,6 +34,7 @@ $transitionDuration: 300ms;
     transition: opacity $transitionDuration;
     z-index: 1;
     pointer-events: none;
+    max-height: calc(100vh - $dialogMarginVertical);
 
     & > * {
         pointer-events: auto;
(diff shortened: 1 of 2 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/components/Dialog/dialog.scss
```diff
@@ -33,6 +34,7 @@ $transitionDuration: 300ms;
     transition: opacity $transitionDuration;
     z-index: 1;
     pointer-events: none;
+    max-height: calc(100vh - $dialogMarginVertical);
```
danrot, line 37: Why isn't that variable called `dialogMarginBottom`, to be similar to the `Overlay` component?

nnatter: The `Overlay` has no `margin-top` because it is positioned at the top of the screen. The `Dialog` is centered, therefore calling it `dialogMarginBottom` feels wrong 🙂

PR #5403, opened 2020-07-11
Fix block types select width and overflowing form width in Firefox

tniebergall:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Set the width of the types select field to 150px to also show longer block names

Bildschirmfoto 2020-07-11 um 23 17 06

Bildschirmfoto 2020-07-11 um 23 16 49

#### Why?

To show longer block names in full length

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Block/block.scss
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Block/block.scss
@@ -47,7 +47,7 @@ $blockHandleWidth: 20px;
 }
 
 .types {
-    width: 150px;
+    min-width: 120px;
 }
 
 .icons {
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Block/block.scss
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Block/block.scss
@@ -47,7 +47,7 @@ $blockHandleWidth: 20px;
 }
 
 .types {
-    width: 120px;
+    width: 150px;
 }
 
 .icons {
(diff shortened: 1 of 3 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/components/Block/block.scss
```diff
@@ -72,7 +72,7 @@ $blockIconsColor: $gray;
 }
 
 .types {
-    width: 120px;
+    width: 150px;
```
nnatter, line 75: i think we can just use `min-width: 120px;` here. that would fix the problem for even longer type names

nnatter:
Hey,
I rebased this PR and implemented the suggestions. Hope that is okay for you.
Thanks again for your effort!

nnatter:
This can be tested with the following block property:
```xml
<block name="blocks" default-type="editor" minOccurs="0">
    <meta>
    </meta>
    <types>
        <type name="editor">
            <meta>
                <title lang="en">Editor with long name</title>
                <title lang="de">Editor mit langem Namen</title>
            </meta>
            <properties>
                <property name="text_editor" type="text_editor">
                    <meta>
                        <title lang="en">Text Editor</title>
                        <title lang="de">Texteditor</title>
                    </meta>
                </property>
            </properties>
        </type>
        <type name="media">
            <meta>
                <title lang="de">Media Selection with even longer name than editor, probably so long that it does not fit on the screen anymore without cropping it</title>
                <title lang="en">Medienauswahl mit noch längerem Namen wie Editor, vermutlich so lang, dass er nicht mehr auf dem Display dargestellt werden kann ohne gecroppt zu werden</title>
            </meta>
            <properties>
                <property name="media_selection" type="media_selection">
                    <meta>
                        <title lang="en">Media Selection</title>
                        <title lang="de">Medienauswahl</title>
                    </meta>
                </property>
            </properties>
        </type>
    </types>
</block>
```

alexander-schranz:
In Firefox the current solution didn't work I needed to adjust some more css that the Rendered Form is not going outside the container and the application view works correctly there.

PR #5406, opened 2020-07-14
Add frontend security for media and pages

### LINKED ISSUE #4051: Allow to set a object security for other systems than Sulu itself
| Q | A
| --- | ---
| Bug? | no
| New Feature? | yes

#### Actual Behavior

#### Expected Behavior

#### Possible Solutions

#### Open questions

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes parts of #4051 
| Related issues/PRs | ---
| License | MIT

#### What's in this PR?

This PR allows to set frontend security for pages and media. These contents will then be filtered out on the homepage, if the current user does not have any permission to access them.

#### Why?

Because this is a often requested featured.

#### BC Breaks/Deprecations

#### To Do

- [x] Show only one webspace system in `RolePermissions` for e.g. pages (only for the webspace the page is contained in)
- [x] Test saving permissions for pages when deleting a role permission
- [x] Implement deletion of roles permissions in PHPCR as well
- [x] Which permission required to see systems in permission tab?
- [x] Load roles from correct system for object specific security
- [x] Load roles from correct system for context specific security
- [x] ~~Replace instantiation of `SecurityCondition`s with Builder implementation~~ Won't be implemented because `SecurityCondition` is not touched now (`system` is determined by `AccessControlManager` instead)
- [x] Double-check that no `'Sulu'` is hardcoded anymore
- [x] ~~Passing `null` as system currently has a different result than before~~ WIll not be implemented since we change the default behavior
- [x] Fix loading of default permissions from roles for webspaces in pages (`#webspace#` is not replaced)
- [x] Determine correct system in `AccessControlManager` instead of hardcoding


src/Sulu/Bundle/SecurityBundle/Resources/js/containers/RolePermissions/RolePermissions.js
```diff
@@ -22,72 +23,80 @@ class RolePermissions extends React.Component<Props> {
...
-                {roles.map((role) => (
-                    <Matrix.Row key={role.id} name={role.id.toString()} title={role.name}>
-                        {actions.map((action) => (
-                            <Matrix.Item icon={getActionIcon(action)} key={action} name={action} />
-                        ))}
-                    </Matrix.Row>
-                ))}
-            </Matrix>
-        );
+        const systems = system ? ['Sulu', system] : securityContextStore.getSystems();
```
danrot, line 65: Hardcoded value

src/Sulu/Bundle/SecurityBundle/Resources/js/stores/securityContextStore/securityContextStore.js
```diff
@@ -1,33 +1,44 @@
...
-    getAvailableActions(resourceKey: string) {
-        for (const systemKey in this.securityContexts) {
-            const system = this.securityContexts[systemKey];
-            for (const groupKey in system) {
-                const group = system[groupKey];
-                for (const permissionKey in group) {
-                    if (permissionKey === this.resourceKeyMapping[resourceKey]) {
-                        return group[permissionKey];
-                    }
+    getAvailableActions(resourceKey: string, system: string = 'Sulu'): Actions {
```
danrot, line 27: Hardcoded value

danrot:
The last thing missing in this PR is to make sure that the permissions are checked for the correct system. My guess would be that this has to be added to our `SecurityCondition` class. Currently a typical security check looks like this:

```php
$this->securityChecker->checkPermission(
    new SecurityCondition(
        'sulu.media.collections',
        null,
        Collection::class,
        $mediaEntity->getCollection()->getId()
    ),
    PermissionTypes::DELETE
);
```

Copied that from a security check for a media collection (the `null` in the `SecurityCondition` is the optional loclizations for the check). We have to integrate the a `system` variable somehow, that would allow us to make sure that only permissions from assigned roles using that system are checked.

Two questions come up:

#### 1. Should this `system` parameter be optional?

I guess it should be, in order to keep BC, but if so, what is the default value? Currently it checks all roles regardless of their system, but that might lead to confusing behavior (e.g. if a website role is not allowed to see a specific collection, and no object permissions for the "Sulu" system have been added, it will not grant permissions in the "Sulu" system as well). Alternatively we could set the default system to "Sulu", but that might break existing applications, because they might have assigned another system to a role used in such a check.

#### 2. Where should this parameter be placed?

I think it has to go somehow into the `SecurityCondition` object, because it also contains other similar information. For BC reasons it would make sense to add it as the last parameter to the constructor, but that might lead to weird situations, because that is a parameter that might be set quite often, leading to code like this:

```php
new SecurityCondition('security.context', null, null, null, 'Sulu');
```

This is not very comfortable to write... The alternative would be two add the parameter at the second position:

```php
new SecurityCondition('security.context', 'Sulu');
```

But that has the downside of being a rather big BC break, affecting almost everybody using the `SecurityChecker`. Therefore I would stick with the first approach of adding it last, but add a `SecurityConditionBuilder` that allows to mix the arguments (as a replacement for the not yet available named parameters, if you will):

```php
$securityCondition = $this->securityConditionBuilderFactory->createSecurityConditionBuilder()
    ->setSecurityContext('security.context')
    ->setSystem('Sulu');
```

The builder would then take care of passing the constructor parameters correctly, while the very first example of this comment stays valid. Alternatively the first example could be written like this (and I would suppose we rewrite the existing code):

```php
$securityCondition = $this->securityConditionBuilderFactory->createSecurityConditionBuilder()
    ->setSecurityContext('sulu.media.collections')
    ->setObjectType(Collection::class)
    ->setObjectId($mediaEntity->getCollection()->getId())
    ->setSystem('Sulu');
```

@sulu/core-team Any inputs on this?

alexander-schranz:
@danrot I see also a second option to add a new static method to the `SecurityCondition` to create a security context for a system e.g.:

```php
SecurityCondition::createForSystem('sulu.media.collections', 'Sulu');
```

This is inspired by [JsonResponse](https://github.com/symfony/symfony/blob/fbe1ee941e775a0879635aea7745fff4e2d03fa9/src/Symfony/Component/HttpFoundation/JsonResponse.php#L53-L89) class which has this kind of static methods. This avoids to create a new addition builder service.

danrot:
@alexander-schranz A factory method would also work, yeah. But the thing is that there are not two totally exclusive ways of building a `SecurityCondition` as there is for the `JsonResponse`. Actually all of the available values except for the security context are optional and the `system` might be added in many different cases, which would end up making stuff like the `createForSystem` factory method to only swap the order of arguments. Or how would you implement it? I would imagine it somehow like this currently:

```php
class SecurityCondition {
    public function __construct($securityContext, $locale = null, $objectType = null, $objectId = null, $system = null) {
        // initialize fields
    }

    public static function createForSystem($securityContext, $system, $locale = null, $objectType = null, $objectId = null) {
        return new self($securityContext, $locale, $objectType, $objectId, $system);
    }
}
```

And that is just one of the available options. Since 4 parameters are optional, and we can't really tell which one are filled by the developer, we would have a function for every possible permutation, if we want to avoid passing `null` a lot. Additionally to the `createForSystem` we would also need a `createForObject($securityContext, $objectType, $objectId)`. If we want to be consistent and want to avoid using the constructor at all we would also have to add a `createForLocalizedObject` method. So as I see it, this method doesn't really scale very well. Also, we might forget to add one of the available permutations.

The variant with the builder enables the developer to decided which of the parameters should be set, and allows to do so for all possible permutations. Therefore I would favor that variant over the factory method.

danrot:
Result of discussion with @chirimoya and @alexander-schranz 

1. The `system` parameter for the security check must be optional and keep the current behavior for BC reasons, but if `null` is passed a deprecation will be thrown.

2. We go with the variant using `Builder` and `Factory` for the `SecurityCondition`.

danrot:
Some notes:

The `system` variable must somehow be considered when getting the permissions for both the **object specific** and **context specific** security checks. These permissions are currently loaded by the [`AccessControlProvider` (PHPCR or Doctrine) in the `AccessControlManager`](https://github.com/sulu/sulu/blob/7fc5cac/src/Sulu/Component/Security/Authorization/AccessControl/AccessControlManager.php#L73).

The easiest solution for **doctrine entities** would be to just add the `system` as a parameter to above function, and adjust the query done in the [`DoctrineAccessControlProvider`](https://github.com/sulu/sulu/blob/7fc5cac3cde666a6b8512e41e5f248bffeff752b/src/Sulu/Component/Security/Authorization/AccessControl/DoctrineAccessControlProvider.php#L97), which would mean that we have to [add a join to the roles in this query](https://github.com/sulu/sulu/blob/7fc5cac3cde666a6b8512e41e5f248bffeff752b/src/Sulu/Bundle/SecurityBundle/Entity/AccessControlRepository.php#L42-L51).

However, the case is not that easy for **phpcr documents**. The problem is, that **roles are only referred by their ID in these documents**, so we don't know to which system they belong. There are two ways of solving that:

- Encode the name of the system into the node property in the [SecuritySubscriber](https://github.com/sulu/sulu/blob/7fc5cac/src/Sulu/Component/Content/Document/Subscriber/SecuritySubscriber.php#L72)
- Filter the permissions before they are returned [in the `PhpcrAccessControlProvider`](https://github.com/sulu/sulu/blob/8dfc342/src/Sulu/Component/Security/Authorization/AccessControl/PhpcrAccessControlProvider.php#L64) by loading the role objects and compare the systems

danrot:
There are two open questions regarding the systems:

1. ~~The first one is how to handle **hardcoded system values like `"Sulu"`** (have already mentioned that in the [comment above](https://github.com/sulu/sulu/pull/5406#discussion_r456438719)). There is already a [constant in the abstract `Admin`](https://github.com/sulu/sulu/blob/5f0edc8/src/Sulu/Bundle/AdminBundle/Admin/Admin.php#L24), but it feels a bit off there, because I think it should somehow be part of the SecurityBundle. There is also a [configuration variable](https://github.com/sulu/sulu/blob/5f0edc812d095a371a0d86c8643795906d8e890d/src/Sulu/Bundle/SecurityBundle/DependencyInjection/Configuration.php#L30), but it seems to be deprecated for a reason. And there are already some hardcoded occurences [before this PR](https://github.com/sulu/sulu/blob/5f0edc812d095a371a0d86c8643795906d8e890d/src/Sulu/Bundle/SecurityBundle/Command/CreateRoleCommand.php#L95).~~ After having a closer look it seems that the constant was introduced in #5199, which I already didn't really like back then :see_no_evil: But in that case I would keep it for now.
2. This is actually the bigger question... In order to get a sensible behavior, we probably have to set a `system` in many different places where we are instantiating `SecurityCondition`s. Including some examples below:
- https://github.com/sulu/sulu/blob/851fc5464886e8cee65e4d2ed523620e16b00a96/src/Sulu/Component/Security/Authorization/SecurityChecker.php#L49
- https://github.com/sulu/sulu/blob/851fc5464886e8cee65e4d2ed523620e16b00a96/src/Sulu/Bundle/SecurityBundle/EventListener/SuluSecurityListener.php#L106
- https://github.com/sulu/sulu/blob/dc5d932e7ef43b894292156fdc4ec805707c62df/src/Sulu/Bundle/PageBundle/Controller/WebspaceController.php#L90
- https://github.com/sulu/sulu/blob/8c23932a4d8e1abc7d17f066e27c91726c588d48/src/Sulu/Bundle/PageBundle/EventListener/WebspaceSerializeEventSubscriber.php#L255
- https://github.com/sulu/sulu/blob/851fc5464886e8cee65e4d2ed523620e16b00a96/src/Sulu/Bundle/PageBundle/Repository/NodeRepository.php#L182
- https://github.com/sulu/sulu/blob/dc5d932e7ef43b894292156fdc4ec805707c62df/src/Sulu/Bundle/PageBundle/Controller/PageController.php#L698
- https://github.com/sulu/sulu/blob/851fc5464886e8cee65e4d2ed523620e16b00a96/src/Sulu/Bundle/MediaBundle/Media/Manager/MediaManager.php#L655

For the controllers it might be fine to just hard code a value of `'Sulu'`, although I thought we have also already recommended to expose routes to these controller e.g. for usage on the website as well. Just using the value of `'Sulu'` there would probably cause some issues. And especially for event listeners and services it feels to me like the might be used in different systems, so we should be careful about hardcoding some values there. Therefore I am wondering if we should introduce a separate service, which tries to determine the matching system for the current environment. If we are currently in the admin context, we could simply return the constant from the `Admin` class containing `'Sulu'`, and if it is used on the website, it could try to use the system defined in the current webspace. Then this service could be used in all the places we need to define a `SecurityCondition`. WDYT?

danrot:
In a short call @chirimoya, @alexander-schranz and I have discussed about the previously stated problem concerning the system in the `SecurityCondition`. Now the question came up, if it is a good idea to **put the `system` into the `SecurityCondition`**.

The idea I described before about extracting a service that will find the correct system based on our context (admin/website) and the webspace could be done directly in the `AccessControlManager`, making it unnecessary to pass that information via the `SecurityCondition`.

Until now I have thought about the `system` and the webspace in a decoupled way. There currently even is a `system` property on the webspace XML, which helps with that kind of thinking. But if we decide that finding the correct system happens directly in the `AccessControlManager` based on the webspace, then the webspace probably should be the `system`. Therefore we should probably even remove (or at least deprecate) these tags on the webspace XML, since it is a 1-to-1 relationship anyway.

src/Sulu/Bundle/SecurityBundle/Resources/config/services.xml
```diff
@@ -286,6 +290,13 @@
             <tag name="sulu.context" context="admin"/>
         </service>
 
+        <service id="sulu_security.system_listener" class="Sulu\Bundle\SecurityBundle\EventListener\SystemListener">
+            <argument type="service" id="sulu_security.system_store" />
+            <argument type="service" id="sulu_core.webspace.request_analyzer"/>
+            <argument>%sulu.context%</argument>
+            <tag name="kernel.event_listener" event="kernel.request" />
```
alexander-schranz, line 297: I would register it with a higher priority so its available as soon as possible. The webspace is set on `32` maybe go with `30`:

```
Registered Listeners for "kernel.request" Event
===============================================

 ------- ---------------------------------------------------------------------------------------------- ----------
  Order   Callable                                                                                       Priority
 ------- ---------------------------------------------------------------------------------------------- ----------
  #1      Symfony\Component\HttpKernel\EventListener\DebugHandlersListener::configure()                  2048
  #2      Symfony\Component\HttpKernel\EventListener\ValidateRequestListener::onKernelRequest()          256
  #3      FOS\RestBundle\EventListener\ZoneMatcherListener::onKernelRequest()                            248
  #4      Symfony\Component\HttpKernel\EventListener\SessionListener::onKernelRequest()                  128
  #5      Symfony\Component\HttpKernel\EventListener\LocaleListener::setDefaultLocale()                  100
  #6      Symfony\Component\HttpKernel\EventListener\FragmentListener::onKernelRequest()                 48
  #7      Sulu\Bundle\WebsiteBundle\EventListener\RouterListener::onKernelRequest()                      32
  #8      Sulu\Bundle\WebsiteBundle\Routing\RequestListener::onRequest()                                 31
  #9      Symfony\Component\HttpKernel\EventListener\LocaleListener::onKernelRequest()                   16
  #10     Symfony\Component\HttpKernel\EventListener\LocaleAwareListener::onKernelRequest()              15
  #11     Sulu\Bundle\WebsiteBundle\EventListener\TranslatorListener::onKernelRequest()                  14
  #12     FOS\RestBundle\EventListener\BodyListener::onKernelRequest()                                   10
  #13     Symfony\Bundle\SecurityBundle\Debug\TraceableFirewallListener::configureLogoutUrlGenerator()   8
  #14     Symfony\Bundle\SecurityBundle\Debug\TraceableFirewallListener::onKernelRequest()               8
  #15     Sulu\Bundle\SecurityBundle\EventListener\UserLocaleListener::copyUserLocaleToRequest()         0
  #16     Sulu\Bundle\SecurityBundle\EventListener\SystemListener::onKernelRequest()                     0
 ------- ---------------------------------------------------------------------------------------------- ----------
```

src/Sulu/Bundle/SecurityBundle/Entity/AccessControlRepository.php
```diff
@@ -37,13 +37,18 @@ public function findByTypeAndIdAndRole($type, $id, $roleId)
         }
     }
 
-    public function findByTypeAndId($type, $id)
+    public function findByTypeAndId($type, $id, $system = null)
```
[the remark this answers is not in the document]
danrot: Not sure if we should do either of those things... This class is probably not touched by anybody... If I could do it again, I would remove the interface and make that class final, because it is not meant to be overridden... Adding code like what you are suggesting is cumbersome, for not a big gain, and documenting it will bloat the `UPGRADE.md`, which will make people read it less carefully...

If the rest of the team disagrees I am find with implementing it the way you've suggested though.

src/Sulu/Component/Security/Authentication/RoleRepositoryInterface.php
```diff
@@ -37,4 +37,9 @@ public function findAllRoles();
      * @return array
      */
     public function getRoleNames();
+
+    /**
+     * @return array
+     */
+    public function findRoleIdsBySystem($system);
```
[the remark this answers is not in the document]
danrot: Same as above, I am worried that we are bloating the `UPGRADE.md` too much, if we add stuff like this :confused: Happy to do if the majority has a different opinion.

src/Sulu/Component/Content/Document/Subscriber/SecuritySubscriber.php
```diff
@@ -23,6 +23,7 @@
  */
 class SecuritySubscriber implements EventSubscriberInterface
 {
+    const SECURITY_PROPERTY_PREFIX = 'sec:role-';
```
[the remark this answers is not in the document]
danrot: Didn't find a rule... but I am fixing it of course.

src/Sulu/Bundle/SecurityBundle/Resources/config/services.xml
danrot: Good catch! Going with `24`, to keep binary methodology and leave some room between the listeners.

alexander-schranz: `24` sounds good for me.

alexander-schranz:
permission

As shown to you. When activate a system permission and change permissions only in another system, the system for the unchanged system are not saved.

See some options how it could behave in that case:

A. It should just save the permission like when I check and uncheck something in that system
B. It should uncheck that system when I edit things in another system
C. It should uncheck that system when form was saved to represent the same status which is shown in reload

PR #3263, opened 2017-03-16
Escape search terms

janmassive:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Search terms are escaped.

#### Why?

Words containing special characters have led to empty search result.

PR #5407, opened 2020-07-14
Fix fuzzy matching in WebsiteSearchController by omiting quotes from query

nnatter:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Related issues/PRs | https://github.com/sulu/SuluHeadlessBundle/pull/37 https://github.com/sulu/sulu/pull/3263
| License | MIT

#### What's in this PR?

This PR adjusts the `WebsiteSearchController` to not add quotes around the query string for the fuzzy matching search terms. 

These removed quotes were introduced in #3263 with the reason `Words containing special characters have led to empty search result.`. I am not sure if this reason was correct because the `preg_replace('/([^\pL\s\d])/u', '?', $queryValue)` should actually remove all special characters.

#### Why?

At the moment, searching for `keywor` creates the following query: `+("keywor" OR "keywor*" OR "keywor~")`
As the fuzzy matching search terms (`"keywor*"`, `"keywor~"`) are enclosed in `"`, the fuzzy characters (`*`. `~`) will be interpreted as literal characters. For example, this means that searching for `keywor` will not include a page with the title `Keyword` in the result.

See: https://github.com/sulu/SuluHeadlessBundle/pull/37

### DIFF
--- a/src/Sulu/Bundle/SearchBundle/Tests/Application/config/webspaces/sulu.io.xml
+++ b/src/Sulu/Bundle/SearchBundle/Tests/Application/config/webspaces/sulu.io.xml
@@ -22,6 +22,10 @@
         <default-template type="homepage">overview</default-template>
     </default-templates>
 
+    <templates>
+        <template type="search">search</template>
+    </templates>
+
     <navigation>
         <contexts>
             <context key="main">
@@ -78,6 +82,15 @@
                     <urls>
                         <url language="de" country="at">sulu.lo</url>
                         <url language="de" country="at">localhost</url>
+                        <url language="de">de.sulu.lo</url>
+                        <url redirect="sulu.lo">sulu-redirect.lo</url>
+                    </urls>
+                </environment>
+                <environment type="test_zend_lucene">
+                    <urls>
+                        <url language="de" country="at">sulu.lo</url>
+                        <url language="de" country="at">localhost</url>
+                        <url language="de">de.sulu.lo</url>
                         <url redirect="sulu.lo">sulu-redirect.lo</url>
                     </urls>
                 </environment>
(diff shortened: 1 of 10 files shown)


alexander-schranz:
We should definitely creating functional test cases else this will in future make more problems.

src/Sulu/Bundle/SearchBundle/Tests/Application/config/webspaces/sulu.io.xml
```diff
@@ -78,6 +82,15 @@
                     <urls>
                         <url language="de" country="at">sulu.lo</url>
                         <url language="de" country="at">localhost</url>
+                        <url language="de">de.sulu.lo</url>
+                        <url redirect="sulu.lo">sulu-redirect.lo</url>
+                    </urls>
+                </environment>
+                <environment type="test_zend_lucene">
```
alexander-schranz, line 89: This should be `test_search_adapter`:

```suggestion
               <environment type="test_search_adapter">
```

src/Sulu/Bundle/SearchBundle/Tests/Functional/Controller/WebsiteSearchControllerTest.php
```diff
@@ -0,0 +1,138 @@
...
+     * @var KernelBrowser
+     */
+    private $websiteClient;
+
+    protected static function getKernelConfiguration(): array
+    {
+        return [
+            'environment' => 'test_zend_lucene',
+        ];
+    }
```
alexander-schranz, line 32: This should be removed

```suggestion
```

PR #5410, opened 2020-07-14
Update FormatOptions.orm.xml

### LINKED ISSUE #5409: FormatOptions entity uses INT to store crop values but could receive FLOATS
| Q | A
| --- | ---
| sulu version | 2.1.0

# Actual Behavior

Sulu supplies FLOATS to the DB and MySQL will auto cast those to INT.

# Expected Behavior

Just save INTs to the DB

# Steps to Reproduce

Use PGSQL and try to save an copped image. It will crash on ->flush().

# Possible Solutions

Postgrsql doesn't auto cast and throws an error.  So somewhere the data should be casted to INT

kfrohwein:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #5409
| License | MIT

Changed the type of the crop fields to FLOAT from INT because my Chrome Browser provided FLOAT now and pgsql will not autocast FLOAT to INT.

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/Resources/config/doctrine/FormatOptions.orm.xml
+++ b/src/Sulu/Bundle/MediaBundle/Resources/config/doctrine/FormatOptions.orm.xml
@@ -9,10 +9,10 @@
         <id name="fileVersion" association-key="true"/>
         <id name="formatKey" type="string" length="255"/>
 
-        <field name="cropX" type="integer"/>
-        <field name="cropY" type="integer"/>
-        <field name="cropWidth" type="integer"/>
-        <field name="cropHeight" type="integer"/>
+        <field name="cropX" type="float"/>
+        <field name="cropY" type="float"/>
+        <field name="cropWidth" type="float"/>
+        <field name="cropHeight" type="float"/>
 
         <many-to-one field="fileVersion" target-entity="Sulu\Bundle\MediaBundle\Entity\FileVersion" inversed-by="formatOptions">
             <join-columns>


alexander-schranz:
Imagine works with whole pixels and so works with int. So `int` is the correct database type here. If we have problems with PostgreSQL not converting the values to int we should cast the given values, so they are correctly converted to int.

kfrohwein:
@alexander-schranz @danrot told me it schould be float. Well anyways so it is not a DB problem but that for some reason the JS Library does provide floats and not ints. And there doesn't seem to be any casting to int. So all in all maybe this is a problem with a later chrome version. But I believe our content managers use the Internet Explorer which means it might not be browser specific.

danrot:
@alexander-schranz Does that mean that imagine does not work with floats at all? The problem is that you migth easily end up with float values, and depending on the size of the image that might cause visible distortions.

alexander-schranz:
@danrot yes imagine doesn't work with floats I also remember that we needed to add cast to int in our calculations that it work correctly, you can not crop images by half pixels, in best it would be sending whole pixels to the api. I don't see it as visible distortions as a generated image can not have half pixels. We had in the past more problems when given unrounded floats to the imagine adapter as it did cast it to an incorrect value see #4818, so I would do the same for the width, height, x and y of the formatOptions setters.

Also the SVG Imagine adapters using strict types https://github.com/contao/imagine-svg/blob/5ab31a68fbba41f359a0e38378d6b5f43df13a01/src/SvgBox.php#L35 there it could end into an type error with half pixels.

danrot:
Well, sounds like we have to go with a client side solution then... Probably we should just round the values before sending them to the API? Maybe it is enough to set the following value to true: https://github.com/sulu/sulu/blob/51b18d8/src/Sulu/Bundle/AdminBundle/Resources/js/components/ImageRectangleSelection/ImageRectangleSelection.js#L124

But I remember that this kind of stuff is quite nasty, because these values are immediately fed back in a loop, to make sure they are always in sync. However, that might lead to other weird situations...

PR #5414, opened 2020-07-16
Fix serialization regression for TargetGroups

danrot:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | schmittjoh/serializer#1239
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR tries to fix some serialization issues introduced with the jms/serializer 3.7.0.

#### Why?

It seems like previously the [`DoctrineObjectConstructor`](https://github.com/schmittjoh/serializer/blob/d6bee10/src/Construction/DoctrineObjectConstructor.php) has not been used at all for a reason. We could simply avoid the usage of this constructor, but when applied correctly it simplifies our code by a lot. So this PR will try to integrate the new behavior.

However, there is still an issue in the JMS serializer, that makes that not possible at the moment. Read more about that in schmittjoh/serializer#1239. Once that is fixed, this code is hopefully working as it should.


src/Sulu/Bundle/CoreBundle/DependencyInjection/Compiler/JmsObjectConstructorCompilerPass.php
```diff
@@ -0,0 +1,30 @@
...
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\CoreBundle\DependencyInjection\Compiler;
+
+use JMS\Serializer\Construction\DoctrineObjectConstructor;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+class JmsObjectConstructorCompilerPass implements CompilerPassInterface
```
alexander-schranz, line 18: Should we mark this as `final` and `@internal`?

alexander-schranz (changes requested):

src/Sulu/Bundle/CoreBundle/DependencyInjection/Compiler/JmsObjectConstructorCompilerPass.php
danrot: Done

alexander-schranz (approved):

PR #5415, opened 2020-07-16
Fix teaser-selection deselecting

### LINKED ISSUE #5412: Adding and removing elements to teaser selection behaves strange
| Q | A
| --- | ---
| Bug? | yes
| New Feature? | no
| Sulu Version | 1.6.33

#### Actual Behavior

By adding more then one page, there is always one page less added to the content-type then selected.
When removing pages in the overlay from the content-type. It still shows those pages.
Same when deleting them in the content-type, saving and reloading the browser.

#### Expected Behavior

Korrekt behavior after selecting, unselecting pages.

#### Steps to Reproduce

Add teaser_selection content-type to a page and start adding / removing pages from it.

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #5412
| Related issues/PRs | none
| License | MIT
| Documentation PR | none

#### What's in this PR?

Fixes the bug #5412


src/Sulu/Bundle/ContentBundle/Resources/public/js/components/teaser-selection/main.js
```diff
@@ -213,11 +213,16 @@ define([
...
+                                        return arrayItem.id === item.id && arrayItem.type === item.type;
+                                    });
+
+                                    if (0 === filteredArray.length) {
+                                        data.push(item);
+                                    } else {
+                                        data = _.without(data, _.findWhere(data, item));
+                                    }
                                 },
-                                deselectCallback: function(item) {
```
alexander-schranz, line 218: I'm little bit confused why this function was removed or why it did exist 🙈

danrot: Just tested it, and it seems to work this way... Looks to me like the deselectCallback is never called :see_no_evil: Therefore this change is fine for me.

PR #5416, opened 2020-07-16
Make disabled MultiItemSelection scrollable

### LINKED ISSUE #5413: Make entries of disabled media_selection scrollable
| Q | A
| --- | ---
| Sulu Version | 2.1

# Actual Behavior

The media list has a max height of `225px`. If there are too many medias selected in a disabled `media_selected`, the media list cannot be scrolled.

# Expected Behavior

Scrolling should work in disabled `media_selections`.

# Steps to Reproduce

~~~xml
<property name="medias" type="media_selection" disabledCondition="true">
    <meta>
        <title>app.medias</title>
    </meta>
</property>
~~~

luca-rath:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #5413
| License | MIT

#### What's in this PR?

Make disabled MultiItemSelection scrollable

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/MultiItemSelection/Header.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/MultiItemSelection/Header.js
@@ -9,6 +9,7 @@ import headerStyles from './header.scss';
 const LOADER_SIZE = 24;
 
 type Props<T, U> = {
+    className?: string,
     emptyList: boolean,
     label?: string,
     leftButton?: ButtonConfig<T>,
@@ -23,6 +24,7 @@ export default class Header<T: string | number, U: string | number> extends Reac
 
     render() {
         const {
+            className,
             label,
             loading,
             emptyList,
@@ -31,6 +33,7 @@ export default class Header<T: string | number, U: string | number> extends Reac
         } = this.props;
 
         const headerClass = classNames(
+            className,
             headerStyles.header,
             {
                 [headerStyles.emptyList]: emptyList,
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/MultiItemSelection/Header.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/MultiItemSelection/Header.js
@@ -9,7 +9,7 @@ import headerStyles from './header.scss';
 const LOADER_SIZE = 24;
 
 type Props<T, U> = {
-    className?: string,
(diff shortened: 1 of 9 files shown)


luca-rath:
@danrot I'm not sure how I should fix the failing pipeline?

This is the problematic code:
```jsx
<section className={blockClass} onClick={this.handleExpand} role="switch">
```

src/Sulu/Bundle/AdminBundle/Resources/js/components/MultiItemSelection/MultiItemSelection.js
```diff
@@ -21,11 +21,26 @@ type Props<T, U, V> = {|
...
 |};
 
-const ItemWrapper = ({children}: Object) => (
-    <li className={multiItemSelectionStyles.listElement}>
-        {children}
-    </li>
-);
+type ItemWrapperProps = {
+    children: Element<typeof Item>,
+    isDisabled: boolean,
```
danrot, line 26: We are usually only using `disabled`.

danrot: Ok, weird that the `SortableElement` does not support that... But it is not that bad, since that component is only used within this file.

src/Sulu/Bundle/AdminBundle/Resources/js/components/MultiItemSelection/Header.js
```diff
@@ -23,6 +24,7 @@ export default class Header<T: string | number, U: string | number> extends Reac
 
     render() {
         const {
+            className,
```
danrot, line 27: Why not making this more explicit, and pass a `disabled` prop instead? Then we also would not have the `header` class applied twice in the tests.

danrot: And it would be consistent with the `ItemWrapper` below.

danrot (changes requested):

src/Sulu/Bundle/AdminBundle/Resources/js/components/MultiItemSelection/Header.js
luca-rath: fixed

danrot (approved):

PR #5418, opened 2020-07-16
Store values from image crop as integers

### LINKED ISSUE #5409: FormatOptions entity uses INT to store crop values but could receive FLOATS

kfrohwein:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #5409
| License | MIT

#### What's in this PR?

I added type casts and return types to the FormatOptions Entity Object.
As JS could provide FLOATs this might be a good place to make sure SULU and it's DB is working with INTs in the end.

By doing this I figured that

`$fileVersion->addFormatOptions($formatOptions);`

relies on

`$formatOptions->setFormatKey($formatKey);`

is already called or the array key will be NULL
No clue how this ever could have been able to run but probably the array of format options is never used.

#### Possible BC break
PHPStorm tells me that return self is PHP > 7 and the 1.6 still supports 5.5

#### Why?

JS might provide FLOATS and at least PGSQL will not save the crop then.

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/Entity/FormatOptions.php
+++ b/src/Sulu/Bundle/MediaBundle/Entity/FormatOptions.php
@@ -53,9 +53,9 @@ class FormatOptions
      *
      * @return FormatOptions
      */
-    public function setCropX($cropX): self
+    public function setCropX($cropX)
     {
-        $this->cropX = (int) $cropX;
+        $this->cropX = (int) \round($cropX);
 
         return $this;
     }
@@ -65,7 +65,7 @@ class FormatOptions
      *
      * @return int
      */
-    public function getCropX(): int
+    public function getCropX()
     {
         return $this->cropX;
     }
@@ -77,9 +77,9 @@ class FormatOptions
      *
      * @return FormatOptions
      */
-    public function setCropY($cropY): self
+    public function setCropY($cropY)
     {
-        $this->cropY = (int) $cropY;
+        $this->cropY = (int) \round($cropY);
 
         return $this;
     }
@@ -89,7 +89,7 @@ class FormatOptions
      *
      * @return int
      */
-    public function getCropY(): int
+    public function getCropY()
     {
         return $this->cropY;
     }
@@ -101,9 +101,9 @@ class FormatOptions
      *
      * @return FormatOptions
      */
-    public function setCropWidth($cropWidth): self
+    public function setCropWidth($cropWidth)
     {
-        $this->cropWidth = (int) $cropWidth;
+        $this->cropWidth = (int) \round($cropWidth);
 
         return $this;
     }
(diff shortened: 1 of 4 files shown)


src/Sulu/Bundle/MediaBundle/Entity/FormatOptions.php
```diff
@@ -101,9 +101,9 @@ public function getCropY()
      *
      * @return FormatOptions
      */
-    public function setCropWidth($cropWidth)
+    public function setCropWidth($cropWidth): self
     {
-        $this->cropWidth = $cropWidth;
+        $this->cropWidth = (int) $cropWidth;
```
alexander-schranz, line 106: same as above should be rounded `(int) \round(cropWidth)`

src/Sulu/Bundle/MediaBundle/Entity/FormatOptions.php
```diff
@@ -125,9 +125,9 @@ public function getCropWidth()
      *
      * @return FormatOptions
      */
-    public function setCropHeight($cropHeight)
+    public function setCropHeight($cropHeight): self
     {
-        $this->cropHeight = $cropHeight;
+        $this->cropHeight = (int) $cropHeight;
```
alexander-schranz, line 130: same as above should be rounded `(int) \round($cropHeight)`

kfrohwein:
@alexander-schranz So what about this PR? Is it okay now?

src/Sulu/Bundle/MediaBundle/Media/FormatOptions/FormatOptionsManager.php
```diff
@@ -121,8 +121,8 @@ public function save($mediaId, $formatKey, array $data)
         if (!isset($formatOptions)) {
             $formatOptions = new FormatOptions();
             $formatOptions->setFileVersion($fileVersion);
-            $fileVersion->addFormatOptions($formatOptions);
             $formatOptions->setFormatKey($formatKey);
+            $fileVersion->addFormatOptions($formatOptions);
```
alexander-schranz, line 125: This change is not necessary as adding can be done before it. In PHP a object is based by reference and not a copy so you can edit it.

alexander-schranz: Ah I see what you mean, ignore my comment :)

danrot:
Just added the changelog line, will merge once the tests are green.

alexander-schranz:
@kfrohwein Thank you!

PR #5424, opened 2020-07-21
Add private image format cache

### LINKED ISSUE #4051: Allow to set a object security for other systems than Sulu itself

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #4051
| Related issues/PRs | #5406
| License | MIT

#### What's in this PR?

Add private image format cache.

#### Why?

We want to allow to have private media per collection.

#### TODO

- [ ] Update Deployer Sulu 2 Recipe https://github.com/deployphp/deployer/pull/2121


src/Sulu/Bundle/MediaBundle/DependencyInjection/Configuration.php
```diff
@@ -133,6 +133,7 @@ public function getConfigTreeBuilder()
             ->arrayNode('format_cache')
                 ->addDefaultsIfNotSet()
                 ->children()
+                    ->scalarNode('private_path')->defaultValue('%kernel.project_dir%/var/images-cache')->end()
```
alexander-schranz, line 136: Somebody maybe ask why I did choose to store it here. There would be two alternatives to this but both make more problems.

Alternative *A*  would be to store it in `var/cache/` directory. This I find not good as many developers remove that cache folder and I think its not good to trigger the rerender images again.

Alternative *B*  would be to store it in `var/uploads` directory. What problem I have with this is that it is mostly be the folder which is also be backed up, as in projects I mostly have other files also there for example:

 - `var/uploads/media`
 - `var/uploads/pdfs`
 - `var/uploads/invoices`
 - `var/uploads/wallets`

Things which need to be backup. Thats why I decided to add a additional folder `var/images` in the root folder. So I can still make whole `var/uploads` part of the backup and don't need to filter out `var/uploads/images` folder.

To more represent that its just a cache I decided to name it `var/images-cache`.

alexander-schranz: After discussion with @chirimoya it was decided to go with `var/media/formats` and also change the urls and folder in the skeleton for the public to `public/media/formats` and `/media/formats/{slug}`

alexander-schranz:
closing this in favor of #5437

PR #5432, opened 2020-07-27
Add possibility for distinct to count field descriptor

alexandersch:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | todo

#### What's in this PR?

Added a distinct option to the existing Count FieldDescriptor

#### Why?

You could not do a COUNT(DISTINCT()) on a field, only COUNT() or set distinct on the whole SELECT.

#### Example Usage

~~~xml
<count-property name="teams" visibility="always" searchability="no" distinct="true">
    <field-name>id</field-name>
    <entity-name>App\Bundle\Entity\Team</entity-name>
    <joins>
        <join>
            <entity-name>App\Bundle\Entity\Team</entity-name>
            <field-name>App\Bundle\Entity\Season.teams</field-name>
        </join>
    </joins>
</count-property>
~~~

#### To Do


src/Sulu/Component/Rest/ListBuilder/Doctrine/FieldDescriptor/DoctrineCountDistinctFieldDescriptor.php
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * This class defines the necessary information for a field to resolve it within a Doctrine Query for the ListBuilder.
+ *
+ * @ExclusionPolicy("all")
+ */
+class DoctrineCountDistinctFieldDescriptor extends DoctrineFieldDescriptor
+{
+    public function getSelect()
+    {
+        return 'COUNT(DISTINCT(' . $this->encodeAlias($this->getEntityName()) . '.' . $this->getFieldName() . '))';
```
alexander-schranz, line 23: Looks for me like the `DoctrineGroupConcatFieldDescriptor` has a `distinct` flag in its constructor for this kind of query, so maybe it would be better to extend the `DoctrineDistinctFieldDescriptor`

https://github.com/sulu/sulu/blob/release/2.1/src/Sulu/Component/Rest/ListBuilder/Doctrine/FieldDescriptor/DoctrineGroupConcatFieldDescriptor.php#L68

/cc @danrot what do you think?

src/Sulu/Component/Rest/ListBuilder/Metadata/CountDistinctPropertyMetadata.php
```diff
@@ -0,0 +1,38 @@
...
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Component\Rest\ListBuilder\Metadata;
+
+/**
+ * Describes a one-to-many field which will be evaluated with COUNT DISTINCT function.
+ */
+class CountDistinctPropertyMetadata extends AbstractPropertyMetadata
```
alexander-schranz, line 17: Based on the above discussion this could be removed.

src/Sulu/Component/Rest/ListBuilder/Metadata/FieldDescriptorFactory.php
```diff
@@ -221,6 +222,32 @@ private function getCountFieldDescriptor(AbstractPropertyMetadata $propertyMetad
         );
     }
 
+    private function getCountDistinctFieldDescriptor(AbstractPropertyMetadata $propertyMetadata)
```
alexander-schranz, line 225: This function is never called think there is something missing at: https://github.com/sulu/sulu/blob/9834ad41870335a6cdc2cc7c799772e9484b1662/src/Sulu/Component/Rest/ListBuilder/Metadata/FieldDescriptorFactory.php#L88-L93

But based on the above discussion its maybe not needed if we extend the exist DoctrineCountFieldDescriptor

src/Sulu/Component/Rest/ListBuilder/Metadata/ListXmlLoader.php
```diff
@@ -203,6 +206,19 @@ private function loadCountPropertyMetadata(\DOMXPath $xpath, \DOMNode $propertyN
         return $propertyMetadata;
     }
 
+    private function loadCountDistinctPropertyMetadata(\DOMXPath $xpath, \DOMNode $propertyNode)
```
alexander-schranz, line 209: Maybe not needed if we change the exist adapter. Lets wait for the response of the other core teams members.

src/Sulu/Component/Rest/ListBuilder/Metadata/Resources/schema/list-2.0.xsd
```diff
@@ -57,6 +57,11 @@
         <xs:attributeGroup ref="propertyAttributeGroup"/>
     </xs:complexType>
 
+    <xs:complexType name="countDistinctPropertyType">
```
alexander-schranz, line 60: Maybe not needed if we change the exist adapter. Lets wait for the response of the other core teams members.

alexandersch:
@alexander-schranz I've created a unit test for DoctrineCountFieldDescriptor::getSelect, one with and one without distinct.

src/Sulu/Component/Rest/ListBuilder/Metadata/Resources/schema/list-2.0.xsd
```diff
@@ -135,6 +135,11 @@
         <xs:attribute name="distinct" type="xs:boolean"/>
     </xs:attributeGroup>
 
+    <xs:attributeGroup name="countPropertyAttributeGroup">
```
alexander-schranz, line 138: /cc @sulu/core-team think we need to create a new .xsd when else it would autocomplete distinct in old versions where distinct does not exist. Or what do you think @sulu/core-team u?

danrot: It is just a minor DX issue and doesn't blow up in any way, right? If we create a new XSD for every single attribute we add, it's going to be very hard to maintain :confused: Therefore I would not do it.

src/Sulu/Component/Rest/ListBuilder/Doctrine/FieldDescriptor/DoctrineCountDistinctFieldDescriptor.php
danrot: If you are talking about the `DoctrineCountFieldDescriptor` that's ok for me :slightly_smiling_face:

alexander-schranz:
@alexandersch Thank you for contribution!

PR #5433, opened 2020-07-27
Add anonymous roles for each system

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes -
| Related issues/PRs | -
| License | MIT

#### What's in this PR?

Add anonymous flag to role.

#### Why?

We need the possibility to set permissions for anonymous users so we need anonymous roles in every system.

#### To Do

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -2,6 +2,25 @@
 
 ## dev-master
 
+### Role Entity changed for anonymous roles
+
+Sulu need to handle anonymous users for this we need additional anonymous roles.
+
+For this a Role Entity need an new anonymous field:
+
+```sql
+ALTER TABLE se_roles ADD anonymous TINYINT(1) NOT NULL;
+```
+
+In this case also the `RoleInterface` has changed and need to implement the new `getAnonymous` and `setAnonymous`
+functions.
+
+Now you need to run the following command to create the anonymous users:
+
+```bash
+bin/adminconsole sulu:security:init
+```
+
 ### Deprecation of securityContextStore methods
 
 The `loadSecurityContextGroups` and `loadAvailableActions` method from the `securityContextStore` have been deprecated.
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -4,16 +4,15 @@
 
 ### Role Entity changed for anonymous roles
 
-Sulu need to handle anonymous users for this we need additional anonymous roles.
+Sulu needs to handle anonymous users, so we need additional anonymous roles.
 
-For this a Role Entity need an new anonymous field:
+Therefore the `Role` Entity needs a new field:
 
 ```sql
 ALTER TABLE se_roles ADD anonymous TINYINT(1) NOT NULL;
 ```
 
-In this case also the `RoleInterface` has changed and need to implement the new `getAnonymous` and `setAnonymous`
-functions.
+The `RoleInterface` has changed and now contains a `getAnonymous` and `setAnonymous` method.
 
 Now you need to run the following command to create the anonymous users:
 
--- a/src/Sulu/Bundle/SecurityBundle/Build/InitSecurityBuilder.php
+++ b/src/Sulu/Bundle/SecurityBundle/Build/InitSecurityBuilder.php
@@ -1,0 +1,38 @@
+<?php
+
+/*
+ * This file is part of Sulu.
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\SecurityBundle\Build;
+
+use Sulu\Bundle\CoreBundle\Build\SuluBuilder;
+
+/**
+ * Builder for creating anonymous roles.
+ */
+class InitSecurityBuilder extends SuluBuilder
+{
+    public function getName()
+    {
+        return 'init_security';
+    }
+
+    public function getDependencies()
+    {
+        return ['fixtures', 'database'];
+    }
+
+    public function build()
+    {
+        $this->execCommand(
+            'Initialize security: ',
+            'sulu:security:init'
+        );
+    }
+}
--- a/src/Sulu/Bundle/SecurityBundle/Command/InitCommand.php
+++ b/src/Sulu/Bundle/SecurityBundle/Command/InitCommand.php
@@ -1,0 +1,114 @@
+<?php
+
+/*
+ * This file is part of Sulu.
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\SecurityBundle\Command;
+
+use Doctrine\ORM\EntityManagerInterface;
+use Sulu\Bundle\AdminBundle\Admin\Admin;
+use Sulu\Bundle\AdminBundle\Admin\AdminPool;
+use Sulu\Component\Security\Authentication\RoleInterface;
+use Sulu\Component\Security\Authentication\RoleRepositoryInterface;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+
+/**
+ * @internal
+ */
+final class InitCommand extends Command
+{
+    protected static $defaultName = 'sulu:security:init';
+
+    /**
+     * @var EntityManagerInterface
+     */
+    private $entityManager;
+
+    /**
+     * @var RoleRepositoryInterface
+     */
+    private $roleRepository;
+
+    /**
+     * @var AdminPool
+     */
+    private $adminPool;
+
+    public function __construct(
+        EntityManagerInterface $entityManager,
+        RoleRepositoryInterface $roleRepository,
+        AdminPool $adminPool
+    ) {
+        parent::__construct();
+
+        $this->entityManager = $entityManager;
+        $this->roleRepository = $roleRepository;
+        $this->adminPool = $adminPool;
+    }
+
+    protected function configure()
+    {
+        $this->setName('sulu:security:init')
+            ->setDescription('Create required sulu security entities.');
+    }
+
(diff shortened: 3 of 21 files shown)


src/Sulu/Bundle/SecurityBundle/Entity/Role.php
```diff
@@ -110,6 +116,10 @@ public function getRole()
 
     public function getIdentifier()
     {
+        if ($this->anonymous) {
+            return AuthenticatedVoter::IS_ANONYMOUS;
```
alexander-schranz, line 120: How should we handle this the identifier of our anonymous roles:

*Option A*:

Override the Identifier function and return the Symfony `IS_ANONYMOUS` role if anonymous flag is set.

*Option B*:

Change the unique constraint from [Role](https://github.com/sulu/sulu/blob/89518be214a030c68d12c16a459f2ef74beced4d/src/Sulu/Bundle/SecurityBundle/Resources/config/doctrine/Role.orm.xml#L12)  from `key` to `key,system`. So a role is unique inside its system.

*Option C*

Create a Role with a Identifier `ROLE_<SYSTEM>_ANONYMOUS`. This sadly lose the connection to the Symfony `IS_ANONYMOUS` flag an its `is_granted('IS_ANONYMOUS')` function, but not sure if we need that.

alexander-schranz: If I think about it if we go with **Option A** or **Option B** the `getRoles` function should only return Roles for the current system. So the UserProvider would need to set a in memory flag on the user with `currentSystem`.

danrot: Currently option A is implemented, right? Seems like the cleanest solution to me, but wouldn't we still have to make only the combination of `system` and `key` unique?

alexander-schranz: > Currently option A is implemented, right?

Yes implemented is currently *Option A*.

> Seems like the cleanest solution to me, but wouldn't we still have to make only the combination of system and key unique?

No as the key is null in the database we don't need todo that for anonymous roles.

danrot: If this works like that, I don't see any reason to change it :shrug: Would probably keep it as it is, and only change if we encounter any problems.

alexander-schranz: Okay then I will got with that solution.

UPGRADE.md
```diff
@@ -2,6 +2,25 @@
 
 ## dev-master
 
+### Role Entity changed for anonymous roles
+
+Sulu need to handle anonymous users for this we need additional anonymous roles.
```
danrot, line 7: "Sulu needs to handle anonymous users, so we need additional anonymous roles."

UPGRADE.md
```diff
@@ -2,6 +2,25 @@
...
+
+For this a Role Entity need an new anonymous field:
```
danrot, line 9: "Therefore the `Role` Entity needs a new field:"

UPGRADE.md
```diff
@@ -2,6 +2,25 @@
...
+
+```sql
+ALTER TABLE se_roles ADD anonymous TINYINT(1) NOT NULL;
+```
+
+In this case also the `RoleInterface` has changed and need to implement the new `getAnonymous` and `setAnonymous`
```
danrot, line 15: "The `RoleInterface` has changed and now contains a `getAnonymous` and `setAnonymous` method."

src/Sulu/Bundle/SecurityBundle/Build/InitSecurityBuilder.php
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+namespace Sulu\Bundle\SecurityBundle\Build;
+
+use Sulu\Bundle\CoreBundle\Build\SuluBuilder;
+
+/**
+ * Builder for creating anonymous roles.
+ */
+class InitSecurityBuilder extends SuluBuilder
```
danrot, line 19: Shouldn't we call that just `SecurityBuilder`? None of the other builders is prefixed with `Init`

src/Sulu/Bundle/SecurityBundle/Build/InitSecurityBuilder.php
```diff
@@ -0,0 +1,38 @@
...
+{
+    public function getName()
+    {
+        return 'init_security';
```
danrot, line 23: Would also call that just `security`.

src/Sulu/Bundle/SecurityBundle/Command/InitCommand.php
```diff
@@ -0,0 +1,114 @@
...
+use Sulu\Component\Security\Authentication\RoleRepositoryInterface;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+
+/**
+ * @internal
+ */
+final class InitCommand extends Command
```
danrot, line 27: +1 for making much more stuff final. Should that be the new default for most classes now? Or what was your reasoning for doing this? The `Builder` class above could e.g. also be `final`.

src/Sulu/Bundle/SecurityBundle/Command/InitCommand.php
```diff
@@ -0,0 +1,114 @@
...
+            ->setDescription('Create required sulu security entities.');
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $systems = \array_keys($this->adminPool->getSecurityContexts());
+        $roles = $this->roleRepository->findAllRoles();
+        $ui = new SymfonyStyle($input, $output);
+
+        $existAnonymousRoles = [];
```
danrot, line 70: `$existingAnonymousRoles`

src/Sulu/Bundle/SecurityBundle/Command/InitCommand.php
```diff
@@ -0,0 +1,114 @@
...
+
+        $count = 0;
+        foreach ($systems as $system) {
+            if (Admin::SULU_ADMIN_SECURITY_SYSTEM !== $system) {
+                continue;
+            }
+
+            if (isset($existAnonymousRoles[$system])) {
+                $ui->text(\sprintf(
+                    '[x] Anonymous role exist in system "%s" as "%s".',
```
danrot, line 85: "as" somehow sounds weird, resp. it is not clear what is meant by that... How about that?

"Anonymous role named "%s" exists in system "%s" already."

src/Sulu/Bundle/SecurityBundle/Command/InitCommand.php
```diff
@@ -0,0 +1,114 @@
...
+                continue;
+            }
+
+            /** @var RoleInterface $role */
+            $role = $this->roleRepository->createNew();
+            $role->setName('Anonymous User ' . $system);
+            $role->setAnonymous(true);
+            $role->setSystem($system);
+
+            $ui->text(\sprintf('[ ] Create anonymous role in system "%s" as "%s".', $system, $role->getName()));
```
danrot, line 99: Same as above:

"Create anonymous role named "%s" in system "%s". And shouldn't there be some kind of symbol between the brackets? I think the document manager init command writes an "*" or "+" when something is created.

src/Sulu/Bundle/SecurityBundle/Entity/UserRole.php
```diff
@@ -124,6 +124,15 @@ public function getUser()
      */
     public function setRole(RoleInterface $role)
     {
+        if ($role->getAnonymous()) {
+            throw new \LogicException(
```
danrot, line 128: Two questions here:

1. If it is not allowed to add an anonymous role to a user, does that mean there will be no anonymous user? I.e. that in the code where we need to check for anonymous, we are going to just load the anonymous role instead of getting the roles from a user object?
2. Shouldn't we introduce a more specific exception for that? Could imagine that the calling code might try to recover from that specific error. Would think about something like a `AssignAnonymousRoleException`

src/Sulu/Component/Security/Authentication/RoleInterface.php
```diff
@@ -176,4 +176,15 @@ public function getSecurityType();
      * @return RoleSettingInterface|null
      */
     public function getSetting($key);
+
+    /**
+     * Returns if the role is of type anonymous and is handle like IS_ANONYMOUS in Symfony.
```
danrot, line 181: "handled"

src/Sulu/Bundle/SecurityBundle/Command/InitCommand.php
alexander-schranz: Yeah think we should mark more classes final and internal where we don't want or can (in case of third party dependencies) provide any bc promises.

src/Sulu/Bundle/SecurityBundle/Command/InitCommand.php
alexander-schranz: Oh sure will also add the "Legend" from that command.

src/Sulu/Bundle/SecurityBundle/Entity/UserRole.php
alexander-schranz: 1. Yes I would think that there should be no Anonymous User. As I would think that would break also currently how we did handle anonymous User in project by just checking for `if (!$user instanceof SuluUserInterface)`

2. Sounds good for me 👍

src/Sulu/Bundle/SecurityBundle/Resources/config/services.xml
```diff
@@ -259,6 +259,10 @@
             <tag name="massive_build.builder" />
         </service>
 
+        <service id="sulu_security.init_security" class="Sulu\Bundle\SecurityBundle\Build\SecurityBuilder">
```
danrot, line 262: The service id should also be renamed.

alexander-schranz: Thx

PR #5436, opened 2020-07-29
Throw proper error message when invalid form XML is loaded

danrot:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR throws a proper exception when a file with a root tag not suitable for a form is interpreted as a form.

#### Why?

Because when people want to make use of XML Includes, they might want to create a `fragments` folder within `config/forms` and put them there. If they do so, it will result in a cryptic error like this:

```
Notice: Trying to get property 'nodeValue' of non-object
```

That's because the program tries to load the key from the form, but the fragment most often does not even have a `key` under a `form` tag, and that's why the code fails.

In order to make that at least a little bit more understandable, we throw a better error message now.

#### Steps to reproduce

Place a file like this in `config/forms/fragments`, and the cryptic error message should be shown:

```xml
<properties xmlns="http://schemas.sulu.io/template/template" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://schemas.sulu.io/template/template http://schemas.sulu.io/template/template-1.1.xsd">
    <property name="test" type="text_line" />
</properties>

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/FormMetadata/FormXmlLoader.php
+++ b/src/Sulu/Bundle/AdminBundle/FormMetadata/FormXmlLoader.php
@@ -11,6 +11,7 @@
 
 namespace Sulu\Bundle\AdminBundle\FormMetadata;
 
+use Sulu\Bundle\AdminBundle\Exception\InvalidRootTagException;
 use Sulu\Bundle\AdminBundle\FormMetadata\FormMetadata as ExternalFormMetadata;
 use Sulu\Bundle\AdminBundle\Metadata\FormMetadata\FormMetadata;
 use Sulu\Bundle\AdminBundle\Metadata\FormMetadata\LocalizedFormMetadataCollection;
@@ -68,6 +69,10 @@ class FormXmlLoader extends AbstractLoader
     {
         // init running vars
         $tags = [];
+
+        if (0 === $xpath->query('/x:form')->count()) {
+            throw new InvalidRootTagException($resource, 'form');
+        }
 
         $form = new ExternalFormMetadata();
         $form->setResource($resource);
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/Sulu/Bundle/AdminBundle/FormMetadata/FormXmlLoader.php [hunk 1]
```diff
@@ -69,6 +70,10 @@ protected function parse($resource, \DOMXPath $xpath, $type): LocalizedFormMetad
         // init running vars
         $tags = [];
 
+        if ($xpath->query('/x:form')->count() === 0) {
+            throw new InvalidRootTagException($resource, 'form');
```
danrot, line 74: Maybe we could simply even skip that file, but not sure if this will hide some bugs, that would be hard to find then :thinking:

alexander-schranz:
@danrot so If i understand you correctly the forms are read recursive from the `config/forms` folder? I personally see this as the issue because in case of `pages` I can create subfolders which are not read. For me only the configured folder should be read and subfolders be ignored, if I want a subfolder also be read I should configure it.

danrot:
@alexander-schranz I agree that it should work in the same way, but I don't think we can currently make this consistent in both cases, because whatever we change would be a BC break. Unless we introduce a Symfony config we will deprecate immediately, as we already did in the [`SuluWebsiteBundle` configuration](https://github.com/sulu/sulu/blob/1bd30e2c890aaaa94b1254816de6508fbae30417/src/Sulu/Bundle/WebsiteBundle/DependencyInjection/Configuration.php#L52). However, I think the change in here is worthwhile anyway, since it will also throw an error with a better explanation if somebody puts the wrong file here on accident.

PR #5437, opened 2020-07-29
Allow to secure media download on website

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes -
| Related issues/PRs | replaces parts of #5424 
| License | MIT
| Documentation PR | -

#### What's in this PR?

Add admin media download url.

#### Why?

That permission can be checked for a media be downloaded we need a admin media url.

#### To Do

- [x] Add Copy Admin Url
- [x] Add Copy Website Url
- [ ] Download History Both Urls?

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/Media/Manager/MediaManagerInterface.php
+++ b/src/Sulu/Bundle/MediaBundle/Media/Manager/MediaManagerInterface.php
@@ -17,6 +17,9 @@ use Sulu\Bundle\MediaBundle\Media\Exception\CollectionNotFoundException;
 use Sulu\Bundle\MediaBundle\Media\Exception\MediaNotFoundException;
 use Symfony\Component\HttpFoundation\File\UploadedFile;
 
+/**
+ * @method string getAdminUrl($id, $fileName, $version);
+ */
 interface MediaManagerInterface
 {
     /**
@@ -133,15 +136,4 @@ interface MediaManagerInterface
      * @return string
      */
     public function getUrl($id, $fileName, $version);
-
-    /**
-     * Returns admin download url for given id and filename.
-     *
-     * @param string|int $id
-     * @param string $fileName
-     * @param string|int $version
-     *
-     * @return string
-     */
-    public function getAdminUrl($id, $fileName, $version);
 }
--- a/src/Sulu/Bundle/MediaBundle/Media/Manager/MediaManagerInterface.php
+++ b/src/Sulu/Bundle/MediaBundle/Media/Manager/MediaManagerInterface.php
@@ -126,11 +126,22 @@ interface MediaManagerInterface
     /**
      * Returns download url for given id and filename.
      *
-     * @param string $id
+     * @param string|int $id
      * @param string $fileName
-     * @param string $version
+     * @param string|int $version
      *
      * @return string
      */
     public function getUrl($id, $fileName, $version);
+
+    /**
+     * Returns admin download url for given id and filename.
+     *
+     * @param string|int $id
+     * @param string $fileName
+     * @param string|int $version
+     *
+     * @return string
+     */
+    public function getAdminUrl($id, $fileName, $version);
 }
--- a/src/Sulu/Bundle/MediaBundle/Resources/js/containers/Form/tests/fields/SingleMediaUpload.test.js
+++ b/src/Sulu/Bundle/MediaBundle/Resources/js/containers/Form/tests/fields/SingleMediaUpload.test.js
@@ -295,10 +295,11 @@ test('Create a MediaUploadStore when constructed with data', () => {
         )
     );
     const data = {
+        adminUrl: '',
         id: 1,
         locale: 'en',
-        title: 'test title',
         mimeType: 'image/jpeg',
+        title: 'test',
         thumbnails: {},
         url: '',
     };
--- a/src/Sulu/Bundle/MediaBundle/Resources/js/containers/MediaVersionUpload/FocusPointOverlay.js
+++ b/src/Sulu/Bundle/MediaBundle/Resources/js/containers/MediaVersionUpload/FocusPointOverlay.js
@@ -89,7 +89,6 @@ class FocusPointOverlay extends React.Component<Props> {
 
     render() {
         const {open} = this.props;
-        const image = this.resourceStore.data.adminUrl ? this.resourceStore.data.adminUrl : this.resourceStore.data.url;
 
         return (
             <Overlay
@@ -105,7 +104,11 @@ class FocusPointOverlay extends React.Component<Props> {
                 <div className={focusPointOverlayStyles.focusPointContainer}>
                     {!!this.resourceStore &&
                         <ImageFocusPoint
-                            image={image}
+                            image={
+                                this.resourceStore.data.adminUrl
+                                    ? this.resourceStore.data.adminUrl
+                                    : this.resourceStore.data.url
+                            }
                             onChange={this.handleFocusPointChange}
                             value={{x: this.focusPointX, y: this.focusPointY}}
                         />
--- a/src/Sulu/Bundle/MediaBundle/Resources/js/containers/MediaVersionUpload/FocusPointOverlay.js
+++ b/src/Sulu/Bundle/MediaBundle/Resources/js/containers/MediaVersionUpload/FocusPointOverlay.js
@@ -89,6 +89,7 @@ class FocusPointOverlay extends React.Component<Props> {
 
     render() {
         const {open} = this.props;
+        const image = this.resourceStore.data.adminUrl ? this.resourceStore.data.adminUrl : this.resourceStore.data.url;
 
         return (
             <Overlay
@@ -104,7 +105,7 @@ class FocusPointOverlay extends React.Component<Props> {
                 <div className={focusPointOverlayStyles.focusPointContainer}>
                     {!!this.resourceStore &&
                         <ImageFocusPoint
-                            image={this.resourceStore.data.adminUrl}
+                            image={image}
                             onChange={this.handleFocusPointChange}
                             value={{x: this.focusPointX, y: this.focusPointY}}
                         />
--- a/src/Sulu/Bundle/MediaBundle/Resources/js/containers/MediaVersionUpload/FocusPointOverlay.js
+++ b/src/Sulu/Bundle/MediaBundle/Resources/js/containers/MediaVersionUpload/FocusPointOverlay.js
@@ -104,7 +104,7 @@ class FocusPointOverlay extends React.Component<Props> {
                 <div className={focusPointOverlayStyles.focusPointContainer}>
                     {!!this.resourceStore &&
                         <ImageFocusPoint
-                            image={this.resourceStore.data.url}
+                            image={this.resourceStore.data.adminUrl}
                             onChange={this.handleFocusPointChange}
                             value={{x: this.focusPointX, y: this.focusPointY}}
                         />
--- a/src/Sulu/Bundle/MediaBundle/Resources/js/containers/MediaVersionUpload/MediaVersionUpload.js
+++ b/src/Sulu/Bundle/MediaBundle/Resources/js/containers/MediaVersionUpload/MediaVersionUpload.js
@@ -108,6 +108,7 @@ class MediaVersionUpload extends React.Component<Props> {
 
         const {
             data: {
+                adminUrl,
                 previewImageId,
                 isImage,
                 url,
@@ -181,7 +182,7 @@ class MediaVersionUpload extends React.Component<Props> {
                 />
                 <CropOverlay
                     id={id}
-                    image={url}
+                    image={adminUrl ? adminUrl : url}
                     locale={locale.get()}
                     onClose={this.handleCropOverlayClose}
                     onConfirm={this.handleCropOverlayConfirm}
--- a/src/Sulu/Bundle/MediaBundle/Resources/js/containers/SingleMediaUpload/tests/SingleMediaUpload.test.js
+++ b/src/Sulu/Bundle/MediaBundle/Resources/js/containers/SingleMediaUpload/tests/SingleMediaUpload.test.js
@@ -11,7 +11,7 @@ jest.mock('../../../stores/MediaUploadStore', () => jest.fn(function(media) {
     this.update = jest.fn();
     this.delete = jest.fn();
     this.getThumbnail = jest.fn((size) => size);
-    this.downloadUrl = media ? media.url : undefined;
+    this.downloadUrl = media?.adminUrl || media?.url;
     this.media = media;
 }));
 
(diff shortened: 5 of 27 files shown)


src/Sulu/Bundle/MediaBundle/EventListener/MediaSecurityListener.php
```diff
@@ -0,0 +1,84 @@
...
+        $routeName = $request->attributes->get('_route');
+
+        if (!\in_array($routeName, ['sulu_media.website.media.download', 'sulu_media.website.media.download_admin'])) {
+            return;
+        }
+
+        $collectionId = $this->collectionRepository->findIdByMediaId((int) $request->attributes->get('id'));
+
+        if (!$collectionId) {
+            throw new NotFoundHttpException();
```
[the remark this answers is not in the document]
alexander-schranz: As the media can not exist without a collection a NotFoundHttpException is correct here, as its common that a old media url could still exist in some http cache or email. And if that media not exist it should return 404.

The other way would be that we don't throw any exception and just do a early return to avoid security check, then it would land in the MediaController but as the Media does not exist it will also correctly there throw a 404 Exception.

src/Sulu/Bundle/MediaBundle/EventListener/MediaSecurityListener.php
```diff
@@ -0,0 +1,84 @@
...
+    {
+        $request = $event->getRequest();
+
+        if (!$event->isMasterRequest()) {
+            return;
+        }
+
+        $routeName = $request->attributes->get('_route');
+
+        if (!\in_array($routeName, ['sulu_media.website.media.download', 'sulu_media.website.media.download_admin'])) {
```
danrot, line 64: Isn't that code called now for *every request*? And both of these routes are referring to the same controller, right? Wouldn't it be better performance-wise to do that in the controller instead?

src/Sulu/Bundle/MediaBundle/Resources/js/containers/MediaVersionUpload/MediaVersionUpload.js
```diff
@@ -181,7 +182,7 @@ class MediaVersionUpload extends React.Component<Props> {
                 />
                 <CropOverlay
                     id={id}
-                    image={url}
+                    image={adminUrl ? adminUrl : url}
```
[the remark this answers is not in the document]
alexander-schranz: Its BC we can not be sure if the Media API was changed or MediaManager was changed by the Project so we always fallback to the url.

src/Sulu/Bundle/MediaBundle/Resources/js/containers/SingleMediaUpload/tests/SingleMediaUpload.test.js
```diff
@@ -4,24 +4,39 @@ import {observable} from 'mobx';
...
+import type {Media} from '../../../types';
 
 jest.mock('../../../stores/MediaUploadStore', () => jest.fn(function(media) {
     this.id = media ? media.id : undefined;
     this.create = jest.fn();
     this.update = jest.fn();
     this.delete = jest.fn();
     this.getThumbnail = jest.fn((size) => size);
-    this.downloadUrl = media ? media.url : undefined;
+    this.downloadUrl = media ? (media.adminUrl ? media.adminUrl : (media.url ? media.url : undefined)) : undefined;
```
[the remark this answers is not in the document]
alexander-schranz: Oh nice didn't know about this :)

src/Sulu/Bundle/MediaBundle/Resources/js/stores/MediaUploadStore/MediaUploadStore.js
```diff
@@ -38,7 +38,11 @@ export default class MediaUploadStore {
             return undefined;
         }
 
-        return media.url;
+        if (!media.adminUrl) {
+            return media.url;
```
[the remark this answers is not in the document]
alexander-schranz: We can not be sure if he did call `setAdminUrl` if he did override MediaManager or MediaController

src/Sulu/Bundle/MediaBundle/Tests/Functional/Controller/MediaFormatControllerTest.php
```diff
@@ -225,6 +225,44 @@ public function testPutWithFormatOptions()
         $this->assertEquals(100, $response->{'small-inset'}->cropHeight);
     }
 
+    public function testPutWithFormatOptionsAsDecimal()
```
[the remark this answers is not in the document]
alexander-schranz: Ups yes false PR 🙈

src/Sulu/Bundle/MediaBundle/EventListener/MediaSecurityListener.php
alexander-schranz: If somebody did override the route/controller, the security check would then not longer work.

I thought about using attributes instead maybe so I could check for `$request->attributes->getBoolean('media_security');` so I can enable the security check on the route definition a default attribute. What do you think?

danrot: I think moving this check to a separate listener is very confusing, when it could also be attached to a single controller. And in addition to that it has a negative impact on the entire system, because this check has to be done for every single route. For these reasons I would not do it this way, but maybe add it to the `UPGRADE.md` instead, so that people know they have to add the security check.

Using attributes instead does not improve performance, does it? And it introduces even more indirection.

PS: I think we should make a lot more stuff `final` :see_no_evil: Would even make it the new default.

src/Sulu/Bundle/MediaBundle/Resources/js/containers/Form/tests/fields/SingleMediaUpload.test.js
```diff
@@ -25,6 +26,20 @@ jest.mock('sulu-admin-bundle/stores/userStore', () => ({
...
 
+function createMedia(data = {}): Media {
+    const exampleMedia = {
+        adminUrl: '',
+        id: 1,
+        locale: 'en',
+        mimeType: 'image/jpeg',
+        title: 'test',
+        thumbnails: {},
+        url: '',
```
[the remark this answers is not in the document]
alexander-schranz: Can change it but I think its just annoying edit all tests everytime a new property is added. Would be better to have a media class in js instead just a type but that should maybe then be the case for every type which represent a entity.

src/Sulu/Bundle/MediaBundle/EventListener/MediaSecurityListener.php
alexander-schranz: As discussed in the daily I will refractor it into the controller

alexander-schranz:
@danrot This is ready for review again.

src/Sulu/Bundle/MediaBundle/Resources/js/containers/Form/tests/fields/SingleMediaUpload.test.js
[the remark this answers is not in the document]
danrot: I get that, but I have just had to touch a few tests in PHP, which were written in a similar way, and they are a pain to maintain, because everytime you touch something way too much tests fail :confused:

For a similar reason we have introduced files like [`fieldTypeDefaultProps`](https://github.com/sulu/sulu/blob/9a8cf157bd8a7b1997d2d39ba4d3941aa0f14cd3/src/Sulu/Bundle/AdminBundle/Resources/js/utils/TestHelper/fieldTypeDefaultProps.js), which only contain *very* default values (almost only only empty objects and undefined). This way you don't have to write everything out, but still might save one or the other keystroke. And the tests practically don't share any data.

danrot:
I somehow can still download the image even if I create a firewall and add permissions to a collection... I think that's caused by some services missing, had the same issue in #5439: https://github.com/sulu/sulu/pull/5439/commits/7292f3d5cc213b1c3c81d5fb4fb91f9d4382067c#diff-f93835774e15516577978298956f1f12

If you can copy this change over, it is also less for you to review in the other PR :see_no_evil:

src/Sulu/Bundle/MediaBundle/Tests/Functional/Controller/MediaStreamControllerAdminTest.php
```diff
@@ -0,0 +1,135 @@
...
+            $this->assertSame(MediaAdmin::SECURITY_CONTEXT, $securityCondition->getSecurityContext());
+            $this->assertSame(null, $securityCondition->getLocale());
+            $this->assertSame(Collection::class, $securityCondition->getObjectType());
+            $this->assertSame($media->getCollection(), $securityCondition->getObjectId());
+
+            return true;
+        }), PermissionTypes::VIEW)
+            ->shouldBeCalled();
+
+        self::$container->set('sulu_security.security_checker', $securityChecker->reveal());
```
danrot, line 64: Shouldn't we make that a fully functional tests? And same for the same controller in the website context? Then the bug that downloading still works would have been recognized.

alexander-schranz: @danrot Need to check that but our TestVoter currently does not allow to do security specific tests.

danrot: Worst case scenario would be to do something like I already did in #5439: https://github.com/sulu/sulu/pull/5439/commits/557c5b8c227b017b050a10dd5f4e9a73b93d2af1#diff-b5ddd621eded67d076a89bcb768bef31R31

danrot:
Looks good to me now, the only thing missing is a functional test :slightly_smiling_face:

src/Sulu/Bundle/MediaBundle/Tests/Functional/Controller/MediaStreamControllerAdminTest.php
alexander-schranz: Test added beneath

src/Sulu/Bundle/MediaBundle/Media/Manager/MediaManagerInterface.php
```diff
@@ -126,11 +126,22 @@ public function addFormatsAndUrl(Media $media);
...
+    /**
+     * Returns admin download url for given id and filename.
+     *
+     * @param string|int $id
+     * @param string $fileName
+     * @param string|int $version
+     *
+     * @return string
+     */
+    public function getAdminUrl($id, $fileName, $version);
```
danrot, line 146: Since we have been very careful about BC breaks in #5439 we probably also should not add this method to the interface, because this is a BC break as well.

src/Sulu/Bundle/MediaBundle/Resources/js/containers/MediaVersionUpload/FocusPointOverlay.js
```diff
@@ -104,7 +104,7 @@ class FocusPointOverlay extends React.Component<Props> {
                 <div className={focusPointOverlayStyles.focusPointContainer}>
                     {!!this.resourceStore &&
                         <ImageFocusPoint
-                            image={this.resourceStore.data.url}
+                            image={this.resourceStore.data.adminUrl}
```
danrot, line 107: Why don't we need the fallback here but in other places?

src/Sulu/Bundle/TestBundle/Testing/TestUserProvider.php
```diff
@@ -58,27 +61,34 @@ class TestUserProvider implements UserProviderInterface
...
         $this->userPasswordEncoderFactory = $userPasswordEncoderFactory;
         $this->userProvider = $userProvider;
+        $this->roleRepository = $roleRepository;
     }
 
-    public function getUser()
+    public function getUser(string $username = self::TEST_USER_USERNAME)
     {
-        if ($this->user) {
+        if ($this->user && self::TEST_USER_USERNAME === $username) {
```
danrot, line 87: Don't you think that this is quite confusing? This class was only supposed to load a test user, now it being "abused" to create other users as well, and just caching one of them. For now I would keep that part in the tests, and if we decide at one point that we want to outsource that part, it should probably be a different service IMO.

src/Sulu/Bundle/TestBundle/Testing/TestUserProvider.php
```diff
@@ -87,21 +97,46 @@ public function getUser()
...
 
-        $this->user = $user;
+        if (self::TEST_USER_USERNAME === $username) {
+            $this->user = $user;
+        }
+
+        return $user;
+    }
+
+    public function getRole(
```
danrot, line 119: As above, I would not include it in this class, because it is not really its responsibility to create roles. The implementation is also a bit confusing IMO, because it's not really a getter nor a creator. Calling it somehow like would cause the second line to change the role of the first line as well:

```php
$role1 = $testUserProvider->getRole('Test', 'sulu', false);
$role2 = $testUserProvider->getRole('Test', 'website', true);
```

src/Sulu/Bundle/TestBundle/Testing/TestUserTrait.php
```diff
@@ -11,19 +11,32 @@
...
 {
     /**
-     * Return the test user (which is provided / created
+     * Return a test user (which is provided / created
      * by the test_user_provider in this Bundle at runtime).
      *
      * @return User
      */
-    protected static function getTestUser()
+    protected static function getTestUser(string $username = TestUserProvider::TEST_USER_USERNAME)
```
danrot, line 26: I think I would leave that file untouched, and create the entities in the classes as necessary, probably creating a `createRole` and/or `createUser` function as required. Then we don't have any side effects among different test cases, which are hard to fix.

src/Sulu/Bundle/TestBundle/Resources/config/test_user_provider.xml
```diff
@@ -12,6 +12,7 @@
             <argument type="service" id="sulu.repository.user"/>
             <argument type="service" id="security.encoder_factory"/>
             <argument type="service" id="sulu_security.user_provider"/>
+            <argument type="service" id="sulu.repository.role"/>
```
danrot, line 15: That's not needed anymore, right?

src/Sulu/Bundle/MediaBundle/Resources/js/containers/MediaVersionUpload/FocusPointOverlay.js
```diff
@@ -89,6 +89,7 @@ class FocusPointOverlay extends React.Component<Props> {
 
     render() {
         const {open} = this.props;
+        const image = this.resourceStore.data.adminUrl ? this.resourceStore.data.adminUrl : this.resourceStore.data.url;
```
danrot, line 92: This errs in flow, the check would be easier if we use destructuring here.

PR #5439, opened 2020-07-30
Hide elements without permission on website

### LINKED ISSUE #4051: Allow to set a object security for other systems than Sulu itself

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #4051
| Related issues/PRs | #5406 
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR will hide elements for which the currently logged in user does not have any permissions.

#### Why?

Because that is the reason you apply security in the first place.

#### BC Breaks/Deprecations

TODO

#### To Do

- [x] Hide navigation items without sufficient permission
- [x] Hide pages from smart content without sufficient permissions
- [x] Hide pages from a selection field without sufficient permissions
- [x] Hide pages from a single selection field without sufficient permissions
- [x] Hide entities from smart content without sufficient permissions
- [x] Hide entities from a selection field without sufficient permissions
- [x] Hide entities from a single selection field without sufficient permissions
- [ ] Check for better solution than copying `getUser` function
- [x] Adjust `AccessControlManager` to fallback to anonymous role if no user is passed
- [x] Make sure the `MediaDataProvider` uses the collection to check permissions
- [x] Hide pages on sitemap not visible to anonymous users
- [x] `DataProviderRepositoryTrait` should also work without a user
- [ ] `DataProviderRepositoryTrait` could work only with `QueryBuilder` instead of returning the end result to be more flexible and reuse a known interface (not implementable now for BC reasons, but create an issue?)
- [x] Deprecate `SecuredEntityRepositoryTrait` and create a `AccessControlQueryEnhancer` service instead
- [x] Inject the `AccessControlQueryEnhancer` service into the `MediaDataRepository` so that it is available to the `DataProviderRepositoryTrait`
- [x] Lazily load anonymous role for current system in `SystemStore` to avoid multiple queries

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/Content/Types/MediaSelectionContentType.php
+++ b/src/Sulu/Bundle/MediaBundle/Content/Types/MediaSelectionContentType.php
@@ -20,7 +20,9 @@ use Sulu\Component\Content\Compat\PropertyParameter;
 use Sulu\Component\Content\ComplexContentType;
 use Sulu\Component\Content\ContentTypeExportInterface;
 use Sulu\Component\Content\PreResolvableContentTypeInterface;
+use Sulu\Component\Security\Authorization\PermissionTypes;
 use Sulu\Component\Util\ArrayableInterface;
+use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface;
 
 /**
  * content type for image selection.
@@ -37,10 +39,26 @@ class MediaSelectionContentType extends ComplexContentType implements ContentTyp
      */
     private $referenceStore;
 
-    public function __construct(MediaManagerInterface $mediaManager, ReferenceStoreInterface $referenceStore)
-    {
+    /**
+     * @var TokenStorageInterface
+     */
+    private $tokenStorage;
+
+    /**
+     * @var array
+     */
+    private $permissions;
+
+    public function __construct(
+        MediaManagerInterface $mediaManager,
+        ReferenceStoreInterface $referenceStore,
+        TokenStorageInterface $tokenStorage = null,
+        array $permissions = null
+    ) {
         $this->mediaManager = $mediaManager;
         $this->referenceStore = $referenceStore;
+        $this->tokenStorage = $tokenStorage;
+        $this->permissions = $permissions;
     }
 
     public function getDefaultParams(PropertyInterface $property = null)
@@ -120,7 +138,9 @@ class MediaSelectionContentType extends ComplexContentType implements ContentTyp
             isset($data['ids']) ? $data['ids'] : [],
             $property->getStructure()->getLanguageCode(),
             $types,
-            $this->mediaManager
+            $this->mediaManager,
+            $this->getUser(),
+            $this->permissions[PermissionTypes::VIEW]
         );
 
         return $container->getData();
@@ -167,5 +187,20 @@ class MediaSelectionContentType extends ComplexContentType implements ContentTyp
         foreach ($data['ids'] as $id) {
             $this->referenceStore->add($id);
         }
+    }
+
+    private function getUser()
+    {
+        if (!$this->tokenStorage) {
+            return null;
+        }
+
+        $user = $this->tokenStorage->getToken()->getUser();
+
+        if (\is_object($user)) {
+            return $user;
+        }
+
+        return null;
     }
 }
--- a/src/Sulu/Bundle/MediaBundle/Content/Types/MediaSelectionContentType.php
+++ b/src/Sulu/Bundle/MediaBundle/Content/Types/MediaSelectionContentType.php
@@ -40,7 +40,7 @@ class MediaSelectionContentType extends ComplexContentType implements ContentTyp
     private $referenceStore;
 
     /**
-     * @var TokenStorageInterface
+     * @var ?TokenStorageInterface
      */
     private $tokenStorage;
 
--- a/src/Sulu/Bundle/MediaBundle/Content/Types/MediaSelectionContentType.php
+++ b/src/Sulu/Bundle/MediaBundle/Content/Types/MediaSelectionContentType.php
@@ -20,7 +20,7 @@ use Sulu\Component\Content\Compat\PropertyParameter;
 use Sulu\Component\Content\ComplexContentType;
 use Sulu\Component\Content\ContentTypeExportInterface;
 use Sulu\Component\Content\PreResolvableContentTypeInterface;
-use Sulu\Component\Security\Authentication\UserInterface;
+use Sulu\Component\Security\Authorization\PermissionTypes;
 use Sulu\Component\Util\ArrayableInterface;
 use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface;
 
@@ -44,14 +44,21 @@ class MediaSelectionContentType extends ComplexContentType implements ContentTyp
      */
     private $tokenStorage;
 
+    /**
+     * @var ?array
+     */
+    private $permissions;
+
     public function __construct(
         MediaManagerInterface $mediaManager,
         ReferenceStoreInterface $referenceStore,
-        TokenStorageInterface $tokenStorage = null
+        TokenStorageInterface $tokenStorage = null,
+        $permissions = null
     ) {
         $this->mediaManager = $mediaManager;
         $this->referenceStore = $referenceStore;
         $this->tokenStorage = $tokenStorage;
+        $this->permissions = $permissions;
     }
 
     public function getDefaultParams(PropertyInterface $property = null)
@@ -132,7 +139,7 @@ class MediaSelectionContentType extends ComplexContentType implements ContentTyp
             $property->getStructure()->getLanguageCode(),
             $types,
             $this->mediaManager,
-            $this->getUser()
+            $this->permissions[PermissionTypes::VIEW]
         );
 
         return $container->getData();
@@ -179,20 +186,5 @@ class MediaSelectionContentType extends ComplexContentType implements ContentTyp
         foreach ($data['ids'] as $id) {
             $this->referenceStore->add($id);
         }
-    }
-
-    private function getUser(): ?UserInterface
-    {
-        if (!$this->tokenStorage) {
-            return null;
-        }
-
-        $user = $this->tokenStorage->getToken()->getUser();
-
-        if ($user instanceof UserInterface) {
-            return $user;
-        }
-
-        return null;
     }
 }
--- a/src/Sulu/Bundle/MediaBundle/Content/Types/MediaSelectionContentType.php
+++ b/src/Sulu/Bundle/MediaBundle/Content/Types/MediaSelectionContentType.php
@@ -20,7 +20,7 @@ use Sulu\Component\Content\Compat\PropertyParameter;
 use Sulu\Component\Content\ComplexContentType;
 use Sulu\Component\Content\ContentTypeExportInterface;
 use Sulu\Component\Content\PreResolvableContentTypeInterface;
-use Sulu\Component\Security\Authorization\PermissionTypes;
+use Sulu\Component\Security\Authentication\UserInterface;
 use Sulu\Component\Util\ArrayableInterface;
 use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface;
 
@@ -44,21 +44,14 @@ class MediaSelectionContentType extends ComplexContentType implements ContentTyp
      */
     private $tokenStorage;
 
-    /**
-     * @var array
-     */
-    private $permissions;
-
     public function __construct(
         MediaManagerInterface $mediaManager,
         ReferenceStoreInterface $referenceStore,
-        TokenStorageInterface $tokenStorage = null,
-        array $permissions = null
+        TokenStorageInterface $tokenStorage = null
     ) {
         $this->mediaManager = $mediaManager;
         $this->referenceStore = $referenceStore;
         $this->tokenStorage = $tokenStorage;
-        $this->permissions = $permissions;
     }
 
     public function getDefaultParams(PropertyInterface $property = null)
@@ -139,8 +132,7 @@ class MediaSelectionContentType extends ComplexContentType implements ContentTyp
             $property->getStructure()->getLanguageCode(),
             $types,
             $this->mediaManager,
-            $this->getUser(),
-            $this->permissions[PermissionTypes::VIEW]
+            $this->getUser()
         );
 
         return $container->getData();
@@ -189,7 +181,7 @@ class MediaSelectionContentType extends ComplexContentType implements ContentTyp
         }
     }
 
-    private function getUser()
+    private function getUser(): ?UserInterface
     {
         if (!$this->tokenStorage) {
             return null;
@@ -197,7 +189,7 @@ class MediaSelectionContentType extends ComplexContentType implements ContentTyp
 
         $user = $this->tokenStorage->getToken()->getUser();
 
-        if (\is_object($user)) {
+        if ($user instanceof UserInterface) {
             return $user;
         }
 
--- a/src/Sulu/Bundle/MediaBundle/Entity/MediaDataProviderRepository.php
+++ b/src/Sulu/Bundle/MediaBundle/Entity/MediaDataProviderRepository.php
@@ -16,6 +16,8 @@ use Doctrine\ORM\Query\Expr\Join;
 use Doctrine\ORM\QueryBuilder;
 use Sulu\Bundle\MediaBundle\Api\Media as MediaApi;
 use Sulu\Bundle\MediaBundle\Media\Manager\MediaManagerInterface;
+use Sulu\Bundle\SecurityBundle\AccessControl\AccessControlQueryEnhancer;
+use Sulu\Component\Security\Authentication\UserInterface;
 use Sulu\Component\SmartContent\Orm\DataProviderRepositoryInterface;
 use Sulu\Component\SmartContent\Orm\DataProviderRepositoryTrait;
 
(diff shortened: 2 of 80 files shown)


danrot:
I am still wondering about [my other comment](https://github.com/sulu/sulu/issues/4051#issuecomment-665044976). 

We somehow have to get the current user in order to do proper queries in the `ContentQueryExecutor` and in the `DataProviderRepositoryTrait`.

In the `ContentQueryExecutor` it would be easier to simply use the `TokenStorage` service in there and use it to retrieve the user and its permissions. However, that is quite a tight coupling, and is not very flexible, since it *only* works with the logged in user. Symfony provides some kind of impersonation, but not sure if this always helps. The alternative is to move the `TokenStorage` on a higher level, and pass the current user as a argument. That would be more flexible, but might require to be done in multiple places (currently the `NavigationTwigExtension` and the `PageDataProvider`).

However, for the `DataProviderRepositoryTrait` it has to be passed as a argument, because it can't be used as a service itself. And if we move it to the next higher level, i.e. that repositories, we have to inject the `TokenStorage` service in every repository, which is also cumbersome.

Therefore I would say it makes sense to inject the `TokenStorage` service in the `NavigationTwigExtension` and the `ContentType` class of the `SmartContent`. This way we might need to duplicate some code, but it feels to be less than in the other case, because for the `SmartContent` we only need it in one class this way. Otherwise the `PageDataProvider` gets that information either from somewhere else and therefore works differently than the other `DataProvider`s, or we have to duplicate that injection for every repository.

src/Sulu/Bundle/MediaBundle/Entity/MediaDataProviderRepository.php
```diff
@@ -74,7 +90,17 @@ public function findByFilters($filters, $page, $pageSize, $limit, $locale, $opti
...
+        $entities = $this->parentFindByFilters(
+            $filters,
+            $page,
+            $pageSize,
+            $limit,
+            $locale,
+            $options,
+            $user,
+            Collection::class,
+            'collection'
```
danrot, line 102: In order to activate permissions for other entities, they also have to override the default `findByFilters` method in this way, in order to pass the last two arguments. Is that good enough?

src/Sulu/Bundle/WebsiteBundle/Twig/Navigation/NavigationTwigExtension.php
```diff
@@ -38,14 +40,21 @@ class NavigationTwigExtension extends AbstractExtension implements NavigationTwi
      */
     private $requestAnalyzer;
 
+    /**
+     * @var TokenStorageInterface
```
alexander-schranz, line 44: Can be null:

```suggestion
     * @var TokenStorageInterface|null
```

src/Sulu/Component/Rest/ListBuilder/Doctrine/DoctrineListBuilder.php
```diff
@@ -33,15 +34,13 @@
...
 use Sulu\Component\Security\Authentication\UserInterface;
-use Sulu\Component\Security\Authorization\AccessControl\SecuredEntityRepositoryTrait;
 use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 
 /**
  * The listbuilder implementation for doctrine.
  */
 class DoctrineListBuilder extends AbstractListBuilder
 {
-    use SecuredEntityRepositoryTrait;
```
alexander-schranz, line 44: Not sure about removing this here if somebody did extend it. Because I remember that this was not always created over the factory or at least did see code where it was instanced manually.

src/Sulu/Component/Rest/ListBuilder/Doctrine/DoctrineListBuilder.php
```diff
@@ -418,7 +424,7 @@ protected function createSubQueryBuilder($select = null)
         $queryBuilder = $this->createQueryBuilder($addJoins)->select($select);
 
         if ($this->user && $this->permission && \array_key_exists($this->permission, $this->permissions)) {
-            $this->addAccessControl(
+            $this->accessControlQueryEnhancer->enhance(
```
alexander-schranz, line 427: Would keep the trait and fallback to it if the `accessControlQueryEnhancer` is not set.

src/Sulu/Component/SmartContent/ContentType.php
```diff
@@ -442,4 +454,19 @@ protected function resolveTags(&$value, $key)
...
+
+    private function getUser()
+    {
+        if (!$this->tokenStorage) {
+            return null;
+        }
+
+        $user = $this->tokenStorage->getToken()->getUser();
+
+        if (\is_object($user)) {
```
alexander-schranz, line 466: Not sure but think our smart content can only work with UserInterface User so we should only return that kind of users:

```suggestion
        if ($user instanceof UserInterface) {
```

src/Sulu/Component/SmartContent/ContentType.php
```diff
@@ -442,4 +454,19 @@ protected function resolveTags(&$value, $key)
             $value[$key] = $ids;
         }
     }
+
+    private function getUser()
```
alexander-schranz, line 458:
```suggestion
    private function getUser(): ?UserInterface
```

src/Sulu/Bundle/MediaBundle/Media/Manager/MediaManager.php
```diff
@@ -212,10 +212,16 @@ public function getEntityById($id)
         return $mediaEntity;
     }
 
-    public function getByIds(array $ids, $locale)
+    public function getByIds(array $ids, $locale, UserInterface $user = null, $permission = null)
```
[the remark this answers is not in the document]
danrot: I just hope that nobody relies on the current behavior...

src/Sulu/Bundle/PageBundle/Sitemap/PagesSitemapProvider.php
```diff
@@ -85,6 +94,17 @@ public function build($page, $scheme, $host)
                     continue;
                 }
 
+                $userPermissions = $this->accessControlManager->getUserPermissionByArray(
+                    $contentPage->getLocale(),
+                    PageAdmin::SECURITY_CONTEXT_PREFIX . $contentPage->getWebspaceKey(),
+                    $contentPage->getPermissions(),
```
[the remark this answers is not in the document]
danrot: It would be quite hard to write a correct query, because in PHPCR we have only the `sec:role-<id>` properties, and we don't know which of the roles are for the correct system.

But this is also not such a big problem, because the `$contentPage` already contains the permissions, as it has already loaded the entire document anyway. And the `getUserPermissionByArray` method only uses that information, and does not do any requests to the `se_access_controls` table (the access control for pages are also not stored there).

src/Sulu/Bundle/MediaBundle/Content/Types/MediaSelectionContentType.php
```diff
@@ -168,4 +188,19 @@ public function preResolve(PropertyInterface $property)
...
+
+    private function getUser()
+    {
+        if (!$this->tokenStorage) {
+            return null;
+        }
+
+        $user = $this->tokenStorage->getToken()->getUser();
+
+        if (\is_object($user)) {
```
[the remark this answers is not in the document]
danrot: I've copied that part from somewhere else, so it has already be done this way in other places. Fixed it there too.

src/Sulu/Bundle/PageBundle/Sitemap/PagesSitemapProvider.php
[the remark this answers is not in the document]
alexander-schranz: Forgot about that Document has the Permissions directly. So this is no problem for me than.

src/Sulu/Bundle/MediaBundle/Media/Manager/MediaManager.php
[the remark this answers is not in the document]
alexander-schranz: Old Behaviour. Did I oversee something? There was no user before in the used interfaces or? Then I would keep it as it is here currently.

src/Sulu/Bundle/PersistenceBundle/DependencyInjection/PersistenceExtensionTrait.php
```diff
@@ -83,6 +83,14 @@ private function getRepositoryDefinition($object, array $services, ContainerBuil
             $this->getClassMetadataDefinition($services['model']),
         ]);
 
+        $repositoryReflectionClass = new \ReflectionClass($repositoryClass);
+        if ($repositoryReflectionClass->hasMethod('setAccessControlQueryEnhancer')) {
```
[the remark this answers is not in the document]
danrot: Shouldn't make any difference, because it is very unlikely that this repository would implement the `setAccessControlQueryEnhancer` method then. However, still added the check, just to be sure.

src/Sulu/Bundle/MediaBundle/Media/Manager/MediaManager.php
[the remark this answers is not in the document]
danrot: No, it was not, but it also `null` differently now. People might expect to get all the entities anyway, but if there is an anonymous role with some missing permissions, this function might not return all the passed ids anymore.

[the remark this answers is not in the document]
alexander-schranz: Okay I understand no then we should definitely send the $permission else it will accidently filter out media when using the command from the CLI context.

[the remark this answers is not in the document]
danrot: I think I would keep it without the `permission` variable for now... Why should somebody use this method to get all the media? There are other methods for that... And as long as they are not introducing any permissions on the security it will behave the same, so they probably cannot even tell if that is a BC break :see_no_evil:

danrot:
In order to keep the behavior we should add a `$permission` variable to all get functions of the following classes:

- `CollectionManager`
- `MediaManager´
- `ContentQueryExecutor`

This is necessary because the behavior of the `null` user changed. It represents the anonymous user now, which might not have permissions for all entities/documents. Before that, the absence of the `$user` variable meant that no security check was taking place, an option that is not available otherwise anymore. With a `$permission` variable that can still be achieved by omitting it, and if a `$user` of null and a `$permission` value is given, then this permission is checked for the anonymous user.

src/Sulu/Bundle/PageBundle/Sitemap/PagesSitemapProvider.php
```diff
@@ -42,14 +44,21 @@ class PagesSitemapProvider extends AbstractSitemapProvider
      */
     private $environment;
 
+    /**
+     * @var ?AccessControlManagerInterface
```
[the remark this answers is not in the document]
danrot: It is already sometimes used in our codebase, and even in phpstan's codebase, so it should understand it. Apart from that I would prefer that way of writing it, because it is shorter and also supported by PHP itself (other than the `|null` approach, which does not work in PHP currently).

danrot:
The `CollectionManager` does not pass the user, but uses the `TokenStorage` to retrieve it. Shouldn`t the `MediaManager` do exactly the same? Initially I thought we need that to avoid that certain entities are not loaded in a CLI context, but we could do that using the `$permission` parameter now.

So the new methods would look like this:

- `CollectionManager::getById($id, $locale, $depth = 0, $breadcrumb = false, $filter = [], $sortBy = [], $children = false, $permission = null)`
- `MediaManager::getByIds(array $ids, $locale, $permission = null)`
- `MediaManager::get($locale, $filter = [], $limit = null, $offset = null, $permission = null)`
- `ContentQueryExecutor::execute($webspaceKey, $locales, ContentQueryBuilderInterface $contentQueryBuilder, $flat = true, $depth = -1, $limit = null, $offset = null, $moveUp = false, $permission = null)

It might feel a bit weird for the `MediaManager`, because in other places the `$userId` is passed instead (e.g. its `save` method), but that inconsistency was already there before this PR.

src/Sulu/Bundle/MediaBundle/Content/Types/MediaSelectionContentType.php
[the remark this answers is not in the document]
alexander-schranz: Maybe it would be better to move the whole `UserInterface` check into the SecurityChecker and remove the `UserInterface` typehints on the other places.

src/Sulu/Bundle/MediaBundle/Media/Manager/MediaManager.php
```diff
@@ -170,7 +164,6 @@ public function __construct(
         TokenStorageInterface $tokenStorage = null,
         SecurityCheckerInterface $securityChecker = null,
         FFProbe $ffprobe = null,
-        $permissions,
```
alexander-schranz, line 173: Removing this would be a bc break if somebody override the service and had called `parent::__construct` and expect permissions to given in its service definition.

src/Sulu/Bundle/MediaBundle/Media/Manager/MediaManagerInterface.php
```diff
@@ -30,7 +30,7 @@ interface MediaManagerInterface
      *
      * @return Media[]
      */
-    public function get($locale, $filter = [], $limit = null, $offset = null);
+    public function get($locale, $filter = [], $limit = null, $offset = null /* $permission = null */);
```
alexander-schranz, line 33: Missing comma

```suggestion
    public function get($locale, $filter = [], $limit = null, $offset = null /*, $permission = null */);
```

src/Sulu/Bundle/MediaBundle/Media/Manager/MediaManagerInterface.php
```diff
@@ -65,7 +65,7 @@ public function getEntityById($id);
      *
      * @return Media[]
      */
-    public function getByIds(array $ids, $locale);
+    public function getByIds(array $ids, $locale /* $permission = null */);
```
alexander-schranz, line 68: Missing comma:

```suggestion
    public function getByIds(array $ids, $locale /*, $permission = null */);
```

src/Sulu/Bundle/MediaBundle/Resources/config/services.xml
```diff
@@ -292,7 +298,6 @@
             <argument type="service" id="doctrine.orm.entity_manager" />
             <argument type="service" id="security.token_storage" on-invalid="null"/>
             <argument type="string">%sulu_media.collection.previews.format%</argument>
-            <argument>%sulu_security.permissions%</argument>
```
alexander-schranz, line 295: Need to be keeped here for backward compatibility.

src/Sulu/Bundle/WebsiteBundle/Navigation/NavigationMapperInterface.php
```diff
@@ -36,7 +36,8 @@ public function getNavigation(
         $depth = 1,
         $flat = false,
         $context = null,
-        $loadExcerpt = false
+        $loadExcerpt = false,
+        $segmentKey = null
```
alexander-schranz, line 40: Was this committed accidentally? Seems this change  has nothing todo with this PR?

src/Sulu/Bundle/WebsiteBundle/Navigation/NavigationMapperInterface.php
```diff
@@ -57,7 +58,8 @@ public function getRootNavigation(
...
```
alexander-schranz, line 62: Was this committed accidentally? Seems this change  has nothing todo with this PR?

src/Sulu/Bundle/WebsiteBundle/Resources/config/services.xml
```diff
@@ -122,6 +123,7 @@
             <argument type="service" id="sulu.content.mapper"/>
             <argument type="service" id="sulu_website.navigation_mapper"/>
             <argument type="service" id="sulu_core.webspace.request_analyzer" on-invalid="ignore"/>
+            <argument type="service" id="security.token_storage" on-invalid="ignore"/>
```
alexander-schranz, line 126: I personally would use `ignore` only if we use setter Injection. Find it a little bit confusing when used with constructor injection, because it sounds for me like the whole service  definition is then ignored, but it behaves in constructor the same as `on-invalid="null",  so I would go with `on-invalid="null"` in constructor injection. But we can also keep it this way, just did confuse me at the first if what the difference is here.

src/Sulu/Component/Content/Mapper/ContentMapper.php
```diff
@@ -144,7 +158,9 @@ public function __construct(
         SessionManagerInterface $sessionManager,
         EventDispatcherInterface $eventDispatcher,
         ResourceLocatorStrategyPoolInterface $resourceLocatorStrategyPool,
-        NamespaceRegistry $namespaceRegistry
+        NamespaceRegistry $namespaceRegistry,
+        AccessControlManagerInterface $accessControlManager,
```
alexander-schranz, line 162: This can be null if website has no security.

src/Sulu/Component/Content/Mapper/ContentMapper.php
```diff
@@ -662,6 +689,20 @@ private function rowToArray(Row $row, $locale, $webspaceKey, $fields, $onlyPubli
             return false;
         }
 
+        if ($document instanceof SecurityBehavior && $permission) {
+            $permissionKey = \array_search($permission, $this->permissions);
+            $permissions = $this->accessControlManager->getUserPermissionByArray(
```
alexander-schranz, line 694: Could be null if website has no security enabled.

src/Sulu/Component/Media/SmartContent/MediaDataProvider.php
```diff
@@ -112,7 +115,8 @@ public function resolveResourceItems(
         array $options = [],
         $limit = null,
         $page = 1,
-        $pageSize = null
+        $pageSize = null,
+        UserInterface $user = null
```
alexander-schranz, line 119: Is this still needed if we set the tokenStorage in the constructor?

src/Sulu/Bundle/MediaBundle/Media/Manager/MediaManager.php
danrot: I would have lived with that... Because it feels very weird to me if a constructor parameter is set and never used. But fine, going to add it again.

src/Sulu/Bundle/WebsiteBundle/Navigation/NavigationMapperInterface.php
danrot: No, I've added it afterwards, because it was forgotten in another PR. That's how the interface is actually looking.

src/Sulu/Component/Content/Mapper/ContentMapper.php
danrot: That's not true, this service is from our `SecurityBundle`, which is always registered because we need our user entities.

src/Sulu/Component/Content/Mapper/ContentMapper.php
danrot: Same as above

src/Sulu/Bundle/MediaBundle/Resources/config/services.xml
```diff
@@ -292,7 +299,7 @@
...
+            <argument type="string">%sulu_security.permissions%</argument>
```
alexander-schranz, line 302: Type string is not correct here also if it seems to have no effect I would remove it:

```suggestion
            <argument>%sulu_security.permissions%</argument>
```

PR #5444, opened 2020-08-05
Add permission inheritance for collections

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes -
| Related issues/PRs | -
| License | MIT

#### What's in this PR?

Add permission inheritance for collections.

#### Why?

It should be possible to set permissions also to the child collections.

#### BC Breaks/Deprecations

See upgrade.

#### To Do


src/Sulu/Bundle/SecurityBundle/Exception/AccessControlDescendantProviderNotFoundException.php
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+namespace Sulu\Bundle\SecurityBundle\Exception;
+
+class AccessControlDescendantProviderNotFoundException extends \Exception
+{
+    public function __construct($type, $code = 0, \Throwable $previous = null)
+    {
+        parent::__construct(
+            \sprintf('Access Control Children Provider not found for type "%s".', $type),
```
danrot, line 19: `AccessControlDescendantProvider`

src/Sulu/Component/Security/Authorization/AccessControl/AccessControlManagerInterface.php
```diff
@@ -26,7 +26,7 @@ interface AccessControlManagerInterface
      * @param string $identifier The identifier of the protected object
      * @param mixed[] $permissions
      */
-    public function setPermissions($type, $identifier, $permissions);
+    public function setPermissions($type, $identifier, $permissions/*, bool $inherit = false */);
```
danrot, line 29: I think in this interface I would not care too much about BC, or do you know about anybody who has extended this class in a way that this would make it break? If we don't do it this way it will be way easier to maintain.

src/Sulu/Component/Security/Authorization/AccessControl/DescendantProviderInterface.php
```diff
@@ -0,0 +1,29 @@
...
+interface DescendantProviderInterface
+{
+    /**
+     * Return the children for access control inherit update.
+     *
+     * @param string|int $id
+     *
+     * @return string[]|int[]
+     */
+    public function getDescendantIdsById($id);
```
danrot, line 23: Arent't these methods mostly named `find...`?

src/Sulu/Component/Security/Authorization/AccessControl/AccessControlManagerInterface.php
alexander-schranz: No does not know any project which did extend this class. So I think we could change it.

src/Sulu/Component/Security/Authorization/AccessControl/DescendantProviderInterface.php
alexander-schranz: Is okay for me. It was just once in the past where we accidentally conflicted with a doctrine function and so the interface combination of our and doctrine repository interface did not longer work, thats why I'm also a little bit worried when naming a function in a repository interface with find 🙈

danrot: Okay, I see... but seems to be highly unlikely in that case, right? Also saw a talk once, where people did not even use the doctrine repository as the base class :see_no_evil:

alexander-schranz: @danrot In our new projects we did that without extending but in the latest release of doctrine they added sadly a typehint so now all registred repositories need to extend from ObjectRepository else it will fail. Did not had a deeper look at it as @nnatter / @luca-rath did run into that issue and then changed to extend from the doctrine repository again.

danrot: But I guess they will only fail if you register them in Doctrine as `repositoryClass` of an entity, right? If we only use Symfony's dependency injection to get the repositories that should not really matter, does it? Just curious, it obviously is out of scope of this PR anyway :see_no_evil:

src/Sulu/Component/Security/Authorization/AccessControl/AccessControlManager.php
```diff
@@ -355,4 +380,13 @@ private function getAccessControlProvider($type)
             }
         }
     }
+
+    private function getChildrenProviders(string $type): ?DescendantProviderInterface
```
danrot, line 384: `getChildrenProvider` (only returns a single one, right?)

danrot:
Integrated in #5457

PR #5455, opened 2020-08-13
Fix mobx regression causing error in development build

danrot:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | mobxjs/mobx#2430, webpack/webpack#11312
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR adds a `toJS` to the JSX passed via a mobx store, because since mobx 4.15.5 the internals of react also seem to be tracked by mobx, and causes an error when react tries to change those later. See also mobxjs/mobx#2430 for more details.

#### Example Usage

1. Make a development build using `npm run watch`
2. Open an existing page in Sulu
3. Before this PR the screen will turn blank and an error is shown in the dev tools

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Toolbar/Toolbar.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Toolbar/Toolbar.js
@@ -1,6 +1,6 @@
 // @flow
 import {observer} from 'mobx-react';
-import {action, computed} from 'mobx';
+import {action, computed, toJS} from 'mobx';
 import React, {Fragment} from 'react';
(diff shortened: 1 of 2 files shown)


danrot:
There would be better solutions as described in https://github.com/mobxjs/mobx/issues/2430#issuecomment-673553487 performance-wise, but they are breaking changes... Added a few comments so that we might fix that in the next major release. If it otherwise works for you it should be fine now.

danrot:
@alexander-schranz Added a few explanatory comments with a suggestion how this could be improved by introducing a breaking change.

PR #5456, opened 2020-08-18
Fix localization ports if the current scheme is not https

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | none
| Related issues/PRs | none
| License | MIT
| Documentation PR | none

#### What's in this PR?

This PR fixes the generation of localizations when using HTTPS.

#### Why?

The localizations generate urls like this `http://sulu.io:443/de/kontakt`.

This happens because of the parameter `router.request_context.scheme` which is always `http` and used as `defaultScheme` in `WebspaceManager`.


alexander-schranz:
As a fix this should target another branch.

src/Sulu/Bundle/WebsiteBundle/Resolver/ParameterResolver.php
```diff
@@ -101,12 +101,14 @@ public function resolve(
         }
         $localizations = [];
 
+        $scheme = $this->requestStack->getCurrentRequest()->getScheme();
```
danrot, line 104: Why aren't we doing that in the `WebspaceManager`? Otherwise the same error might occur in other places as well. And there is a test missing :wink:

wachterjohannes:
@alexander-schranz the change for the "localizations" are only done on master

src/Sulu/Bundle/WebsiteBundle/Resolver/ParameterResolver.php
wachterjohannes: then the "WebspaceManager" depends on the request - do we want that?

alexander-schranz: The webspace already has the `currentRequest` in sulu 2.0 we already us it for the port see https://github.com/sulu/sulu/pull/5374. So I would fix this in the WebspaceManager and so target same as we targeted for the port fix the `release/2.0` branch.

alexander-schranz: Think the `defaultScheme` should first look at the currentRequest: https://github.com/sulu/sulu/blob/c17ca6b5e366f2ca6e541240a9eecb62590461e2/src/Sulu/Component/Webspace/Manager/WebspaceManager.php#L471 before falling back to the default scheme.

danrot:
@alexander-schranz Tests are failing

alexander-schranz:
@danrot fixed

PR #5457, opened 2020-08-19
Inherit page and collection permissions to children

### LINKED ISSUE #2391: Inheriting permissions of entities to sub entities
| Q | A |
| --- | --- |
| Bug? | no |
| New Feature? | yes |
#### Actual Behavior

Sulu does not offer any possibilities to inherit the permissions.
#### Expected Behavior

Sulu should allow to inherit the permissions from one entity to all of its sub entities. This would e.g. be useful for collections.

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #2391
| Related issues/PRs | dependant on #5458
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR will allow to inherit permissions to children. It is also implemented for page and collections.

#### To Do

- [ ] Only ask if permissions should be inherited in collection if collection has sub collections
- [x] Only ask if permisisons should be inherited in pages if page has sub pages


danrot:
Should be rebased after #5458 is merged.

src/Sulu/Bundle/MediaBundle/Entity/CollectionRepository.php
```diff
@@ -389,4 +390,22 @@ public function setAccessControlQueryEnhancer(AccessControlQueryEnhancer $access
     {
         $this->accessControlQueryEnhancer = $accessControlQueryEnhancer;
     }
+
+    public function findDescendantIdsById($id)
+    {
+        $queryBuilder = $this->createQueryBuilder('subCollection')
+            ->select('subCollection.id')
+            ->leftJoin($this->_entityName, 'collection', Join::WITH, 'collection.id = :id')
```
alexander-schranz, line 398: Should this not be a inner join as it need to exist and if I remember correctly inner joins are faster then left joins?

src/Sulu/Component/Content/Repository/ContentRepository.php
```diff
@@ -890,4 +922,16 @@ private function resolveHasChildren(Row $row)
...
+    public function supportsDescendantType(string $type): bool
+    {
+        try {
+            $class = new \ReflectionClass($type);
+        } catch (\ReflectionException $e) {
+            // in case the class does not exist there is no support
+            return false;
+        }
+
+        return $class->implementsInterface(SecurityBehavior::class);
```
alexander-schranz, line 935: We should avoid reflection and use here the [is_a](https://www.php.net/manual/de/function.is-a.php) function of php:

```php
return is_a($type, SecurityBehavior::class, true)
```

alexander-schranz (changes requested):
Did have a look at the commits would still be great to have #5458 merged and then have another review on this changes.

src/Sulu/Component/Content/Repository/ContentRepository.php
alexander-schranz: Think `is_a` is better in performance then doing a reflection and think we should avoid reflection if not really needed. Maybe we should then also update the other code snippets or does is_a handle something different?

src/Sulu/Bundle/MediaBundle/Entity/CollectionRepository.php
danrot: I think the confusing part was the misusage of a `join` in general... This is not a join in the general sense, but a cartesian product, whereby the elimination happens only based on the `where` clause happening later. Rewrote it a bit, to make that part more clear.

src/Sulu/Component/Content/Repository/ContentRepository.php
danrot: From what I know the fear of reflections being slow comes from a pre-PHP7 era, and in some PHP version they got much faster. Therefore I would prefer to do a benchmark before changing all occurences based on a guess. We can change this occurence including all others in a separate PR, if the difference is really that big.

alexander-schranz (approved):

PR #5458, opened 2020-08-19
Introduce SaveWithFormDialogToolbarAction to show forms on save

danrot:
| Q | A 
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | prerequesite of #5457
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR introduces a new `SaveWithFormDialogToolbarAction`, which shows a form when a user hits the save button. The form is displayed in a dialog, and whatever the user chooses is appended to the save request.

#### Why?

Because we need this functionality for #5457.

#### Example Usage

See the [`PageAdmin` in the other PR](https://github.com/sulu/sulu/pull/5457/files#diff-35e269d96646ae05db89f97f6115d238). Use e.g. `tag_details` instead of `permission_inheritance`, as the latter does not exist in this PR.

#### BC Breaks/Deprecations

The form submit now does not take a single `action` string anymore, but an `options` object instead. The old way still works, but is deprecated and logs a warning.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/views/Form/toolbarActions/AbstractFormToolbarAction.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/views/Form/toolbarActions/AbstractFormToolbarAction.js
@@ -1,5 +1,6 @@
 // @flow
 import type {Node} from 'react';
+import ResourceStore from '../../../stores/ResourceStore';
 import {ResourceFormStore} from '../../../containers/Form';
 import type {ToolbarItemConfig} from '../../../containers/Toolbar/types';
 import Router from '../../../services/Router';
@@ -11,19 +12,22 @@ export default class AbstractFormToolbarAction {
     router: Router;
     locales: ?Array<string>;
     options: {[key: string]: mixed};
+    parentResourceStore: ResourceStore;
 
     constructor(
         resourceFormStore: ResourceFormStore,
         form: Form,
         router: Router,
         locales: ?Array<string>,
-        options: {[key: string]: mixed}
+        options: {[key: string]: mixed},
+        parentResourceStore: ResourceStore
     ) {
         this.resourceFormStore = resourceFormStore;
         this.form = form;
         this.router = router;
         this.locales = locales;
         this.options = options;
+        this.parentResourceStore = parentResourceStore;
     }
 
     setLocales(locales: Array<string>) {
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/views/Form/toolbarActions/AbstractFormToolbarAction.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/views/Form/toolbarActions/AbstractFormToolbarAction.js
@@ -37,4 +37,8 @@ export default class AbstractFormToolbarAction {
     getToolbarItemConfig(): ?ToolbarItemConfig<*> {
         throw new Error('The getToolbarItemConfig method must be implemented by the sub class!');
     }
+
+    destroy() {
+
+    }
 }
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/views/Form/toolbarActions/SaveWithFormDialogToolbarAction.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/views/Form/toolbarActions/SaveWithFormDialogToolbarAction.js
@@ -3,29 +3,35 @@ import React from 'react';
 import type {ElementRef} from 'react';
 import {action, observable} from 'mobx';
 import Dialog from '../../../components/Dialog';
-import {default as FormContainer, memoryFormStoreFactory} from '../../../containers/Form';
+import {default as FormContainer, memoryFormStoreFactory, ResourceFormStore} from '../../../containers/Form';
 import type {FormStoreInterface} from '../../../containers/Form';
+import Router from '../../../services/Router';
 import {translate} from '../../../utils/Translator';
+import Form from '../Form';
 import AbstractFormToolbarAction from './AbstractFormToolbarAction';
 
 export default class SaveWithFormDialogToolbarAction extends AbstractFormToolbarAction {
     @observable showDialog: boolean = false;
     dialogForm: ?FormContainer;
-    dialogFormStore: ?FormStoreInterface;
+    dialogFormStore: FormStoreInterface;
 
-    getDialogFormStore = () => {
-        if (!this.dialogFormStore) {
-            const {formKey} = this.options;
+    constructor(
+        resourceFormStore: ResourceFormStore,
+        form: Form,
+        router: Router,
+        locales: ?Array<string>,
+        options: {[key: string]: mixed}
+    ) {
+        super(resourceFormStore, form, router, locales, options);
 
-            if (typeof formKey !== 'string') {
-                throw new Error('The "formKey" option of the SaveWithFormDialogToolbarAction must be a string!');
-            }
+        const {formKey} = options;
 
-            this.dialogFormStore = memoryFormStoreFactory.createFromFormKey(formKey);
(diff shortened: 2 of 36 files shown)


danrot:
It just came to my mind that we said the overlay should only appear if the item has children... Now I realized that this is not easily possible to achieve :confused:

The problem is that the `SaveWithFormDialogToolbarAction` only has access to the `ResourceFormStore` of the current form, and the current form is not a page form, but a permissions form. And the permissions form does not really know anything about the pages, so I can't ask if the page has children...

src/Sulu/Bundle/AdminBundle/Resources/js/views/Form/toolbarActions/AbstractFormToolbarAction.js
```diff
@@ -37,4 +37,8 @@ export default class AbstractFormToolbarAction {
     getToolbarItemConfig(): ?ToolbarItemConfig<*> {
         throw new Error('The getToolbarItemConfig method must be implemented by the sub class!');
     }
+
+    destroy() {
```
alexander-schranz, line 41: @danrot we had in the past projects also stores inside the ListToolbarActions, think it would be good and consistent to also have there the `destroy` function. What do you think?

alexander-schranz (changes requested):
I think the list and form toolbar actions should be consistent and have both the `destroy` function.

src/Sulu/Bundle/AdminBundle/Serializer/Subscriber/SaveWithFormDialogToolbarActionSubscriber.php
```diff
@@ -0,0 +1,54 @@
...
+
+namespace Sulu\Bundle\AdminBundle\Serializer\Subscriber;
+
+use JMS\Serializer\EventDispatcher\Events;
+use JMS\Serializer\EventDispatcher\EventSubscriberInterface;
+use JMS\Serializer\EventDispatcher\ObjectEvent;
+use Sulu\Bundle\AdminBundle\Admin\View\SaveWithFormDialogToolbarAction;
+use Symfony\Contracts\Translation\TranslatorInterface;
+
+class SaveWithFormDialogToolbarActionSubscriber implements EventSubscriberInterface
```
alexander-schranz, line 20: I think it would be good if we have a functional test case for this e.g.:

```php
$this->assertSame($expectedJson, $this->jmsSerializer->serialize($toolbarAction);
```

src/Sulu/Bundle/AdminBundle/Resources/js/views/Form/toolbarActions/SaveWithFormDialogToolbarAction.js
```diff
@@ -0,0 +1,99 @@
...
+                onCancel={this.handleCancel}
+                onConfirm={this.handleConfirm}
+                open={this.showDialog}
+                title={title}
+            >
+                <FormContainer
+                    onSubmit={this.handleSubmit}
+                    ref={this.setDialogFormRef}
+                    store={this.dialogFormStore}
+                />
```
alexander-schranz, line 78: In the projects we did mostly lazy initialize the form store and loading of the data maybe that would also make sense here for performance/memory reasons:

```js

                {
                    this.showDialog ?
                <FormContainer
                    onSubmit={this.handleSubmit}
                    ref={this.setDialogFormRef}
                    store={this.getDialogFormStore()}
                /> : null
}

    getDialogFormStore() {
        if (!this.dialogFormStore) {
            this.dialogFormStore = ...
        }

        return this.dialogFormStore;
    }

    destroy() {
        if (this.dialogFormStore) {
            this.dialogFormStore.destroy();
        }
    }
```

alexander-schranz:
@danrot to your [comment](https://github.com/sulu/sulu/pull/5458#issuecomment-677246759) I think the only way we could solve this problem is that we introduce something like a `parentResourceStore` which allows to access the resource data of the [tabs](https://github.com/sulu/sulu/blob/master/src/Sulu/Bundle/PageBundle/Admin/PageAdmin.php#L218), else I see know other way to solve this if we want this behaviour.

danrot:
The problem is that this solution does not scale :see_no_evil: Maybe some day somebody would need the `parentParentResourceStore` :grimacing:

But another thing came to my mind: The request for the permissions returns a JSON object with a `permissions` key. Maybe we could return another key `hasChildren`, so that we always get that returned as expected. However, that means we would have to introduce a new method on the `AccessControlManager`. A problem that would probably arise is that getting this information might differ between different entities. Maybe the `parentResourceStore` is the easier solution, and we just live with the fact that it only works if the form has only one parent view.

alexander-schranz:
@danrot not sure if possible but maybe then a `parent` parameter would work so you could access the `resourceStore` of the `parent` with: `parentView.resourceStore` and the `parent` itself could have a `parent.parent` but not sure if we ever will run into that usecase having tabs inside tabs would it be or I'm wrong? So I think I would go with `parentResourceStore` at current state. 

Could the `hasChildren` flag not use the introduced `findDescendantIdsById` function to return based on that values true or false, so it not need to have additional function.

src/Sulu/Bundle/AdminBundle/Resources/js/views/Form/toolbarActions/SaveWithFormDialogToolbarAction.js
danrot: Tried to implement it this way, and while it might be better for performance/memory reasons, it comes with two downsides, which make me think that it's not worth the tradeoff:

- When the dialog is opened it will take some more time to load the metadata, otherwise it is already loaded and can be immediately displayed (better perceived performance)
- When the dialog is closed the form disappears before the dialog, which makes it feel a bit uneasy (could be fixed, but makes the code much more complicated by holding two different `showDialog` states

alexander-schranz: Okay for me.

src/Sulu/Bundle/AdminBundle/Tests/Functional/Admin/View/SaveWithFormDialogToolbarActionTest.php
```diff
@@ -0,0 +1,28 @@
...
+namespace Sulu\Bundle\AdminBundle\Tests\Functional\Admin\View;
+
+use Sulu\Bundle\AdminBundle\Admin\View\SaveWithFormDialogToolbarAction;
+use Sulu\Bundle\TestBundle\Testing\SuluTestCase;
+
+class SaveWithFormDialogToolbarActionTest extends SuluTestCase
+{
+    public function testSerializer()
+    {
+        $saveWithFormDialogToolbarAction = new SaveWithFormDialogToolbarAction('title', 'form');
```
alexander-schranz, line 21: Would be great to use a real translation key here so we are sure in this test that the translated value is correctly set.

src/Sulu/Bundle/AdminBundle/Tests/Functional/Admin/View/SaveWithFormDialogToolbarActionTest.php
```diff
@@ -0,0 +1,28 @@
...
+
+        $this->assertEquals(
+            '{"type":"sulu_admin.save_with_form_dialog","options":{"formKey":"form","title":"title"}}',
```
alexander-schranz, line 24: As above this should use a real translation key and we should test then agains the translated value. Else we can not be sure if subscriber listener did correctly set the translated value.

alexander-schranz (changes requested):

alexander-schranz (approved):

PR #5460, opened 2020-08-19
Add configuration for additional indexes for website search

luca-rath:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This pull request adds configuration for additional indexes for the website search

#### Why?

Without this, you always have to override the `WebsiteSearchController` to allow searching for documents in different indexes.

#### Example Usage

~~~yaml
# config/packages/sulu_search.yaml

sulu_search:
  website:
    additional_indexes:
      - examples_published
~~~

### DIFF
--- a/src/Sulu/Bundle/SearchBundle/Controller/WebsiteSearchController.php
+++ b/src/Sulu/Bundle/SearchBundle/Controller/WebsiteSearchController.php
@@ -47,16 +47,23 @@ class WebsiteSearchController
      */
     private $twig;
 
+    /**
+     * @var string[]
+     */
+    private $additionalIndexes;
+
     public function __construct(
         SearchManagerInterface $searchManager,
         RequestAnalyzerInterface $requestAnalyzer,
         ParameterResolverInterface $parameterResolver,
-        Environment $twig
+        Environment $twig,
+        array $additionalIndexes
     ) {
         $this->searchManager = $searchManager;
         $this->requestAnalyzer = $requestAnalyzer;
         $this->parameterResolver = $parameterResolver;
         $this->twig = $twig;
+        $this->additionalIndexes = $additionalIndexes;
     }
 
     /**
@@ -90,7 +97,10 @@ class WebsiteSearchController
         $hits = $this->searchManager
             ->createSearch($queryString)
             ->locale($locale)
-            ->index('page_' . $webspace->getKey() . '_published')
+            ->indexes(array_merge(
+                ['page_' . $webspace->getKey() . '_published'],
+                $this->additionalIndexes
+            ))
             ->execute();
 
         $template = $webspace->getTemplate('search', $request->getRequestFormat());
--- a/src/Sulu/Bundle/SearchBundle/DependencyInjection/Configuration.php
+++ b/src/Sulu/Bundle/SearchBundle/DependencyInjection/Configuration.php
@@ -45,6 +45,15 @@ class Configuration implements ConfigurationInterface
                         ->end()
                     ->end()
                 ->end()
+                ->arrayNode('website')
+                    ->children()
+                        ->arrayNode('additional_indexes')
+                            ->prototype('scalar')
+                                ->defaultValue([])
+                            ->end()
+                        ->end()
+                    ->end()
+                ->end()
             ->end();
 
         return $treeBuilder;
--- a/src/Sulu/Bundle/SearchBundle/Tests/Unit/Controller/WebsiteSearchControllerTest.php
+++ b/src/Sulu/Bundle/SearchBundle/Tests/Unit/Controller/WebsiteSearchControllerTest.php
@@ -70,7 +70,8 @@ class WebsiteSearchControllerTest extends TestCase
             $this->searchManager->reveal(),
             $this->requestAnalyzer->reveal(),
             $this->parameterResolver->reveal(),
-            $this->twig->reveal()
+            $this->twig->reveal(),
+            []
         );
     }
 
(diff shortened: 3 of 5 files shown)


wachterjohannes:
@luca-rath will you add a docs PR?

wachterjohannes (approved):

src/Sulu/Bundle/SearchBundle/Controller/WebsiteSearchController.php
```diff
@@ -47,16 +47,23 @@ class WebsiteSearchController
...
+     */
+    private $additionalIndexes;
+
     public function __construct(
         SearchManagerInterface $searchManager,
         RequestAnalyzerInterface $requestAnalyzer,
         ParameterResolverInterface $parameterResolver,
-        Environment $twig
+        Environment $twig,
+        array $additionalIndexes
```
alexander-schranz, line 60: For BC this needs to be:

```suggestion
        array $additionalIndexes = []
```

src/Sulu/Bundle/SearchBundle/DependencyInjection/Configuration.php
```diff
@@ -45,6 +45,16 @@ public function getConfigTreeBuilder()
                         ->end()
                     ->end()
                 ->end()
+                ->arrayNode('website')
+                    ->addDefaultsIfNotSet()
+                    ->children()
+                        ->arrayNode('additional_indexes')
```
alexander-schranz, line 51: Not sure if I would go with `additional_indexes` by array. It will not allow to override things or prepend it in the bundles so I would go more with a configuration like this:

```yaml
website:
     indexes:
          pages: 'page_#webspace#_published' 
          products: 'my_products_published' 
          news: 'my_news_published' 
```

This make it to possible override the index or disable a prepended index by using set it to `null`.

alexander-schranz (changes requested):
@wachterjohannes What do you think about this config?

src/Sulu/Bundle/SearchBundle/Controller/WebsiteSearchController.php
```diff
@@ -90,7 +97,15 @@ public function queryAction(Request $request)
         $hits = $this->searchManager
             ->createSearch($queryString)
             ->locale($locale)
-            ->index('page_' . $webspace->getKey() . '_published')
+            ->indexes(
+                \str_replace(
```
alexander-schranz, line 101: I personally would move this logic into the `SuluSearchExtension.php`

luca-rath: @alexander-schranz just the `str_replace`? or also the `array_filter` and `array_values`? because then the keys are lost

alexander-schranz: Yes would move all into the Extension. Think the controller should not know about the key as that is configuration logic.

luca-rath: @alexander-schranz I just realized, that the current webspace isn't available when setting the parameter

alexander-schranz: Thats correct would still move the other logic into the Extension.

alexander-schranz (approved):

src/Sulu/Bundle/SearchBundle/Tests/Unit/Controller/WebsiteSearchControllerTest.php
```diff
@@ -93,7 +94,54 @@ public function testQueryAction()
...
+    }
+
+    public function testQueryActionWithDifferentIndexes()
+    {
+        $this->websiteSearchController = new WebsiteSearchController(
+            $this->searchManager->reveal(),
+            $this->requestAnalyzer->reveal(),
+            $this->parameterResolver->reveal(),
+            $this->twig->reveal(),
+            ['examples_published']
```
danrot, line 122: Would make this a test with multiple indices and a `#webspace#' placeholder, you can't really tell until now if that is working by the tests, right?

luca-rath: fixed

src/Sulu/Bundle/SearchBundle/DependencyInjection/Configuration.php
```diff
@@ -45,6 +45,15 @@ public function getConfigTreeBuilder()
                         ->end()
                     ->end()
                 ->end()
+                ->arrayNode('website')
+                    ->addDefaultsIfNotSet()
+                    ->children()
+                        ->arrayNode('indexes')
+                            ->useAttributeAsKey('key')
```
danrot, line 52: Does it make sense to set this? If I read the [Symfony documentation](https://symfony.com/doc/3.2/components/config/definition.html#array-node-options) correctly, the `array_values` call in our code would be unnecessary if we just omit that, because when the config is evaluated it will already be returned as simple array. Not totally sure if overriding as you've suggested in this PR would still work though.

danrot:
I would still prefer to keep this an array, because the developer has to come up with some keys, that are not used at all...

An alternative approach could be to keep it as an array, and make the other bundles configurable in that way:

```yaml
sulu_article:
    search_integration: false # not sure about the naming, but you get the idea
```

Might be more logical for the developer working with that configuration, but I am also fine with the current solution if you prefer that.

luca-rath:
> I would still prefer to keep this an array, because the developer has to come up with some keys, that are not used at all...
> 
> An alternative approach could be to keep it as an array, and make the other bundles configurable in that way:
> 
> ```yaml
> sulu_article:
>     search_integration: false # not sure about the naming, but you get the idea
> ```
> 
> Might be more logical for the developer working with that configuration, but I am also fine with the current solution if you prefer that.

I would prefer to keep it like it is, because your suggested configuration could also be implemented in the ArticleBundle using the current solution. Otherwise we depend on each bundle offering such a configuration (also the PageBundle). /cc @sulu/core-team

src/Sulu/Bundle/SearchBundle/DependencyInjection/Configuration.php
luca-rath: just tried, overriding doesn't work then

danrot:
That's indeed a downside, but personally I see that as an edge case anyway. I don't see why people would deactivate pages or articles on the website search. But as said, the current solution is also OK for me, just have a slight preference for the other one.

luca-rath:
And another thing, with the current solution it's still possible to configure the `indexes` using a flat array, if you want to. Then `0`, `1`, ... will be used as keys.

PR #5461, opened 2020-08-20
Use permissions of parents when creating new collections or pages

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | needs #5457 
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR uses the parent permissions when a new page or media collection is created. 

#### Why?

This way we can make sure that a page being created under a restricted page is no publicly visible by accident.
#### To Do

- [x] Implement for pages


alexander-schranz (changes requested):
I personally would handle entities same as documents over a listener this also allows to easier implement security parent behaviour for custom entities and avoids us the bc break in the collection manager.

I also think it would be great to have functional test checking when parent collection has no permission in the database it should also create no permissions for the new collection.

alexander-schranz (approved):

PR #5464, opened 2020-08-24
Warn when a page or collection with permissions is moved

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | depends on #5461
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR will ask the user if it is OK to move the item without touching the different securities of an element and its new parent. The user can only confirm or cancel that operation.

#### Why?

Because that will be enough, since the User can still move and change the permissions afterwards. Also, we think that this won't happen very often.

#### To Do

- [x] Ask for confirmation
- [x] Implement the merging of the permissions
- [x] Add missing translations `sulu_security.merge_permission_title` and `sulu_security.merge_permission_warning`


alexander-schranz (approved):
Code and functionality looks good, just translations missing.

PR #5470, opened 2020-08-27
Fix function to get user from token-storage if no token is set

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | none
| Related issues/PRs | none
| License | MIT
| Documentation PR | none

#### What's in this PR?

This PR fixes the navigation twig extension if the security is enabled but no token isset.

#### Why?

This can happen if the current route is not inside a firewall or no token was initialized.

### DIFF
--- a/src/Sulu/Bundle/WebsiteBundle/Twig/Navigation/NavigationTwigExtension.php
+++ b/src/Sulu/Bundle/WebsiteBundle/Twig/Navigation/NavigationTwigExtension.php
@@ -210,7 +210,12 @@ class NavigationTwigExtension extends AbstractExtension implements NavigationTwi
             return null;
         }
 
-        $user = $this->tokenStorage->getToken()->getUser();
+        $token = $this->tokenStorage->getToken();
+        if (!$token) {
+            return null;
+        }
+
+        $user = $token->getUser();
 
         if ($user instanceof UserInterface) {
             return $user;
--- a/src/Sulu/Bundle/MediaBundle/Media/Manager/MediaManager.php
+++ b/src/Sulu/Bundle/MediaBundle/Media/Manager/MediaManager.php
@@ -905,12 +905,12 @@ class MediaManager implements MediaManagerInterface
             return;
         }
 
-        if (!$this->tokenStorage->getToken()) {
+        $token = $this->tokenStorage->getToken();
+        if (!$token) {
             return;
         }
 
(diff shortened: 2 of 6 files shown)


src/Sulu/Bundle/WebsiteBundle/Twig/Navigation/NavigationTwigExtension.php
```diff
@@ -210,7 +210,12 @@ private function getUser(): ?UserInterface
             return null;
         }
 
-        $user = $this->tokenStorage->getToken()->getUser();
+        $token = $this->tokenStorage->getToken();
```
alexander-schranz, line 213: If I remember correctly this is not the only code part where the token is read and the user accessed. Can you check for the other `$this->tokenStorage->getToken()` calls and fix it also there.

alexander-schranz (changes requested):

alexander-schranz:
@wachterjohannes How to reproduce this? Should not be the token be anonymous? Or is this the case if I'm on a url with doesn't match any firewall?

src/Sulu/Bundle/WebsiteBundle/Twig/Navigation/NavigationTwigExtension.php
wachterjohannes: there are severall places where we do that :) but in the most its correct (some are now fixed) - only in some admin services i dont think its neccessary to change it - there is the user required and it seems that this "edge case" cannot happen there.

wachterjohannes:
@alexander-schranz correct a URL which is not in the firewall - or the firewall is lazy and not initialized

alexander-schranz (approved):
`lazy` should not be the problem as the `tokenStorage` is the one which will initialize the firewall [here](https://github.com/symfony/symfony/blob/2c4e215ee9b29e45ed0b2f7fa4f705c628be98e5/src/Symfony/Component/Security/Core/Authentication/Token/Storage/TokenStorage.php#L37). 

Code looks good from my side but want that @danrot have also a look at it.

danrot:
Code looks good to me. @alexander-schranz Have you already tested it?

alexander-schranz:
@danrot good, I did test it and it did work with this PR if a url was called without a matching firewall.

PR #5472, opened 2020-08-28
Allow to copy page between webspaces

### LINKED ISSUE #4502: Cannot copy/move content between webspaces
| Q | A
| --- | ---
| Bug? | no
| New Feature? | no
| Sulu Version | 1.6.25
| Introduced with issue |  #2104

#### Actual Behavior

Can not copy/move content between webspaces.

#### Expected Behavior

Would like to copy/move content between webspaces.

#### Steps to Reproduce

- Set up two webspaces; 
- create content in either of them
- use the column navigation dropdown context menu and select copy/move

Only the tree of the current webspace is displayed.

#### Possible Solutions

The responsible line of code is here:

https://github.com/sulu/sulu/blob/ae060d1b1403f789653cff5c2478da5154a6b04a/src/Sulu/Bundle/ContentBundle/Resources/public/js/components/content/column/main.js#L610

When "webspace-nodes=all" is used (instead of 'single'), then all webspaces are listed, allowing to copy/move pages between them freely.

Is it deliberately pre-configured to only allow copying/moving content within the same webspace?

When not, can we change the default behaviour to list all configured webspaces in the tree?

The page selection within the teaser selection content type already lists the complete tree - including all webspaces.

#### Use case

I have a 'Staging' and a 'Production' webspace in one CMS that allow the website editor to test out content in the Staging webspace (on its dedicated hostname) and when done, copy/move the changes/new pages to the Production webspace, draft/preview it and when all is fine, publish it.

A kind of 'Content Staging' solution that is not available in sulu.

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #4502 
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR allows to copy pages between different webspaces.

#### Why?

Because this feature is often requested.

#### To Do

- [x] Go to just copied page also if the destination was in the other webspace (introduce `onCopyFinished` callback that gets called in `List` container? Or return some HTTP 4xx if `expandedIds` does not exist in given `webspace`?)
- [x] Allow to fix pages if the template is not available in the other webspace
- [x] Disallow setting non-existing or for this webspace excluded template as default for any type (#5484)
- [x] Also check ghost page with the loaded language being a non-existing template
- [x] Fix `update-context` call of preview when switching templates (seems like it was already broken before that PR)
- [x] Add missing translations (`sulu_admin.missing_type_dialog_title`, `sulu_admin.missing_type_dialog_description`, `sulu_admin.choose_template`)
- [ ] Show exclamation mark icon if current template of page is not existing in its webspace


danrot:
Somehow I ended up using `MissingTypeDialog` and `hasInvalidType` as identifiers... Which one do you think suit better? :thinking: /cc @sulu/core-team

danrot:
Just to be clear: These identifiers are used in two different situations... The question is if the term "Missing" or "Invalid" is a better match for a page's template, that does not exist, i.e. was either deleted or excluded in the page's webspace.

alexander-schranz:
`MIssing` sounds better for me but have no strong opinion on that.

danrot:
Thought about it since yesterday, and I think I would go with "Invalid". "Missing" sounds like the template definition is not existing in the system, which is not true necessarily. The template can exist, but it is just "invalid" in the current webspace, because it has been excluded.

alexander-schranz:
> Thought about it since yesterday, and I think I would go with "Invalid". "Missing" sounds like the template definition is not existing in the system, which is not true necessarily. The template can exist, but it is just "invalid" in the current webspace, because it has been excluded.

Sounds valid for me 👍

src/Sulu/Component/Content/Compat/Structure/StructureBridge.php
```diff
@@ -322,7 +322,8 @@ public function toArray($complete = true)
             'nodeType' => $this->getNodeType(),
             'nodeState' => $this->getNodeState(),
             'internal' => false,
-            'contentLocales' => $this->inspector->getLocales($document),
+            'availableLocales' => $this->inspector->getLocales($document),
```
nnatter, line 325: whats the reason for this change?

config/webspaces/sulu-blog.xml
```diff
@@ -15,7 +15,7 @@
     </localizations>
```
nnatter, line 15: whats the reason for adjusting this file?

nnatter (commented):
code looks good otherwise

nnatter:
There is an unchecked `Show exclamation mark icon if current template of page is not existing in its webspace ` todo in the pr description. Should this already work?

nnatter:
I encounter the following error when switching the template of a page:
```
Too few arguments to function Sulu\Bundle\PreviewBundle\Preview\Preview::updateContext(), 4 passed in /Users/niklas/Development/projects/sulu-skeleton/vendor/sulu/sulu/src/Sulu/Bundle/PreviewBundle/Controller/PreviewController.php on line 120 and exactly 5 expected
```

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/MissingTypeDialog.js
```diff
@@ -0,0 +1,72 @@
...
+                confirmDisabled={!this.selectedType}
+                confirmText={translate('sulu_admin.ok')}
+                onCancel={this.handleCancel}
+                onConfirm={this.handleConfirm}
+                open={open}
+                title={translate('sulu_admin.missing_type_dialog_title')}
+            >
+                <p>{translate('sulu_admin.missing_type_dialog_description')}</p>
+                <Form>
+                    <Form.Field colSpan={6}>
```
nnatter, line 57: why not make this full width?

src/Sulu/Component/Content/Compat/Structure/StructureBridge.php
danrot: Because the value of these properties were different when you were requesting multiple pages vs a single page. Adjusted it to be consistent.

config/webspaces/sulu-blog.xml
danrot: To have all the features somehow implemented in this repository. We've decided some time ago to include as much as possible here to make testing easier. In this repository it does not matter if the configuration gets a bit more complicated, because these files are not used in a real project.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/MissingTypeDialog.js
danrot: In order to match the design of the dialog for the ghost locale. I also think it looks better, because the name of the template usually does not get that long.

danrot (commented):
> There is an unchecked `Show exclamation mark icon if current template of page is not existing in its webspace ` todo in the pr description. Should this already work?

I have got a new Icon set also including the permissions icon, which we also want to add. I thought I am going to add both of those icons in a separate PR.

> I encounter the following error when switching the template of a page. Not sure if this is caused by this PR.
> 
> ```
> Too few arguments to function Sulu\Bundle\PreviewBundle\Preview\Preview::updateContext(), 4 passed in /Users/niklas/Development/projects/sulu-skeleton/vendor/sulu/sulu/src/Sulu/Bundle/PreviewBundle/Controller/PreviewController.php on line 120 and exactly 5 expected
> ```

This issue has already been fixed in #5485, should just be a rebase.

nnatter (approved):
> I have got a new Icon set also including the permissions icon, which we also want to add. I thought I am going to add both of those icons in a separate PR.

Sounds good 👍

PR #5482, opened 2020-09-03
Extract validation of FieldMetadata to FieldMetadataValidator

luca-rath:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Add FieldMetadataValidator service to allow validation of property names e.g.

#### Why?

To make the current validation more flexible


src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/StructureFormMetadataLoader.php
```diff
@@ -121,6 +129,11 @@ public function warmUp($cacheDir)
         foreach ($this->locales as $locale) {
             foreach ($structuresMetadataByTypes as $structureType => $structuresMetadata) {
                 $structure = $this->mapStructureMetadata($structuresMetadata, $locale);
+
+                foreach ($structure->getForms() as $formMetadata) {
+                    $this->validateFormMetadata($formMetadata);
```
luca-rath, line 134: I have no idea how to test this, because this is in the `warmUp()` method. Calling this method in a test doesn't throw an exception

src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/XmlFormMetadataLoader.php
```diff
@@ -87,7 +95,13 @@ public function warmUp($cacheDir)
         }
 
         foreach ($formsMetadataCollection as $key => $formMetadataCollection) {
+            /**
+             * @var string $locale
+             * @var FormMetadata $formMetadata
+             */
             foreach ($formMetadataCollection->getItems() as $locale => $formMetadata) {
+                $this->validateFormMetadata($formMetadata);
```
luca-rath, line 103: I have no idea how to test this, because this is in the `warmUp()` method. Calling this method in a test doesn't throw an exception

src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/Validation/BlockFieldMetadataValidator.php
```diff
@@ -0,0 +1,46 @@
...
+        }
+
+        foreach ($fieldMetadata->getTypes() as $type) {
+            foreach ($type->getItems() as $item) {
+                if (!$item instanceof FieldMetadata) {
+                    continue;
+                }
+
+                if (\in_array($item->getName(), static::RESERVED_PROPERTY_NAMES, true)) {
+                    throw new InvalidFieldMetadataException(
```
danrot, line 35: I'd prefer to pass the information directly to the constructor, instead of generating the message everywhere the exception is created. Less code duplication and it would allow to retrieve the information from the exception in a `catch` block without parsing the exception message. So IMO something like this should be possible:

```php
try {
    throw new InvalidFieldMetadataException($fieldMetadata->getName(), $item->getName());
} catch (InvalidFieldMetadataException $e) {
    $e->getFieldName();
    $e->getPropertyName();
}
```

TBH I am also wondering why you have introduced another `Exception` instead of reusing the `ReservedPropertyNameException`. That one would also explain better what that exception is about.

src/Sulu/Component/Content/Tests/Unit/Metadata/Loader/StructureXmlLoaderTest.php
```diff
@@ -294,22 +293,6 @@ public function testLoadRequiredTagOtherType()
...
-    }
-
-    public function testLoadFormWithBlockSettingsProperty()
-    {
-        $this->expectException(ReservedPropertyNameException::class);
-        $this->expectExceptionMessageRegExp('"settings"');
-
-        $this->load('template_with_block_settings_property.xml');
-    }
-
```
danrot, line 312: Why have you removed these tests?

danrot (changes requested):

src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/Validation/BlockFieldMetadataValidator.php
```diff
@@ -0,0 +1,46 @@
+<?php
+
+declare(strict_types=1);
```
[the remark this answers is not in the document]
luca-rath: wasn't on purpose 🙈

src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/Validation/BlockFieldMetadataValidator.php
luca-rath: One reason why we decided to introduce this kind of validation is to be able to validate more than just the property name, right? So I guess there should be a generic exception for all validation errors. But I would be okay with the `ReservedPropertyNameException` if it would extend from the `InvalidFieldMetadataException`. But currently we don't get the form key in the validator and the `ReservedPropertyNameException` needs that one too.

danrot: Yeah, but we could also introduce other more specific exceptions in the future. For me it would also be ok for the `ReservedPropertyNameException` to inherit from the `InvalidFieldMetadataException`.

But passing the form key is essential to this exception, because otherwise it is very hard for the developer to figure out which form is affected. If we introduce a new inheritance hierarchy I would even add that key to the `InvalidFieldMetadataException`.

src/Sulu/Bundle/AdminBundle/Tests/Unit/FormMetadata/FormXmlLoaderTest.php
```diff
@@ -506,28 +505,6 @@ public function testLoadFormWithSizedSections()
...
-    {
-        $this->expectException(ReservedPropertyNameException::class);
-        $this->expectExceptionMessageRegExp('"settings"');
-        $this->expectExceptionMessageRegExp('"form_with_block_settings_property"');
-
-        $this->loader->load(
-            $this->getFormDirectory() . '../invalid-forms/form_with_block_settings_property.xml'
-        );
-    }
-
```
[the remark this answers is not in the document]
luca-rath: No they don't, because this is a unit test and the `StructureFormMetadataLoader` (which is calling the validator) never gets called.

src/Sulu/Component/Content/Tests/Unit/Metadata/Loader/StructureXmlLoaderTest.php
luca-rath: They don't work anymore, because this is a unit test and the `StructureFormMetadataLoader` (which is calling the validator) never gets called.

src/Sulu/Bundle/AdminBundle/Tests/Unit/FormMetadata/StructureFormMetadataLoaderTest.php
```diff
@@ -0,0 +1,163 @@
...
+            $formMetadata->addItem($item);
+        }
+
+        return $formMetadata;
+    }
+
+    /**
+     * @param ItemMetadata[] $items
+     */
+    private function createSectionMetadata(string $name = 'some_section', array $items = []): SectionMetadata
```
danrot, line 114: Same as above, would make the first paramter non-optional.

src/Sulu/Bundle/AdminBundle/Tests/Unit/FormMetadata/StructureFormMetadataLoaderTest.php
```diff
@@ -0,0 +1,163 @@
...
+            false
+        );
+    }
+
+    /**
+     * @param FormMetadata[] $types
+     */
+    private function createFieldMetadata(
+        string $name = 'some_property',
+        string $type = 'text_line',
```
danrot, line 81: I think I would not make the first two parameters default values. It makes it much harder to understand the tests below, and from what I've seen you have never called these function without at least one parameter.

src/Sulu/Bundle/AdminBundle/Tests/Unit/FormMetadata/StructureFormMetadataLoaderTest.php
```diff
@@ -0,0 +1,163 @@
...
+            $fieldMetadata->addType($type);
+        }
+
+        return $fieldMetadata;
+    }
+
+    /**
+     * @param ItemMetadata[] $items
+     */
+    private function createFormMetadata(string $name = 'some_form', string $key = 'some_form_key', array $items = []): FormMetadata
```
danrot, line 97: Same as above, wouldn't add a default parameter to the first two.

src/Sulu/Bundle/AdminBundle/Tests/Unit/FormMetadata/Validation/BlockFieldMetadataValidatorTest.php
```diff
@@ -0,0 +1,155 @@
...
+    {
+        $this->blockFieldMetadataValidator = new BlockFieldMetadataValidator();
+    }
+
+    /**
+     * @param FormMetadata[] $types
+     */
+    private function createFieldMetadata(
+        string $name = 'some_property',
+        string $type = 'block',
```
danrot, line 39: Same as above, would avoid default values.

src/Sulu/Bundle/AdminBundle/Tests/Unit/FormMetadata/Validation/BlockFieldMetadataValidatorTest.php
```diff
@@ -0,0 +1,155 @@
...
+            $fieldMetadata->addType($type);
+        }
+
+        return $fieldMetadata;
+    }
+
+    /**
+     * @param ItemMetadata[] $items
+     */
+    private function createFormMetadata(string $name = 'some_form', array $items = []): FormMetadata
```
danrot, line 55: Same as above, would avoid the first default value.

src/Sulu/Bundle/AdminBundle/Tests/Unit/FormMetadata/Validation/BlockFieldMetadataValidatorTest.php
```diff
@@ -0,0 +1,155 @@
...
+            $formMetadata->addItem($item);
+        }
+
+        return $formMetadata;
+    }
+
+    /**
+     * @param ItemMetadata[] $items
+     */
+    private function createSectionMetadata(string $name = 'some_section', array $items = []): SectionMetadata
```
danrot, line 70: Same as above, would avoid the first default value.

src/Sulu/Bundle/AdminBundle/Tests/Unit/FormMetadata/Validation/BlockFieldMetadataValidatorTest.php
```diff
@@ -0,0 +1,155 @@
...
+        $sectionMetadata = new SectionMetadata($name);
+
+        foreach ($items as $item) {
+            $sectionMetadata->addItem($item);
+        }
+
+        return $sectionMetadata;
+    }
+
+    private function expectReservedPropertyNameException(string $formKey, string $blockName, string $propertyName)
```
danrot, line 81: Nice idea :slightly_smiling_face:

src/Sulu/Bundle/AdminBundle/Tests/Unit/FormMetadata/XmlFormMetadataLoaderTest.php
```diff
@@ -0,0 +1,133 @@
...
+            false
+        );
+    }
+
+    /**
+     * @param FormMetadata[] $types
+     */
+    private function createFieldMetadata(
+        string $name = 'some_property',
+        string $type = 'text_line',
```
danrot, line 66: Same as above

src/Sulu/Bundle/AdminBundle/Tests/Unit/FormMetadata/XmlFormMetadataLoaderTest.php
```diff
@@ -0,0 +1,133 @@
...
+            $fieldMetadata->addType($type);
+        }
+
+        return $fieldMetadata;
+    }
+
+    /**
+     * @param ItemMetadata[] $items
+     */
+    private function createFormMetadata(string $name = 'some_form', string $key = 'some_form_key', array $items = []): FormMetadata
```
danrot, line 82: Same as above

src/Sulu/Bundle/AdminBundle/Tests/Unit/FormMetadata/XmlFormMetadataLoaderTest.php
```diff
@@ -0,0 +1,133 @@
...
+            $formMetadata->addItem($item);
+        }
+
+        return $formMetadata;
+    }
+
+    /**
+     * @param ItemMetadata[] $items
+     */
+    private function createSectionMetadata(string $name = 'some_section', array $items = []): SectionMetadata
```
danrot, line 99: Same as above

src/Sulu/Bundle/AdminBundle/Tests/Unit/FormMetadata/dummy-forms/dummy-form.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" ?>
+<form xmlns="http://schemas.sulu.io/template/template"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://schemas.sulu.io/template/template http://schemas.sulu.io/template/form-1.0.xsd"
+>
+    <key>dummy-form</key>
```
danrot, line 6: Why is this form necessary?

danrot (changes requested):

src/Sulu/Bundle/AdminBundle/Tests/Unit/FormMetadata/dummy-forms/dummy-form.xml
luca-rath: because I can't mock the `Finder`, I need to provide a real directory with a real xml file

luca-rath: theoretically it could also be a completely empty xml file, but that would be very confusing IMO

src/Sulu/Bundle/AdminBundle/Tests/Unit/FormMetadata/StructureFormMetadataLoaderTest.php
```diff
@@ -0,0 +1,163 @@
...
+            false
+        );
+    }
+
+    /**
+     * @param FormMetadata[] $types
+     */
+    private function createFieldMetadata(
+        string $name,
+        string $type = 'text_line',
```
danrot, line 81: I think I would even make that one optional... Otherwise it wouldn't be clear if code like that should succeed or fail without looking at the function definition, which might be in a complete different location:

```php
$fieldMetadata = $this->createFieldMetadata('test');
$this->assertEquals('text_line', $fieldMetadata->getType());
```

src/Sulu/Bundle/AdminBundle/Tests/Unit/FormMetadata/Validation/BlockFieldMetadataValidatorTest.php
```diff
@@ -0,0 +1,155 @@
...
+    {
+        $this->blockFieldMetadataValidator = new BlockFieldMetadataValidator();
+    }
+
+    /**
+     * @param FormMetadata[] $types
+     */
+    private function createFieldMetadata(
+        string $name,
+        string $type = 'block',
```
danrot, line 39: Same as above. But if you prefer we could also introduce another method like `createBlockFieldMetadata` I would also be fine with that. Then a situation like mentioned above, when you don't really know what's happening from reading the code, cannot happen.

src/Sulu/Bundle/AdminBundle/Tests/Unit/FormMetadata/XmlFormMetadataLoaderTest.php
```diff
@@ -0,0 +1,133 @@
...
+            false
+        );
+    }
+
+    /**
+     * @param FormMetadata[] $types
+     */
+    private function createFieldMetadata(
+        string $name,
+        string $type = 'text_line',
```
danrot, line 66: Same as above

danrot (changes requested):
Sorry, I was talking about both parameters of these functions (items with an empty set is fine, this is really not needed in many cases and should not add any confusion from my current POV)

danrot:
@luca-rath Have fixed the remaining issues quickly myself and will merge as soon as tests are passing.

danrot (approved):

PR #5484, opened 2020-09-03
Throw exception if invalid default template is configured in webspace

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT

#### What's in this PR?

This PR throws an exception if a template is configured as default using the `default-template` tag in a webspace configuration, but this template does not exist in this webspace.

#### Why?

Because this leads just to confusing behavior, and the developer does not immediately know what's going wrong.


nnatter (approved):
looks good, but should we target `release/2.0` or `release/2.1`?

danrot:
@nnatter I think it shouldn't do any harm... It only throws an error earlier. Therefore I changed the target to 2.0.

PR #5487, opened 2020-09-07
Add target groups to block settings

### LINKED ISSUE #4997: Show different set of blocks based on target groups
# Problem description

We have implemented the Audience Targeting feature (#3148), which allows to segment your visitors in different target groups. Based on this target groups the SmartContent can show content that is interesting for the current visitor. However, this is currently only working for the smart content, which sometimes is not the correct field type to use.

In order to make this feature even more useful, it would be great if you could also assign target groups on blocks and only show the blocks interesting for the current visitor.

# Proposed solution

There should be a selection component similar to what is seen in the "Excerpt & Taxonomy" tab on the page, where you can choose for which target groups this block is interesting. This should probably be hidden behind some kind of overlay. The website will then skip the blocks not being relevant for the current visitor based on their target group.

What is a bit problematic about this feature, is that the number of blocks can explode, if you create multiple different blocks for many different target groups. Either we accept that and see how that goes, or think about a solution to this problem. Two particular ones come to mind:

1. **The target group should be handled like the language**, i.e. as different dimension. Then completely different pages for each target group can be created, not only involving blocks but any kind of property. The question is if this is desirable, because after our last research we came to the conclusion that it is bad for the search maching ranking, if the page changes too much.
2. **Always show all target groups and allow to filter them.** Then it is easier to get an overview how stuff looks like. However, it is not competely clear to me how we would have to design a dropdown for target groups or something like that in order to not be confused with actual content.

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #4997 
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR adds a selection of target groups to block settings. Then the audience targeting feature will filter the blocks accordingly.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Block/block.scss
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Block/block.scss
@@ -33,7 +33,6 @@ $blockIconsColor: $gray;
 
         .icons {
             border-right: 1px solid $blockBorderColor;
-            padding-right: 10px;
             margin-right: 10px;
         }
     }
@@ -73,6 +72,7 @@ $blockIconsColor: $gray;
 
 .types {
     min-width: 120px;
+    margin-right: 10px;
 }
 
 .type {
@@ -87,7 +87,7 @@ $blockIconsColor: $gray;
     font-size: 14px;
 
     > * {
-        margin-left: 10px;
+        margin-right: 10px;
     }
 }
 
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
@@ -119,7 +119,7 @@ class FieldBlocks extends React.Component<FieldTypeProps<Array<BlockEntry>>> {
 
             const blockSettingsTag = schemaEntry.tags.find((tag) => tag.name === SETTINGS_TAG);
 
-            if (blockSettingsTag) {
+            if (blockSettingsTag && schemaEntry.visible !== false) {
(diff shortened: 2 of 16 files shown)


nnatter (commented):
Code looks good! But would like to improve two things:

* If I disable the `AudienceTargetingBundle` after setting the target-group for some blocks, the administration interface still displays the user icon in the block header. Would be nice to check the `visible` property of the `schemaEntry` in the `iconsMapping` getter of the `FieldBlocks`.

Screenshot 2020-09-17 at 13 59 03

* The first icon in the block header has a margin and therefore is intended compared to the other elements when the block is expanded (see screenshot above). It would be nice if this aligns with the properties of the block. This should be possible by the following diff:

```diff
diff --git a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Block/block.scss b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Block/block.scss
index f5f9b3f9c3..e27729f055 100644
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Block/block.scss
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Block/block.scss
@@ -33,7 +33,6 @@ $blockIconsColor: $gray;

         .icons {
             border-right: 1px solid $blockBorderColor;
-            padding-right: 10px;
             margin-right: 10px;
         }
     }
@@ -87,7 +86,7 @@ $blockIconsColor: $gray;
     font-size: 14px;

     > * {
-        margin-left: 10px;
+        margin-right: 10px;
     }
 }
```

danrot:
Fixed both issues, but considering the following:

- For the first issue you should also test if the icons are correctly places if some types are available on the block (your solution would not have worked for that).
- Had to duplicate the `visibleCondition`, because the item having the tag for the icon didn't have it, and therefore couldn't be checked.

nnatter (approved):

PR #5428, opened 2020-07-23
Fix portal route matching

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #5489
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

Fix portal route matching.

#### Why?

Currently routes e.g.: `/_profiler/search`  does also match our search, only routes with a specific portal prefix should match.

#### To Do

- [ ] Add Tests

PR #5498, opened 2020-09-15
Adjust portal route type to match path prefix against webspace configuration

### LINKED ISSUE #5489: File name "search" activates WebsiteSearchController
| Q | A
| --- | ---
| Sulu Version | 2.1.2

# Actual Behavior

```
Sulu\Component\Rest\Exception\MissingParameterException:
Missing parameter q in Sulu\Bundle\SearchBundle\Controller\WebsiteSearchController

  at vendor/sulu/sulu/src/Sulu/Component/Rest/RequestParametersTrait.php:38
  at Sulu\Bundle\SearchBundle\Controller\WebsiteSearchController->getRequestParameter(object(Request), 'q', true)
     (vendor/sulu/sulu/src/Sulu/Bundle/SearchBundle/Controller/WebsiteSearchController.php:69)
  at Sulu\Bundle\SearchBundle\Controller\WebsiteSearchController->queryAction(object(Request))
     (vendor/symfony/http-kernel/HttpKernel.php:158)
  at Symfony\Component\HttpKernel\HttpKernel->handleRaw(object(Request), 1)
     (vendor/symfony/http-kernel/HttpKernel.php:80)
  at Symfony\Component\HttpKernel\HttpKernel->handle(object(Request), 1, true)
     (vendor/symfony/http-kernel/Kernel.php:201)
  at Symfony\Component\HttpKernel\Kernel->handle(object(Request))
     (public/index.php:63)
```

# Expected Behavior

show the image that was requested via ```/build/website/icons/search.svg```

# Steps to Reproduce

In a Sulu System call any file in any subfolder named "search" (+ any extension or none at all)

# Possible Solutions

No concrete ideas, but the WebsiteSearchController should not respond every time a file has the name "search". Workaround for me was to suffix all icons with '-icon'

nnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #5489
| Related issues/PRs | #5428
| License | MIT

#### What's in this PR?

This PR adjusts the `PortalLoader` to generate routes that match only if the path prefix is configured for a webspace.
It builds upon #5428, as I am not able to test this on `release/1.6`. Would be nice if someone else is motivated to backport it 🙂 

#### Why?

At the moment, the path `/_profiler/search` does also match the search portal route. This is wrong, as only routes with a specific portal prefix should match.

### DIFF
--- a/src/Sulu/Bundle/WebsiteBundle/Routing/PortalLoader.php
+++ b/src/Sulu/Bundle/WebsiteBundle/Routing/PortalLoader.php
@@ -11,7 +11,6 @@
 
 namespace Sulu\Bundle\WebsiteBundle\Routing;
 
-use Sulu\Component\Webspace\Analyzer\RequestAnalyzerInterface;
 use Sulu\Component\Webspace\Manager\WebspaceManagerInterface;
 use Symfony\Component\Config\Loader\Loader;
 use Symfony\Component\Routing\Route;
@@ -38,11 +37,11 @@ class PortalLoader extends Loader
         $prefixes = [];
         foreach ($this->webspaceManager->getPortalInformations() as $portalInformation) {
             // cast null to string as prefix can be empty string
-            $prefixes[] = preg_quote((string) $portalInformation->getPrefix());
+            $prefixes[] = \preg_quote((string) $portalInformation->getPrefix());
         }
 
         // need to omit prefix from path if it must be empty to pass symfony route validation
-        $prefixPattern = implode('|', array_unique($prefixes));
+        $prefixPattern = \implode('|', \array_unique($prefixes));
         $pathPrefix = empty($prefixPattern) ? '' : '{prefix}';
         $requirements = empty($prefixPattern) ? [] : ['prefix' => $prefixPattern];
 
--- a/src/Sulu/Bundle/WebsiteBundle/Routing/PortalLoader.php
+++ b/src/Sulu/Bundle/WebsiteBundle/Routing/PortalLoader.php
@@ -12,6 +12,7 @@
 namespace Sulu\Bundle\WebsiteBundle\Routing;
 
 use Sulu\Component\Webspace\Analyzer\RequestAnalyzerInterface;
+use Sulu\Component\Webspace\Manager\WebspaceManagerInterface;
 use Symfony\Component\Config\Loader\Loader;
 use Symfony\Component\Routing\Route;
 use Symfony\Component\Routing\RouteCollection;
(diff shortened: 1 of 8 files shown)


src/Sulu/Bundle/WebsiteBundle/Tests/Functional/Routing/PortalLoaderTest.php
```diff
@@ -0,0 +1,49 @@
...
+     * @var KernelBrowser
+     */
+    private $client;
+
+    public function setUp(): void
+    {
+        $this->client = static::createWebsiteClient();
+    }
+
+    public function testPortalRoute()
```
nnatter, line 29: unfortunately, the behaviour for `http://example.lo/valid-english-prefix/portal-route` is still wrong. but i dont see an easy way to fix that 😕

src/Sulu/Bundle/WebsiteBundle/Routing/PortalLoader.php
```diff
@@ -22,33 +23,45 @@
...
+        $prefixes = [];
+        foreach ($this->webspaceManager->getPortalInformations() as $portalInformation) {
+            // cast null to string as prefix can be empty string
+            $prefixes[] = preg_quote((string) $portalInformation->getPrefix());
+        }
+
+        // need to omit prefix from path if it must be empty to pass symfony route validation
+        $prefixPattern = implode('|', array_unique($prefixes));
+        $pathPrefix = empty($prefixPattern) ? '' : '{prefix}';
+        $requirements = empty($prefixPattern) ? [] : ['prefix' => $prefixPattern];
```
alexander-schranz, line 47: prefix need always exist else it will break things.

src/Sulu/Bundle/WebsiteBundle/Tests/Functional/Routing/PortalLoaderTest.php
nnatter, line 29: unfortunately, the behaviour for http://example.lo/valid-english-prefix/portal-route is still wrong. but i dont see an easy way to fix that 😕  i think we would need to register one route for each `PortalInformation`

src/Sulu/Bundle/WebsiteBundle/Routing/PortalLoader.php
nnatter: when would this happen? 🤔 
if there is a single webspace that uses `{host}` as url, the prefix requirement will be an empty string. this causes the following error:
```
Routing requirement for "prefix" cannot be empty in @SuluSearchBundle/Resources/config/routing_website.yml
```

alexander-schranz: Seems like then another BC Break then because at current state with empty resourceLocatorPrefix url generation works correctly:

```yaml
{{ path('sulu_community.password_forget', { prefix: request.resourceLocatorPrefix }) }}
```

or 

```yaml
{{ path('sulu_community.password_forget', { prefix: '' }) }}
```

As said we can not remove the `prefix` as this would be a bc break which accidentally adding a query parameter which should not be there. The route path should not change because I add in future another webspace or change the webspace configuration. We should just edit the current regex that it allow only the configured prefixes.

nnatter: its not about the generation - the problem is that symfony does not accept a route with an empty requirement.. in the case described above, the regex (`$prefixPattern`) would be an empty string. this leads to the exception

alexander-schranz: Think we will find a regex which will match this and symfony doesn't throw an error.

alexander-schranz: The following should work:

```
static:
    path: /{prefix}static-route
    controller: App\Controller\DefaultController::index
    requirements:
        prefix: '(^$)?|test/'
```

So the prefix for empty resourceLocatorPrefix should be `(^$)?`.

nnatter: you are right 👍  i figured that `()` works too and updated the PR

alexander-schranz (approved):

PR #5504, opened 2020-09-21
Add updating of exist anonymous roles

### LINKED ISSUE #5499: sulu:security:init does not add anonymous access for new webspaces
| Q | A
| --- | ---
| Sulu Version | 74d39b37aa536d76dbc8adb7fd5a44653fd95998

# Actual Behavior

When you already updated sulu and initialized the security. When adding new webspaces anonymous users wont be able to access those pages because the wont have the role permissions. when executing sulu:security:init again it wont add those permissions.

# Expected Behavior

Add anonymous permissions for new webspaces or add toggler to the role overview to show the anonymous role and edit their permission.

# Steps to Reproduce

init security add new webspace and init security again there wont be any access.

# Temporary Solution
Add the same permissions for the new webspace like the old one in se_permissions for the anonymous role.

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #5499
| Related issues/PRs | 
| License | MIT

#### What's in this PR?

Add updating of exist anonymous roles.

#### Why?

As the content manager can not add permission for anonymous user a new added webspace need to add permission over the `sulu:security:init` command else the anonymous user will not have access to a new webspace in the same security system.

#### Example Usage

1. `cp sulu-blog.xml sulu-blog2.xml`
2. Change `<key>` to `sulu-blog2`
3. `bin/console sulu:security:init`

Exist anonymous role for same system should be updated and anonymous role should now have access to the new webspace.

### DIFF
--- a/src/Sulu/Bundle/SecurityBundle/Command/InitCommand.php
+++ b/src/Sulu/Bundle/SecurityBundle/Command/InitCommand.php
@@ -102,14 +102,14 @@ final class InitCommand extends Command
             }
 
             $permissionAdded = false;
-            $existSecurityContexts = [];
+            $existingSecurityContexts = [];
 
             foreach ($role->getPermissions() as $permission) {
-                $existSecurityContexts[] = $permission->getContext();
+                $existingSecurityContexts[] = $permission->getContext();
             }
 
             foreach ($securityContextsFlat as $securityContext) {
-                if (\in_array($securityContext, $existSecurityContexts)) {
+                if (\in_array($securityContext, $existingSecurityContexts)) {
                     continue;
                 }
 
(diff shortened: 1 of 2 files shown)


src/Sulu/Bundle/SecurityBundle/Command/InitCommand.php
```diff
@@ -110,27 +101,62 @@ protected function execute(InputInterface $input, OutputInterface $output)
                 }
             }
 
+            $permissionAdded = false;
+            $existSecurityContexts = [];
```
danrot, line 105: `$existingSecurityContexts`

danrot (changes requested):
Just a minor change in order to be consistent. And I think we should also write tests for commands... Especially this one contains some logic, and they break so easily.

danrot:
@alexander-schranz I would still add a test for that command... There is already quite some logic involved, and it is not tested at all :confused:

alexander-schranz:
@danrot Test added!

PR #5505, opened 2020-09-23
Update BlockCollection arrays on setting new block-data

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Add a `reaction` to call the `fillArrays` function when the `blocks` data is updated e.g. via the `ResourceFormStore`.

#### Why?

For a project I have the use-case, that I have to copy data from `Tab 1` to `Tab 2`.
The `fillArrays` method is only called in the `constructor`, thus after copying the `blocks` from `Tab 1` to `Tab 2` the two arrays `generatedBlockIds` and `expandedBlocks` are not updated accordingly.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/BlockCollection.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/BlockCollection.js
@@ -37,10 +37,10 @@ class BlockCollection extends React.Component<Props> {
         super(props);
 
         this.fillArrays();
-        reaction(() => this.props.value.length, this.fillArrays.bind(this));
+        reaction(() => this.props.value.length, this.fillArrays);
     }
 
-    fillArrays() {
+    fillArrays = () => {
         const {defaultType, onChange, minOccurs, value} = this.props;
         const {expandedBlocks, generatedBlockIds} = this;
 
@@ -75,7 +75,7 @@ class BlockCollection extends React.Component<Props> {
                 ),
             ]);
         }
-    }
+    };
 
     @action handleAddBlock = () => {
         const {defaultType, onChange, value} = this.props;
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/BlockCollection.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/BlockCollection.js
@@ -1,6 +1,6 @@
 // @flow
 import React from 'react';
-import {action, observable, toJS} from 'mobx';
+import {action, observable, toJS, reaction} from 'mobx';
 import {observer} from 'mobx-react';
 import {arrayMove} from 'react-sortable-hoc';
 import {translate} from '../../utils/Translator';
@@ -37,14 +37,24 @@ class BlockCollection extends React.Component<Props> {
         super(props);
 
         this.fillArrays();
+        reaction(() => this.props.value.length, this.fillArrays.bind(this));
     }
 
     fillArrays() {
         const {defaultType, onChange, minOccurs, value} = this.props;
         const {expandedBlocks, generatedBlockIds} = this;
 
-        if (!value) {
+        if ((!value || expandedBlocks.length === value.length || generatedBlockIds.length === value.length)
+                && value.length !== 0) {
             return;
+        }
+
+        if (expandedBlocks.length > value.length) {
+            expandedBlocks.length = 0;
+        }
+
+        if (generatedBlockIds.length > value.length) {
+            generatedBlockIds.length = 0;
         }
 
         expandedBlocks.push(...new Array(value.length - expandedBlocks.length).fill(false));
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/BlockCollection.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/BlockCollection.js
@@ -44,17 +44,16 @@ class BlockCollection extends React.Component<Props> {
         const {defaultType, onChange, minOccurs, value} = this.props;
         const {expandedBlocks, generatedBlockIds} = this;
 
-        if ((!value || expandedBlocks.length === value.length || generatedBlockIds.length === value.length)
-                && value.length !== 0) {
+        if (!value) {
             return;
         }
 
         if (expandedBlocks.length > value.length) {
-            expandedBlocks.length = 0;
+            expandedBlocks.splice(value.length);
         }
 
         if (generatedBlockIds.length > value.length) {
-            generatedBlockIds.length = 0;
+            generatedBlockIds.splice(value.length);
         }
 
         expandedBlocks.push(...new Array(value.length - expandedBlocks.length).fill(false));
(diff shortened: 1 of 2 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/BlockCollection.js
```diff
@@ -37,16 +37,26 @@ class BlockCollection extends React.Component<Props> {
...
             return;
         }
 
+        if (expandedBlocks.length > value.length) {
+            expandedBlocks.length = 0;
+        }
+
+        if (generatedBlockIds.length > value.length) {
+            generatedBlockIds.length = 0;
+        }
```
danrot, line 58: I don't really get why these two `if`s are necessary... Can you elaborate?

danrot (changes requested):
And add the missing test :slightly_smiling_face:

src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/BlockCollection.js
Prokyonn: Those are necessary for the case, when at first we have e.g. 5 `blocks` in the `value` array, thus `generatedBlockIds` and `expandedBlocks`, all 3 arrays have a length of 5

then we change the data in `value` e.g. through a toolbarAction to only 2 `blocks` and now we also have to adjust the `expandedBlocks` and `generatedBlockIds` otherwise we would run into an error in the following lines

`expandedBlocks.push(...new Array(value.length - expandedBlocks.length).fill(false));` and 
`        generatedBlockIds.push(
            ...new Array(value.length - generatedBlockIds.length).fill(false).map(() => ++BlockCollection.idCounter)
        );`

danrot: Wouldn't it be better to handle this case differently? Just sketching something for only the `expandedBlocks` array:

```javascript
if (expandedBlocks.length > value.length) {
    expandedBlocks.splice(value.length);
}
```

This would have the advantage, that the old values stay. As I see it, the `reaction` you've wrote is also fired when a new block is added using the button, and then all blocks would have to collapse. Have you tested that?

src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/tests/BlockCollection.test.js
```diff
@@ -406,3 +406,52 @@ test('Should apply renderBlockContent before rendering the block content includi
...
         .toEqual(prefix + value[1].content + typePrefix + 'type1');
 });
+
+test('Should adjust expandedBlocks and generatedBlockIds after updating the value variable', () => {
+    const types = {
+        type1: 'Type 1',
+        type2: 'Type 2',
+    };
+
+    let value: any = [
```
danrot, line 416: Why is `any` necessary here? Isn't there a better way to handle this?

src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/tests/BlockCollection.test.js
```diff
@@ -406,3 +406,52 @@ test('Should apply renderBlockContent before rendering the block content includi
...
+            type: 'type1',
+            content: 'Test 1',
+        },
+    ];
+
+    expect(blockCollection.props().value.length).toBe(3);
+    expect(blockCollection.instance().expandedBlocks.length).toBe(3);
+    expect(blockCollection.instance().generatedBlockIds.length).toBe(3);
+
+    blockCollection.setProps({value});
```
danrot, line 452: Can't we just insert the content of the `value` variable here? Shortly got confused by the changing of the `value` variable, because I've expected the previous assertions to be already updated :see_no_evil:

src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/tests/BlockCollection.test.js
danrot: Quickly tested it, does not seem to be necessary at all :see_no_evil:

src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/BlockCollection.js
Prokyonn: I guess you are right, this would also work. I will change it later.

Yes I have tested it.
If it is fired when a block is added or deleted, this condition is true 
```
 if ((!value || expandedBlocks.length === value.length || generatedBlockIds.length === value.length)
                && value.length !== 0) {
            return;	            return;
        }	        
}
```

because `expandedBlocks` and `generatedBlockIds` were already updated during the `add block` method.

danrot: Ah I see... That was not obvious on the first look. I think the other way (without making the initial `if` with the early return more complicated) would be easier to understand, do you agree? Plus it would have the advantage that the values stay the same, which might not be important in your use case, but maybe in a different one.

danrot (changes requested):

src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/BlockCollection.js
Prokyonn: Yes I agree, fixed it :)

src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/tests/BlockCollection.test.js
```diff
@@ -406,3 +406,56 @@ test('Should apply renderBlockContent before rendering the block content includi
     expect(blockCollection.find('Block').at(1).prop('children'))
         .toEqual(prefix + value[1].content + typePrefix + 'type1');
 });
+
+test('Should adjust expandedBlocks and generatedBlockIds after updating the value variable', () => {
```
danrot, line 410: Can you also write a test for the other cases (when the same amount resp. more elements are set in `setProps`)?

danrot (changes requested):
Otherwise it looks fine now, will test it after the last changes.

src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/tests/BlockCollection.test.js
```diff
@@ -406,3 +406,170 @@ test('Should apply renderBlockContent before rendering the block content includi
...
+        ],
+    });
+
+    expect(blockCollection.props().value.length).toBe(3);
+    expect(blockCollection.instance().expandedBlocks.length).toBe(3);
+    expect(blockCollection.instance().generatedBlockIds.length).toBe(3);
+    expect(blockCollection.instance().expandedBlocks[0]).toBe(true);
+});
+
+test('Should not adjust expandedBlocks and generatedBlockIds after updating the value variable with the same ' +
```
danrot, line 516: This function misses the closing parantheses. And can we slight rename it, so that everything fits in one line? Using this `+` looks weird :see_no_evil: How about something more concise like:

Updating value with same length should not adjust expandedBlocks and generatedBlockIds.

danrot (changes requested):

src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/tests/BlockCollection.test.js
Prokyonn: Fixed, thanks for the better naming :)

danrot (approved):

PR #5508, opened 2020-09-28
Add selectable option to List view

nnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This PR adds a `selectable` option to the `List` view. The option is implemented in the same way as the existing `searchable` option.

#### Why?

Because there are valid usecases for the `List` view in which the items dont need to be selectable. In these cases, it is confusing from an UX perspective to render them as selectable.


src/Sulu/Bundle/AdminBundle/Admin/View/ListViewBuilderTrait.php
```diff
@@ -62,6 +62,11 @@ private function setSearchableToView(View $route, bool $searchable): void
         $route->setOption('searchable', $searchable);
     }
 
+    private function setSelectableToView(View $route, bool $searchable): void
```
danrot, line 65: The second parameter should probably be `$selectable`

danrot (changes requested):

danrot:
Just tested it, everything works fine :+1: Will merge when the above change is done and tests succeed again.

src/Sulu/Bundle/AdminBundle/Admin/View/ListViewBuilderTrait.php
nnatter: 👍

danrot:
Rebased it, to make sure the circle ci build is also working (failed because of the eslint rules we've fixed).

PR #3463, opened 2017-08-01
added border box styling and created global style file

3rfan:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | yes
| Deprecations? | no
| License | MIT

#### What's in this PR?

In this PR the box-sizing css property is globally set to `border-box`. The application component now includes a `global.scss` file and an `application.scss`.

#### Why?

Because the default box-sizing sucks.

#### BC Breaks/Deprecations

All layout size properties (width, height, padding, etc.) has now to be set with `border-box` in mind but no issues were found in the current state of the application.

PR #5513, opened 2020-09-30
Remove null-loader because it is not used anymore

danrot:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR removes the `null-loader` dependency.

#### Why?

Because is not used anymore since #3463

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,7 +67,6 @@
         "mobx": "^4.0.0",
         "mobx-react": "^5.0.0",
         "moment-timezone": "^0.5.14",
-        "null-loader": "^3.0.0",
         "optimize-css-assets-webpack-plugin": "^5.0.3",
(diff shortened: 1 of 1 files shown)


alexander-schranz:
@danrot looks for me we could remove this also from the `2.0` branch?

danrot:
@alexander-schranz We probably could, but if somebody relies on our repository to install that, and they forgot to add it in their own `package.json`, it might end up being a BC break... Therefore I would do it only on `master`, especially because having it still there does not hurt too much, except for the need that we would have to update it all the time.

PR #5516, opened 2020-10-01
Include filters and search to list export button

### LINKED ISSUE #4943: Filters are not taken into account during export csv lists
| Q | A
| --- | ---
| Bug? | yes
| New Feature? | no
| Sulu Version | 1.6.29

#### Actual Behavior

During lists csv export, filters are not taken into account.

#### Expected Behavior

When a filter is selected the export should take it into account.

#### Steps to Reproduce

Set up filters and csv export function for a list of items.
https://docs.sulu.io/en/1.6/cookbook/integrating-custom-filters.html

sanhofman:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #4943 
| Related issues/PRs | #4943 
| License | MIT

#### What's in this PR?

Pass filters and search parameters to export.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/views/List/toolbarActions/ExportToolbarAction.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/views/List/toolbarActions/ExportToolbarAction.js
@@ -130,7 +130,7 @@ export default class ExportToolbarAction extends AbstractListToolbarAction {
             enclosure: this.enclosure,
             newLine: this.newLine,
             ...this.listStore.options,
-            filter: filter,
+            filter,
             search,
         }));
         this.showOverlay = false;
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/views/List/toolbarActions/ExportToolbarAction.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/views/List/toolbarActions/ExportToolbarAction.js
@@ -136,7 +136,7 @@ export default class ExportToolbarAction extends AbstractListToolbarAction {
             newLine: this.newLine,
             ...this.listStore.options,
             filter: filters,
-            search: search
+            search: search,
         }));
         this.showOverlay = false;
     };
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/views/List/toolbarActions/ExportToolbarAction.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/views/List/toolbarActions/ExportToolbarAction.js
@@ -116,15 +116,10 @@ export default class ExportToolbarAction extends AbstractListToolbarAction {
     };
 
     @action handleConfirm = () => {
-        let filters = {};
-        if (Object.keys(this.listStore.filterQueryOption).length > 0) {
-            filters = this.listStore.filterQueryOption;
-        }
+        const {filterQueryOption} = this.listStore;
+        const filter = Object.keys(filterQueryOption).length > 0 ? filterQueryOption : undefined;
 
-        let search = null;
-        if (this.listStore.searchTerm.get()) {
-            search = this.listStore.searchTerm.get();
-        }
+        const search = this.listStore.searchTerm.get();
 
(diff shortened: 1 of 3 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/views/List/toolbarActions/ExportToolbarAction.js
```diff
@@ -125,6 +135,8 @@ export default class ExportToolbarAction extends AbstractListToolbarAction {
             enclosure: this.enclosure,
             newLine: this.newLine,
             ...this.listStore.options,
+            filter: filters,
+            search: search,
```
alexander-schranz, line 139: Lint is failing this should just be:

```suggestion
            search,
```

alexander-schranz (changes requested):
@sanhofman Nice 👍, thank you for working on this.

I think we should also add a test for this, maybe @danrot can help you here.

src/Sulu/Bundle/AdminBundle/Resources/js/views/List/toolbarActions/ExportToolbarAction.js
```diff
@@ -116,6 +116,16 @@ export default class ExportToolbarAction extends AbstractListToolbarAction {
     };
 
     @action handleConfirm = () => {
+        let filters = {};
+        if (Object.keys(this.listStore.filterQueryOption).length > 0) {
+            filters = this.listStore.filterQueryOption;
+        }
```
danrot, line 122: Instead of using `let` we try to make use of `const` as much as possible. You can achieve that using a ternary operator. We also make use of destructuring a lot, to avoid repeating certain stuff. So I would change that section to:

```javascript
const {filterQueryOption} = this.listStore;
const filter = Object.keys(filterQueryOption).length > 0 ? filterQueryOption : undefined;
```

src/Sulu/Bundle/AdminBundle/Resources/js/views/List/toolbarActions/ExportToolbarAction.js
```diff
@@ -116,6 +116,16 @@ export default class ExportToolbarAction extends AbstractListToolbarAction {
...
+
+        let search = null;
+        if (this.listStore.searchTerm.get()) {
+            search = this.listStore.searchTerm.get();
+        }
```
danrot, line 127: Since `get()` should return `undefined` already if nothing is passed, something like that should suffice:

```javascript
const search = this.listStore.searchTerm.get();
```

danrot:
Although @alexander-schranz is right that a test would be nice, I had to realize that there are no tests yet :open_mouth: Therefore I won't ask you to add one, just fix the issues above, and I'll add the tests afterwards.

danrot (changes requested):

src/Sulu/Bundle/AdminBundle/Resources/js/views/List/toolbarActions/ExportToolbarAction.js
sanhofman: Thanks for the tip @danrot! Refactored in latest commit.

src/Sulu/Bundle/AdminBundle/Resources/js/views/List/toolbarActions/ExportToolbarAction.js
sanhofman: Thanks for the tip @danrot! Refactored in latest commit.

src/Sulu/Bundle/AdminBundle/Resources/js/views/List/toolbarActions/ExportToolbarAction.js
sanhofman: Thanks for the tip @alexander-schranz! Refactored in latest commit.

It seems Lint still failing, due to filter: filter?

danrot:
Tested, it works as expected :-) Added the tests as promised.

danrot (approved):

nnatter (approved):

PR #5522, opened 2020-10-05
Fix image map on safari

luca-rath:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Fixes `CircleSelection`/`RectangleSelection` performance issues on safari

#### Why?

Because safari is a supported browser and therefore users should not experience any issues when using it.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/RectangleSelection/ModifiableRectangle.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/RectangleSelection/ModifiableRectangle.js
@@ -31,6 +31,7 @@ class ModifiableRectangle extends React.Component<Props> {
 
     @observable moveMode = false;
     @observable resizeMode = false;
+    @observable clickAnchor = {pageY: 0, pageX: 0};
 
     componentDidMount() {
         window.addEventListener('mouseup', this.handleMouseUp);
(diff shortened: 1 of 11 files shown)


src/Sulu/Bundle/MediaBundle/Resources/js/containers/ImageMap/imageRenderer.scss
```diff
@@ -6,14 +6,15 @@
 .image-renderer-wrapper {
     position: relative;
     overflow: hidden;
-    display: inline-flex;
```
luca-rath, line 9: safari somehow doesn't like `inline-flex`

src/Sulu/Bundle/AdminBundle/Resources/js/components/RectangleSelection/ModifiableRectangle.js
```diff
@@ -67,26 +75,24 @@ class ModifiableRectangle extends React.Component<Props> {
 
     @action handleMouseMove = (event: MouseEvent) => {
         const {onChange} = this.props;
-
-        if (!onChange) {
-            return;
-        }
-
-        const {movementX, movementY} = event;
```
luca-rath, line 75: safari's `movementX` behaves different than chrome's `movementX` does. On chrome, `movementX` stops increasing, if the cursor reached the edge of the screen. On safari, `movementX` increases as long as you scroll on the touch pad, even if the cursor already stopped at the edge of the screen.

luca-rath: ofc same for `movementY`

luca-rath:
@danrot I guess the failing appveyor task isn' related to this pr

danrot:
This also seems to fix the issues I've had with Chrome on linux, which cause the `RectangleSelection` and `CircleSelection` to be moving about twice the speed of the cursor.

danrot (approved):

danrot:
Going to merge it, since it solves also my chrome problem.

PR #5525, opened 2020-10-05
Update react-datepicker

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR updates to the latest version of react-datepicker. However, they have introduce a BC break I tried to fix in https://github.com/arqex/react-datetime/pull/741, let's see if they merge it. We have to wait for it, because currently the datepicker does not close itself automatically after a date has been chosen.

#### Why?

Because we want to use the latest versions.


alexander-schranz:
closed in favor of #5956

PR #5526, opened 2020-10-05
Fix media image extractor with unseekable resources

### LINKED ISSUE #5468: Flysystem s3 update breaks thumbnailing behavior
| Q | A
| --- | ---
| Sulu Version | Since 2.0.0
| Browser Version | any

# Actual Behavior

You get the following warning `Warning: mime_content_type(): stream does not support seeking` on https://github.com/sulu/sulu/blob/master/src/Sulu/Bundle/MediaBundle/Media/ImageConverter/MediaImageExtractor.php#L52 when doing any media resize operation. This is caused by https://flysystem.thephpleague.com/v1/docs/adapter/aws-s3-v3/#streamed-reads.

# Expected Behavior

media operations should work with newest version of flysystem-aws-s3-v3 (1.0.28+)

# Steps to Reproduce

Base sulu installation where the media is stored on s3.

**config/packages/sulu_media.yaml**
```
sulu_media:
    storage: s3
    storages:
        s3:
            key: '%env(S3_MEDIA_BUCKET_KEY)%'
            secret: '%env(S3_MEDIA_BUCKET_SECRET)%'
            bucket_name: '%env(S3_MEDIA_BUCKET_NAME)%'
            endpoint: '%env(S3_MEDIA_BUCKET_ENDPOINT)%'
            path_prefix: 'media'
            region: 'eu-west-1'
            arguments:
                use_path_style_endpoint: true
                bucket_endpoint: false
```
after that make sure league/flysystem-aws-s3-v3 is at 1.0.28 and you should get an error on any media resize operation

# Possible Solutions

### solution 1 
For now I locked in the version of flysystem-aws-s3-v3
```
"league/flysystem-aws-s3-v3": ">1 <=1.0.24",
```

### solution 2
Flysystems already made a suggestion that it's possible return to the previous behavior by calling the AwsS3Adapter with an additional false.
```
$adapter = new AwsS3Adapter($client, 'your-bucket-name', 'optional/path/prefix', [], false /** disable streamed reads **/);
```

### solution 3
A more permanent solution might be to read out the wrapper_data of the stream as it already contains the information for which the `mime_content_type` is currently called for.
Screenshot 2020-08-27 at 12 07 01
I'm not sure if this will work for other (cloud) storage systems

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #5468
| Related issues/PRs | -
| License | MIT

#### What's in this PR?

Use the saved file version mimetype instead of try to seek it from the blob.

#### Why?

Fix media image extractor with unseekable resources.


src/Sulu/Bundle/MediaBundle/Media/ImageConverter/ImagineImageConverter.php
```diff
@@ -136,12 +136,15 @@ public function getSupportedOutputImageFormats(?string $mimeType): array
 
     public function convert(FileVersion $fileVersion, $formatKey, $imageFormat)
     {
-        $imageResource = $this->mediaImageExtractor->extract($this->storage->load($fileVersion->getStorageOptions()));
+        $imageResource = $this->mediaImageExtractor->extract(
+            $this->storage->load($fileVersion->getStorageOptions()),
+            $fileVersion->getMimeType()
```
nnatter, line 141: this is set on upload right? how/where do we detect it then? 🙂

src/Sulu/Bundle/MediaBundle/Media/ImageConverter/MediaImageExtractor.php
```diff
@@ -47,9 +47,18 @@ public function __construct(
         $this->videoThumbnail = $videoThumbnail;
     }
 
-    public function extract($resource)
+    public function extract($resource/*, string $resourceMimeType*/)
```
nnatter, line 50: do you think this is better than making the `$resourceMimeType` nullable? i think ide support and static code analysis would be better with a nullable parameter

alexander-schranz: That would be a bc break as the class is not final and somebody could extend it.

src/Sulu/Bundle/MediaBundle/Media/ImageConverter/ImagineImageConverter.php
alexander-schranz: First the media is uploaded to PHP Server there we detect the mimetype before it is then store over flyssystem in a external storage.

alexander-schranz:
@nnatter travis should now run correctly!

niklasnatter (approved):

PR #5529, opened 2020-10-06
Update history package

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR updates the `history` package to its latest version. There a [lots of undocumented breaking changes](https://github.com/ReactTraining/history/issues/811), but the most important ones for us are:

- The `entries` are not available anymore externally, therefore I had to rewrite some tests
- The `push` function does not reset the `search` property of the `location` anymore (they argue that the browser API itself handles it the same way), had to adjust the `Router` because of that

#### Why?

Because we want to use up to date dependencies.


nnatter (approved):
Have not found any obvious bugs during a short test 🙂

Issue #5288, opened 2020-05-13
Sort page templates alphabetically in current language


man-andy:
# Problem description

Currently the page templates are sorted by the order in the xml template, instead I think it would be great if the dropdown is sorted alphabetically by the users language for better user experience.

Related Issue:sulu/sulu/issues/5086

# Proposed solution

A. Sort block types alphabetically in current language by default

B. Make sorting of block types configureable e.g.: `<types sort-by="title">`

danrot:
Just to clarify this: Your proposed solution does not really affect what you are describing here, does it? I think currently the page templates are sorted by their keys, and since every template has its own file, which does not know anything about the others, there is no way to configure this via some XML configuration.

The way I see it, we would have to change the default sorting. I think I would support that,
because sorting them by key does not really make sense. The only disadvantage I can think about, is that the positioning of these elements is different for every language then, so if two people would be talking about the "first template" they could mean something different. But I don't think that this would be really a problem.

@sulu/core-team Does anything speak against sorting the page templates by the localized title? If not, I would quickly implement this.

nnatter:
Sorting by title instead of key sounds sensible to me. If somebody wants to have the same order for all languages, he could still make things explicit and prefix the title with numbers for his desired order.

Not sure how webspaces are sorted at the moment, but we should probably implement the same sorting there.

danrot:
@man-andy Thanks for raising the issue! This will be fixed in Sulu 2.2.

PR #5532, opened 2020-10-07
Sort webspaces by name in administration interface

nnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Related issues/PRs | #5288
| License | MIT

#### What's in this PR?

This PR adjusts the `PageAdmin` to pass the webspaces to the frontend sorted by their name. This changes the order of the webspaces for the following components: `WebspaceTabs`, `SegmentSelect`, `Preview`, `Permissions`

#### Why?

Because there should be a understandable sorting. See #5288


src/Sulu/Bundle/PageBundle/Tests/Unit/Admin/PageAdminTest.php
```diff
@@ -124,9 +124,29 @@ public function testGetConfigWithVersioning()
...
+        $this->teaserProviderPool->getConfiguration()->willReturn([]);
+
         $config = $admin->getConfig();
 
-        $this->assertEquals(true, $config['versioning']);
+        $this->assertEquals(
+            [
+                'teaser' => [],
+                'versioning' => true,
+                'webspaces' => [$webspace1->reveal(), $webspace2->reveal()],
```
danrot, line 146: Shouldn't `$webspace2` be listed here first, because `f` comes before `s` in the alphabet? :thinking:

src/Sulu/Bundle/PageBundle/Tests/Unit/Admin/PageAdminTest.php
```diff
@@ -124,9 +124,29 @@ public function testGetConfigWithVersioning()
             true
         );
 
+        $webspace1 = $this->prophesize(Webspace::class);
+        $webspace1->getName()->willReturn('second webspace');
+
+        $webspace2 = $this->prophesize(Webspace::class);
+        $webspace2->getName()->willReturn('first webspace');
```
danrot, line 131: I know that this is mixed in this test file, but I think I would prefer to use the real `Webspace` implementation here. It is the exact lines of code (even less characters), and we could be absolutely sure, that this is working with the real implementation, not only with the mock:

```php
$webspace2 = new Webspace();
$webspace2->setName('first webspace');
```

danrot (changes requested):

src/Sulu/Bundle/PageBundle/Tests/Unit/Admin/PageAdminTest.php
nnatter: you are right 🙈 the problem was that we keep the array key because it is set to the webspace-key. adjusted the test and added an additional check

src/Sulu/Bundle/PageBundle/Tests/Unit/Admin/PageAdminTest.php
nnatter: 👍

src/Sulu/Bundle/PageBundle/Tests/Unit/Admin/PageAdminTest.php
```diff
@@ -124,9 +124,37 @@ public function testGetConfigWithVersioning()
...
+                'versioning' => true,
+                'webspaces' => [
+                    'two' => $webspace2,
+                    'one' => $webspace1,
+                ],
+            ],
+            $config
+        );
+
+        $this->assertEquals(
```
nnatter, line 154: the `assertEquals` above will not check the order of the webspaces

danrot (approved):

PR #5534, opened 2020-10-07
Sort templates by their title by adding sort_by option to TypeToolbarAction

### LINKED ISSUE #5288: Sort page templates alphabetically in current language

nnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #5288
| License | MIT

#### What's in this PR?

This PR adds a `sort_by` option to the `TypeToolbarAction`. Furthermore it adjusts the `PageAdmin` and the `SnippetAdmin` to enable this option for page templates and snippet types. 

#### Why?

Because there should be a understandable sorting. See #5288

### DIFF
--- a/src/Sulu/Bundle/PageBundle/Admin/PageAdmin.php
+++ b/src/Sulu/Bundle/PageBundle/Admin/PageAdmin.php
@@ -121,7 +121,7 @@ class PageAdmin extends Admin
             new ToolbarAction(
                 'sulu_admin.type',
                 [
-                    'sort_by_title' => true,
+                    'sort_by' => 'title',
                     'disabled_condition' => '(_permissions && !_permissions.edit)',
                 ]
             ),
--- a/src/Sulu/Bundle/PageBundle/Admin/PageAdmin.php
+++ b/src/Sulu/Bundle/PageBundle/Admin/PageAdmin.php
@@ -121,6 +121,7 @@ class PageAdmin extends Admin
             new ToolbarAction(
                 'sulu_admin.type',
                 [
+                    'sort_by_title' => true,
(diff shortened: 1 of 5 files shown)


src/Sulu/Bundle/PageBundle/Admin/PageAdmin.php
```diff
@@ -121,6 +121,7 @@ public function configureViews(ViewCollection $viewCollection): void
             new ToolbarAction(
                 'sulu_admin.type',
                 [
+                    'sort_by_title' => true,
```
danrot, line 124: Don't you think it would make more sense to name this option `sort_by`? I think it would scale better, because we could use `key` or `title` as a property, and it would still work in the future if we might have other properties to sort by. Could e.g. imagine that there will be a `order` property in the future, so that people can decide completely on their own how they want to order it.

danrot (changes requested):

src/Sulu/Bundle/PageBundle/Admin/PageAdmin.php
nnatter: sounds good - adjusted this 🙂

danrot (approved):

PR #5535, opened 2020-10-08
Add add_button_text param to adjust text of add button to block content-type

nnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT
| Documentation PR | sulu/sulu-docs#567

#### What's in this PR?

This PR adds a `add_button_text ` param to the `block` content-type. The `add_button_text ` param allows to set the text of the add button which is displayed in the administration interface. The param is implemented analogous to the `label` param of the `checkbox` content-type.

#### Why?

The `block` content type is often used in a specific context. For example, in the Sulu demo, the `block` content-type is used for managing tracks of an album. In these cases, it would be nice to set a specific text instead of the generic `Add block` to the add button. 

Screenshot 2020-10-08 at 13 45 29

#### Example Usage

~~~xml
<params>
    <param name="add_button_text">
        <meta>
            <title lang="en">Add track</title>
            <title lang="de">Track hinzufügen</title>
        </meta>
    </param>
</params>
~~~

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/BlockCollection.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/BlockCollection.js
@@ -10,6 +10,7 @@ import blockCollectionStyles from './blockCollection.scss';
 import type {BlockEntry, RenderBlockContentCallback} from './types';
 
 type Props = {|
+    addText?: ?string,
     defaultType: string,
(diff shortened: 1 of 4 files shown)


alexander-schranz:
Would be as specific as possible and use `add_button_text`.

danrot (approved):

PR #5536, opened 2020-10-09
Fix portal route loader

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes -
| Related issues/PRs | -
| License | MIT

#### What's in this PR?

Fix portal route loader by use another regex for empty prefix match.

#### Why?

When merging 2.0 to 2.1 this did fail in symfony 4.4 / 5.0 so we need to adjust it.

#### Example Usage

```php
rm -rf src/Sulu/Bundle/WebsiteBundle/Tests/Application/var/cache/

vendor/bin/phpunit src/Sulu/Bundle/WebsiteBundle/ -c src/Sulu/Bundle/WebsiteBundle/phpunit.xml.dist --filter=ErrorControllerTest::test404ErrorTemplate
```

### DIFF
--- a/src/Sulu/Bundle/WebsiteBundle/Routing/PortalLoader.php
+++ b/src/Sulu/Bundle/WebsiteBundle/Routing/PortalLoader.php
@@ -41,8 +41,8 @@ class PortalLoader extends Loader
 
         $prefixes = [];
         foreach ($this->webspaceManager->getPortalInformations() as $portalInformation) {
-            // symfony does not accept an empty regex as requirement, therefore we use '()' to match an empty prefix
-            $prefixes[] = \preg_quote($portalInformation->getPrefix()) ?: '()';
+            // symfony does not accept an empty regex as requirement, therefore we use '(^$)?' to match an empty prefix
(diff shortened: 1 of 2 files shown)


src/Sulu/Bundle/WebsiteBundle/Routing/PortalLoader.php
```diff
@@ -42,7 +42,7 @@ public function load($resource, $type = null)
         $prefixes = [];
         foreach ($this->webspaceManager->getPortalInformations() as $portalInformation) {
             // symfony does not accept an empty regex as requirement, therefore we use '()' to match an empty prefix
```
nnatter, line 44:
```suggestion
            // symfony does not accept an empty regex as requirement, therefore we use '(^$)?' to match an empty prefix
```

alexander-schranz:
MediaBundle tests are failing because of #5537 but I will rebase this to 2.0 so the regex is the same for all versions.

Issue #5478, opened 2020-09-02
The commands for translation export/import are misleading


thenetexperts:
| Q | A
| --- | ---
| Sulu Version | 2.x
| Browser Version | -

# Actual Behavior

I was under the impression (without looking at the source) that the two commands

`sulu:webspaces:export` and `sulu:webspaces:import`

are actually "backing" up data.

# Expected Behavior

I expected the commands to backup/restore data in regards to websites.

# Steps to Reproduce

-

# Possible Solutions

Maybe a simple adjustment to the command naming would help clarify. Something like `sulu:translation:*`

alexander-schranz:
This also effects the following commands:

 - `sulu:snippet:export`
 - `sulu:article:export`
 - `sulu:webspace:export`
 - `sulu:snippet:import`
 - `sulu:article:import`
 - `sulu:webspace:import`

Maybe we should try to avoid the word `export` and `import`, but not sure what good replacement for them are:

 - `sulu:webspace:translation-create`
 - `sulu:webspace:translation-update`

Maybe adding translation is also enough and we can stay with something like:

 - `sulu:webspace:translation-export`
 - `sulu:webspace:translation-impot`

PR #5373, opened 2020-06-30
Improve download build command with skeleton file downloads and npm build

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

Improve download build command with skeleton file downloads and ask for manual build with cleanup block files.

#### Why?

Most people run into the not matching files as they forget to update the skeleton files and then do a manual npm run build. Then mostly they have a strange package-lock or just forget to update the dependencies.

PR #5543, opened 2020-10-12
Rename export/import commands

TheCadien:
See: https://github.com/sulu/sulu/issues/5478


nnatter:
Thanks for your effort! Unfortunately, we cannot change this in a minor version because it would break the backward compatibility. 

I think we would need to do something similar like @alexander-schranz did in #5373. But as far as I see, there is still an open suggestion for using decoration instead of inheritance.

TheCadien:
@alexander-schranz what do you Think is that something for 3.X Now ?

PR #5551, opened 2020-10-14
Add refresh button to ResourceLocator field-type

### LINKED ISSUE #4221: Regenerate resource locator slug based on new page title

### LINKED ISSUE #5547: Sulu doen't offer the URL prefix when ResourceLocator field is clicked first
| Q | A
| --- | ---
| Sulu Version | 2.1.0
| Browser Version | Chrome 86.0.4240.75

# Actual Behavior

Sulu doen't offer the URL prefix when `ResourceLocator` field is clicked first.

# Expected Behavior

Sulu offer me the correct URL prefix even if I click on `ResourceLocator` field first.

# Steps to Reproduce

When I click (without writing anything) on the `ResourceLocator` field first and then enter the title, Sulu doen't offer me the URL prefix.
image

When I enter the title first and then click on the `ResourceLocator` field, it'll offer me the correct prefix.
image

nnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #4221, fixes #5547
| Related issues/PRs | #5340
| License | MIT

#### What's in this PR?

This PR adds a *Refresh URL* button to the `ResourceLocator` field-type. The button allows to regenerate the URL based on the current values of the Form.

Screenshot 2020-10-14 at 12 09 57

#### Why?

Because it is frequently requested feature 🙂 
See #4221 and #5340.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/ResourceLocator.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/ResourceLocator.js
@@ -1,6 +1,6 @@
 // @flow
 import React from 'react';
-import {action, observable, toJS} from 'mobx';
+import {action, comparer, computed, observable, reaction, toJS} from 'mobx';
 import {observer} from 'mobx-react';
 import ResourceLocatorComponent from '../../../components/ResourceLocator';
 import ResourceLocatorHistory from '../../../containers/ResourceLocatorHistory';
@@ -18,59 +18,12 @@ const HOMEPAGE_RESOURCE_LOCATOR = '/';
 class ResourceLocator extends React.Component<FieldTypeProps<?string>> {
     @observable mode: string;
     @observable enableRefreshButton: boolean = false;
+    @observable wasModifiedManually: boolean = false;
 
-    constructor(props: FieldTypeProps<?string>) {
-        super(props);
+    partsChangeDisposer: () => mixed;
 
+    @computed get parts(): {[string]: mixed} {
         const {
-            dataPath,
-            fieldTypeOptions: {
-                generationUrl,
-                modeResolver,
-            },
-            formInspector,
-            value: initialValue,
-        } = this.props;
-
-        if (!modeResolver) {
-            throw new Error('The "modeResolver" must be a function returning a promise with the desired mode');
-        }
-
-        modeResolver(this.props).then(action((mode) => this.mode = mode));
-
-        if (!generationUrl) {
-            return;
-        }
-
-        if (typeof generationUrl !== 'string') {
-            throw new Error('The "generationUrl" fieldTypeOption must be a string!');
-        }
-
-        if (initialValue === HOMEPAGE_RESOURCE_LOCATOR) {
-            return;
-        }
-
-        formInspector.addFinishFieldHandler(action((finishedFieldDataPath, finishedFieldSchemaPath) => {
-            const {tags: finishedFieldTags} = formInspector.getSchemaEntryByPath(finishedFieldSchemaPath) || {};
-            if (!finishedFieldTags || !finishedFieldTags.some((tag) => tag.name === PART_TAG)) {
-                return;
-            }
-
-            if (initialValue === undefined && !formInspector.isFieldModified(dataPath)) {
-                this.refreshResourceLocator();
-            } else {
-                this.enableRefreshButton = true;
-            }
-        }));
-    }
-
-    refreshResourceLocator = () => {
-        const {
-            onChange,
-            fieldTypeOptions: {
-                generationUrl,
-                resourceStorePropertiesToRequest = {},
-            },
             formInspector,
         } = this.props;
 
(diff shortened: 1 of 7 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/ResourceLocator.js
```diff
@@ -122,31 +133,36 @@ class ResourceLocator extends React.Component<FieldTypeProps<?string>> {
...
-            <div className={resourceLocatorStyles.resourceLocatorContainer}>
-                <div className={resourceLocatorStyles.resourceLocator}>
-                    <ResourceLocatorComponent
-                        disabled={!!disabled}
-                        id={dataPath}
-                        mode={this.mode}
-                        onBlur={this.handleBlur}
-                        onChange={onChange}
-                        value={value}
+            <div>
```
danrot, line 136: Wouldn't a `Fragment` do the trick as well? Or do you need a `div` without a `class` here for some reason?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/ResourceLocator.js
```diff
@@ -122,31 +133,36 @@ class ResourceLocator extends React.Component<FieldTypeProps<?string>> {
...
+                <ResourceLocatorComponent
+                    disabled={!!disabled}
+                    id={dataPath}
+                    mode={this.mode}
+                    onBlur={this.handleBlur}
+                    onChange={onChange}
+                    value={value}
+                />
+                <div className={resourceLocatorStyles.buttonContainer}>
```
danrot, line 145: Wouldn't `buttonsContainer` make more sense?

danrot (changes requested):

danrot:
I have just tested it, and to me it feels a little bit weird from a user perspective...

Let's assume we have just the title and the URL field. The URL field uses the value from the title field to generate the URL. I've realized two weird behaviors:

1. The refresh button won't be activated until I the user blurs the title field.
2. The refresh button also activates, if the user focuses the title field and blurs it without applying any change.

The first one requires an additional click, which should not be necessary IMO, and the second one might be confusing to the user, because nothing is happening after the button click.

That made me wonder: Wouldn't it make sense to set the `enableRefreshButton` when a change is happening? We could do that by adding a `addChangeFieldHandler`. I guess alternatively that could also be done by using a `@computed` value for the url parts and use an `autorun`.

nnatter:
Yes, I think the behaviour is not perfect too. I think adding a `addChangeFieldHandler` would be nice, but I am not sure how to implement it. 

As far as I see, the `dataPath` of the field is passed to the given `FinishFieldHandler`. Also, the the `FormStoreInterface::finishField` method has a `dataPath` parameter. In contrast, the `FormStoreInterface::change` method uses a `name` parameter. Is there a specific reason for this disparity? Or would we implement the `FormStoreInterface::change` method with a `dataPath` too, if we would implement it again? WhatShould the `ChangeFieldHandler` receive a `name` or a `dataPath`? 🤔 

Furthermore, the `Form` container calls the `FormInspector::finishField` method. The `FormInspector` then calls the `FormStoreInterface::finishField` method. Should we implement a similar `FormInspector::changeField` method? That feels kinda wrong to me as "`Inspector`" sounds like it is a read-only thing, doesnt it? Maybe I am just overthinking this? 😅 

One solution for the second problem would be adding a `FormStoreInterface::notifyFieldFinish` and a `FormStoreInterface::notifyFieldChange` method and deprecating the `FormStoreInterface::finishField` method.

nnatter:
This PR depends on #5558

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/ResourceLocator.js
nnatter: 👍

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/ResourceLocator.js
nnatter: 👍

nnatter:
I do not understand why `circleci` is failing, I have not adjusted this files as far as I see 😕

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/ResourceLocator.js
```diff
@@ -45,68 +91,83 @@ class ResourceLocator extends React.Component<FieldTypeProps<?string>> {
...
             }
+        }));
+    }
 
-            const {tags: finishedFieldTags} = formInspector.getSchemaEntryByPath(finishedFieldSchemaPath) || {};
-            if (!finishedFieldTags || !finishedFieldTags.some((tag) => tag.name === PART_TAG)) {
-                return;
-            }
+    componentWillUnmount() {
+        this.partsChangeDisposer();
```
alexander-schranz, line 115: Could be not be set.

alexander-schranz (changes requested):
A error happens on the homepage when the `partsChangeDisposer` is not set.

danrot:
> I do not understand why `circleci` is failing, I have not adjusted this files as far as I see

Probably related to https://github.com/sulu/sulu/pull/5569, but there is also a jest fix on its way. Not sure if we should merge #5569 or wait for the fix.

alexander-schranz (approved):
Works now as expected. Thx for adding the unmount test 👍

alexander-schranz (approved):

PR #5552, opened 2020-10-14
Remove jsx-indent-props rule because indent rule covers it already

danrot:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

It seems like there was a change in the `jsx-indent-props` rule. I would argue that we don't want that change, and even if I apply that change, `eslint` still fails, because it seems that the standard `indent` rule of eslint already checks JSX by itself. So I decided to remove two of the react rules from eslint, because invalid formatted JSX still fails.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -174,9 +174,14 @@
             }
         ],
         "react/jsx-key": "error",
-        "react/jsx-indent": "error",
+        "react/jsx-indent": [
+            "error",
+            4,
+            {
+                "indentLogicalExpressions": true
+            }
+        ],
         "react/jsx-no-bind": "error",
-        "react/jsx-indent-props": "error",
         "react/jsx-equals-spacing": "error",
         "react/jsx-handler-names": "error",
         "react/jsx-first-prop-new-line": [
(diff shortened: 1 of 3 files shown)


alexander-schranz (approved):

.eslintrc.json
```diff
@@ -174,9 +174,7 @@
             }
         ],
         "react/jsx-key": "error",
-        "react/jsx-indent": "error",
```
nnatter, line 177: as far as i tested it, this rule should not make problem at the moment, right?

would propose the following config:
```
        "react/jsx-indent": [
            "error",
            4,
            {
                "indentLogicalExpressions": true
            }
        ],
```

nnatter (commented):
With this change, the following is considered as valid:
```
                {formInspector.id &&
            <div className={customUrlStyles.resourceLocatorHistory}>
                <ResourceLocatorHistory
                    id={formInspector.id}
                    options={{webspace: formInspector.options.webspace}}
                    resourceKey="custom_url_routes"
                />
            </div>
                }
```

.eslintrc.json
danrot: No, that one didn't make any problems, but I didn't find a case where it made a difference, that's why I removed it. But it seems like you did :slightly_smiling_face: Testing what you are suggesting, and will add it afterwards.

danrot: Just added it, the additional configuration is superb :slightly_smiling_face:

nnatter (approved):

PR #5553, opened 2020-10-14
Display error bar when media upload fails in MediaOverview

### LINKED ISSUE #5540: Error handling frontend of media library
# Problem description

I’ve extended the `MediaManager.php `. When i throw a:
```
throw new MediaException('Blocked file type', MediaException::EXCEPTION_CODE_BLOCKED_FILE_TYPE);
```
In the `save()` function, no error is triggered in the CMS.

It just glitches and the user has no clue what happened.

# Proposed solution

Ability to display a red error bar in the media library after throwing an error from a controller.
Underneath answer by @nnatter 

> it looks like there is no error handling at all implemented in the frontend at the moment. at least, we need to adjust the MediaUploadStore, the MultiMediaDropzone, the MediaCollection and the MediaOverview

nnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #5540
| License | MIT

#### What's in this PR?

This PR adjusts the `MediaOverview` view to display an error bar when the upload of a file fails. To do this, the `MediaUploadStore`, `MultiMediaDropzone` and `MediaCollection` were adjusted to support the handling of upload errors.

#### Why?

Because there is no feedback to the user if an error happens at the moment. See #5540.

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/Resources/js/containers/MediaCollection/MediaCollection.js
+++ b/src/Sulu/Bundle/MediaBundle/Resources/js/containers/MediaCollection/MediaCollection.js
@@ -21,7 +21,7 @@ type Props = {|
     mediaListStore: ListStore,
     onCollectionNavigate: (collectionId: ?string | number) => void,
     onMediaNavigate?: (mediaId: string | number) => void,
-    onUploadError?: (errorResponse: string) => void,
+    onUploadError?: (errorResponses: Array<string>) => void,
     onUploadOverlayClose: () => void,
     onUploadOverlayOpen: () => void,
     overlayType: OverlayType,
@@ -63,11 +63,11 @@ class MediaCollection extends React.Component<Props> {
         );
     };
 
-    @action handleUploadError = (errorResponse: string) => {
+    @action handleUploadError = (errorResponses: Array<string>) => {
         const {mediaListStore, onUploadError} = this.props;
 
         if (onUploadError) {
-            onUploadError(errorResponse);
+            onUploadError(errorResponses);
         }
 
         mediaListStore.reset();
--- a/src/Sulu/Bundle/MediaBundle/Resources/js/containers/MediaCollection/MediaCollection.js
+++ b/src/Sulu/Bundle/MediaBundle/Resources/js/containers/MediaCollection/MediaCollection.js
@@ -21,6 +21,7 @@ type Props = {|
     mediaListStore: ListStore,
     onCollectionNavigate: (collectionId: ?string | number) => void,
     onMediaNavigate?: (mediaId: string | number) => void,
+    onUploadError: (error: any) => void,
     onUploadOverlayClose: () => void,
     onUploadOverlayOpen: () => void,
     overlayType: OverlayType,
@@ -62,6 +63,15 @@ class MediaCollection extends React.Component<Props> {
         );
     };
 
+    @action handleUploadError = (error: any) => {
+        const {mediaListStore, onUploadError} = this.props;
+
+        onUploadError(error);
+
+        mediaListStore.reset();
+        mediaListStore.reload();
+    };
+
     render() {
         const {
             collectionListStore,
@@ -92,6 +102,7 @@ class MediaCollection extends React.Component<Props> {
                 onClose={onUploadOverlayClose}
                 onOpen={onUploadOverlayOpen}
                 onUpload={this.handleUpload}
+                onUploadError={this.handleUploadError}
                 open={uploadOverlayOpen}
             >
                 <CollectionSection
--- a/src/Sulu/Bundle/MediaBundle/Resources/js/containers/MediaCollection/MediaCollection.js
+++ b/src/Sulu/Bundle/MediaBundle/Resources/js/containers/MediaCollection/MediaCollection.js
@@ -21,7 +21,7 @@ type Props = {|
     mediaListStore: ListStore,
     onCollectionNavigate: (collectionId: ?string | number) => void,
     onMediaNavigate?: (mediaId: string | number) => void,
-    onUploadError: (error: any) => void,
+    onUploadError?: (error: any) => void,
     onUploadOverlayClose: () => void,
     onUploadOverlayOpen: () => void,
     overlayType: OverlayType,
@@ -66,7 +66,9 @@ class MediaCollection extends React.Component<Props> {
     @action handleUploadError = (error: any) => {
         const {mediaListStore, onUploadError} = this.props;
 
-        onUploadError(error);
+        if (onUploadError) {
+            onUploadError(error);
+        }
 
         mediaListStore.reset();
         mediaListStore.reload();
--- a/src/Sulu/Bundle/MediaBundle/Resources/js/containers/MediaCollection/MediaCollection.js
+++ b/src/Sulu/Bundle/MediaBundle/Resources/js/containers/MediaCollection/MediaCollection.js
@@ -21,7 +21,7 @@ type Props = {|
     mediaListStore: ListStore,
     onCollectionNavigate: (collectionId: ?string | number) => void,
     onMediaNavigate?: (mediaId: string | number) => void,
-    onUploadError?: (error: any) => void,
+    onUploadError?: (errorResponse: string) => void,
     onUploadOverlayClose: () => void,
     onUploadOverlayOpen: () => void,
     overlayType: OverlayType,
@@ -63,11 +63,11 @@ class MediaCollection extends React.Component<Props> {
         );
     };
 
-    @action handleUploadError = (error: any) => {
+    @action handleUploadError = (errorResponse: string) => {
         const {mediaListStore, onUploadError} = this.props;
 
         if (onUploadError) {
-            onUploadError(error);
+            onUploadError(errorResponse);
         }
 
         mediaListStore.reset();
--- a/src/Sulu/Bundle/MediaBundle/Resources/js/containers/MultiMediaDropzone/MultiMediaDropzone.js
+++ b/src/Sulu/Bundle/MediaBundle/Resources/js/containers/MultiMediaDropzone/MultiMediaDropzone.js
@@ -17,6 +17,7 @@ type Props = {
     onClose: () => void,
     onOpen: () => void,
     onUpload: (media: Array<Object>) => void,
+    onUploadError: (error: any) => void,
     open: boolean,
 };
 
@@ -64,6 +65,9 @@ class MultiMediaDropzone extends React.Component<Props> {
         const {
             locale,
             collectionId,
+            onClose,
+            onUpload,
+            onUploadError,
         } = this.props;
         const uploadPromises = [];
 
@@ -80,11 +84,17 @@ class MultiMediaDropzone extends React.Component<Props> {
         });
 
         return Promise.all(uploadPromises).then((...media) => {
-            this.props.onUpload(...media);
+            onUpload(...media);
 
             setTimeout(() => {
-                this.props.onClose();
+                onClose();
                 this.destroyMediaUploadStores();
+            }, 1000);
+        }).catch((e) => {
+            setTimeout(() => {
+                onClose();
+                this.destroyMediaUploadStores();
+                onUploadError(e);
             }, 1000);
         });
     };
--- a/src/Sulu/Bundle/MediaBundle/Resources/js/containers/MultiMediaDropzone/MultiMediaDropzone.js
+++ b/src/Sulu/Bundle/MediaBundle/Resources/js/containers/MultiMediaDropzone/MultiMediaDropzone.js
@@ -17,7 +17,7 @@ type Props = {
     onClose: () => void,
     onOpen: () => void,
     onUpload: (media: Array<Object>) => void,
-    onUploadError: (errorResponse: string) => void,
+    onUploadError: (errorResponses: Array<string>) => void,
     open: boolean,
 };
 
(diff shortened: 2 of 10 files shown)


danrot (changes requested):

src/Sulu/Bundle/MediaBundle/Resources/js/stores/MediaUploadStore/MediaUploadStore.js
```diff
@@ -173,7 +173,13 @@ export default class MediaUploadStore {
 
             xhr.open('POST', url);
 
-            xhr.onload = (event: any) => resolve(JSON.parse(event.target.response));
+            xhr.onload = (event: any) => {
+                if (event.target.status >= 200 && event.target.status <= 299) {
```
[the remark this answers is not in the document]
nnatter: unfortunately not: https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest
but this behaviour matches the behaviour of the `Response` object that is returned when using `fetch`: https://developer.mozilla.org/en-US/docs/Web/API/Response/ok

src/Sulu/Bundle/MediaBundle/Resources/translations/admin.de.json
```diff
@@ -43,6 +43,7 @@
     "sulu_media.formats": "Formate",
     "sulu_media.all_collections": "Alle Kollektionen",
     "sulu_media.move_media": "Medien verschieben",
+    "sulu_admin.upload_server_error": "Beim Upload einer Datei ist ein Fehler aufgetreten.",
```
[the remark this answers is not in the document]
nnatter: the error message shows up a single time because the `MultiMediaDropzone` catches the error via `Promise.all(uploadPromises)`

src/Sulu/Bundle/MediaBundle/Resources/js/containers/MediaCollection/MediaCollection.js
```diff
@@ -62,6 +63,17 @@ class MediaCollection extends React.Component<Props> {
...
 
+    @action handleUploadError = (error: any) => {
+        const {mediaListStore, onUploadError} = this.props;
+
+        if (onUploadError) {
+            onUploadError(error);
+        }
+
+        mediaListStore.reset();
+        mediaListStore.reload();
```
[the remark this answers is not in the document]
nnatter: i thought that this is useful because some images might have been uploaded successfully. but dont have a strong opinion on the behaviour

src/Sulu/Bundle/MediaBundle/Resources/js/containers/MultiMediaDropzone/MultiMediaDropzone.js
```diff
@@ -80,12 +84,18 @@ class MultiMediaDropzone extends React.Component<Props> {
...
+            onUpload(...media);
 
             setTimeout(() => {
-                this.props.onClose();
+                onClose();
                 this.destroyMediaUploadStores();
             }, 1000);
+        }).catch((e) => {
+            setTimeout(() => {
+                onClose();
```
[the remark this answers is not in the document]
nnatter: when not closing the popover, the "upload items" will not be removed therefore still be present if the popover is closed and opened a second time. but you are right about the error message too, it is not very visible while the popover is openend

src/Sulu/Bundle/MediaBundle/Resources/js/containers/MediaCollection/MediaCollection.js
```diff
@@ -21,6 +21,7 @@ type Props = {|
     mediaListStore: ListStore,
     onCollectionNavigate: (collectionId: ?string | number) => void,
     onMediaNavigate?: (mediaId: string | number) => void,
+    onUploadError?: (error: any) => void,
```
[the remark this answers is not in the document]
nnatter: Changed this to string 🙂 

We are passing `event.target.response` to the callback in the `MediaUploadStore`: https://github.com/sulu/sulu/blob/master/src/Sulu/Bundle/MediaBundle/Resources/js/stores/MediaUploadStore/MediaUploadStore.js#L181

Looks like this is a string, if we do not set any `responseType`: https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/response

nnatter:
I know the UX of this is not perfect, but I think it is already **a lot better** than just dont inform the user about the error at all like it is done now.
If somebody has a better idea in terms of UX, I am obviously happy to have a look at it 🙂  @sulu/core-team

src/Sulu/Bundle/MediaBundle/Resources/js/containers/MediaCollection/MediaCollection.js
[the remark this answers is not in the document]
danrot: Okay, and that's the same behavior as when all images are successfully uploaded, right? Then it makes sense to me :+1:

src/Sulu/Bundle/MediaBundle/Resources/js/containers/MultiMediaDropzone/MultiMediaDropzone.js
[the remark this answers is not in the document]
danrot: I would say if it would just have been about the uploaded items it would have to be solved in a different way, but since the error message should be visible it's probably better this way anyway.

src/Sulu/Bundle/MediaBundle/Resources/js/stores/MediaUploadStore/MediaUploadStore.js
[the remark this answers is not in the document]
danrot: Okay, it's fine then.

src/Sulu/Bundle/MediaBundle/Resources/translations/admin.de.json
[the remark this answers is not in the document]
danrot: Should we maybe phrase it a little bit different then? Maybe "Beim Upload mancher Dateien ist ein Fehler augetreten."? Or maybe even better: "Nicht alle Dateien wurden erfolgreich hochgeladen.".

Would be a tiny change, but explains the error a bit better IMO.

src/Sulu/Bundle/MediaBundle/Resources/js/containers/MultiMediaDropzone/tests/MultiMediaDropzone.test.js
```diff
@@ -20,9 +20,15 @@ jest.mock('sulu-admin-bundle/utils', () => ({
 }));
 
 jest.mock('../../../stores/MediaUploadStore', () => jest.fn(function() {
-    this.create = jest.fn().mockReturnValue(Promise.resolve({
-        id: 123,
-    }));
+    this.create = jest.fn((_, file) => {
```
danrot, line 23: Can't we put that into the test we do this for? This way it is hard to tell which tests need that, and it gets messier the bigger this file gets.

src/Sulu/Bundle/MediaBundle/Resources/js/stores/MediaUploadStore/tests/MediaUploadStore.test.js
```diff
@@ -51,6 +51,78 @@ test('Calling the "update" method should make a "POST" request to the media upda
...
+    const fileData = new File([''], 'fileName');
+
+    const updatePromise = mediaUploadStore.update(fileData);
+
+    window.XMLHttpRequest.mock.instances[0].onload({target: {status: 400, response: 'invalid-format'}});
+
+    return expect(updatePromise).rejects.toEqual('invalid-format');
+});
+
+test('Promise returned by "update" method should be resolved if request is not successful', () => {
```
danrot, line 102: "should be rejected", right?

danrot (changes requested):

src/Sulu/Bundle/MediaBundle/Resources/js/containers/MultiMediaDropzone/tests/MultiMediaDropzone.test.js
nnatter: I am not sure how to do that unfortunately 😕 
The store is created inside of the `handleDrop` method of the component and the `MediaUploadStore.create` method is called in the next statement. Is there a way to mock this inside of the testcase?

https://github.com/sulu/sulu/blob/master/src/Sulu/Bundle/MediaBundle/Resources/js/containers/MultiMediaDropzone/MultiMediaDropzone.js#L75-L76

src/Sulu/Bundle/MediaBundle/Resources/js/stores/MediaUploadStore/tests/MediaUploadStore.test.js
nnatter: yes 👍

src/Sulu/Bundle/MediaBundle/Resources/translations/admin.en.json
```diff
@@ -43,7 +43,7 @@
     "sulu_media.formats": "Formats",
     "sulu_media.all_collections": "All collections",
     "sulu_media.move_media": "Move media",
-    "sulu_admin.upload_server_error": "There was an error when uploading a file to the server.",
+    "sulu_media.upload_server_error": "There was an error while uploading one or multiple file to the server.",
```
danrot, line 46: "one or multiple files"

src/Sulu/Bundle/MediaBundle/Resources/js/containers/MultiMediaDropzone/tests/MultiMediaDropzone.test.js
danrot: The `MediaUploadStore` has already been imported above, so you should be able to change the return value of the mocked function in the test itself somehow like this (not tested):

```javascript
MediaUploadStore.mock.instances[index].create.mockReturnValue(...);
```

danrot (changes requested):

src/Sulu/Bundle/MediaBundle/Resources/js/containers/MultiMediaDropzone/tests/MultiMediaDropzone.test.js
nnatter: unfortunately calling `console.log(MediaUploadStore.mock.instances);` before the `onDrop` call logs an empty array (`[]`) 🙈

danrot: Then the only other way I see around this, is to replace the entire constructor mock... But yeah, that might be a bit cumbersome too... Let's merge it like this, and hope it never makes any problems :see_no_evil:

src/Sulu/Bundle/MediaBundle/Resources/js/containers/MultiMediaDropzone/MultiMediaDropzone.js
```diff
@@ -80,12 +84,18 @@ class MultiMediaDropzone extends React.Component<Props> {
         });
 
         return Promise.all(uploadPromises).then((...media) => {
```
danrot, line 86: This seems to behave a a bit strange, because this promise already seems to be resolved when the first request fails... I guess that could lead to some inconsistent state after reloading the view, because some uploads might still finish afterwards. Shouldn't we maybe use something like [`Promise.allSettled`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled) instead?

danrot (changes requested):

src/Sulu/Bundle/MediaBundle/Resources/js/containers/MultiMediaDropzone/MultiMediaDropzone.js
nnatter: yes, that would be better 🙂 
what should we pass to the `onUploadError` in that case? the first error or all errors?

danrot: All errors would probably make more sense, wouldn't it? But I guess it doesn't make a huge difference anyway, since the handling will not have a look at every single error, right?

nnatter: no in this case it will not, but maybe we will reuse the component somewhere and i am not sure what i would expect then 🙂
another possibility would be calling the callback for each error that happens? 🤔

danrot: Personally I would have expected all the errors at once... But that's subjective I guess. The advantage of that would be that it would be easier to handle in the calling component. If the callback is called every single time you would probably have to keep an array of values in the calling component, making that one more complicated than it needs to be.

nnatter: alright, going to use an array of errors then 🙂 just saw that this is also consistent with the argument `onUpload` callback

danrot (approved):
Will merge as soon as all the builds are green.

PR #5555, opened 2020-10-16
AdminBundle: Add ColorFieldTransformer

FarouHub:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Add a ColorFieldTransformer form the AdminBundle.

With this transformer we will have the ability to view/show the color choose with the "Color content type" into the admin table list.

#### Why?

Many users are asking to this option.

#### Example Usage

~~~xml
<!-- New type "color" -->
<property name="color" visibility="always" searchability="yes" translation="sulu_admin.color" type="color">
    <field-name>color</field-name>
    <entity-name>App\Entity\Opening</entity-name>
</property>
~~~


src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/fieldTransformers/ColorFieldTransformer.js
```diff
@@ -0,0 +1,10 @@
+// @flow
+import React from 'react';
+import type {Node} from 'react';
+import type {FieldTransformer} from '../types';
+
+export default class ColorFieldTransformer implements FieldTransformer {
+    transform(value: *): Node {
+        return <div style={{'backgroundColor': value, 'width': '40px', 'height': '20px'}}></div>;
```
alexander-schranz, line 8: Would add a `className` and create a `.scss` which is imported, which set the `width` and `height` attribute.

alexander-schranz (changes requested):
@FarouHub Thank you for this PR. We should also add a test for this new transformer see for example [IconTransformer.test.js](https://github.com/sulu/sulu/blob/0753565e3d7bd205adcfabf4a1f4f90a71c7d8a3/src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/tests/fieldTransformers/IconFieldTransformer.test.js#L65-L69)

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/fieldTransformers/ColorFieldTransformer.js
alexander-schranz: /cc @chirimoya any preferences how the color should be displayed?

FarouHub: My first commit was to fast and to light :)
Here come a new commit with your requirements and more control in the transformer.
First time i handle all this tools : was very cool 👍

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/fieldTransformers/ColorFieldTransformer.js
```diff
@@ -0,0 +1,21 @@
+// @flow
+import React from 'react';
+import log from 'loglevel';
+import type {Node} from 'react';
+import type {FieldTransformer} from '../types';
+import colorFieldTransformerStyles from './colorFieldTransformer.scss';
+
+export default class ColorFieldTransformer implements FieldTransformer {
+    transform(value: *): Node {
+        //test si contain # et 3 ou 6 charateres
```
nnatter, line 10: would remove the comment as the regex below is quite ease to understand imo. also, personally i dont understand french 😇 

```suggestion
```

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/fieldTransformers/ColorFieldTransformer.js
```diff
@@ -0,0 +1,21 @@
...
+        if (!/^#[0-9a-f]{3}([0-9a-f]{3})?$/i.test(value)) {
+            log.error(`Transformer parameter "${value}" needs to be of type hexadecimal color.`);
+            return value;
```
nnatter, line 13: would return `null` here to be consistent with the `NumberFieldTransformer` and `DateFieldTransformer`
```suggestion

            return null;
```

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/fieldTransformers/ColorFieldTransformer.js
```diff
@@ -0,0 +1,21 @@
+// @flow
+import React from 'react';
+import log from 'loglevel';
+import type {Node} from 'react';
+import type {FieldTransformer} from '../types';
+import colorFieldTransformerStyles from './colorFieldTransformer.scss';
+
+export default class ColorFieldTransformer implements FieldTransformer {
+    transform(value: *): Node {
```
nnatter, line 9: i think we dont want to log an error if the value is not set 🙂 

```suggestion
    transform(value: *): Node {
        if (!value) {
            return null;
        }
```

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/fieldTransformers/colorFieldTransformer.scss
```diff
@@ -0,0 +1,4 @@
+.color-box {
+    width: 40px;
+    height: 20px;
```
nnatter, line 3: would propose the following style that matches the dimensions of the checkbox:

```suggestion
    width: 16px;
    height: 16px;
    border-radius: 2px;
```

nnatter (commented):
Looks good to me otherwise - thanks a lot for your effort! 🙂

nnatter (approved):
I adjusted the error message a but, hope that is okay for you. Thanks for your effort! 🙂

nnatter:
The CircleCI error is not related to this PR (see #5569)

PR #5556, opened 2020-10-19
Lazy loaded Twig extensions

plozmun:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | ?
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

There are some references that indicate that using runtime for twig extensions improves performance. Especially for the less used ones. 

I will try well with blackfire if the improvement is really good

References:
https://symfony.com/doc/current/templating/twig_extension.html#creating-lazy-loaded-twig-extensions
https://github.com/liip/LiipImagineBundle/issues/1261

#### Why?

Because we want to improve performance in some cases.

#### BC Breaks/Deprecations

Maybe is someone extend the extension. I think it's better to make runtime final.

#### To Do

- [ ] Change all extensions that can be improved
- [ ] Make tests

### DIFF
--- a/src/Sulu/Bundle/CategoryBundle/Resources/config/services.xml
+++ b/src/Sulu/Bundle/CategoryBundle/Resources/config/services.xml
@@ -42,12 +42,16 @@
         </service>
 
         <service id="sulu_category.twig_extension" class="Sulu\Bundle\CategoryBundle\Twig\CategoryTwigExtension">
+            <tag name="twig.extension"/>
+        </service>
+
+        <service id="sulu_category.twig_runtime" class="Sulu\Bundle\CategoryBundle\Twig\CategoryRuntime">
             <argument type="service" id="sulu_category.category_manager"/>
             <argument type="service" id="sulu_category.category_request_handler"/>
             <argument type="service" id="sulu_core.array_serializer"/>
             <argument type="service" id="sulu_core.cache.memoize"/>
 
-            <tag name="twig.extension"/>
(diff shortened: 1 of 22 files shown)


alexander-schranz:
I sadly must say at current state I don't think that we can change this. As this implements a lot of bc breaks and would crash any project which did override this twig extension services. I think it would be better to analyze which dependency make it slow and maybe just define that service as `lazy`.
But I definitely think we should have a look at twig runtime for the next major. So it would be great if you have tested it with blackfire if you could make an issue with your profile changes.

plozmun:
True , it's a BC break clearly but i will test it well to check if performance can be improved in a mayor version. I suppose that the mayor improvement will be in custom controllers and maybe not a big change in sulu pages

plozmun:
Hi! I have added more twig runtime clases but there is no great improvement with these changes. There is a bit in admin section but maybe not too much. 

Close this PR to not make noise if you want 

Captura de pantalla 2020-10-25 a las 12 16 13
Captura de pantalla 2020-10-25 a las 12 19 14

alexander-schranz:
I close this PR as written above this is currently a to big BC Break. I still think its very interesting and something we could have a look at it in `3.0`, just I think that it maybe something like lazy services could be also a good way to avoid the overhead of initializing all injected services in twig. Thank you for this doing this benchmarks maybe we can move this into an issue? Do you want to create it?

Issue #5548, opened 2020-10-13
Translated pages do not use translated parent pages in path prefix


thenetexperts:
| Q | A
| --- | ---
| Sulu Version | 2.1.2
| Browser Version | Firefox

# Actual Behavior

The Resourcelocator of the newly created child page is using either the path prefix of the (copied from) page or removes the path prefix completely.

# Expected Behavior

The Resourcelocator should reflect the path prefix from the parent in the same language

# Steps to Reproduce

1. Create a new page in the default language (DE) named "Deutsche Seite", save.
2. Switch to english language (EN) and allow copy of the (DE) locale
3. Rename this english page to "English Page", save.
4. Switch back to default language (DE)
5. Create a new child page below "Deutsche Seite" and name it "Deutsche Unterseite" (The Resourcelocator will be correct)
6. Switch to english language (EN) and allow copy of the (DE) locale
7. Rename the english page to "English Subpage".

You will now see the ResourceLocator will keep the name "deutsche-seite". If you click in the Article Textfield the prefix might even vanish, leaving the english child page with no path prefix (of the parent).

nnatter:
This should hopefully be fixed by #5558 and #5573 🙂

PR #5558, opened 2020-10-19
Include parentId in request of resource_locator field-type after copying language

nnatter:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Related issues/PRs | #5548
| License | MIT

#### What's in this PR?

This PR adjusts the `resource_locator` field-type to include the `parentId` in the request for generating the URL on the edit form. This is needed for including the parent in the generated URL when editing an **existing page in a new language**.

This fixes one aspect of #5548

#### Why?

On the **add form**, the `parentId` is included via the `FormInspector::options` object. The `parentId` is set to the `FormInspector::options` object because of the `addRouterAttributesToFormRequest` call in the `PageAdmin`.

This strategy does not work on the **edit form**, because the `routerAttributes` do not contain the `parentId` in that case.

#### Reproducing the Problem

1. Save and publish a page `England` in the `en` locale
2. Switch to the `de` locale
3. Save and publish the page as `Deutschland`
3. Add a child page `London` in the `en` locale
4. Switch to the `de` locale
5. Click a random button on the copy dialog (the answer does not matter)
6. Adjust the title of the page to `Berlin` and click outside of the field
7. **The generated URL does not include the expected `/deutschland` prefix**

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/ResourceLocator.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/ResourceLocator.js
@@ -81,8 +81,8 @@ class ResourceLocator extends React.Component<FieldTypeProps<?string>> {
 
             const requestOptions = {...formInspector.options};
 
-            Object.entries(resourceStorePropertiesToRequest).forEach(([parameterName, propertyName]) => {
-                const propertyValue = toJS(formInspector.getValueByPath('/' + String(propertyName)));
+            Object.entries(resourceStorePropertiesToRequest).forEach(([propertyName, parameterName]) => {
+                const propertyValue = toJS(formInspector.getValueByPath('/' + propertyName));
                 if (propertyValue !== undefined) {
                     requestOptions[parameterName] = propertyValue;
                 }
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/ResourceLocator.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/ResourceLocator.js
@@ -1,6 +1,6 @@
 // @flow
 import React from 'react';
-import {action, observable} from 'mobx';
+import {action, observable, toJS} from 'mobx';
 import {observer} from 'mobx-react';
 import ResourceLocatorComponent from '../../../components/ResourceLocator';
 import ResourceLocatorHistory from '../../../containers/ResourceLocatorHistory';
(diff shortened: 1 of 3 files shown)


thenetexperts:
(this PR and #5559 solved the issue for me - thank you!)

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/tests/fields/ResourceLocator.test.js
```diff
@@ -378,6 +378,9 @@ test('Should request a new URL including the options from the ResourceFormStore
                 generationUrl: '/admin/api/resourcelocators?action=generate',
                 historyResourceKey: 'page_resourcelocators',
                 modeResolver: () => Promise.resolve('leaf'),
+                resourceStorePropertiesToRequest: {
+                    requestParamKey: 'propertyName',
```
danrot, line 382: Shouldn't that be implemented in the other way? The property is called `resourceStorePropertiesToRequest`, so the left side here should be the name of the property from the `ResourceStore`, and right side should be the name that goes into the request.

I've done a quick check, and that's also how it works e.g. for the `routerAttributesToFormMetadata` option in the `FormViewBuilder`.

danrot (changes requested):

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/ResourceLocator.js
```diff
@@ -19,8 +19,17 @@ class ResourceLocator extends React.Component<FieldTypeProps<?string>> {
...
         super(props);
 
-        const {dataPath, onChange, fieldTypeOptions, formInspector, value} = this.props;
-        const {generationUrl, modeResolver} = fieldTypeOptions;
+        const {
+            dataPath,
+            fieldTypeOptions: {
+                generationUrl,
+                modeResolver,
+                resourceStorePropertiesToRequest = {},
```
danrot, line 27: Oh, and wouldn't it make more sene to call this `resourceStorePropertiesToGenerationRequest`? Then it would be clearer when this is used.

danrot (changes requested):

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/ResourceLocator.js
nnatter: I chose this name because it is already used in the `SingleSelection` and `Selection` field-type. but dont have a strong opinion

danrot: Ok, I don't mind too much either... Was just a suggestion.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/tests/fields/ResourceLocator.test.js
nnatter: Thats okay for me too 🙂 

Just for completion, the `selection` and `single_selection` field-type allow for the following configuration:
```
    <params>
        <param name="resource_store_properties_to_request" type="collection">
            <param name="parameterKey" value="testParameter" />
        </param>
    </params>
```

danrot: Yeah, it's structured differently there... But when it is defined in other places the name seems mostly to read from left to right, and I think that's also what we agreed on once :slightly_smiling_face:

danrot (approved):

PR #5559, opened 2020-10-19
Load content for new locale when copying content from another locale

### LINKED ISSUE #5548: Translated pages do not use translated parent pages in path prefix

nnatter:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #5548
| Related issues/PRs | #5558
| License | MIT

#### What's in this PR?

This PR adjusts the `ResourceStore` to load and return the content of the destination locale. 

#### Why?

Because the `copy-locale` API returns the content of the source locale. At the moment, when copying the data of an existing source locale, the data of the source locale is set to the form. This causes problems because the data of the source locale contains the **wrong resource locator (see #5548) and the wrong `publishedState`**.

#### Reproducing the Problem

1. Save and publish a page `England` in the `en` locale
2. Switch to the `de` locale
3. Save and publish the page as `Deutschland`
3. Add a child page `London` in the `en` locale
4. Switch to the `de` locale
5. Confirm the copy dialog
6. **The form displays the wrong resource locator and wrongly indicates that the locale is published**

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/stores/ResourceStore/ResourceStore.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/stores/ResourceStore/ResourceStore.js
@@ -232,13 +232,22 @@ export default class ResourceStore {
             throw new Error('Copying from another locale does only work for objects with locales!');
         }
 
+        const destinationLocale = this.locale.get();
+
         return ResourceRequester
             .post(
                 this.resourceKey,
                 {},
-                {...options, action: 'copy-locale', dest: this.locale.get(), id: this.id, locale}
-            ).then(action((response) => {
+                {...options, action: 'copy-locale', dest: destinationLocale, id: this.id, locale}
+            ).then(action(() => {
+                // api will return data of there source locale, therefore we need to load the data of the new locale
+                return ResourceRequester.get(
+                    this.resourceKey,
+                    {...options, ...this.loadOptions, locale: destinationLocale, id: this.id}
+                );
+            })).then(action((response) => {
                 this.setMultiple(response);
+
                 return response;
             }));
     }
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/stores/ResourceStore/tests/ResourceStore.test.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/stores/ResourceStore/tests/ResourceStore.test.js
@@ -514,22 +514,29 @@ test('Saving should consider the passed idQueryParameter flag and reset it after
 
 test('Copy the content from different locale', () => {
     const resourceStore = new ResourceStore('pages', 4, {locale: observable.box('en')}, {webspace: 'sulu_io'});
-    resourceStore.set('content', 'old content');
-    expect(resourceStore.data).toEqual({content: 'old content'});
-    const oldData = resourceStore.data;
+    resourceStore.set('content', 'initial content');
+    expect(resourceStore.data).toEqual({content: 'initial content'});
+    const resourceStoreDataObject = resourceStore.data;
 
-    const germanContent = {id: 3, content: 'new content'};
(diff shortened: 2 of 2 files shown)


thenetexperts:
(this PR and #5558 solved the issue for me - thank you!)

src/Sulu/Bundle/AdminBundle/Resources/js/stores/ResourceStore/ResourceStore.js
```diff
@@ -232,13 +232,22 @@ export default class ResourceStore {
...
         return ResourceRequester
             .post(
                 this.resourceKey,
                 {},
-                {...options, action: 'copy-locale', dest: this.locale.get(), id: this.id, locale}
-            ).then(action((response) => {
+                {...options, action: 'copy-locale', dest: destinationLocale, id: this.id, locale}
+            ).then(action(() => {
+                // api will return data of there source locale, therefore we need to load the data of the new locale
+                return ResourceRequester.get(
```
danrot, line 244: Making another request feels like some overhead here... Can't we fix that in another way? I know that the API returns the source locale, because when you copy the locale with the edit menu from the toolbar you can also copy it to multiple localizations, and then it would not be clear which one to return...

The reason I am worrying about this, is that sending too many requests was probably one of the main reasons the old administration felt much slower...

What if we change the API for the `copy-locale` action in a way, that let's the caller decide in which locale the content is returned? The current request looks like this, and it would probably be wise to keep this one working in the same way:

```
/admin/api/pages/d3354303-a11b-443f-9cb7-cace67a4e57c?webspace=sulu-test&action=copy-locale&dest=de&locale=en
```

In this example the english version would be returned from the API, although the german version was the target. This is driven by the `locale` query parameter. So we should probably keep the behavior, that the `locale` defines which locale is returned in the response.

An idea that just came to my mind would be that we introduce another query parameter, which is called `src` (opposed to `dest`). This query parameter could then be used to define from where the content is copied, and the `dest` parameter can be used to determine the target of the copy process. The the `locale` would be free to decide in which locale the content will be returned.

In order to keep BC, we can say that the `src` parameter is optional, and the value always falls back to the `locale` value. So the URL from above would still work, but at the same time something like this would be possible, that would save us the request created in this PR:

```
/admin/api/pages/d3354303-a11b-443f-9cb7-cace67a4e57c?webspace=sulu-test&action=copy-locale&src=en&dest=de&locale=de
```

WDYT?

nnatter: Yes, I think adjusting the API like you described would be the better solution.
Unfortunately, I dont have time to implement this at the moment, at least not this week. So it would be nice if someone else could take over here 😇  @sulu/core-team

danrot:
Replaced by #5573

PR #5560, opened 2020-10-19
Check permissions only if permission-check is configured in webspace security

### LINKED ISSUE #5480: Cache (fos_http_cache) completely deactivated when security is activated
| Q | A
| --- | ---
| Sulu Version | dev-master
| Browser Version | all

# Actual Behavior

Currently the NavigationTwigExtension uses the user to determine the current role of the user.

Therefor on all pages which uses a navigation has a session.

When a session is startet symfony's `AbstractSessionListener` set all responses to private.

# Expected Behavior

The responses should not be private only when i use a Navigation.

# Steps to Reproduce

Install sulu-skeleton and call a navigation twig extension.

# Possible Solutions

No idea

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #5480
| Related issues/PRs | #5439 
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR will only check for permissions e.g. in navigation and smart content if the `permission-check` attribute of the webspace `security` tag is set to true.

#### Why?

See discussion in #5480.

#### To Do

Copied the ones below from #5480 to remember where I have to check for the new flag

- [x] Hide navigation items without sufficient permission
- [x] Hide pages from smart content without sufficient permissions
- [x] Hide pages from a selection field without sufficient permissions
- [x] Hide pages from a single selection field without sufficient permissions
- [x] Hide entities from smart content without sufficient permissions
- [x] Hide entities from a selection field without sufficient permissions
- [x] Hide entities from a single selection field without sufficient permissions
- [x] Hide pages on sitemap not visible to anonymous users

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/Content/Types/SingleMediaSelection.php
+++ b/src/Sulu/Bundle/MediaBundle/Content/Types/SingleMediaSelection.php
@@ -23,6 +23,7 @@ use Sulu\Component\Content\SimpleContentType;
 use Sulu\Component\Security\Authorization\PermissionTypes;
 use Sulu\Component\Security\Authorization\SecurityCheckerInterface;
 use Sulu\Component\Security\Authorization\SecurityCondition;
+use Sulu\Component\Webspace\Analyzer\RequestAnalyzerInterface;
 
 class SingleMediaSelection extends SimpleContentType implements PreResolvableContentTypeInterface
 {
@@ -37,6 +38,11 @@ class SingleMediaSelection extends SimpleContentType implements PreResolvableCon
     private $mediaReferenceStore;
 
     /**
+     * @var RequestAnalyzerInterface
+     */
+    private $requestAnalyzer;
+
+    /**
      * @var ?SecurityCheckerInterface
      */
     private $securityChecker;
@@ -44,10 +50,12 @@ class SingleMediaSelection extends SimpleContentType implements PreResolvableCon
     public function __construct(
         MediaManagerInterface $mediaManager,
         ReferenceStoreInterface $referenceStore,
+        RequestAnalyzerInterface $requestAnalyzer,
         SecurityCheckerInterface $securityChecker = null
     ) {
         $this->mediaManager = $mediaManager;
         $this->mediaReferenceStore = $referenceStore;
+        $this->requestAnalyzer = $requestAnalyzer;
         $this->securityChecker = $securityChecker;
 
         parent::__construct('SingleMediaSelection', '{"id": null}');
@@ -66,7 +74,10 @@ class SingleMediaSelection extends SimpleContentType implements PreResolvableCon
             return null;
         }
 
-        if ($this->securityChecker
+        $webspace = $this->requestAnalyzer->getWebspace();
+
+        if ($webspace->hasWebsiteSecurity()
+            && $this->securityChecker
             && !$this->securityChecker->hasPermission(
                 new SecurityCondition(
                     MediaAdmin::SECURITY_CONTEXT,
--- a/src/Sulu/Bundle/MediaBundle/Content/Types/SingleMediaSelection.php
+++ b/src/Sulu/Bundle/MediaBundle/Content/Types/SingleMediaSelection.php
@@ -76,7 +76,8 @@ class SingleMediaSelection extends SimpleContentType implements PreResolvableCon
 
         $webspace = $this->requestAnalyzer->getWebspace();
 
-        if ($webspace->hasWebsiteSecurity()
+        if ($webspace
+            && $webspace->hasWebsiteSecurity()
             && $this->securityChecker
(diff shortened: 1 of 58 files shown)


src/Sulu/Bundle/MediaBundle/Content/Types/SingleMediaSelection.php
```diff
@@ -66,7 +74,10 @@ public function getContentData(PropertyInterface $property): ?Media
             return null;
         }
 
-        if ($this->securityChecker
+        $webspace = $this->requestAnalyzer->getWebspace();
+
+        if ($webspace->hasWebsiteSecurity()
```
alexander-schranz, line 79: Think in some cases the `$webspace` can here be also null so we need to check here for:

```suggestion
        if ($webspace
            && $webspace->hasWebsiteSecurity()
```

src/Sulu/Bundle/MediaBundle/Tests/Unit/MediaSelectionContentTypeTest.php
```diff
@@ -41,12 +53,15 @@ protected function setUp(): void
     {
         $this->mediaManager = $this->prophesize(MediaManagerInterface::class);
         $this->mediaReferenceStore = $this->prophesize(ReferenceStoreInterface::class);
-        $this->tokenStorage = $this->prophesize(TokenStorageInterface::class);
+        $this->requestAnalyzer = $this->prophesize(RequestAnalyzerInterface::class);
+
+        $this->webspace = new Webspace();
+        $this->requestAnalyzer->getWebspace()->willReturn($this->webspace);
```
alexander-schranz, line 59: Can we add a test where the `getWebspace` is null?

src/Sulu/Component/Content/Mapper/ContentMapper.php
```diff
@@ -689,13 +696,18 @@ private function rowToArray(
             return false;
         }
 
-        if ($document instanceof SecurityBehavior && $permission) {
+        if ($document instanceof SecurityBehavior && $this->security && $permission && $webspace->hasWebsiteSecurity()) {
```
alexander-schranz, line 699: `$webspace` key could be null, think we should here also do:

```suggestion
        if ($document instanceof SecurityBehavior && $this->security && $permission && $webspace && $webspace->hasWebsiteSecurity()) {
```

Else we would need to throw some kind of exception for an unexpected call.

alexander-schranz (changes requested):
Overall it looks good, think at some places we should add the check if `$webspace` is not null, to avoid things calling function on null error.

alexander-schranz:
I created a new page in my example sulu community bundle installation using this branch without any permission check enabled:

```xml
    <security>
        <system>Website</system>
    </security>
```

In the admin UI after creating the page the Permission tab look like this:

Bildschirmfoto 2020-10-27 um 12 15 31

I think there should be no permissions here, I think also the whole system should not be shown as there is no permission-check it should also not be able to set permissions for the page.

danrot:
@alexander-schranz done

src/Sulu/Bundle/PageBundle/Admin/PageAdmin.php
```diff
@@ -316,7 +316,7 @@ public function getSecurityContexts()
             }
 
             $system = $security->getSystem();
-            if (!$system) {
+            if (!$webspace->hasWebsiteSecurity()) {
```
alexander-schranz, line 319: I don't think this is the correct place to handle this. I think the system if configured should still exist to create roles but the section not be shown on the permission tab. The problem with this solution is also if I have a webspace A with:

```xml
        <system>Website</system>
    </security>
```

and a webspace B with:

```xml
    <security>
        <system>Website</system>
    </security>
```

both webspace would show the `Website` Permission Tab. So I think we need to handle this in the JS component, where we currently handles that only the specific system is shown, we need to check for `hasWebsiteSecurity`, don't see currently another solution to handle this.

alexander-schranz (changes requested):
See commented part. I think we need another solution here.

src/Sulu/Bundle/PageBundle/Admin/PageAdmin.php
danrot: @alexander-schranz Wanted to do it like that at first, but somehow was not able to, the break seems to had a good influence :see_no_evil: 

However, now I've implemented it that way, and not really happy either... It feels like our config request returns something wrong (because it looks like you can set these permissions on these systems) and you have to fix that in the frontend by combining it with some other information. From that point of view I liked the other solution better, although what you described might be a problem (although I am quite sure that the setup you've described there would break at other points as well).

alexander-schranz (approved):

PR #5561, opened 2020-10-21
Set default options to routes created in the route-providers

nnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes https://github.com/sulu/SuluArticleBundle/issues/521
| License | MIT

#### What's in this PR?

This PR adjusts the following route providers to include the symfony route default options:
* `RouteProvider` of the `RouteBundle`
* `ContentRouteProvider` of the `WebsiteBundle`
* `CustomUrlRouteProvider` of the `CustomUrlRouteBundle`

The route default options were introduced in Symfony 4.2 to handle the `framework.router.utf8` configuration: https://github.com/symfony/symfony/pull/27774. This options allows to set the `utf8` option to all router per default. This is needed when using UTF8 symbols inside of the path or the requirement of a route. Unfortunately, the `symfony-cmf/routing` package cannot handle this for us (see discussion below).

The `utf8` option of the `Route` class was introduced in Symfony 3.2:
* https://symfony.com/blog/new-in-symfony-3-2-unicode-routing-support
* https://github.com/symfony/symfony/pull/19604/files

#### Why?

Because routes cannot use UTF-8 characters without setting the `utf8` option (See https://github.com/sulu/SuluArticleBundle/issues/521). For example, the path `accééeeents` leads to the following error: 
```
Cannot use UTF-8 route patterns without setting the "utf8" option for route "/en/accééeeents3".
```

This is caused by the following lines:
https://github.com/symfony/symfony/blob/35dad22f9e01b06b538837b8c729d67f26c6f6ed/src/Symfony/Component/Routing/RouteCompiler.php#L113-L118


nnatter:
I guess we should only add the `utf8` option if `framework.router.utf8` is configured to true. Do you have an idea how to do that? 🤔

alexander-schranz:
Maybe this is something which should the Symfony CMF Routing Bundle should take care of if the `utf8: true` is set all the routes should also get that option. /cc @dbu what do you think?

dbu:
@alexander-schranz that sounds reasonable. i guess at the time we built this (symfony 2) this option probably did not exist. do you want to do a pull request for it? we might need a compiler pass to "copy" the configuration over.

nnatter:
@dbu Thanks for your quick response. I just had a look the code of the `symfony-cmf/routing` package, but I am not sure where I could set the `utf8: true` option to the route.

My idea was to implement this similar to the symfony implementation: https://github.com/symfony/symfony/pull/27774. In that implementation, the `DelegatingLoader` loops over the loaded route collection and sets the default options (which include the `utf8: true` option): https://github.com/symfony/symfony/pull/27774/files#diff-c6bf228eca2bee9991fbaa09eefbbf51bb3396387f0724c46f19f4cc80285660R78-R80

What would be the correct service to do this in the `symfony-cmf/routing` or `symfony-cmf/routing-bundle` package? I already noticed that the `NestedMatcher` gathers the route collection from the provider: https://github.com/symfony-cmf/Routing/blob/3c97e7b7709b313cecfb76d691ad4cc22acbf3f5/src/NestedMatcher/NestedMatcher.php#L140

But as far as I see, this class is not registered as service in the `symfony-cmf/routing-bundle` but inside of the project. This means that I cannot pass any default options to this class, right? 

Thanks for your time!

dbu:
the idea of the cmf routing is that routes are stored in a database, as opposed to being statically configured. 

is the utf8 option a thing on the route object, or more general? and what exactly is the effect of the flag? 
if it is on the route object, storing it in the object would be possible but likely be weird as you probably want to have that globally. it might affect the route loading mechanism [candidates](https://github.com/symfony-cmf/Routing/blob/master/src/Candidates/Candidates.php#L129) and [validity check](https://github.com/symfony-cmf/routing-bundle/blob/master/src/Doctrine/Phpcr/PrefixCandidates.php#L73)... depending on the storage system, we might also need to encode the paths and decode them again, and then encode in the candidates strategy - i think at least PHPCR does not fully handle utf-8 in node names.

i hope this helps you. i am afraid that this could be quite a chain of changes in the end... we do not necessarily need it for all storage solutions  provided by the cmf bundle, they already do not have exactly the same features.

nnatter:
@dbu Thanks for your answer!

Yes, the `utf8: true` is just an option that is passed to the/saved on the `Route`. For example, it can be set when constructing a route manually (`new Route('path', [], [], ['utf8' => true])`) or by using the following annotation:
```
/**
 * @Route(
 *     "/category/{name}",
 *     "options" = { "utf8": true }
 * )
 */
```

The effect of the flag is that the `PCRE_UTF8` modifier is added to the regex of the route. This allows to use unicode characters in the path and in the requirements of the route:
https://github.com/symfony/symfony/blob/35dad22f9e01b06b538837b8c729d67f26c6f6ed/src/Symfony/Component/Routing/RouteCompiler.php#L234-L241 

With this in mind, here comes the problem: In 4.2, Symfony added a `framework.router.utf8` configuration that sets `['utf8' => true]` to all loaded routes per default. My goal would be to somehow respect this option in the `symfony-cmf/routing-bundle`. Meaning that I would like to set `['utf8' => true]` to all routes returned by the registered providers. 

At the moment, setting this default value to the routes when the `framework.router.utf8` configuration is set would be enough for us. I think we dont need to change anything about storing/loading the routes 🙂 

Do you see a place for handling this in the `symfony-cmf/routing-bundle`? Or should we just adjust the `RouteProvider` implementation inside of Sulu to respect the `framework.router.utf8` configuration?

dbu:
i see. then i think you should do it in the step that loads the route candidates from storage. (you will need to double-check if the candidates strategy works with utf8 routes though, as the point in the cmf route loading it to not always load potentially thousands or tens of thousands of routes)

nnatter:
@sulu/core-team I adjusted this PR to to use compiler passes for copying the route default options to our route providers. I also updated the title and description to explain why this is necessary 🙂

src/Sulu/Bundle/WebsiteBundle/DependencyInjection/Compiler/RouteDefaultOptionsCompilerPass.php
```diff
@@ -0,0 +1,36 @@
...
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\WebsiteBundle\DependencyInjection\Compiler;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+class RouteDefaultOptionsCompilerPass implements CompilerPassInterface
```
alexander-schranz, line 17: Maybe instead of having multiple compilerpasses we should have one compilerpass in the route or core bundle which sets a parameter which could be just used by the other service definitions.

src/Sulu/Bundle/WebsiteBundle/DependencyInjection/Compiler/RouteDefaultOptionsCompilerPass.php
```diff
@@ -0,0 +1,36 @@
...
+
+        if (!$container->hasDefinition('sulu_website.provider.content')) {
+            return;
+        }
+
+        // copy default route options which are set by the symfony FrameworkExtension based on the config:
+        // https://github.com/symfony/symfony/pull/31900
+        $container->getDefinition('sulu_website.provider.content')->replaceArgument(
+            7,
+            $container->getDefinition('routing.loader')->getArgument(1)
```
alexander-schranz, line 33: Is it not possible to use `getArgument('$defaultOptions')` or something to be sure that we load no other argument instead?

alexander-schranz (changes requested):
I think to be sure that the ContentProvider is getting the correct defaultOptions we should have some kind of functional test case to test if the route have utf8=>true set. A mocked test I think is in this cases not enough as it would not test if the compilerpass works correctly when symfony does change the routing.loader service definition.

src/Sulu/Bundle/WebsiteBundle/DependencyInjection/Compiler/RouteDefaultOptionsCompilerPass.php
nnatter: Yes i thought about that too, but where should I place it? 🙂

src/Sulu/Bundle/WebsiteBundle/DependencyInjection/Compiler/RouteDefaultOptionsCompilerPass.php
nnatter: Unfortunately not. `The argument "$defaultOptions" doesn't exist.`

alexander-schranz: Okay a functional test that the defaultOptions is correctly set is then for me more important, else in a symfony upgrade this is maybe forgotten to be updated.

nnatter: added a functional test case 🙂

alexander-schranz (approved):

alexander-schranz (approved):

PR #5562, opened 2020-10-22
Register route field-type only on first login

### LINKED ISSUE #5565: Second log out throws JS error
| Q | A
| --- | ---
| Sulu Version 2.1.1

# Actual Behavior

Log out page loads infinitly after second log out. JS error thrown:
`Uncaught (in promise) Error: The key "route" has already been used for another field`

# Steps to Reproduce

Log in into admin panel
Log out
Log in again
Log out

nnatter:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #5565
| License | MIT

#### What's in this PR?

This PR adjusts the `index.js` of the `RouteBundle` to register the `route` field-type one on the first login.

#### Why?

Because otherwise, this will throw an error if somebody logs in, logs out and logs in again without reloading the application.

#### Reproducing the Problem

1. Login into the administration interface
2. Logout
3. Login again
4. Try to Logout again
5. **A third Login is not possible anymore**

### DIFF
--- a/src/Sulu/Bundle/PageBundle/Resources/js/stores/webspaceStore/webspaceStore.js
+++ b/src/Sulu/Bundle/PageBundle/Resources/js/stores/webspaceStore/webspaceStore.js
@@ -1,12 +1,12 @@
 // @flow
-import {computed, observable} from 'mobx';
+import {action, computed, observable} from 'mobx';
 import log from 'loglevel';
 import type {Webspace} from './types';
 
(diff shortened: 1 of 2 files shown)


alexander-schranz (approved):
Looks good from my view. Still would like if @danrot could have a look at it if this is the correct way to handle it.

PR #5564, opened 2020-10-22
Fix image_map example template

luca-rath:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Fixes image_map example template

#### Why?

If no image set in admin, an error is thrown in twig.

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/Content/Types/ImageMapContentType.php
+++ b/src/Sulu/Bundle/MediaBundle/Content/Types/ImageMapContentType.php
@@ -340,23 +340,35 @@ class ImageMapContentType extends ComplexContentType implements ContentTypeExpor
 
     public function getViewData(PropertyInterface $property)
     {
-        return $this->prepareData(
+        $data = $this->prepareData(
             $property,
             function(ContentTypeInterface $contentType, $property) {
                 return $contentType->getViewData($property);
             },
             false
         );
+
+        if (!($data['image'] ?? null)) {
+            return null;
+        }
+
+        return $data;
     }
 
     public function getContentData(PropertyInterface $property)
     {
-        return $this->prepareData(
+        $data = $this->prepareData(
             $property,
             function(ContentTypeInterface $contentType, $property) {
                 return $contentType->getContentData($property);
             }
         );
+
+        if (!($data['image'] ?? null)) {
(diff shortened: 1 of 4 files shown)


luca-rath:
@danrot The `media_selection`'s empty value isn't `null` either, it's an empty array

alexander-schranz:
@luca-rath its not about the `media_selection` its about the `single_media_selection` which is null if the image is not set. So this should be the case for the `image_map`, else developers will run always in the same error as you did in the example. 

See the SingleMediaSelection Content Type here: https://github.com/sulu/sulu/blob/aaa1273abefcd3016243766b0199729097dd074b/src/Sulu/Bundle/MediaBundle/Content/Types/SingleMediaSelection.php#L59-L61

templates/pages/example.html.twig
```diff
@@ -85,7 +85,5 @@
     <p>{{ content.contact_account_selection|map(contact => contact.name ?? contact.fullName)|join(', ') }}</p>
     <p>{{ content.text_area }}</p>
 
-    {% if content.image_map %}
```
alexander-schranz, line 88: This is correct and should work.

alexander-schranz (changes requested):

danrot:
A test is missing.

src/Sulu/Bundle/MediaBundle/Content/Types/ImageMapContentType.php
```diff
@@ -351,12 +351,18 @@ function(ContentTypeInterface $contentType, $property) {
...
     {
-        return $this->prepareData(
+        $data = $this->prepareData(
             $property,
             function(ContentTypeInterface $contentType, $property) {
                 return $contentType->getContentData($property);
             }
         );
+
+        if (!($data['image'] ?? null)) {
```
danrot, line 361: Doesn't `!isset($data['image'])` work as well? This line looks a bit confusing and unusual to me.

danrot:
The CircleCI build is currently also failing on master and this PR is not changing an JS at all... So I am going to merge that as soon as the other tests are passing.

danrot (approved):

PR #5567, opened 2020-10-23
Make select focusable

luca-rath:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Make select component controllable using arrows and text input

#### Why?

Better UX

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Action.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Action.js
@@ -13,7 +13,7 @@ type Props<T> = {|
 |};
 
 export default class Action<T> extends React.PureComponent<Props<T>> {
-    handleButtonClick = () => {
+    triggerButton = () => {
         const {
             onClick,
             afterAction,
@@ -24,6 +24,18 @@ export default class Action<T> extends React.PureComponent<Props<T>> {
 
         if (afterAction) {
             afterAction();
+        }
+    };
+
+    handleButtonClick = () => {
+        this.triggerButton();
+    };
+
+    handleButtonKeyDown = (event: KeyboardEvent) => {
+        if (event.key === 'Enter' || event.key === 'Space') {
+            event.preventDefault();
+            event.stopPropagation();
+            this.triggerButton();
         }
     };
 
@@ -47,6 +59,7 @@ export default class Action<T> extends React.PureComponent<Props<T>> {
                 <button
                     className={actionStyles.action}
                     onClick={this.handleButtonClick}
+                    onKeyDown={this.handleButtonKeyDown}
                     ref={this.setButtonRef}
                 >
                     {this.props.children}
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Action.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Action.js
@@ -1,6 +1,6 @@
 // @flow
 import React from 'react';
-import type {ElementRef} from "react";
+import type {ElementRef} from 'react';
 import actionStyles from './action.scss';
 
 type Props<T> = {|
@@ -35,7 +35,7 @@ export default class Action<T> extends React.PureComponent<Props<T>> {
         }
     };
 
-    handleMouseMove = (event: MouseEvent) => {
+    handleMouseMove = () => {
         if (this.props.requestFocus) {
             this.props.requestFocus();
         }
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Action.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Action.js
@@ -1,11 +1,14 @@
 // @flow
 import React from 'react';
+import type {ElementRef} from "react";
 import actionStyles from './action.scss';
 
 type Props<T> = {|
     afterAction?: () => void,
+    buttonRef?: (buttonRef: ?ElementRef<'button'>) => void,
     children: string,
     onClick: (value: ?T) => void,
+    requestFocus?: () => void,
     value?: T,
 |};
 
@@ -24,10 +27,28 @@ export default class Action<T> extends React.PureComponent<Props<T>> {
         }
     };
 
+    setButtonRef = (ref: ?ElementRef<'button'>) => {
+        const {buttonRef} = this.props;
+
+        if (buttonRef) {
+            buttonRef(ref);
+        }
+    };
+
+    handleMouseMove = (event: MouseEvent) => {
+        if (this.props.requestFocus) {
+            this.props.requestFocus();
+        }
+    };
+
     render() {
         return (
-            <li>
-                <button className={actionStyles.action} onClick={this.handleButtonClick}>
+            <li onMouseMove={this.handleMouseMove}>
+                <button
+                    className={actionStyles.action}
+                    onClick={this.handleButtonClick}
+                    ref={this.setButtonRef}
+                >
                     {this.props.children}
                 </button>
             </li>
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Select.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Select.js
@@ -231,8 +231,8 @@ class Select<T> extends React.Component<Props<T>> {
             this.clearSearchText();
 
             focusedElementIndex = Math.max(
-                ...this.availableButtonIndices.filter(i => i < this.focusedElementIndex)
-            )
+                ...this.availableButtonIndices.filter((i) => i < this.focusedElementIndex)
+            );
         }
 
         if (event.key === 'ArrowDown') {
@@ -240,8 +240,8 @@ class Select<T> extends React.Component<Props<T>> {
             this.clearSearchText();
 
             focusedElementIndex = Math.min(
-                ...this.availableButtonIndices.filter(i => i > this.focusedElementIndex)
-            )
+                ...this.availableButtonIndices.filter((i) => i > this.focusedElementIndex)
+            );
         }
 
         if (focusedElementIndex !== this.focusedElementIndex) {
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Select.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Select.js
@@ -3,6 +3,7 @@ import React from 'react';
 import type {Element, ElementRef} from 'react';
 import {action, observable} from 'mobx';
 import {observer} from 'mobx-react';
+import debounce from 'debounce';
 import {translate} from '../../utils/Translator';
 import Popover from '../Popover';
 import Menu from '../Menu';
@@ -43,6 +44,12 @@ class Select<T> extends React.Component<Props<T>> {
 
     @observable selectedOptionRef: ?ElementRef<'li'>;
 
+    @observable buttonRefs: Map<number, ElementRef<'button'>> = new Map();
+
+    @observable searchText: string = '';
+
+    @observable focusIndex: ?number;
+
     @observable open: boolean;
 
     @action openOptionList = () => {
@@ -71,7 +78,35 @@ class Select<T> extends React.Component<Props<T>> {
         }
     };
 
-    cloneOption(originalOption: Element<Class<Option<T>>>): Element<Class<Option<T>>> {
+    setButtonRef = (index: number) => action((ref: ?ElementRef<'li'>) => {
+        if (ref) {
+            this.buttonRefs.set(index, ref);
+        } else if (this.buttonRefs.has(index)) {
+            this.buttonRefs.delete(index);
+        }
+    });
+
+    @action clearSearchText = () => {
+        this.searchText = '';
+    };
+
+    debouncedClearSearchText = debounce(this.clearSearchText, 500);
+
+    @action appendSearchText = (searchText: string) => {
+        this.searchText += searchText;
+
+        const hit = Array.from(this.buttonRefs.entries()).find(([index, ref]) => {
+            return ref.textContent.startsWith(this.searchText);
+        });
+
+        if (hit) {
+            this.setFocusIndex(hit[0])
+        }
+
+        this.debouncedClearSearchText();
+    };
+
+    cloneOption(originalOption: Element<Class<Option<T>>>, index: number): Element<Class<Option<T>>> {
         const anchorWidth = this.displayValueRef ? this.displayValueRef.getBoundingClientRect().width : 0;
 
         return React.cloneElement(originalOption, {
@@ -79,7 +114,9 @@ class Select<T> extends React.Component<Props<T>> {
             onClick: this.handleOptionClick,
             selected: this.props.isOptionSelected(originalOption),
             selectedVisualization: this.props.selectedVisualization,
+            setFocusIndex: this.handleSetFocusIndex(index),
             optionRef: this.setSelectedOptionRef,
+            buttonRef: this.setButtonRef(index),
         });
     }
 
@@ -90,14 +127,14 @@ class Select<T> extends React.Component<Props<T>> {
     }
 
     cloneChildren(): SelectChildren<T> {
-        return React.Children.map(this.props.children, (child: any) => {
+        return React.Children.map(this.props.children, (child: any, index: number) => {
             if (!child) {
                 return child;
             }
 
             switch (child.type) {
                 case Option:
-                    return this.cloneOption(child);
+                    return this.cloneOption(child, index);
                 case Action:
                     return this.cloneAction(child);
                 default:
@@ -118,6 +155,69 @@ class Select<T> extends React.Component<Props<T>> {
 
     handleOptionListClose = this.closeOptionList;
 
+    @action setFocusIndex = (focusIndex: number) => {
+        if (focusIndex < 0 || focusIndex >= this.buttonRefs.size) {
+            return;
+        }
+
+        this.focusIndex = focusIndex;
+
+        if (this.buttonRefs.has(focusIndex)) {
+            const ref = this.buttonRefs.get(focusIndex);
+            ref.focus();
+        }
+    };
+
+    handleSetFocusIndex = (focusIndex: number) => () => {
+        this.setFocusIndex(focusIndex);
+    };
+
+    @action handleKeyDown = (event: KeyboardEvent) => {
+        if (!this.open) {
+            return;
+        }
+
+        if (event.key === 'Escape' || event.code === 27) {
+            event.preventDefault();
+            this.closeOptionList();
+
+            return;
+        }
+
+        if (event.key === 'Tab' || event.code === 9) {
+            event.preventDefault();
+            this.clearSearchText();
+
+            return;
+        }
+
+        if (event.key === 'ArrowUp' || event.code === 38) {
+            event.preventDefault();
+            this.setFocusIndex(this.focusIndex - 1);
+            this.clearSearchText();
+
+            return;
+        }
+
+        if (event.key === 'ArrowDown' || event.code === 40) {
+            event.preventDefault();
+            this.setFocusIndex(this.focusIndex + 1);
+            this.clearSearchText();
+
+            return;
+        }
+
+    };
+
+    handleKeyPress = (event: KeyboardEvent) => {
+        if (!this.open) {
+            return;
+        }
+
+        event.preventDefault();
+        this.appendSearchText(event.key);
+    };
+
     render() {
         const {
             icon,
@@ -128,7 +228,7 @@ class Select<T> extends React.Component<Props<T>> {
         const clonedChildren = this.cloneChildren();
 
         return (
-            <div className={selectStyles.select}>
+            <div className={selectStyles.select} onKeyDown={this.handleKeyDown} onKeyPress={this.handleKeyPress}>
                 <DisplayValue
                     disabled={disabled}
                     displayValueRef={this.setDisplayValueRef}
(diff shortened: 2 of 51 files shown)


luca-rath:
@danrot This example works, but I think it's a lot slower than it was before. I think it's very difficult to implement this nicely, because the value is not a flat array of options or something like that, instead it's a list of `Option` components passed as children. So could you please have a look at the code and point me into the right direction how to optimize and enhance all of that.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/ResourceLocatorHistory/tests/__snapshots__/ResourceLocatorHistory.test.js.snap
```diff
@@ -99,7 +99,7 @@ Array [
                       <div
                         class="cellContent"
                       >
-                        4/10/2019, 1:06:16 PM
+                        2019-4-10 13:06:16
```
luca-rath, line 102: I'm unsure about this change?

src/Sulu/Bundle/MediaBundle/Resources/js/views/MediaHistory/tests/__snapshots__/MediaHistory.test.js.snap
```diff
@@ -100,7 +100,7 @@ exports[`Render a MediaHistory view 1`] = `
             <div
               class="cellContent"
             >
-              10/23/2018, 10:25:00 AM
+              2018-10-23 10:25:00
```
luca-rath, line 103: Same here

src/Sulu/Bundle/MediaBundle/Resources/js/views/MediaHistory/tests/__snapshots__/MediaHistory.test.js.snap
```diff
@@ -150,7 +150,7 @@ exports[`Render a MediaHistory view 1`] = `
             <div
               class="cellContent"
             >
-              10/23/2018, 10:18:00 AM
+              2018-10-23 10:18:00
```
luca-rath, line 153: Same here

src/Sulu/Bundle/AdminBundle/Resources/js/containers/ResourceLocatorHistory/tests/__snapshots__/ResourceLocatorHistory.test.js.snap
```diff
@@ -135,7 +135,7 @@ Array [
                       <div
                         class="cellContent"
                       >
-                        4/10/2019, 4:01:12 PM
+                        2019-4-10 16:01:12
```
luca-rath, line 138: Same here

src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Select.js
```diff
@@ -128,7 +257,12 @@ class Select<T> extends React.Component<Props<T>> {
         const clonedChildren = this.cloneChildren();
 
         return (
-            <div className={selectStyles.select}>
+            <div
+                className={selectStyles.select}
+                onKeyDown={this.handleKeyDown}
```
nnatter, line 262: is there a specific reason for binding `onKeyDown` and `onKeyPress`? 🙂

src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Select.js
```diff
@@ -118,6 +180,73 @@ class Select<T> extends React.Component<Props<T>> {
...
+            do {
+                ++focusedElementIndex;
+            } while (!this.buttonRefs.has(focusedElementIndex) && focusedElementIndex <= this.highestButtonIndex);
+        }
+
+        if (focusedElementIndex !== this.focusedElementIndex) {
+            this.requestFocus(focusedElementIndex);
+        }
+
+        this.clearSearchText();
```
nnatter, line 238: doesnt this always reset the search term? so the searchterm can never be longer than 1 character? 🤔

src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Action.js
```diff
@@ -24,10 +27,28 @@ export default class Action<T> extends React.PureComponent<Props<T>> {
...
+        if (this.props.requestFocus) {
+            this.props.requestFocus();
+        }
+    };
+
     render() {
         return (
-            <li>
-                <button className={actionStyles.action} onClick={this.handleButtonClick}>
+            <li onMouseMove={this.handleMouseMove}>
```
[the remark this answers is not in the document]
luca-rath: okay: if you hover an item, then press arrow down a few times and then hover the mouse a little bit without leaving the current hovered item, that item isn't being focus again

src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Select.js
```diff
@@ -43,10 +44,38 @@ class Select<T> extends React.Component<Props<T>> {
...
+    @observable focusedElementIndex: number = -1;
+
     @observable open: boolean;
 
+    @computed get buttonTexts() {
+        return Array.from(this.buttonRefs.entries())
+            .map<[number, string]>(([index, ref]) => [index, ref.textContent]);
+    }
+
+    @computed get selectedIndex() {
```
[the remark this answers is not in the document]
luca-rath: yes, but I think it's better than using the first item as first focused element, because if you have a very long select list (like countries) the current selected item is always centered. so if you would just press arrow down, the list would jump to the top

src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Select.js
luca-rath: keyPress ignores non printable keys, and that's exactly what is needed ... Since arrow down/up and escape are not captured by keyPress, I need the keyDown as well

nnatter: ah okay did not know this 👍

src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Action.js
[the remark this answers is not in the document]
nnatter: alright - makes sense

src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Select.js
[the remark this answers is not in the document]
nnatter: thats true, but i think i would call this `firstSelectedIndex` then

src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Select.js
```diff
@@ -118,6 +180,73 @@ class Select<T> extends React.Component<Props<T>> {
...
+
+            return;
+        }
+
+        let focusedElementIndex = this.focusedElementIndex;
+
+        if (event.key === 'ArrowUp') {
+            do {
+                --focusedElementIndex;
+            } while (!this.buttonRefs.has(focusedElementIndex) && focusedElementIndex > 0);
```
danrot, line 225: Is that some kind of busy wait? Or why do we need a loop here? Is maybe that the reason for some of the performance issues you are experiencing?

nnatter: think this is needed to skip `<Divider>` children, right? But it might be easier to understand when using something like `focusedElementIndex  = Math.min(...availableIndices.filter(i => i > this.focusedElementIndex), 0)`

luca-rath: yes, exactly

luca-rath: will try your suggestions @nnatter

src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Select.js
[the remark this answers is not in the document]
luca-rath: fixed

src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Select.js
```diff
@@ -43,10 +44,38 @@ class Select<T> extends React.Component<Props<T>> {
 
     @observable selectedOptionRef: ?ElementRef<'li'>;
 
+    @observable buttonRefs: Map<number, ElementRef<'button'>> = new Map();
```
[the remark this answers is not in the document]
luca-rath: fixed

src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Select.js
```diff
@@ -43,10 +44,38 @@ class Select<T> extends React.Component<Props<T>> {
...
+
+    @observable searchText: string = '';
+
+    @observable focusedElementIndex: number = -1;
+
     @observable open: boolean;
 
+    @computed get buttonTexts() {
```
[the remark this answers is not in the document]
luca-rath: fixed

src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Select.js
```diff
@@ -118,6 +180,73 @@ class Select<T> extends React.Component<Props<T>> {
 
     handleOptionListClose = this.closeOptionList;
 
+    @computed get highestButtonIndex() {
+        const max = Math.max(...Array.from(this.buttonRefs.keys()));
+        return max;
+    }
+
+    @action requestFocus = (focusedElementIndex: number) => {
```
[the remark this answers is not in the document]
luca-rath: fixed

src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Select.js
```diff
@@ -118,6 +180,73 @@ class Select<T> extends React.Component<Props<T>> {
...
+        if (focusedElementIndex < 0 || focusedElementIndex > this.highestButtonIndex) {
```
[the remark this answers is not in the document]
luca-rath: fixed

src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Select.js
```diff
@@ -118,6 +180,73 @@ class Select<T> extends React.Component<Props<T>> {
...
+            this.focusedElementIndex = focusedElementIndex;
+            const ref = this.buttonRefs.get(focusedElementIndex);
+
+            if (ref) {
+                ref.focus();
+            }
+        }
+    };
+
+    handleRequestFocus = (focusedElementIndex: number) => () => {
```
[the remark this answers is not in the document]
luca-rath: fixed

src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Select.js
luca-rath: fixed

src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Select.js
```diff
@@ -118,6 +189,75 @@ class Select<T> extends React.Component<Props<T>> {
 
     handleOptionListClose = this.closeOptionList;
 
+    @computed get highestButtonIndex() {
```
nnatter, line 192: think this is not used anymore

src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Select.js
```diff
@@ -43,10 +44,45 @@ class Select<T> extends React.Component<Props<T>> {
...
 
+    @observable buttonRefsByIndex: Map<number, ElementRef<'button'>> = new Map();
+
+    @observable searchText: string = '';
+
+    @observable focusedElementIndex: number = -1;
+
     @observable open: boolean;
 
+    @computed get buttonTexts(): Map<number, string> {
```
nnatter, line 55: now we need to call this `buttonTextsByIndex` 😇

src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Select.js
```diff
@@ -118,6 +189,75 @@ class Select<T> extends React.Component<Props<T>> {
...
+        }
+
+        let focusedElementIndex = this.focusedElementIndex;
+
+        if (event.key === 'ArrowUp') {
+            event.preventDefault();
+            this.clearSearchText();
+
+            focusedElementIndex = Math.max(
+                ...this.availableButtonIndices.filter((i) => i < this.focusedElementIndex)
```
nnatter, line 234: will this work if you are already at the top of the list? 🤔

luca-rath: yes, because the `requestFocus()` checks, if the `buttonRefsByIndex` has that key

src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Select.js
luca-rath: fixed

src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Select.js
luca-rath: fixed

src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Select.js
```diff
@@ -43,10 +44,45 @@ class Select<T> extends React.Component<Props<T>> {
...
+                firstSelectedIndex = index;
+            }
+        });
+
+        return firstSelectedIndex;
+    }
+
     @action openOptionList = () => {
         this.open = true;
+        this.focusedElementIndex = this.firstSelectedIndex;
```
nnatter, line 85: shouldnt we use `requestFocus()` here? this way the option is not styled

src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Select.js
```diff
@@ -118,6 +189,70 @@ class Select<T> extends React.Component<Props<T>> {
...
+        if (ref) {
+            ref.focus();
+        }
+    };
+
+    handleRequestFocus = (elementIndex: number) => () => {
+        this.requestFocus(elementIndex);
+    };
+
+    @action handleKeyDown = (event: KeyboardEvent) => {
```
nnatter, line 209: i think we need to handle `Enter` here too. otherwise i am not able to submit my selection

nnatter: also the native select is opened when pressing `ArrowUp` or `ArrowDown` while it is focused. should we implement that too?
https://www.w3schools.com/tags/tryit.asp?filename=tryhtml_select

src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Select.js
```diff
@@ -43,10 +44,45 @@ class Select<T> extends React.Component<Props<T>> {
 
     @observable selectedOptionRef: ?ElementRef<'li'>;
 
+    @observable buttonRefsByIndex: Map<number, ElementRef<'button'>> = new Map();
+
+    @observable searchText: string = '';
+
+    @observable focusedElementIndex: number = -1;
```
nnatter, line 51: do you think we need to store this inside of the component? as far as i see, the only time it is used is on line 242 where we could just request the focue without the check

src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Select.js
```diff
@@ -118,6 +189,70 @@ class Select<T> extends React.Component<Props<T>> {
...
+        if (event.key === 'ArrowDown') {
+            event.preventDefault();
+            this.clearSearchText();
+
+            focusedElementIndex = Math.min(
+                ...this.availableButtonIndices.filter((i) => i > this.focusedElementIndex)
+            );
+        }
+
+        if (focusedElementIndex !== this.focusedElementIndex) {
```
nnatter, line 242: not sure if we need this check. if we omit it, we could remove the `this.focusedElementIndex` getter and also the `focusedElementIndex` variable in this method

src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Select.js
```diff
@@ -118,6 +189,70 @@ class Select<T> extends React.Component<Props<T>> {
...
+            focusedElementIndex = Math.max(
+                ...this.availableButtonIndices.filter((i) => i < this.focusedElementIndex)
+            );
+        }
+
+        if (event.key === 'ArrowDown') {
+            event.preventDefault();
+            this.clearSearchText();
+
+            focusedElementIndex = Math.min(
```
nnatter, line 237: if we remove the `this.focusedElementIndex`, we could just directly call `requestFocus`

src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Select.js
luca-rath: It's also used in line 222

src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Select.js
luca-rath: no, we need the focusedElementIndex, otherwise we can't know which one is the previous/next element

src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Select.js
nnatter: you are right, forgot about that 🙈

src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Select.js
nnatter: maybe we should to set the focus in the `setButtonRef` method if the index matches?

src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Select.js
nnatter: yes, i forgot about that. but we could still just call `requestFocus` here directly without the need of a variable 🙂

src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Select.js
luca-rath: fixed

src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Select.js
luca-rath: fixed

src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Select.js
luca-rath: fixed

src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Select.js
luca-rath: fixed

src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Select.js
```diff
@@ -43,10 +44,70 @@ class Select<T> extends React.Component<Props<T>> {
...
+
+    @computed get firstSelectedIndex() {
+        let firstSelectedIndex = -1;
+
+        React.Children.forEach(this.props.children, (child: any, index: number) => {
+            if (!child || child.type !== Option) {
+                return;
+            }
+
+            if (this.props.isOptionSelected(child)) {
```
nnatter, line 75: think we need to check if `firstSelectedIndex` was already set. otherwise we will return the last one

src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Select.js
```diff
@@ -57,67 +118,158 @@ class Select<T> extends React.Component<Props<T>> {
...
+    handleDisplayValueClick = this.openOptionList;
+
+    handleOptionListClose = this.closeOptionList;
+
+    handleRequestFocus = (elementIndex: number) => () => {
+        this.requestFocus(elementIndex);
+    };
+
+    handleKeyDown = (event: KeyboardEvent) => {
+        if (['Enter', 'ArrowUp', 'ArrowDown'].includes(event.key)) {
```
nnatter, line 176: would be nice to test this 🙂

nnatter (commented):
looks good to me otherwise 🙂

src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Select.js
luca-rath: fixed

src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Select.js
luca-rath: fixed

nnatter (approved):
Looks good to me! Can we merge this or should we wait for the release? @danrot

danrot:
@nnatter Since I've already created a release/2.2 branch it would be ok to merge this into master now.

nnatter:
While talking to @alexander-schranz and @chirimoya we decided to merge this into `release/2.1` after `2.2.0` was released.

PR #5569, opened 2020-10-27
Fix jest regression in expect.objectContaining for observable variables

danrot:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | https://github.com/facebook/jest/pull/10508
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR avoids the usage of `expect.objectContaining` in some cases.

#### Why?

Because https://github.com/facebook/jest/pull/10508 introduced a regression. As soon as `expect.objectContaining` is called, it seems like variables passed to it are sometimes changed. For that reason the interceptor on the `excludeGhostsAndShadows` variables was changed, so the `intercept` call stopped working.


danrot:
Will probably be fixed by https://github.com/facebook/jest/issues/10689, which will probably released today... So maybe we don't need to merge that.

nnatter:
Think I would wait for the fix 🙂

danrot:
Closing this, since we are waiting for the release. Hopefully they release it soon...

PR #5573, opened 2020-10-27
Load correct language after copying content for ghost page from other language

### LINKED ISSUE #5548: Translated pages do not use translated parent pages in path prefix

danrot:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #5548
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR makes the request after confirming the `GhostDialog` return the language of the destination language. In addition it adjusts the page REST API in a way to make this possible without a BC break.

This also replaces #5559.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/stores/ResourceStore/tests/ResourceStore.test.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/stores/ResourceStore/tests/ResourceStore.test.js
@@ -525,7 +525,7 @@ test('Copy the content from different locale', () => {
     resourceStore.copyFromLocale('de');
 
     expect(ResourceRequester.post)
-        .toBeCalledWith('pages', {}, {action: 'copy-locale', id: 4, locale: 'de', dest: 'en'});
+        .toBeCalledWith('pages', {}, {action: 'copy-locale', id: 4, locale: 'en', dest: 'en', src: 'de'});
 
     return promise.then(() => {
         expect(resourceStore.data).toEqual(germanContent);
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -2,6 +2,29 @@
 
 ## dev-release/2.1
 
+### GhostDialog
+
+When a page was opened in a ghost language, the `GhostDialog`, which allows to copy content from another locale,
+appeared. But after copying the language, the API returned the wrong locale, and therefore the UI was broken. We fixed
+that by separating the `src` and `locale` parameter in the pages API. This results in a different request being sent to
+that API:
+
+Before:
+
+```
+/admin/api/pages/d3354303-a11b-443f-9cb7-cace67a4e57c?webspace=sulu-test&action=copy-locale&dest=de&locale=en
+```
+
+After:
+
+```
+/admin/api/pages/d3354303-a11b-443f-9cb7-cace67a4e57c?webspace=sulu-test&action=copy-locale&src=en&dest=de&locale=de
+```
+
+Mind the change of the `locale` query parameter and the addition of the `src` parameter. In case you have reused that
+functionality with some of your custom entities, you need to adjust your API so that they work with the new query
+parameters.
+
 ### DateTime filter type
 
 The DateTime filter type does now support time by default. If you want to reuse the "old" behaviour we have introduced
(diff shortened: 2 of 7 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/stores/ResourceStore/tests/ResourceStore.test.js
```diff
@@ -525,7 +525,7 @@ test('Copy the content from different locale', () => {
     resourceStore.copyFromLocale('de');
 
     expect(ResourceRequester.post)
-        .toBeCalledWith('pages', {}, {action: 'copy-locale', id: 4, locale: 'de', dest: 'en'});
+        .toBeCalledWith('pages', {}, {action: 'copy-locale', id: 4, locale: 'en', dest: 'en', src: 'de'});
```
alexander-schranz, line 528: This is a bc break as we are now sending the `dest` locale as locale and this will break existing bundles as the now get a false locale for `copy-locale` action.

alexander-schranz (changes requested):
I really think we can't avoid the second request, this is the only we can sure not breaking exist bundles using this behaviour and also fix it there.

src/Sulu/Bundle/AdminBundle/Resources/js/stores/ResourceStore/tests/ResourceStore.test.js
danrot: Other existing bundles are probably broken in the same way as the pages are, so they need fixing anyway. Therefore I would maybe mention that in the `UPGRADE.md`, but I would still not make two requests, since this is bad performance-wise.

nnatter (commented):
Code looks good and the change works for me! 👍 
I do not have a strong opinion on the BC break, so I think I am going to wait for that discussion to resolve before merging this 🙂

src/Sulu/Bundle/AdminBundle/Resources/js/stores/ResourceStore/tests/ResourceStore.test.js
alexander-schranz: We still can not `change` the `locale` parameter here, it will break more things as before for this bundles as it will try then to copy from `en` to `en`. The `locale` and `dest` should stay the same and we could add another additional parameter `return-dest=true` or something, else it will break things more as it was done before.

danrot: I think there are not too many bundles out there relying on that (probably just the ones we've created on our own), therefore I would not go down that road and build a very complicated API.

In addition to that `return-dest` won't work for all cases, since the API also accepts multiple destination locales, and we can only return one. We could keep the `locale` and `dest` query parameters untouched and introduce a `return-locale` variable instead, which decided the locale of the response. But it is harder to grasp what is used as source, and it is inconsistent to me, since the `locale` usually describes the locale of the response. And in order to get rid of the bug the API still has to be touched for the other bundles.

For these reasons I would accept the BC break for the sake of clarity and consistency. It is also documented in the `UPGRADE.md`, so there is a hint for people having implemented something like that. Nevertheless I am willing to change it using a `return-locale` query parameter instead, if @sulu/core-team prefers that.

nnatter (approved):

Issue #4987, opened 2020-01-17
Multiple default values do not work inside of blocks


alexander-schranz:
| Q | A
| --- | ---
| Sulu Version | 2.0.3
| Browser Version | -

# Actual Behavior

Currently the default value is only set on the last single_select.

<img width="518" alt="Bildschirmfoto 2020-01-17 um 15 00 50" src="https://user-images.githubusercontent.com/1698337/72617855-2c291e00-393a-11ea-8b8f-a15d0e61226f.png">

# Expected Behavior

Default value should be correct on both single_selects.

# Steps to Reproduce

<details>
<summary>Example Template</summary>

```xml
<?xml version="1.0" ?>
<template xmlns="http://schemas.sulu.io/template/template"
          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xmlns:xi="http://www.w3.org/2001/XInclude"
          xsi:schemaLocation="http://schemas.sulu.io/template/template http://schemas.sulu.io/template/template-1.0.xsd">

    <key>default</key>

    <view>pages/default</view>
    <controller>Sulu\Bundle\WebsiteBundle\Controller\DefaultController::indexAction</controller>
    <cacheLifetime>604800</cacheLifetime>

    <meta>
        <title lang="en">Default</title>
        <title lang="de">Standard</title>
    </meta>

    <properties>
        <property name="title" type="text_line" mandatory="true">
            <meta>
                <title lang="de">Titel</title>
                <title lang="en">Title</title>
            </meta>

            <params>
                <param name="headline" value="true"/>
            </params>

            <tag name="sulu.rlp.part"/>
        </property>

        <property name="url" type="resource_locator" mandatory="true">
            <meta>
                <title lang="de">Adresse</title>
                <title lang="en">Resourcelocator</title>
            </meta>

            <tag name="sulu.rlp"/>
        </property>

        <property name="description" type="text_editor">
            <meta>
                <title lang="en">Text</title>
                <title lang="de">Text</title>
            </meta>

            <tag name="sulu.search.field" role="description"/>
        </property>

        <block name="test" default-type="test">
            <types>
                <type name="test">
                    <properties>
                            <property name="position" type="single_select" colspan="6">
                            <meta>
                                <title lang="en">Position 1</title>
                                <title lang="nl">Positie 1</title>
                            </meta>
                            <params>
                                <param name="default_value" value="right"/>
                                <param name="values" type="collection">
                                    <param name="left">
                                        <meta>
[...]

PR #5576, opened 2020-10-28
Fix default-values for multiple-select in blocks

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes https://github.com/sulu/sulu/issues/4987
| License | MIT

#### What's in this PR?

Fixes the default-values in blocks. See https://github.com/sulu/sulu/issues/4987

An additional `value` attribute is introduced to save the state of the `value`. Previously the `value` was saved in the properties (`this.props`), but the properties are updated after the `default-value` is set for the next `single_select`, thus the `value` is not up to date during the rendering of the other components.

Now the new `value` attribute is manually updated before the next component is rendered.

#### Why?

The default value was only working for the last component.

#### Todo

 - [ ] Tests
 - [x] Rebase to 2.1
 - [ ] Check Imagemap

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
@@ -14,7 +14,7 @@ const MISSING_BLOCK_ERROR_MESSAGE = 'The "block" field type needs at least one t
 const BLOCK_PREVIEW_TAG = 'sulu.block_preview';
 
 @observer
-export default class FieldBlocks extends React.Component<FieldTypeProps<Array<BlockEntry>>> {
+class FieldBlocks extends React.Component<FieldTypeProps<Array<BlockEntry>>> {
     @observable value: Object;
 
     constructor(props: FieldTypeProps<Array<BlockEntry>>) {
@@ -27,7 +27,9 @@ export default class FieldBlocks extends React.Component<FieldTypeProps<Array<Bl
         const {defaultType, onChange, types, value} = this.props;
         const {types: oldTypes} = prevProps;
 
-        this.updateValue(value);
+        if (!equals(prevProps.value, value) ){
+            this.updateValue(value);
+        }
 
         if (!types || !oldTypes) {
             throw new Error(MISSING_BLOCK_ERROR_MESSAGE);
@@ -254,3 +256,5 @@ export default class FieldBlocks extends React.Component<FieldTypeProps<Array<Bl
         );
     }
 }
+
+export default FieldBlocks;
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
@@ -2,7 +2,7 @@
 import equals from 'fast-deep-equal';
 import jsonpointer from 'json-pointer';
 import React, {Fragment} from 'react';
-import {observable, action, toJS} from 'mobx';
+import {action, observable, toJS} from 'mobx';
 import {observer} from 'mobx-react';
 import BlockCollection from '../../components/BlockCollection';
 import type {BlockEntry} from '../../components/BlockCollection/types';
@@ -27,7 +27,7 @@ class FieldBlocks extends React.Component<FieldTypeProps<Array<BlockEntry>>> {
         const {defaultType, onChange, types, value} = this.props;
         const {types: oldTypes} = prevProps;
 
-        if (!equals(prevProps.value, value) ){
+        if (!equals(prevProps.value, value)){
             this.updateValue(value);
         }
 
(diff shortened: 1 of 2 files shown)


nnatter:
Reminder: Check if we need to do something similar for the `ImageMap` when merging this 🙂

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
```diff
@@ -2,7 +2,8 @@
 import equals from 'fast-deep-equal';
 import jsonpointer from 'json-pointer';
 import React, {Fragment} from 'react';
-import {toJS} from 'mobx';
+import {observable, action, toJS} from 'mobx';
```
danrot, line 5: Sort alphabetically.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
```diff
@@ -12,11 +13,24 @@ import FieldRenderer from './FieldRenderer';
...
+        super(props);
+
+        this.value = this.props.value;
+    }
+
     componentDidUpdate(prevProps: FieldTypeProps<Array<BlockEntry>>) {
         const {defaultType, onChange, types, value} = this.props;
         const {types: oldTypes} = prevProps;
 
+        if (!equals(prevProps.value, value) ){
```
danrot, line 30: Remove the space between the two closing brackets

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
```diff
@@ -57,6 +76,8 @@ export default class FieldBlocks extends React.Component<FieldTypeProps<Array<Bl
         const newValues = toJS(oldValues);
         jsonpointer.set(newValues[index], '/' + name, value);
 
+        this.updateValue(newValues);
```
danrot, line 79: Just for my understanding: This is absolutely necessary, right? The `componentDidUpdate` call you've also touched is not executed between multiple `onChange` calls, right?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/tests/FieldBlocks.test.js
```diff
@@ -898,6 +901,130 @@ test('Should call onFinish when the order of the blocks has changed', () => {
...
 
+test('Should set correct default values for multiple single_select in blocks', () => {
+    const formInspector = new FormInspector(new ResourceFormStore(new ResourceStore('test'), 'test'));
+
+    const types = {
+        default: {
+            title: 'Default',
+            form: {
+                position_center: {
+                    label: 'Position Left',
```
danrot, line 912: Rename to "Position Center"? Confused me a bit at first :see_no_evil:

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/tests/FieldBlocks.test.js
```diff
@@ -898,6 +901,130 @@ test('Should call onFinish when the order of the blocks has changed', () => {
...
+            defaultType="default"
+            formInspector={formInspector}
+            minOccurs={1}
+            onChange={changeSpy}
+            types={types}
+            value={observable([{type: 'default'}])}
+        />
+    );
+
+    fieldBlocks.find('Block').at(0).simulate('click');
```
danrot, line 1015: Out of curiosity: Is that click really necessary? Shouldn't the call below happen anyway? Because if it doesn't, there might be some edge case, which might cause some weird behavior.

danrot (changes requested):
I think that's a better solution than [the two I came up with](https://github.com/sulu/sulu/issues/4987#issuecomment-576152129) :+1:

So I am generally fine with that, still added a few minor issues resp. questions for better understanding.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
Prokyonn: Yep, we debugged through it and the `componentDidUpdate` is called only after the second `onChange` is called. Thus the `value` is not yet updated during the second `onChange` call.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/tests/FieldBlocks.test.js
Prokyonn: No the call does not happen. The blocks are collapsed and the rendering is only called after they are expanded.

danrot: Ok, when a new block is added, it is already opened anyway... So that should not be that bad. I guess the only thing that might not work as expected is if a new field is added to a block type, which is already used somewhere. The value would probably not be updated, until the block is expanded in the UI. But I think that this is an edge case, which does not hurt.

danrot (approved):

PR #5580, opened 2020-10-30
Fix MediaDataProvider if no security is configured

danrot:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | #5439 
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR fixes the signature of the `MediaDataProvider` constructor to also accept `null` for the `Security` class.

#### Why?

Because if no security is configured in `config/packages/security_website.yaml` the `Security` class is not available and therefore null. In that case the website currently breaks, if a SmartContent displaying media is on the page.

### DIFF
--- a/src/Sulu/Component/Media/SmartContent/MediaDataProvider.php
+++ b/src/Sulu/Component/Media/SmartContent/MediaDataProvider.php
@@ -45,7 +45,7 @@ class MediaDataProvider extends BaseDataProvider
         ArraySerializerInterface $serializer,
         RequestStack $requestStack,
         ReferenceStoreInterface $referenceStore,
-        Security $security,
+        Security $security = null,
         RequestAnalyzerInterface $requestAnalyzer,
         $permissions
     ) {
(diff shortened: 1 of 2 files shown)


src/Sulu/Component/Media/SmartContent/MediaDataProvider.php
```diff
@@ -45,7 +45,7 @@ public function __construct(
         ArraySerializerInterface $serializer,
         RequestStack $requestStack,
         ReferenceStoreInterface $referenceStore,
-        Security $security,
+        Security $security = null,
```
alexander-schranz, line 48: optional parameters in the middle is not longer allowed: https://php.watch/versions/8.0/deprecate-required-param-after-optional

So we would need to add it to the end or use `?Security $security`

alexander-schranz (changes requested):
Optional parameters should be at the end, else it will fail in the [next version of php](https://php.watch/versions/8.0/deprecate-required-param-after-optional). So maybe we just maek it nullable using `?Security $security,`

src/Sulu/Component/Media/SmartContent/MediaDataProvider.php
danrot: Woah, that's a heavy change. Gonna fix it.

alexander-schranz (approved):

PR #5581, opened 2020-10-30
Implement single_snippet_selection content-type

### LINKED ISSUE #5568: Add a single_snippet_selection content-type
# Problem description

Sulu already includes a `snippet_selection` content-type. In a lot of cases, it would be useful to allow for selecting only a single snippet in the administration interface. 

We already have a `single_category_selection`, `single_page_selection`, `single_contact_selection`, `single_account_selection` and `single_media_selection`. Therefore it would be nice to provide a `single_snippet_selection` too.

nnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #5568
| License | MIT
| Documentation PR | https://github.com/sulu/sulu-docs/pull/576

#### What's in this PR?

This PR implements a `single_snippet_selection` content type.

#### Why?

Sulu already includes a `snippet_selection` content-type. In a lot of cases, it would be useful to allow for selecting only a single snippet in the administration interface. See #5568

#### Example Usage

```xml
<property name="single_snippet_selection" type="single_snippet_selection">
    <meta>
        <title lang="en">Single Snippet Selection</title>
        <title lang="de">Schnipsel</title>
    </meta>

    <params>
        <param name="types" value="default"/>
        <param name="loadExcerpt" value="true"/>
    </params>
</property>
```

### DIFF
--- a/src/Sulu/Bundle/SnippetBundle/Content/SingleSnippetSelection.php
+++ b/src/Sulu/Bundle/SnippetBundle/Content/SingleSnippetSelection.php
@@ -54,7 +54,7 @@ class SingleSnippetSelection extends SimpleContentType implements PreResolvableC
         $resolvedSnippet = $this->resolveSnippet($property);
 
         if (null === $resolvedSnippet) {
-            return $this->defaultValue;
+            return null;
         }
 
         return $resolvedSnippet['content'];
--- a/src/Sulu/Bundle/SnippetBundle/Content/SingleSnippetSelection.php
+++ b/src/Sulu/Bundle/SnippetBundle/Content/SingleSnippetSelection.php
@@ -1,0 +1,107 @@
+<?php
+
+/*
+ * This file is part of Sulu.
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\SnippetBundle\Content;
+
+use Sulu\Bundle\SnippetBundle\Snippet\SnippetResolverInterface;
+use Sulu\Bundle\WebsiteBundle\ReferenceStore\ReferenceStoreInterface;
+use Sulu\Component\Content\Compat\PropertyInterface;
+use Sulu\Component\Content\Compat\Structure\PageBridge;
+use Sulu\Component\Content\PreResolvableContentTypeInterface;
+use Sulu\Component\Content\SimpleContentType;
+
+class SingleSnippetSelection extends SimpleContentType implements PreResolvableContentTypeInterface
+{
+    /**
+     * @var SnippetResolverInterface
+     */
+    private $snippetResolver;
+
+    /**
+     * @var ReferenceStoreInterface
+     */
+    private $snippetReferenceStore;
+
+    public function __construct(
+        SnippetResolverInterface $snippetResolver,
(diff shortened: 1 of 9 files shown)


src/Sulu/Bundle/SnippetBundle/Content/SingleSnippetSelection.php
```diff
@@ -0,0 +1,107 @@
...
+    ) {
+        $this->snippetResolver = $snippetResolver;
+        $this->snippetReferenceStore = $snippetReferenceStore;
+
+        parent::__construct('SingleSnippetSelection', null);
+    }
+
+    public function getContentData(PropertyInterface $property)
+    {
+        $resolvedSnippet = $this->resolveSnippet($property);
```
alexander-schranz, line 45: The single_snippet_selection should handle also the default snippets like the `snippet_selection` content type. Please have a look at the exist implementation of the Snippets Content Type to implement the same behaviour as there.

alexander-schranz (changes requested):

src/Sulu/Bundle/SnippetBundle/Content/SingleSnippetSelection.php
nnatter: are you sure? i dont think that feature is very helpful. especially because it is not implemented at all in the admin interface. also, if you want to implement that behaviour, you can just fallback to the default snippet in your template

nnatter: added the fallback mechanism. still not a fan of that feature 😇

alexander-schranz:
Think this should target the `master` branch as a new feature.

nnatter:
I think I could argue that this was forgotten when releasing 2.0 and therefore its more like a bugfix 😅

src/Sulu/Bundle/SnippetBundle/Content/SingleSnippetSelection.php
```diff
@@ -0,0 +1,132 @@
...
+
+        if (null === $resolvedSnippet) {
+            return $this->defaultValue;
```
alexander-schranz, line 57: Would not use `$this->defaultValue` here instead:

```suggestion
            return  null;
```

alexander-schranz (approved):

alexander-schranz (approved):

src/Sulu/Bundle/SnippetBundle/Content/SnippetContent.php
```diff
@@ -26,9 +26,6 @@
 use Sulu\Component\Content\ContentTypeExportInterface;
 use Sulu\Component\Content\PreResolvableContentTypeInterface;
 
-/**
- * ContentType for Snippets.
- */
 class SnippetContent extends ComplexContentType implements ContentTypeExportInterface, PreResolvableContentTypeInterface
```
TheCadien, line 29: Maybe its usefull to rename the `SnippetContent` Class to `SnippetSelection` for the uniformity. 
Think that should have only little effect on the end user, because the content type is still called "snippet_selection, but think thats cleaner =)

nnatter: yes that would be really nice to do, but unfortunately it is bc break. if somebody has extended that class, his application will break if we rename the class/file 😕

PR #5582, opened 2020-10-31
Check if segments key exists in webspace

plozmun:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

Fix for optional segment

#### Why?

Segments is a optional configuration but if segments is not set an error occurs
```
Notice: Undefined index: segments
```


plozmun:
Sorry, my bad. I had override excerpt template to add some custom parameters

PR #5583, opened 2020-10-31
Improve phpdoc in AdminBundle

BlackScorp:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | AdminBundle phpstan level 5
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Implemented together with @rogoit some typhint fixes acording to phpstan and increased the phpstan level from 1 to 5 in the AdminBundle

#### Why?

Its always better to have better sofware and also because of Hacktoberfest2020 event, if you merge the PR please add a label "hacktoberfest-accepted" so i can win a t-shirt

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Admin/AdminPool.php
+++ b/src/Sulu/Bundle/AdminBundle/Admin/AdminPool.php
@@ -26,7 +26,7 @@ class AdminPool
     /**
      * Adds a new admin.
      *
-     * @var Admin $admin
+     * @var Admin
      */
     public function addAdmin($admin)
     {
--- a/src/Sulu/Bundle/AdminBundle/Admin/AdminPool.php
+++ b/src/Sulu/Bundle/AdminBundle/Admin/AdminPool.php
@@ -19,14 +19,16 @@ class AdminPool
     /**
      * The array for all the admin-objects.
      *
-     * @var array
+     * @var Admin[]
      */
     private $pool = [];
 
     /**
      * Adds a new admin.
+     *
+     * @var Admin $admin
      */
-    public function addAdmin(Admin $admin)
+    public function addAdmin($admin)
     {
         $this->pool[] = $admin;
     }
--- a/src/Sulu/Bundle/AdminBundle/Admin/AdminPool.php
+++ b/src/Sulu/Bundle/AdminBundle/Admin/AdminPool.php
@@ -25,8 +25,6 @@ class AdminPool
 
     /**
      * Adds a new admin.
-     *
-     * @param Admin $admin
      */
     public function addAdmin(Admin $admin)
     {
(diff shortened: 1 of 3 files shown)


src/Sulu/Bundle/AdminBundle/Admin/AdminPool.php
```diff
@@ -25,10 +25,8 @@ class AdminPool
 
     /**
      * Adds a new admin.
-     *
-     * @param $admin
      */
-    public function addAdmin($admin)
+    public function addAdmin(Admin $admin)
```
nnatter, line 29: i think we cannot add typehints here because of backwards compatibility. if somebody extended this class and overwrote this method, his application will break because the method signature does not match anymore 😕

alexander-schranz: Aggree here we can from BC promise not add typehints here. This should be added to the phpdoc above:

```
     * @param Admin $admin
```

alexander-schranz (changes requested):

src/Sulu/Bundle/AdminBundle/Admin/AdminPool.php
```diff
@@ -19,14 +19,14 @@ class AdminPool
...
-     * @var array
+     * @var Admin[]
      */
     private $pool = [];
 
     /**
      * Adds a new admin.
      *
-     * @param $admin
+     * @var Admin
```
nnatter, line 29: i think this should be `@param Admin $admin`

src/Sulu/Bundle/AdminBundle/Admin/View/ViewRegistry.php
```diff
@@ -115,14 +123,14 @@ private function mergeViewOptions(array $views, string $parent = null)
...
-        /** @var View $parentView */
+        /** @var View[] $parentViews */
         $parentViews = \array_values(\array_filter($views, function(View $view) use ($parent) {
             return $view->getName() === $parent;
         }));
 
         $parentView = null;
         if (!empty($parentViews)) {
-            $parentView = $parentViews[0];
+            $parentView = \current($parentViews);
```
nnatter, line 133: not sure if it makes a difference in this case, but `current` does not always return the first element. i think we should use `reset` if it is necessary to change this

staabm: maybe just `list($parentView) = $parentViews;

nnatter: looks like `list($test)` will trigger a `Undefined variable` notice if the array is empty. Therefore I would use `reset` 🙂 
https://3v4l.org/Cg2bs

nnatter (approved):
Thanks for your effort!

PR #5589, opened 2020-11-03
Collect tags from types only if data is an array

danrot:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | #5434 
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR avoids collecting tags in the `AbstractFormStore`, if the passed data is not an array.

#### Why?

The code only works with arrays, and it throws an error if the passed data is of a different type. This didn't happen before, but started happening with the `image_map` field type now, since this is the first one making use of the `types` other than the blocks. The way this is implemented now, the system skips tags within the `image_map` field type. This is not an optimal solution, but the best I can come up with.

The problem is that now every property can make use of types, and we don't really have an idea where they are used. E.g. for the `image_map` field type we would have to loop over the `hotspots` property of the object it creates. For another field type it might have a different name, and another field type might not even use it in combination with an array. I guess we need to externalize that logic, in order to properly fix that, but for now that's the quickest fix to release 2.2.1.

#### To Do

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/AbstractFormStore.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/AbstractFormStore.js
@@ -1,5 +1,5 @@
 // @flow
-import {action, computed, observable, set, toJS, untracked, when} from 'mobx';
+import {action, computed, isObservableArray, observable, set, toJS, untracked, when} from 'mobx';
 import type {IObservableValue} from 'mobx';
 import jexl from 'jexl';
 import jsonpointer from 'json-pointer';
@@ -124,7 +124,11 @@ function collectTagPathsWithPriority(
             continue;
         }
 
-        if (types && Object.keys(types).length > 0 && data[key]) {
+        if (types
+            && Object.keys(types).length > 0
+            && data[key]
+            && (Array.isArray(data[key]) || isObservableArray(data[key]))
+        ) {
             for (const childKey of data[key].keys()) {
                 const childData = data[key][childKey];
                 pathsWithPriority.push(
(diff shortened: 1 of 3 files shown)


alexander-schranz:
@danrot The solution is okay for me as a fast one. A general solution I could think about is adding options to the `fieldRegistry` e.g.:

```js
fieldRegistry.add('image_map', ImageMap, { typesPath: 'hotspots' });
```

This would tell where to find the data of the types. What do you think about that?

danrot:
That would not solve all of the problems... As said, we don't even know for what the types are used. It might e.g. be that only one type can be selected at a time, and therefore the result is not even stored as array. Then just passing a path where we want the "array" to be, won't work either.

But apart from that I would say we create a separate issue for that, and continue the discussion there. Otherwise this information is more or less lost.

alexander-schranz:
@danrot Okay I'm not completely understand it but as said solution is okay for me at current state, can you create an issue about image map and tags not working now correctly?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/AbstractFormStore.js
```diff
@@ -124,7 +124,11 @@ function collectTagPathsWithPriority(
             continue;
         }
 
-        if (types && Object.keys(types).length > 0 && data[key]) {
+        if (types
+            && Object.keys(types).length > 0
```
nnatter, line 128: maybe we should test if the `type` is `block` rather than this? looks like the handling below is for blocks, right? 🤔

danrot: It would work for any other type as well, that saves the different type as an array. I think I would not be too restrictive here, since this is a part of the code you can't (really) touch as 3rd party developer.

alexander-schranz (approved):

Issue #5301, opened 2020-05-19
Paste into resourceLocator field does not filter out special characters


alexander-schranz:
| Q | A
| --- | ---
| Sulu Version | 2.0.6
| Browser Version | -

# Actual Behavior

When you example copy `/example/thank-you` or something like `Hello World /// Test` into the url field it will add that directly without filtering special characters out.

# Expected Behavior

As you can not type `/` or other characters you should also not be able to paste this characters into the url field.

# Steps to Reproduce

![paste-url-with-slash](https://user-images.githubusercontent.com/1698337/82315808-2ae12e00-99cc-11ea-9615-266e0598960b.gif)

# Possible Solutions

Filter out special characters also when pasted.

danrot:
Good catch! It is also questionable what happens if you type an umlaut in the URL field. Should nothing happen? Or should e.g. "ä" automatically be replaced by "ae"?

alexander-schranz:
@danrot thats a good question. I thought we are allowing `äöü` but currently it seems to end in a ResourceLocatorNotValidException atelast in language `en` and `de`.

kfrohwein:
In our case the string was "ResourceLocator '/20200714-reise-zum-nordkap–bis-zum-ende-elektrisch' is not valid" maybe it failed because of the leading slash.

danrot:
Fixed in #5591

Issue #5462, opened 2020-08-21
Problem saving pages when there is an uppercase letter in the URL field.


jordygroote:
# Actual Behavior

When there is a capital letter in the URL. We get a red error in the top of the CMS and can't save the page.
With no logic error message or hint on what to do.

# Expected Behavior

It saves the page or displays usefull error.

# Steps to Reproduce

See this video:

https://www.loom.com/share/080c6f426d5c4425b2785e1931ba164b

# Possible Solutions

Automaticly convert uppercase letters to lowercase letters.

Or

Display a usefull error message saying something like:
"There is a uppercase letter in your URl structure please change this to lowercase letter."

danrot:
Fixed in #5591

PR #5591, opened 2020-11-03
Replace invalid characters in resource locator

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes https://github.com/sulu/sulu/issues/5301 and https://github.com/sulu/sulu/issues/5462
| License | MIT

#### What's in this PR?

Fixes https://github.com/sulu/sulu/issues/5301 and https://github.com/sulu/sulu/issues/5462
Adds validation for the `ResourceLocator` input field. Upper case letters are replaced with lower case letters and slashes are replaced with dashes in `leaf` mode.

#### Why?

The input field of the `ResourceLocator` has unexpected behaviour when inserting strings with `/` and capital letters result in a validation error on the backend.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/ResourceLocator/ResourceLocator.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/ResourceLocator/ResourceLocator.js
@@ -17,8 +17,6 @@ type Props = {|
 |};
 
 const replacerMap = new Map([
-    // replace multiple dash with one
-    [/([-]+)/g, '-'],
     // remove dash before slash
     [/[-]+\//g, '/'],
     // remove dash after slash
@@ -29,6 +27,8 @@ const replacerMap = new Map([
     [/([/]+)/g, '/'],
     // replace spaces with dashes
     [/ /g, '-'],
+    // replace multiple dash with one
+    [/([-]+)/g, '-'],
     // remove special characters
     [/[^a-z0-9-_/]/g, ''],
 ]);
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/ResourceLocator/ResourceLocator.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/ResourceLocator/ResourceLocator.js
@@ -2,12 +2,14 @@
 import React from 'react';
 import {action, computed, observable} from 'mobx';
 import {observer} from 'mobx-react';
+import type {IObservableValue} from 'mobx';
 import Input from '../Input';
 import resourceLocatorStyles from './resourceLocator.scss';
 
 type Props = {|
     disabled: boolean,
     id?: string,
+    locale: IObservableValue<string>,
     mode: string,
     onBlur?: () => void,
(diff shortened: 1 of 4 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/ResourceLocator.js
```diff
@@ -145,6 +145,7 @@ class ResourceLocator extends React.Component<FieldTypeProps<?string>> {
                     <ResourceLocatorComponent
                         disabled={!!disabled}
                         id={dataPath}
+                        locale={formInspector.locale}
```
danrot, line 148: A test for that is misisng.

danrot (changes requested):

danrot:
I think spaces should also been replaced with dashes if they are directly typed in the resource locator. They also cause an error, and are also replaced when the server generates the resource locator.

danrot:
I just unintentionally found more errors like this... Also if you type in something like `'`, `!` or umlauts like `ä` it will not save the page... I think the problem is that we would have to reimplement our entire [`PathCleanup` class](https://github.com/sulu/sulu/blob/3f2225e3defcc8cfc218380c2e65d65b1084bf31/src/Sulu/Component/PHPCR/PathCleanup.php), and I am starting to wondering if this is the right way to address this. That would result in most of the key presses to be shown as something else (all special characters, uppercase characters, ...). Also, on the server we would replace multiple consecutive dashes to be a single dash only.

Maybe we should stop for a second and think about a completely different approach here? Maybe the characters not allowed should be simply deleted instead of replaced? So that the user is not able at all to type in an invalid character? Or if an invalid character is entered it is displayed, but we see an error message explaining that only certain characters are allowed? Maybe we should also consult an UX expert for that. I think what we have done on the server side in the `PathCleanup` class works quite well if we have some real text and need to convert it to a URL, but doing that on the fly while a user is typing feels confusing to me.

nnatter:
I think this will also fix https://github.com/sulu/SuluArticleBundle/issues/521 once it is merged 🙂

src/Sulu/Bundle/AdminBundle/Resources/js/components/ResourceLocator/ResourceLocator.js
```diff
@@ -2,18 +2,37 @@
...
+    locale: IObservableValue<string>,
     mode: string,
     onBlur?: () => void,
     onChange: (value: ?string) => void,
     value: ?string,
 |};
 
+const replacerMap = new Map([
+    // replace multiple dash with one
+    [/([-]+)/g, '-'],
```
danrot, line 21: I think the order here is important. That should probably be one of the last entries here. Because if I am copying now a text like "That's a /// test!" in the resource locator field, it will still result in multiple dashes.

danrot (changes requested):

danrot (approved):

PR #5592, opened 2020-11-03
Fix link inside of table cell in ckeditor

### LINKED ISSUE #5250: External and internal links in CKEditor get deleted in table
| Q | A
| --- | ---
| Sulu Version | 2.0.4
| Browser Version | Chrome Version 69.0.3497.81 (Official Build) (64-bit)

# Actual Behavior

When I create a block with a textfield and add a table with a link inside, everything goes right and I can save it.
Only when I re-open the same block it doesn't add the link to the text, so when I save it the link is gone.

# Expected Behavior

If you re-open the textblock the link inside the table should still be available

# Steps to Reproduce

1. Add a block with a text_field.
2. Create a table with items.
3. Make one of the items a Link.
4. Save the page.
5. Re-open the same block (the link(s) will be gone)
6. Change something inside the block. 
7. Save the page again.

nnatter:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #5250
| License | MIT

#### What's in this PR?

This PR adjusts the upcast conversion that is created in the`addLinkConversion` helper method to match the [code of the official `ckeditor5-link` package](https://github.com/ckeditor/ckeditor5/blob/5fe1d6394bbaf4aaf7285c64401296cb58642018/packages/ckeditor5-link/src/linkediting.js#L82-L94). The method is used by the `ExternalLinkPlugin` and `InternalLinkPlugin` for the ckeditor.

#### Why?

See #5250.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/CKEditor5/utils.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/CKEditor5/utils.js
@@ -1,13 +1,21 @@
 // @flow
 function addLinkConversion(editor: Object, tag: string, internalAttribute: string, tagAttribute: string) {
+    // implementation is inspired by the official ckeditor5-link package:
+    // https://github.com/ckeditor/ckeditor5/blob/v23.0.0/packages/ckeditor5-link/src/linkediting.js
+
     editor.model.schema.extend('$text', {allowAttributes: internalAttribute});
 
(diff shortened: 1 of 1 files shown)


danrot:
Interesting, I thought I tasted mimicking that file already, but it didn't work for me :thinking: But I tested it and it works for me too :+1:

danrot (approved):

PR #5599, opened 2020-11-04
Fix sulu_content_load for none pages

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes -
| Related issues/PRs | -
| License | MIT

#### What's in this PR?

Fix sulu_content_load for none pages.

#### Why?

As reported by @patrickpotthoff `sulu_content_load` can be used to load articles or I think also snippets and currently as they don't have a webspacekey it will fail.

> request.CRITICAL: Uncaught PHP Exception Error: "Call to a member function getSecurity() on null" at /var/www/example.org/releases/16/vendor/sulu/sulu/src/Sulu/Bundle/WebsiteBundle/Twig/Content/ContentTwigExtension.php line 116 {"exception":"[object] (Error(code: 0): Call to a member function getSecurity() on null at /var/www/example.org/releases/16/vendor/sulu/sulu/src/Sulu/Bundle/WebsiteBundle/Twig/Content/ContentTwigExtension.php:116)"} []

#### To Do

- [ ] Add Test
- [ ] Adopt to check permission only for page structure

### DIFF
--- a/src/Sulu/Bundle/WebsiteBundle/Twig/Content/ContentTwigExtension.php
+++ b/src/Sulu/Bundle/WebsiteBundle/Twig/Content/ContentTwigExtension.php
@@ -112,24 +112,28 @@ class ContentTwigExtension extends AbstractExtension implements ContentTwigExten
                 $locale
             );
 
-            $targetWebspace = $this->webspaceManager->findWebspaceByKey($contentStructure->getWebspaceKey());
-            $security = $targetWebspace->getSecurity();
-            $system = $security ? $security->getSystem() : null;
+            $targetWebspaceKey = $contentStructure->getWebspaceKey();
 
-            if ($targetWebspace->hasWebsiteSecurity()
-                && $this->securityChecker
-                && !$this->securityChecker->hasPermission(
-                    new SecurityCondition(
-                        PageAdmin::SECURITY_CONTEXT_PREFIX . $contentStructure->getWebspaceKey(),
-                        $locale,
-                        SecurityBehavior::class,
-                        $uuid,
-                        $system
-                    ),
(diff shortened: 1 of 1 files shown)


src/Sulu/Bundle/WebsiteBundle/Twig/Content/ContentTwigExtension.php
```diff
@@ -112,24 +112,28 @@ public function load($uuid)
...
-                        $uuid,
-                        $system
-                    ),
-                    PermissionTypes::VIEW
-                )
-            ) {
-                return null;
+            $targetWebspaceKey = $contentStructure->getWebspaceKey();
+
+            if ($targetWebspaceKey) {
```
alexander-schranz, line 117: Maybe we should here also check if `contentStructure` is from the correct structure type `page`, else the security context could be called with a none page uuid which just returns a webspace key.

alexander-schranz: So it should maybe be something like instead of check for `getWebspaceKey`:

```
if ('page' !== $contentStructure->getType()) {
    return $this->structureResolver->resolve($contentStructure);
}
```

plozmun:
Hi! This patch works for me. Is there any workaround meanwhile this is published?

alexander-schranz:
@plozmun I'm currently short in time. I think a better solution revert this change and do what I described in https://github.com/sulu/sulu/pull/5599/files#r517475689. Do you want to take the PR over?

alexander-schranz:
closed in favor of #5610

PR #5600, opened 2020-11-04
Fix add request format to redirect url

### LINKED ISSUE #5598: Redirects do not respect request format
| Q | A
| --- | ---
| Sulu Version | 2.2.1
| Browser Version | Chrome 86.0.4240.183

# Actual Behavior

Currently if you fetch a page in JSON format, a redirect response will not respect the request format (e.g. .json)

# Expected Behavior

I expect it to redirect including the request format.

# Steps to Reproduce

I have a page with the slug `test-new`. I also have an old route `test-old`.
`http://sulu.test/test-old.json` will then redirect to `http://sulu.test/test-new` without the .json format.

image

alexandersch:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #5598 
| License | MIT

#### What's in this PR?

I've implemented a fix for issue #5598. I have appended the requestFormat in the getRedirectRoute method.

### DIFF
--- a/src/Sulu/Bundle/WebsiteBundle/Routing/ContentRouteProvider.php
+++ b/src/Sulu/Bundle/WebsiteBundle/Routing/ContentRouteProvider.php
@@ -76,7 +76,7 @@ class ContentRouteProvider implements RouteProviderInterface
     {
         $collection = new RouteCollection();
 
-        if ('' === $request->getRequestFormat()) {
+        if ('' === $request->getRequestFormat(null)) {
             return $collection;
         }
 
--- a/src/Sulu/Bundle/WebsiteBundle/Routing/ContentRouteProvider.php
+++ b/src/Sulu/Bundle/WebsiteBundle/Routing/ContentRouteProvider.php
@@ -232,15 +232,15 @@ class ContentRouteProvider implements RouteProviderInterface
      */
     protected function getRedirectRoute(Request $request, $url)
     {
-        $requestFormat = $request->getRequestFormat();
-        $format = 'html' !== $requestFormat ? '.' . $requestFormat : '';
+        $requestFormat = $request->getRequestFormat(null);
+        $formatSuffix = $requestFormat ? '.' . $requestFormat : '';
 
         // redirect to linked page
         return new Route(
             $this->decodePathInfo($request->getPathInfo()),
             [
                 '_controller' => 'SuluWebsiteBundle:Redirect:redirect',
-                'url' => $url . $format,
+                'url' => $url . $formatSuffix,
             ]
         );
     }
--- a/src/Sulu/Bundle/WebsiteBundle/Routing/ContentRouteProvider.php
+++ b/src/Sulu/Bundle/WebsiteBundle/Routing/ContentRouteProvider.php
@@ -232,12 +232,15 @@ class ContentRouteProvider implements RouteProviderInterface
      */
     protected function getRedirectRoute(Request $request, $url)
     {
+        $requestFormat = $request->getRequestFormat();
+        $format = 'html' !== $requestFormat ? '.' . $requestFormat : '';
+
         // redirect to linked page
         return new Route(
             $this->decodePathInfo($request->getPathInfo()),
             [
                 '_controller' => 'SuluWebsiteBundle:Redirect:redirect',
-                'url' => $url,
+                'url' => $url . $format,
             ]
         );
     }
--- a/src/Sulu/Bundle/WebsiteBundle/Tests/Unit/Sulu/Bundle/WebsiteBundle/Routing/ContentRouteProviderTest.php
+++ b/src/Sulu/Bundle/WebsiteBundle/Tests/Unit/Sulu/Bundle/WebsiteBundle/Routing/ContentRouteProviderTest.php
@@ -528,7 +528,7 @@ class ContentRouteProviderTest extends \PHPUnit_Framework_TestCase
             ->willThrow(new ResourceLocatorMovedException('/new-test', '123-123-123'));
 
         $request = new Request(
-            [], [], ['_sulu' => $attributes->reveal(), '_format'=> 'json'], [], [], ['REQUEST_URI' => \rawurlencode('/de/qwertz/')]
+            [], [], ['_sulu' => $attributes->reveal(), '_format' => 'json'], [], [], ['REQUEST_URI' => \rawurlencode('/de/qwertz/')]
         );
 
         // Test the route provider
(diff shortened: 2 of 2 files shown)


src/Sulu/Bundle/WebsiteBundle/Routing/ContentRouteProvider.php
```diff
@@ -282,7 +282,7 @@ protected function getRedirectRoute(Request $request, $url)
             $this->decodePathInfo($request->getPathInfo()),
             [
                 '_controller' => 'sulu_website.redirect_controller:redirectAction',
-                'url' => $url,
+                'url' => $url . '.' . $request->getRequestFormat(),
```
alexander-schranz, line 285:
```
$request->getRequestFormat()
```

Will return `html` and `.html` should not be added to the redirect. We need to check for that case.

alexander-schranz (changes requested):
@alexandersch Thank you for working on this. As this seems to be a bug which also exist in older version it should target the lowest supported versions. We currently making bugfix release for `1.6`, `2.1` and `2.2`. As I think this is also the case for `1.6` this should target the `1.6` Branch.

Its also needed as commented that `.html` is not added as `.html` is always without the format in sulu.

alexandersch:
> @alexandersch Thank you for working on this. As this seems to be a bug which also exist in older version it should target the lowest supported versions. We currently making bugfix release for `1.6`, `2.1` and `2.2`. As I think this is also the case for `1.6` this should target the `1.6` Branch.

If I rebase onto release/1.6 then I get a conflict on the _controller name. The lowest I can get is release/2.0
I can make a separate PR for the 1.6 release, if you want?

I have implemented a check around html format.

src/Sulu/Bundle/WebsiteBundle/Routing/ContentRouteProvider.php
alexandersch: I have implemented a check around html format.

alexander-schranz:
If you create the `1.6` pull request we will fix the conflict when merging it to the `2.1` branch. For your info with the release of `2.2` we will not longer create bugfixes for `2.0` Branch so currently maintained versions are:

 - `1.6` - supported for bugfixes until `3.0` is released
 - `2.1` - supported for bugfixes until `2.3` is released
 - `2.2` - supported for bugfixes until `2.4` is released
 - `master` - branch for new features for `2.3`

It would also be good to have a test for this so it does not get accidentally removed.

src/Sulu/Bundle/WebsiteBundle/Routing/ContentRouteProvider.php
```diff
@@ -237,12 +237,14 @@ private function checkResourceLocator($resourceLocator, $resourceLocatorPrefix)
      */
     protected function getRedirectRoute(Request $request, $url)
     {
+        $format = 'html' !== $request->getRequestFormat() ? '.' . $request->getRequestFormat() : '';
```
alexander-schranz, line 240: You never know which process a function is triggering or could trigger in future, so I would avoid a duplicated call of `$request->getRequestFormat()` and create a variable for it:

```
$requestFormat = $request->getRequestFormat();
$format = 'html' !== $requestFormat ? '.' . $requestFormat : '';
```

alexandersch: Changed and rebased onto 1.6

alexandersch:
- [ ] Write a test

alexandersch:
I am going to make a separate PR based off release/2.1. because I need a patch in our project.

src/Sulu/Bundle/WebsiteBundle/Routing/ContentRouteProvider.php
```diff
@@ -232,12 +232,15 @@ private function checkResourceLocator($resourceLocator, $resourceLocatorPrefix)
...
+        $requestFormat = $request->getRequestFormat();
+        $format = 'html' !== $requestFormat ? '.' . $requestFormat : '';
+
         // redirect to linked page
         return new Route(
             $this->decodePathInfo($request->getPathInfo()),
             [
                 '_controller' => 'SuluWebsiteBundle:Redirect:redirect',
-                'url' => $url,
+                'url' => $url . $format,
```
nnatter, line 243: have not tested this, but maybe we could also set `_format` similar to `_controller`? https://symfony.com/doc/current/routing.html#special-parameters

maybe we would not need to handle the `html` case then because symfony would do that for us. but not sure about this

alexandersch: Ah, I could try this of course. I'll let you know.

alexandersch: I've tried adding `_format` and removing the format from the `url`, but unfortunately it does not work as expected. It does not add `.json` to the redirected URL. I would suggest leaving it like it is, unless you guys can come up with a better solution.

src/Sulu/Bundle/WebsiteBundle/Routing/ContentRouteProvider.php
```diff
@@ -232,8 +232,8 @@ private function checkResourceLocator($resourceLocator, $resourceLocatorPrefix)
      */
     protected function getRedirectRoute(Request $request, $url)
     {
-        $requestFormat = $request->getRequestFormat();
-        $format = 'html' !== $requestFormat ? '.' . $requestFormat : '';
+        $requestFormat = $request->getRequestFormat(null);
+        $formatSuffix = $requestFormat ? '.' . $requestFormat : '';
```
alexander-schranz, line 236: @nnatter I'm not sure with your change this will add `.html` and we should never redirect to a `.html` url as that should always be stripped.

nnatter: no i tested that - `$request->getRequestFormat(null)` will return `null` if the url does not contain any format.

even if you access an url with `.html` in the end (eg `http://sulu.localhost:8000/new-test.html`), the `RedirectExceptionSubscriber` will remove that before this method is called 🙂

alexander-schranz: okay never saw the `RedirectExceptionSubscriber` 🙈

src/Sulu/Bundle/WebsiteBundle/Routing/ContentRouteProvider.php
```diff
@@ -76,7 +76,7 @@ public function getRouteCollectionForRequest(Request $request)
     {
         $collection = new RouteCollection();
 
-        if ('' === $request->getRequestFormat()) {
+        if ('' === $request->getRequestFormat(null)) {
```
nnatter, line 79: there is a bug in older symfony versions that caches the default value, even if you pass a different default value in the next call. thes was fixed in `2.8.18`: https://github.com/symfony/http-foundation/commit/d45f940f3b55a0c99eae8482c726be264ad1c8dc

alexander-schranz (approved):

nnatter:
@alexandersch Thanks for working on this!

Issue #4995, opened 2020-01-21
Choose different presentations on a block


danrot:
# Problem description

Sometimes a content manager wants to choose a different presentation on blocks, but keep the content exactly the same. This often only changes e.g. the background color or some other styling of the block on the website.

Currently there are two different ways of achieving this:

1. **You could create two different block types** and maybe use xincludes to avoid duplicating this information in the XML files.
2. **Add another property to the template** and use its value from the `content` variable in twig.

Both options have some drawbacks, because they are cumbersome to implement, and it somehow breaks our distinction between `content` (actual content) and `view` (how is the content displayed on the website).

# Proposed solution

In a separate section, e.g. an overlay containing multiple settings for the block, it should be possible to choose from a predefined list of values which presentation should be used from a dropdown.

This option should be passed to the `view` variable in our twig template, which will allow the frontend developer to choose how to present this block. This is very similar to the [`present_as` option for the smartcontent](http://docs.sulu.io/en/2.0/reference/content-types/smart_content.html#parameters). The presentation should be configurable on a type level in the XML file:

```xml
<block name="blocks" default-type="editor" minOccurs="0">
    <types>
        <type name="editor">
            <params>
                <param name="present_as" type="collection">
                    <param name="two">
                        <meta>
                            <title lang="en">Two columns</title>
                        </meta>
                    </param>
                    <param name="one">
                        <meta>
                            <title lang="en">One column</title>
                        </meta>
                    </param>
                </param>
            </params>
            <properties>
                <property name="article" type="text_editor"></property>
            </properties>
        </type>
        <type name="editor_image">
            <params>
                <param name="present_as" type="collection">
                    <param name="dark">
                        <meta>
                            <title lang="en">Dark</title>
                        </meta>
                    </param>
                    <param name="light">
                        <meta>
                            <title lang="en">Light</title>
                        </meta>
                    </param>
                </param>
            </params>
            <properties>
                <property name="article" type="text_editor" colspan="9"></property>
            </properties>
        </type>
    </types>
</block>
```

danrot:
Two questions:

1. Should simply the first value be the default value?
2. And if we put the `present_as` parameter on the `type` tag, it means that this is the only field in the block settings that changes when the block changes its type. Is that OK from a UX perspective?

alexander-schranz:
@danrot when switching and the target type does not have the current type in it, it should use the first type, else it should stay as the same type, would be my expected behaviour. And yes the first type should be returned if none is set in phpcr.

PR #5608, opened 2020-11-06
Add scheduling of blocks

### LINKED ISSUE #4994: Configure when a block should be shown on the website
# Problem description

If e.g. a product (stored as a block on a page) should be featured on the website for a certain period of time (something like the 25th of January from 1pm to 8pm), then the content manager has to manually add and delete the block exactly at these times.

# Proposed solution

The solution proposed at #4964 would already help with that, because then the content manager would only have to change a toggler, which would be less work than creating an entire block. However, the content manager still has to make this change manually.

Therefore we would suggest to also add another option, which allows to define at which times the block should be shown. What comes to mind are the following options (ordered by my personal priority):

- **Complete start and end date** - Two datetime fields, if the current time is in between these two times the block will be shown on the website
- **Daily times** - Two time fields, the block will be shown every day at the given times
- **Weekly** - Select from a dropdown on which days of the week the blog should be shown along with two time fields
- **Monthly** - Select from a dropdown on which days of the month the blog should be shown along with two time fields. Mind that we have to also find solution for e.g. the last day of the month, since this is differing from month to month.
- **Yearly** - Probably very hard to configure... Maybe it is also not necessary, because people only plan in advance for a full year not too often?

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #4994 
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR will add the possibility to schedule when a block should be displayed.

#### BC Breaks/Deprecations

None yet

#### To Do

- [x] Add UI to configure scheduling in block settings
- [x] Store the data in PHPCR
- [x] Filter output of blocks on website by configured schedule by not passing them to the twig template
- [x] Calculate the correct cache expiration and set it on the response
- [ ] Add skin to make blocks look like in schedules design
- [x] Make blocks non-collapsable
- [x] Make blocks non-movable (#5652)
- [x] Add missing translations in `page_block_settings.xml`
- [x] Time in `WeeklySchedule` does not send data correctly to server (the time field returns a date object, which probably should not be the case)
- [x] Extract multiple "strategies" from `BlockContentType` to split up segments, target groups and scheduling?
- [x] How to handle time scheduling feature in preview?
- [x] Check "weekly" with time spanning multiple days (e.g. 23:00-01:00)
- [x] Doublecheck timezone behavior
- [ ] Dark skin for date picker
- [x] Delete Schedules field type
- [ ] Think about deleting newly introduced date functions


danrot:
The UI is already working, and it is also storing the data in PHPCR. Now two more tasks are left to be doing, whereby I'd like to discuss them:

### Hiding blocks on the website

The idea is that blocks are not even passed to the twig template. There is already [some code that hides the blocks in similar situations](https://github.com/sulu/sulu/blob/11fdd6d8de723374529bc0b3e4bfeecd1f6c28aa/src/Sulu/Component/Content/Types/BlockContentType.php#L377-L404), so that's probably where the time scheduling should also be handled.

The main idea is, that the current time has to be between the start and end time of one of the defined schedules for each block (including checking other constraints, like the day of the week for the weekly option). If the block does not have any schedule defined that includes the current date, then it should be skipped. The thing is that we could inline that like the checks for the `hidden` field, the target groups and the segments, but that would mean that it would be hard to extend that functionality (although adding some stuff to the [`page_block_settings` form](https://github.com/sulu/sulu/blob/11fdd6d8de723374529bc0b3e4bfeecd1f6c28aa/src/Sulu/Bundle/PageBundle/Resources/config/forms/page_block_settings.xml) is quite easy).

Extracting all these functionality to some generic interface would allow to make the code in the `BlockContentType` easier to understand and make the extension of the block settings overlay easier. I'd be fine with leaving that out for now, and inline that code in the `BlockContentType` for now. But I've already put some thought into this, so I am writing that down as well:

Every setting we have done so far for the blocks, is about hiding some of them. Therefore it might make sense to include a `BlockSkipperInterface` (not really happy with the name, but couldn't come up with something better):

```php
interface BlockPropertyTypeSkipper {
    public function shouldBeSkipped(): bool;
}
```

So [the code currently skipping the blocks](https://github.com/sulu/sulu/blob/11fdd6d8de723374529bc0b3e4bfeecd1f6c28aa/src/Sulu/Component/Content/Types/BlockContentType.php#L377-L404) would be replaced by multiple implementations of the above interface. But that would not be enough, since #4995 is also left to be implemented. While also being a block setting, it does not skip blocks, but just set a variable in the `view` twig variable. In that special case it might make sense to just include all the view data from the `page_block_settings` form, but eventually there might be more stuff we want to add. Maybe it would also make sense to include an interface like this:

```php
interface BlockPropertyTypeViewModifier {
    public function modifyBlock(array &$block): void; // or pass by value and return a new array?
}
```

Implementations of this interface would be called in the [`getViewData` method of the `BlockContentType`]. A similar (or the same) interface could exist for the `getContentData` method as well.

The reason why I am mentioning this now, is that it would be nice to have a single interface for both the previous use cases, because otherwise it feels like we are going to introduce far too many interfaces. But I couldn't come up with a good one... I could imagine a generic `BlockPropertyTypeModifier` which looks like that:

```php
interface BlockPropertyEnhancer {
    public function enhanceBlock(BlockPropertyTypeInterface $blockPropertyType): ?BlockPropertyType;
}
```

Returning `null` here would mean the block gets skipped (calling `array_filter` afterwards would be necessary), the exact same block could be returned, or the enhancer returns a modified block. So this interface would be a replacement of the above `BlockPropertyTypeViewModifier` and `BlockPropertyTypeSkipper` suggestions. However, this suggestion also does not feel very nice, because the implementations would have to fiddle around with the `BlockPropertyType` variable, where it is probably not so easy to add arbitrary data (because later the [`getViewData` of each content type in the block is called](https://github.com/sulu/sulu/blob/11fdd6d8de723374529bc0b3e4bfeecd1f6c28aa/src/Sulu/Component/Content/Types/BlockContentType.php#L340), which means we would have to introduce that data in some kind of dummy content type).

Maybe introducing interfaces like this makes only sense as a part of a bigger refactoring of the blocks. So I'd be fine with hardcode also the time schedules for now, and wait until an even better design for these interfaces emerge.

### Set the correct cache expiration

What is also missing in this PR, is the correct handling of caching. Basically we would have to find the next time some thing changes in any of the schedules, i.e. when is the closest start or end time of any time schedule for any blocks on this page.

The only real question is where this is done. The most straight forward approach would be to add this functionality to the [`CacheLifetimeEnhancer`](https://github.com/sulu/sulu/blob/11fdd6d8de723374529bc0b3e4bfeecd1f6c28aa/src/Sulu/Bundle/HttpCacheBundle/CacheLifetime/CacheLifetimeEnhancer.php#L46-L69). But that means it has to loop over all blocks of the currently loaded structure again, and find the closest date that changes something. Additionally this approach means that other content types can't have an influence on the cache lifetime, unless it is integrated in Sulu itself, or the `CacheLifetimeEnhancer` is overriden, which is not a nice extension point.

So I thought instead of doing that in the `CacheLifetimeEnhancer`, we should introduce a generic possibility to influence the caching. An idea that came to my mind is an interface like this:

```php
interface CacheLifetimeStore {
    public function setCacheLifetime($cacheLifetime: int): void;
    public function getCacheLifetime(): int;
}
```

The trick here would be that anything in Sulu can call the `setCacheLifetime` method of this interface, but the store only remembers the soonest. Then the `getCacheLifetime` method would return the correct cache lifetime at the very end, when we set the cache lifetime in a `ResponseListener`.

The advantages would be that we could call that method for every block without thinking about it, and the internals of the `CacheLifetimeStore` would make sure that the correct lifetime will be delivered at the very end. Additionally this class could also be used in other content types, because it is not tied to the blocks at all.

wachterjohannes:
@danrot "Set the correct cache expiration" sounds like a good idea - also because we could reuse this for other things in customer projects

"Hiding blocks on the website" - i would also devide the logic into several services to keep it easier to extend (extend the block settings form and add a "BlockPropertyTypeSkipper" service is quite straight forward to do this. The second thing of this topic isn't 100% clear to me. First of all `modifyBlock` should return the newly created (or extended) array. but why doesnt we inject all the settings data into the view variable? i would this also this would make it easier to extend from outside (project world). and i dont think that should be a problem

danrot:
@wachterjohannes I have even mentioned above, that we could just include all the `view` variable in my comment above, but I am already taking that thought a bit further, to try making that design a bit more future-proof.

If we just implement the `BlockPropertyTypeSkipperInterface` it already falls apart, as soon as one of the settings wants to do more than just hiding a block or rely on the value in the `view` variable. An interface that would allow to modify the existing or return a new block would be much more powerful, and allow to implement more use cases.

However, that is not easy to implement because we have a `getContentData` and `getViewData` method, which both work on top of what the `prepareData` returns. That means either the `BlockPropertyEnhancer` above works with a not so comfortable data structure (requiring to make use of a dummy content type in certain situations) or we are introducing even more interfaces than just one. Both of these options don't seem very elegant, so I am wondering if we should bother about that now, or fix that when the entire `BlockContentType` is refactored, so that an interface like the `BlockPropertyEnhancer` could be implemented more easily.

danrot:
#### Caching

The `CacheLifetimeStore` is maybe not the best idea, because there might be subrequests using the same store. Instead we have to store that lifetime on the `Request` object somehow. Then we are not talking about a store anymore, so we have to come up with a better name.

#### Hiding a blocks

We implement only the skip part for now, because skipping and enhancing the block (if ever implemented) should still be separated, since if we skip the block, the data of this block (like smart content, links, ...) do not need to be resolved.

Also, we are renaming the interface:

```php
interface BlockSkipperInterface {
    public function shouldBeSkipped(Block $block): bool;
}
```

danrot:
I have now introduced the `BlockSkipperInterface`, which works so far. So I've started implementing a `ScheduleBlockSkipper`, that should skip the block if none of the defined schedules match the current time. I've had some problems with testing this code, since it requires the `ClockMock` functionality of the PHPUnit Bridge of Symfony.

However, while discussing that issue, we realized that we might also want to support setting the preview to a certain time to enable the content manager to test what has been configured. We imagined something like a timeline or a datepicker in the preview toolbar, that allows to set the time manually.

If we implement something like this, it would kind of solve my problems with the tests, because it means that we cannot use something like `new DateTime` in the `ScheduleBlockSkipper`, because then it would not be possible to implement that preview feature. Leaving the UI of the new feature aside, we would have to make sure that the that we can pass a query parameter containing the set time in the request to the [`PreviewController`](https://github.com/sulu/sulu/blob/11fdd6d8de723374529bc0b3e4bfeecd1f6c28aa/src/Sulu/Bundle/PreviewBundle/Controller/PreviewController.php#L87-L102). That raises a two questions to me right now:

1. The `PreviewController` would have to accept another query parameter like `targetGroup` or `segment`, called `datetime` or something like that. If we do it the way it was done until now, we would have to add another parameter to [a few methods of the `PreviewInterface`](https://github.com/sulu/sulu/blob/11fdd6d8de723374529bc0b3e4bfeecd1f6c28aa/src/Sulu/Bundle/PreviewBundle/Preview/Preview.php#L89-L95). This interface in general feels a bit brittle, since we have to change it everytime we add a new feature that somehow affects the rendering of the website. Should we still add this one parameter again, or take some time to think about a better solution?
2. I guess ultimately we have to set the "current time of the website" somehow on the request like the [target group header in the `PreviewRenderer`](https://github.com/sulu/sulu/blob/11fdd6d8de723374529bc0b3e4bfeecd1f6c28aa/src/Sulu/Bundle/PreviewBundle/Preview/Renderer/PreviewRenderer.php#L178). Then we could introduce a `DateTimeRequestProcessor` which is called by the [`RequestAnalyzer`](https://github.com/sulu/sulu/blob/11fdd6d8de723374529bc0b3e4bfeecd1f6c28aa/src/Sulu/Component/Webspace/Analyzer/RequestAnalyzer.php). The question is if we pass that information somehow to the `BlockSkipperInterface`, so that it is available in the `ScheduleBlockSkipper`, or if the `ScheduleBlockSkipper` will get the `RequestAnalyzer` injected. I think in other similar situations we've decided to go with the injection, to keep the interface simple, and we avoid the need to have more interface changes in the future. Therefore I would go with that variant again.

danrot:
@wachterjohannes @nnatter @alexander-schranz The last commit is about the cache handling, would be nice if you could check if that is kind of ok (code is not finished/polished yet, but that would be the general idea I guess).

src/Sulu/Bundle/HttpCacheBundle/CacheLifetime/CacheLifetimeEnhancer.php
```diff
@@ -33,14 +33,21 @@ class CacheLifetimeEnhancer implements CacheLifetimeEnhancerInterface
...
+     * @var CacheLifetimeRequestEnhancer
+     */
+    private $cacheLifetimeRequestEnhancer;
+
     public function __construct(
         CacheLifetimeResolverInterface $cacheLifetimeResolver,
         $maxAge,
-        $sharedMaxAge
+        $sharedMaxAge,
+        CacheLifetimeRequestEnhancer $cacheLifetimeRequestEnhancer
```
alexander-schranz, line 45: For bc this should be:

```suggestion
       ?CacheLifetimeRequestEnhancer $cacheLifetimeRequestEnhancer = null
```

and throw a deprecation message when not set.

src/Sulu/Bundle/HttpCacheBundle/CacheLifetime/CacheLifetimeEnhancer.php
```diff
@@ -55,6 +62,12 @@ public function enhance(Response $response, StructureInterface $structure)
             $cacheLifetimeData['value']
         );
 
+        $requestCacheLifetime = $this->cacheLifetimeRequestEnhancer->getCacheLifetime();
```
alexander-schranz, line 65: for BC:

```suggestion
        $requestCacheLifetime = $this->cacheLifetimeRequestEnhancer ? $this->cacheLifetimeRequestEnhancer->getCacheLifetime() : null;
```

src/Sulu/Bundle/HttpCacheBundle/CacheLifetime/CacheLifetimeRequestEnhancer.php
```diff
@@ -0,0 +1,56 @@
...
+    {
+        $this->requestStack = $requestStack;
+    }
+
+    public function setCacheLifetime(int $cacheLifetime)
+    {
+        $request = $this->requestStack->getCurrentRequest();
+
+        if (!$request) {
+            return;
```
alexander-schranz, line 35: should we store it in a private variable when no current request is set? Not sure if the cache information could be interested in some kind of a CLI script 🤔 

/cc @wachterjohannes

src/Sulu/Component/Content/Types/BlockContentType.php
```diff
@@ -392,6 +400,42 @@ private function prepareData(PropertyInterface $property, callable $dataCallback
...
+                            $end = new \DateTime($schedule['end']);
+                            if ($end > $now && (!$soonestFutureSchedule || $end < $soonestFutureSchedule)) {
+                                $soonestFutureSchedule = $end;
+                            }
+                            break;
+                        case 'weekly':
+                            // TODO
+                            break;
+                    }
+                }
```
alexander-schranz: Should we move this into an own service for have a better way to create Unit Tests for this and reuse the logic?

alexander-schranz (commented):
@danrot Did have a look at the caching part and just some minor comments but at all it looks good from my side 👍 .

danrot:
@alexander-schranz That's the kind of stuff I omitted for now to get a working proof of concept. Going to address the comments anyway :slightly_smiling_face:

src/Sulu/Bundle/HttpCacheBundle/CacheLifetime/CacheLifetimeRequestEnhancer.php
danrot: Since the rest of the caching information is probably also not available on the CI I guess that is not a big deal. I guess if we want to make that available in the CLI, then we should make the final cache lifetime available, not this intermediate one.

src/Sulu/Component/Content/Types/BlockContentType.php
danrot: Was wondering about that as well... That's also one of the reasons I was skeptical about the `BlockSkipperInterface`. We could integrate that in the `ScheduleBlockSkipperInterface`, but that seems weird, because the name doesn't suggest that anything like that is happening. The other opportunity would be to create a new service (with or without interface), but it would be nice to have one interface that could also do stuff like that. Maybe we are talking about a `BlockIterator` here? :thinking:

danrot: When thinking about it, just this part we are talking about here, would be a nice use case for the `Visitor` pattern... We are just doing something based on that information, but do not want to add that to the data structure we are iterating over.

But I still think that it would be nice to combine that in a way, that we can skip a block (that's why I came up with a `BlockIterator`, because in Java you can call `remove` on an `Iterator`, but it does not really match here, because we want to extract some logic, for which the `BlockIterator` is not the right tool.

If we don't find a nice way to combine these behaviors, I would probably stick with the `BlockSkipperInterface` and also introduce a `BlockVisitor`. The visitor could do what is happening in this code passage, but we still would have to do some stuff twice (namely iterating over the `ScheduleBlockSkipper` and the `CacheLifeTimeBlockVisitor`).

danrot: I've read a bit about the `Visitor` pattern, and I think this is quite a good match.

Actually it seems to be built for traversing entire object structures, which would also make sense in our case. It would be very nice to be able to traverse over the entire metadata of a page that way. Then we could e.g. build a `CacheLifetimeVisitor` and then every content type, not only the blocks, could have an effect on the cache lifetime. But I don't think it makes a lot of sense to build that now, especially since I am quite sure that would need a few more adjustments.

The pattern consists of 2 important elements:

1. **`Element`**: Is a part of the object structure that will be traversed. Usually has an `accept` method, which takes the visitor and calls the appropriate function.
2. **`Visitor`**: Has a `visit` method for each element that can be passed, and will do something with them.

I think the `accept` method would not be really nice in our existing design, and according to the GoF book it also is not absolutely necessary. As an alternative approach it mentions that the object structure itself iterates over all elements and call the `Visitor` themselves. This means there are less possibilities to dispatch different `Visitor` implementations, but I think that is not necessary in our case. What we will do instead is to have multiple `Visitor` implementions.

That's what I've got so far today, and I wanted to jot it down somewhere. Will continue to work on that tomorrow and hopefully come up with a design.

danrot: The `Visitor` pattern would probably start showing of its full strength, if it would be used for the entire `StructureInterface`. If it would be used like that we could have a `CacheLifetimeVisitor`, `ContentDataVisitor` and a `ViewDataVisitor`, making the logic being split nicely and traversing the entire tree would only have to be implemented once.

But now probably is not the right time to do that, therefore I would suggest to implement something like this in the `BlockContentType` (omitting some code to keep the example brief):

```php
private function prepareData($property) {
    // ...
    for ($i = 0; $i < $blockProperty->getLength(); ++$i) {
        $blockPropertyType = $blockProperty->getProperties($i);
        foreach ($blockVisitor as $this->blockVisitors) { // strictly speaking the visitor could hold some internal state, so this could be a factory instead
            $blockPropertyType = $blockVisitor->visit($blockPropertyType);

            if (!$blockPropertyType) {
                continue 2;
            }
        }

        // read data from block
    }
}
```

In this scenario all the currently existing `BlockSkipperInterface` implementations can be replaced by a `BlockVisitor`, that returns `null` if the block should be skipped. Additionally a `CacheLifetimeBlockVisitor` could find the soonest future date something changes again, and call the `CacheLifetimeRequestEnhancer` for each block with that date.

Question is if we should introduce a `BlockVisitoryFactoryInterface` as well, to allow the `BlockVisitorInterface` implementations to have internal state as well. This would not be possible if the `BlockVisitorInterface` is registered as a service, since state would also be shared among different requests, etc.

alexander-schranz: Code looks fine from my side. I personally would not introduce any factory as this mostly make it more complicated to create this visitors and allow to autowire them in the symfony eco system. So would just go with the `BlockVisitorInterface` and avoid the introduction of the Factory.

danrot: In the Symfony ecosystem the `Factory` could also be autowired, so I don't think that should be a big problem.

The problem IMO is that the visitor pattern is generally understood that way, and if somebody wants to build a visitor holding some internal state in the future, we have already blocked that by this decision. But it's also ok for me to do it like this for now and live with these consequences. Would just mention it in a PHPDoc of the interface.

It's especially ok for me, since we are slightly changing the pattern already, because we are returning something from the `visit` function, which is not the case in the original pattern. There all `visit` functions would be returning nothing, and the visitor itself has another function returning the end result. That way it would maybe make more sense if we took the different approach I mentioned above (traversing the entire structure and also having a `ContentDataVisitor` and a `ViewDataVisitor`). But maybe this way we can see it as dipping our toes in the water :slightly_smiling_face:

danrot:
I thought that I will fix the preview first, because that might have an impact on the `BlockSkipperInterface` resp. the new `BlockVisitorInterface` (especially on how the current time is passed to these classes). But the question how exactly we are going to solve the `Preview` problem remains. There are two parts of this:

#### How should the `PreviewInterface` look like?

This boils down to two sub questions:
1. What parameters passed to the `PreviewInterface` are really mandatory and have to be passed every single time?
2. How do we pass optional parameters?

Let's have a look at the current [`PreviewInterface`](https://github.com/sulu/sulu/blob/2eb020fdbecd534c99e5476969c2193ee4c62356/src/Sulu/Bundle/PreviewBundle/Preview/PreviewInterface.php) first:

```php
interface PreviewInterface
{
    public function start(string $providerKey, string $id, string $locale, int $userId, array $data = []): string;

    public function stop(string $token): void;

    public function exists(string $token): bool;

    public function update(
        string $token,
        string $webspaceKey,
        array $data,
        ?int $targetGroupId,
        ?string $segmentKey
    ): string;

    public function updateContext(
        string $token,
        string $webspaceKey,
        array $context,
        ?int $targetGroupId,
        ?string $segmentKey
    ): string;

    public function render(
        string $token,
        string $webspaceKey,
        string $locale,
        ?int $targetGroupId,
        ?string $segmentKey
    ): string;
}
```

I think the `stop` and `exists` methods are ok, but the others probably include fields that are not mandatory, **especially if we also consider that the preview might be used in a non-website context** (then not even the `$webspaceKey` is mandatory!).

Considering this I would suggest a `PreviewInterface` like the following:

```php
interface PreviewInterface
{
    public function start(string $providerKey, string $id, int $userId, array $data = [], array $options = []): string;

    public function stop(string $token): void;

    public function exists(string $token): bool;

    public function update(string $token, array $data, array $options = []): string;

    public function updateContext(string $token, array $context, array $options): string;

    public function render(string $token): string;
}
```

The following parameters have been removed, because I would consider them mandatory in every case. These parameters would be move to the newly created `$options` parameter, that can take any number of parameters, and we could easily add more of them in the future, without having to adjust the interface every single time.

- `$locale`: There might be unlocalized entities using the preview
- `$webspaceKey`: The preview could be used for entities without any knowledge about webspaces
- `$targetGroupId`: It was already optional, so it should be moved to `$options`
- `$segmentKey`: It was already optional, so it should be moved to `$options`

I know we have shortly discussed reusing the `$context` instead of introducing a new `$options` variable, but to me it seems like the context already has some special handling in place, so I am not sure if it would be totally safe to reuse it for everything else as well. By splitting them only the `updateContext` function can care about the context and if we realize at some point, that what is written in the context can also be passed as an option instead, we can get rid of the `updateContext` function as a whole (As a short side note, the context is currently only used for the current template of the page).

#### How should the request to the `PreviewController` look like?

The above question also has impact on this one. By splitting the `PreviewInterface` into mandatory parameters and options `$options`, we can try to do the same split in the `PreviewController` and avoid [getting every parameter on its own](https://github.com/sulu/sulu/blob/2eb020fdbecd534c99e5476969c2193ee4c62356/src/Sulu/Bundle/PreviewBundle/Controller/PreviewController.php#L87-L94).

We could do that by sending the mandatory parameters still via query parameters, and the optional ones via the POST body. I have already checked, and the POST body already sends the data, but encapsulated behind a `data` property, so there is still room for other data, which I would like to use for these options. So the new update request for the preview could look like this:

```http
POST /admin/preview/update?id=<id>&provider=pages&token=<token>

{
    "options": {"locale": "en", "targetGroup": -1, "webspace": "example"},
    "data": {"title": "Test"}
}
```

This way the [`PreviewController::updateAction`](https://github.com/sulu/sulu/blob/2eb020fdbecd534c99e5476969c2193ee4c62356/src/Sulu/Bundle/PreviewBundle/Controller/PreviewController.php#L85-L103) could be simplified to something like this, leaving us with one less place to adjust if some new preview option is added:

```php
    public function updateAction(Request $request): Response
    {
        $provider = $this->getRequestParameter($request, 'provider', true);
        $id = $this->getRequestParameter($request, 'id', true);
        $token = $this->getRequestParameter($request, 'token', true);
        $data = $this->getRequestParameter($request, 'data', true);
        $options = $this->getRequestParameter($request, 'options', true);

        if (!$this->preview->exists($token)) {
            $token = $this->preview->start($provider, $id, $this->getUserId(), $options);
        }

        $content = $this->preview->update($token, $data, $options;

        return new JsonResponse(['content' => $content]);
    }
```

The only downside is, that we would have still to correctly read values from the `$options` array in or for the `PreviewRenderer`, giving us less type-safety. I would consider this good enough for now, because we could pass the entire `$options` array to the `PreviewRenderer`, leading o more stable interfaces. We can think in the future how we can make the `PreviewRenderer` more extensible.

wachterjohannes:
@danrot i am not sure if "especially if we also consider that the preview might be used in a non-website context" happens sometimes - we always renders the entities in a "webspace" to get the theme information and other things.

the rest is OK for me - also the interfaces

src/Sulu/Component/Content/Types/Block/HiddenBlockSkipper.php
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Component\Content\Types\Block;
+
+use Sulu\Component\Content\Compat\Block\BlockPropertyType;
+
+class HiddenBlockSkipper implements BlockSkipperInterface
```
nnatter, line 16: would it be possible to extract this functionality and include it into a `2.1.*` release? that would really simplify supporting this in the `SuluHeadlessBundle` 🙂

src/Sulu/Component/Content/Types/BlockContentType.php
```diff
@@ -374,32 +386,9 @@ private function prepareData(PropertyInterface $property, callable $dataCallback
...
-                }
-
-                if (isset($blockPropertyTypeSettings['target_groups_enabled'])
-                    && $blockPropertyTypeSettings['target_groups_enabled']
-                    && isset($blockPropertyTypeSettings['target_groups'])
-                    && $this->targetGroupStore
-                    && !\in_array($this->targetGroupStore->getTargetGroupId(), $blockPropertyTypeSettings['target_groups'])
-                ) {
-                    continue;
+            foreach ($this->blockSkippers as $blockSkipper) {
```
nnatter, line 389: would it be possible to extract this functionality and include it into a `2.1.*` release? that would really simplify supporting this in the `SuluHeadlessBundle` 🙂

danrot: I think I would not do it... It's a new feature, and therefore should not be released in a bugfix release IMO. Also, what is there to support in 2.1 regarding this? IIRC in Sulu 2.1 the UI controls for defining this stuff did not even exist for the functionality being implemented in these interfaces, so I would say it is safe to just not register these services if this interface does not exist.

nnatter: I am sorry - you are right. I was thinking about a `2.2.*` release, not a `2.1.*` release 🙈  I adjusted the comment above.

danrot: Right, some of this hiding functionality is already in Sulu 2.2, forgot about that :see_no_evil: Personally I would still prefer to merge it only in master, to avoid changing too much in existing releases... Apart from that the interface will change once more to be something like a `Visitor` anyway...

I think I would definitely go to merge everything in this PR, and hopefully this is tested somewhere, so that we see if everything works as expected. When you then try to implement it in the SuluHeadlessBundle, I would try to go first without requiring a change in sulu/sulu, and if it is really that much of a hassle we can still backport that part.

src/Sulu/Component/Content/Types/Block/ScheduleBlockVisitor.php
```diff
@@ -0,0 +1,116 @@
...
+                    $month = $now->format('m');
+                    $day = $now->format('d');
+
+                    $start = new \DateTime($schedule['start']);
+                    $start->setDate($year, $month, $day);
+                    $end = new \DateTime($schedule['end']);
+                    $end->setDate($year, $month, $day);
+
+                    if ($end < $start) {
+                        $start->modify('-1 day');
```
nnatter, line 81: shouldnt this be `$end->modify('+1 day')`? 🤔

src/Sulu/Component/Content/Types/Block/ScheduleBlockVisitor.php
```diff
@@ -0,0 +1,116 @@
...
+            }
+        }
+
+        return $returnBlock ? $block : null;
+    }
+
+    private function matchWeekday(\DateTime $datetime, $schedule)
+    {
+        if (!\is_array($schedule['days'])) {
+            return true;
```
nnatter, line 111: whats the reason for this? 🤔

src/Sulu/Component/Content/Types/Block/ScheduleBlockVisitor.php
```diff
@@ -0,0 +1,116 @@
...
+                    $i = 0;
+                    do {
+                        if ($this->matchWeekday($start, $schedule)) {
+                            break;
+                        }
+
+                        $start->modify('+1 day');
+                        $end->modify('+1 day');
+                        ++$i;
+                    } while ($i < 7);
```
nnatter, line 93: what does this do? can we add a comment or extract it into a function with a descriptive name? 🙈

src/Sulu/Component/Content/Types/BlockContentType.php
```diff
@@ -369,39 +381,26 @@ private function prepareData(PropertyInterface $property, callable $dataCallback
...
             $blockPropertyType = $blockProperty->getProperties($i);
-            $blockPropertyTypeSettings = $blockPropertyType->getSettings();
 
-            if (
-                \is_array($blockPropertyTypeSettings)
-                && !empty($blockPropertyTypeSettings['hidden'])
-            ) {
-                continue;
-            }
+            foreach ($this->blockVisitors as $blockVisitor) {
```
nnatter, line 388: i would really like to extract this into a seperate PR that targets `release/2.2`

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/types.js
```diff
@@ -2,6 +2,11 @@
 import type {Node} from 'react';
 import type {SchemaEntry} from '../Form/types';
 
+export type BlockEntry<T> = {
```
nnatter, line 5: this is the type for the `FieldBlocks`, right? is there a reason for making this generic?

src/Sulu/Bundle/HttpCacheBundle/CacheLifetime/CacheLifetimeEnhancer.php
```diff
@@ -33,14 +33,21 @@ class CacheLifetimeEnhancer implements CacheLifetimeEnhancerInterface
      */
     private $sharedMaxAge;
 
+    /**
+     * @var CacheLifetimeRequestEnhancer
```
nnatter, line 37: This sounds really similar to the `CacheLifetimeEnhancer`. I think using something like `RequestCacheLifetimeStore` would be easier to distinguish

src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/Preview.js
```diff
@@ -216,6 +216,10 @@ class Preview extends React.Component<Props> {
         this.selectedDeviceOption = value;
     };
 
+    @action handleDateTimeChange = debounce((value: ?Date) => {
+        this.previewStore.setDateTime(value || new Date());
```
nnatter, line 220: do we need to debounce this because of the datepicker component we are using?

src/Sulu/Bundle/HttpCacheBundle/CacheLifetime/CacheLifetimeEnhancer.php
nnatter: or maybe add a `addMaxCacheLifetime` method to this class?

src/Sulu/Bundle/HttpCacheBundle/CacheLifetime/CacheLifetimeRequestEnhancer.php
```diff
@@ -0,0 +1,58 @@
...
+     * @var RequestStack
+     */
+    private $requestStack;
+
+    public function __construct(RequestStack $requestStack)
+    {
+        $this->requestStack = $requestStack;
+    }
+
+    public function setCacheLifetime(int $cacheLifetime)
```
nnatter, line 30: should we call this `addMaxCacheLifetime`? think i would expect to overwrite the current value if it is called `setCacheLifetime `

src/Sulu/Bundle/AdminBundle/Resources/js/components/Form/tests/Form.test.js
```diff
@@ -4,5 +4,10 @@ import {render} from 'enzyme';
 import Form from '../Form';
 
 test('Render a form', () => {
-    expect(render(<Form>Test</Form>)).toMatchSnapshot();
```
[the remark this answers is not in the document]
danrot: Actually I am always trying to keep the number snapshot tests minimal, because they cause a lot of noise when something fails... This is about rendering the component without a class, I think flow got already covered most of that by making sure that something is set resp. that it works without a `skin` being set.

src/Sulu/Bundle/AdminBundle/Resources/js/components/Block/Block.js
```diff
@@ -39,7 +39,7 @@ export default class Block extends React.Component<Props> {
         }
     };
 
-    handleTypeChange = (type: string | number) => {
+    handleTypeChange = (type: *) => {
```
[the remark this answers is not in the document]
danrot: Ideally it should, but if I change it I get some errors... Feels to me like there is something wrong in some other place, and I don't think that I have enough time to hunt it down :see_no_evil:

src/Sulu/Bundle/AdminBundle/Resources/js/components/Schedules/Schedules.js
```diff
@@ -0,0 +1,65 @@
...
+    };
+
+    renderBlockContent = (value: ScheduleEntry, type: ScheduleType, index: number) => {
+        switch (value.type) {
+            case 'weekly':
+                return <WeeklySchedule index={index} onChange={this.handleBlockChange} value={value} />;
+            case 'fixed':
+                return <FixedSchedule index={index} onChange={this.handleBlockChange} value={value} />;
+            default:
+                throw new Error('"' + type + '" is not a valid schedule type!');
```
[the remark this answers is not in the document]
danrot: I avoided doing that, because this error can only happen if somebody alters that manually in the DB, and I didn't want to manually repeat these types in here, because it might easily be forgotten if we add new ones.

src/Sulu/Bundle/AdminBundle/Resources/js/components/Schedules/Schedules.js
```diff
@@ -0,0 +1,65 @@
...
+        }
+    };
+
+    render() {
+        const {onChange, value} = this.props;
+
+        return (
+            // $FlowFixMe
```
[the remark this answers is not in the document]
danrot: See https://github.com/sulu/sulu/pull/5608#discussion_r544090942

src/Sulu/Bundle/AdminBundle/Resources/js/components/Schedules/Schedules.js
```diff
@@ -0,0 +1,65 @@
...
+            <BlockCollection
+                collapsable={false}
+                defaultType="fixed"
+                movable={false}
+                onChange={onChange}
+                renderBlockContent={this.renderBlockContent}
+                // $FlowFixMe
```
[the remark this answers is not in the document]
danrot: Yeah, also see https://github.com/sulu/sulu/pull/5608#discussion_r544090942

src/Sulu/Component/Content/Types/Block/ScheduleBlockVisitor.php
danrot: No, then it does not work if you e.g. set the schedule from friday 20:00 to saturday 04:00 and you visit on saturday 03:00, because the start date is used to check if the selected week days are matching.

src/Sulu/Component/Content/Types/Block/ScheduleBlockVisitor.php
danrot: If the user has not selected any weekday, then the given times work for every weekday. Have added a comment.

src/Sulu/Component/Content/Types/Block/ScheduleBlockVisitor.php
danrot: Added a comment. Basically it tries to find the right weekday, especially to set the correct cache lifetime.

src/Sulu/Component/Content/Types/BlockContentType.php
danrot: If we need that I would still do that after this PR has been merged.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/types.js
danrot: Yeah, the `FieldBlocks` are used for the schedules in the overlay, and this way we can say that only `fixed` and `weekly` is allowed as type, not any string. Sadly this does not work all the way, but in some parts it is already useful.

src/Sulu/Bundle/HttpCacheBundle/CacheLifetime/CacheLifetimeEnhancer.php
danrot: Still don't think that it is a `Store` though... But happy to change it, if you think it makes more sense.

Regarding the method name, I don't think that `addMaxCacheLifetime` is a good one... `add` seems like a non-destructive operation, so nothing that has already been there is lost, but this is not the case... `set` might not be perfect either, but IMO it's closer. Something like `setCacheLifetimeIfShorter` would be the clearest, but it does not feel very nice to me, and it couples the name of the function to the implementation.

src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/Preview.js
danrot: If we don't do that, the preview rerenders on every change. That includes every click on the up and down button for the time picker, so if you click ten times the minute up button to travel 10 minutes, a request for rerendering the preview will be sent 10 times to the server, if we don't debounce.

src/Sulu/Bundle/HttpCacheBundle/CacheLifetime/CacheLifetimeRequestEnhancer.php
danrot: It *does* overwrite the current value, but only some check passes. Feels to me much more like a setter than an adder.

src/Sulu/Bundle/AdminBundle/Resources/js/components/Block/Block.js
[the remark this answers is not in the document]
danrot: Partially fixed the flow issues in 23b9e72.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/types.js
nnatter: maybe i missed something, but i think the `Schedules` does use only the `BlockCollection`, not the `FieldBlocks` 🤔

src/Sulu/Bundle/AdminBundle/Resources/js/components/Block/Block.js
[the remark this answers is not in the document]
nnatter: still strange that we need these. but yeah, i think we can use that time better in other places 🙂

src/Sulu/Bundle/AdminBundle/Resources/js/components/Form/tests/Form.test.js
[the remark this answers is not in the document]
nnatter: okay, dont hae a strong opinion on that 🙂

src/Sulu/Bundle/HttpCacheBundle/CacheLifetime/CacheLifetimeEnhancer.php
nnatter: No really happy with `store` too. But I dont have a better idea and i think it makes things easier to understand. So i would change it 🙂

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/types.js
danrot: Right, it seems like it indeed would not be absolutely necessary in this PR. IIRC it was necessary in the beginning of my PR, but some refactoring seems to make it obsolete.

But don't you think it would be a good idea anyway? Would allow for better typing when the `FieldBlocks` component is used in some other place.

src/Sulu/Bundle/AdminBundle/Resources/js/components/Block/Block.js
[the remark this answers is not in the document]
danrot: The particular issue you have listed here has already been fixed anyway :slightly_smiling_face:

src/Sulu/Bundle/HttpCacheBundle/CacheLifetime/CacheLifetimeEnhancer.php
danrot: I think `Enhancer` was easier to understand, because it is also kind of related to the `RouteEnhancer`s from the `CmfRoutingBundle`. But if we rename it, I would go with `CacheLifetimeRequestStore`. When the words `Request` and `Store` are closer together I would think it is more likely to assume that the information is stored on the request.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/types.js
nnatter: i dont think it changes anything when the `FieldBlocks` is used in some other place. The `FieldBlock` sets the generic parameter anyway (`extends React.Component<FieldTypeProps<Array<BlockEntry<string>>>>`). Would make this non-generic because it is easier to understand 🙂

src/Sulu/Bundle/HttpCacheBundle/CacheLifetime/CacheLifetimeEnhancer.php
nnatter: > I think Enhancer was easier to understand, because it is also kind of related to the RouteEnhancers from the CmfRoutingBundle

If there wasnt already a `CacheLifetimeEnhancer`, `CacheLifetimeRequestEnhancer` would be good for me too. But we already have a `CacheLifetimeEnhancer` that does something different, therefore I would prefer`CacheLifetimeRequestStore` 🙂  Yes, `CacheLifetimeRequestStore::setCacheLifetime` is okay for me

src/Sulu/Component/Content/Types/Block/ScheduleBlockVisitor.php
```diff
@@ -0,0 +1,123 @@
...
+            switch ($schedule['type']) {
+                case 'fixed':
+                    $start = new \DateTime($schedule['start']);
+                    $end = new \DateTime($schedule['end']);
+
+                    $this->cacheLifetimeRequestEnhancer->setCacheLifetime($start->getTimestamp() - $nowTimestamp);
+                    $this->cacheLifetimeRequestEnhancer->setCacheLifetime($end->getTimestamp() - $nowTimestamp);
+
+                    if ($now >= $start && $now <= $end) {
+                        $returnBlock = true;
```
nnatter, line 66: cant we just return here instead of setting a variable? the variable will never be set back to false right? 🤔

src/Sulu/Component/Content/Types/Block/ScheduleBlockVisitor.php
```diff
@@ -0,0 +1,123 @@
...
+                        $start->modify('+1 day');
+                        $end->modify('+1 day');
+                        ++$i;
+                    } while ($i < 7);
+
+                    $this->cacheLifetimeRequestEnhancer->setCacheLifetime($start->getTimestamp() - $nowTimestamp);
+                    $this->cacheLifetimeRequestEnhancer->setCacheLifetime($end->getTimestamp() - $nowTimestamp);
+
+                    if ($now >= $start && $now <= $end) {
+                        $returnBlock = true;
```
nnatter, line 105: dont we need `continue 2` here too? or maybe we can just return as asked above?

src/Sulu/Component/Content/Types/Block/ScheduleBlockVisitor.php
nnatter: oky already figured that we probably cannot return because of the cache lifetime, right?

src/Sulu/Component/Content/Types/Block/ScheduleBlockVisitor.php
```diff
@@ -0,0 +1,123 @@
...
+
+                    if ($end < $start) {
+                        // If the end date is smaller than the start date, it means that the user has entered a time
+                        // combination that spans multiple days. In order to make sure that the start date is before the
+                        // end date the start date has to be set to yesterday.
+                        $start->modify('-1 day');
+                    }
+
+                    $i = 0;
+                    // This loop checks for the coming 7 days for matches of the weekdays selected by the users. This is
```
nnatter, line 88:
```suggestion
                    // This loop checks the coming 7 days for matches of the weekdays selected by the user. This is
```

src/Sulu/Component/Content/Types/Block/ScheduleBlockVisitor.php
danrot: Exactly, that's the reason. We have to make sure that all of the cache lifetimes get called, otherwise we end up with incorrect cache invalidations.

src/Sulu/Component/Content/Types/Block/ScheduleBlockVisitor.php
danrot: We could, but since nothing is happen after this `if`, it does not really matter. Would find it even confusing to have a `continue` at the very end :see_no_evil:

src/Sulu/Bundle/PageBundle/Resources/config/forms/page_block_settings.xml
```diff
@@ -26,6 +26,32 @@
             <tag name="sulu.block_setting_icon" icon="su-hide" />
         </property>
 
+        <section name="schedules">
+            <properties>
+                <property name="schedules_enabled" type="checkbox">
+                    <params>
+                        <param name="description">
+                            <meta>
+                                <title>sulu_page.schedules_description</title>
```
nnatter, line 35: looks like this is not translated yet

src/Sulu/Bundle/PageBundle/Resources/config/forms/page_block_settings.xml
```diff
@@ -26,6 +26,32 @@
...
+                            </meta>
+                        </param>
+                        <param name="icon" value="su-clock" />
+                        <param name="label">
+                            <meta>
+                                <title>sulu_page.schedules_label</title>
```
nnatter, line 41: looks like this is not translated yet

src/Sulu/Bundle/AdminBundle/Resources/js/components/Schedules/Schedules.js
```diff
@@ -0,0 +1,71 @@
...
+                throw new Error('"' + type + '" is not a valid schedule type!');
+        }
+    };
+
+    render() {
+        const {disabled, onChange, value} = this.props;
+
+        return (
+            // $FlowFixMe
+            <BlockCollection
```
nnatter, line 56: would set the `addButtonText` to something like `Add schedule`

src/Sulu/Bundle/AdminBundle/Resources/js/components/Schedules/WeeklySchedule.js
```diff
@@ -0,0 +1,65 @@
...
+        onChange(index, {...value, end: end ? transformDateToTimeString(end) : undefined});
+    };
+
+    render() {
+        const {value} = this.props;
+
+        return (
+            <Form>
+                <Form.Field colSpan={6} label={translate('sulu_admin.weekdays')} spaceAfter={6}>
+                    <MultiSelect onChange={this.handleDaysChange} values={value.days || []}>
```
nnatter, line 38: we should set the `noneSelectedText` to tell the user that the times are effective for all days per default

src/Sulu/Bundle/AdminBundle/Resources/js/components/Schedules/WeeklySchedule.js
```diff
@@ -0,0 +1,65 @@
...
+        const {index, onChange, value} = this.props;
+        onChange(index, {...value, end: end ? transformDateToTimeString(end) : undefined});
+    };
+
+    render() {
+        const {value} = this.props;
+
+        return (
+            <Form>
+                <Form.Field colSpan={6} label={translate('sulu_admin.weekdays')} spaceAfter={6}>
```
nnatter, line 37: i know the `colSpan` is like this in the design, but dont you think it would be useful to use `colSpan={12}` to see as many days as possible if the select is closed?

src/Sulu/Component/Content/Types/Block/ScheduleBlockVisitor.php
```diff
@@ -0,0 +1,123 @@
...
+        $now = $this->requestAnalyzer->getDateTime();
+        $nowTimestamp = $now->getTimestamp();
+
+        $returnBlock = false;
+
+        foreach ($blockPropertyTypeSettings['schedules'] as $schedule) {
+            switch ($schedule['type']) {
+                case 'fixed':
+                    $start = new \DateTime($schedule['start']);
+                    $end = new \DateTime($schedule['end']);
```
nnatter, line 60: this will lead to an exception if the user does not set `start` or `end` in the admin interface. can we make them mandatory somehow?

src/Sulu/Component/Content/Types/Block/ScheduleBlockVisitor.php
```diff
@@ -0,0 +1,123 @@
...
+                    break;
+                case 'weekly':
+                    $year = $now->format('Y');
+                    $month = $now->format('m');
+                    $day = $now->format('d');
+
+                    $start = new \DateTime($schedule['start']);
+                    $start->setDate($year, $month, $day);
+                    $end = new \DateTime($schedule['end']);
+                    $end->setDate($year, $month, $day);
```
nnatter, line 78: as above, this will throw an error if the used does not select a `start` or an `end`

src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/Preview.js
```diff
@@ -301,6 +305,24 @@ class Preview extends React.Component<Props> {
                                 options={this.availableDeviceOptions}
                                 value={this.selectedDeviceOption}
                             />
+                            <Toolbar.Popover
+                                icon="su-calendar"
+                                label={translate('sulu_admin.set_time')}
```
nnatter, line 310: this is not translated yet

src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/Preview.js
```diff
@@ -301,6 +305,24 @@ class Preview extends React.Component<Props> {
...
+                            >
+                                {() => (
+                                    <div className={previewStyles.dateTimeForm}>
+                                        <Form skin="dark">
+                                            <Form.Field label={translate('sulu_admin.date_time')}>
```
nnatter, line 315: this is not translated yet

src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/Preview.js
```diff
@@ -301,6 +305,24 @@ class Preview extends React.Component<Props> {
                                 options={this.availableDeviceOptions}
                                 value={this.selectedDeviceOption}
                             />
+                            <Toolbar.Popover
+                                icon="su-calendar"
+                                label={translate('sulu_admin.set_time')}
+                            >
+                                {() => (
+                                    <div className={previewStyles.dateTimeForm}>
```
nnatter, line 313: Did you implement a specific design here? Is it expected that there is a gap between the toolbar and the popover?
Screenshot 2020-12-18 at 16 52 11

src/Sulu/Component/Content/Types/Block/ScheduleBlockVisitor.php
danrot: The "for" is doubled this way, which is not correct.

src/Sulu/Bundle/AdminBundle/Resources/js/components/Schedules/WeeklySchedule.js
danrot: What would you suggest as a text? Writing "All selected" or something like that although nothing is selected seems also confusing to me.

src/Sulu/Component/Content/Types/Block/ScheduleBlockVisitor.php
danrot: Actually I think we've said that you should be able to only set a start or end date... I think I am going to implement it this way.

src/Sulu/Component/Content/Types/Block/ScheduleBlockVisitor.php
danrot: I think in the above case it should be possible to set only a start or only an end date, but in this case the UI should probably already require the user to enter both... Will have a look.

src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/Preview.js
danrot: Are you using Firefox? This seems to work in Chrome, but not in Firefox... Will have a look.

danrot: And I just realized, that there seems already to be an issue like that before this PR... The segments select also has a gap, although it is smaller than this one, but I guess it is for the same reason.
Screenshot from 2020-12-21 09-11-29

danrot:
I am not sure how I should solve the issue with invalid entries in the schedules... I think validation of block settings in general is something that needs a deeper look. For the other settings we didn't really need that yet, so that didn't get enough attention. I think currently I would not check any of that in the UI and catch all the error when evaluating them on the server, then I should also be able to finish that. Is that OK?

src/Sulu/Bundle/AdminBundle/Resources/js/components/Schedules/WeeklySchedule.js
nnatter: Would ust use `Everyday` as `allSelectedText` and also `noneSelectedText` 🙂

src/Sulu/Component/Content/Types/Block/ScheduleBlockVisitor.php
nnatter: but the `for` is doubled right now, isnt it? 🤔

src/Sulu/Bundle/AdminBundle/Resources/js/components/Schedules/Schedules.js
```diff
@@ -0,0 +1,72 @@
...
+                collapsable={false}
+                defaultType="fixed"
+                disabled={disabled}
+                movable={false}
+                onChange={onChange}
+                renderBlockContent={this.renderBlockContent}
+                // $FlowFixMe
+                types={this.types}
+                value={value || []}
+            />
```
nnatter, line 67: think it would be nice to set `minOccurs` to `1`

nnatter:
> I think validation of block settings in general is something that needs a deeper look. 

I think we would need to register a `PropertyMetadataMapper` for the `schedules` field to allow for validation, right?

> I think currently I would not check any of that in the UI and catch all the error when evaluating them on the server, then I should also be able to finish that. Is that OK?

Its okay for me to not validate anything in the UI. But then I would use `00:00`(from)  and `24:00/23:59` (to) as placeholder in the UI and fallback on the server. that feels nicer than just swallowing errors 🙂

src/Sulu/Bundle/AdminBundle/Resources/js/components/Schedules/FixedSchedule.js
```diff
@@ -0,0 +1,48 @@
...
+        onChange(index, {...value, end: transformDateToDateTimeString(end)});
+    };
+
+    render() {
+        const {value} = this.props;
+
+        return (
+            <Form>
+                <Form.Field colSpan={6} label={translate('sulu_admin.start')}>
+                    <DatePicker
```
nnatter, line 32: would use `00:00` as placeholder here and fallback to that value on the server if nothing is set

src/Sulu/Bundle/AdminBundle/Resources/js/components/Schedules/FixedSchedule.js
```diff
@@ -0,0 +1,48 @@
...
+                        onChange={this.handleStartChange}
+                        options={{dateFormat: true, timeFormat: true}}
+                        value={value.start ? new Date(value.start) : undefined}
+                    />
+                </Form.Field>
+                <Form.Field colSpan={6} label={translate('sulu_admin.end')}>
+                    <DatePicker
```
nnatter, line 39: would use `24:00` (or `23:59`) as placeholder here and fallback to that value on the server if nothing is set

src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/Preview.js
nnatter: No I am using `Chrome 87.0.4280.88`

src/Sulu/Component/Content/Types/Block/ScheduleBlockVisitor.php
nnatter: 👍

src/Sulu/Component/Content/Types/Block/ScheduleBlockVisitor.php
nnatter: as written below, would go for sensible defaults if we cannot require a value 🙂

src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/Preview.js
```diff
@@ -301,6 +305,24 @@ class Preview extends React.Component<Props> {
...
                             />
+                            <Toolbar.Popover
+                                icon="su-calendar"
+                                label={(this.previewStore?.dateTime || new Date()).toLocaleString()}
+                            >
+                                {() => (
+                                    <div className={previewStyles.dateTimeForm}>
+                                        <Form skin="dark">
+                                            <Form.Field label={translate('sulu_admin.preview_date_time')}>
+                                                <DatePicker
```
nnatter, line 316: would be nice to add something like a small `x` icon for clearing the field and reset the time to new. but not sure if thats easly possible

src/Sulu/Component/Content/Types/Block/ScheduleBlockVisitor.php
danrot: Ah, sorry, read the diff the wrong way :see_no_evil:

src/Sulu/Bundle/AdminBundle/Resources/js/components/Schedules/WeeklySchedule.js
danrot: Still think that this is confusing... Since you have already suggested using default values below, I would say we do the same here, and make all days selected by default.

src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/Preview.js
danrot: Can you check if the segment dropdown also has a small gap in `master`? Because then it is related to the `PopoverPositioner` and not at all to this PR, would suggest to fix it in another PR then.

danrot:
Now that you've mentioned writing a separate `PropertyMetadataMapper` I realized that it might not even be necessary to have a separate `Schedules` type at all? Why don't we just use a block in the `page_block_settings` form XML? Would make a few things easier, including the validation.

I am really not sure anymore why we said we are going to make that a separate field type, but I could imagine that we initially said it is too different from the form blocks, because also the design is different. But while working on this PR we've included a few params to make use of the parametrization we need for the `Schedules` component anyway, and I think by this we made the `Schedules` component obsolete :thinking:

src/Sulu/Bundle/AdminBundle/Resources/js/components/Schedules/WeeklySchedule.js
nnatter: is okay for me too, but then we should only match the selected days on the server and do not implement a special handling for an empty selection.

src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/Preview.js
nnatter: no, i dont see a gap below the segment select 🙈 

Screenshot 2020-12-22 at 09 40 00

nnatter:
> Now that you've mentioned writing a separate PropertyMetadataMapper I realized that it might not even be necessary to have a separate Schedules type at all? Why don't we just use a block in the page_block_settings form XML? Would make a few things easier, including the validation.

Yes you might be right about that 🙈  Its okay for me to change it if you find the time, but its also okay for me if you dont

src/Sulu/Bundle/AdminBundle/Resources/js/components/Schedules/WeeklySchedule.js
danrot: Made it a mandatory field and set all days by default.

danrot:
@nnatter Sorry, had to rebase :see_no_evil:

src/Sulu/Bundle/AdminBundle/Resources/js/components/Schedules/FixedSchedule.js
danrot: That would not be the correct behavior... It should also be possible to omit start or end, and interpret it as an open interval then...

The only way I see making this work with validation is to use the `schema` node we have introduced for the [`page_settings` form](https://github.com/sulu/sulu/blob/ad20c49418e7cb2a0fd54b0656c527f34b0cc432/src/Sulu/Bundle/PageBundle/Resources/config/forms/page_settings.xml#L8-L49), but I don't think that I am able to finish that on time... This schema is to simple until now, because it does not support any arrays, which we would need because this is within a block.

I think the best bet is to leave it as it is, and improve the schema XML tag in a separate PR.

danrot:
@nnatter Everything I can achieve within the rest of the year is done now I think... Open issues are:

- The [gap in the preview](https://github.com/sulu/sulu/pull/5608#discussion_r5459247569), which I think is an issue that already existed before to a certain extent (maybe the new datetime field is just exploiting it)
- The [not existing validation for the start and end field in the fixed schedule](https://github.com/sulu/sulu/pull/5608#discussion_r546733211), which can probably only be properly fixed by extending the `schema` tag in the form xml, because that's the only place where we can do dependant validation
- The [delete button for the datetime](https://github.com/sulu/sulu/pull/5608#discussion_r546737894), which is not included in the design, and I probably won't get that before the holiday. Also, it is possible to delete the content of the date field manually and reset the date to right now this way.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
```diff
@@ -364,7 +396,10 @@ class FieldBlocks extends React.Component<FieldTypeProps<Array<BlockEntry>>> {
 
     @action handleSettingsOverlayConfirm = () => {
         this.formRef?.submit();
-        this.closeSettingsOverlay();
+
+        if (!this.blockSettingsFormStore.hasErrors) {
```
nnatter, line 400: would it make sense to disable the button if there are errors?

src/Sulu/Bundle/PageBundle/Resources/config/forms/page_block_settings.xml
```diff
@@ -26,6 +26,129 @@
...
+                        <param name="type" value="toggler" />
+                    </params>
+
+                    <tag name="sulu.block_setting_icon" icon="su-clock" />
+                </property>
+
+                <block name="schedules" minOccurs="1" visibleCondition="schedules_enabled == true">
+                    <params>
+                        <param name="collapsable" value="false" />
+                        <param name="movable" value="false" />
```
nnatter, line 54: would still be nice to set a `add_button_text`

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
```diff
@@ -121,6 +121,38 @@ class FieldBlocks extends React.Component<FieldTypeProps<Array<BlockEntry>>> {
         return addButtonText;
     }
 
+    @computed get collapsable() {
```
nnatter, line 124: we need to document this

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
```diff
@@ -121,6 +121,38 @@ class FieldBlocks extends React.Component<FieldTypeProps<Array<BlockEntry>>> {
...
+        } = this.props;
+
+        if (collapsable !== undefined && typeof collapsable !== 'boolean') {
+            throw new Error('The "block" field types only accepts booleans as "collapsable" schema option!');
+        }
+
+        return collapsable;
+    }
+
+    @computed get movable() {
```
nnatter, line 140: we need to document this

src/Sulu/Bundle/PageBundle/Resources/config/forms/page_block_settings.xml
```diff
@@ -26,6 +26,129 @@
...
+                            </meta>
+                        </param>
+                        <param name="skin" value="heading" />
+                        <param name="type" value="toggler" />
+                    </params>
+
+                    <tag name="sulu.block_setting_icon" icon="su-clock" />
+                </property>
+
+                <block name="schedules" minOccurs="1" visibleCondition="schedules_enabled == true">
```
nnatter, line 51: should we set the default type to `fixed` like before?

src/Sulu/Bundle/AdminBundle/Resources/js/components/Schedules/FixedSchedule.js
nnatter: you are right, i thought i added this comment to the `WeeklySchedule` 🙈

src/Sulu/Bundle/PageBundle/Resources/translations/admin.en.json
```diff
@@ -57,5 +57,8 @@
     "sulu_page.segment_label": "Assign segment",
     "sulu_page.segment_description": "The block will be automatically filtered on the website by the chosen segment.",
     "sulu_page.target_groups_label": "Assign target groups",
-    "sulu_page.target_groups_description": "The block will only be visible for the selected target groups."
+    "sulu_page.target_groups_description": "The block will only be visible for the selected target groups.",
+    "sulu_page.add_schedule": "Add schedule",
+    "sulu_page.schedules_label": "Assign time schedule",
+    "sulu_page.schedules_description": "The block will only be visible if at least one condition is valid."
```
nnatter, line 63:
```suggestion
    "sulu_page.schedules_description": "The block will only be visible if at least one condition is fulfilled."
```

src/Sulu/Bundle/PageBundle/Resources/config/forms/page_block_settings.xml
danrot: Sorry, already implemented it but refactored that part afterwards :see_no_evil:

src/Sulu/Bundle/PageBundle/Resources/translations/admin.en.json
danrot: This translation is from Bernd, and he didn't like fulfilled.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
danrot: Done in https://github.com/sulu/sulu-docs/pull/614

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
danrot: Done in https://github.com/sulu/sulu-docs/pull/614

nnatter (approved):

PR #5610, opened 2020-11-06
Fix sulu_content_load for non pages

plozmun:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | https://github.com/sulu/sulu/pull/5599
| License | MIT
| Documentation PR | -

#### What's in this PR?

Fix for sulu_content_load when is used in articles and snippets
@alexander-schranz getStructureType does not exists and no there is no similar function. I use PageInterface to check it , WDT? 

#### Why?

Because now the error `Call to a member function getSecurity() on null` occurs

### DIFF
--- a/src/Sulu/Bundle/WebsiteBundle/Twig/Content/ContentTwigExtension.php
+++ b/src/Sulu/Bundle/WebsiteBundle/Twig/Content/ContentTwigExtension.php
@@ -115,7 +115,10 @@ class ContentTwigExtension extends AbstractExtension implements ContentTwigExten
 
             $document = $contentStructure->getDocument();
 
-            if (!$document || !$document instanceof WebspaceBehavior) {
+            if (!$this->securityChecker
+                || !$document instanceof WebspaceBehavior
+                || !$document instanceof SecurityBehavior
+            ) {
                 return $this->structureResolver->resolve($contentStructure);
             }
 
@@ -124,7 +127,6 @@ class ContentTwigExtension extends AbstractExtension implements ContentTwigExten
             $system = $security ? $security->getSystem() : null;
 
             if ($targetWebspace->hasWebsiteSecurity()
-                && $this->securityChecker
                 && !$this->securityChecker->hasPermission(
                     new SecurityCondition(
                         PageAdmin::SECURITY_CONTEXT_PREFIX . $contentStructure->getWebspaceKey(),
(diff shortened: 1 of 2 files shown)


src/Sulu/Bundle/WebsiteBundle/Twig/Content/ContentTwigExtension.php
```diff
@@ -112,6 +113,10 @@ public function load($uuid)
                 $locale
             );
 
+            if (false === $contentStructure instanceof PageInterface) {
```
alexander-schranz, line 116: Can you check if something like:

```
$document = $contentStructure->getDocument();

if ($document && 'page' === $document->getStructureType()) {
```

Works here.

I'm afraid that some custom bundles extended from the `PageBridge` and they would run into the same issue if you check for the Interface.

plozmun:
I will check better this weekend but `getStructureType` returns the template key 

Another approach is to use WebspaceBehavior 

```php
            use Sulu\Component\Content\Document\Behavior\WebspaceBehavior;
          
            ...

            $document = $contentStructure->getDocument();

            if (!$document || !$document instanceof WebspaceBehavior) {
                return $this->structureResolver->resolve($contentStructure);
            }
            
            $webspaceKey = $document->getWebspaceName();
```

plozmun:
@alexander-schranz can you take a look? I've added some tests. Thx

thomasduenser:
@plozmun @alexander-schranz None of the suggestions above are really working for my case.

I've tested it for **Pages** and for **Articles** and the only thing that worked for me was [this](https://github.com/sulu/sulu/pull/5599/files#diff-1fcfaa0b00e894a0e8940edfa289a8ff7cc169c53dcd360384d504f8e1bb4f95R115) code part:

```
$targetWebspaceKey = $contentStructure->getWebspaceKey();

if ($targetWebspaceKey) {
```

plozmun:
> @plozmun @alexander-schranz None of the suggestions above are really working for my case.
> 
> I've tested it for **Pages** and for **Articles** and the only thing that worked for me was [this](https://github.com/sulu/sulu/pull/5599/files#diff-1fcfaa0b00e894a0e8940edfa289a8ff7cc169c53dcd360384d504f8e1bb4f95R115) code part:
> 
> ```
> $targetWebspaceKey = $contentStructure->getWebspaceKey();
> 
> if ($targetWebspaceKey) {
> ```

It’s strange because I have the same issue and it works properly. Please , can you tell me what kind of document you get $structure->getDocument()’

plozmun:
There are another WebspaceBehavior in ArticleBundle ( it's strange this ), maybe have you used this one?

alexander-schranz:
@plozmun and @thomasduenser I updated the PR can you test if this does work for your use cases?

nnatter:
Code looks good, but I am not confident to be able to catch every case when testing this TBH. Would be nice to get some feedback from @plozmun and @thomasduenser before merging this 🙂

thomasduenser:
It works in our case:
- no website security
- for articles
- for pages

plozmun:
> It works in our case:
> 
> * no website security
> * for articles
> * for pages

Same for me 👍

nnatter:
Thanks for your feedback! 🙂

nnatter (approved):

PR #5615, opened 2020-11-09
Isolate styling of modals from their children

danrot:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | #5608
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR uses specific classes to style some stuff of the `Dialog` and `Overlay` components instead of relying on the tag name.

#### Why?

Because using the tag names might introduce problems (as happened in #5608). When some children are passed to the `Dialog` or `Overlay` their styling is affected if they use some `header`, `article` or `footer` tags.


danrot:
Backported this to Sulu 2.1, because this can happen to any installation where `Dialog` or `Overlay` is combined with some children containiner `header`, `article` or `footer` tags.

nnatter (approved):

Issue #5028, opened 2020-02-03
Sections doesn't work inside blocks


alexander-schranz:
| Q | A
| --- | ---
| Sulu Version | 2.0.3, dev-master
| Browser Version | -

# Actual Behavior

Section inside blocks currently crash on 2.0.3 on the frontend and on dev-master in the SchemaMetadata.

**dev-master**

> Argument 1 passed to Sulu\Bundle\AdminBundle\Metadata\SchemaMetadata\SchemaMetadata::Sulu\Bundle\AdminBundle\Metadata\SchemaMetadata\{closure}() must be an instance of Sulu\Bundle\AdminBundle\Metadata\SchemaMetadata\PropertyMetadata, array given

**2.0.3**

>  uncaught exception: Object

Properties not rendered.

# Expected Behavior

Section should work also inside blocks. They are needed to control the grid better or structure the block content.

# Steps to Reproduce

```xml
        <block name="mytestblocks" default-type="my-test-type">
            <types>
                <type name="my-test-type">
                    <properties>
                        <section name="section1">
                            <properties>
                                <property name="test1" type="text_line" colspan="4">
                                    <meta>
                                        <title>Test1</title>
                                    </meta>
                                </property>

                                <property name="test2" type="text_editor" colspan="4">
                                    <meta>
                                        <title>Test2</title>
                                    </meta>
                                </property>

                                <property name="test3" type="text_line" colspan="4">
                                    <meta>
                                        <title>Test3</title>
                                    </meta>
                                </property>
                            </properties>
                        </section>

                        <section name="section2">
                            <properties>
                                <property name="test4" type="text_line" colspan="4">
                                    <meta>
                                        <title>Test4</title>
                                    </meta>
                                </property>

                                <property name="test5" type="text_editor" colspan="4">
                                    <meta>
                                        <title>Test5</title>
                                    </meta>
                                </property>

                                <property name="test6" type="text_line" colspan="4">
                                    <meta>
                                        <title>Test6</title>
                                    </meta>
                                </property>
                            </properties>
                        </section>
                    </properties>
                </type>
            </types>
        </block>
```

# Possible Solutions

If you have already ideas how to solve the issue, add them here.
(remove this section if not needed)

danrot:
@alexander-schranz We decided to use the assignee when somebody really starts working on something. All of us have almost historic issues assigned, and we still don't feel responsible for them, therefore I would rather use the assignee option this way.

TheCadien:
[...]

Issue #4786, opened 2019-10-03
A section at the top of a page XML results in too much space.


kfrohwein:
I add a < section > element to capsule title and url. This results in too much space. Seems that the tab bar has some weird displaced <div> elemet that pushes everything down. And maybe the section should have less space above.

![sulu_too_much_space](https://user-images.githubusercontent.com/245385/66135802-bf904a80-e5fa-11e9-9a2d-6a794842da7c.png)

danrot:
Fixed in #5618

PR #5618, opened 2020-11-10
Fix sections in blocks

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes https://github.com/sulu/sulu/issues/5028
| License | MIT

#### What's in this PR?

Removes the `SectionProperties` for blocks similar to the sections in forms and thus fixing the bug from https://github.com/sulu/sulu/issues/5028

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Block/block.scss
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Block/block.scss
@@ -26,7 +26,7 @@ $blockHandleWidth: 20px;
             display: flex;
             align-items: center;
             padding: 0;
-            margin: 10px;
+            margin: 1em;
             height: 18px;
         }
     }
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Form/grid.scss
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Form/grid.scss
@@ -1,6 +1,7 @@
 @import '../../components/Grid/variables.scss';
 @import '../../components/Form/variables.scss';
 @import './variables.scss';
+@import '../Divider/divider.scss';
 
 .grid {
     margin-bottom: -$sectionMarginBottom;
@@ -11,7 +12,7 @@
 
     &:first-child {
         .divider-container {
-            margin-top: -50px;
+            margin-top: -$dividerTopMargin;
         }
     }
 
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Form/grid.scss
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Form/grid.scss
@@ -9,6 +9,12 @@
 .grid-section {
     margin-bottom: -$fieldMarginBottom;
 
+    &:first-child {
+        .divider-container {
+            margin-top: -50px;
+        }
+    }
+
     &:last-child {
         margin-bottom: 0;
     }
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Form/grid.scss
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Form/grid.scss
@@ -1,7 +1,7 @@
 @import '../../components/Grid/variables.scss';
 @import '../../components/Form/variables.scss';
 @import './variables.scss';
-@import '../Divider/divider.scss';
+@import '../Divider/variables.scss';
 
 .grid {
     margin-bottom: -$sectionMarginBottom;
--- a/src/Sulu/Component/Content/Metadata/PropertiesMetadata.php
+++ b/src/Sulu/Component/Content/Metadata/PropertiesMetadata.php
@@ -113,7 +113,7 @@ class PropertiesMetadata extends ItemMetadata
 
             if ($child instanceof BlockMetadata) {
                 // Clone before removing sections, because the metadata including the sections is required elsewhere
-                $clone = \unserialize(\serialize($child));
+                $clone = clone $child;
                 foreach ($clone->getComponents() as $component) {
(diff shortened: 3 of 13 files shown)


nnatter:
I think this also fixes #4786

danrot:
@Prokyonn It seems like some snapshots are failing.

Prokyonn:
> @Prokyonn It seems like some snapshots are failing.

@danrot should be fixed now :)

src/Sulu/Component/Content/Metadata/PropertiesMetadata.php
```diff
@@ -111,6 +111,19 @@ private function removeSectionProperties(array $children)
                 continue;
             }
 
+            if ($child instanceof BlockMetadata) {
+                // deep cloning is necessary because the components must only be manipulated for these properties
```
danrot, line 115: Still don't really understand why this is necessary :thinking:

src/Sulu/Component/Content/Metadata/PropertiesMetadata.php
```diff
@@ -111,6 +111,19 @@ private function removeSectionProperties(array $children)
...
+                $clone = \unserialize(\serialize($child));
```
danrot, line 116: Doesn't [`clone`](https://www.php.net/manual/en/language.oop5.cloning.php) work?

src/Sulu/Bundle/AdminBundle/Resources/js/components/Form/grid.scss
```diff
@@ -9,6 +9,12 @@
 .grid-section {
     margin-bottom: -$fieldMarginBottom;
 
+    &:first-child {
+        .divider-container {
+            margin-top: -50px;
```
danrot, line 14: I think I would not just use a magic number here, but instead move that into an own variable in the `Divider` component. Then this will continue to work if we change the top margin of the divider component.

danrot (changes requested):

src/Sulu/Bundle/AdminBundle/Resources/js/components/Block/block.scss
```diff
@@ -26,7 +26,7 @@ $blockHandleWidth: 20px;
             display: flex;
             align-items: center;
             padding: 0;
-            margin: 10px;
+            margin: 1em;
```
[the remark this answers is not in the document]
Prokyonn: @chirimoya wants that both the `title` and the `block-type` are on one line, when the collapsed `block` only shows one line

src/Sulu/Component/Content/Metadata/PropertiesMetadata.php
Prokyonn: The problem is, that we need the properties with the `sections` for the `metadata` request and without the `sections` for the data-properties and it seems like both are referring to the same `BlockMetaData` object.

If we don't clone the `BlockMetaData` object and remove the `sections` then all the `sections` are also removed in the `metadata` request and the `section` is not rendered in the frontend. This occurs because the same object is used and thus should not be modified.

For the `sections` outside of a block this is not an issue because we don't have to modify an object, we just extract the inner properties and append them.
But to get rid of the `sections` inside of the `BlockMetaData` we have to modify the `components` directly on the object via the `$component->setChildren(..)` method.

src/Sulu/Component/Content/Metadata/PropertiesMetadata.php
Prokyonn: Nope because `clone` copies the object but uses the same references for the child objects

src/Sulu/Component/Content/Metadata/PropertiesMetadata.php
danrot: Ok, I see... Maybe write the comment then more like that? What do you think about this?

```php
// Clone before removing sections, because the metadata including the sessions is required elsewhere
```

src/Sulu/Component/Content/Metadata/PropertiesMetadata.php
danrot: Why not override the `__clone` as it is shown in the [first example of the clone manual](https://www.php.net/manual/en/language.oop5.cloning.php)? I think that's what I would expect here to happen.

src/Sulu/Bundle/AdminBundle/Resources/js/components/Form/grid.scss
```diff
@@ -1,6 +1,7 @@
 @import '../../components/Grid/variables.scss';
 @import '../../components/Form/variables.scss';
 @import './variables.scss';
+@import '../Divider/divider.scss';
```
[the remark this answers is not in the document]
Prokyonn: Oh sorry my bad, I wasn't sure if I should make a separate file, because the other variables were already in the `divider.css` file.

src/Sulu/Component/Content/Metadata/PropertiesMetadata.php
Prokyonn: Ok I will try it 👍

src/Sulu/Component/Content/Metadata/PropertiesMetadata.php
chirimoya: `... sessions ...` ?

src/Sulu/Component/Content/Metadata/PropertiesMetadata.php
Prokyonn: @danrot fixed it

danrot (approved):

PR #5622, opened 2020-11-11
Do not discard existing data in setMultiple method of MemoryFormStore

nnatter:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### Why?

Right now, it is not possible to use a `text_line` property inside of the `page_block_settings` form because of this. See https://github.com/sulu/sulu-demo/pull/86 for how to extend the `page_block_settings` form.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/MemoryFormStore.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/MemoryFormStore.js
@@ -47,7 +47,7 @@ export default class MemoryFormStore extends AbstractFormStore implements FormSt
     }
 
     @action setMultiple(data: Object) {
-        this.data = data;
+        this.data = {...this.data, ...data};
 
         super.setMultiple();
     }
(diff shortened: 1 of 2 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/MemoryFormStore.js
```diff
@@ -47,7 +47,7 @@ export default class MemoryFormStore extends AbstractFormStore implements FormSt
     }
 
     @action setMultiple(data: Object) {
-        this.data = data;
+        this.data = {...this.data, ...data};
```
danrot, line 50: Actually this is quite a change in the semantics... Would it at leaset be consistent with the behavior of the `ResourceFormStore`?

nnatter: Yes.

https://github.com/sulu/sulu/blob/master/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js#L155

https://github.com/sulu/sulu/blob/master/src/Sulu/Bundle/AdminBundle/Resources/js/stores/ResourceStore/ResourceStore.js#L268

danrot (approved):

PR #5625, opened 2020-11-12
Fix account/contact media list

luca-rath:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

A bugfix for account/contact media list controller

#### Why?

The media list view of accounts and contacts didn't contain all necessary data. Also, filtering by `type` e.g. resulted in a 500 error

### DIFF
--- a/src/Sulu/Bundle/ContactBundle/Controller/AbstractMediaController.php
+++ b/src/Sulu/Bundle/ContactBundle/Controller/AbstractMediaController.php
@@ -17,13 +17,17 @@ use Sulu\Bundle\ContactBundle\Contact\AbstractContactManager;
 use Sulu\Bundle\MediaBundle\Api\Media;
 use Sulu\Bundle\MediaBundle\Entity\MediaRepositoryInterface;
 use Sulu\Bundle\MediaBundle\Media\ListRepresentationFactory\MediaListRepresentationFactory;
+use Sulu\Bundle\MediaBundle\Media\Manager\MediaManagerInterface;
 use Sulu\Component\Rest\AbstractRestController;
 use Sulu\Component\Rest\Exception\EntityNotFoundException;
 use Sulu\Component\Rest\Exception\RestException;
 use Sulu\Component\Rest\ListBuilder\CollectionRepresentation;
+use Sulu\Component\Rest\ListBuilder\Doctrine\DoctrineListBuilderFactoryInterface;
 use Sulu\Component\Rest\ListBuilder\Doctrine\FieldDescriptor\DoctrineFieldDescriptor;
 use Sulu\Component\Rest\ListBuilder\Doctrine\FieldDescriptor\DoctrineJoinDescriptor;
 use Sulu\Component\Rest\ListBuilder\FieldDescriptorInterface;
+use Sulu\Component\Rest\ListBuilder\ListRepresentation;
+use Sulu\Component\Rest\RestHelperInterface;
 use Sulu\Component\Security\Authentication\UserInterface;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
@@ -34,7 +38,27 @@ use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInt
  */
 abstract class AbstractMediaController extends AbstractRestController
 {
+    protected static $collectionEntityName = 'SuluMediaBundle:Collection';
+
+    protected static $fileVersionEntityName = 'SuluMediaBundle:FileVersion';
+
+    protected static $fileEntityName = 'SuluMediaBundle:File';
+
+    protected static $fileVersionMetaEntityName = 'SuluMediaBundle:FileVersionMeta';
+
     protected static $mediaEntityKey = 'media';
+
+    protected $fieldDescriptors = null;
+
+    /**
+     * @var RestHelperInterface
+     */
+    private $restHelper;
+
+    /**
+     * @var DoctrineListBuilderFactoryInterface
+     */
+    private $listBuilderFactory;
 
     /**
      * @var EntityManagerInterface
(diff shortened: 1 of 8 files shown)


src/Sulu/Bundle/MediaBundle/Media/ListRepresentationFactory/MediaListRepresentationFactory.php
```diff
@@ -0,0 +1,241 @@
...
+use Sulu\Component\Rest\ListBuilder\Doctrine\FieldDescriptor\DoctrineFieldDescriptor;
+use Sulu\Component\Rest\ListBuilder\ListBuilderInterface;
+use Sulu\Component\Rest\ListBuilder\ListRepresentation;
+use Sulu\Component\Rest\ListBuilder\Metadata\FieldDescriptorFactoryInterface;
+use Sulu\Component\Rest\RestHelperInterface;
+use Sulu\Component\Security\Authentication\UserInterface;
+use Sulu\Component\Security\Authorization\PermissionTypes;
+use Sulu\Component\Security\Authorization\SecurityCheckerInterface;
+
+class MediaListRepresentationFactory
```
danrot, line 29: Not sure if I would build it this way... Don't you think it would make sense to split that into multiple classes? I could imagine that if we split it into a `MediaListRepresentationFactory` and a `MediaListBuilderFactory` it would be easier to understand what is both components are doing. And I am not sure if I would include a `getFieldDescriptors` method, that contains a one-liner. Feels a bit odd to me (and it is again a function we would not allow to remove anymore). We could even split it further to have a `MediaFieldDescriptorFactory`, but since it is only a function call being passed `media`, I am not sure that would be worth it. /cc @alexander-schranz WDYT? Since you've built most of the `MediaBundle` :slightly_smiling_face:

danrot (changes requested):

src/Sulu/Bundle/ContactBundle/Controller/AbstractMediaController.php
```diff
@@ -206,50 +222,119 @@ protected function removeMediaFromEntity($entityName, $id, $mediaId)
...
+                        $entityName,
+                        $routeName,
+                        $contactId,
+                        $request,
+                        $locale
+                    );
+                } else {
+                    $fieldDescriptors = $this->mediaListRepresentationFactory->getFieldDescriptors();
+
+                    $fieldDescriptors['contactId'] = new DoctrineFieldDescriptor(
```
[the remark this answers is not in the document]
luca-rath: `contactId` is just a name and represents the id of either a contact or an account. It also was like that before. Feel free to suggest a more generic name which is good for both, contacts and accounts. But since both, the `AccountManager` and the `ContactManager` extends from the `AbstractContactManager`, I don't think this naming is too bad here.

alexander-schranz (changes requested):

src/Sulu/Bundle/MediaBundle/Media/ListRepresentationFactory/MediaListRepresentationFactory.php
alexander-schranz: Splitting make sense for me and remove the single line function.

src/Sulu/Bundle/ContactBundle/Controller/AbstractMediaController.php
```diff
@@ -206,50 +222,119 @@ protected function removeMediaFromEntity($entityName, $id, $mediaId)
...
+     * Returns a list representation containing all media of an entity.
+     *
+     * @param string $entityName
+     * @param string $routeName
+     * @param Request $request
+     * @param string $locale
+     *
+     * @return ListRepresentation
+     */
+    protected function legacyGetMultipleView($entityName, $routeName, $contactId, $request, $locale)
```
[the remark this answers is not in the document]
luca-rath: this is a newly introduced method, but you're right, it should be private. I will rename it to `getListRepresentation`

src/Sulu/Bundle/ContactBundle/Controller/AbstractMediaController.php
```diff
@@ -95,6 +117,16 @@ public function __construct(
         $this->mediaRepository = $mediaRepository;
         $this->mediaClass = $mediaClass;
         $this->mediaManager = $mediaManager;
+        $this->mediaListBuilderFactory = $mediaListBuilderFactory;
+        $this->mediaListRepresentationFactory = $mediaListRepresentationFactory;
+        $this->fieldDescriptorFactory = $fieldDescriptorFactory;
+
+        if (null === $this->mediaListBuilderFactory) {
```
danrot, line 124: Shouldn't we add the same check for the other parameters?

src/Sulu/Bundle/ContactBundle/Controller/AbstractMediaController.php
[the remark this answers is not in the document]
danrot: I am not happy with the contact and account naming all together, especially since they are called differently in the UI :see_no_evil: But if it was like that before, I would say we ignore that issue for this PR.

src/Sulu/Bundle/MediaBundle/Controller/MediaController.php
```diff
@@ -142,6 +157,15 @@ public function __construct(
         $this->fieldDescriptorFactory = $fieldDescriptorFactory;
         $this->mediaClass = $mediaClass;
         $this->collectionClass = $collectionClass;
+        $this->mediaListBuilderFactory = $mediaListBuilderFactory;
+        $this->mediaListRepresentationFactory = $mediaListRepresentationFactory;
+
+        if (null === $this->mediaListBuilderFactory) {
```
danrot, line 163: Same as above, we probably have to do the same check for the other newly introduced arguments as well.

danrot (approved):

PR #5628, opened 2020-11-13
Fix block preview with sections

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Currently preview of collapsed blocks with sections doesn't work properly.
This PR removes the sections from the `blockSchemaTypeForm` for the block preview rendering and fixes this bug.

Additionally a check was added to skip potential preview fields without a value. So that the first fields with a value are shown in the preview.

#### Why?

Because the block preview should also work as properly with sections.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
@@ -235,7 +235,7 @@ class FieldBlocks extends React.Component<FieldTypeProps<Array<BlockEntry>>> {
 
     removeSections(blockSchemaTypeForm: Object) {
         let filteredForm = {};
-        Object.keys(blockSchemaTypeForm).map((key) => {
+        Object.keys(blockSchemaTypeForm).forEach((key) => {
             if (blockSchemaTypeForm[key]['type'] === 'section') {
                 filteredForm = {...filteredForm, ...this.removeSections(blockSchemaTypeForm[key]['items'])};
                 return false;
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
@@ -167,12 +167,14 @@ class FieldBlocks extends React.Component<FieldTypeProps<Array<BlockEntry>>> {
     // eslint-disable-next-line no-unused-vars
     renderCollapsedBlockContent = (value: Object, type: string, index: number) => {
         const blockSchemaType = this.getBlockSchemaType(type);
-        const blockSchemaTypeForm = blockSchemaType.form;
+        const blockSchemaTypeForm = this.removeSections(blockSchemaType.form);
 
         const previewPropertyNames = Object.keys(blockSchemaTypeForm)
             .filter((schemaKey) => {
                 const schemaEntryTags = blockSchemaTypeForm[schemaKey].tags;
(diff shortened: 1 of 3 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
```diff
@@ -167,12 +167,14 @@ class FieldBlocks extends React.Component<FieldTypeProps<Array<BlockEntry>>> {
...
         const blockSchemaType = this.getBlockSchemaType(type);
-        const blockSchemaTypeForm = blockSchemaType.form;
+        const blockSchemaTypeForm = this.removeSections(blockSchemaType.form);
 
         const previewPropertyNames = Object.keys(blockSchemaTypeForm)
             .filter((schemaKey) => {
                 const schemaEntryTags = blockSchemaTypeForm[schemaKey].tags;
-                return schemaEntryTags && schemaEntryTags.some((tag) => tag.name === BLOCK_PREVIEW_TAG);
+                return schemaEntryTags &&
+                    value[schemaKey] &&
```
danrot, line 176: Why is that necessary now?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
```diff
@@ -201,7 +203,7 @@ class FieldBlocks extends React.Component<FieldTypeProps<Array<BlockEntry>>> {
         if (previewPropertyNames.length === 0) {
             for (const fieldTypeKey of blockPreviewTransformerRegistry.blockPreviewTransformerKeysByPriority) {
                 for (const propertyName of Object.keys(blockSchemaTypeForm)) {
-                    if (blockSchemaTypeForm[propertyName].type === fieldTypeKey) {
+                    if (blockSchemaTypeForm[propertyName].type === fieldTypeKey && value[propertyName]) {
```
danrot, line 206: Why is that necessary?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
```diff
@@ -231,6 +233,20 @@ class FieldBlocks extends React.Component<FieldTypeProps<Array<BlockEntry>>> {
         );
     };
 
+    removeSections(blockSchemaTypeForm: Object) {
+        let filteredForm = {};
+        Object.keys(blockSchemaTypeForm).map((key) => {
```
danrot, line 238: The `map` call does not really make sense here, since it returns a new array, which is not used. We could use `forEach` instead, or use `reduce`, which would avoid the `let` variable at all:

```javascript
return Object.keys(blockSchemaTypeForm).reduce((blockSchema, key) => {
    if (blockSchemaTypeForm[key]['type'] === 'section') {
        blockSchema = {...blockSchema, ...this.removeSections(blockSchemaTypeForm[key]['items']};
    } else {
        filteredForm[key] = blockSchemaTypeForm[key];
    }

    return blockSchema;
}, {});
```

(Untested, might need small adjustements)

danrot (changes requested):

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
Prokyonn: It is not necessary for the block preview.

Currently the first fields are rendered no matter if these fields have values in it or not, with this change the first fields with a value are rendered.

danrot (approved):

PR #5635, opened 2020-11-17
Render anchor tag instead of text if target of sulu-link is not published in preview

nnatter:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This PR adjusts the implementation of the `<sulu-link>` tag to render an `<a>` tag without any `href` if the target of the link is not published. Right now, only the text of the link is rendered, this breaks the styling for unpublished links in the preview.

#### Why?

Right now, only the text of the link is rendered. This means that the style for `<a>` elements will not be applied to unpublished links. This is behaviour is a bit confusing for the content manager.

### DIFF
--- a/src/Sulu/Bundle/MarkupBundle/Markup/LinkTag.php
+++ b/src/Sulu/Bundle/MarkupBundle/Markup/LinkTag.php
@@ -40,22 +40,22 @@ class LinkTag implements TagInterface
         $result = [];
         foreach ($attributesByTag as $tag => $attributes) {
             $provider = $this->getValue($attributes, 'provider', self::DEFAULT_PROVIDER);
-            if (!isset($attributes['href'])
-                || !\array_key_exists($provider . '-' . $attributes['href'], $contents)
-            ) {
-                $result[$tag] = $this->getContent($attributes);
 
-                continue;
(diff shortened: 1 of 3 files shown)


alexander-schranz:
> `<a>` tag without any href if the target of the link is not published

This is by design that unpublished sulu links are not rendered by design and just the text is shown. What you mean it break the design? I would think that changing this behaviour would break other design too.

/cc @wachterjohannes

nnatter:
If a developer decides to add a global style for `<a>` tags (for example add a background color), this style will not be applied to unpublished links. Also, if the developer uses the `<sulu-link>` tag in a `.twig` template like in the following example, the `my-awesome-button-link` class will not be added if the target is unpublished:
```twig
<sulu-link href="123-123-123" title="test-title" class="my-awesome-button-link" />
```

I think we can argue that this is the desired behaviour on the public website, because we dont want to render "dead" links in terms of usability and SEO. But I also think this is a confusing behaviour inside of the preview for the content manager. Therefore I will adjust this PR to render dead links inside of the preview.

PR #5636, opened 2020-11-17
Log error to javascript console if administration build does not match Sulu version

nnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This PR adjusts the `AdminBundle` to log an error if the version of the Sulu backend does not match the version of the admin javascript build.

#### Why?

Because a lot of users forget to update the javascript build of the administration interface when upgrading their application.

### DIFF
--- a/flow-typed/globals.js
+++ b/flow-typed/globals.js
@@ -1,0 +1,1 @@
+declare var SULU_ADMIN_BUILD_VERSION: string;
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/index.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/index.js
@@ -310,6 +310,16 @@ function processConfig(config: Object) {
 }
 
 function startAdmin() {
+    if (Config.suluVersion !== SULU_ADMIN_BUILD_VERSION) {
+        log.error(
+            'Sulu administration interface: JavaScript build version mismatch' +
+            '\nJavaScript build of the Sulu administration interface does not match the version of the Sulu backend.' +
+            '\nBackend version: ' + Config.suluVersion + ', JavaScript build version: ' + SULU_ADMIN_BUILD_VERSION +
+            '\n\nHave you forgotten to update the build while upgrading your application?' + '' +
+            '\nhttps://docs.sulu.io/en/latest/upgrades/upgrade-2.x.html'
+        );
+    }
+
     const router = new Router(createHashHistory());
     router.addUpdateAttributesHook(updateRouterAttributesFromView);
     router.addUpdateAttributesHook(updateRouterAttributesFromUserStoreContentLocale);
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -20,11 +20,12 @@ module.exports = (env, argv) => { // eslint-disable-line no-undef
         publicDir = composerConfig.extra['public-dir'];
     }
 
-    let suluVersion = undefined;
+    // default value for version must match default value in SuluVersionPass.php
+    let suluVersion = '_._._';
     if (fs.existsSync(path.resolve(projectRootPath, 'composer.lock'))) {
         const composerLock = JSON.parse(fs.readFileSync(path.resolve(projectRootPath, 'composer.lock')));
         const suluPackage = composerLock.packages.find((packageItem) => packageItem.name === 'sulu/sulu');
-        suluVersion = suluPackage ? suluPackage.version : undefined;
+        suluVersion = suluPackage ? suluPackage.version : suluVersion;
     }
 
     const CleanObsoleteChunksPlugin = require(path.resolve(nodeModulesPath, 'webpack-clean-obsolete-chunks'));
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -20,9 +20,12 @@ module.exports = (env, argv) => { // eslint-disable-line no-undef
         publicDir = composerConfig.extra['public-dir'];
     }
 
-    const composerLock = JSON.parse(fs.readFileSync(path.resolve(projectRootPath, 'composer.lock')));
-    const suluPackage = composerLock.packages.find((p) => p.name === 'sulu/sulu');
-    const suluVersion = suluPackage ? suluPackage.version : 'unknown';
+    let suluVersion = undefined;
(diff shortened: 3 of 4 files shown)


flow-typed/globals.js
```diff
@@ -0,0 +1 @@
+declare var SULU_ADMIN_BUILD_VERSION: string;
```
danrot, line 1: That folder is kind of reserved for typings that are actually loaded from [flow-typed](https://github.com/flow-typed/flow-typed), so I think we have to find another place for that.

webpack.config.js
```diff
@@ -19,6 +20,10 @@ module.exports = (env, argv) => { // eslint-disable-line no-undef
         publicDir = composerConfig.extra['public-dir'];
     }
 
+    const composerLock = JSON.parse(fs.readFileSync(path.resolve(projectRootPath, 'composer.lock')));
+    const suluPackage = composerLock.packages.find((p) => p.name === 'sulu/sulu');
```
danrot, line 24: `package` instead of just `p`, we have not used any abbreviations so far.

webpack.config.js
```diff
@@ -19,6 +20,10 @@ module.exports = (env, argv) => { // eslint-disable-line no-undef
...
+    const suluVersion = suluPackage ? suluPackage.version : 'unknown';
```
danrot, line 25: Wouldn't it make more sense to return `undefined` in case the package version is not known? Then it would be easier to handle that case differently, in case that is necessary at some point.

webpack.config.js
```diff
@@ -47,6 +52,9 @@ module.exports = (env, argv) => { // eslint-disable-line no-undef
                 fileName: outputPath + '/manifest.json',
             }),
             new CleanObsoleteChunksPlugin(),
+            new webpack.DefinePlugin({
+                SULU_ADMIN_BUILD_VERSION: JSON.stringify(suluVersion),
```
danrot, line 56: Why does the version need to be stringified? Shouldn't that already be a string? :thinking:

danrot (changes requested):

webpack.config.js
nnatter: have not thought about that - `undefined` sounds good too

webpack.config.js
nnatter: the plugin does a direct text replacement, therefore the quotes must be included in the value. see the blue box in the docs: https://webpack.js.org/plugins/define-plugin/

flow-typed/globals.js
nnatter: as far as i understand, `flow-typed` will store its definitions in `flow-typed/npm/*`. but i am fine with any other path too - what would you suggest?

nnatter: as far as i understand, `flow-typed` will store its definitions in `flow-typed/npm/*`. the flow documentation just states that the `flow-typed` directory is a convention for storing definitions. but i am fine with any other path too - what would you suggest?

danrot: Just googled, that's the way it is supposed to be, even found a [flow documentation page](https://flow.org/en/docs/libdefs/creation/) about it.

src/Sulu/Bundle/AdminBundle/Resources/js/index.js
```diff
@@ -310,6 +310,16 @@ function processConfig(config: Object) {
 }
 
 function startAdmin() {
+    if (Config.suluVersion !== SULU_ADMIN_BUILD_VERSION) {
```
danrot, line 313: I guess now every developer always gets error messages like this:

```
Sulu administration interface: JavaScript build version mismatch
JavaScript build of the Sulu administration interface does not match the version of the Sulu backend.
Backend version: _._._, JavaScript build version: undefined

Have you forgotten to update the build while upgrading your application?
https://docs.sulu.io/en/latest/upgrades/upgrade-2.x.html
```

Also, the `composer.lock` file is probably not up to date, if I switch the branch locally, which might lead to false positives. Wondering if we can't implement that more gracefully...

Options that come to my mind:

- Ignore mismatch when one of both operands does not seem to be a version
- Ignore mismatch when webpack runs in development mode
- Maybe only show a warning instead of an error? (Might still work, but not sure if people will read the warning)

Thinking about the last point, I am wondering if (especially) PHP developers look at the console tab in the dev tools of the browser. I often have the feeling they don't care about what's written there... That being said, maybe we can move that check somehow to the server side? Although that might be even less graceful...

nnatter: > Ignore mismatch when one of both operands does not seem to be a version

Think I would rather use `_._._` as default value. to solve this That would be consistent to the `SuluVersionPass`.

> Ignore mismatch when webpack runs in development mode

Not sure about this, as the mechanism will not work anymore if somebody commits a watch build by mistake. I think this happens quite often unfortunately 🙈 

> Maybe only show a warning instead of an error? (Might still work, but not sure if people will read the warning)

I used a warning initially. Bun I dont think people will pay attention to thatunfortunatelyy 🙈 

> Thinking about the last point, I am wondering if (especially) PHP developers look at the console tab in the dev tools of the browser.

Probably not very often. But I hope that some of them will have a look at the console if something does not work in the frontend 🙂

danrot: Changing the default to `_._._` would probably be the minimum we'd have to do... Maybe let's just try that, and see how it behaves then.

I could still imagine that false positives appear during development. I think we have to be careful that this error is not shown too often, otherwise it will just be accepted that it is there :see_no_evil: Therefore I would tend to show it too less than too often.

danrot (changes requested):

src/Sulu/Bundle/AdminBundle/Resources/js/index.js
nnatter: Changed the default value 🙂  

If this matching-strategy proves to be working, we could still add additional logs in the backend in the future. Maybe it would also make sense to display an icon beside the version in the administration interface. But I think I would not yet do this in this PR.

danrot (approved):

Issue #5623, opened 2020-11-12
Versioning function missing in 2.2


GeorgePeter:
# Problem description

Versioning cannot be enabled as described in https://docs.sulu.io/en/2.2/cookbook/jackrabbit.html
When following the guide it seems to work fine, but there is no possibility to revert to an earlier version.

# Proposed solution

Create a possibility to revert a page to a previous version as stated in the documentation here:
https://docs.sulu.io/en/2.2/bundles/page/versioning.html

danrot:
Are you talking about the articles or pages?

PR #5639, opened 2020-11-18
Adjust SettingsVersions field type for better configurability

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | https://github.com/sulu/sulu/issues/5623
| License | MIT

#### What's in this PR?

Adjusts the `PageSettingsVersions` field type for better configurability. This is needed for the versioning and restoring in articles.

#### Why?

Because now the field type can be used to restore other documents like `articles`.

#### Example Usage

~~~xml
                <property name="versions" type="settings_versions" onInvalid="ignore">
                    <params>
                        <param name="resource_key" type="string" value="page_versions"/>
                        <param name="list_key" type="string" value="page_versions"/>
                        <param name="user_settings_key" type="string" value="page_versions"/>
                    </params>
                </property>
~~~

#### BC Breaks/Deprecations

`PageSettingsVersions` has been renamed to `SettingsVersions`.

Some hard coded values have been changed to be configurable through the `schemaOptions`.


src/Sulu/Bundle/PageBundle/Resources/js/index.js
```diff
@@ -44,7 +44,7 @@ initializer.addUpdateConfigHook('sulu_page', (config: Object, initialized: boole
     );
 
     if (config.versioning) {
-        fieldRegistry.add('page_settings_versions', PageSettingsVersions);
+        fieldRegistry.add('settings_versions', SettingsVersions);
```
nnatter, line 47: not sure if we should move this to another bundle now that it is reusable 🤔

src/Sulu/Bundle/PageBundle/Resources/js/containers/Form/fields/SettingsVersions.js
```diff
@@ -40,6 +40,75 @@ class PageSettingsVersions extends React.Component<Props> {
...
+        const {
+            schemaOptions: {
+                list_key: {
+                    value: listKey,
+                } = {},
+            },
+        } = this.props;
+
+        if (listKey === undefined || typeof listKey !== 'string') {
+            throw new Error(
```
danrot, line 72: Just an idea: Shouldn't we fallback to the `resourceKey` in case a separate `listKey` is not given? Would make the configuration in XML a bit easier, if all three introduced paramters are the same (which might be the case most of the time).

src/Sulu/Bundle/PageBundle/Resources/js/containers/Form/fields/SettingsVersions.js
```diff
@@ -40,6 +40,75 @@ class PageSettingsVersions extends React.Component<Props> {
...
+        const {
+            schemaOptions: {
+                user_settings_key: {
+                    value: userSettingsKey,
+                } = {},
+            },
+        } = this.props;
+
+        if (userSettingsKey === undefined || typeof userSettingsKey !== 'string') {
+            throw new Error(
```
danrot, line 91: Same idea as above, what about a fallback (probably to the `listKey` in this case).

src/Sulu/Bundle/PageBundle/Resources/js/containers/Form/fields/SettingsVersions.js
```diff
@@ -40,6 +40,75 @@ class PageSettingsVersions extends React.Component<Props> {
...
+            );
+        }
+
+        return userSettingsKey;
+    }
+
+    @computed get parentRoute(): string {
+        const {router} = this.props;
+
+        if (!router || !router.route || !router.route.parent || !router.route.parent.name) {
```
danrot, line 103: I think it is safe to use the `?.` operator now, so this could be simplified to `!router?.route?.parent?.name`, if you want to type less :slightly_smiling_face:

danrot: But when reading this, are you sure this is true for all cases? :thinking: I guess we can leave it like this for now, because this probably works for pages and articles, but maybe we have to introduce a separate parameter for that as well at some point.

src/Sulu/Bundle/PageBundle/Resources/js/index.js
danrot: Hm, not sure about this... I think I would leave it here for the time being.

The thing is that the `versioning` parameter checked in the line before is also in the PageBundle, and we have to avoid registering it when this configuration is set to false, in order for the `on-invalid` attribute in the XML to work correctly.

I am also not sure where to move it instead, and I think the ArticleBundle is already depending a lot on the PageBundle (am I right?). Don't like that fact, but I think it is still better to keep it this way, than introducing another dependency right away.

danrot (changes requested):

src/Sulu/Bundle/PageBundle/Resources/js/containers/Form/fields/SettingsVersions.js
Prokyonn: Yeah I like the idea :)

src/Sulu/Bundle/PageBundle/Resources/js/containers/Form/fields/SettingsVersions.js
Prokyonn: No it probably doesn't, because not everyone has a parent route...
At first I thought we could also have a `route` as a param in the xml configuration, but this is also not possible because the article bundle uses a different route for each `articleType`.

For now I would also go with this approach.

src/Sulu/Bundle/PageBundle/Resources/js/index.js
nnatter: is alright for me 🙂

src/Sulu/Bundle/PageBundle/Resources/js/index.js
```diff
@@ -44,7 +44,7 @@ initializer.addUpdateConfigHook('sulu_page', (config: Object, initialized: boole
     );
 
     if (config.versioning) {
-        fieldRegistry.add('page_settings_versions', PageSettingsVersions);
+        fieldRegistry.add('page_settings_versions', SettingsVersions);
```
danrot, line 47: You've changed it back to `page_settings_versions`, right? Renaming it to `settings_versions` was for me, but we can also keep it like this. But then I would also not rename the `PageSettingsVersions` components. Either way is fine for me, but I would keep it consistent in this place.

Prokyonn: Oops that was by mistake, I would go with `settings_versions`. Otherwise it is kinda confusing that we use `page_settings_versions` in the `article-template'

danrot (approved):
Just tested it, and it works fine for me :+1: Can whoever tests the PR in the ArticleBundle merge both of them together? Just to make sure everything works as expected, and we don't have to create another PR.

Issue #5426, opened 2020-07-23
Mandatory field not working correctly.


GillesKnockaert:
| Q | A
| --- | ---
| Sulu Version | Specific version or SHA of a commit
Sulu 2.1

| Browser Version | Browser name and version
Latest version of Chrome

# Actual Behavior

A mandatory field is mandatory when you create the page. But when you go back to the page you can just remove the value of the mandatory field so it is empty and save it.

# Expected Behavior

A mandatory field should always have a value.

# Steps to Reproduce

Create a temlpate with a mandatory field on it let's say a single media select. then when you create a page using that template, you need to add a image or you can't save it. So when you saved the page, go back to the page and remove the image. now you can save the page but the mandatory field is empty???

# Possible Solutions

Make sure that mandatory fields always have a value else you can't save them.

alexander-schranz:
@GillesKnockaert currently mandatory fields only working for simple fields which are saved as empty string. Field Types like selection which use array or objects be saved are currently not possible to be mandatory as current limitation of the json schema which is not type specific. We hope we can in future provide here something to fix this problems.

danrot:
This has been for the mentioned media selection in #5640. Thanks for reporting!

Issue #2938, opened 2016-09-22
"mandatory" & "minOccurs" attributes are ineffective in non-text-only properties


Edvinas9:
| Q | A |
| --- | --- |
| Bug? | yes |
| New Feature? | no |
| Sulu Version | 1.3.0 |
| Browser Version | Chrome 53.0.2785.116 m (64-bit) on Windows 10 |
#### Actual Behavior

Neither `mandatory="true"` nor `minOccurs="1"` attributes on contact selection content type in a page template doesn't prevent user from saving the page when no contacts has been selected.
#### Expected Behavior

Either `mandatory="true"` or `minOccurs="1"` attributes on contact selection content type in a page template should ensure that at least 1 contact has been selected and prevent user from saving the page if the latter condition hasn't been fulfilled.
#### Steps to Reproduce

Open any page template and add new field of type `contact`:

``` xml
<property name="client" type="contact" mandatory="true" minOccurs="1">
    <meta>
        <title lang="en">Client</title>
    </meta>
    <params>
        <param name="contact" value="false"/>
        <param name="account" value="true"/>
    </params>
</property>
```

In the CMS, open any page which uses that template. Make sure the contact field is empty (no contacts are selected) and hit `Save` button. In such scenario, `mandatory` field has to ensure at least 1 contact has been selected and prevent `Save` action from occurring if not. It doesn't though.

wachterjohannes:
the usage of `minOccurs` (also `maxOccurs`) would be nice and i agree your thoughts about that. but with the `mandatory` i have a little problem because are multiple selected fields also valid (if mandatory) if no item is selected? `mandatory` was introduced originally to ensure that the value is not null and a empty array is also not null. /cc @chirimoya @danrot what do you think?

danrot:
IMO the `mandatory` attribute in the XML file has a more semantical meaning than just not null. If there is no contact assigned I would consider it empty, making the `mandatory` validation fail. Otherwise it has really no meaning when added on these fields.

wachterjohannes:
ok that seems to be an argument this should then also be implemented in following selections:

* Contact
* Media
* Internal link

danrot:
There seems to be the same problem for the single internal link, see https://github.com/sulu/sulu/issues/3249

alexander-schranz:
I personally think we should get rid of general max and min occurs on attributes on the property tag. And the field types which support such kind of things should have a param for it e.g.:

```xml
<property name="pages" type="page_selection">
    <params>
[...]

PR #5640, opened 2020-11-18
Add json-schema validation for media_selection

luca-rath:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes parts of #5426 and #2938 
| License | MIT

#### What's in this PR?

Add json-schema validation for `media_selection`

#### Why?

Because mandatory was not working for `media_selection`


luca-rath:
After talking to @danrot and @nnatter, we came up with two solutions to fix this problem:

# Solution 1

A `PropertyMetadataMapper` is a service, which maps an `ItemMetadata` to a `PropertyMetadata` instance. There is a `PropertyMetadataMapperRegistry` which holds all tagged `PropertyMetadataMapper`s. Every `PropertyMetadataMapper` needs to be tagged with the name of a field type. This ensures, that every field type has either zero or one corresponding `PropertyMetadataMapper`.

The advantages of this approch are, that the `PropertyMetadata` is exchangeable. If required, the `PropertyMetadataMapper` of the `media_selection` field type could just return an instance of `MediaSelectionPropertyMetadata`, which extends from `PropertyMetadata`.

A disadvantage is, that schema creation logic can't be shared easily, because every field type has it's own corresponding `PropertyMetadataMapper`.

So if you have a custom field type in your project, which want's to reuse the `pattern` schema of the `text_line` field type, you would have to create a custom `PropertyMetadataMapper`, which extends from the `TextLinePropertyMetadataMapper` (this is only possible, if that's an own class. If the `TextLineContentType` implements the `PropertyMetadataMapperInterface`, that's not possible anymore). But if the `text_line` field type has some json schema you don't want for your field type, you have to manually remove that. Also, the json schema of `text_line` can change from time to time, so you always have to react to those changes. Otherwise this could break your custom field type.

This is the only option to reuse some json schema of a field type. Otherwise you have to duplicate that code in your custom `PropertyMetadataMapper`.

# Solution 2

A `PropertyMetadataEnhancer` is a service, which creates a new instance of `PropertyMetadata` based on the previous `PropertyMetadata` and a corresponding `ItemMetadata`. But unlike solution 1, a `PropertyMetadataEnhancer` is not directly linked to a field type. This allows multiple field types to share a specific `PropertyMetadataEnhancer` but still keeps the possibility to also use a custom `PropertyMetadataEnhancer` specific for that field type. This behaviour can be achieved with a public `supports()` method on the `PropertyMetadataEnhancer`.

Assuming there is a `PatternPropertyMetadataEnhancer` which is registered in sulu for the `text_line` field type. If you want your custom field type to also use the `PatternPropertyMetadataEnhancer`, you just have to decorate that service and override the `supports()` method. If you have some specific json schema, you can just register a new `PropertyMetadataEnhancer` service which supports only your custom field type.

danrot:
I've got the feeling that you are already biased towards solution 2, and therefore have sneaked in a few assumptions in Solution 1, that are not true in my opinion.

In solution 2 you've said that

> If you want your custom field type to also use the PatternPropertyMetadataEnhancer, you just have to decorate that service and override the supports() method.

which is practically almost the same as in solution 1, where you have mentioned that this is only possible if the `PropertyMetadataMapper` is a separate class and not part of the concrete `ContentType`. That's the most crucial part that is not true IMO, because you can still decorate the service exactly in the same way as you mentioned in Solution 2:

```php
interface PropertyMetadataMapper {
    public function mapProperty(ItemMetadata $itemMetadata): PropertyMetadata
}

class PatternPropertyMetadataMapper implements PropertyMetadataMapper { /* ... */ }

class TextLinePropertyMetadataMapper implements PropertyMetadataMapperInterface {
    private $innerPropertyMetadataMapper;

    // Maybe it would make even more sense to use the composite pattern here, and make that parameter an array of PropertyMetadataMapperInterfaces
    public function __construct(PropertyMetadataMapperInterface $innerPropertyMetadataMapper) {
        $this->innerPropertyMetadataMapper = $innerPropertyMetadataMapper;
    }

    public function mapProperty(ItemMetadata $itemMetadata): PropertyMetadata
    {
        $propertyMetadata = $this->innerPropertyMetadataMapper->mapProperty($itemMetadata);

        /* ... */
    }
}

$metadataMapper = new TextLinePropertyMetadataMapper(
    new PatternPropertyMetadataMapper() // This could be anything, even a `ContentType`, as long as it implements the `PropertyMetadataMapperInterface
);
```

So it is not necessary at all to use inheritance with that approach, as you've suggested in your text above. Also, in both solutions you would have to extract the `PatternPropertyMetadataMapper` resp. `PatternPropertyMetadataEnhancer` if you want to reuse that properly. That is also exactly the same pattern you've suggested as an advantage in solution 2.

I would still have a slight difference towards Solution 1, because for most use cases it seems to be the simpler one. It would probably also be more efficient, because not every `PropertyMetadataEnhancer` would have to be called for every property. However, the latter advantage might not be that important, because the result is cached afterwards anyway. That being said, I think we still have to make sure that the `PropertyMetadata` is immutable, which can be done in both solutions.

But since you seem to feel so much better about the second solution, it is also OK for me to proceed with that one.

luca-rath:
I don't think your code example would work.

If you decorate the `MediaSelectionContentType` with that `TextLinePropertyMetadataMapper`, there will definitely be an error, because the `TextLinePropertyMetadataMapper` doesn't implement the `ContentTypeInterface`, the `ContentTypeExportInterface` nor the `PreResolvableContentTypeInterface`. So sulu would e.g. try to call `TextLinePropertyMetadataMapper::read()`, which is not a function.

danrot:
1. I am only talking about the general concept of decoration, that does not mean that the symfony decoration (which indeed replaces the service) has to be used.
2. That is only true, if the `ContentType` implements the `PropertyMetadataMapperInterface` instead of having that in a separate class (and the symfony decoration is used as mentioned in point 1). And that also applies to your second solution: If the `PropertyMetadataEnhancerInterface` is implemented by the `ContentType`, you'll have the exact same problem, if you try to decorate that `ContentType` with a pure `PropertyMetadataEnhancer`.

luca-rath:
yes, but with solution 1, you will probably always have specific `PropertyMetadataMapper`s, because you can only have one. With solution 2, it's more likely, that the `PropertyMetadataEnhancer`s are specific to functionality (like pattern e.g.), not to the field type.

danrot:
You can build single implementations per content type or extract some of the most common logic into separate implementations being reused by others in both solutions, so I would say that's not a point for either one of them.

IMO the main difference between those two is the way you would reuse existing implementations for custom content types, and if you think about it, doing that is about the same amount of work in both cases (using `Custom` as placeholder):

- Solution 1
    - Create a `CustomPropertyMetadataMapper` decorating the `PatternPropertyMetadataMapper`
    - Adding a new service definition for the `CustomPropertyMetadataMapper` with a tag referencing the correct content type
- Solution 2
    - Create a `CustomPropertyMetadataEnhancer` decorating the `PatternPropertyMetadataEnhancer` and overriding the `supports` method
    - Adding a new service decorator definition for the `CustomPropertyMetadataEnhancer`

So also from that point of view I would not say that one solution is much better than the other, they are more or less equal IMO.

If I try to make a list of exclusive advantages to both solutions I come up with this:

- Solution 1
    - Better performance because less methods have to be called
    - Implementations are more explicit (because I see what the `TextLinePropertyMetadataMapper` depends on in its constructor arguments)
    - Easier to debug (a very underrated consideration IMO), because I don't have to step through every `PropertyMetadataEnhancerInterface` to just see that it returns without doing anything
- Solution 2
    - Might lead to a less classes
        - Although if we only need the `PatternPropertyMetadataMapper` in solution 1 we can add a tag to that service for the new content type without creating a new class, a complete new class in solution 1 is only necessary if multiple existing `PropertyMetadataMapper`s are needed
        - Creating less classes is not the highest priority on my list, would rank stuff like explicity and ease-to-debug higher

I think I've made my points. If you don't understand some of my arguments or think that they are wrong I am happy to discuss them in a call, otherwise we can try to make a decision on top of this in arrangement with @nnatter.

nnatter:
After thinking about this again last evening, I think i would go for **Solution 1** because of the following reasons:
- Its easier to understand for (external) developers
- Its consistent to the association between content-type and field-type
- The registry approach is already used in a lot of other parts of the system
- The process for extension is consistent to other parts of the system

From my understanding, the additional complexity of solution 2 comes with the following advantages:
- Allows to provide/set the schema for different content types in a single class
  - Do we have any usecase for this in the core?
- Allows multiple mappers for a single content type
  - This also means that the mapper needs to understand how to merge the schema of previous mappers
  - Do we have any usecase of this in the core?

luca-rath:
To answer your questions:

1) A usecase would be all `selection` and `single_selection` field types.
2) A usecase would be the `tag_selection`, because it would override the schema of the `SelectionPropertyMetadataEnhancer`, because it's values are strings, not ints

nnatter:
1. In the case of the `selection` and `single_selection` field types, we could also just register the same `PropertyMetadataMapper` class with different tags.

2. Thats only necessary because of answer 1, right? 🤔

luca-rath:
1) yes we could, but then every developer who created a custom selection field type would have to register that by himself. With solution 2, we could just use the FieldTypeRegistry to check if the `PropertyMetadataEnhancer` supports that field type.

2) yes

luca-rath:
But since you both prefer solution 1, I'm also fine with doing so.

nnatter:
> yes we could, but then every developer who created a custom selection field type would have to register that by himself. With solution 2, we could just use the FieldTypeRegistry to check if the PropertyMetadataEnhancer supports that field type.

If we can check this in the `PropertyMetadataEnhancer`, it should be possible too to register the services with the respective tags in the `AdminBundle` 🙂

luca-rath:
> IMO the main difference between those two is the way you would reuse existing implementations for custom content types, and if you think about it, doing that is about the same amount of work in both cases (using `Custom` as placeholder):
> 
> * Solution 1
>   
>   * Create a `CustomPropertyMetadataMapper` decorating the `PatternPropertyMetadataMapper`
>   * Adding a new service definition for the `CustomPropertyMetadataMapper` with a tag referencing the correct content type
> * Solution 2
>   
>   * Create a `CustomPropertyMetadataEnhancer` decorating the `PatternPropertyMetadataEnhancer` and overriding the `supports` method
>   * Adding a new service decorator definition for the `CustomPropertyMetadataEnhancer`

@danrot When using solution 1, there will probably never be a `PatternPropertyMetadataMapper`, because each field type can just have one single `PropertyMetadataMapper` and because the `text_line` field type requires other json schema as well, there will be a `TextLinePropertyMetadataMapper`, not a `PatternPropertyMetadataMapper`

luca-rath:
> If we can check this in the `PropertyMetadataEnhancer`, it should be possible to register the services with the respective tags in the `AdminBundle` too 🙂

You're right, we probably can do that.

danrot:
> > yes we could, but then every developer who created a custom selection field type would have to register that by himself. With solution 2, we could just use the FieldTypeRegistry to check if the PropertyMetadataEnhancer supports that field type.
> 
> If we can check this in the `PropertyMetadataEnhancer`, it should be possible to register the services with the respective tags in the `AdminBundle` too

I think the elegance in the approach of solution 1 is that the logic can be reused by simply adding another tag to the service. In solution two there are two possibilities:

1. We are talking about a core feature, then it can be hardcoded
2. We are talking about a custom field, then we have to add a service definition AND implement a new class.

> > IMO the main difference between those two is the way you would reuse existing implementations for custom content types, and if you think about it, doing that is about the same amount of work in both cases (using `Custom` as placeholder):
> > 
> > * Solution 1
> >   
> >   * Create a `CustomPropertyMetadataMapper` decorating the `PatternPropertyMetadataMapper`
> >   * Adding a new service definition for the `CustomPropertyMetadataMapper` with a tag referencing the correct content type
> > * Solution 2
> >   
> >   * Create a `CustomPropertyMetadataEnhancer` decorating the `PatternPropertyMetadataEnhancer` and overriding the `supports` method
> >   * Adding a new service decorator definition for the `CustomPropertyMetadataEnhancer`
> 
> @danrot When using solution 1, there will probably never be a `PatternPropertyMetadataMapper`, because each field type can just have one single `PropertyMetadataMapper` and because the `text_line` field type requires other json schema as well, there will be a `TextLinePropertyMetadataMapper`, not a `PatternPropertyMetadataMapper`

There could be a `PatternPropertyMetadataMapper`, which can be passed as a constructor argument to each other `PropertyMetadataMapper`, so that logic could be reused. However, it is questionable to me if it is worth it.

luca-rath:
Sounds reasonable. I implemented solution 1 in the current revision.

src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/PropertyMetadata.php
```diff
@@ -23,10 +23,16 @@ class PropertyMetadata
...
+    /**
+     * @var array<string, mixed>|null
+     */
+    private $jsonSchema;
+
+    public function __construct(string $name, bool $mandatory, ?array $jsonSchema = null)
     {
         $this->name = $name;
         $this->mandatory = $mandatory;
+        $this->jsonSchema = $jsonSchema;
```
danrot, line 35: Not sure what to think about this... I get that this is the fastest way of doing stuff, but we are mixing how these classes should be used. When following the current paradigm, then there would have to be a `ObjectMetadata` class along with the already existing `ArrayMetadata` and `ConstMetadata`, and all of them should inherit from the `PropertyMetadata`...

I think we could go for an approach where we are completely using arrays like in this property, but then we'd probably have to go for arrays completely. By that I mean no `SchemaMetadata` or `PropertyMetadata` at all, and handling all differently. A reason that would be problematic is that the [`SchemaMetadata` is holding some logic](https://github.com/sulu/sulu/blob/11fdd6d8de723374529bc0b3e4bfeecd1f6c28aa/src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/SchemaMetadata.php#L40-L85), which is actually quite nice.

So either approach would be fine for me, but mixing them does not feel like a good idea to me.

danrot (changes requested):

src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/PropertyMetadata.php
luca-rath: I would go for option 1, but without renaming the `PropertyMetadataMapper` to `ObjectMetadataMapper`, right? Because it still maps a `PropertyMetadata` to an `ObjectMetadata`?

src/Sulu/Bundle/AdminBundle/Tests/Unit/Metadata/SchemaMetadata/PropertyMetadataMapperRegistryTest.php
```diff
@@ -0,0 +1,91 @@
...
+        $propertyMetadataMapper = $this->prophesize(PropertyMetadataMapperInterface::class);
+
+        $this->serviceLocator->get('mapper')->willReturn($propertyMetadataMapper->reveal());
+
+        $this->assertSame($propertyMetadataMapper->reveal(), $this->propertyMetadataMapperRegistry->get('mapper'));
+    }
+
+    public function testGetNotFound(): void
+    {
+        $this->serviceLocator->get('mapper')->willThrow($this->createNotFoundException());
```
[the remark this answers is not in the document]
luca-rath: If you have a look at the `ServiceLocatorTrait` at https://github.com/symfony/symfony/blob/49eafee03323253328d76ec9cb4c515f94d2f202/src/Symfony/Contracts/Service/ServiceLocatorTrait.php#L117, it also throws such an exception

src/Sulu/Bundle/AdminBundle/Resources/config/services.xml
```diff
@@ -92,9 +92,17 @@
             <argument type="service" id="sulu_admin.form_metadata.form_metadata_mapper"/>
         </service>
 
+        <service
+            id="sulu_admin.property_metadata_mapper_registry"
+            class="Sulu\Bundle\AdminBundle\Metadata\SchemaMetadata\PropertyMetadataMapperRegistry">
+            <argument type="tagged_locator" tag="sulu_admin.property_metadata_mapper" index-by="type"/>
```
nnatter, line 98: not sure if there are any advantage/disadvantages, but we used `type="tagged"` with a `default-index-method` when implementing the list filter feature:
https://github.com/sulu/sulu/pull/5035/files#diff-634082d921f72e34116b5f1b8e0bc715867d2b93182f5b6ab90445caf74f02bdR33-R43

it would be nice to keep these things consistent for new features

luca-rath: +1 for using the `default-index-method`, but that might be a bit confusing, since probably most `PropertyMetadataMapper`s are implemented in their corresponding content types.

About the `type="tagged"`, I would prefer keeping the `tagged_locator` for performance reasons, because the `ServiceLocator` is a lazy service container.

WDYT?

src/Sulu/Bundle/AdminBundle/Tests/Unit/Metadata/SchemaMetadata/PropertyMetadataMapperRegistryTest.php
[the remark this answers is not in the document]
danrot: That's weird :see_no_evil: But ok, then we have to do it like that.

src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/PropertyMetadataMapperInterface.php
```diff
@@ -0,0 +1,19 @@
...
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\AdminBundle\Metadata\SchemaMetadata;
+
+use Sulu\Component\Content\Metadata\PropertyMetadata as ContentPropertyMetadata;
+
+interface PropertyMetadataMapperInterface
+{
+    public function mapPropertyMetadata(ContentPropertyMetadata $propertyMetadata): ObjectMetadata;
```
danrot, line 18: That could return any `PropertyMetadata`, not only `ObjectMetadata`, or am I wrong?

src/Sulu/Bundle/MediaBundle/Content/Types/SingleMediaSelection.php
```diff
@@ -123,4 +126,27 @@ protected function decodeValue($value)
 
         return \json_decode($value, true);
     }
+
+    public function mapPropertyMetadata(PropertyMetadata $propertyMetadata): ObjectMetadata
+    {
+        $mandatory = $propertyMetadata->isRequired();
+
+        $jsonSchema = [
+            'type' => 'object',
```
danrot, line 135: The fact that you have to pass `object` as a type here, although you are generating a `ObjectMetadata` below might lead to some confusing situations... What if we create some like that:

```php
return new ObjectMetadata($propertyMetadata->getName(), $mandatory, ['type' => 'string']);
```

That's why I think it might be confusing to build it this way. Would it be good enough for the use cases we have to build it like this:

```php
new ObjectMetadata(
    $propertyMetadata->getName(),
    $mandatory,
    [
        new PropertyMetadata('id', true, 'number'),
        new PropertyMetadata('displayOption', false, 'string'),
    ]
);
```

In that example I assume that we are adding a third argument to the `PropertyMetadata` for the type. Instead we could also add something like a `StringMetadata` and `NumberMetadata`.

Then it would be consistent IMO. I know that this might be a little bit cumbersome if somebody is a JSON schema pro, and they just want to pass an array as schema. I would totally understand if we want to support that better, but then I would deprecate all of the classes (including `SchemaMetadata`, `PropertyMetadata`, ...) and only pass one - potentially huge - array instead.

My suggestion would be to try it the way I have mentioned above, and if that turns out to be too cumbersome in the future, we could still deprecate all these classes, and build it in a way that just an array is passed (would probably need a service then, that would help with merging schemas etc., basically replacing the logic that is happening in the `SchemaMetadata` classes).

src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/PropertyMetadata.php
danrot: I think I kind of answered that in https://github.com/sulu/sulu/pull/5640#discussion_r530859212

src/Sulu/Bundle/AdminBundle/Resources/config/services.xml
danrot: Does it really have performance advantages? I think Symfony only creates that service when it is used (isn't that the advantage of the container being split since Symfony 3.4(?))? It might be a bit faster if there are loads of `PropertyMetadataMapper` registered, but I could imagine that it is slower if most of the `PropertyMetadataMapper`s are used (which I guess will be the case, especially in the beginning), because the laziness comes with some overhead.

That being said, I don't think that consistency is super important here, and if it brings a performance advantage I would keep it this way.

Regarding the `default-index-method`: That approach really gets a bit problematic, if we add more interfaces to the `ContentType`, at some point there might be naming collisions. But if we say we want to split the `ContentType` classes into multiple classes in the future, it would be fine for me to do it this way at the moment.

src/Sulu/Bundle/MediaBundle/Content/Types/SingleMediaSelection.php
luca-rath: Your suggestions works very well for simple json schemas ... But if you have a short look at https://github.com/sulu/sulu/pull/5640/files#diff-6f4f6054791755b1732dec2b54938638b8a7dd63b892bcf3b0ef7d63e1358946R200-R215, it already gets a little bit more complicated. There is already an `ArrayMetadata`, but that would have to be fully modified to represent e.g. the schema needed for the `ids` property. I don't even think, that this can be done without a bc break

danrot: Are you talking about the `uniqueItems` and `minItems` properties? These can be optional, then it's not a BC break, or am I missing something obvious :thinking:

src/Sulu/Bundle/AdminBundle/Resources/config/services.xml
luca-rath: Ok, probably you're right, that using a `ServiceLocator` won't have performance advantages with so few services, but it's also easier to work with in the `PropertyMetadataMapperRegistry`, because the `ServiceLocator` already comes with exactly those methods the `PropertyMetadataMapperRegistry` also has ...

Because the `ContentType` will probably be removed in `3.0`, we can assume there will be multiple classes per field type in future, so I guess naming collisions shouldn't be a problem

src/Sulu/Bundle/MediaBundle/Content/Types/SingleMediaSelection.php
luca-rath: The `ArrayMetadata` can't handle scalar arrays as far as I see

luca-rath: Also, we would have to add a parameter to `ArrayMetadata` and to all other `Metadata`s for every json schema keyword we want to support, so this would be a big limitation

danrot: Wouldn't something like that work:

```php
new ObjectMetadata(
    $propertyMetadata->getName(),
    $mandatory,
    [
        new ArrayMetadata('ids', true, new SchemaMetadata([], [], [], 'string')),
        new PropertyMetadata('displayOption', false, 'string'),
    ]
);
```

I know that we would have to add a new parameter to the `SchemaMetadata`, but that could also easily be hidden in factory methods (or we just leave it that way for now). That would need a small change, but I still don't think that it would be a breaking one.

Regarding the limitation: That's correct, but I am not sure if that will be a real deal breaker. Especially since in order to properly handle new keywords we sometimes might have to add some stuff in the `Form` javascript implementation.

As said, that's why I would at least try to implement it that way. My guess is that mixing classes for some stuff and plain arrays in other cases would cause more confusion that it helps. With this solution we would keep this consistent for now, and I am happy to change it later to a pure array variant (deprecating all the classes in the `Sulu\Bundle\AdminBundle\Metadata\SchemaMetadata`) if it really turns out to be cumbersome. But I think it would be a pity to lose consistency now.

luca-rath: ok 👍

src/Sulu/Bundle/AdminBundle/Resources/config/services.xml
luca-rath: @nnatter would you be okay with that?

nnatter: yeah sure - as far as i understand `type="tagged"` and `type="tagged_locator"` does not make any difference for the developer anyway 🙂

luca-rath: exactly, it just simplifies the `PropertyMetadataMapperRegistry`

danrot:
@luca-rath Code LGTM now, but the tests are failing. Can you fix that?

luca-rath:
> @luca-rath Code LGTM now, but the tests are failing. Can you fix that?

I know, already working on it

danrot:
Added #5426 and #2938 to the descrption, because this PR fixes parts of these issues. But they mention even more fields than just `MediaSelection` and `SingleMediaSelection`, I think we should later add this kind of validation for them as well.

Going to test it right now, and if I don't find any errors I am going to merge it.

danrot (approved):

src/Sulu/Bundle/MediaBundle/Resources/config/services.xml
```diff
@@ -323,6 +324,7 @@
 
             <tag name="sulu.content.type" alias="single_media_selection"/>
             <tag name="sulu.content.export" format="1.2.xliff" translate="false" />
+            <tag name="sulu_admin.property_metadata_mapper" type="single_media_selection"/>
```
alexander-schranz, line 327: The `type="single_media_selection"` is something I personally would not longer do as it make an interface not be able to be autoconfigureable I would add a static `getType` to the Interface instead this make it able to autowire/autoconfigure the Interface and no service definition need to be created when used in a project.

luca-rath: But because those `PropertyMetadataMapper`s are mostly implemented in their corresponding content types, that may lead to confusion or even naming conflicts, because most content types implement quite a lot interfaces

nnatter: havent we added a `default-index-method` like discussed in https://github.com/sulu/sulu/pull/5640#discussion_r529658444? 🤔  
if i remember right, you dont need to implement the method if you set the index explicitly on the service

danrot: You two seemed to agree in that comment, that's why I thought it is OK as it is now :see_no_evil: But apart from that, you can still add a static `getDefaultIndexName` method now, and it will work with autoconfiguring, if I am not mistaken.

luca-rath: I thought, it would only make sense, if that method is also declared in the interface ... But I have to touch those files again very soon to implement a `PropertyMetadataMapper` for `selection` and `single_selection` field types, so I can add the `default-index-method` there without declaring the corresponding method in the `PropertyMetadataMapperInterface`

nnatter: you are right, we cannot add it to the `PropertyMetadataMapperInterface` and it looks like adding a `getDefaultIndexName` should already work: https://symfony.com/doc/current/service_container/tags.html#tagged-services-with-index

dont have a strong opioniern on adding am explicit `default-index-method` 🙂

danrot: Well, I thought you *can* add a static method to the `PropertyMetadataMapperInterface`, but then you also force people to implement it, although they might want to register the service manually. Don't know how often that's the case, and the currently implemented variant works for me too.

PR #5645, opened 2020-11-20
Run tests with github actions

luca-rath:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Run tests with github actions

#### Why?

Because we want to migrate from travis to github actions


danrot:
LGTM, do you have any plan on how to handle that for the other branches? I am sure there'll be a conflict when backmerging that...

Maybe we can do backmerges to all branches first, then merge this PR, and you will create PRs for the other branches including a merge commit? Or should we backmerge and just ignore this changes, since in the `release/2.1` branch the travis file looked different anyway, and we probably need to have a different github actions file as well?

luca-rath:
If we just merge this now and then backmerge only to `release/2.1`, there will be only a short amount of time, where the `release/2.1` branch is failing until i fixed that. Backmerging the pr for `release/2.1` to `release/2.2` and `master` shouldn't then be a problem anymore.

.github/workflows/test-application.yaml
```diff
@@ -28,3 +33,121 @@ jobs:
...
+                test-flags: [ '' ]
+                php-extensions: [ 'ctype, iconv, mysql, gd' ]
+                tools: [ 'composer:v1' ]
+                env:
+                    - {
+                          SYMFONY__PHPCR__TRANSPORT: doctrinedbal,
+                          SYMFONY__DATABASE__DRIVER: pdo_mysql,
+                          SYMFONY__DATABASE__PORT: 3306,
+                    }
+                include:
```
nnatter, line 64: not sure if possible, but i feel like it would be more explicit and predictable if we define all "environments" in the `include`

luca-rath: I don't know ... will try that

.github/workflows/test-application.yaml
```diff
@@ -28,3 +33,121 @@ jobs:
...
+
+            - name: Install and configure PHP
+              uses: shivammathur/setup-php@v2
+              with:
+                  php-version: ${{ matrix.php-version }}
+                  extensions: ${{ matrix.php-extensions }}
+                  tools: ${{ matrix.tools }}
+                  coverage: none
+
+            - name: Cache jackrabbit runnable
```
nnatter, line 109: maybe we should use our jackrabbit image as service?

luca-rath: Jackrabbit can't be used as a service, because the `bin/runtests` script needs to be able to restart jackrabbit after each test suite, which is not possible with the current `bin/runtests` script, if jackrabbit is a docker container. In sulu 2.1, it shouldn't be needed anymore to restart jackrabbit, therefore we can then use it as a service

.github/workflows/test-application.yaml
```diff
@@ -28,3 +33,130 @@ jobs:
...
+                      tools: 'composer:v1'
+                      env:
+                          SYMFONY__PHPCR__TRANSPORT: doctrinedbal
+                          SYMFONY__DATABASE__DRIVER: pdo_mysql
+                          SYMFONY__DATABASE__PORT: 3306
+
+                    - php-version: '7.0'
+                      database: mysql
+                      phpcr-transport: doctrinedbal
+                      composer-flags: ''
```
nnatter, line 65: think the travis ci used `--prefer-lowest --prefer-dist --no-interaction` here

.github/workflows/test-application.yaml
```diff
@@ -28,3 +33,130 @@ jobs:
...
+                      test-flags: ''
+                      php-extensions: 'ctype, iconv, mysql, imagick'
```
nnatter, line 67: think there was not `imagick` in the travis build but not sure if that was intended 😅

.github/workflows/test-application.yaml
danrot: I have already mentioned that in our slack channel, if we add much more tests which talk to jackrabbit, that might be necessary again at some point, because it needs to be restarted when jackrabbit has too much load (don't ask me why that is). So it might also be necessary for 2.1 and later at some point, therefore I would stick with the restartable standalone jar in that case.

.github/workflows/test-application.yaml
danrot: It might be that it was preinstalled on travis... But the above build is running with gd, might make sense to test both cases anyway :slightly_smiling_face:

luca-rath: The `7.4` case also uses `imagick`, so both cases are tested anyway. But i changed `7.0` to `gd` to be consistent with the travis config

.github/workflows/test-application.yaml
luca-rath: I'm fine with that as well

nnatter: i think it would be nice to use the image. first because it would be consistent to the other services, secondly because it would test the image 🙂 

has anybody tested if these errors till occur? as far as i see, the version of jackrabbit that was previously used on travis is at least 4 years old (https://archive.apache.org/dist/jackrabbit/2.12.0/) maybe this was an issue that was fixed? also, if the issue still persists, i suspect that it is rather a problem with our tests than with jackrabbit? 🙈

danrot: Haven't tested it, and i don't know if it also happens with a later version.

luca-rath: I would just use jackrabbit as service starting from `release/2.1`, because no matter if we use it as service or as standalone package, the same problems will occurr, because the current `bin/runtests` script doesn't handle that yet. So if we ever encounter such an error, we have to modify both, the `bin/runtests` and the `test-application.yaml`

danrot:
For me that is fine to merge. Still not sure how to handle the conflict when merging it back though. Should I just revert all the changes, so that these changes are not in the `release/2.1` branch? And we will create a separate PR, that can take over some of the changes made here?

luca-rath:
Sounds good to me, but will there be problems with backmerging then, if we want to change e.g. the `download-languages` job in `release/1.6`?

nnatter:
I think i would do it like this:
- Merge this into `1.6`
- Pull `1.6` into `2.1`
- Edit `test-application.yaml` in new PR targeting `2.1`

danrot:
> Sounds good to me, but will there be problems in the future with backmerging, if we want to change e.g. the `php-cs-fixer` job in `release/1.6`?

I don't think that this should be more of a problem than doing it the other way round. Git will only apply the changes that are new after the merge commit, and when these files diverge a lot between the two branches it will be a mess anyway :see_no_evil:

danrot:
> I think i would do it like this:
> 
> * Merge this into `1.6`
> * Pull `1.6` into `2.1`
> * Edit `test-application.yaml` in new PR targeting `2.1`

This approach would mean that I have to somehow solve the emerging conflicts and push the changes without a review (I think squash merging the merge commit in a PR might not work as intended...).

nnatter:
> This approach would mean that I have to somehow solve the emerging conflicts and push the changes without a review (I think squash merging the merge commit in a PR might not work as intended...).

There shouldn't be a lot of merge conflicts if we keep the existing `.travis.yml` file in the merge commit, right? The content of the `test-application.yaml` might be incorrect for the `release/2.1` branch, but thats what the new PR would change. Anyway, I dont have a strong opinion on this. Guess it should not matter that much 🙂

PR #5648, opened 2020-11-21
Phpstan/custom bundle

rogoit:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

PHPStan level 2 improvements for CustomUrlBundle

### DIFF
--- a/src/Sulu/Bundle/CustomUrlBundle/Controller/CustomUrlController.php
+++ b/src/Sulu/Bundle/CustomUrlBundle/Controller/CustomUrlController.php
@@ -110,15 +110,15 @@ class CustomUrlController extends AbstractRestController implements SecuredContr
      * Create a new custom-url object.
      *
      * @param string $webspace
+     * @param Request $request
      *
      * @return Response
      */
-    public function postAction($webspace, Request $request)
+    public function postAction(string $webspace, Request $request)
     {
         $document = $this->customUrlManager->create(
             $webspace,
-            $request->request->all(),
-            $this->getRequestParameter($request, 'targetLocale', true)
+            $request->request->all()
         );
         $this->documentManager->flush();
 
--- a/src/Sulu/Bundle/CustomUrlBundle/Controller/CustomUrlController.php
+++ b/src/Sulu/Bundle/CustomUrlBundle/Controller/CustomUrlController.php
@@ -109,9 +109,6 @@ class CustomUrlController extends AbstractRestController implements SecuredContr
     /**
      * Create a new custom-url object.
      *
-     * @param string $webspace
-     * @param Request $request
-     *
      * @return Response
      */
     public function postAction(string $webspace, Request $request)
--- a/src/Sulu/Bundle/CustomUrlBundle/Controller/CustomUrlController.php
+++ b/src/Sulu/Bundle/CustomUrlBundle/Controller/CustomUrlController.php
@@ -115,6 +115,7 @@ class CustomUrlController extends AbstractRestController implements SecuredContr
      */
     public function postAction($webspace, Request $request)
     {
+        // throw helpful error message if targetLocale is not set
         $this->getRequestParameter($request, 'targetLocale', true);
 
         $document = $this->customUrlManager->create(
--- a/src/Sulu/Bundle/CustomUrlBundle/Controller/CustomUrlController.php
+++ b/src/Sulu/Bundle/CustomUrlBundle/Controller/CustomUrlController.php
@@ -115,6 +115,8 @@ class CustomUrlController extends AbstractRestController implements SecuredContr
      */
     public function postAction($webspace, Request $request)
     {
+        $this->getRequestParameter($request, 'targetLocale', true);
+
         $document = $this->customUrlManager->create(
             $webspace,
             $request->request->all()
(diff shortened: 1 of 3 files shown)


src/Sulu/Bundle/CustomUrlBundle/Controller/CustomUrlController.php
```diff
@@ -109,16 +109,13 @@ public function getAction($webspace, $id, Request $request)
     /**
      * Create a new custom-url object.
      *
-     * @param string $webspace
-     *
      * @return Response
      */
-    public function postAction($webspace, Request $request)
+    public function postAction(string $webspace, Request $request)
```
nnatter, line 114: we cannot change the method signature because it would be a bc break, so we need to revert this 😕 
```suggestion
    public function postAction($webspace, Request $request)
```

src/Sulu/Bundle/CustomUrlBundle/Controller/CustomUrlController.php
```diff
@@ -109,16 +109,13 @@ public function getAction($webspace, $id, Request $request)
     /**
      * Create a new custom-url object.
      *
-     * @param string $webspace
```
nnatter, line 112: would not remove this

src/Sulu/Bundle/CustomUrlBundle/Request/CustomUrlRequestProcessor.php
```diff
@@ -163,11 +163,9 @@ private function matchCustomUrl($url, PortalInformation $portalInformation, Requ
      * Server encodes the url and symfony does not encode it
      * Symfony decodes this data here https://github.com/symfony/symfony/blob/3.3/src/Symfony/Component/Routing/Matcher/UrlMatcher.php#L91.
      *
-     * @param $pathInfo
-     *
      * @return string
      */
-    private function decodeUrl($pathInfo)
+    private function decodeUrl(string $pathInfo)
```
nnatter, line 168: we cannot change the method signature here too

src/Sulu/Bundle/CustomUrlBundle/Request/CustomUrlRequestProcessor.php
```diff
@@ -163,11 +163,9 @@ private function matchCustomUrl($url, PortalInformation $portalInformation, Requ
      * Server encodes the url and symfony does not encode it
      * Symfony decodes this data here https://github.com/symfony/symfony/blob/3.3/src/Symfony/Component/Routing/Matcher/UrlMatcher.php#L91.
      *
-     * @param $pathInfo
```
nnatter, line 166: would not remove this but add a `string` phpdoc type

src/Sulu/Bundle/CustomUrlBundle/Controller/CustomUrlController.php
rogoit: The function what is called here only has two parameters

src/Sulu/Bundle/CustomUrlBundle/Controller/CustomUrlController.php
rogoit: In other function in this file the parameters are also descriped like this. And when we want to go up to higher PHPStan levels or to PHP 8 we also have to do this

src/Sulu/Bundle/CustomUrlBundle/Controller/CustomUrlController.php
nnatter: yes, changing the call below is fine 🙂  but we cannot add the `string` typehint for the `$webspace`

src/Sulu/Bundle/CustomUrlBundle/Request/CustomUrlRequestProcessor.php
rogoit: How should we go forward with this topic?

nnatter: remove the typehint and add it as phpdoc 🙂

src/Sulu/Bundle/CustomUrlBundle/Controller/CustomUrlController.php
nnatter: the problem is not that the change would be wrong, the problem is that we cannot break backwards compatibility of this class. if somebody has overwritten/extended the class, his project will break if we change the typehints of the methods. therefore we need to use phpdoc types instead of typehints

nnatter: > And when we want to go up to higher PHPStan levels or to PHP 8 we also have to do this

I really doubt this. Can you provide any source for this?

src/Sulu/Bundle/CustomUrlBundle/Controller/CustomUrlController.php
```diff
@@ -117,8 +117,7 @@ public function postAction($webspace, Request $request)
     {
         $document = $this->customUrlManager->create(
             $webspace,
-            $request->request->all(),
-            $this->getRequestParameter($request, 'targetLocale', true)
+            $request->request->all()
         );
```
alexander-schranz, line 121: I think this should for bc reasons something like:

```suggestion
        $this->getRequestParameter($request, 'targetLocale', true);

        $document = $this->customUrlManager->create(
            $webspace,
            $request->request->all()
       );
```

to validate and throw a 400 error if targetLocale is not sent to the api. If we remove the getRequestParameter we will not get any validation about the required parameter.

nnatter:
Thank you! 🙂

Issue #1456, opened 2015-08-05
Add type filter in smartcontent


alexander-schranz:
E.g Content Structure
- news (overview template)
  - 2014 (overview template)
    - july (overview template)
      - Page 1 (news template)
      - Page 2 (news template)
  - 2015 (overview template)
    - august (overview template)
      - Page 1 (news template)
      - Page 2 (news template)
    - september (overview template)
      - Page 1 (news template)
      - Page 2 (news template)

I want to have a smartcontent for a specific template. Because I have nested overviews and I don't want that the overview templates are shown in the smartcontent.

danrot:
I've swapped template with type in the title. Would prefer that naming, because then it is easier to be reused in other contexts as well.

alexander-schranz:
Would think about the following in the Smart Content DataProvider [Configuration](https://github.com/sulu/sulu/blob/2be30ce7561911b243884ae27e86907899fa2639/src/Sulu/Component/Content/SmartContent/PageDataProvider.php#L112-L133):

```xml
    ->enableTemplates('page'); // the pages template "type" key
```

And for articles:

```php
    ->enableTemplates('article'); // the articles template "type" key
```

danrot:
In order to not couple this to pages and articles, I would prefer to call this function `enableTypes`. Maybe we want to filter for only a specific type of contacts in the future.

And I think that is something that should be exposed to the content manager. Then a call like the following should be sufficient:

```php
->enableTypes()
```

alexander-schranz:
@danrot enableTypes sounds good for me as it are template types. At current state I would just implement it for templates as there should it be as simple as possible. For contacts or other types we would need some kind of providers or resources for that types but think in first case I would just go with template types.

danrot:
When you e.g. load the metadata for the `pages` resource key, you already get all available types. I would just reuse that information, should work for other entities in a similar fashion.

alexander-schranz:
@danrot thats what I would implement currently, but for other entities this types do not exist as they don't have different template types. Media just have media types which would make sense in the smart content filter but they have nothing todo with form metadata types that is what I meant with it that there you would need something else.

danrot:
Ok, thought of this types as something different :see_no_evil: We could make the parameter in `enableTypes` optional, and if nothing is passed it tries to get that information from the metadata. Optionally the available types can be passed, for cases like media.

danrot:
Fixed in #5649

danrot:
[...]

PR #5649, opened 2020-11-23
Add type filtering in smart content

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | yes
| Deprecations? | no
| Fixed tickets | fixes https://github.com/sulu/sulu/issues/1456
| License | MIT

#### What's in this PR?

Adds type (template) filtering to smart contents.

#### Why?

See https://github.com/sulu/sulu/issues/1456

#### Example Usage

~~~php
    protected function configure(BuilderInterface $builder): void
    {
        $builder
            ->enableTags()
            ->...()
            ->enableTypes(
                [
                    ['type' => 'default', 'title' => 'my-translation-key'],
                    ...
                ]
            );
    }
~~~

#### BC Breaks/Deprecations

`BuilderInterface` has been extended with the `enableTypes` method.
`ProviderConfigurationInterface` has been extended with the `getTypes` and `hasTypes` methods.

#### To Do


src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/tests/fields/SmartContent.test.js
```diff
@@ -82,6 +82,7 @@ test('Should correctly initialize SmartContentStore', () => {
         sortMethod: undefined,
         tagOperator: undefined,
         tags: undefined,
+        types: undefined,
```
danrot, line 85: Shouldn't there at least be one test that contains some values?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/SmartContent/FilterOverlay.js
```diff
@@ -316,6 +327,26 @@ class FilterOverlay extends React.Component<Props> {
...
+                                <h3>{translate('sulu_admin.filter_by_types')}</h3>
+                                <div className={filterOverlayStyles.types}>
+                                    <MultiSelect
+                                        allSelectedText={translate('sulu_admin.all_types')}
+                                        noneSelectedText={translate('sulu_admin.no_types')}
+                                        onChange={this.handleTypesChange}
+                                        values={this.types || []}
+                                    >
+                                        {types.map((type) => (
+                                            <MultiSelect.Option key={type['value']} value={type['value']}>
```
danrot, line 341: Any specific reason we are using `type['value']` instead of `type.value` here?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/SmartContent/FilterOverlay.js
```diff
@@ -316,6 +327,26 @@ class FilterOverlay extends React.Component<Props> {
...
+                                                {type['name']}
```
danrot, line 342: Same as above, why not using `type.name` here?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/SmartContent/stores/smartContentConfigStore.js
```diff
@@ -30,6 +30,9 @@ class SmartContentConfigStore {
             sortBy: config.sorting.length > 0 ? config.sorting[0].name : undefined,
             sortMethod: config.sorting.length > 0 ? 'asc' : undefined,
             tagOperator: config.tags ? 'or' : undefined,
+            types: config.types && config.types.length > 0 ? config.types.map((type)=>{
```
danrot, line 33: Why do we need to check for the existince of `config.types`, but not for `config.sorting`? And the `map` could be written easier:

```javascript
config.types.map((type) => type.value)
```

And if we need to break the line, I would do that at the `?` and `:` of the ternary operator, IMO it reads easier that way:

```javascript
config.types.length > 0
    ? config.types.map((type) => type.value)
    : undeifned
```

src/Sulu/Component/Content/SmartContent/PageDataProvider.php
```diff
@@ -400,6 +419,28 @@ function(
         );
     }
 
+    private function getTypes(): array
+    {
+        $types = [];
+        if ($this->tokenStorage && null != $this->tokenStorage->getToken()) {
```
danrot, line 425: `null !==`

src/Sulu/Component/Content/SmartContent/PageDataProvider.php
```diff
@@ -400,6 +419,28 @@ function(
...
+            $user = $this->tokenStorage->getToken()->getUser();
+
+            // user may be "anon."
+            if (!$user instanceof UserInterface) {
```
danrot, line 429: Is that a real possibility? Because if this should not happen, I would throw an exception in this `if`. Would make it easier to spot errors later on.

src/Sulu/Component/SmartContent/Configuration/BuilderInterface.php
```diff
@@ -27,6 +27,13 @@ public static function create();
      */
     public function enableTags(bool $enable = true);
 
+    /**
+     * @param mixed[] $types
+     *
+     * @return BuilderInterface
+     */
+    public function enableTypes(array $types = []);
```
danrot, line 35: Strictly speaking that would be a BC break... Not sure if it matters too much here, I don't think so, because the `Builder` itself is just used by other `SmartContentDataProviders` and this interface is probably not implemented outside of Sulu.

src/Sulu/Component/SmartContent/Configuration/ProviderConfigurationInterface.php
```diff
@@ -52,6 +52,18 @@ public function setAudienceTargeting(bool $audienceTargeting);
...
+     * Returns types for types filtering.
+     *
+     * @return null|PropertyParameter[]
+     */
+    public function getTypes(): ?array;
+
+    /**
+     * Returns TRUE if types should be displayed.
+     */
+    public function hasTypes(): bool;
```
danrot, line 65: Same as above, strictly speaking that would be a BC break, but it is not necessary to implement this interface to add your own `DataProvider`, so I think it should be fine.

danrot (changes requested):
And didn't we say that this should also be implemented for media and snippets? Shouldn't be so much to add anymore, right? Then we would also immediately see if it works in these cases as well.

Prokyonn:
> And didn't we say that this should also be implemented for media and snippets? Shouldn't be so much to add anymore, right? Then we would also immediately see if it works in these cases as well.

@danrot yeah we did, I would have made the implementation for snippets and media in another PR, but I can also push it to this branch if you want

danrot:
I guess it is not a lot of code, or is it? If it just a few more lines I would have done it in here directly, to make sure everything is working, and we don't need more adjustements afterwards.

src/Sulu/Component/Content/SmartContent/PageDataProvider.php
Prokyonn: The `PageDataProvider` configuration is initialized on the first `/admin` request. If the user is not logged in, then the `user` is "anon."

Prokyonn:
> I guess it is not a lot of code, or is it? If it just a few more lines I would have done it in here directly, to make sure everything is working, and we don't need more adjustements afterwards.

No, it shouldn't be much more. I will add it to this pr :)

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/tests/fields/SmartContent.test.js
```diff
@@ -82,7 +82,7 @@ test('Should correctly initialize SmartContentStore', () => {
         sortMethod: undefined,
         tagOperator: undefined,
         tags: undefined,
-        types: undefined,
+        types: ['default', 'homepage'],
```
danrot, line 85: When I said that, I thought that should also be asserted afterwards somehow :see_no_evil: Don't we at least check if the props of the child components are passed correctly in this test?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/SmartContent/stores/smartContentConfigStore.js
```diff
@@ -30,9 +30,7 @@ class SmartContentConfigStore {
             sortBy: config.sorting.length > 0 ? config.sorting[0].name : undefined,
             sortMethod: config.sorting.length > 0 ? 'asc' : undefined,
             tagOperator: config.tags ? 'or' : undefined,
-            types: config.types && config.types.length > 0 ? config.types.map((type)=>{
-                return type['value'];
-            }) : undefined,
+            types: config.types.length > 0 ? config.types.map((type)=> type['value']) : undefined,
```
danrot, line 33: Why still not `type.value` instead of `type['value']`?

src/Sulu/Component/Content/SmartContent/PageDataProvider.php
danrot: Are you saying that the `initConfiguration` function is also called if the user is not logged in? I thought the `/config` request is only sent if the user has been logged in :thinking:

danrot (changes requested):

src/Sulu/Component/Content/SmartContent/PageDataProvider.php
Prokyonn: Yes, when the `AudienceTargeting` is not active, the configuration for every `dataProvider` is initialized.
See: https://github.com/sulu/sulu/blob/master/src/Sulu/Component/SmartContent/DataProviderPool.php#L43-L45

src/Sulu/Bundle/AdminBundle/Resources/js/containers/SmartContent/stores/smartContentConfigStore.js
Prokyonn: sorry my bad, I'm more used to other notation

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/tests/fields/SmartContent.test.js
Prokyonn: No the only check we do in this test is 
`expect(SmartContentStore).toBeCalledWith('media', value, undefined, 'collections', undefined, schemaOptions, 'sulu_io');`
but we don't explicitly check the props

src/Sulu/Bundle/SnippetBundle/Content/SnippetDataProvider.php
```diff
@@ -220,6 +240,37 @@ function(
...
+            // the user is required for the metadata locale because of the translated template titles
+            $user = $this->tokenStorage->getToken()->getUser();
+
+            // user is "anon." if the user is not logged in
+            // this happens on the initial admin request (/admin) if the AudienceTargeting is deactivated
+            if (!$user instanceof UserInterface) {
+                return $types;
+            }
+
+            try {
```
Prokyonn, line 259: The `try/catch` is necessary because in some tests like https://github.com/sulu/sulu/blob/master/src/Sulu/Bundle/AudienceTargetingBundle/Tests/Functional/Controller/AdminControllerTest.php#L32 there is no `SnippetDataProvider` registered, but I don't really have an idea why the providers are not registered.
Do you have any idea? @danrot @wachterjohannes

src/Sulu/Bundle/SnippetBundle/Content/SnippetDataProvider.php
```diff
@@ -67,20 +72,34 @@ class SnippetDataProvider implements DataProviderInterface
...
+        FormMetadataProvider $formMetadataProvider = null,
+        TokenStorageInterface $tokenStorage = null
     ) {
         $this->contentQueryExecutor = $contentQueryExecutor;
         $this->snippetQueryBuilder = $snippetQueryBuilder;
         $this->nodeHelper = $nodeHelper;
         $this->proxyFactory = $proxyFactory;
         $this->documentManager = $documentManager;
         $this->referenceStore = $referenceStore;
+        $this->formMetadataProvider = $formMetadataProvider;
```
danrot, line 101: The `tokenStorage` probably is really nullable, but the `formMetadataProvider` should trigger a deprecation notice here if not set, shouldn't it?

src/Sulu/Bundle/SnippetBundle/Content/SnippetDataProvider.php
```diff
@@ -220,6 +240,37 @@ function(
...
+
+            // user is "anon." if the user is not logged in
+            // this happens on the initial admin request (/admin) if the AudienceTargeting is deactivated
+            if (!$user instanceof UserInterface) {
+                return $types;
+            }
+
+            try {
+                /** @var TypedFormMetadata $metadata */
+                $metadata = $this->formMetadataProvider->getMetadata('snippet', $user->getLocale(), []);
```
danrot, line 261: We have to handle the case in which the `formMetadataProvider` is `null`, right?

src/Sulu/Bundle/SnippetBundle/Tests/Unit/Content/SnippetDataProviderTest.php
```diff
@@ -84,6 +91,79 @@ public function setUp(): void
...
+        $formMetadata2->getName()
+            ->shouldBeCalled()
+            ->willReturn('template-2');
+        $formMetadata2->getTitle()
+            ->shouldBeCalled()
+            ->willReturn('translated-template-2');
+
+        $typedFormMetadata->getForms()
+            ->shouldBeCalled()
+            ->willReturn([$formMetadata1, $formMetadata2]);
```
danrot, line 143: I think it would be easier to simply use these classes, wouldn't it? They are quite simple value objects without a lot of logic. Would also guarantee that it works properly with the real implementation.

src/Sulu/Component/Content/SmartContent/PageDataProvider.php
```diff
@@ -89,14 +103,18 @@ public function __construct(
...
+        FormMetadataProvider $formMetadataProvider = null,
+        TokenStorageInterface $tokenStorage = null
     ) {
         $this->contentQueryBuilder = $contentQueryBuilder;
         $this->contentQueryExecutor = $contentQueryExecutor;
         $this->documentManager = $documentManager;
         $this->proxyFactory = $proxyFactory;
         $this->session = $session;
         $this->referenceStore = $referenceStore;
+        $this->formMetadataProvider = $formMetadataProvider;
```
danrot, line 116: Same as above, it should trigger a deprecation if `$formMetadataProvider` is set to null.

src/Sulu/Component/Content/SmartContent/PageDataProvider.php
```diff
@@ -400,6 +419,33 @@ function(
...
+            $user = $this->tokenStorage->getToken()->getUser();
+
+            // user is "anon." if the user is not logged in
+            // this happens on the initial admin request (/admin) if the AudienceTargeting is deactivated
+            if (!$user instanceof UserInterface) {
+                return $types;
+            }
+
+            /** @var TypedFormMetadata $metadata */
+            $metadata = $this->formMetadataProvider->getMetadata('page', $user->getLocale(), []);
```
danrot, line 439: We have to handle the case that the `formMetadataProvider` is null here.

src/Sulu/Component/Content/Tests/Unit/SmartContent/PageDataProviderTest.php
```diff
@@ -132,6 +139,81 @@ public function testGetConfiguration()
...
+        $formMetadata2->getName()
+            ->shouldBeCalled()
+            ->willReturn('template-2');
+        $formMetadata2->getTitle()
+            ->shouldBeCalled()
+            ->willReturn('translated-template-2');
+
+        $typedFormMetadata->getForms()
+            ->shouldBeCalled()
+            ->willReturn([$formMetadata1, $formMetadata2]);
```
danrot, line 191: Same as above, would use the real classes here.

src/Sulu/Component/Media/Tests/Unit/SmartContent/MediaDataProviderTest.php
```diff
@@ -106,6 +111,73 @@ public function testGetConfiguration()
...
+            ->willReturn('image');
+
+        /** @var MediaType|ObjectProphecy $mediaType2 */
+        $mediaType2 = $this->prophesize(MediaType::class);
+        $mediaType2->getId()
+            ->shouldBeCalled()
+            ->willReturn(2);
+        $mediaType2->getName()
+            ->shouldBeCalled()
+            ->willReturn('audio');
```
danrot, line 142: If easily possible would also use the real classes here.

src/Sulu/Bundle/SnippetBundle/Content/SnippetDataProvider.php
wachterjohannes: @Prokyonn is the snippet bundle registered in the test app?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/tests/fields/SmartContent.test.js
Prokyonn: I'm not sure but I don't see any other possibility to test this. The `SmartContent` field passes the `types` via the `value` object to the `SmartContentStore` and this is already checked with the line from my comment above. The field-type doesn't have any other reference to the `types`.

danrot (changes requested):

src/Sulu/Component/Media/SmartContent/MediaDataProvider.php
```diff
@@ -162,4 +172,24 @@ protected function getSerializationContext()
...
+    {
+        $types = [];
+
+        if (!$this->entityManager) {
+            return $types;
+        }
+
+        $repository = $this->entityManager->getRepository(MediaType::class);
+        /** @var MediaType $mediaType */
+        foreach ($repository->findAll() as $mediaType) {
```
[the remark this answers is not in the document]
Prokyonn: I talked with @alexander-schranz yesterday and there seems to be no way except for hard coding it. Another benefit of the database query is, that if developers extended the media types, they can automatically filter with those types as well.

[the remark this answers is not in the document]
danrot: Okay, then it is fine.

src/Sulu/Component/Content/SmartContent/QueryBuilder.php
```diff
@@ -359,7 +359,12 @@ private function buildTypesWhere($types, string $languageCode)
     {
         $sql2Where = [];
         foreach ($types as $type) {
-            $sql2Where[] = 'page.[i18n:' . $languageCode . '-template] = ' . $type;
+            if ('snippet' === static::$structureType) {
```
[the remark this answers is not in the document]
Prokyonn: Yes you are right, but this would require bigger changes and most likely BC breaks. I already discussed it with @wachterjohannes and we didn't had a better idea.

src/Sulu/Bundle/SnippetBundle/Content/SnippetDataProvider.php
Prokyonn: Yes, the `Kernel` extends the `SuluTestKernel` and this kernel registers the `SnippetBundle`

src/Sulu/Component/Media/SmartContent/MediaDataProvider.php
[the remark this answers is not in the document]
alexander-schranz: The problem here is that the IDs are hardcoded here and sadly constants where created for them:

https://github.com/sulu/sulu/blob/11fdd6d8de723374529bc0b3e4bfeecd1f6c28aa/src/Sulu/Bundle/MediaBundle/DataFixtures/ORM/LoadMediaTypes.php#L28-L46

But theoretical you could still add additional types by defining the mimetypes for the new type in the sulu_media extension and create the type in your database.

Another way would be not depend on the `id` of the types or going with the `type.name` and go on the `sulu_media.media.types` parameter:

> bin/console debug:container --parameter sulu_media.media.types --format json

This I think would need in the media case an additional join but not sure I'm always a but confused by the smart content providers 🙈. But I thought going with the database way would be the easiest way and think types should also be possible to be provided from the database so having this usage in the sulu core we are sure it also works loading types from the db. Maybe we can use a custom query and use the query result cache but not sure, just wanted to share my thoughts here with you.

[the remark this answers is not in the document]
danrot: Yeah, the smart content providers are confusing me as well... But at the current state there already is a join to the types introduced :see_no_evil: Does that change your mind somehow?

src/Sulu/Bundle/SnippetBundle/Content/SnippetDataProvider.php
danrot: Do you have a link to the failing test?

src/Sulu/Component/Content/SmartContent/QueryBuilder.php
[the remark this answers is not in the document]
danrot: Why can't we just override the `buildTypesWhere` function in the `SnippetQueryBuilder`?

src/Sulu/Bundle/SnippetBundle/Tests/Unit/Content/SnippetDataProviderTest.php
```diff
@@ -84,6 +91,69 @@ public function setUp(): void
...
+        $formMetadata1->setName('template-1');
+        $formMetadata1->setTitle('translated-template-1');
+
+        $formMetadata2 = new FormMetadata();
+        $formMetadata2->setName('template-2');
+        $formMetadata2->setTitle('translated-template-2');
+
+        $typedFormMetadata->getForms()
+            ->shouldBeCalled()
+            ->willReturn([$formMetadata1, $formMetadata2]);
```
danrot, line 133: Actually I meant this object as well :grimacing: Could be easily replaced with:

```php
$typedFormMetadata = new TypedFormMetadata();
$typedFormMetadata->add('template-1', $formMetadata1);
$typedFormMetadata->add('template-2', $formMetadata2);
```

src/Sulu/Component/Content/Tests/Unit/SmartContent/PageDataProviderTest.php
```diff
@@ -132,6 +139,71 @@ public function testGetConfiguration()
...
+        $formMetadata1->setName('template-1');
+        $formMetadata1->setTitle('translated-template-1');
+
+        $formMetadata2 = new FormMetadata();
+        $formMetadata2->setName('template-2');
+        $formMetadata2->setTitle('translated-template-2');
+
+        $typedFormMetadata->getForms()
+            ->shouldBeCalled()
+            ->willReturn([$formMetadata1, $formMetadata2]);
```
danrot, line 181: Same as above

src/Sulu/Component/Media/Tests/Unit/SmartContent/MediaDataProviderTest.php
```diff
@@ -106,6 +111,63 @@ public function testGetConfiguration()
         $this->assertTrue($configuration->hasAudienceTargeting());
     }
 
+    public function testGetTypesConfiguration()
+    {
+        /** @var EntityManagerInterface|ObjectProphecy $entityManager */
+        $entityManager = $this->prophesize(EntityManagerInterface::class);
+
+        /** @var ObjectRepository|ObjectProphecy $objectRepository */
+        $objectRepository = $this->prophesize(ObjectRepository::class);
```
danrot, line 120: Wouldn't it make more sense to call this `$mediaTypeRepository`? `$objectRepository` could be anything.

danrot (changes requested):

src/Sulu/Bundle/SnippetBundle/Content/SnippetDataProvider.php
Prokyonn: Not in the CI, but the `testRulesConfig` of the audiencetargeting bundle fails https://github.com/sulu/sulu/blob/master/src/Sulu/Bundle/AudienceTargetingBundle/Tests/Functional/Controller/AdminControllerTest.php#L32

src/Sulu/Component/Content/SmartContent/QueryBuilder.php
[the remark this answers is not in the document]
Prokyonn: Because the `buildTypes..` methods are private, but we could make them protected and override it.

src/Sulu/Bundle/SnippetBundle/Content/SnippetDataProvider.php
Prokyonn: @alexander-schranz found a solution. The error happens because in those bundles no snippet templates exist. Now the question is if we should add a default snippet template in each test bundle or if it is a valid case, if there is really no snippet template available. WDYT @danrot

Prokyonn:
Rebased onto `release/2.1`

src/Sulu/Bundle/SnippetBundle/Content/SnippetDataProvider.php
danrot: I would say the system should not break just because there are no snippets defined... The thing is that the [`StructureFormMetadataLoader`](https://github.com/sulu/sulu/blob/master/src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/StructureFormMetadataLoader.php#L96) has to return `null` here, because the mechanism relies on `MetadataLoader`s to not return anything, in case there is nothing... And the `StructureMetadataFactory` does not seem to return `snippet` as a structure type if no snippets are given...

Maybe (not sure, have not tested it) it would be enough to instead of [implicitly creating the array](https://github.com/sulu/sulu/blob/0da80a9562d5249dd629a45d0fe18281fdf85f10/src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/StructureFormMetadataLoader.php#L129) make that explicit, i.e. create the array like this after the foreach in line 123:

```php
$structuresMetadataByTypes[$structureType] = [];
```

Then snippets should be returned as an option even if no snippets are available. Not sure if that will create other problems though :see_no_evil: But I guess that would be the proper fix to this problem.

UPGRADE.md
```diff
@@ -1,5 +1,34 @@
...
+Furthermore the `PageDataProvider`, the `SnippetDataProvider` and the `MediaDataProvider` have been updated
+with additional services. The injected services are optionally but usage of these providers without them will throw
+a deprecation warning.
+
+`PageDataProvider` and `SnippetDataProvider` now additionally requires the `FormMetadataProvider` and the `TokenStorageInterface` services.
+The `MediaDataprovider` requires the `EntityManagerInterface` and the `TranslatorInterface`.
+
+For the type filtering to work properly all those mentioned services are necessary.
+
+
```
danrot, line 31: Only one new line.

danrot (changes requested):

danrot (approved):

UPGRADE.md
```diff
@@ -1,5 +1,33 @@
 # Upgrade
 
+## 2.1.6
+
+### Smartcontent Type Filtering
```
nnatter, line 5: think this is nicely written! 🙂

PR #5650, opened 2020-11-23
Run tests with github actions in sulu 2

luca-rath:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Run tests with github actions for sulu 2.

#### Why?

Because we want to migrate from travis to github actions


src/Sulu/Bundle/TestBundle/Testing/PurgeDatabaseTrait.php
```diff
@@ -34,6 +34,10 @@ protected static function purgeDatabase(): void
             $connection->executeUpdate('SET foreign_key_checks = 0;');
         }
 
+        if ($connection->getDriver() instanceof \Doctrine\DBAL\Driver\PDOPgSql\Driver) {
+            $connection->executeUpdate('SET session_replication_role = "replica";');
```
luca-rath, line 38: @danrot This solution is working, what do you think of that?

.github/workflows/test-application.yaml
```diff
@@ -51,3 +52,138 @@ jobs:
               run: curl "https://cs.symfony.com/download/php-cs-fixer-v2.phar" --output ${{ steps.php-cs-fixer-path.outputs.path }} && chmod +x ${{ steps.php-cs-fixer-path.outputs.path }}
             - name: Run php-cs-fixer
               run: ./${{ steps.php-cs-fixer-path.outputs.path }} fix --dry-run --diff
+
+    test:
+        name: "Run tests with php ${{ matrix.php-version }} (${{ matrix.phpcr-transport }})"
```
alexander-schranz, line 57: would be nice to also add the database name e.g. mysql / postgresql

luca-rath: should be done in `release/1.6` then ... I can create another pr there

src/Sulu/Bundle/TestBundle/Testing/PurgeDatabaseTrait.php
danrot: What exactly does it do, resp. where have you found that solution? But as long as it is working I am fine with almost everything :see_no_evil:

.github/workflows/test-application.yaml
danrot: I guess so... Create a PR there, then we can also see how back merging works :see_no_evil:

danrot:
@luca-rath Is it now almost the same as #5645? Or anything that we should have a special look at?

src/Sulu/Bundle/TestBundle/Testing/PurgeDatabaseTrait.php
luca-rath: https://postgresqlco.nf/en/doc/param/session_replication_role/ says:

> The intended use of this setting is that logical replication systems set it to replica when they are applying replicated changes. The effect of that will be that triggers and rules (that have not been altered from their default configuration) will not fire on the replica. See the sql-altertable clauses ENABLE TRIGGER and ENABLE RULE for more information.

> Since foreign keys are implemented as triggers, setting this parameter to replica also disables all foreign key checks, which can leave data in an inconsistent state if improperly used.

I found it here https://stackoverflow.com/a/49584660

luca-rath:
> @luca-rath Is it now almost the same as #5645? Or anything that we should have a special look at?

The biggest differences are, that jackrabbit is a service now and `phpstan` and `lint-container` scripts have been added.

src/Sulu/Bundle/TestBundle/Testing/PurgeDatabaseTrait.php
danrot: Sounds good to me :+1:

danrot (approved):

PR #5652, opened 2020-11-24
Add flag to disable moving in BlockCollection

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | #5608
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR adds a flag to the `BlockCollection` component, that allows to disable the moving feature in it.

#### Why?

Because in #5608 we need the blocks, but it should not be possible to move them.


wachterjohannes (approved):

alexander-schranz:
I think in data management this could be interested and we should make it possible to set a [parameter](https://docs.sulu.io/en/2.2/reference/content-types/block.html#parameters) for this in the xml, so it can also be disable from the form.xmls.

PR #5654, opened 2020-11-24
Allow components as children of Select Option

binarious:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | 
| Related issues/PRs | 
| License | MIT
| Documentation PR | 

#### What's in this PR?

This PR allows to use html elements or React components with the `DisplayValue` component.

#### Why?

Currently `DisplayValue` only supports strings. When used in combination with other components like `Icon` the resulting display value would be `[object Object]`.

#### Example Usage

~~~js
<SingleSelect disabled={disabled} value={value} onChange={this.handleInputChange} id={dataPath} valid={!error}>
  <Option value="fa-info-circle"><Icon name="fa-info-circle"/></Option>
  <Option value="fa-user"><Icon name="fa-user"/></Option>
</SingleSelect>
~~~

#### To Do

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/DisplayValue.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/DisplayValue.js
@@ -60,8 +60,8 @@ export default class DisplayValue extends React.PureComponent<Props> {
                 {!!icon &&
                     <Icon className={displayValueStyles.frontIcon} name={icon} />
                 }
-                {typeof children === 'string'
-                    ? <CroppedText>{children}</CroppedText>
+                {typeof children === 'string' || typeof children === 'number'
+                    ? <CroppedText>{String(children)}</CroppedText>
                     : children
                 }
                 <Icon className={displayValueStyles.toggle} name="su-angle-down" />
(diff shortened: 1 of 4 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/DisplayValue.js
```diff
@@ -8,7 +8,7 @@ import displayValueStyles from './displayValue.scss';
 import type {Skin} from './types';
 
 type Props = {|
-    children: string,
+    children: any,
```
binarious, line 11: I don't know the correct type for any kind of html element here. `React.Element` didn't work, because it needs a generic type.

nnatter: i think the correct type is `Node`, at least that is what was used in other places 🙂 
https://flow.org/en/docs/react/types/#toc-react-node

nnatter:
Thanks a lot for working on this. I rebased the change onto `release/2.1` 🙂

src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/DisplayValue.js
binarious: Ah I see. It look like you've already fixed it in https://github.com/sulu/sulu/pull/5654/commits/447f9ed2a087949abfe575d7106cfdc4f3665ecb :+1:. So this seems to be resolved.

nnatter (approved):

PR #5655, opened 2020-11-25
Add flag to disable collapsing of blocks

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | --
| Related issues/PRs | #5608 
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR adds a flag to make blocks not collapsable.

#### Why?

We need that for #5608, because the schedule blocks should not be collapsable.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Block/Block.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Block/Block.js
@@ -12,8 +12,8 @@ type Props = {
     dragHandle?: Node,
     expanded: boolean,
     icons?: Array<string>,
-    onCollapse: () => void,
-    onExpand: () => void,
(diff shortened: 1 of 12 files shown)


wachterjohannes (approved):

alexander-schranz:
Same as for #5652 think this should be also be configureable over the xml to reuse it for custom entities.

danrot:
Yeah, might make sense, but right now I am concentrating on what I need to be done for #5608. Also, I am not totally sure if the `block` XML element can have a `params` tag as child.

alexander-schranz:
@danrot the block has exist params: https://docs.sulu.io/en/2.2/reference/content-types/block.html#parameters so its possible :)

PR #5656, opened 2020-11-25
Remove AudienceTargeting reference in DataProviderPool

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Removes the reference to the `AudienceTargetingBundle` in the `DataProviderPool`. Now every provider, which requires the audience targeting bundle has to enable it by itself only when the `AudienceTargetingBundle` is really enabled.

### DIFF
--- a/src/Sulu/Bundle/SnippetBundle/Tests/Unit/Content/SnippetDataProviderTest.php
+++ b/src/Sulu/Bundle/SnippetBundle/Tests/Unit/Content/SnippetDataProviderTest.php
@@ -136,6 +136,7 @@ class SnippetDataProviderTest extends TestCase
             $this->proxyFactory->reveal(),
             $this->documentManager->reveal(),
             $this->referenceStore->reveal(),
+            false,
             $formMetadataProvider->reveal(),
             $tokenStorage->reveal()
         );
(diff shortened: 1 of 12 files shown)


src/Sulu/Bundle/SnippetBundle/Tests/Unit/Content/SnippetDataProviderTest.php
```diff
@@ -136,6 +136,7 @@ public function testGetTypesConfiguration()
             $this->proxyFactory->reveal(),
             $this->documentManager->reveal(),
             $this->referenceStore->reveal(),
+            false,
```
danrot, line 139: Shouldn't we also add tests with that parameter set to `true`?

src/Sulu/Component/Content/Tests/Unit/SmartContent/PageDataProviderTest.php
```diff
@@ -182,6 +182,7 @@ public function testGetTypesConfiguration()
             $this->getSession(),
             new ReferenceStore(),
             ['view' => 64],
+            false,
```
danrot, line 185: Same as above

danrot (changes requested):

danrot (approved):

danrot:
LGTM and works on my local machine, except for the `cache:warmup --env=test` command, which is still not working for me, if the DB does not exist... But I get no error anymore with just `cache:warmup`, I guess that should be good enough for testing other bundles.

danrot (approved):

PR #5663, opened 2020-11-26
add better php doc for phpstan lvl2 HttpCacheBundle

Julian-B90:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Improvements for PHPStan level 2 on

#### Why?

Better type hint in our ide´s

### DIFF
--- a/src/Sulu/Bundle/HttpCacheBundle/EventSubscriber/InvalidationSubscriber.php
+++ b/src/Sulu/Bundle/HttpCacheBundle/EventSubscriber/InvalidationSubscriber.php
@@ -180,7 +180,7 @@ class InvalidationSubscriber implements EventSubscriberInterface
     /**
      * Invalidates the structure of the given document.
      *
-     * @param $document
+     * @param object $document
      */
     private function invalidateDocumentStructure($document)
     {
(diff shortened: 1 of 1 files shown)


src/Sulu/Bundle/HttpCacheBundle/EventSubscriber/InvalidationSubscriber.php
```diff
@@ -180,7 +180,7 @@ public function invalidateDocumentBeforeRemoving(RemoveEvent $event)
     /**
      * Invalidates the structure of the given document.
      *
-     * @param $document
+     * @param object $document
```
nnatter, line 183: as far as i see, all calls to this method look like this:
```
        if ($document instanceof StructureBehavior) {
            $this->invalidateDocumentStructure($document);
        }
```

so i think we should use `StructureBehavior`
```suggestion
     * @param StructureBehavior $document
```

alexander-schranz: @nnatter a document can be any object so `object` is correct here.

nnatter (approved):
Thanks for working on this! The test cases are failing because of the Symfony 5.2 release today.

Issue #5084, opened 2020-02-27
Preview not refreshed completely on language change


alexander-schranz:
| Q | A
| --- | ---
| Sulu Version | 2.0.4
| Browser Version | -

# Actual Behavior

Preview not refreshed on language change

# Expected Behavior

When switching the language of a page or other content with preview, the preview should reload the whole site in the new selected language.

# Steps to Reproduce

1. Open https://sulu.rocks/admin/#/webspaces/demo/pages/en/5eab85ce-7a3b-4e71-a3ad-203bd5002144/details
2. Switch language
3. Only things inside content block are refreshed not the menu outside or header

niklasnatter:
I can reproduce the problem by adding a translation to my template. The translation will not be updated in the preview when switching the locale:

```twig
{{ 'sulu_website.cache_clear'|trans([], 'admin') }}
```

It looks like the locale that is sent to the server by the `PreviewStore` is not updated when switching the locale in the administration interface.

alexander-schranz:
This issue was fixed by #5924 at current state we can only fix this in the new 2.3 release.

PR #5665, opened 2020-11-27
Refactor preview to avoid changing interface for every addition

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | #5608
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR refactors some interfaces as described in https://github.com/sulu/sulu/pull/5608#issuecomment-734724110.

#### Why?

Because everytime we add some new option to the preview toolbar (this time it will be a `DatePicker` for previewing the block scheduling) we had to touch the `PreviewInterface` and some other classes.

#### BC Breaks/Deprecations

Some interfaces in the PreviewBundle will be changed, which is a BC break.

#### To Do

### DIFF
--- a/src/Sulu/Bundle/PreviewBundle/Controller/PreviewController.php
+++ b/src/Sulu/Bundle/PreviewBundle/Controller/PreviewController.php
@@ -56,7 +56,7 @@ class PreviewController
 
         return new JsonResponse(
             [
-                'token' => $this->preview->start($provider, $id, $locale, $this->getUserId()),
+                'token' => $this->preview->start($provider, $id, $this->getUserId()),
             ]
         );
     }
@@ -71,16 +71,18 @@ class PreviewController
         $targetGroup = $this->getRequestParameter($request, 'targetGroup', false, null);
         $segment = $this->getRequestParameter($request, 'segment', false, null);
 
+        $options = [
+            'targetGroupId' => $targetGroup,
+            'segmentKey' => $segment,
+            'webspaceKey' => $webspace,
+            'locale' => $locale,
+        ];
+
         if (!$this->preview->exists($token)) {
-            $token = $this->preview->start($provider, $id, $locale, $this->getUserId());
+            $token = $this->preview->start($provider, $id, $this->getUserId(), $options);
         }
 
-        $content = $this->preview->render(
-            $token,
-            $webspace,
-            $locale,
-            ['targetGroupId' => $targetGroup, 'segmentKey' => $segment]
-        );
+        $content = $this->preview->render($token, $options);
 
         $this->disableProfiler();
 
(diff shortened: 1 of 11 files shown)


danrot:
Just found another issue with the preview, that makes implementing it by sending the options as POST body harder:

There are still [two](https://github.com/sulu/sulu/blob/e13d40a362f27fa2fb4defbbff55404e9ccca707/src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/Preview.js#L249) [places](https://github.com/sulu/sulu/blob/e13d40a362f27fa2fb4defbbff55404e9ccca707/src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/Preview.js#L286) that refer to the `PreviewStore.renderRoute` attribute. Also, the [`PreviewController::renderAction` needs the query parameters](https://github.com/sulu/sulu/blob/e13d40a362f27fa2fb4defbbff55404e9ccca707/src/Sulu/Bundle/PreviewBundle/Controller/PreviewController.php#L69-L72). Now I thought it would be nice to implement the `renderAction` in a way, that only passing the token would suffice. Then we would also have to worry less in places where this route is used, because only one parameter has to be set correctly. But that would probably mean we have to add more information to the `PreviewCacheItem`, so that all information is available to rerender that page. And I guess the `renderAction` could not start the preview itself this way, but is this really necessary?  I think the preview should always already be started if the route is used in the UI.

/cc @wachterjohannes WDYT?

wachterjohannes:
@danrot i am not sure if this is a good thing 🙈  we already stored that in the cache a while ago and have decided to remove as much as possible from the cache. would currently not change this and keep it as it is

danrot:
Don't see any explanation here :see_no_evil:

I would agree with avoiding the cache, if we want to get rid of it all together. But since the navigation twig function is so expensive and we want to avoid rendering that on every preview update, I don't see that happening anytime soon.

Resp. let me phrase the question differently: Could we replace that `PreviewCacheItem` all together? Because currently we take some information from it (`locale`, `id`) although this information is also passed as query parameter to the requests, and other information (`webspaceKey`, `targetGroupId` and `segmentKey`) is taken from the query parameters. It is hard to understand this way, and not obvious where to put information when building new features for the preview. And if we don't get this right this time, there will be another discussion when a new key is added, because then again some information is read from the cache and some from the query parameters, and the distinction seems to be arbitrary. The way I see it, it would be sufficient to just store the HTML in the cache with the generated token as a key, and the rest could always be read from the requests. Or am I missing something?

wachterjohannes:
I am not 100% sure but yes we could remove the locale / id from the cache (with the id i would have to take a deeper look). But yes you are right we should keep it konsistent all in or all out. i would say remove as much as you can 👍

danrot:
Is the [`Preview.php` file](https://github.com/sulu/sulu/blob/e13d40a362f27fa2fb4defbbff55404e9ccca707/src/Sulu/Bundle/PreviewBundle/Preview/Preview.php) the only one using the `PreviewCacheItem`? Would help with tracking that down, and this is the only class I can find.

nnatter:
I dont know if #5084 is related to this at all. But maybe we are lucky and this PR fixes that bug as a side effect. Leaving this comment as a reminder for checking this 🙂

danrot:
@nnatter That probably did not change, I've only changed the interfaces, but the same things should be happening as before.

PR #5666, opened 2020-11-27
Add json schema to selection and single_selection

### LINKED ISSUE #2938: "mandatory" & "minOccurs" attributes are ineffective in non-text-only properties

luca-rath:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #2938
| License | MIT

#### What's in this PR?

Add json schema to selection and single_selection to make `mandatory`, `minOccurs` and `maxOccurs` working


src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/ArrayMetadata.php
```diff
@@ -33,12 +38,14 @@ public function __construct(
         bool $mandatory,
         SchemaMetadata $schemaMetadata,
         int $minItems = null,
+        int $maxItems = null,
         bool $uniqueItems = null
```
luca-rath, line 42: Since the parameters `$minItems` and `$uniqueItems` are not released yet, I decided to add `$maxItems` in the middle of them instead of at the end

alexander-schranz:
I personally would not reuse the `minOccurs` and `maxOccurs` properties. I personally would remove them completely in the future and use params for this.

luca-rath:
Why removing them? I think for multi-value-properties they are just fine, because every multi-value-property should have a `minItems` and a `maxItems` check (and if we use `minOccurs` and `maxOccurs` it's consistent along all multi-value-properties, otherwise blocks are different again) ... But I would not overuse it, e.g. the min/max characters of a text_line property has nothing to do with `minOccurs` and `maxOccurs`

alexander-schranz:
Because not all properties have items or have possibility to have multiple items thats why I would remove them and I think currently missleading in many cases like text and so on that why I would not build up something new on them. I also commented at the linked issue https://github.com/sulu/sulu/issues/2938#issuecomment-629097147 I would go with a min and max attribute for the types which support min or max.

luca-rath:
Okay, that sounds good to me 👍

src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/SelectionPropertyMetadataMapper.php
```diff
@@ -0,0 +1,51 @@
...
+namespace Sulu\Bundle\AdminBundle\Metadata\SchemaMetadata;
+
+use Sulu\Component\Content\Metadata\PropertyMetadata as ContentPropertyMetadata;
+
+class SelectionPropertyMetadataMapper implements PropertyMetadataMapperInterface
+{
+    public function mapPropertyMetadata(ContentPropertyMetadata $propertyMetadata): PropertyMetadata
+    {
+        $mandatory = $propertyMetadata->isRequired();
+        $minItems = $propertyMetadata->getParameter('min', true)['value'] ?? null;
```
luca-rath, line 21: Currently, there is still a problem here ... if a property has `mandatory: false` and `min: 2`, I would expect to be able to either select `0` or `>= 2` items ... But unfortunately that's not possible with only json-schema, at least I don't know how. Probably we would have to change the `MultiSelection` to return `null` instead of an empty array, if no items are selected, but that's just a guess.

luca-rath: @danrot WDYT about that?

src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/SelectionPropertyMetadataMapper.php
```diff
@@ -0,0 +1,51 @@
...
+        $maxItems = $propertyMetadata->getParameter('max', true)['value'] ?? null;
```
nnatter, line 22: i think this is a nice feature! should we call it something like `maxItems` or `maxSelections`? we also need to document it once this is merged 😇

luca-rath: @alexander-schranz suggested `min` and `max` and I also think that's a good idea, because then we can use `min` and `max` also for other multi-value-properties like blocks ... if we would use `minSelections`, we could probably only use that param for selections, because it's name doesn't fit for blocks

src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/ObjectMetadata.php
```diff
@@ -18,23 +18,40 @@ class ObjectMetadata extends PropertyMetadata
...
+     * @var SchemaMetadata[]
+     */
+    private $allOfs;
+
     /**
      * @param PropertyMetadata[] $properties
+     * @param SchemaMetadata[] $anyOfs
+     * @param SchemaMetadata[] $allOfs
      */
-    public function __construct(string $name, bool $mandatory, array $properties = [])
```
danrot, line 24: Are these additions a leftover?

src/Sulu/Component/Content/Metadata/ItemMetadata.php
```diff
@@ -245,20 +245,33 @@ public function getTitle($locale)
...
+                $this->parameters,
+                function($parameter) use ($name) {
+                    if (!\is_array($parameter)) {
+                        return false;
+                    }
+
+                    return $parameter['name'] === $name;
+                }
+            )
+        );
```
danrot, line 265: Why is that necessary?

src/Sulu/Component/Content/Metadata/ItemMetadata.php
```diff
@@ -245,20 +245,33 @@ public function getTitle($locale)
     }
 
     /**
-     * Return the paramter with the given name.
+     * Return the parameter with the given name.
      *
      * @param string $name
      */
-    public function getParameter($name)
+    public function getParameter($name, bool $nullIfNotExists = false)
```
danrot, line 252: Not a big fan of these kind of flags... I have tried to see if there are any usages of this function, and there don't seem to be any... Can't we just return null if nothing exists without a flag then?

danrot (changes requested):

src/Sulu/Bundle/AdminBundle/DependencyInjection/Compiler/AddPropertyMetadataMapperPass.php
```diff
@@ -0,0 +1,53 @@
...
+        );
+    }
+
+    /**
+     * @param string[] $fieldTypes
+     */
+    private function registerPropertyMetadataMappers(ContainerBuilder $container, string $className, array $fieldTypes)
+    {
+        foreach ($fieldTypes as $fieldType) {
+            $definition = new Definition($className);
```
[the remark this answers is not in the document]
luca-rath: But then you can't just decorate the `sulu_admin.property_data_mapper.custom_entity_selection` service, if you want some specific schema for your custom entity selection.

src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/ObjectMetadata.php
luca-rath: No, they are used in the `SingleSelectionPropertyMetadataMapper`

src/Sulu/Bundle/AdminBundle/DependencyInjection/Compiler/AddPropertyMetadataMapperPass.php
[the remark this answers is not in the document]
danrot: You still can, just not using the decorate services mechanism from symfony. And I think it does not make it a lot easier in that case anyway.

With symfony decorator:
```xml
    <service id="Some\Mapper"
        decorates="Sulu\Mapper"
    >
        <argument type="service" id=".inner"/>
    </service>
```

Without:

```xml
    <service id="Some\Mapper">
        <argument type="service" id="Sulu\Mapper"/>
        <tag name="property_metadata_mapper" type="custom_entity_selection />
    </service>
```

Haven't used all the right names, because I was too lazy to look them up :see_no_evil: But I think you get my idea :slightly_smiling_face:

src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/SelectionPropertyMetadataMapper.php
nnatter: i think thats the advantage of using params here - we dont need to use the same param name for different things. 
for example, we could use descriptive names such as `minItemCount` for selections and `minBlockCount` for blocks. but if everybody else thinks we should try to reuse param names, thats okay for me too 🙂

src/Sulu/Component/Content/Metadata/ItemMetadata.php
luca-rath: `params` of properties look like this: `[['name' => 'foo', 'type' => 'string', 'value' => 'bar']]` ... But this method previously handled the `$this->parameters` array, as it would look like this: `['foo' => 'bar']` ... So I had to somehow combine that logic

src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/SelectionPropertyMetadataMapper.php
luca-rath: I don't agree on that (my personal opinion) ... /cc @sulu/core-team

src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/SelectionPropertyMetadataMapper.php
luca-rath: I think it would be easier for developers, if they don't have to remember lots of different param names for different properties, if their purpose is the same

src/Sulu/Bundle/AdminBundle/DependencyInjection/Compiler/AddPropertyMetadataMapperPass.php
```diff
@@ -0,0 +1,53 @@
...
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Definition;
+
+class AddPropertyMetadataMapperPass implements CompilerPassInterface
+{
+    public function process(ContainerBuilder $container)
+    {
+        $fieldTypeOptionRegistry = $container->get('sulu_admin.field_type_option_registry');
+        $baseFieldTypes = $fieldTypeOptionRegistry->toArray();
```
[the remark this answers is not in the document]
luca-rath: Or can we maybe add a priority to the compiler pass? Would that even be a better solution than doing this in the `SuluAdminExtension`?

src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/SelectionPropertyMetadataMapper.php
nnatter: not sure if there is a correct behaviour, think i would go with whats easier to implement 😅

src/Sulu/Bundle/AdminBundle/DependencyInjection/Compiler/AddPropertyMetadataMapperPass.php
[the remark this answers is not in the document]
luca-rath: In theory, yes. But I think most developers are used to use symfony decoration (like me ... until a few days ago I never thought about the possibility to manually use the decoration pattern) and will then be confused, why it's not working as they expect

src/Sulu/Component/Content/Metadata/ItemMetadata.php
luca-rath: Should I also refactor this whole method to ignore all keys and just use the `['name' => 'foo']` property?

src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/SelectionPropertyMetadataMapper.php
luca-rath: Let's compare it with the `minLength` attribute of the `text_line`: if `mandatory: false` and `minLength: 2`, null or empty string will pass validation, one character will fail and all above to characters will pass again ... So i guess this is the correct behaviour ... If the value of the selection would be null, validation would also pass, if the field is not required but has `minItems: 2`

danrot: That's not what I would expect, even if it is currently implemented like that in the `text_line`. As said, if I put a `min` attribute on it, I am doing that because I want there to be always at least to characters. And no characters is also less than 2, so the constraint I put on this element is not fulfilled, and therefore should not validate IMO.

src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/SelectionPropertyMetadataMapper.php
danrot: I also think that `min` and `max` are precise enough in the context of the content type, so I don't think I would give them any longer names.

src/Sulu/Bundle/AdminBundle/DependencyInjection/Compiler/AddPropertyMetadataMapperPass.php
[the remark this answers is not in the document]
danrot: It would solve the problem of not all options being added (unless somebody registers a compiler pass with an even lower priority).

But I think it is still not a good idea... It opens the door to errors like we recently had, which caused the `cache:clear` command to fail because there was no database configured. Calling `$container->get` is already calling the constructor, which might not be ready to be called. This might be fine for now, but who know what happens with the `FieldTypeOptionRegistry` in the future, maybe this will break at some point...

src/Sulu/Bundle/AdminBundle/DependencyInjection/Compiler/AddPropertyMetadataMapperPass.php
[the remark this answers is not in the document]
danrot: Can we at least try to find a different name than the `className` then? As said, this way we make this service autowirable, and I don't think that's a good idea, since I wouldn't consider that a public service.

That being said, I would personally still do it the other way, in order to keep a few unnecessary services from being created.

src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/SelectionPropertyMetadataMapper.php
```diff
@@ -0,0 +1,114 @@
...
+    /**
+     * Result object has following structure:
+     * <code>
+     * ?int $min - value of min parameter
+     * ?int $max - value of max parameter
+     * </code>.
+     *
+     * @throws \InvalidArgumentException
+     */
+    public static function getValidatedMinMaxValue(
```
danrot, line 46: What's the reason for this function to be static?

src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/SelectionPropertyMetadataMapper.php
```diff
@@ -0,0 +1,114 @@
...
+
+        if (null !== $min) {
+            Assert::nullOrGreaterThanEq($max, $min, \sprintf(
+                'Because parameter "%1$s" of property "%2$s" has value "%4$d", parameter "%3$s" needs to be greater than or equal "%4$d"',
+                $minParamName,
+                $propertyMetadata->getName(),
+                $maxParamName,
+                $min
+            ));
+        }
```
danrot, line 107: I just checked, and `webmozart/assert` does not seem to be a direct dependency of Sulu, so we can't really use it here... Unless we add it as dependency, but I guess we would have to decide as a team if we want that.

src/Sulu/Bundle/AdminBundle/Tests/Unit/Metadata/SchemaMetadata/SelectionPropertyMetadataMapperTest.php
```diff
@@ -0,0 +1,290 @@
...
+    public function testGetValidatedMinMaxValue(): void
+    {
+        $propertyMetadata = new PropertyMetadata();
+        $propertyMetadata->setName('property-name');
+        $propertyMetadata->setParameters([
+            ['name' => 'min', 'value' => 2],
+            ['name' => 'max', 'value' => 3],
+        ]);
+
+        $minMaxValue = SelectionPropertyMetadataMapper::getValidatedMinMaxValue($propertyMetadata);
```
danrot, line 110: Why not using the public `mapPropertyMetadata` function here? This might cause the tests to break, if we refactor the `PropertyMetadataMapper`, although everything might still work fine.

src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/SelectionPropertyMetadataMapper.php
luca-rath: To summarize that - there are two options:

1. All other constraints than `mandatory` (like `minLength`) do always apply, even if a field is optional and has no value.
Using this option, it's not possible to save a field with `mandatory: false` and `min: 2`, even if the field is optional (IMO optional fields don't need to be filled, no matter what constraints they have)
2. All other constraints than `mandatory` only apply, if a field has a value.
This option has the advantage to say: "The amount of teasers in a `text-with-teasers` block type can either be `0` or exactly `3`."

I think this is a use case that's needed quite often.
@sulu/core-team please share your opinions

src/Sulu/Bundle/MediaBundle/Content/Types/MediaSelectionContentType.php
```diff
@@ -200,10 +201,17 @@ public function preResolve(PropertyInterface $property)
     public function mapPropertyMetadata(ContentPropertyMetadata $propertyMetadata): PropertyMetadata
     {
         $mandatory = $propertyMetadata->isRequired();
-        $minItems = $mandatory ? 1 : 0;
+        $minMaxValue = SelectionPropertyMetadataMapper::getValidatedMinMaxValue($propertyMetadata);
```
danrot, line 204: Sorry to say, but this feels wrong to me... If we want to reuse stuff, the Symfony way would be to use DI for that, not static methods... That introduces hard coupling, makes testing harder, etc. I think there is a reason we are using DI nowadays.

src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/ObjectMetadata.php
danrot: Ah, seems like I have overseen that... And I think I was also confused, because the `SingleSelectionProperty` does not return an object, right? Is it correct to use that class then? I would say either the `PropertyMetadata` or the `SchemaMetadata` should be the fallback containing these additional properties, don't you agree?

src/Sulu/Component/Content/Metadata/ItemMetadata.php
danrot: So are you saying this didn't work at all before? :see_no_evil:

src/Sulu/Component/Content/Metadata/ItemMetadata.php
danrot: I think that's related to my question in https://github.com/sulu/sulu/pull/5666#discussion_r534032459. If I understand you correctly, this function was not working at all before? Resp. are there sometimes different representations, and the current implementation works for both of them?

src/Sulu/Bundle/AdminBundle/DependencyInjection/Compiler/AddPropertyMetadataMapperPass.php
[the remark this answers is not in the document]
luca-rath: I don't really have such a strong opinion on that, so I'm also up to implement your initial suggestion. But then I'd like to define the service in the `services.xml` and just add the proper tags in the`SuluAdminExtension`, does that sound good to you?

src/Sulu/Bundle/AdminBundle/DependencyInjection/Compiler/AddPropertyMetadataMapperPass.php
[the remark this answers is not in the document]
luca-rath: Okay, I'm also fine with doing this in the extension

src/Sulu/Component/Content/Metadata/ItemMetadata.php
luca-rath: I think so 😄

src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/SelectionPropertyMetadataMapper.php
alexander-schranz: @luca-rath In symfony "null" and "empty strings" are still valid for example constraints like:

 - https://symfony.com/doc/current/reference/constraints/Length.html 
 - https://symfony.com/doc/current/reference/constraints/Email.html
 - and many others

See there note:

> As with most of the other constraints, null and empty strings are considered valid values. This is to allow them to be optional values. If the value is mandatory, a common solution is to combine this constraint with NotBlank.

And I think we should go the same way, as it is the most flexible. So if something is not mandatory it can be still valid but I can tell that if it is set it need to have at least 10 letters in it.

src/Sulu/Component/Content/Metadata/ItemMetadata.php
luca-rath: Yes, that implementation did work for both representations, but I think it's not necessary, since `getParameter()` was not used somewhere else ...

src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/SelectionPropertyMetadataMapper.php
luca-rath: Since it's required by our dependencies either, I'm up for adding it to the `composer.json`.

src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/SelectionPropertyMetadataMapper.php
nnatter: Think its a good idea to make this consistent with symfony constraints if possible 👍

alexander-schranz: For JSON Schema when making a string also null able it would also work:

<details>

<summary>Example</summary>

```json
{
  "$schema": "http://json-schema.org/draft-04/schema#",
  "type": "object",
  "properties": {
    "test": {
      "type": ["string", "null"],
      "minLength": 2,
      "maxLength": 5
    },
    "test2": {
      "type": ["string", "null"],
      "minLength": 2,
      "maxLength": 5
    }
  },
  "required": [
    "test"
  ]
}
```

Example invalid:

```json
{
	"test": "invalid",
	"test2": "i",
}
```

```json
{
	"test": "valid",
	"test2": "12",
}
```

```json
{
	"test": "valid",
	"test2": null,
}
```

</details>

danrot: Well, if Symfony also implements it like this, it is OK for me, although it still feels weird to me.

src/Sulu/Bundle/AdminBundle/DependencyInjection/Compiler/AddPropertyMetadataMapperPass.php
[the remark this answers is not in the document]
danrot: Works for me, if that is also OK for you I would go with it :-)

src/Sulu/Component/Content/Metadata/ItemMetadata.php
danrot: Well, then it's a good thing it has been properly fixed now :see_no_evil: I guess what has been written in the unit test before was not how the data is actually structured, that's the disadvantages of these kind of tests :see_no_evil: In order to avoid something like that in the future, this parameter is now also called in some of the functional tests, right?

src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/SelectionPropertyMetadataMapper.php
luca-rath: okay, i think we got a decision ... will implement option 2

src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/SelectionPropertyMetadataMapper.php
danrot: Since it also seems to be a popular package with multiple maintainers, I think it is a safe choice. We just have to make sure that everybody knows it is available, so that we also use it in other situations :slightly_smiling_face: /cc @sulu/core-team

nnatter: is okay for me 🙂

src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/SelectionPropertyMetadataMapper.php
danrot: I was just thinking... Isn't that the kind of stuff that would make sense in a separate `PropertyMetadataMapper`, e.g. a `MinMaxPropertyMetadataMapper`? And every `MetadataMapper` that needs that, can inject that one? Code could then look something like this:

```php
class SelectionPropertyMetadataMapper
{
    private $minMaxPropertyMetadataMapper;

    public function __construct(MinMaxPropertyMetadataMapper $minMaxPropertyMetadataMapper)
    {
        $this->minMaxPropertyMetadataMapper = $minMaxPropertyMetadataMapper;
    }

    public function mapPropertyMetadata(ContentPropertyMetadata $propertyMetadata): PropertyMetadata
    {
        return $this->minMaxPropertyMetadataMapper
            ->mapPropertyMetadata($propertyMetadata) // returns a ArrayMetadata object
            ->withSchemaMetadata( // new function on the ArrayMetadataObject, that returns a new instance
                 new SchemaMetadata(
                     [],
                     [
                         new SchemaMetadata([], [], [], 'string'),
                         new SchemaMetadata([], [], [], 'number'),
                     ]
                 )
             );
    }
}
```

luca-rath: Just for documentation: I didn't implement it this way, because we probably need the `MinMaxValueResolver` in more places in the future ... One example could be `min_characters` and `max_characters` of the `text_line` field type

src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/SelectionPropertyMetadataMapper.php
luca-rath: Required it now, if someone has an objection, please share it here

src/Sulu/Bundle/AdminBundle/Resources/translations/admin.de.json
```diff
@@ -29,6 +29,8 @@
     "sulu_admin.error_required": "Dieses Feld muss ausgefüllt werden",
     "sulu_admin.error_minlength": "Die minimale Länge dieses Feldes wurde unterschritten",
     "sulu_admin.error_minitems": "Dieses Feld hat zu wenig Zuweisungen",
+    "sulu_admin.error_maxitems": "Dieses Feld hat zu viele Zuweisungen",
+    "sulu_admin.error_type": "Ein Fehler ist aufgetreten",
```
luca-rath, line 33: This error should not occurr and is likely a bug, therefore I decided to use a generic error message

src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/SelectionPropertyMetadataMapper.php
```diff
@@ -0,0 +1,53 @@
...
+            $minMaxValue->max,
+            true
+        );
+
+        if (!$mandatory) {
+            $selectionMetadata = new AnyOfsMetadata([
+                new NullMetadata(),
+                new EmptyArrayMetadata(),
+                $selectionMetadata,
+            ]);
```
luca-rath, line 48: The order of the `anyOf` schemas is important, because the admin ui will always display the error of the last `anyOf` schema

UPGRADE.md
```diff
@@ -2,6 +2,30 @@
...
+  * `new ArrayMetadata($name, $mandatory, $itemsSchema)` has to be changed to `new PropertyMetadata($name, $mandatory, new ArrayMetadata($itemsSchema))` and
+  * `new ConstMetadata($name, $mandatory, $value)` has to be changed to `new PropertyMetadata($name, $mandatory, new ConstMetadata($value))`.
+
+* Various new metadata classes (like `ObjectMetadata`, `StringMetadata`, `NumberMetadata`, ...) have been introduced to allow better schema definitions of a property.
+
+### The constructor of the `MediaSelectionContentType` requires a new `$propertyMetadataMinMaxValueResolver` for full functionality
+
+Without this service, `min` and `max` parameters of a `media_selection` property will not work.
+
+### The implementation of `ItemMetadata::getParameter($name)` has changed significantly
```
danrot, line 24: Isn't that the method, for which the previous implementation did not work at all? Because if it didn't work, I would not even mention it in the `UPGRADE.md`, since it seems like nobody has used it anyway.

src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/ConstMetadata.php
```diff
@@ -21,16 +21,16 @@ class ConstMetadata extends PropertyMetadata
...
+        parent::__construct();
+
         $this->value = $value;
     }
 
-    public function toJsonSchema(): ?array
+    public function toJsonSchema(): array
     {
         return [
-            'name' => $this->getName(),
```
danrot, line 33: In the `ArrayMetadata` you've called `parent::toJsonSchema` as well. Any reason this is not happening in this case?

src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/SchemaMetadata.php
```diff
@@ -95,6 +90,24 @@ function(PropertyMetadata $propertyMetadata) {
...
+                \array_filter(
+                    $this->properties,
+                    function(PropertyMetadata $propertyMetadata) {
+                        return $propertyMetadata->isMandatory();
+                    }
+                )
+            )
+        );
+
+        if (\count($required) > 0 || empty($jsonSchema)) {
```
danrot, line 107: Why the empty check?

src/Sulu/Bundle/AdminBundle/Tests/Unit/Metadata/SchemaMetadata/SingleSelectionPropertyMetadataMapperTest.php
```diff
@@ -0,0 +1,70 @@
...
+    {
+        $propertyMetadata = new PropertyMetadata();
+        $propertyMetadata->setName('property-name');
+
+        $jsonSchema = $this->singleSelectionPropertyMetadataMapper->mapPropertyMetadata($propertyMetadata)->toJsonSchema();
+
+        $this->assertEquals([
+            'name' => 'property-name',
+            'anyOf' => [
+                $this->getNullSchema(),
```
danrot, line 47: Any specific reason to extract that? It's only used once, the next two lines would have to be extracted then as well, and writing this line is longer than the content the function returns... I think I would inline it.

src/Sulu/Bundle/MediaBundle/Content/Types/MediaSelectionContentType.php
```diff
@@ -56,16 +60,23 @@ class MediaSelectionContentType extends ComplexContentType implements ContentTyp
...
         RequestAnalyzerInterface $requestAnalyzer = null,
-        $permissions = null
+        $permissions = null,
+        PropertyMetadataMinMaxValueResolver $propertyMetadataMinMaxValueResolver = null
     ) {
         $this->mediaManager = $mediaManager;
         $this->referenceStore = $referenceStore;
         $this->requestAnalyzer = $requestAnalyzer;
         $this->permissions = $permissions;
+        $this->propertyMetadataMinMaxValueResolver = $propertyMetadataMinMaxValueResolver;
```
danrot, line 79: Shouldn't a deprecation be triggered if the given value is null?

src/Sulu/Bundle/MediaBundle/Content/Types/MediaSelectionContentType.php
```diff
@@ -56,16 +60,23 @@ class MediaSelectionContentType extends ComplexContentType implements ContentTyp
...
+     */
+    private $propertyMetadataMinMaxValueResolver;
+
     public function __construct(
         MediaManagerInterface $mediaManager,
         ReferenceStoreInterface $referenceStore,
         RequestAnalyzerInterface $requestAnalyzer = null,
-        $permissions = null
+        $permissions = null,
+        PropertyMetadataMinMaxValueResolver $propertyMetadataMinMaxValueResolver = null
```
danrot, line 73: If we make it nullable we need to prepend the type with `?`, because IIRC this will trigger a deprecation in PHP 8.

src/Sulu/Bundle/MediaBundle/Resources/config/services.xml
```diff
@@ -310,6 +310,7 @@
             <argument type="service" id="sulu_media.reference_store.media"/>
             <argument type="service" id="sulu_core.webspace.request_analyzer"/>
             <argument>%sulu_security.permissions%</argument>
+            <argument type="service" id="sulu_admin.property_metadata_min_max_value_resolver" on-invalid="null"/>
```
danrot, line 313: Is the `on-invalid` really necessary? I think the service will always be there, won't it?

danrot (changes requested):

src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/ConstMetadata.php
luca-rath: The SchemaMetadata returns `['required' => []]`, if it's real schema is empty, because it always needs to return an associative array, otherwise the admin ui would treat it as array instead of an object and would break.

The `ArrayMetadata` initializes `SchemaMetadata` with type `array`, so `parent::toJsonSchema()` will return ['type' => 'array'].

But the `ConstMetadata` initializes `SchemaMetadata` with type `null`, so `parent::toJsonSchema()` would return ['required' => []] and we don't want the final json schema to look like `['required' => [], 'const' => 'any-value']`

src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/SchemaMetadata.php
luca-rath: Mostly explained here https://github.com/sulu/sulu/pull/5666#discussion_r535236464 ... In most cases, `['required' => []]` makes no sense, it's just used to tell the admin ui that the schema is an object instead of an array

src/Sulu/Bundle/AdminBundle/Tests/Unit/Metadata/SchemaMetadata/SingleSelectionPropertyMetadataMapperTest.php
luca-rath: In here it's used just once, right. But in some other classes (like `SelectionPropertyMetadataMapperTest`) it's used more often and I think it makes the code a lot more readable. And for consistency I thought I'm gonna add this utility function to all tests needing a `null`-schema

src/Sulu/Bundle/MediaBundle/Content/Types/MediaSelectionContentType.php
luca-rath: Since the `PropertyMetadataMinMaxValueResolver` is registered in the `AdminBundle`, it's only available in the admin context. Therefore this argument will always be `null` in the website context, so I can't trigger a deprecation unless I would also inject the context to check

src/Sulu/Bundle/MediaBundle/Resources/config/services.xml
luca-rath: Explained here https://github.com/sulu/sulu/pull/5666#discussion_r535240358

src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/ConstMetadata.php
danrot: Ok, that's a bit implicit, but I get it... Let's leave it this way.

src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/SchemaMetadata.php
danrot: Would add this as a comment here, otherwise it is not easy to understand.

src/Sulu/Bundle/AdminBundle/Tests/Unit/Metadata/SchemaMetadata/SingleSelectionPropertyMetadataMapperTest.php
danrot: Wouldn't have done it in the other places either, but I don't care about it too much, so it is ok for me.

src/Sulu/Bundle/MediaBundle/Content/Types/MediaSelectionContentType.php
danrot: Ok, makes sense.

src/Sulu/Bundle/AdminBundle/Resources/translations/admin.en.json
```diff
@@ -29,6 +29,8 @@
     "sulu_admin.error_required": "This field must have a value",
     "sulu_admin.error_minlength": "The minimal length of this field was undershot",
     "sulu_admin.error_minitems": "This field has too few selections",
+    "sulu_admin.error_maxitems": "This field has too much selections",
```
danrot, line 32: `too many selections`

danrot:
Not sure why, but I can't seem to assign an account in a `single_account_selection` configured like that:

```xml
        <property name="single_account_selection" type="single_account_selection" mandatory="true">
            <meta>
                <title lang="en">Single Account Selection</title>
                <title lang="de">Organisation</title>
            </meta>
        </property>
```

It always shows "An error occured".

danrot (approved):

PR #5670, opened 2020-11-30
Fallback to content-locale of user in ResourceLocator field-type

nnatter:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This PR adjusts the `ResourceLocator` field-type to use the `userStore.contentLocale` if a form is not localized.

#### Why?

Because the `ResourceLocator` components expects a non-null locale. Also, the API that is called in the `ResourceLocatorHistory` expects a non-null locale. Because of this, the `ResourceLocator` field-type is not usable for unlocalized entities at the moment (see https://github.com/sulu/sulu-demo/pull/87 for example).

We used the same fallback strategy in multiple other field-types including the `SingleSelection`, `TextEditor`, `MediaSelection` and `TeaserSelection`.


wachterjohannes (approved):
@nnatter but the PR fails

nnatter:
@wachterjohannes  The failing pipeline is not related to this PR - every pipeline will fail until https://github.com/sulu/sulu/pull/5669 is merged. 🙂

PR #5672, opened 2020-11-30
Add json schema validation for text_line and text_area field type

### LINKED ISSUE #4976: TagSelection not work with Tag containing a comma

luca-rath:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #4976
| License | MIT

#### What's in this PR?

Add json schema validation for `min_characters`, `max_characters` and `pattern` parameters of `text_line` and `text_area` field type


src/Sulu/Bundle/AdminBundle/Tests/Unit/Metadata/SchemaMetadata/SelectionPropertyMetadataMapperTest.php
```diff
@@ -186,32 +186,6 @@ public function testMapPropertyMetadataMinAndMaxMaxOnly(): void
...
-                        'anyOf' => [
-                            ['type' => 'string'],
-                            ['type' => 'number'],
-                        ],
-                    ],
-                    'uniqueItems' => true,
-                ],
-            ],
-        ], $jsonSchema);
-    }
```
luca-rath, line 213: This exact same test is already covered above in `testMapPropertyMetadata()`

src/Sulu/Bundle/MediaBundle/Tests/Unit/MediaSelectionContentTypeTest.php
```diff
@@ -649,44 +649,6 @@ public function testMapPropertyMetadataMinAndMaxMaxOnly(): void
...
-                        ],
-                        'displayOption' => [
-                            'type' => 'string',
-                            'name' => 'displayOption',
-                        ],
-                    ],
-                ],
-            ],
-        ], $jsonSchema);
-    }
```
luca-rath, line 688: This exact same test is already covered above in `testMapPropertyMetadata()`

src/Sulu/Bundle/PageBundle/Tests/Functional/Controller/AdminControllerTest.php
```diff
@@ -163,15 +163,15 @@ public function testPagesFormMetadataAction()
...
         $this->assertEquals('array', $overviewType->schema->properties->block->type);
         $this->assertCount(2, $overviewType->schema->properties->block->items->anyOf);
         $this->assertEquals(['title', 'type'], $overviewType->schema->properties->block->items->anyOf[0]->required);
-        $this->assertCount(1, (array) $overviewType->schema->properties->block->items->anyOf[0]->properties);
+        $this->assertCount(3, (array) $overviewType->schema->properties->block->items->anyOf[0]->properties);
         $this->assertEquals('type', $overviewType->schema->properties->block->items->anyOf[0]->properties->type->name);
         $this->assertEquals('type1', $overviewType->schema->properties->block->items->anyOf[0]->properties->type->const);
         $this->assertEquals(['image', 'type'], $overviewType->schema->properties->block->items->anyOf[1]->required);
-        $this->assertCount(1, (array) $overviewType->schema->properties->block->items->anyOf[1]->properties);
+        $this->assertCount(3, (array) $overviewType->schema->properties->block->items->anyOf[1]->properties);
```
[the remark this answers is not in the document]
luca-rath: The `SchemaMetadata` filters out every property without a schema. And because now every `text_line` property has a schema, it's not filtered out anymore.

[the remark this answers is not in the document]
danrot: Ah, ok... I think these tests are not well written, because it is pretty much unclear what's happening here :see_no_evil: But it is fine for now.

config/templates/pages/example.xml
```diff
@@ -126,6 +126,21 @@
...
+                <title lang="en">Text Line</title>
+                <title lang="de">Text</title>
+                <info_text lang="en">There are no commas allowed in this field</info_text>
+                <info_text lang="de">In diesem Feld sind keine Beistriche erlaubt</info_text>
+            </meta>
+
+            <params>
+                <param name="pattern" value="^[^,]*$"/>
+                <param name="min_characters" value="2"/>
+                <param name="max_characters" value="3"/>
```
danrot, line 140: There is already a soft limit on length in the `page_seo` form, why can I still save the page with too many character there, but it is not possible here? Seems a bit confusing to me :thinking:

luca-rath: because it's not a `text_line` 🙈

danrot: Looks like a text line to me :thinking: 
https://github.com/sulu/sulu/blob/cf138452aeb5cb1b252f8921806088e2a8997279/src/Sulu/Bundle/PageBundle/Resources/config/forms/page_seo.xml#L14

luca-rath: oops, my fault ... i was thinking of the seo description 😄

This is the schema, which is being generated
```json
{
  "type": "object",
  "properties": {
    "ext/seo/title": {
      "const": "foo"
    }
  }
}
```
and this the corresponding data
```json
{
  "ext": {
    "seo": {
      "title": "bar"
    }
  }
}
```
My guess is that ajv searches for a property called `ext/seo/title`, but that doesn't exist. there's just an `ext` property, which is the object we see above. And since we allow additional properties, the validation passes.

alexander-schranz: Does this override the soft limit? I think we should not override the exist softlimit behaviour as this is by design a soft limit because its just a recommendation. I think it would better to use `minLength` and `maxLength` for hard limit text_line and keep the soft limit as it is, maybe deprecate it and rename it but it should still be possible to have soft limit for the seo things.

luca-rath: But then I would rename the `max_characters` to something like `max_character_recommendation`, to make sure, that this is not a hard constraint. And also we should not use the `CharacterCounter` component for the `maxLength` then, to clearly differentiate between those two concepts.

luca-rath: But yes, sounds like a good idea to me

danrot: Yeah, using the `CharacterCounter` felt also weird to me... I think I would let the current param work as it did (no hard constraint) and deprecate it right now. In order to be consistent with the other params I would name the new ones `min` and `max`, and these should be the hard limit, and the form should fail if they happen. The deprecated ones could be replaced with something like `soft_min` and `soft_max`. I think `soft` has enough meaning in order to convey what it means and it is a lot less to type than `max_character_recommendation` (remember that in this place there also is no auto completion).

But I am also open to other suggestions.

luca-rath: I don't think I would name these parameters `min` and `max`, because that doesn't describe the resulting constraints very well. For selections it's pretty clear what `min` and `max` is meaning, the same would be for a number field. But I think for text fields i would call it `minLength` and `maxLength`, because there could also be `minSegments` and `maxSegments` ...

alexander-schranz: I would also go with `minLength` and `maxLength`. Because `min` and `max` for the `input` is something different https://www.w3schools.com/tags/tag_input.asp which would be used for `number` input.

danrot: Fine for me. What about the soft limit? `minSoftLength` and `maxSoftLength`?

luca-rath: is okay for me ... but currently we only have `max_characters` but no `min_characters`, so should we really implement a `minSoftLength`?

nnatter: guess i would use `softMinLength`, but `minSoftLength` is okay for me too 🙂

alexander-schranz: >  minSoftLength and maxSoftLength?

Is fine for me.

danrot: > is okay for me ... but currently we only have `max_characters` but no `min_characters`, so should we really implement a `minSoftLength`?

From my POV we can also leave it out... Not sure how that should be handled if you have set a soft min and max value :thinking:

luca-rath:
@danrot Just thought about it again, should we maybe create a more generic `TextPropertyMetadataMapper` for both, the `text_line` and the `text_area` at the same time? Or do you think there will be different constraints in the future?

danrot:
@luca-rath Sounds good to me... I guess one or the other property are less likely on a `text_area` than on a `text_line`, but I also think it does not cause any harm. And we can be sure that everything is consistent :+1: Is that already implemented as you have imagined it?

luca-rath:
> Is that already implemented as you have imagined it?

Yes it is 🙂

config/templates/pages/example.xml
```diff
@@ -126,6 +126,21 @@
...
+                <title lang="en">Text Line</title>
+                <title lang="de">Text</title>
+                <info_text lang="en">There are no commas allowed in this field</info_text>
+                <info_text lang="de">In diesem Feld sind keine Beistriche erlaubt</info_text>
+            </meta>
+
+            <params>
+                <param name="pattern" value="^[^,]*$"/>
+                <param name="minLength" value="2"/>
+                <param name="maxLength" value="3"/>
```
danrot, line 140: Don't we use underscores in parameters like this in other cases? :thinking: E.g. there is `present_as` for the smart content.

danrot (changes requested):

config/templates/pages/example.xml
alexander-schranz: I personal would go with [maxlength](https://www.w3schools.com/tags/att_input_maxlength.asp) like its written in html inputs without any case or underscore.

luca-rath: Since we never just concatenated words without a separator, I would not use `maxlength`. I agree with @danrot on keeping consistency by using snake case.
And we should probably also transform the camel case parameters of the `MediaSelection` into snake case.

luca-rath:
Failing appveyor has nothing to do with this pr, as you can see in the logs

danrot (approved):

PR #5677, opened 2020-12-01
Add resolveConflict parameter to createOrUpdateByAttributes method of RouteManager

### LINKED ISSUE #5579: Add resolveConflict parameter to createOrUpdateByAttributes method of RouteManager
# Problem description

When using the createOrUpdateByAttributes function of the RouteManager there is no option to check for conficting routes. If a conflict exists it cannot be resolved afterwards as the function already persists the route.

This problem can for example occur when using the ContentBundle (https://github.com/sulu/SuluContentBundle/blob/4d62683790248bda95fa243ad34b59078c79a2a2/Content/Application/ContentDataMapper/DataMapper/RoutableDataMapper.php#L168)

# Proposed solution

Add a parameter analogous to the create and update functions.

nnatter:
| Q | A
| --- | ---
| New feature? | yes
| Fixed tickets | fixes #5579
| License | MIT

#### What's in this PR?

This PR adds a `resolveConflict` parameter to the `RouteManagerInterface::createOrUpdateByAttributes` method. This makes the available parameter consistent to the `RouteManagerInterface::create` and `RouteManagerInterface::update`.

#### Why?

When using the `RouteBundle` for a custom entity, it is not possible to implement the `RoutableInterface` in some cases (eg for unlocalized entities or when using the SuluContentBundle). You need to use the `RouteManagerInterface::createOrUpdateByAttributes` method for these entities. We should allow to resolve possible conflicts in these cases too.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,16 @@
 # Upgrade
 
+## dev-master
+
+### Added resolveConflict parameter to RouteManagerInterface::createOrUpdateByAttributes
+
+The `createOrUpdateByAttributes` method of the `RouteManagerInterface` was adjusted to include a `resolveConflict`
+parameter. This makes the available parameters consistent to the `RouteManagerInterface::create` method and the
+`RouteManagerInterface::update` method.
+
+If you have implemented this interface in your project, you need to add the parameter to the 
+`createOrUpdateByAttributes` method of your implementation.
+
 ## 2.2.2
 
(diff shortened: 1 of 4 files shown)


nnatter:
Reminder to myself:
- test conflict with other entity if entity has no route yet
- test conflict with history route that is not current route

wachterjohannes (approved):
@nnatter add testcase for that

Issue #5676, opened 2020-12-01
Allow to configure route generation for custom entities consistent to pages


nnatter:
### Basics

There are two basic strategies when it comes to route generation at the moment.

#### Strategy 1: Live generation in Form based on `sulu.rlp.part` tag (used for pages)

In this strategy, properties that have a `sulu.rlp.part` tag are sent to the server via AJAX request when the form is edited. The value of the route is updated directly in the form, even without saving. On save, the already generated route-path is included in the data that sent to the API. This strategy is used for pages. 

#### Strategy 2: Generation based on entity (used for articles)

In this strategy, the route field in the form is updated only after save. When saving the form, the respective entity is persisted and the route bundle is used for generating the route-path for the persisted entity. The route bundle will generate the route-path by evaluating the configured route schema with the persisted entity. In contrast to strategy 1, this strategy allows to include the id of the entity in the route. Unfortunately, this strategy has considerable worse UX (see for example https://github.com/sulu/SuluArticleBundle/issues/514).

### Current Solution

When implementing a custom entity, I would like to use live generation for the route of the entity because it has better UX. 

This is already possible by the following steps:

1. Add a `sulu.rlp.part` tag to a property in the form of the entity. Now, if this property is changed, the ResourceLocator field-type will send an AJAX to the `RouteController::postAction`. The `RouteController` will generate the route-path by evaluating the configured route schema with the parts that were sent to the controller.

2. Because the `RouteController` will evaluate the configured route schema with the parts sent to the controller, we can now use a route schema like this:

```yml
sulu_route:
    mappings:
        App\Entity\Event:
            generator: schema
            options:
                route_schema: '/events/{object["title"]}-{object["subtitle"]}'
            resource_key: events
```

#### ~~Problem 1: Need explicitly include all parts in the configured route schema~~ 

~~When generating the route-path for a page. The parts that are tagged with `sulu.rlp.part` are sorted by their priority and concatenated using `-` between parts. I cannot easily mimic this behaviour for a custom entity without explicitly including all parts in the route schema.~~

Solved by #5675

#### Problem 2: Error when route schema is evaluated with entity 

Depending on how the custom entity is implemented, there might be parts of the system that will try to evaluate the route schema using the entity (not with the `parts` array like the `RouteController`). For example, this can happen when using the `bin/console sulu:route:update` command. 
In these cases, the generation will fail because the route schema (`'/events/{object["title"]}'`) expects that `object` is an array.

[...]

PR #5678, opened 2020-12-01
Resolve conflicts when generating route for entity in RouteController

nnatter:
| Q | A
| --- | ---
| Bug fix? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This PR adjusts the `RouteController` to resolve conflicts when generating the route path for an entity. This allows the content manager to see if a route is already occupied and prevents failed requests when saving the form. 

~~To do this, a `resolveByAttributes` method was added to the `ConflictResolverInterface`.~~ I adjusted the PR to use a temporary route for resolving conflicts instead of changing the interface.

#### Why?

Right now, when using live generation for the route of a custom entity as described in #5676, the generated route can conflict with an existing route in the system. If this happens, depending on the implementation of the entity, saving the form will result in an exception.

The current is inconsistent to the behaviour when creating a page. When creating a page, conflicts are resolved before the route is returned to the form (the form will display routes like `/parent-page/title-1` in this case). This provides a better UX because it leads to less exceptions and the content manager sees if a route is already occupied before saving the form.

### DIFF
--- a/src/Sulu/Bundle/RouteBundle/Controller/RouteController.php
+++ b/src/Sulu/Bundle/RouteBundle/Controller/RouteController.php
@@ -55,7 +55,7 @@ class RouteController extends AbstractRestController implements ClassResourceInt
     private $routeGenerator;
 
     /**
-     * @var ConflictResolverInterface
+     * @var ConflictResolverInterface|null
      */
     private $conflictResolver;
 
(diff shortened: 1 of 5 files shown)


src/Sulu/Bundle/RouteBundle/Manager/ConflictResolverInterface.php
```diff
@@ -24,4 +24,9 @@ interface ConflictResolverInterface
      * @return RouteInterface
      */
     public function resolve(RouteInterface $route);
+
+    /**
+     * @return string
+     */
+    public function resolveByAttributes(?string $entityClass, ?string $id, string $locale, string $path);
```
nnatter, line 31: not sure if adding this is a good idea 😕  maybe it would be better to create a temporary route in the `RouteController`? that would be no bc break and we would not need to maintain a `resolveByAttributes` method. the problem is that we might not have an id at that point

src/Sulu/Bundle/RouteBundle/Controller/RouteController.php
```diff
@@ -53,18 +54,25 @@ class RouteController extends AbstractRestController implements ClassResourceInt
...
+     * @var ConflictResolverInterface
+     */
+    private $conflictResolver;
+
     public function __construct(
         ViewHandlerInterface $viewHandler,
         RouteRepositoryInterface $routeRepository,
         EntityManagerInterface $entityManager,
         RouteGeneratorInterface $routeGenerator,
+        ConflictResolverInterface $conflictResolver,
```
nnatter, line 67: @alexander-schranz what is the correct way for doing this? can you give me a hint about an example somewhere? 🙂

nnatter: Fount one: https://github.com/sulu/sulu/blob/e13d40a362f27fa2fb4defbbff55404e9ccca707/src/Sulu/Bundle/MediaBundle/Controller/MediaController.php#L163-L168

PR #5679, opened 2020-12-02
Fix query issue when sorting and limiting

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Fixes the query built with the `DataProviderRepositoryTrait`.

#### Why?

Currently limiting is done first and therefore sorting only works on the first limited results.

### DIFF
--- a/src/Sulu/Component/SmartContent/Orm/DataProviderRepositoryTrait.php
+++ b/src/Sulu/Component/SmartContent/Orm/DataProviderRepositoryTrait.php
@@ -58,29 +58,26 @@ trait DataProviderRepositoryTrait
     {
         $parameter = [];
 
-        $queryBuilder = $this->createQueryBuilder('c')
-            ->select('c.id')
+        $queryBuilder = $this->createQueryBuilder('entity')
+            ->select('entity.id')
             ->distinct()
-            ->orderBy('c.id', 'ASC');
+            ->orderBy('entity.id', 'ASC');
 
-        $tagRelation = $this->appendTagsRelation($queryBuilder, 'c');
-        $categoryRelation = $this->appendCategoriesRelation($queryBuilder, 'c');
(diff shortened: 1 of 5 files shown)


src/Sulu/Component/SmartContent/Orm/DataProviderRepositoryTrait.php
```diff
@@ -58,29 +58,26 @@ private function findByFiltersIds($filters, $page, $pageSize, $limit, $locale, $
...
-            $sortMethod = \array_key_exists('sortMethod', $filters) ? $filters['sortMethod'] : 'asc';
-            $this->appendSortBy($filters['sortBy'], $sortMethod, $queryBuilder, 'c', $locale);
-
-            foreach ($filters['sortBy'] as $sortColumn) {
-                $queryBuilder->addSelect($sortColumn);
-            }
+        if (isset($filters['sortBy'])) {
+            $sortMethod = $filters['sortMethod'] ?? 'asc';
+            $this->appendSortBy($filters['sortBy'], $sortMethod, $queryBuilder, 'entity', $locale);
+            $queryBuilder->addSelect(str_contains($filters['sortBy'], '.') ? $filters['sortBy'] : 'entity' . '.' . $filters['sortBy']);
```
nnatter, line 72: i cant think about a reason why this is needed. going to remove it and try if the tests will still work 🙂

nnatter: the reason why this is needed is the `->distinct()` a few line above. we cannot remove that because returning duplicated ids would break pagination. therefore i added the select again

nnatter (approved):

PR #5690, opened 2020-12-09
Fix for image hight calculation

rogoit:
For version 1.6

#### What's in this PR?
A new line for the height calculation in Focus.php

#### Why?

I had problem on a organisation logo with the scale 250*60 px. Also when i upload a image in exact the same sice it was not calculated right. Result was 246*59px

#### Special
I do not test the running result by myself. It only was a Slack conversation and @alexander-schranz asked for the PR and this solution ;)


alexander-schranz:
@rogoit Thank you for the PR can you add also a testcase. Its simple for the focustest just add a line with the expected things here: https://github.com/sulu/sulu/blob/cf138452aeb5cb1b252f8921806088e2a8997279/src/Sulu/Bundle/MediaBundle/Tests/Unit/Media/ImageConverter/Focus/FocusTest.php#L57

alexander-schranz (approved):
Looks good for me. Will wait for the tests to complete.

nnatter (approved):
Thanks a lot for your effort! 🙂

PR #5703, opened 2020-12-12
Add dimensions properties to image media

RealAestan:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | N/A
| Related issues/PRs | 
| License | MIT

#### What's in this PR?

This PR adds the width and height of the image files to the media file properties.

#### Why?

I need the dimensions of the image files for a javascript library.

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/Media/Manager/MediaManager.php
+++ b/src/Sulu/Bundle/MediaBundle/Media/Manager/MediaManager.php
@@ -15,6 +15,7 @@ use Doctrine\DBAL\DBALException;
 use Doctrine\ORM\EntityManager;
 use FFMpeg\Exception\ExecutableNotFoundException;
 use FFMpeg\FFProbe;
+use Imagine\Exception\InvalidArgumentException;
 use Imagine\Image\ImagineInterface;
 use Sulu\Bundle\AudienceTargetingBundle\Entity\TargetGroupRepositoryInterface;
 use Sulu\Bundle\CategoryBundle\Entity\CategoryRepositoryInterface;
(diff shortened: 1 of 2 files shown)


src/Sulu/Bundle/MediaBundle/Media/Manager/MediaManager.php
```diff
@@ -311,9 +311,15 @@ private function getProperties(UploadedFile $uploadedFile)
                 } catch (\RuntimeException $e) {
                     // Exception is thrown if the dimension could not be extracted
                 }
+            } catch (ExecutableNotFoundException $e) {
+                // Exception is thrown if ffmpeg is not installed -> video properties are not set
+            }
+        } elseif (\fnmatch('image/*', $mimeType)) {
+            $dimensions = \getimagesize($uploadedFile->getPathname());
```
alexander-schranz, line 318: As written in the comment above `getimagesize` is a gd specific function to support imagick, gmagick or vips we need to replace this with the `imagine` service which should be injected as additonal optional service:

```php
if ($this->imagine) {
    $image = $this->imagine->open(..);
    $properties['width'] = $image->getSize()->getHeight();
    $properties['height'] = $image->getSize()->getWidth();
}
```

alexander-schranz (changes requested):
@RealAestan thank you for your PR, this is definitely interested information today with lazy loaded images.

Sulu supports different image processes using the [imagine library](https://github.com/avalanche123/Imagine). Imagine is an abstraction around gd, imagick, gmagick, vips image processors in php. So we need to inject the `imagine` service read this information because exist project maybe use imagick instead of gd. 

So to inject the imagine service we need to add [here](https://github.com/sulu/sulu/blob/7ca6a5e0d22d69c69d45f39ca52a156e75d2adfc/src/Sulu/Bundle/MediaBundle/Resources/config/services.xml#L273) the:

```xml
<argument type="service" id="sulu_media.adapter"/>
```

service. To keep Backward compatibility this argument need to be optional by doing:

```
__construct(
     ...
     ?ImagineInterface $imagine = null
) {
       $this->imagine = $imagine;

       if (!$imagine) {
              // throw deprecation message
       }
}
```

Let me know if you need any help to implement this.

src/Sulu/Bundle/MediaBundle/Media/Manager/MediaManager.php
RealAestan: All modifications done, sorry for the delay.

alexander-schranz (approved):

niklasnatter:
@RealAestan Thanks for your effort! 🙂

PR #5705, opened 2020-12-14
Refactor visibleCondition and disabledCondition to allow different blocks having different evaluations

### LINKED ISSUE #5291: Allow to use relative visibleCondition (or disableCondition) in block
| Q | A
| --- | ---
| Sulu version | 2.1.0-RC1 
| Browser version | Chrome for Mac versie 81.0.4044.138 (Officiële build) (64-bits)

# Actual Behavior

We're trying to hide/show a field on condition of a checkbox in SULU, but nothing happens when toggling the checkbox. This is 2 levels into nested blocks.

```
<property name="loadSubPages" type="checkbox">
    <meta>
        <title lang="nl">Alle onderliggende pagina's laden</title>
    </meta>
</property>

<block name="subMenu" default-type="subMenuItem" minOccurs="0" visibleCondition="!loadSubPages">
```

# Expected Behavior

When the checkbox is checked the block subMenu should be hidden, and vice versa.

# Steps to Reproduce

Configure a XML with the above mentioned properties, nest it in another block. Then try to toggle the loadSubPages, nothing happens.

# Possible Solutions

When the checkbox is toggled, hide the other property.

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #5291
| Related issues/PRs | ---
| License | MIT
| Documentation PR | sulu/sulu-docs#609

#### What's in this PR?

This PR will allow to have different evaluations of multiple blocks.

#### BC Breaks/Deprecations

TODO

#### To Do

- [x] Implement `visibleCondition` for sections
- [x] `Renderer` should retrieve all data via the `data` prop and only the current one via `value`, so that the same semantics as before the refactoring can be established
- [x] Double check if no `visible` and `disabled` flags are used anymore in the tests
- [x] Support multiple nesting levels like `__parent.__parent.__parent.title`

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
@@ -146,7 +146,7 @@ class FieldBlocks extends React.Component<FieldTypeProps<Array<BlockEntry>>> {
 
             const blockSettingsTag = schemaEntry.tags.find((tag) => tag.name === SETTINGS_TAG);
 
-            if (blockSettingsTag && schemaEntry.visible !== false) {
+            if (blockSettingsTag) {
                 iconsMapping[SETTINGS_PREFIX + schemaKey] = blockSettingsTag.attributes.icon;
             }
 
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
@@ -246,6 +246,7 @@ class FieldBlocks extends React.Component<FieldTypeProps<Array<BlockEntry>>> {
 
     renderExpandedBlockContent = (value: Object, type: string, index: number) => {
         const {
+            data,
             dataPath,
             error,
             formInspector,
@@ -261,7 +262,7 @@ class FieldBlocks extends React.Component<FieldTypeProps<Array<BlockEntry>>> {
 
         return (
             <FieldRenderer
-                data={value}
+                data={data}
                 dataPath={dataPath + '/' + index}
                 errors={errors && errors.length > index && errors[index] ? errors[index] : undefined}
                 formInspector={formInspector}
@@ -273,6 +274,7 @@ class FieldBlocks extends React.Component<FieldTypeProps<Array<BlockEntry>>> {
                 schema={blockSchemaType.form}
                 schemaPath={schemaPath + '/types/' + type + '/form'}
                 showAllErrors={showAllErrors}
+                value={value}
             />
         );
     };
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/Field.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/Field.js
@@ -3,15 +3,18 @@ import React from 'react';
 import {computed} from 'mobx';
 import {observer} from 'mobx-react';
 import log from 'loglevel';
+import jexl from 'jexl';
 import Router from '../../services/Router';
 import {translate} from '../../utils';
-import FieldComponent from '../../components/Form/Field';
+import Form from '../../components/Form';
+import conditionDataProviderRegistry from './registries/conditionDataProviderRegistry';
 import fieldRegistry from './registries/fieldRegistry';
 import fieldStyles from './field.scss';
 import FormInspector from './FormInspector';
 import type {Error, ErrorCollection, SchemaEntry} from './types';
 
 type Props = {|
+    data: Object,
     dataPath: string,
     error?: Error | ErrorCollection,
     formInspector: FormInspector,
@@ -32,10 +35,42 @@ class Field extends React.Component<Props> {
         showAllErrors: false,
     };
 
-    handleChange = (value: *) => {
-        const {name, onChange, schema} = this.props;
+    @computed get conditionData() {
+        const {data, formInspector} = this.props;
+        const {locale, metadataOptions, options} = formInspector;
 
-        if (schema.disabled) {
+        return conditionDataProviderRegistry.getAll().reduce(
+            function(data, conditionDataProvider) {
+                return {...data, ...conditionDataProvider(data, options, metadataOptions)};
+            },
+            {...data, __locale: locale}
+        );
+    }
+
+    @computed get disabled() {
+        const {schema} = this.props;
+
+        if (!schema.disabledCondition) {
+            return false;
+        }
+
+        return jexl.evalSync(schema.disabledCondition, this.conditionData);
+    }
+
+    @computed get visible() {
+        const {schema} = this.props;
+
+        if (!schema.visibleCondition) {
+            return true;
+        }
+
+        return jexl.evalSync(schema.visibleCondition, this.conditionData);
+    }
+
+    handleChange = (value: *) => {
+        const {name, onChange} = this.props;
+
+        if (this.disabled) {
             return;
         }
 
@@ -87,6 +122,10 @@ class Field extends React.Component<Props> {
     }
 
     render() {
+        if (!this.visible) {
+            return null;
+        }
+
         const {
             dataPath,
             error,
@@ -103,7 +142,6 @@ class Field extends React.Component<Props> {
         const {
             defaultType,
             description,
-            disabled,
             label,
             maxOccurs,
             minOccurs,
@@ -125,7 +163,7 @@ class Field extends React.Component<Props> {
             log.error(e);
 
             return (
-                <FieldComponent
+                <Form.Field
                     colSpan={schema.colSpan}
                     spaceAfter={schema.spaceAfter}
                 >
@@ -140,7 +178,7 @@ class Field extends React.Component<Props> {
                             </div>
                         </div>
                     </div>
-                </FieldComponent>
+                </Form.Field>
             );
         }
         const fieldTypeOptions = fieldRegistry.getOptions(type);
@@ -148,7 +186,7 @@ class Field extends React.Component<Props> {
         const errorKeyword = this.findErrorKeyword(error);
 
         return (
-            <FieldComponent
+            <Form.Field
                 colSpan={schema.colSpan}
                 description={description}
                 error={errorKeyword ? translate('sulu_admin.error_' + errorKeyword.toLowerCase()) : undefined}
@@ -162,7 +200,7 @@ class Field extends React.Component<Props> {
                         <FieldType
                             dataPath={dataPath}
                             defaultType={defaultType}
-                            disabled={disabled}
+                            disabled={this.disabled}
                             error={error}
                             fieldTypeOptions={fieldTypeOptions}
                             formInspector={formInspector}
@@ -181,7 +219,7 @@ class Field extends React.Component<Props> {
                         />
                     </div>
                 </div>
-            </FieldComponent>
+            </Form.Field>
         );
     }
 }
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/Field.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/Field.js
@@ -36,14 +36,13 @@ class Field extends React.Component<Props> {
     };
 
     @computed get conditionData() {
-        const {data, formInspector} = this.props;
-        const {locale, metadataOptions, options} = formInspector;
+        const {data, dataPath, formInspector} = this.props;
 
         return conditionDataProviderRegistry.getAll().reduce(
             function(data, conditionDataProvider) {
-                return {...data, ...conditionDataProvider(data, options, metadataOptions)};
+                return {...data, ...conditionDataProvider(data, dataPath, formInspector)};
             },
-            {...data, __locale: locale}
+            {...data}
         );
     }
 
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/Field.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/Field.js
@@ -127,6 +127,7 @@ class Field extends React.Component<Props> {
         }
 
         const {
+            data,
             dataPath,
             error,
             formInspector,
@@ -198,6 +199,7 @@ class Field extends React.Component<Props> {
                 <div className={fieldStyles.fieldContainer}>
                     <div className={fieldStyles.field}>
                         <FieldType
+                            data={data}
                             dataPath={dataPath}
                             defaultType={defaultType}
                             disabled={this.disabled}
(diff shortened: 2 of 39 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/Field.js
```diff
@@ -32,10 +35,42 @@ class Field extends React.Component<Props> {
...
+
+    @computed get visible() {
+        const {schema} = this.props;
+
+        if (!schema.visibleCondition) {
+            return true;
+        }
+
+        return jexl.evalSync(schema.visibleCondition, this.conditionData);
+    }
```
danrot, line 68: I've tried to add the evaluations in this file, because in here the two conditions can be evaluated for this exact field, without the need to work with array/objects or whatever. Since the `Field` is already an observer, only the necessary fields are reevaluated.

danrot: But there is one problem with that approach. The [`Renderer` component](https://github.com/sulu/sulu/blob/dc4b04b77af71286cae13aad44d704a83e1755ee/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/Renderer.js#L44) also renders the form sections, so the `visibleCondition` handling would have to be copied there. But I think I would move the rendering of `Section` in this `Field` component, to reuse that logic.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/Field.js
```diff
@@ -3,15 +3,18 @@ import React from 'react';
...
-import FieldComponent from '../../components/Form/Field';
+import Form from '../../components/Form';
+import conditionDataProviderRegistry from './registries/conditionDataProviderRegistry';
 import fieldRegistry from './registries/fieldRegistry';
 import fieldStyles from './field.scss';
 import FormInspector from './FormInspector';
 import type {Error, ErrorCollection, SchemaEntry} from './types';
 
 type Props = {|
+    data: Object,
```
danrot, line 17: Here the `Field` component gets the entire data from the form, which allows us to move the evaluation of the conditions into this component, which works quite well. However, there is still the issue with the `Renderer` component being reused in e.g. blocks. These `Renderer` instances get [only the data from the block](https://github.com/sulu/sulu/blob/dc4b04b77af71286cae13aad44d704a83e1755ee/src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js#L264), not from the entire form. So the `data` prop I have introduce here for all the data already exists in the `Renderer`, but it only gets a part of the data, and I would like to have stuff consistent here. Basically two ways are possible for handling this:

1. The `data` prop in this line can be renamed to something like `allData` or `rootData`, then the same name could be used in the `Renderer` component.
2. We change the meaning of the `data` prop of the `Renderer` in order to match it with the one from the `Field`, and rename the current `data` attribute from the `Renderer` to `value`, then it would match the `value` prop of the `Field`.

danrot (commented):
@alexander-schranz @nnatter WDYT about these issues?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/Field.js
nnatter: I think i would prefer option 2. Option 1 would mean that the `Renderer` has a `rootData`, `dataPath` and a `data` prop right?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/Field.js
nnatter: Would be okay for me, but I am not sure if `Field` is still the correct name then

danrot: Not really reimplement, I think it will end up looking a bit different in general, but the `renderItem` method in the `Renderer` would be gone.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/Field.js
danrot: Yeah, option 1 would have that consequence.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/Field.js
danrot: And regarding the name, what would you suggest? Initially I would have handle the section exactly the same as another field type, in order to avoid a situation like we have now :see_no_evil:

nnatter: maybe just `Item`?

danrot: I've introduced another `Section` component in here, and duplicated the `visibleCondition` getter for now... Getting both of them to work as a single component is not that easy, because they are sharing quite some data via recursion... I'll have another look at that at the end, if there is some time left.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
```diff
@@ -146,7 +146,7 @@ class FieldBlocks extends React.Component<FieldTypeProps<Array<BlockEntry>>> {
 
             const blockSettingsTag = schemaEntry.tags.find((tag) => tag.name === SETTINGS_TAG);
 
-            if (blockSettingsTag && schemaEntry.visible !== false) {
```
danrot, line 149: This allowed to hide icons when the field for which the icon should be displayed was invisible. This is not so easy to rebuild anymore, because we only have the `visibleCondition` here. Therefore I just removed that check, I guess that wasn't really used anyway.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/conditionDataProviders/parentConditionDataProvider.js
```diff
@@ -0,0 +1,10 @@
+// @flow
+import jsonpointer from 'json-pointer';
+
+export default function(data: Object, dataPath: ?string): {[string]: any} {
+    if (!dataPath) {
+        return {__parent: data};
+    }
+
+    return {__parent: jsonpointer.get(data, dataPath.substring(0, dataPath.lastIndexOf('/')))};
```
danrot, line 9: I think it is not easily possible to implement this `__parent` variable recursively while still keeping the advantages of mobx's reactivity... I have tried something like this:

```javascript
// @flow
import {isObservableArray} from 'mobx';

export default function(data: Object, dataPath: ?string): {[string]: any} {
    if (!dataPath) {
        return {__parent: data};
    }

    let parentDataPath = dataPath;
    const conditionData = {};
    let currentConditionData = conditionData;

    do {
        parentDataPath = parentDataPath.substring(0, parentDataPath.lastIndexOf('/'));
        const evaluatedData = jsonpointer.get(data, parentDataPath);

        if (Array.isArray(evaluatedData) || isObservableArray(evaluatedData)) {
            continue;
        }

        currentConditionData.__parent = evaluatedData;
        currentConditionData = currentConditionData.__parent;
    } while (parentDataPath.match(/^\/.*\//));

    return conditionData;
}
```

The problem is that the reference from the `ResourceStore` is manipulated here (adding the `__parent` variable). This ends in an endless loop, I guess that's because of some side effect... Even if we solve that issue, the side effects would still remain.

I guess using `toJS` somehow would work, but then too much code would be executed, because mobx cannot observe the single properties of that object correctly anymore, and making the rest a lot harder to implement. Also, I am not sure if that worse performance is a good tradeoff now, since we don't even know if somebody would make use of that...

danrot: Maybe using a [Proxy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) would work :thinking: But then IE11 support is not possible anymore for sure...

danrot: Nevermind, seems to work if the `evaluatedData` variable is spread. Also, still only the `render` method of the fields impacted by the currently edited field are called.

nnatter: you already solved this right? 🙂 
if not, maybe it would also work to evaluate the `__parent` lazy inside of a [getter on the object](https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Functions/get)?

danrot: That wouldn't solve the problem, because we would still adjust the existing object. And yeah, I solved it by creating the objects using the `...` spread operator.

src/Sulu/Bundle/MediaBundle/Resources/js/containers/Form/fields/ImageMap/ImageMap.js
```diff
@@ -110,7 +111,7 @@ class ImageMap extends React.Component<FieldTypeProps<Value>> {
 
         return (
             <FieldRenderer
-                data={value}
+                data={data}
```
nnatter, line 114: we need to document this in the `UPGRADE.md`

src/Sulu/Bundle/PageBundle/Resources/js/containers/Form/conditionDataProviders/webspaceConditionDataProvider.js
```diff
@@ -1,12 +1,14 @@
 // @flow
 import {toJS} from 'mobx';
+import {FormInspector} from 'sulu-admin-bundle/containers';
 import webspaceStore from '../../../stores/webspaceStore';
 
 export default function(
     data: {[string]: any},
-    options: {[string]: any},
-    metadataOptions: ?{[string]: any}
+    dataPath: ?string,
```
nnatter, line 8: we need to document this in the `UPGRADE.md`

danrot:
Writing upgrades was an open task anyway :see_no_evil: Should be done now. Will quickly check if there is some documentation to check.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/tests/conditionDataProviders/parentConditionDataProvider.test.js
```diff
@@ -0,0 +1,58 @@
...
+
+    expect(parentConditionDataProvider(data, '/blocks/1/title')).toEqual({__parent: {title: 'Block title 2'}});
+});
+
+test('Return parent for nested second block', () => {
+    const data = {
+        title: 'Title',
+        blocks: [
+            {
+                blocks: [
```
nnatter, line 39: would be nice to add another property to the outer block for understandability 🙂

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/conditionDataProviders/parentConditionDataProvider.js
```diff
@@ -0,0 +1,27 @@
...
+        parentDataPath = parentDataPath.substring(0, parentDataPath.lastIndexOf('/'));
+        const evaluatedData = jsonpointer.get(data, parentDataPath);
+
+        if (Array.isArray(evaluatedData) || isObservableArray(evaluatedData)) {
+            continue;
+        }
+
+        currentConditionData.__parent = {...evaluatedData};
+        currentConditionData = currentConditionData.__parent;
+    } while (parentDataPath.match(/^\/.*\//));
```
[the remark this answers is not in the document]
danrot: Yeah, that's what it does. We could use the code snippet you've provided, but isn't that even more confusing? If I just read that line I would have to put some thinking into it, to realize that it checks if there are two slashes.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/AbstractFormStore.js
```diff
@@ -1,15 +1,13 @@
...
 import jsonpointer from 'json-pointer';
 import log from 'loglevel';
-import type {RawSchema, RawSchemaEntry, Schema, SchemaEntry} from '../types';
-import conditionDataProviderRegistry from '../registries/conditionDataProviderRegistry';
+import type {Schema, SchemaEntry} from '../types';
 
 const SECTION_TYPE = 'section';
 
-function addSchemaProperties(data: Object, key: string, schema: RawSchema) {
+function addSchemaProperties(data: Object, key: string, schema: Schema) {
```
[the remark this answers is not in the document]
danrot: Strictly speaking yes, will add it.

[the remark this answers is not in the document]
danrot: Actually... Wasn't the `RawSchema` used only internally? AFAIK the `schema` property of the `AbstractFormStore` was already returning a `Schema`. Also I don't want to spam the `UPGRADE.md` file too much, because people don't read it entirely already :see_no_evil:

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/AbstractFormStore.js
```diff
@@ -177,17 +104,13 @@ export default class AbstractFormStore
     +metadataOptions: ?{[string]: any};
     +loading: boolean;
     +locale: ?IObservableValue<string>;
-    @observable rawSchema: RawSchema;
+    schema: Schema;
     @observable evaluatedSchema: Schema = {};
```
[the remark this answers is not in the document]
danrot: No, just forgot to delete it.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/AbstractFormStore.js
```diff
@@ -249,25 +166,6 @@ export default class AbstractFormStore
...
-                    locale,
-                    '',
-                    this.options,
-                    this.metadataOptions
-                )
-            )
-        );
-    };
-
     @action setEvaluatedSchema(evaluatedSchema: Schema) {
```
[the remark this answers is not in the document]
danrot: No, also forgot to delete it.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/conditionDataProviders/parentConditionDataProvider.js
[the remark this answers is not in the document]
nnatter: you are right, it probably does not make a big difference 😅

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/AbstractFormStore.js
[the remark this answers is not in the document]
nnatter: i dont know -  its okay for me to not document it as well. i really doubt that anybody relies on this change 🙂

nnatter (approved):

PR #5719, opened 2020-12-23
Test frontend application using GitHub Actions instead of CircleCI

nnatter:
Unfortunately, the pipeline on CircleCI regularly fails because of scarce memory. See for example: https://app.circleci.com/pipelines/github/sulu/sulu/2510/workflows/58ee3915-b72c-4de1-9a1c-a4bfcc584f6a/jobs/10383


nnatter:
Replaced by https://github.com/sulu/sulu/pull/5720

PR #5720, opened 2020-12-23
Test frontend application using GitHub Actions instead of CircleCI

nnatter:
Unfortunately, the pipeline on CircleCI regularly fails because of scarce memory. See for example: https://app.circleci.com/pipelines/github/sulu/sulu/2510/workflows/58ee3915-b72c-4de1-9a1c-a4bfcc584f6a/jobs/10383


.github/workflows/test-application.yaml
```diff
@@ -61,7 +61,59 @@ jobs:
...
+        name: Test JavaScript application
+        runs-on: ubuntu-latest
+        steps:
+            - name: Checkout project
+              uses: actions/checkout@v2
+
+            - name: Install and configure Node
+              uses: actions/setup-node@v2-beta
+              with:
+                  node-version: '12'
```
luca-rath, line 74: Should we maybe add a matrix to test with node 12 and 14?

.github/workflows/test-application.yaml
```diff
@@ -61,7 +61,59 @@ jobs:
...
+
+            - name: Get npm cache directory
+              id: npm-cache-dir
+              run: echo "::set-output name=dir::$(npm config get cache)"
+
+            - name: Cache dependencies
+              uses: actions/cache@v2
+              with:
+                  path: ${{ steps.npm-cache-dir.outputs.dir }}
+                  key: ${{ runner.os }}-composer-${{ hashFiles('**/package-lock.json') }}
```
luca-rath, line 84: Rename to `nm`

.github/workflows/test-application.yaml
```diff
@@ -61,7 +61,59 @@ jobs:
...
+                  npm run lint:js
+                  npm run lint:scss
+                  npm run flow -- check
+                  npm run depcruise
+
+            - name: Test JavaScript code
+              run: npm test -- --maxWorkers=4
+
+            - name: Test Styleguidist build
+              run: npm run styleguide:build
```
luca-rath, line 103: Is this really necessary since there is an additional workflow building the styleguide for `jsdocs.sulu.io`?

nnatter: i think so - the `jsdocs.sulu.io` action is not executed for pull request, right? 🙂

luca-rath: ah yes, you're right 👍

luca-rath (approved):

PR #5737, opened 2021-01-08
Update Inflector to new Version 1.4 || 2.0

bramtervoort:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets |  - 
| Related issues/PRs |   - 
| License | MIT
| Documentation PR | -

#### What's in this PR?

Fixed deprecation messages for Doctrine\Common\Inflector\Inflector these calls now get the new Inflecto api

#### Why?

This should make future updates a little easier.

#### To Do

- [x] update composer.json "doctrine/inflector": "^1.4 || ^2.0",

### DIFF
--- a/src/Sulu/Bundle/CoreBundle/Entity/ApiEntity.php
+++ b/src/Sulu/Bundle/CoreBundle/Entity/ApiEntity.php
@@ -11,7 +11,7 @@
 
 namespace Sulu\Bundle\CoreBundle\Entity;
 
-use Doctrine\Common\Inflector\Inflector;
+use Doctrine\Inflector\InflectorFactory;
 use JMS\Serializer\Annotation\Accessor;
 use JMS\Serializer\Annotation\Exclude;
 
@@ -81,7 +81,9 @@ abstract class ApiEntity
         // if no apiPath is not set generate it from basepath
         if (\is_null($this->getApiPath())) {
             $class = \explode('\\', \get_class($this));
-            $plural = Inflector::pluralize(\strtolower(\end($class)));
+            // this is deprecated already so no use to do fancy to get the factory
+            $inflector = InflectorFactory::create()->build();
+            $plural = $inflector->pluralize(\strtolower(\end($class)));
             $this->apiPath = $this->apiBasePath . '/' . $plural;
         }
 
--- a/src/Sulu/Bundle/CoreBundle/Entity/ApiEntity.php
+++ b/src/Sulu/Bundle/CoreBundle/Entity/ApiEntity.php
@@ -81,7 +81,6 @@ abstract class ApiEntity
         // if no apiPath is not set generate it from basepath
         if (\is_null($this->getApiPath())) {
             $class = \explode('\\', \get_class($this));
-            // this is deprecated already so no use to do fancy to get the factory
             $inflector = InflectorFactory::create()->build();
             $plural = $inflector->pluralize(\strtolower(\end($class)));
             $this->apiPath = $this->apiBasePath . '/' . $plural;
--- a/src/Sulu/Component/Content/Metadata/Factory/StructureMetadataFactory.php
+++ b/src/Sulu/Component/Content/Metadata/Factory/StructureMetadataFactory.php
@@ -11,7 +11,7 @@
 
 namespace Sulu\Component\Content\Metadata\Factory;
 
-use Doctrine\Common\Inflector\Inflector;
+use Doctrine\Inflector\InflectorFactory;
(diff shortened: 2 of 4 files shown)


alexander-schranz:
@bramtervoort Thx for the PR. Can you target the `2.1` branch for this change? And changing the `composer.json` is fine. We need here `^1.4.1 || ^2.0.1` (we can so remove the conflicts to 1.4.0 and 2.0.0 as they are not longer in the range).
This is no breaking change as we have a requirement to `handcraftedinthealps/rest-routing-bundle` which defines min version of `doctrine/inflector` to `1.4`.

bramtervoort:
Thank you. I rebased it. if there is anything else needed i'd be haopy to do that.

src/Sulu/Bundle/CoreBundle/Entity/ApiEntity.php
```diff
@@ -81,7 +81,9 @@ public function createSelfLink()
         // if no apiPath is not set generate it from basepath
         if (\is_null($this->getApiPath())) {
             $class = \explode('\\', \get_class($this));
-            $plural = Inflector::pluralize(\strtolower(\end($class)));
+            // this is deprecated already so no use to do fancy to get the factory
+            $inflector = InflectorFactory::create()->build();
```
alexander-schranz: Would remove this comment does more confuse me as it helps 🙈 

```suggestion
            $inflector = InflectorFactory::create()->build();
```

src/Sulu/Component/Content/Metadata/Factory/StructureMetadataFactory.php
```diff
@@ -52,6 +52,11 @@ class StructureMetadataFactory implements StructureMetadataFactoryInterface
      */
     private $cache = [];
 
+    /**
+     * @var \Doctrine\Inflector\Inflector
```
alexander-schranz: add here `use \Doctrine\Inflector\Inflector;` for this:

```suggestion
     * @var Inflector
```

src/Sulu/Component/DocumentManager/Subscriber/Behavior/Path/AliasFilingSubscriber.php
```diff
@@ -28,13 +28,19 @@ class AliasFilingSubscriber extends AbstractFilingSubscriber
      */
     private $metadataFactory;
 
+    /**
+     * @var \Doctrine\Inflector\Inflector
```
alexander-schranz: same as above, add here `use \Doctrine\Inflector\Inflector;` for this:

```suggestion
     * @var Inflector
```

alexander-schranz (changes requested):
Some little change requests. Else it looks fine from my side.

bramtervoort:
Just applied the suggestions. Github does make that very easy. Curious if this work correctly with the tests since I did not verify the php cs of this.

alexander-schranz (approved):

src/Sulu/Component/Content/Metadata/Factory/StructureMetadataFactory.php
```diff
@@ -52,6 +52,11 @@ class StructureMetadataFactory implements StructureMetadataFactoryInterface
      */
     private $cache = [];
 
+    /**
+     * @var Inflector
```
alexander-schranz, line 57: Looks for me like the `use` statement is missing here for this class

src/Sulu/Component/DocumentManager/Subscriber/Behavior/Path/AliasFilingSubscriber.php
```diff
@@ -28,13 +28,19 @@ class AliasFilingSubscriber extends AbstractFilingSubscriber
      */
     private $metadataFactory;
 
+    /**
+     * @var Inflector
```
alexander-schranz, line 33: Looks for me like the `use` statement is missing here for this class

alexander-schranz (changes requested):
@bramtervoort you can also just edit the files I just write the suggestion to maybe make it easier for you but if manual change it does work better for you use that. Currently I think the use statements are missing so the IDE would fail to find the Inflector class.

bramtervoort:
yes it does work better in the editor. Thank you. i updated the use statements

alexander-schranz (approved):

alexander-schranz:
@bramtervoort Thank your for upgrading this! 🚀

PR #5749, opened 2021-01-14
Fix if no segements are set

Quehnie:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Fixed segments index missing in controller.

#### Why?

Update from sulu 2.0 to sulu 2.2.3 segments are not given.

#### To Do

### DIFF
--- a/src/Sulu/Bundle/WebsiteBundle/Routing/ContentRouteProvider.php
+++ b/src/Sulu/Bundle/WebsiteBundle/Routing/ContentRouteProvider.php
@@ -208,7 +208,7 @@ class ContentRouteProvider implements RouteProviderInterface
                 return $collection;
             } else {
                 if ($document instanceof ExtensionBehavior) {
-                    $documentSegments = $document->getExtensionsData()['excerpt']['segments'];
+                    $documentSegments = $document->getExtensionsData()['excerpt']['segments'] ?? [];
                     $documentSegmentKey = $documentSegments[$portal->getWebspace()->getKey()] ?? null;
                     $segment = $this->requestAnalyzer->getSegment();
 
--- a/src/Sulu/Bundle/RouteBundle/Routing/RouteProvider.php
+++ b/src/Sulu/Bundle/RouteBundle/Routing/RouteProvider.php
@@ -148,7 +148,7 @@ class RouteProvider implements RouteProviderInterface
 
         if ($routeObject instanceof ExtensionBehavior) {
             $portal = $attributes->getAttribute('portal');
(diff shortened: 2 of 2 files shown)


Quehnie:
was a mistake on our site overwriting the excerpt tab...

alexander-schranz:
@nnatter how should we handle this. If somebody overide the excerpt and has the "segements" not in it `segements` are not set in the excerptdata. Currently it did just fail in the contentrouteprovider but not sure if there are other cases we need to test if the segments field type is removed from excerpt.

src/Sulu/Bundle/WebsiteBundle/Routing/ContentRouteProvider.php
```diff
@@ -208,7 +208,7 @@ public function getRouteCollectionForRequest(Request $request)
                 return $collection;
             } else {
                 if ($document instanceof ExtensionBehavior) {
-                    $documentSegments = $document->getExtensionsData()['excerpt']['segments'];
```
niklasnatter, line 211: please adjust this in the `Sulu\Bundle\RouteBundle\Routing\RouteProvider` too

niklasnatter:
> Currently it did just fail in the contentrouteprovider but not sure if there are other cases we need to test if the segments field type is removed from excerpt.

How can you remove a field from the excerpt? wouldnt you just set the `visibleCondition` to `false`? 🤔 
That said, I think it should not hurt anybody to fallback to an empty array. So this is okay for me.

alexander-schranz:
@nnatter excerpt override need to override the `excerpt.xml` in your  `pages` folder and if you don't add there the segments again in that excerpt xml it will totally get lost.

alexander-schranz:
@nnatter we worked around this problem with `xinclude` here https://github.com/sulu/sulu-demo/pull/76/files#diff-e63b89f4647abb90d9c7fee3a8be28b6be23875590a74317a71a35a25ecb0417R16-R17
but @Quehnie did not use xinclude instead a whole copy of the excerpt.xml.

niklasnatter:
> @nnatter excerpt override need to override the excerpt.xml in your pages folder and if you don't add there the segments again in that excerpt xml it will totally get lost.

You are right 🙂

src/Sulu/Bundle/WebsiteBundle/Routing/ContentRouteProvider.php
alexander-schranz: @nnatter did adjust it.

PR #5758, opened 2021-01-18
Add badges to tabs

luca-rath:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | https://github.com/sulu/SuluAutomationBundle/issues/29
| License | MIT

#### What's in this PR?

Add badges to tabs

#### Why?

To allow SuluAutomationBundle to show amount of tasks on the `Automation` tab

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -10,7 +10,7 @@ with this source code in the file LICENSE.
 EOF;
 
 $finder = PhpCsFixer\Finder::create()
-    ->exclude(['var/cache', 'tests/Resources/cache'])
+    ->exclude(['var/cache', 'tests/Resources/cache', 'node_modules'])
     ->in(__DIR__);
 
 return PhpCsFixer\Config::create()
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -10,7 +10,7 @@ with this source code in the file LICENSE.
 EOF;
 
 $finder = PhpCsFixer\Finder::create()
-    ->exclude(['var/cache', 'tests/Resources/cache', 'node_modules'])
+    ->exclude(['var/cache', 'tests/Resources/cache'])
     ->in(__DIR__);
 
 return PhpCsFixer\Config::create()
--- a/package.json
+++ b/package.json
@@ -88,8 +88,7 @@
         "webpack": "^4.20.2",
         "webpack-clean-obsolete-chunks": "^0.4.0",
         "webpack-cli": "^3.1.1",
-        "webpack-manifest-plugin": "^2.0.2",
-        "whatwg-fetch": "^3.5.0"
+        "webpack-manifest-plugin": "^2.0.2"
     },
     "jest": {
         "moduleNameMapper": {
--- a/package.json
+++ b/package.json
@@ -88,7 +88,8 @@
         "webpack": "^4.20.2",
         "webpack-clean-obsolete-chunks": "^0.4.0",
         "webpack-cli": "^3.1.1",
-        "webpack-manifest-plugin": "^2.0.2"
+        "webpack-manifest-plugin": "^2.0.2",
+        "whatwg-fetch": "^3.5.0"
     },
     "jest": {
         "moduleNameMapper": {
--- a/src/Sulu/Bundle/AdminBundle/Admin/View/Badge.php
+++ b/src/Sulu/Bundle/AdminBundle/Admin/View/Badge.php
@@ -11,66 +11,63 @@
 
 namespace Sulu\Bundle\AdminBundle\Admin\View;
 
+use JMS\Serializer\Annotation as Serializer;
+
 class Badge
 {
     /**
      * @var string
+     *
+     * @Serializer\Expose()
+     * @Serializer\Groups({"fullView"})
      */
     private $routeName;
 
     /**
      * @var string|null
+     *
+     * @Serializer\Expose()
+     * @Serializer\Groups({"fullView"})
      */
     private $dataPath;
 
     /**
      * @var string|null
+     *
+     * @Serializer\Expose()
+     * @Serializer\Groups({"fullView"})
      */
     private $visibleCondition;
 
     /**
      * @var array<string, mixed>
+     *
+     * @Serializer\Expose()
+     * @Serializer\Groups({"fullView"})
      */
-    private $attributesToRequest = [];
+    private $requestParameters = [];
 
     /**
      * @var array<string, mixed>
+     *
+     * @Serializer\Expose()
+     * @Serializer\Groups({"fullView"})
      */
     private $routerAttributesToRequest = [];
 
-    public function __construct(string $routeName, ?string $dataPath = null)
+    public function __construct(string $routeName, ?string $dataPath = null, ?string $visibleCondition = null)
     {
         $this->routeName = $routeName;
         $this->dataPath = $dataPath;
-    }
-
-    public function setRouteName(string $routeName): self
-    {
-        $this->routeName = $routeName;
-
-        return $this;
-    }
-
-    public function setDataPath(?string $dataPath): self
-    {
-        $this->dataPath = $dataPath;
-
-        return $this;
-    }
-
-    public function setVisibleCondition(?string $visibleCondition): self
-    {
         $this->visibleCondition = $visibleCondition;
-
-        return $this;
     }
 
     /**
-     * @param array<string, mixed> $attributesToRequest
+     * @param array<string, mixed> $requestParameters
      */
-    public function addAttributesToRequest(array $attributesToRequest): self
+    public function addRequestParameters(array $requestParameters): self
     {
-        $this->attributesToRequest = \array_merge($this->attributesToRequest, $attributesToRequest);
+        $this->requestParameters = \array_merge($this->requestParameters, $requestParameters);
 
         return $this;
     }
@@ -83,21 +80,5 @@ class Badge
         $this->routerAttributesToRequest = \array_merge($this->routerAttributesToRequest, $routerAttributesToRequest);
 
         return $this;
-    }
-
-    /**
-     * @internal
-     *
-     * @return array<string, mixed>
-     */
-    public function getConfiguration(): array
-    {
-        return [
-            'routeName' => $this->routeName,
-            'dataPath' => $this->dataPath,
-            'visibleCondition' => $this->visibleCondition,
-            'attributesToRequest' => $this->attributesToRequest,
-            'routerAttributesToRequest' => $this->routerAttributesToRequest,
-        ];
     }
 }
--- a/src/Sulu/Bundle/AdminBundle/Admin/View/Badge.php
+++ b/src/Sulu/Bundle/AdminBundle/Admin/View/Badge.php
@@ -1,0 +1,103 @@
+<?php
+
+/*
+ * This file is part of Sulu.
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\AdminBundle\Admin\View;
+
+class Badge
+{
+    /**
+     * @var string
+     */
+    private $routeName;
+
+    /**
+     * @var string|null
+     */
+    private $dataPath;
+
+    /**
+     * @var string|null
+     */
+    private $visibleCondition;
+
+    /**
+     * @var array<string, mixed>
+     */
+    private $attributesToRequest = [];
+
+    /**
+     * @var array<string, mixed>
+     */
+    private $routerAttributesToRequest = [];
+
+    public function __construct(string $routeName, ?string $dataPath = null)
+    {
+        $this->routeName = $routeName;
+        $this->dataPath = $dataPath;
+    }
+
+    public function setRouteName(string $routeName): self
+    {
+        $this->routeName = $routeName;
+
+        return $this;
+    }
+
+    public function setDataPath(?string $dataPath): self
+    {
+        $this->dataPath = $dataPath;
+
+        return $this;
+    }
+
+    public function setVisibleCondition(?string $visibleCondition): self
+    {
+        $this->visibleCondition = $visibleCondition;
+
+        return $this;
+    }
+
+    /**
+     * @param array<string, mixed> $attributesToRequest
+     */
+    public function addAttributesToRequest(array $attributesToRequest): self
+    {
+        $this->attributesToRequest = \array_merge($this->attributesToRequest, $attributesToRequest);
+
+        return $this;
+    }
+
+    /**
+     * @param array<string, mixed> $routerAttributesToRequest
+     */
+    public function addRouterAttributesToRequest(array $routerAttributesToRequest): self
+    {
+        $this->routerAttributesToRequest = \array_merge($this->routerAttributesToRequest, $routerAttributesToRequest);
+
+        return $this;
+    }
+
+    /**
+     * @internal
+     *
+     * @return array<string, mixed>
+     */
+    public function getConfiguration(): array
+    {
+        return [
+            'routeName' => $this->routeName,
+            'dataPath' => $this->dataPath,
+            'visibleCondition' => $this->visibleCondition,
+            'attributesToRequest' => $this->attributesToRequest,
+            'routerAttributesToRequest' => $this->routerAttributesToRequest,
+        ];
+    }
+}
--- a/src/Sulu/Bundle/AdminBundle/Admin/View/Badge.php
+++ b/src/Sulu/Bundle/AdminBundle/Admin/View/Badge.php
@@ -76,7 +76,7 @@ class Badge
     }
 
     /**
-     * @param array<string, mixed> $routerAttributesToRequest
+     * @param array<string|int, string> $routerAttributesToRequest
      */
     public function addRouterAttributesToRequest(array $routerAttributesToRequest): self
     {
--- a/src/Sulu/Bundle/AdminBundle/Admin/View/Badge.php
+++ b/src/Sulu/Bundle/AdminBundle/Admin/View/Badge.php
@@ -63,6 +63,29 @@ class Badge
         $this->visibleCondition = $visibleCondition;
     }
 
+    public function getRouteName(): string
+    {
+        return $this->routeName;
+    }
+
+    public function getDataPath(): ?string
+    {
+        return $this->dataPath;
+    }
+
+    public function getVisibleCondition(): ?string
+    {
+        return $this->visibleCondition;
+    }
+
+    /**
+     * @return array<string, mixed>
+     */
+    public function getRequestParameters(): array
+    {
+        return $this->requestParameters;
+    }
+
     /**
      * @param array<string, mixed> $requestParameters
      */
@@ -71,6 +94,14 @@ class Badge
         $this->requestParameters = \array_merge($this->requestParameters, $requestParameters);
 
         return $this;
+    }
+
+    /**
+     * @return array<string|int, string>
+     */
+    public function getRouterAttributesToRequest(): array
+    {
+        return $this->routerAttributesToRequest;
     }
 
     /**
--- a/src/Sulu/Bundle/AdminBundle/Admin/View/Badge.php
+++ b/src/Sulu/Bundle/AdminBundle/Admin/View/Badge.php
@@ -11,47 +11,48 @@
 
 namespace Sulu\Bundle\AdminBundle\Admin\View;
 
-use JMS\Serializer\Annotation as Serializer;
+use JMS\Serializer\Annotation\Expose;
+use JMS\Serializer\Annotation\Groups;
 
 class Badge
 {
     /**
      * @var string
      *
-     * @Serializer\Expose()
-     * @Serializer\Groups({"fullView"})
+     * @Expose()
+     * @Groups({"fullView"})
      */
     private $routeName;
 
     /**
      * @var string|null
      *
-     * @Serializer\Expose()
-     * @Serializer\Groups({"fullView"})
+     * @Expose()
+     * @Groups({"fullView"})
      */
     private $dataPath;
 
     /**
      * @var string|null
      *
-     * @Serializer\Expose()
-     * @Serializer\Groups({"fullView"})
+     * @Expose()
+     * @Groups({"fullView"})
      */
     private $visibleCondition;
 
     /**
      * @var array<string, mixed>
      *
-     * @Serializer\Expose()
-     * @Serializer\Groups({"fullView"})
+     * @Expose()
+     * @Groups({"fullView"})
      */
     private $requestParameters = [];
 
     /**
      * @var array<string, mixed>
      *
-     * @Serializer\Expose()
-     * @Serializer\Groups({"fullView"})
+     * @Expose()
+     * @Groups({"fullView"})
      */
     private $routerAttributesToRequest = [];
 
--- a/src/Sulu/Bundle/AdminBundle/Admin/View/TabViewChildBuilderTrait.php
+++ b/src/Sulu/Bundle/AdminBundle/Admin/View/TabViewChildBuilderTrait.php
@@ -33,10 +33,15 @@ trait TabViewChildBuilderTrait
         $view->setOption('tabPriority', $tabPriority);
     }
 
-    private function addTabBadgeToView(View $view, Badge $badge): void
+    private function addTabBadgeToView(View $view, Badge $badge, ?string $key = null): void
     {
         $tabBadges = $view->getOption('tabBadges') ?? [];
-        $tabBadges[] = $badge->getConfiguration();
+
+        if (null === $key) {
+            $tabBadges[] = $badge->getConfiguration();
+        } else {
+            $tabBadges[$key] = $badge->getConfiguration();
+        }
 
         $view->setOption('tabBadges', $tabBadges);
     }
--- a/src/Sulu/Bundle/AdminBundle/Admin/View/TabViewChildBuilderTrait.php
+++ b/src/Sulu/Bundle/AdminBundle/Admin/View/TabViewChildBuilderTrait.php
@@ -32,4 +32,12 @@ trait TabViewChildBuilderTrait
     {
         $view->setOption('tabPriority', $tabPriority);
     }
+
+    private function addTabBadgeToView(View $view, Badge $badge): void
+    {
+        $tabBadges = $view->getOption('tabBadges') ?? [];
+        $tabBadges[] = $badge->getConfiguration();
+
+        $view->setOption('tabBadges', $tabBadges);
+    }
 }
--- a/src/Sulu/Bundle/AdminBundle/Admin/View/TabViewChildBuilderTrait.php
+++ b/src/Sulu/Bundle/AdminBundle/Admin/View/TabViewChildBuilderTrait.php
@@ -33,16 +33,18 @@ trait TabViewChildBuilderTrait
         $view->setOption('tabPriority', $tabPriority);
     }
 
-    private function addTabBadgeToView(View $view, Badge $badge, ?string $key = null): void
+    private function addTabBadgeToView(View $view, Badge $badge): void
     {
         $tabBadges = $view->getOption('tabBadges') ?? [];
+        $tabBadges[] = $badge;
 
-        if (null === $key) {
-            $tabBadges[] = $badge->getConfiguration();
-        } else {
-            $tabBadges[$key] = $badge->getConfiguration();
-        }
+        $view->setOption('tabBadges', $tabBadges);
+    }
 
+    private function setTabBadgeToView(View $view, string $key, Badge $badge): void
+    {
+        $tabBadges = $view->getOption('tabBadges') ?? [];
+        $tabBadges[$key] = $badge;
         $view->setOption('tabBadges', $tabBadges);
     }
 }
--- a/src/Sulu/Bundle/AdminBundle/Admin/View/TabViewChildBuilderTrait.php
+++ b/src/Sulu/Bundle/AdminBundle/Admin/View/TabViewChildBuilderTrait.php
@@ -33,18 +33,16 @@ trait TabViewChildBuilderTrait
         $view->setOption('tabPriority', $tabPriority);
     }
 
-    private function addTabBadgeToView(View $view, Badge $badge): void
+    /**
+     * @param View $view
+     * @param Badge[] $badges
+     */
+    private function addTabBadgesToView(View $view, array $badges): void
     {
-        $tabBadges = $view->getOption('tabBadges') ?? [];
-        $tabBadges[] = $badge;
-
-        $view->setOption('tabBadges', $tabBadges);
-    }
-
-    private function setTabBadgeToView(View $view, string $key, Badge $badge): void
-    {
-        $tabBadges = $view->getOption('tabBadges') ?? [];
-        $tabBadges[$key] = $badge;
-        $view->setOption('tabBadges', $tabBadges);
+        $oldBadges = $view->getOption('tabBadges');
+        $newBadges = $oldBadges
+            ? \array_merge($oldBadges, $badges)
+            : $badges;
+        $view->setOption('tabBadges', $newBadges);
     }
 }
--- a/src/Sulu/Bundle/AdminBundle/Admin/View/TabViewChildBuilderTrait.php
+++ b/src/Sulu/Bundle/AdminBundle/Admin/View/TabViewChildBuilderTrait.php
@@ -34,7 +34,6 @@ trait TabViewChildBuilderTrait
     }
 
     /**
-     * @param View $view
      * @param Badge[] $badges
      */
     private function addTabBadgesToView(View $view, array $badges): void
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Badge/badge.scss
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Badge/badge.scss
@@ -1,0 +1,12 @@
+@import '../../containers/Application/colors.scss';
+
+.badge {
+    border-radius: 2px;
+    background: $persianRed;
+    color: white;
+    padding: 2px 4px;
+    width: min-content;
+    height: min-content;
+    font-size: 9px;
+    font-weight: bold;
+}
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Badge/badge.scss
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Badge/badge.scss
@@ -2,7 +2,7 @@
 
 .badge {
     border-radius: 2px;
-    background: $persianRed;
+    background: $blueZodiac;
     color: white;
     padding: 2px 4px;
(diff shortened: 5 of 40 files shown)


luca-rath:
@alexander-schranz @nnatter Should this pr target the `2.1` or `2.x` branch?

.php_cs.dist
```diff
@@ -10,7 +10,7 @@ with this source code in the file LICENSE.
 EOF;
 
 $finder = PhpCsFixer\Finder::create()
-    ->exclude(['var/cache', 'tests/Resources/cache'])
+    ->exclude(['var/cache', 'tests/Resources/cache', 'node_modules'])
```
luca-rath: php-cs-fixer always tries to fix some php files in `node_modules/flatted`

alexander-schranz:
From my side we could target `2.x` and do a:

```
if (method_exists($viewBuilder, 'addBadge')) {
    $viewBuilder->addBadge(...);
}
```

In the automation bundle to make it possible to install with a lower Sulu version as 2.3.

luca-rath:
Yes, was thinking about doing the same. We also did something similar for the state icons in the list view.

.php_cs.dist
luca-rath: I guess this should be fixed in 2.1, so I'll create an additional pull request ... Didn't think about that before

src/Sulu/Bundle/AdminBundle/Admin/View/Badge.php
```diff
@@ -0,0 +1,103 @@
...
+    {
+        $this->visibleCondition = $visibleCondition;
+
+        return $this;
+    }
+
+    /**
+     * @param array<string, mixed> $attributesToRequest
+     */
+    public function addAttributesToRequest(array $attributesToRequest): self
```
niklasnatter: i think we called this `requestParameters` in most other places

src/Sulu/Bundle/AdminBundle/Admin/View/Badge.php
```diff
@@ -0,0 +1,103 @@
...
+
+        return $this;
+    }
+
+    /**
+     * @internal
+     *
+     * @return array<string, mixed>
+     */
+    public function getConfiguration(): array
```
niklasnatter: as far as i see, we used serializer annotations in all other classes

src/Sulu/Bundle/AdminBundle/Admin/View/TabViewChildBuilderTrait.php
```diff
@@ -32,4 +32,17 @@ private function setTabPriorityToView(View $view, int $tabPriority): void
     {
         $view->setOption('tabPriority', $tabPriority);
     }
+
+    private function addTabBadgeToView(View $view, Badge $badge, ?string $key = null): void
```
niklasnatter: what the purpose of the `$key`? would like to kee this consistent with the `ToolbarAction` and  `ListItemAction`

src/Sulu/Bundle/AdminBundle/Resources/js/components/Badge/badge.scss
```diff
@@ -0,0 +1,12 @@
+@import '../../containers/Application/colors.scss';
+
+.badge {
+    border-radius: 2px;
+    background: $persianRed;
```
niklasnatter: did we use red in older versions too?
doesnt this look like an error? 🤔

src/Sulu/Bundle/AdminBundle/Resources/js/components/Tabs/Tab.js
```diff
@@ -1,10 +1,11 @@
 // @flow
 import React from 'react';
-import type {ElementRef} from 'react';
+import type {Element, ElementRef} from 'react';
 import classNames from 'classnames';
 import tabStyles from './tab.scss';
 
 type Props = {
+    badges: Element<*>[],
```
niklasnatter, line 8: shouldnt we use something like `Element<typeof Badge>>`?

src/Sulu/Bundle/AdminBundle/Resources/js/components/Tabs/tab.scss
```diff
@@ -70,3 +70,11 @@ $buttonSmallHorizontalPadding: 10px;
         }
     }
 }
+
+.badges {
+    margin-left: 4px;
+    display: inline-flex;
+    vertical-align: text-top;
+    gap: 1px;
+    transform: translateY(-3px);
```
niklasnatter: whats the reason for this? would prefer solutions without hardcoded numbers for centering

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Badge/stores/BadgeStore.js
```diff
@@ -0,0 +1,93 @@
...
+import Requester from '../../../services/Requester';
+
+export default class BadgeStore {
+    router: Router;
+    routeName: string;
+    dataPath: ?string;
+    visibleCondition: ?string;
+    attributesToRequest: Object;
+    routerAttributesToRequest: Object;
+    @observable text: ?string = null;
```
niklasnatter: think i would call this `value`

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Badge/stores/BadgeStore.js
```diff
@@ -0,0 +1,93 @@
...
+        routerAttributesToRequest: Object
+    ) {
+        this.router = router;
+        this.routeName = routeName;
+        this.dataPath = dataPath;
+        this.visibleCondition = visibleCondition;
+        this.attributesToRequest = attributesToRequest;
+        this.routerAttributesToRequest = routerAttributesToRequest;
+
+        this.disposer = autorun(() => {
```
niklasnatter: when should this `autorun`?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Badge/Badge.js
```diff
@@ -0,0 +1,63 @@
...
+    componentWillUnmount() {
+        this.store.destroy();
+    }
+
+    render() {
+        if (!this.store || this.store.text === null || this.store.text === undefined) {
+            return null;
+        }
+
+        return <BadgeComponent>{this.store.text}</BadgeComponent>;
```
niklasnatter: would expect that the visible condition is evaluated here instead of inside of the store

src/Sulu/Bundle/AdminBundle/Resources/js/views/Tabs/Tabs.js
```diff
@@ -7,8 +7,17 @@ import TabsComponent from '../../components/Tabs';
 import type {ViewProps} from '../../containers/ViewRenderer';
 import {translate} from '../../utils/Translator';
 import {Route} from '../../services/Router';
+import Badge from '../../containers/Badge';
 import tabsStyles from './tabs.scss';
 
+type BadgeType = {
```
niklasnatter: would move this to a `types.js` and call it `Badge` or `BadgeOptions`

src/Sulu/Bundle/AdminBundle/Resources/js/views/Tabs/Tabs.js
```diff
@@ -134,8 +143,27 @@ class Tabs<T> extends React.Component<Props<T>> {
                     <TabsComponent onSelect={this.handleSelect} selectedIndex={selectedTabIndex}>
                         {this.sortedTabRoutes.map((tabRoute) => {
                             const tabTitle = tabRoute.options.tabTitle;
+                            const tabBadges = tabRoute.options.tabBadges;
```
niklasnatter: using `[]` would simplify the logic below

src/Sulu/Bundle/AdminBundle/Resources/js/views/Tabs/Tabs.js
```diff
@@ -134,8 +143,27 @@ class Tabs<T> extends React.Component<Props<T>> {
...
+
+                            let badges = undefined;
+                            if (tabBadges) {
+                                badges = (Object.values(tabBadges): any).map((badge: BadgeType, index: number) => {
+                                    return (
```
niklasnatter: we check the type of the object in other places:
https://github.com/sulu/sulu/blob/2.x/src/Sulu/Bundle/AdminBundle/Resources/js/views/Form/Form.js#L280-L295
https://github.com/sulu/sulu/blob/2.x/src/Sulu/Bundle/AdminBundle/Resources/js/views/List/List.js#L246-L250

src/Sulu/Bundle/AdminBundle/Admin/View/Badge.php
```diff
@@ -0,0 +1,103 @@
...
+     */
+    private $routerAttributesToRequest = [];
+
+    public function __construct(string $routeName, ?string $dataPath = null)
+    {
+        $this->routeName = $routeName;
+        $this->dataPath = $dataPath;
+    }
+
+    public function setRouteName(string $routeName): self
```
niklasnatter: we do not provide setters in the `ToolbarAction` class or the  `ListItemAction` class. would like to keep this consistent

src/Sulu/Bundle/AdminBundle/Admin/View/Badge.php
```diff
@@ -0,0 +1,103 @@
...
+     * @var array<string, mixed>
+     */
+    private $attributesToRequest = [];
+
+    /**
+     * @var array<string, mixed>
+     */
+    private $routerAttributesToRequest = [];
+
+    public function __construct(string $routeName, ?string $dataPath = null)
```
niklasnatter: whats the reason for not adding the `visibleCondition` to the constructor?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Badge/stores/BadgeStore.js
```diff
@@ -0,0 +1,93 @@
...
+        const {
+            router: {
+                attributes: routerAttributes,
+            },
+            routeName,
+            attributesToRequest,
+            routerAttributesToRequest,
+        } = this;
+
+        let requestAttributes = {};
```
niklasnatter: i think we evaluated this inside of the component in other places right? not sure if there was a specific reason for that

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Badge/stores/BadgeStore.js
```diff
@@ -0,0 +1,93 @@
...
+        const {dataPath, visibleCondition} = this;
+
+        let enhancedData = data;
+        if (dataPath !== null) {
+            enhancedData = jsonPointer.get(data, dataPath);
+        }
+        const text = enhancedData.toString();
+
+        if (visibleCondition !== null) {
+            const result = jexl.evalSync(visibleCondition, {text});
```
niklasnatter: would expect this to happen in the component

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Badge/tests/Badge.test.js
```diff
@@ -0,0 +1,55 @@
...
+}));
+
+jest.mock('../../../services/Router', () => jest.fn(function() {
+    this.attributes = {
+        id: 5,
+        locale: 'en',
+    };
+}));
+
+test('Should create a new BadgeStore', () => {
```
niklasnatter: we should test if the correct props are passed to the children

src/Sulu/Bundle/AdminBundle/Admin/View/TabViewChildBuilderTrait.php
luca-rath: To add the possibility to override a badge. In the SuluAutomationBundle, I can only add the badge in the `AutomationViewBuilder::setEntityClass()` method. So if someone would call that method twice, 2 badges would be added to the tab

src/Sulu/Bundle/AdminBundle/Resources/js/components/Badge/badge.scss
luca-rath: I'm pretty sure it was also red before. /cc @sulu/core-team what do you think which background color the badge should have?

src/Sulu/Bundle/AdminBundle/Resources/js/components/Tabs/Tab.js
luca-rath: The Problem is that the `<Badge/>` which is being passed here is a `container`, not a `component`. And I think there is a rule that `components` should not have dependencies on `containers`, right?

src/Sulu/Bundle/AdminBundle/Resources/js/components/Tabs/tab.scss
luca-rath: To lift the badge up a little bit ...

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Badge/stores/BadgeStore.js
luca-rath: Initially and on tab change ...

src/Sulu/Bundle/AdminBundle/Resources/js/views/Tabs/Tabs.js
luca-rath: I don't know what you are meaning

src/Sulu/Bundle/AdminBundle/Admin/View/Badge.php
luca-rath: Just forgot about it

src/Sulu/Bundle/AdminBundle/Admin/View/TabViewChildBuilderTrait.php
niklasnatter: as said, i would like to keep this consistent with the `ToolbarAction` and  `ListItemAction`. maybe we can implement a `setTabBadgeToView` for all of them that has a mandatory `key` parameter?

src/Sulu/Bundle/AdminBundle/Resources/js/components/Tabs/Tab.js
niklasnatter: okay thats possible indeed 😕

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Badge/stores/BadgeStore.js
niklasnatter: but why is it run on tab change this way? 🤔

src/Sulu/Bundle/AdminBundle/Resources/js/views/Tabs/Tabs.js
niklasnatter: i forgot some word above 😓  
using `const tabBadges = tabRoute.options.tabBadges || []` would allow to omit the if condition below

src/Sulu/Bundle/AdminBundle/Admin/View/TabViewChildBuilderTrait.php
luca-rath: so you mean a `addTabBadgeToView(View $view, Badge $badge)` and a `setTabBadgeToView(View $view, string $key, Badge $badge)`?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Badge/stores/BadgeStore.js
luca-rath: hmm ... I have no idea 😄

src/Sulu/Bundle/AdminBundle/Admin/View/TabViewChildBuilderTrait.php
niklasnatter: yes, it feels strange to overwrite something in the `addTabBadgeToView` method. and a `set...ToView` method is probably useful for the `ToolbarAction` and `ListItemAction` too 🙂

luca-rath: you can just do `$viewBuilder->addToolbarActions(['foo' => new ToolbarAction('save')])` and override it with `$viewBuilder->addToolbarActions(['foo' => new ToolbarAction('save_with_publishing')])` ... right? So maybe we should just rename the method to `addTabBadges(array $badges)`?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Badge/stores/BadgeStore.js
luca-rath: Currently, the autorun only triggers when changing tabs, not when a new Task is added or deleted. Do you have an idea what else we should listen to to update the badge when new data is available? But we don't have access to the ResourceStore nor the ListStore

luca-rath: I think we only have access to the router

src/Sulu/Bundle/AdminBundle/Admin/View/TabViewChildBuilderTrait.php
niklasnatter: yes that sounds good too 🙂

src/Sulu/Bundle/AdminBundle/Resources/js/components/Badge/badge.scss
luca-rath: image
image

yes, it was red too

src/Sulu/Bundle/AdminBundle/Admin/View/Badge.php
```diff
@@ -0,0 +1,84 @@
...
+
+use JMS\Serializer\Annotation as Serializer;
+
+class Badge
+{
+    /**
+     * @var string
+     *
+     * @Serializer\Expose()
+     * @Serializer\Groups({"fullView"})
```
niklasnatter: we just used `* @Groups({"frontend"})` on the `ToolbarAction` and `ListItemAction`. not sure if it makes any difference. would you mind to investigate that and make things consistent?

src/Sulu/Bundle/AdminBundle/Resources/js/components/Badge/badge.scss
niklasnatter: oh did not know that this was superscripted 🙂 
looks good for me, but dont have a strong opinion about the color/styling - what do you think? @sulu/core-team

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Badge/stores/BadgeStore.js
```diff
@@ -0,0 +1,87 @@
...
+    ) {
+        this.router = router;
+        this.routeName = routeName;
+        this.dataPath = dataPath;
+        this.requestParameters = requestParameters;
+        this.routerAttributesToRequest = routerAttributesToRequest;
+
+        this.disposer = autorun(() => {
+            // Needed to tell autorun to listen on route changes
+            this.router.route;
```
niklasnatter: would use a `reaction` instead of `autorun` to make it explicit what triggers the load 🙂

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Badge/stores/BadgeStore.js
niklasnatter: the automation bundle uses a `FormOverlayList` right? i guess we could add an `attribute` to the `router` when the overlay is opened? then we could listen to the `attributes` to trigger this

imo, it would make sense to do this in the `FormOverlayList` anyway, because it would allow to send a link that opens the overlay for the correct entity. but would do that in a separate pull request 🙂

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Badge/stores/BadgeStore.js
```diff
@@ -0,0 +1,87 @@
...
+
+    @action setData(data: any) {
+        const {dataPath} = this;
+
+        let enhancedData = data;
+        if (dataPath) {
+            enhancedData = jsonPointer.get(data, dataPath);
+        }
+
+        this.value = enhancedData.toString();
```
niklasnatter: think this should be abit more robust

```suggestion
        this.value = String(enhancedData);
```

src/Sulu/Bundle/AdminBundle/Admin/View/Badge.php
```diff
@@ -0,0 +1,84 @@
...
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\AdminBundle\Admin\View;
+
+use JMS\Serializer\Annotation as Serializer;
+
+class Badge
```
niklasnatter, line 17: think the class is not tested

src/Sulu/Bundle/AdminBundle/Resources/js/components/Tabs/Tab.js
```diff
@@ -46,6 +50,11 @@ export default class Tab extends React.PureComponent<Props> {
                     title={children}
                 >
                     {children}
+                    {!!badges && !!badges.length &&
```
niklasnatter, line 53: think this is not tested

src/Sulu/Bundle/AdminBundle/Resources/js/views/Tabs/Tabs.js
```diff
@@ -134,8 +135,30 @@ class Tabs<T> extends React.Component<Props<T>> {
...
+                                        router={router}
+                                        routerAttributesToRequest={badge.routerAttributesToRequest}
+                                        visibleCondition={badge.visibleCondition}
+                                    />
+                                );
+                            });
+
                             return (
-                                <TabsComponent.Tab key={tabRoute.name}>
+                                <TabsComponent.Tab badges={badges} key={tabRoute.name}>
```
niklasnatter, line 161: think this is not tested

src/Sulu/Bundle/AdminBundle/Resources/js/views/Tabs/Tabs.js
```diff
@@ -134,8 +135,30 @@ class Tabs<T> extends React.Component<Props<T>> {
                     <TabsComponent onSelect={this.handleSelect} selectedIndex={selectedTabIndex}>
                         {this.sortedTabRoutes.map((tabRoute) => {
                             const tabTitle = tabRoute.options.tabTitle;
+                            const tabBadges = tabRoute.options.tabBadges || [];
+
+                            const badges = (Object.values(tabBadges): any).map((badge: BadgeOptions, index: number) => {
+                                if (typeof badge !== 'object') {
+                                    throw new Error(
+                                        'Badges need to be of type object. This should not happen and is likely a bug.'
```
niklasnatter: would make this consistent with the toolbar-actions

```suggestion
                                        'The value of a badge entry must be an object, but ' + typeof badge + ' was given!'
```

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Badge/stores/BadgeStore.js
luca-rath: The bigger problem is, that the badge isn't updated after deleting a task, because deleting is triggered by the `DeleteToolbarAction` on the list view, so the url won't change

src/Sulu/Bundle/AdminBundle/Admin/View/Badge.php
luca-rath: Doesn't make any difference, but I'll change it for consistency

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Badge/stores/BadgeStore.js
```diff
@@ -0,0 +1,122 @@
...
+        if (['POST', 'PUT', 'PATCH', 'DELETE'].includes(options.method.toUpperCase())) {
+            this.load();
+        }
+    };
+
+    initialize = () => {
+        this.load();
+
+        // Needed to tell autorun to listen on route changes
+        this.disposer = reaction(() => this.router.route, () => {
```
niklasnatter: would call this `routeChangeDisposer`

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Badge/stores/BadgeStore.js
```diff
@@ -0,0 +1,122 @@
...
+            this.load();
+        });
+
+        if (!Requester.handleResponseHooks.includes(this.responseHook)) {
+            Requester.handleResponseHooks.push(this.responseHook);
+        }
+    };
+
+    destroy = () => {
+        if (Requester.handleResponseHooks.includes(this.responseHook)) {
```
niklasnatter, line 115: think we should call `this.routeChangeDisposer` too 🙂

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Badge/stores/BadgeStore.js
```diff
@@ -0,0 +1,122 @@
...
+        if (response.url.includes('/admin/preview/')) {
+            return;
+        }
+
+        if (['POST', 'PUT', 'PATCH', 'DELETE'].includes(options.method.toUpperCase())) {
+            this.load();
+        }
+    };
+
+    initialize = () => {
```
niklasnatter: any reason for not doing this in the constructor? did we do it like this in other places too? 🙂

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Badge/stores/BadgeStore.js
```diff
@@ -0,0 +1,122 @@
...
+        }
+
+        this.value = String(enhancedData);
+    }
+
+    load = debounce(() => {
+        Requester.get(this.url).then((response: Object) => {
+            this.setData(response);
+        });
+    }, 3000, true);
```
niklasnatter, line 88: isnt this a bit long?

src/Sulu/Bundle/AdminBundle/Resources/js/services/Requester/Requester.js
```diff
@@ -99,9 +99,9 @@ function transformRequestData(data: Object | Array<Object>) {
     return transformRequestObject(data);
 }
 
-function handleResponse(response: Response): Promise<Object | Array<Object>> {
+function handleResponse(response: Response, options): Promise<Object | Array<Object>> {
```
niklasnatter:
```suggestion
function handleResponse(response: Response, options: ?Object): Promise<Object | Array<Object>> {
```

src/Sulu/Bundle/AdminBundle/Resources/js/services/Requester/Requester.js
```diff
@@ -122,8 +122,8 @@ function handleResponse(response: Response): Promise<Object | Array<Object>> {
     });
 }
 
-function handleObjectResponse(response: Response): Promise<Object> {
-    return handleResponse(response).then((response) => {
+function handleObjectResponse(response: Response, options): Promise<Object> {
```
niklasnatter:
```suggestion
function handleObjectResponse(response: Response, options: ?Object): Promise<Object> {
```

src/Sulu/Bundle/AdminBundle/Resources/js/components/Tabs/Tab.js
```diff
@@ -1,9 +1,11 @@
 // @flow
 import React from 'react';
+import type {Element} from 'react';
 import classNames from 'classnames';
 import tabStyles from './tab.scss';
 
 type Props = {
+    badges: ?(Element<*>[] | Element<*>),
```
niklasnatter: think allowing `Element<*>` instead of just allowing an array makes it harder to not render the container below 🙂

src/Sulu/Bundle/AdminBundle/Resources/js/components/Tabs/Tab.js
```diff
@@ -46,6 +50,11 @@ export default class Tab extends React.PureComponent<Props> {
                     title={children}
                 >
                     {children}
+                    {!!badges &&
+                        <div className={tabStyles.badges}>
```
niklasnatter, line 54: this will render an empty container if an empty array is passed right? 🤔

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Badge/stores/BadgeStore.js
luca-rath: I think you said you would register the `HandleResponseHook` in the `componentDidMount` ... Should i only keep that in the `initialize()` method and keep the rest in the `constructor()`

src/Sulu/Bundle/AdminBundle/Resources/js/views/Tabs/tests/__snapshots__/Tabs.test.js.snap
```diff
@@ -15,6 +15,9 @@ Array [
           title="tabTitle1"
         >
           tabTitle1
+          <div
```
niklasnatter: this will add a margin even if it is empty right?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Badge/stores/BadgeStore.js
luca-rath: The `true` at the end executes the request immediately and stops all following requests in the next 3 seconds, so I think this should be fine

src/Sulu/Bundle/AdminBundle/Admin/View/Badge.php
```diff
@@ -0,0 +1,85 @@
...
+{
+    /**
+     * @var string
+     *
+     * @Expose()
+     * @Groups({"fullView"})
+     */
+    private $routeName;
```
niklasnatter, line 25: would add getters for these properties just to be consistent 🙂

src/Sulu/Bundle/AdminBundle/Tests/Unit/Admin/View/BadgeTest.php
```diff
@@ -0,0 +1,34 @@
...
+use Sulu\Bundle\AdminBundle\Admin\View\Badge;
+
+class BadgeTest extends TestCase
+{
+    public function testAddRequestParameters(): void
+    {
+        $this->expectNotToPerformAssertions();
+
+        $badge = new Badge('app.notification_count', 'count', 'count != 0');
+        $badge->addRequestParameters(['foo' => 'bar', 'bar' => 'baz']);
```
niklasnatter, line 40: if we add getters, we can test this properly

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Badge/stores/BadgeStore.js
niklasnatter: okay 🙂

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Badge/tests/stores/BadgeStore.test.js
```diff
@@ -0,0 +1,187 @@
...
+
+    // Initial request
+    expect(Requester.get).toHaveBeenCalledTimes(1);
+    expect(Requester.get).toHaveBeenNthCalledWith(1, 'foo?entityId=5&locale=en&limit=0&entityClass=Foo');
+
+    // Should not perform request because of a "GET" request
+    const response1 = new Response();
+    response1.url = 'http://sulu.lo/admin/api/anything';
+    Requester.handleResponseHooks[0](
+        response1,
```
niklasnatter, line 153:
```suggestion
        (response1: any),
```

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Badge/tests/stores/BadgeStore.test.js
```diff
@@ -0,0 +1,187 @@
...
+    badgeStore.load();
+
+    expect(Requester.get).toBeCalledWith('foo?entityId=5&locale=en&limit=0&entityClass=Foo');
+
+    return promise.then(() => {
+        expect(badgeStore.value).toEqual('hello');
+    });
+});
+
+test('Should load data on response hook callback', () => {
```
niklasnatter, line 120: test for route change is missing. we need to use `extendObservable` to test the reaction 🙈

src/Sulu/Bundle/AdminBundle/Resources/js/package.json
```diff
@@ -45,7 +45,8 @@
     },
     "devDependencies": {
         "enzyme": "^3.3.0",
-        "moment-timezone": "^0.5.14"
+        "moment-timezone": "^0.5.14",
+        "whatwg-fetch": "^3.5.0"
```
niklasnatter: would prefer not to install this 🙂

package.json
```diff
@@ -88,7 +88,8 @@
         "webpack": "^4.20.2",
         "webpack-clean-obsolete-chunks": "^0.4.0",
         "webpack-cli": "^3.1.1",
-        "webpack-manifest-plugin": "^2.0.2"
+        "webpack-manifest-plugin": "^2.0.2",
+        "whatwg-fetch": "^3.5.0"
```
niklasnatter: would prefer not to install this 🙂

src/Sulu/Bundle/AdminBundle/Resources/js/services/Requester/tests/Requester.test.js
```diff
@@ -17,6 +17,7 @@ test('Should execute GET request and reject with response when the response cont
     expect(window.fetch).toBeCalledWith('/some-url', {
```
niklasnatter, line 17: can we add a test to check if the hooks are called? looks like this is not tested at the moment

src/Sulu/Bundle/AdminBundle/Resources/js/views/Tabs/tests/Tabs.test.js
```diff
@@ -57,6 +57,59 @@ test('Should render the children after the tabs', () => {
...
+    Router.mockImplementation(function() {
+        this.attributes = attributes;
+        this.redirect = jest.fn();
+        this.route = route;
+    });
+
+    const router = new Router({});
+
+    const Child = () => (<h1>Child</h1>);
+    expect(render(<Tabs route={route} router={router}>{() => (<Child />)}</Tabs>)).toMatchSnapshot();
```
niklasnatter: is it possible that this does not render any badge? 🤔

niklasnatter (approved):

Issue #3283, opened 2017-03-27
[2.0] Media Bundle Refractoring


alexander-schranz:
| Q | A
| --- | ---
| Bug? | no
| New Feature? | yes
| Sulu Version | 2.0

#### Actual Behavior

Currently its hard to use the media in import scripts or on a submit forms.

#### Expected Behavior/Possible Solutions

 - [ ] A. Media Manager should not handle Storage and Formats (remove Storage and FormatManager from MediaManager)
 - [ ] B. Loading of the media entity should be done directly over Repository and not by Media Manager 
 - [ ] C. Remove media api object
   - [ ] C1. Image formats should be generated by symfony route or twig extension (FormatManager)
   - [ ] C2. Download url should be generated by symfony route or twig extension
 - [ ] D. Move security checks from media repository to controller
 - [ ] E. Add chain pool for media property providers (e.g. Video(duration, resolution, ...))
 - [ ] F. Add Image property provider to add height, width as properties
 - [ ] G. Move Video property providers into an own bundle to make php-ffmpeg optional
 - [ ] H. Add Media image extractor pool to allow add of new image extractor (needed for video bundle)

#### Flowcharts

TODO: coming soon

danrot:
@alexander-schranz At least some of these things have already been done, right? How up to date is that issue? Does it make sense to keep it open?

danrot:
@alexander-schranz ping

abdellahrk:
Hello @alexander-schranz, any update on this issue?

mamazu:
I think this should be reevaluated what we can impement in Sulu 3 without any BC breaks.

abdellahrk:
> I think this should be reevaluated what we can impement in Sulu 3 without any BC breaks.

It makes sense.

mamazu:
What exactly would you like to see from this list?

abdellahrk:
I was asking if there's anything not done on the list that I can send a PR for it.

mamazu:
I think J would be nice to have although it's probably not that easy.

alexander-schranz:
@mamazu I think that we will keep the collection based security and not add more security to media itself. There might be plans in future a way to make media entities `private` but not like Security, more like the published / unpublished media.

I even think more about currently closing the whole issue. /cc @Prokyonn what do you think?

mamazu:
I'd also vote for closing this and the related issue linked. They're really outdated and not very helpful.

Issue #1659, opened 2015-10-06
Allow register services by mime type for media properties and thumbnail generation


alexander-schranz:
The video integration e.g. #1653 is hardcoded into the media manager.
Also the pdf and pds to image converter is directly in the FormatManager. 
I think the best solution would be to handle this over service tags.

So you register the services over tags: 

``` xml
<tag name="thumbnail_generator" type="application/pdf" />
```

``` xml
<tag name="media_properties_reader" type="video/*" />
```

Advantages:
 - easier to add new thumbnail generations for other mime types
 - fewer dependencies on the media and format manager
 - no need for the preview mime types parameter if no service is registered for a mime type it can't be generated

wachterjohannes:
sounds reasonable (: :+1

PR #5766, opened 2021-01-20
Add extension point for providing additional properties for a media

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? |  yes
| BC breaks? | no
| Deprecations? | yes 
| Fixed tickets |  part of #1659 and part of #3283
| Related issues/PRs |
| License | MIT
| Documentation PR | https://github.com/sulu/sulu-docs/pull/620

#### What's in this PR?

Add media properties provider.

#### Why?

Make it easier for developers add properties to media.

#### To Do

- [ ] Regenerate Command


alexander-schranz:
@nnatter not sure about the interface naming. Can you have a look at them and give a feedback?

src/Sulu/Bundle/MediaBundle/Media/PropertiesProvider/MediaPropertiesProvider.php
```diff
@@ -0,0 +1,45 @@
...
+     * @var iterable<PropertiesProviderInterface>
+     */
+    private $providers;
+
+    public function __construct(iterable $providers)
+    {
+        $this->providers = $providers;
+    }
+
+    public function provide(File $file): array
```
niklasnatter: not sure if we need this service. we could just inject all providers into the `MediaManager` and do this there. that way, we would not need two interfaces

src/Sulu/Bundle/MediaBundle/Media/PropertiesProvider/PropertiesProviderInterface.php
```diff
@@ -0,0 +1,24 @@
...
+use Symfony\Component\HttpFoundation\File\File;
+
+interface PropertiesProviderInterface
+{
+    /**
+     * @return array<string, mixed>
+     */
+    public function provide(File $file): array;
+
+    public static function supports(File $file): bool;
```
niklasnatter: do you see an advantage of using a `supports` method? the provider could just return an empty array if he cannot analyze the file right?

niklasnatter (commented):
not sure if i would add a `PropertiesProviderInterface` and a second `MediaPropertiesProviderInterface` 🤔

src/Sulu/Bundle/MediaBundle/Media/PropertiesProvider/MediaPropertiesProvider.php
alexander-schranz: @nnatter I don't like it that the user of the service need to handle the `array_merge`. I would really like a service which also can be used in imports which and return all properties.

src/Sulu/Bundle/MediaBundle/Media/PropertiesProvider/PropertiesProviderInterface.php
alexander-schranz: @nnatter the concept is that it does not initialize the provider service until the service does support it, thats why the supports is static here. But did not yet test it if the lazy mechanic works as expected and does not initialize the ffmpeg service when an image is uploaded.

src/Sulu/Bundle/MediaBundle/Media/PropertiesProvider/MediaPropertiesProvider.php
alexander-schranz: Orientated here by the `ContentMergerInterface` and `MergerInterface` from the ContentBundle. So `MediaPropertiesProviderInterface` and `PropertiesProviderInterface` 🙈 .

src/Sulu/Bundle/MediaBundle/Resources/config/ffmpeg.xml
```diff
@@ -23,5 +23,11 @@
             </argument>
             <argument id="logger" type="service" on-invalid="null"/>
         </service>
+
+        <service id="sulu_media.video_properties_provider" class="Sulu\Bundle\MediaBundle\Media\PropertiesProvider\VideoPropertiesProvider">
+            <argument type="service" id="sulu_media.ffprobe"/>
+
+            <tag name="sulu_media.media_properties_provider"/>
```
niklasnatter, line 29: think i would call this just `properties_provider` because the interface is called `PropertiesProviderInterface `

```suggestion
            <tag name="sulu_media.properties_provider"/>
```

src/Sulu/Bundle/MediaBundle/Media/PropertiesProvider/MediaPropertiesProvider.php
niklasnatter: do you think the user will need this service? 
using two interfaces is okay for me too, but then we should start to use this pattern everywhere 🙂 
for example, we did not use two interfaces for the `BlockVisitor` concept

src/Sulu/Bundle/MediaBundle/Media/PropertiesProvider/PropertiesProviderInterface.php
niklasnatter: okay - not sure if the optimization is worth the more complex interface. but dont have a too strong opinion 🙂

src/Sulu/Bundle/MediaBundle/Resources/config/ffmpeg.xml
alexander-schranz: I did use the `MediaPropertiesProviderInterface` now so I think I would keep here `sulu_media.media_properties_provider`.

src/Sulu/Bundle/MediaBundle/Media/Exception/CollectionNotFoundException.php
```diff
@@ -16,8 +16,12 @@ class CollectionNotFoundException extends MediaException
     /**
      * @param string $id
      */
-    public function __construct($id)
+    public function __construct($id, \Throwable $previous = null)
```
niklasnatter, line 19: its also okay for me to just update the baseline for such errors in the future

src/Sulu/Bundle/MediaBundle/Resources/config/ffmpeg.xml
niklasnatter: sounds good 🙂

niklasnatter (approved):

Issue #5394, opened 2020-07-06
Preview image for PDF with dots in the filename is regenerated on each pageload of media-list


kjueni:
| Q | A
| --- | ---
| Sulu Version | 2.1.0
| Browser Version | -

# Actual Behavior

When a file is uploaded and it has dots (.) in its filename (e.g. my.cool.file.pdf), the filename for the preview image results in my.cool.jpg (Media list). Because it's saved correctly (my.cool.file.jpg) on the filesystem, the preview-file is not found when the media list is rendered and the preview is always regenerated.

# Expected Behavior

The filename of the preview urls should be my.cool.file.jpg

# Steps to Reproduce

Upload a pdf with dots in its filename, e.g. my.cool.file.pdf and check the media list where this file should be displayed --> preview is generated on each load.

# Possible Solutions
As far as I could reproduce, pathinfo is called two times in the process, once at the beginning (filename becomes my.cool.file) and once before getMediaUrl is called in the replaceExtension method (filename becomes my.cool)
```
public function getFormats($id, $fileName, $version, $subVersion, $mimeType)
{
        // filename before pathinfo: my.cool.file.pdf
        $fileName = \pathinfo($fileName)['filename'];
        // filename after pathinfo: my.cool.file
        $formats = [];

        $extensions = $this->converter->getSupportedOutputImageFormats($mimeType);

        if (empty($extensions)) {
            return [];
        }

        $originalExtension = $extensions[0];
        foreach ($this->formats as $format) {
            foreach ($extensions as $extension) {
                $formatUrl = $this->formatCache->getMediaUrl(
                    $id,
                    // Because there is another pathinfo call in replaceExtension, filename becomes my.cool and then my.cool.jpg
                    $this->replaceExtension($fileName, $extension),
                    $format['key'],
                    $version,
                    $subVersion
                );

                if ($extension === $originalExtension) {
                    $formats[$format['key']] = $formatUrl;
                }

                $formats[$format['key'] . '.' . $extension] = $formatUrl;
            }
        }

        return $formats;
}
```

niklasnatter:
Thanks a lot for reporting! This should hopefully be fixed in #5768

PR #5768, opened 2021-01-21
Fix format manager generated urls with multiple dots

binarious:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no 
| Deprecations? | no 
| Fixed tickets | 
| Related issues/PRs | 
| License | MIT
| Documentation PR | 

#### What's in this PR?

This PR removes the call to `replaceExtension` in the `FormatManager` of the MediaBundle, because it's not useful in the specific case.

#### Why?

`$fileName` is already without an extension, because it's read via `pathinfo`'s `filename` attribute, which doesn't include an extension. Calling `replaceExtension` removes normal filename parts containing dots. This lead to a wrong requested filename, when the original file had dots in the filename part (like `myfile.2000x1000.jpg`). 

Before this patch, `image.thumbnails['myname']` resulted in a filename without `.2000x1000`, which did render the image, but the file is always generated by sulu, because in `public/uploads/media` the file is saved *with* `.2000x1000` and therefore not directly served via the webserver.

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/Media/FormatManager/FormatManager.php
+++ b/src/Sulu/Bundle/MediaBundle/Media/FormatManager/FormatManager.php
@@ -176,8 +176,6 @@ class FormatManager implements FormatManagerInterface
 
     public function getFormats($id, $fileName, $version, $subVersion, $mimeType)
     {
-        $fileName = \pathinfo($fileName)['filename'];
-
         $formats = [];
 
         $extensions = $this->converter->getSupportedOutputImageFormats($mimeType);
@@ -191,7 +189,7 @@ class FormatManager implements FormatManagerInterface
             foreach ($extensions as $extension) {
                 $formatUrl = $this->formatCache->getMediaUrl(
                     $id,
-                    "$fileName.$extension",
+                    $this->replaceExtension($fileName, $extension),
                     $format['key'],
                     $version,
                     $subVersion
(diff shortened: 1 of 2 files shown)


niklasnatter:
Thanks a lot for the PR and the great explanation! I think this will fix #5394 🎉 
@alexander-schranz Do we need to rebase this onto `1.6`?

alexander-schranz:
@binarious thx for the PR. This looks like a bug which also exists in the [2.1 Version  of the FormatManager](https://github.com/sulu/sulu/blob/2.1/src/Sulu/Bundle/MediaBundle/Media/FormatManager/FormatManager.php#L194) can you rebase your PR to 2.1?
It would also be great if you could create a Unit Test for it, if not I can provide some.

binarious:
@alexander-schranz I've added a test that fails with the old solution and passes with my fix. Hope that fits with your requirements. About rebasing: I'm getting a lot of unrelated conflicts which I'm not suited to solve.

plozmun:
What do you think about removing this line instead to keep same behaviour in all functions?https://github.com/sulu/sulu/blob/2.x/src/Sulu/Bundle/MediaBundle/Media/FormatManager/FormatManager.php#L179

alexander-schranz:
@binarious I will rebase it for you. Just for you, when you rebase to a lower version you should do a interactive rebase and use only the commits which are yours of the PR. To make this easier you can use `--onto` on the rebase so if you do:

e.g.:

```bash
git fetch origin # fetch sulu remote
git rebase --interactive --onto origin/2.1 origin/2.x
```

It should then just show the 2 commits of your branch and have no conflicts in most cases or easy to solve conflicts.

alexander-schranz:
@plozmun good input 👍

alexander-schranz (approved):

niklasnatter (approved):

alexander-schranz:
@binarious @plozmun Thank you! This will be in the next release.

PR #5772, opened 2021-01-22
Fix buggy scroll behaviour

### LINKED ISSUE #5765: Sulu auto scrolls to bottom of page
| Q | A
| --- | ---
| Sulu Version | 2.2.3 / 2.1.7
| PHP Version | 7.4.12
| Browser Version | Chrome 88.0.4324.96 (Official Build) (x86_64)

# Actual Behavior

While using longer lists and increasing the page limit to 100 for example, relevant results are showing while scrolling which is great. But when another tab in browser is openend and you return to Sulu tab it seems Sulu always scrolls down to the bottom of the page. Instead of staying at the top or same position the current user left it in.

The same for longer forms, after returning to the Sulu tab it's pointing at the bottom of the page.

This also seems to be the case in demo project, see [screencast](https://we.tl/t-CxfZWotKh6).

screencast_sulu_scroll

# Expected Behavior

Stay in same place where current user left it.

# Steps to Reproduce

Create a long list or long form for which you have to scroll. Stay on top of page and open new tab, return to previous tab and you wil notice Sulu scrolled to bottom of page incorrectly.

### LINKED ISSUE #5769: Scroll position changes when the browser window loses focus
| Q | A
| --- | ---
| Sulu Version | 2.2.3
| PHP Version | 7.4.11
| Browser Version | Tested in _Chrome 87.0.4280.141_ & _Firefox 84.0.1_

_**Edit**: probably a duplicate of https://github.com/sulu/sulu/issues/5765_

# Actual Behavior

When you are editing content and move focus to another window or tab (e.g. to copy content that you'd like to paste inside the field you're editing) the scroll-position in Sulu changes. So when you come back to paste your copied content, you have to scroll back to the block you were editing.

Issue

# Expected Behavior

You'd expect to scroll-position to not change when leaving the window.

# Steps to Reproduce

Add a few blocks (at least 3?) on a Sulu page. Then leave the window (or focus on another one).
The scroll position will have moved up.

_**Edit**: the quickest way to reproduce might be adding a new page on https://sulu.rocks/admin, adding 4 or 5 blocks and leaving the window._ 

Template I used:
```
<?xml version="1.0" ?>

    <key>test</key>

    <view>pages/test</view>
    <controller>Sulu\Bundle\WebsiteBundle\Controller\DefaultController::indexAction</controller>
    <cacheLifetime>604800</cacheLifetime>

    <meta>
        <title lang="en">Test</title>
    </meta>

    <properties>
            <meta>
            </meta>
            <params>
            </params>

            <tag name="sulu.rlp.part"/>
        </property>

        <property name="url" type="resource_locator" mandatory="true">
            <meta>
            </meta>

            <tag name="sulu.rlp"/>
        </property>

        <section name="content">
            <meta>
            </meta>

            <properties>
                <block name="blocks" default-type="title" minOccurs="0">
                    <!-- @see https://docs.sulu.io/en/2.1/reference/content-types/block.html -->

                    <meta>
                        <title lang="en">Blocks</title>
                    </meta>

                    <types>
                        <type name="form">
                            <meta>
                                <title lang="en">Form</title>
                            </meta>

                            <properties>
                                <property name="form" type="single_form_selection" mandatory="true">
                                    <meta>
                                        <title lang="en">Form</title>
                                    </meta>

                                    <params>
                                        <param name="resourceKey" value="page"/>
                                    </params>
                                </property>

                                <property name="contact-info" type="checkbox">
                                    <meta>
                                        <title lang="en">Show contact info</title>
                                    </meta>
                                    <params>
                                        <param name="type" value="toggler"/>
                                        <param name="default_value" value="false"/>
                                    </params>
                                </property>
                            </properties>
                        </type>

                        <type name="title">
                            <meta>
                            </meta>

                            <properties>
                                    <meta>
                                    </meta>
                                </property>

                                <property name="size" type="single_select">
                                    <meta>
                                        <title lang="en">Size</title>
                                    </meta>

                                    <params>
                                        <param name="default_value" value="h2"/>

                                            <param name="h1">
                                                <meta>
                                                    <title>H1</title>
                                                </meta>
                                            </param>
                                            <param name="h2">
                                                <meta>
                                                    <title>H2</title>
                                                </meta>
                                            </param>
                                            <param name="h3">
                                                <meta>
                                                    <title>H3</title>
                                                </meta>
                                            </param>
                                        </param>
                                    </params>
                                </property>
                            </properties>
                        </type>
                    </types>
                </block>
            </properties>
        </section>
    </properties>
</template>
```

# Extra info

I made the above recording in Firefox, a browser which I rarely use and which has no plugins installed. But the same issue happens in Chrome.
I've also tried disabling our custom javascript (in the frontend) to make sure that wasn't interfering.

luca-rath:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #5765, fixes #5769
| Related issues/PRs | #5567
| License | MIT
| Documentation PR | -

#### What's in this PR?

Fix buggy scroll behaviour

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Popover/tests/Popover.test.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Popover/tests/Popover.test.js
@@ -102,8 +102,10 @@ test('The popover should request to be closed when the backdrop is clicked', () 
 });
 
 test('The popover should not request to be closed if it is already closed', () => {
+    const windowListeners = {};
+    window.addEventListener = jest.fn((event, cb) => windowListeners[event] = cb);
     const onCloseSpy = jest.fn();
-    const popover = shallow(
+    mount(
         <Popover
             anchorElement={getMockedAnchorEl()}
             onClose={onCloseSpy}
@@ -118,7 +120,8 @@ test('The popover should not request to be closed if it is already closed', () =
             }
         </Popover>
     );
-    popover.find('Backdrop').simulate('click');
+    expect(windowListeners.blur).toBeDefined();
+    windowListeners.blur();
     expect(onCloseSpy).not.toBeCalled();
 });
 
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Popover/tests/Popover.test.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Popover/tests/Popover.test.js
@@ -101,7 +101,7 @@ test('The popover should request to be closed when the backdrop is clicked', () 
     expect(onCloseSpy).toBeCalled();
 });
 
-test('The popover should not request to be closed if it\' already closed', () => {
+test('The popover should not request to be closed if it is already closed', () => {
     const onCloseSpy = jest.fn();
     const popover = shallow(
         <Popover
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Popover/tests/Popover.test.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Popover/tests/Popover.test.js
@@ -101,6 +101,27 @@ test('The popover should request to be closed when the backdrop is clicked', () 
     expect(onCloseSpy).toBeCalled();
 });
 
+test('The popover should not request to be closed if it\' already closed', () => {
+    const onCloseSpy = jest.fn();
+    const popover = shallow(
+        <Popover
+            anchorElement={getMockedAnchorEl()}
+            onClose={onCloseSpy}
+            open={false}
+        >
+            {
+                (setPopoverRef, styles) => (
+                    <div ref={setPopoverRef} style={styles}>
+                        <div>My item 1</div>
+                    </div>
+                )
+            }
+        </Popover>
+    );
+    popover.find('Backdrop').simulate('click');
+    expect(onCloseSpy).not.toBeCalled();
+});
+
 test('The popover should request to be closed when the window is blurred', () => {
     const windowListeners = {};
     window.addEventListener = jest.fn((event, cb) => windowListeners[event] = cb);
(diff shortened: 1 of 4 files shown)


sanhofman:
I can confirm this fixes https://github.com/sulu/sulu/issues/5765.

Thanks in advance @luca-rath! 👍

niklasnatter:
Nice, this fixes #5765 and #5769 for me too 🙂 

As far as i understand, the root cause is that the `Popover` component calls its `onClose` prop everytime the window is blurred or resized. Therefore I would prefer solving this in the `Popover` component.

luca-rath:
> Nice, this fixes #5765 and #5769 for me too 🙂
> 
> As far as i understand, the root cause is that the `Popover` component calls its `onClose` prop everytime the window is blurred or resized. Therefore I would prefer solving this in the `Popover` component.

I didn't want to solve it like that to minimize possible side effects. But if you prefer that, that's also okay for me

niklasnatter:
Yes, I would prefer that. Looks like we are unnecessarily executing a quite big number of callbacks on resize/blur at the moment 🙈  

If a component depends on the current behaviour of executing the `onClose` callback even if the `Popover` is already closed, we should adjust the component. But hopefully there is no such component 🙂

luca-rath:
Fixed it, but I kept the previous changes, because they still make sense.

src/Sulu/Bundle/AdminBundle/Resources/js/components/Popover/tests/Popover.test.js
```diff
@@ -101,6 +101,27 @@ test('The popover should request to be closed when the backdrop is clicked', ()
     expect(onCloseSpy).toBeCalled();
 });
 
+test('The popover should not request to be closed if it\' already closed', () => {
```
niklasnatter:
```suggestion
test('The popover should not request to be closed if it is already closed', () => {
```

luca-rath: oops 🙈

niklasnatter (approved):

PR #5773, opened 2021-01-22
Fix original media download link in the admin UI

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixe
| Related issues/PRs |
| License | MIT

#### What's in this PR?

The `adminUrl` generation was missing in `MediaListRepresentationFactory` for images. Thus downloading the original image in the admin UI was not possible due to an `undefined` url.

#### Why?

Downloading the original image was not possible in the admin UI.

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/Media/ListRepresentationFactory/MediaListRepresentationFactory.php
+++ b/src/Sulu/Bundle/MediaBundle/Media/ListRepresentationFactory/MediaListRepresentationFactory.php
@@ -63,6 +63,12 @@ class MediaListRepresentationFactory
                 $listResponse[$i]['version']
             );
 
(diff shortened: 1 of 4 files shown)


alexander-schranz (changes requested):
@Prokyonn looks for me in the MediaCartAdapter also the fallback to the `.url` is missing we did normally in all cases fallback to `.url` if `.adminUrl` was not given by the api: https://github.com/sulu/sulu/pull/5437/files#diff-c31ab70d434dc1866c5c8b40a59aa2168f31ca4e5e733d87e29062ae51ce2987

PR #5778, opened 2021-01-28
Adjust wording of info texts about soft-limits in SEO tab

niklasnatter:


src/Sulu/Bundle/PageBundle/Resources/translations/admin.en.json
```diff
@@ -35,11 +35,11 @@
     "sulu_page.edited": "edited",
     "sulu_page.search_result": "Search Engine Result Pages Preview (SERP)",
     "sulu_page.meta_title": "Title (Meta title)",
-    "sulu_page.meta_title_info_text": "The meta title is limited to 55 characters.",
+    "sulu_page.meta_title_info_text": "The meta title is recommended to use 55 characters at most.",
```
niklasnatter: the limits on the tab are not hard-limits but recommendations. we should make that clear in the respective info texts 🙂

luca-rath:
```suggestion
    "sulu_page.meta_title_info_text": "The meta title should be no longer than 55 characters.",
```

src/Sulu/Bundle/PageBundle/Resources/translations/admin.en.json
```diff
@@ -35,11 +35,11 @@
...
     "sulu_page.meta_description": "Description (Meta description)",
-    "sulu_page.meta_description_info_text": "The meta description is limited to 320 characters.",
+    "sulu_page.meta_description_info_text": "The meta description is recommended to use 320 characters at most.",
```
luca-rath:
```suggestion
    "sulu_page.meta_description_info_text": "The meta description should be no longer than 320 characters.",
```

src/Sulu/Bundle/PageBundle/Resources/translations/admin.en.json
Prokyonn: Personally I would recommend to mention the seo guidelines in the info-text, something like `Seo guidelines recommend a title of 55 characters maximum.` So that it is clear that this is not only a recommendation from our side.

luca-rath:
```suggestion
    "sulu_page.meta_title_info_text": "According to the seo guidelines, the meta title should be no longer than 55 characters.",
```

src/Sulu/Bundle/PageBundle/Resources/translations/admin.en.json
luca-rath:
```suggestion
    "sulu_page.meta_description_info_text": "According to the seo guidelines, the meta description should be no longer than 320 characters.",
```

src/Sulu/Bundle/PageBundle/Resources/translations/admin.en.json
alexander-schranz: I think I would go with the word `recommended` here, other things sounds more like a hard limit from my side.

niklasnatter: pushed a new proposal - what do you think? 🙂

luca-rath (approved):

src/Sulu/Bundle/PageBundle/Resources/translations/admin.en.json
luca-rath: looks good to me

PR #5782, opened 2021-01-31
Add marker interface for DocumentManager event subscribers to enable autoconfiguration

luca-rath:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

Add marker interface for DocumentManager event subscribers

#### Why?

To enable autoconfiguration for DocumentManager event subscribers

### DIFF
--- a/src/Sulu/Bundle/DocumentManagerBundle/DependencyInjection/SuluDocumentManagerExtension.php
+++ b/src/Sulu/Bundle/DocumentManagerBundle/DependencyInjection/SuluDocumentManagerExtension.php
@@ -13,6 +13,7 @@ namespace Sulu\Bundle\DocumentManagerBundle\DependencyInjection;
 
 use Sulu\Bundle\DocumentManagerBundle\DataFixtures\DocumentFixtureInterface;
 use Sulu\Bundle\DocumentManagerBundle\Session\Session;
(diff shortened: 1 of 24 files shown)


luca-rath:
ping @sulu/core-team what do you think about this change?

alexander-schranz:
@luca-rath I personally would prefer change to use the symfony event dispatcher, but that would be more work. If we really want to implement a marker Interface, I would prefer not using the MarkerInterface internal for service which are not autowired, because the normal EventSubscriberInterface should still work the same way.

niklasnatter:
Havent really used this yet. Therefore I dont have a strong opinion 😇

luca-rath:
I'm totally okay with using symfony's `EventSubscriberInterface` for internal services, but adding a marker interface definitely doesn't hurt. But if we decide on using symfony's `EventDispatcher`, that's fine for me too. I justspent 10 minutes figuring out why my event subscriber wasn't doing it's work, so we should at least do something 😄

alexander-schranz:
Would just use still use the `EventSubscriberInterface` for our services and only create the interface and autoconfiguration in this PR and let the other services untouched.

luca-rath:
ping @alexander-schranz

PR #5783, opened 2021-01-31
* PHP 8.x Fix deprecation notices for optional function parameters declared before required parameter.

StaffNowa:
…clared before required parameter.

| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | yes
| Fixed tickets |
| Related issues/PRs | 
| License | MIT
| Documentation PR |

#### What's in this PR?

PHP Deprecated:  Required parameter $roleRepository follows optional parameter $descendantProviders in /home/domain/domains/domain_name.com/vendor/sulu/sulu/src/Sulu/Component/Security/Authorization/AccessControl/AccessControlManager.php on line 61

#### Why?

Fix deprecation notices for optional function parameters declared before required parameter.

### DIFF
--- a/.github/workflows/test-application.yaml
+++ b/.github/workflows/test-application.yaml
@@ -62,7 +62,7 @@ jobs:
               run: ./${{ steps.php-cs-fixer-path.outputs.path }} fix --dry-run --diff
 
     js-css:
-        name: Test JavaScript application
+        name: "Test JavaScript application with Node ${{ matrix.node-version }}"
         runs-on: ubuntu-latest
 
         strategy:


src/Sulu/Component/Security/Authorization/AccessControl/AccessControlManager.php
```diff
@@ -62,8 +62,8 @@ public function __construct(
         MaskConverterInterface $maskConverter,
         EventDispatcherInterface $eventDispatcher,
         SystemStoreInterface $systemStore,
-        iterable $descendantProviders = [],
-        RoleRepositoryInterface $roleRepository
+        RoleRepositoryInterface $roleRepository,
+        iterable $descendantProviders = []
```
alexander-schranz: we should not swap any parameter in the constructor. In this case I would just remove the `= []` as it doe snot make any sense

src/Sulu/Bundle/SecurityBundle/Resources/config/services.xml
```diff
@@ -121,8 +121,8 @@
             <argument type="service" id="sulu_security.mask_converter"/>
             <argument type="service" id="event_dispatcher"/>
             <argument type="service" id="sulu_security.system_store"/>
-            <argument type="tagged_iterator" tag="sulu_security.access_control_descendant_provider"/>
             <argument type="service" id="sulu.repository.role"/>
+            <argument type="tagged_iterator" tag="sulu_security.access_control_descendant_provider"/>
```
alexander-schranz: we should not swap this parameters

alexander-schranz (changes requested):
Thx for the fix. We will provide PHP 8 support for 2.1.x branch. Can you adjust the change request and rebasse your branch to 2.1?

src/Sulu/Component/Security/Authorization/AccessControl/AccessControlManager.php
StaffNowa: 2.1.x not have these changes at all https://github.com/StaffNowa/sulu/blob/release/2.1/src/Sulu/Component/Security/Authorization/AccessControl/AccessControlManager.php

StaffNowa:
something broken in this part https://github.com/sulu/sulu/pull/5789

Issue #5596, opened 2020-11-04
Check of Email uniquness in UserManager::processEmail should be case insensitiv


rs2487:
| Q | A
| --- | ---
| Sulu Version | 2.1.1

# Actual Behavior

Sulu throws an exception when updating a user which email address only changed by different upper or lower case of character. 

# Expected Behavior

The comparison should be case insensitive. 

# Possible Solutions

Use strcasecmp here: https://github.com/sulu/sulu/blob/master/src/Sulu/Bundle/SecurityBundle/UserManager/UserManager.php#L684

nnatter:
Hey, thanks for reporting. This sounds like a useful thing to do 🙂 Would you like to create a pull request?

luca-rath:
Fixed by #6170, will be released as 2.2.12 and 2.3.2

PR #5797, opened 2021-02-05
Display helpful error message if uploaded media is blocked or exceeds maximum filesize

niklasnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Fixed tickets | fixes #5417
| License | MIT

#### What's in this PR?

This PR adjusts the `MediaOverview` view to display helpful error messages if the upload of a file fails.

Builds upon https://github.com/sulu/sulu/pull/5799

Issue #5312, opened 2020-05-31
Allow to display custom validation and error messages in Form


d3v4nt:
# Problem description
I'm extending the admin area with custom entities and having issues setting unique fields. The unique check works but it's completely hidden for the user.
On the screenshoots I'm trying to save item with already used title.

![entity](https://user-images.githubusercontent.com/33395021/83356558-456dbc80-a367-11ea-8442-b6be30b5c16c.png)
![admin](https://user-images.githubusercontent.com/33395021/83356551-40a90880-a367-11ea-858f-99d7279ee7f7.png)

# Proposed solution
Provide a possibility of using custom error messages or at least marking wrong fields.

alexander-schranz:
Thank you for creating this proposal 👍 .

We definitely should add a way to add an error message from the backend. I think we should implement something like the [rfc7807](https://tools.ietf.org/html/rfc7807). This is also supported by the FOSRestBundle in newer version (2.8 / 3.0). The json format is very flexible as the rfc allows to add custom fields to the response json.
The rfc has an own mimetype `application/problem+json` but this is all supported by FOSRest we just would need to implement the JS part that the message is displayed.

antonstsk:
Wouldn't it be great to create a project in the sulu repository for people like me that don't know much about Sulu but still can work with Symfony + React and can help on so necessary functionality 😅

antonstsk:
I wish I can help... If someone that knows Sulu feels like can help on planning the job needed to have that functionality please be my guest! I can definitely invest some time on this!

alexander-schranz:
@antonstsk 

> Wouldn't it be great to create a project in the sulu repository for people like me that don't know much about Sulu but still can work with Symfony + React and can help on so necessary functionality 😅

Nobody directly works in the sulu repository also every core developer has there own sulu fork. So you click in this repository on the right top on *Fork*  then *clone* your fork, there you *create a new branch* and then you *push it* into your fork and *create PR* to the main sulu repository. Or what kind of project are you searching for?

> I wish I can help... If someone that knows Sulu feels like can help on planning the job needed to have that functionality please be my guest! I can definitely invest some time on this!

I think in this case the best one which is most familiar how this could be implemented in the JS is @danrot.

alexander-schranz:
Some additional information about the RFC7807 how it looks like and how we could use its data:

Example Full RFC7807 response as json:

```json
{
    "type": "about:blank",
    "title": "Title",
    "detail": "More Detail",
[...]

Issue #5336, opened 2020-06-08
Api not return error messages on prod


alexander-schranz:
| Q | A
| --- | ---
| Sulu Version | effects currently all versions
| Browser Version | -

# Actual Behavior

Api not return error messages on production.

# Expected Behavior

The api should return the error messages on production.

# Steps to Reproduce

1. Disable debug

```env
APP_DEBUG=0
```

2. Login on sulu

3. Visit e.g.: https://sulu.rocks/admin/api/media/1

# Possible Solutions

That under prod/stage the correct error message e.g.: [MissingArgumentException](https://github.com/sulu/sulu/blob/release/2.1/src/Sulu/Component/Rest/Exception/MissingArgumentException.php) we would need to add this exception to the fos_rest messages mapping, else fos rest will just return the status code text [see logic here](https://github.com/FriendsOfSymfony/FOSRestBundle/blob/3d8501dbdfa48811ef942f5f93c358c80d5ad8eb/Serializer/Normalizer/AbstractExceptionNormalizer.php#L54-L60).

danrot:
Wondering why they have chosen to implement it this way... I could imagine that they consider this a possible security vulnerability. Especially in your example the following string would be shown to the requester:

```
Missing parameter locale in Sulu\\Bundle\\MediaBundle\\Controller\\MediaController
```

This would allow to infer that Sulu and therefore PHP is used, delivering valuable information to potential attackers. This is only one example, I could easily imagine, that other exceptions would even contain more information about the underlying system.

Actually the person calling this API would probably not even really be interested in what controller was called, but only what they were doing wrong. Therefore I think that in the above case a message like this would be sufficient:

```
Missing parameter locale
```

Furthermore I would suggest checking all subclasses of `RestException`, make sure their messages don't include any sensitive information, and add them to the `fos_rest.exception.messages` configuration.

niklasnatter:
As described in https://github.com/sulu/sulu/issues/5312#issuecomment-777291327, implementing the `TranslationErrorMessageExceptionInterface` will add a translated error message to the response. Every other exception will lead to a generic error message such as `"message": "Bad Request"`. This is done by the `FlattenExceptionNormalizer` of the FOSRestBundle:
https://github.com/FriendsOfSymfony/FOSRestBundle/blob/8779ceebf715d1c60bd10286fce9d32ed03c484a/Serializer/Normalizer/FlattenExceptionNormalizer.php#L49-L53

PR #5799, opened 2021-02-05
Allow to display specific error message returned by server in Form view

### LINKED ISSUE #4822: Duplicates can be created when working with drafts (save and not publish)
| Q | A
| --- | ---
| Bug? | yes
| New Feature? | no
| Sulu Version | adcfb1b5c4aa3d40a8adde5e2c2579d1aa8fc601
| Browser Version | all

#### Actual Behavior

When u create a new page as a concept, so u don't publish it, the url is not automatically changed with a number. If u create multiple pages with the same title and so the same url and u want to publish them later, it says the location is not valid. Because of the duplicate. 

#### Expected Behavior

When u are creating a page with the same title and same url, add a number to the url as you would do when actually published.

#### Steps to Reproduce

Create concepts (save without publishing) with the same title (and also same url). Try to publish them later. The first one will be published, the second one will not because of duplicate.

#### Possible Solutions

When u are creating a page with the same title and same url, add a number to the url as you would do when actually published.

niklasnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Fixed tickets | fixes #4822
| Related issues/PRs | #5312
| License | MIT

#### What's in this PR?

~This PR will add a `errorCodeMessages` option to the `Form` view that allows to **display a specific error message based on the response of the server**, when the save request of the form fails.~

This PR adds a `ErrorMessageExceptionInterface` that allows to **set a translation key that is displayed as error message in the administration interface** when the save request of a form fails. The interface can be reused to display error messages in other parts of the system (eg https://github.com/sulu/sulu/pull/5797) in a future PR.

Furthermore, the PR utilizes the `ErrorMessageExceptionInterface` to display a helpful message in the following cases:

- [x] Submitted URL is already assigned to another page
- [x] Submitted name is already assigned to another role
- [x] Submitted key is already assigned to another role
- [x] Submitted email is already assigned to another user
- [x] Submitted username is already assigned to another user
- [x] Submitted title is already assigned to another custom url

#### Why?

At the moment, the error message in all cases listed above is just `There was an error when trying to save the form`. This does not really help the user to solve the problem. Also see #4822 and #5312


src/Sulu/Bundle/SecurityBundle/UserManager/UserManager.php
```diff
@@ -681,7 +681,7 @@ private function processEmail(UserInterface $user, $email, $contact = null)
             }
         } else {
             if (null !== $email) {
-                if ($email !== $user->getEmail() &&
+                if (strtolower($email) !== strtolower($user->getEmail()) &&
```
niklasnatter: fixes #5596

src/Sulu/Bundle/AdminBundle/Admin/View/FormOverlayListViewBuilderInterface.php
```diff
@@ -97,4 +97,6 @@ public function addMetadataRequestParameters(array $metadataRequestParameters):
      * @param Badge[] $badges
      */
     public function addTabBadges(array $badges): self;
+
+    public function addErrorCodeMessages(array $errorCodeMessages): self;
```
alexander-schranz: I must say I don't lik it this way to configure this on the view level. The error `codes` are designed to be unique and every package had its own [number](https://github.com/sulu/docs/tree/master/detail-specification). Not sure if I would implement this things this way, maybe a global mapping between error code and error message would be better. I think still the whole correct translated messages should directly come from the error response and no mapping between an error code or other things should be needed to be configured somewhere. And if it should we maybe should configure it at a global place and not for every view again. As documented in the issue I really would prefer that we use a well formated error format like its done in the RFC and and not had additional mapping on the view here. I'm really afraid that this will just lead that the API itself will never return nice error messages just error codes which need to be looked up.

/cc @sulu/core-team

src/Sulu/Bundle/SecurityBundle/UserManager/UserManager.php
alexander-schranz: Think a test for this fix would be great 👍

src/Sulu/Bundle/AdminBundle/Resources/js/views/Form/Form.js
```diff
@@ -396,17 +396,34 @@ class Form extends React.Component<Props> {
...
+    handleSaveError = (error: Object) => {
+        const {
+            route: {
+                options: {
+                    errorCodeMessages = {},
+                },
+            },
+        } = this.props.router;
+
+        const errorMessage = (error.code && errorCodeMessages[error.code])
```
alexander-schranz: I'm not sure if this error.code is the exception code under production environment because there are 2 different serializer of exception for dev and prod in fos rest and under prod I think `code` will just return the `statusCode` (400, 500) and no error code of the exception 
https://github.com/sulu/sulu/issues/5336. I don't think we will get around that we are creating a ExceptionInterface which will add somekind of `errorCode` additonal maybe `errorMessage` create our own exception serializer which I think already exist for develop.

src/Sulu/Bundle/AdminBundle/Admin/View/FormOverlayListViewBuilderInterface.php
niklasnatter: The idea behind this approach is that we dont need to use a specific format for the `code`. It allows to display messages for numerical codes of the existing API, but we can also use string codes such as `blocked-filetype` in newly implemented APIs.

Another idea is to make it possible to display different error messages for the same code based on the current view. For example, this makes it easily possible to **display an error message based on the HTTP status code of the response**. Lets be honest - most developers will just throw a HTTP status code to indicate an error. 

Finally, with this solution we can fallback to a generic error message and it is easy to find out if a translation is still used by just searching through the codebase. These things are not possible when generating the translation key dynamically without a mapping.

> The error codes are designed to be unique and every package had its own number.

This does not scale very well 🙈  Who decides which third party bundle uses which code range?

> maybe a global mapping between error code and error message would be better.

As said above, a global mapping does not scale very well. Also it prevents to display a error message for a specific status code. What advantages do you see with that?

> I think still the whole correct translated messages should directly come from the error response and no mapping between an error code or other things should be needed to be configured somewhere.

Returning a localized message is kind of strange for unlocalized APIs. Do you know an API that returns translated messages? I would prefer to return local independent key such as `blocked-filetype` in newly implemented APIs. 

But we can still add an additional property (eg `localizedErrorMessage`) in the future that will be displayed directly without any mapping. **This solution does not prevent that at all**, it just implements the handling for the existing APis which we cannot change anyway.

> As documented in the issue I really would prefer that we use a well formated error format like its done in the RFC and and not had additional mapping on the view here.

We should still implement a well structured error format that allows to display an error for a specific field in a later PR.  As stated above, we can also add a `localizedErrorMessage` property that will be displayed without any mapping in a later PR 🙂

src/Sulu/Bundle/AdminBundle/Resources/js/views/Form/Form.js
niklasnatter: It should be available in `prod`; I get the following response on `sulu.rocks`:
```
{"code":1103,"message":"Conflict"}
```

src/Sulu/Bundle/AdminBundle/Admin/View/FormOverlayListViewBuilderInterface.php
niklasnatter: Fact is, we cannot change the existing APIs and we should really implement a solution for this. It is quite embarrassing that Sulu still does not allow to display a custom error message at all. 

This solution **does not require big changes or BC breaks**, **works with the existing API** and it is flexible enough to be used inside of projects or bundle. Furthermore **it works now** and it does not prevent to implement a better solution for displaying field-specific errors in the future at all.

src/Sulu/Bundle/AdminBundle/Resources/js/views/Form/Form.js
alexander-schranz: You are correct seems like this was fixed also the https://sulu.rocks/admin/api/media does work now as expected under prod. I mean code is 0 but its not longer the status code.

src/Sulu/Bundle/AdminBundle/Admin/View/FormOverlayListViewBuilderInterface.php
alexander-schranz: > but we can also use string codes such as blocked-filetype in newly implemented APIs.

Definitly we should not go on "string" codes. I also think string codes would be the false way. I think the API should return a translated error message for the user system language. So also API only users know really what the error was.

> Returning a localized message is kind of strange for unlocalized APIs.  I would prefer to return local independent key such as blocked-filetype in newly implemented APIs.

Adding a key is maybe a good idea to add logic on it but as we have the code for it I would not add something like a key.

> Returning a localized message is kind of strange for unlocalized APIs.

Error messages are not returned in the "content" language of the API they are returned in the "system" languages. SO I don't see here a problem with unlocalized APIs.

> Do you know an API that returns translated messages?

The most API i did work with always had both. Some kind of error code or error string key and a translated message, so you never needed to lookup errors in some documentation. 

> This solution does not require big changes or BC breaks, works with the existing API 

But we are providing then different ways to achieve the same thing. Which I think is not a great idea. And what we do in future when there is a error message returned which has the priority the mapping in the view or the error message returned by the API?

> BC breaks

I don't think that adding properties to the error response is a BC Break. So I did never understand that argument from daniel as we did add properties on all apis without any versioning and problems. So the current state is:

```json
{
     "code": "0",
     "message": "Internal Server Error"
}
```

And then add here the rfc7807 message parameters would be no bc break:

```json
{
     "code": "0",
     "message": "...",
     "type": "..",
     "title": ".."
}
```

I just wanted to mention my fears about implementing it this way. Also I'm afraid that we will never implement it to provide nice error messages over our APIs which I think is required for a good Developer Experience. I hope you understand my point of view on this topic. I really appreciate that you are working on this and if the others don't share my fears I'm okay with the solution.

src/Sulu/Bundle/SecurityBundle/UserManager/UserManager.php
niklasnatter: Yes I havent found time to add any tests yet. Will add them when we agree on how to implement it 🙂

src/Sulu/Bundle/AdminBundle/Admin/View/FormOverlayListViewBuilderInterface.php
niklasnatter: > Definitly we should not go on "string" codes. I also think string codes would be the false way. I think the API should return a translated error message for the user system language.

I think string codes are nicer than number codes because the make the code easier to understand without the need of constants, we can still add a translated message in another key. This could look like this:
```
{
    "code": "blocked-filetype"
    "message": "The filetype of the uploaded file is configured to be blocked."
}
```

> But we are providing then different ways to achieve the same thing. Which I think is not a great idea.

That is right, there would be two possibilities:

- Specify the message for an error code or HTTP status code in a mapping
- Specify the error message that should be displayed directly

We dont need to use the first possibility in the core, but I would still like to implement it because I think most projects will just use HTTP status codes for indicating errors.

> And what we do in future when there is a error message returned which has the priority the mapping in the view or the error message returned by the API?

I would use the following priority:

1. Translated error message in the response (eg `message` property)
2. Mapped error message for the `code` property of the response
3. Mapped error message for the HTTP status code of the response

It would be okay for me to get rid of the second type. But as already said, I think we should allow to define a mapping for HTTP status codes 🙂 

> Also I'm afraid that we will never implement it to provide nice error messages over our APIs which I think is required for a good Developer Experience.

I am afraid of that too. Do you see an easy solution to implement this for the existing APIs? 

If we find an easy way for returning a translated error message in the cases where we already return a code now, I would really like to implement that.

> I really appreciate that you are working on this and if the others don't share my fears I'm okay with the solution.

Thats nice to hear and I appreciatee the feedback! 🙂  
I am sorry if my wording is a bit harsh, but I am really annoyed about this topic and that we still have not made any progress about it. 

**I want to solve this, at least for simple cases, in the next release**. I think this PR is not perfect, but it would solve things for most usecases. If we find a way to return additional translated error messages, it would be even better. But I think I need your help for that 🙂

src/Sulu/Bundle/AdminBundle/Resources/js/views/Form/Form.js
```diff
@@ -396,17 +396,34 @@ class Form extends React.Component<Props> {
...
+            ? errorCodeMessages[error.code]
```
nnatter, line 421: TODO: check if there is a message for the http status code of the error

src/Sulu/Bundle/SecurityBundle/UserManager/UserManager.php
niklasnatter: Reverted this change because of a failing test case. Guess we need to fix this in another PR 🙂

src/Sulu/Component/Rest/FlattenExceptionNormalizer.php
```diff
@@ -29,18 +31,30 @@ class FlattenExceptionNormalizer implements NormalizerInterface
...
 
     public function normalize($exception, $format = null, array $context = [])
     {
         $data = $this->normalizer->normalize($exception, $format, $context);
+        $data['code'] = $exception->getCode();
 
-        if (\is_array($data)) {
-            $data['code'] = $exception->getCode();
+        $contextException = $context['exception'] ?? null;
+        if ($contextException instanceof ErrorMessageExceptionInterface) {
```
nnatter, line 52: TODO: add test case

src/Sulu/Component/Rest/Exception/ErrorMessageExceptionInterface.php
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Component\Rest\Exception;
+
+interface ErrorMessageExceptionInterface
```
nnatter, line 14: TODO. rename to `TranslationErrorMessageExceptionInterface`

src/Sulu/Component/Rest/FlattenExceptionNormalizer.php
```diff
@@ -22,28 +24,37 @@ class FlattenExceptionNormalizer implements NormalizerInterface
...
-        $this->debug = $debug;
+    public function __construct(
+        NormalizerInterface $decoratedNormalizer,
+        TranslatorInterface $translator
+    ) {
+        $this->decoratedNormalizer = $decoratedNormalizer;
+        $this->translator = $translator;
     }
 
     public function normalize($exception, $format = null, array $context = [])
```
niklasnatter, line 42: method is called by the `SerializerErrorRenderer`: https://github.com/symfony/symfony/blob/5.x/src/Symfony/Component/ErrorHandler/ErrorRenderer/SerializerErrorRenderer.php#L74-L77

alexander-schranz (approved):
This implementation looks good from my side. Hope that in future not all ToolbarActions need to handle this themself, but at current state its nice to see that it works for the Form View. Good Job 👍

PR #5803, opened 2021-02-08
Fix DeviceDetectorSubscriber with newer matomo device detector version

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

The method `deviceDetector->setUserAgent(..)` expects a string, if no `User-Agent` was set in the request-header an exception is thrown because null is not allowed to set.

#### Why?

The exception should not be thrown on missing `User-Agent` headers.

### DIFF
--- a/src/Sulu/Bundle/AudienceTargetingBundle/EventListener/DeviceDetectorSubscriber.php
+++ b/src/Sulu/Bundle/AudienceTargetingBundle/EventListener/DeviceDetectorSubscriber.php
@@ -39,7 +39,7 @@ class DeviceDetectorSubscriber implements EventSubscriberInterface
 
     public function setUserAgent(RequestEvent $event)
     {
-        $this->deviceDetector->setUserAgent($event->getRequest()->headers->get('User-Agent'));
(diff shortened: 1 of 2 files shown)


src/Sulu/Bundle/AudienceTargetingBundle/EventListener/DeviceDetectorSubscriber.php
```diff
@@ -39,7 +39,7 @@ public static function getSubscribedEvents()
 
     public function setUserAgent(RequestEvent $event)
     {
-        $this->deviceDetector->setUserAgent($event->getRequest()->headers->get('User-Agent'));
+        $this->deviceDetector->setUserAgent($event->getRequest()->headers->get('User-Agent') ?? '');
```
alexander-schranz: The default can be handled also over the second parameter:

```suggestion
        $this->deviceDetector->setUserAgent($event->getRequest()->headers->get('User-Agent', '');
```

alexander-schranz (changes requested):
As discussed a test which really calls the DeviceDetector instance should be created here.

alexander-schranz (approved):

PR #5805, opened 2021-02-10
Adjust data-format used by auto_complete type of single_selection to be consistent with other selections

### LINKED ISSUE #4053: Single selection autocomplete field type should work with id only
| Q | A
| --- | ---
| Bug? | no
| New Feature? | no
| Sulu Version | for 2.0

#### Actual Behavior

Actually the api need always need to load the whole object. This currently means if I implemented a single selection content type I need to call the repository in `read` and I need to call `getContentData` as read is called before getContentData this means everytime on the website the database is called also twice.

<details>
   <summary>Example Contentype </summary>

   ```php
<?php

namespace AppBundle\Content\Types;

use AppBundle\Model\Event\Event;
use AppBundle\Model\Localization;
use AppBundle\Repository\EventRepository;
use PHPCR\NodeInterface;
use Sulu\Component\Content\Compat\PropertyInterface;
use Sulu\Component\Content\SimpleContentType;

class SingleEventSelection extends SimpleContentType
{
    /**
     * @var EventRepository
     */
    private $eventRepository;

    public function __construct(EventRepository $eventRepository)
    {
        parent::__construct('SingleEventSelection', null);
        $this->eventRepository = $eventRepository;
    }

    /**
     * TODO this should be able to be removed and just return the id.
     *
     * {@inheritdoc}
     */
    public function read(NodeInterface $node, PropertyInterface $property, $webspaceKey, $languageCode, $segmentKey)
    {
        if (!$node->hasProperty($property->getName())) {
            return null;
        }

        $event = $this->eventRepository->findById($node->getPropertyValue($property->getName()), new Localization($languageCode));

        $value = null;
        if ($event) {
            $value = [
                'id' => $event->getId(),
                'title' => $event->getTitle(),
            ];
        }

        $property->setValue($value);

        return $value;
    }

    /**
     * TODO this should be able to be removed and just return the value.
     *
     * {@inheritdoc}
     */
    protected function encodeValue($value)
    {
        if (is_array($value)) {
            return $value['id'];
        }

        return $value;
    }

    public function getContentData(PropertyInterface $property): ?Event
    {
        // TODO value should return id only and not the whole object but read need this for the current single selection.
        $value = $property->getValue();

        if (!isset($value['id'])) {
            return null;
        }

        return $this->eventRepository->findById($value['id'], new Localization($property->getStructure()->getLanguageCode()));
    }
}
```

</details>

#### Expected Behavior

The single selection should load the object itself. This should in may case be the default behaviour of a single selection.

<details>
<summary>Expected Content type</summary>

```php
<?php

namespace AppBundle\Content\Types;

use AppBundle\Model\Event\Event;
use AppBundle\Model\Localization;
use AppBundle\Repository\EventRepository;
use Sulu\Component\Content\Compat\PropertyInterface;
use Sulu\Component\Content\SimpleContentType;

class SingleEventSelection extends SimpleContentType
{
    private $eventRepository;

    public function __construct(EventRepository $eventRepository)
    {
        parent::__construct('SingleEventSelection', null);
        $this->eventRepository = $eventRepository;
    }

    public function getContentData(PropertyInterface $property): ?Event
    {
        $value = $property->getValue();

        if (!isset($value)) {
            return null;
        }

        return $this->eventRepository->findById($value, new Localization($property->getStructure()->getLanguageCode()));
    }
}
```

</details>

#### Possible Solutions

The single selection should load the data from the api.

### LINKED ISSUE #4960: auto_complete ignores locale
| Q | A
| --- | ---
| Bug? | yes
| New Feature? | no
| Sulu Version | Sulu 2.0.2
| Browser Version | Firefox 71.0 (64 bits)

#### Actual Behavior

It's about the `auto_complete` widget which can be used when editing a content in back-office. In the AJAX query performed to feed suggestions, locale parameter is not sent to endpoint (unlike for  `list_overlay` widget).

Below, an example of AJAX query for `list_overlay` widget:
>  /admin/api/authors?flat=true&page=1&locale=fr&limit=10&fields=first_name
> _Notice the part "&locale=fr"_

And below, an example of AJAX query for `auto_complete` widget:
>  /admin/api/theme?flat=true&excludedIds=&limit=10&page=1&searchFields=title&search=the

#### Expected Behavior

To send `locale` parameter to endpoint.

#### Steps to Reproduce

Edits a content with an `auto_complete` and an `list_overlay` widgets. Analyzes network when updating fields.

#### Possible Solutions
From Sulu slack conversation:
> niklas  14 days ago
> looks like the SearchStore does not know anything about locales - this should probably be adjusted :slightly_smiling_face: https://github.com/sulu/sulu/blob/da3240d/src/Sulu/Bundle/AdminBundle/Resources/js/containers/MultiAutoComplete/MultiAutoComplete.js#L53

niklasnatter:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | yes
| Deprecations? | yes
| Fixed tickets | fixes #4053, fixes #4960
| License | MIT

#### What's in this PR?

This PR adjusts the data-format used by the `auto_complete` type of the `single_selection` field-type to be consistent with other `single_selection` and `selection` types. Furthermore it adjusts the implementation of the `SingleAutoComplete` container to be consistent with other selection components. 

In practice, the changes adjust the data that is sent to the server when using a `single_account_selection` to be consistent with all other `single_*_selection` field-types:
```diff
{
     "other_property": null,
-    "single_account_selection_property": {
-        "id": 1,
-        "name": "Test Organisation",
-        ...
-    },
+    "single_account_selection_property": 1,
}
```
The old behaviour of the `single_account_selection` can be brought back by setting the `use_deprecated_object_data_format` param like this:
```diff
     <property name="single_account_selection_property" type="single_account_selection" colspan="6">
         <meta>
             <title>sulu_contact.organization</title>
         </meta>
+        <params>
+            <param name="use_deprecated_object_data_format" value="true" />
+        </params>
     </property>
```

#### Why?

Because the data-format that is used by the `single_selection` field type should be the same for all types.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,6 +1,80 @@
 # Upgrade
 
-## dev-master
+## 2.3
+
+### Changed data-format used by the single_account_selection field-type
+
+The `single_account_selection` field-type was adjusted to process an id instead of a serialized account entity.
+This makes the data-format used by the `single_account_selection` field-type consistent to all other `single_*_selection` field types.
+
+The data that is sent to the server by the field-type was changed like this:
+
+```diff
+{
+-    "single_account_selection_property": {
+-        "id": 1,
+-        "name": "Test Account",
+-        ...
+-    },
++    "single_account_selection_property": 1,
+}
+```
+
+If you have used the `single_account_selection` field-type in a form configuration for your custom entity,
+you should adjust the API of the custom entity to be compatible with the new data-format. 
+If you cannot adjust the API, you can use the `use_deprecated_object_data_format` param to bring back the old behaviour:
+
+```diff
+     <property name="single_account_selection_property" type="single_account_selection">
+         <!-- .. -->
++        <params>
++            <param name="use_deprecated_object_data_format" value="true" />
++        </params>
+     </property>
+```
+
+### Changed data-format used by the auto_complete type of single_selection field-type
+
+The `auto_complete` type of `single_selection` field-type was adjusted to process an id instead of a serialized object.
+This makes the data-format used by the `auto_complete` type consistent to all other `single_selection` types and therefore
+allows to switch between different types.
+
+If you have configured a `single_selection` field-type with the `auto_complete` type for your custom entity, 
+the data that is sent to the server by the field-type is changed like this:
+
+```diff
+{
+-    "auto_complete_single_selection_property": {
+-        "id": 1,
+-        "name": "...",
+-        ...
+-    },
++    "auto_complete_single_selection_property": 1,
+}
+```
+
+If you have used such a field-type in a form configuration for your custom entity, you should adjust the API of the 
+custom entity to be compatible with the new data-format.
+If you cannot adjust the API, you can use the `use_deprecated_object_data_format` param to bring back the old behaviour:
+
+```diff
+     <property name="auto_complete_single_selection_property" type="single_entity_selection">
+         <!-- .. -->
++        <params>
++            <param name="use_deprecated_object_data_format" value="true" />
++        </params>
+     </property>
+```
+
+### Adjusted SingleAutoComplete component to accept SingleSelectionStore
+
+The `SingleAutoComplete` container component was adjusted to accept a `SingleSelectionStore` instance via the `store` prop. 
+Furthermore, the `onChange`, `resourceKey` and `value` prop were replaced by the `store` prop and have been removed.
+This makes the implementation of the `SingleAutoComplete` consistent to the implementation of the `MultiAutoComplete`
+and makes it easier to reuse the component.
+
+If you are using the `SingleAutoComplete` container component in your custom javascript code, you need to adjust your 
+code to instantiate a `SingleSelectionStore` object and pass it to the `store` prop.
 
 ### conditionDataProvider interface changed
 
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/SingleSelection.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/SingleSelection.js
@@ -1,6 +1,6 @@
 // @flow
 import React from 'react';
-import {computed, observable, toJS} from 'mobx';
+import {computed, observable, reaction, toJS} from 'mobx';
 import type {IObservableValue} from 'mobx';
 import log from 'loglevel';
 import jsonpointer from 'json-pointer';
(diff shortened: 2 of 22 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/containers/SingleAutoComplete/SingleAutoComplete.js
```diff
@@ -3,16 +3,15 @@ import React from 'react';
 import {observer} from 'mobx-react';
 import SingleAutoCompleteComponent from '../../components/SingleAutoComplete';
 import SearchStore from '../../stores/SearchStore';
+import SingleSelectionStore from '../../stores/SingleSelectionStore';
 
 type Props = {|
     disabled: boolean,
     displayProperty: string,
     id?: string,
-    onChange: (value: ?Object) => void,
```
nnatter, line 11: TODO: add these changes to `UPGRADE.md`

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/SingleSelection.js
```diff
@@ -106,39 +129,51 @@ class SingleSelection extends React.Component<Props>
...
+
+        if (this.autoCompleteSelectionStore.loading) {
+            return;
+        }
+
+        if (this.value !== selectedItem?.id) {
+            if (this.useDeprecatedObjectDataFormat) {
+                this.handleChange((selectedItem: any));
+            } else {
+                this.handleChange(selectedItem?.id);
```
nnatter, line 147: TODO: add this change to `UPGRADE.md`

niklasnatter:
@sulu/core-team This would make the data-format consistent between all `selection` types and fix #4053 and #4960. What do you think about these changes? I am afraid we cannot do this without a BC break, therefore I chose `2.x` as target branch.

UPGRADE.md
```diff
@@ -1,6 +1,80 @@
...
+```diff
+     <property name="auto_complete_single_selection_property" type="single_entity_selection">
+         <!-- .. -->
++        <params>
++            <param name="use_deprecated_object_data_format" value="true" />
++        </params>
+     </property>
+```
+
+### Adjusted SingleAutoComplete component to accept SingleSelectionStore
```
niklasnatter, line 69: We changed a similar thing for the `MultiAutoComplete` in `2.1`: https://github.com/sulu/sulu/blob/2.x/UPGRADE.md#passing-multiselectionstore-to-multiautocomplete

src/Sulu/Bundle/ContactBundle/Content/Types/SingleAccountSelection.php
```diff
@@ -70,9 +57,9 @@ public function write(
         $languageCode,
         $segmentKey
     ) {
-        $account = $property->getValue();
-        if (null != $account) {
-            $node->setProperty($property->getName(), $account['id']);
+        $value = $property->getValue();
+        if (null != $value) {
+            $node->setProperty($property->getName(), is_array($value) ? $value['id'] : $value);
```
alexander-schranz: I think we should deprecate the $value['id'] and throw a deprecation message here as it should not longer be send this way to the backend?

niklasnatter: sounds good 👍

alexander-schranz (approved):

PR #5806, opened 2021-02-10
Validate Webspaces with SuluThemeBundle

plozmun:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

With SuluThemeBundle 3.0 validation webspace fails due to old config

#### Why?

Because we want to validate all templates in themes with
```
bin/console sulu:content:validate:webspaces
```

### DIFF
--- a/src/Sulu/Bundle/PageBundle/Command/ValidateWebspacesCommand.php
+++ b/src/Sulu/Bundle/PageBundle/Command/ValidateWebspacesCommand.php
@@ -16,6 +16,10 @@ use Sulu\Component\Content\Metadata\Factory\StructureMetadataFactoryInterface;
 use Sulu\Component\Webspace\Manager\WebspaceManagerInterface;
 use Sulu\Component\Webspace\StructureProvider\WebspaceStructureProvider;
 use Sulu\Component\Webspace\Webspace;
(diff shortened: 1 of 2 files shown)


alexander-schranz:
The Problem here is that Sulu does know something about SuluThemeBundle or the SyliusThemeBundle and it also should know nothing about them.

I could think about that we could provide a event which the SuluThemeBundle could listen to set to webspace. If we would introduce a event like `WebspaceContextSet` or something similar which we could also trhow in the RequestAnalyzer, Preview and this Command, the SuluThemeBundle would not longer have the need to listen to different events just forthis single event.

/cc @sulu/core-team what do you think?

plozmun:
I certainly think that the solution you propose is better. I can work on this if you want.

plozmun:
Maybe i can decorate the service to change some methods because I have not tested with theme inheritance

alexander-schranz:
I hope this is currently not blocking for you. We need first discuss this in the team about which effects it could have and what alternatives are there.

plozmun:
@alexander-schranz no, it is not currently blocking. If I can help, let me know

PR #5811, opened 2021-02-11
Fix psr18 client min requirements for using symfony http client

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? |  yes
| New feature? | no 
| BC breaks? | no 
| Deprecations? | no  
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Fix min requirements for the psr implementation by conflicting php-http/discovery which is lower 1.8: https://github.com/php-http/discovery/releases/tag/1.8.0

#### Why?

Currently the CachingTest fails because of I think a bug in older version of the PSR implementation for the symfony/http-client.

The 2.2 Branch did fail as there it seems the only case where the http cache bundle is tested. To see the test run on 2.2 Click here: https://github.com/sulu/sulu/actions/runs/557809206


alexander-schranz:
The 2.2 Branch did fail as there it seems the only case where the http cache bundle is tested. To see the test run on 2.2 Click here: https://github.com/sulu/sulu/actions/runs/557809206

niklasnatter (approved):

Issue #5681, opened 2020-12-03
Browser crashes on input of faulty, complex URL in URL Field within backend


saphrus:
| Q | A
| --- | ---
| Sulu Version | 2.1.5
| Browser Version | Chrome 87.0.4280.88 / Firefox 83.0

# Actual Behavior

When you enter a complex, faulty URL (e.g. `https://secret-secret-secret-secret-secret.secret-secret.com/de/publicservicedetail?q=elterngeld beantragen&id=8969217&organisationid=208445263` (the space character between "elterngeld" and "beantragen" being the fault)) in an URL input field within the backend, your browser crashes.

# Expected Behavior

no browser crashes

# Steps to Reproduce

- Log into the backend
- Have an URL input field
- Enter a faulty, complex URL (e.g. `https://secret-secret-secret-secret-secret.secret-secret.com/de/publicservicedetail?q=elterngeld beantragen&id=8969217&organisationid=208445263`)

# Possible Solutions

You can quick check the the behaviour here https://regexr.com/5hisv
Since this seems to be no sulu-specific bug, but a browser issue, I see only 2 options:
- optimize/simplify the Regex to work around the faulty JavaScript RegEx implementation
- Fix the JavaScript RegEx implementation

alexander-schranz:
I personally would remove the regex as today there with kind of things like applinks e.g. `myapp://` or other things I don't really see why we should add any validation. Also I think the regex does not support IPv6 which would also be more complicated as see for example the [Symfony Url Validator](https://github.com/symfony/symfony/blob/31ee43a8c67897a2f61782060965f5c8a81f58a2/src/Symfony/Component/Validator/Constraints/UrlValidator.php#L24-L40). 

The browser `<input type="url">` itself a valid url is all which has a [urlscheme://restofurl](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/url) the schema does only support there only `a-z` and `-`, the url can have any character with our select for schema I really don't see any need to validate the url as in our case the urlscheme is dropdown and can not be input false and the rest should support any characters.

luca-rath:
Since we've added the possibility to define json-schema per field type recently, we could use that for validation. The `uri` format is even included in the spec (https://json-schema.org/understanding-json-schema/reference/string.html#resource-identifiers)

luca-rath:
[...]

PR #5815, opened 2021-02-11
Remove email validation from js component

luca-rath:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

Remove email validation from js component

#### Why?

To be able to display proper error messages and prevent form saving


src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/MemoryFormStore.js
```diff
@@ -2,7 +2,7 @@
 import {action, observable} from 'mobx';
 import type {IObservableValue} from 'mobx';
 import jsonpointer from 'json-pointer';
-import {createAjv} from '../../../utils';
+import {createAjv} from '../../../utils/Ajv';
```
luca-rath: @nnatter Do you have an idea why this change fixes this error? https://github.com/sulu/sulu/pull/5815/checks?check_run_id=1880372111

niklasnatter: because these test classes mock the `sulu-admin-bundle/utils` modules instead of the specific module (for example `sulu-admin-bundle/utils/Translator`) 🙈  
see for example https://github.com/sulu/sulu/blob/466edcb9beb7debe5c64e75bade7d04bbb6164de/src/Sulu/Bundle/MediaBundle/Resources/js/containers/MediaCollection/tests/MediaCollection.test.js#L191-L193

luca-rath: 🙈

src/Sulu/Bundle/AdminBundle/Resources/js/utils/Ajv/formats/idnEmailValidator.js
```diff
@@ -0,0 +1,9 @@
+// @flow
+import IsEmail from 'isemail';
+import {FormatValidator} from 'ajv';
+
+const idnEmailValidator: FormatValidator = (data: string): boolean => {
```
niklasnatter, line 5: would prefer to update `ajv` and add the `ajv-formats` and `ajv-formats-draft2019` package. just had a look at the changelog, i think the update should not cause problems 🙂

src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/StringMetadata.php
```diff
@@ -53,6 +63,10 @@ public function toJsonSchema(): array
             $jsonSchema['pattern'] = $this->pattern;
         }
 
+        if (null !== $this->format) {
+            $jsonSchema['format'] = $this->format;
```
niklasnatter, line 67: this is not tested, right?

src/Sulu/Bundle/PageBundle/Content/Types/Email.php
```diff
@@ -11,15 +11,48 @@
...
  */
-class Email extends SimpleContentType
+class Email extends SimpleContentType implements PropertyMetadataMapperInterface
 {
     public function __construct()
     {
         parent::__construct('Email', '');
     }
+
+    public function mapPropertyMetadata(ContentPropertyMetadata $propertyMetadata): PropertyMetadata
```
niklasnatter, line 33: not sure if we tested this method for other `PropertyMetadataMapper`s

src/Sulu/Bundle/AdminBundle/Resources/js/utils/Ajv/formats/idnEmailValidator.js
luca-rath: I would prefer not to install `ajv-formats-draft2019`, because it contains formats which are not working as expected, see https://github.com/sulu/sulu/issues/5681#issuecomment-777442172 ... So we still need to implement the `idn-email` format by ourselves, but it's not that complicated with using `isemail`, so that should be fine

src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/StringMetadata.php
luca-rath: The whole class is not tested, but I'll write a test for it

src/Sulu/Bundle/AdminBundle/Resources/js/utils/Ajv/formats/idnEmailValidator.js
niklasnatter: just saw that the `ajv-formats-draft2019` is a community package. is okay for me too 🙂

src/Sulu/Bundle/AdminBundle/Resources/js/utils/Ajv/keywords/nameDefinition.js
```diff
@@ -0,0 +1,11 @@
+// @flow
+import type {KeywordDefinition} from 'ajv';
+
+const nameDefinition: KeywordDefinition = {
```
niklasnatter, line 4: not really sure what this does? can you give me a hint? 😇

niklasnatter (approved):
looks good otherwise!

src/Sulu/Bundle/AdminBundle/Resources/js/utils/Ajv/keywords/nameDefinition.js
luca-rath: The json schema returned by our api often looks like this

~~~json
{
  "type": "object",
  "properties": {
    "foo": {
      "name": "foo",
      "type": "string"
    }
  }
}
~~~

But `name` is not a valid json schema keyword, so we need to explicitly allow it.

luca-rath: In ajv v7, the strict mode is enabled by default, therefore this wasn't needed before. We could also disable that, but I think it's generally good to have strict mode enabled

niklasnatter: alright sounds good 👍

Issue #4576, opened 2019-05-22
RFC: Provide a more performant alternative to sulu_content_load


alexander-schranz:
| Q | A
| --- | ---
| Bug? | no
| New Feature? | yes
| Sulu Version | for 2.0
| Browser Version | -

#### Actual Behavior

If you have only a uuid of a page because e.g. custom entities or currently in single_page_selection in many cases `sulu_content_load` is used to get to the url. This is basically leaks in performance as `sulu_content_load` resolve all its content types so many data is loaded which is basically not needed.

You can use currently the `sulu:link` (now `sulu-link`) tag to avoid the sulu_content_load but if you need to url in another case then a tag e.g. form tag given it to a js component there is currently not alternative to sulu_content_load. 

#### Expected Behavior

We should provide a better twig extension which force you to define which fields should really be loaded / resolved by this twig extension.

#### Possible Solutions

In most cases only the url is needed so something like should be ok:

```php
sulu_page_load_url(uuid)
sulu_page_load_url(uuid, 'en') // optional locale
```

but in some cases you maybe need a little bit more information:

```php
sulu_page_load(uuid, ['title', 'url', 'excerptTitle' => 'excerpt.title'])
sulu_page_load(uuid, ['title', 'url', 'excerptTitle' => 'excerpt.title'], 'en')  // optional locale
```

Another way could also be going over the teaser provider which provides a [teaser](https://github.com/sulu/sulu/blob/develop/src/Sulu/Bundle/PageBundle/Teaser/Teaser.php) object:

```php
sulu_load_teaser('page', uuid)
sulu_load_teaser('page', uuid, locale) // optional locale
```

Also a way could be go over the link provider which is used for the sulu-link tag to get its data which provides a title and url [information](https://github.com/sulu/sulu/blob/develop/src/Sulu/Bundle/MarkupBundle/Markup/Link/LinkItem.php):

```php
sulu_load_link('page', uuid)
sulu_load_link('page', uuid, locale) // optional locale
```

niklasnatter:
I think this was fixed be https://github.com/sulu/sulu/pull/5820. Please feel free to reopen if that is not the case 🙂

alexander-schranz:
Is fine for me!

PR #5820, opened 2021-02-12
Allow to pass included properties to sulu_content_load

niklasnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Related issues/PRs | #4576
| License | MIT
| Documentation PR | sulu/sulu-docs#627

#### What's in this PR?

This PR adds an additional `properties` parameter to the `sulu_content_load` and `sulu_content_load_parent` twig function that allows to set the properties of the target that should be resolved. The added parameter behaves the same as the `properties` param of the `page_selection` content type.

#### Why?

The `sulu_content_load` and `sulu_content_load_parent` twig function are used to render specific properties (eg the title and the description) of a page in most cases. At the moment, the functions **load an resolve all extensions and all properties of the page**. This is an expensive operation that is not necessary in most cases. We even [warn about using the function in out documentation](https://docs.sulu.io/en/2.2/reference/twig-extensions/functions/sulu_content_load.html). 

While some usecases can be implemented by using the [`<sulu-link>` tag](https://docs.sulu.io/en/2.2/bundles/markup/link.html) instead of the `sulu_content_load` function, there are cases where there is no alternative to using these functions. This change allows to improve performance in these cases by specifying the properties that are needed.

#### Example Usage

```twig
{# old behaviour #}
sulu_content_load(content.single_page_selection)

{# new behaviour #}
sulu_content_load(content.single_page_selection, ['title'])
sulu_content_load(content.single_page_selection, {'title': 'title', 'excerptTitle': 'excerpt.title'})
```

### DIFF
--- a/src/Sulu/Bundle/WebsiteBundle/Resolver/StructureResolver.php
+++ b/src/Sulu/Bundle/WebsiteBundle/Resolver/StructureResolver.php
@@ -45,7 +45,7 @@ class StructureResolver implements StructureResolverInterface
 
     public function resolve(StructureInterface $structure, bool $loadExcerpt = true/*, array $includedProperties = null*/)
     {
-        $includedProperties = (\func_num_args() > 2) ? null : \func_get_arg(2);
+        $includedProperties = (\func_num_args() > 2) ? \func_get_arg(2) : null;
 
         $data = [
             'view' => [],
--- a/src/Sulu/Bundle/WebsiteBundle/Resolver/StructureResolver.php
+++ b/src/Sulu/Bundle/WebsiteBundle/Resolver/StructureResolver.php
@@ -43,7 +43,7 @@ class StructureResolver implements StructureResolverInterface
         $this->extensionManager = $structureManager;
     }
 
-    public function resolve(StructureInterface $structure, bool $loadExcerpt = true)
+    public function resolve(StructureInterface $structure, bool $loadExcerpt = true, array $includedProperties = null)
     {
         $data = [
             'view' => [],
@@ -96,17 +96,21 @@ class StructureResolver implements StructureResolverInterface
 
         // pre-resolve content-types
         foreach ($structure->getProperties(true) as $property) {
-            $contentType = $this->contentTypeManager->get($property->getContentTypeName());
+            if ($includedProperties === null || in_array($property->getName(), $includedProperties)) {
+                $contentType = $this->contentTypeManager->get($property->getContentTypeName());
 
-            if ($contentType instanceof PreResolvableContentTypeInterface) {
-                $contentType->preResolve($property);
+                if ($contentType instanceof PreResolvableContentTypeInterface) {
+                    $contentType->preResolve($property);
+                }
             }
         }
 
         foreach ($structure->getProperties(true) as $property) {
-            $contentType = $this->contentTypeManager->get($property->getContentTypeName());
-            $data['view'][$property->getName()] = $contentType->getViewData($property);
-            $data['content'][$property->getName()] = $contentType->getContentData($property);
+            if ($includedProperties === null || in_array($property->getName(), $includedProperties)) {
+                $contentType = $this->contentTypeManager->get($property->getContentTypeName());
+                $data['view'][$property->getName()] = $contentType->getViewData($property);
+                $data['content'][$property->getName()] = $contentType->getContentData($property);
+            }
         }
 
         return $data;
--- a/src/Sulu/Bundle/WebsiteBundle/Resolver/StructureResolver.php
+++ b/src/Sulu/Bundle/WebsiteBundle/Resolver/StructureResolver.php
@@ -43,8 +43,10 @@ class StructureResolver implements StructureResolverInterface
         $this->extensionManager = $structureManager;
     }
 
-    public function resolve(StructureInterface $structure, bool $loadExcerpt = true, array $includedProperties = null)
+    public function resolve(StructureInterface $structure, bool $loadExcerpt = true/*, array $includedProperties = null*/)
     {
+        $includedProperties = (\func_num_args() > 2) ? null : \func_get_arg(2);
+
         $data = [
             'view' => [],
             'content' => [],
--- a/src/Sulu/Bundle/WebsiteBundle/Resolver/StructureResolver.php
+++ b/src/Sulu/Bundle/WebsiteBundle/Resolver/StructureResolver.php
@@ -96,7 +96,7 @@ class StructureResolver implements StructureResolverInterface
 
         // pre-resolve content-types
         foreach ($structure->getProperties(true) as $property) {
-            if ($includedProperties === null || in_array($property->getName(), $includedProperties)) {
+            if (null === $includedProperties || \in_array($property->getName(), $includedProperties)) {
                 $contentType = $this->contentTypeManager->get($property->getContentTypeName());
 
                 if ($contentType instanceof PreResolvableContentTypeInterface) {
@@ -106,7 +106,7 @@ class StructureResolver implements StructureResolverInterface
         }
 
         foreach ($structure->getProperties(true) as $property) {
-            if ($includedProperties === null || in_array($property->getName(), $includedProperties)) {
+            if (null === $includedProperties || \in_array($property->getName(), $includedProperties)) {
                 $contentType = $this->contentTypeManager->get($property->getContentTypeName());
                 $data['view'][$property->getName()] = $contentType->getViewData($property);
                 $data['content'][$property->getName()] = $contentType->getContentData($property);
--- a/src/Sulu/Bundle/WebsiteBundle/Resolver/StructureResolverInterface.php
+++ b/src/Sulu/Bundle/WebsiteBundle/Resolver/StructureResolverInterface.php
@@ -26,5 +26,5 @@ interface StructureResolverInterface
      *
      * @return array
      */
-    public function resolve(StructureInterface $structure, bool $loadExcerpt = true, array $includedProperties = null);
+    public function resolve(StructureInterface $structure, bool $loadExcerpt = true/*, array $includedProperties = null*/);
 }
--- a/src/Sulu/Bundle/WebsiteBundle/Resolver/StructureResolverInterface.php
+++ b/src/Sulu/Bundle/WebsiteBundle/Resolver/StructureResolverInterface.php
@@ -26,5 +26,5 @@ interface StructureResolverInterface
      *
      * @return array
      */
-    public function resolve(StructureInterface $structure, bool $loadExcerpt = true);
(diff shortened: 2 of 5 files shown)


niklasnatter:
@sulu/core-team Just a suggestion - what do you think?

src/Sulu/Bundle/WebsiteBundle/Resolver/StructureResolverInterface.php
```diff
@@ -26,5 +26,5 @@ interface StructureResolverInterface
      *
      * @return array
      */
-    public function resolve(StructureInterface $structure, bool $loadExcerpt = true);
+    public function resolve(StructureInterface $structure, bool $loadExcerpt = true, array $includedProperties = null);
```
niklasnatter: TODO: add parameter in a non-breaking way like in the `MediaManagerInterface`:
https://github.com/sulu/sulu/blob/466edcb9beb7debe5c64e75bade7d04bbb6164de/src/Sulu/Bundle/MediaBundle/Media/Manager/MediaManagerInterface.php#L36

src/Sulu/Bundle/WebsiteBundle/Resolver/StructureResolver.php
```diff
@@ -43,7 +43,7 @@ public function __construct(
         $this->extensionManager = $structureManager;
     }
 
-    public function resolve(StructureInterface $structure, bool $loadExcerpt = true)
+    public function resolve(StructureInterface $structure, bool $loadExcerpt = true, array $includedProperties = null)
```
nnatter, line 46: we could even deprecate calling this without a `includedProperties` parameter. not sure if there are a lot of usecases for that

src/Sulu/Bundle/WebsiteBundle/Twig/Content/ContentTwigExtension.php
```diff
@@ -77,7 +77,7 @@ public function getFunctions()
         ];
     }
 
-    public function load($uuid)
+    public function load($uuid, bool $includeExtensions = true, array $includedProperties = null)
```
niklasnatter: another possibility would be a `array $properties` parameter that matches the semantic of the `properties` param of the `page_selection`

alexander-schranz: I would prefer that the `sulu_content_load` would work the same way as `page_selection` currently not sure if the `page_selection` outputs any `view` data. I think it would be good if it would do that (maybe changes in the page_selection is needed to be more consistent). Still I think the mapping configuration with `excerptImages => excerpt.images` is a nice feature of the page_selection and make it easier for twig developer to access this things.

luca-rath:
I assume one reason for creating this pull request was a recent [slack question](https://sulu-io.slack.com/archives/C5Q2MG1DY/p1613123929024300) regarding using `sulu_content_load()` with a page with a paginated smart content, if the current page has also a paginated smart_content, which led to a `PageOutOfBoundsException`, right? If not, ignore the following.

I think this is definitely a good enhancement of the `sulu_content_load()` function, but I think I would fix my previously described problem differently. My suggestion is duplicating the current request and remove it's query parameters before calling the StructureResolver, see this code snippet:

~~~php
// ContentTwigExtension.php

$request = $this->requestStack->getCurrentRequest();
$this->requestStack->push($request->duplicate([]));
$resolvedStructure = $this->structureResolver->resolve($contentStructure);
$this->requestStack->pop();

return $resolvedStructure;
~~~

niklasnatter:
> My suggestion is duplicating the current request and remove it's query parameters before calling the StructureResolver, see this code snippet

That would solve the `PageOutOfBoundsException` problem, but i feel like it is a really dangerous thing to do. I would suspect that there are other places of the code that access the request and do not expect to find an empty request. Also I am not sure if this is compatible with our `RequestAnalyzer` 🙈

luca-rath:
It wouldn't be an empty request, because it would be a duplicate of the old one. It would just have empty query parameters and maybe also empty post parameters and files etc. ... I guess when doing `sulu_content_load()` you would not expect that any of the query parameters of the current request are taken into account ... and the `$requestStack->pop()` makes sure, that after executing the `sulu_content_load()` the old request stack is restored like before

alexander-schranz:
@luca-rath Duplicating the `Request` would also not solve the issue here when you have a structure with 2 smart contents you will get the same `PageOutOfBoundsException`. I would remove the whole `PageOutOfBoundsException` as I think its never good to return another statuscode because of a false page query parameter added. But lets discuss this in an own issue and not here in this context.

luca-rath:
`$request->duplicate([])` accepts the new query parameters as first argument, therefore it works, i tried ... But yes, discussing that in an own issue is probably better

alexander-schranz:
Please move discussion to about the `PageOutOfBoundsException` to #5821

src/Sulu/Bundle/WebsiteBundle/Twig/Content/ContentTwigExtension.php
```diff
@@ -90,15 +90,48 @@ public function load($uuid)
                 $this->requestAnalyzer->getCurrentLocalization()->getLocale()
             );
 
-            return $this->structureResolver->resolve($contentStructure);
+            if ($properties === null) {
+                return $this->structureResolver->resolve($contentStructure);
```
niklasnatter: we could even deprecate calling this without a `properties` parameter. not sure if there are usecases for loading all extensions and properties

src/Sulu/Bundle/WebsiteBundle/Resolver/StructureResolverInterface.php
alexander-schranz: ~~If we go with the mapping maybe which would include also the `excerpt` data we could go with:~~

```php
/**
 * @param bool|array $includedProperties
 */
public function resolve(StructureInterface $structure, $includedProperties = true)
{
      if (is_bool($includedProperties)) {
      
      }
}
```

src/Sulu/Bundle/WebsiteBundle/Twig/Content/ContentTwigExtension.php
niklasnatter: done 🙂

src/Sulu/Bundle/WebsiteBundle/Resolver/StructureResolverInterface.php
alexander-schranz: ohh didn't see this is the StructureResolver and not the extension, so forget my comment 🙈

src/Sulu/Bundle/WebsiteBundle/Twig/Content/ContentTwigExtension.php
alexander-schranz: Isn't this what the WebsiteController is dong calling the structureResolver without any options?

alexander-schranz: The WebsiteController is calling ParameterResolver which does this: https://github.com/sulu/sulu/blob/466edcb9beb7debe5c64e75bade7d04bbb6164de/src/Sulu/Bundle/WebsiteBundle/Resolver/ParameterResolver.php#L83

niklasnatter: Yes - i moved the comment from the `StructureResolver` to the `ContentTwigExtension` 🙂  
Would only deprecate it in the `ContentTwigExtension`

alexander-schranz: yeah deprecating it there sounds good for me.

src/Sulu/Bundle/WebsiteBundle/Twig/Content/ContentTwigExtension.php
```diff
@@ -89,16 +90,25 @@ public function load($uuid)
...
             );
-
-            return $this->structureResolver->resolve($contentStructure);
         } catch (DocumentNotFoundException $e) {
             $this->logger->error((string) $e);
 
             return;
         }
+
+        if (null === $properties) {
```
nnatter, line 99: TODO: adjust documentation 🙂

src/Sulu/Bundle/WebsiteBundle/Twig/Content/ContentTwigExtension.php
```diff
@@ -108,6 +118,51 @@ public function loadParent($uuid)
...
+        }
+
+        $resolvedStructure = $this->structureResolver->resolve(
+            $contentStructure,
+            !empty($extensionProperties),
+            \array_keys($contentProperties)
+        );
+
+        foreach ($contentProperties as $sourceProperty => $targetProperty) {
+            if ($sourceProperty !== $targetProperty) {
```
alexander-schranz: Not all templates will have all source properties, it should not fail when sourceProperties do not exists:

```suggestion
            if (!\array_key_exists($sourceProperty, $resolvedStructure['content'])) {
                $resolvedStructure['content'][$targetProperty] = null;
                $resolvedStructure['view'][$targetProperty] = [];
            } elseif ($sourceProperty !== $targetProperty) {
```

Not sure how this is currently handled in the PageSelection if it is set to null or the property is never set but I think it should be consistent.

alexander-schranz (changes requested):
For me the usage here with:

```twig
sulu_content_load(content.single_page_selection, {'title': 'title', 'excerpt.title': 'excerptTitle'})
```

Looks the other way around as here:

```xml
<property name="links" type="page_selection">
    <meta>
        <title lang="en">Links</title>
    </meta>

    <params>
        <param name="properties" type="collection">
            <param name="title" value="title"/>
            <param name="article" value="article"/>
            <param name="excerptTitle" value="excerpt.title"/>
            <param name="excerptDescription" value="excerpt.description"/>
        </param>
    </params>
</property>
```

I think I would go here with:

```twig
sulu_content_load(content.single_page_selection, {'title': 'title', 'excerptTitle': 'excerpt.title'})
```

And it in the twig extension, else I think this could be confusing for the twig developer which know the mapping from the page_selection and smart_content pages provider.

Also with the target as key a source can be mapped into 2 properties e.g.:

```xml
<param name="title" value="title"/>
<param name="excerptTitle" value="title"/>
```

Which would only possible if we also switch the arguments here. Think the usage for the developer should be consistent here also we need to [flip](https://www.php.net/manual/en/function.array-flip.php) our arguments in the twig extension for them.

src/Sulu/Bundle/WebsiteBundle/Twig/Content/ContentTwigExtension.php
niklasnatter: the `page_selection` just omits the property. implemented it like this here too 🙂

alexander-schranz (approved):

PR #5826, opened 2021-02-17
Add validation for iban and bic fields

luca-rath:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | #5815 
| License | MIT
| Documentation PR | -

#### What's in this PR?

Add validation for iban fields


src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/SchemaMetadata.php
```diff
@@ -47,9 +47,18 @@ public function merge(self $schema): self
...
     public function toJsonSchema(): array
     {
+        $propertiesSchema = $this->propertiesMetadata->toJsonSchema();
+
+        if (!empty($propertiesSchema)) {
+            $propertiesSchema = \array_merge(
+                ['type' => 'object'],
+                $propertiesSchema
+            );
+        }
```
luca-rath: This is necessary to remove the warning from ajv (because of strict mode, `properties` or `required` keyword must not appear without `"type": "object"`)

src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/SchemaMetadata.php
```diff
@@ -59,7 +68,10 @@ public function toJsonSchema(): array
          * empty schema object as array instead of an object and would break
          */
         if (empty($jsonSchema)) {
-            $jsonSchema['required'] = [];
+            return [
+                'type' => 'object',
+                'required' => [],
+            ];
```
luca-rath, line 74: Same here

src/Sulu/Bundle/ContactBundle/Resources/js/index.js
```diff
@@ -245,12 +256,10 @@ initializer.addUpdateConfigHook('sulu_contact', (config: Object, initialized: bo
                         iban: {
                             label: translate('sulu_contact.iban'),
                             required: true,
-                            colSpan: 8,
                             type: 'iban',
                         },
                         bic: {
                             label: translate('sulu_contact.bic'),
-                            colSpan: 4,
```
luca-rath, line 253: With `colSpan: 4`, the field is not wide enough to properly display the field error, because the error is too long

src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/SchemaMetadata.php
```diff
@@ -59,7 +68,9 @@ public function toJsonSchema(): array
          * empty schema object as array instead of an object and would break
          */
         if (empty($jsonSchema)) {
-            $jsonSchema['required'] = [];
+            return [
+                'type' => ['number', 'string', 'boolean', 'object', 'array', 'null'],
```
luca-rath, line 76: This schema matches every value, like `['required' => []]` did, but doesn't throw a warning because no type was specified

package.json
```diff
@@ -40,6 +40,9 @@
         "@babel/preset-react": "^7.0.0",
         "@ckeditor/ckeditor5-dev-utils": "^23.5.1",
         "@ckeditor/ckeditor5-theme-lark": "^23.0.0",
+        "ajv": "^7.1.0",
+        "ajv-formats": "^1.5.1",
+        "ajv-keywords": "^4.0.0",
```
luca-rath:
```suggestion
```

src/Sulu/Bundle/AdminBundle/Resources/js/utils/Ajv/formats/index.js
```diff
@@ -1,8 +1,12 @@
 // @flow
 import {FormatValidator} from 'ajv';
+import bicValidator from './bicValidator';
+import ibanValidator from './ibanValidator';
 import idnEmailValidator from './idnEmailValidator';
 
 const formats: {[string]: FormatValidator} = {
+    'bic': bicValidator,
+    'iban': ibanValidator,
```
niklasnatter: these are not included in any specification, right? 
if yes, i think we should not add them as format. this will couple the generated json schema to our frontend implementation and make it harder to use in other places (which was the reason for using json schema in the first place) 😕

src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/SchemaMetadata.php
niklasnatter: not perfectly sue, but i think we can use `return new \stdClass();` to return an empty schema that will be serialized to an empty object instead of an array. that would solve the problem described above and we dont need to hardcode all tpes here 🙂

src/Sulu/Bundle/AdminBundle/Resources/js/utils/Ajv/createAjv.js
```diff
@@ -6,7 +6,7 @@ import type {KeywordDefinition} from 'ajv';
 import customKeywords from './keywords';
 import customFormats from './formats';
 
-const createAjv = (options: Options = {allErrors: true}) => {
+const createAjv = (options: Options = {allErrors: true, allowUnionTypes: true}) => {
```
niklasnatter, line 7: not sure if `allowUnionTypes` is needed if we return an empty schema above

src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/SchemaMetadata.php
niklasnatter: would be nice to add the comment to the code

src/Sulu/Bundle/AdminBundle/Resources/js/utils/Ajv/formats/index.js
luca-rath: I could just use the `pattern` keyword then for both, would that be better?

luca-rath: But a iban can't be fully checked with regex only, because it contains some sort of verification number, so we could only check the iban for proper format, not for validity

src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/SchemaMetadata.php
luca-rath: The `toJsonSchema()` method has `array` as return type, and I think I would not change that ... But these hard coded types won't change, because those are all valid json types

src/Sulu/Bundle/AdminBundle/Resources/js/utils/Ajv/formats/index.js
luca-rath: We could prefix the non-standard formats with e.g. `custom-` or `sulu-`

src/Sulu/Bundle/AdminBundle/Metadata/SchemaMetadata/SchemaMetadata.php
niklasnatter: you are right. this way is okay for me too 🙂

src/Sulu/Bundle/AdminBundle/Resources/js/utils/Ajv/formats/index.js
niklasnatter: yes i would go with `pattern`. i dont think anybody expects the inputted iban to be checked for validity in depth and it makes the generated schema better while keeping it compatible with other validators beside our frontend 🙂

luca-rath: But it would be nice to have instant feedback if you made a typo ... especially because it was the same behaviour before

niklasnatter: Thats true, but we decided to use JSON schema for our validation because it is a standard and we can reuse it in other places. That reason is no longer valid if we start to generate schemas that are not compatible to that standard.

luca-rath: I guess most validators could be configured to ignore unknown formats like `iban` or `bic` ... Probably most validators also do not implement `idn-email` (like ajv), therefore we had to implement that by ourselves

niklasnatter: `idn-email` is part of the specification: https://json-schema.org/understanding-json-schema/reference/string.html#built-in-formats

luca-rath: Yes, but it's still not implemented by ajv

niklasnatter: > I guess most validators could be configured to ignore unknown formats like `iban` or `bic`

Thats possible, but unlike when using `format`, these validators will not validate the data at all then. Also, it does not change the fact that the generated schema does not match the JSON schema specification

luca-rath: So if we want to validate data on the server, we need a json schema library which implements `idn-email`, otherwise we would have to implement that by ourselves as well. The exact same applies for `iban`

niklasnatter: > Yes, but it's still not implemented by ajv

`ajv` is just one implementation that tries to implement the specification. While it is not implemented yet, it probably will be implemented in the future because it is part of the specification.

luca-rath: Our json schema isn't valid at all because all object `properties` contain a `name` property ... This is not defined in the spec either

luca-rath: If I would not had introduced that definition of the `name` keyword in an earlier pull request, ajv would have thrown a javascript error and would not have validated something at all

niklasnatter: As already stated above, we decided to generate a JSON schema for validation because it is a standard and we are able to reuse the schema in other places. It took quite an effort and increased complexity to implement things this way compared to just validating things in the frontend. 

Because of this, **I am really against generating schemas that do not match the specification and are coupled to our validator implementation in the frontend.** If we start to do this now for the `iban`, we will do the same thing in other places in the future, If we do that, we just could have implemented validation by ourselves without using a standard. 

> Our json schema isn't valid at all because all object properties contain a name property ... This is not defined in the spec either

> If I would not had introduced that definition of the name keyword in an earlier pull request, ajv would have thrown a javascript error and would not have validated something at all

Then we should improve that and maybe use `title` instead of `name` like in the [specification](https://json-schema.org/draft/2019-09/json-schema-core.html)? That said, this is not a reason for violating the specification in other places and should not be topic of this discussion.

I think i made my point. I am against violating the specification to slightly improve the validation of an IBAN. If a majority of the @sulu/core-team thinks otherwise, I will obviously accept that 🙂

luca-rath: I think we definitely should check the iban for validity, an option would be to use the `pattern` keyword **AND** a custom `iban` format. So probably all json-schema validators will validate the pattern and those we implemented the `iban` format will additionally verify the validity of given iban

niklasnatter: If we really want to do this, I would be okay with adding an additional `sulu-format` property beside the existing `format` property. But we should prioritize to generate the best schema possible inside of the JSON schema specification and use the `sulu-format` property only for things that cannot be defined inside of the specification.

alexander-schranz: I think I would not strictly validate here if there is no clear schema support for iban or bic. Because at the end what is maybe not valid in international use, could be valid in a test system or a simulation software. Like my comment for the url validation, at the end a too strict validation would harm us more then a not so strict validation.
If we really want to add add any patterns we should find a way to make them configurable over the admin configuration.

luca-rath: Defining the pattern in the admin configuration is not easily possible, because the json schema for `iban` and `bic` is defined [here](https://github.com/sulu/sulu/blob/2.x/src/Sulu/Bundle/ContactBundle/Resources/js/index.js#L228-L230)

luca-rath: fixed

src/Sulu/Bundle/AdminBundle/Resources/js/package.json
```diff
@@ -24,6 +24,7 @@
         "fast-deep-equal": "^3.1.1",
         "font-awesome": "^4.7.0",
         "history": "^5.0.0",
+        "iban": "^0.0.14",
```
niklasnatter: think this is not used anymore 🙂

niklasnatter (commented):
looks good otherwise

src/Sulu/Bundle/ContactBundle/Resources/js/index.js
```diff
@@ -226,6 +226,17 @@ initializer.addUpdateConfigHook('sulu_contact', (config: Object, initialized: bo
                     addOverlayTitle: 'sulu_contact.add_bank_account',
                     editOverlayTitle: 'sulu_contact.edit_bank_account',
                     jsonSchema: {
+                        type: 'object',
+                        properties: {
+                            iban: {
+                                type: 'string',
+                                pattern: '^[A-Z]{2}[0-9]{2}(?:[ ]?[0-9]{4}){4}(?:[ ]?[0-9]{1,2})?$',
```
niklasnatter:
```suggestion
                                type: 'string',
                                // regex copied from: https://stackoverflow.com/a/44657292
                                // eslint-disable-next-line max-len
                                pattern: '^([A-Z]{2}[ \\-]?[0-9]{2})(?=(?:[ \\-]?[A-Z0-9]){9,30}$)((?:[ \\-]?[A-Z0-9]{3,5}){2,7})([ \\-]?[A-Z0-9]{1,3})?$',
```

src/Sulu/Bundle/ContactBundle/Resources/js/index.js
```diff
@@ -226,6 +226,17 @@ initializer.addUpdateConfigHook('sulu_contact', (config: Object, initialized: bo
...
+                            },
+                            bic: {
+                                type: 'string',
+                                pattern: '^([A-Z]{6}[A-Z2-9][A-NP-Z1-9])(X{3}|[A-WY-Z0-9][A-Z0-9]{2})?$',
```
niklasnatter:
```suggestion
                                type: 'string',
                                // eslint-disable-next-line max-len
                                // regex copied from: https://github.com/jquery-validation/jquery-validation/blob/master/src/additional/bic.js
                                pattern: '^([A-Z]{6}[A-Z2-9][A-NP-Z1-9])(X{3}|[A-WY-Z0-9][A-Z0-9]{2})?$',
```

niklasnatter (commented):
just tested some examples, looks like the iban format differs between countries 🙈 
copied another regex from stackoverflow

niklasnatter (approved):

PR #5827, opened 2021-02-17
Adjust webpack configuration to prefer nearest node_modules folder when resolving packages

niklasnatter:
At the moment, webpack will prefer the `assets/admin/node_modules` directory over the `vendor/sulu/sulu/src/Sulu/Bundle/AdminBundle/Resources/js/node_modules` directory when resolving a package from a file inside of `vendor/sulu/sulu/src/Sulu/Bundle/AdminBundle/Resources/js`. See: https://webpack.js.org/configuration/resolve/#resolvemodules

The current behaviour leads to problems if the `assets/admin/node_modules` directory contains another version of a package than the `vendor/sulu/sulu/src/Sulu/Bundle/AdminBundle/Resources/js/node_modules` directory. We encounter this problem on the `2.x` branch after updating the `ajv` package in #5815

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -67,10 +67,10 @@ module.exports = (env, argv) => { // eslint-disable-line no-undef
                     'vendor/friendsofsymfony/jsrouting-bundle/Resources/public/js'
                 ),
             },
-            modules: [nodeModulesPath, 'node_modules'],
+            modules: ['node_modules', nodeModulesPath],
         },
         resolveLoader: {
-            modules: [nodeModulesPath, 'node_modules'],
+            modules: ['node_modules', nodeModulesPath],
         },
         module: {
             rules: [


Prokyonn (approved):

alexander-schranz:
@nnatter at the end we should make sure that we don't have 2 different versions of a library which is no dev dependency installed in our setup, this always would lead to strange behaviours. The current behaviour had the effect that always the root version of a library was used and so no multiple version of a library was compiled into our builds, changing this behaviour will now lead into the effect that we will have again multiple version of multiple libraries in our build and the build will increase in its size and some libraries could also stop working.

alexander-schranz:
I think this change really not good as npm does not work well with our `../../vendor` dependencies it will on some version install duplicated version in `node_modules` and `assets/admin/node_modules`. Also with the current LTS we now have a lot more of duplicated libraries in our build: https://www.diffchecker.com/bpqA0xHN

luca-rath:
To fix the specific problem with the ajv library, I could imagine that requiring `ajv@7` in the skeleton and in sulu/sulu should be sufficient

alexander-schranz:
Yeah I know what was fixed here, but was there some time investigated why there where 2 versions of `ajv` who did require an older version of `ajv`?

luca-rath:
I think it's because of some dev dependencies of `assets/admin/package.json`

alexander-schranz:
That dependency should be find out and be fixed that there are no 2 versions installed.

luca-rath:
This snippet is from `assets/admin/node_modules/ajv/package.json`

~~~json
{
  "_requiredBy": [
    "/file-loader/schema-utils",
    "/mini-css-extract-plugin/schema-utils",
    "/postcss-loader/schema-utils",
    "/raw-loader/schema-utils",
    "/schema-utils",
    "/terser-webpack-plugin/schema-utils",
    "/webpack",
    "/webpack/schema-utils"
  ]
}
~~~

I guess there is no other option than requiring `ajv@7` in the skeleton

niklasnatter:
This caused quite a fuss 🙈  
Will try to address the different topics one by one:

> The current behaviour had the effect that always the root version of a library was used and so no multiple version of a library was compiled into our builds, changing this behaviour will now lead into the effect that we will have again multiple version of multiple libraries in our build and the build will increase in its size and some libraries could also stop working.

Its right that the previous behaviour had that effect. But this works directly against the way how `npm` handles version conflicts: https://medium.com/@imdongchen/how-does-npm-handle-conflicting-package-versions-44f90950cca5
I really doubt that this was configured like this intentionally.

The new behaviour matches the behavior in a "normal" project setup where the `vendor/sulu/sulu/src/Sulu/Bundle/AdminBundle/Resources/js/node_modules` folder has a higher priority than the root `node_modules` folder when building a file insidee of `vendor/sulu/sulu/src/Sulu/Bundle/AdminBundle/Resources/js`. See: https://webpack.js.org/configuration/resolve/#resolvemodules

> Also with the current LTS we now have a lot more of duplicated libraries in our build: https://www.diffchecker.com/bpqA0xHN

I dont know the linking works in depth, but this change does not affect the size of the build.

Size of the build before this change (2.x branch):
```
du -h ../../public/build/admin/*
2.5M    ../../public/build/admin/fonts
224K    ../../public/build/admin/images
204K    ../../public/build/admin/main.b3b9c9f20c4ec6b02730.css
3.9M    ../../public/build/admin/main.b3b9c9f20c4ec6b02730.js
 11M    ../../public/build/admin/main.b3b9c9f20c4ec6b02730.js.map
4.0K    ../../public/build/admin/manifest.json
```

Size of the build after this change (2.x branch):
```
du -h ../../public/build/admin/*
2.5M    ../../public/build/admin/fonts
224K    ../../public/build/admin/images
204K    ../../public/build/admin/main.103b106d2f798a5c4dd2.css
3.9M    ../../public/build/admin/main.103b106d2f798a5c4dd2.js
 11M    ../../public/build/admin/main.103b106d2f798a5c4dd2.js.map
4.0K    ../../public/build/admin/manifest.json
```

> To fix the specific problem with the ajv library, I could imagine that requiring ajv@7 in the skeleton and in sulu/sulu should be sufficient

That is a workaround that only works because the flawed configuration before this change. This might break things for the packages that expect to be linked with `ajv@6`.

> Yeah I know what was fixed here, but was there some time investigated why there where 2 versions of ajv who did require an older version of ajv?

The `sulu-admin-bundle` package now requires `ajv@7` while the following packages (or one of theirdependenciess) require `ajv@6`: `postcss-loader`, `file-loader`, `mini-css-extract-plugin`, `raw-loader`

> That dependency should be find out and be fixed that there are no 2 versions installed.

Unlike to PHP, conflicting versions of dependencies are not a problem in the Javascript ecosystem.  npm has a [strategy for installing them](https://medium.com/@imdongchen/how-does-npm-handle-conflicting-package-versions-44f90950cca5) and webpack will [resolve the expected version for the package](https://webpack.js.org/configuration/resolve/#resolvemodules) if configured correctly.

> I guess there is no other option than requiring ajv@7 in the skeleton

I would avoid doing that. Firstly this is wrong semantically, The application inside of `assets/admin` has no dependency on `ajv`. Secondly, with the old configuration, this will break packages that expect to be linked with `ajv@6`.

luca-rath:
Thanks for the detailed explanation! 🙂 All of your arguments look reasonable to me, so I'm okay with that. Just one thing, i don't think requiring `ajv@7` in the skeleton will not break packages expecting `ajv@6`, because those packages would still install `ajv@6` because of the way how npm handles conflicts, like you described before.

niklasnatter:
> i don't think requiring ajv@7 in the skeleton will not break packages expecting ajv@6, because those packages would still install ajv@6 because of the way how npm handles conflicts, like you described before.

Yes, npm will install the correct version. But with the configuration before this change, webpack will resolve the version that was installed in the `assets/admin/node_models` folder instead of the version that was installed for that package 🙂

luca-rath:
ahh right

alexander-schranz:
@nnatter I understand how npm works and handles conflicts but as npm was not build to work with `../../` local dependencies we had in the past problems that npm did install, 2 libraries with the same version one inside `assets/admin/node_modules` and one inside `vendor/sulu/sulu/src/Sulu/Bundle/AdminBundle/Resources/js`. Now you need to make sure that `vendor/sulu/sulu/src/Sulu/Bundle/AdminBundle/Resources/js` or a custom bundle does not accidently had `npm install` run in there directory else  the now new webpack config will load react and mobx from there and you will end into this problems. Webpack is doing a great job merge them together if they are definitely the same version but if your npm install in the vendor would be older then the on in your project would end up that both version which in case of react and mobx isn't good could end in the build.

niklasnatter:
@alexander-schranz But this is the expected behaviour for every project that uses linked packages/workspaces? 🙂 

Also, this behaviour is the reason why we (and other projects) use `peerDependencies` for the `react` and `mobx` package: https://github.com/sulu/sulu/blob/2.x/src/Sulu/Bundle/AdminBundle/Resources/js/package.json#L52-L57
This is also described in the blogpost linked above: https://medium.com/@imdongchen/how-does-npm-handle-conflicting-package-versions-44f90950cca5

> I understand how npm works and handles conflicts but as npm was not build to work with `../../` local dependencies we had in the past problems that npm did install, 2 libraries with the same version one inside `assets/admin/node_modules` and one inside `vendor/sulu/sulu/src/Sulu/Bundle/AdminBundle/Resources/js`

Its possible that npm does not handle the structure of the `sulu/skeleton` perfectly, I have not investigated that in depth. But that is not a good reason to use a webpack configuration that works against the strategies that are used throughout the ecosystem 🙈 

> Now you need to make sure that `vendor/sulu/sulu/src/Sulu/Bundle/AdminBundle/Resources/js` or a custom bundle does not accidently had `npm install` run in there directory

Even with the old configuration we had to remove all `node_modules` folders in the `vendor` directory when switching branches to generate the correct build. So the old configuration had this problem too. 

Also, as said above, if these packages would require `mobx` in `dependencies` (which they dont, they use `peerDependencies` because of this reason), it is the expected behaviour in the JS ecosystem that the required version of `mobx` is linked, not the one that is installed in the `assets/admin/node_modules` folder. Everything els will lead to problems sooner or later.

alexander-schranz:
@nnatter 

> Also, this behaviour is the reason why we (and other projects) use peerDependencies for the react and mobx package: https://github.com/sulu/sulu/blob/2.x/src/Sulu/Bundle/AdminBundle/Resources/js/package.json#L52-L57
This is also described in the blogpost linked above: https://medium.com/@imdongchen/how-does-npm-handle-conflicting-package-versions-44f90950cca5

Yeah but if you have the peerdendency because you did test your bundle or things like that it will use the version from that node_modules folder instead of that.

> Its possible that npm does not handle the structure of the sulu/skeleton perfectly, I have not investigated that in depth. But that is not a good reason to use a webpack configuration that works against the strategies that are used throughout the ecosystem 🙈

Yes we are working again how npm does work here because it did make as problems in the past. That the directory structure is not widely spread in the ecosystem we are aware of and that we did work against it to get sulu running better like this change. 

> Also, as said above, if these packages would require mobx in dependencies (which they dont, they use peerDependencies because of this reason), it is the expected behaviour in the JS ecosystem that the required version of mobx is linked, not the one that is installed in the assets/admin/node_modules folder. Everything els will lead to problems sooner or later.

Yeah but mobx and react are 2 things which definitly say that they only work when there is only 2 version installed and not 2 different. Thats why its recommended to use peerDependency but peerDependency and local packages aren't working so well. 

And we definitely have now about [~20](https://www.diffchecker.com/1NPQLBQx) more dependencies which are duplicated in our build and there was the size increased maybe it wasn't so huge but there was a change:

Bildschirmfoto 2021-02-17 um 12 12 52

niklasnatter:
> Yes we are working again how npm does work here because it did make as problems in the past

Well this workaround caused new problems that are hard to fix without removing the workaround now 🙂 
What problems did it solve in the past? The one that you had to remove the `node_modules` folder i the linked packages?

> Thats why its recommended to use peerDependency but peerDependency and local packages aren't working so well.

But its the widely accepted solution in the JS ecosystem..

> And we definitely have now about ~20 more dependencies which are duplicated in our build and there was the size increased maybe it wasn't so huge but there was a change:

I dont really undestand the diff - do you know which packages were duplicated exactly?
Is it possible that we already resolved the wrong versions of these dependencies and just havent noticed it because it did not result in an obvious error?

alexander-schranz:
> What problems did it solve in the past?

The problems using the main version installed of a library inside the project instead of duplicated version inside the dependencies node_modules folder.

> But its the widely accepted solution in the JS ecosystem

Which doesn't support the way of local dependencies we are using, maybe workspaces would fix all your problems but think we are not near enough there for this.

> I dont really undestand the diff - do you know which packages were duplicated exactly?

The diff should list all duplicated generated by webpack stats.json. 

> Is it possible that we already resolved the wrong versions of these dependencies and just havent noticed it because it did not result in an obvious error?

This is what I want to tell it is not the wrong versions, it is a duplicated library which got injected. Because I know older npm versions did for the local packages install all dependencies in the node_modules folder of the linked package also not only the one which where not in the main directory.

niklasnatter:
> The problems using the main version installed of a library inside the project instead of duplicated version inside the dependencies node_modules folder.

> Which doesn't support the way of local 
dependencies we are using, maybe workspaces would fix all your problems but think we are not near enough there for this.

Okay so the line of reasoning is that our setup is incompatible with `npm` and `npm` will install duplicated dependencies in our linked packages. These duplicated dependencies will then end up in our build. To prevent this, we configure webpack to always use dependencies installed in the root `node_modules` directory (regardless if the package has the right version). Do I understand this right?

- Are you sure that installing duplicated dependencies is still happening? I am not on my work machine anymore so I cant test this right now.
- If duplicated dependencies (same package version) end up in our build, what problems does it cause? 

> This is what I want to tell it is not the wrong versions, it is a duplicated library which got injected

I really dont understand the diff 🙈 Are you sure about that? Do you know which library was injected more than one time?

niklasnatter:
How would you solve the problem of wrongly linked package versions without this change? 
I dont think we can fully prevent having conflicting versions somewhere in our dependency tree.

PR #5830, opened 2021-02-18
Fix tags auto_complete

### LINKED ISSUE #5828: Tags are limited to 10 in the tag selection content type
| Q | A
| --- | ---
| Sulu Version | 2.1.2
| PHP Version | 7.4.9
| DB Version | 5.7.31
| Browser Version | Any browser

# Actual Behavior

Only 10 tags are loaded in a tag selection in the Sulu admin area.

# Expected Behavior

Either there is something like a pagination for tags or the limit of tags that are loaded for a tag selection should be configurable in the xml.

# Steps to Reproduce

1. Add 11 tags to a tag selection in Sulu.
2. Save (all 11 tags will be saved).
3. Reload the page.
4. Only 10 tags will be displayed in the tag selection.
5. If you save the page, those 10 tags will be saved, the 11th is missing afterwards.

luca-rath:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #5828 
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

If specific tag names are passed to the `TagController::cgetAction`, the limit should be set to the amount of tags

#### Why?

Otherwise the autocomplete field in the admin will only show the first 10 tags

### DIFF
--- a/src/Sulu/Bundle/TagBundle/Tests/Functional/Controller/TagControllerTest.php
+++ b/src/Sulu/Bundle/TagBundle/Tests/Functional/Controller/TagControllerTest.php
@@ -130,7 +130,7 @@ class TagControllerTest extends SuluTestCase
         $response = \json_decode($this->client->getResponse()->getContent());
 
         $this->assertEquals(1, $response->total);
-        $this->assertEquals(10, $response->limit);
+        $this->assertEquals(1, $response->limit);
         $this->assertEquals('tag1', $response->_embedded->tags[0]->name);
     }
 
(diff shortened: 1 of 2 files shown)


luca-rath:
The other option would be sending `?limit=123` in the `autocomplete` selection field type

src/Sulu/Bundle/TagBundle/Tests/Functional/Controller/TagControllerTest.php
```diff
@@ -130,9 +130,40 @@ public function testListFilteredByNames()
         $response = \json_decode($this->client->getResponse()->getContent());
 
         $this->assertEquals(1, $response->total);
+        $this->assertEquals(10, $response->limit);
```
luca-rath:
```suggestion
        $this->assertEquals(1, $response->limit);
```

niklasnatter (approved):

niklasnatter:
The `ListRestHelper` does something similar when the `ids` parameter is set. So this option looks good for me 🙂

PR #5833, opened 2021-02-18
Load page teasers from phpcr

### LINKED ISSUE #5785: Teaser Selection does not show secured pages.
| Q | A
| --- | ---
| Sulu Version | 2.2.2

# Actual Behavior

When pages are secured with the website roles and you select them in a teaser selection they wont load any of the page details and in result also not shown on the page even when logged in. I know that it is not that easy because the teaser_selection is saved in the index and not like the other fields.

# Expected Behavior

Also secured pages should be selectable in the teaser_selection

# Steps to Reproduce

Secure some pages with website roles and try to add them to a teaser_selection

luca-rath:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | yes
| Fixed tickets | fixes #5785
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

This pull request refactors the `PageTeaserProvider` to load the pages from `phpcr` instead of the `massive-search` index. This allows a permission check for pages like in the `PageSelection`.

### DIFF
--- a/src/Sulu/Bundle/PageBundle/Teaser/PHPCRPageTeaserProvider.php
+++ b/src/Sulu/Bundle/PageBundle/Teaser/PHPCRPageTeaserProvider.php
@@ -11,10 +11,13 @@
 
 namespace Sulu\Bundle\PageBundle\Teaser;
 
+use Sulu\Bundle\MediaBundle\Api\Media;
 use Sulu\Bundle\PageBundle\Admin\PageAdmin;
 use Sulu\Bundle\PageBundle\Teaser\Configuration\TeaserConfiguration;
 use Sulu\Bundle\PageBundle\Teaser\Provider\TeaserProviderInterface;
 use Sulu\Component\Content\Compat\PropertyParameter;
+use Sulu\Component\Content\Metadata\Factory\StructureMetadataFactoryInterface;
+use Sulu\Component\Content\Metadata\StructureMetadata;
 use Sulu\Component\Content\Query\ContentQueryBuilderInterface;
 use Sulu\Component\Content\Query\ContentQueryExecutorInterface;
 use Sulu\Component\Security\Authorization\PermissionTypes;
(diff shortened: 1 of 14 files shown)


wachterjohannes (approved):

src/Sulu/Bundle/PageBundle/Teaser/PHPCRPageTeaserProvider.php
```diff
@@ -0,0 +1,316 @@
...
+use Sulu\Component\Content\Compat\PropertyParameter;
+use Sulu\Component\Content\Metadata\Factory\StructureMetadataFactoryInterface;
+use Sulu\Component\Content\Metadata\StructureMetadata;
+use Sulu\Component\Content\Query\ContentQueryBuilderInterface;
+use Sulu\Component\Content\Query\ContentQueryExecutorInterface;
+use Sulu\Component\Security\Authorization\PermissionTypes;
+use Symfony\Contracts\Translation\TranslatorInterface;
+
+/**
+ * @deprecated This class will replace the `PageTeaserProvider` and will be renamed to `PageTeaserProvider` afterwards
```
alexander-schranz: We should never deprecate something before the alternative does exist. This confuse developers alot, this is currently the correct teaser provider and so should not be deprecated.

alexander-schranz (changes requested):
Why is this targetting 2.2 instead of 2.1? If its a patch release fix it should also be in `2.1`.

src/Sulu/Bundle/PageBundle/Teaser/PHPCRPageTeaserProvider.php
luca-rath: Okay, so just remove the `@deprecated` but keep the comment? I think this is an important information

luca-rath:
@alexander-schranz This pull request is targeting `2.2`, because #5439 was released with sulu 2.2 and not sulu 2.1

alexander-schranz:
@luca-rath but we just don't need to inject the permissions parameter if I remeber correctly in 2.1. Should not be much work todo this and add in 2.2 the permission parameter again.

src/Sulu/Bundle/PageBundle/Teaser/PHPCRPageTeaserProvider.php
alexander-schranz: Don't see any need for this comment until its really done.

Issue #5821, opened 2021-02-12
Add better handling for PageOutOfBoundsException


alexander-schranz:
# Problem description

Currently the PageOutOfBoundsException does hurt more then it will actually do something. It will throw an exception which no developer can catch easily. 

# Proposed solution

I would remove to throw "PageOutOfBoundsException"  as I see no advantage for it. If you have multiple smart contents over example a block it will throw this exception this smart contents are paginated.

Example:

```xml
<block name="block" minOccurs="2">
     <types>
             <type name="smart">
                   <properties>
                        <property name="pages" type="smart_content" />
                  </properties>
             </type>
     </types>
</block>
```

niklasnatter:
Cannot think about a usecase where the exception is needed, so removing the exception would be okay for me 👍

luca-rath:
Is it likely that someone caught that exception somewhere?

alexander-schranz:
@luca-rath the only place to catch the exception would be a custom controller but then the developer will not be able to display any content as content was never resolved when the exception is thrown, so they could just maybe show a nicer error page then the default once. I don't think that the most know about this exception and that nobody really did ever catch it.

niklasnatter:
Fixed in #5835 🎉

PR #5835, opened 2021-02-22
Remove smart content PageOutOfBoundsException

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes https://github.com/sulu/sulu/issues/5821
| License | MIT

#### What's in this PR?

Removes the `PageOutOfBoundsException` from the SmartContent-Contenttype.

#### Why?

See https://github.com/sulu/sulu/issues/5821


src/Sulu/Component/SmartContent/ContentType.php
```diff
@@ -285,13 +284,9 @@ public function getContentData(PropertyInterface $property)
                 $params,
                 $options,
                 (!empty($limit) ? \intval($limit) : null),
-                $page,
+                $page < 1 ? 1 : ($page > \PHP_INT_MAX ? \PHP_INT_MAX : $page),
```
alexander-schranz: why this change?

Is `page` a string here? would a cast to int not do the same? I think I would do a return (int)  in the getCurrentPage method.

Prokyonn: In the tests we expect `$page` to be a valid integer > 0 and < `PHP_INT_MAX`. Before any `$page` outside these boundaries would have thrown the exception and thus I thought it would also make sense here to show the first page instead of trying to show a negative page value. A negative page value results in a doctrine exception `Offset must be a positive integer or zero`

alexander-schranz: Would still do the logic in the `getCurrentPage` method then.

Prokyonn: Yeah makes more sense, I think we could also omit the `$page > PHP_INT_MAX` check and only check for negative integers.

alexander-schranz: At the end we read a query parameter here and a query parameter will always be a string representation. 

Its just funny that there is a difference in PHP converting a float to int and converting a string to int:

```bash
php -r 'var_dump((int) 9223372036854775808); var_dump((int) "9223372036854775807");';

int(-9223372036854775808)
int(9223372036854775807)
```

So from real world usage the first one can not appear (as query strings are always strings).

I would go in the getCurrentPage with:

```php
if ($page <= 1) {
     return 1;
}

if ( $page > \PHP_INT_MAX) {
      return \PHP_INT_MAX;
}

return (int) $page;
```

alexander-schranz: I would also adopt the tests if **easy possible** that the `$request->query->getParameter(` will return a string value not a float or int. As this matches the real world usage.

With something like:

```php
    ->willReturn((string) $page);
```

src/Sulu/Component/SmartContent/Tests/Unit/ContentTypeTest.php
```diff
@@ -679,7 +673,7 @@ public function testGetContentDataPagedLimit(
             ],
             ['webspaceKey' => 'sulu_io', 'locale' => 'de'],
             $limitResult,
-            $page,
+            $page < 1 ? 1 : ($page > \PHP_INT_MAX ? \PHP_INT_MAX : $page),
```
alexander-schranz, line 676: @Prokyonn as now handled in getCurrentPage I think this can be reverted to $page:

```suggestion
          $page,
```

Prokyonn: We still need this, because from the request we want to have the invalid values, but on the method call we have to check for the valid values. In the tests the `$page` comes directly from the `pageProvider` which generates also the invalid values

alexander-schranz: Oh sorry thought we were here in the ContentType not in the test 🙈

PR #5837, opened 2021-02-23
Replace appveyor with github actions

luca-rath:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

Replace appveyor with github actions

#### Why?

Because appveyor often fails


alexander-schranz (approved):

.github/workflows/test-application.yaml
```diff
@@ -161,3 +161,81 @@ jobs:
...
+            SYMFONY__DATABASE__PASSWORD: root
+
+        strategy:
+            fail-fast: false
+            matrix:
+                include:
+                    - php-version: '7.2'
+                      phpcr-transport: jackrabbit
+                      dependency-versions: 'highest'
+                      test-flags: '--jackrabbit-restart' # restart jackrabbit after each suite see: https://github.com/sulu-io/sulu/issues/2137
```
niklasnatter: does this really work on windows? 😮

luca-rath: I'm not sure, maybe this isn't even needed ... Just had a look at the old `appveyor.yml`, this option wasn't passed there

niklasnatter: i would suspect that at least the following command does not work on windows 🙈 
https://github.com/sulu/sulu/blob/1.6/bin/runtests#L292

so if it works without the flag, i think i would remove it

luca-rath:
```suggestion
```

.github/workflows/test-application.yaml
```diff
@@ -161,3 +161,81 @@ jobs:
...
+                  composer remove jackalope/jackalope-doctrine-dbal --dev --no-interaction --no-update
+                  composer require jackalope/jackalope-jackrabbit:~1.2  --no-interaction --no-update
+
+            - name: Install composer dependencies
+              uses: ramsey/composer-install@v1
+              with:
+                  dependency-versions: ${{ matrix.dependency-versions }}
+
+            - name: Run tests
+              run: php bin/runtests -i -a ${{ matrix.test-flags }}
```
luca-rath:
```suggestion
              run: php bin/runtests -i -a
```

niklasnatter (approved):
🎉

PR #5841, opened 2021-02-24
Replace appveyor with github actions on sulu 2.1

luca-rath:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | #5837
| License | MIT
| Documentation PR | -

#### What's in this PR?

Replace appveyor with github actions


composer.json
```diff
@@ -207,30 +207,30 @@
...
+            "@php bin/adminconsole lint:container --env stage",
+            "@php bin/websiteconsole lint:container --env stage",
+            "@php bin/adminconsole lint:container --env prod",
+            "@php bin/websiteconsole lint:container --env prod"
         ],
         "lint-doctrine": [
-            "bin/adminconsole doctrine:schema:validate --skip-sync",
-            "bin/adminconsole doctrine:ensure-production-settings --env prod"
+            "@php bin/adminconsole doctrine:schema:validate --skip-sync",
+            "@php bin/adminconsole doctrine:ensure-production-settings --env prod"
```
alexander-schranz, line 233: Think this is good as `@php` would also use the same php version which composer was run e.g.:

```
php7.3 composer.phar lint-doctrine
```

will use php7.3 and not then the global php version to run the subcommand.

👍 for this change, we should make this consistent in skeleton the bundles.

alexander-schranz (approved):

composer.json
niklasnatter: sounds good! anybody interested in updating this in all bundles? 😅

luca-rath: not really 🙈 The next one who does something in a bundle can just change this as well

luca-rath: /cc @sulu/core-team

.github/workflows/test-application.yaml
```diff
@@ -167,3 +167,99 @@ jobs:
...
+              with:
+                  dependency-versions: ${{ matrix.dependency-versions }}
+
+            - name: Bootstrap test environment
+              run: composer bootstrap-test-environment
+
+            - name: Execute test cases
+              run: composer test
+
+            - name: Install javascript dependencies
```
niklasnatter: not sure if this is compatible with windows, but we used the `bahmutov/npm-install` action in the other workflow:
https://github.com/sulu/sulu/blob/8f6c63ab7e4e606e3fea04c1d7e1dfc9b738c1b5/.github/workflows/test-application.yaml#L34-L37

niklasnatter (approved):

PR #5848, opened 2021-02-25
Add configuration for sulu_http_cache to enable xkey/purgekey feature

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? |  yes
| BC breaks? |  yes
| Deprecations? | no
| Fixed tickets | none
| Related issues/PRs | none
| License | MIT
| Documentation PR | https://github.com/sulu/sulu-docs/pull/634

#### What's in this PR?

This PR add the configuration to enable XKey/PurgeKeys feature of varnish.

#### Why?

This will allow to use XKey feature of varnish and use it to clear the whole cache. In combination with the GraceMode there is always a HIT and the cache entry will be updated in the background.

#### Example Usage

```vcl
vcl 4.0;

include "/<path-to-project/vendor/sulu/sulu/src/Sulu/Bundle/HttpCacheBundle/Resources/varnish/sulu.vcl";
include "/<path-to-project/vendor/friendsofsymfony/http-cache/resources/config/varnish/fos_tags_xkey.vcl";

acl invalidators {
    "0.0.0.0"/0;
}

backend default {
    .host = "host.docker.internal";
    .port = "13080";
}

sub vcl_recv {
    call fos_tags_xkey_recv;
    call sulu_recv;

    # Force the lookup, the backend must tell not to cache or vary on all
    # headers that are used to build the hash.
    return (hash);
}

sub vcl_backend_response {
     set beresp.grace = 2m;

     call sulu_backend_response;
}

sub vcl_deliver {
    call fos_tags_xkey_deliver;
    call sulu_deliver;
}
```

```yaml
# config/packages/prod/sulu_http_cache.yaml
sulu_http_cache:
    proxy_client:
        varnish:
            enabled: true
            servers:
                - '%env(resolve:VARNISH_SERVER)%'
            tag_mode: purgekeys
    tags:
        enabled: true
```

#### To Do

- [x] Unit Tests


src/Sulu/Bundle/WebsiteBundle/Cache/CacheClearer.php
```diff
@@ -69,17 +70,23 @@ public function __construct(
         $this->eventDispatcher = $eventDispatcher;
     }
 
-    public function clear()
+    public function clear(?string $webspaceKey = null)
     {
         if ($this->cacheManager && $this->cacheManager->supportsInvalidate()) {
+            if ($webspaceKey) {
+                $this->cacheManager->invalidateTag(WebspaceReferenceStore::WEBSPACE_REFERENCE_ALIAS . '-' . $webspaceKey);
```
alexander-schranz: this here is a check missing for `supportsInvalidateTag`

src/Sulu/Bundle/WebsiteBundle/Cache/CacheClearer.php
```diff
@@ -69,17 +70,23 @@ public function __construct(
         $this->eventDispatcher = $eventDispatcher;
     }
 
-    public function clear()
+    public function clear(?string $webspaceKey = null)
```
alexander-schranz: Not sure about that the cache clearer. Should we provide a `tag` array instead to be more flexible here.

```
    public function clear(/*?array $tags = []*/)
```

When the webspacekey is given this would actually support that we can invalidate all domains of the wespace and not just `$request->getHost()` which could actually not being the domain when sulu admin runs on a seperate domain.

BC layer need to implemented here also with `func_get_arg`.

src/Sulu/Bundle/WebsiteBundle/Cache/CacheClearer.php
```diff
@@ -69,17 +70,23 @@ public function __construct(
...
             if (!$request) {
                 return;
             }
 
             $this->cacheManager->invalidateDomain($request->getHost());
-
-            $this->eventDispatcher->dispatch(new CacheClearEvent(), Events::CACHE_CLEAR);
+            $this->eventDispatcher->dispatch(new CacheClearEvent($webspaceKey), Events::CACHE_CLEAR);
 
             return;
```
alexander-schranz, line 102: interesting return 👀

src/Sulu/Bundle/HttpCacheBundle/DependencyInjection/Configuration.php
```diff
@@ -85,6 +85,14 @@ public function getConfigTreeBuilder()
                                     ->defaultNull()
                                     ->info('Default host name and optional path for path based invalidation.')
                                 ->end()
+                                ->enumNode('tag_mode')
+                                    ->info('If you can enable the xkey module in Varnish, use the purgekeys mode for more efficient tag handling')
+                                    ->values(['ban', 'purgekeys'])
+                                    ->defaultValue('ban')
+                                ->end()
+                                ->scalarNode('tags_header')
```
alexander-schranz, line 93: I more and more for it that we will remove this config in sulu 3.0 and let the dev configure directly define foshttpcachebundle. But at current state okay for me.

src/Sulu/Bundle/WebsiteBundle/Cache/CacheClearerInterface.php
```diff
@@ -19,5 +19,5 @@ interface CacheClearerInterface
     /**
      * Clear the website cache.
      */
-    public function clear();
+    public function clear(/* ?string $webspaceKey = null */);
```
alexander-schranz: same need to be done in the implemented class as the class is not final and could be extended

src/Sulu/Bundle/WebsiteBundle/ReferenceStore/WebspaceReferenceStore.php
```diff
@@ -0,0 +1,48 @@
...
+    private $requestAnalyzer;
+
+    public function __construct(?RequestAnalyzerInterface $requestAnalyzer = null)
+    {
+        $this->requestAnalyzer = $requestAnalyzer;
+    }
+
+    public function add($id)
+    {
+        // wont be called
```
alexander-schranz: throw exception `new \LogicException('Should never be called');` then?

src/Sulu/Bundle/WebsiteBundle/ReferenceStore/WebspaceReferenceStore.php
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\WebsiteBundle\ReferenceStore;
+
+use Sulu\Component\Webspace\Analyzer\RequestAnalyzerInterface;
+
+class WebspaceReferenceStore implements ReferenceStoreInterface
```
alexander-schranz, line 17: Add unit test class for this

src/Sulu/Bundle/WebsiteBundle/Controller/CacheController.php
```diff
@@ -59,7 +59,8 @@ public function clearAction()
             return new JsonResponse(null, 403);
         }
 
-        $this->cacheClearer->clear();
+        // TODO webspace-key
+        $this->cacheClearer->clear('sulu');
```
alexander-schranz: what about:

```
$this->cacheClearer->clear(['webspace-sulu']);
```

So only specific x-tags could be cleared?

src/Sulu/Bundle/WebsiteBundle/Controller/CacheController.php
```diff
@@ -59,7 +59,8 @@ public function clearAction()
             return new JsonResponse(null, 403);
         }
 
-        $this->cacheClearer->clear();
+        // TODO webspace-key
```
alexander-schranz: Another way would be that the purge is like yet on the `domain` instead of the webspace key but would prefer on webspace. Think the clear button would just need to forward it as a post or get parameter.

src/Sulu/Bundle/WebsiteBundle/Resources/config/services.xml
```diff
@@ -292,6 +292,13 @@
             <argument type="collection"/>
         </service>
 
+        <service id="sulu_website.reference_store.webspace"
```
alexander-schranz: we did go for new service mostly with one point only for the bundle prefix for the service name `sulu_website.webspace_reference_store` but not important for me when the other reference stores follow the same schema currently.

src/Sulu/Bundle/WebsiteBundle/Controller/CacheController.php
alexander-schranz: When we have the webspace key the above thing can also be changed:

```
if (!$this->checkLivePermissionForAllWebspaces()) {
````

**Review by alexander-schranz — COMMENTED** [review]
Did have a first look at this and added a few comments!

File: src/Sulu/Bundle/WebsiteBundle/Cache/CacheClearer.php (hunk 3 above)
Comment (wachterjohannes) [reply src/Sulu/Bundle/WebsiteBundle/Cache/CacheClearer.php:102]: why? we need it - dont want to delete the folder also

File: src/Sulu/Bundle/WebsiteBundle/Controller/CacheController.php (hunk 8 above)
Comment (wachterjohannes) [reply src/Sulu/Bundle/WebsiteBundle/Controller/CacheController.php]: 👍

File: src/Sulu/Bundle/WebsiteBundle/Resources/config/services.xml (hunk 10 above)
Comment (wachterjohannes) [reply src/Sulu/Bundle/WebsiteBundle/Resources/config/services.xml]: this service names are never used - so i will change it

File: src/Sulu/Bundle/HttpCacheBundle/DependencyInjection/Configuration.php (hunk 4 above)
Comment (wachterjohannes) [reply src/Sulu/Bundle/HttpCacheBundle/DependencyInjection/Configuration.php:93]: i am not sure - we can then here define which features works well - see this PR we have to change things to allow this configuration - but i think we have to discuss this in a bigger round

File: src/Sulu/Bundle/WebsiteBundle/Cache/CacheClearer.php (hunk 3 above)
Comment (alexander-schranz) [reply src/Sulu/Bundle/WebsiteBundle/Cache/CacheClearer.php:102]: oh thought it was end of the function  😄

File: src/Sulu/Bundle/WebsiteBundle/Controller/AnalyticsController.php [hunk 11]
```diff
@@ -110,7 +110,7 @@ public function postAction(Request $request, $webspace)
 
         $entity = $this->analyticsManager->create($webspace, $data);
         $this->entityManager->flush();
-        $this->cacheClearer->clear();
+        $this->cacheClearer->clear($webspace);
```
alexander-schranz: think should be an array now or?

wachterjohannes: 👍

wachterjohannes:
@alexander-schranz @nnatter can be reviewed now in detail - dont froget the docs PR

src/Sulu/Bundle/HttpCacheBundle/Cache/CacheManager.php
```diff
@@ -79,4 +79,9 @@ public function supportsInvalidate(): bool
     {
         return $this->fosCacheManager->supports(FOSCacheManager::INVALIDATE);
     }
+
+    public function supportsInvalidateTag(): bool
```
niklasnatter: why break from the `FOSCacheManager` naming?

```suggestion
    public function supportsTags(): bool
```

src/Sulu/Bundle/HttpCacheBundle/Cache/CacheManagerInterface.php
```diff
@@ -37,4 +37,9 @@ public function invalidateReference(string $alias, string $id): void;
      * Returns true if current proxy client supports invalidation.
      */
     public function supportsInvalidate(): bool;
+
+    /**
+     * Returns true if current proxy client supports tags-invalidation.
+     */
+    public function supportsInvalidateTag(): bool;
```
niklasnatter: see above

```suggestion
    public function supportsTags(): bool;
```

src/Sulu/Bundle/HttpCacheBundle/DependencyInjection/Configuration.php
```diff
@@ -85,6 +85,14 @@ public function getConfigTreeBuilder()
                                     ->defaultNull()
                                     ->info('Default host name and optional path for path based invalidation.')
                                 ->end()
+                                ->enumNode('tag_mode')
+                                    ->info('If you can enable the xkey module in Varnish, use the purgekeys mode for more efficient tag handling')
```
niklasnatter:
```suggestion
                                    ->info('Use the purgekeys mode for more efficient tag handling, if your Varnish server supports the xkey module')
```

src/Sulu/Bundle/HttpCacheBundle/Tests/Unit/Cache/CacheManagerTest.php
```diff
@@ -36,6 +36,20 @@ public function setUp(): void
...
     }
 
+    public function testSupportsInvalidate()
+    {
+        // proxy client doesn't support tag invalidation
+        $this->fosCacheManager->supports(FOSCacheManager::INVALIDATE)->willReturn(true);
+        $this->assertTrue($this->cacheManager->supportsInvalidate());
+    }
+
+    public function testSupportsInvalidateTag()
```
niklasnatter:
```suggestion
    public function testSupportsTags()
```

src/Sulu/Bundle/WebsiteBundle/Cache/CacheClearer.php
```diff
@@ -69,17 +69,28 @@ public function __construct(
...
+
         if ($this->cacheManager && $this->cacheManager->supportsInvalidate()) {
+            if (!empty($tags) && $this->cacheManager->supportsInvalidateTag()) {
+                foreach ($tags as $tag) {
+                    $this->cacheManager->invalidateTag($tag);
+                }
+
+                $this->eventDispatcher->dispatch(new CacheClearEvent($tags), Events::CACHE_CLEAR);
+
+                return;
```
niklasnatter, line 91: would move the code below into a `else` block. that way its easier to understand that we either call `invalidateTag` or `invalidateDomain`

alexander-schranz (approved):

src/Sulu/Bundle/WebsiteBundle/Cache/CacheClearer.php
```diff
@@ -69,17 +69,28 @@ public function __construct(
         $this->eventDispatcher = $eventDispatcher;
     }
 
-    public function clear()
+    public function clear(/*?array $tags = []*/)
     {
+        $tags = \func_num_args() >= 1 ? \func_get_arg(0) : null;
+
         if ($this->cacheManager && $this->cacheManager->supportsInvalidate()) {
+            if (!empty($tags) && $this->cacheManager->supportsInvalidateTag()) {
```
niklasnatter: i think i would expect that nothing is cleared if i pass an empty array here

src/Sulu/Bundle/WebsiteBundle/Controller/CacheController.php
```diff
@@ -53,13 +55,21 @@ public function __construct(
...
+            return new JsonResponse(null, 403);
+        }
+
+        $tags = [];
+        if ($webspaceKey) {
+            $tags[] = \sprintf('%s-%s', WebspaceReferenceStore::WEBSPACE_REFERENCE_ALIAS, $webspaceKey);
         }
 
-        $this->cacheClearer->clear();
+        $this->cacheClearer->clear($tags);
```
niklasnatter: its strange to pass an empty array here to clear the cache for all webspaces, isnt it? 
i would pass `null` to clear everything

src/Sulu/Bundle/WebsiteBundle/ReferenceStore/WebspaceReferenceStore.php
```diff
@@ -0,0 +1,48 @@
...
+    private $requestAnalyzer;
+
+    public function __construct(?RequestAnalyzerInterface $requestAnalyzer = null)
+    {
+        $this->requestAnalyzer = $requestAnalyzer;
+    }
+
+    public function add($id)
+    {
+        throw new \LogicException('Should never be called');
```
niklasnatter:
```suggestion
        throw new \LogicException('Webspace tags cannot be set manually. They are set to match the current webspace automatically.');
```

src/Sulu/Bundle/WebsiteBundle/ReferenceStore/WebspaceReferenceStore.php
```diff
@@ -0,0 +1,48 @@
...
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\WebsiteBundle\ReferenceStore;
+
+use Sulu\Component\Webspace\Analyzer\RequestAnalyzerInterface;
+
+class WebspaceReferenceStore implements ReferenceStoreInterface
+{
+    const WEBSPACE_REFERENCE_ALIAS = 'webspace';
```
niklasnatter, line 19: wouldnt it be easier to provide a static method that returns the tag for a given `webspackeKey` instead of building the tag using this constant in every service?

src/Sulu/Bundle/WebsiteBundle/Resources/config/services.xml
```diff
@@ -292,6 +292,13 @@
             <argument type="collection"/>
         </service>
 
+        <service id="sulu_website.webspace_reference_store"
+                 class="Sulu\Bundle\WebsiteBundle\ReferenceStore\WebspaceReferenceStore">
+            <argument type="service" id="sulu_core.webspace.request_analyzer" on-invalid="null"/>
+
+            <tag name="sulu_website.reference_store" alias="webspace"/>
```
niklasnatter, line 300: can we use the constant for the `alias` value here?

src/Sulu/Bundle/WebsiteBundle/Resources/js/containers/CacheClearToolbarAction/CacheClearToolbarAction.js
```diff
@@ -8,9 +8,14 @@ import {translate} from 'sulu-admin-bundle/utils';
 export default class CacheClearToolbarAction {
     static clearCacheEndpoint: string;
 
+    searchParams: Object;
```
niklasnatter: we called this `requestParameters` in nearly all other places. would like to keep this consistent

src/Sulu/Bundle/WebsiteBundle/Resources/js/containers/CacheClearToolbarAction/CacheClearToolbarAction.js
```diff
@@ -44,7 +49,13 @@ export default class CacheClearToolbarAction {
 
     @action handleConfirm = () => {
         this.cacheClearing = true;
-        Requester.delete(CacheClearToolbarAction.clearCacheEndpoint).then(action(() => {
+
+        let url = CacheClearToolbarAction.clearCacheEndpoint;
+        if (this.searchParams) {
+            url += '?' + (new URLSearchParams(this.searchParams)).toString();
```
niklasnatter: there is a `buildQueryString` util that we use for this in other places

src/Sulu/Bundle/WebsiteBundle/Cache/CacheClearer.php
alexander-schranz: is fine for me but we would need to check if `$tags = null` because then all should be cleared

src/Sulu/Bundle/WebsiteBundle/Cache/CacheClearer.php
alexander-schranz: Not sure with an else here I think else block make it less readable then the early return we mostly did go with early returns instead of else statements.

src/Sulu/Bundle/WebsiteBundle/Controller/CacheController.php
alexander-schranz:
```suggestion
        $this->cacheClearer->clear(\empty($tags) ? null : $tags);
```

src/Sulu/Bundle/WebsiteBundle/Cache/CacheClearer.php
niklasnatter: dont have a too strong opinion on this, we can leave it like this too 🙂

src/Sulu/Bundle/WebsiteBundle/Cache/CacheClearer.php
```diff
@@ -69,17 +69,28 @@ public function __construct(
         $this->eventDispatcher = $eventDispatcher;
     }
 
-    public function clear()
+    public function clear(/*?array $tags = []*/)
```
niklasnatter: would be okay for me to add this without a comment and document it in the `UPGRADE.md` because it will be released in a minor release (not sure what @alexander-schranz thinks about that). 

but if we add it like this, we should trigger a deprecation if the param is not given. otherwise we will forget about changing it in the next major version

src/Sulu/Bundle/HttpCacheBundle/DependencyInjection/Configuration.php
wachterjohannes: copied from the library :) but will change it

src/Sulu/Bundle/WebsiteBundle/Cache/CacheClearer.php
wachterjohannes: i am of the same opinion as @alexander-schranz - will keep it

src/Sulu/Bundle/WebsiteBundle/Resources/config/services.xml
wachterjohannes: thats not possible

wachterjohannes:
@alexander-schranz @nnatter fixed

src/Sulu/Bundle/HttpCacheBundle/Tests/Unit/Cache/CacheManagerTest.php
```diff
@@ -36,7 +36,21 @@ public function setUp(): void
         $this->cacheManager = new CacheManager($this->fosCacheManager->reveal());
     }
 
-    public function testInvalidateTag()
+    public function testSupportsInvalidate(): void
+    {
+        // proxy client doesn't support tag invalidation
```
niklasnatter: not sure what this comment means? 😇

niklasnatter (approved):

niklasnatter:
Think you need to rebase the branch and update the `phpstan-baseline.neon` file with `vendor/bin/phpstan analyse --generate-baseline` 🙂

src/Sulu/Bundle/HttpCacheBundle/Tests/Unit/Cache/CacheManagerTest.php
wachterjohannes: that the prox client does not support the tag invalidation 😂  not sure how i should write that different 🙈

wachterjohannes:
@nnatter seems not to help have rebased and generated the baseline - locally php has no errors

Prokyonn:
@wachterjohannes you have to use php8 locally to generate the baseline

wachterjohannes:
@Prokyonn thanks 🙈  hopefully now it works

src/Sulu/Bundle/HttpCacheBundle/Tests/Unit/Cache/CacheManagerTest.php
niklasnatter: but why does this matter for this test case? is the proxy client even used? 🤔

wachterjohannes:
@nnatter @alexander-schranz should go to green now

src/Sulu/Bundle/HttpCacheBundle/Tests/Unit/Cache/CacheManagerTest.php
niklasnatter: @wachterjohannes ?

niklasnatter (approved):
PHPStan error is unrelated to this PR and already fixed on the base branch

Issue #5852, opened 2021-03-01
Add postinstall script to validate ckeditor plugin versions.


alexander-schranz:
# Problem description

Currently when a project add a ckeditor plugin like [Font](https://github.com/sulu/sulu/blob/60be7bdf58fa71ac807ca8a61332b7e3aad365c7/src/Sulu/Bundle/AdminBundle/Resources/js/containers/CKEditor5/CKEditor5.js#L161-L176). They need to be sure that the ckeditor plugin is installed in the version sulu does support else it could end in strange errors.

# Proposed solution

Like the current `preinstall` script add a `postinstall` script which validates that all installed `@ckeditor` packages matches the same version "major" version, which the `sulu-admin-bundle` requires.

PR #5856, opened 2021-03-04
Add pre and post install script for symlink

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no 
| BC breaks? | no 
| Deprecations? | no  
| Fixed tickets | fixes # 
| Related issues/PRs | https://github.com/sulu/skeleton/pull/96 Replaced by #5859
| License | MIT

#### What's in this PR?

Add pre and post install script for symlink.

#### Why?

Fixes current problems with CKEditor https://github.com/sulu/sulu/issues/5852. Problems with NPM 7: https://github.com/sulu/skeleton/issues/88 and hopefully issues with the windows build also: https://github.com/sulu/skeleton/issues/68


UPGRADE.md
```diff
@@ -2,6 +2,10 @@
 
 ## 2.1.9
 
+### A new argument `$requestStack` has been added to the `ContentTwigExtension`
+
+Instantiating a `ContentTwigExtension` without the `$requestStack` argument is deprecated.
+
```
luca-rath: Why did you add this? It's already [here](https://github.com/sulu/sulu/blob/2.1/UPGRADE.md)

luca-rath: Ah, this is not the only wrong thing in here 😂

preinstall.js
```diff
@@ -3,3 +3,5 @@
 if (process.env.npm_execpath.indexOf('npm') === -1) {
     throw new Error('\x1b[31mYou must use "npm install", yarn is not supported\x1b[0m');
 }
+
+require('./symlink.js');
```
niklasnatter, line 10: think we should only require this if the script is executed inside of the `assets/admin` folder

postinstall.js
```diff
@@ -0,0 +1,4 @@
+/* eslint-disable flowtype/require-valid-file-annotation */
+
+require('./symlink.js');
```
niklasnatter, line 3: think we should only require this if the script is executed inside of the `assets/admin` folder

webpack.config.js
```diff
@@ -3,7 +3,6 @@
 /* eslint-disable import/no-dynamic-require */
 const fs = require('fs');
 const path = require('path');
-const babelConfig = JSON.parse(fs.readFileSync(path.resolve(__dirname, '.babelrc'))); // eslint-disable-line no-undef
```
niklasnatter, line 6: would use a `babel.config.js` here too

niklasnatter: and move the `postcss.config.cs` into the bundles 🙂

alexander-schranz:
Replaced by #5859

PR #5859, opened 2021-03-08
Add symlink to vendor directory in node_modules directory in preinstall and postinstall script

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no 
| BC breaks? | no 
| Deprecations? | no  
| Fixed tickets | fixes # 
| Related issues/PRs | https://github.com/sulu/skeleton/pull/96
| License | MIT

#### What's in this PR?

Add pre and post install script for symlink.

#### Why?

Fixes current problems with CKEditor https://github.com/sulu/sulu/issues/5852 and problems with NPM 7: https://github.com/sulu/skeleton/issues/88


symlink.js
```diff
@@ -0,0 +1,25 @@
+/* eslint-disable flowtype/require-valid-file-annotation */
+/* eslint-disable import/no-nodejs-modules */
+
+const {exec} = require('child_process'); //
+const path = require('path');
+
+const parameters = process.argv.slice(2);
+
+const from = parameters.length > 1 ? parameters[0] : '../../vendor';
```
niklasnatter: is there a reason for making these configurale? can we pass an argument when requiring the script in another script? 🤔

symlink.js
```diff
@@ -0,0 +1,25 @@
+/* eslint-disable flowtype/require-valid-file-annotation */
```
niklasnatter: think i would be specific and call the file `symlink-vendor-directory.js` 🙂

styleguide.config.js
```diff
@@ -151,7 +149,10 @@ module.exports = { // eslint-disable-line
                     exclude: /node_modules\/(?!(sulu-(.*)-bundle|@ckeditor|lodash-es)\/)/,
                     use: {
                         loader: 'babel-loader',
-                        options: babelConfig,
+                        options: {
```
niklasnatter, line 152: 🎉

symlink-vendor-directory.js
```diff
@@ -0,0 +1,25 @@
...
+/* eslint-disable import/no-nodejs-modules */
+
+const {exec} = require('child_process'); //
+const path = require('path');
+
+const parameters = process.argv.slice(2);
+
+const from = parameters.length > 1 ? parameters[0] : '../../vendor';
+const to = parameters.length > 2 ? parameters[1] : 'node_modules/@sulu/vendor';
+const ignoreCheck = parameters.length > 3 && parameters[2] === 'ignoreCheck';
```
niklasnatter: think i would call this `--force` if we want to make this configurable 🙂 
```suggestion
const doNotCheckCurrentDirectory = parameters.length > 3 && parameters[2] === '--force';
```

webpack.config.js
```diff
@@ -59,6 +58,16 @@ module.exports = (env, argv) => { // eslint-disable-line no-undef
             new webpack.DefinePlugin({
                 SULU_ADMIN_BUILD_VERSION: JSON.stringify(suluVersion),
             }),
+            new webpack.NormalModuleReplacementPlugin(/^@ckeditor/, (resource) => {
```
niklasnatter: would keep this only if we need it for windows compatibility

webpack.config.js
```diff
@@ -59,6 +58,16 @@ module.exports = (env, argv) => { // eslint-disable-line no-undef
...
+                // npm might install the @ckeditor/* packages multiple times if they are required in a project because
+                // the assets/admin folder is not a parent of the vendor folder. because of this, the packages might be
+                // included in the build multiple times and throw an error. to prevent such duplicated packages in the
+                // build, we load the package that is installed in the assets/admin/node_modules folder if possible
```
niklasnatter:
```suggestion
                // when using windows, npm installs the @ckeditor/* packages multiple times if they are required in a 
                // project because the assets/admin folder is not a parent of the vendor folder. unfortunately, the 
                // node_modules/@sulu/vendor symlink does not prevent this behaviour on windows systems.
                // because of this, the packages are included in the build multiple times and throw an error. 
                // to prevent such duplicated packages in the build, we load the package that is installed in the 
                // assets/admin/node_modules folder if possible
```

symlink-vendor-directory.js
```diff
@@ -0,0 +1,25 @@
+/* eslint-disable flowtype/require-valid-file-annotation */
+/* eslint-disable import/no-nodejs-modules */
+
+const {exec} = require('child_process'); //
+const path = require('path');
+
+const parameters = process.argv.slice(2);
```
niklasnatter, line 24:
```suggestion
// the sulu/skeleton uses the assets/admin directory as root directory for the administration interface javascript
// application. this application imports the sulu/sulu code by requiring the bundles from the the vendor directory
// in its package.json.
//
// unfortunately this setup causes various problems in the javascript ecosystem because the assets/admin directory
// is not an ancestor of the vendor/sulu/sulu directory. for example, npm 7 does not install the dependencies of the
// bundle packages (https://github.com/npm/cli/issues/2339) and older npm versions are not able to dedupe packages that
// are required by a bundle and in the assets/admin directory.
// because of this, packages that were not deduped might be included in the webpack build multiple times. this
// increases the size of the build and will lead to an error in case of the @ckeditor packages:
// https://ckeditor.com/docs/ckeditor5/latest/framework/guides/support/error-codes.html#error-ckeditor-duplicated-modules
//
// to prevent these errors, this file creates a assets/admin/node_modules/node_modules/@sulu/vendor symlink and use
// the symlink to require the sulu bundles in the assets/admin/package.json. this makes the vendor directory a child
// of the assets/admin directory and allows npm to correctly dedupe the installed packages.

const parameters = process.argv.slice(2);
```

symlink-vendor-directory.js
```diff
@@ -0,0 +1,25 @@
...
+
+const from = parameters.length > 1 ? parameters[0] : '../../vendor';
+const to = parameters.length > 2 ? parameters[1] : 'node_modules/@sulu/vendor';
+const ignoreCheck = parameters.length > 3 && parameters[2] === 'ignoreCheck';
+
+if (
+    ignoreCheck
```
niklasnatter:
```suggestion
    doNotCheckCurrentDirectory
```

src/Sulu/Bundle/AdminBundle/Command/UpdateBuildCommand.php
```diff
@@ -101,7 +101,8 @@ protected function execute(InputInterface $input, OutputInterface $output)
             'index.js',
             'package.json',
             'webpack.config.js',
-            '.babelrc',
+            'babel.config.json',
```
alexander-schranz, line 105: @nnatter I think we need to rename `.babelrc` if it exists to `babel.config.json`

symlink-vendor-directory.js
```diff
@@ -0,0 +1,41 @@
...
+// is not an ancestor of the vendor/sulu/sulu directory. for example, npm 7 does not install the dependencies of the
+// bundle packages (https://github.com/npm/cli/issues/2339) and older npm versions are not able to dedupe packages that
+// are required by a bundle and in the assets/admin directory.
+// because of this, packages that were not deduped might be included in the webpack build multiple times. this
+// increases the size of the build and will lead to an error in case of the @ckeditor packages:
+// https://ckeditor.com/docs/ckeditor5/latest/framework/guides/support/error-codes.html#error-ckeditor-duplicated-modules
+//
+// to prevent these errors, this file creates a assets/admin/node_modules/node_modules/@sulu/vendor symlink and use
+// the symlink to require the sulu bundles in the assets/admin/package.json. this makes the vendor directory a child
+// of the assets/admin directory and allows npm to correctly dedupe the installed packages.
```
niklasnatter:
```suggestion
// the sulu/skeleton uses the assets/admin directory as root directory for the administration interface javascript
// application. this application imports the sulu/sulu code by requiring the bundles from the the vendor directory
// in its package.json.
//
// unfortunately this setup causes various problems in the javascript ecosystem because the assets/admin directory
// is not an ancestor of the vendor/sulu/sulu directory which is used to require the code of bundles. for example,
// npm 7 does not install the dependencies of the bundle packages (https://github.com/npm/cli/issues/2339) and older
// npm versions are not able to dedupe packages that are required by a bundle and in the assets/admin directory.
// because of this, packages that were not deduped might be included in the webpack build multiple times. this
// increases the size of the build and will lead to an error in case of the @ckeditor packages:
// https://ckeditor.com/docs/ckeditor5/latest/framework/guides/support/error-codes.html#error-ckeditor-duplicated-modules
//
// to prevent these problems, this file creates a assets/admin/node_modules/node_modules/@sulu/vendor symlink and we
// use the symlinked vendor directory to require the bundles in the assets/admin/package.json.
// this makes the directory that is used for requiring the bundles a descendant of the assets/admin directory and
// allows npm to correctly dedupe the installed packages.
```

symlink-vendor-directory.js
```diff
@@ -0,0 +1,42 @@
...
+// application. this application imports the sulu/sulu code by requiring the bundles from the the vendor directory
+// in its package.json.
+//
+// unfortunately this setup causes various problems in the javascript ecosystem because the assets/admin directory
+// is not an ancestor of the vendor/sulu/sulu directory which is used to require the code of bundles. for example,
+// npm 7 does not install the dependencies of the bundle packages (https://github.com/npm/cli/issues/2339) and older
+// npm versions are not able to dedupe packages that are required by a bundle and in the assets/admin directory.
+// because of this, packages that were not deduped might be included in the webpack build multiple times. this
+// increases the size of the build and will lead to an error in case of the @ckeditor packages:
+// https://ckeditor.com/docs/ckeditor5/latest/framework/guides/support/error-codes.html#error-ckeditor-duplicated-modules
```
niklasnatter:
```suggestion
// http://ckeditor.com/docs/ckeditor5/latest/framework/guides/support/error-codes.html#error-ckeditor-duplicated-modules
```

src/Sulu/Bundle/AdminBundle/Command/UpdateBuildCommand.php
```diff
@@ -101,10 +102,25 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
             'postcss.config.js',
         ];
 
+        $renamedFiles = [
+            '.babelrc' => 'babel.config.json',
+        ];
+
+        foreach ($renamedFiles as $oldFile => $newFile){
+            if (!$filesystem->exists($this->projectDir . $oldFile)) {
+                if ('y' !== \strtolower(
```
niklasnatter:
```suggestion
                if ('y' === \strtolower(
```

niklasnatter (approved):

symlink.js
alexander-schranz: yes if somebody have a different directory structure it should not fail or atleast he should be able to change it:
works for me correctly:

```
        "preinstall": "node ../../vendor/sulu/sulu/preinstall.js ../../vendor node_modules/@sulu/vendor --force",
```

PR #5864, opened 2021-03-09
Adjust smart_content sortBy migration to not check for limitResult key

niklasnatter:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| License | MIT

#### Why?

There we two independent reports that the check for `limitResult` caused problems when updating a project from 1.6 to 2.0 in our Slack channel. Also, the migration does not migrate `smart_content` properties on the homepage at the moment.


alexander-schranz (approved):

alexander-schranz (approved):
We need todo the same in the SuluArticleBundle

niklasnatter:
Done in https://github.com/sulu/SuluArticleBundle/pull/552 🙂

PR #5877, opened 2021-03-12
Allow to define empty option for single_select property in form configuration

niklasnatter:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This PR adjusts the `SingleSelect` field-type to allow for defining an empty option in the form configuration xml like shown below. The empty option will appear as selected when a new page is created and allows to "unselect" the selected option later on.

```xml
<property name="single_select" type="single_select">
    <meta>
        <title lang="en">Single Select</title>
        <title lang="de">Einzelauswahl</title>
    </meta>

    <params>
            <param name="">
                <meta>
                    <title lang="en">No selection</title>
                    <title lang="de">Keine Auswahl</title>
                </meta>
            </param>
            <param name="value-1">
                <meta>
                    <title lang="en">Value 1</title>
                    <title lang="de">Wert 1</title>
                </meta>
            </param>
            <param name="value-2">
                <meta>
                    <title lang="en">Value 2</title>
                    <title lang="de">Wert 2</title>
                </meta>
            </param>
        </param>
    </params>
</property>
```

#### Why?

At the moment, it is not possible to define an empty option in the form configuration xml. The `single_select` property will always show `Please choose` for new pages if no default value is set.

### DIFF
--- a/config/templates/pages/example.xml
+++ b/config/templates/pages/example.xml
@@ -116,19 +116,25 @@
 
             <params>
                 <param name="values" type="collection">
+                    <param name="">
+                        <meta>
+                            <title lang="en">No selection</title>
+                            <title lang="de">Keine Auswahl</title>
+                        </meta>
+                    </param>
                     <param name="value-1">
                         <meta>
                             <title lang="en">Value 1</title>
                             <title lang="de">Wert 1</title>
                         </meta>
                     </param>
-                    <param name="value-2" value="Value 2">
+                    <param name="value-2">
                         <meta>
                             <title lang="en">Value 2</title>
                             <title lang="de">Wert 2</title>
                         </meta>
                     </param>
-                    <param name="value-3" value="Value 3">
+                    <param name="value-3">
                         <meta>
                             <title lang="en">Value 3</title>
                             <title lang="de">Wert 3</title>
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/SingleSelect.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/SingleSelect.js
@@ -16,7 +16,7 @@ export default class SingleSelect extends React.Component<FieldTypeProps<string 
             } = {},
         } = schemaOptions;
 
-        if (defaultValue === undefined || defaultValue === null) {
+        if (defaultValue === undefined || defaultValue === null || defaultValue === '') {
             return;
         }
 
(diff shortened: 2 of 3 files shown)


config/templates/pages/example.xml
```diff
@@ -116,19 +116,25 @@
 
             <params>
                 <param name="values" type="collection">
+                    <param name="">
```
alexander-schranz, line 119: I think this option should not be needed to be set by the dev itself. It should always be there if there is no `default_value` set.

niklasnatter: yeah thats a valid point. havent thought about that a lot, but i think we should discuss to do it in a separate pr 🙂  
but at the moment it is not possible to create such an option at all, therefore it would be nice to merge this anyway

niklasnatter: its also okay for me to remove the option in the `example.xml` and just adjust the javascript to handle `''` as `null` in this PR 🙂  
i think we need to do that anyway to allow for setting a custom text for the `null` option

alexander-schranz (approved):

config/templates/pages/example.xml
alexander-schranz: is okay for me to keep the `example.xml` at current state.

DanielRuf:
This should be documented. So far the current documentation does not describe this.

https://docs.sulu.io/en/2.5/reference/content-types/single_select.html

config/templates/pages/example.xml
DanielRuf: > It should always be there if there is no default_value set.

At the moment if no `default_value` param is used, no option for unselecting is possible. So I guess this comment is obsolete?

niklasnatter:
@DanielRuf You are right. It would be great if you could create a pull request on https://github.com/sulu/sulu-docs

DanielRuf:
> @DanielRuf You are right. It would be great if you could create a pull request on https://github.com/sulu/sulu-docs

Oops, sorry.

Deleted my comment since it works now.

So currently this is not documented:

> if you do not add the `default_value` param, Sulu will show a `Default` select option (as last option) with an empty value to unselect the currently selected option

Or similar, the wording is not that good.

DanielRuf:
> @DanielRuf You are right. It would be great if you could create a pull request on https://github.com/sulu/sulu-docs

Oops, sorry.

Deleted my comment since it works like in your example

```
                                <param name="">
                                    <meta>
                                        <title lang="en">Default</title>
                                    </meta>
                                </param>
```

So currently this feature is currently not there yet:

> I think this option should not be needed to be set by the dev itself. It should always be there if there is no default_value set.

PR #5878, opened 2021-03-12
Display title of internal link instead of link destination in webspace overview

### LINKED ISSUE #3721: Internal links: use internal title instead of redirect title in the content tree
| Q | A
| --- | ---
| Bug? | yes
| New Feature? | no
| Sulu Version | 1.6.13

#### Actual Behavior

When browsing the content tree, internal links use the title of their respective redirect target. 

#### Expected Behavior

Internal links should use their own title ("internal-title") in the nav tree. The link icon will still be visible and thus identify the node type ("link").

#### Steps to Reproduce

Create internal link in the content tree. Add subcontents to this link node. Internal link will show the title of its target in the nav tree. It will be very hard for content editors to understand why the subcontents are arranged under a node with a different name.

### Example

Imagine a content structure like this one:

Contact
Contact/Form
Contact/Contacts 
Contact/Map

Contact is an empty structure node, configured as internal link to "Form" and therefore use "Form" as its title. It will not make sense for content editors, why Form, Contacts and Map are subcontents of a node named "Form".

### LINKED ISSUE #3877: Title of assigned pages in InternalLinks FieldType is not consistent
| Q | A
| --- | ---
| Bug? | yes
| New Feature? | no
| Sulu Version | ---
| Browser Version | ---

#### Actual Behavior

If you open a page with an `internal_links` content type included, the behavior is a little bit strange when the homepage of a webspace is entered. Directly after adding it it will have the name of the webspace shown in the listing, and after saving and reloading it will have the actual title assigned in the Homepage (by default "Homepage").

#### Expected Behavior

The behavior should be consistent if the page is assigned already or is being newly assigned.

#### Steps to Reproduce

1. Open a page with an internal link field type
2. Click the plus button to open the overlay
3. Add the homepage shown in the overlay with the webspace name
4. Click confirm
5. The homepage is now listed in the field type using the name of the webspace
6. Save the page
7. Reload the page
8. After loading the items for the field type you'll see that the name of the homepage is now whatever title it has ("Homepage" if you haven't changed it since the initialization of Sulu)

#### Possible Solutions

The problem is that the API returns the title differently in both requests (once it is "/admin/api/pages?locale=en&ids=44c991c1-044e-40e9-bbd0-03fc25035ce3,6ec0fbd4-abef-4de5-a7c1-9299aacb67a7,92615b87-3ae7-4fb8-b635-0e0f50faf5a4&page=1&flat=true" and the other time it is "/admin/api/pages?locale=en&flat=true". Passing the `ids` seem to change this behavior. I think we have to be consistent in both cases.

niklasnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | yes
| Fixed tickets | fixes #3721, fixes #3877
| License | MIT

#### What's in this PR?

This PR adjusts the `ContentRepository` service to return the title of the source page instead of the title of the destination page for internal links. This affects the title that is displayed for internal links in the webspace overview.

#### Why?

This makes the displayed title consistent to the title returned by the `sulu_navigation_*` functions. Furthermore, it makes the behaviour of the `ContentRepository` consistent to the `ContentMapper`.

### DIFF
--- a/src/Sulu/Component/Content/Repository/ContentRepository.php
+++ b/src/Sulu/Component/Content/Repository/ContentRepository.php
@@ -715,7 +715,8 @@ class ContentRepository implements ContentRepositoryInterface
         $data = $linkedContent->getData();
 
         // return value of source node instead of link destination for title and non-fallback-properties
-        $sourceNodeValueProperties = array_merge(['title'], self::$nonFallbackProperties);
+        $sourceNodeValueProperties = self::$nonFallbackProperties;
+        $sourceNodeValueProperties[] = 'title';
         $properties = \array_intersect($sourceNodeValueProperties, \array_keys($data));
         foreach ($properties as $property) {
             $data[$property] = $this->resolveProperty($row, $property, $locale);
--- a/src/Sulu/Component/Content/Repository/ContentRepository.php
+++ b/src/Sulu/Component/Content/Repository/ContentRepository.php
@@ -714,9 +714,9 @@ class ContentRepository implements ContentRepositoryInterface
         $linkedContent = $this->find($row->getValue('internalLink'), $locale, $webspaceKey, $mapping);
         $data = $linkedContent->getData();
 
-        // properties which are in the intersection of the data and non
-        // fallback properties should be handled on the original row.
-        $properties = \array_intersect(self::$nonFallbackProperties, \array_keys($data));
+        // return value of source node instead of link destination for title and non-fallback-properties
+        $sourceNodeValueProperties = array_merge(['title'], self::$nonFallbackProperties);
+        $properties = \array_intersect($sourceNodeValueProperties, \array_keys($data));
         foreach ($properties as $property) {
             $data[$property] = $this->resolveProperty($row, $property, $locale);
         }
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,13 @@
 # Upgrade
 
+## 2.1.10
+
+### Changed ContentRepository to return title of source instead of link destination for internal link pages
+
+The `ContentRepository` service was changed to return the title of the source page instead of the title of the destination
+page for internal links. This makes the behaviour consistent with external links and the `ContentMapper` service. 
+This change only affects you if you are using the `ContentRepository` service with a mapping that includes the `title` property.
+
 ## 2.1.9
 
 ### A new argument `$requestStack` has been added to the `ContentTwigExtension`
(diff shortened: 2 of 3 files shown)


src/Sulu/Component/Content/Repository/ContentRepository.php
```diff
@@ -714,9 +714,9 @@ public function resolveInternalLinkContent(
         $linkedContent = $this->find($row->getValue('internalLink'), $locale, $webspaceKey, $mapping);
         $data = $linkedContent->getData();
 
-        // properties which are in the intersection of the data and non
-        // fallback properties should be handled on the original row.
-        $properties = \array_intersect(self::$nonFallbackProperties, \array_keys($data));
+        // return value of source node instead of link destination for title and non-fallback-properties
+        $sourceNodeValueProperties = array_merge(['title'], self::$nonFallbackProperties);
```
alexander-schranz: `array_merge` should only be used when merging array together. If we just want to add something to an array it is more efficient to use the `[]` operator:

```suggestion
        $sourceNodeValueProperties = self::$nonFallbackProperties;
        $sourceNodeValueProperties[] = 'title';
```

You can see the difference when running:

```bash
php -r '$startTime = microtime(true);

              $array = [];

              for ($i = 0; $i <= 10000; ++$i) {
                  $array = array_merge([\'title\'], $array);
              }

              $endTime = microtime(true);

              var_dump($endTime - $startTime);'

# float(0.34501719474792)
```

vs:

```bash
php -r '
              $startTime = microtime(true);

              $array = [];

              for ($i = 0; $i <= 10000; ++$i) {
                  $array[] = \'title\';
              }

              $endTime = microtime(true);

              var_dump($endTime - $startTime);'
              
# float(0.00098204612731934)
```

Using in this benchmark 100000 will make the array_merge a lot slower.

```
# array_merge:
# float(98.648282051086)

# []:
# float(0.0087049007415771)
```

Maybe not so relevant in this case but I would always go the `[]` way when its possible and we don't need to `merge` keys together.

alexander-schranz (changes requested):

alexander-schranz (changes requested):
A change request which can be discussed, some tests are failing and I didn't yet test it locally which effects it could have on internal_links, smart content, navigation and breadcrumb.

niklasnatter:
@alexander-schranz Tests should pass now 🙂 
As far as i see, this should not change anything for `smart_content`, `navigation` and `breadcrumb` because they use the `ContentQueryExecuter` which uses the `ContentMapper`. 

The PR adjusts the `ContentRepository` which is used by the `PageLinkProvider`, `PagesSitemapProvider`, `CustomUrlRepository` and `PageController`

alexander-schranz (approved):

Issue #5868, opened 2021-03-10
Block settings are applying in all blocks


plozmun:
| Q | A
| --- | ---
| Sulu Version | >= 2.2.4 
| PHP Version | -
| DB Version | - 
| Browser Version | Checked on Chrome Versión 88.0.4324.192 

# Actual Behavior

If some setting is applied in a block when you open block settings in another block same settings is applied ( not saved if you don't push apply )

# Steps to Reproduce

- Go to sulu.rocks , edit a block settings and check "Don't display on public website" 
- Edit settings in next block and the check is marked

niklasnatter:
Can confirm that this is a bug - thanks for reporting!
We should probably reset the data of the `blockSettingsFormStore` before opening (or after closing) the overlay here:
https://github.com/sulu/sulu/blob/aa692e8255750d21f5a287a420304c544c2a8619/src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js#L383-L395

niklasnatter:
Thanks for reporting! Should be fixed by #5882

PR #5882, opened 2021-03-17
Clear data of previous block when opening block settings overlay

niklasnatter:
Fixes https://github.com/sulu/sulu/issues/5868

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
@@ -350,7 +350,7 @@ class FieldBlocks extends React.Component<FieldTypeProps<Array<BlockEntry>>> {
         this.blockSettingsFormStore?.destroy();
         this.blockSettingsFormStore = memoryFormStoreFactory.createFromFormKey(
             settingsFormKey,
-            this.value[index][SETTINGS_KEY],
+            {...this.value[index][SETTINGS_KEY]},
             this.props.formInspector.locale,
             undefined,
             this.props.formInspector.options
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
@@ -21,7 +21,7 @@ const SETTINGS_TAG = 'sulu.block_setting_icon';
 
 @observer
 class FieldBlocks extends React.Component<FieldTypeProps<Array<BlockEntry>>> {
-    @observable blockSettingsOpen: ?number;
+    @observable openedBlockSettingsIndex: ?number;
     @observable blockSettingsFormStore: ?FormStoreInterface;
     @observable value: Object;
 
@@ -356,7 +356,7 @@ class FieldBlocks extends React.Component<FieldTypeProps<Array<BlockEntry>>> {
             this.props.formInspector.options
         );
 
-        this.blockSettingsOpen = index;
+        this.openedBlockSettingsIndex = index;
     };
 
     handleSettingsOverlayClose = () => {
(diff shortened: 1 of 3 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
```diff
@@ -351,9 +351,16 @@ class FieldBlocks extends React.Component<FieldTypeProps<Array<BlockEntry>>> {
             return;
         }
 
+        const newFormStoreData = {};
+        Object.keys(this.blockSettingsFormStore.data).forEach((key) => {
```
nnatter, line 355: TODO: add test case that tests this

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
```diff
@@ -351,9 +351,16 @@ class FieldBlocks extends React.Component<FieldTypeProps<Array<BlockEntry>>> {
             return;
         }
 
+        const newFormStoreData = {};
```
alexander-schranz: Would it not be better to instant for every new settings a new form store here with the correct data:

```js
const {formInspector} = this.props;
this.blockSettingsFormStore = memoryFormStoreFactory.createFromFormKey(
     this.settingsFormKey,
     this.value[index][SETTINGS_KEY] || {},
     formInspector.locale,
     undefined,
     formInspector.options
);
```

For me it feels just false to share the same instance of FormStore between different blocks, just afraid that we again share some state between them which we maybe don't want.

alexander-schranz (changes requested):
Not sure if we should use a global `blockSettingsFormStore` over all blocks, instead of creating a new formStore for every block when it is opened.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
```diff
@@ -35,14 +33,13 @@ class FieldBlocks extends React.Component<FieldTypeProps<Array<BlockEntry>>> {
 
     @action componentDidMount() {
         if (this.settingsFormKey) {
-            const {formInspector} = this.props;
-
+            // initialize empty blockSettingsFormStore because schema of the store is used for determining iconsMapping
             this.blockSettingsFormStore = memoryFormStoreFactory.createFromFormKey(
```
niklasnatter, line 37: unfortunately we need to initialize an empty `blockSettingsFormStore` because the schema of the store is used to determine which icons are displayed for the block in the `iconsMapping` getter

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/tests/FieldBlocks.test.js
```diff
@@ -1284,80 +1285,22 @@ test('Should open and close block settings overlay when confirm button is clicke
...
-        fieldBlocks.update();
-        expect(changeSpy).not.toBeCalled();
-
-        fieldBlocks.find('Overlay Button[children="sulu_admin.apply"]').simulate('click');
-        expect(fieldBlocks.find('Overlay').prop('open')).toEqual(false);
-        expect(changeSpy).toBeCalledWith([{type: 'default'}, {settings: {}, type: 'default'}]);
-    });
-});
-
-test('Should open and close block settings overlay when confirm button is clicked with changed data', () => {
```
niklasnatter, line 1304: this testcase is not possible anymore because the `FormOverlay` will disable the confirm button if there were no changes

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
```diff
@@ -347,46 +340,56 @@ class FieldBlocks extends React.Component<FieldTypeProps<Array<BlockEntry>>> {
...
         }
 
-        this.blockSettingsOpen = index;
-        this.blockSettingsFormStore.setMultiple(this.value[index][SETTINGS_KEY] ?? {});
-        this.blockSettingsFormStore.dirty = false;
+        // create new formstore to make sure that overlay displays correct data
+        this.blockSettingsFormStore?.destroy();
+        this.blockSettingsFormStore = memoryFormStoreFactory.createFromFormKey(
+            settingsFormKey,
+            this.value[index][SETTINGS_KEY],
```
alexander-schranz: Think here a deep copy would be needed that the `X` just close the overlay and does not apply its changes.

alexander-schranz: Could solve it by doing:

```suggestion
            {...this.value[index][SETTINGS_KEY]},
```

not sure if we should use `toJS` in this case.

niklasnatter: fixed 🙂

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
alexander-schranz: okay then it is that case its sad but we can not currently do it aslong as this lives in the FieldBlocks.js

alexander-schranz (approved):

Issue #5038, opened 2020-02-06
Allow to upload media in root collection


alexander-schranz:
| Q | A
| --- | ---
| Sulu Version | 2.0.3
| Browser Version | .

# Actual Behavior

Currently if you open the media module you can not upload files directly there you need first choice a collection.

# Expected Behavior

You should be able to upload files always then a overlay should be shown with a single collection selection.

# Steps to Reproduce

1. Try to upload a file without choosing  a collection

# Possible Solutions

Show overlay with single collection selection (should exist as used inside the media smart content provider).

nnatter:
As a quickfix, we could display a warning/tooltip if the user hovers the disabled `Upload` toolbar action or tries to drag-and-drop something 🙂

alexander-schranz:
We did got a design for this. The upload should be possible but when the media is dropped and no collection is selected the following overlay should appear to select the correct collection:

![All Media - Choose Location](https://user-images.githubusercontent.com/1698337/111486823-1a0e5f80-8738-11eb-881d-f6d9d13355b7.png)

alexander-schranz:
This is fixed by #6052 and will be part of the 2.3 release.

PR #5883, opened 2021-03-17
Add upload button to media selection

### LINKED ISSUE #5039: Add Upload button to Media Selection Overlay
# Problem description

Not all end users are familiar with Drag&Drop so this users think they have to always change to the media module to upload new files to sulu instead of Drag&Drop them into the overlay.

I personally example does not use Drag&Drop when using Windows as you mostly have all Windows in Fullscreen mode which is on Mac something you mostly don't have. So think windows users are mostly not so familiar with Drag&Drop because its not so intergrated in there OS or other Software there.

# Proposed solution

The Media Selection Overlay should also show a upload button a user can click, so the users which are not familiar with Drag&Drop can use that.

hedwiggggg:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | fixes #5039 
| Related issues/PRs | #5039 
| License | MIT

#### What's in this PR?

Added the ability to open the `MultiMediaDropzone` by clicking the newly added upload button.

#### Why?

Because it kinda confused me, that I am only able to upload new medias in the media collection and not in the media selection. Until I realized, you **can** upload in the media section, but only by dragging a file inside the modal.

I added an `uploadable` boolean to the `CollectionSection` component. Basing on the investigation I made in the `MultiMediaDropzone`, you can upload a file if addable is true and a collectionId is defined.

Then I passed the onUploadOverlayOpen to the `CollectionSection` component, to trigger the overlay open state.

#### To Do

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Button/Button.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Button/Button.js
@@ -91,12 +91,19 @@ export default class Button<T> extends React.PureComponent<Props<T>> {
                 {icon &&
                     <Icon className={iconClass} name={icon} />
                 }
+
+                {icon && children && (
+                    <span className={buttonStyles.spacer} />
+                )}
+
                 {children &&
                     <span className={buttonStyles.buttonText}>{children}</span>
                 }
+
                 {showDropdownIcon &&
                     <Icon className={buttonStyles.dropdownIcon} name="su-angle-down" />
                 }
+
                 {loading &&
                     <div className={buttonStyles.loader}>
                         <Loader size={LOADER_SIZE} />
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Button/button.scss
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Button/button.scss
@@ -114,6 +114,10 @@ $iconColorActive: $white;
     }
 }
 
+.spacer {
+    margin: auto 5px;
+}
+
 .text {
     padding: 0;
     border: 0;
@@ -146,7 +150,7 @@ $iconColorActive: $white;
     color: $iconBorderColor;
     cursor: pointer;
     font-size: $iconFontSize;
-    padding: 0 5px;
+    padding: 0 10px;
 
     &:hover:not(:disabled) {
         background-color: $silver;
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/List.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/List.js
@@ -38,6 +38,7 @@ type Props = {|
     adapters: Array<string>,
     allowActivateForDisabledItems: boolean,
     copyable: boolean,
+    customButtons?: Array<Node>,
     deletable: boolean,
     disabled: boolean,
     disabledIds: Array<string | number>,
@@ -65,6 +66,7 @@ class List extends React.Component<Props> {
     static defaultProps = {
         allowActivateForDisabledItems: true,
         copyable: true,
+        customButtons: [],
         deletable: true,
         disabled: false,
         disabledIds: [],
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/List.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/List.js
@@ -515,6 +515,7 @@ class List extends React.Component<Props> {
             adapterOptions,
             adapters,
             copyable,
+            customButtons,
             deletable,
             disabled,
             header,
@@ -601,7 +602,7 @@ class List extends React.Component<Props> {
                             </Fragment>
                         }
 
-                        {this.props.customButtons}
+                        {customButtons}
 
                         <AdapterSwitch
                             adapters={adapters}
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/List.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/List.js
@@ -598,6 +598,9 @@ class List extends React.Component<Props> {
                                 />
                             </Fragment>
                         }
+
+                        {this.props.customButtons}
+
                         <AdapterSwitch
                             adapters={adapters}
                             currentAdapter={this.currentAdapterKey}
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/List.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/List.js
@@ -37,8 +37,8 @@ type Props = {|
     adapterOptions?: {[adapterKey: string]: {[key: string]: mixed}},
     adapters: Array<string>,
     allowActivateForDisabledItems: boolean,
+    buttons?: Array<Node>,
     copyable: boolean,
-    customButtons?: Array<Node>,
     deletable: boolean,
     disabled: boolean,
     disabledIds: Array<string | number>,
@@ -65,8 +65,8 @@ const USER_SETTING_ADAPTER = 'adapter';
 class List extends React.Component<Props> {
     static defaultProps = {
         allowActivateForDisabledItems: true,
+        buttons: [],
         copyable: true,
-        customButtons: [],
         deletable: true,
         disabled: false,
         disabledIds: [],
@@ -514,8 +514,8 @@ class List extends React.Component<Props> {
         const {
             adapterOptions,
             adapters,
+            buttons,
             copyable,
-            customButtons,
             deletable,
             disabled,
             header,
@@ -602,7 +602,7 @@ class List extends React.Component<Props> {
                             </Fragment>
                         }
 
-                        {customButtons}
+                        {buttons}
 
                         <AdapterSwitch
                             adapters={adapters}
--- a/src/Sulu/Bundle/MediaBundle/Resources/js/containers/MediaCollection/CollectionSection.js
+++ b/src/Sulu/Bundle/MediaBundle/Resources/js/containers/MediaCollection/CollectionSection.js
@@ -22,12 +22,11 @@ type Props = {
     listStore: ListStore,
     locale: IObservableValue<string>,
     onCollectionNavigate: (collectionId: ?string | number) => void,
+    onOpenDropZone: () => void,
     overlayType: OverlayType,
     resourceStore: ResourceStore,
     securable: boolean,
-
     uploadable: boolean,
-    onOpenDropZone: () => void,
 };
 
 @observer
@@ -196,11 +195,11 @@ class CollectionSection extends React.Component<Props> {
             editable,
             listStore,
             locale,
+            onOpenDropZone,
             overlayType,
             resourceStore,
             securable,
             uploadable,
-            onOpenDropZone,
         } = this.props;
 
         const operationType = this.openedCollectionOperationOverlayType;
--- a/src/Sulu/Bundle/MediaBundle/Resources/js/containers/MediaCollection/CollectionSection.js
+++ b/src/Sulu/Bundle/MediaBundle/Resources/js/containers/MediaCollection/CollectionSection.js
@@ -25,6 +25,9 @@ type Props = {
     overlayType: OverlayType,
     resourceStore: ResourceStore,
     securable: boolean,
+
+    uploadable: boolean,
+    onOpenDropZone: () => void,
 };
 
 @observer
@@ -196,6 +199,8 @@ class CollectionSection extends React.Component<Props> {
             overlayType,
             resourceStore,
             securable,
+            uploadable,
+            onOpenDropZone,
         } = this.props;
 
         const operationType = this.openedCollectionOperationOverlayType;
@@ -216,6 +221,11 @@ class CollectionSection extends React.Component<Props> {
                                 {addable &&
                                     <Button icon="su-plus" onClick={this.handleAddCollectionClick} />
                                 }
+
+                                {uploadable && (
+                                    <Button icon="su-upload" onClick={onOpenDropZone} />
+                                )}
+
                                 {!!resourceStore.id && (editable || deletable || editable || securable) &&
                                     <DropdownButton icon="su-cog">
                                         {editable &&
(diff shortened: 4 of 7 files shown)


hedwiggggg:
Implemented the UX team's proposal.

TODO:
- [x] check lint:js
- [x] check lint:scss
- [x] check flow
- [x] check depcruise
- [ ] add / modify tests
- [ ] add i18n

Do I miss something else?
What do you think about the general implementation?

@alexander-schranz 

https://user-images.githubusercontent.com/42947316/111503541-21893500-8747-11eb-9eb5-efc4121c1b99.mp4

hedwiggggg:
Im not really deep into the project structure. Discovered sulu yesterday & never worked with symfony before. But I'm quite familiar with react. So if you question the decisions I made, feel free to comment!

src/Sulu/Bundle/MediaBundle/Resources/js/containers/MediaCollection/CollectionSection.js
```diff
@@ -214,7 +214,9 @@ class CollectionSection extends React.Component<Props> {
                         <div className={collectionSectionStyles.right}>
                             <ButtonGroup>
                                 {addable &&
-                                    <Button icon="su-plus" onClick={this.handleAddCollectionClick} />
+                                    <Button icon="su-plus" onClick={this.handleAddCollectionClick}>
+                                        Add folder
```
alexander-schranz: We should use translation here. `t('sulu_media.add_collection')`. Would use this translation will check this again with the UX Team but think it should be consistent to the overlay title.

src/Sulu/Bundle/MediaBundle/Resources/js/containers/MediaCollection/MediaSection.js
```diff
@@ -20,11 +23,23 @@ export default class MediaSection extends React.PureComponent<Props> {
...
         } = this.props;
 
         return (
             <List
                 adapters={adapters}
+                customButtons={[
+                    uploadable && (
+                        <ButtonGroup key="upload-media">
+                            <Button icon="su-upload" onClick={onUploadOverlayOpen}>
+                                Upload File
```
alexander-schranz: We should use translation here. `t('sulu_media.upload_file')` seems like a new translation which we would need to add in the `admin.de.json` and `admin.en.json` of MediaBundle

hedwiggggg:
Bildschirmfoto von 2021-03-17 18-33-08

Also the UX of this section can be improved. I don't know the english translation right now, but the text should be "oder **hier** klicken um hinzuzufügen" because the current translation "oder klicken um hinzuzufügen" implies that clicking anywhere should work. But clicking anywhere except the "oder klicken um hinzuzufügen" text will close the overlay.

And I would add some css rules:

```css
cursor: pointer;
text-decoration: underline;
```

This would be my proposal:

https://user-images.githubusercontent.com/42947316/111512136-b7c15900-874f-11eb-8690-d1f0a586bbbe.mp4

src/Sulu/Bundle/MediaBundle/Resources/js/containers/MediaCollection/MediaCollection.js
```diff
@@ -97,6 +97,8 @@ class MediaCollection extends React.Component<Props> {
         const securable = !locked
             && (permissions.security !== undefined ? permissions.security : MediaCollection.securable);
 
+        const hasCollectionId = ['string', 'number'].includes(typeof collectionStore.id);
```
alexander-schranz: would use the same check as in other cases used `!collectionStore.id` so I would remove this variable. see next comment.

src/Sulu/Bundle/MediaBundle/Resources/js/containers/MediaCollection/MediaCollection.js
```diff
@@ -125,6 +127,8 @@ class MediaCollection extends React.Component<Props> {
                         listStore={mediaListStore}
                         mediaListRef={mediaListRef}
                         onMediaClick={this.handleMediaClick}
+                        onUploadOverlayOpen={onUploadOverlayOpen}
```
alexander-schranz: I personally would not add 2 params here a `uploadable=true`  doesn't make sense when we don't set `onUploadOverlayOpen` so this would conflict so I would go more then with:

```js
                        onUploadOverlayOpen={(addable && collectionStore.id) ? onUploadOverlayOpen : null}
 ```

File: src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/List.js [hunk 5]
```diff
@@ -513,6 +515,7 @@ class List extends React.Component<Props> {
             adapterOptions,
             adapters,
             copyable,
+            customButtons,
```
alexander-schranz: not sure about this param and its place not sure if the buttons should be adapter specific. Here @nnatter had I think a better overview what is consistence.

At current state I personally would call it just `buttons` don't see a usage to add a `custom` prefix here.

alexander-schranz (changes requested):
@hedwiggggg thx for having a look at this issue 👍 I wanted to give you as fast as possible feedback for PR. Best would be that @nnatter will have a look at this. He has the best overview about current props apis and have a good knowledge to keep the things consistent to other, but I think he will not be able todo a review this week. So hopefully he can have a look at this PR next week.

src/Sulu/Bundle/MediaBundle/Resources/js/containers/MediaCollection/MediaCollection.js
alexander-schranz: Also this will be fixed with https://github.com/sulu/sulu/issues/5038 where we got also the design today.

alexander-schranz:
@hedwiggggg 

> This would be my proposal

Thx I forwarded it to the UX Team maybe they have some feedback, your proposal is I think not wrong if we add that CSS. But lets check if there will be some feedback from the UX team.

src/Sulu/Bundle/MediaBundle/Resources/js/containers/MediaCollection/MediaSection.js
hedwiggggg: `t('sulu_media.upload')` should work too. Or adding `t('sulu_media.upload_files')` I guess, because multiple files are possible too.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/List.js
hedwiggggg: My impression was that the adapters are only responsible for the content & the toolbar is adapter independent. But I didn't look that closely either.

alexander-schranz:
Looks like they also did add a design for this stage.

All Media - Drag Image

The whole area is clickable to upload and the overlay will only be closed over the `X` button, the close button should have the size to the dashed line.

hedwiggggg:
The last suggestions have been implemented, but not yet the one of the new dropzone. I don't really know what to do, since the D&D behavior still needs to be adjusted. (See video)

https://user-images.githubusercontent.com/42947316/111616329-0f100980-87e2-11eb-96b3-46f2d5147932.mp4

I don't know if the current behavior is intentional (`DragLeave` event closes the DropZone, `DragLeave` triggers when `DragEnter`ing over child elements).

And how the new proposal will affect this behaviour.
Logical for me would be a dropzone to upload and a dropzone to cancel, that would also be the least hacky solution.

Otherwise, the behavior would have to be adjusted somehow, do you have a suggestion? @alexander-schranz

alexander-schranz:
@hedwiggggg from my perspective if the overlay is shown by drag it should close on dragleave and if it is open by click it should be closed by click. So I think the overlay need to know how it was opened, to register the correct handling for closing. Maybe adding `closingBehaviour="click"` or something to it. /cc @nnatter 
I did not have yet time for a talk with the UX Team but hopefully will have some at end of the week or start of the next week.

src/Sulu/Bundle/AdminBundle/Resources/js/components/Button/Button.js
```diff
@@ -91,12 +91,19 @@ export default class Button<T> extends React.PureComponent<Props<T>> {
                 {icon &&
                     <Icon className={iconClass} name={icon} />
                 }
+
+                {icon && children && (
```
niklasnatter, line 95: would prefer to add an additional class to the `iconClass`or `buttonClass` instead of adding a new element

src/Sulu/Bundle/AdminBundle/Resources/js/components/Button/button.scss
```diff
@@ -146,7 +150,7 @@ $iconColorActive: $white;
     color: $iconBorderColor;
     cursor: pointer;
     font-size: $iconFontSize;
-    padding: 0 5px;
+    padding: 0 10px;
```
niklasnatter, line 153: i think this will affect all existing buttons, right? i am a bit afraid that this will break something - why do we have to change it?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/List.js
```diff
@@ -37,6 +37,7 @@ type Props = {|
     adapterOptions?: {[adapterKey: string]: {[key: string]: mixed}},
     adapters: Array<string>,
     allowActivateForDisabledItems: boolean,
+    buttons?: Array<Node>,
```
niklasnatter, line 40: i would prefer to pass some kind of metadata for rendering the buttons instead of component instances to be consistent with other parts of the system such as the [`Table` component](https://github.com/sulu/sulu/blob/d033d250dee104f0db96818c33366e425719545b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Table/types.js#L6-L10), [`ColumnList` component](https://github.com/sulu/sulu/blob/d033d250dee104f0db96818c33366e425719545b/src/Sulu/Bundle/AdminBundle/Resources/js/components/ColumnList/types.js#L2-L6) or [`Toolbar` component](https://github.com/sulu/sulu/blob/d033d250dee104f0db96818c33366e425719545b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Toolbar/types.js#L8-L11).

The `List` already has a `itemActionsProvider` prop. To keep the naming consistent, I think i would call this new prop `actions` and use a `ActionConfig` type like this:

```
// set default value to [] in defaultProps
actions: Array<ActionConfig>,
```

```
export type ActionConfig = {|
    disabled?: boolean,
    icon: string,
    label?: string,
    onClick: ?() => void,
|};
```

src/Sulu/Bundle/MediaBundle/Resources/js/containers/MediaCollection/MediaCollection.js
```diff
@@ -97,6 +97,10 @@ class MediaCollection extends React.Component<Props> {
         const securable = !locked
             && (permissions.security !== undefined ? permissions.security : MediaCollection.securable);
 
+        const onUploadClick = (addable && collectionStore.id)
```
niklasnatter, line 100: would be nice to include checks for `disabled` from the `MediaOverview`:
https://github.com/sulu/sulu/blob/2.x/src/Sulu/Bundle/MediaBundle/Resources/js/views/MediaOverview/MediaOverview.js#L272-L282

to do this, I would pass an `actions` prop instead of the `onUploadClick` to the `MediaSection`

src/Sulu/Bundle/MediaBundle/Resources/js/containers/MediaCollection/MediaCollection.js
```diff
@@ -125,6 +129,7 @@ class MediaCollection extends React.Component<Props> {
                         listStore={mediaListStore}
                         mediaListRef={mediaListRef}
                         onMediaClick={this.handleMediaClick}
+                        onUploadClick={onUploadClick}
```
niklasnatter, line 132: it looks like the `MediaSection` does not really do anything. because of this, I would prefer to directly pass a `listActions` prop that is forwarded to the `List` instead of creating an abstraction here. this would make it easier to remove the `MediaSection` component altogether in the future

src/Sulu/Bundle/MediaBundle/Resources/js/containers/MediaCollection/MediaSection.js
```diff
@@ -2,29 +2,49 @@
...
             adapters,
             listStore,
             mediaListRef,
+            onUploadClick,
         } = this.props;
 
         return (
             <List
                 adapters={adapters}
+                buttons={[
```
niklasnatter, line 38: would forward a `listActions` prop instead of abstracting this behind a `onUploadClick` prop.

IMO we should remove this component in the future as it does not really do anything. but we need to think about backwards compatibility when doing that. so i would leave it for now and not add any logic to it to keep things simple 🙂

niklasnatter (commented):
First of all, I am sorry for the late answer. I have not found the time to really look into this until now 😕  
Thanks a lot for working on it - the code looks good and we really appreciate your effort! 

I would be nice if you find the time to have a look at the comments. If not, we will update the PR when we find the time for it. I hope that is okay for you. It would be great to include this into the next minor release 🙂

niklasnatter:
Regarding to the behaviour of the `Dropzone`: I dont think its a big problem that the `Dropzone` is closed on `DragLeave`. But i think we should prevent that `DragLeave` is triggered while the cursor is inside of the displayed area. So in this case, the `DragLeave`  event would only close the `Dropzone` if the cursor is outside of the page.

But it looks like these problems exist event without the changes in this PR, right? If yes, I would solve them in a separate PR anyway 🙂

src/Sulu/Bundle/MediaBundle/Resources/js/containers/MediaCollection/MediaCollection.js
niklasnatter: No that I think of it - I would remove the button in the `MediaOverview` completely and make the new button the only way to open the upload dropzone. I think its a bit confusing if there are two buttons that do the same thing 🙂

hedwiggggg:
Hey, sorry for the late reply. Currently i'm focusing on my b.sc. and some other time consuming things, but I will look into this as soon as I find the time.

niklasnatter:
@hedwiggggg Dont worry, we appreciate your time that is spent on this!

niklasnatter:
Thanks again for working on this! We finally fixed the problem in https://github.com/sulu/sulu/pull/6050 🎉

Issue #5886, opened 2021-03-18
User Roles field is marked as mandatory but is not


johann-mok:
| Q | A
| --- | ---
| Sulu Version | 2.2.4
| PHP Version | 7.4.15

# Actual Behavior

Its possible to save users permissions without selecting any user roles.

# Expected Behavior

There should be an error message.

# Steps to Reproduce

Edit an user inside of the permissions tab and remove all user roles.

alexander-schranz:
@johann-mok the role is not required. You can create users and then later define which role you want to add to them. The user is not a required field. It could also be just a user of the website which don't use the sulu roles and security system. So this is correct behaviour that you can create a user without role.

I will keep this open as the form currently seems to tell that there need to be a user role but this is incorrect, there could be users without a role also when its not recommended to be used this way its possible when using on your website a normal [doctrine user provider](https://symfony.com/doc/current/security/user_provider.html#entity-user-provider). Beside the [`role_assignement`](https://github.com/sulu/sulu/blob/aa692e8255750d21f5a287a420304c544c2a8619/src/Sulu/Bundle/SecurityBundle/Resources/config/forms/user_details.xml#L60) field type seems currently not support the mandatory flag.

niklasnatter:
As @alexander-schranz already explained, there are valid usecases creating a user without a roles. To reflect this, the misleading `mandatory` flag was removed in https://github.com/sulu/sulu/pull/5887.

The mandatory flag does not work for the `role_assignments` field-type because of the reason described in https://github.com/sulu/sulu/issues/5426 and https://github.com/sulu/sulu/issues/2938. This could be fixed in the `2.3` release by implementing a `PropertyMetadataMapper` like in https://github.com/sulu/sulu/pull/5666. 

I think the `role_assignments` field-type is not something that should be reused in another form and therefore I dont think we should invest a lot of effort into this. But we will happily review a pull request if somebody is interested in contributing this! 🙂

johann-mok:
@alexander-schranz @nnatter it makes absolutely sence. The mandatory true flag was the confusing point. :)

PR #5887, opened 2021-03-18
Remove mandatory of user roles

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no 
| BC breaks? | no 
| Deprecations? | no 
| Fixed tickets | part of #5886 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Remove mandatory of user roles.

#### Why?

The user roles are not mandatory, currently they are marked as mandatory. There is another open issue that we should check why the role_assignement doesn't support mandatory, but I think its okay the user roles are not mandatory.

### DIFF
--- a/src/Sulu/Bundle/SecurityBundle/Resources/config/forms/user_details.xml
+++ b/src/Sulu/Bundle/SecurityBundle/Resources/config/forms/user_details.xml
@@ -57,7 +57,7 @@
                 <title>sulu_security.permissions</title>
             </meta>
             <properties>
-                <property name="userRoles" mandatory="true" type="role_assignments">
+                <property name="userRoles" type="role_assignments">
                     <meta>
                         <title>sulu_security.roles</title>
                     </meta>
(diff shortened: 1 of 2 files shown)


niklasnatter:
Tests are failing at the moment 😕

alexander-schranz:
@nnatter did adjust the metadata test.

niklasnatter:
> There is another open issue that we should check why the role_assignement doesn't support mandatory

The reason for this is the same as in https://github.com/sulu/sulu/issues/5426 and https://github.com/sulu/sulu/issues/2938.
We could fix this for the `2.3` release by implementing a `PropertyMetadataMapper` (like in https://github.com/sulu/sulu/pull/5666). But I would argue that the `role_assignments` field-type is not really meant to be reused and therefore I would invest that time into something else 🙂

niklasnatter (approved):

PR #5888, opened 2021-03-18
Set aliases for PersistenceBundle repositories in PHP instead of XML

niklasnatter:
I think doing this in the code is a bit easier to understand because we can create the aliases after calling the `PersistenceBundleTrait::buildPersistence` method 🙂 

Documentation PR: https://github.com/sulu/sulu-docs/pull/637

### DIFF
--- a/src/Sulu/Bundle/AudienceTargetingBundle/SuluAudienceTargetingBundle.php
+++ b/src/Sulu/Bundle/AudienceTargetingBundle/SuluAudienceTargetingBundle.php
@@ -13,13 +13,9 @@ namespace Sulu\Bundle\AudienceTargetingBundle;
 
 use Sulu\Bundle\AudienceTargetingBundle\DependencyInjection\Compiler\AddRulesPass;
 use Sulu\Bundle\AudienceTargetingBundle\Entity\TargetGroupConditionInterface;
-use Sulu\Bundle\AudienceTargetingBundle\Entity\TargetGroupConditionRepositoryInterface;
 use Sulu\Bundle\AudienceTargetingBundle\Entity\TargetGroupInterface;
-use Sulu\Bundle\AudienceTargetingBundle\Entity\TargetGroupRepositoryInterface;
 use Sulu\Bundle\AudienceTargetingBundle\Entity\TargetGroupRuleInterface;
-use Sulu\Bundle\AudienceTargetingBundle\Entity\TargetGroupRuleRepositoryInterface;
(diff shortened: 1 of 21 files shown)


src/Sulu/Bundle/AudienceTargetingBundle/SuluAudienceTargetingBundle.php
```diff
@@ -40,6 +44,15 @@ public function build(ContainerBuilder $container)
             $container
         );
 
+        $container->addAliases(
```
niklasnatter: not sure if it causes problems to do this in the `Bundle` class. but it would be nice to do it here because we also need to call the `buildPersistence` method with all model interfaces. this makes it easier to think about this 🙂

alexander-schranz: The `Bundle` is where no services should be in my opinion defined. The Bundle is place to define CompilerPasses. Service should in our case be defined in the `xml` aslong as we use `xml` for service definitions, some aliases maybe in the `Extensions:load` which need to be programmitically, Compilerpasses mostly should only in edge cases define new services.

alexander-schranz (changes requested):
I personally not a fan of moving them out of the `xml` if I'm searching for the service definitions I would go into the `xml` files. I would aslong as we use `xml` to define our service also define our aliases which are possible to be defined there still in the `xml` files.

src/Sulu/Bundle/AudienceTargetingBundle/SuluAudienceTargetingBundle.php
niklasnatter: alright - thanks for the explanation. i moved the logic into the `Extension` classes 🙂

niklasnatter: I agree with you that we should use the `xml` files for defining services in general. But i think in this case it would be nice to do it in the `Extension` class after the `configurePersistence` call. This makes it easier to understand that the `sulu.repository.*` services are created by the `PersistenceBundle`. Also we cannot forget about creating the alias when copying the code from another bundle this way 😉 

But I dont have a too strong opinion here. The current solution works for me too, just think this would make things easier to understand 🙂

alexander-schranz: is okay for if errors could prevent with this in future we could use in this case php

alexander-schranz (approved):

PR #5840, opened 2021-02-24
Add EventLogBundle for dispatching and recording events that happen in the application

niklasnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no 
| Related issues/PRs | #5545
| License | MIT

#### What's in this PR?

This PR implements a proof-of-concept of a basic `EventLogBundle`. The goal of the bundle is to provide a centralized hub that is used for dispatching and handling events that describe what is happening in the application. It aims to fulfill the following usecases:

* Allow to listen for specific events in a project (eg updating a project specific search index when a category is modified)
* Persist events to allow for querying the events for a specific resource (eg for displaying a "changelog"/"history" for a specific page)

PR #5898, opened 2021-03-26
Make phpstan configuration stricter and generate baseline for existing code

niklasnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| License | MIT

#### What's in this PR?

This PR adjusts the `phpstan.neon` to match the configuration of other projects such as the [sulu-demo](https://github.com/sulu/sulu-demo) or the [SuluHeadlessBundle](https://github.com/sulu/SuluHeadlessBundle).

Unfortunately, it is not realistic that the existing code will be updated to comply with this strict configuration. Therefore I generated a [phpstan baseline](https://phpstan.org/user-guide/baseline) to ignore the existing errors in the codebase. This `phpstan-baseline.neon` file needs to be updated when adjusting the respective parts of the existing code with the following command:

```
vendor/bin/phpstan analyse --generate-baseline
```

#### Why?

We are not able to update the complete codebase to comply with a strict configuration, but we want to make sure that newly added code is analyzed as strict as possible. **This PR forces us to write strictly typed code when implementing new features** such as https://github.com/sulu/sulu/pull/5840.


src/Sulu/Bundle/AdminBundle/Tests/Unit/Admin/AdminPoolTest.php
```diff
@@ -24,12 +25,12 @@ class AdminPoolTest extends TestCase
     protected $adminPool;
 
     /**
-     * @var Admin
+     * @var Admin|ObjectProphecy
```
niklasnatter, line 28: used this to test if the `phpstan-prophecy` is working

alexander-schranz (approved):

src/Sulu/Bundle/CoreBundle/SuluCoreBundle.php
```diff
@@ -51,7 +50,6 @@ public function build(ContainerBuilder $container)
             )
         );
 
-        $container->addCompilerPass(new ExceptionHandlerCompilerPass());
```
niklasnatter, line 54: the `FOS\RestBundle\Serializer\Normalizer\ExceptionHandler` was deprecated in https://github.com/FriendsOfSymfony/FOSRestBundle/pull/2161 and removed in `3.0`. we have already disabled this class via `fos_rest.exception.serialize_exceptions` in https://github.com/sulu/sulu/pull/4798

src/Sulu/Bundle/MediaBundle/Tests/Fixtures/DefaultStructureCache.php
```diff
@@ -1,154 +0,0 @@
...
-use Sulu\Component\Content\Block\BlockProperty;
-use Sulu\Component\Content\Block\BlockPropertyType;
-use Sulu\Component\Content\Property;
-use Sulu\Component\Content\PropertyTag;
-use Sulu\Component\Content\Section\SectionProperty;
-
-/**
- * DefaultStructureCache.
- */
-class DefaultStructureCache extends \Sulu\Component\Content\Metadata
```
niklasnatter, line 23: the `\Sulu\Component\Content\Metadata` does not exist and i havent found out if there is a replacement for it. but it looks like this is not used anyway 🙈

src/Sulu/Bundle/CoreBundle/SuluCoreBundle.php
alexander-schranz: Are you sure it works still correctly when having FosRest 2.8 installed?

alexander-schranz: Because as I remember there the old mechanism is used where the service is used

src/Sulu/Bundle/MediaBundle/Tests/Fixtures/DefaultStructureCache.php
alexander-schranz: as it is part of the `Tests` this can be safely be removed if its not used in any test longer

src/Sulu/Bundle/CoreBundle/SuluCoreBundle.php
niklasnatter: i have not tested it yet, but as far as i see is the `fos_rest.serializer.exception_normalizer.jms` service removed if `fos_rest.exception.serialize_exceptions` is set to `false` here:
https://github.com/FriendsOfSymfony/FOSRestBundle/blob/369149f1de7f6519960fb1bb54c73e1b8c7dddcf/DependencyInjection/FOSRestExtension.php#L431-L439

we do set this to `false` here:
https://github.com/sulu/sulu/blob/2.1/src/Sulu/Bundle/CoreBundle/DependencyInjection/SuluCoreExtension.php#L101

alexander-schranz: is okay for me then to remove this call

PR #5900, opened 2021-03-26
Dispatch domain events in CategoryBundle

niklasnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Related issues/PRs | builds upon #5840
| License | MIT

#### What's in this PR?

This PR adjusts the `CategoryManager` to dispatch domain events.

### DIFF
--- a/src/Sulu/Bundle/CategoryBundle/Category/CategoryManager.php
+++ b/src/Sulu/Bundle/CategoryBundle/Category/CategoryManager.php
@@ -269,9 +269,10 @@ class CategoryManager implements CategoryManagerInterface
         }
 
         $defaultTranslation = $entity->findTranslationByLocale($entity->getDefaultLocale());
+        $categoryName = $defaultTranslation ? $defaultTranslation->getTranslation() : null;
 
         $this->em->remove($entity);
-        $this->domainEventCollector->collect(new CategoryRemovedEvent($id, $defaultTranslation->getTranslation()));
+        $this->domainEventCollector->collect(new CategoryRemovedEvent($id, $categoryName));
         $this->em->flush();
 
         // throw a category.delete event
--- a/src/Sulu/Bundle/CategoryBundle/Category/CategoryManager.php
+++ b/src/Sulu/Bundle/CategoryBundle/Category/CategoryManager.php
@@ -32,7 +32,6 @@ use Sulu\Bundle\CategoryBundle\Exception\CategoryKeyNotUniqueException;
 use Sulu\Bundle\CategoryBundle\Exception\CategoryNameMissingException;
 use Sulu\Bundle\EventLogBundle\Application\Collector\DomainEventCollectorInterface;
 use Sulu\Bundle\MediaBundle\Entity\MediaInterface;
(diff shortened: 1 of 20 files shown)


src/Sulu/Bundle/CategoryBundle/Domain/Event/CategoryCreatedEvent.php
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    public function getEventPayload(): array
+    {
+        return $this->payload;
+    }
+
+    public function getResourceKey(): string
+    {
+        return 'categories';
```
nnatter, line 64: define a constant for this

src/Sulu/Bundle/CategoryBundle/Domain/Event/CategoryCreatedEvent.php
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    public function getResourceLocale(): ?string
+    {
+        return $this->locale;
+    }
+
+    public function getResourceTitle(): ?string
+    {
+        return $this->category->findTranslationByLocale($this->locale)->getTranslation();
```
alexander-schranz: `$this->category->findTranslationByLocale($this->locale)` could return `null`

src/Sulu/Bundle/CategoryBundle/Domain/Event/CategoryMovedEvent.php
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    public function getResourceId(): string
+    {
+        return (string) $this->category->getId();
+    }
+
+    public function getResourceTitle(): ?string
+    {
+        return $this->category->findTranslationByLocale($this->category->getDefaultLocale())->getTranslation();
```
alexander-schranz: same above the `findTranslationByLocale` can return null would more be a race condition  but maybe we should still not fail hard if it happens

alexander-schranz (changes requested):

src/Sulu/Bundle/CategoryBundle/Domain/Event/CategoryMovedEvent.php
```diff
@@ -0,0 +1,77 @@
...
+    public function getEventType(): string
+    {
+        return 'moved';
+    }
+
+    public function getEventContext(): array
+    {
+        return [
+            'previousParentId' => $this->previousParentId,
+            'newParentId' => $this->category->getParent()->getId(),
```
niklasnatter: i suspect that `$this->category->getParent()` could return null

src/Sulu/Bundle/CategoryBundle/Category/CategoryManager.php
```diff
@@ -272,7 +290,10 @@ public function move($id, $parent)
             throw new CategoryIdNotFoundException($parent);
         }
 
+        $previousParentId = $category->getParent()->getId();
```
alexander-schranz: @nnatter think this `getParent()` call could also be `null` when it is a root Category moved into a child

alexander-schranz (approved):

wachterjohannes (approved):

PR #5902, opened 2021-03-27
Remove context based service definitions

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? |  yes
| BC breaks? | no 
| Deprecations? | yes 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Remove context based service definitions.

#### Why?

It should be possible to access services like list builder and form metadata also from the website context. The reasons why the 2 Kernel Setup was introduced are not longer relevant today. As the main reasons that the firewall was not lazy is not longer the case.

#### Example Usage

```php
class Kernel extends SuluKernel implements HttpCacheProvider
{
    public static $SULU_CONTEXT_DEFINITION_DISABLED = true;
   
    // ...
}
```

#### To Do


src/Sulu/Bundle/PreviewBundle/EventSubscriber/CacheCommandSubscriber.php
```diff
@@ -69,6 +69,8 @@ public function onCommand(ConsoleCommandEvent $event)
             return;
         }
 
+        return; // TODO FIXME currently not possible to clear preview and admin cache
```
alexander-schranz: ends in infinity loop in `cache:clear` command jackalope factory autoloader

alexander-schranz:
This is something we can only do step by step so I will close this PR, as it is easier and less error prone step by step enabling some bundles and service in both context. Current biggest problem seems to be that the `CMFRoutingBundle` makes problems when activated in the `Admin` context. So I will first focus on bringing currently the admin services in the website, because that is the important thing that new form and list metadata can be in future be used by the content bundle.

PR #5903, opened 2021-03-29
Include schema defined in XML template in StructureFormMetadataLoader

niklasnatter:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| License | MIT

#### What's in this PR?

This PR adjusts the `StructureFormMetadataLoader` to include the schema that is defined in the `<schema>` node inside of the XML template. This is already implemented in the `FormXmlLoader` and therefore works for non-template forms such as the [`analytic_details.xml`](https://github.com/sulu/sulu/blob/2.x/src/Sulu/Bundle/WebsiteBundle/Resources/config/forms/analytic_details.xml#L8-L56). 

#### Why?

Because it might be useful to use this inside of page template for something like shown below. Also, the template XML schema already allows to define a `<schema>` node, but content of the node is ignored at the moment. 

```xml
<schema>
    <!-- template is valid if at least one of the firstProperty and secondProperty is set -->
    <anyOf>
        <schema>
            <properties>
                <property name="firstProperty" mandatory="true"/>
            </properties>
        </schema>
        <schema>
            <properties>
                <property name="secondProperty" mandatory="true"/>
            </properties>
        </schema>
    </anyOf>
</schema>
```

### DIFF
--- a/src/Sulu/Bundle/RouteBundle/Command/UpdateRouteCommand.php
+++ b/src/Sulu/Bundle/RouteBundle/Command/UpdateRouteCommand.php
@@ -67,7 +67,7 @@ class UpdateRouteCommand extends Command
     {
         $this->addArgument('entity', InputArgument::REQUIRED)
             ->addArgument('locale', InputArgument::REQUIRED)
-            ->addOption('batch-size', null, InputOption::VALUE_REQUIRED, '', 1000)
+            ->addOption('batch-size', null, InputOption::VALUE_REQUIRED, '', '1000')
             ->setDescription('Update the routes for all entities.')
             ->setHelp(
                 <<<'EOT'
@@ -82,7 +82,7 @@ EOT
     {
         $this->translator->setLocale($input->getArgument('locale'));
 
-        $batchSize = $input->getOption('batch-size');
+        $batchSize = (int) $input->getOption('batch-size');
 
         /** @var EntityRepository $repository */
         $repository = $this->entityManager->getRepository($input->getArgument('entity'));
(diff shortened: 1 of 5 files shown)


alexander-schranz (approved):

src/Sulu/Bundle/RouteBundle/Command/UpdateRouteCommand.php
```diff
@@ -67,7 +67,7 @@ protected function configure()
     {
         $this->addArgument('entity', InputArgument::REQUIRED)
             ->addArgument('locale', InputArgument::REQUIRED)
-            ->addOption('batch-size', null, InputOption::VALUE_REQUIRED, '', 1000)
+            ->addOption('batch-size', null, InputOption::VALUE_REQUIRED, '', '1000')
```
niklasnatter, line 70: this was detected as an error by phpstan with the new `symfony/symfony:5.2.6`

niklasnatter:
I think the failing test cases are caused by the new `symfony/symfony:5.2.6` and ar not related to this PR:

almare:
Is it possible that these changes break the xincludes? Since I updated to 2.2.7 all my includes are broken which work before in 2.2.6. I moved all my custom blocks into subtemplates which I load then inside the content types.

niklasnatter:
@almare As far as i understand, this should not affect xincludes. But maybe I am mistaken.
Can you please create a new issue in this repository that describes how to reproduce the problem using the `sulu/skeleton`?

alexander-schranz:
@almare I can not reproduce it. Can you create an issue for it with an example based on the current `sulu/skeleton` repository how you are including files in your templates.

almare:
Sure, thank you for your fast response!!

almare:
I found the issue. It was caused by me due to a accidentality added sting to the xml file. Sorry for the wrong alert! Works as expected.

Issue #5759, opened 2021-01-19
Profile overlay shows data of previous user when switching users without reload


jordygroote:
| Q | A
| --- | ---
| Sulu Version | Latest

# Actual Behavior

The edit profile keeps hanging on administrator eventhough i am not logged in as administrator.
Also the form is broken and cannot save.

# Expected Behavior

The edit profile modal is showing the correct info of the account and the form can be saved.

# Steps to Reproduce
- Be logged in as administrator
- Create account
- Login as new account
- Click left top menu bar to open edit profile

I added screenshots (screenshots are from sulu.rocks).
![Schermafbeelding 2021-01-19 om 13 43 39](https://user-images.githubusercontent.com/8653367/105036413-ba0f7b80-5a5c-11eb-8c24-2df6441cae60.png)
![Schermafbeelding 2021-01-19 om 13 43 44](https://user-images.githubusercontent.com/8653367/105036416-bbd93f00-5a5c-11eb-94cf-2b7bddf8b02c.png)

alexander-schranz:
Seems like the form is not cleared. After refresh the new data is shown. I think the overlay should always load the data from the endpoint and not cache it.

niklasnatter:
> Seems like the form is not cleared. After refresh the new data is shown. I think the overlay should always load the data from the endpoint and not cache it.

Sounds good. Solving this might be a good opportunity form implementing a reusable `FormOverlay` container in the core. I implemented something like that in a project recently:

```javascript
/* ResourceFormOverlay.js */

import React from 'react';
import {observer} from 'mobx-react';
import {translate} from 'sulu-admin-bundle/utils/Translator';
import Form from 'sulu-admin-bundle/containers/Form';
import Overlay from 'sulu-admin-bundle/components/Overlay';
import Snackbar from 'sulu-admin-bundle/components/Snackbar';
import ResourceFormStore from 'sulu-admin-bundle/containers/Form/stores/ResourceFormStore';
import ResourceStore from 'sulu-admin-bundle/stores/ResourceStore';
import {action, observable} from 'mobx';
import Loader from 'sulu-admin-bundle/components/Loader';
import resourceFormOverlayStyles from './resourceFormOverlay.scss';

@observer
class ResourceFormOverlay extends React.Component {
    static defaultProps = {
        formStoreMetadataOptions: {},
        formStoreOptions: {},
    };

    formRef;

    @observable formStore;
    @observable formErrors = [];

    handleOverlayConfirm = () => {
[...]

PR #5905, opened 2021-03-30
Extract FormOverlay component from FormOverlayList view

niklasnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Related issues/PRs | #5868, #5759
| License | MIT

#### What's in this PR?

This PR extracts a reusable `FormOverlay` from the `FormOverlayList` view.

#### Why?

As a preparation for fixing #5882 and #5759. We display a form inside of an overlay in various places of the administration interface. We should use a reusable component in these places to o keep things consistent.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/FormOverlay/FormOverlay.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/FormOverlay/FormOverlay.js
@@ -8,6 +8,7 @@ import {translate} from '../../utils';
 import Snackbar from '../../components/Snackbar';
 import Form from '../Form';
 import type {FormStoreInterface} from '../Form/types';
+import type {ResourceFormStore} from '../Form';
 import type {Size} from '../../components/Overlay/types';
 import formOverlayStyles from './formOverlay.scss';
 
@@ -15,7 +16,7 @@ type Props = {|
     confirmDisabled: boolean,
     confirmLoading: boolean,
     confirmText: string,
-    formStore: FormStoreInterface,
+    formStore: FormStoreInterface | ResourceFormStore,
     onClose: () => void,
     onConfirm: () => void,
     open: boolean,
@@ -44,8 +45,7 @@ class FormOverlay extends React.Component<Props> {
         const {confirmLoading, formStore} = this.props;
 
         // disable confirm button while saving if formstore is instance of ResourceFormStore
-        // $FlowFixMe
-        const formStoreSaving = formStore.hasOwnProperty('saving') && formStore.saving;
+        const formStoreSaving = (typeof formStore.saving === 'boolean') && formStore.saving;
 
         return confirmLoading || formStoreSaving;
     }
@@ -74,7 +74,7 @@ class FormOverlay extends React.Component<Props> {
         } = this.props;
 
         // save data before calling onConfirm callback if formstore is instance of ResourceFormStore
-        if (formStore.hasOwnProperty('save')) {
+        if (typeof formStore.save === 'function') {
             // $FlowFixMe
             formStore.save()
                 .then(() => {
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/FormOverlay/FormOverlay.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/FormOverlay/FormOverlay.js
@@ -46,7 +46,10 @@ class FormOverlay extends React.Component<Props> {
 
     @computed get confirmLoading() {
         const {confirmLoading, formStore} = this.props;
-        const formStoreSaving = (formStore instanceof ResourceFormStore) ? formStore.saving : false;
+
+        // disable confirm button while saving if formstore is instance of ResourceFormStore
+        // $FlowFixMe
+        const formStoreSaving = (formStore && formStore.hasOwnProperty('saving')) ? formStore.saving : false;
 
         return confirmLoading || formStoreSaving;
     }
@@ -82,7 +85,9 @@ class FormOverlay extends React.Component<Props> {
             onConfirm,
         } = this.props;
 
-        if (formStore instanceof ResourceFormStore) {
+        // save data before calling onConfirm callback if formstore is instance of ResourceFormStore
+        if (formStore && formStore.hasOwnProperty('saving')) {
+            // $FlowFixMe
             formStore.save()
                 .then(() => {
                     onConfirm();
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/FormOverlay/FormOverlay.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/FormOverlay/FormOverlay.js
@@ -1,0 +1,155 @@
+// @flow
+import React from 'react';
+import {action, computed, observable} from 'mobx';
+import type {ElementRef} from 'react';
+import {observer} from 'mobx-react';
+import Overlay from '../../components/Overlay';
+import {translate} from '../../utils';
+import Snackbar from '../../components/Snackbar';
+import Loader from '../../components/Loader';
+import Form, {ResourceFormStore} from '../Form';
+import type {FormStoreInterface} from '../Form/types';
+import type {Size} from '../../components/Overlay/types';
+import formOverlayStyles from './formOverlay.scss';
+
+type Props = {|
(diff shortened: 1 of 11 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/containers/FormOverlay/FormOverlay.js
```diff
@@ -0,0 +1,155 @@
...
+import type {FormStoreInterface} from '../Form/types';
+import type {Size} from '../../components/Overlay/types';
+import formOverlayStyles from './formOverlay.scss';
+
+type Props = {|
+    confirmDisabled: boolean,
+    confirmLoading: boolean,
+    confirmText: string,
+    formStore?: ?FormStoreInterface,
+    loading: boolean,
```
alexander-schranz: This seems not to be used. And should the `loading` state not be read from the [`formStore::loading`](https://github.com/sulu/sulu/blob/3dd8b83bccf488196dcd526ede19dda0e4b52932/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/AbstractFormStore.js#L105) property?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FormOverlay/FormOverlay.js
```diff
@@ -0,0 +1,155 @@
...
+import Form, {ResourceFormStore} from '../Form';
+import type {FormStoreInterface} from '../Form/types';
+import type {Size} from '../../components/Overlay/types';
+import formOverlayStyles from './formOverlay.scss';
+
+type Props = {|
+    confirmDisabled: boolean,
+    confirmLoading: boolean,
+    confirmText: string,
+    formStore?: ?FormStoreInterface,
```
alexander-schranz: Can this really be null? Should then not a normal overlay used instead, if there is no formStore?

alexander-schranz (changes requested):
Know it is still WIP but just some question I'm not sure about the current implementation.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FormOverlay/FormOverlay.js
niklasnatter: just to make rendering easier 🙂  
i think in most cases the component will be used like this:
```
<FormOverlay
    formStore={this.formStore}
    open={!!this.formStore}
/>
```

if we make the prop non-nullable, you need to render it like this:
```
{this.formStore && (
    <FormOverlay
        formStore={this.formStore}
        open={!!this.formStore}
    />
)}
```

but dont have a too strong opinion. we can make it non-nullable too

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FormOverlay/FormOverlay.js
niklasnatter: it is used here: https://github.com/sulu/sulu/pull/5905/files#diff-916771f5f650152bc3fb03ed379df868f07312e27fabca909b31946d8aa8518bR138-R148

the `<Form>` container will handle the `formStore::loading` property for us. but in the case of the `ProfileFormOverlay`, we load the data by ourselves (because the profile does not have an id and we cannot use the `ResourceFormStore`). it would be nice to display a loader in the overlay in this case while loading the data

niklasnatter: but i am okay with removing the prop for now and think about this when fixing #5759 🙂

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FormOverlay/FormOverlay.js
alexander-schranz: What happens if the FormOverlay is open and has no formStore, would it then not just render nothing. Should we not avoid incompatible property combinations? I would more go making it required, did look at the `ListOverlay` there the [`ListStore`](https://github.com/sulu/sulu/blob/3dd8b83bccf488196dcd526ede19dda0e4b52932/src/Sulu/Bundle/AdminBundle/Resources/js/containers/ListOverlay/ListOverlay.js#L22) is also required.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FormOverlay/FormOverlay.js
alexander-schranz: Okay I understand, the MemoryFormStore seems to have no loading state and get the data from outside. Could think that given the MemoryFormStore a promise instead of data and that it has so a `.loading` state would be nice. Also that the `profile` can not be a resource is something which isn't good but we would there need to implement something like `resource_type` which would fix our problem with the [Form Views hacks](https://github.com/sulu/sulu-demo/pull/88/files#diff-eabd835211e4b74baa3e6829bf5075a55cbca853d261d11a0b1491ba644736f4R51) currently. Overall for me its currently okay to let the `loading` prop there.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FormOverlay/FormOverlay.js
```diff
@@ -0,0 +1,143 @@
...
+    };
+
+    handleFormSubmit = () => {
+        const {
+            formStore,
+            onConfirm,
+        } = this.props;
+
+        // save data before calling onConfirm callback if formstore is instance of ResourceFormStore
+        if (formStore.hasOwnProperty('save')) {
```
niklasnatter: cannot use `formStore instanceof ResourceFormStore` here because i have not found a way to make that check work in the tests 😕

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FormOverlay/FormOverlay.js
```diff
@@ -0,0 +1,143 @@
...
+
+        return confirmDisabled || !formStore.dirty;
+    }
+
+    @computed get confirmLoading() {
+        const {confirmLoading, formStore} = this.props;
+
+        // disable confirm button while saving if formstore is instance of ResourceFormStore
+        // $FlowFixMe
+        const formStoreSaving = formStore.hasOwnProperty('saving') && formStore.saving;
```
niklasnatter: cannot use `formStore instanceof ResourceFormStore` here because i have not found a way to make that check work in the tests 😕

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FormOverlay/FormOverlay.js
alexander-schranz: think we need here `if (formStore.save) {` or `if (typeof formStore.save === 'function') {`. the minified code fails here

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FormOverlay/tests/FormOverlay.test.js
```diff
@@ -0,0 +1,388 @@
...
+});
+
+test('Should pass correct props to Overlay component', () => {
+    const formStore = new MemoryFormStore({}, {}, undefined, undefined);
+    formStore.dirty = true;
+
+    const closeSpy = jest.fn();
+
+    const formOverlay = shallow(<FormOverlay
+        confirmDisabled={false}
```
alexander-schranz: would maybe add a test wit the minimal props and check that the default values are correctly used because changing default values should fail the tests as it would be a bc break

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FormOverlay/tests/FormOverlay.test.js
```diff
@@ -0,0 +1,388 @@
...
+    />);
+
+    const form = formOverlay.find(Form);
+
+    expect(form.props()).toEqual(expect.objectContaining({
+        store: formStore,
+    }));
+});
+
+test('Should disable confirm button if FromStore is not dirty', () => {
```
alexander-schranz: `FromStore` -> `FormStore` also on other tests

alexander-schranz (approved):

PR #5911, opened 2021-03-31
Add twig blocks to admin twig template

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? yes 
| BC breaks? | no 
| Deprecations? | no  
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Add blocks to twig templates.

#### Why?

This allows use to move the [`observer` of the sulu demo](https://github.com/sulu/sulu-demo/blob/8140793ba355f3f2ef3a2934c9ff4e7d03311803/assets/admin/index.js#L25-L43) into that template. And use the build provided by sulu/skeleton repository.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/views/Admin/main.html.twig
+++ b/src/Sulu/Bundle/AdminBundle/Resources/views/Admin/main.html.twig
@@ -16,7 +16,7 @@
 
     <script>
         {%- set config = {
-            initialLoginState: is_granted('IS_AUTHENTICATED_REMEMBERED') ? 'true' : 'false',
+            initialLoginState: is_granted('IS_AUTHENTICATED_REMEMBERED'),
             translations: translations,
             fallbackLocale: fallback_locale,
             endpoints: endpoints,
--- a/src/Sulu/Bundle/AdminBundle/Resources/views/Admin/main.html.twig
+++ b/src/Sulu/Bundle/AdminBundle/Resources/views/Admin/main.html.twig
@@ -1,30 +1,37 @@
 <!DOCTYPE html>
 <html>
-    <head>
-        <meta charset="utf-8">
-        <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
-        <title>SULU</title>
-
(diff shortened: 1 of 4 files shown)


src/Sulu/Bundle/AdminBundle/Resources/views/Admin/main.html.twig
```diff
@@ -1,30 +1,37 @@
...
+    {%- endblock ~%}
+</head>
+<body>
+    {% block application -%}
+        <div id="application"></div>
+    {% endblock %}
+
+    <script>
+        {%- set config = {
+            initialLoginState: is_granted('IS_AUTHENTICATED_REMEMBERED') ? 'true' : 'false',
```
niklasnatter: think we need to use booleans instead of strings here? 🤔

alexander-schranz: good catch 👍 will fix it!

alexander-schranz: @nnatter fixed and a test case added for this endpoint

src/Sulu/Bundle/AdminBundle/Tests/Application/config/routing.yml
```diff
@@ -7,6 +13,10 @@ sulu_admin_api:
     type: rest
     prefix: /admin/api
 
+sulu_security:
+    resource: "@SuluSecurityBundle/Resources/config/routing.yml"
+    prefix: /admin/security
```
alexander-schranz, line 18: required for testing the `/admin/` route

src/Sulu/Bundle/AdminBundle/Tests/Functional/Controller/AdminControllerTest.php
```diff
@@ -56,6 +56,52 @@ public function testGetConfig()
         $this->assertEquals('de_at', $response->sulu_admin->localizations[3]->localization);
     }
 
+    public function testTemplateConfig(): void
```
niklasnatter, line 59: 🎉

Issue #5349, opened 2020-06-15
Deprecating security groups


NeuralClone:
| Q | A
| --- | ---
| Sulu Version | 2.0.8
| Bug? | Unclear
| Browser Version | Doesn't appear to matter

# Actual Behavior

Groups in the SecurityBundle lack a UI to manage them in the admin panel.

# Steps to Reproduce

Use the instructions found in the [Getting Started guide](https://docs.sulu.io/en/2.0/book/getting-started.html) in the Sulu documentation and create a user with permissions to view, add, edit, and delete roles, groups, and users.

# Additional Information

I've been looking through the current Sulu documentation and I've run into a bit of trouble with user groups. In the [SecurityBundle documentation](https://docs.sulu.io/en/2.0/bundles/security.html), it says:

> In addition to that a user can have some roles and groups assigned, whereby a group can consist of multiple other groups and roles.

And then the feature isn't elaborated upon. The concept of groups being assigned roles is fairly self-explanatory but as far as I've been able to tell, there is absolutely no trace of groups in the admin panel except when editing permissions for a specific role.

I'm using the default configuration found in [sulu/skeleton](https://github.com/sulu/skeleton). I've looked through the source code and I've found the code that handles user groups. Even the appropriate Doctrine model exists. The actual database tables have also been automatically created.

But how are groups actually used and/or enabled? I've double checked the permissions for my admin user and the user has the ability to view, add, edit, and delete groups along with roles and users. Yet I can't actually create groups through the admin panel and there isn't a console command that handles that. They don't appear in the demo on the Sulu website either.

![Security](https://user-images.githubusercontent.com/678221/84695868-880fc700-af19-11ea-8b67-36da0e63b909.png)

Are user groups supposed to be a feature that's functional by default or are they something we're supposed to manually add to the admin panel if we want to use them? Are there settings somewhere that let us enable and/or configure them?

It's very unclear how groups are intended to be used without additional information. I feel like I'm missing something really obvious.

alexander-schranz:
@NeuralClone Thank you for your report. 
I'm not sure if there where some projects using the groups but in the core I think they are not implemented and have no effect atleast in the current versions, so here definitly the Documentaiton is missleading and should be adopted. Do you have any usecase for groups? Can you provide some information what you want to handle over groups and can not be greated with the current roles?

/cc @sulu/core-team

NeuralClone:
Hi @alexander-schranz,

[...]

PR #5912, opened 2021-03-31
Deprecate unused group functionality of SecurityBundle

niklasnatter:
Fixes https://github.com/sulu/sulu/issues/5349


src/Sulu/Bundle/SecurityBundle/Admin/SecurityAdmin.php
```diff
@@ -107,12 +110,6 @@ public function getSecurityContexts()
                         PermissionTypes::EDIT,
                         PermissionTypes::DELETE,
                     ],
-                    static::GROUP_SECURITY_CONTEXT => [
```
niklasnatter, line 110: as there is not interface for using the group functionality, these permissions do not have any consequence in the administration interface in Sulu 2

alexander-schranz (approved):

alexander-schranz (approved):

alexander-schranz (approved):

PR #5914, opened 2021-04-02
Fix media caching with security

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| License | MIT

#### What's in this PR?

Prevents accessing the `currentUser` if no permissions are given. Trying to get the `user` results in a session and thus the response header is automatically set to `private` by symfony.

#### Why?

Activating a security context (without `check-permission`) results in non cacheable response.

#### To Do

- [ ] Write a test for this behaviour

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/Media/Manager/MediaManager.php
+++ b/src/Sulu/Bundle/MediaBundle/Media/Manager/MediaManager.php
@@ -238,7 +238,7 @@ class MediaManager implements MediaManagerInterface
             ['pagination' => false, 'ids' => $ids],
             null,
             null,
-            $this->getCurrentUser(),
+            $permission ? $this->getCurrentUser() : null,
             $permission
         );
         $this->count = \count($mediaEntities);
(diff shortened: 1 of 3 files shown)


Prokyonn:
Currently the permissions in the `MediaManager` are a bit confusing. There is a property `$this->permissions` but this is never used, because the permissions are always given via method parameters like in `getByIds(array $ids, $locale, $permission = null)`. Can we somehow delete the property without introducing a BC or mark it deprecated? WDYT @alexander-schranz

alexander-schranz:
We should add `@deprecated` to the `var` and maybe remove it in `2.3` as there will be a necessary BC Break for introducing the SuluEventLogBundle into the manager.

alexander-schranz (approved):

src/Sulu/Bundle/MediaBundle/Media/Manager/MediaManager.php
```diff
@@ -197,6 +199,7 @@ public function __construct(
         $this->tokenStorage = $tokenStorage;
         $this->securityChecker = $securityChecker;
         $this->ffprobe = $ffprobe;
+        // TODO permissions are deprecated and should be removed in 2.3
```
niklasnatter, line 202: @Prokyonn can you create a PR for this? i am afraid we will forget about it if we dont do that now 🙈

PR #5915, opened 2021-04-02
Display 'Recently added' headline when no collection is selected in media library

niklasnatter:
I noticed that the initial design for the media library includes a `Recently added` headline. I think it would be nice to display this headline if no collection is selected to explain that area below will show all recently added medias:

Screenshot 2021-04-02 at 14 11 27


alexander-schranz (approved):

niklasnatter:
@alexander-schranz noticed that the headline might be misleading if the user switches to the table view and changes the sorting of the medias 😕

niklasnatter:
Closing this because the headline is misleading if the user switches to the table view and changes the sorting of the medias

PR #5893, opened 2021-03-22
WIP: Improve CI Performance

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? | no 
| BC breaks? | no 
| Deprecations? | no  
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT
| Documentation PR | sulu/sulu-docs# 

#### What's in this PR?

Improve CI Performance

#### Why?

To CI on windows need about 32min which is a lot:

 - https://github.com/sulu/sulu/pull/5891/checks?check_run_id=2167382988
 - https://github.com/sulu/sulu/runs/2167182511

PR #5917, opened 2021-04-02
Run windows test cases with doctrinedbal instead of jackrabbit to improve performance

niklasnatter:
Related: https://github.com/sulu/sulu/pull/5893


niklasnatter:
Not sure if this will make a big difference, but I guess it is worth a try 🙂

niklasnatter:
From a first look, I think this decreases the time for the job from `~33 minutes` to about `~18 minutes`. These times will obviously vary between runs, but I think this should increase performance substantially 🙂

alexander-schranz (approved):

PR #5919, opened 2021-04-05
 Referencing controllers with a single colon is deprecated

StaffNowa:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | yes
| License | MIT

#### What's in this PR?
symfony/http-kernel 5.1 return deprecation warning

#### Why?

Change : to ::

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/config/routing.yml
+++ b/src/Sulu/Bundle/AdminBundle/Resources/config/routing.yml
@@ -1,22 +1,22 @@
 sulu_admin:
     path:  /
     defaults:
-        _controller: sulu_admin.admin_controller:indexAction
+        _controller: sulu_admin.admin_controller::indexAction
 
 sulu_admin.config:
     path: /config
     defaults:
-        _controller: sulu_admin.admin_controller:configAction
+        _controller: sulu_admin.admin_controller::configAction
 
 sulu_admin.metadata:
     path: /metadata/{type}/{key}
     defaults:
-        _controller: sulu_admin.admin_controller:metadataAction
+        _controller: sulu_admin.admin_controller::metadataAction
 
 sulu_admin.translation:
     path: /translations
     defaults:
-        _controller: sulu_admin.admin_controller:translationsAction
+        _controller: sulu_admin.admin_controller::translationsAction
 
 sulu_admin.login_check:
     path: /login
@@ -26,4 +26,4 @@ sulu_admin.logout:
 
 sulu_admin.reset:
     path: /reset/{token}
-    defaults: { _controller: SuluAdminBundle:Security:reset }
+    defaults: { _controller: SuluAdminBundle::Security:reset }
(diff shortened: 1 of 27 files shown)


src/Sulu/Bundle/AdminBundle/Resources/config/routing.yml
```diff
@@ -26,4 +26,4 @@ sulu_admin.logout:
 
 sulu_admin.reset:
     path: /reset/{token}
-    defaults: { _controller: SuluAdminBundle:Security:reset }
+    defaults: { _controller: SuluAdminBundle::Security:reset }
```
alexander-schranz: This route looks strange. I also don' think this controller does exist, so maybe the whole route definition can be removed?

alexander-schranz (changes requested):
@StaffNowa thx for the pull request can you rebase then the phpstan error should be gone.

src/Sulu/Bundle/AdminBundle/Resources/config/routing.yml
StaffNowa: Possible we need to have 

```
defaults: { _controller: SuluAdminBundle::Security::reset }
```

not this one

```
defaults: { _controller: SuluAdminBundle:Security:reset }
```

alexander-schranz: The controller does not longer exist since sulu 2.x so this route can be removed as it never did anything in 2.x.

StaffNowa: This one in Sulu 2.1 I found :)

alexander-schranz: What you mean you found? In any 2.x version there does the SecurityController not longer exist see: 

 - [https://github.com/sulu/sulu/tree/2.1/src/Sulu/Bundle/AdminBundle/Controller](https://github.com/sulu/sulu/tree/2.1/src/Sulu/Bundle/AdminBundle/Controller)
 - [https://github.com/sulu/sulu/tree/2.2/src/Sulu/Bundle/AdminBundle/Controller](https://github.com/sulu/sulu/tree/2.2/src/Sulu/Bundle/AdminBundle/Controller)
 - [https://github.com/sulu/sulu/tree/2.x/src/Sulu/Bundle/AdminBundle/Controller](https://github.com/sulu/sulu/tree/2.x/src/Sulu/Bundle/AdminBundle/Controller)
 
 only in [1.6](https://github.com/sulu/sulu/tree/1.6/src/Sulu/Bundle/AdminBundle/Controller) this did exist. So this route can safely be removed as it currently ends in a 500 error when called e.g.: [http://sulu.localhost/admin/reset/test](http://sulu.localhost/admin/reset/test) and so does nothing. This way completed replaced by the security bundle resetting route.

StaffNowa: You are right. We do not have this controller at all

niklasnatter (approved):
Looks good to me - thanks a lot! 

The failing pipeline is cause by a new version of the `friendsofphp/php-cs-fixer` package and not related to this PR.

alexander-schranz (approved):

PR #5920, opened 2021-04-06
Fix route generation with correct translator locale

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes https://sulu-io.slack.com/archives/C0430GGCV/p1617448633344600
| Related issues/PRs | none
| License | MIT
| Documentation PR | none

#### What's in this PR?

This PR fixes the generation of localized routes by setting the locale to the translator in the RouteContoller

#### Why?

See discussion in slack https://sulu-io.slack.com/archives/C0430GGCV/p1617448633344600

#### To Do

- [ ] Add Tests

### DIFF
--- a/src/Sulu/Bundle/RouteBundle/Controller/RouteController.php
+++ b/src/Sulu/Bundle/RouteBundle/Controller/RouteController.php
@@ -26,6 +26,8 @@ use Sulu\Component\Rest\RequestParametersTrait;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
+use Symfony\Contracts\Translation\LocaleAwareInterface;
+use Symfony\Contracts\Translation\TranslatorInterface;
 
 /**
  * Provides api to handle routes.
@@ -59,13 +61,19 @@ class RouteController extends AbstractRestController implements ClassResourceInt
      */
     private $conflictResolver;
 
+    /**
+     * @var TranslatorInterface|null
+     */
+    private $translator;
+
     public function __construct(
         ViewHandlerInterface $viewHandler,
         RouteRepositoryInterface $routeRepository,
         EntityManagerInterface $entityManager,
         RouteGeneratorInterface $routeGenerator,
         array $resourceKeyMappings,
-        ConflictResolverInterface $conflictResolver = null
+        ConflictResolverInterface $conflictResolver = null,
+        TranslatorInterface $translator = null
     ) {
         parent::__construct($viewHandler);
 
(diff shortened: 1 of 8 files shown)


src/Sulu/Bundle/RouteBundle/Controller/RouteController.php
```diff
@@ -191,4 +208,19 @@ private function generateUrlResponse(Request $request)
 
         return $this->handleView($this->view(['resourcelocator' => $route]));
     }
+
+    private function generateUrl(string $locale, array $parts, array $options): string
+    {
+        if (!$this->translator || !$this->translator instanceof LocaleAwareInterface){
+            return $this->routeGenerator->generate($parts, $options);
```
niklasnatter: what do you think about passing the `locale` as option inside of the `$options` parameter here? that way we dont need to think about this in other places that use the `RouteGenerator`

niklasnatter: think it would be nice to move the handling of the locale from the entity to `RouteGenerator` too when doing this:
https://github.com/sulu/sulu/blob/5c7a15c250c01ee4337f1fd1c44a34b2e9035147/src/Sulu/Bundle/RouteBundle/Generator/SymfonyExpressionTokenProvider.php#L57-L59

alexander-schranz: I updated the PR to adding the locale to the options which us give to the generator.

src/Sulu/Bundle/RouteBundle/Controller/RouteController.php
```diff
@@ -167,20 +167,23 @@ public function cdeleteAction(Request $request)
...
         $resourceKeyMapping = $this->resourceKeyMappings[$resourceKey] ?? null;
 
         /** @var array $parts */
         $parts = $this->getRequestParameter($request, 'parts', true);
 
         $route = '/' . \implode('-', $parts);
         if ($resourceKeyMapping) {
+            $options = $this->resourceKeyMappings[$resourceKey]['options'];
+            $options['locale'] = $locale;
             $route = $this->routeGenerator->generate($parts, $this->resourceKeyMappings[$resourceKey]['options']);
```
niklasnatter: think you need to pass `$options` here? 🤔

src/Sulu/Bundle/RouteBundle/Tests/Unit/Generator/RouteGeneratorTest.php
```diff
@@ -46,8 +47,8 @@ public function testGenerate()
     {
         $entity = $this->prophesize(RoutableInterface::class);
 
-        $this->tokenProvider->provide($entity->reveal(), 'object.getTitle()')->willReturn('Test Title');
```
niklasnatter, line 49: think we can use `['route_schema' => '/prefix/{object.getTitle()}/postfix/{object.getId()}']` instead of `Argument::any()`?

src/Sulu/Bundle/RouteBundle/Controller/RouteController.php
niklasnatter: would be nice to add a functional test case in the `RouteControllerTest`

src/Sulu/Bundle/RouteBundle/Controller/RouteController.php
niklasnatter: 🎉

src/Sulu/Bundle/RouteBundle/Controller/RouteController.php
alexander-schranz: I think we need todo this in the #5965

PR #5926, opened 2021-04-08
Don't remove trailing zeroes when parsing values in Router.js

Xheradon:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Added an extra check to see if a value is a number with trailing zeroes and if so, return it as a string instead of parsing it as a number/float and getting rid of those zeroes.

#### Why?

Route generation is stripping trailing zeroes, which leads to 404 errors when trying to edit an object.

Before: `/admin/#/products/9771/details`
After:   `/admin/#/products/09771/details`

This is only a problem when the id is saved in the database as a string and not a zerofilled integer.

It might also be possible to not check if the value is a number and return the value early, or parse values starting with `.` only. For now the commit only handles values starting with `0` and not followed by a `.`.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/services/Router/Router.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/services/Router/Router.js
@@ -38,7 +38,7 @@ function tryParse(value: ?string) {
         }
     }
 
-    if (isNaN(value)) {
+    if (isNaN(value) || (value && value.match(/0+[^.].*/))) {
         return value;
     }
 
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/services/Router/Router.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/services/Router/Router.js
@@ -38,7 +38,11 @@ function tryParse(value: ?string) {
         }
     }
 
-    if (isNaN(value) || (value && value.match(/0+[^.].*/))) {
+    if (isNaN(value)) {
+        return value;
+    }
+
+    if (value && value.match(/0+[^.].*/)) {
         return value;
     }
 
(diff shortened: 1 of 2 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/services/Router/Router.js
```diff
@@ -39,7 +39,7 @@ function tryParse(value: ?string) {
         }
     }
 
-    if (isNaN(value)) {
```
niklasnatter: would prefer to do a `value && value.startsWith('0')` in a separate check. also it would be great to add a test to the `Router.test.js` to make sure that this will not be changed in the future 🙂

niklasnatter (commented):
Thanks for your time!

src/Sulu/Bundle/AdminBundle/Resources/js/services/Router/Router.js
Xheradon: Using `value.startsWith('0')` will make any float that starts with 0 return as a string, I think that might create some problems with strict comparisons, that's why I used a regular expression. I think we could skip the `isNaN(value)` check and always return as a string at that point if we changed to loose comparisons but that might cause some problems on its own.

I'll try to write some tests on monday and make the check separately.

Thanks for the feedback!

niklasnatter: you are right - thanks for thinking this through 🙂

niklasnatter (approved):
Changes look good. I have rebased the branch to target `2.1`. Thanks for working on this!

PR #5930, opened 2021-04-09
Not analyze route if doesn't match

plozmun:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no 
| Deprecations? | -
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

After symfony route match a 404 or redirect route, some routes are analyzed and throw a UrlMatchNotFoundException showing a basic 404 symfony error. 

This should solve this kind or error https://sulu.rocks/fdsafd.txt ( Sulu error page must be shown )

#### Why?

Because we want to show a sulu error page or redirect

### DIFF
--- a/src/Sulu/Bundle/WebsiteBundle/EventListener/RouterListener.php
+++ b/src/Sulu/Bundle/WebsiteBundle/EventListener/RouterListener.php
@@ -11,6 +11,7 @@
 
 namespace Sulu\Bundle\WebsiteBundle\EventListener;
 
+use Sulu\Component\Webspace\Analyzer\Exception\UrlMatchNotFoundException;
 use Sulu\Component\Webspace\Analyzer\RequestAnalyzerInterface;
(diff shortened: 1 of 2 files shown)


alexander-schranz:
This is at current state expected that you get an error when no webspace is matched as a webspace is required also to render 404 pages. If you have a route that can be called without any webspace like a redirect route make sure that the following is set on your route: https://github.com/sulu/sulu/blob/5c7a15c250c01ee4337f1fd1c44a34b2e9035147/src/Sulu/Bundle/MediaBundle/Resources/config/routing_website.yml#L21

plozmun:
Solved , thanks @alexander-schranz . Sorry for the noise

alexander-schranz:
No problem 👍

PR #5936, opened 2021-04-14
Add tag and category filter on contact, account and media list

### LINKED ISSUE #5469: Filter on tags and categories for media, people and organizations in the Admin UI List overview
# Problem description

Currently I cannot search or filter on a tag or category when searching for a media, people or organization in the Admin UI List overview.

# Proposed solution

My proposal is to add two filters for media, people and organizations; tags and categories in the Admin UI List overview.

niklasnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Fixed tickets | fixes #5469
| Related issues/PRs | depends on #5805, depends on #5943
| License | MIT

#### What's in this PR?

This PR implements a POC for adding a tag filter and a category filter on the list view for contacts, accounts and medias:
Screenshot 2021-04-14 at 17 23 19

#### Why?

See #5469

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/fieldFilterTypes/SelectionFieldFilterType.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/fieldFilterTypes/SelectionFieldFilterType.js
@@ -3,6 +3,7 @@ import React from 'react';
 import type {ElementRef} from 'react';
 import {action, autorun, computed, observable, toJS, untracked, when} from 'mobx';
 import equals from 'fast-deep-equal';
+import userStore from '../../../stores/userStore';
 import MultiSelectionStore from '../../../stores/MultiSelectionStore';
 import MultiAutoComplete from '../../MultiAutoComplete';
 import ResourceCheckboxGroup from '../../ResourceCheckboxGroup';
@@ -30,7 +31,11 @@ class SelectionFieldFilterType extends AbstractFieldFilterType<?Array<string | n
     ) {
         super(onChange, parameters, value);
 
-        this.selectionStore = new MultiSelectionStore(this.resourceKey, []);
+        this.selectionStore = new MultiSelectionStore(
+            this.resourceKey,
+            [],
+            computed(() => userStore.contentLocale)
+        );
 
         this.selectionStoreDisposer = autorun(() => {
             const {onChange, selectionStore} = this;
--- a/src/Sulu/Bundle/ContactBundle/Controller/AccountController.php
+++ b/src/Sulu/Bundle/ContactBundle/Controller/AccountController.php
@@ -391,6 +391,7 @@ class AccountController extends AbstractRestController implements ClassResourceI
         if ('true' == $request->get('flat')) {
             $fieldDescriptors = $this->getFieldDescriptors();
             $listBuilder = $this->generateFlatListBuilder();
+            $listBuilder->distinct(true);
             $this->restHelper->initializeListBuilder($listBuilder, $fieldDescriptors);
             $this->applyRequestParameters($request, $listBuilder);
 
--- a/src/Sulu/Bundle/ContactBundle/Controller/AccountController.php
+++ b/src/Sulu/Bundle/ContactBundle/Controller/AccountController.php
@@ -389,9 +389,9 @@ class AccountController extends AbstractRestController implements ClassResourceI
         $locale = $this->getUser()->getLocale();
 
         if ('true' == $request->get('flat')) {
-            $fieldDescriptors = $this->getFieldDescriptors();
             $listBuilder = $this->generateFlatListBuilder();
             $listBuilder->distinct(true);
+            $fieldDescriptors = $this->getFieldDescriptors();
             $this->restHelper->initializeListBuilder($listBuilder, $fieldDescriptors);
             $this->applyRequestParameters($request, $listBuilder);
 
--- a/src/Sulu/Bundle/ContactBundle/Controller/AccountController.php
+++ b/src/Sulu/Bundle/ContactBundle/Controller/AccountController.php
@@ -390,7 +390,6 @@ class AccountController extends AbstractRestController implements ClassResourceI
 
         if ('true' == $request->get('flat')) {
             $listBuilder = $this->generateFlatListBuilder();
-            $listBuilder->distinct(true);
             $fieldDescriptors = $this->getFieldDescriptors();
             $this->restHelper->initializeListBuilder($listBuilder, $fieldDescriptors);
             $this->applyRequestParameters($request, $listBuilder);
--- a/src/Sulu/Bundle/ContactBundle/Resources/config/lists/accounts.xml
+++ b/src/Sulu/Bundle/ContactBundle/Resources/config/lists/accounts.xml
@@ -265,5 +265,28 @@
                 </params>
             </filter>
         </property>
+
+        <property
+            name="categoryId"
+            visibility="never"
+            translation="sulu_category.categories"
+        >
+            <field-name>id</field-name>
+            <entity-name>SuluCategoryBundle:Category</entity-name>
+
+            <joins>
+                <join>
+                    <entity-name>SuluCategoryBundle:Category</entity-name>
+                    <field-name>%sulu.model.account.class%.categories</field-name>
+                </join>
+            </joins>
+
+            <filter type="selection">
+                <params>
+                    <param name="displayProperty" value="name" />
+                    <param name="resourceKey" value="categories" />
+                </params>
+            </filter>
+        </property>
     </properties>
 </list>
--- a/src/Sulu/Bundle/ContactBundle/Resources/config/lists/accounts.xml
+++ b/src/Sulu/Bundle/ContactBundle/Resources/config/lists/accounts.xml
@@ -22,6 +22,13 @@
         </join>
     </joins>
 
+    <joins name="tags">
+        <join>
+            <entity-name>SuluTagBundle:Tag</entity-name>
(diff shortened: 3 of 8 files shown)


src/Sulu/Bundle/ContactBundle/Resources/config/lists/accounts.xml
```diff
@@ -247,12 +254,25 @@
...
-                    <entity-name>SuluTagBundle:Tag</entity-name>
-                    <field-name>%sulu.model.account.class%.tags</field-name>
-                </join>
-            </joins>
+            <joins ref="tags"/>
         </group-concat-property>
+
+        <property
+            name="tagId"
+            visibility="never"
```
nnatter, line 262: it is important that this property will never be included in the `SELECT` statement, otherwise accounts with multiple tags will be duplicated in the response

src/Sulu/Bundle/ContactBundle/Controller/AccountController.php
```diff
@@ -391,6 +391,7 @@ public function cgetAction(Request $request)
         if ('true' == $request->get('flat')) {
             $fieldDescriptors = $this->getFieldDescriptors();
             $listBuilder = $this->generateFlatListBuilder();
+            $listBuilder->distinct(true);
```
niklasnatter: without enable `DISTINCT`, accounts with multiple tags will be duplicated in the response if the tag filter is active

niklasnatter: TODO: add a test cases for the controllers to check this

alexander-schranz (approved):

src/Sulu/Bundle/ContactBundle/Resources/config/lists/accounts.xml
```diff
@@ -247,12 +254,25 @@
...
+            translation="sulu_tag.tags"
+        >
+            <field-name>id</field-name>
+            <entity-name>SuluTagBundle:Tag</entity-name>
+
+            <joins ref="tags"/>
+
+            <filter type="selection">
```
nnatter, line 270: we cannot do this for `categories` at the moment because the `SelectionFieldFilterType` does not send any locale to the list api

nnatter: TODO: adjust the `SelectionFieldFilterType` to be consistent with the `Selection`:
https://github.com/sulu/sulu/blob/33b1356d979ac5095ff9f4d32be0ad0aac24856e/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/Selection.js#L220-L224

nnatter: This change depends on https://github.com/sulu/sulu/pull/5805, because the `locale` parameter of the `SearchStore` is added in https://github.com/sulu/sulu/pull/5805

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/fieldFilterTypes/SelectionFieldFilterType.js
```diff
@@ -30,7 +31,11 @@ class SelectionFieldFilterType extends AbstractFieldFilterType<?Array<string | n
     ) {
         super(onChange, parameters, value);
 
-        this.selectionStore = new MultiSelectionStore(this.resourceKey, []);
+        this.selectionStore = new MultiSelectionStore(
+            this.resourceKey,
+            [],
+            computed(() => userStore.contentLocale)
```
niklasnatter, line 37: this is necessary because the `CategoryController` will throw an error if the request does not include any locale. unfortunately this change depends on #5805, because the `SearchStore` does not have any locale before that PR:
https://github.com/sulu/sulu/pull/5805/files#diff-9f4b5bb9b4570d8c7e2c3314b256e334a86bb15875300447290886166230715f

Because of this, this PR needs to target `2.x`.

niklasnatter: TODO: add a test cases for this change

src/Sulu/Bundle/ContactBundle/Controller/AccountController.php
niklasnatter: this is necessary because all joins of all expressions fields are added to the query by the `DoctrineListBuilder` at the moment:
https://github.com/sulu/sulu/blob/7833cdced3878fe25303c0f93c99daeba1e5e6b8/src/Sulu/Component/Rest/ListBuilder/Doctrine/DoctrineListBuilder.php#L370

i am not sure if there is a reason for this yet. as far as i understand, it should be enough to add the joins for `searchFields` and `expressionFields` in the query that selects the ids only:
https://github.com/sulu/sulu/blob/2.x/src/Sulu/Component/Rest/ListBuilder/Doctrine/DoctrineListBuilder.php#L267-L300

src/Sulu/Component/Rest/ListBuilder/Doctrine/DoctrineListBuilder.php
```diff
@@ -367,7 +367,7 @@ protected function assignGroupBy($queryBuilder)
     protected function getJoins()
     {
         $joins = [];
-        $fields = \array_merge($this->sortFields, $this->selectFields, $this->searchFields, $this->expressionFields);
```
alexander-schranz, line 370: I don't think we can remove the `$this->expressionFields` as when we have a `where` condition on a join which was not sorted or selected the where will fail: https://github.com/sulu/sulu/blob/a9a5deb1d306f36960a00e56f15e7d64c8819435/src/Sulu/Component/Rest/ListBuilder/Doctrine/DoctrineListBuilder.php#L553

niklasnatter: i havent had a deeper look at this yet to be honest. in theory, this `WHERE` condition should only be needed on the query that selects the ids (not on the the main query), right?

niklasnatter: i think the part of the code is only called in these two places, so its not used for the "main" query. but i will need to add an additional check for the first call
https://github.com/sulu/sulu/blob/7833cdced3878fe25303c0f93c99daeba1e5e6b8/src/Sulu/Component/Rest/ListBuilder/Doctrine/DoctrineListBuilder.php#L208-L213
https://github.com/sulu/sulu/blob/7833cdced3878fe25303c0f93c99daeba1e5e6b8/src/Sulu/Component/Rest/ListBuilder/Doctrine/DoctrineListBuilder.php#L426

alexander-schranz: Yes. Seems like the case without limit also checks that there is no expression: https://github.com/sulu/sulu/blob/a9a5deb1d306f36960a00e56f15e7d64c8819435/src/Sulu/Component/Rest/ListBuilder/Doctrine/DoctrineListBuilder.php#L209-L212
So I think this is save. I would like to backport this change then to the `2.1` branch.

niklasnatter: at least it looks like there are no failing test cases. but i think i will need to add an additional parameter because we still need to join the search fields in this case 🙈 
https://github.com/sulu/sulu/blob/7833cdced3878fe25303c0f93c99daeba1e5e6b8/src/Sulu/Component/Rest/ListBuilder/Doctrine/DoctrineListBuilder.php#L208-L213

will backport the change onto 2.1 once it works here 🙂

src/Sulu/Component/Rest/ListBuilder/Doctrine/DoctrineListBuilder.php
```diff
@@ -362,12 +362,18 @@ protected function assignGroupBy($queryBuilder)
     /**
      * Returns all the joins required for the query.
      *
+     * @param bool $includeFilterFields Define if joins of filtering FieldDescriptors should be returned
+     *
      * @return DoctrineJoinDescriptor[]
      */
-    protected function getJoins()
+    protected function getJoins($includeFilterFields = true)
```
nnatter, line 369: TODO: backport this onto `2.1`

src/Sulu/Bundle/ContactBundle/Resources/config/lists/accounts.xml
```diff
@@ -238,12 +245,52 @@
             <field-name>id</field-name>
             <entity-name>SuluTagBundle:Tag</entity-name>
 
+            <joins ref="tags"/>
+        </group-concat-property>
+
+        <!-- tagId property is only used for filtering and should never be included as column in a list -->
+        <!-- when used as column, the property will lead to duplicated rows for entities with multiple tags -->
```
nnatter, line 252: TODO: add a testcase to ensure that this does not happen when only used for filtering

nnatter:
Depends on #5943

src/Sulu/Component/Rest/ListBuilder/Doctrine/DoctrineListBuilder.php
niklasnatter: backported in #5943

src/Sulu/Bundle/ContactBundle/Resources/config/lists/accounts.xml
nnatter: this depends on #5943

niklasnatter:
Replaced by #5943 and #5944

PR #5937, opened 2021-04-16
Add country to twig localization parameter

plozmun:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? | yes
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | https://github.com/sulu/sulu-docs/pull/643

#### What's in this PR?

Add country parameter in localization if exists

#### Why?

Because is useful for developers to build complex language switchers by country

#### Example Usage

```twig
{% for localization in localizations %}
    {% if localization.country is defined and localization.country == 'de' %}
        ....
    {% endif %}
{% endfor %}
```

#### To Do

### DIFF
--- a/src/Sulu/Bundle/WebsiteBundle/Resolver/ParameterResolver.php
+++ b/src/Sulu/Bundle/WebsiteBundle/Resolver/ParameterResolver.php
@@ -12,9 +12,9 @@
 namespace Sulu\Bundle\WebsiteBundle\Resolver;
 
 use Sulu\Component\Content\Compat\StructureInterface;
-use Sulu\Component\Localization\Localization;
 use Sulu\Component\Webspace\Analyzer\RequestAnalyzerInterface;
 use Sulu\Component\Webspace\Manager\WebspaceManagerInterface;
+use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\RequestStack;
 
 /**
@@ -37,7 +37,7 @@ class ParameterResolver implements ParameterResolverInterface
      */
     private $webspaceManager;
 
-    /*
+    /**
      * @var RequestStack
      */
     private $requestStack;
@@ -85,6 +85,10 @@ class ParameterResolver implements ParameterResolverInterface
             $structureData = [];
         }
 
+        if (!$requestAnalyzer) {
+            return \array_merge($parameter, $structureData);
+        }
+
         $requestAnalyzerData = $this->requestAnalyzerResolver->resolve($requestAnalyzer);
         $webspace = $requestAnalyzer->getWebspace();
 
@@ -102,7 +106,6 @@ class ParameterResolver implements ParameterResolverInterface
         $localizations = [];
 
         foreach ($allLocalizations as $localization) {
-            /* @var Localization $localization */
             $locale = $localization->getLocale();
 
             if (\array_key_exists($locale, $pageUrls)) {
@@ -114,6 +117,7 @@ class ParameterResolver implements ParameterResolverInterface
             $localizations[$locale] = [
                 'locale' => $locale,
                 'url' => $url,
+                'country' => $localization->getCountry(),
             ];
         }
 
@@ -128,7 +132,10 @@ class ParameterResolver implements ParameterResolverInterface
 
         $structureData['localizations'] = $localizations;
 
-        $url = $this->requestStack->getCurrentRequest()->getUri();
+        /** @var Request $request */
+        $request = $this->requestStack->getCurrentRequest();
+
+        $url = $request->getUri();
 
         $segments = [];
         foreach ($webspace->getSegments() as $segment) {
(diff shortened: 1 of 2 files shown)


src/Sulu/Bundle/WebsiteBundle/Resolver/ParameterResolver.php
```diff
@@ -85,6 +85,10 @@ public function resolve(
             $structureData = [];
         }
 
+        if (!$requestAnalyzer) {
+            return [];
```
plozmun: I dont know if its better throw a LogicException here if requestAnalyzer does not exists

src/Sulu/Bundle/WebsiteBundle/Resolver/ParameterResolver.php
```diff
@@ -115,6 +118,10 @@ public function resolve(
                 'locale' => $locale,
                 'url' => $url,
             ];
+
+            if ($country = $localization->getCountry()) {
```
plozmun: I have set optional if exists. It should be always present with null value?

src/Sulu/Bundle/WebsiteBundle/Resolver/ParameterResolver.php
niklasnatter: not sure if there is a valid usecase where this is called without a `$requestAnalyzer` 🙈 
think i would just return the `$parameter` array in this case to be safe

src/Sulu/Bundle/WebsiteBundle/Resolver/ParameterResolver.php
niklasnatter: think i would just always set it in the assignment above. or do you see any problems with that?

niklasnatter (commented):
Thanks for working on this. I fixed the unrelated phpstan error on the base branch 🙂 

Could you rebase this onto the `2.2` branch?
```
git rebase --interactive --onto origin/2.2 origin/2.x
```

src/Sulu/Bundle/WebsiteBundle/Resolver/ParameterResolver.php
plozmun: No, there is no problem, i think it's better. I'll change it

plozmun:
@nnatter updated :)

src/Sulu/Bundle/WebsiteBundle/Resolver/ParameterResolver.php
```diff
@@ -114,6 +117,7 @@ public function resolve(
             $localizations[$locale] = [
                 'locale' => $locale,
                 'url' => $url,
+                'country' => $localization->getCountry(),
```
niklasnatter, line 120: just tested this - of we add `country`.  i think it would be nice to add `'language' => $localization->getLanguage()` here too. would you mind doing that in this PR? 🙂

plozmun: > just tested this - of we add `country`. i think it would be nice to add `'language' => $localization->getLanguage()` here too. would you mind doing that in this PR? 🙂

@nnatter done!

niklasnatter (approved):
Thanks - tested this and it looks good to me! 
But I am not really an expert on this part of the system - would be nice if @alexander-schranz  could have a quick look at it too 🙂

src/Sulu/Bundle/WebsiteBundle/Tests/Unit/Resolver/ParameterResolverTest.php
```diff
@@ -245,10 +262,94 @@ public function testResolveWithoutUrlsParameter()
...
+            'view' => [],
+            'extension' => [
+                'seo' => [],
+                'excerpt' => [],
+            ],
+            'localizations' => [
+                'en_gb' => [
+                    'locale' => 'en_gb',
+                    'url' => '/en/test',
+                    'language' => 'en',
```
alexander-schranz: I discussed this with @nnatter I would not add the `language` at current state. I'm afraid that somebody use the `language` instead of `locale` in `sulu_content_path` and then they are not sure why it don't work like expected. We had in the past a lot of confusen with path and url which where also very similar and a lot did fall into this pitfall so alson as there is no really need for the language I would avoid creating this pitfall here.

Sorry for inconveniences!

alexander-schranz (changes requested):

src/Sulu/Bundle/WebsiteBundle/Tests/Unit/Resolver/ParameterResolverTest.php
niklasnatter: @plozmun i am sorry - i did not think about this 🙈 i think it makes sense to no add the `language` to avoid confusion

plozmun: Ok! no problem. Maybe we can change these two parameters to `iso_language` and `iso_region` witch defines better the ISO specification. WDYT?

alexander-schranz: Actually we don't know if they did input a iso languages or iso countries its not validated and sadly did see projects using none iso strings in there localization, so I would just stay with `country` here at current state.

plozmun: Ok, language removed :)

alexander-schranz (approved):

niklasnatter:
Thanks for your patience! 🙂

PR #5939, opened 2021-04-16
Add domain events for MediaBundle

luca-rath:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | yes
| Deprecations? | yes 
| Fixed tickets | -
| Related issues/PRs | #5840
| License | MIT
| Documentation PR | -

#### What's in this PR?

Add domain events for MediaBundle


luca-rath:
Currently, if a collection with children is deleted, all children get deleted as well. This behaviour is handled by the gedmo tree extension. But this leads to the problem, that if a collection with children is deleted, sulu is only throwing a `removed` event for the specific collection, which was deleted by the user, not for the children of that collection. So how should that be fixed?
/cc @sulu/core-team 

PS: The same applies for categories

alexander-schranz:
> Currently, if a collection with children is deleted, all children get deleted as well.

Thomas and I had something similar discussed today. That tree removes should not relay on cascading and some kind of batch process should be done here. In some cases this batch process can be take a lot of time so it would be some kind of batch processing with multiple requests from the client side, also permission is here a big topic which is currently not keeped in mind with cascading.

So at the end in future we should throw exception if we try to remove a collection which has child collection or medias.

At current state for this PR with the domain events I think its enough that we throw just the CollectionRemovedEvent on collection manager remove. So I don't see the children part now something we are fixing in this PR, it should be an own PR.

luca-rath:
> At current state for this PR with the domain events I think its enough that we throw just the CollectionRemovedEvent on collection manager remove. So I don't see the children part now something we are fixing in this PR, it should be an own PR.

sounds good 👍

src/Sulu/Bundle/MediaBundle/Media/Manager/MediaManager.php
```diff
@@ -951,4 +999,56 @@ private function getNormalizedFileName($originalFileName)
...
+        }
+
+        if ($currentFileVersion === $file->getLatestFileVersion()) {
+            throw new BadRequestHttpException('Can\'t delete active version of a media.');
+        }
+
+        $this->em->remove($currentFileVersion);
+        foreach ($currentFileVersion->getFormatOptions() as $formatOptions) {
+            $this->em->remove($formatOptions);
+        }
```
luca-rath, line 1042: I don't know why this is necessary here, but without this, https://github.com/sulu/sulu/pull/5939/checks?check_run_id=2380363925 one of those errors occurrs

src/Sulu/Bundle/MediaBundle/Tests/Functional/Controller/CollectionControllerTest.php
```diff
@@ -1193,6 +1193,8 @@ public function testDeleteById()
     {
         $collection1Id = $this->collection1->getId();
 
+        $this->em->clear();
```
luca-rath, line 1196: I don't know why this is necessary here, but without this, https://github.com/sulu/sulu/pull/5939/checks?check_run_id=2380363925 one of those errors occurrs

src/Sulu/Bundle/MediaBundle/Collection/Manager/CollectionManager.php
```diff
@@ -396,17 +410,40 @@ public function save(array $data, ?int $userId = null, bool $breadcrumb = false)
...
             $collection,
             $data
         );
 
-        $collectionEntity = $collection->getEntity();
         $this->em->persist($collectionEntity);
+
+        if ($isNewLocale) {
+            $this->domainEventCollector->collect(
+                new CollectionLocaleAddedEvent($collectionEntity, $collection->getLocale(), $data)
```
niklasnatter: think this should be `CollectionTranslationAddedEvent`

src/Sulu/Bundle/MediaBundle/Controller/MediaController.php
```diff
@@ -456,16 +461,43 @@ public function deleteAction($id)
      */
     public function deleteVersionAction(Request $request, $id, $version)
     {
-        $locale = $this->getRequestParameter($request, 'locale', true);
-        $media = $this->mediaManager->getById($id, $locale);
+        if (!\method_exists($this->mediaManager, 'removeFileVersion')) {
```
niklasnatter: not sure if this can happen. the constructor argument has a `MediaManagerInterface` typehint, so `$this->mediaManager` must be a `MediaManagerInterface`. i think php will throw an error if you dont implement a method of an interface, so i dont think this condition can be true🙂

src/Sulu/Bundle/MediaBundle/Controller/MediaPreviewController.php
```diff
@@ -94,6 +110,18 @@ public function postAction($id, Request $request)
             $mediaEntity->setPreviewImage($previewImage->getEntity());
             $this->mediaManager->addFormatsAndUrl($media);
 
+            // Because the `MediaManager::save()` method calls `$entityManager->flush()` itself, the `created` event of
+            // the preview image and the `preview_image_created`/`preview_image_modified` event are not in the same badge.
```
niklasnatter:
```suggestion
            // the preview image and the `preview_image_created`/`preview_image_modified` event are not in the same batch.
```

dont think this is critical, sguess we will just live with it 🙂

src/Sulu/Bundle/MediaBundle/Domain/Event/CollectionCreatedEvent.php
```diff
@@ -0,0 +1,120 @@
...
+
+    private function getCollectionMeta(): ?CollectionMeta
+    {
+        /** @var CollectionMeta|null $meta */
+        $meta = $this->collection->getDefaultMeta();
+        foreach ($this->collection->getMeta() as $collectionMeta) {
+            if ($collectionMeta->getLocale() === $this->locale) {
+                $meta = $collectionMeta;
+
+                break;
```
niklasnatter:
```suggestion
                return $collectionMeta;
```

src/Sulu/Bundle/MediaBundle/Domain/Event/CollectionCreatedEvent.php
```diff
@@ -0,0 +1,120 @@
...
+    {
+        $collectionMeta = $this->getCollectionMeta();
+
+        return $collectionMeta ? $collectionMeta->getLocale() : null;
+    }
+
+    private function getCollectionMeta(): ?CollectionMeta
+    {
+        /** @var CollectionMeta|null $meta */
+        $meta = $this->collection->getDefaultMeta();
```
niklasnatter, line 99: good idea to fallback to the default meta here 🙂  would you mind to adjust this for the category events too?

src/Sulu/Bundle/MediaBundle/Domain/Event/CollectionLocaleAddedEvent.php
```diff
@@ -0,0 +1,120 @@
...
+
+namespace Sulu\Bundle\MediaBundle\Domain\Event;
+
+use Sulu\Bundle\EventLogBundle\Domain\Event\DomainEvent;
+use Sulu\Bundle\MediaBundle\Admin\MediaAdmin;
+use Sulu\Bundle\MediaBundle\Entity\Collection;
+use Sulu\Bundle\MediaBundle\Entity\CollectionInterface;
+use Sulu\Bundle\MediaBundle\Entity\CollectionMeta;
+
+class CollectionLocaleAddedEvent extends DomainEvent
```
niklasnatter: think this should be `CollectionTranslationAddedEvent `

src/Sulu/Bundle/MediaBundle/Domain/Event/MediaCreatedEvent.php
```diff
@@ -0,0 +1,129 @@
...
+    private function getFileVersionMeta(): ?FileVersionMeta
+    {
+        $file = $this->media->getFiles()[0] ?? null;
+        $fileVersion = $file ? $file->getLatestFileVersion() : null;
+        $meta = $fileVersion ? $fileVersion->getDefaultMeta() : null;
+
+        if (null !== $fileVersion) {
+            foreach ($fileVersion->getMeta() as $fileVersionMeta) {
+                if ($fileVersionMeta->getLocale() === $this->locale) {
+                    $meta = $fileVersionMeta;
```
niklasnatter:
```suggestion
                    return $fileVersionMeta;
```

src/Sulu/Bundle/MediaBundle/Domain/Event/MediaLocaleAddedEvent.php
```diff
@@ -0,0 +1,129 @@
...
+
+namespace Sulu\Bundle\MediaBundle\Domain\Event;
+
+use Sulu\Bundle\EventLogBundle\Domain\Event\DomainEvent;
+use Sulu\Bundle\MediaBundle\Admin\MediaAdmin;
+use Sulu\Bundle\MediaBundle\Entity\Collection;
+use Sulu\Bundle\MediaBundle\Entity\FileVersionMeta;
+use Sulu\Bundle\MediaBundle\Entity\MediaInterface;
+
+class MediaLocaleAddedEvent extends DomainEvent
```
niklasnatter: think this should be `MediaTranslationAdded`

src/Sulu/Bundle/MediaBundle/Domain/Event/MediaPreviewImageCreatedEvent.php
```diff
@@ -0,0 +1,110 @@
...
+
+namespace Sulu\Bundle\MediaBundle\Domain\Event;
+
+use Sulu\Bundle\EventLogBundle\Domain\Event\DomainEvent;
+use Sulu\Bundle\MediaBundle\Admin\MediaAdmin;
+use Sulu\Bundle\MediaBundle\Entity\Collection;
+use Sulu\Bundle\MediaBundle\Entity\FileVersionMeta;
+use Sulu\Bundle\MediaBundle\Entity\MediaInterface;
+
+class MediaPreviewImageCreatedEvent extends DomainEvent
```
niklasnatter: think we should caa this `MediaPreviewImageAddedEvent` because the media that is added as preview image might have already existed before

niklasnatter: or maybe just use the `MediaPreviewImageModifiedEvent` and set `oldPreviewImageId` to null? i think that would be even simpler

src/Sulu/Bundle/MediaBundle/Domain/Event/MediaVersionCreatedEvent.php
```diff
@@ -0,0 +1,105 @@
...
+
+namespace Sulu\Bundle\MediaBundle\Domain\Event;
+
+use Sulu\Bundle\EventLogBundle\Domain\Event\DomainEvent;
+use Sulu\Bundle\MediaBundle\Admin\MediaAdmin;
+use Sulu\Bundle\MediaBundle\Entity\Collection;
+use Sulu\Bundle\MediaBundle\Entity\FileVersionMeta;
+use Sulu\Bundle\MediaBundle\Entity\MediaInterface;
+
+class MediaVersionCreatedEvent extends DomainEvent
```
niklasnatter: i know i suggested this name initially, but i think i would call this `MediaVersionAddedEvent` 🙈  what do you think?

niklasnatter: `added` would imply that it is added to an existing entity and would be consistent to the `translation_added` type. in hindsight, i think i would also use `keyword_added` instead of `keyword_created` in the category bundle now

src/Sulu/Bundle/MediaBundle/Media/Manager/MediaManager.php
```diff
@@ -351,6 +370,8 @@ private function modifyMedia($uploadedFile, $data, $user)
             throw new FileVersionNotFoundException($mediaEntity->getId(), $version);
         }
 
+        $shouldThrowModifiedEvent = true;
```
niklasnatter:
```suggestion
        $shouldOmitModifiedEvent = true;
```

src/Sulu/Bundle/MediaBundle/Media/Manager/MediaManager.php
niklasnatter: it is necessary because the `cascade` configurations in the media-bundle doctrine mappings are random and nobody understands them anymore 😅

src/Sulu/Bundle/MediaBundle/Media/Manager/MediaManagerInterface.php
```diff
@@ -19,6 +19,7 @@
 
 /**
  * @method string getAdminUrl($id, $fileName, $version);
+ * @method void removeFileVersion(int $mediaId, int $version);
```
niklasnatter: okay with this the check above actually makes sense. but as we are releasing a minor version, i would just add the method to the interface. should not be hard to implement this in a project and will save us a bit of pain in the future 🙂

src/Sulu/Bundle/MediaBundle/Domain/Event/MediaPreviewImageModifiedEvent.php
```diff
@@ -0,0 +1,118 @@
...
+
+    public function getEventType(): string
+    {
+        return 'preview_image_modified';
+    }
+
+    public function getEventContext(): array
+    {
+        return [
+            'oldPreviewImageId' => $this->oldPreviewImageId,
```
niklasnatter: i think you used `previous` instead of `old` in some other events

src/Sulu/Bundle/MediaBundle/Domain/Event/MediaVersionCreatedEvent.php
niklasnatter: for completion, think i will create a pr after this to rename the `keyword_created` event 😅

src/Sulu/Bundle/MediaBundle/Controller/MediaController.php
luca-rath: ~~The `MediaManagerInterface` doesn't define this method, it's just defined as phpdoc~~

src/Sulu/Bundle/MediaBundle/Media/Manager/MediaManager.php
luca-rath: I guess you meant `emit` 🙂

src/Sulu/Bundle/MediaBundle/Domain/Event/MediaPreviewImageCreatedEvent.php
luca-rath: Then I would rename this to `MediaPreviewImageChangedEvent`, because it's not a modification of the preview image itself, it's really just changing the preview image

src/Sulu/Bundle/MediaBundle/Domain/Event/CollectionCreatedEvent.php
luca-rath: Would do this in an extra pr

src/Sulu/Bundle/MediaBundle/Domain/Event/MediaPreviewImageCreatedEvent.php
luca-rath: @nnatter WDYT?

niklasnatter: yes sounds good to me! 🙂

src/Sulu/Bundle/MediaBundle/Media/Manager/MediaManager.php
niklasnatter: you are right 😅  `emit` or `dispatch`. `throw` sounds a bit like an exception

src/Sulu/Bundle/MediaBundle/Domain/Event/CollectionCreatedEvent.php
niklasnatter: yes that sounds good

src/Sulu/Bundle/MediaBundle/Domain/Event/MediaCropModifiedEvent.php
```diff
@@ -0,0 +1,125 @@
...
+
+namespace Sulu\Bundle\MediaBundle\Domain\Event;
+
+use Sulu\Bundle\EventLogBundle\Domain\Event\DomainEvent;
+use Sulu\Bundle\MediaBundle\Admin\MediaAdmin;
+use Sulu\Bundle\MediaBundle\Entity\Collection;
+use Sulu\Bundle\MediaBundle\Entity\FileVersionMeta;
+use Sulu\Bundle\MediaBundle\Entity\MediaInterface;
+
+class MediaCropModifiedEvent extends DomainEvent
```
niklasnatter, line 20: think we should call this `MediaCropChangedEvent` if we use `MediaPreviewImageChangedEvent` for the preview image 🙂

niklasnatter: okay not sure about this anymore, i guess it will be confusing when to use `Change` and when to use `Modify` 🙈 
maybe wo should just always use `Modify` to prevent this confusion?

src/Sulu/Bundle/MediaBundle/Domain/Event/MediaPreviewImageCreatedEvent.php
niklasnatter: not sure about this anymore. i am afraid that it will get confusing when to use `Modify` vs `Change` because they basically mean the same thing 🙈  
lets discuss this in https://github.com/sulu/sulu/pull/5939/files#r624992008

src/Sulu/Bundle/MediaBundle/Domain/Event/MediaCropModifiedEvent.php
luca-rath: I don't agree on that one ... I would use `changed` only, if you switch from one item in a predefined list to another ... For the preview image, the predefined list is the media gallery. So you just choose one image and change your selection later. But for crops, you really modify the crop ... The user is moving the square around and setting the exact position and size of that square as crop

luca-rath: Ah, github didn't show your latest comment ... my comment was related to the first one 🙈

niklasnatter: > I would use changed only, if you switch from one item in a predefined list to another ... For the preview image, the predefined list is the media gallery

Yeah I see that the differentiation would make sense in this case.. But I am afraid that the terms are too similar and once both terms are used in the codebase, somebody will use the wrong term unintentionally and we will end up with an inconsistent naming 🙈 
To prevent this, we could use a more explicit term like `preview_image_assigned`. But I am not sure if this is worth the effort. I think the simplest thing would be just using `modified` everywhere

niklasnatter (approved):

PR #5941, opened 2021-04-16
Use computed contentLocale from userStore instead of boxing value a single time

niklasnatter:
`computed(() => userStore.contentLocale)` will return an observable that is updated if the `userStore.contentLocale` is changed. 

I dont think this changes the behaviour anywhere, because `userStore.contentLocale` is only updated on view change. But we should still use the more appropriate method in my opinion 🙂

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/ResourceLocator.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/ResourceLocator.js
@@ -1,6 +1,6 @@
 // @flow
 import React from 'react';
-import {action, observable, toJS} from 'mobx';
+import {action, computed, observable, toJS} from 'mobx';
 import {observer} from 'mobx-react';
 import ResourceLocatorComponent from '../../../components/ResourceLocator';
 import ResourceLocatorHistory from '../../../containers/ResourceLocatorHistory';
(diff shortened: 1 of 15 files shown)


alexander-schranz (approved):
Looks good for me. Was there any bug or something we need test?

niklasnatter:
> Looks good for me. Was there any bug or something we need test?

Not really, I just figured that this would be the better way when implementing https://github.com/sulu/sulu/pull/5936. I think the change does not affect the behaviour because we update the `userStore.contentLocale` only when changing views and this will rerender everything anyway 🙂

alexander-schranz:
@nnatter looking again on this, could this not end in strange sideeffects. I open a FormStore which opens an overlay with a Form in another locale and then accidently my main form has a new contentLocale? I'm not sure if its good to use `computed` here and refresh the forms when some subform maybe change the user.contentLocale.

alexander-schranz:
I think using `userStore.contentLocale` on initialization is okay but refreshing based on it can have strange sideeffects when having multiple forms on a view.

niklasnatter:
The `userStore.contentLocale` is used on unlocalized forms only, so not sure if this could happen. But its possible that there are cases similar to what you have described. Will close this PR for now as this does no cause any problems at the moment 🙂

alexander-schranz:
Aslong as it currently doesn't make any problems I would let it as it is.

PR #5942, opened 2021-04-18
Enable Admin Bundle in website context

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? | yes
| BC breaks? | no 
| Deprecations? | no
| Fixed tickets | fixes # 
| Related issues/PRs | part of #5902
| License | MIT

#### What's in this PR?

Enable Admin Bundle in website context.

#### Why?

The SuluAdminBundle contains services like list and form metadata which also make sense to be used on the website context, also in future we should make sure that sulu works with a single kernel so step by step we should make sure that there is no longer any context based service.

#### To Do


src/Sulu/Bundle/TestBundle/Kernel/SuluTestKernel.php
```diff
@@ -103,7 +104,6 @@ public function registerBundles()
 
         if (self::CONTEXT_ADMIN === $this->getContext()) {
             $bundles[] = new \Symfony\Bundle\SecurityBundle\SecurityBundle();
-            $bundles[] = new \Sulu\Bundle\AdminBundle\SuluAdminBundle();
```
niklasnatter, line 106: shouldnt we do this for the other bundles in this block too? 🤔

alexander-schranz: its a step by step, take some work todo for all.

niklasnatter:
Related to https://github.com/sulu/skeleton/pull/110

src/Sulu/Component/HttpKernel/SuluKernel.php
```diff
@@ -82,6 +82,14 @@ protected function configureContainer(ContainerBuilder $container, LoaderInterfa
         $container->addResource(new FileResource($this->getProjectDir() . '/config/bundles.php'));
         $confDir = $this->getProjectDir() . '/config';
 
+        // this is a bc layer that sulu_admin.yaml is only loaded when bundle is registered
+        // this can be removed when there is no longer context based configuration
+        $bundles = $container->getParameter('kernel.bundles');
+        if (isset($bundles['SuluAdminBundle'])) {
```
alexander-schranz, line 89: @nnatter this fixes your comment here: https://github.com/sulu/skeleton/pull/110#issuecomment-822359021

PR #5943, opened 2021-04-19
Add tag filter on contact, account and media list

niklasnatter:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Related issues/PRs | preparation for #5936
| License | MIT

#### What's in this PR?

This PR add a tag filter on the list view for contacts, accounts and medias. To prevent duplicated rows when filtering, it adjusts the `DoctrineListBuilder` to not add joins related to expressions and search to the main query builder.

Screenshot 2021-04-14 at 17 23 19

#### Why?

See #5469

The changes to the `DoctrineListBuilder` are needed because a contact (same for account/media) can be assigned to multiple tags. If we want to filter contacts for a specific tag, we need to join all tags of the contact in the "preselect" query. If we add this join in the "main" query too, the result will contain duplicated rows for contacts with multiple tags.

Because of this, joins that are only required for filtering should be added to the "preselect" query that is created in the `findIdsByGivenCriteria()` method only.

### DIFF
--- a/src/Sulu/Bundle/ContactBundle/Resources/config/lists/accounts.xml
+++ b/src/Sulu/Bundle/ContactBundle/Resources/config/lists/accounts.xml
@@ -22,6 +22,13 @@
         </join>
     </joins>
 
+    <joins name="tags">
+        <join>
+            <entity-name>SuluTagBundle:Tag</entity-name>
+            <field-name>%sulu.model.account.class%.tags</field-name>
+        </join>
+    </joins>
+
     <properties>
         <property name="logo" visibility="always"
             translation="sulu_contact.logo"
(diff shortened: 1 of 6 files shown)


src/Sulu/Bundle/ContactBundle/Resources/config/lists/accounts.xml
```diff
@@ -247,12 +254,27 @@
...
-                <join>
-                    <entity-name>SuluTagBundle:Tag</entity-name>
-                    <field-name>%sulu.model.account.class%.tags</field-name>
-                </join>
-            </joins>
+            <joins ref="tags"/>
         </group-concat-property>
+
+        <!-- tagId property is only used for filtering and should never be included as column in a list -->
+        <!-- when used as column, the property will lead to duplicated rows for entities with multiple tags -->
```
nnatter, line 261: TODO: add a test case to ensure that this does not happen when the property is used only for filtering

src/Sulu/Component/Rest/ListBuilder/Doctrine/DoctrineListBuilder.php
```diff
@@ -198,7 +198,7 @@ public function execute()
         $this->expressionFields = $this->getUniqueExpressionFieldDescriptors($this->expressions);
 
         if (!$this->limit && empty($this->expressions)) {
-            $queryBuilder = $this->createFullQueryBuilder($this->createQueryBuilder());
+            $queryBuilder = $this->createFullQueryBuilder($this->createQueryBuilder($this->getJoins(true)));
```
alexander-schranz: I would not extend the `getJoins` method. I only would maybe change this `if` statement if this really needed: 

```php
if (!$this->limit && empty($this->expressions) && !$this->search) {
```

alexander-schranz: as I think we are now adding `expressionFields` in a query where no `expressions` are.

niklasnatter: i think its a bit confusing that the `getJoins()` method does not return the joins for the `this->expressionFields` then. the `getAllFields` method already has a similar `onlyReturnFilterFields` parameter. but on the other hand, the class is already confusing anyway, so i am okay with both ways 🙂

alexander-schranz: I'm just a little bit afraid about add new parameters to an protected function on a not final class. And yes I find the class also very confusing at current state.

niklasnatter: is okay for me - removed the parameter 🙂

alexander-schranz (approved):

PR #5945, opened 2021-04-20
Add resourceTitleLocale to DomainEvent

### LINKED ISSUE #5940: Resource title of domain event is not localized
| Q | A
| --- | ---
| Sulu Version | 2.3

# Actual Behavior

The resource title of a `DomainEvent` is not localized. This will probably be a problem for `removed` events as soon as there should be an audit log showing entity deletions, because then the actual entity is not available anymore to retrieve the title in the correct language.

# Possible Solutions

All available translations of the resource title could be saved on the `DomainEvent`'s event context.

luca-rath:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #5940
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

Add `resourceTitleLocale` field to `DomainEvent` and `EventRecord`

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -4,7 +4,7 @@
 
 ### Added resourceTitleLocale field to EventRecord
 
-Becase a new `resourceTitleLocale` field has been added to the `EventRecord` entity,
+Because a new `resourceTitleLocale` field has been added to the `EventRecord` entity,
 you need to update your database schema:
 
 ```sql
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -2,6 +2,15 @@
 
 ## 2.3
 
+### Added resourceTitleLocale field to EventRecord
+
+Becase a new `resourceTitleLocale` field has been added to the `EventRecord` entity,
+you need to update your database schema:
+
+```sql
+ALTER TABLE el_event_records ADD resourceTitleLocale VARCHAR(191) DEFAULT NULL;
+```
+
 ### Changed constructor of multiple services to integrate them with the SuluEventLogBundle
 
 To integrate the `SuluEventLogBundle` with the existing services, the constructor of the following services was 
--- a/src/Sulu/Bundle/CategoryBundle/Domain/Event/CategoryCreatedEvent.php
+++ b/src/Sulu/Bundle/CategoryBundle/Domain/Event/CategoryCreatedEvent.php
@@ -14,6 +14,7 @@ namespace Sulu\Bundle\CategoryBundle\Domain\Event;
 use Sulu\Bundle\CategoryBundle\Admin\CategoryAdmin;
 use Sulu\Bundle\CategoryBundle\Entity\CategoryInterface;
 use Sulu\Bundle\EventLogBundle\Domain\Event\DomainEvent;
+use Webmozart\Assert\Assert;
 
 class CategoryCreatedEvent extends DomainEvent
 {
@@ -79,14 +80,12 @@ class CategoryCreatedEvent extends DomainEvent
 
     public function getResourceTitle(): ?string
     {
-        $translation = $this->category->findTranslationByLocale($this->getResourceTitleLocale());
+        $resourceTitleLocale = $this->getResourceTitleLocale();
+        Assert::notNull($resourceTitleLocale);
+
+        $translation = $this->category->findTranslationByLocale($resourceTitleLocale);
 
         return $translation ? $translation->getTranslation() : null;
-    }
-
-    public function getResourceTitleLocale(): string
-    {
-        return $this->getResourceLocale();
     }
 
     public function getResourceSecurityContext(): ?string
--- a/src/Sulu/Bundle/CategoryBundle/Domain/Event/CategoryCreatedEvent.php
+++ b/src/Sulu/Bundle/CategoryBundle/Domain/Event/CategoryCreatedEvent.php
@@ -72,16 +72,21 @@ class CategoryCreatedEvent extends DomainEvent
         return (string) $this->category->getId();
     }
 
-    public function getResourceLocale(): ?string
+    public function getResourceLocale(): string
     {
         return $this->locale;
     }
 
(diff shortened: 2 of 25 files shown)


src/Sulu/Bundle/CategoryBundle/Domain/Event/CategoryCreatedEvent.php
```diff
@@ -72,18 +72,23 @@ public function getResourceId(): string
...
-        $translation = $this->category->findTranslationByLocale($this->locale);
+        $translation = $this->category->findTranslationByLocale($this->getResourceTitleLocale());
 
         return $translation ? $translation->getTranslation() : null;
     }
 
+    public function getResourceTitleLocale(): string
+    {
+        return $this->getResourceLocale();
+    }
```
niklasnatter: would move this as default implementation into the `DomainEvent` and only overwrite the method in the event classes that need another implementation

src/Sulu/Bundle/CategoryBundle/Domain/Event/CategoryCreatedEvent.php
```diff
@@ -72,18 +72,23 @@ public function getResourceId(): string
...
-    public function getResourceLocale(): ?string
+    public function getResourceLocale(): string
     {
         return $this->locale;
     }
 
     public function getResourceTitle(): ?string
     {
-        $translation = $this->category->findTranslationByLocale($this->locale);
+        $translation = $this->category->findTranslationByLocale($this->getResourceTitleLocale());
```
niklasnatter: not sure if this makes it harder to understand what is happening, but dont have a too strong opinion 🙂

UPGRADE.md
```diff
@@ -2,6 +2,15 @@
 
 ## 2.3
 
+### Added resourceTitleLocale field to EventRecord
+
+Becase a new `resourceTitleLocale` field has been added to the `EventRecord` entity,
```
```suggestion
Because a new `resourceTitleLocale` field has been added to the `EventRecord` entity,
```

src/Sulu/Bundle/CategoryBundle/Domain/Event/CategoryCreatedEvent.php
luca-rath: I just want to make sure, that the locale of the `resourceTitle` is identical to the result of `DomainEvent::getResourceTitleLocale()`

src/Sulu/Bundle/CategoryBundle/Domain/Event/CategoryCreatedEvent.php
```diff
@@ -72,14 +73,17 @@ public function getResourceId(): string
...
+    public function getResourceLocale(): string
     {
         return $this->locale;
     }
 
     public function getResourceTitle(): ?string
     {
-        $translation = $this->category->findTranslationByLocale($this->locale);
+        $resourceTitleLocale = $this->getResourceTitleLocale();
+        Assert::notNull($resourceTitleLocale);
```
luca-rath: Because the implementation of the `getResourceTitleLocale(): ?string` method is in the `DomainEvent` class now, we can't typehint it with `getResourceTitleLocale(): string` anymore, therefore we need to ensure, that the `resourceTitleLocale` is not null

niklasnatter: i think this would be a second reason for using `$this->locale` 🙂

niklasnatter: our tests should ensure that the locale used by the `getResourceTitle()` is the same as the one returned by `getResourceTitleLocale()`

luca-rath: I really think we should use the `getResourceTitleLocale()` method here:
1. Errors will occurr earlier and won't lead to inconsitent data in the database, if someone forgot to implement the `getResourceTitleLocale()` method
2. I think the name of the `getResourceTitleLocale()` method is quite self explanatory, so IMO, here it's easier to understand what's actually going on

alexander-schranz: I would also not call the public `getResourceLocale` internal and directly accessing `$this->locale` or the correct locale here.

niklasnatter: > 1. Errors will occurr earlier and won't lead to inconsitent data in the database, if someone forgot to implement the `getResourceTitleLocale()` method

If somebody forgets to implement the `getResourceTitleLocale()` method, it means that he is implementing his own domain event. this means he will not use this class and its implementation at all, right? 🤔 

> 2. I think the name of the `getResourceTitleLocale()` method is quite self explanatory, so IMO, here it's easier to understand what's actually going on

Might be a matter of taste. I feel like its harder to understand because I need to have a look at the base class to check what the `getResourceTitleLocale()` is doing. Also I will probably wonder about why the assertion is needed at first 🙈

luca-rath: > If somebody forgets to implement the getResourceTitleLocale() method, it means that he is implementing his own domain event. this means he will not use this class and its implementation at all, right? 🤔

But he will probably use this or a similar class as reference

alexander-schranz: looks like you did forget to change it here to $this->locale?

luca-rath: oops 🙈 fixed now

niklasnatter (approved):
🎉

PR #5948, opened 2021-04-22
Add copyLocale method to DocumentManager

luca-rath:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | yes
| Deprecations? | yes
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

Add `copyLocale` method to `DocumentManagerInterface`

#### Why?

To allow listening to a new `CopyLocaleEvent` (this is needed for the domain events)

### DIFF
--- a/src/Sulu/Bundle/PageBundle/Controller/PageController.php
+++ b/src/Sulu/Bundle/PageBundle/Controller/PageController.php
@@ -234,6 +234,7 @@ class PageController extends AbstractRestController implements ClassResourceInte
                 case 'copy-locale':
                     $srcLocale = $this->getRequestParameter($request, 'src', false, $locale);
                     $destLocales = $this->getRequestParameter($request, 'dest', true);
+                    $destLocales = \explode(',', $destLocales);
 
                     $document = $this->documentManager->find($id, $srcLocale);
 
--- a/src/Sulu/Bundle/PageBundle/Controller/PageController.php
+++ b/src/Sulu/Bundle/PageBundle/Controller/PageController.php
@@ -238,7 +238,10 @@ class PageController extends AbstractRestController implements ClassResourceInte
 
                     $document = $this->documentManager->find($id, $srcLocale);
 
-                    $this->documentManager->copyLocale($document, $srcLocale, $destLocales);
+                    foreach ($destLocales as $destLocale) {
+                        $this->documentManager->copyLocale($document, $srcLocale, $destLocale);
+                    }
+
                     $this->documentManager->flush();
 
                     $data = $document;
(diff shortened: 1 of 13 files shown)


src/Sulu/Bundle/PageBundle/Controller/PageController.php
```diff
@@ -233,16 +233,18 @@ public function postTriggerAction(Request $request, string $id): Response
...
+                    $destLocales = \explode(',', $destLocales);
 
-                    // call repository method
-                    $data = $this->nodeRepository->copyLocale($id, $userId, $webspace, $srcLocale, \explode(',', $destLocale));
+                    $document = $this->documentManager->find($id, $srcLocale);
 
-                    if ($srcLocale !== $locale) {
-                        $data = $this->nodeRepository->getNode($id, $webspace, $locale);
-                    }
+                    $this->documentManager->copyLocale($document, $srcLocale, $destLocales);
```
niklasnatter: not sure if we should pass a single destination locale here. as far as i see, we do not pass an array of locales anywhere else. what do you think?

niklasnatter: maybe related to this: should there be a domain event for each destination locale?

luca-rath: > not sure if we should pass a single destination locale here. as far as i see, we do not pass an array of locales anywhere else. what do you think?

fine for me, I don't have a strong opinion about that

> maybe related to this: should there be a domain event for each destination locale?

I would say yes, but that doesn't really affect the above decision IMO, because we could just loop over the `$destLocales` and throw a specific domain event for each locale without any effort

/cc @sulu/core-team

alexander-schranz: I have no strong opinion here using a single local is okay from my side.

src/Sulu/Component/Content/Document/Subscriber/CopyLocaleSubscriber.php
```diff
@@ -0,0 +1,140 @@
...
+                $destDocument->getTitle(),
+                $parentUuid,
+                $webspaceKey,
+                $destLocale
+            );
+
+            $destDocument->setResourceSegment($resourceLocator);
+        }
+
+        $this->documentManager->persist($destDocument, $destLocale);
```
wachterjohannes, line 138: @luca-rath havent we said that there should be an option to disable the domain event? will that be implemented in the follow up PR?

wachterjohannes (approved):
like the implementation with a single locale in the interface :) beside of the small question everything works and looks fine

src/Sulu/Component/Content/Document/Subscriber/CopyLocaleSubscriber.php
luca-rath: exactly

PR #5950, opened 2021-04-22
Update ckeditor to 27.1.0

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? | yes 
| BC breaks? | yes (webpack config updated needed) 
| Deprecations? | no 
| Fixed tickets | fixes #
| Related issues/PRs | #
| License | MIT

#### What's in this PR?

Update ckeditor to 27.1.0.

#### Why?

New version contain more bugfixes and features.


package.json
```diff
@@ -121,7 +121,7 @@
             ]
         },
         "transformIgnorePatterns": [
-            "node_modules/(?!(@ckeditor|lodash-es)/)"
+            "node_modules/(?!(@ckeditor|ckeditor5|lodash-es)/)"
```
alexander-schranz, line 124: See https://github.com/ckeditor/ckeditor5/issues/9554 there seems to be a strange new core packages just called `ckeditor5` which also need to transpiled

webpack.config.js
```diff
@@ -75,7 +75,7 @@ module.exports = (env, argv) => { // eslint-disable-line no-undef
             rules: [
                 {
                     test: /\.js$/,
-                    exclude: /node_modules[/\\](?!(sulu-(.*)-bundle|@ckeditor|lodash-es)[/\\])/,
+                    exclude: /node_modules[/\\](?!(sulu-(.*)-bundle|@ckeditor|ckeditor5|lodash-es)[/\\])/,
```
alexander-schranz, line 78: See https://github.com/ckeditor/ckeditor5/issues/9554 there seems to be a strange new core packages just called `ckeditor5` which also need to be transpiled

alexander-schranz: It was not required for the build but was afraid that something could go wrong when I only add it to jest and not also here

PR #5953, opened 2021-04-22
Upgrade react-leaflet to 3.0.0 version

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? | no 
| BC breaks? | yes
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Upgrade react-leaflet to newest version.

#### Why?

New version contains bugfixes and features.

### DIFF
--- a/src/Sulu/Bundle/LocationBundle/Resources/js/containers/Location/LocationOverlay.js
+++ b/src/Sulu/Bundle/LocationBundle/Resources/js/containers/Location/LocationOverlay.js
@@ -48,8 +48,8 @@ class LocationOverlay extends React.Component<Props> {
 
         this.updateDataOnOpenDisposer = reaction(() => this.props.open, (newOpenValue) => {
             if (newOpenValue === true) {
-                this.lat = this.props.value ? this.props.value.lat : 0;
-                this.long = this.props.value ? this.props.value.long : 0;
+                this.lat = this.props.value ? this.props.value.lat : null;
+                this.long = this.props.value ? this.props.value.long : null;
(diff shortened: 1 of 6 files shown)


src/Sulu/Bundle/LocationBundle/Resources/js/containers/Location/LocationOverlay.js
```diff
@@ -218,7 +218,7 @@ class LocationOverlay extends React.Component<Props> {
                         </Form.Field>
 
                         <Form.Field>
-                            <Map
+                            <MapContainer
```
niklasnatter, line 225: unfortunately this change is not enough for me 😕  
it looks like the zoom level and the coordinates in the map are not updated when the map is zoomed or the marker is moved. also the map outside of the overlay is not updated correctly when the overlay is confirmed. 

i tried to adjust the things mentioned in the [`CHANGELOG.md` of the package](https://github.com/PaulLeCam/react-leaflet/blob/master/CHANGELOG.md). but i was not able to fix the problems on the first try and dont have the time to investigate this deeper at the moment 😕

niklasnatter: i think we need to at least adjust the following things:
- use `eventHandlers` prop instead of `on*` props for all `react-leaflet` components
- use [MapConsumer](https://react-leaflet.js.org/docs/api-map#mapconsumer) component to gather underlying `Map` instance to change center/zoom of map instead of using `center` and `zoom` prop of `MapContainer`

PR #5954, opened 2021-04-23
Add CategoryTranslationAddedEvent

luca-rath:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | -
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

This pull request adds a `CategoryLocaleAddedEvent`, which is being thrown instead of the `CategoryModifiedEvent`, if the saved locale didn't exist before

### DIFF
--- a/src/Sulu/Bundle/CategoryBundle/Category/CategoryManager.php
+++ b/src/Sulu/Bundle/CategoryBundle/Category/CategoryManager.php
@@ -15,6 +15,7 @@ use Doctrine\DBAL\Exception\UniqueConstraintViolationException;
 use Doctrine\ORM\EntityManagerInterface;
 use Sulu\Bundle\CategoryBundle\Api\Category as CategoryWrapper;
 use Sulu\Bundle\CategoryBundle\Domain\Event\CategoryCreatedEvent;
+use Sulu\Bundle\CategoryBundle\Domain\Event\CategoryLocaleAddedEvent;
 use Sulu\Bundle\CategoryBundle\Domain\Event\CategoryModifiedEvent;
 use Sulu\Bundle\CategoryBundle\Domain\Event\CategoryMovedEvent;
 use Sulu\Bundle\CategoryBundle\Domain\Event\CategoryRemovedEvent;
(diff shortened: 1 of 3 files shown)


src/Sulu/Bundle/CategoryBundle/Category/CategoryManager.php
```diff
@@ -167,9 +168,18 @@ private function findOrCreateCategoryTranslation(
     public function save($data, $userId, $locale, $patch = false)
     {
         $isNewCategory = !$this->getProperty($data, 'id');
+        $isNewLocale = true;
 
         if (!$isNewCategory) {
             $categoryEntity = $this->findById($this->getProperty($data, 'id'));
+
+            foreach ($categoryEntity->getTranslations() as $translation) {
```
niklasnatter: think you can use `$categoryEntity->findTranslationByLocale` instead of looping by yourself

src/Sulu/Bundle/CategoryBundle/Domain/Event/CategoryLocaleAddedEvent.php
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    public function getCategory(): CategoryInterface
+    {
+        return $this->category;
+    }
+
+    public function getEventType(): string
+    {
+        return 'locale_added';
```
niklasnatter: @sulu/core-team should we call this `locale_added` or `translation_added`? the type will also be used for medias, articles and pages

niklasnatter (commented):
looks good otherwise 🙂

src/Sulu/Bundle/CategoryBundle/Domain/Event/CategoryLocaleAddedEvent.php
niklasnatter: i guess we should also use the same term for the `locale_copied` and `locale_removed` event

niklasnatter: i feel like `translation` makes more sense in the context of categories and from the perspective of the user. on the other hand, the `DocumentManager` already has a `copyLocale` and `removeLocale` method. so i guess both terms are okay for me 😅

luca-rath: Since internally the term `locale` is used a lot more than `translation`, I would stick with that. On the frontend, the messages are translated anyway, so we can still use `Translation copied` and `Übersetzung kopiert` e.g.

src/Sulu/Bundle/CategoryBundle/Category/CategoryManager.php
luca-rath: fixed

niklasnatter (approved):

PR #5955, opened 2021-04-23
Add resourceSecurityObjectId to domain events

luca-rath:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

Add a `getResourceSecurityObjectId()` method to the `DomainEvent`

#### Why?

There is no object security on media itself, but on collections. Therefore it's required to store the `collectionId` somehow on the `EventRecord`, because that one is used to build the `SecurityCondition`

### DIFF
--- a/src/Sulu/Bundle/EventLogBundle/Domain/Event/DomainEvent.php
+++ b/src/Sulu/Bundle/EventLogBundle/Domain/Event/DomainEvent.php
@@ -130,4 +130,9 @@ abstract class DomainEvent
     {
         return null;
     }
+
+    public function getResourceSecurityObjectId(): ?string
+    {
+        return $this->getResourceId();
+    }
 }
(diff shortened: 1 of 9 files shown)


src/Sulu/Bundle/EventLogBundle/Domain/Event/DomainEvent.php
```diff
@@ -130,4 +130,9 @@ public function getResourceSecurityType(): ?string
     {
         return null;
     }
+
+    public function getResourceSecurityObjectId(): ?string
```
niklasnatter, line 134: the `SecurityCondition` has a `getObjectType` and a `getObjectId` method. would be nice to have a consistent prefix in the `DomainEvent` and `EventRecord` too. i guess i have not thought about this enough when i implemented the `getResourceSecurityType` method 🙈 

there are two ways to have a consistent prefix:

- rename `resourceSecurityType` to `resourceSecurityObjectType`
- rename `resourceSecurityObjectId` to `resourceSecurityId` 

i dont think it will make a big difference in the future, so both ways are okay for me. but i would like to be consistent 🙂

niklasnatter (approved):

niklasnatter:
Failing pipeline is not related to this PR.

PR #5956, opened 2021-04-23
Upgrade react-datetime to 3.0.0

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? | no 
| BC breaks? |  yes
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Upgrade react-datetime to 3.0.0. https://github.com/arqex/react-datetime/blob/master/migrateToV3.md.

There is a problem currently with the 3.0 version as it seems not to close the overlay when choosing the time: https://github.com/arqex/react-datetime/pull/741#issuecomment-825612866

#### Why?

New version provide bugfixes and new features.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/DatePicker/DatePicker.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/DatePicker/DatePicker.js
@@ -80,6 +80,16 @@ class DatePicker extends React.Component<Props> {
         this.props.onChange(date);
 
         this.setShowError(!!this.value && !date);
+
+        // can be removed when is fixed the following is fixed:
+        // https://github.com/arqex/react-datetime/pull/741
+        if (
+            (!this.value && date) ||
+            (this.value && !date) ||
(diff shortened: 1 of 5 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/components/DatePicker/DatePicker.js
```diff
@@ -80,6 +80,16 @@ class DatePicker extends React.Component<Props> {
         this.props.onChange(date);
 
         this.setShowError(!!this.value && !date);
+
+        // can be removed when is fixed the following is fixed:
+        // https://github.com/arqex/react-datetime/pull/741
```
niklasnatter, line 85: i guess we can remove the `closeOnSelect` if we implement it like this 🙂

niklasnatter:
I dont think https://github.com/arqex/react-datetime/pull/741 will ever be fixed, so I would be okay with implementing the workaround for closing the Popover 🙂

PR #5957, opened 2021-04-23
Fix Permission check in PageController for the Multi Webspace Column View Overlay

loic-cariou:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes
| Related issues/PRs |
| License | MIT

#### What's in this PR?

Check if the user has the right to view a webspace so as not to return the webspace trees that he cannot view

#### Why?

When a user does not have the right to "view" on a webspace he can still view the complete tree structure of a webpsace

### DIFF
--- a/src/Sulu/Bundle/PageBundle/Controller/PageController.php
+++ b/src/Sulu/Bundle/PageBundle/Controller/PageController.php
@@ -16,6 +16,7 @@ use FOS\RestBundle\View\ViewHandlerInterface;
 use HandcraftedInTheAlps\RestRoutingBundle\Routing\ClassResourceInterface;
 use PHPCR\ItemNotFoundException;
 use PHPCR\PropertyInterface;
+use Sulu\Bundle\PageBundle\Admin\PageAdmin;
 use Sulu\Bundle\PageBundle\Repository\NodeRepositoryInterface;
 use Sulu\Component\Content\Document\Behavior\SecurityBehavior;
 use Sulu\Component\Content\Form\Exception\InvalidFormException;
@@ -406,7 +407,7 @@ class PageController extends AbstractRestController implements ClassResourceInte
                     $this->documentManager->remove($document);
                     $this->documentManager->flush();
                 } catch (DocumentNotFoundException $ex) {
-                    throw new EntityNotFoundException('Content', $id);
+                    throw new EntityNotFoundException('Content', $id, $ex);
                 }
             }
             );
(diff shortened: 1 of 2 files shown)


alexander-schranz:
@loic-cariou Thank you for your pull request.

This feature is implemented in another way in sulu. To activate this feature you need to configure at first user specific caching:

https://docs.sulu.io/en/2.2/cookbook/user-context-caching.html

And then activate that you want to check for permissions in your security of your webspace:

https://github.com/sulu/sulu/blob/5fee2ca21b9c6d039180efedec4fd1eba289f4e0/config/webspaces/sulu-test.xml#L9-L11

Read more about it here: https://sulu.io/blog/sulu-release-2-2-0

alexander-schranz:
@loic-cariou If this make any problems for you let us know. If you want to hide a whole webspace you should create firewall in your `security_website.yaml` and allow only specific roles in the firewall. The security check is `system` based not webspace specific, the webspace is currently only for the admin area and has no effects on the website as I know.

loic-cariou:
@alexander-schranz 
The goal is precisely in the backoffice at the level of smartcontent and links for example not to display the tree structure of a webspace whose administrator does not have the "view" authorization on the webspace in question

loic-cariou:
@alexander-schranz  
Currently when an administrator has no rights on a webspace he still sees the complete tree in all webspaces when he uses a single_page_selection. this pull request is used to display only the tree structure for which the administrator has at least the view right on a webspace

alexander-schranz:
@loic-cariou sorry I mistake it with the website controller. Thank you for the clarification we will take a look at it soon.

src/Sulu/Bundle/PageBundle/Controller/PageController.php
```diff
@@ -621,10 +621,12 @@ private function getWebspaceNodes(
         $webspaces = [];
         /** @var Webspace $webspace */
         foreach ($this->webspaceManager->getWebspaceCollection() as $webspace) {
-            if (null === $webspace->getLocalization($locale)) {
+            if (null === $webspace->getLocalization($locale) || false === $this->securityChecker->hasPermission(
+                new SecurityCondition('sulu.webspaces.' . $webspace->getKey(), $locale, SecurityBehavior::class),
```
luca-rath: Use `PageAdmin::getPageSecurityContext($webspace->getKey())` instead of `'sulu.webspaces' . $webspace->getKey()`

src/Sulu/Bundle/PageBundle/Controller/PageController.php
```diff
@@ -643,6 +645,13 @@ private function getWebspaceNode(
         UserInterface $user
     ) {
         $webspace = $this->webspaceManager->findWebspaceByKey($webspaceKey);
+        if (null === $webspace->getLocalization($locale) || false === $this->securityChecker->hasPermission(
+            new SecurityCondition('sulu.webspaces.' . $webspace->getKey(), $locale, SecurityBehavior::class),
```
luca-rath: Same here, use `PageAdmin::getPageSecurityContext($webspace->getKey())` instead of `'sulu.webspaces' . $webspace->getKey()`

luca-rath (changes requested):

src/Sulu/Bundle/PageBundle/Controller/PageController.php
alexander-schranz: This does not exist on `2.1` but will port it back to `2.1` branch so we can also use it there.

src/Sulu/Bundle/PageBundle/Controller/PageController.php
alexander-schranz: PR updated!

src/Sulu/Bundle/PageBundle/Controller/PageController.php
alexander-schranz: PR updated!

luca-rath (approved):
Code looks good to me, but didn't test

Prokyonn (approved):
tested it and works as expected :)

alexander-schranz:
@loic-cariou Thank you for bringing this up 👍

PR #5958, opened 2021-04-23
Use buildQueryString util instead of external query-string package in PreviewBundle

niklasnatter:


src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/tests/stores/PreviewStore.test.js
```diff
@@ -26,7 +26,7 @@ test('Should request server on start preview', () => {
     previewStore.start();
 
     return requestPromise.then(() => {
-        expect(Requester.post).toBeCalledWith('/start?id=123-123-123&locale=en&provider=pages');
```
niklasnatter, line 29: looks like the `query-string` library sorted the parameters by the parameter key

alexander-schranz (approved):

PR #5961, opened 2021-04-23
Update flow-bin package to version 0.142

niklasnatter:


.flowconfig
```diff
@@ -1,4 +1,5 @@
 [options]
+types_first=false
```
niklasnatter, line 2: unfortunately, classic-mode was removed in the 0.143 release of `flow-bin`:
https://github.com/facebook/flow/releases/tag/v0.143.0

enabling the `types_first` mode leads to over 1800 errors are the moment. i am afraid we cannot update to version higher than 0.142 without spending a lot of time on this. (looking at the popularity of flow, i feel like this time would be better spent on migrating to typescript 🙈 )

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/FormInspector.js
```diff
@@ -1,6 +1,6 @@
 // @flow
 import {computed} from 'mobx';
-import type {IObservableValue} from 'mobx';
+import type {IObservableValue} from 'mobx/lib/mobx';
```
niklasnatter, line 3: flow does not recognize the declaration file (`mobx.js.flow`) wen using `'mobx'` 😕

alexander-schranz (approved):

PR #5962, opened 2021-04-23
Add general activity view

luca-rath:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

Add global activity log to admin interface

### DIFF
--- a/src/Sulu/Bundle/ActivityBundle/Domain/Model/ActivityInterface.php
+++ b/src/Sulu/Bundle/ActivityBundle/Domain/Model/ActivityInterface.php
@@ -15,9 +15,11 @@ use Sulu\Component\Security\Authentication\UserInterface;
 
 interface ActivityInterface
 {
+    const RESOURCE_KEY = 'activities';
+
     public function getType(): string;
 
-    public function setType(string $eventType): ActivityInterface;
+    public function setType(string $type): ActivityInterface;
 
     /**
      * @return mixed[]
@@ -25,9 +27,9 @@ interface ActivityInterface
     public function getContext(): array;
 
     /**
-     * @param mixed[] $eventContext
+     * @param mixed[] $context
      */
-    public function setContext(array $eventContext): ActivityInterface;
+    public function setContext(array $context): ActivityInterface;
 
     /**
      * @return mixed[]|null
@@ -35,17 +37,17 @@ interface ActivityInterface
     public function getPayload(): ?array;
 
     /**
-     * @param mixed[]|null $eventPayload
+     * @param mixed[]|null $payload
      */
-    public function setPayload(?array $eventPayload): ActivityInterface;
+    public function setPayload(?array $payload): ActivityInterface;
 
     public function getTimestamp(): \DateTimeImmutable;
 
-    public function setTimestamp(\DateTimeImmutable $eventDateTime): ActivityInterface;
+    public function setTimestamp(\DateTimeImmutable $timestamp): ActivityInterface;
 
     public function getBatch(): ?string;
 
-    public function setBatch(?string $eventBatch): ActivityInterface;
+    public function setBatch(?string $batch): ActivityInterface;
 
     public function getUser(): ?UserInterface;
 
--- a/src/Sulu/Bundle/ActivityBundle/UserInterface/Controller/ActivityController.php
+++ b/src/Sulu/Bundle/ActivityBundle/UserInterface/Controller/ActivityController.php
@@ -334,7 +334,7 @@ class ActivityController extends AbstractRestController implements ClassResource
                 $translationParameters[$prefix . $key . $suffix] = (string) $value;
             } elseif (\is_array($value)) {
                 foreach ($this->getTranslationParameters($value, '', '') as $translationParameterKey => $translationParameterValue) {
-                    $translationParameters[$prefix . $key . '_' . $translationParameterKey . $suffix] = $translationParameterValue;
+                    $translationParameters[$prefix . $key . '.' . $translationParameterKey . $suffix] = $translationParameterValue;
                 }
             }
         }
--- a/src/Sulu/Bundle/ActivityBundle/UserInterface/Controller/ActivityController.php
+++ b/src/Sulu/Bundle/ActivityBundle/UserInterface/Controller/ActivityController.php
@@ -349,7 +349,7 @@ class ActivityController extends AbstractRestController implements ClassResource
     {
         $translationParameters = $this->getTranslationParameters($activity);
 
-        $user = $activity['user'] ?? $this->translator->trans('sulu_activity.someone', [], 'admin', $translationLocale);
+        $user = $activity['userFullName'] ?? $this->translator->trans('sulu_activity.someone', [], 'admin', $translationLocale);
 
         $resourceTitleLocale = $activity['resourceTitleLocale'] ?? null;
         $resourceTitle = '"' . $activity['resourceTitle'] . '"';
@@ -358,7 +358,7 @@ class ActivityController extends AbstractRestController implements ClassResource
             $resourceTitle .= ' [' . \strtoupper($resourceTitleLocale) . ']';
         }
 
-        $translationParameters['{user}'] = $user;
+        $translationParameters['{userFullName}'] = $user;
         $translationParameters['{resourceTitle}'] = $resourceTitle;
 
         return $this->translator->trans(
@@ -403,8 +403,8 @@ class ActivityController extends AbstractRestController implements ClassResource
             'resourceSecurityContext' => $this->createFieldDescriptor('resourceSecurityContext'),
             'resourceSecurityObjectType' => $this->createFieldDescriptor('resourceSecurityObjectType'),
             'resourceSecurityObjectId' => $this->createFieldDescriptor('resourceSecurityObjectId'),
-            'user' => $this->createConcatenationFieldDescriptor(
-                'user',
+            'userFullName' => $this->createConcatenationFieldDescriptor(
+                'userFullName',
                 [
                     $this->createFieldDescriptor('firstName', null, $this->contactClass, $userJoins),
                     $this->createFieldDescriptor('lastName', null, $this->contactClass, $userJoins),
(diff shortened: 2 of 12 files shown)


src/Sulu/Bundle/EventLogBundle/UserInterface/Controller/EventRecordController.php
```diff
@@ -0,0 +1,450 @@
...
+        /** @var DoctrineFieldDescriptor $resourceSecurityObjectTypeFieldDescriptor */
+        $resourceSecurityObjectTypeFieldDescriptor = $fieldDescriptors['resourceSecurityObjectType'];
+        $listBuilder->addPermissionCheckField($resourceSecurityObjectTypeFieldDescriptor);
+
+        $listBuilder->setPermissionCheckWithDynamicEntityClass(
+            $this->getUser(),
+            PermissionTypes::VIEW,
+            'resourceSecurityObjectType',
+            'resourceSecurityObjectId'
+        );
```
niklasnatter: would be consistent with everything else to extract this into a method similar to the `add*Condition` methods 😇

src/Sulu/Bundle/EventLogBundle/UserInterface/Controller/EventRecordController.php
```diff
@@ -0,0 +1,450 @@
...
+    }
+
+    /**
+     * @param array<string, mixed> $eventRecord
+     */
+    private function getEventText(array $eventRecord, string $translationLocale): string
+    {
+        $translationParameters = [];
+        foreach ($eventRecord as $key => $value) {
```
nnatter, line 314: think it would be nice to do the same for the context of the event. right now its not possible to access that data in the translation right?

src/Sulu/Bundle/EventLogBundle/UserInterface/Controller/EventRecordController.php
```diff
@@ -0,0 +1,450 @@
...
+use Sulu\Component\Security\Authentication\UserInterface;
+use Sulu\Component\Security\Authorization\PermissionTypes;
+use Sulu\Component\Security\Authorization\SecurityCheckerInterface;
+use Sulu\Component\Security\SecuredControllerInterface;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface;
+use Symfony\Contracts\Translation\TranslatorInterface;
+
+class EventRecordController extends AbstractRestController implements ClassResourceInterface, SecuredControllerInterface
```
niklasnatter: nice to see things coming together. also good to see that this works in a quite clean way without any hacks 🙂

src/Sulu/Bundle/EventLogBundle/Infrastructure/Sulu/Admin/EventLogAdmin.php
```diff
@@ -0,0 +1,101 @@
...
+        $this->localizationManager = $localizationManager;
+    }
+
+    public function configureNavigationItems(NavigationItemCollection $navigationItemCollection): void
+    {
+        if (!$this->securityChecker->hasPermission(static::SECURITY_CONTEXT, PermissionTypes::VIEW)) {
+            return;
+        }
+
+        $eventLogNavigationItem = new NavigationItem('sulu_event_log.event_log');
```
niklasnatter: not sure how we should call this in the frontend yet? @sulu/core-team 

on one hand, i think `activity` would be a nice term. but on the other hand we dont use that term in the backend anywhere and i really want to avoid a confusion similar to the differences between `account`/`organization` 🙈

niklasnatter: as we renamed the bundle to `SuluActivityBundle` in #6003, we can use `Activity` now 🙂

src/Sulu/Bundle/ActivityBundle/Domain/Model/ActivityInterface.php
```diff
@@ -15,37 +15,39 @@
 
 interface ActivityInterface
 {
+    const RESOURCE_KEY = 'activities';
+
     public function getType(): string;
 
-    public function setType(string $eventType): ActivityInterface;
+    public function setType(string $type): ActivityInterface;
```
niklasnatter, line 22: thanks! 🙂

src/Sulu/Bundle/ActivityBundle/UserInterface/Controller/ActivityController.php
```diff
@@ -0,0 +1,472 @@
...
+        /** @var DoctrineListBuilder $listBuilder */
+        $listBuilder = $this->listBuilderFactory->create($this->activityClass);
+        $this->restHelper->initializeListBuilder($listBuilder, $fieldDescriptors);
+        $listBuilder->setSelectFields($fieldDescriptors);
+        $listBuilder->sort($fieldDescriptors['timestamp'], ListBuilderInterface::SORTORDER_DESC);
+
+        $translationLocale = $user->getLocale();
+
+        /** @var string|null $resourceLocale */
+        $resourceLocale = $this->getRequestParameter($request, 'locale');
```
niklasnatter, line 139: in other APIs, the `locale` parameter will change the language of the data is returned, in this it will act similar to a filter. maybe we should call it `resourceLocale` because of the different behaviour?

niklasnatter: that would allow us to add a `locale` parameter to change the locale that is used by the translator if we need to do that in the future

src/Sulu/Bundle/ActivityBundle/UserInterface/Controller/ActivityController.php
```diff
@@ -0,0 +1,472 @@
...
+            'resourceKey' => $this->createFieldDescriptor('resourceKey'),
+            'resourceId' => $this->createFieldDescriptor('resourceId'),
+            'resourceLocale' => $this->createFieldDescriptor('resourceLocale'),
+            'resourceWebspaceKey' => $this->createFieldDescriptor('resourceWebspaceKey'),
+            'resourceTitle' => $this->createFieldDescriptor('resourceTitle'),
+            'resourceTitleLocale' => $this->createFieldDescriptor('resourceTitleLocale'),
+            'resourceSecurityContext' => $this->createFieldDescriptor('resourceSecurityContext'),
+            'resourceSecurityObjectType' => $this->createFieldDescriptor('resourceSecurityObjectType'),
+            'resourceSecurityObjectId' => $this->createFieldDescriptor('resourceSecurityObjectId'),
+            'user' => $this->createConcatenationFieldDescriptor(
```
niklasnatter: maybe call this `userFullName`?

src/Sulu/Bundle/ActivityBundle/UserInterface/Controller/ActivityController.php
luca-rath: If I remember correctly, we talked about that from face to face and agreed, that the locale of the activity text should match the locale of the navigation items, toolbar actions, etc... So that locale can be changed by chaning the user locale. And I'm also not sure, if the admin frontend allows to just rename the locale parameter?

niklasnatter: Yeah i think its good that it matches the locale of the user 🙂  What I wanted to say is that I think I would expect the `locale` parameter to affect the language of the text like in other APIs, not filter the items.

I think its possible to configure the list view to send the locale of the route (eg. in the page activities) via another parameter with something like `->addRouterAttributesToListRequest(['locale' => 'resourceLocale'])`

src/Sulu/Bundle/ActivityBundle/UserInterface/Controller/ActivityController.php
```diff
@@ -0,0 +1,472 @@
...
+use Sulu\Component\Security\Authentication\UserInterface;
+use Sulu\Component\Security\Authorization\PermissionTypes;
+use Sulu\Component\Security\Authorization\SecurityCheckerInterface;
+use Sulu\Component\Security\SecuredControllerInterface;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface;
+use Symfony\Contracts\Translation\TranslatorInterface;
+
+class ActivityController extends AbstractRestController implements ClassResourceInterface, SecuredControllerInterface
```
niklasnatter, line 39: would be nice to add functional test cases for this 🙂

src/Sulu/Bundle/ActivityBundle/UserInterface/Controller/ActivityController.php
niklasnatter: Unfortunately, we figured that `addRouterAttributesToListRequest` does not work if the locale is switched after the initial rendering of the list. Therefore we are going to use `locale` 🙂

src/Sulu/Bundle/ActivityBundle/UserInterface/Controller/ActivityController.php
```diff
@@ -0,0 +1,472 @@
...
+                $translationParameters[$prefix . $key . $suffix] = 'null';
+            } elseif (\is_bool($value)) {
+                $translationParameters[$prefix . $key . $suffix] = $value ? 'true' : 'false';
+            } elseif (\is_numeric($value)) {
+                $translationParameters[$prefix . $key . $suffix] = $value;
+            } elseif (\is_scalar($value)) {
+                $translationParameters[$prefix . $key . $suffix] = (string) $value;
+            } elseif (\is_array($value)) {
+                foreach ($this->getTranslationParameters($value, '', '') as $translationParameterKey => $translationParameterValue) {
+                    $translationParameters[$prefix . $key . '_' . $translationParameterKey . $suffix] = $translationParameterValue;
```
niklasnatter: could we use `.` instead of `_` here?  think that would be a bit nicer 🙂

niklasnatter (approved):

PR #5964, opened 2021-04-26
Duplicate phpcr permissions in database

luca-rath:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

This pull request automatically duplicates object permissions of phpcr documents into the `AccessControl` entity.

#### Why?

To be able to filter on permissions for pages using the `DoctrineListBuilder`

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -2,6 +2,16 @@
 
 ## 2.3
 
+### Change entityId field in AccessControl entity to string
+
+To allow entities with uuid's instead of auto generated ids to have object permissions,
+the `entityId` field of the `AccessControl` entity had to be changed from `integer` to `string`.
+Therefore the following sql statement needs to be executed.
+
+```sql
+ALTER TABLE se_access_controls CHANGE entityId entityId VARCHAR(36) NOT NULL;
+```
+
 ### Added resourceSecurityObjectId field to EventRecord
 
 Because a new `resourceSecurityObjectId` field has been added to the `EventRecord` entity
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -2,6 +2,18 @@
 
 ## 2.3
 
+### Sync object permissions stored in phpcr to doctrine
+
+To enable permission checking on database level for resources with object permissions stored in phpcr,
+you need to sync these permissions into doctrine. A command is available for that:
+
+```bash
+bin/adminconsole sulu:security:sync-phpcr-permissions
+```
+
+This command needs to be executed just once when upgrading sulu to 2.3,
+in the future the permissions are being synced automatically.
+
 ### Change entityId field in AccessControl entity to string
 
 To allow entities with uuid's instead of auto generated ids to have object permissions,
--- a/src/Sulu/Bundle/CoreBundle/DependencyInjection/SuluCoreExtension.php
+++ b/src/Sulu/Bundle/CoreBundle/DependencyInjection/SuluCoreExtension.php
@@ -11,6 +11,7 @@
 
 namespace Sulu\Bundle\CoreBundle\DependencyInjection;
 
+use Oro\ORM\Query\AST\Functions\Cast;
 use Oro\ORM\Query\AST\Functions\String\GroupConcat;
 use Sulu\Bundle\ContactBundle\Entity\Account;
 use Sulu\Bundle\ContactBundle\Entity\AccountInterface;
@@ -19,7 +20,6 @@ use Sulu\Bundle\MediaBundle\Entity\CollectionInterface;
 use Sulu\Component\Content\Types\Block\BlockVisitorInterface;
 use Sulu\Component\HttpKernel\SuluKernel;
 use Sulu\Component\Rest\Csv\ObjectNotSupportedException;
-use Sulu\Component\Rest\DQL\Cast;
 use Sulu\Component\Rest\Exception\InvalidHashException;
 use Sulu\Component\Rest\Exception\MissingParameterException;
 use Sulu\Component\Rest\ListBuilder\Filter\InvalidFilterTypeOptionsException;
--- a/src/Sulu/Bundle/MediaBundle/Entity/FileVersionMetaRepository.php
+++ b/src/Sulu/Bundle/MediaBundle/Entity/FileVersionMetaRepository.php
@@ -33,7 +33,7 @@ class FileVersionMetaRepository extends EntityRepository implements FileVersionM
                 'accessControl',
                 'WITH',
                 'accessControl.entityClass = :entityClass '
-                . 'AND accessControl.entityId = CAST(collection.id AS VARCHAR)'
+                . 'AND accessControl.entityId = CAST(collection.id AS CHAR(36))'
             )
             ->where('file.version = fileVersion.version')
             ->andWhere('accessControl.id is null');
@@ -61,7 +61,7 @@ class FileVersionMetaRepository extends EntityRepository implements FileVersionM
                 'accessControl',
                 'WITH',
                 'accessControl.entityClass = :entityClass '
-                . 'AND accessControl.entityId = CAST(collection.id AS VARCHAR)'
+                . 'AND accessControl.entityId = CAST(collection.id AS CHAR(36))'
             )
             ->where('file.version = fileVersion.version')
             ->andWhere('accessControl.id is null')
--- a/src/Sulu/Bundle/MediaBundle/Entity/FileVersionMetaRepository.php
+++ b/src/Sulu/Bundle/MediaBundle/Entity/FileVersionMetaRepository.php
@@ -33,7 +33,7 @@ class FileVersionMetaRepository extends EntityRepository implements FileVersionM
                 'accessControl',
                 'WITH',
                 'accessControl.entityClass = :entityClass '
-                . 'AND accessControl.entityId = CAST(collection.id AS CHAR(36))'
+                . 'AND accessControl.entityId = CAST(collection.id AS STRING)'
             )
             ->where('file.version = fileVersion.version')
             ->andWhere('accessControl.id is null');
@@ -61,7 +61,7 @@ class FileVersionMetaRepository extends EntityRepository implements FileVersionM
                 'accessControl',
                 'WITH',
                 'accessControl.entityClass = :entityClass '
-                . 'AND accessControl.entityId = CAST(collection.id AS CHAR(36))'
+                . 'AND accessControl.entityId = CAST(collection.id AS STRING)'
             )
             ->where('file.version = fileVersion.version')
             ->andWhere('accessControl.id is null')
--- a/src/Sulu/Bundle/MediaBundle/Entity/FileVersionMetaRepository.php
+++ b/src/Sulu/Bundle/MediaBundle/Entity/FileVersionMetaRepository.php
@@ -20,6 +20,11 @@ class FileVersionMetaRepository extends EntityRepository implements FileVersionM
 {
     public function findLatestWithoutSecurity()
     {
+        $entityIdCondition = 'accessControl.entityId = collection.id';
+        if ('postgresql' === $this->getDatabasePlatformName()) {
+            $entityIdCondition = 'accessControl.entityId = CAST(collection.id AS VARCHAR)';
+        }
+
         $queryBuilder = $this->createQueryBuilder('fileVersionMeta')
             ->addSelect('fileVersion')
             ->addSelect('file')
@@ -33,7 +38,8 @@ class FileVersionMetaRepository extends EntityRepository implements FileVersionM
                 AccessControl::class,
                 'accessControl',
                 'WITH',
-                'accessControl.entityClass = :entityClass AND accessControl.entityId = collection.id'
+                'accessControl.entityClass = :entityClass '
+                . 'AND ' . $entityIdCondition
             )
             ->where('file.version = fileVersion.version')
             ->andWhere('accessControl.id is null');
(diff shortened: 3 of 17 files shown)


src/Sulu/Bundle/SecurityBundle/Resources/config/doctrine/AccessControl.orm.xml
```diff
@@ -13,7 +13,7 @@
         </id>
 
         <field name="permissions" type="smallint" column="permissions"/>
-        <field name="entityId" type="integer" column="entityId"/>
+        <field name="entityId" type="string" column="entityId" length="191"/>
```
alexander-schranz: Would set the length to `32`. This is in our cases never longer as a uuid we should not use more space as needed to not get to some kind of rowmaxsize error in some databases.

alexander-schranz: sorry uuids are `36`

UPGRADE.md
```diff
@@ -2,6 +2,16 @@
...
 ## 2.3
 
+### Change entityId field in AccessControl entity to string
+
+To allow entities with uuid's instead of auto generated ids to have object permissions,
+the `entityId` field of the `AccessControl` entity had to be changed from `integer` to `string`.
+Therefore the following sql statement needs to be executed.
+
+```sql
+ALTER TABLE se_access_controls CHANGE entityId entityId VARCHAR(191) NOT NULL;
```
```suggestion
ALTER TABLE se_access_controls CHANGE entityId entityId VARCHAR(36) NOT NULL;
```

luca-rath: oops 🙈 forgot about that

luca-rath:
@alexander-schranz Do you have an idea why those tests are failing on php7.2?

src/Sulu/Bundle/SecurityBundle/Entity/AccessControl.php
```diff
@@ -38,7 +38,7 @@ class AccessControl implements AccessControlInterface
     /**
      * The id of the model this access control rule applies to.
      *
-     * @var int
+     * @var string
```
alexander-schranz, line 41: I don't think we can change this to string I think for BC I would go here with `string|int` and go in the `getEntityId()` with a `is_numeric` check, because smebody could made here a typesafe check with `===` to the access control object entityId so I would go here with:

```php
public function getEntityId()
{
     if (\is_numeric($this->entityId)) {
           return (int) $this->entityId;
     }
     
     return $this->entityId;
}
```

src/Sulu/Bundle/MediaBundle/Entity/FileVersionMetaRepository.php
```diff
@@ -20,6 +20,11 @@ class FileVersionMetaRepository extends EntityRepository implements FileVersionM
 {
     public function findLatestWithoutSecurity()
     {
+        $entityIdCondition = 'accessControl.entityId = collection.id';
+        if ('postgresql' === $this->getDatabasePlatformName()) {
+            $entityIdCondition = 'accessControl.entityId = CAST(collection.id AS VARCHAR)';
```
niklasnatter: does this make problems with other databases than postgres? 
think i would not make this dependent on the `DatabasePlatformName` of not necessary - in theory, the repository should only know about doctrine and not about the specific database. doctrine should be responsible for abstracting the differences between different platforms

src/Sulu/Bundle/SecurityBundle/Entity/AccessControl.php
```diff
@@ -76,12 +76,16 @@ public function setPermissions($permissions)
 
     public function getEntityId()
     {
+        if (\is_numeric($this->entityId)) {
```
niklasnatter, line 79: dont think there will ever be a float/double value, but i guess i would use `is_int` anyway just to be safe

src/Sulu/Bundle/SecurityBundle/EventListener/PhpcrSecuritySubscriber.php
```diff
@@ -0,0 +1,62 @@
...
+    {
+        $type = $event->getType();
+        $identifier = $event->getIdentifier();
+        $permissions = $event->getPermissions();
+
+        if (!$this->phpcrAccessControlProvider->supports($type)) {
+            return;
+        }
+
+        $this->doctrineAccessControlProvider->setPermissions($type, $identifier, $permissions);
```
niklasnatter, line 61: would be nice to add a comment to explain the reason for this class somewhere
```
// sync permissions stored in phpcr to doctrine to make them usable in doctrine queries and the list builder
```

src/Sulu/Bundle/SecurityBundle/Command/SyncPhpcrPermissionsCommand.php
```diff
@@ -0,0 +1,147 @@
...
+        $this->entityManager = $entityManager;
+        $this->documentManager = $documentManager;
+        $this->doctrineAccessControlProvider = $doctrineAccessControlProvider;
+
+        parent::__construct();
+    }
+
+    public function configure(): void
+    {
+        $this->setDescription('Sync object permissions from phpcr into database');
```
niklasnatter:
```suggestion
        $this->setDescription('Sync existing object permissions from phpcr into database to make them usable in database queries');
```

UPGRADE.md
```diff
@@ -2,6 +2,16 @@
 
 ## 2.3
 
+### Change entityId field in AccessControl entity to string
```
niklasnatter, line 17: think we need to add a section for executing the `sulu:security:sync-phpcr-permissions` command

src/Sulu/Bundle/MediaBundle/Entity/FileVersionMetaRepository.php
luca-rath: the problem is, that casting decreases the performance ... So I don't want to cast those columns when using mysql, because it's not necessary

src/Sulu/Bundle/SecurityBundle/Entity/AccessControl.php
luca-rath: `is_int` won't match something like `"2"`, only `is_numeric` does

src/Sulu/Bundle/MediaBundle/Entity/FileVersionMetaRepository.php
niklasnatter: have you tested if this actually decreases performance? i am quite sure that mysql will implicitly cast this anyway as part of the comparison 🙂

src/Sulu/Bundle/SecurityBundle/Command/SyncPhpcrPermissionsCommand.php
```diff
@@ -0,0 +1,147 @@
...
+        $count = 0;
+        foreach ($documents as $document) {
+            $io->progressAdvance();
+
+            if (!$document instanceof SecurityBehavior || !$document instanceof UuidBehavior) {
+                continue;
+            }
+
+            $this->doctrineAccessControlProvider->setPermissions(
+                SecurityBehavior::class,
```
alexander-schranz, line 98: should we have here not a difference between pages and articles, else it could make problems when we migrating articles or pages away from phpcr as there is no real identifier which permission are pages and which are articles.

src/Sulu/Bundle/SecurityBundle/Command/SyncPhpcrPermissionsCommand.php
```diff
@@ -0,0 +1,147 @@
...
+     * @return QueryResultCollection<object>
+     */
+    protected function findDocuments(array $roleIds): QueryResultCollection
+    {
+        $roleIdsConstraint = \implode(' OR ', \array_map(function($roleId) {
+            return \sprintf('[%s%s] IS NOT NULL', SecuritySubscriber::SECURITY_PROPERTY_PREFIX, $roleId);
+        }, $roleIds));
+
+        $sql2 = \sprintf(
+            'SELECT * FROM [nt:unstructured] AS a WHERE (%s)',
```
alexander-schranz, line 141: Should we do here some batching to avoid problems if a lot of nodes are loaded here?

alexander-schranz (changes requested):

src/Sulu/Bundle/MediaBundle/Entity/FileVersionMetaRepository.php
luca-rath: No I haven't tested ... But I don't have a strong opinion here, I'll change it

src/Sulu/Bundle/SecurityBundle/Entity/AccessControl.php
niklasnatter: you are right - did not think about that 🙈

src/Sulu/Bundle/SecurityBundle/Command/SyncPhpcrPermissionsCommand.php
luca-rath: I don't think we can do this easily now?

src/Sulu/Bundle/SecurityBundle/Command/SyncPhpcrPermissionsCommand.php
luca-rath: I don't think this is necessary, because the `QueryResultCollection` hydrates the documents one by one in the `current()` method, not all at the beginning

src/Sulu/Bundle/SecurityBundle/Command/SyncPhpcrPermissionsCommand.php
alexander-schranz: Why can we not use `get_class` here? I would just merge page and home documents together and use so:

```php
$document instanceOf BasePageDocument ? BasePageDocument::class : get_class($document);
```

luca-rath: But in the `PhpcrSecuritySubscriber`, we don't have the actual document. We just have `$type` which is `SecurityBehavior`

UPGRADE.md
luca-rath: fixed

src/Sulu/Bundle/SecurityBundle/EventListener/PhpcrSecuritySubscriber.php
luca-rath: fixed

src/Sulu/Bundle/MediaBundle/Entity/FileVersionMetaRepository.php
luca-rath: fixed

src/Sulu/Bundle/SecurityBundle/Command/SyncPhpcrPermissionsCommand.php
alexander-schranz: Okay didn't know that. Then its okay for me to stay with `SecurityBehavior` at current state and check the permission migration when we are migrating pages to doctrine, will just make the update statement little bit more complex but at the end its just a join.

src/Sulu/Bundle/SecurityBundle/Command/SyncPhpcrPermissionsCommand.php
alexander-schranz: Sounds good and as we are only selecting pages which actually have permission this should not be a too big problem

src/Sulu/Bundle/CoreBundle/DependencyInjection/SuluCoreExtension.php
```diff
@@ -11,6 +11,7 @@
 
 namespace Sulu\Bundle\CoreBundle\DependencyInjection;
 
+use Oro\ORM\Query\AST\Functions\Cast;
```
niklasnatter, line 14: should we mark the `Sulu\Component\Rest\DQL\Cast` as deprecated?

src/Sulu/Bundle/MediaBundle/Entity/FileVersionMetaRepository.php
```diff
@@ -33,7 +32,8 @@ public function findLatestWithoutSecurity()
                 AccessControl::class,
                 'accessControl',
                 'WITH',
-                'accessControl.entityClass = :entityClass AND accessControl.entityId = collection.id'
+                'accessControl.entityClass = :entityClass '
+                . 'AND accessControl.entityId = CAST(collection.id AS CHAR(36))'
```
niklasnatter: the `CAST` function of the `oroinc/doctrine-extensions` package supports `string` as type. maybe we could use that instead of hardcoding a length here?

niklasnatter: think the `CAST` function will use `char` on mysql and `varchar` on postgres when using `string` 🙂 

- https://github.com/oroinc/doctrine-extensions/blob/master/src/Oro/ORM/Query/AST/Platform/Functions/Mysql/Cast.php
- https://github.com/oroinc/doctrine-extensions/blob/master/src/Oro/ORM/Query/AST/Platform/Functions/Postgresql/Cast.php

src/Sulu/Bundle/CoreBundle/DependencyInjection/SuluCoreExtension.php
luca-rath: Would definitely do this

niklasnatter (approved):

PR #5966, opened 2021-04-26
Enhance DoctrineListBuilder to make permission check more dynamic

luca-rath:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

Enhance DoctrineListBuilder and AccessControlQueryEnhancer to allow permission check with dynamic `securedEntityName`

### DIFF
--- a/src/Sulu/Bundle/SecurityBundle/AccessControl/AccessControlQueryEnhancer.php
+++ b/src/Sulu/Bundle/SecurityBundle/AccessControl/AccessControlQueryEnhancer.php
@@ -40,9 +40,13 @@ class AccessControlQueryEnhancer
         QueryBuilder $queryBuilder,
         UserInterface $user = null,
         int $permission,
-        string $entityClass,
-        string $entityAlias
+        ?string $entityClass,
+        string $entityAlias,
+        ?string $entityClassField = null,
+        ?string $entityIdField = 'id'
     ) {
+        $entityIdField = $entityIdField ?? 'id';
+
         $systemRoleQueryBuilder = $this->entityManager->createQueryBuilder()
             ->from(RoleInterface::class, 'systemRoles')
             ->select('systemRoles.id')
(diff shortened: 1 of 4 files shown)


src/Sulu/Bundle/SecurityBundle/AccessControl/AccessControlQueryEnhancer.php
```diff
@@ -40,9 +40,13 @@ public function enhance(
         QueryBuilder $queryBuilder,
         UserInterface $user = null,
         int $permission,
-        string $entityClass,
-        string $entityAlias
+        ?string $entityClass,
+        string $entityAlias,
+        ?string $entityClassField = null,
+        ?string $entityIdField = 'id'
```
niklasnatter: its a bit confusing what parameter does what here. but dont have a better idea, so its okay for me 😇

niklasnatter: one thing i can think of would be leaving this method like it is and add a second `enhanceWithFields` method:
```
    public function enhanceWithFields(
        QueryBuilder $queryBuilder,
        UserInterface $user = null,
        int $permission,
        string $entityClassField,
        string $entityIdField = 'id'
        string $entityAlias
    )
```

alexander-schranz:
Its a very special case that not only the `id` but also the `securityContext` is loaded from the a table column. I'm not sure if we should extend in this case the listBuilder and instead should add the required joins and where conditions ourself in this case to the listBuilder in the controller directly instead of using here the buildin security check.

luca-rath:
> Its a very special case that not only the `id` but also the `securityContext` is loaded from the a table column. I'm not sure if we should extend in this case the listBuilder and instead should add the required joins and where conditions ourself in this case to the listBuilder in the controller directly instead of using here the buildin security check.

I'm pretty sure, if someone changes the `AccessControlQueryEnhancer` at some point in the future, he will most likely forget to to the same in the `EventRecordController`. I think it wouldn't hurt to implement the `enhanceWithDynamicFields` there and add a corresponding method to the `DoctrineListBuilder`

alexander-schranz:
> I'm pretty sure, if someone changes the AccessControlQueryEnhancer at some point in the future, he will most likely forget to to the same in the EventRecordController. I think it wouldn't hurt to implement the enhanceWithDynamicFields there and add a corresponding method to the DoctrineListBuilder

Still  this case is so special about security I don't think that it is part of the enhance `AccessControlQueryEnhancer` which is build that the context is given from outside and not from a table. I think it is just mixing to much things and make the interface for the listBuilder less understandable for just one usecase which we probably will never have again for another entity with security.

I'm also thinking about not even use the listBuilder here and just create the query ourselve.

niklasnatter (approved):
Looks good for me 🙂 
@alexander-schranz What do you think?

alexander-schranz (approved):

PR #5972, opened 2021-04-29
Test Symfony 5.3 compatibility

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? | no 
| BC breaks? | no 
| Deprecations? | no  
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Test against minimum stability.

#### Why?

Test run against symfony 5.3

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -246,5 +246,5 @@
         },
         "sort-packages": true
     },
-    "minimum-stability": "dev"
+    "minimum-stability": "beta"
 }
--- a/composer.json
+++ b/composer.json
@@ -245,5 +245,6 @@
             "*": "dist"
         },
         "sort-packages": true
-    }
+    },
+    "minimum-stability": "beta"
 }
--- a/composer.json
+++ b/composer.json
@@ -246,5 +246,5 @@
         },
         "sort-packages": true
     },
-    "minimum-stability": "beta"
+    "minimum-stability": "dev"
 }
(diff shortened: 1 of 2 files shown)


alexander-schranz:
@sulu/core-team Seems like the only issue sulu did have with Symfony 5.3 was already fixed in https://github.com/symfony/symfony/pull/41009. Tests are running on `beta` version 🎉 . There is currently an issue with running on `dev` but this is related to `doctrine-bundle` and `doctrine/orm` 2.9 which they are already working on it  (https://github.com/doctrine/DoctrineBundle/pull/1327). So not related to the Symfony 5.3 version.

.github/workflows/test-application.yaml
```diff
@@ -94,7 +94,7 @@ jobs:
                       dependency-versions: 'highest'
                       php-extensions: 'ctype, iconv, mysql, imagick'
                       tools: 'composer:v2'
-                      lint: true
+                      lint: false # disabled to ignore phpstan changes
```
alexander-schranz: phpstan will need some work after Symfony 5.3 is released as they added better typehints to some classes we use, we can discuss doing a baseline or invest here some time add missing phpdocs.

StaffNowa: Yes, Symfony 5.3 will be shorty released 🎉

alexander-schranz: Mostly last day of may 🎉

StaffNowa: Yes :) 🎉

alexander-schranz:
Close as this is not longer needed and the above issues were fixed in Symfony before 5.3 was released. Also doctrine bundle is compatible after they release.

PR #5973, opened 2021-04-29
Add domain events for contacts

luca-rath:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

Add domain events for contacts

### DIFF
--- a/src/Sulu/Bundle/ContactBundle/Controller/AbstractMediaController.php
+++ b/src/Sulu/Bundle/ContactBundle/Controller/AbstractMediaController.php
@@ -14,10 +14,6 @@ namespace Sulu\Bundle\ContactBundle\Controller;
 use Doctrine\ORM\EntityManagerInterface;
 use FOS\RestBundle\View\ViewHandlerInterface;
 use Sulu\Bundle\ContactBundle\Contact\AbstractContactManager;
-use Sulu\Bundle\ContactBundle\Domain\Event\ContactMediaAddedEvent;
-use Sulu\Bundle\ContactBundle\Domain\Event\ContactMediaRemovedEvent;
-use Sulu\Bundle\ContactBundle\Entity\ContactInterface;
-use Sulu\Bundle\EventLogBundle\Application\Collector\DomainEventCollectorInterface;
 use Sulu\Bundle\MediaBundle\Api\Media;
 use Sulu\Bundle\MediaBundle\Entity\MediaRepositoryInterface;
 use Sulu\Bundle\MediaBundle\Media\ListBuilderFactory\MediaListBuilderFactory;
@@ -101,11 +97,6 @@ abstract class AbstractMediaController extends AbstractRestController
      */
     private $fieldDescriptorFactory;
 
-    /**
-     * @var DomainEventCollectorInterface|null
-     */
-    private $domainEventCollector;
-
     public function __construct(
         ViewHandlerInterface $viewHandler,
         TokenStorageInterface $tokenStorage,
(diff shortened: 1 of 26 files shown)


src/Sulu/Bundle/ContactBundle/Controller/AbstractMediaController.php
```diff
@@ -135,10 +135,11 @@ public function __construct(
      * @param string $entityName
      * @param string $id
      * @param string $mediaId
+     * @param callable|null $dispatchDomainEvent
      *
      * @return Media
      */
-    protected function addMediaToEntity($entityName, $id, $mediaId)
+    protected function addMediaToEntity($entityName, $id, $mediaId, $dispatchDomainEvent = null)
```
niklasnatter: think it would be a bit easier to understand to pass the domain event and dispatch it here instead of passing a callable? 🙂

src/Sulu/Bundle/ContactBundle/Resources/config/services.xml
```diff
@@ -45,7 +45,6 @@
             <argument type="service" id="sulu.repository.media" />
             <argument type="service" id="sulu_event_log.domain_event_collector" />
             <argument type="service" id="sulu.repository.user" />
-            <argument type="string">%sulu.model.account.class%</argument>
```
niklasnatter, line 48: 😂

src/Sulu/Bundle/ContactBundle/Controller/AbstractMediaController.php
luca-rath: you mean passing the `ContactDomainEvent::class` string? This would force every domain event passed here to have a constructor like `__construct($contact, $media)` ... I guess passing a callable is cleaner here, also the whole ContactBundle is built with callables like this one.

src/Sulu/Bundle/ContactBundle/Resources/config/services.xml
luca-rath: It's always a pleasure to fix something like that 🙈

src/Sulu/Bundle/ContactBundle/Controller/AbstractMediaController.php
niklasnatter: no i thought that we can already create the event. but i figured that we cannot do that because we dont have the entities there, so this way is fine for me 🙂

luca-rath: 👍

niklasnatter: think i would call it `dispatchDomainEventCallback` then - guess you will need to rebase this pr anyway once #5990 is merged 🙂

src/Sulu/Bundle/ContactBundle/Controller/AbstractMediaController.php
```diff
@@ -185,13 +191,14 @@ protected function addMediaToEntity($entityName, $id, $mediaId)
      * @param string $entityName
      * @param string $id
      * @param string $mediaId
+     * @param callable|null $dispatchDomainEvent
      *
      * @return Response
      */
-    protected function removeMediaFromEntity($entityName, $id, $mediaId)
+    protected function removeMediaFromEntity($entityName, $id, $mediaId, $dispatchDomainEvent = null)
```
niklasnatter: think i would call this `dispatchDomainEventCallback`

src/Sulu/Bundle/ContactBundle/Controller/AbstractMediaController.php
luca-rath: Yes you're right, that would be a better name ... Will change it

PR #5978, opened 2021-05-01
Single file for our dev security config

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? | no 
| BC breaks? | no 
| Deprecations? | no  
| Fixed tickets | fixes # 
| Related issues/PRs | https://github.com/sulu/skeleton/pull/114
| License | MIT

#### What's in this PR?

Move security.yaml toghet as the skeleton now enable security bundle by default for website context.

#### Why?

In the last years symfony did improve a lot and the firewall is lazy and does not has any effect on caching. To keep moving forward to a single kernel setup this PR enabled the security bundle also in the website.


config/packages/test/security.yaml
```diff
@@ -1,12 +1,12 @@
...
     providers:
-        testprovider:
+        sulu:
             id: test_user_provider
 
     firewalls:
         admin:
             http_basic:
-                provider: testprovider
+                provider: sulu
```
alexander-schranz, line 9: This fixes:

```
  Not configuring explicitly the provider for the "form_login" listener on "s  
  ulu_test" firewall is ambiguous as there is more than one registered provid  
  er.     
```

In 1.6 this was needed as the test user provider did only handle the test user but in 2.x now thes testuserprovider does fallback now to the sulu user provider so we can just override here in the test env that the `test_user_provider` can be used.

PR #5188, opened 2020-04-07
Add support for Symfony RoutingConfigurator in Kernel

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes (does mostly not effect any projects)
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

Add support for Symfony RoutingConfigurator in Kernel.

#### Why?

In Symfony 5 the configureRoutes will get a RoutingConfigurator instead of RouteCollectionBuilder. To support both we need to remove the typehint.

#### To Do

- [x] Add breaking changes to UPGRADE.md

PR #5979, opened 2021-05-01
Fix route collection builder is deprecation message

StaffNowa:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | yes
| Fixed tickets |
| Related issues/PRs |
| License | MIT
| Documentation PR |

#### What's in this PR?

Since symfony/routing 5.1: The "Symfony\Component\Routing\RouteCollectionBuilder" class is deprecated, use "Symfony\Component\Routing\Loader\Configurator\RoutingConfigurator" instead.

These versions is no longer maintained.
https://symfony.com/releases/4.0
https://symfony.com/releases/4.1
https://symfony.com/releases/4.2
https://symfony.com/releases/4.3

### DIFF
--- a/src/Sulu/Component/HttpKernel/SuluKernel.php
+++ b/src/Sulu/Component/HttpKernel/SuluKernel.php
@@ -151,6 +151,9 @@ abstract class SuluKernel extends Kernel
         }
     }
 
+    /**
+     * * @param RouteCollectionBuilder|RoutingConfigurator $routes Is a RouteCollectionBuilder for Symfony <= 4.4
+     */
     protected function import($routes, $confDir, $pattern)
     {
         $configExtensions = $this->getConfigExtensions();
--- a/src/Sulu/Component/HttpKernel/SuluKernel.php
+++ b/src/Sulu/Component/HttpKernel/SuluKernel.php
@@ -124,6 +124,9 @@ abstract class SuluKernel extends Kernel
         return $class;
     }
 
+    /**
+     * * @param RouteCollectionBuilder|RoutingConfigurator $routes Is a RouteCollectionBuilder for Symfony <= 4.4
+     */
     protected function configureRoutes(RoutingConfigurator $routes)
     {
         $confDir = $this->getProjectDir() . '/config';
@@ -149,7 +152,7 @@ abstract class SuluKernel extends Kernel
     }
 
     /**
-     * @param RoutingConfigurator $routes
+     * @param RouteCollectionBuilder|RoutingConfigurator $routes Is a RouteCollectionBuilder for Symfony <= 4.4
      */
     protected function import($routes, $confDir, $pattern)
     {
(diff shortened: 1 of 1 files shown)


src/Sulu/Component/HttpKernel/SuluKernel.php
```diff
@@ -150,7 +149,9 @@ protected function load(LoaderInterface $loader, $confDir, $pattern)
     }
 
     /**
-     * @param RouteCollectionBuilder|RoutingConfigurator $routes Is a RouteCollectionBuilder for Symfony <= 4.4
+     * @param RoutingConfigurator $routes
+     * @param $confDir
+     * @param $pattern
```
alexander-schranz: We do not add `@param` which add any value, thats why phpstan is here failing now.

```suggestion
```

src/Sulu/Component/HttpKernel/SuluKernel.php
```diff
@@ -125,7 +124,7 @@ protected function generateContainerClass($class)
         return $class;
     }
 
-    protected function configureRoutes(RouteCollectionBuilder $routes)
+    protected function configureRoutes(RoutingConfigurator $routes)
```
alexander-schranz: The `4.4` micro kernel seems still integrate this with `RouteCollectionBuilder`(https://github.com/symfony/symfony/blob/30dd60857692d182aaa0a0d38ee69cb2848fdd3b/src/Symfony/Bundle/FrameworkBundle/Kernel/MicroKernelTrait.php#L33). And the `RoutingConfigurator` seems not to extend from the new class: https://github.com/symfony/symfony/blob/4.4/src/Symfony/Component/Routing/RouteCollectionBuilder.php So this can lead to some issues.

I think we need to go here with a `phpdoc` like it was original done here (https://github.com/sulu/sulu/pull/5188) to avoid that the deprecated class is loaded.

alexander-schranz: Seems like this deprecated class was added again in a false resolved merge conflict. https://github.com/sulu/sulu/commit/7e5f7ea02d7eb207edb688122a30ec7c19cc07e8#diff-6aaa8aaea84882e3b2580d844f6b1cdacbf4fea47c6950e011cb3c41c38acdbb So I would here go with the original solution with the phpdoc to avoid problems with the typehint.

src/Sulu/Component/HttpKernel/SuluKernel.php
```diff
@@ -149,9 +151,6 @@ protected function load(LoaderInterface $loader, $confDir, $pattern)
         }
     }
 
-    /**
-     * @param RouteCollectionBuilder|RoutingConfigurator $routes Is a RouteCollectionBuilder for Symfony <= 4.4
-     */
     protected function import($routes, $confDir, $pattern)
```
alexander-schranz:
```suggestion
    /**
     * * @param RouteCollectionBuilder|RoutingConfigurator $routes Is a RouteCollectionBuilder for Symfony <= 4.4
     */
    protected function import($routes, $confDir, $pattern)
```

phpstan requires this.

alexander-schranz (approved):

alexander-schranz:
@StaffNowa Thx for bringing this up 👍 . Did rebase and fix the missing things, so this is part of the next release.

StaffNowa:
@alexander-schranz thanks :+1:

PR #5980, opened 2021-05-01
Run php lint task in own step

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? | no 
| BC breaks? | no 
| Deprecations? | no  
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Run test before linting tasks.

#### Why?

Its for PR more important that the test are green then phpstan or php-cs is green. This make in case of external contributions better to give feedback is something is failing in both in lint and test.

### DIFF
--- a/.github/workflows/test-application.yaml
+++ b/.github/workflows/test-application.yaml
@@ -61,7 +61,7 @@ jobs:
                   npm run build
 
     php:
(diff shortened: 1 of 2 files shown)


niklasnatter:
Think i would prefer to create a separate job for the `composer lint` script instead of changing the order 🙂

alexander-schranz:
@nnatter Yeah think that would be the best will change the PR

niklasnatter (approved):

PR #5984, opened 2021-05-03
Use PersistenceBundleTrait to resolve perisitence bundle entities

niklasnatter:


src/Sulu/Bundle/CoreBundle/DependencyInjection/SuluCoreExtension.php
```diff
@@ -177,10 +177,6 @@ public function prepend(ContainerBuilder $container)
                                 'CAST' => Cast::class,
                             ],
                         ],
-                        'resolve_target_entities' => [
-                            CollectionInterface::class => Collection::class,
```
niklasnatter, line 181: this will cause problems if the collection class is overwritten in a project

src/Sulu/Bundle/CoreBundle/DependencyInjection/SuluCoreExtension.php
```diff
@@ -178,8 +174,6 @@ public function prepend(ContainerBuilder $container)
                             ],
                         ],
                         'resolve_target_entities' => [
```
niklasnatter: think this can be fully removed

PR #5985, opened 2021-05-03
Remove UserManager compiler pass

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | yes
| License | MIT

#### What's in this PR?

Removes the compilerpass for the `UserManager`


niklasnatter (approved):

niklasnatter:
maybe we should target `2.1`?

Prokyonn:
> maybe we should target `2.1`?

I wasn't sure because of the BC break in the constructor, but I can rebase it :)

PR #5993, opened 2021-05-05
Update babel.config.json to resolve warning

niklasnatter:
When building the administration interface application, the following warning is logged multiple times at the moment:
```
Though the "loose" option was set to "false" in your @babel/preset-env config, it will not be used for @babel/plugin-proposal-private-methods since the "loose" mode option was set to "true" for @babel/plugin-proposal-class-properties.
The "loose" option must be the same for @babel/plugin-proposal-class-properties, @babel/plugin-proposal-private-methods and @babel/plugin-proposal-private-property-in-object (when they are enabled): you can silence this warning by explicitly adding
	["@babel/plugin-proposal-private-methods", { "loose": true }]
to the "plugins" section of your Babel config.
```
For example, see https://github.com/sulu/sulu/runs/2501494184

### DIFF
--- a/babel.config.json
+++ b/babel.config.json
@@ -2,8 +2,9 @@
     "presets": ["@babel/preset-env", "@babel/preset-react"],
     "plugins": [
         ["@babel/plugin-proposal-decorators", {"legacy": true}],
-        "@babel/plugin-proposal-object-rest-spread",
-        "@babel/plugin-transform-flow-strip-types",
-        ["@babel/plugin-proposal-class-properties", {"loose": true}]
-    ]
+        "@babel/plugin-transform-flow-strip-types"
+    ],
+    "assumptions": {
+        "setPublicClassFields": true
+    }
 }
--- a/package.json
+++ b/package.json
@@ -32,9 +32,7 @@
     },
     "devDependencies": {
         "@babel/core": "^7.5.5",
-        "@babel/plugin-proposal-class-properties": "^7.5.5",
         "@babel/plugin-proposal-decorators": "^7.4.4",
-        "@babel/plugin-proposal-object-rest-spread": "^7.5.5",
         "@babel/plugin-transform-flow-strip-types": "^7.4.4",
         "@babel/preset-env": "^7.5.5",
         "@babel/preset-react": "^7.0.0",


babel.config.json
```diff
@@ -2,8 +2,6 @@
     "presets": ["@babel/preset-env", "@babel/preset-react"],
     "plugins": [
         ["@babel/plugin-proposal-decorators", {"legacy": true}],
-        "@babel/plugin-proposal-object-rest-spread",
```
niklasnatter, line 5: this is included in the `@babel/preset-env`: https://babeljs.io/docs/en/babel-plugin-proposal-object-rest-spread

babel.config.json
```diff
@@ -2,8 +2,6 @@
...
-        "@babel/plugin-transform-flow-strip-types",
-        ["@babel/plugin-proposal-class-properties", {"loose": true}]
```
niklasnatter, line 7: i removed `{"loose": true}` and migrated to the `setPublicClassFields` assumption like described in https://babeljs.io/docs/en/babel-plugin-proposal-class-properties#loose

when removing `{"loose": true}`, we can remove the whole line because `@babel/plugin-proposal-class-properties` is included in `@babel/preset-env`

alexander-schranz, line 7: As I remember this babel loose and legacy are part of mobx4 requirements, sure that mobx4 will still work when we remove this?

niklasnatter: not sure if it is still needed, unfortunately i do not find any mobx documentation anymore..
i have built the application and tested it a bit. have not found any issues yet 🙈

niklasnatter: i think the alternative for solving the warning would be enabling `{"loose": true}` for the whole `@babel/preset-env`

niklasnatter: i think we could also set the `setPublicClassFields` assumption. not sure whats the best solution 🙈 
what do you think?
https://babeljs.io/docs/en/assumptions#setpublicclassfields

alexander-schranz: In the upgrade docs and in the mobx6 docs it still tells that mobx6 now need `"loose": false` which was different for mobx4 and mobx5. https://mobx.js.org/migrating-from-4-or-5.html / https://mobx.js.org/enabling-decorators.html.

I would not change this aslong as we don't upgrade to mobx4. Before we will not get ride of this warnings that this is deprecated.

alexander-schranz: @nnatter here is still the official mobx4 documentation: https://github.com/mobxjs/mobx/blob/4.15.7/docs/best/decorators.md

alexander-schranz: @nnatter https://github.com/mobxjs/mobx/blob/2ef73472434bab633b8f2517a7c7e441db47f6d7/CHANGELOG.md#550--450 here is a babel config provided which has loose true but additional decorator export.

niklasnatter: okay i would prefer to use the `setPublicClassFields` assumption then. the babel docs recommend to migrate to this: https://babeljs.io/docs/en/babel-plugin-proposal-class-properties#loose

what do you think?

alexander-schranz: not sure about that changelog as it seems it was not in a release.

alexander-schranz: The `setPublicClassFields` sounds for me good if it has the same effect.

alexander-schranz: I asked @mweststrate on [twitter](https://twitter.com/alex_s_/status/1389889380010139648) if he have experience with this change, so I will wait here if we maybe got an answer from him.

alexander-schranz:
Can @nnatter can you create a skeleton PR for this?

PR #5996, opened 2021-05-05
Add domain events for default snippets

luca-rath:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

Add domain events for default snippets

### DIFF
--- a/src/Sulu/Bundle/SnippetBundle/Domain/Event/DefaultSnippetModifiedEvent.php
+++ b/src/Sulu/Bundle/SnippetBundle/Domain/Event/DefaultSnippetModifiedEvent.php
@@ -1,0 +1,96 @@
+<?php
+
+/*
+ * This file is part of Sulu.
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\SnippetBundle\Domain\Event;
+
+use Sulu\Bundle\EventLogBundle\Domain\Event\DomainEvent;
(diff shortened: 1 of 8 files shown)


src/Sulu/Bundle/SnippetBundle/Domain/Event/DefaultSnippetModifiedEvent.php
```diff
@@ -0,0 +1,96 @@
...
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\SnippetBundle\Domain\Event;
+
+use Sulu\Bundle\EventLogBundle\Domain\Event\DomainEvent;
+use Sulu\Bundle\SnippetBundle\Admin\SnippetAdmin;
+use Sulu\Bundle\SnippetBundle\Document\SnippetDocument;
+
+class DefaultSnippetModifiedEvent extends DomainEvent
```
niklasnatter: think i would call this `WebspaceDefaultSnippetModifiedEvent` because all other events are prefixed with the resource too. or maybe we should use another resource-key? 🙈

luca-rath: I thought we agreed on using `webspaces` as resource, so I renamed it 🙈

niklasnatter: sounds good to me 🙂

PR #6006, opened 2021-05-07
Use boolean default value in RequestParametersTrait::getBooleanRequestParameter method

niklasnatter:


src/Sulu/Component/Rest/RequestParametersTrait.php
```diff
@@ -54,7 +54,7 @@ protected function getRequestParameter(Request $request, $name, $force = false,
      *
      * @return bool
      */
-    protected function getBooleanRequestParameter($request, $name, $force = false, $default = null)
+    protected function getBooleanRequestParameter($request, $name, $force = false, $default = false)
```
alexander-schranz, line 57: This could be a bc break if somebody used it and checked for `if (null === ...)`

alexander-schranz: I personal would just deprecate this methods and replace them in our code with `$request->query->getBoolean` from symfony: https://github.com/symfony/symfony/blob/883899eb8798936137ed12a9620d42fff4efb73f/src/Symfony/Component/HttpFoundation/ParameterBag.php#L167

niklasnatter: You are right about the BC break - will close this 🙈

PR #6007, opened 2021-05-07
Add activity translations for page events

luca-rath:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | #5962
| License | MIT
| Documentation PR | -

#### What's in this PR?

This pr adds activity translations for page events.
This pr should be used as reference to implement the translations for the activity types of the other resources.


src/Sulu/Bundle/PageBundle/Resources/translations/admin.en.json
```diff
@@ -61,5 +61,18 @@
...
-    "sulu_page.schedules_description": "The block will only be visible if at least one condition is valid."
+    "sulu_page.schedules_description": "The block will only be visible if at least one condition is valid.",
+    "sulu_activity.text.pages.created": "{userFullName} has created the page {resourceTitle}",
+    "sulu_activity.text.pages.modified": "{userFullName} has changed the page {resourceTitle}",
+    "sulu_activity.text.pages.removed": "{userFullName} has removed the page {resourceTitle}",
+    "sulu_activity.text.pages.published": "{userFullName} has published the page {resourceTitle}",
+    "sulu_activity.text.pages.unpublished": "{userFullName} has unpublished the page {resourceTitle}",
+    "sulu_activity.text.pages.draft_removed": "{userFullName} has discarded the draft of the page {resourceTitle}",
+    "sulu_activity.text.pages.copied": "{userFullName} has copied the page {resourceTitle}",
+    "sulu_activity.text.pages.moved": "{userFullName} has moved the page {resourceTitle} from \"{context.previousParentTitle}\" to \"{context.newParentTitle}\"",
```
luca-rath: @nnatter The problem here is, that this translation can't take the `{context.previousParentTitleLocale}` and the `{context.newParentTitleLocale}` into account ... How should we handle that?

niklasnatter: i think you mean that we cannot easily check if the `context.previousParentTitleLocale` is different to the `resourceLocale`, right? i think it would be okay to just always/never display them 🙂

luca-rath: not different to the `resourceLocale`, different to the locale of the user ... I would still prefer something like we did in the `ActivityController` for the `resourceTitleLocale`, if there is an "easy" way to do this ...

luca-rath: Maybe we could override all `context.*Title`, if both `context.*Title` and `context.*TitleLocale` is given? I know this is kind of hacky, but I could live with that ...

niklasnatter: i guess we could check if there is a translation parameter with the same name and a `Locale` suffix for each of the translations parameters to generalize it. 

i think i would not add the quotes like we do in the `resourceTitle` at the moment, but just append the locale like this `parameter value [locale]`. i think its a bit confusing at the moment that the `resourceTitle` already contains quotes but all other parameters do not

luca-rath: But the result should be `"Page A" [EN]` and this is not possible if we don't add the quotes already to the translation parameter

niklasnatter: i think it would be okay to display `"Page A [EN]"` instead of `"Page A" [EN]` in the UI if it makes things easier/more consistent to use 🙂

luca-rath: Then I think we should change that for the `resourceTitle` too and remove the quotes from the translation parameter

niklasnatter: yes would be okay for me 🙂

src/Sulu/Bundle/ActivityBundle/UserInterface/Controller/ActivityController.php
```diff
@@ -331,10 +331,19 @@ private function getTranslationParameters(array $array, string $prefix = '{', st
...
+
+                    $value = $this->getLocalizedValue($value, $valueLocale, $translationLocale);
+                }
+
+                $translationParameters[$prefix . $key . $suffix] = $value;
             } elseif (\is_array($value)) {
-                foreach ($this->getTranslationParameters($value, '', '') as $translationParameterKey => $translationParameterValue) {
-                    $translationParameters[$prefix . $key . '.' . $translationParameterKey . $suffix] = $translationParameterValue;
+                foreach ($this->getTranslationParameters($value, $translationLocale, '', '') as $translationParameterKey => $translationParameterValue) {
+                    $translationParameters[$prefix . $key . '_' . $translationParameterKey . $suffix] = $translationParameterValue;
```
luca-rath, line 346: @nnatter Just realized, that the js library `intl-messageformat` can't handle translations containing something like `{context.foo}` ... An error is thrown in the javascript console then. Therefore I changed it back to `{context_foo}`

niklasnatter: alright - thats a pity. but looks good to me otherwise 🙂

PR #6009, opened 2021-05-07
Update ajv package to new major version

niklasnatter:
See release notes for breaking changes: https://github.com/ajv-validator/ajv/releases/tag/v8.0.0


src/Sulu/Bundle/AdminBundle/Resources/js/package.json
```diff
@@ -15,9 +15,8 @@
         "@ckeditor/ckeditor5-paragraph": "^27.1.0",
         "@ckeditor/ckeditor5-table": "^27.1.0",
         "@ckeditor/ckeditor5-ui": "^27.1.0",
-        "ajv": "^7.1.0",
-        "ajv-formats": "^1.5.1",
-        "ajv-keywords": "^4.0.0",
```
niklasnatter, line 20: i noticed that the [`ajv-keywords` package](https://github.com/ajv-validator/ajv-keywords) adds custom keywords that are not part of the JSON schema specification. we added it in #5815 and it was never part of a released version, so i think we can remove it without worrying about backwards compatibility 🙂

alexander-schranz (approved):

PR #6014, opened 2021-05-10
Fix PortalRouteLoader for none bundle paths

### LINKED ISSUE #5858: Can not add portal routes
| Q | A
| --- | ---
| Sulu Version | 2.2.5
| PHP Version | PHP 7.3.24
| DB Version | 5.7.30
| Browser Version | Firefox 78.8.0esr

# Actual Behavior

The file "routes_website_portal.yaml" does not exist (in: "") in routes_website_portal.yaml (which is being imported from "/var/www/project/config/routes_website.yaml"). Make sure there is a loader supporting the "portal" type.

# Expected Behavior

Load the file routes_website_portal.yaml from the routes_website.yaml using type portal.

# Steps to Reproduce

Add this to  config/routes_website.yaml:
```YAML
project:
    type: portal
    resource: "routes_website_portal.yaml"
```

Add the file routes_website_portal.yaml into the config folder, too.

Notice: As soon as the "type: portal" is removed the error is gone (but then the type portal is not considered).

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? |  yes
| New feature? | no 
| BC breaks? | no 
| Deprecations? | no  
| Fixed tickets | fixes #5858
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Fix PortalRouteLoader for none bundle paths.

#### Why?

Currently a bundle was needed to create portal routes.

### DIFF
--- a/src/Sulu/Bundle/WebsiteBundle/Routing/PortalLoader.php
+++ b/src/Sulu/Bundle/WebsiteBundle/Routing/PortalLoader.php
@@ -12,23 +12,40 @@
 namespace Sulu\Bundle\WebsiteBundle\Routing;
 
 use Sulu\Component\Webspace\Manager\WebspaceManagerInterface;
-use Symfony\Component\Config\Loader\Loader;
+use Symfony\Component\Config\FileLocator;
+use Symfony\Component\Config\FileLocatorInterface;
+use Symfony\Component\Config\Loader\FileLoader;
 use Symfony\Component\Routing\Route;
 use Symfony\Component\Routing\RouteCollection;
 
 /**
  * This route loader is responsible for loading routes from a routing file, and adjust the routes in a way, so that
  * there will be an own route for every portal registered.
+ *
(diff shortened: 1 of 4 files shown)


src/Sulu/Bundle/WebsiteBundle/Routing/PortalLoader.php
```diff
@@ -12,32 +12,48 @@
...
     /**
      * @var WebspaceManagerInterface
      */
     private $webspaceManager;
 
-    public function __construct(WebspaceManagerInterface $webspaceManager)
-    {
+    public function __construct(
+        WebspaceManagerInterface $webspaceManager,
+        ?FileLocatorInterface $fileLocator = null
```
niklasnatter, line 36: this is not a bc break this way, right? should we target `2.1`?

alexander-schranz: yes this should not be a bc break this way

PR #6026, opened 2021-05-11
Add activity view to pages and media

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| License | MIT

#### What's in this PR?

Adds activity view to `PageAdmin` and `MediaAdmin`.

### DIFF
--- a/src/Sulu/Bundle/ActivityBundle/Infrastructure/Sulu/Admin/ActivityAdmin.php
+++ b/src/Sulu/Bundle/ActivityBundle/Infrastructure/Sulu/Admin/ActivityAdmin.php
@@ -98,7 +98,7 @@ class ActivityAdmin extends Admin
         if ($viewCollection->has(PageAdmin::EDIT_FORM_VIEW)) {
             $viewCollection->add(
                 $this->viewBuilderFactory
-                    ->createResourceTabViewBuilder(static::EDIT_FORM_ACTIVITY_VERSION_TAB_VIEW, '/history')
+                    ->createResourceTabViewBuilder(static::EDIT_FORM_ACTIVITY_VERSION_TAB_VIEW, '/activity')
                     ->setResourceKey(BasePageDocument::RESOURCE_KEY)
                     ->setTabOrder(6144)
                     ->setTabTitle($this->versioningEnabled ? 'sulu_admin.activity_versions' : 'sulu_admin.activity')
--- a/src/Sulu/Bundle/ActivityBundle/Infrastructure/Sulu/Admin/ActivityAdmin.php
+++ b/src/Sulu/Bundle/ActivityBundle/Infrastructure/Sulu/Admin/ActivityAdmin.php
@@ -17,6 +17,10 @@ use Sulu\Bundle\AdminBundle\Admin\Navigation\NavigationItem;
 use Sulu\Bundle\AdminBundle\Admin\Navigation\NavigationItemCollection;
 use Sulu\Bundle\AdminBundle\Admin\View\ViewBuilderFactoryInterface;
 use Sulu\Bundle\AdminBundle\Admin\View\ViewCollection;
+use Sulu\Bundle\MediaBundle\Admin\MediaAdmin;
+use Sulu\Bundle\MediaBundle\Entity\MediaInterface;
+use Sulu\Bundle\PageBundle\Admin\PageAdmin;
+use Sulu\Bundle\PageBundle\Document\BasePageDocument;
 use Sulu\Component\Security\Authorization\PermissionTypes;
 use Sulu\Component\Security\Authorization\SecurityCheckerInterface;
 
@@ -25,6 +29,8 @@ class ActivityAdmin extends Admin
     const SECURITY_CONTEXT = 'sulu.activities.activities';
 
     const LIST_VIEW = 'sulu_activity.activities.list';
+
+    const EDIT_FORM_ACTIVITY_VERSION_TAB_VIEW = 'sulu_page.page_edit_form.activity_version_tab';
 
     /**
      * @var ViewBuilderFactoryInterface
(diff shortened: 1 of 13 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/components/Table/table.scss
```diff
@@ -27,7 +27,6 @@ $borderRadius: 3px;
 .table {
     border-collapse: separate;
     border-spacing: 0 5px;
-    margin: -5px 0;
```
niklasnatter, line 30: 😮

src/Sulu/Bundle/MediaBundle/Admin/MediaAdmin.php
```diff
@@ -150,12 +153,37 @@ public function configureViews(ViewCollection $viewCollection): void
...
             );
+
             $viewCollection->add(
                 $this->viewBuilderFactory
                     ->createViewBuilder(static::EDIT_FORM_HISTORY_VIEW, '/history', 'sulu_media.history')
                     ->setOption('tabTitle', 'sulu_media.history')
                     ->setParent(static::EDIT_FORM_VIEW)
             );
+
+            $viewCollection->add(
```
niklasnatter: i think i would add this view in the `ActivityAdmin`. we did something similar in the `SecurityAdmin` or in the `AutomationAdmin `

niklasnatter: would use the `$viewCollection->has()` method to check if the parent view is present

src/Sulu/Bundle/MediaBundle/Admin/MediaAdmin.php
```diff
@@ -150,12 +153,37 @@ public function configureViews(ViewCollection $viewCollection): void
...
+                        ],
+                    ])
+                    ->disableTabGap()
+                    ->disableSearching()
+                    ->disableSelection()
+                    ->disableColumnOptions()
+                    ->disableFiltering()
+                    ->addResourceStorePropertiesToListRequest(['id' => 'resourceId'])
+                    ->addRequestParameters(['resourceKey' => MediaInterface::RESOURCE_KEY])
+                    ->addRouterAttributesToListRequest(['locale'])
```
niklasnatter: not sure if this is needed, i think the locale should be sent to the server automatically. think it could event break things to set it like this when changing the locale 🙈

src/Sulu/Bundle/PageBundle/Admin/PageAdmin.php
```diff
@@ -316,6 +319,39 @@ public function configureViews(ViewCollection $viewCollection): void
                     ->setTabOrder(5120)
                     ->setParent(static::EDIT_FORM_VIEW)
             );
+
+            $viewCollection->add(
+                $this->viewBuilderFactory
+                    ->createResourceTabViewBuilder(static::EDIT_FORM_ACTIVITY_VERSION_TAB_VIEW, '/history')
```
niklasnatter: not sure about `/history`. but i guess we need to set something right?

src/Sulu/Bundle/PageBundle/Admin/PageAdmin.php
```diff
@@ -316,6 +319,39 @@ public function configureViews(ViewCollection $viewCollection): void
...
+                        ],
+                    ])
+                    ->disableTabGap()
+                    ->disableSearching()
+                    ->disableSelection()
+                    ->disableColumnOptions()
+                    ->disableFiltering()
+                    ->addResourceStorePropertiesToListRequest(['id' => 'resourceId'])
+                    ->addRequestParameters(['resourceKey' => BasePageDocument::RESOURCE_KEY])
+                    ->addRouterAttributesToListRequest(['locale'])
```
niklasnatter: same as above

src/Sulu/Bundle/PageBundle/Admin/PageAdmin.php
Prokyonn: yeah I'm also not happy with the naming but I didn't came up with a better one. The other alternative would be `log` or something like that.

src/Sulu/Bundle/PageBundle/Resources/js/containers/Form/fields/SettingsVersions.js
```diff
@@ -156,9 +156,11 @@ class SettingsVersions extends React.Component<Props> {
             <Fragment>
                 <List
                     adapters={['table']}
+                    filterable={false}
                     itemActionsProvider={this.getListItemActions}
                     searchable={false}
                     selectable={false}
+                    showColumnOptions={false}
```
niklasnatter, line 163: 🎉

src/Sulu/Bundle/ActivityBundle/Infrastructure/Sulu/Admin/ActivityAdmin.php
```diff
@@ -81,6 +94,65 @@ public function configureViews(ViewCollection $viewCollection): void
                 ])
                 ->addToolbarActions([])
         );
+
+        if ($viewCollection->has(PageAdmin::EDIT_FORM_VIEW)) {
+            $viewCollection->add(
+                $this->viewBuilderFactory
+                    ->createResourceTabViewBuilder(static::EDIT_FORM_ACTIVITY_VERSION_TAB_VIEW, '/history')
```
niklasnatter: think i would just call it `/activity` and use `activity/activity` and `activity/versions` later on. that way we would not introduce a new term `history`

```suggestion
                    ->createResourceTabViewBuilder(static::EDIT_FORM_ACTIVITY_VERSION_TAB_VIEW, '/activity')
```

PR #6040, opened 2021-05-14
guzzle.xml - required package + version on deprecated

StaffNowa:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | yes
| License | MIT

#### What's in this PR?

Since symfony/dependency-injection 5.1: Not setting the attribute "version" of the node "deprecated" in "/var/www/project/vendor/sulu/sulu/src/Sulu/Bundle/LocationBundle/DependencyInjection/../Resources/config/guzzle.xml" is deprecated.

#### Why?

Since symfony/dependency-injection 5.1: Not setting the attribute "package" of the node "deprecated" in "/var/www/project/vendor/sulu/sulu/src/Sulu/Bundle/LocationBundle/DependencyInjection/../Resources/config/guzzle.xml" is deprecated.

### DIFF
--- a/src/Sulu/Bundle/LocationBundle/Resources/config/guzzle.xml
+++ b/src/Sulu/Bundle/LocationBundle/Resources/config/guzzle.xml
@@ -8,7 +8,7 @@
     <services>
         <!-- Guzzle client -->
         <service id="sulu_location.geolocator.guzzle.client" class="%sulu_location.guzzle.client.class%">
-            <deprecated />
+            <deprecated package="sulu/sulu" version="2.2" />
         </service>
     </services>
 </container>


alexander-schranz:
@StaffNowa we can not add this as we still support Symfony 4.4 and Symfony 4.4 will fail here, so most deprecation which where introduced after Symfony 4.4 can be ignored and will be tackled when we will drop support for Symfony 4.4, this will mostly be when Symfony 6 will be released and we will then upgrade to 5.4 / 6.0 as min requirement, until there we will stay with 4.4 / 5.* aslong as it doesn't make any problems. I will close this Pull Request, still thank you for your effort!

StaffNowa:
@alexander-schranz very good explanation 👍 thanks 🙂

PR #6050, opened 2021-05-17
Add upload action to media list displayed in media selection overlay

### LINKED ISSUE #5039: Add Upload button to Media Selection Overlay

niklasnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Fixed tickets | fixes #5039
| Related issues/PRs | #5883
| License | MIT

#### What's in this PR?

This pull request add a `actions` prop to the `List` container that allows to display additional actions in the list toolbar beside the adapter switcher. It uses this `actions` prop to display an `Upload file` button in the media list that is displayed in the media selection overlay. 

Screenshot 2021-05-17 at 18 20 42

#### Why?

At the moment, there is no button for uploading a new file inside of the overlay of the `media_selection` field type. See #5039 and #5883.

Additionally, the `actions` prop of the `List` container is a nice extension point and might be useful for other things in the future too 🙂


src/Sulu/Bundle/MediaBundle/Resources/js/containers/MediaCollection/MediaCollection.js
```diff
@@ -108,25 +118,24 @@ class MediaCollection extends React.Component<Props> {
...
                 <Divider />
-                <div>
-                    <MediaSection
-                        adapters={mediaListAdapters}
-                        listStore={mediaListStore}
-                        mediaListRef={mediaListRef}
-                        onMediaClick={this.handleMediaClick}
-                    />
-                </div>
+                <List
```
niklasnatter, line 135: decided to directly render a `List` instead of the `MediaSection` because the `MediaSection` does not really do anything and I would need to add a `actions` prop to the `MediaSection`

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/List.js
```diff
@@ -65,6 +67,7 @@ const USER_SETTING_ADAPTER = 'adapter';
 @observer
 class List extends React.Component<Props> {
     static defaultProps = {
+        actions: [],
```
alexander-schranz, line 70: think this should be added to the `README.md`

src/Sulu/Bundle/MediaBundle/Resources/js/views/MediaOverview/MediaOverview.js
```diff
@@ -260,33 +259,19 @@ export default withToolbar(MediaOverview, function() {
...
         locked: collectionLocked,
     } = this.collectionStore;
 
-    const addPermission = collectionPermissions.add !== undefined ? collectionPermissions.add : routeAddPermission;
     const deletePermission = collectionPermissions.delete !== undefined
         ? collectionPermissions.delete
         : routeDeletePermission;
     const editPermission = collectionPermissions.edit !== undefined ? collectionPermissions.edit : routeEditPermission;
 
-    if (addPermission) {
```
alexander-schranz, line 273: I would not remove this button in the toolbar.

alexander-schranz (changes requested):
@nnatter as discussed I would still show the Upload button in the toolbar, but to avoid confusen we will show the button only when no toolbar is available.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/List.js
niklasnatter: yes you are right, but it looks like the `README.md` does not contain most other props like `adapterOptions`/`paginated`/`selectable`/... too 😕

src/Sulu/Bundle/MediaBundle/Resources/js/views/MediaOverview/MediaOverview.js
niklasnatter: done. the button is now displayed in the toolbar instead of the list 🙂

alexander-schranz (approved):

PR #6055, opened 2021-05-18
Add versions tab to activity view in page details

niklasnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| License | MIT

#### What's in this PR?

This PR moves the interface for restoring previous page versions from the settings tab to a new tab inside of the activity view.

#### Why?

Because page versions are related to the activity of the page.


niklasnatter:
This PR is not finished yet. We really need to think about how to style this - at the moment, the tab looks like this:
Screenshot 2021-05-18 at 13 06 53

niklasnatter:
The tab looks like this now:
Screenshot 2021-05-26 at 15 44 14

alexander-schranz (approved):

alexander-schranz (approved):

PR #6056, opened 2021-05-18
Add @babel/plugin-proposal-nullish-coalescing-operator to transpile code of react-leaflet package

niklasnatter:
#### What's in this PR?

This PR adds the `@babel/plugin-proposal-nullish-coalescing-operator` plugin to our babel configuration to transpile the code of the `react-leaflet` package. Without this change, the build errors with the following message when installing the newest version of the `react-leaflet` package:

```
ERROR in /sulu-skeleton/vendor/sulu/sulu/node_modules/react-leaflet/esm/Pane.js 25:37
Module parse failed: Unexpected token (25:37)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
|   }
| 
>   const parentPaneName = props.pane ?? context.pane;
|   const parentPane = parentPaneName ? context.map.getPane(parentPaneName) : undefined;
|   const element = context.map.createPane(name, parentPane);
 @ /sulu-skeleton/vendor/sulu/sulu/node_modules/react-leaflet/esm/index.js 13:0-30 13:0-30
 @ /sulu-skeleton/vendor/sulu/sulu/src/Sulu/Bundle/LocationBundle/Resources/js/containers/Location/Location.js
 @ /sulu-skeleton/vendor/sulu/sulu/src/Sulu/Bundle/LocationBundle/Resources/js/containers/Location/index.js
 @ /sulu-skeleton/vendor/sulu/sulu/src/Sulu/Bundle/LocationBundle/Resources/js/containers/Form/fields/Location.js
 @ /sulu-skeleton/vendor/sulu/sulu/src/Sulu/Bundle/LocationBundle/Resources/js/containers/Form/index.js
 @ /sulu-skeleton/vendor/sulu/sulu/src/Sulu/Bundle/LocationBundle/Resources/js/index.js
 @ ./index.js
```

#### Why?

The reason for this problem is quite interesting; [`react-leaflet`](https://github.com/PaulLeCam/react-leaflet/releases/tag/v3.2.0) uses the [nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator) in the code of the newly released version 3.2.0. This operator is supported by most modern browsers, but unfortunately `webpack@4` does not support the operator (https://github.com/webpack/webpack/issues/10227, https://github.com/acornjs/acorn/pull/890). Because of this, webpack crashes while processing the code of the `react-leaflet` package.

T solve the problem, the changes in this PR configure babel to transpile the nullish coalescing operator before the files are processed with webpack.

### DIFF
--- a/babel.config.json
+++ b/babel.config.json
@@ -2,7 +2,8 @@
     "presets": ["@babel/preset-env", "@babel/preset-react"],
     "plugins": [
         ["@babel/plugin-proposal-decorators", {"legacy": true}],
-        "@babel/plugin-transform-flow-strip-types"
+        "@babel/plugin-transform-flow-strip-types",
+        "@babel/plugin-proposal-nullish-coalescing-operator"
     ],
     "assumptions": {
         "setPublicClassFields": true
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
     "devDependencies": {
         "@babel/core": "^7.5.5",
         "@babel/plugin-proposal-decorators": "^7.4.4",
+        "@babel/plugin-proposal-nullish-coalescing-operator": "^7.14.2",
         "@babel/plugin-transform-flow-strip-types": "^7.4.4",
         "@babel/preset-env": "^7.5.5",
         "@babel/preset-react": "^7.0.0",
@@ -120,7 +121,7 @@
             ]
         },
         "transformIgnorePatterns": [
-            "node_modules/(?!(@ckeditor|ckeditor5|lodash-es)/)"
+            "node_modules/(?!(@ckeditor|ckeditor5|lodash-es|@react-leaflet|react-leaflet)/)"
         ],
         "testPathIgnorePatterns": [
             "vendor/friendsofsymfony"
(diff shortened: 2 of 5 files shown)


babel.config.json
```diff
@@ -2,7 +2,8 @@
     "presets": ["@babel/preset-env", "@babel/preset-react"],
     "plugins": [
         ["@babel/plugin-proposal-decorators", {"legacy": true}],
-        "@babel/plugin-transform-flow-strip-types"
+        "@babel/plugin-transform-flow-strip-types",
+        "@babel/plugin-proposal-nullish-coalescing-operator"
```
niklasnatter, line 6: TODO: adjust this in the `babel.config.json` of the `sulu/skeleton` too

niklasnatter: we would not need to do this if the `react-leaflet` package releases a new version that does not use the nullish coalescing operator. i have created https://github.com/PaulLeCam/react-leaflet/issues/883 to suggest this and would propose to set this pull request on hold until that issue is either resolved or rejected 🙂

alexander-schranz: @nnatter Looks like we will not get any feedback from @PaulLeCam. Looks like we need to hack this into here or what you think?

niklasnatter: yes i think so too 😢 
will add a pull request for the skeleton

niklasnatter: Have created a PR here: https://github.com/sulu/skeleton/pull/128

PR #6061, opened 2021-05-19
Make Sulu runnable with Roadrunner

chirimoya:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT

#### What's in this PR?

Make Sulu runnable with Roadrunner.

#### Why?

Because is fast.

#### To Do

- [ ] More testing


alexander-schranz:
The `ResetInterface` should also be injected into following services and call the following methods:

 - [x] `DocumentManager::clear` this will clear an internal document manager array cache for the current process / request
 - [x] `Memoize` (this is an array cache which need to be reseted between requests) used by different twig extensions and as this caches contain request specific data it need to be reseted
 - [x] `SystemStore` maybe not required and can be cashed between requests as the `setSystem` would automatically reset it
 - [ ] ~`SystemCollectionManager` should maybe reset its `$this->systemCollections` array~

alexander-schranz (approved):

alexander-schranz (approved):

PR #6064, opened 2021-05-20
Include id in activity list response to prevent error when rendering rows

niklasnatter:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| License | MIT

#### What's in this PR?

This PR includes the `id` of the activity in the list response of the `ActivityController"`.

#### Why?

Our list view uses the id as key for rendering the rows of the list. If not id is returned from the server, the following error is logged to the browser console:

Screenshot 2021-05-20 at 15 02 25

### DIFF
--- a/src/Sulu/Bundle/ActivityBundle/UserInterface/Controller/ActivityController.php
+++ b/src/Sulu/Bundle/ActivityBundle/UserInterface/Controller/ActivityController.php
@@ -173,7 +173,7 @@ class ActivityController extends AbstractRestController implements ClassResource
                         ]
                     ),
                     function(string $key) use ($configurationFieldDescriptors) {
-                        return \array_key_exists($key, $configurationFieldDescriptors);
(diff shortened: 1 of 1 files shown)


src/Sulu/Bundle/ActivityBundle/UserInterface/Controller/ActivityController.php
```diff
@@ -173,7 +173,7 @@ function(array $activity) use ($translationLocale, $configurationFieldDescriptor
                         ]
                     ),
                     function(string $key) use ($configurationFieldDescriptors) {
-                        return \array_key_exists($key, $configurationFieldDescriptors);
+                        return $key === 'id' || \array_key_exists($key, $configurationFieldDescriptors);
```
alexander-schranz:
```suggestion
                        return 'id' === $key || \array_key_exists($key, $configurationFieldDescriptors);
```

alexander-schranz (approved):

PR #6068, opened 2021-05-21
Fix spacing at bottom of media overview when using table adapter

niklasnatter:


src/Sulu/Bundle/AdminBundle/Resources/js/components/InfiniteScroller/infiniteScroller.scss
```diff
@@ -2,7 +2,6 @@
 
 .indicator {
     margin-top: 30px;
-    height: 50px;
```
niklasnatter, line 5: this causes duplicated spacing in the media selection overlay and when the adapter is used in the default list view

PR #6069, opened 2021-05-25
Fix phpstan baseline and fix import/order eslint rule

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? | no 
| BC breaks? | no 
| Deprecations? | no  
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Fix phpstan baseline. And manual configure the `import/order` rule.

#### Why?

New doctrine/orm release.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -112,7 +112,18 @@
             "error",
             "prefer-double"
         ],
-        "import/order": "error",
+        "import/order": [
+            "error", {
+                "groups": [
+                    "builtin",
+                    "external",
+                    "parent",
+                    "sibling",
+                    "index",
+                    "type"
+                ]
+            }
+        ],
         "import/no-dynamic-require": "error",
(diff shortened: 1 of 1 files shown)


.eslintrc.json
```diff
@@ -112,7 +112,18 @@
...
-        "import/order": "error",
+        "import/order": [
+            "error", {
+                "groups": [
+                    "builtin",
+                    "external",
+                    "parent",
+                    "sibling",
+                    "index",
+                    "type"
```
alexander-schranz, line 123: Type was added: https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/order.md#groups-array so they are again at the end. In the last releases they did add it and then revert it. I did now set a specific rule so it will not longer change here and I think having `type` at the end will make it easier in future for us.

Prokyonn (approved):

PR #6071, opened 2021-05-25
Remove false x-default alternate url

### LINKED ISSUE #6027: X-Default handling in meta and sitemap
| Q | A
| --- | ---
| Sulu Version | 2.x 
| PHP Version | 
| DB Version | 
| Browser Version | 

# Actual Behavior

Currently the x-default doesn't work:

 - when the webspace configure url per locale (x-default is from language to language different)
 - when the default language does not exist the `x-default` should not be returned
 - to be discussed x-default should never be set as it is a international landingpage

# Expected Behavior

Correctly handling of x-default handling in meta and sitemap.

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes 
| New feature? | no 
| BC breaks? | no 
| Deprecations? | no  
| Fixed tickets | part of fixes #6027 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Remove x-default alternate url.

#### Why?

The `x-default` is for `[international pages](https://developers.google.com/search/blog/2013/04/x-default-hreflang-for-international-pages)` which means a page which you could select the language, so basically only one x-default url should exist, see also [google xdefault docs](https://developers.google.com/search/docs/advanced/crawling/localized-versions#xdefault) and [yandex xdefault docs](https://yandex.com/support/webmaster/yandex-indexing/locale-pages.html#auto).

If a website has international landingpage they should now override the `sitemap.xml` file to add it as an alternate link to the current webspace.

### DIFF
--- a/src/Sulu/Bundle/WebsiteBundle/Resources/views/Sitemap/sitemap.xml.twig
+++ b/src/Sulu/Bundle/WebsiteBundle/Resources/views/Sitemap/sitemap.xml.twig
@@ -7,35 +7,38 @@
         {% if domain in location %}
             <url>
                 {% block url %}
-                    <loc>{{ location }}</loc>
-                    {% if entry.lastmod %}
-                        <lastmod>{{ entry.lastmod|date('Y-m-d') }}</lastmod>
-                    {% endif %}
-                    {% if entry.changefreq %}
-                        <changefreq>{{ entry.changefreq }}</changefreq>
-                    {% endif %}
-                    {% if entry.priority %}
-                        <priority>{{ entry.priority }}</priority>
-                    {% endif %}
+                    {% block loc %}
+                        <loc>{{ location }}</loc>
+                    {% endblock %}
 
-                    {% set amount = 0 %}
-                    {% set defaultHref = null %}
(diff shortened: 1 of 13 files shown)


src/Sulu/Bundle/WebsiteBundle/Resources/views/Sitemap/sitemap.xml.twig
```diff
@@ -7,35 +7,38 @@
...
-                    {% if entry.lastmod %}
-                        <lastmod>{{ entry.lastmod|date('Y-m-d') }}</lastmod>
-                    {% endif %}
-                    {% if entry.changefreq %}
-                        <changefreq>{{ entry.changefreq }}</changefreq>
-                    {% endif %}
-                    {% if entry.priority %}
-                        <priority>{{ entry.priority }}</priority>
-                    {% endif %}
+                    {% block loc %}
```
alexander-schranz, line 10: I just added more blocks to override the template here and removed the `x-default` behaviour as it is incorrect.

Prokyonn (approved):

PR #6072, opened 2021-05-25
Avoid false alternate link to homepage

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? |  yes
| New feature? | no 
| BC breaks? | no 
| Deprecations? | no  
| Fixed tickets | fixes part of #6027
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Remove fallback to the homepage as alternate links.

#### Why?

Avoid false alternate link to homepage in the sitemap. We can currently not change this on the template itself, this will be fixed in a seperate PR on the 2.2 branch where this is possible over the localization PR.

### DIFF
--- a/src/Sulu/Bundle/PageBundle/Sitemap/PagesSitemapProvider.php
+++ b/src/Sulu/Bundle/PageBundle/Sitemap/PagesSitemapProvider.php
@@ -135,6 +135,10 @@ class PagesSitemapProvider extends AbstractSitemapProvider
         );
 
         foreach ($contentPage->getUrls() as $urlLocale => $href) {
+            if (null === $href) {
+                continue;
+            }
+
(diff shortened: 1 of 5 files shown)


src/Sulu/Bundle/PageBundle/Sitemap/PagesSitemapProvider.php
```diff
@@ -135,6 +135,10 @@ private function generateSitemapUrl(
         );
 
         foreach ($contentPage->getUrls() as $urlLocale => $href) {
+            if (null === $href) {
```
alexander-schranz, line 138: This is the required change. The rest is more typo and for better debugging.

Prokyonn (approved):

Issue #5927, opened 2021-04-08
Save button is not enabled after changing the template of a page


michaelzangerle:
| Q | A
| --- | ---
| Sulu Version | 2.2.6
| PHP Version | 7.4.16
| DB Version | mySQL 5.7
| Browser Version | Version 89.0.4389.114 (Official Build) (64-bit)

# Actual Behavior

The safe button for a page is not clickable.

# Expected Behavior

The safe button for a page should be clickable if all required fields are present and have a proper value.

# Steps to Reproduce

I was able to reproduce it also on sulu.rocks

- Go to webspaces in https://sulu.rocks/admin/
- Create a new page
- Add a title
- Change the template to e.g. `Homepage`
- Try to click the safe button (it's deactivated)
- As a workaround you can make a change to some field or just click inside the resource locator field and the save button will be enabled

# Possible Solutions

niklasnatter:
Thanks for reporting! 

It looks like the reason for this is a bit complicated 🙈  
The button is disabled because the form is marked as not dirty in this callback:
https://github.com/sulu/sulu/blob/5c7a15c250c01ee4337f1fd1c44a34b2e9035147/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/Form.js#L58-L63

The callback is executed because the `Form` container is rerendered because the value of `store.loading` changes:
https://github.com/sulu/sulu/blob/5c7a15c250c01ee4337f1fd1c44a34b2e9035147/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/Form.js#L163-L165

The value of `store.loading` changes because these lines are executed when the type of the form is changed:
https://github.com/sulu/sulu/blob/5c7a15c250c01ee4337f1fd1c44a34b2e9035147/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js#L82-L86

alexander-schranz:
This is fixed by #6115

PR #6077, opened 2021-05-26
Merge local data with remote data when switching template of page

### LINKED ISSUE #5983: Different templates with different blocks but same block names
| Q | A
| --- | ---
| Sulu Version | current master

First of all I am not sure this is a bug or just a missing feature or something else. As I ran into it because of an error I chose the bug template to create this issue. I talked to @wachterjohannes and we decided to create an issue to discuss it here because we uncovered some more scenarios as well.

**Situation** 
- we have two templates T1 and T2
- we have blocks `headline`, `products`, `products-automatic`, `teaser`

# Actual Behavior
While switching from T1 with a configured `headline` to T2 I ran into the following error `The block type "headline" has not been registered. Known block types are: [products, products-automatic, teaser]`

# Expected Behavior
I was expecting / we were discussing the following scenarios:
- if `headline` is only registered in T1 but not in T2 it shouldn't have any effect on the new template and get removed
- if `headline` is registered in both templates and has the same definition it should be possible to kept the information
- if `headline` is registered in both templates and they have at least a partially compatible definition as much as possible information should be kept
- if `headline` is registered in both templates but the have incompatible definitions no information should be kept

Those were the things we discussed very briefly and it seems like some of them are already implemented but not all of them. Is this something you want to do? If not what other solution would you suggest to prevent this error from happening? Prefixing block names in templates?

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes https://github.com/sulu/sulu/issues/5983
| Related issues/PRs | builds upon https://github.com/sulu/sulu/pull/6076
| License | MIT

#### What's in this PR?

See https://github.com/sulu/sulu/issues/5983

* on template change, obsolete data will be removed from the `resourceStore`
* if data uses block-types which are not available in the new template, these blocks are set to the default block-type
* if the user switches back to the origin template, the data will be fetched from the backend and merged with the current data of the frontend

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js
@@ -1,5 +1,5 @@
 // @flow
-import {action, autorun, computed, get, isArrayLike, observable, when} from 'mobx';
+import {action, autorun, computed, get, isArrayLike, observable, toJS, when} from 'mobx';
 import jsonpointer from 'json-pointer';
 import {createAjv} from '../../../utils/Ajv';
 import ResourceStore from '../../../stores/ResourceStore';
@@ -16,14 +16,14 @@ const ajv = createAjv();
 export default class ResourceFormStore extends AbstractFormStore implements FormStoreInterface {
     resourceStore: ResourceStore;
     formKey: string;
-    options: {[string]: any};
+    options: { [string]: any };
     @observable type: string;
-    @observable types: {[key: string]: SchemaType} = {};
+    @observable types: { [key: string]: SchemaType } = {};
     @observable schemaLoading: boolean = true;
     @observable typesLoading: boolean = true;
     schemaDisposer: ?() => void;
     typeDisposer: ?() => void;
-    metadataOptions: ?{[string]: any};
+    metadataOptions: ?{ [string]: any };
 
     constructor(resourceStore: ResourceStore, formKey: string, options: Object = {}, metadataOptions: ?Object) {
         super();
@@ -91,151 +91,108 @@ export default class ResourceFormStore extends AbstractFormStore implements Form
         this.validator = jsonSchema ? ajv.compile(jsonSchema) : undefined;
         this.pathsByTag = {};
 
-        return this.resourceStore.requestData().then((data: Object) => {
-            this.removeObsoleteData(this.schema, schema, this.data);
-            const result = this.mergeData(this.schema, schema, this.data, data);
-            this.setMultiple(result);
-
+        return this.requestRemoteDataAndMerge(this.schema, schema).then(action(() => {
             this.schema = schema;
             this.addMissingSchemaProperties();
             this.validate();
             this.setSchemaLoading(false);
-        });
+        }));
+    };
+
+    requestRemoteDataAndMerge = (localSchema: Schema, remoteSchema: Schema) => {
+        if (localSchema) {
+            return this.resourceStore.requestRemoteData({template: this.type}).then((data: Object) => {
+                const result = this.mergeData(localSchema, remoteSchema, this.data, data);
+                this.setMultiple(result);
+            });
+        }
+        return Promise.resolve();
     };
 
     mergeData(
-        oldSchema: Schema,
-        newSchema: Schema,
-        oldData: Object,
-        newData: Object,
-        parentPath: string[] = []
+        localSchema: Schema,
+        remoteSchema: Schema,
+        localData: Object,
+        remoteData: Object
     ) {
         let result = {};
-        if (!oldSchema || !newSchema) {
+        if (!localSchema || !remoteSchema) {
             return result;
         }
 
-        for (const key in newSchema) {
-            const {items: newItems, type: newType, types: newTypes} = newSchema[key];
-            const {items: oldItems, type: oldType, types: oldTypes} = oldSchema[key] || {};
+        for (const name in remoteSchema) {
+            const {items: remoteItems, type: remoteType, types: remoteTypes} = remoteSchema[name];
+            const {items: localItems, type: localType, types: localTypes} = localSchema[name] || {};
 
-            if (newType === SECTION_TYPE && newItems &&
-                oldType === SECTION_TYPE && oldItems) {
+            if (remoteType === SECTION_TYPE && remoteItems &&
+                localType === SECTION_TYPE && localItems) {
                 result = this.mergeData(
-                    oldItems,
-                    newItems,
-                    oldData,
-                    newData,
-                    parentPath
+                    localItems,
+                    remoteItems,
+                    localData,
+                    remoteData
                 );
                 continue;
             }
-            if (newTypes && oldTypes
-                && Object.keys(newTypes).length > 0 && Object.keys(oldTypes).length > 0
-                && oldData[key] && newData[key]
-                && isArrayLike(oldData[key]) && isArrayLike(newData[key])
+            if (remoteTypes && localTypes
+                && Object.keys(remoteTypes).length > 0 && Object.keys(localTypes).length > 0
+                && localData[name] && remoteData[name]
+                && isArrayLike(localData[name]) && isArrayLike(remoteData[name])
             ) {
-                for (const childKey of newData[key].keys()) {
-                    const newChildData = newData[key][childKey];
-                    const oldChildData = oldData[key].length > childKey ? oldData[key][childKey] || {} : {};
+                for (const key of remoteData[name].keys()) {
+                    const remoteChildData = remoteData[name][key];
+                    const localChildData = toJS(localData[name].length > key ? localData[name][key] || {} : {});
 
-                    const oldChildSchema = oldTypes[oldChildData.type]?.form;
-                    const newChildSchema = newTypes[newChildData.type].form;
+                    const localChildSchema = localTypes[localChildData.type]?.form;
+                    const remoteChildSchema = remoteTypes[remoteChildData.type].form;
 
-                    if (Object.keys(oldChildData).length === 0 && !oldChildSchema) {
-                        // set originData
-                        if (!result[key]) {
-                            result[key] = [];
+                    if (!(localChildData.type in remoteTypes)) {
+                        //set default type
+                        localChildData.type = remoteSchema[name].defaultType;
+                    }
+
+                    const resultChildData = this.mergeData(
+                        localChildSchema,
+                        remoteChildSchema,
+                        localChildData,
+                        remoteChildData
+                    );
+
+                    if (!result[name]) {
+                        result[name] = [];
+                    }
+
+                    if (Object.keys(resultChildData).length === 0) {
+                        if (localChildData.type === remoteChildData.type){
+                            result[name][key] = localChildData;
+                        } else {
+                            result[name][key] = remoteChildData;
                         }
-                        result[key][childKey] = newChildData;
-
-                        continue;
+                    } else {
+                        result[name][key] = resultChildData;
                     }
-
-                    if (!result[key]){
-                        result[key] = [];
-                    }
-                    if (!(oldChildData.type in newTypes)) {
-                        result[key][childKey] = this.mergeData(
-                            oldChildSchema,
-                            newChildSchema,
-                            oldChildData,
-                            newChildData,
-                            parentPath.concat([key, childKey])
-                        );
-
-                        continue;
-                    }
-                    result[key][childKey] = oldChildData;
                 }
 
                 continue;
             }
 
-            if (!oldData[key] || newSchema[key].type !== oldSchema[key]?.type) {
-                // set originData
-                result[key] = newData[key];
+            if (!localData[name]) {
+                result[name] = remoteData[name];
+                if (remoteData.type) {
+                    result.type = remoteData.type;
+                }
+
+                if (remoteData.settings) {
+                    result.settings = localData?.settings || remoteData.settings;
+                }
+            }
+
+            if (localData[name] && remoteSchema[name].type !== localSchema[name]?.type) {
+                result[name] = undefined;
             }
         }
 
         return result;
-    }
-
-    removeObsoleteData(oldSchema: Schema, newSchema: Schema, data: Object, parentPath: string[] = []) {
-        if (!oldSchema || !newSchema) {
-            return;
-        }
-
-        for (const key in oldSchema) {
-            const {items: newItems, type: newType, types: newTypes} = newSchema[key] || {};
-            const {items: oldItems, type: oldType, types: oldTypes} = oldSchema[key];
-
-            if (newType === SECTION_TYPE && newItems &&
-                oldType === SECTION_TYPE && oldItems) {
-                this.removeObsoleteData(
-                    oldItems,
-                    newItems,
-                    data,
-                    parentPath
-                );
-                continue;
-            }
-
-            if (newTypes && oldTypes
-                && Object.keys(newTypes).length > 0 && Object.keys(oldTypes).length > 0
-                && data[key] && isArrayLike(data[key])
-            ) {
-                for (const childKey of data[key].keys()) {
-                    const childData = data[key][childKey];
-                    const oldChildSchema = oldTypes[childData.type]?.form;
-                    let newChildSchema = newTypes[childData.type]?.form;
-
-                    if (!newChildSchema) {
-                        const defaultType = newSchema[key]?.defaultType;
-                        if (defaultType) {
-                            newChildSchema = newTypes[defaultType].form;
-                            //set default type
-                            const path = parentPath.concat([key, childKey]).join('/') + '/type';
-                            this.change(path, defaultType);
-                        }
-                    }
-
-                    this.removeObsoleteData(
-                        oldChildSchema,
-                        newChildSchema,
-                        childData,
-                        parentPath.concat([key, childKey])
-                    );
-                }
-
-                continue;
-            }
-
-            if (!newSchema[key] || newSchema[key]['type'] !== oldSchema[key]['type']) {
-                // remove obsolete data from previous schema
-                this.remove(parentPath.concat(key).join('/'));
-            }
-        }
     }
 
     @computed get hasTypes(): boolean {
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js
@@ -88,25 +88,27 @@ export default class ResourceFormStore extends AbstractFormStore implements Form
         });
     };
 
-    @action handleSchemaResponse = ([schema, jsonSchema]: [Schema, Object]) => {
+    handleSchemaResponse = ([schema, jsonSchema]: [Schema, Object]) => {
         this.validator = jsonSchema ? ajv.compile(jsonSchema) : undefined;
         this.pathsByTag = {};
 
-        if (this.schema && this.type === this.data['originTemplate']){
-            this.loadAndMergeServerData(this.schema, schema);
-        }
-        this.removeObsoleteData(this.schema, schema, this.data);
-
-        this.schema = schema;
-        this.addMissingSchemaProperties();
-        this.setSchemaLoading(false);
+        this.loadAndMergeOriginData(this.schema, schema).then(action(()=> {
+            this.removeObsoleteData(this.schema, schema, this.data);
+            this.schema = schema;
+            this.addMissingSchemaProperties();
+            this.setSchemaLoading(false);
+        }));
     };
 
-    loadAndMergeServerData(oldSchema: Schema, newSchema: Schema) {
-        this.resourceStore.loadData().then((data: Object) => {
-            this.mergeData(oldSchema, newSchema, this.data, data);
-            this.validate();
-        });
+    loadAndMergeOriginData(oldSchema: Schema, newSchema: Schema) {
+        if (oldSchema && this.type === this.data['originTemplate']){
+            return this.resourceStore.loadData().then((data: Object) => {
+                this.mergeData(oldSchema, newSchema, this.data, data);
+                this.validate();
+            });
+        }
+
+        return Promise.resolve();
     }
 
     mergeData(oldSchema: Schema, newSchema: Schema, oldData: Object, newData: Object, parentPath: string[] = []) {
@@ -139,7 +141,7 @@ export default class ResourceFormStore extends AbstractFormStore implements Form
                     const newChildData = newData[key][childKey];
                     const oldChildData = oldData[key][childKey] || {};
 
-                    const oldChildSchema = oldTypes[oldChildData.type].form;
+                    const oldChildSchema = oldTypes[oldChildData.type]?.form;
                     const newChildSchema = newTypes[newChildData.type].form;
 
                     this.mergeData(
@@ -188,7 +190,7 @@ export default class ResourceFormStore extends AbstractFormStore implements Form
             ) {
                 for (const childKey of data[key].keys()) {
                     const childData = data[key][childKey];
-                    const oldChildSchema = oldTypes[childData.type].form;
+                    const oldChildSchema = oldTypes[childData.type]?.form;
                     let newChildSchema = newTypes[childData.type]?.form;
 
                     if (!newChildSchema) {
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js
@@ -1,5 +1,5 @@
 // @flow
-import {action, autorun, computed, get, isObservableArray, observable, toJS, when} from 'mobx';
+import {action, autorun, computed, get, isArrayLike, observable, toJS, when} from 'mobx';
 import jsonpointer from 'json-pointer';
 import {createAjv} from '../../../utils/Ajv';
 import ResourceStore from '../../../stores/ResourceStore';
@@ -101,8 +101,9 @@ export default class ResourceFormStore extends AbstractFormStore implements Form
     };
 
     loadAndMergeOriginData(oldSchema: Schema, newSchema: Schema) {
-        if (oldSchema && toJS(this.type) === this.data['originTemplate']){
-            return this.resourceStore.loadData().then((data: Object) => {
+        // origin data must be loaded only when switching back to the origin template
+        if (oldSchema && toJS(this.type) === this.data['originTemplate']) {
+            return this.resourceStore.requestData().then((data: Object) => {
                 this.mergeData(oldSchema, newSchema, this.data, data);
                 this.validate();
             });
@@ -118,7 +119,7 @@ export default class ResourceFormStore extends AbstractFormStore implements Form
 
         for (const key in newSchema) {
             const {items: newItems, type: newType, types: newTypes} = newSchema[key];
-            const {items: oldItems, type: oldType, types: oldTypes} = oldSchema ? oldSchema[key] || {} : {};
+            const {items: oldItems, type: oldType, types: oldTypes} = oldSchema[key] || {};
 
             if (newType === SECTION_TYPE && newItems &&
                 oldType === SECTION_TYPE && oldItems) {
@@ -134,8 +135,7 @@ export default class ResourceFormStore extends AbstractFormStore implements Form
             if (newTypes && oldTypes
                 && Object.keys(newTypes).length > 0 && Object.keys(oldTypes).length > 0
                 && oldData[key] && newData[key]
-                && (Array.isArray(oldData[key]) || isObservableArray(oldData[key]))
-                && (Array.isArray(newData[key]) || isObservableArray(newData[key]))
+                && isArrayLike(oldData[key]) && isArrayLike(newData[key])
             ) {
                 for (const childKey of newData[key].keys()) {
                     const newChildData = newData[key][childKey];
@@ -191,8 +191,7 @@ export default class ResourceFormStore extends AbstractFormStore implements Form
 
             if (newTypes && oldTypes
                 && Object.keys(newTypes).length > 0 && Object.keys(oldTypes).length > 0
-                && data[key]
-                && (Array.isArray(data[key]) || isObservableArray(data[key]))
+                && data[key] && isArrayLike(data[key])
             ) {
                 for (const childKey of data[key].keys()) {
                     const childData = data[key][childKey];
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js
@@ -38,17 +38,6 @@ function mergeData(
             types: localTypes,
         } = localSchema[name] || {};
 
-        if (remoteType === SECTION_TYPE && remoteItems &&
-            localType === SECTION_TYPE && localItems) {
-            result = mergeData(
-                localItems,
-                remoteItems,
-                localData,
-                remoteData
-            );
-            continue;
-        }
-
         if (remoteType === SECTION_TYPE && remoteItems) {
             result = mergeData(
                 localSchema,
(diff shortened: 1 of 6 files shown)


src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/Preview.js
```diff
@@ -137,9 +137,9 @@ class Preview extends React.Component<Props> {
         );
 
         this.typeDisposer = reaction(
```
niklasnatter:
```suggestion
        this.schemaDisposer = reaction(
```

src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/Preview.js
```diff
@@ -137,9 +137,9 @@ class Preview extends React.Component<Props> {
...
-            () => toJS(formStore.type),
-            (type) => {
-                previewStore.updateContext(type, formStore.data).then(this.setContent);
+            () => toJS(formStore.schema),
```
niklasnatter, line 140: i guess this needs to be changed to update the context only after the new schema is loaded, right?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js
```diff
@@ -87,15 +88,145 @@ export default class ResourceFormStore extends AbstractFormStore implements Form
...
+        this.loadAndMergeOriginData(this.schema, schema).then(action(()=> {
+            this.removeObsoleteData(this.schema, schema, this.data);
+            this.schema = schema;
+            this.addMissingSchemaProperties();
+            this.setSchemaLoading(false);
+        }));
     };
 
+    loadAndMergeOriginData(oldSchema: Schema, newSchema: Schema) {
+        if (oldSchema && toJS(this.type) === this.data['originTemplate']){
```
niklasnatter: can we add a comment that explains the `toJS(this.type) === this.data['originTemplate']` condition?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js
```diff
@@ -87,15 +88,145 @@ export default class ResourceFormStore extends AbstractFormStore implements Form
...
+    }
+
+    mergeData(oldSchema: Schema, newSchema: Schema, oldData: Object, newData: Object, parentPath: string[] = []) {
+        if (!oldSchema || !newSchema) {
+            return;
+        }
+
+        for (const key in newSchema) {
+            const {items: newItems, type: newType, types: newTypes} = newSchema[key];
+            const {items: oldItems, type: oldType, types: oldTypes} = oldSchema ? oldSchema[key] || {} : {};
```
niklasnatter: think `oldSchema` is always defined at this point

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js
```diff
@@ -87,15 +88,145 @@ export default class ResourceFormStore extends AbstractFormStore implements Form
...
+                    oldData,
+                    newData,
+                    parentPath
+                );
+                continue;
+            }
+            if (newTypes && oldTypes
+                && Object.keys(newTypes).length > 0 && Object.keys(oldTypes).length > 0
+                && oldData[key] && newData[key]
+                && (Array.isArray(oldData[key]) || isObservableArray(oldData[key]))
```
niklasnatter: you can use `isArrayLike` here

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js
```diff
@@ -87,15 +88,145 @@ export default class ResourceFormStore extends AbstractFormStore implements Form
...
-    @action handleSchemaResponse = ([schema, jsonSchema]: [Schema, Object]) => {
+    handleSchemaResponse = ([schema, jsonSchema]: [Schema, Object]) => {
         this.validator = jsonSchema ? ajv.compile(jsonSchema) : undefined;
         this.pathsByTag = {};
 
-        this.schema = schema;
-        this.addMissingSchemaProperties();
-        this.setSchemaLoading(false);
+        this.loadAndMergeOriginData(this.schema, schema).then(action(()=> {
+            this.removeObsoleteData(this.schema, schema, this.data);
```
nnatter, line 96: do we need to do this in the frontend? shouldnt the server do it on save 🤔

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js
```diff
@@ -87,15 +88,145 @@ export default class ResourceFormStore extends AbstractFormStore implements Form
...
+            this.schema = schema;
+            this.addMissingSchemaProperties();
+            this.setSchemaLoading(false);
+        }));
     };
 
+    loadAndMergeOriginData(oldSchema: Schema, newSchema: Schema) {
+        if (oldSchema && toJS(this.type) === this.data['originTemplate']){
+            return this.resourceStore.loadData().then((data: Object) => {
+                this.mergeData(oldSchema, newSchema, this.data, data);
```
niklasnatter: i think i would expect this method to return a new data array and not change the data of the store

niklasnatter: if the method returns a new array that only contains the keys of the new schema, we might also not need the `removeObsoleteData`?

src/Sulu/Bundle/AdminBundle/Resources/js/stores/ResourceStore/ResourceStore.js
```diff
@@ -98,6 +87,27 @@ export default class ResourceStore {
             }));
     };
 
+    loadData = () => {
```
niklasnatter: maybe `requestData`?

src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/Preview.js
Prokyonn: yes otherwise the preview is reloaded to early

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js
```diff
@@ -87,15 +88,163 @@ export default class ResourceFormStore extends AbstractFormStore implements Form
...
+                        parentPath.concat([key, childKey])
+                    );
+                }
+
+                continue;
+            }
+
+            if (!newSchema[key] || newSchema[key]['type'] !== oldSchema[key]['type']) {
+                // remove obsolete data from previous schema
+                this.remove(parentPath.concat(key).join('/'));
```
Prokyonn: I still think we need this in the frontend, otherwise the admin crashes in the following scenario:

We have template `A` with property `title` as `text_line` and template `B` with property `title` as `single_media_selection` .

If template `A` is selected with a value in `title` and the user switches to template `B` the admin crashes. For blocks we can avoid this with a switch to the `default-block-type` but for the other properties I don't see an easy way to replace it other than removing the property completely.

If we would switch back to the `originTemplate` this would also be no problem, because we can just use the `originData` from the server, but this is only possible for the `originTemplate`.

src/Sulu/Bundle/AdminBundle/Resources/js/stores/ResourceStore/ResourceStore.js
```diff
@@ -279,6 +294,31 @@ export default class ResourceStore {
         );
     }
 
+    @action setMultipleRecursive(data: Object, parentPath: string = '') {
```
Prokyonn: This is necessary, because `setMultiple` sets only the root properties and thus `blocks` is replaced by `blocks`. 

But to set the `diff` from the `calculateDifference` method, we have to recursively navigate through the individual blocks and replace only single blocks instead of all of them.

I implemented a new method for this, because the performance of the simple `setMultiple` is obviously better and the recursive method is only required during the data merge of this pr.

Prokyonn:
after this is merged, we should also prioritise https://github.com/sulu/sulu/issues/5927 

then we really change the `data`, but the `Save` button is still disabled

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js
```diff
@@ -10,6 +10,7 @@ import type {IObservableValue} from 'mobx/lib/mobx';
 
 // TODO do not hardcode "template", use some kind of metadata instead
 const TYPE = 'template';
+const SECTION_TYPE = 'section';
```
niklasnatter: think this constant is already defined in the `AbstractFormStore`

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js
niklasnatter: you are right - makes sense 🙂

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js
```diff
@@ -87,15 +88,163 @@ export default class ResourceFormStore extends AbstractFormStore implements Form
...
+        this.removeObsoleteData(this.schema, schema, this.data);
+        this.loadAndMergeOriginData(this.schema, schema).then(action(()=> {
+            this.schema = schema;
+            this.addMissingSchemaProperties();
+            this.setSchemaLoading(false);
+        }));
     };
 
+    loadAndMergeOriginData(oldSchema: Schema, newSchema: Schema) {
+        // origin data must be loaded only when switching back to the origin template
```
niklasnatter: i think im missing something here 🙈  
what is the `origin template` exactly? why dont we need to merge the data for other templates?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js
```diff
@@ -87,15 +88,163 @@ export default class ResourceFormStore extends AbstractFormStore implements Form
...
 
-    @action handleSchemaResponse = ([schema, jsonSchema]: [Schema, Object]) => {
+    handleSchemaResponse = ([schema, jsonSchema]: [Schema, Object]) => {
         this.validator = jsonSchema ? ajv.compile(jsonSchema) : undefined;
         this.pathsByTag = {};
 
-        this.schema = schema;
-        this.addMissingSchemaProperties();
-        this.setSchemaLoading(false);
+        this.removeObsoleteData(this.schema, schema, this.data);
```
niklasnatter: do we need the old schema in this method? 
maybe we could implement it similar to the `addMissingSchemaProperties` method and call it `removeUnknownSchemaProperties`?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js
```diff
@@ -87,15 +88,163 @@ export default class ResourceFormStore extends AbstractFormStore implements Form
...
+            this.addMissingSchemaProperties();
+            this.setSchemaLoading(false);
+        }));
     };
 
+    loadAndMergeOriginData(oldSchema: Schema, newSchema: Schema) {
+        // origin data must be loaded only when switching back to the origin template
+        if (oldSchema && toJS(this.type) === this.data['originTemplate']) {
+            return this.resourceStore.requestData().then((data: Object) => {
+                const result = this.calculateDifference(oldSchema, newSchema, this.data, data);
```
niklasnatter: do we need to calculate the difference here? could we return the merged data instead of just the difference? 🤔 
that way we would not need the `setMultipleRecursive` method

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js
```diff
@@ -87,15 +88,163 @@ export default class ResourceFormStore extends AbstractFormStore implements Form
...
-    @action handleSchemaResponse = ([schema, jsonSchema]: [Schema, Object]) => {
+    handleSchemaResponse = ([schema, jsonSchema]: [Schema, Object]) => {
         this.validator = jsonSchema ? ajv.compile(jsonSchema) : undefined;
         this.pathsByTag = {};
 
-        this.schema = schema;
-        this.addMissingSchemaProperties();
-        this.setSchemaLoading(false);
+        this.removeObsoleteData(this.schema, schema, this.data);
+        this.loadAndMergeOriginData(this.schema, schema).then(action(()=> {
```
niklasnatter: i think this would be a bit easier to understand if we first load the data and then manipulate the data. but maybe i am missing some reason why that not possible. i would expect the flow to be something like this:

```js
requestData().then((data) => {
    const mergedData = this.mergeData(oldSchema, newSchema, this.data, data);
    this.setMultiple(mergedData);

    this.schema = schema;
    this.removeUnknownSchemaProperties();
    this.addMissingSchemaProperties();
    this.validate();
    this.setSchemaLoading(false);
)
```

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js
```diff
@@ -87,15 +88,163 @@ export default class ResourceFormStore extends AbstractFormStore implements Form
...
+                for (const childKey of data[key].keys()) {
+                    const childData = data[key][childKey];
+                    const oldChildSchema = oldTypes[childData.type]?.form;
+                    let newChildSchema = newTypes[childData.type]?.form;
+
+                    if (!newChildSchema) {
+                        const defaultType = newSchema[key]?.defaultType;
+                        if (defaultType) {
+                            newChildSchema = newTypes[defaultType].form;
+                            //set default type
```
niklasnatter: would expect that this method only removes things. shouldnt this be set in the `calculateDifference` (`mergeData`) method? 🙈

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js
niklasnatter: think i would also move the method to the `AbstractFormStore` if the comment above is possible

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js
Prokyonn: yes that is also possible, seems like I misunderstood you the last time, I thought you specifically wanted to calculate only the difference 😅

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js
Prokyonn: the `originTemplate` is the template which is saved in the backend. We only need to merge the data, when the user goes back to the `originTemplate` because then we can load the `originData` for this template.

If the user switches between other templates, we just remove the invalid properties for the new schema, but we have no other `data` to make a merge.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js
Prokyonn: For this flow we would have to load the server data on each template switch, not only when going back to the `originTemplate`. I don't see any benefits when loading the `originData` on each template switch. Most likely the data is not compatible with the schema anyway.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js
Prokyonn: I'm also not happy with the method name, but if we split setting the `defaultValue` for unknown blocks and removing the obsolete properties into two methods, we would also have to iterate twice through the `oldSchema`.

The `calculateDifference` iterates over the `newSchema` and the `removeObsoleteData` iterates over the `oldSchema` to find the invalid data entries.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js
```diff
@@ -87,15 +87,157 @@ export default class ResourceFormStore extends AbstractFormStore implements Form
...
+                for (const childKey of newData[key].keys()) {
+                    const newChildData = newData[key][childKey];
+                    const oldChildData = oldData[key].length > childKey ? oldData[key][childKey] || {} : {};
+
+                    const oldChildSchema = oldTypes[oldChildData.type]?.form;
+                    const newChildSchema = newTypes[newChildData.type].form;
+
+                    if (Object.keys(oldChildData).length === 0 && !oldChildSchema) {
+                        // set originData
+                        if (!result[key]) {
```
niklasnatter: think we could do this as soon as we know that its a block. then we dont need to do it two times 🙂

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js
```diff
@@ -87,15 +87,157 @@ export default class ResourceFormStore extends AbstractFormStore implements Form
...
-    @action handleSchemaResponse = ([schema, jsonSchema]: [Schema, Object]) => {
+    handleSchemaResponse = ([schema, jsonSchema]: [Schema, Object]) => {
         this.validator = jsonSchema ? ajv.compile(jsonSchema) : undefined;
         this.pathsByTag = {};
 
-        this.schema = schema;
-        this.addMissingSchemaProperties();
-        this.setSchemaLoading(false);
+        return this.resourceStore.requestData().then((data: Object) => {
+            this.removeObsoleteData(this.schema, schema, this.data);
```
niklasnatter: after talking about this again in person, i think we dont need to do this 🙈

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js
niklasnatter: we talked about this again and figured that this is not a problem anymore because we always load the remote data when switching the template now

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js
```diff
@@ -87,15 +87,157 @@ export default class ResourceFormStore extends AbstractFormStore implements Form
...
+            this.addMissingSchemaProperties();
+            this.validate();
+            this.setSchemaLoading(false);
+        });
     };
 
+    mergeData(
+        oldSchema: Schema,
+        newSchema: Schema,
+        oldData: Object,
```
niklasnatter: `localData`

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js
```diff
@@ -87,15 +87,157 @@ export default class ResourceFormStore extends AbstractFormStore implements Form
...
+        newData: Object,
```
niklasnatter: `remoteData`

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js
```diff
@@ -87,15 +87,157 @@ export default class ResourceFormStore extends AbstractFormStore implements Form
...
+
+            this.schema = schema;
+            this.addMissingSchemaProperties();
+            this.validate();
+            this.setSchemaLoading(false);
+        });
     };
 
+    mergeData(
+        oldSchema: Schema,
```
niklasnatter: `localSchema`

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js
```diff
@@ -87,15 +87,157 @@ export default class ResourceFormStore extends AbstractFormStore implements Form
...
+        newSchema: Schema,
```
niklasnatter: `remoteSchema`

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js
```diff
@@ -87,15 +87,157 @@ export default class ResourceFormStore extends AbstractFormStore implements Form
...
+        let result = {};
+        if (!oldSchema || !newSchema) {
+            return result;
+        }
+
+        for (const key in newSchema) {
+            const {items: newItems, type: newType, types: newTypes} = newSchema[key];
+            const {items: oldItems, type: oldType, types: oldTypes} = oldSchema[key] || {};
+
+            if (newType === SECTION_TYPE && newItems &&
```
niklasnatter: i think we need to splite this into two conditions; `newType === SECTION_TYPE && newItems` and a separate `oldType === SECTION_TYPE && oldItems`

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js
```diff
@@ -87,15 +87,157 @@ export default class ResourceFormStore extends AbstractFormStore implements Form
...
+        oldData: Object,
+        newData: Object,
+        parentPath: string[] = []
+    ) {
+        let result = {};
+        if (!oldSchema || !newSchema) {
+            return result;
+        }
+
+        for (const key in newSchema) {
```
niklasnatter: think i would call this `name` to make it different to the `key` used for blocks below

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js
```diff
@@ -87,15 +87,157 @@ export default class ResourceFormStore extends AbstractFormStore implements Form
...
+                        // set originData
+                        if (!result[key]) {
+                            result[key] = [];
+                        }
+                        result[key][childKey] = newChildData;
+
+                        continue;
+                    }
+
+                    if (!result[key]){
```
niklasnatter: we dont need to do this here if we move it up

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js
```diff
@@ -87,15 +87,157 @@ export default class ResourceFormStore extends AbstractFormStore implements Form
...
+                        result[key] = [];
+                    }
+                    if (!(oldChildData.type in newTypes)) {
```
niklasnatter: i think we always need to merge the data, regardless of the type

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js
```diff
@@ -87,15 +87,157 @@ export default class ResourceFormStore extends AbstractFormStore implements Form
...
+                        result[key][childKey] = this.mergeData(
+                            oldChildSchema,
+                            newChildSchema,
+                            oldChildData,
+                            newChildData,
+                            parentPath.concat([key, childKey])
+                        );
+
+                        continue;
```
niklasnatter: think we could set the type to the default type here if the local schema does not contain the type of the result

src/Sulu/Bundle/AdminBundle/Resources/js/stores/ResourceStore/ResourceStore.js
```diff
@@ -255,6 +264,11 @@ export default class ResourceStore {
             }));
     }
 
+    @action remove(path: string) {
```
niklasnatter: think this is not used anymore

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js
```diff
@@ -141,6 +240,10 @@ export default class ResourceFormStore extends AbstractFormStore implements Form
             });
     }
 
+    remove(name: string) {
```
niklasnatter: think this is not used anymore

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js
```diff
@@ -87,15 +87,114 @@ export default class ResourceFormStore extends AbstractFormStore implements Form
...
-        this.setSchemaLoading(false);
+        return this.requestRemoteDataAndMerge(this.schema, schema).then(action(() => {
+            this.schema = schema;
+            this.addMissingSchemaProperties();
+            this.validate();
+            this.setSchemaLoading(false);
+        }));
     };
 
+    requestRemoteDataAndMerge = (localSchema: Schema, remoteSchema: Schema) => {
```
niklasnatter: `loadAndMergeRemoteData`? 😇

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js
```diff
@@ -87,15 +87,114 @@ export default class ResourceFormStore extends AbstractFormStore implements Form
...
+        if (localSchema) {
```
nnatter, line 103:
```suggestion
        // load data only after initial schema was set to prevent duplicate requests during initialization
        if (localSchema) {
```

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js
```diff
@@ -87,15 +87,114 @@ export default class ResourceFormStore extends AbstractFormStore implements Form
...
+        let result = {};
+        if (!localSchema || !remoteSchema) {
+            return result;
+        }
+
+        for (const name in remoteSchema) {
+            const {items: remoteItems, type: remoteType, types: remoteTypes} = remoteSchema[name];
+            const {items: localItems, type: localType, types: localTypes} = localSchema[name] || {};
+
+            if (remoteType === SECTION_TYPE && remoteItems &&
```
nnatter, line 127: think we need to split this into two condition blocks to handle a case where only one of the templates has a section 🙈

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js
```diff
@@ -87,15 +87,114 @@ export default class ResourceFormStore extends AbstractFormStore implements Form
...
+                && isArrayLike(localData[name]) && isArrayLike(remoteData[name])
+            ) {
+                for (const key of remoteData[name].keys()) {
+                    const remoteChildData = remoteData[name][key];
+                    const localChildData = toJS(localData[name].length > key ? localData[name][key] || {} : {});
+
+                    const localChildSchema = localTypes[localChildData.type]?.form;
+                    const remoteChildSchema = remoteTypes[remoteChildData.type].form;
+
+                    if (!(localChildData.type in remoteTypes)) {
```
niklasnatter: do we need to do this befor the merge? think i would move it below the merge and change the `resultChildData` otherwise

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js
```diff
@@ -87,15 +87,114 @@ export default class ResourceFormStore extends AbstractFormStore implements Form
...
+                        remoteChildSchema,
+                        localChildData,
+                        remoteChildData
+                    );
+
+                    if (!result[name]) {
+                        result[name] = [];
+                    }
+
+                    if (Object.keys(resultChildData).length === 0) {
```
niklasnatter: not sure when this happen - maybe we can add a comment?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js
niklasnatter: guess we need to move it into the `else` above `result[name][key] = resultChildData;` because of the `Object.keys(resultChildData).length` check

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js
```diff
@@ -87,15 +87,114 @@ export default class ResourceFormStore extends AbstractFormStore implements Form
...
+
+            if (!localData[name]) {
+                result[name] = remoteData[name];
+                if (remoteData.type) {
+                    result.type = remoteData.type;
+                }
+
+                if (remoteData.settings) {
+                    result.settings = localData?.settings || remoteData.settings;
+                }
```
niklasnatter: think i would move this into the `block` case above to make clear that we need to do this only for blocks 🙈

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js
```diff
@@ -87,15 +87,139 @@ export default class ResourceFormStore extends AbstractFormStore implements Form
...
+        if (localSchema) {
+            return this.resourceStore.requestRemoteData({template: this.type}).then((data: Object) => {
+                const result = this.mergeData(localSchema, remoteSchema, this.data, data);
+                this.setMultiple(result);
+            });
+        }
+        return Promise.resolve();
+    };
+
+    mergeData(
```
niklasnatter: think i would define this method outside of the class like we did in the `AbstractFormStore`: https://github.com/sulu/sulu/blob/5e4f11ac8d160c4a1242b8454265925c72a9cf30/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/AbstractFormStore.js#L10-L25

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js
```diff
@@ -87,15 +87,139 @@ export default class ResourceFormStore extends AbstractFormStore implements Form
...
+                types: remoteTypes,
+            } = remoteSchema[name];
+            const {
+                items: localItems,
+                defaultType: localDefaultType,
+                type: localType,
+                types: localTypes,
+            } = localSchema[name] || {};
+
+            if (remoteType === SECTION_TYPE && remoteItems &&
```
niklasnatter: think this is not needed because the conditions below will handle this case implicitly

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js
```diff
@@ -87,15 +87,139 @@ export default class ResourceFormStore extends AbstractFormStore implements Form
...
+                    );
+
+                    if (!result[name]) {
+                        result[name] = [];
+                    }
+
+                    if (Object.keys(resultChildData).length > 0) {
+                        resultChildData.type = resultType;
+
+                        if (resultChildData.settings) {
```
niklasnatter: i think this condition should be negated, right?

niklasnatter (commented):
looks good for me otherwise 🙂

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js
Prokyonn: I don't think so, because in the conditions below we pass either `localSchema`/`remoteItems` or `localItems`/`remoteItems` but never `localItems/remoteItems` and only one condition is called because of the `continue`

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js
Prokyonn: I think the condition can be removed completely, the `settings` are never set before

PR #6079, opened 2021-05-26
Use light skin for lists inside of tabs

niklasnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| License | MIT

#### What's in this PR?

This PR adjusts all lists inside of tab views to use the `light` skin.

##### Before

Screenshot 2021-05-26 at 16 37 59

##### After

Screenshot 2021-05-26 at 16 38 35


niklasnatter:
We decided to use the `light` skin only if there is another important element like the webspace switcher over the list.

PR #6080, opened 2021-05-26
Fix margin of nested tabs and display back button on media activity tab

niklasnatter:


src/Sulu/Bundle/ActivityBundle/Infrastructure/Sulu/Admin/ActivityAdmin.php
```diff
@@ -31,7 +31,7 @@ class ActivityAdmin extends Admin
 
     const LIST_VIEW = 'sulu_activity.activities.list';
 
-    const EDIT_FORM_ACTIVITY_VERSION_TAB_VIEW = 'sulu_page.page_edit_form.activity_version_tab';
+    const PAGE_EDIT_FORM_ACTIVITY_TAB_VIEW = 'sulu_page.page_edit_form.activity_version_tab';
```
niklasnatter: @alexander-schranz should we move these views into the `PageAdmin`/`MediaAdmin` or should we leave them here? 🤔

niklasnatter: not only the constant but also the `$viewCollection->add()` calls

alexander-schranz: They should exist in `PageAdmin` and `MediaAdmin` or be there own `PageActivityAdmin` and  `PageMediaAdmin` classes but still in there specific bundles

alexander-schranz: I think this can be removed now?

PR #6084, opened 2021-05-27
Add Upgrade Information for Composer

TheCadien:
Simple information about Composer 2.0.14 requirement in Ocramius/ProxyManager 2.12.0

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -7,6 +7,11 @@
 The method `Preview::updateContext` has been extended with the `data` argument. The argument is necessary
 to make sure that the rendered data are consistent with the template.
 
+### Ocramius/ProxyManager
+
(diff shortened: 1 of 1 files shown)


alexander-schranz:
I think I will not document this in our `UPGRADE.md` as this is not a change from our side. If somebody updates `composer` or `ocramius/proxy-manager` they would need also take a look at there repository and changes not at sulu/sulu repository else we would need to add any upgrade from all of our dependencies to our and I don't think this is something which we should do.

If you want you could create a [Discussion](https://github.com/sulu/sulu/discussions/categories/q-a)  and add there a answer so it can be found here.

TheCadien:
Okay :)

alexander-schranz:
Still thank you for effort 👍

alexander-schranz:
We decided to switch to the lts version of the proxy manager here so this problem should then not longer exist in all composer versions: https://github.com/sulu/sulu/pull/6093/files

PR #6085, opened 2021-05-27
Prevent loading config files multiple times

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no 
| BC breaks? | no 
| Deprecations? | no  
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Avoid multiple loading of config files

#### Why?

Avoids that sulu_admin config is loaded multiple times.

### DIFF
--- a/src/Sulu/Component/HttpKernel/SuluKernel.php
+++ b/src/Sulu/Component/HttpKernel/SuluKernel.php
@@ -42,6 +42,16 @@ abstract class SuluKernel extends Kernel
     private $reversedContext = self::CONTEXT_WEBSITE;
 
     /**
+     * @var string[]
+     */
+    private $loadedFiles = [];
+
+    /**
+     * @var string[]
+     */
+    private $importedFiles = [];
+
+    /**
      * Overload the parent constructor method to add an additional
      * constructor argument.
      *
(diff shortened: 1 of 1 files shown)


src/Sulu/Component/HttpKernel/SuluKernel.php
```diff
@@ -155,8 +169,10 @@ protected function load(LoaderInterface $loader, $confDir, $pattern)
         $excludedConfigFiles = $this->glob($confDir, $pattern . $reversedConfigExtensions);
 
         foreach ($configFiles as $resource) {
-            if (!\in_array($resource, $excludedConfigFiles)) {
+            if (!\in_array($resource, $excludedConfigFiles) && !\in_array($resource, $this->loadedFiles)) {
```
niklasnatter, line 172: for completion, we need this line to prevent loading the `sulu_admin.yaml` more two times in the `admin` context: this is caused by the following lines:
https://github.com/sulu/sulu/blob/5e0db29766f99ae62528123a5ca78200bbc21be0/src/Sulu/Component/HttpKernel/SuluKernel.php#L85-L92

PR #6088, opened 2021-05-28
Add ActivityViewBuilderFactory

luca-rath:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | yes
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

Add ActivityViewBuilderFactory to help implement an activity tab for custom entities.

#### Example Usage

```php
public function configureViews(ViewCollection $viewCollection): void
{
    // ...

    if ($this->activityViewBuilderFactory->hasPermissionForActivityListView()) {
        $viewCollection->add(
            $this->activityViewBuilderFactory->createActivityListViewBuilder(
                static::EDIT_FORM_VIEW,
                $resourceKey
            )
        );
    }
}
```

The above example will lead to an activity tab like this:

image

#### To Do


luca-rath:
@sulu/core-team how should the activity list for custom entities look like, if there aren't tabs in tabs? Which header? `flat` skin?

Prokyonn:
> @sulu/core-team how should the activity list for custom entities look like, if there aren't tabs in tabs? Which header? `flat` skin?

I think it should look the same as in the `Media` tab. At the moment there is also no tab in tab

niklasnatter:
> @sulu/core-team how should the activity list for custom entities look like, if there aren't tabs in tabs? Which header? flat skin?

We decided to display no header on the media tab: https://sulu.rocks/admin/#/media/en/32/activity?limit=100
Not really happy with that look, but I would prefer to keep things consistent and we havent found a better solution before the release 😕  

It would be nice to display a grey header if there are no tabs, but we would need to implement an additional skin/adapter-option for that.

luca-rath:
> I think it should look the same as in the `Media` tab. At the moment there is also no tab in tab

Yes, you're right 👍 forgot about the activity tab for media 🙈

src/Sulu/Bundle/ActivityBundle/Infrastructure/Sulu/Admin/View/ActivityViewBuilder.php
```diff
@@ -0,0 +1,52 @@
...
+
+namespace Sulu\Bundle\ActivityBundle\Infrastructure\Sulu\Admin\View;
+
+use Sulu\Bundle\ActivityBundle\Domain\Model\ActivityInterface;
+use Sulu\Bundle\ActivityBundle\Infrastructure\Sulu\Admin\ActivityAdmin;
+use Sulu\Bundle\AdminBundle\Admin\View\ListViewBuilder;
+use Sulu\Component\Security\Authorization\PermissionTypes;
+use Sulu\Component\Security\Authorization\SecurityCheckerInterface;
+
+class ActivityViewBuilder extends ListViewBuilder
```
niklasnatter: think i would call this `ActivityListViewBuilder`

src/Sulu/Bundle/ActivityBundle/Infrastructure/Sulu/Admin/View/ActivityViewBuilder.php
```diff
@@ -0,0 +1,52 @@
...
+{
+    public function __construct(string $parent, string $resourceKey, string $resourceIdRouterAttribute = 'id')
```
niklasnatter: think we should use it for the existing activity lists? 🙂

src/Sulu/Bundle/ActivityBundle/Infrastructure/Sulu/Admin/View/ActivityViewBuilder.php
```diff
@@ -0,0 +1,52 @@
+<?php
```
niklasnatter: i am not sure about this implementation because it is kind of inconsistent to the other `ViewBuilder` in the core. as far as i see, all other `ViewBuilder` classes are instantiated via the `ViewBuilderFactory`. not sure about the initial reasons for this, but it would be nice to keep the pattern because of consistency. unfortunately i dont think we want to add a `ActivityBundle` specific method to the `ViewBuilderFactory`. 

we could implement an `ActivityViewBuilderFactory` service that uses the `ViewBuilderFactory`. this way things would still work if somebody overwrites the `ViewBuilderFactory::createListViewBuilder` method in the project. i guess that would be more consistent to the existing code. what do you think?

niklasnatter: looks like we used a somehow similar pattern in the `ContentBundle`: https://github.com/sulu/SuluContentBundle/blob/0.x/Content/Infrastructure/Sulu/Admin/ContentViewBuilderFactory.php

src/Sulu/Bundle/ActivityBundle/Infrastructure/Sulu/Admin/View/ActivityViewBuilderFactory.php
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    public function createActivityListViewBuilder(
+        string $parent,
+        string $resourceKey,
+        string $resourceIdRouterAttribute = 'id'
+    ): ListViewBuilderInterface {
+        /** @var ListViewBuilderInterface $activityListViewBuilder */
+        $activityListViewBuilder = $this->viewBuilderFactory
+            ->createListViewBuilder($parent . '.activity', '/activities')
```
niklasnatter: think i would not hardcode the `name` and `path` like this but pass it explicitly like in the `ViewBuilderFactoryInterface`. there might be a usecase where it makes sense to use another path or where there is no parent

src/Sulu/Bundle/ActivityBundle/Infrastructure/Sulu/Admin/View/ActivityViewBuilderFactory.php
```diff
@@ -0,0 +1,76 @@
...
+            ->setListKey(ActivityInterface::LIST_KEY)
+            ->setTabTitle('sulu_admin.activity')
+            ->addListAdapters(['table'])
+            ->addAdapterOptions([
+                'table' => [
+                    'skin' => 'flat',
+                    'show_header' => false,
+                ],
+            ])
+            ->setTabOrder(4096)
```
niklasnatter: think i would not set a general value for this here but set it in the specific admin class if necessary 🙂

src/Sulu/Bundle/ActivityBundle/Infrastructure/Sulu/Admin/View/ActivityViewBuilderFactory.php
```diff
@@ -0,0 +1,76 @@
...
+            ->disableTabGap()
+            ->disableSearching()
+            ->disableSelection()
+            ->disableColumnOptions()
+            ->disableFiltering()
+            ->addRouterAttributesToListRequest([$resourceIdRouterAttribute => 'resourceId'])
+            ->addRequestParameters(['resourceKey' => $resourceKey])
+            ->setParent($parent);
```
niklasnatter: there might be a usecase where there is no parent. think i would set the parent in the specific admin class

src/Sulu/Bundle/ActivityBundle/Infrastructure/Sulu/Admin/View/ActivityViewBuilderFactory.php
```diff
@@ -0,0 +1,76 @@
...
+
+        return $activityListViewBuilder;
+    }
+
+    public function hasPermissionForActivityListView(): bool
```
niklasnatter: maybe call this `hasActivityListPermission`? 🤔

niklasnatter (approved):
Looks good to me 🙂

PR #6089, opened 2021-05-30
Avoid fetching of node in markup listener to improve performance

### LINKED ISSUE #5935: ContentRepository: Only fetching needed properties to increase Sulu performance
# ContentRepository::resolvePermissions is fetching unused properties

Hi everyone :slightly_smiling_face: We are currently working on several performance fixes, starting with one of our biggest sites with the awesome Sulu 2 :+1: 

On many pages we have now already implemented cache for the navigation and it has some effect - it works well.
But we looked further and noticed that much time is spent in specific methods. We did several Blackfire tests and we noticed that calling the `ContentRepository::findByUuids` takes quite a long time (see screenshot, prod mode).

We looked further and we noticed two things: One has to do with query cache could be used to increase the performance.
See: https://github.com/doctrine/DoctrinePHPCRBundle/pull/359 

The other is about the handling of the **result permissions**:

You can see in the screenshot that this method is constantly called:

result_permission_time

The `ContentRepository::getNode` method fetches the complete node with often about 200-800 properties. And in our case: Only a few or none are left over in the end. ( `sec:*` filter ) 

.. to find in this method: https://github.com/sulu/sulu/blob/c0e3b0b4d64d82111618a48de374e10265f3fb89/src/Sulu/Component/Content/Repository/ContentRepository.php#L455

permission_code
stack_properties

# Possible solution 1: Make it possible to only fetch needed properties

If we could somehow have the possibility to fetch only the properties that are needed in the first place, we would have a performance boost. So with an early return (`[]`) in this function our code **runs 20%-30% faster**. 

# Possible solution 2: Configuration option to disable permissions in website mode

Somehow disable permissions at all (in website mode). I guess on many projects the most pages are anonymous anyway - without firewall involved - so granular checks on permission aren't needed.

Idea: Is a configuration where you can disable permissions completely in that situations useful? Like `permission_enabled: true`
Of course this flag should be enabled by default - but when desired - permissions could be disabled in frontend.

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? | no 
| BC breaks? | no 
| Deprecations? | no yes 
| Fixed tickets | fixes #5935 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Avoid fetching of node in markup listener.

#### Why?

Currently all nodes are lazy loaded inside the security check which has the effect that a lot of queries unneeded queries on the database is done and unneeded hydration is also done, which ends in a lot bigger response time as expected.

#### TODO

 - [x] Save Permission in one collected field
 - [x] Handle SetPermission on Document Hydrate
 - [x] Handle Saving of Permission
 - [x] Output object permission only when one role has permission set
 - [x] Target 2.2 Branch!
 - [x] Update TestCases
 - [x] Create Migration
 
 Before:
 
```bash
| sec:role-2                  | STRING (0)   |                                                                             |
| sec:role-3                  | STRING (0)   |                                                                             |
| sec:role-4                  | STRING (4)   | [0] view                                                                    |
```

Additional new Field:

```bash
| sec:permissions              | STRING (28)  | {"2":[],"3":[],"4":["view"]}                                                |
 ```
 
 
 #### Profiles
 
Unoptimized: https://blackfire.io/profiles/fc104978-ad6f-4f12-a73b-70dfe3df72a6/graph?settings%5Bdimension%5D=wt&settings%5Bdisplay%5D=landscape&settings%5BtabPane%5D=nodes&selected=Sulu%5CComponent%5CContent%5CRepository%5CContentRepository%3A%3AresolveResultPermissions&callname=main()

Bildschirmfoto 2021-06-01 um 10 03 19

Optimized Permissions: https://blackfire.io/profiles/6ae6b900-f331-4dbe-8291-af0930232bc3/graph?settings%5Bdimension%5D=wt&settings%5Bdisplay%5D=landscape&settings%5BtabPane%5D=nodes&selected=Sulu%5CComponent%5CWebspace%5CManager%5CWebspaceManager%3A%3AfindUrlByResourceLocator&callname=main()

Optimized Permissions WebspaceManager d7a679c86aff575b598532d84d63705cbadd1a67: https://blackfire.io/profiles/7754f2dc-ab51-4c51-beb4-01ad83d06f1d/graph?settings%5Bdimension%5D=wt&settings%5Bdisplay%5D=landscape&settings%5BtabPane%5D=nodes&selected=Sulu%5CComponent%5CWebspace%5CManager%5CWebspaceManager%3A%3AfindUrlByResourceLocator&callname=main()

Optimized Permissions WebspaceManager Type b284d3952adb769cf587fadf1a09508f67793657: https://blackfire.io/profiles/20d266c7-e2e3-470c-ac08-f6e317dfaa8c/graph?settings%5Bdimension%5D=wt&settings%5Bdisplay%5D=landscape&settings%5BtabPane%5D=nodes&selected=Sulu%5CComponent%5CWebspace%5CManager%5CWebspaceManager%3A%3AfindUrlByResourceLocator&callname=main()

Optimized Permissions WebspaceManager Type Always 1befd77ef019012feb9e45165670c4101afb2dbd: https://blackfire.io/profiles/d1b7718c-50d6-4593-b6dc-c18b32e094ee/graph?settings%5Bdimension%5D=wt&settings%5Bdisplay%5D=landscape&settings%5BtabPane%5D=nodes&selected=Sulu%5CComponent%5CWebspace%5CManager%5CWebspaceManager%3A%3AfindUrlByResourceLocator&callname=main()

Optimized Permissions WebspaceManager Type Once cdc62a5a9622fa972a1225b37cba6c2fd2daefca: https://blackfire.io/profiles/519c4d61-dd4e-4fbe-a2b9-b160d9441f20/graph?settings%5Bdimension%5D=wt&settings%5Bdisplay%5D=landscape&settings%5BtabPane%5D=nodes&selected=&callname=main()

Bildschirmfoto 2021-06-01 um 01 36 17

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,16 @@
 # Upgrade
 
+## 2.2.11
+
+### Migrate role permissions properties
+
+For performance reasons we need to migrate the permission fields into a single field in PHPCR.
+Run the phpcr migration command which will handle migration:
+
+```bash
+bin/console phpcr:migrations:migrate
+```
+
 ## 2.2.6
 
 ### Changed ContentRepository to return title of source instead of link destination for internal link pages
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -2,10 +2,14 @@
 
 ## 2.2.11
 
-### Migrate role permissions properties
+### Migrate permissions properties for pages
 
-For performance reasons we need to migrate the permission fields into a single field in PHPCR.
-Run the phpcr migration command which will handle migration:
+The role-specific PHPCR properties used for storing page permissions decrease performance when
+used in combination with website security. To mitigate the problem and improve performance, all permissions 
+are now stored in a single property. 
+
+If you use page-specific permissions in your project, you need to migrate the existing data by running the 
+phpcr migration command:
 
 ```bash
 bin/console phpcr:migrations:migrate
--- a/src/Sulu/Component/Content/Document/Subscriber/SecuritySubscriber.php
+++ b/src/Sulu/Component/Content/Document/Subscriber/SecuritySubscriber.php
@@ -12,7 +12,6 @@
 namespace Sulu\Component\Content\Document\Subscriber;
 
 use PHPCR\NodeInterface;
-use PHPCR\PropertyInterface;
 use PHPCR\SessionInterface;
 use Sulu\Component\Content\Document\Behavior\SecurityBehavior;
 use Sulu\Component\DocumentManager\Behavior\Mapping\PathBehavior;
@@ -28,7 +27,12 @@ use Symfony\Component\EventDispatcher\EventSubscriberInterface;
  */
 class SecuritySubscriber implements EventSubscriberInterface
 {
+    /**
+     * @deprecated use the SECURITY_PERMISSION_PROPERTY to access the permissions
+     */
     const SECURITY_PROPERTY_PREFIX = 'sec:role-';
+
+    const SECURITY_PERMISSION_PROPERTY = 'sec:permission';
 
     /**
      * @var array
@@ -150,13 +154,23 @@ class SecuritySubscriber implements EventSubscriberInterface
             }
         }
 
+        $allowedPermissions = [];
         foreach ($permissions as $roleId => $permission) {
-            $allowedPermissions = $this->getAllowedPermissions($permission);
+            $allowedRolePermissions = $this->getAllowedPermissions($permission);
             // TODO use PropertyEncoder, once it is refactored
-            $node->setProperty(static::SECURITY_PROPERTY_PREFIX . $roleId, $allowedPermissions);
+            $node->setProperty(static::SECURITY_PROPERTY_PREFIX . $roleId, $allowedRolePermissions);
             if ($liveNode) {
-                $liveNode->setProperty(static::SECURITY_PROPERTY_PREFIX . $roleId, $allowedPermissions);
+                $liveNode->setProperty(static::SECURITY_PROPERTY_PREFIX . $roleId, $allowedRolePermissions);
             }
+
+            $allowedPermissions[$roleId] = $allowedRolePermissions;
+        }
+
+        $jsonAllowedPermissions = \json_encode($allowedPermissions);
+
+        $node->setProperty(static::SECURITY_PERMISSION_PROPERTY, $jsonAllowedPermissions);
+        if ($liveNode) {
+            $liveNode->setProperty(static::SECURITY_PERMISSION_PROPERTY, $jsonAllowedPermissions);
         }
     }
 
@@ -172,13 +186,20 @@ class SecuritySubscriber implements EventSubscriberInterface
             return;
         }
 
+        if (!$node->hasProperty(static::SECURITY_PERMISSION_PROPERTY)) {
+            return;
+        }
+
+        $nodePermissionJson = $node->getPropertyValue(static::SECURITY_PERMISSION_PROPERTY);
+
+        if (!$nodePermissionJson) {
+            return;
+        }
+
+        $nodePermissions = \json_decode($nodePermissionJson, true);
+
         $permissions = [];
-        foreach ($node->getProperties('sec:*') as $property) {
-            /** @var PropertyInterface $property */
-            $roleId = \substr($property->getName(), 9); // remove the "sec:role-" prefix
-
-            $allowedPermissions = $property->getValue();
-
+        foreach ($nodePermissions as $roleId => $allowedPermissions) {
             foreach ($this->permissions as $permission => $value) {
                 $permissions[$roleId][$permission] = \in_array($permission, $allowedPermissions);
             }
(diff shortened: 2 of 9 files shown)

### CODE REVIEW

File: src/Sulu/Component/Content/Document/Subscriber/SecuritySubscriber.php [hunk 1]
```diff
@@ -28,8 +27,13 @@
  */
 class SecuritySubscriber implements EventSubscriberInterface
 {
+    /**
+     * @deprecated use the SECURITY_PERMISSION_PROPERTY to access the permissions
+     */
     const SECURITY_PROPERTY_PREFIX = 'sec:role-';
 
+    const SECURITY_PERMISSION_PROPERTY = 'sec:permission';
```
alexander-schranz: TODO rename to `permissions`

src/Sulu/Component/Content/Document/Subscriber/SecuritySubscriber.php
```diff
@@ -150,13 +154,23 @@ public function handlePersist(PersistEvent $event)
...
+            $node->setProperty(static::SECURITY_PROPERTY_PREFIX . $roleId, $allowedRolePermissions);
             if ($liveNode) {
-                $liveNode->setProperty(static::SECURITY_PROPERTY_PREFIX . $roleId, $allowedPermissions);
+                $liveNode->setProperty(static::SECURITY_PROPERTY_PREFIX . $roleId, $allowedRolePermissions);
             }
+
+            $allowedPermissions[$roleId] = $allowedRolePermissions;
+        }
+
+        $jsonAllowedPermissions = \json_encode($allowedPermissions);
```
niklasnatter, line 168: should we add a comment here why we duplicate the information?
```
// additionally store all permissions in a single property to allow for more performant access
```

niklasnatter: or maybe move this above the deprecated block and add a comment there?
```
// store role permissions in separated properties for backwards compatibility
```

src/Sulu/Component/Content/Repository/ContentRepository.php
```diff
@@ -458,17 +458,22 @@ private function resolveResultPermissions(array $result, UserInterface $user = n
 
         foreach ($result as $index => $row) {
             $permissions[$index] = [];
-            $securityProperties = $row->getNode()->getProperties(SecuritySubscriber::SECURITY_PROPERTY_PREFIX . '*');
```
niklasnatter, line 461: for completion - this `$row->getNode()` caused the bad performance described in https://github.com/sulu/sulu/issues/5935, right? 🙂

src/Sulu/Component/Webspace/Manager/WebspaceManager.php
```diff
@@ -247,48 +252,89 @@ public function findUrlByResourceLocator(
...
-                        \array_unshift($fullMatchedUrls, $url);
-                    } else {
-                        $fullMatchedUrls[] = $url;
+
+            if (!$sameLocalization) {
+                continue;
+            }
+
+            $portalUrl = $portalInformation->getUrl();
+            if (RequestAnalyzerInterface::MATCH_TYPE_FULL === $portalInformation->getType()) {
```
niklasnatter, line 300: need to trust you with this 🙂

src/Sulu/Component/Webspace/Manager/WebspaceManager.php
```diff
@@ -247,48 +252,89 @@ public function findUrlByResourceLocator(
...
+                if ($this->isFromDomain('http://' . $portalUrl, $domain)) {
```
niklasnatter, line 301: not sure if we need to do this for `https` too?

UPGRADE.md
```diff
@@ -1,5 +1,16 @@
 # Upgrade
 
+## 2.2.11
+
+### Migrate role permissions properties
+
+For performance reasons we need to migrate the permission fields into a single field in PHPCR.
+Run the phpcr migration command which will handle migration:
```
```suggestion
### Migrate permissions properties for pages

The role-specific PHPCR properties used for storing page permissions decrease performance when
used in combination with website security. To mitigate the problem and improve performance, all permissions 
are now stored in a single property. 

If you use page-specific permissions in your project, you need to migrate the existing data by running the 
phpcr migration command:
```

src/Sulu/Component/Content/Repository/ContentRepository.php
alexander-schranz: yes thats the bad performance as it will try to load and hydrate all properties.

src/Sulu/Component/Webspace/Manager/WebspaceManager.php
alexander-schranz: the scheme is not important for this call it just need to check if its the `sameDomain` or not for this it needs to be just a valid url.

src/Sulu/Component/Content/Document/Subscriber/SecuritySubscriber.php
alexander-schranz: Added the comment above for bc

niklasnatter (approved):

PR #6090, opened 2021-05-31
Add possibility to define anchor for internal links

### LINKED ISSUE #3440: Manage target and anchor for page of type internal-link
| Q | A
| --- | ---
| Bug? | no
| New Feature? | yes

Hy,
It would be great if we could have the possibility to

- manage the target (self / blank) of a link (content page > parameters)
- add anchor part on a redirection (content page > parameters > target existing content page)

The cases I'm often facing : 
- create an anchor menu link (target one section on the homepage for example)
--> actually I create a content with an external link with an absolute url and a anchor, that's break the domain logic in the webspace definition.
--> we could have a input for the anchor when selecting "internal link"

sulu-content-parameters

Thks

### LINKED ISSUE #5661: Allow to set anchor hash when using <sulu-link /> tag
# Problem description

The `<sulu-link />` tag allows to create a link to another page by its uuid inside of a twig template. Unfortunately, it is not possible to set the `hash` of such a link.

# Proposed solution

When using the default anchor tag, the hash of a link is just appended to the value of the `href` attribute. For example: 
```html
<a href="https://docs.sulu.io/en/2.0/book/getting-started.html#webspaces">Webspaces</a>
```

Therefore, it would be nice to support something like 
```html
<sulu-link href="123-123-123#hash-value">Link Text</sulu-link>
```

### LINKED ISSUE #5662: Allow to set anchor hash for internal link in text editor
# Problem description

When creating a link to another page in the text editor, it is not possible to set the `hash` of the link.

Screenshot 2020-11-26 at 18 00 37

# Proposed solution

I think it would be nice to add an additional input for setting the `hash` of the link in the overlay visible in the screenshot above.

# Related Issues
- https://github.com/sulu/sulu/issues/5661
- https://github.com/sulu/sulu/issues/3440

luca-rath:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #5661, fixes #5662, fixes #3440
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

Add possibility to define anchor for internal links

### DIFF
--- a/src/Sulu/Bundle/PageBundle/Markup/Link/PageLinkProvider.php
+++ b/src/Sulu/Bundle/PageBundle/Markup/Link/PageLinkProvider.php
@@ -103,17 +103,8 @@ class PageLinkProvider implements LinkProviderInterface
             $domain = $request->getHost();
         }
 
-        $uuids = \array_filter(
-            \array_map(
-                function($href) {
-                    return \explode('#', $href, 2)[0] ?? null;
-                },
-                $hrefs
-            )
-        );
-
         $contents = $this->contentRepository->findByUuids(
-            \array_unique(\array_values($uuids)),
+            \array_unique(\array_values($hrefs)),
             $locale,
             MappingBuilder::create()
                 ->setResolveUrl(true)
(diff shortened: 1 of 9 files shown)


src/Sulu/Bundle/PageBundle/Markup/Link/PageLinkProvider.php
```diff
@@ -103,8 +103,17 @@ public function preload(array $hrefs, $locale, $published = true)
             $domain = $request->getHost();
         }
 
+        $uuids = \array_filter(
```
alexander-schranz: I would handle this things outside of the specific link providers. So the `PageLinkProvider` does only get the `uuids` without the `#` and adding the `#hash` at the end is in the more general service it will in future work for all LinkProviders without the LinkProvider need to take care of the `#` anchors.

alexander-schranz: Think the `LinkTag` service would I think the best place for it.

niklasnatter:
Think I would prefer to target `2.3` here because this is something that pops up quite often, it is still a long time until the `2.4` release and it does not feel like a "big" feature that requires a new minor release to me 🙂

niklasnatter (approved):
Looks good to me 🙂 

Not sure if we should target `2.2` or `2.3`? @sulu/core-team  
Both options are okay for me.

PR #6096, opened 2021-06-02
Fixed single media selection import and checkbox empty import

Quehnie:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

When importing a xliff file into sulu the single_media_selection value will escape the string and lead to an error in the /admin interface.

Also when importing a checkbox with an empty target the set value was "true" and not "false".

#### Why?

Wrong logic in the import.


alexander-schranz:
@Quehnie this should be rebased to 2.2 as bugfix. @wachterjohannes I'm not sure about the changes can you check them if it make sense for you?

wachterjohannes:
@alexander-schranz yeah makes sense - i have discussed that with @Quehnie - and yes has to be rebased to 2.2

PR #6100, opened 2021-06-07
Prevent undefined array key error in PreviewRenderer if locale is not set

niklasnatter:
The locale was passed as explicit parameter that could be null in `2.2`:
https://github.com/sulu/sulu/blob/eac666bd7c9beb0e72e3fd8de8bfde7553100449/src/Sulu/Bundle/PreviewBundle/Preview/Renderer/PreviewRenderer.php#L116

We fallback to `null` in the `Preview` class in multiple places:
https://github.com/sulu/sulu/blob/5e4f11ac8d160c4a1242b8454265925c72a9cf30/src/Sulu/Bundle/PreviewBundle/Preview/Preview.php#L95

### DIFF
--- a/src/Sulu/Bundle/PreviewBundle/Preview/Renderer/PreviewRenderer.php
+++ b/src/Sulu/Bundle/PreviewBundle/Preview/Renderer/PreviewRenderer.php
@@ -115,8 +115,8 @@ class PreviewRenderer implements PreviewRendererInterface
         $partial = false,
         $options = []
     ) {
-        $webspaceKey = $options['webspaceKey'];
-        $locale = $options['locale'];
+        $webspaceKey = $options['webspaceKey'] ?? null;
+        $locale = $options['locale'] ?? null;
 
         if (!$this->routeDefaultsProvider->supports(\get_class($object))) {
             throw new RouteDefaultsProviderNotFoundException($object, $id, $webspaceKey, $locale);


src/Sulu/Bundle/PreviewBundle/Preview/Renderer/PreviewRenderer.php
```diff
@@ -115,8 +115,8 @@ public function render(
         $partial = false,
         $options = []
     ) {
-        $webspaceKey = $options['webspaceKey'];
-        $locale = $options['locale'];
+        $webspaceKey = $options['webspaceKey'] ?? null;
+        $locale = $options['locale'] ?? null;
```
alexander-schranz, line 119: Where this maybe make sense in point of view from the `admin` this doesn't reflect how sulu would behave on the website where a webspace key and locale is always required.

niklasnatter: yeah thats right.. but i think it should not hurt to add a fallback here and we do the same thing in the `Preview` class 🙂 
https://github.com/sulu/sulu/blob/5e4f11ac8d160c4a1242b8454265925c72a9cf30/src/Sulu/Bundle/PreviewBundle/Preview/Preview.php#L95

but i dont have a strong opinion, it is also okay for me to close this

alexander-schranz (approved):

PR #6104, opened 2021-06-10
Prevent deletion of collections with child collections or media

luca-rath:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | yes
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

Prevent deletion of collections with child collections or media

#### Why?

When children are automatically deleted through cascading, it's not possible to throw `CollectionRemovedEvent`'s or `MediaRemovedEvent`'s.


src/Sulu/Bundle/AdminBundle/Exception/DeletionImpossibleChildPermissionsException.php
```diff
@@ -0,0 +1,59 @@
...
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\AdminBundle\Exception;
+
+class DeletionImpossibleChildPermissionsException extends \Exception
```
luca-rath: I'm still unhappy with those exception class names, but I can't find a better one. Does somebody have an idea?

src/Sulu/Bundle/AdminBundle/Exception/DeletionImpossibleChildPermissionsException.php
```diff
@@ -0,0 +1,59 @@
...
+{
+    public const EXCEPTION_CODE = 12346; // TODO change
```
luca-rath: Is there a convention on what exception codes should be used?

src/Sulu/Bundle/AdminBundle/Exception/DeletionImpossibleChildrenException.php
```diff
@@ -0,0 +1,56 @@
...
+        $this->totalChildren = $totalChildren;
+
+        parent::__construct(
+            \sprintf('Resource cannot be deleted, because it has %d children', $this->totalChildren),
+            static::EXCEPTION_CODE
+        );
+    }
+
+    /**
+     * @return array<int, array<array{id: int, resourceKey: string}>>
```
alexander-schranz: Id can also be a `string`

src/Sulu/Bundle/AdminBundle/Exception/DeletionImpossibleChildPermissionsException.php
```diff
@@ -0,0 +1,59 @@
...
+            \sprintf(
+                'Resource cannot be deleted, because the user doesn\'t have permissions for %d children',
+                $this->totalUnauthorizedChildren
+            ),
+            static::EXCEPTION_CODE
+        );
+    }
+
+    /**
+     * @return array<array{id: int, resourceKey: string}>
```
alexander-schranz: `id` can also be a string

src/Sulu/Bundle/MediaBundle/Collection/Manager/CollectionManager.php
```diff
@@ -797,4 +846,277 @@ private function getCollectionMeta(CollectionInterface $collection, ?string $loc
...
+            ->from(CollectionInterface::class, 'collection')
+            ->leftJoin(CollectionInterface::class, 'parent', Join::WITH, 'collection.lft > parent.lft AND collection.rgt < parent.rgt')
+            ->where('parent.id = :id')
+            ->setParameter('id', $id);
+    }
+
+    /**
+     * @return array<array{id: int, resourceKey: string, depth: int}>
+     */
+    private function findChildCollections(int $id): array
```
alexander-schranz: Best would be that all methods which communicate with the database live in the `Repositories`. The `Repository` should be in the future architecture the only service creating this kind of queries / accessing doctrine.

We need to discuss here the naming as it are queries for a specific usecase which are not extendable queries.

I think here about `findChildCollectionIdsForPermissionCheck` or `findChildCollectionIdsForDeletion`. /cc @sulu/core-team 

Another way would also that we use the `findFlat($filters, $fields)` method which should use the field descriptors and so we could there add the required filters e.g.: `findFlat(['not-permission' => 'delete', 'rootCollectionId' => $id], ['id', 'depth'])` and `findFlat(['rootCollectionId' => $id], ['id', 'depth'])`.

src/Sulu/Bundle/MediaBundle/Tests/Functional/Controller/CollectionControllerTest.php
```diff
@@ -1210,6 +1228,155 @@ public function testDeleteById()
...
+        $child2 = $this->createCollection($this->collectionType1, [], $collection);
+        $this->addMedia($child2, 1);
+
+        $this->em->clear();
+
+        $this->client->jsonRequest('DELETE', '/api/collections/' . $collectionId);
+        $this->assertHttpStatusCode(409, $this->client->getResponse());
+
+        $response = \json_decode((string) $this->client->getResponse()->getContent());
+        $this->assertEquals(DeletionImpossibleChildrenException::EXCEPTION_CODE, $response->code);
```
alexander-schranz: Would `assertSame` the whole json without the provided stacktrace e.g.::

```
$jsonData = \json_decode((string) $this->client->getResponse()->getContent());
unset($jsonData['errors']);
$expectedJson = <<<EOT
{
    "..": ".."
}
EOT;

$this->assertSame($expectedJson, \json_encode($jsonData, JSON_PRETTY_PRINT));
```

src/Sulu/Bundle/MediaBundle/Tests/Functional/Controller/CollectionControllerTest.php
```diff
@@ -1210,6 +1228,155 @@ public function testDeleteById()
...
+
+        $child2 = $this->createCollection($this->collectionType1, [], $collection);
+        $this->addMedia($child2, 1);
+
+        $this->em->clear();
+
+        $this->client->jsonRequest('DELETE', '/api/collections/' . $collectionId);
+        $this->assertHttpStatusCode(409, $this->client->getResponse());
+
+        $response = \json_decode((string) $this->client->getResponse()->getContent());
```
alexander-schranz: `$response` is missleading here

src/Sulu/Bundle/MediaBundle/Tests/Functional/Controller/CollectionControllerTest.php
```diff
@@ -1210,6 +1228,155 @@ public function testDeleteById()
...
+        $child2 = $this->createCollection($this->collectionType1, ['de' => 'Child 2'], $collection);
+        $this->accessControlManager->setPermissions(Collection::class, (string) $child2->getId(), $permissions);
+
+        $this->em->clear();
+
+        $this->client->jsonRequest('DELETE', '/api/collections/' . $collectionId);
+        $this->assertHttpStatusCode(403, $this->client->getResponse());
+
+        $response = \json_decode((string) $this->client->getResponse()->getContent());
+        $this->assertEquals(DeletionImpossibleChildPermissionsException::EXCEPTION_CODE, $response->code);
```
alexander-schranz: Don't use constants here. It should fail when the contants is changed.

src/Sulu/Bundle/MediaBundle/Tests/Functional/Controller/CollectionControllerTest.php
```diff
@@ -1210,6 +1228,155 @@ public function testDeleteById()
...
+        $this->assertEquals(DeletionImpossibleChildrenException::EXCEPTION_CODE, $response->code);
+        $this->assertTrue(isset($response->message));
+        $this->assertObjectHasAttribute('totalChildren', $response);
+        $this->assertSame(7, $response->totalChildren);
+        $this->assertObjectHasAttribute('children', $response);
+        $this->assertEquals([
+            [
+                (object) [
+                    'id' => $child11->getMedia()->first()->getId(),
+                    'resourceKey' => MediaInterface::RESOURCE_KEY,
```
alexander-schranz: Don't use constants in tests it should fail when a constant is changed

src/Sulu/Bundle/MediaBundle/Tests/Functional/Controller/CollectionControllerTest.php
```diff
@@ -1210,6 +1228,155 @@ public function testDeleteById()
...
+
+        $this->em->clear();
+
+        $this->client->jsonRequest('DELETE', '/api/collections/' . $collectionId);
+        $this->assertHttpStatusCode(403, $this->client->getResponse());
+
+        $response = \json_decode((string) $this->client->getResponse()->getContent());
+        $this->assertEquals(DeletionImpossibleChildPermissionsException::EXCEPTION_CODE, $response->code);
+        $this->assertTrue(isset($response->message));
+        $this->assertObjectHasAttribute('totalUnauthorizedChildren', $response);
```
alexander-schranz: same as above would jsut assertSame the pretty printed json and maybe remove before the stacktrace from it.

src/Sulu/Component/Rest/FlattenExceptionNormalizer.php
```diff
@@ -64,6 +66,16 @@ public function normalize($exception, $format = null, array $context = [])
             $data['errors'] = [$errors];
         }
 
+        if ($contextException instanceof DeletionImpossibleChildPermissionsException) {
```
alexander-schranz: we should create an Interface for it like we did for the `TranslationErrorMessageExceptionInterface`

src/Sulu/Component/Rest/FlattenExceptionNormalizer.php
```diff
@@ -64,6 +66,16 @@ public function normalize($exception, $format = null, array $context = [])
...
+            $data['totalUnauthorizedChildren'] = $contextException->getTotalUnauthorizedChildren();
+            $data['unauthorizedChildren'] = $contextException->getUnauthorizedChildResources();
+        }
+
+        if ($contextException instanceof DeletionImpossibleChildrenException) {
```
alexander-schranz: Like above we should create an interface for it.

alexander-schranz (changes requested):

src/Sulu/Bundle/MediaBundle/Collection/Manager/CollectionManager.php
luca-rath: Didn't implement this suggestion for now, until we have a decision on how to do it

src/Sulu/Bundle/MediaBundle/Tests/Functional/Controller/CollectionControllerTest.php
luca-rath: @alexander-schranz What about comparing two arrays instead of two json strings? Comparing two json strings is harder, because variables like `$child1->getMedia()->first()->getId()` can't be used that easily

src/Sulu/Bundle/MediaBundle/Collection/Manager/CollectionManagerInterface.php
```diff
@@ -121,7 +121,7 @@ public function save(array $data, ?int $userId = null, bool $breadcrumb = false)
      *
      * @param int $id the id of the category to delete
      */
-    public function delete($id);
+    public function delete($id, bool $forceRemoveChildren = false);
```
alexander-schranz: Should not be added in the interface we should handle this over `func_num_args` and `func_get_arg` like in the other methods for bc reasons.

src/Sulu/Bundle/AdminBundle/Exception/DeletionImpossibleChildPermissionsException.php
alexander-schranz: Every bundle had its prefix number in the past example media was `5` and so 5000-5999 area was reserved for media exception codes thats the only one I remember.

src/Sulu/Bundle/AdminBundle/Exception/DeletionImpossibleChildPermissionsException.php
alexander-schranz: Also not sure maybe @nnatter has an idea how we could call this exceptions.

Just something which came to my mind:

 - DenyRemoveByChildPermissionsException
 - ConflictRemoveByExistChildrenException

src/Sulu/Bundle/MediaBundle/Tests/Functional/Controller/CollectionControllerTest.php
alexander-schranz: is okay for me at current state

src/Sulu/Bundle/MediaBundle/Collection/Manager/CollectionManagerInterface.php
luca-rath: We also adjusted some interfaces for sulu 2.3 and agreed, that it's okay to make such a bc break in a minor version. Why should we handle this differently now?

src/Sulu/Bundle/AdminBundle/Exception/DeletionImpossibleChildPermissionsException.php
luca-rath: what about those?

- `DeletionWithChildrenNotAllowedException`
- `InsufficientChildPermissionsException`

niklasnatter: havent had a good idea either 🙈 

i would prefer `InsufficientChildPermissionsException`/`MissingChildPermissionsException`, but i guess we need to include `Deletion` because of the message right?
maybe `DeletionBlockedByInsufficientChildPermissionsException`?

src/Sulu/Bundle/MediaBundle/Collection/Manager/CollectionManagerInterface.php
alexander-schranz: We avoid bc breaks where it is possible and this is easy possible.

src/Sulu/Bundle/AdminBundle/Exception/DeletionImpossibleChildPermissionsException.php
luca-rath: No we don't need to make this exception specific for deletion. I guess we could also show this exception, if a user changes the permissions of a collection and checks `Override permissions of sub elements`, but doesn't have the `security` permission of one of the child collections.

luca-rath: Additionally, we could also add a `permission` property, which can be used to determine, if the user tried to delete or modify

src/Sulu/Bundle/MediaBundle/Collection/Manager/CollectionManager.php
luca-rath: should be fixed now

luca-rath:
Linting fails because of phpstan, but that has nothing to do with this pull request

src/Sulu/Bundle/MediaBundle/Collection/Manager/CollectionManager.php
```diff
@@ -797,4 +841,54 @@ private function getCollectionMeta(CollectionInterface $collection, ?string $loc
...
+        $result = \array_merge($childCollections, $childMedia);
+
+        return $this->groupResourcesByDepth($result);
+    }
+
+    private function countAllChildrenByRootCategory(int $id): int
+    {
+        $countChildCollections = $this->collectionRepository->countChildCollectionsOfRootCollection($id);
+
+        $childCollectionIds = \array_merge([$id], $this->collectionRepository->findChildCollectionIdsOfRootCollection($id));
```
alexander-schranz: I don't see a need for this method it should just give the `rootColectionId` into the `MediaRepository` and use the nested tree to get all children.

src/Sulu/Bundle/MediaBundle/Entity/CollectionRepository.php
```diff
@@ -409,4 +411,131 @@ public function supportsDescendantType(string $type): bool
...
+            ->distinct()
+            ->orderBy('collection.id', 'ASC')
+            ->getQuery()
+            ->getArrayResult();
+    }
+
+    /**
+     * @return int[]
+     */
+    public function findChildCollectionIdsOfRootCollection(int $rootCollectionId): array
```
alexander-schranz: think this method is not needed the `mediaRepository` should just get a `rootCollectionId`

src/Sulu/Bundle/MediaBundle/Collection/Manager/CollectionManager.php
```diff
@@ -797,4 +841,54 @@ private function getCollectionMeta(CollectionInterface $collection, ?string $loc
...
+
+        \krsort($grouped);
+
+        return \array_values($grouped);
+    }
+
+    /**
+     * @return array<int, array<array{id: int, resourceKey: string}>>
+     */
+    private function findAllChildResourcesByRootCategory(int $rootCollectionId): array
```
alexander-schranz: `Category`?

src/Sulu/Bundle/MediaBundle/Collection/Manager/CollectionManager.php
```diff
@@ -582,7 +587,46 @@ public function delete($id)
...
+
+                    throw new InsufficientChildPermissionsException(
+                        $childCollectionResourcesWithoutPermissions,
+                        $totalChildResourcesWithoutPermissions,
+                        PermissionTypes::DELETE
+                    );
+                }
+            }
+
+            $totalChildResources = $this->countAllChildrenByRootCategory($id);
```
alexander-schranz: I think we don't need this count as we could at the end just do `count($childResources)` to return a total so we don't need this additonal method on the repositories to just count something and trigger similar queries against the database.

src/Sulu/Bundle/MediaBundle/Entity/CollectionRepository.php
```diff
@@ -409,4 +411,131 @@ public function supportsDescendantType(string $type): bool
...
+    }
+
+    /**
+     * @return array<array{id: int, resourceKey: string, title: string|null}>
+     */
+    public function findUnauthorizedChildCollectionResourcesOfRootCollection(
+        int $id,
+        UserInterface $user,
+        int $permission,
+        ?int $maxResults = null
```
alexander-schranz: we could this parameter in other repositories `$limit`

src/Sulu/Bundle/MediaBundle/Entity/CollectionRepository.php
```diff
@@ -409,4 +411,108 @@ public function supportsDescendantType(string $type): bool
...
+        $permittedChildCollectionsQb = $this
+            ->createPermittedChildCollectionsQueryBuilder('permittedCollection', $id, $user, $permission)
+            ->select('permittedCollection.id')
+            ->distinct()
+            ->orderBy('permittedCollection.id', 'ASC');
+
+        $qb->andWhere($alias . '.id NOT IN (' . $permittedChildCollectionsQb->getDQL() . ')');
+
+        /** @var Query\Parameter $parameter */
+        foreach ($permittedChildCollectionsQb->getParameters() as $parameter) {
```
niklasnatter: maybe we could add all parameters with a single `setParameters` call

src/Sulu/Bundle/MediaBundle/Entity/CollectionRepository.php
```diff
@@ -409,4 +411,108 @@ public function supportsDescendantType(string $type): bool
...
+                $alias . '.lft > ' . $rootCollectionAlias . '.lft AND ' . $alias . '.rgt < ' . $rootCollectionAlias . '.rgt'
+            )
+            ->where($rootCollectionAlias . '.id = :id')
+            ->setParameter('id', $rootCollectionId);
+    }
+
+    /**
+     * @return array<array{id: int, resourceKey: string, depth: int}>
+     */
+    public function findChildCollectionResourcesOfRootCollection(int $rootCollectionId): array
```
niklasnatter: i think i would use `descendant`/`ancestor`/ instead of `root`. we dont need to include `child` then and can use `findDescendantCollectionResources`/`findCollectionResourcesByAncestor`

src/Sulu/Bundle/MediaBundle/Entity/CollectionRepository.php
```diff
@@ -409,4 +411,108 @@ public function supportsDescendantType(string $type): bool
     {
         return $this->getClassName() === $type;
     }
+
+    private function createChildCollectionsQueryBuilder(string $alias, int $rootCollectionId): QueryBuilder
```
niklasnatter: think these are descendants - `createDescendantCollectionsQueryBuilder`

src/Sulu/Bundle/MediaBundle/Entity/CollectionRepository.php
```diff
@@ -409,4 +411,108 @@ public function supportsDescendantType(string $type): bool
...
+            ->select('collection.id AS id')
+            ->addSelect('\'' . CollectionInterface::RESOURCE_KEY . '\' AS resourceKey')
+            ->addSelect('collection.depth AS depth')
+            ->distinct()
+            ->orderBy('collection.id', 'ASC')
+            ->getQuery()
+            ->getArrayResult();
+    }
+
+    private function createPermittedChildCollectionsQueryBuilder(
```
niklasnatter: would use the same term i the different mehtod names. so either `authorized`+`unauthorized` or`permitted`+`unpermitted`

src/Sulu/Bundle/MediaBundle/Entity/MediaRepository.php
```diff
@@ -445,6 +445,30 @@ private function getIds(
         return $subQuery->getScalarResult();
     }
 
+    /**
+     * @return array<array{id: int, resourceKey: string, depth: int}>
+     */
+    public function findMediaResourcesOfRootCollection(int $rootCollectionId): array
```
niklasnatter: not sure what would be better word for `root` here.. maybe `findMediaResourcesByAncestorCollection`

niklasnatter: or call the method `findMediaResourcesByCollection` and add a `includeDescendantCollections = true` parameter?

src/Sulu/Component/Rest/Exception/DependantResourcesFoundException.php
```diff
@@ -0,0 +1,51 @@
...
+{
+    /**
+     * @var array<int, array<array{id: int|string, resourceKey: string}>>
+     */
+    private $dependantResources;
+
+    /**
+     * @var int
+     */
+    private $totalDependantResources;
```
niklasnatter: `dependantResourceCount`

src/Sulu/Component/Rest/Exception/InsufficientChildPermissionsException.php
```diff
@@ -0,0 +1,65 @@
...
+{
+    /**
+     * @var array<array{id: int|string, resourceKey: string}>
+     */
+    private $unauthorizedChildResources;
+
+    /**
+     * @var int
+     */
+    private $totalUnauthorizedChildResources;
```
niklasnatter:
```suggestion
    private $unauthorizedChildResourceCount;
```

src/Sulu/Bundle/MediaBundle/Collection/Manager/CollectionManager.php
```diff
@@ -582,7 +587,46 @@ public function delete($id)
...
+                        3
+                    );
+
+                    throw new InsufficientChildPermissionsException(
+                        $childCollectionResourcesWithoutPermissions,
+                        $totalChildResourcesWithoutPermissions,
+                        PermissionTypes::DELETE
+                    );
+                }
+            }
```
alexander-schranz: maybe we could move this into a method `checkPermissionForDelete();` to use early returns instead of have this nested if statements e.g.:

```php
            $user = $this->getCurrentUser();

            if (null === $user) {
                 return;
            }
            
            $totalChildResourcesWithoutPermissions = $this->collectionRepository->countUnauthorizedChildCollectionsOfRootCollection(
                $id,
                $user,
                $this->permissions[PermissionTypes::DELETE]
            );

            if ($totalChildResourcesWithoutPermissions) {
                 return;
            }
            
           $childCollectionResourcesWithoutPermissions = $this->collectionRepository->findUnauthorizedChildCollectionResourcesOfRootCollection(
                $id,
                $user,
                $this->permissions[PermissionTypes::DELETE],
                3
           );

           throw new InsufficientChildPermissionsException(
                $childCollectionResourcesWithoutPermissions,
                $totalChildResourcesWithoutPermissions,
                PermissionTypes::DELETE
           );
```

src/Sulu/Bundle/MediaBundle/Collection/Manager/CollectionManager.php
```diff
@@ -582,7 +587,46 @@ public function delete($id)
...
+            $childResources = $this->findAllChildResourcesByRootCollection($id);
+
+            if (!empty($childResources)) {
+                $totalChildResources = $this->countGroupedResources($childResources);
+
+                throw new DependantResourcesFoundException(
+                    $childResources,
+                    $totalChildResources
+                );
+            }
```
alexander-schranz: If we do the above I would do the same for this lines `checkDependantResourcesForDelete()`

alexander-schranz (changes requested):

src/Sulu/Component/Rest/Exception/DependantResourcesFoundException.php
```diff
@@ -0,0 +1,51 @@
...
+    /**
+     * @param array<int, array<array{id: int|string, resourceKey: string}>> $dependantResources
+     */
+    public function __construct(array $dependantResources, int $totalDependantResources)
+    {
+        $this->dependantResources = $dependantResources;
+        $this->totalDependantResources = $totalDependantResources;
+
+        parent::__construct(
+            \sprintf('Resource has %d dependant resources', $this->totalDependantResources),
```
alexander-schranz:
```suggestion
            \sprintf('Resource has %d dependant resources.', $this->totalDependantResources),
```

src/Sulu/Component/Rest/Exception/RestExceptionInterface.php
```diff
@@ -0,0 +1,49 @@
...
+    const EXCEPTION_CODE_INVALID_HASH = 1102;
+
+    /**
+     * Cannot delete resource, because it has children.
+     *
+     * @see DependantResourcesFoundExceptionInterface
+     *
+     * @var int
+     */
+    const EXCEPTION_CODE_DEPENDANT_RESOURCES_FOUND = 1103;
```
alexander-schranz: 1103 seems already be used for `ResourceLocatorAlreadyExistsException`

alexander-schranz (changes requested):

luca-rath:
~~I have no idea why the tests are failing on jackrabbit 🤷~~

Jackrabbit tests are failing on the `2.x` branch as well

alexander-schranz (approved):
@luca-rath looks good from my side. Can you rebase think then should the 7.2 tests run.

PR #6107, opened 2021-06-15
Fix writing 0 as numeric value

alengodev:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Fixes the issue saving 0 as a value


alexander-schranz:
@alengodev Thank you for the PR. Did add a test case and rebase it to 2.2 branch so its also fixed there.

alexander-schranz (approved):

Issue #6106, opened 2021-06-15
Wrong exception when lang is configured on database but not in sulu_core


ehibes:
| Q | A
| --- | ---
| Sulu Version | 2.3
| PHP Version | 7.4
| DB Version | MariaDB
| Browser Version | Chrome

# Actual Behavior

Context : trying to update a 2.2 project to 2.3 project. When I update the database part, Sulu throws an exception : 
``` There is no Metadata available for the type "form" with the key "snippet". ```
This is because a lang is set on updated database (fr in my case), but currently not configured on sulu_core.

# Expected Behavior

**The app should not crash, or a better exception should be thrown.**

# Steps to Reproduce

Sulu 2.2 project with alternative lang configured in `sulu_core`. 
Make a 2.3 fresh install, link the 2.2 database, update it.

# Possible Solutions

Update `sulu_admin.yml` file, adding missing locale on `sulu_core` part. 
The exception is not thrown anymore.

alexander-schranz:
@ehibes Thank you for creating this issue. We decided to add here a fallback to the default locale. It can be tested here #6108

PR #6108, opened 2021-06-15
Add fallback for form metadata for none exist user locale

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? |  yes
| New feature? | no 
| BC breaks? | no 
| Deprecations? | no  
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Add fallback for form metadata for none exist user locale

#### Why?

See https://github.com/sulu/sulu/issues/6106.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/FormMetadataProvider.php
+++ b/src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/FormMetadataProvider.php
@@ -28,12 +28,29 @@ class FormMetadataProvider implements MetadataProviderInterface
      */
     private $formMetadataLoaders;
 
+    /**
+     * @var string
+     */
+    private $defaultLocale;
+
+    /**
+     * @param string[]|null $locales
+     */
     public function __construct(
         ExpressionLanguage $expressionLanguage,
-        iterable $formMetadataLoaders
+        iterable $formMetadataLoaders,
+        ?array $locales = []
     ) {
         $this->expressionLanguage = $expressionLanguage;
         $this->formMetadataLoaders = $formMetadataLoaders;
+
+        if (!$locales) {
+            @\trigger_error('The usage of the "FormMetadataProvider" without "$locales" is deprecated. Please add "$locales" instead.', \E_USER_DEPRECATED);
+
+            $locales = ['en'];
+        }
+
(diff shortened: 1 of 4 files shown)


src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/FormMetadataProvider.php
```diff
@@ -28,12 +28,29 @@ class FormMetadataProvider implements MetadataProviderInterface
...
+    private $defaultLocale;
+
+    /**
+     * @param string[]|null $locales
+     */
     public function __construct(
         ExpressionLanguage $expressionLanguage,
-        iterable $formMetadataLoaders
+        iterable $formMetadataLoaders,
+        ?array $locales = []
```
niklasnatter: i think we need to add this parameter to the service definition? 🤔

src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/FormMetadataProvider.php
```diff
@@ -28,12 +28,29 @@ class FormMetadataProvider implements MetadataProviderInterface
      */
     private $formMetadataLoaders;
 
+    /**
+     * @var string
+     */
+    private $defaultLocale;
```
niklasnatter: would call this `fallbackLocale`

src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/FormMetadataProvider.php
niklasnatter: i guess you are injecting the configured locales here - maybe we can call it `systemLocales`?

niklasnatter: or maybe directly use the `sulu_core.fallback_locale`? that one is already used in the `AdminController`

alexander-schranz: Oh did search for default_locale but not for fallback_locale, will directly use here the fallback locale instead of sulu_core.locales.

PR #6115, opened 2021-06-16
Introduce ChangeContext to prevent marking form as dirty when setting default value

### LINKED ISSUE #5927: Save button is not enabled after changing the template of a page

niklasnatter:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Fixed tickets | fixes #5927
| License | MIT

#### What's in this PR?

This PR adds an additional `FieldChangeContext` argument to the `onChange` callback that is passed to all field-types of the `Form` component. The argument allows to pass additional information about the change to the `Form` component.

The newly added argument is utilized to set a default value from a field-type to the store of the form without marking the form as dirty. 

#### Why?

At the moment, to allow for setting default values, the form is marked as not-dirty after  the initial rendering. Because of this, the form is marked as not-dirty too after switching the template of a form (see https://github.com/sulu/sulu/issues/5927)

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/Form.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/Form.js
@@ -10,7 +10,7 @@ import Renderer from './Renderer';
 import FormInspector from './FormInspector';
 import GhostDialog from './GhostDialog';
 import MissingTypeDialog from './MissingTypeDialog';
-import type {FormStoreInterface} from './types';
+import type {FieldChangeContext, FormStoreInterface} from './types';
 
 type Props = {|
     onError?: (errors: Object) => void,
@@ -55,13 +55,6 @@ class Form extends React.Component<Props> {
         this.displayGhostDialogDisposer();
     }
 
-    @action setRendererRef = () => {
-        const {store} = this.props;
-        // This avoids starting with an already dirty form
-        // That can happen if a field type calls the onChange handler to set a default value in their its constructor
-        store.dirty = false;
-    };
-
     @computed get formInspector(): FormInspector {
         return new FormInspector(this.props.store);
     }
@@ -96,8 +89,12 @@ class Form extends React.Component<Props> {
         }
     };
 
-    handleChange = (name: string, value: mixed) => {
-        this.props.store.change(name, value);
+    handleChange = (name: string, value: mixed, context: FieldChangeContext) => {
+        if (context.isDefaultValue) {
+            this.props.store.set(name, value);
+        } else {
+            this.props.store.change(name, value);
+        }
     };
 
     @action showGhostDialog() {
@@ -189,7 +186,6 @@ class Form extends React.Component<Props> {
                         onChange={this.handleChange}
                         onFieldFinish={this.handleFieldFinish}
                         onSuccess={onSuccess}
-                        ref={this.setRendererRef}
                         router={router}
                         schema={store.schema}
                         schemaPath=""
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/Form.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/Form.js
@@ -119,7 +119,7 @@ class Form extends React.Component<Props> {
     @action handleMissingTypeDialogConfirm = (type: string) => {
         const {store} = this.props;
 
-        store.setType(type);
+        store.changeType(type);
     };
 
     @action handleMissingTypeDialogCancel = () => {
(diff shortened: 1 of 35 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/Form.js
```diff
@@ -96,8 +89,12 @@ class Form extends React.Component<Props> {
         }
     };
 
-    handleChange = (name: string, value: mixed) => {
-        this.props.store.change(name, value);
+    handleChange = (name: string, value: mixed, context: FieldChangeContext) => {
+        if (context.isDefaultValue) {
+            this.props.store.set(name, value);
```
alexander-schranz: Maybe we could in future migrate `set` and `change` into one method and give the `context` into that methods. In that case maybe `isDefaultValue` could be misleading and maybe would need then be `noDirty: true`.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/types.js
```diff
@@ -121,12 +121,17 @@ export interface FormStoreInterface {
...
     +resourceKey: ?string,
     +schema: Object,
+    +set: (name: string, value: mixed) => void,
     +setMultiple: (data: Object) => void,
     +setType: (type: string) => void,
     +types: {[key: string]: SchemaType},
     +validate: () => boolean,
 }
 
+export type FieldChangeContext = {
```
niklasnatter: maybe call this `FormChangeContext` and pass it to the `FormStoreInterface::change` method

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/types.js
```diff
@@ -121,12 +121,17 @@ export interface FormStoreInterface {
...
+    isDefaultValue?: boolean,
```
niklasnatter: maybe call this `markFormAsDirty`

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/types.js
```diff
@@ -121,12 +121,17 @@ export interface FormStoreInterface {
     +options: SchemaOptions,
     +resourceKey: ?string,
     +schema: Object,
+    +set: (name: string, value: mixed) => void,
```
niklasnatter: we should get rid of the `FormStoreInterface::set` and `FormStoreInterface::setMultiple` methods and use `change` and `changeMultiple` with a context instead

niklasnatter: parameters of these methods should use `dataPath` instead of `name`

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/types.js
alexander-schranz: I would go the other way `markFormNotDirty: true` as markFormAsDirty should be the default.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/Form.js
```diff
@@ -126,7 +119,7 @@ class Form extends React.Component<Props> {
     @action handleMissingTypeDialogConfirm = (type: string) => {
         const {store} = this.props;
 
-        store.setType(type);
+        store.changeType(type);
```
niklasnatter, line 122: i think it makes sense to enable the save button in this case

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/types.js
alexander-schranz: I'm also okay to keep it this way, so we are maybe flexible how to behave on default vs server values.

alexander-schranz (approved):

PR #6121, opened 2021-06-18
Fix audience targeting loop with user context based caching

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? |  yes
| New feature? | no 
| BC breaks? | no 
| Deprecations? | no  
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Fix audience targeting loop with user context based caching.

#### Why?

Avoid infinite loop between audience targeting and user context based caching.

#### Example Usage

Kernel.php on prod:

```php
            // Activate the following for user based caching see also:
            // https://foshttpcachebundle.readthedocs.io/en/latest/features/user-context.html
            $this->httpCache->addSubscriber(
                new \FOS\HttpCache\SymfonyCache\UserContextListener([
                    'session_name_prefix' => 'SULUSESSID',
                ])
            );

            $this->httpCache->addSubscriber(new AudienceTargetingCacheListener());
```

### DIFF
--- a/src/Sulu/Bundle/AudienceTargetingBundle/EventListener/AudienceTargetingCacheListener.php
+++ b/src/Sulu/Bundle/AudienceTargetingBundle/EventListener/AudienceTargetingCacheListener.php
@@ -44,7 +44,8 @@ class AudienceTargetingCacheListener implements EventSubscriberInterface
     {
         $request = $cacheEvent->getRequest();
 
-        // requests like "/_sulu_target_group" or "/_fos_user_context_hash" should be ignored
+        // the friendsofsymfony/http-cache-bundle package uses the "internalRequest" attribute to mark internal requests
+        // return early in this case to prevent loops on requests "/_sulu_target_group" or "/_fos_user_context_hash"
         if ($request->attributes->get('internalRequest', false)) {
             return;
         }
--- a/src/Sulu/Bundle/AudienceTargetingBundle/EventListener/AudienceTargetingCacheListener.php
+++ b/src/Sulu/Bundle/AudienceTargetingBundle/EventListener/AudienceTargetingCacheListener.php
@@ -44,6 +44,7 @@ class AudienceTargetingCacheListener implements EventSubscriberInterface
     {
         $request = $cacheEvent->getRequest();
 
+        // requests like "/_sulu_target_group" or "/_fos_user_context_hash" should be ignored
         if ($request->attributes->get('internalRequest', false)) {
             return;
         }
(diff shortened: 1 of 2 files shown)


src/Sulu/Bundle/AudienceTargetingBundle/EventListener/AudienceTargetingCacheListener.php
```diff
@@ -42,24 +42,28 @@ class AudienceTargetingCacheListener implements EventSubscriberInterface
 
     public function preHandle(CacheEvent $cacheEvent)
     {
-        if (self::TARGET_GROUP_URL === $cacheEvent->getRequest()->getRequestUri()) {
+        $request = $cacheEvent->getRequest();
+
+        // requests like "/_sulu_target_group" or "/_fos_user_context_hash" should be ignored
+        if ($request->attributes->get('internalRequest', false)) {
```
niklasnatter, line 49: is this attribute set anywhere else too? or do we need to set it only in the `AudienceTargetingCacheListener`?

niklasnatter: i guess the `UserContextListener` sets this too? if not, i would call it `isTargetGroupRequest`

alexander-schranz: Yes this is required to be named `internalRequest` as it need to be match with the attribute the `UserContextListener` is setting here: https://github.com/FriendsOfSymfony/FOSHttpCache/blob/39999a6c60d8bdc756436dd0739202d897d29351/src/SymfonyCache/UserContextListener.php#L225

src/Sulu/Bundle/AudienceTargetingBundle/EventListener/AudienceTargetingCacheListener.php
```diff
@@ -42,24 +42,28 @@ class AudienceTargetingCacheListener implements EventSubscriberInterface
 
     public function preHandle(CacheEvent $cacheEvent)
     {
-        if (self::TARGET_GROUP_URL === $cacheEvent->getRequest()->getRequestUri()) {
+        $request = $cacheEvent->getRequest();
+
+        // requests like "/_sulu_target_group" or "/_fos_user_context_hash" should be ignored
```
niklasnatter:
```suggestion
        // the friendsofsymfony/http-cache-bundle package uses the "internalRequest" attribute to mark internal requests
        // return early in this case to prevent loops on requests "/_sulu_target_group" or "/_fos_user_context_hash"
```

Issue #5931, opened 2021-04-09
The future of repository interfaces for sulu models


alexander-schranz:
# Problem description

In the last time @nnatter and I did discuss many times how a repository interfaces should look like in future for our domain models (contacts, medias, categories, ...) in Sulu, I think we did find a good solution in current projects. In this issue I want to write down my thoughts about how a RepositoryInterfaces for our Models should look like. It is focused focus to be flexible, easy to extend and consistent between all of them.

# Proposed solution

In the following Example the name `Model` could be replaced with  `Contact`, `Media`, `Category` or any other Entity in Sulu.

## A. 💉 Injecting instead of extending the doctrine entity repository

Extending the doctrine entity repository did make us some problems when doctrine add new features like in the past the `count` method we had the rename our `count` method with a [bc break](https://github.com/sulu/sulu/pull/3694/files). Also our class should be not be connected to the infrastructure code and only make this method available which are part of our interface. So we should inject the doctrine repository instead of extending from it. This can be done in repositories the following way:

```php
class ModelRepository implements ModelRepositoryInterface {
    /** @var EntityRepository<ModelInterface> */
    protected $entityRepository;

    /** @var EntityManagerInterface */
    protected $entityManager;
    
    public function __construct(
        EntityManagerInterface $entityManager
    ) {
        $this->entityRepository = $entityManager->getRepository(ModelInterface::class);
        $this->entityManager = $entityManager;
    }
}
```

This means that the Sulu ModelRepository will not be available over `$entityManager->getRepository` and the only correct way to inject the repository is injecting the ModelRepository over its service id or over the ModelRepositoryInterface. 

## B. 🏭 Factory on the Repository

This is currently done by us this way over the persistence bundle and I think I would keep it this way I don't see a need for a custom ModelFactory service. This is very Sulu specific as we use doctrine metadata as to allow override entities.

```php
class ModelRepositoryInterface {
      public function create(): ModelInterface;
}
```

## C. 📗 Add and Remove over Repository

Currently a lot of entityManager->persist and entityManager->remove are done in services. This should be avoided and our repository should have a `add` and a `remove` method:

```php
class ModelRepositoryInterface {
      public function add(ModelInterface $model): void;
      
      public function remove(ModelInterface $model): void;
}
```

The `flush` should be done over middlewares or called manually by the dev. We should not automatical flush something in handler, managers. The "UserInterface" layer should be the one which tells when it is the right time to flush.

[...]

PR #6132, opened 2021-06-24
Implement prototype for SuluTrashBundle

### LINKED ISSUE #5150: Trashcan functionality
# Problem description

If you delete a content it is deleted with all it's children from the database. All data is immediately lost.

Bringing back deleted nodes is a common question in the slack channel and today one of my content managers did remove a full tree of news. Nasty thing to happen.

# Proposed solution

Do not delete anything at first. Just flag content as "deleted" or "trash" or something like this. You could just add a flag to the data table or add a "thrash" workspace to the PHPCR. Don't know technically what is the better thing to do.

I guess that the document tree of the trash might be a problem here with duplicated routes and items without a parent node. Ad hoc I have no idea about this. Maybe just add a hash or timestamp.

Then add a trash can view to the backoffice. Here you can see all deleted content. You can restore or permanently delete things.

The "Permanently Delete" option should not open a modal but lead to a real "Are you sure page?" and log the deletion. Maybe even dump the data to the log.

Maybe add an "empty all" option.

Maybe add a CLI command for cron usage.

For the restore option it might be wise to use the "move" dialog where you want to restore the content if the tree has changed.

luca-rath:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | yes
| Deprecations? | no
| Fixed tickets | fixes #5150
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

Implement prototype for SuluTrashBundle


src/Sulu/Bundle/TagBundle/Tag/TagManager.php
```diff
@@ -53,16 +58,23 @@ class TagManager implements TagManagerInterface
      */
     private $domainEventCollector;
 
+    /**
+     * @var TrashManagerInterface
```
alexander-schranz: Should be nullable the whole `TrashBundle` should be able to be removed from the `config/bundles` and all bundles should work without it. 

```suggestion
     * @var TrashManagerInterface|null
```

src/Sulu/Bundle/TagBundle/Tag/TagManager.php
```diff
@@ -53,16 +58,23 @@ class TagManager implements TagManagerInterface
...
+     */
+    private $trashManager;
+
     public function __construct(
         TagRepositoryInterface $tagRepository,
         ObjectManager $em,
         EventDispatcherInterface $eventDispatcher,
-        DomainEventCollectorInterface $domainEventCollector
+        DomainEventCollectorInterface $domainEventCollector,
+        TrashManagerInterface $trashManager
```
alexander-schranz:
```suggestion
        TrashManagerInterface $trashManager = null
```

src/Sulu/Bundle/TagBundle/Tag/TagManager.php
```diff
@@ -158,6 +218,8 @@ public function delete($id)
             throw new TagNotFoundException($id);
         }
 
+        $this->trashManager->store(TagInterface::RESOURCE_KEY, $tag);
```
alexander-schranz:
```suggestion
        if ($this->trashManager) {
            $this->trashManager->store(TagInterface::RESOURCE_KEY, $tag);
        }
```

src/Sulu/Bundle/TrashBundle/Application/TrashItemHandler/SupportableTrashItemHandlerInterface.php
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\TrashBundle\Application\TrashItemHandler;
+
+interface SupportableTrashItemHandlerInterface
```
alexander-schranz: Not sure if I would create this interface, I would move the method into the 2 other interface.

src/Sulu/Bundle/TrashBundle/Application/TrashManager/TrashManager.php
```diff
@@ -0,0 +1,133 @@
...
+
+        throw new StoreTrashItemHandlerNotFoundException($resourceKey);
+    }
+
+    public function restore(TrashItemInterface $trashItem, array $restoreFormData): object
+    {
+        $resourceKey = $trashItem->getResourceKey();
+
+        foreach ($this->restoreTrashItemHandlers as $restoreTrashItemHandler) {
+            if (!$restoreTrashItemHandler->supports($resourceKey)) {
```
alexander-schranz: We are currently not sure if we would go here with the `supports` pattern because there should only exist 1 handler per resourceKey it would be better that the index of the trashItemHandler directly is the resource key over the [`default_index_method`](https://symfony.com/doc/current/service_container/tags.html#tagged-services-with-index) e.g.:

```php
$restoreTrashItemHandler = $this->restoreTrashItemHandlers[$resourceKey];
```

For this we would need to introduce a `static` method instead of `supports` method not sure yet about the name we got here:

 - `public static function getTrashItemResourceKey(): string`
 - `public static function handleResourceKey(): string`
 - `public static function responsibleResourceKey(): string`
 - `public static function supportedResourceKey(): string`
 - `public static function getResourceKey(): string`
 
/cc @sulu/core-team

src/Sulu/Bundle/TagBundle/Resources/config/services.xml
```diff
@@ -15,6 +15,7 @@
             <argument type="service" id="doctrine.orm.entity_manager"/>
             <argument type="service" id="event_dispatcher"/>
             <argument type="service" id="sulu_activity.domain_event_collector"/>
+            <argument type="service" id="sulu_trash.trash_manager"/>
```
alexander-schranz:
```suggestion
            <argument type="service" id="sulu_trash.trash_manager" on-invalid="null"/>
```

src/Sulu/Bundle/TagBundle/Resources/config/services.xml
```diff
@@ -62,5 +63,15 @@
 
             <tag name="sulu.context" context="admin"/>
         </service>
+
+        <!-- Trash -->
+        <service id="sulu_tag.tag_trash_item_handler"
```
alexander-schranz: should only be registered when the tag bundle extension is enabled. So I would move this in an own `services_trash.xml`  and load that when the TrashBundle extension is enabled.

src/Sulu/Bundle/TrashBundle/Domain/Factory/TrashItemFactory.php
```diff
@@ -0,0 +1,87 @@
...
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\TrashBundle\Domain\Factory;
+
+use Sulu\Bundle\TrashBundle\Domain\Model\TrashItemInterface;
+use Sulu\Component\Security\Authentication\UserInterface;
+use Symfony\Component\Security\Core\Security;
```
alexander-schranz: The Security service make this service infrastructure specific so only the `TrashItemFactoryInterface` can be in the `Domain` namespace and the `TrashItemFactory` needs to be moved to the `Infrastructure` namespace.

src/Sulu/Bundle/TrashBundle/Domain/Factory/TrashItemFactory.php
```diff
@@ -0,0 +1,87 @@
...
+
+class TrashItemFactory implements TrashItemFactoryInterface
```
alexander-schranz: Would mark this service as final as we will not support extending here as official extension point:

```suggestion
final class TrashItemFactory implements TrashItemFactoryInterface
```

(if we keep this as own service and don't add the create method on the repository)

src/Sulu/Bundle/TrashBundle/Domain/Factory/TrashItemFactoryInterface.php
```diff
@@ -0,0 +1,31 @@
...
+
+use Sulu\Bundle\TrashBundle\Domain\Model\TrashItemInterface;
+
+interface TrashItemFactoryInterface
+{
+    /**
+     * @param mixed[] $restoreData
+     * @param string|array<string, string> $resourceTitle
+     */
+    public function create(
```
alexander-schranz: I'm thinking about if we should move this into the `Repository` like we have a `create` or `createNew` method currently there for all our entities, because in our case we did merge Model Factory and Repository together.

alexander-schranz: /cc @sulu/core-team what do you think?

src/Sulu/Bundle/TagBundle/Trash/TagTrashItemHandler.php
```diff
@@ -0,0 +1,73 @@
...
+            throw new \InvalidArgumentException();
+        }
+
+        return $this->trashItemFactory->create(
+            TagInterface::RESOURCE_KEY,
+            (string) $tag->getId(),
+            [
+                'name' => $tag->getName(),
+            ],
+            $tag->getName(),
```
alexander-schranz, line 53: this is not correct for translated items. We need to give an array of titles into it e.g.:

```php
$titles = [];
foreach ($someEntity->getTranslations() as $someEntityTranslation) {
     $titles[$someEntityTranslation->getLocale()] = $someEntityTranslation->getTitle();
}

return $this->trashItemFactory->create(
    // ...
    ],
    $titles,
```

src/Sulu/Bundle/TrashBundle/Domain/Model/TrashItem.php
```diff
@@ -0,0 +1,225 @@
...
+
+    public function getTranslation(?string $locale = null): TrashItemTranslation
+    {
+        /** @var TrashItemTranslation|false $translation */
+        $translation = $this->translations->filter(
+            function(TrashItemTranslation $translation) use ($locale) {
+                return $translation->getLocale() === $locale;
+            }
+        )->first()
+            ?: $this->translations->first();
```
alexander-schranz: I personally not a fan of automatic fallbacks or fallbacks which can not be disabled in model methods so I would here go with:

```suggestion
    public function getTranslation(?string $locale = null, bool $fallback = false): TrashItemTranslation
    {
        /** @var TrashItemTranslation|false $translation */
        $translation = $this->translations->filter(
            function(TrashItemTranslation $translation) use ($locale) {
                return $translation->getLocale() === $locale;
            }
        )->first();
        
        if (!$translation && $fallback) {
            $translation = $this->translations->first();
        }
```

src/Sulu/Bundle/TagBundle/Trash/TagTrashItemHandler.php
```diff
@@ -0,0 +1,73 @@
...
+
+use Sulu\Bundle\TagBundle\Admin\TagAdmin;
+use Sulu\Bundle\TagBundle\Tag\TagInterface;
+use Sulu\Bundle\TagBundle\Tag\TagManagerInterface;
+use Sulu\Bundle\TrashBundle\Application\TrashItemHandler\RestoreTrashItemHandlerInterface;
+use Sulu\Bundle\TrashBundle\Application\TrashItemHandler\StoreTrashItemHandlerInterface;
+use Sulu\Bundle\TrashBundle\Domain\Factory\TrashItemFactoryInterface;
+use Sulu\Bundle\TrashBundle\Domain\Model\TrashItemInterface;
+
+class TagTrashItemHandler implements StoreTrashItemHandlerInterface, RestoreTrashItemHandlerInterface
```
alexander-schranz: Would mark this as final as I don't see this as extension point and I think this service never need to be mocked else we should create a `TagTrashItemHandlerInterface`:

```suggestion
final class TagTrashItemHandler implements StoreTrashItemHandlerInterface, RestoreTrashItemHandlerInterface
```

src/Sulu/Bundle/TrashBundle/Application/TrashManager/TrashManager.php
```diff
@@ -0,0 +1,133 @@
...
+use Sulu\Bundle\TrashBundle\Application\TrashItemHandler\StoreTrashItemHandlerInterface;
+use Sulu\Bundle\TrashBundle\Domain\Event\TrashItemCreatedEvent;
+use Sulu\Bundle\TrashBundle\Domain\Event\TrashItemRemovedEvent;
+use Sulu\Bundle\TrashBundle\Domain\Event\TrashItemRestoredEvent;
+use Sulu\Bundle\TrashBundle\Domain\Exception\RestoreTrashItemHandlerNotFoundException;
+use Sulu\Bundle\TrashBundle\Domain\Exception\StoreTrashItemHandlerNotFoundException;
+use Sulu\Bundle\TrashBundle\Domain\Model\TrashItemInterface;
+use Sulu\Bundle\TrashBundle\Domain\Repository\TrashItemRepositoryInterface;
+
+class TrashManager implements TrashManagerInterface
```
alexander-schranz: Would mark this service as final as extending is no official extension point here:

```suggestion
final class TrashManager implements TrashManagerInterface
```

src/Sulu/Bundle/TrashBundle/Infrastructure/Doctrine/Repository/TrashItemRepository.php
```diff
@@ -0,0 +1,67 @@
...
+
+namespace Sulu\Bundle\TrashBundle\Infrastructure\Doctrine\Repository;
+
+use Doctrine\ORM\EntityManager;
+use Doctrine\ORM\EntityRepository;
+use Sulu\Bundle\TrashBundle\Domain\Exception\TrashItemNotFoundException;
+use Sulu\Bundle\TrashBundle\Domain\Model\TrashItemInterface;
+use Sulu\Bundle\TrashBundle\Domain\Repository\TrashItemRepositoryInterface;
+
+class TrashItemRepository implements TrashItemRepositoryInterface
```
alexander-schranz: At current state I would not make this extendable so I would go here also with final:

```suggestion
final class TrashItemRepository implements TrashItemRepositoryInterface
```

src/Sulu/Bundle/TrashBundle/Infrastructure/Doctrine/Repository/TrashItemRepository.php
```diff
@@ -0,0 +1,67 @@
...
+{
+    /**
+     * @var EntityManager
```
alexander-schranz: Should be `EntityManagerInterface`
```suggestion
     * @var EntityManagerInterface
```

src/Sulu/Bundle/TrashBundle/Infrastructure/Doctrine/Repository/TrashItemRepository.php
```diff
@@ -0,0 +1,67 @@
...
+     */
+    protected $entityManager;
+
+    /**
+     * @var EntityRepository<TrashItemInterface>
+     */
+    protected $entityRepository;
+
+    public function __construct(EntityManager $entityManager)
```
alexander-schranz:
```suggestion
    public function __construct(EntityManagerInterface $entityManager)
```

src/Sulu/Bundle/TrashBundle/Infrastructure/Doctrine/Repository/TrashItemRepository.php
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\TrashBundle\Infrastructure\Doctrine\Repository;
+
+use Doctrine\ORM\EntityManager;
```
alexander-schranz:
```suggestion
use Doctrine\ORM\EntityManagerInterface;
```

alexander-schranz (changes requested):
Overall looks good from my side. Good job 👍 . Only some things we need to discuss. Would be great if @nnatter could have a look at the Admin JS code.

src/Sulu/Bundle/TagBundle/Tag/TagManager.php
luca-rath: why not just setting the `sulu_trash.storage.repository` to `null` instead of `doctrine`, like we did in the `ActivityBundle`?

src/Sulu/Bundle/TrashBundle/Application/TrashItemHandler/SupportableTrashItemHandlerInterface.php
luca-rath: Yes normally I would think the same, but since most bundles will implement both `StoreTrashItemHandlerInterface` and `RestoreTrashItemHandlerInterface` in a single class, i thought there should also be a single declaration for the `supports()` method. What about renaming the interface to `TrashItemHandlerInterface`?

src/Sulu/Bundle/TrashBundle/Domain/Factory/TrashItemFactoryInterface.php
luca-rath: There were two reasons for me to introduce the `TrashItemFactory`

1. Developers don't need to inject the `Repository` in the `TrashItemHandler` and therefore don't have to worry about, if they should `addAndCommit()` the new `TrashItem` or if this is handled automatically for them.
2. To not duplicate the logic of setting the current user

src/Sulu/Bundle/TagBundle/Trash/TagTrashItemHandler.php
luca-rath: The `create()` method can handle both, strings and arrays, because I thought it's kind of weird for untranslated entities to pass `[null => $tag->getName()]`

src/Sulu/Bundle/TrashBundle/Domain/Factory/TrashItemFactory.php
luca-rath: But in which subdirectory of `Infrastructure`? `Infrastructure/Sulu/Trash`?

src/Sulu/Bundle/TrashBundle/Infrastructure/Doctrine/Repository/TrashItemRepository.php
```diff
@@ -0,0 +1,67 @@
...
+use Sulu\Bundle\TrashBundle\Domain\Exception\TrashItemNotFoundException;
+use Sulu\Bundle\TrashBundle\Domain\Model\TrashItemInterface;
+use Sulu\Bundle\TrashBundle\Domain\Repository\TrashItemRepositoryInterface;
+
+final class TrashItemRepository implements TrashItemRepositoryInterface
+{
+    /**
+     * @var EntityManager
+     */
+    protected $entityManager;
```
luca-rath:
```suggestion
    private $entityManager;
```

src/Sulu/Bundle/TrashBundle/Infrastructure/Doctrine/Repository/TrashItemRepository.php
```diff
@@ -0,0 +1,67 @@
...
+
+    /**
+     * @var EntityRepository<TrashItemInterface>
+     */
+    protected $entityRepository;
```
luca-rath:
```suggestion
    private $entityRepository;
```

src/Sulu/Bundle/TagBundle/Tag/TagManager.php
luca-rath: Would rather not do this, since all future constructor arguments after `$trashManager` also need `= null` then

luca-rath: `?TrashManagerInterface $trashManager` in combination with `on-invalid="null"` should do the job too

src/Sulu/Bundle/TagBundle/Trash/TagTrashItemHandler.php
luca-rath: And I thought it's kind of weird for untranslated entities to pass `[null => $tag->getName()]`

src/Sulu/Bundle/TrashBundle/Domain/Factory/TrashItemFactory.php
alexander-schranz: It would be `Infrastructure/Symfony/Security`

src/Sulu/Bundle/TrashBundle/Domain/Model/TrashItemInterface.php
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\TrashBundle\Domain\Model;
+
+use Sulu\Component\Security\Authentication\UserInterface;
+
+interface TrashItemInterface
```
alexander-schranz, line 16: we could extend from the AuditableInterface so we don't need to handle the user our self

src/Sulu/Bundle/TrashBundle/Infrastructure/Doctrine/Repository/TrashItemRepository.php
```diff
@@ -0,0 +1,67 @@
...
+
+    public function __construct(EntityManagerInterface $entityManager)
+    {
+        $this->entityManager = $entityManager;
+        $this->entityRepository = $entityManager->getRepository(TrashItemInterface::class);
+    }
+
+    public function addAndCommit(TrashItemInterface $trashItem): void
+    {
+        $this->entityManager->persist($trashItem);
```
alexander-schranz, line 89: I think we don't need `addAndCommit` here as it should just be flushed with when the entity is really removed. So this can just be `add`

src/Sulu/Bundle/TrashBundle/Infrastructure/Doctrine/Repository/TrashItemRepository.php
```diff
@@ -0,0 +1,67 @@
...
+        $this->entityManager->flush();
```
alexander-schranz: should not be flushed here.

src/Sulu/Bundle/TrashBundle/Infrastructure/Doctrine/Repository/TrashItemRepository.php
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    public function removeAndCommit(TrashItemInterface $trashItem): void
+    {
+        $this->entityManager->remove($trashItem);
+        $this->entityManager->flush();
```
alexander-schranz: should not be flushed here.

alexander-schranz (changes requested):

src/Sulu/Bundle/TrashBundle/Domain/Factory/TrashItemFactory.php
luca-rath: This feels a little bit strange ... What if the `TrashItemFactory` would have a dependency on the `Security` service and on the `EventDispatcher` service e.g. ... I guess the `Infrastructure/Symfony/Security` namespace would be the right place for services implementing an interface from symfony's security component

src/Sulu/Bundle/TrashBundle/Domain/Model/TrashItemInterface.php
luca-rath: Since trash items could theoretically also be stored in elasticsearch, the `AuditableInterface` wouldn't work anymore, because it's bound to doctrine

src/Sulu/Bundle/TagBundle/Tag/TagManager.php
alexander-schranz: The future constructor arguments should be `null` and trigger a deprecation or correct error message from BC promise. Adding arguments to service constructors make the container hard fail so this should be if possible always be avoided to avoid errors when a symfony application is deployed without green blue deployment.

src/Sulu/Bundle/TrashBundle/Infrastructure/Doctrine/Repository/TrashItemRepository.php
luca-rath: This is only true, if the storage adapter is `doctrine`, but won't work with `elasticsearch` anymore

src/Sulu/Bundle/TrashBundle/Infrastructure/Doctrine/Repository/TrashItemRepository.php
luca-rath: This is only true, if the storage adapter is `doctrine`, but won't work with `elasticsearch` anymore

src/Sulu/Bundle/TrashBundle/Infrastructure/Doctrine/Repository/TrashItemRepository.php
luca-rath: This is only true, if the storage adapter is `doctrine`, but won't work with `elasticsearch` anymore

src/Sulu/Bundle/TagBundle/Tag/TagManager.php
alexander-schranz: So I definitely would add ` = null` as it does not hurt and the __construct doesn't care about inheritance matching.

src/Sulu/Bundle/TrashBundle/Application/TrashItemHandler/SupportableTrashItemHandlerInterface.php
alexander-schranz: Don't see any usage of this Interface so I would really just move the method to both interfaces. This seperate interface has not advantage in my view and will not be used in our application logic.

src/Sulu/Bundle/TrashBundle/Infrastructure/Doctrine/Repository/TrashItemRepository.php
alexander-schranz: We should in no case flush inside a repository or a handler as flushes are sideeffects.

In the doctrine implementation the repository should not flush if we implement ever a elasticsearch that is an own implementation which need to handle the flush therself.

luca-rath: But the doctrine implementation and the elasticsearch implementation would only have a different repository, right? All other services should stay the same. So where should be "committed" instead?

src/Sulu/Bundle/TrashBundle/Domain/Factory/TrashItemFactory.php
luca-rath: Removed the `TrashItemFactory` and implemented a `create()` method in the `TrashItemRepository` instead

src/Sulu/Bundle/TrashBundle/Domain/Model/TrashItemInterface.php
luca-rath: And a `TrashItem` also can't have a `changer`

src/Sulu/Bundle/TrashBundle/Application/TrashManager/TrashManager.php
luca-rath: Implemented it as `getResourceKey()` now, but I don't have a strong opinion about that

src/Sulu/Bundle/TagBundle/Tag/TagManager.php
```diff
@@ -143,6 +155,54 @@ public function save($data, $id = null)
...
+        $this->em->persist($tag);
+
+        $this->domainEventCollector->collect(
+            new TagRestoredEvent($tag, $data)
+        );
+
+        $this->em->flush();
+
+        $metadata->setIdGeneratorType($oldIdGeneratorType);
+        $metadata->setIdGenerator($oldIdGenerator);
```
luca-rath, line 201: I'm not sure, if it's a good idea to set the id like this ... but is there a better way?

src/Sulu/Bundle/TagBundle/Tag/TagManager.php
```diff
@@ -143,6 +155,56 @@ public function save($data, $id = null)
...
+
+            $this->domainEventCollector->collect(
+                new TagRestoredEvent($tag, $data)
+            );
+
+            $this->em->flush();
+        } finally {
+            $metadata->setIdGeneratorType($oldIdGeneratorType);
+            $metadata->setIdGenerator($oldIdGenerator);
+        }
```
luca-rath: I'm not sure, if it's a good idea to set the id like this ... but is there a better way?

src/Sulu/Bundle/TagBundle/Tag/TagManager.php
```diff
@@ -143,6 +155,56 @@ public function save($data, $id = null)
...
     }
 
+    public function restore(int $id, array $data): TagInterface
+    {
+        $name = $data['name'];
+
+        /** @var TagInterface|null $existingTag */
+        $existingTag = $this->tagRepository->findTagById($id);
+        if (null !== $existingTag) {
+            throw new TagAlreadyExistsException($existingTag->getName());
```
niklasnatter, line 165: think this exception is a bit misleading here. should we restore the tag with a new id in this case?

src/Sulu/Bundle/TagBundle/Trash/TagTrashItemHandler.php
```diff
@@ -0,0 +1,73 @@
...
+
+    public function __construct(TrashItemRepositoryInterface $trashItemRepository, TagManagerInterface $tagManager)
+    {
+        $this->trashItemRepository = $trashItemRepository;
+        $this->tagManager = $tagManager;
+    }
+
+    public function store(object $tag): TrashItemInterface
+    {
+        if (!$tag instanceof TagInterface) {
```
niklasnatter: maybe use the `Assert` library here?

src/Sulu/Bundle/TrashBundle/Infrastructure/Symfony/DependencyInjection/Configuration.php
```diff
@@ -0,0 +1,55 @@
...
+    private function addObjectsSection(ArrayNodeDefinition $node): void
+    {
+        $node->children()
+            ->arrayNode('objects')
+                ->addDefaultsIfNotSet()
+                ->children()
+                    ->arrayNode('trash_item')
+                        ->addDefaultsIfNotSet()
+                        ->children()
+                            ->scalarNode('model')->defaultValue(TrashItem::class)->end()
```
niklasnatter, line 48: not sure if you have omitted the repository intentionally here

src/Sulu/Bundle/TrashBundle/Infrastructure/Symfony/DependencyInjection/SuluTrashExtension.php
```diff
@@ -0,0 +1,119 @@
...
+        }
+    }
+
+    public function load(array $configs, ContainerBuilder $container): void
+    {
+        $configuration = new Configuration();
+        $config = $this->processConfiguration($configuration, $configs);
+
+        $this->loadServices($container);
+        $this->configurePersistence($config['objects'], $container);
```
niklasnatter, line 92: if we make the repository overwritable, we should add an alias for the interface here like in other bundles:
https://github.com/sulu/sulu/blob/9f893118c6dd247ac8704093a338b42e9a5b1668/src/Sulu/Bundle/TagBundle/DependencyInjection/SuluTagExtension.php#L82-L86

src/Sulu/Bundle/TrashBundle/Domain/Model/TrashItem.php
```diff
@@ -0,0 +1,228 @@
...
+
+    /**
+     * @var string|null
+     */
+    private $resourceSecurityObjectId;
+
+    /**
+     * @var \DateTimeImmutable
+     */
+    private $timestamp;
```
niklasnatter: maybe call this `storeTimestamp` or `created`?

src/Sulu/Bundle/TrashBundle/Domain/Model/TrashItem.php
```diff
@@ -0,0 +1,228 @@
...
+ */
+
+namespace Sulu\Bundle\TrashBundle\Domain\Model;
+
+use Doctrine\Common\Collections\ArrayCollection;
+use Doctrine\Common\Collections\Collection;
+use Sulu\Bundle\TrashBundle\Domain\Exception\TrashItemTranslationNotFoundException;
+use Sulu\Component\Security\Authentication\UserInterface;
+
+class TrashItem implements TrashItemInterface
```
niklasnatter, line 21: translation handling looks nice! 🙂

src/Sulu/Bundle/TrashBundle/Resources/js/views/List/itemActions/RestoreItemAction.js
```diff
@@ -0,0 +1,95 @@
...
+            icon: 'su-process',
+            onClick: item?.id ? () => this.handleRestoreClick(item.id, item?.resourceKey) : undefined,
+            disabled: !item?.id,
+        };
+    }
+
+    getNode(): Node {
+        return (
+            <React.Fragment key="restore">
+                <Dialog
```
niklasnatter, line 74: took me some time to understand this. think it would be a bit easier to understand when using conditional rendering like
```
this.restoreFormKey ? (<RestoreFormOverlay ...) : (<Dialog ...)
```

src/Sulu/Bundle/TrashBundle/Resources/js/views/List/itemActions/RestoreItemAction.js
```diff
@@ -0,0 +1,95 @@
...
+            }))
+            .catch(action((response) => {
+                this.restoring = false;
+                this.idToBeRestored = undefined;
+                this.resourceKeyToBeRestored = undefined;
+
+                this.listStore.setShouldReload(true);
+
+                response.json().then(action((error) => {
+                    this.list.errors.push(error.detail || error.message || translate('sulu_trash.restore_error'));
```
niklasnatter: `error.message` will be very generic in a prod environment. would use the same properties as everywhere else:
```
                this.errors.push(error.detail || error.title || translate('sulu_admin.form_save_server_error'));
```

src/Sulu/Bundle/TrashBundle/Infrastructure/Doctrine/Repository/TrashItemRepository.php
alexander-schranz: This seems to be not set yet?

src/Sulu/Bundle/TagBundle/Trash/TagTrashItemHandler.php
alexander-schranz: Is fine for me.

src/Sulu/Bundle/TrashBundle/Domain/Model/TrashItem.php
```diff
@@ -0,0 +1,228 @@
...
+    }
+
+    public function setUser(?UserInterface $user): TrashItemInterface
+    {
+        $this->user = $user;
+
+        return $this;
+    }
+
+    public function getTranslation(?string $locale = null, bool $fallback = true): TrashItemTranslation
```
alexander-schranz: Would prefer having fallback as false by default and explicit set to true where needed.

```suggestion
    public function getTranslation(?string $locale = null, bool $fallback = true): TrashItemTranslation
```

src/Sulu/Bundle/TrashBundle/Infrastructure/Sulu/Admin/TrashAdmin.php
```diff
@@ -0,0 +1,141 @@
...
+use Sulu\Bundle\AdminBundle\Admin\View\ListItemAction;
+use Sulu\Bundle\AdminBundle\Admin\View\ToolbarAction;
+use Sulu\Bundle\AdminBundle\Admin\View\ViewBuilderFactoryInterface;
+use Sulu\Bundle\AdminBundle\Admin\View\ViewCollection;
+use Sulu\Bundle\TrashBundle\Domain\Model\TrashItemInterface;
+use Sulu\Component\Localization\Manager\LocalizationManagerInterface;
+use Sulu\Component\Security\Authorization\PermissionTypes;
+use Sulu\Component\Security\Authorization\SecurityCheckerInterface;
+
+class TrashAdmin extends Admin
```
alexander-schranz: Would also make our Admin classes in future `final`:

```suggestion
final class TrashAdmin extends Admin
```

src/Sulu/Bundle/TrashBundle/Infrastructure/Symfony/DependencyInjection/Configuration.php
luca-rath: As written in #5931, our repositories should be more than just doctrine repositories. Therefore they should also not extend from `EntityRepository`. But when configuring the `repository` here, the `TrashItemRepository` **is** the doctrine repository. I guess it would be better to instead inject the doctrine repository into the `TrashItemRepository`, because both of them are and should be different services. If someone wants to override the `TrashItemRepository`, he can just decorate the `TrashItemRepository` service. Also, `$em->getRepository(TrashItem::class)` should return the doctrine repository instead of the `TrashItemRepository`, because the latter one isn't an `EntityRepository` and I would expect `$em->getRepository()` to return an `EntityRepository`

niklasnatter: sounds reasonable 🙂  
i guess it would be better to use existing symfony mechanisms (eg. decoration) instead of implementing a custom mechanism in the `PersistenceBundle` for overwriting repositories in the future

src/Sulu/Bundle/TrashBundle/Infrastructure/Symfony/DependencyInjection/SuluTrashExtension.php
alexander-schranz: also when not overwriteable the `RepositoryInterface` should alias to the repository service id.

luca-rath: It does, see https://github.com/sulu/sulu/pull/6132/files#diff-e3d3425492946f3707c8da36ea5693c56f29ce37ccadc2a47d88551650eb8fa7R11

src/Sulu/Bundle/TrashBundle/Resources/js/views/List/itemActions/RestoreItemAction.js
luca-rath: Dialogs should not be rendered conditionally, because animations would not work when closing the dialog, right?

src/Sulu/Bundle/TrashBundle/Domain/Model/TrashItem.php
luca-rath: Would be great to finish the conversation first, if the AuditableInterface should be used or not. If so, this is not necessary anymore.

src/Sulu/Bundle/TagBundle/Tag/TagManager.php
niklasnatter: thought about this again; it is not possible to create two tags with the same name, therefore it should not be possible to restore one too. 

thinks its okay to throw an exception here. would be nice if the exception displays a helpful error message in the user interface 🙂

src/Sulu/Bundle/TrashBundle/Domain/Model/TrashItem.php
niklasnatter: dont have a strong opinion on that 🙂  is okay for me to not use it because `changer` and `changed` are not really useful for trash items

src/Sulu/Bundle/TrashBundle/Domain/Model/TrashItemInterface.php
niklasnatter: as said in https://github.com/sulu/sulu/pull/6132/files#r701771320, it would be okay for me to not use the `AuditableInterface` because `changer` and `changed` are kind of confusing on a trash item

src/Sulu/Bundle/TrashBundle/Resources/config/lists/trash_items.xml
```diff
@@ -0,0 +1,96 @@
...
+            <entity-name>%sulu.model.trash_item.class%_translation</entity-name>
+            <field-name>%sulu.model.trash_item.class%.translations</field-name>
+            <condition>%sulu.model.trash_item.class%_translation.locale = :locale</condition>
+        </join>
+    </joins>
+
+    <joins name="fallbackTranslation">
+        <join>
+            <entity-name>%sulu.model.trash_item.class%_fallbackTranslation</entity-name>
+            <field-name>%sulu.model.trash_item.class%.translations</field-name>
```
niklasnatter, line 27: doesnt this join multiple translation if the trash item has multiple translations? i think sucha a trash item would appear in the list multiple times? 🤔

niklasnatter: but it would be okay for me to merge it like this and think about it when implementing the functionality for a localized entity

src/Sulu/Bundle/TrashBundle/Resources/js/views/List/itemActions/RestoreItemAction.js
niklasnatter: i dont know to be honest, but its possible. is okay for me like this too 🙂

UPGRADE.md
```diff
@@ -2,6 +2,50 @@
...
+
+Finally, you need to include the routes of the bundle in your `config/routes/sulu_admin.yaml`:
+
+```yaml
+sulu_trash_api:
+    resource: "@SuluTrashBundle/Resources/config/routing_api.yml"
+    type: rest
+    prefix: /admin/api
+```
+
```
niklasnatter, line 48: i think its also necessary to add the javascript to the `package.json` and `index.js`

UPGRADE.md
```diff
@@ -2,6 +2,62 @@
...
+A new bundle was added to the `sulu/sulu` package. The `SuluTrashBundle` implements the functionality to move resources
+to trash and restore them from trash. To register the services of the bundle in your project, you need
+to add the bundle to your `config/bundles.php` file:
+
+```diff
++    Sulu\Bundle\TrashBundle\SuluTrashBundle::class => ['all' => true],
+```
+
+Additionally, you need to update your database schema to include the tables that are used by the bundle:
```
```suggestion
To update your database schema to include the tables that are used by the bundle, you need to execute the following SQL statements:
```

UPGRADE.md
```diff
@@ -2,6 +2,62 @@
...
+CREATE TABLE tr_trash_item_translations (id INT AUTO_INCREMENT NOT NULL, locale VARCHAR(191) DEFAULT NULL, title VARCHAR(191) NOT NULL, trashItemId INT NOT NULL, INDEX IDX_8264DAF45C8D7CA (trashItemId), UNIQUE INDEX UNIQ_8264DAF45C8D7CA4180C698 (trashItemId, locale), PRIMARY KEY(id)) DEFAULT CHARACTER SET utf8mb4 COLLATE `utf8mb4_unicode_ci` ENGINE = InnoDB;
+ALTER TABLE tr_trash_items ADD CONSTRAINT FK_102989B64B64DCC FOREIGN KEY (userId) REFERENCES se_users (id) ON DELETE SET NULL;
+ALTER TABLE tr_trash_item_translations ADD CONSTRAINT FK_8264DAF45C8D7CA FOREIGN KEY (trashItemId) REFERENCES tr_trash_items (id) ON DELETE CASCADE;
+```
+
+> For MYSQL 5.6 and lower, the `JSON` type of the `restoreData` column must be replaced with `TEXT`. See the [doctrine/dbal type](https://www.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/types.html#json) documentation.
+
+Finally, you need to include the routes of the bundle in your `config/routes/sulu_admin.yaml`:
```
```suggestion
Additionally, you need to include the routes of the bundle in your `config/routes/sulu_admin.yaml`:
```

UPGRADE.md
```diff
@@ -2,6 +2,62 @@
...
+
+```yaml
+sulu_trash_api:
+    resource: "@SuluTrashBundle/Resources/config/routing_api.yml"
+    type: rest
+    prefix: /admin/api
+```
+
+then add the following to the dependencies section of your `assets/admin/package.json`:
```
```suggestion
Furthermore, you need to add the Javascript code of the bundle to the dependencies section in your `assets/admin/package.json`:
```

UPGRADE.md
```diff
@@ -2,6 +2,62 @@
...
+
+```json
+"sulu-trash-bundle": "file:node_modules/@sulu/vendor/sulu/sulu/src/Sulu/Bundle/TrashBundle/Resources/js",
+```
+
+and import the `SuluTrashBundle` assets in your `assets/admin/index.js`:
```
```suggestion
Finally, you need to import the `SuluTrashBundle` Javascript code in your `assets/admin/index.js` file:
```

PR #6136, opened 2021-06-30
Add default sort order for categories

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| License | MIT

#### What's in this PR?

Adds a default sort order to the category `listBuilder`.

#### Why?

Without this default sort order, the result order may differ between identical requests.


Andreas-Schoenefeldt:
It would be lovely to have this merged :)

alexander-schranz:
@Prokyonn would be nice if we add a test by creating `B`, `A`, `C` which will then be outputted as `A`, `B`, `C`

alexander-schranz (approved):

alexander-schranz (approved):

PR #6139, opened 2021-07-02
Fix phpstan errors

luca-rath:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

Fix phpstan errors

#### Why?

```
 ------ --------------------------------------------------------------------- 
  Line   src/Sulu/Bundle/ActivityBundle/Domain/Model/Activity.php             
 ------ --------------------------------------------------------------------- 
  31     Property Sulu\Bundle\ActivityBundle\Domain\Model\Activity::$context  
         type mapping mismatch: database can contain                          
         array|bool|float|int|JsonSerializable|stdClass|string|null but       
         property expects array.                                              
  36     Property Sulu\Bundle\ActivityBundle\Domain\Model\Activity::$payload  
         type mapping mismatch: database can contain                          
         array|bool|float|int|JsonSerializable|stdClass|string|null but       
         property expects array|null.                                         
 ------ --------------------------------------------------------------------- 
```


src/Sulu/Bundle/ActivityBundle/Domain/Model/Activity.php
```diff
@@ -26,12 +26,12 @@ class Activity implements ActivityInterface
     private $type;
 
     /**
-     * @var mixed[]
+     * @var mixed
```
niklasnatter: i think `mixed[]` is the correct type here - why did you change it? 🤔

luca-rath: Since the `phpstan/phpstan-doctrine` `0.12.40` release, the error in the pull request description above occurrs for json fields ... And it's true, the database "could" contain also a `bool` in that column

alexander-schranz: Would add it as ignored error and not add casting to array here.

alexander-schranz: This kind of things is okay to be a runtime error, the checks and casts are unnecessary costs in this cases.

PR #6142, opened 2021-07-02
PreviewRenderer.php bugfix

StaffNowa:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #6100
| Related issues/PRs | #6100
| License | MIT

We do not have webspaceKey, but we have webspace
```
PreviewRenderer.php on line 120:
array:3 [▼
  "locale" => "lt"
  "targetGroup" => "-1"
  "webspace" => "kuro-kainos"
]
```

Error
```
Sulu\Component\Webspace\Manager\WebspaceManager::findPortalInformationsByWebspaceKeyAndLocale(): Argument #1 ($webspaceKey) must be of type string, null given, called in /var/www/project/vendor/sulu/sulu/src/Sulu/Bundle/PreviewBundle/Preview/Renderer/PreviewRenderer.php on line 128
```

### DIFF
--- a/src/Sulu/Bundle/PreviewBundle/Preview/Renderer/PreviewRenderer.php
+++ b/src/Sulu/Bundle/PreviewBundle/Preview/Renderer/PreviewRenderer.php
@@ -115,7 +115,7 @@ class PreviewRenderer implements PreviewRendererInterface
         $partial = false,
         $options = []
     ) {
-        $webspaceKey = $options['webspaceKey'] ?? null;
+        $webspaceKey = $options['webspace'] ?? null;
         $locale = $options['locale'] ?? null;
 
         if (!$this->routeDefaultsProvider->supports(\get_class($object))) {


luca-rath:
Looks like you updated sulu but forgot to update the admin javascript build, could that be possible? Because the [PreviewStore](https://github.com/sulu/sulu/blob/2.x/src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/stores/PreviewStore.js#L49) sends `webspaceKey` instead of `webspace` since sulu 2.3

StaffNowa:
O will try it and answer here 🙄🙃

luca-rath:
When upgrading sulu, always have a look at https://docs.sulu.io/en/2.3/upgrades/upgrade-2.x.html

StaffNowa:
@luca-rath thanks for help after
```bin/console sulu:admin:update-build``` again worsk

PR #6144, opened 2021-07-05
Change Admin Reset Email to text/html

loic-cariou:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| License | MIT

#### What's in this PR?

Allows you to customize the password reset email with an HTML email

### DIFF
--- a/src/Sulu/Bundle/SecurityBundle/Controller/ResettingController.php
+++ b/src/Sulu/Bundle/SecurityBundle/Controller/ResettingController.php
@@ -445,7 +445,7 @@ class ResettingController
             ->setSubject($this->getSubject())
             ->setFrom($from)
             ->setTo($to)
-            ->setBody($this->getMessage($user, $token));
+            ->setBody($this->getMessage($user, $token), 'text/html');
 
         $mailer->send($message);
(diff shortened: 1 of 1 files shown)


src/Sulu/Bundle/SecurityBundle/Controller/ResettingController.php
```diff
@@ -445,7 +445,7 @@ private function sendTokenEmail(UserInterface $user, string $from, string $to, s
             ->setSubject($this->getSubject())
             ->setFrom($from)
             ->setTo($to)
-            ->setBody($this->getMessage($user, $token));
+            ->setBody($this->getMessage($user, $token), 'text/html');
```
alexander-schranz, line 448: If we change this to `text/html` the email template should be updated I think to use `<br/>` instead of a normal line break.

alexander-schranz (changes requested):
@loic-cariou Thank you for your Pull Request. I think there is also a small change in the `reset_password.html.twig` template then need. The current target branch is 2.x which is for the 2.4 release, if you want this to have in earlier patch release please target the `2.2` branch for this. You can rebase to it with: `git rebase --onto origin/2.2 origin/2.x`

loic-cariou:
I will close this pull request to reopen it properly on the right branch

PR #6149, opened 2021-07-09
Add meta and title block to admin main twig file

ymc-thzi:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no 

#### What's in this PR?

Add a block around metadata so that some customers can at least change the hardcoded title or add some custom metadata.


alexander-schranz:
@ymc-thzi Thx for the PR I updated it to match other templates and rebased it to 2.2

PR #6123, opened 2021-06-20
Implement UI for batch delete

luca-rath:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | #6104, #6105, #6109 
| License | MIT
| Documentation PR | -

#### What's in this PR?

Implement UI for batch delete

PR #6155, opened 2021-07-19
Add props for displaying Snackbar inside of Dialog

luca-rath:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | #6123
| License | MIT
| Documentation PR | -

#### What's in this PR?

This pull request adds the functionality for displaying error messages and warnings inside dialogs.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Dialog/Dialog.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Dialog/Dialog.js
@@ -40,20 +40,18 @@ class Dialog extends React.Component<Props> {
 
     @observable open: boolean = false;
     @observable visible: boolean = false;
-    @observable snackbarType: SnackbarType;
 
     constructor(props: Props) {
         super(props);
 
-        const {open, snackbarType} = this.props;
+        const {open} = this.props;
 
         this.open = open;
         this.visible = open;
-        this.snackbarType = snackbarType;
     }
 
     @action componentDidUpdate(prevProps: Props) {
-        const {open, snackbarMessage, snackbarType} = this.props;
+        const {open} = this.props;
 
         if (prevProps.open === false && open === true) {
             this.visible = true;
@@ -63,10 +61,6 @@ class Dialog extends React.Component<Props> {
             afterElementsRendered(action(() => {
                 this.open = open;
             }));
-        }
-
-        if (snackbarMessage && this.snackbarType !== snackbarType) {
-            this.snackbarType = snackbarType;
         }
     }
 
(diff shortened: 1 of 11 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/components/Dialog/Dialog.js
```diff
@@ -56,6 +64,10 @@ class Dialog extends React.Component<Props> {
                 this.open = open;
             }));
         }
+
+        if (snackbarMessage && this.snackbarType !== snackbarType) {
+            this.snackbarType = snackbarType;
```
niklasnatter: whats the reason for storing this in a class variable instead of using the `snackbarType` prop directly? 🤔

src/Sulu/Bundle/AdminBundle/Resources/js/components/Dialog/README.md
```diff
@@ -52,3 +52,44 @@ const onConfirm = () => {
...
+    <button onClick={() => setOpen(true)}>Open dialog</button>
+    <Dialog
+        title="Question?"
+        onCancel={onCancel}
+        onConfirm={onConfirm}
+        cancelText="No"
+        confirmText="Yes"
+        snackbarType={snackbarType}
+        snackbarMessage={
+            snackbarType === 'error'
```
niklasnatter: think I would extract this into a `snackbarMessage` variable above

niklasnatter (commented):
Looks good to me otherwise 🙂

src/Sulu/Bundle/AdminBundle/Resources/js/components/Dialog/Dialog.js
luca-rath: Let's assume the dialog renders a warning. When the user clicks on the snackbar, the `onSnackbarClick()` prop will be called. Then, the container will probably do something like resetting it's `warning` state. From that moment, the container will pass `undefined` for both, `snackbarType` and `snackbarMessage` props to the dialog. But as soon as `snackbarType` is `undefined`, the snackbar would turn red (because the default is `error`) while still swiping up

niklasnatter: makes sense - i think there is a similar scenario for the `message` of the snackbar that is already handled in the `Snackbar` component:
https://github.com/sulu/sulu/blob/0238c55bd6a4d965aa7ebd12b2ba7467f33db957/src/Sulu/Bundle/AdminBundle/Resources/js/components/Snackbar/Snackbar.js#L42-L44

what do you think about moving this check into the `Snackbar` component too? that way both checks would be in the same place nad it might be easier to understand why this is needed 🙂

niklasnatter (approved):

PR #6158, opened 2021-07-23
UserManager dependency ContactManagerInterface in order to ContactMan…

ehibes:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Changing UserManager ContactManager dependency

#### Why?

If we make a new ClassManager in our project, the UserManager class must have a ContactManagerInterface dependency in order to ContactManager.


luca-rath:
Unfortunately it's not enough to just change the typehints, because the `UserManager` calls some methods like `findById()` and `getContactEntityName()` of the `ContactManager`, but those are not defined in the `ContactManagerInterface` ... This would need a little bit more refactoring (and bc breaks, since the `ContactManagerInterface` would need to be adjusted)

For now, if you need to extend the `ContactManager`, it's probably best to extend from sulu's `ContactManager` instead of just implementing the interface

PR #6160, opened 2021-07-26
Use RequestAnalyzer to determine webspace in snippet selection

luca-rath:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | yes
| Fixed tickets | fixes sulu/SuluArticleBundle#190
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

Use RequestAnalyzer to determine webspace in snippet selection

### DIFF
--- a/src/Sulu/Bundle/SnippetBundle/Content/SingleSnippetSelection.php
+++ b/src/Sulu/Bundle/SnippetBundle/Content/SingleSnippetSelection.php
@@ -19,6 +19,7 @@ use Sulu\Component\Content\Compat\PropertyInterface;
 use Sulu\Component\Content\Compat\Structure\PageBridge;
 use Sulu\Component\Content\PreResolvableContentTypeInterface;
 use Sulu\Component\Content\SimpleContentType;
+use Sulu\Component\Webspace\Analyzer\RequestAnalyzerInterface;
 
 class SingleSnippetSelection extends SimpleContentType implements PreResolvableContentTypeInterface
 {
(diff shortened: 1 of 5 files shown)


src/Sulu/Bundle/SnippetBundle/Content/SingleSnippetSelection.php
```diff
@@ -88,7 +100,12 @@ private function resolveSnippet(PropertyInterface $property)
 
         /** @var PageBridge $page */
         $page = $property->getStructure();
+
         $webspaceKey = $page->getWebspaceKey();
+        if ($this->requestAnalyzer) {
+            $webspaceKey = $this->requestAnalyzer->getWebspace()->getKey();
```
alexander-schranz, line 106: This would be a bc break and can cause now that a false default snippet is used instead of the expected structure webspace default snippet. What is the case? Is the `getWebspaceKey` not correctly returned by articles? We should basically avoid accessing request specific things inside a content type as this can end in unexpected results.

alexander-schranz: What we could do is adding a fallback if no webspace key is returned for the structure but not sure about it 🤔

luca-rath: Another possibility would be to just call `$articleBridge->setWebspaceKey()` somewhere

PR #6163, opened 2021-08-02
Remove obsolete .scrutinizer.yml file

niklasnatter:
#### Why?

We moved to GitHub Actions and therefore do not use Scrutinizer CI anymore 🙂

### DIFF
--- a/.github/workflows/test-application.yaml
+++ b/.github/workflows/test-application.yaml
@@ -146,7 +146,7 @@ jobs:
             - name: Run php-cs-fixer
               if: ${{ matrix.php-cs-fixer }}
               run: |
-                  composer global require friendsofphp/php-cs-fixer --prefer-dist --no-interaction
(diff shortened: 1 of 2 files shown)


.github/workflows/test-application.yaml
```diff
@@ -146,7 +146,7 @@ jobs:
             - name: Run php-cs-fixer
               if: ${{ matrix.php-cs-fixer }}
               run: |
-                  composer global require friendsofphp/php-cs-fixer --prefer-dist --no-interaction
+                  composer global require friendsofphp/php-cs-fixer:^2.17 --prefer-dist --no-interaction
```
luca-rath, line 149: Should we consider upgrading to 3.0?

niklasnatter: yes it would be nice to do that at some point. but i would do that in a separate PR 🙂

luca-rath (approved):

PR #6166, opened 2021-08-03
Add generics phpdoc to repositories and update phpstan baseline

niklasnatter:


src/Sulu/Bundle/ContactBundle/Contact/ContactManager.php
```diff
@@ -487,7 +487,7 @@ public function getContact($contact, $locale)
      *
      * @throws EntityNotFoundException
      */
-    public function setMainAccount(Contact $contact, $data)
+    public function setMainAccount(ContactInterface $contact, $data)
```
luca-rath, line 490: BC break

luca-rath (changes requested):

src/Sulu/Bundle/ContactBundle/Contact/ContactManager.php
niklasnatter: i think every `Contact` should be a `ContactInterface`, right?

alexander-schranz: still a bc break when somebody extended the ContactManager, the change only works in the other way from ContactInterface to Contact not from `Contact` to more open `ContactInterface`.

niklasnatter: But it would probably be better to remove the typehint and add the type only as phpdoc - will change that 🙂

luca-rath: yes, but it's a signature change of a inheritable method ... but after thinking about it again, this would only result in a `Declaration incompatible` error, if someone has overridden the `setMainAccount()` method in a custom `ContactManager` extending sulu's `ContactManager`, which is most likely not the case

luca-rath: Removing the typehint also results in a `Declaration incompatible` error, btw

niklasnatter: You are right, I dont think its possible to change the signature without breaking projects that overwrote this method. On the other hand, the current typehint is definately wrong. I dont have a strong opinion what to do - shall I just revert the change?

luca-rath: Like I said before, I think it's very unlikely that someone has overridden this specific method, so doing this bc break here would be okay for me ... but it's still a bc break and I'm not the one deciding this 🙂 /cc @sulu/core-team

luca-rath (approved):

PR #6168, opened 2021-08-03
Reload items of current page after deleting selected items via ListStore

### LINKED ISSUE #6140: Unable to navigate to other pages after deleting all entries of one page in a list (e.g. snippets)
| Q | A
| --- | ---
| Sulu Version | 2.3.1
| PHP Version | 7.4.16
| DB Version | 5.7.32
| Browser Version | Chrome 91.0.4472.114

# Actual Behavior

After deleting all items of a page in a list view (e.g. snippets) the pagination navigation is not shown any more:

snippets-deleted

The pagination navigation is shown again after a refresh of the page:

snippets-reload

# Expected Behavior

The pagination toolbar should be shown after deleting all items of a page. It would be nice if not only the toolbar is shown but if remaining items from another page would be shown. E.g. if I delete all items from page 1 I want to see all items of page 2 afterwards.

# Steps to Reproduce

1. Create 11 snippets.
2. Use the checkbox in the table header of the list to select all 10 items of the first page.
3. Delete the snippets.
4. The pagination navigation ist not visible any more and also the remaining 1 snippet is not shown.

### LINKED ISSUE #6154: media items reload per page after batch delete
| Q | A
| --- | ---
| Sulu Version | 2.3.1
| PHP Version | 7.4.16
| DB Version | MySQL 5.7.35
| Browser Version | Firefox 89.0.2

# Actual Behavior
I'm in a media folder with more than 100 items.
It shows me 10 items per page on default.
If I select one or more items and delete them, then there is no reload like a refill until the page maximum is reached.
Also, if I select and delete all, in my example all 10 items, pagination disappears! So I have to do a page reload.

# Expected Behavior

After the batch delete, it should also refill the items per page and update the pagination.
Maybe a site reload is easier!? what happen on a error with the message...

# Steps to Reproduce

Bildschirmfoto 2021-07-15 um 18 43 28

niklasnatter:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Fixed tickets | fixes #6140, fixes #6154
| License | MIT

#### What's in this PR?

This PR adjusts the `ListStore` to reload the items for the currently loaded page after deleting the selected items.

#### Why?

See #6140 and #6154

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Pagination/Pagination.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Pagination/Pagination.js
@@ -33,23 +33,20 @@ class Pagination extends React.Component<Props> {
     @action componentDidMount() {
         const {currentPage} = this.props;
 
-        if (!currentPage) {
-            return;
-        }
-
-        this.currentInputValue = this.props.currentPage;
+        this.currentInputValue = currentPage;
+        this.validateAndSubmitInputValue();
     }
 
     @action componentDidUpdate(prevProps: Props) {
-        const {currentPage} = this.props;
+        const {currentPage, totalPages} = this.props;
 
         if (prevProps.currentPage !== currentPage) {
-            if (!currentPage) {
-                this.currentInputValue = 1;
-                return;
-            }
-
             this.currentInputValue = currentPage;
+            this.validateAndSubmitInputValue();
+        }
+
+        if (prevProps.totalPages !== totalPages) {
+            this.validateAndSubmitInputValue();
         }
     }
 
@@ -112,16 +109,16 @@ class Pagination extends React.Component<Props> {
     };
 
     handleInputBlur = () => {
-        this.submitInputValue();
+        this.validateAndSubmitInputValue();
     };
 
     handleInputKeyPress = (key: ?string) => {
         if (key === 'Enter') {
-            this.submitInputValue();
+            this.validateAndSubmitInputValue();
         }
     };
 
-    @action submitInputValue = () => {
+    @action validateAndSubmitInputValue = () => {
         const {currentPage, onPageChange, totalPages} = this.props;
         let page = this.currentInputValue;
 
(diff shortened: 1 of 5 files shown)


luca-rath:
When you're on the last page and you delete all items of that page, the pagination says `Page [2] of 1` and the list remains empty

niklasnatter:
Thats right. I thought about implementing a redirect to the previous page, but the previous page contains more elements than the page that is displayed before the deletion. It felt kind of strange to see more elements after the deletion than before. I think displaying an empty page feels more natural (and is easier to implement 😉 ), but its okay for me to change it too.

luca-rath:
@nnatter I don't have a strong opinion about that, so for me it's also okay to keep it like that ... But in my opinion, `page 2 of 1` is an invalid state and should therefore not be possible.

/cc @sulu/core-team

src/Sulu/Bundle/AdminBundle/Resources/js/components/Pagination/Pagination.js
```diff
@@ -33,23 +33,20 @@ class Pagination extends React.Component<Props> {
...
-                this.currentInputValue = 1;
-                return;
-            }
-
             this.currentInputValue = currentPage;
+            this.validateAndSubmitInputValue();
+        }
+
+        if (prevProps.totalPages !== totalPages) {
+            this.validateAndSubmitInputValue();
```
niklasnatter, line 49: not excited to do this here, but i have not found a better place 😕 

doing it directly in the `ListStore` is not possible because we do not want to do this for adapters that use the `InfiniteLoadingStrategy` (eg. the `MediaCardAdapter`).

doing it in the `DefaultLoadingStrategy` is not possible because the loading strategy cannot access the `ListStore`.

niklasnatter:
I cannot reproduce the `lint:js` error on my local machine at the moment. Hopefully this is just a hiccup that will sort itself out on its own 🙈

src/Sulu/Bundle/AdminBundle/Resources/js/views/List/toolbarActions/UploadToolbarAction.js
```diff
@@ -144,7 +145,7 @@ export default class UploadToolbarAction extends AbstractListToolbarAction {
         this.list.errors = [...this.list.errors, error];
     };
 
-    handleError = (fileRejections: FileRejection[]) => {
+    handleError = (fileRejections: any[]) => {
```
niklasnatter, line 148: the types of the `react-dropzone` package are typescript only unfortunately 😕

luca-rath (approved):

PR #6170, opened 2021-08-03
Use case-insensitive comparison when comparing old and new email in UserManager

### LINKED ISSUE #5596: Check of Email uniquness in UserManager::processEmail should be case insensitiv

niklasnatter:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Fixed tickets | fixes #5596
| License | MIT

#### What's in this PR?

This PR adjusts the `UserManager` service to use a case-insensitive comparison for comparing the old and the new email address when changing the email address of a user.

#### Why?

See #5596

### DIFF
--- a/src/Sulu/Bundle/SecurityBundle/Tests/Functional/Controller/ProfileControllerTest.php
+++ b/src/Sulu/Bundle/SecurityBundle/Tests/Functional/Controller/ProfileControllerTest.php
@@ -11,6 +11,8 @@
 
 namespace Sulu\Bundle\SecurityBundle\Tests\Functional\Controller;
 
+use Sulu\Bundle\ContactBundle\Entity\Contact;
+use Sulu\Bundle\SecurityBundle\Entity\User;
 use Sulu\Bundle\TestBundle\Testing\SuluTestCase;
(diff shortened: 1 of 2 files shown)


luca-rath:
@nnatter The `PHP Lint` workflow failed

luca-rath:
@nnatter Not sure why the workflows aren't starting, can you push again?

niklasnatter:
@luca-rath Pushed again, but looks like the problem persists. Guess I will try again tomorrow 🙈

nnatter:
@luca-rath Looks like they were completed now 🙂

luca-rath (approved):

PR #6176, opened 2021-08-10
Update friendsofphp/php-cs-fixer package to version 3

niklasnatter:
#### What's in this PR?

This PR updates the `friendsofphp/php-cs-fixer` to version 3. The currently used version 2.19 is deprecated.

See https://github.com/FriendsOfPHP/PHP-CS-Fixer/releases/tag/v3.0.0

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/DependencyInjection/Compiler/AddAdminPass.php
+++ b/src/Sulu/Bundle/AdminBundle/DependencyInjection/Compiler/AddAdminPass.php
@@ -19,7 +19,7 @@ use Symfony\Component\DependencyInjection\ContainerBuilder;
  */
 class AddAdminPass implements CompilerPassInterface
 {
-    const ADMIN_TAG = 'sulu.admin';
+    public const ADMIN_TAG = 'sulu.admin';
 
     public function process(ContainerBuilder $container)
     {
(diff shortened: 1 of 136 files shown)


src/Sulu/Bundle/AdminBundle/DependencyInjection/Compiler/AddAdminPass.php
```diff
@@ -19,7 +19,7 @@
  */
 class AddAdminPass implements CompilerPassInterface
 {
-    const ADMIN_TAG = 'sulu.admin';
```
niklasnatter, line 22: this change is not compatible with php 5.6. not sure if it is possible to configure `php-cs-fixer` to be compatible with older php versions

niklasnatter: okay looks like `php-cs-fixer` will target the php version that is used for running it. as the new `php-cs-fixer` version is not compatible with php versions smaller than 7.1, we cannot automatically fix the `1.6` branch without breaking php 5.6 compatibility 😕

niklasnatter:
Closing this because of https://github.com/sulu/sulu/pull/6176/files#r685881238

PR #6183, opened 2021-08-17
Move project specific javascript code into separate app.js file

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? | no 
| BC breaks? | no 
| Deprecations? | no
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Cleanup node_modules directories always to avoid that after the update still a package-lock or node modules is there.
Add the `app.js` which can be modified but when editited it will do a manual build.

#### Why?

Improve sulu update for custom JS.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Command/UpdateBuildCommand.php
+++ b/src/Sulu/Bundle/AdminBundle/Command/UpdateBuildCommand.php
@@ -16,6 +16,7 @@ use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Style\SymfonyStyle;
 use Symfony\Component\Filesystem\Filesystem;
+use Symfony\Component\HttpClient\Exception\ClientException;
 use Symfony\Component\Process\Process;
 use Symfony\Contracts\HttpClient\HttpClientInterface;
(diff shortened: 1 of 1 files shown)


wachterjohannes (approved):

src/Sulu/Bundle/AdminBundle/Command/UpdateBuildCommand.php
```diff
@@ -239,9 +257,14 @@ private function getLocalFile(string $path)
...
     {
         $path = \str_replace(\DIRECTORY_SEPARATOR, '/', $path);
-        $response = $this->httpClient->request('GET', $remoteRepository . $path);
 
-        return $response->getContent();
+        try {
+            $response = $this->httpClient->request('GET', $remoteRepository . $path);
+
+            return $response->getContent();
+        } catch (ClientException $e) {
```
niklasnatter, line 265: can we catch a more specific exception here? 🤔 
think this could lead to confusing behaviour if an error happens because of an unreliable internet connection

alexander-schranz: The exception represents 4xx status code.

PR #6186, opened 2021-08-18
Use generic DropdownToolbarAction instead of SaveWithPublishingToolbarAction in PageAdmin

niklasnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| License | MIT

#### What's in this PR?

This PR adjusts the `PageAdmin` to use a generic `DropdownToolbarAction` instead of the specific `SaveWithPublishingToolbarAction`. To do this, the PR adds a reusable `PublishToolbarAction`.

#### Why?

Using a `DropdownToolbarAction` makes it easier to add an additional item to the dropdown. Also, this change forces us to implement all dropdown options (save as draft, save and publish, publish) in a reusable manner. This makes it easier to use this functionality in other places.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/views/Form/toolbarActions/DropdownToolbarAction.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/views/Form/toolbarActions/DropdownToolbarAction.js
@@ -83,9 +83,10 @@ export default class DropdownToolbarAction extends AbstractFormToolbarAction {
             throw new Error('The "icon" option must be a string!');
         }
 
+        // use "Boolean" to filter undefined: https://github.com/facebook/flow/issues/1414#issuecomment-251422935
         const childToolbarItemConfigs: Array<ToolbarItemConfig<*>> = this.toolbarActions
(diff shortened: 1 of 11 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/views/Form/toolbarActions/DropdownToolbarAction.js
```diff
@@ -82,41 +83,40 @@ export default class DropdownToolbarAction extends AbstractFormToolbarAction {
             throw new Error('The "icon" option must be a string!');
         }
 
-        const options: Array<DropdownOption> = this.toolbarActions
-            .reduce((toolbarActions, toolbarAction) => {
-                const toolbarItemConfig = toolbarAction.getToolbarItemConfig();
+        // use "Boolean" to filter undefined: https://github.com/facebook/flow/issues/1414#issuecomment-251422935
```
alexander-schranz, line 86: 🙈

alexander-schranz (approved):

Issue #5441, opened 2020-07-31
Supplying * asterix into tag selection UI fails with blank screen


mad-wnn:
| Q | A
| --- | ---
| Sulu Version | 2.1.0
| Browser Version | Firefox 68.2

# Actual Behavior
supplying `*` into `tag_selection` UI fails with blank screen

browser console output:
```
SyntaxError: "nothing to repeat"
    <anonymous> Suggestion.js:28
    value AutoCompletePopover.js:116
    value AutoCompletePopover.js:114
    value Suggestion.js:82
```

alexander-schranz:
@mad-wnn thank you for report. This seems to be also an issue in the 2.0 version. The `*` seems not be escaped for reges at https://github.com/sulu/sulu/blob/bfc18fd0a2a37db25c16ac9d21c76ee984ae7cec/src/Sulu/Bundle/AdminBundle/Resources/js/components/AutoCompletePopover/Suggestion.js#L28. And that why it fails.

I think we should use [replaceAll](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll) instead of a regular expression here, would require to core-js polyfill as it seems Edge does not yet support it. /cc @danrot what do you think? A package for escape string for regex seems to be an overhead for me and replaceAll polyfill would be something we could remove in future. 

Do you want to create a pull request on the `release/2.0` branch to fix this?

danrot:
Sorry, but I cannot even reproduce that error. I have tried to enter stuff like "asdf*" or only "*" into the tags field in the page excerpt tab, and at least there it seems to work for me. Am I missing something? Tried it in Chrome and Firefox.

alexander-schranz:
@danrot You need first create a Tag with has a asterix inside it over `Settings` -> `Tags` and then go back on a page and search for `*`. The suggestion.js will try to highlight the `*` in that autocomplete dropdown and will crash.

danrot:
Okay, got it, I get the following error now, that makes totally sense:

```
SyntaxError: Invalid regular expression: /*/: Nothing to repeat
```

What @alexander-schranz suggested sounds good to me, I don't see why we shouldn't be able to use `replaceAll`.

mad-wnn:
Hello.May look into it in 3 days for more insight as I haven't got any tags containing asterixes, just ASCII strings.

alexander-schranz:
This issue seems still exist and seems not be fixed by #6191.

Reproducer:

[...]

PR #6191, opened 2021-08-23
Display suggestions when focusing input field of autocomplete even without search term

niklasnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Fixed tickets | fixes https://github.com/sulu/sulu/issues/5441
| License | MIT

#### What's in this PR?

This PR adjusts the behaviour of the `MultiAutoComplete` and `SingleAutoComplete` component to display suggestions as soon as the user focuses the respective input field. At the moment, suggestions are displayed only after the user have entered a search term that consists of at least one character.

#### Why?

Displaying suggestions as soon as the input field is focused shows the user how possible results look like. This makes it easier to formulate an appropriate search term. Moreover, if there are not man options, the initial suggestions will already contain the desired option and the user is able to select it without entering any search term.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/AutoCompletePopover/AutoCompletePopover.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/AutoCompletePopover/AutoCompletePopover.js
@@ -12,6 +12,7 @@ type Props = {
     anchorElement: ElementRef<*>,
     idProperty: string,
     minWidth: number,
+    onClose?: () => void,
     onSelect: (suggestion: Object) => void,
     open: boolean,
     query: ?string,
@@ -44,9 +45,8 @@ export default class AutoCompletePopover extends React.Component<Props> {
     };
 
     handlePopoverClose = () => {
-        const {onSelect, suggestions} = this.props;
-        if (suggestions.length > 0) {
-            onSelect(suggestions[0]);
+        if (this.props.onClose) {
+            this.props.onClose();
         }
     };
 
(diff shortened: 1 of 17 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/components/AutoCompletePopover/AutoCompletePopover.js
```diff
@@ -44,9 +45,8 @@ export default class AutoCompletePopover extends React.Component<Props> {
     };
 
     handlePopoverClose = () => {
-        const {onSelect, suggestions} = this.props;
-        if (suggestions.length > 0) {
```
niklasnatter, line 48: closing the autocomplete popover by clicking somewhere else selected the first suggestion previously. with the changes in this PR, this would automatically select the first suggestion when a user clicks outside of the popover after he focused the input field once.

src/Sulu/Bundle/AdminBundle/Resources/js/components/Popover/Popover.js
```diff
@@ -54,22 +57,38 @@ class Popover extends React.Component<Props> {
             this.setPopoverSize(0, 0);
             this.updateDimensions();
         });
+
+        if (this.props.open) {
```
niklasnatter, line 61: this implementation corresponds to the implementation in the `Overlay` component: https://github.com/sulu/sulu/blob/a00ad7ad1a1c5cffe245c586dbcd360e4beabff8/src/Sulu/Bundle/AdminBundle/Resources/js/components/Overlay/Overlay.js#L55-L57

src/Sulu/Bundle/AdminBundle/Resources/js/components/MultiAutoComplete/MultiAutoComplete.js
```diff
@@ -41,8 +42,14 @@ class MultiAutoComplete extends React.Component<Props> {
 
     @observable labelRef: ElementRef<'label'>;
     @observable inputRef: ElementRef<'input'>;
+
+    @observable displaySuggestions = false;
```
niklasnatter, line 45: the suggestion list will be closed when clicking outside of the popover or when pressing `Escape`. it will be opened again when the search term in the input field is changed

alexander-schranz (approved):

PR #6193, opened 2021-08-24
Fix username for password_resetted event in activity list

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? |  yes
| New feature? | no 
| BC breaks? | no 
| Deprecations? | no  
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Fix activity list without resourceTitle.

#### Why?

It can be that there is no resourceTitle which can cause the method `getLocalizedValue` to crash as it expects a string instead of null.

### DIFF
--- a/src/Sulu/Bundle/ActivityBundle/UserInterface/Controller/ActivityController.php
+++ b/src/Sulu/Bundle/ActivityBundle/UserInterface/Controller/ActivityController.php
@@ -383,7 +383,7 @@ class ActivityController extends AbstractRestController implements ClassResource
             : $this->translator->trans('sulu_activity.someone', [], 'admin', $translationLocale);
 
         $translationParameters['{resourceTitle}'] = $this->getLocalizedValue(
-            $activity['resourceTitle'] ?? '',
+            $activity['resourceTitle'],
             $activity['resourceTitleLocale'] ?? null,
             $translationLocale
         );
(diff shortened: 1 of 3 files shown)


src/Sulu/Bundle/ActivityBundle/UserInterface/Controller/ActivityController.php
```diff
@@ -383,7 +383,7 @@ private function getActivityDescription(array $activity, string $translationLoca
             : $this->translator->trans('sulu_activity.someone', [], 'admin', $translationLocale);
 
         $translationParameters['{resourceTitle}'] = $this->getLocalizedValue(
-            $activity['resourceTitle'],
+            $activity['resourceTitle'] ?? '',
```
niklasnatter: maybe we should adjust the typehint of the `getLocalizedValue` instead of passing an empty string here? 🤔

niklasnatter:
Maybe we can add the following methods to the `UserPasswordResettedEvent` to prevent this case for the `password_resetted` event in the future:

```
    public function getResourceUser(): UserInterface
    {
        return $this->resourceUser;
    }

    public function getResourceTitle(): ?string
    {
        return $this->resourceUser->getUsername();
    }
```

src/Sulu/Bundle/ActivityBundle/UserInterface/Controller/ActivityController.php
alexander-schranz: updated!

alexander-schranz:
@nnatter fixed also the password reset event containing now the username

PR #6196, opened 2021-08-24
Fix check for ext-zip extension in UpdateBuildCommand

niklasnatter:


alexander-schranz:
Fixes error from #6183

PR #6205, opened 2021-09-02
Include properties inside of nested blocks in search index

mgwestwerk:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

This PR allows properties inside nested blocks to be used as search fields

#### Why?

Currently only properties that are one level deep inside a block will be indexed if marked as search field


niklasnatter (approved):
I do not have a lot of experience with this part of the code. But from my perspective, the changed code looks good! 

I have tested the changes on my machine and the result looks correct to me 🙂  
For completion, the following picture shows a comparison between an indexed page before (left) and after (right) the changes in this PR. 

Screenshot 2021-09-30 at 11 56 12

Thanks a lot for working on this!

PR #6208, opened 2021-09-02
Add PHP 8.1 Support

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? |  yes
| BC breaks? | no 
| Deprecations? | no  
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Add support for PHP 8.1

#### Why?

The first RC of PHP 8.1 was released https://www.php.net/archive/2021.php#2021-09-02-1, and it would great if sulu could support PHP 8.1. This PR currently should tell us where we need todo some work to get compatible with PHP 8.1.

#### Checklist:

Packages:

 - [x] composer/composer (https://github.com/composer/composer/pull/10088)
 - [x] phpspec/prophecy (https://github.com/phpspec/prophecy/pull/533) (1.14.0)
 - [x] laminas/laminas-code (https://github.com/laminas/laminas-code/pull/86)
    - [x] 4.5 release
 - [x] ~~https://github.com/goodby/csv/pull/87 (probably unmaintained)~~
   - [x] Replaced by fork https://github.com/handcraftedinthealps/goodby-csv
 - [x] MassiveSearch Metadata Classes: https://github.com/massiveart/MassiveSearchBundle/blob/7e9c329a85a8baaca6bd299e050a1209f0915d04/Search/Metadata/ClassMetadata.php#L81-L93
 
 Nice to have:
 
 - [x] friendsofphp/php-cs-fixer (https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/5945) (not required)
    - [x] https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/5993

 Changes:
 
 - [x] User Serializable (https://php.watch/versions/8.1/serializable-deprecated)  (https://github.com/sulu/sulu/pull/6208/commits/74cf4fd217dcbdc0ab2a8ed58723b8a640aa1df8)
 - [x] No optional Parameter before required (https://github.com/sulu/sulu/pull/6208/commits/9a6a7ead70e71b806e507c8bc274b7fe2536b07d)


src/Sulu/Bundle/SecurityBundle/Entity/User.php
```diff
@@ -306,17 +306,22 @@ public function eraseCredentials()
...
-                $this->salt,
-                $this->username,
-                $this->locked,
-                $this->enabled,
-            ]
+            $this->__serialize()
         );
     }
 
+    public function __serialize(): array
```
alexander-schranz: Actually we could also remove the deprecated `Serializable` interface here and remove the `serialize` and `unserialize` methods. They are in our case not longer used as we did add the `EquatableInterface` in the symfony 5 upgrade to check equality between the session serialized user and the loaded user. Removing it would be a small bc break. @sulu/core-team what do you think?

niklasnatter: Would be okay for me to remove the `Serializable` interface from the `User` entity 🙂

composer.json
```diff
@@ -51,7 +61,7 @@
         "friendsofsymfony/jsrouting-bundle": "^2.3",
         "friendsofsymfony/rest-bundle": "^2.8 || ^3.0",
         "gedmo/doctrine-extensions": "^2.4 || ^3.0",
-        "goodby/csv": "^1.3",
+        "goodby/csv": "dev-patch-2 as 1.4.0",
```
alexander-schranz: As I think goodby csv is unmaintained and we will not get a release for this fix we have to create a decission here. The package is only used at this place in our code: https://github.com/sulu/sulu/blob/2.x/src/Sulu/Component/Rest/Csv/CsvHandler.php

A. Replace CSV Handler with a raw fputcsv and so implement that logic ourself.
B. Fork the goodby/csv and create a compatible version.

/cc @wachterjohannes what would you do?

wachterjohannes: @alexander-schranz i would try to replace it with standard csv functions of php - `fputcsv` for example - the streaming functionality of `goodby/csv` cant be really used by the the handler because we load all the data in front. so throw it away 😂

alexander-schranz: As discussed we fork it to https://github.com/handcraftedinthealps/goodby-csv.

src/Sulu/Bundle/SecurityBundle/Entity/User.php
alexander-schranz: Is removed

.github/workflows/test-application.yaml
```diff
@@ -86,7 +90,7 @@ jobs:
                       phpcr-transport: jackrabbit
                       dependency-versions: 'lowest'
                       php-extensions: 'ctype, iconv, mysql, gd'
-                      tools: 'composer:v1'
+                      tools: 'composer:v2'
```
alexander-schranz: required to install new packages: https://blog.packagist.com/deprecating-composer-1-support/

/cc @chirimoya

alexander-schranz: To make it clear this would mean newer versions can only be installed when using composer v2. What do you think?

alexander-schranz: As discussed with @chirimoya we will fix the 8.1 issues on our side in the 2.2 and 2.3 branches. But only the 2.4 branch we will use the fork in 2.2 and 2.3. So with begin of Sulu 2.4 we require composer 2.

chirimoya (approved):

alexander-schranz:
For all subscribing this PR sulu itself will be compatible to PHP 8.1 with the next Versions of 2.2 and 2.3. But if you want to use PHP 8.1 in this projects you need to require an additional package:

```
composer require handcraftedinthealps/goodby-csv:^1.4
```

Which will provide PHP 8.1 support for `goodby/csv` package which we forked. Begin from Version 2.4 we will require that package. As it is a new Packagist package from the Begin of Sulu 2.4 sulu is only compatible with Composer 2. See the following blog post by the composer maintainers: https://blog.packagist.com/deprecating-composer-1-support/.

alexander-schranz:
[Laminas-code 4.5](https://github.com/laminas/laminas-code/releases) is now also released so its now 8.1 can be used with stable packages.

PR #6209, opened 2021-09-02
Upgrade font-awesome from 4.7 to 5.15

alexandersch:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | yes
| License | MIT

#### What's in this PR?

Upgraded the font awesome library from version 4 to stable version 5. I've followed the [upgrade docs](https://fontawesome.com/v5.15/how-to-use/on-the-web/setup/upgrading-from-version-4) from font awesome themselves. I've made the `solid` and `brand` styles available. Version 4 just had one prefix — fa. Version 5 has four prefixes. 2 of them are free for commercial use: `fas` and `fab`. 

I've also added v4-shims to enable backwards compatibility for icons without prefix and changed icon names. This way it won't be a BC but just a deprecation of v4 icons. 

A list of icons that have changes (name or prefix) can be found [here](https://fontawesome.com/v5.15/how-to-use/on-the-web/setup/upgrading-from-version-4#name-changes).

#### Why?

Because new icons could not be used. Also upgrading to version 6 should be a piece of cake now when it comes available on npm. Version 6 is still in beta though.

#### Example Usage

```php
// If you want to use a font awesome icon with a 'fas' prefix
$navigationItem->setIcon('fas fa-door-open');

// If you want to use a font awesome icon with a 'fab' prefix
$navigationItem->setIcon('fab fa-facebook');

// Backwards compatible without 'fas' or 'fab' prefix
$navigationItem->setIcon('fa-facebook-f'); 
```

#### To Do

### DIFF
--- a/src/Sulu/Bundle/CategoryBundle/DependencyInjection/SuluCategoryExtension.php
+++ b/src/Sulu/Bundle/CategoryBundle/DependencyInjection/SuluCategoryExtension.php
@@ -153,7 +153,7 @@ class SuluCategoryExtension extends Extension implements PrependExtensionInterfa
                                         'list_key' => 'categories',
                                         'display_properties' => ['name'],
                                         'empty_text' => 'sulu_category.no_category_selected',
-                                        'icon' => 'fa-tags',
+                                        'icon' => 'su-tag',
                                         'overlay_title' => 'sulu_category.single_category_selection_overlay_title',
                                     ],
                                 ],
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Icon/Icon.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Icon/Icon.js
@@ -1,5 +1,6 @@
 // @flow
-import 'font-awesome/css/font-awesome.min.css';
+import '@fortawesome/fontawesome-free/js/all.js';
+import '@fortawesome/fontawesome-free/js/v4-shims.js';
 import './sulu-icon.css';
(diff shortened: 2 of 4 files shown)


src/Sulu/Bundle/CategoryBundle/DependencyInjection/SuluCategoryExtension.php
```diff
@@ -153,7 +153,7 @@ public function prepend(ContainerBuilder $container)
                                         'list_key' => 'categories',
                                         'display_properties' => ['name'],
                                         'empty_text' => 'sulu_category.no_category_selected',
-                                        'icon' => 'fa-tags',
+                                        'icon' => 'su-tag',
```
niklasnatter, line 156: i noticed that we use the fontawesome icon instead of the sulu icon (that is already used for the [search index](https://github.com/sulu/sulu/blob/a31dab9f85a446d1a315eb0be639c9ceaca6aa70/src/Sulu/Bundle/CategoryBundle/DependencyInjection/SuluCategoryExtension.php#L174))

alexander-schranz: Interesting, yeah we should use in the core also our own icons. Good catch 👍

alexander-schranz (approved):

niklasnatter (approved):
Thanks for spending your time on this!

PR #6210, opened 2021-09-03
Set route path property on live node in RoutableSubscriber when article is published

### LINKED ISSUE #593: RoutePath not saved with articles
| Q | A
| --- | ---
| Bug? | yes
| New Feature? | no
| SuluArticleBundle Version | 2.3.1
| Sulu Version | 2.3.1, 2.3.2 
| PHP Version | 8.0.9
| Elasticsearch | 7.14.0
| DB Version | MySQL 8.0.23
| Browser Version | Chromium 92.0.4515

#### Actual Behavior
After reindexing the articles the frontend crashes for articles where no resourcelocator was set.

`Sulu\Bundle\ArticleBundle\Content\ArticleResourceItem::getRoutePath(): Return value must be of type string, null returned`

#### Expected Behavior

A reindex should not affect the resource locator.

#### Steps to Reproduce

- Add a new article in the Sulu backend:
  - Leave the URL field empty
  - Save the article

- Reindex the articles:
  - `bin/adminconsole sulu:article:reindex --drop --no-interaction`
  - `bin/websiteconsole sulu:article:reindex --drop --no-interaction`

- Open a page in the frontend where the `{{ routePath }}` is used.
  - The article has no routePath (HTTP 500)

#### Furthermore

The PHPCR node properties differ between the `live` and the `default` version.
Among other things there is no `routePath` property in the subset of the live session.

To reproduce this:
- Save an article without URL
- Copy its UUID
- Execute the following commands and compare the properties:
  - `bin/console  doctrine:phpcr:node:dump --identifiers "your articles UUID" --props --session default`
  - `bin/console  doctrine:phpcr:node:dump --identifiers "your articles UUID" --props --session live`
- There should be no `routePath` in the `live` version.

niklasnatter:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no yes
| Fixed tickets | fixes https://github.com/sulu/SuluArticleBundle/issues/593
| License | MIT>

#### What's in this PR?

This PR adjusts the `RoutableSubscriber` to set the route path property on the live node instead of the draft node in the `handlePublish` method. This is done by using the node of the event instead of fetching the node via the `DocumentInspector::getNode()` method.

#### Why?

At the moment, the route path property is only saved on the draft node if an article is created and published in the same step. See https://github.com/sulu/SuluArticleBundle/issues/593 for more information.

### DIFF
--- a/src/Sulu/Bundle/RouteBundle/Document/Subscriber/RoutableSubscriber.php
+++ b/src/Sulu/Bundle/RouteBundle/Document/Subscriber/RoutableSubscriber.php
@@ -212,8 +212,6 @@ class RoutableSubscriber implements EventSubscriberInterface
             return;
         }
 
-        $node = $this->documentInspector->getNode($document);
-
         try {
             $route = $this->createOrUpdateRoute($document, $event->getLocale());
         } catch (RouteIsNotUniqueException $exception) {
@@ -223,7 +221,7 @@ class RoutableSubscriber implements EventSubscriberInterface
         $document->setRoutePath($route->getPath());
         $this->entityManager->persist($route);
 
-        $node->setProperty(
+        $event->getNode()->setProperty(
             $this->getRoutePathPropertyName($document, $event->getLocale()),
             $route->getPath()
         );
@@ -257,10 +255,6 @@ class RoutableSubscriber implements EventSubscriberInterface
                 $this->getRoutePathPropertyName($localizedDocument, $locale),
                 $route->getPath()
             );
-
-            $propertyName = $this->getRoutePathPropertyName($localizedDocument, $locale);
-            $node = $this->documentInspector->getNode($localizedDocument);
-            $node->setProperty($propertyName, $route->getPath());
         }
     }
 


src/Sulu/Bundle/RouteBundle/Document/Subscriber/RoutableSubscriber.php
```diff
@@ -257,10 +255,6 @@ public function handleCopy(CopyEvent $event): void
                 $this->getRoutePathPropertyName($localizedDocument, $locale),
                 $route->getPath()
             );
-
-            $propertyName = $this->getRoutePathPropertyName($localizedDocument, $locale);
```
niklasnatter, line 261: exactly the same logic is executed 5 lines above

src/Sulu/Bundle/RouteBundle/Document/Subscriber/RoutableSubscriber.php
```diff
@@ -223,7 +221,7 @@ public function handlePublish(PublishEvent $event): void
         $document->setRoutePath($route->getPath());
         $this->entityManager->persist($route);
 
-        $node->setProperty(
```
niklasnatter, line 226: the `PublishSubscriber` sets the live node to the event. all other subscribers should get the node from the event to process the correct node:
https://github.com/sulu/sulu/blob/59d69249772a78634505aae92152af91e6d86457/src/Sulu/Bundle/PageBundle/Document/Subscriber/PublishSubscriber.php#L79

alexander-schranz (approved):

PR #6213, opened 2021-09-07
Integrate the TrashBundle functionality into the CategoryBundle

niklasnatter:
| Q | A
| --- | ---
| New feature? | yes
| Related issues/PRs | #5150 #6132 
| License | MIT

#### What's in this PR?

This PR integrates the `SuluTrashBundle` into the `SuluCategoryBundle`. With the changes in this PR, categories will be moved into the trash when being deleted and can be restored from the trash at a later point in time.


src/Sulu/Bundle/CategoryBundle/Trash/CategoryTrashItemHandler.php
```diff
@@ -0,0 +1,234 @@
...
+            $categoryTitles,
+            CategoryAdmin::SECURITY_CONTEXT,
+            null,
+            null
+        );
+    }
+
+    public function restore(TrashItemInterface $trashItem, array $restoreFormData): object
+    {
+        $data = $trashItem->getRestoreData();
```
niklasnatter, line 181: i thought about moving this logic into the `CategoryManager` (similar to the `TagManager::restore` method). but that way the `CategoryManager` would need to know about the structure of the data returned by `$trashItem->getRestoreData()`. keeping the restore logic here means that the `CategoryTrashItemHandler` is the only class that needs to know the structure of this data.

src/Sulu/Bundle/TagBundle/Trash/TagTrashItemHandler.php
```diff
@@ -59,6 +60,7 @@ public function store(object $tag): TrashItemInterface
 
     public function restore(TrashItemInterface $trashItem, array $restoreFormData): object
     {
+        // TODO: maybe inline functionality of $tagManager->restore() to keep handling of $trashItem->getRestoreData() in this class?
```
niklasnatter: see https://github.com/sulu/sulu/pull/6213/files#r703604625

src/Sulu/Bundle/CategoryBundle/Trash/CategoryTrashItemHandler.php
alexander-schranz: I would keep it this way and not move it to the managers. So all about the trash is in this service and only the store method need to be called in the remove method. Also the CategoryTrashItemHandler should just not be registered when Trash is available. And in future the RemoveHandler should also just call the store method.

niklasnatter: sounds good to me 🙂  i will remove the `TagManager::restore` too, i dont think that it is a method should be used outside of the `TagTrashItemHandler `

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FormOverlay/FormOverlay.js
```diff
@@ -34,12 +34,6 @@ class FormOverlay extends React.Component<Props> {
 
     @observable formErrors: Array<string> = [];
 
-    @computed get confirmDisabled() {
-        const {confirmDisabled, formStore} = this.props;
-
-        return confirmDisabled || !formStore.dirty;
```
niklasnatter, line 40: i figured that disabling the confirm button while the formstore is not empty is not the correct behaviour in every case. eg. wen restoring a category, i want to be able to confirm the overlay without selecting a parent category. 

because of this, i think we should remove this here. if we need this behaviour somewhere, we can still set the `confirmDisabled` prop from the outside.

src/Sulu/Bundle/CategoryBundle/Entity/CategoryInterface.php
```diff
@@ -236,6 +236,14 @@ public function getParent();
      */
     public function setCreator(UserInterface $creator = null);
 
+    /**
+     * Set created.
+     * Note: This property is set automatically by the TimestampableSubscriber if not set manually.
+     *
+     * @return CategoryInterface
+     */
+    public function setCreated(\DateTime $created);
```
luca-rath, line 245: Do we need to add this to `UPGRADE.md`?

src/Sulu/Bundle/CategoryBundle/Resources/translations/admin.de.json
```diff
@@ -6,8 +6,11 @@
     "sulu_category.keyword": "Schlüsselwort",
     "sulu_category.multiple_usage": "Mehrfachverwendung",
     "sulu_category.all_categories": "Alle Kategorien",
+    "sulu_category.parent_category": "Übergeordnete Kategorie",
+    "sulu_category.restore_parent_category_info_text": "Wählen Sie eine Kategorie um die gelöschte Kategorie als Kind einer bestehenden Kategorie wiederherzustellen",
```
luca-rath:
```suggestion
    "sulu_category.restore_parent_category_info_text": "Wählen Sie eine Kategorie um die gelöschte Kategorie als Kind der ausgewählten Kategorie wiederherzustellen",
```

src/Sulu/Bundle/CategoryBundle/Resources/translations/admin.en.json
```diff
@@ -6,8 +6,11 @@
     "sulu_category.keyword": "Keyword",
     "sulu_category.multiple_usage": "Multiple Usage",
     "sulu_category.all_categories": "All categories",
+    "sulu_category.parent_category": "Parent category",
+    "sulu_category.restore_parent_category_info_text": "Select a category to restore the removed category as child of an existing category",
```
luca-rath:
```suggestion
    "sulu_category.restore_parent_category_info_text": "Select a category to restore the removed category as child of the selected category",
```

src/Sulu/Bundle/CategoryBundle/Trash/CategoryTrashItemHandler.php
```diff
@@ -0,0 +1,284 @@
...
+                    'creatorId' => $creator ? $creator->getId() : null,
+                ];
+            }
+
+            $data['translations'][] = $translationData;
+        }
+
+        $categoryTitles = [];
+        foreach ($category->getTranslations() as $translation) {
+            $categoryTitles[$translation->getLocale()] = $translation->getTranslation();
```
luca-rath: Why looping the `$category->getTranslations()` twice? This could be done easily in the above loop

src/Sulu/Bundle/CategoryBundle/Trash/CategoryTrashItemHandler.php
```diff
@@ -0,0 +1,284 @@
...
+    }
+
+    public function restore(TrashItemInterface $trashItem, array $restoreFormData): object
+    {
+        $id = (int) $trashItem->getResourceId();
+        $data = $trashItem->getRestoreData();
+        $parentId = $restoreFormData['parent'];
+
+        if ($data['key'] && null !== $this->categoryRepository->findCategoryByKey($data['key'])) {
+            throw new CategoryKeyNotUniqueException($data['key']);
```
luca-rath, line 185: Should we maybe restore the category without key in this case? Or allow the user to specify the `key` in the `restore_category` form?

src/Sulu/Bundle/TrashBundle/Application/DoctrineRestoreHelper/DoctrineRestoreHelper.php
```diff
@@ -0,0 +1,60 @@
...
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\TrashBundle\Application\DoctrineRestoreHelper;
+
+use Doctrine\ORM\EntityManagerInterface;
+use Doctrine\ORM\Id\AssignedGenerator;
+use Doctrine\ORM\Mapping\ClassMetadata;
+
+final class DoctrineRestoreHelper implements DoctrineRestoreHelperInterface
```
luca-rath, line 23: Would mark this class as `@internal` until it's a little bit battle tested, wdyt?

src/Sulu/Bundle/TrashBundle/Resources/config/doctrine/TrashItem.orm.xml
```diff
@@ -14,6 +14,7 @@
         <field name="resourceSecurityObjectType" column="resourceSecurityObjectType" type="string" nullable="true" length="191"/>
         <field name="resourceSecurityObjectId" column="resourceSecurityObjectId" type="string" nullable="true" length="191"/>
         <field name="storeTimestamp" column="storeTimestamp" type="datetime_immutable" nullable="false"/>
+        <field name="defaultLocale" column="defaultLocale" type="string" nullable="true" length="191"/>
```
luca-rath, line 17: Add this field to the `CREATE TABLE` statement in the `UPGRADE.md` or add a new `ALTER TABLE` statement

src/Sulu/Bundle/TagBundle/Tests/Functional/Trash/TagTrashItemHandlerTest.php
```diff
@@ -0,0 +1,89 @@
...
+
+    public function testStoreAndRestore(): void
+    {
+        /** @var Tag $tag1 */
+        $tag1 = $this->tagRepository->createNew();
+        $tag1->setName('First Tag');
+        $tag1->setCreated(new \DateTime('2020-10-10'));
+        $this->entityManager->persist($tag1);
+
+        $tag2 = $this->tagRepository->createNew();
```
luca-rath, line 62: This tag isn't used, right? Why did you create it?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FormOverlay/FormOverlay.js
luca-rath: I'm not sure if this has side effects somewhere

luca-rath (changes requested):

src/Sulu/Bundle/TagBundle/Tests/Functional/Trash/TagTrashItemHandlerTest.php
luca-rath: Ah to test that there are no id collisions when restoring the tag, right? ... nevermind then

luca-rath:
Would be nice if we could access the restore data in the restore form, so the old parent id could be filled in the `single_category_selection` directly

luca-rath:
Another thought, what about opening the detail view of the restored element after restoring? Should be possible using the deep link functionality, right?

alexander-schranz:
> Would be nice if we could access the restore data in the restore form, so the old parent id could be filled in the single_category_selection directly

I thought we already had some discussion about this that we just set the all restored data as form data so this kind of things are preselected if they still exist, if not the selection should set it to null and the content manager can still select one or change it where to restore.

luca-rath:
> I thought we already had some discussion about this that we just set the all restored data as form data so this kind of things are preselected if they still exist, if not the selection should set it to null and the content manager can still select one or change it where to restore.

To be honest, I can't remember 🙈 But yes, this would be a huge benefit for the content manager

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FormOverlay/FormOverlay.js
niklasnatter: the `FormOverlay` is used in the `ProfileFormOverlay`, `FormOverlayList` and for displaying the settings overlay of the `FieldBlocks`. i will set the `confirmDisabled` prop in these components to keep the previous behaviour 🙂

src/Sulu/Bundle/CategoryBundle/Entity/CategoryInterface.php
niklasnatter: will add it

src/Sulu/Bundle/CategoryBundle/Trash/CategoryTrashItemHandler.php
niklasnatter: thats right - will do it there

src/Sulu/Bundle/TrashBundle/Application/DoctrineRestoreHelper/DoctrineRestoreHelper.php
niklasnatter: 👍

src/Sulu/Bundle/TagBundle/Tests/Functional/Trash/TagTrashItemHandlerTest.php
niklasnatter: yes thats right 🙂

src/Sulu/Bundle/CategoryBundle/Trash/CategoryTrashItemHandler.php
niklasnatter: at the moment, this exceptions results in the following error in the administration interface:
Screenshot 2021-09-16 at 18 42 26

i think i would not expect that the key of the category is changed automatically. we could add the key to the restore form, but this means that the input will always be visible (even before an error occurred). 

i think i would just display the error message for now. but thats not a strong opinion and i am open to change it if you prefer another solution 🙂  @sulu/core-team

niklasnatter:
> Would be nice if we could access the restore data in the restore form, so the old parent id could be filled in the `single_category_selection` directly

Done 🙂

niklasnatter:
> Another thought, what about opening the detail view of the restored element after restoring? Should be possible using the deep link functionality, right?

unfortunately, i think we would need to add an additional configuration like we did for the search configuration and the selection field type configuration for this 😕 
https://github.com/sulu/sulu/blob/40b4785014b6936cbf74f3c88b3396c112d7b65e/src/Sulu/Bundle/ContactBundle/DependencyInjection/SuluContactExtension.php#L48-L54
https://github.com/sulu/sulu/blob/40b4785014b6936cbf74f3c88b3396c112d7b65e/src/Sulu/Bundle/ContactBundle/DependencyInjection/SuluContactExtension.php#L174-L178

i am open to do this, but we would need to decide where to add the configuration. also, i think i would prefer to implement this in a separate pr 🙂

alexander-schranz:
I would not add an additional `sulu_trash` bundle configuration. Its always strange and not a nice DX that a bundle prepend another bundles configuration.

And I think it would be better that this lives in an Interface which the CategoryTrashItemHandler can additionally implement when there is a form e.g.:

```php
public static function getRestoreFormConfiguration(): RestoreFormConfiguration
{
      return new RestoreFormConfiguration(
           'restore_category', 
           [
                'parent' => 'parent',
           ],
      );
}
```

niklasnatter:
> And I think it would be better that this lives in an Interface which the CategoryTrashItemHandler can additionally implement when there is a form

I am open to change this 🙂  
Would propose to talk about it in person on monday next week? Would that be okay for you too @luca-rath?

luca-rath:
> Would that be okay for you too @luca-rath?

Yes, sounds good to me, don't have a strong opinion about that

niklasnatter:
> > And I think it would be better that this lives in an Interface which the CategoryTrashItemHandler can additionally implement when there is a form
> 
> I am open to change this 🙂
> Would propose to talk about it in person on monday next week? Would that be okay for you too @luca-rath?

We have decided to change this in a separate pull request 🙂

luca-rath (approved):

alexander-schranz (approved):

PR #6214, opened 2021-09-09
Improve cache clearer to avoid remove directory directly for symfony cache proxy

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | none
| Related issues/PRs | none
| License | MIT
| Documentation PR | none

#### What's in this PR?

This PR contains an improvement for the cache-clearer which enables tag purge for symfony cache proxy. This avoids delete the director directly.

#### Why?

Removing the directory can fail when in the same moment a user opens the page. But that moment is not really short when the cache is big - So for a bigger page this happens often.

### DIFF
--- a/src/Sulu/Bundle/WebsiteBundle/Cache/CacheClearer.php
+++ b/src/Sulu/Bundle/WebsiteBundle/Cache/CacheClearer.php
@@ -80,24 +80,27 @@ class CacheClearer implements CacheClearerInterface
 
         $tags = \func_num_args() >= 1 ? \func_get_arg(0) : null;
 
-        if ($this->cacheManager && $this->cacheManager->supportsInvalidate()) {
-            if (null !== $tags && $this->cacheManager->supportsTags()) {
-                foreach ($tags as $tag) {
-                    $this->cacheManager->invalidateTag($tag);
-                }
-
-                $this->eventDispatcher->dispatch(new CacheClearEvent($tags), Events::CACHE_CLEAR);
-
-                return;
+        if (null !== $tags && $this->cacheManager && $this->cacheManager->supportsTags()) {
+            foreach ($tags as $tag) {
+                $this->cacheManager->invalidateTag($tag);
             }
 
-            $request = $this->requestStack->getCurrentRequest();
-            if (!$request) {
-                return;
-            }
-
-            $this->cacheManager->invalidateDomain($request->getHost());
             $this->eventDispatcher->dispatch(new CacheClearEvent($tags), Events::CACHE_CLEAR);
+
+            return;
+        }
+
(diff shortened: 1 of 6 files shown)


src/Sulu/Bundle/WebsiteBundle/Cache/CacheClearer.php
```diff
@@ -80,24 +80,27 @@ public function clear(/*?array $tags = null*/)
...
+            $this->eventDispatcher->dispatch(new CacheClearEvent($tags), Events::CACHE_CLEAR);
+
+            return;
+        }
 
+        $request = $this->requestStack->getCurrentRequest();
+        if (null !== $request && $this->cacheManager && $this->cacheManager->supportsInvalidate()) {
             $this->cacheManager->invalidateDomain($request->getHost());
-            $this->eventDispatcher->dispatch(new CacheClearEvent($tags), Events::CACHE_CLEAR);
+            $this->eventDispatcher->dispatch(new CacheClearEvent(), Events::CACHE_CLEAR);
```
niklasnatter, line 96: we read the webspace from the tags to dispatch the `CacheCleared` domain event:
https://github.com/sulu/sulu/blob/0322de13b6c2860db6f0550d702b57acc1207392/src/Sulu/Bundle/WebsiteBundle/EventSubscriber/DomainEventEventSubscriber.php#L60-L63

if we dont pass any tag here, there will be no domain event 😕

niklasnatter: maybe we should adjust the `DomainEventEventSubscriber` to create a domain event for all webspaces, if no tags are set?

src/Sulu/Bundle/WebsiteBundle/EventSubscriber/DomainEventEventSubscriber.php
```diff
@@ -53,6 +62,11 @@ public function onCacheClear(CacheClearEvent $event): void
     {
         $tags = $event->getTags();
         if (null === $tags || 0 === \count($tags)) {
+            /** @var Webspace $webspace */
+            foreach ($this->webspaceManager->getWebspaceCollection() as $webspace) {
```
niklasnatter, line 58: looks good to me, but i think the events are not dispatched because there is no flush 🙈

niklasnatter: think we could use the `DomainEventDispatcher` instead of the `DomainEventCollector` here (and also for the events below)

niklasnatter (approved):

PR #5951, opened 2021-04-22
Update postcss and required dependencies

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? | no 
| BC breaks? | yes (webpack config update needed) 
| Deprecations? | no  
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT
| Documentation PR | sulu/sulu-docs# 

#### What's in this PR?

Update postcss and required dependencies.

A updated to newest dependencies is not possible aslong as ckeditor is stuck with postcss 7: https://github.com/ckeditor/ckeditor5-dev/blob/0f984366c53a56ffdc81a44d9e2c79defb4b85ff/packages/ckeditor5-dev-utils/package.json#L17-L21 (https://github.com/ckeditor/ckeditor5-dev/pull/695)

#### Why?

New versions provide bugfixes and new features.

PR #6217, opened 2021-09-10
Downgrade style-loader to fix very long render-time of react-styleguidist

luca-rath:
…dist

| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | yes
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | #5951
| License | MIT
| Documentation PR | -

#### What's in this PR?

Downgrade style-loader to fix very long render-time of react-styleguidist

#### Why?

Since #5951 react-styleguidist takes extremely long to render

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,7 +80,7 @@
         "react-styleguidist": "^11.0.5",
         "react-test-renderer": "^17.0.0",
         "regenerator-runtime": "^0.13.3",
-        "style-loader": "^2.0.0",
+        "style-loader": "^1.0.0",
         "stylelint": "^13.2.1",
         "stylelint-config-standard": "^22.0.0",
         "url-search-params-polyfill": "^8.0.0",


package.json
```diff
@@ -80,7 +80,7 @@
         "react-styleguidist": "^11.0.5",
         "react-test-renderer": "^17.0.0",
         "regenerator-runtime": "^0.13.3",
-        "style-loader": "^2.0.0",
+        "style-loader": "^1.0.0",
```
alexander-schranz, line 83: I think we should stay on style-loader and maybe only in the jsdocs generation use then style-loader 1. Do we know what it happens is there any default settings changed from 1 to 2. As for me its strange that a webpack plugin should have any effect on a build version of a JS.

luca-rath: Unfortunately I don't know what the exact problem with style-loader@v2 is ... I tried installing style-loader@v3, but then the styleguidist build just threw an error

luca-rath: Using style-loader@v1 only for styleguidist solves the problem just partially, because then it would only work for 
[jsdocs.sulu.io](jsdocs.sulu.io) and not if a developer runs `npm run styleguide:build` locally

alexander-schranz: Would maybe know great an issue on styleguide to find out why this is happening with style loader 2 and not with style loader 1.
I think would npm run styleguide:build is something which is used that often and when then only by us. Maybe we could also move all about styleguidist out of sulu core to the jsdocs repository.

luca-rath: See https://github.com/styleguidist/react-styleguidist/issues/1883

niklasnatter: i just compared the source code of `jsdocs.sulu.io/2.2` with `jsdocs.sulu.io/2.3`. it looks like the ckeditor styling is not resolved correctly - maybe this is related to our performance issue? 🤔 
https://www.diffchecker.com/aKosEcyC

luca-rath:
Closing in favour of #6223

PR #6218, opened 2021-09-10
Integrate the TrashBundle functionality for the Media entity

### LINKED ISSUE #5512: "Soft" Delete for Media
# Problem description

Currently in Sulu, when a media file is deleted, all database entries are deleted completely. 

# Proposed solution

I think it would be a good feature if the deletion would not happen directly on the database entries, but the entries would only be flagged with a row like "delete" true false or something else.

In this way, a deleted data record can at least once be restored manually, or it can be traced by whom it was deleted

niklasnatter:
| Q | A
| --- | ---
| New feature? | yes
| Fixed tickets | fixes #5512
| Related issues/PRs | builds upon #6213
| License | MIT

#### What's in this PR?

This PR integrates the `SuluTrashBundle` with the build in media entity. With the changes in this PR, media entities will be moved into the trash when being deleted and can be restored from the trash at a later point in time.

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/Media/Manager/MediaManager.php
+++ b/src/Sulu/Bundle/MediaBundle/Media/Manager/MediaManager.php
@@ -46,6 +46,7 @@ use Sulu\Bundle\MediaBundle\Media\PropertiesProvider\VideoPropertiesProvider;
 use Sulu\Bundle\MediaBundle\Media\Storage\StorageInterface;
 use Sulu\Bundle\MediaBundle\Media\TypeManager\TypeManagerInterface;
 use Sulu\Bundle\TagBundle\Tag\TagManagerInterface;
+use Sulu\Bundle\TrashBundle\Application\TrashManager\TrashManagerInterface;
 use Sulu\Component\PHPCR\PathCleanupInterface;
 use Sulu\Component\Security\Authentication\UserInterface;
 use Sulu\Component\Security\Authentication\UserRepositoryInterface;
@@ -168,6 +169,11 @@ class MediaManager implements MediaManagerInterface
     private $mediaPropertiesProviders;
 
     /**
+     * @var TrashManagerInterface|null
+     */
+    private $trashManager;
+
+    /**
      * @param CollectionRepository $collectionRepository
      * @param null|FFprobe|MediaPropertiesProviderInterface[] $mediaPropertiesProviders
      * @param string $downloadPath
@@ -191,7 +197,8 @@ class MediaManager implements MediaManagerInterface
         $mediaPropertiesProviders,
         $downloadPath,
         ?TargetGroupRepositoryInterface $targetGroupRepository,
-        $adminDownloadPath = null
+        $adminDownloadPath = null,
+        ?TrashManagerInterface $trashManager = null
     ) {
         $this->mediaRepository = $mediaRepository;
         $this->collectionRepository = $collectionRepository;
@@ -209,6 +216,7 @@ class MediaManager implements MediaManagerInterface
         $this->tokenStorage = $tokenStorage;
         $this->securityChecker = $securityChecker;
         $this->downloadPath = $downloadPath;
+        $this->trashManager = $trashManager;
 
         if (!$adminDownloadPath) {
             @\trigger_error(
@@ -742,6 +750,10 @@ class MediaManager implements MediaManagerInterface
                 ),
                 PermissionTypes::DELETE
             );
+        }
+
+        if (null !== $this->trashManager) {
+            $this->trashManager->store(MediaInterface::RESOURCE_KEY, $mediaEntity);
         }
 
         /** @var File $file */
--- a/src/Sulu/Bundle/MediaBundle/Media/Storage/FlysystemStorage.php
+++ b/src/Sulu/Bundle/MediaBundle/Media/Storage/FlysystemStorage.php
@@ -48,7 +48,7 @@ abstract class FlysystemStorage implements StorageInterface
 
         $this->createDirectories($storageOptions);
 
-        $parentPath = \implode('/', \array_filter([$storageOptions['directory'], $storageOptions['segment']]));
+        $parentPath = $this->getFilePath(\array_merge($storageOptions, ['fileName' => null]));
         $storageOptions['fileName'] = $this->getUniqueFileName($parentPath, $fileName);
 
         $filePath = $this->getFilePath($storageOptions);
@@ -91,9 +91,12 @@ abstract class FlysystemStorage implements StorageInterface
         }
     }
 
-    public function move(array $sourceStorageOptions, array $targetStorageOptions): void
+    public function move(array $sourceStorageOptions, array $targetStorageOptions): array
     {
         $this->createDirectories($targetStorageOptions);
+
+        $targetParentPath = $this->getFilePath(\array_merge($targetStorageOptions, ['fileName' => null]));
+        $targetStorageOptions['fileName'] = $this->getUniqueFileName($targetParentPath, $targetStorageOptions['fileName']);
 
         $targetFilePath = $this->getFilePath($targetStorageOptions);
         if ($this->filesystem->has($targetFilePath)) {
@@ -101,6 +104,8 @@ abstract class FlysystemStorage implements StorageInterface
         }
 
         $this->filesystem->rename($this->getFilePath($sourceStorageOptions), $targetFilePath);
+
+        return $targetStorageOptions;
     }
 
     protected function getUniqueFileName(string $parentPath, string $fileName, int $counter = 0): string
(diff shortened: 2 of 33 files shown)


src/Sulu/Bundle/MediaBundle/Trash/MediaTrashItemHandler.php
```diff
@@ -0,0 +1,344 @@
...
+
+            $fileData = [
+                'version' => $file->getVersion(),
+                'created' => $file->getCreated()->format('c'),
+                'creatorId' => $creator ? $creator->getId() : null,
+                'fileVersions' => [],
+            ];
+
+            /** @var FileVersion $fileVersion */
+            foreach ($file->getFileVersions() as $fileVersion) {
```
alexander-schranz, line 124: /cc @chirimoya should we move all versions into the trash?

src/Sulu/Bundle/MediaBundle/Media/Manager/MediaManager.php
```diff
@@ -745,6 +753,10 @@ public function delete($id, $checkSecurity = false)
             );
         }
 
+        if (null !== $this->trashManager) {
+            $this->trashManager->store(MediaInterface::RESOURCE_KEY, $mediaEntity);
+        }
```
alexander-schranz, line 757: The original file needs to be backup or moved into private directory / storage.

```php
        $removeOriginalFile = true;
        if (null !== $this->trashManager) {
             $removeOriginalFile = true;
             $this->trashManager->store(MediaInterface::RESOURCE_KEY, $mediaEntity);
        }
```

src/Sulu/Bundle/MediaBundle/Trash/MediaTrashItemHandler.php
niklasnatter: we decided to store/move all versions for now 🙂

src/Sulu/Bundle/MediaBundle/Media/Manager/MediaManager.php
niklasnatter: done in https://github.com/sulu/sulu/pull/6218/files#diff-2b8ee69c9133c03ab57c7f6a946d7a4adfcfa726793aac1aa1191079a5b36b78R126 🙂

src/Sulu/Bundle/MediaBundle/Trash/MediaTrashItemHandler.php
```diff
@@ -0,0 +1,390 @@
...
+    }
+
+    /**
+     * @param array<string, string|null> $originalStorageOptions
+     *
+     * @return array<string, string|null>
+     */
+    private function getTrashStorageOptions(array $originalStorageOptions): array
+    {
+        return \array_merge($originalStorageOptions, ['directory' => 'trash']);
```
alexander-schranz: we would need to prefix here the `fileName` with a uuid or the fileVersion id to avoid conflicts.

src/Sulu/Bundle/MediaBundle/Media/Storage/FlysystemStorage.php
```diff
@@ -60,41 +63,47 @@ public function save(string $tempPath, string $fileName, array $storageOptions =
...
         }
     }
 
-    protected function getUniqueFileName(string $folder, string $fileName, int $counter = 0): string
+    public function move(array $sourceStorageOptions, array $targetStorageOptions): void
+    {
+        $this->createDirectories($targetStorageOptions);
+
+        $targetFilePath = $this->getFilePath($targetStorageOptions);
+        if ($this->filesystem->has($targetFilePath)) {
```
alexander-schranz, line 98: as discussed we should resolve the conflict and return here the new storageOptions

niklasnatter: done 🙂

src/Sulu/Bundle/MediaBundle/Trash/MediaTrashItemHandler.php
niklasnatter: we decided to resolve conflicts in the `StorageInterface::move` method instead 🙂

src/Sulu/Bundle/MediaBundle/Media/Storage/LocalStorage.php
```diff
@@ -53,32 +53,28 @@ public function __construct(
...
         if ($this->filesystem->exists($filePath)) {
             throw new FilenameAlreadyExistsException($filePath);
         }
         $this->filesystem->copy($tempPath, $filePath);
 
-        return [
-            'segment' => $segment,
-            'fileName' => $fileName,
-        ];
+        return $storageOptions;
```
alexander-schranz, line 75: Return should stay the same for the save method and not contain an empty `directory`.

src/Sulu/Bundle/MediaBundle/Media/Storage/FlysystemStorage.php
```diff
@@ -60,41 +63,52 @@ public function save(string $tempPath, string $fileName, array $storageOptions =
             throw new FilenameAlreadyExistsException($filePath);
         }
 
-        return [
-            'segment' => $segment,
-            'fileName' => $fileName,
-        ];
+        return $storageOptions;
```
alexander-schranz, line 62: same as for LocalStorage. Saveshould not contain an empty `directory` key.

alexander-schranz (changes requested):

src/Sulu/Bundle/MediaBundle/Media/Storage/FlysystemStorage.php
```diff
@@ -105,15 +119,52 @@ protected function getUniqueFileName(string $folder, string $fileName, int $coun
...
+     */
     protected function getStorageOption(array $storageOptions, string $key): ?string
     {
         return $storageOptions[$key] ?? null;
     }
+
+    /**
+     * @param array<string, string|null> $storageOptions
+     */
+    protected function createDirectories(array $storageOptions): void
```
alexander-schranz: Would make this private

src/Sulu/Bundle/MediaBundle/Media/Storage/FlysystemStorage.php
```diff
@@ -105,15 +119,52 @@ protected function getUniqueFileName(string $folder, string $fileName, int $coun
...
+
+        if ($segmentPath && !$this->filesystem->has($segmentPath)) {
+            $this->filesystem->createDir($segmentPath);
+        }
+    }
+
+    /**
+     * @param array<string, string|null> $storageOptions
+     */
+    protected function getFilePath(array $storageOptions): string
```
alexander-schranz: Would make this private

src/Sulu/Bundle/MediaBundle/Media/Storage/LocalStorage.php
```diff
@@ -144,21 +155,41 @@ private function getUniqueFileName(string $folder, string $fileName, int $counte
...
     {
-        return \implode('/', \array_filter([$this->uploadPath, $segment, $fileName]));
+        return \array_key_exists($key, $storageOptions) ? $storageOptions[$key] : null;
     }
 
-    private function getStorageOption(array $storageOption, string $key): ?string
+    /**
+     * @param array<string, string|null> $storageOptions
+     */
+    private function createDirectories(array $storageOptions): void
```
alexander-schranz, line 175: This method is unnecessary the previous  `$segmentPath` would just need to keep the `directory` parameter in count. 

https://github.com/sulu/sulu/blob/02161cb45e0c7175c267ceaea3ed310a32c996fe/src/Sulu/Bundle/MediaBundle/Media/Storage/LocalStorage.php#L61

Then already at:

https://github.com/sulu/sulu/blob/02161cb45e0c7175c267ceaea3ed310a32c996fe/src/Sulu/Bundle/MediaBundle/Media/Storage/LocalStorage.php#L69

The correct directory would be created recursively by the symfony filesystem component: https://symfony.com/doc/current/components/filesystem.html#mkdir

alexander-schranz (changes requested):

src/Sulu/Bundle/MediaBundle/Media/Storage/FlysystemStorage.php
niklasnatter: i need to access this in the `GoogleCloudStorage` and `S3Storage` at the moment. think i would prefer to keep it protected and not duplicate the logic ?

src/Sulu/Bundle/MediaBundle/Media/Storage/LocalStorage.php
niklasnatter: okay i have removed the check for the `directoryPath`. think i would keep the private method because we need to do this in the `save` method and the `move` method

alexander-schranz (approved):

PR #6219, opened 2021-09-10
Add single_collection_selection and collection_selection content type

### LINKED ISSUE #5694: Add media `single_collection_selection` field type
| Q | A
| --- | ---
| Sulu Version | for `2.x`
| Browser Version | -

# Actual Behavior

Currently there is no field type or content types for media collections.

# Expected Behavior

In some cases it would be great to have a `single_collection_selection` field type or also a `collection_selection`. Example the sulu form bundle could use it to make it possible to set the target collection for uploaded media (could be restricted collection).

Other use cases should be a media overview on the website with very special filters for tags or categories or custom fields, where the developer want to make the query itself and can not use the smart content feature for it.

# Possible Solutions

Register selection types for the collection resource. And maybe also provide a content type for it which returns the api/collection using the collection manager service.

niklasnatter:
| Q | A
| --- | ---
| Fixed tickets | fixes #5694
| License | MIT
| Documentation PR | sulu/sulu-docs#673

#### What's in this PR?

This PR adds a `single_collection_selection` content type and a `collection_selection`. The content types allow to select collections that are managed in the media section of the administration interface.

#### Why?

We need a `single_collection_selection` to select the target collection when restoring a media entity in #6218. Also, these field type might be useful for other usecases (see #5694). 

#### Example Usage

```xml
<property name="collection_selection" type="collection_selection">
    <meta>
        <title lang="en">collection_selection</title>
    </meta>
</property>

<property name="single_collection_selection" type="single_collection_selection">
    <meta>
        <title lang="en">single_collection_selection</title>
    </meta>
</property>
```


src/Sulu/Bundle/MediaBundle/Content/Types/CollectionSelection.php
```diff
@@ -0,0 +1,79 @@
...
+        $locale = $property->getStructure()->getLanguageCode();
+
+        if (!\is_array($ids) || empty($ids)) {
+            return [];
+        }
+
+        $collections = [];
+        foreach ($ids as $id) {
+            try {
+                $collections[] = $this->collectionManager->getById($id, $locale);
```
alexander-schranz, line 59: Not sure about this but we currently seems not to have another method to load the collections.

alexander-schranz (approved):

src/Sulu/Bundle/MediaBundle/Content/Types/CollectionSelection.php
niklasnatter: unfortunately thats right 🙈

PR #6223, opened 2021-09-13
Fix startup performance and ckeditor styling of react-styleguidist integration

niklasnatter:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Related issues/PRs | #6217
| License | MIT

#### What's in this PR?

This PR adds the `css-loader` before the `style-loader` for processing ckeditor css files in the webspace configuration of the `react-styleguidist` package.

I looks the processing of the ckeditor css files worked without the `css-loader` while we were using `style-loader@v1`, but fails and leads to a big performance penalty when using `style-loader@v2` (see #6217).

### DIFF
--- a/styleguide.config.js
+++ b/styleguide.config.js
@@ -178,6 +178,7 @@ module.exports = { // eslint-disable-line
                                 injectType: 'singletonStyleTag',
                             },
                         },
+                        'css-loader',
                         {
                             loader: 'postcss-loader',
                             options: styles.getPostCssConfig({


styleguide.config.js
```diff
@@ -178,6 +178,7 @@ module.exports = { // eslint-disable-line
                                 injectType: 'singletonStyleTag',
                             },
                         },
+                        'css-loader',
```
niklasnatter, line 181: i figured that his was missing by comparing the configuration inside of the `webpack.config.js` with the configuration used here. we use the 'css-loader' before the `MiniCssExtractPlugin` in the `webpack.config.js` here:
https://github.com/sulu/sulu/blob/a8b771793decf61248288dead37407f0950f0690/webpack.config.js#L121-L139

PR #6225, opened 2021-09-14
Remove separate webpack configuration in react-styleguidist configuration

niklasnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Related issues/PRs | #6223
| License | MIT

#### What's in this PR?

This PR removes the separate webpack configuration in the `styleguide.config.js` and uses the default webpack configuration instead. This allows us to maintain a single webpack configuration and therefore prevents issues like #6223.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -8,7 +8,7 @@ module.exports = (env, argv) => { // eslint-disable-line no-undef
     env = env ? env : {};
     argv = argv ? argv : {};
 
-    let publicDir = 'public';
+    const publicPath = env && env.public_path ? env.public_path : '/';
     const outputPath = env && env.output_path ? env.output_path : path.join('build', 'admin');
     // eslint-disable-next-line no-undef
     const projectRootPath = env && env.project_root_path ? env.project_root_path : __dirname;
@@ -16,6 +16,7 @@ module.exports = (env, argv) => { // eslint-disable-line no-undef
         ? env.node_modules_path
         : path.resolve(projectRootPath, 'node_modules');
 
+    let publicDir = 'public';
     const composerConfig = require(path.resolve(projectRootPath, 'composer.json'));
     if (composerConfig.extra && composerConfig.extra['public-dir']) {
         publicDir = composerConfig.extra['public-dir'];
@@ -42,6 +43,7 @@ module.exports = (env, argv) => { // eslint-disable-line no-undef
         output: {
             path: path.resolve(projectRootPath, publicDir),
             filename: outputPath + '/[name].[chunkhash].js',
+            publicPath,
         },
         devtool: argv.mode === 'development' ? 'eval-source-map' : 'source-map',
         plugins: [
@@ -94,7 +96,8 @@ module.exports = (env, argv) => { // eslint-disable-line no-undef
                         {
                             loader: MiniCssExtractPlugin.loader,
                             options: {
-                                publicPath: '../../', // @see https://github.com/sulu/sulu/pull/6225
+                                // set path to public root from bundled css: https://github.com/sulu/sulu/pull/6225
+                                publicPath: path.relative(outputPath, '.'),
                             },
                         },
                         'css-loader',
@@ -106,7 +109,8 @@ module.exports = (env, argv) => { // eslint-disable-line no-undef
                         {
                             loader: MiniCssExtractPlugin.loader,
                             options: {
-                                publicPath: '../../', // @see https://github.com/sulu/sulu/pull/6225
+                                // set path to public root from bundled css: https://github.com/sulu/sulu/pull/6225
+                                publicPath: path.relative(outputPath, '.'),
                             },
                         },
                         {
@@ -132,7 +136,8 @@ module.exports = (env, argv) => { // eslint-disable-line no-undef
                         {
                             loader: MiniCssExtractPlugin.loader,
                             options: {
-                                publicPath: '../../', // @see https://github.com/sulu/sulu/pull/6225
+                                // set path to public root from bundled css: https://github.com/sulu/sulu/pull/6225
+                                publicPath: path.relative(outputPath, '.'),
                             },
                         },
                         {
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -5,6 +5,9 @@ const fs = require('fs');
 const path = require('path');
 
 module.exports = (env, argv) => { // eslint-disable-line no-undef
+    env = env ? env : {};
+    argv = argv ? argv : {};
+
     let publicDir = 'public';
     const outputPath = env && env.output_path ? env.output_path : path.join('build', 'admin');
     // eslint-disable-next-line no-undef
@@ -88,14 +91,26 @@ module.exports = (env, argv) => { // eslint-disable-line no-undef
                     test: /\.css/,
                     exclude: /ckeditor5-[^/\\]+[/\\]theme[/\\].+\.css$/,
                     use: [
-                        MiniCssExtractPlugin.loader,
+                        {
+                            loader: MiniCssExtractPlugin.loader,
+                            options: {
+                                // https://github.com/webpack-contrib/mini-css-extract-plugin/issues/44#issuecomment-391009221
+                                publicPath: '../../',
+                            },
+                        },
                         'css-loader',
                     ],
                 },
                 {
                     test: /\.(scss)$/,
                     use: [
-                        MiniCssExtractPlugin.loader,
+                        {
+                            loader: MiniCssExtractPlugin.loader,
+                            options: {
+                                // https://github.com/webpack-contrib/mini-css-extract-plugin/issues/44#issuecomment-391009221
+                                publicPath: '../../',
+                            },
+                        },
                         {
                             loader: 'css-loader',
                             options: {
@@ -116,7 +131,13 @@ module.exports = (env, argv) => { // eslint-disable-line no-undef
                 {
                     test: /ckeditor5-[^/\\]+[/\\]theme[/\\].+\.css$/,
                     use: [
-                        MiniCssExtractPlugin.loader,
+                        {
+                            loader: MiniCssExtractPlugin.loader,
+                            options: {
+                                // https://github.com/webpack-contrib/mini-css-extract-plugin/issues/44#issuecomment-391009221
+                                publicPath: '../../',
+                            },
+                        },
                         {
                             loader: 'css-loader',
                         },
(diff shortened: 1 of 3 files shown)


webpack.config.js
```diff
@@ -5,6 +5,9 @@ const fs = require('fs');
 const path = require('path');
 
 module.exports = (env, argv) => { // eslint-disable-line no-undef
+    env = env ? env : {};
+    argv = argv ? argv : {};
```
niklasnatter, line 9: these fallbacks are already present on the `2.3` branch:
https://github.com/sulu/sulu/blob/100091add44ce79fa37454abf1db58ee955537cf/webpack.config.js#L7-L9

webpack.config.js
```diff
@@ -88,14 +91,26 @@ module.exports = (env, argv) => { // eslint-disable-line no-undef
                     test: /\.css/,
                     exclude: /ckeditor5-[^/\\]+[/\\]theme[/\\].+\.css$/,
                     use: [
-                        MiniCssExtractPlugin.loader,
+                        {
+                            loader: MiniCssExtractPlugin.loader,
+                            options: {
+                                // https://github.com/webpack-contrib/mini-css-extract-plugin/issues/44#issuecomment-391009221
+                                publicPath: '../../',
```
alexander-schranz: this looks false in this webpack config file

nnatter, line 98: we configure the `MiniCssExtractPlugin` to output the bundled css into `build/admin/main.HASH.css`:
https://github.com/sulu/sulu/blob/1b6837b2ca7f64f8baf833508f1a7bb0c7ff0630/webpack.config.js#L50-L52

when we reference another resource inside of a `.scss` file, the path is replaced with the public path returned by the respective loader. for example, the `./tablet.png` is loaded by the `file-loader` and replaced with `build/admin/images/tablet.HASH.png`:
https://github.com/sulu/sulu/blob/602d71aa0c23b864b9d0de06be78d26f5e9a483d/src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/preview.scss#L30

this will look something like this i the bundled `build/admin/main.HASH.css`:
```
background-image: url(build/admin/images/tablet.HASH.png);
```

as the browser resolves `url()` inside of a stylesheet relative to the location of the stylesheet, the browser tries to resolve the url relative to the `build/admin/main.HASH.css` instead of the correct url:
```
WRONG URL:      build/admin/build/admin/images/tablet.HASH.png
CORRECT URL:    build/admin/images/tablet.HASH.png
```

in short, the problem is that the `file-loader` returns the correct public url for the `tablet.png`, but the `MiniCssExtractPlugin.loader` does not adjust this url to be relative to the location of the bundled css.

to fix this, we need to tell the `MiniCssExtractPlugin.loader` where to find the public path of the project relative to the bundled css by setting the `publicPath` option. after doing this, the `MiniCssExtractPlugin.loader` will adjust the url returned by the `file-loader` and the bundled css will look something like this:
```
background-image: url(../../build/admin/images/tablet.HASH.png);
```

the following issue might be helpful for understanding the solution too: https://github.com/webpack-contrib/mini-css-extract-plugin/issues/44#issuecomment-391009221

niklasnatter: why do you think its wrong? this defines the path to the public directory relative to the bundled css file. i tried to explain whats happening here: https://github.com/sulu/sulu/pull/6225/files#r708393479

webpack.config.js
```diff
@@ -140,7 +161,7 @@ module.exports = (env, argv) => { // eslint-disable-line no-undef
                         {
                             loader: 'file-loader',
                             options: {
-                                name: '/' + outputPath + '/fonts/[name].[hash].[ext]',
+                                name: outputPath + '/fonts/[name].[hash].[ext]',
```
niklasnatter, line 166: using absolute paths here was a workaround for the issue described in https://github.com/sulu/sulu/pull/6225/files#r708209437

when using an absolute path, the `file-loader` will return an absolute path and the `url()` inside of the `build/admin/main.HASH.css` **will not be resolved relative to the file**. the problem with this workaround is that it only works if the fonts are accessible via `domain.com/fonts/name.HASH.EXT`. 

for example, the workaround does not work when accessing a styleguide build inside of a subfolder like `jsdocs.sulu.io/2.3/`. in this case, the browser would try to load the font via `jsdocs.sulu.io/fonts/name.HASH.EXT` instead of `jsdocs.sulu.io/2.3/fonts/name.HASH.EXT`

webpack.config.js
```diff
@@ -88,14 +91,24 @@ module.exports = (env, argv) => { // eslint-disable-line no-undef
                     test: /\.css/,
                     exclude: /ckeditor5-[^/\\]+[/\\]theme[/\\].+\.css$/,
                     use: [
-                        MiniCssExtractPlugin.loader,
+                        {
+                            loader: MiniCssExtractPlugin.loader,
+                            options: {
+                                publicPath: '../../', // @see https://github.com/sulu/sulu/pull/6225
```
nnatter, line 97: we configure the `MiniCssExtractPlugin` to output the bundled css into `build/admin/main.HASH.css`:
https://github.com/sulu/sulu/blob/1b6837b2ca7f64f8baf833508f1a7bb0c7ff0630/webpack.config.js#L50-L52

when we reference another resource inside of a `.scss` file, the path is replaced with the public path returned by the respective loader. for example, the `./tablet.png` is loaded by the `file-loader` and replaced with `build/admin/images/tablet.HASH.png`:
https://github.com/sulu/sulu/blob/602d71aa0c23b864b9d0de06be78d26f5e9a483d/src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/preview.scss#L30

this will look something like this i the bundled `build/admin/main.HASH.css`:
```
background-image: url(build/admin/images/tablet.HASH.png);
```

as the browser resolves `url()` inside of a stylesheet relative to the location of the stylesheet, the browser tries to resolve the url relative to the `build/admin/main.HASH.css` instead of the correct url:
```
WRONG URL:      build/admin/build/admin/images/tablet.HASH.png
CORRECT URL:    build/admin/images/tablet.HASH.png
```

in short, the problem is that the `file-loader` returns the correct public url for the `tablet.png`, but the `MiniCssExtractPlugin.loader` does not adjust this url to be relative to the location of the bundled css.

to fix this, we need to tell the `MiniCssExtractPlugin.loader` where to find the public path of the project relative to the bundled css by setting the `publicPath` option. after doing this, the `MiniCssExtractPlugin.loader` will adjust the url returned by the `file-loader` and the bundled css will look something like this:
```
background-image: url(../../build/admin/images/tablet.HASH.png);
```

the following issue might be helpful for understanding whats happening too: https://github.com/webpack-contrib/mini-css-extract-plugin/issues/44#issuecomment-391009221

webpack.config.js
niklasnatter: now that i think about it, the previous workaround probably causes issues when running sulu in a subfolder (https://github.com/sulu/sulu-docs/issues/495) too

webpack.config.js
alexander-schranz: I think this will not longer make it possible to overwrite the public directory. Not tested it but I think using  `path.resolve(projectRootPath, publicDir)` the configured directory like above should also end in the correct result?

niklasnatter: i dont think this is related to the public directory 🙂 
this config only affects the path in the bundled css for resources that are referenced inside of `.scss` files. i am not sure how to explain it better than in https://github.com/sulu/sulu/pull/6225/files#r708393479 - maybe we should talk about it in person?

webpack.config.js
```diff
@@ -88,14 +93,26 @@ module.exports = (env, argv) => { // eslint-disable-line no-undef
                     test: /\.css/,
                     exclude: /ckeditor5-[^/\\]+[/\\]theme[/\\].+\.css$/,
                     use: [
-                        MiniCssExtractPlugin.loader,
+                        {
+                            loader: MiniCssExtractPlugin.loader,
+                            options: {
+                                // set path to public root from bundled css: https://github.com/sulu/sulu/pull/6225
+                                publicPath: path.relative(outputPath, '.'),
```
niklasnatter, line 100: we configure the `MiniCssExtractPlugin` to output the bundled css into `build/admin/main.HASH.css`:
https://github.com/sulu/sulu/blob/1b6837b2ca7f64f8baf833508f1a7bb0c7ff0630/webpack.config.js#L50-L52

when we reference another resource inside of a `.scss` file, the path is replaced with the public path returned by the respective loader. for example, the `./tablet.png` is loaded by the `file-loader` and replaced with `build/admin/images/tablet.HASH.png`:
https://github.com/sulu/sulu/blob/602d71aa0c23b864b9d0de06be78d26f5e9a483d/src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/preview.scss#L30

this will look something like this i the bundled `build/admin/main.HASH.css`:
```
background-image: url(build/admin/images/tablet.HASH.png);
```

as the browser resolves `url()` inside of a stylesheet relative to the location of the stylesheet, the browser tries to resolve the url relative to the `build/admin/main.HASH.css` instead of the correct url:
```
WRONG URL:      build/admin/build/admin/images/tablet.HASH.png
CORRECT URL:    build/admin/images/tablet.HASH.png
```

in short, the problem is that the `file-loader` returns the correct public url for the `tablet.png`, but the `MiniCssExtractPlugin.loader` does not adjust this url to be relative to the location of the bundled css.

to fix this, we need to tell the `MiniCssExtractPlugin.loader` how to find the public directory relative to the bundled css by setting the `publicPath` option. after doing this, the `MiniCssExtractPlugin.loader` will adjust the url returned by the `file-loader` and the bundled css will look something like this:
```
background-image: url(../../build/admin/images/tablet.HASH.png);
```

the following issue might be helpful for understanding whats happening too: https://github.com/webpack-contrib/mini-css-extract-plugin/issues/44#issuecomment-391009221

webpack.config.js
alexander-schranz: Where does then the `../../` come from when I configured sulu into another directory and so have example has 3 directories to the publicDir: `build/another-directory/admin` then I think the `../../../` should be here? Do we need this in the general webpack config can we introduce instead a parameter for styleguidist specific things or overwrite it in there configuration.

niklasnatter: yes, if you set `env.output_path` to `build/another-directory/admin`, we need to use `../../../`. i changed the code to `path.relative(outputPath, '.')` to do this automatically 🙂

niklasnatter: > Do we need this in the general webpack config can we introduce instead a parameter for styleguidist specific things or overwrite it in there configuration.

as long as you are not using sulu inside of a subfolder, hardcoding `/` here (like it was done before: https://github.com/sulu/sulu/pull/6225/files#r708207169) works. if we want to allow running sulu in a subfolder, we should define it like this

alexander-schranz: Looks then good for me.

alexander-schranz (approved):

PR #6231, opened 2021-09-16
Fix application crash if autocomplete query is not a valid regular expression

### LINKED ISSUE #5441: Supplying * asterix into tag selection UI fails with blank screen

niklasnatter:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Fixed tickets | fixes #5441
| License | MIT

#### What's in this PR?

This pull request adjusts the `Suggestion` component that is used by the `AutoComplete` components to be able to handle a query that is not a valid regular expression.

#### Why?

At the moment, the `Suggestion` component throws an error that crashes the administration interface, if the user enters a query that is not a valid regular expression like `*`. See #5441

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,7 +96,8 @@
         },
         "setupFiles": [
             "jest-canvas-mock",
-            "regenerator-runtime/runtime"
+            "regenerator-runtime/runtime",
+            "core-js/features/string/replace-all"
         ],
         "setupFilesAfterEnv": [
             "./tests/js/testSetup.config.js"
--- a/package.json
+++ b/package.json
@@ -105,20 +105,6 @@
             "enzyme-to-json/serializer"
         ],
         "clearMocks": true,
-        "transform": {
-            "\\.js$": [
-                "babel-jest",
-                {
-                    "babelrcRoots": [
-                        ".",
-                        "./src/Sulu/Bundle/*/Resource/js"
-                    ],
-                    "presets": [
-                        "@babel/preset-env"
-                    ]
-                }
-            ]
-        },
         "transformIgnorePatterns": [
             "node_modules/(?!(@ckeditor|lodash-es)/)"
         ],
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/AutoCompletePopover/Suggestion.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/AutoCompletePopover/Suggestion.js
@@ -37,6 +37,7 @@ export default class Suggestion extends React.PureComponent<Props> {
             matcher = this.props.query;
         }
 
+        // $FlowFixMe: flow does not recognize the replaceAll method: https://github.com/facebook/flow/issues/560
         const highlightedText = text.replaceAll(matcher, '<strong>$&</strong>');
 
         return (
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/AutoCompletePopover/Suggestion.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/AutoCompletePopover/Suggestion.js
@@ -37,7 +37,7 @@ export default class Suggestion extends React.PureComponent<Props> {
             matcher = this.props.query;
         }
 
-        const highlightedText = text.replace(matcher, '<strong>$&</strong>');
+        const highlightedText = text.replaceAll(matcher, '<strong>$&</strong>');
 
         return (
             <span dangerouslySetInnerHTML={{__html: highlightedText}} />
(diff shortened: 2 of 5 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/components/AutoCompletePopover/Suggestion.js
```diff
@@ -24,21 +24,23 @@ export default class Suggestion extends React.PureComponent<Props> {
...
+        let matcher;
+        try {
+            // try to match all highlighted parts using case insensitive regular expression
+            matcher = new RegExp(this.props.query, 'gi');
+        } catch (e) {
+            // fallback to highlight first exact match if given query is an invalid regular expression like "*"
+            matcher = this.props.query;
+        }
+
+        const highlightedText = text.replace(matcher, '<strong>$&</strong>');
```
niklasnatter: unfortunately, we cannot easily use `replaceAll` here because it is not supported before `Node 15`:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll

niklasnatter: `replace` will highlight all matches if the query is a valid regular expression. if it is not a valid regular expression, only the first exact match will be highlighted. 

i think this should be okay, because the second case happens very rarely (it is the same case that crashed the application until now).

alexander-schranz: can we have different babel settings for tests? or import the core-js polyfill only in the tests? https://github.com/zloirock/core-js ?

niklasnatter: not sure about babel, but i think we could import the polyfill inside of the `testSetup.config.js` 🙂 
but as far as i see, `core-js` is not declared anywhere in our dependencies. i think i would need to add it?

alexander-schranz: We could add it as test dependency if we did not yet need it anywhere. But I'm also then fine to keep it with this solution if all works and testa are green.

niklasnatter: maybe we could set another babel configuration using the jest `transform` option:
https://jestjs.io/docs/configuration#transform-objectstring-pathtotransformer--pathtotransformer-object

will have another looks at this when i find the time

niklasnatter: it looks like there is a bug in babel and babel does not transpile `String.replaceAll()` at all at the moment:
https://github.com/babel/babel/issues/13701

because of this, we cannot solve this by using a different babel config for tests 😕

package.json
```diff
@@ -106,20 +106,6 @@
             "enzyme-to-json/serializer"
         ],
         "clearMocks": true,
-        "transform": {
```
niklasnatter, line 108: i figured that this is not needed anymore while testing https://github.com/sulu/sulu/pull/6231#discussion_r710073550 🙂

src/Sulu/Bundle/AdminBundle/Resources/js/components/AutoCompletePopover/Suggestion.js
alexander-schranz: I think currently babel is not correctly configured as I think we should use `usage` in the usebuiltins babel configuration.
https://babeljs.io/docs/en/babel-preset-env#usebuiltins-usage

niklasnatter: thats right! i created a separate pull request for this: https://github.com/sulu/sulu/pull/6241

nnatter: thats right! i created a separate pull request for this: https://github.com/sulu/sulu/pull/6241

niklasnatter: i finally managed to change this to `replaceAll` 🥳

alexander-schranz (approved):

PR #6234, opened 2021-09-20
Fix InfiniteLoadingStrategy to prevent missing and duplicated list items

niklasnatter:
| Q | A
| --- | ---
| Bug fix? | yes
| License | MIT

#### What's in this PR?

This PR improves the page handling of the `InfiniteLoadingStrategy` that is used in the media section of the administration interface. The `InfiniteLoadingStrategy` now remembers the last page that was loaded. If the strategy is called to load a page that is not the expected next page, it will reload the items of all previous pages to make sure that there are no missing or duplicated list items.

This change solves the following problems:

- If the administration interface is refreshed with a page query parameter like `?mediaPage=4`, the list will contain only items of the 4th page. Items from page 1 to 3 will be missing and will not be loaded while scrolling.
- If a list item is deleted, the current page is reloaded. Inside of the media section, this adds remaining items of the current page a second time and therefore leads to duplicate list items

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/Resources/js/containers/MediaCollection/CollectionSection.js
+++ b/src/Sulu/Bundle/MediaBundle/Resources/js/containers/MediaCollection/CollectionSection.js
@@ -113,6 +113,8 @@ class CollectionSection extends React.Component<Props> {
     };
 
     handleSaveResponse = (resourceStore: ResourceStore) => {
+        this.closeCollectionOperationOverlay();
+
         if (this.openedCollectionOperationOverlayType === 'update') {
             this.props.resourceStore.setMultiple(resourceStore.data);
         } else {
@@ -120,7 +122,6 @@ class CollectionSection extends React.Component<Props> {
         }
 
         resourceStore.destroy();
-        this.closeCollectionOperationOverlay();
     };
 
     handleCollectionOverlayClose = () => {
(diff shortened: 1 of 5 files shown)


src/Sulu/Bundle/MediaBundle/Resources/js/containers/MediaCollection/CollectionSection.js
```diff
@@ -113,14 +113,15 @@ class CollectionSection extends React.Component<Props> {
     };
 
     handleSaveResponse = (resourceStore: ResourceStore) => {
+        this.closeCollectionOperationOverlay();
```
niklasnatter, line 116: the `this.props.onCollectionNavigate(resourceStore.id)` call below will clear the data of the `CollectionFormOverlay`. we should close the overlay before the data is cleared, otherwise the user sees that the input fields are cleared before the overlay is fully closed 🙂

src/Sulu/Bundle/MediaBundle/Resources/js/containers/MediaCollection/MediaCollection.js
```diff
@@ -47,7 +47,6 @@ class MediaCollection extends React.Component<Props> {
     @action handleUpload = (media: Array<Object>) => {
         const {mediaListStore} = this.props;
 
-        mediaListStore.reset();
```
niklasnatter, line 50: calling `reset` here was necessary to prevent the media list from displaying the last loaded page only. with the changes in this pr, this is not necessary anymore

alexander-schranz (approved):

alexander-schranz:
This seems also fix that if you are on infinite on the last page. Open a media, click back button only items of the last page are loaded again and not all data.

PR #6241, opened 2021-09-22
Configure babel to automatically add polyfills based on usage in code and targets in .browserslistrc

niklasnatter:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| License | MIT

#### What's in this PR?

This pull request adjusts the `babel.config.json` to automatically add corejs polyfills for features used in the source code based on the targets in our `.browserslistrc` file.

### DIFF
--- a/babel.config.json
+++ b/babel.config.json
@@ -2,8 +2,9 @@
     "presets": ["@babel/preset-env", "@babel/preset-react"],
     "plugins": [
         ["@babel/plugin-proposal-decorators", {"legacy": true}],
-        "@babel/plugin-proposal-object-rest-spread",
-        "@babel/plugin-transform-flow-strip-types",
-        ["@babel/plugin-proposal-class-properties", {"loose": true}]
-    ]
+        "@babel/plugin-transform-flow-strip-types"
+    ],
+    "assumptions": {
+        "setPublicClassFields": true
+    }
 }
--- a/babel.config.json
+++ b/babel.config.json
@@ -1,5 +1,8 @@
 {
-    "presets": ["@babel/preset-env", "@babel/preset-react"],
+    "presets": [
+        ["@babel/preset-env", {"useBuiltIns": "usage", "corejs": "3.18"}],
+        "@babel/preset-react"
+    ],
     "plugins": [
         ["@babel/plugin-proposal-decorators", {"legacy": true}],
         "@babel/plugin-transform-flow-strip-types"
--- a/index.js
+++ b/index.js
@@ -1,8 +1,5 @@
 // @flow
 
-// Polyfills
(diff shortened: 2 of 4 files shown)


babel.config.json
```diff
@@ -1,9 +1,13 @@
...
+    ],
     "plugins": [
         ["@babel/plugin-proposal-decorators", {"legacy": true}],
-        "@babel/plugin-proposal-object-rest-spread",
-        "@babel/plugin-transform-flow-strip-types",
-        ["@babel/plugin-proposal-class-properties", {"loose": true}]
-    ]
+        "@babel/plugin-transform-flow-strip-types"
+    ],
+    "assumptions": {
```
niklasnatter, line 10: this prevents the annoying `@babel/plugin-proposal-class-properties` warnings during build. we already added this on the `2.3` branch in #5993 🙂

index.js
```diff
@@ -1,8 +1,5 @@
 // @flow
 
-// Polyfills
-import 'regenerator-runtime/runtime';
```
niklasnatter, line 4: this will be added automatically by `@babel/preset-env` if necessary

babel.config.json
```diff
@@ -1,9 +1,13 @@
 {
-    "presets": ["@babel/preset-env", "@babel/preset-react"],
+    "presets": [
+        ["@babel/preset-env", {"useBuiltIns": "usage", "corejs": "3.18"}],
```
niklasnatter, line 3: without setting the `useBuiltIns` option, `@babel/preset-env` will only transform new syntax (eg. [optional chaining syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining)), but not polyfills for features like [`String.prototype.includes()`](https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/String/includes)

niklasnatter: interestingly, setting the `useBuiltIns` option does not add any polyfills to our build at the moment. the reason for this is that our `.browserslistrc` contains only modern browsers that support all of the features used in our code (and that the [`op_mini` target is ignored by `@babel/preset-env`](https://babeljs.io/docs/en/babel-preset-env#ineffective-browserslist-queries)).

niklasnatter: to test that the feature works, you can add `chrome 80` to the .browserslistrc. with this target, the `es.array.reduce ` polyfill is included in the build. it is also possible to see all included polyfills by enabling the `debug` option of `@babel/preset-env`:
```suggestion
        ["@babel/preset-env", {"useBuiltIns": "usage", "corejs": "3.18", "debug": true}],
```

alexander-schranz (approved):

PR #6248, opened 2021-09-24
Replace sulu_trash configuration with RestoreConfigurationProviderInterface services

niklasnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Related issues/PRs | builds upon #6218
| License | MIT

#### What's in this PR?

This PR adds a `RestoreConfigurationProviderInterface` that allows to set the restore configuration for a specific resource key. This replaces the previously used `sulu_trash` symfony configuration.

Providing this configuration prevents bundles from needing to prepend the `sulu_trash` configuration. Furthermore, we use a similar approach for collecting [smart content provider configurations](https://github.com/sulu/sulu/blob/a5efc81b692a5723910a2ec86d2651b1b79f04c9/src/Sulu/Component/SmartContent/DataProviderInterface.php#L27) and [teaser provider configurations](https://github.com/sulu/sulu/blob/a5efc81b692a5723910a2ec86d2651b1b79f04c9/src/Sulu/Bundle/PageBundle/Teaser/Provider/TeaserProviderInterface.php#L27).

### DIFF
--- a/src/Sulu/Bundle/TrashBundle/Resources/js/containers/RestoreFormOverlay/RestoreFormOverlay.js
+++ b/src/Sulu/Bundle/TrashBundle/Resources/js/containers/RestoreFormOverlay/RestoreFormOverlay.js
@@ -5,8 +5,9 @@ import React from 'react';
 import FormOverlay from 'sulu-admin-bundle/containers/FormOverlay';
 import {translate} from 'sulu-admin-bundle/utils/Translator';
 import {FormStoreInterface} from 'sulu-admin-bundle/containers/Form/types';
-import {memoryFormStoreFactory} from 'sulu-admin-bundle/containers/Form';
 import {ResourceRequester} from 'sulu-admin-bundle/services';
+import SchemaFormStoreDecorator from 'sulu-admin-bundle/containers/Form/stores/SchemaFormStoreDecorator';
+import MemoryFormStore from 'sulu-admin-bundle/containers/Form/stores/MemoryFormStore';
 
 type Props = {
     confirmLoading: boolean,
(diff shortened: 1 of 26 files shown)


src/Sulu/Bundle/TrashBundle/Resources/js/index.js
```diff
@@ -6,5 +6,5 @@ import RestoreItemAction from './views/List/itemActions/RestoreItemAction';
 listItemActionRegistry.add('sulu_trash.restore', RestoreItemAction);
 
 initializer.addUpdateConfigHook('sulu_trash', (config: Object) => {
-    RestoreItemAction.restoreFormMapping = config.restoreFormMapping;
+    RestoreItemAction.restoreConfigurationMapping = config.restoreConfigurationMapping;
```
niklasnatter, line 14: this should be adjusted to prevent the following issue: https://github.com/sulu/sulu/issues/6275

alexander-schranz: should this not be able to prevent with:

```js
if (!config.restoreConfigurationMapping) {
    return;
}
```

niklasnatter: yes, it should be quite easy to fix this. just added this comment as reminder 🙂

src/Sulu/Bundle/TrashBundle/Resources/js/containers/RestoreFormOverlay/RestoreFormOverlay.js
```diff
@@ -58,8 +59,15 @@ class RestoreFormOverlay extends React.Component<Props> {
             return;
         }
 
-        const formStore = memoryFormStoreFactory.createFromFormKey(formKey);
-        formStore.loading = true;
+        const formStore = new SchemaFormStoreDecorator(
+            (schema, jsonSchema) => {
+                const store = new MemoryFormStore({}, schema, jsonSchema);
+                store.loading = true;
```
niklasnatter, line 65: setting `store.loading` to `true` in the initializer function instead of the next statement prevents a race condition (loading is finished before next statement is executed) that leads to an infinite loader

alexander-schranz (approved):

PR #5399, opened 2020-07-09
Fix behaviour of crops when minimum dimensions are bigger than selectable area

nnatter:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #5372, fixes #5385
| License | MIT

#### What's in this PR?

This PR fixes two different issues that occur when defining crops for images:

#### Crop rectangle should change after selecting a format without a defined crop

At the moment, the selected area is not updated to match the ratio of the selected format, when a format that has no defined crop is selected. 

The problem here is that `RectangleSelection` component rendered its content before updating the `normalizers` that are used for determining the selected area in its `componentDidUpdate` method. After the `normalizers` were updated, the component was not rerendered because the `this.normalizers` property is no observable.

#### Incorrect behaviour when minimum dimensions are bigger than selectable area 

Currently, the `RectangleSelection` behaves incorrectly when the minimum dimensions are bigger than the selectable area. This is caused by the order in which the normalizers of are applied. For example, if i want to define the crop for an image with the dimensions `500x500` and a format with the dimensions `2000x1000`, the following will happen:

1. Current crop is set to `{left: 0, top: 0, height: 250, width: 500}`
2. User tries to set crop to `{left: 0, top: 10, height: 250, width: 500}` by moving selection
3. `SizeNormalizer` normalizes this to `{left: 0, top: 10, height: 500, width: 500}`
4. `PositionNormalizer` normalizes this to `{left: 0, top: 0, height: 500, width: 500}`
5. `RatioNormalizer` normalizes this to `{left: 0, top: 0, height: 250, width: 500}`
5. User thinks he cannot move selection towards bottom

This PR adjusts the order of the normalizers to apply the `RatioNormalizer` before the `PositionNormalizer`. This leads to the following behaviour:

1. Current crop is set to `{left: 0, top: 0, height: 250, width: 500}`
2. User tries to set crop to `{left: 0, top: 10, height: 250, width: 500}` by moving selection
3. `SizeNormalizer` normalizes this to `{left: 0, top: 10, height: 500, width: 500}`
4. `RatioNormalizer` normalizes this to `{left: 0, top: 3, height: 250, width: 500}`
5. `PositionNormalizer` normalizes this to `{left: 0, top: 3, height: 250, width: 500}`

PR #6249, opened 2021-09-24
Fix rectangle selection if container is smaller than given minValues

luca-rath:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

Fix rectangle selection if container is smaller than given minValues

#### Why?

It's important that the `RatioNormalizer` is added before the `PositionNormalizer`, because it may alter the width and height of the selection and the PositionNormalizer needs the final calculated width and height to correctly calculate all valid positions

#### Example Usage

```jsx
let image = new Image();
image.src = 'https://unsplash.it/640/360';

const [imageLoaded, setImageLoaded] = React.useState(image.complete);
const [selection, setSelection] = React.useState({width: 100, height: 100, top: 50, left: 50});

image.onload = () => setImageLoaded(true);

imageLoaded
    ? <div>
        <RectangleSelection
            minWidth={1280}
            minHeight={180}
            onChange={setSelection}
            value={selection}
        >
            <img src="https://unsplash.it/640/360"/>
        </RectangleSelection>

        <p>
            Width: {selection.width},
            Height: {selection.height},
            Top: {selection.top},
            Left: {selection.left}
        </p>
    </div>
    : null
```

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/RectangleSelection/RectangleSelection.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/RectangleSelection/RectangleSelection.js
@@ -88,13 +88,12 @@ class RawRectangleSelectionComponent extends React.Component<Props> {
                 minWidth,
                 minHeight
             ),
-            new PositionNormalizer(
-                containerWidth,
-                containerHeight
-            ),
         ];
 
         if (minWidth && minHeight) {
+            // It's important that the RatioNormalizer is added before the PositionNormalizer, because it may alter the
+            // width and height of the selection and the PositionNormalizer needs the final calculated width and height
+            // to correctly calculate all valid positions
             normalizers.push(
                 new RatioNormalizer(
                     containerWidth,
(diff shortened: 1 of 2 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/components/RectangleSelection/RectangleSelection.js
```diff
@@ -88,13 +88,12 @@ class RawRectangleSelectionComponent extends React.Component<Props> {
...
                 minHeight
             ),
-            new PositionNormalizer(
-                containerWidth,
-                containerHeight
-            ),
         ];
 
         if (minWidth && minHeight) {
+            // It's important that the RatioNormalizer is added before the PositionNormalizer, because it may alter the
```
niklasnatter, line 94: It looks like we already changed this in the past (#5399), but the changes got lost in some other PR 🙈 
Hopefully the comment will prevent this from happening again 🙂

PR #6250, opened 2021-09-27
Fix performance of media list query

### LINKED ISSUE #6238: Performance issue in the admin media area
| Q | A
| --- | ---
| Sulu Version | 2.3.1
| PHP Version | 7.4.22
| DB Version | 5.7.32
| Browser Version | All

# Actual Behavior

Loading the media area in the Sulu admin is really slow since our last update to Sulu 2.3.1 in a project with ~14.000 files. According to the profiler the request to `/admin/api/media?page=1&locale=de&fields=thumbnails,type,name,size,mimeType,title,id&limit=10&flat=true` results in at least one really slow query:

profiler

Running the following query takes approx. 25 seconds:

```
SELECT m0_.id AS id_0, m0_.created AS created_1 
FROM me_media m0_ 
LEFT JOIN me_collections m1_ ON m0_.idCollections = m1_.id 
LEFT JOIN se_access_controls s2_ ON (s2_.entityClass = 'Sulu\\Bundle\\MediaBundle\\Entity\\Collection' AND CAST(s2_.entityId AS char) = CAST(m1_.id AS char) AND s2_.idRoles IN (SELECT s3_.id FROM se_roles s3_ WHERE s3_.securitySystem = 'Sulu')) 
LEFT JOIN se_roles s4_ ON s2_.idRoles = s4_.id 
WHERE ((s2_.permissions & 64) = 64 OR s2_.permissions IS NULL) AND (s4_.id IN (1, 2) OR s4_.id IS NULL) 
ORDER BY created_1 DESC 
LIMIT 10;
```

After a hint from @alexander-schranz I added a new column with type integer to the table `su_access_controls` and added the `entity_id` casted to an integer to it (`UPDATE se_access_controls SET entityIdInt = CAST(entityId AS SIGNED);`). Running the query from above again without the casts and with the new column `entityIdInt` the query takes ~10 seconds. After adding an index to the new column the result is available in only a few milliseconds.

This does maybe not only affect the media area but also other parts of Sulu.

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? |  yes
| New feature? | no 
| BC breaks? | no 
| Deprecations? | no  
| Fixed tickets | fixes #6238
| License | MIT

#### What's in this PR?

Fix performance of media list query.

#### Why?

The casts are creating a performance issue so we need to avoid the problem.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,36 @@
 # Upgrade
 
+## 2.3.5
+
+### Migrate access control entityIdInteger field
+
+Because of performance problems, an additional integer representation of the
+entityId need to be added to the access control list.
+
+MySQL:
+
+```sql
+ALTER TABLE se_access_controls ADD entityIdInteger INT DEFAULT NULL;
+CREATE INDEX IDX_C526DC524473BB7A ON se_access_controls (entityIdInteger);
+
+UPDATE se_access_controls
+SET entityIdInteger = entityId
+WHERE entityIdInteger IS NULL
+    AND LENGTH(entityId) != 36;
+```
+
+PostgreSQL:
+
+```sql
+ALTER TABLE se_access_controls ADD entityIdInteger INT DEFAULT NULL;
+CREATE INDEX IDX_C526DC524473BB7A ON se_access_controls (entityIdInteger);
+
+UPDATE se_access_controls
+SET entityIdInteger = CAST(entityId AS int)
+WHERE entityIdInteger IS NULL
+    AND LENGTH(entityId) != 36;
+```
+
 ## 2.3.1
 
 ### Migrate permissions properties for pages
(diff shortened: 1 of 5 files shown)


alexander-schranz:
@sabinebaer can you test this PR: #6238 see also [UPGRADE.md the UPDATE statement](https://github.com/sulu/sulu/pull/6250/files#diff-434625111321abb225365806085a3753cb854d098ed9b4b6dc2001235c7f7bc4) is required!

UPGRADE.md
```diff
@@ -1,5 +1,22 @@
...
+
+Because of performance problems, an additional integer representation of the
+entityId need to be added to the access control list.
+
+```sql
+ALTER TABLE se_access_controls ADD entityIdInteger INT DEFAULT NULL;
+CREATE INDEX IDX_C526DC524473BB7A ON se_access_controls (entityIdInteger);
+
+UPDATE se_access_controls
+SET entityIdInteger = entityId -- postgresql:  `... = CAST(entityId AS int)`
```
wachterjohannes: place both SQLs into the upgrade file

wachterjohannes (approved):

UPGRADE.md
alexander-schranz: updated!

PR #6262, opened 2021-10-01
Set correct workflow stage and published date to shadow locale

niklasnatter:
| Q | A
| --- | ---
| Bug fix? | yes
| License | MIT

#### What's in this PR?

This PR adjusts the `WorkflowStageSubscriber` to use the original locale of the document to determine the workflow stage and the published date for a document.

#### Why?

At the moment, when loading a shadow page, the workflow stage and published date is set based on the data of the target locale.

#### How to reproduce

1. Create page in locale `de` and publish it
2. Switch to locale `en`, add a title and configure `en` as shadow page
3. Save the locale `en` as draft
3b. The form displays the correct draft status
4. Reload the administration interface
4b. The form now displays the locale as published, but the locale is not accessible on the website

### DIFF
--- a/src/Sulu/Component/Content/Document/Subscriber/WorkflowStageSubscriber.php
+++ b/src/Sulu/Component/Content/Document/Subscriber/WorkflowStageSubscriber.php
@@ -92,7 +92,7 @@ class WorkflowStageSubscriber implements EventSubscriberInterface
         }
 
         $node = $event->getNode();
-        $locale = $event->getLocale();
+        $locale = $this->documentInspector->getOriginalLocale($document);
 
         $document->setWorkflowStage(
             $node->getPropertyValueWithDefault(
--- a/src/Sulu/Bundle/PageBundle/Tests/Functional/SmartContent/QueryBuilderTest.php
+++ b/src/Sulu/Bundle/PageBundle/Tests/Functional/SmartContent/QueryBuilderTest.php
@@ -1529,6 +1529,20 @@ class QueryBuilderTest extends SuluTestCase
                 Structure::STATE_TEST
             )
         );
+        $nodesEn = \array_merge(
+            $nodesEn,
+            $this->save(
+                [
+                    'title' => 'Alex',
+                    'url' => '/team/alex',
+                ],
+                'en',
+                null,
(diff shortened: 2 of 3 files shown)


src/Sulu/Component/Content/Document/Subscriber/WorkflowStageSubscriber.php
```diff
@@ -92,7 +92,7 @@ public function setWorkflowStageOnDocument(HydrateEvent $event)
         }
 
         $node = $event->getNode();
-        $locale = $event->getLocale();
+        $locale = $this->documentInspector->getOriginalLocale($document);
```
alexander-schranz, line 95: Looks strange for me, to go always on the document original locale in this listener. What value did have the `$event->getLocale()`? and what value did have the `$this->documentInspector->getOriginalLocale($document);` return?
Should it not be a fallback done if the locale is not correct in that case e.g.:

```suggestion
        $locale = $event->getLocale() ?: $this->documentInspector->getOriginalLocale($document);
```

niklasnatter: if you are loading the locale `en` that is a shadow page with the target `de`, the values look like this:
```
$event->getLocale() -> de
$this->documentInspector->getOriginalLocale($document) -> en
```

the reason for this is the `ShadowLocaleSubscriber`:
https://github.com/sulu/sulu/blob/bb4e0bad01a3bafd7cb9d162e377a159928832e0/src/Sulu/Component/Content/Document/Subscriber/ShadowLocaleSubscriber.php#L128-L131

niklasnatter: i got the idea to use the `DocumentInspector` from the `ResourceSegmentSubscriber`:
https://github.com/sulu/sulu/blob/bb4e0bad01a3bafd7cb9d162e377a159928832e0/src/Sulu/Component/Content/Document/Subscriber/ResourceSegmentSubscriber.php#L133-L140

should we add a fallback there too?

alexander-schranz (approved):

PR #6265, opened 2021-10-01
Fix media properties width and height for svg

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? |  yes
| New feature? | no 
| BC breaks? | no 
| Deprecations? | no  
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Fix media properties width and height for svg

#### Why?

Currently it tries to use the none svg adapter for svg images which ends that in gd case it will fail to read the width and heights correctly.

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/Media/PropertiesProvider/ImagePropertiesProvider.php
+++ b/src/Sulu/Bundle/MediaBundle/Media/PropertiesProvider/ImagePropertiesProvider.php
@@ -23,9 +23,12 @@ class ImagePropertiesProvider implements MediaPropertiesProviderInterface
      */
     private $imagine;
 
-    public function __construct(ImagineInterface $imagine)
-    {
+    public function __construct(
+        ImagineInterface $imagine,
+        ?ImagineInterface $svgImagine = null
+    ) {
         $this->imagine = $imagine;
+        $this->svgImagine = $svgImagine;
     }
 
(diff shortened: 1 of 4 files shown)


src/Sulu/Bundle/MediaBundle/Media/PropertiesProvider/ImagePropertiesProvider.php
```diff
@@ -23,9 +23,12 @@ class ImagePropertiesProvider implements MediaPropertiesProviderInterface
...
     private $imagine;
 
-    public function __construct(ImagineInterface $imagine)
-    {
+    public function __construct(
+        ImagineInterface $imagine,
+        ?ImagineInterface $svgImagine = null
+    ) {
         $this->imagine = $imagine;
+        $this->svgImagine = $svgImagine;
```
niklasnatter, line 36: we should define this property on the class (see phpstan error )

niklasnatter (commented):
Looks good otherwise 🙂

src/Sulu/Bundle/MediaBundle/Media/PropertiesProvider/ImagePropertiesProvider.php
alexander-schranz: done!

PR #6266, opened 2021-10-01
Integrate Analytics with SuluTrashBundle

luca-rath:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | #5150, #6248
| License | MIT
| Documentation PR | -

#### What's in this PR?

Integrate Analytics with SuluTrashBundle

**This PR depends on #6248**

### DIFF
--- a/src/Sulu/Bundle/WebsiteBundle/Trash/AnalyticsTrashItemHandler.php
+++ b/src/Sulu/Bundle/WebsiteBundle/Trash/AnalyticsTrashItemHandler.php
@@ -1,0 +1,237 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of Sulu.
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\WebsiteBundle\Trash;
+
+use Doctrine\Common\Collections\Collection;
+use Doctrine\ORM\EntityManagerInterface;
+use Doctrine\ORM\NoResultException;
+use Sulu\Bundle\ActivityBundle\Application\Collector\DomainEventCollectorInterface;
+use Sulu\Bundle\TrashBundle\Application\DoctrineRestoreHelper\DoctrineRestoreHelperInterface;
+use Sulu\Bundle\TrashBundle\Application\RestoreConfigurationProvider\RestoreConfiguration;
+use Sulu\Bundle\TrashBundle\Application\RestoreConfigurationProvider\RestoreConfigurationProviderInterface;
+use Sulu\Bundle\TrashBundle\Application\TrashItemHandler\RestoreTrashItemHandlerInterface;
+use Sulu\Bundle\TrashBundle\Application\TrashItemHandler\StoreTrashItemHandlerInterface;
+use Sulu\Bundle\TrashBundle\Domain\Model\TrashItemInterface;
+use Sulu\Bundle\TrashBundle\Domain\Repository\TrashItemRepositoryInterface;
+use Sulu\Bundle\WebsiteBundle\Admin\WebsiteAdmin;
+use Sulu\Bundle\WebsiteBundle\Analytics\AnalyticsManager;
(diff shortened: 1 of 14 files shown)


src/Sulu/Bundle/WebsiteBundle/Trash/AnalyticsTrashItemHandler.php
```diff
@@ -0,0 +1,237 @@
...
+     * @return mixed
+     */
+    private function getValue(array $data, string $name, $default = null)
+    {
+        if (!\array_key_exists($name, $data)) {
+            return $default;
+        }
+
+        return $data[$name];
+    }
```
luca-rath: These methods are pretty much the same as in the `AnalyticsManager` ... Should we share them somehow?

luca-rath:
@alexander-schranz Du you have an idea why the parameter `kernel.bundles` is not available in the `SuluWebsiteExtension::load()`?

niklasnatter:
i think you could rebase this onto `2.x` 🙂

src/Sulu/Bundle/WebsiteBundle/Trash/AnalyticsTrashItemHandler.php
```diff
@@ -0,0 +1,237 @@
...
+        if ($analytics->isAllDomains()) {
+            $analytics->clearDomains();
+
+            return;
+        }
+
+        $domains = [];
+        $domainCollection = $analytics->getDomains();
+        if ($domainCollection) {
+            $domains = $domainCollection->toArray();
```
niklasnatter: is there a case where this is not empty?

src/Sulu/Bundle/WebsiteBundle/Trash/AnalyticsTrashItemHandler.php
```diff
@@ -0,0 +1,237 @@
...
+    private function setData(AnalyticsInterface $analytics, array $data): void
+    {
+        $analytics->setTitle($this->getValue($data, 'title'));
+        $analytics->setType($this->getValue($data, 'type'));
+        $analytics->setWebspaceKey($this->getValue($data, 'webspaceKey'));
+        $analytics->setContent($this->getValue($data, 'content', ''));
+        $analytics->setAllDomains($this->getValue($data, 'allDomains', false));
+
+        if ($analytics->isAllDomains()) {
+            $analytics->clearDomains();
```
niklasnatter: this will always be empty at this point, right?

src/Sulu/Bundle/WebsiteBundle/Trash/AnalyticsTrashItemHandler.php
```diff
@@ -0,0 +1,237 @@
...
+                continue;
+            }
+
+            $domainEntity = $this->findOrCreateNewDomain($domain);
+            $analytics->addDomain($domainEntity);
+        }
+
+        foreach ($domains as $domain) {
+            $domainEntity = $this->findOrCreateNewDomain($domain);
+            $analytics->removeDomain($domainEntity);
```
niklasnatter: i dont think this can happen when restoring a domain

src/Sulu/Bundle/WebsiteBundle/Trash/AnalyticsTrashItemHandler.php
niklasnatter: personally, i would not try to do that. the code for creating the entity looks straight forward, i think its okay to do that in two different places. extracting something reusable means we need to maintain it in the future

niklasnatter: also, it should be possible to simplify the code in this handler because some cases will never happen when restoring an entity 🙂

alexander-schranz:
@luca-rath here you need to define which parameters you expect to exist for the ExtensionTests: https://github.com/sulu/sulu/blob/3d19a0e66dc85a3ab5db8f0605c322bf9a1f3610/src/Sulu/Bundle/WebsiteBundle/Tests/Unit/DependencyInjection/SuluWebsiteExtensionTest.php#L28

niklasnatter (approved):

PR #6268, opened 2021-10-01
Integrate the TrashBundle functionality for the Page entity

niklasnatter:
| Q | A
| --- | ---
| Related issues/PRs | builds upon #6248
| License | MIT

#### What's in this PR?

This PR integrates the `SuluTrashBundle` with the built-in page entity. With the changes in this PR, pages will be moved into the trash when being deleted and can be restored from the trash at a later point in time.


src/Sulu/Component/Content/Document/Subscriber/ShadowLocaleSubscriber.php
```diff
@@ -204,10 +204,12 @@ public function handlePersistUpdateUrl(PersistEvent $event)
             return;
         }
 
-        $event->getAccessor()->set(
-          'resourceSegment',
-          $shadowLocator
-        );
+        if (!$event->getAccessor()->get('resourceSegment')) {
```
niklasnatter, line 207: we want to set the resource segment to the segment of the shadow locale only if the document has no segment set. if the document already has a segment, we dont want to overwrite it

src/Sulu/Component/Content/Document/Subscriber/BlameSubscriber.php
```diff
@@ -147,11 +147,9 @@ public function setBlamesOnNode(
 
         $creatorPropertyName = $this->propertyEncoder->encode($encoding, static::CREATOR, $locale);
         if (!$node->hasProperty($creatorPropertyName)) {
-            $accessor->set(self::CREATOR, $userId);
-            $node->setProperty(
-                $creatorPropertyName,
-                $document->getCreator()
-            );
+            $creator = $document->getCreator() ?: $userId;
```
niklasnatter, line 150: we want to use the creator from the document, if it is set. we already use the same strategy in the `TimestampSubscriber`:
https://github.com/sulu/sulu/blob/1af4ee3cc12dd9ecca2a39f90ec57d136326e677/src/Sulu/Component/DocumentManager/Subscriber/Behavior/Audit/TimestampSubscriber.php#L150-L154

src/Sulu/Bundle/PageBundle/EventListener/PageTrashSubscriber.php
```diff
@@ -0,0 +1,70 @@
...
+        if (!$document instanceof BasePageDocument) {
+            return;
+        }
+
+        $this->trashManager->store(BasePageDocument::RESOURCE_KEY, $event->getDocument());
+    }
+
+    public function flushTrashItem(FlushEvent $event): void
+    {
+        $this->entityManager->flush();
```
alexander-schranz, line 83:
```suggestion
        if ($this->requiresFlush) {
            $this->entityManager->flush();
        }
```

There are 1 additional event which need to reset the flag:

```php
            Events::RESET => 'reset',
```

```php
public function resetFlag()
{
     $this->requiresFlush = false;
}
```

src/Sulu/Bundle/PageBundle/EventListener/PageTrashSubscriber.php
```diff
@@ -0,0 +1,70 @@
...
+
+use Doctrine\ORM\EntityManagerInterface;
+use Sulu\Bundle\PageBundle\Document\BasePageDocument;
+use Sulu\Bundle\TrashBundle\Application\TrashManager\TrashManagerInterface;
+use Sulu\Component\DocumentManager\Event\FlushEvent;
+use Sulu\Component\DocumentManager\Event\RemoveEvent;
+use Sulu\Component\DocumentManager\Events;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+class PageTrashSubscriber implements EventSubscriberInterface
```
alexander-schranz:
```suggestion
/**
 * @internal
 * @final
 */
class PageTrashSubscriber implements EventSubscriberInterface
```

Would mark this as internal and final.

src/Sulu/Bundle/PageBundle/Trash/PageTrashItemHandler.php
```diff
@@ -0,0 +1,212 @@
...
+        $this->documentInspector = $documentInspector;
+        $this->documentDomainEventCollector = $documentDomainEventCollector;
+    }
+
+    /**
+     * @param PageDocument $page
+     */
+    public function store(object $page): TrashItemInterface
+    {
+        Assert::isInstanceOf($page, PageDocument::class);
```
alexander-schranz: Why not `BasePageDocument`?

niklasnatter: because its not really possible to remove the homepage via the interface 🙂  should we move the homepage to the trash too?

src/Sulu/Bundle/PageBundle/EventListener/PageTrashSubscriber.php
niklasnatter: 👍

src/Sulu/Bundle/PageBundle/EventListener/PageTrashSubscriber.php
niklasnatter: 👍

src/Sulu/Bundle/PageBundle/Trash/PageTrashItemHandler.php
alexander-schranz: Would adjust the `PageTrashSubscriber.php` to check for PageDocument or allow here `BasePageDocument`.

niklasnatter: adjusted the `PageTrashItemHandler` to allow `BasePageDocument` 🙂

alexander-schranz (approved):

PR #6286, opened 2021-10-06
Initialize reflection on parent metadata in MetadataSubscriber of PersistenceBundle

### LINKED ISSUE #6279: MetadataSubscriber constructing parent class metadata without reflection class
| Q | A
| --- | ---
| Sulu Version | 2.3.5
| PHP Version | 8.0.11
| DB Version | 10.5.8-MariaDB
| Browser Version | Firefox 93.0b9 (64-Bit)

# Actual Behavior

`Sulu\Component\Persistence\EventSubscriber\ORM\MetadataSubscriber` creates `Doctrine\ORM\Mapping\ClassMetadata` instances for parent classes of a given entity, but the instances have no reflection class value:
```php
// Sulu\Component\Persistence\EventSubscriber\ORM\MetadataSubscriber.php:88-98
foreach (\class_parents($metadata->getName()) as $parent) {
    $parentMetadata = new ClassMetadata(
        $parent,
        $configuration->getNamingStrategy()
    );
    // $parentMetadata->getReflectionClass() === null
    // ...
    $configuration->getMetadataDriverImpl()->loadMetadataForClass($parent, $parentMetadata);
}
```

When using PHP8 attributes for Doctrine ORM, `EntityManager::getConfiguration::getMetadataDriverImpl` returns `Doctrine\ORM\Mapping\Driver\AttributeDriver`, whose `loadMetadataForClass` method requires a `$reflClass` field value, else it throws an exception:

```php
// Doctrine\ORM\Mapping\Driver\AttributeDriver.php:58-64
public function loadMetadataForClass($className, ClassMetadata $metadata): void
{
    assert($metadata instanceof ClassMetadataInfo);

    $reflectionClass = $metadata->getReflectionClass();

    $classAttributes = $this->reader->getClassAnnotations($reflectionClass); // throws Exception: Doctrine\ORM\Mapping\Driver\AttributeReader::getClassAnnotations(): Argument #1 ($class) must be of type ReflectionClass, null given, called in /path/to/project/vendor/doctrine/orm/lib/Doctrine/ORM/Mapping/Driver/AttributeDriver.php on line 64
    
    // ...
}
```

# Expected Behavior

`Sulu\Component\Persistence\EventSubscriber\ORM\MetadataSubscriber` should create parent class metadata with a reflection class instance so that `Doctrine\ORM\Mapping\Driver\AttributeDriver` can properly process class annotations

# Steps to Reproduce

1. Configure Doctrine ORM to use PHP8 attributes:
```yaml
# /path/to/project/config/packages/doctrine.yaml
doctrine:
# ...
  orm:
    # ...
    mappings:
      App:
        # ...
        type: attribute
```
2. Extend `Sulu\Bundle\SecurityBundle\Entity\User` with a class table inheritance entity and create corresponding entities for the descriminator map. Use PHP8 attributes for ORM definition:
```php
#[ORM\Entity]
#[ORM\Table(name: 'se_users')]
#[ORM\InheritanceType('JOINED')]
#[ORM\DiscriminatorColumn(name: 'user_type', type:'string', length: 50)]
#[ORM\DiscriminatorMap([
    'a' => ATypeUser::class,
    'b' => BTypeUser::class,
])]
abstract class User extends SuluUser
{
    // ...
}

#[ORM\Entity]
#[ORM\Table(name: 'proj_user_a')]
class ATypeUser extends User
{
    // ...
}

#[ORM\Entity]
#[ORM\Table(name: 'proj_user_b')]
class BTypeUser extends User
{
    // ...
}
```
3. Perform any activity that involves class metadata handling - in my case it was enough to try to generate a new Doctrine migration file with `php bin/adminconsole doc:mig:diff`.

# Possible Solutions

* `Sulu\Component\Persistence\EventSubscriber\ORM\MetadataSubscriber` could use `Doctrine\ORM\Mapping\ClassMetadataFactory::loadMetadataFor` to create parent class metadata, which already instantiates a `ReflectionClass` field value.

R3dLantern:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #6279
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

Make `MetadataSubscriber` utilize `ClassMetadataFactory` to create `ClassMetadata` objects of parent classes instead of creating them via `ClassMetadata::__construct`.
#### Why?

`MetadataSubscriber` created `ClassMetadata` objects of parent classes without a `$reflClass` property value. As of now, while Doctrine's `AnnotationDriver` has a hacky workaround, the new `AttributeDriver` for PHP8 attributes strictly requires the reflection class to exist in the metadata object (see #6279 for more details).
Creating a `ClassMetadata` object with `ClassMetadataFactory::getMetadataFor` initializes a `ReflectionClass` object as the `$reflClass` property value, preventing an exception from being thrown.

### DIFF
--- a/src/Sulu/Component/Persistence/EventSubscriber/ORM/MetadataSubscriber.php
+++ b/src/Sulu/Component/Persistence/EventSubscriber/ORM/MetadataSubscriber.php
@@ -18,6 +18,7 @@ use Doctrine\ORM\Events;
 use Doctrine\ORM\Mapping\ClassMetadata;
 use Doctrine\ORM\Mapping\ClassMetadataInfo;
 use Doctrine\ORM\Mapping\ClassMetadataFactory;
+use Doctrine\Persistence\Mapping\ReflectionService;
 
 /**
  * Doctrine subscriber used to manipulate metadata.
@@ -60,7 +61,11 @@ class MetadataSubscriber implements EventSubscriber
 
         if (!$metadata->isMappedSuperclass) {
             $em = $event->getEntityManager();
-            $this->setAssociationMappings($metadata, $em->getMetadataFactory(), $em->getConfiguration());
+            $this->setAssociationMappings(
+                $metadata,
+                $em->getConfiguration(),
+                $em->getMetadataFactory()->getReflectionService()
+            );
         } else {
             $this->unsetAssociationMappings($metadata);
         }
@@ -83,8 +88,8 @@ class MetadataSubscriber implements EventSubscriber
 
     private function setAssociationMappings(
         ClassMetadataInfo $metadata,
-        ClassMetadataFactory $classMetadataFactory,
-        Configuration $configuration
+        Configuration $configuration,
+        ReflectionService $reflectionService
     ) {
         if (!\class_exists($metadata->getName())) {
             return;
@@ -94,9 +99,9 @@ class MetadataSubscriber implements EventSubscriber
             if (!\in_array($parent, $this->getAllClassNames($configuration))) {
                 continue;
             }
-            
-            /** @var ClassMetadata $parentMetadata */
-            $parentMetadata = $classMetadataFactory->getMetadataFor($parent);
+
+            $parentMetadata = new ClassMetadata($parent, $configuration->getNamingStrategy());
+            $parentMetadata->initializeReflection($reflectionService);
 
             $configuration->getMetadataDriverImpl()->loadMetadataForClass($parent, $parentMetadata);
             if (!$parentMetadata->isMappedSuperclass) {
--- a/src/Sulu/Component/Persistence/EventSubscriber/ORM/MetadataSubscriber.php
+++ b/src/Sulu/Component/Persistence/EventSubscriber/ORM/MetadataSubscriber.php
@@ -91,12 +91,12 @@ class MetadataSubscriber implements EventSubscriber
         }
 
         foreach (\class_parents($metadata->getName()) as $parent) {
-            /** @var ClassMetadata $parentMetadata */
-            $parentMetadata = $classMetadataFactory->getMetadataFor($parent);
-
             if (!\in_array($parent, $this->getAllClassNames($configuration))) {
                 continue;
             }
+            
+            /** @var ClassMetadata $parentMetadata */
+            $parentMetadata = $classMetadataFactory->getMetadataFor($parent);
 
             $configuration->getMetadataDriverImpl()->loadMetadataForClass($parent, $parentMetadata);
             if (!$parentMetadata->isMappedSuperclass) {
--- a/src/Sulu/Component/Persistence/EventSubscriber/ORM/MetadataSubscriber.php
+++ b/src/Sulu/Component/Persistence/EventSubscriber/ORM/MetadataSubscriber.php
@@ -17,6 +17,7 @@ use Doctrine\ORM\Event\LoadClassMetadataEventArgs;
 use Doctrine\ORM\Events;
 use Doctrine\ORM\Mapping\ClassMetadata;
 use Doctrine\ORM\Mapping\ClassMetadataInfo;
+use Doctrine\ORM\Mapping\ClassMetadataFactory;
 
 /**
  * Doctrine subscriber used to manipulate metadata.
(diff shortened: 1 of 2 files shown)


niklasnatter:
Hey,
thanks for creating this pull request. Unfortunately, the changes of the pull request crash the application at the moment (see failing pipelines). Could you have another look at this?

src/Sulu/Component/Persistence/EventSubscriber/ORM/MetadataSubscriber.php
```diff
@@ -79,17 +81,18 @@ private function process(ClassMetadataInfo $metadata)
...
             return;
         }
 
         foreach (\class_parents($metadata->getName()) as $parent) {
-            $parentMetadata = new ClassMetadata(
-                $parent,
-                $configuration->getNamingStrategy()
-            );
+            /** @var ClassMetadata $parentMetadata */
+            $parentMetadata = $classMetadataFactory->getMetadataFor($parent);
```
alexander-schranz: I think you need to move this after:

```php
            if (!\in_array($parent, $this->getAllClassNames($configuration))) {
                continue;
            }
            
            $parentMetadata = $classMetadataFactory->getMetadataFor($parent); // <--
```

R3dLantern: Yes, I think that makes sense. Commit coming up

niklasnatter:
Thanks for the fast update! I tried to overwrite the `Account` entity with the following files:

**config/packages/sulu_contact.yaml:**
```yaml
sulu_contact:
    objects:
        account:
            model: App\Entity\Account
```

**src/Entity/Account.php:**
```php
<?php

declare(strict_types=1);

namespace App\Entity;

use Doctrine\ORM\Mapping as ORM;
use Sulu\Bundle\ContactBundle\Entity\Account as SuluAccount;

/**
 * @ORM\Table(name="co_accounts")
 * @ORM\Entity
 */
class Account extends SuluAccount
{
    /**
     * @var string|null
     *
     * @ORM\Column(name="additionalProperty", type="string", nullable=true)
     */
    private $additionalProperty = null;
}
```

Unfortunately, this leads the to following error when executing `bin/console doctrine:schema:update --dump-sql --complete`:

```
In MappingException.php line 658:
                                                                                                                                         
  Duplicate definition of column 'lft' on entity 'Sulu\Bundle\ContactBundle\Entity\Account' in a field or discriminator column mapping.  
```

R3dLantern:
@nnatter Thanks for the feedback, I'll look into that shortly. Is there a specific reason extensible entities are declared as mapped superclasses?

niklasnatter:
No, unfortunately I am not sure if there is a specific reason for that 😕  
This part of the codebase was not really changed in the recent years - maybe this was a requirement at the time it was wrote and there would be better solutions now. Unfortunately I am not familiar with the inner workings of doctrine to assess that.

alexander-schranz:
> Is there a specific reason extensible entities are declared as mapped superclasses?

@scarcloud They are by design mapped superclasses (inheritance mapping https://www.doctrine-project.org/projects/doctrine-orm/en/2.9/reference/inheritance-mapping.html#mapped-superclasses) as they are overwriteable .

See https://docs.sulu.io/en/2.3/cookbook/extend-entities.html

All entities which are overwriteable use the mapped superclass. To avoid Base mapped Superclass and a Entity Implementation this `MetadataSubscriber` was created as doctrine has some limitations here. Which will on container build convert the mapped superclass if not extended to a entity. If it is extend it will keep it as mapped superclass and move some properties to the extend entity.

So yes this are by design mapped superclasses and this should not be changed. We should just make sure to support the new metadata and set it correctly. It is important that https://docs.sulu.io/en/2.3/cookbook/extend-entities.html still works after the changes we are doing here.

src/Sulu/Component/Persistence/EventSubscriber/ORM/MetadataSubscriber.php
alexander-schranz: Looks like it works this way can you fix the PHP Lint error?

R3dLantern:
@alexander-schranz Thanks for the detailed answer, now a lot of things make sense. 😄

Since `AbstractClassMetadataFactory::loadMetadataFor` seems to be rather strict about constructing new metadata objects and, as of #6279, the only problem is the missing reflection class in case of attributes, I think it's for the better to keep the manual construction, but also manually initialize the reflection class. commit in a sec.

niklasnatter (approved):
I have tested the new changes with the example in https://github.com/sulu/sulu/pull/6286#issuecomment-937798698 and they lead to the expected result. I have pushed an additional commit to fix the linting errors. Thanks a lot for working on this!

alexander-schranz:
@scarcloud Thank you for your contribution 👍

PR #6288, opened 2021-10-07
Remount List view on login to prevent infinite loading screen after 401 response

### LINKED ISSUE #6199: List view crashes when user is logged out because of expired session
| Q | A
| --- | ---
| Sulu Version | 2.*

# Actual Behavior

Sulu registers a request listener that listens for `401` responses and overlays the current view with the `Login` view if such a response is detected. The `Login` view is displayed over the current view intentionally to allow the user to login again without losing the inputted data if his session expires while editing a form. 

Unfortunately, the `List` view does not handle errors gracefully at the moment. If the `List` view receives a `401` response, the view will just keep displaying a loading screen. In combination with the request listener explained above, this means that **the user will see an infinite loading screen after logging in again if his session expired before opening the `List` view**. 

For completion, the error can easily be solved by reloading the page in the browser. But an inexperienced user does not know this and will be stuck on an infinite loading screen.

# Expected Behavior / Possible Solutions

There are two things that could be improved:

1. The `List` view should handle errors gracefully. If an error happens while loading the (meta-)data of the list, the view should display a helpful error message instead of displaying a loader infinitely. This error message could include a button that allows to reload the data of the list.

2. The `List` view could observe the `UserStore::loggedIn` property. If the value of the property changes from `false` to `true`, the view could automatically reload the data of the current page.

# Steps to Reproduce

1. Login into the administration interface
2. Remove the `SULUSESSID` cookie
3. Navigate to the contact list
4. Login via the displayed login form
5. RESULT: The `List` view displays a loader infinitely

niklasnatter:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Fixed tickets | fixes #6199
| License | MIT

#### What's in this PR?

This pull request adds an optional `remountViewOnLogin` property on the `View` type definition. If the `remountViewOnLogin` property is set, the view will be remounted if a login happens. A remount will completely reinitialize the view.

#### Why?

To prevent an infinite loading screen in the `List` view when logging in after a 401 response. See #6199 for a full explanation.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/ViewRenderer/ViewRenderer.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/ViewRenderer/ViewRenderer.js
@@ -1,7 +1,9 @@
 // @flow
 import React from 'react';
 import {observer} from 'mobx-react';
+import {observable, reaction} from 'mobx';
 import Router, {getViewKeyFromRoute, Route} from '../../services/Router';
+import userStore from '../../stores/userStore';
 import viewRegistry from './registries/viewRegistry';
 import type {Element} from 'react';
 import type {View} from './types';
@@ -14,6 +16,10 @@ const UPDATE_ROUTE_HOOK_PRIORITY = 1024;
 
 @observer
 class ViewRenderer extends React.Component<Props> {
+    @observable loginCount: number = 0;
+
+    updateLoginCountDisposer: ?() => *;
+
     componentDidMount() {
         const {router} = this.props;
 
@@ -25,6 +31,21 @@ class ViewRenderer extends React.Component<Props> {
 
             return true;
         }, UPDATE_ROUTE_HOOK_PRIORITY);
+
+        this.updateLoginCountDisposer = reaction(
+            () => (userStore.loggedIn),
+            (newIsLoggedIn) => {
+                if (newIsLoggedIn) {
+                    this.loginCount = this.loginCount + 1;
+                }
+            }
+        );
+    }
+
+    componentWillUnmount() {
+        if (this.updateLoginCountDisposer) {
+            this.updateLoginCountDisposer();
+        }
     }
 
     getView = (route: Route): View => {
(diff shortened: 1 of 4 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/containers/ViewRenderer/types.js
```diff
@@ -13,4 +13,8 @@ interface GetDerivedRouteAttributesInterface {
     +getDerivedRouteAttributes?: (route: Route, attributes: AttributeMap) => Object,
 }
 
-export type View = Class<Component<ViewProps & *>> & GetDerivedRouteAttributesInterface;
+interface RemountViewOnLoginInterface {
+    +remountViewOnLogin?: boolean,
```
niklasnatter, line 17: we could also make this opt-out instead of opt-in by using a `preventRemountOnLogin` property. i guess that remounting the view is the better behaviour for most views. the only exception that comes to my mind is the form view that should keep the values inputted by the user.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/ViewRenderer/ViewRenderer.js
```diff
@@ -41,8 +62,13 @@ class ViewRenderer extends React.Component<Props> {
...
         const View = this.getView(route);
 
+        let viewKey = getViewKeyFromRoute(route, router.attributes);
+        if (View.remountViewOnLogin) {
+            viewKey = viewKey + '__' + this.loginCount;
+        }
+
         const element = (
-            <View key={getViewKeyFromRoute(route, router.attributes)} route={route} router={router}>
+            <View key={viewKey} route={route} router={router}>
```
alexander-schranz, line 71: Will this not rerender the whole View and all inputted data get lost in the forms. Which we actively wanted to avoid.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/ViewRenderer/types.js
alexander-schranz: I would stay with OptIn instead of OptOut.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/ViewRenderer/ViewRenderer.js
alexander-schranz: Oh now I understand, how it works now. Pagination stays the same. So page 2 works on the list also and active filters?

niklasnatter: no, because the `Form` view does not set the `remountViewOnLogin` property :slightly_smiling_face:

alexander-schranz (approved):

PR #6295, opened 2021-10-12
Using the DocumentMangerInterface in every constructor

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no (maintenance)
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?
This pull request replaces all usages of the Sulu hardcoded document manger with the respective interface. I have only changed the constructor calls as changing the interface in the fixtures for example would cause a BC break.

#### Why?
I was trying to add debug information to the DocumentManager by decorating it and noticed that the application hard hard dependencies on the Sulu class and not the correct interface.

#### Example Usage
Use code like this:
```php
use Sulu\Component\DocumentManager\DocumentManagerInterface

class SomeClass {
    public function __construct(DocumentManagerInterface $documentManager) {
    }
}
```
And not like this:
```php
use Sulu\Component\DocumentManager\DocumentManager;

class SomeClass {
    public function __construct(DocumentManager $documentManager) {
    }
}
```

#### To Do
In the future there should be more classes that need to be abstracted away by interfaces to make the system easier to extend.

### DIFF
--- a/src/Sulu/Component/Content/Form/DataTransformer/DocumentToUuidTransformer.php
+++ b/src/Sulu/Component/Content/Form/DataTransformer/DocumentToUuidTransformer.php
@@ -60,7 +60,6 @@ class DocumentToUuidTransformer implements DataTransformerInterface
         }
 
         $document = $this->documentManager->find($uuid);
-
         if (null === $document) {
             throw new TransformationFailedException(\sprintf(
                 'Could not find document with UUID "%s"', $uuid
--- a/src/Sulu/Component/Content/Form/DataTransformer/DocumentToUuidTransformer.php
+++ b/src/Sulu/Component/Content/Form/DataTransformer/DocumentToUuidTransformer.php
@@ -13,15 +13,18 @@ namespace Sulu\Component\Content\Form\DataTransformer;
 
 use PHPCR\Util\UUIDHelper;
 use Sulu\Component\DocumentManager\Behavior\Mapping\UuidBehavior;
-use Sulu\Component\DocumentManager\DocumentManager;
+use Sulu\Component\DocumentManager\DocumentManagerInterface;
 use Symfony\Component\Form\DataTransformerInterface;
 use Symfony\Component\Form\Exception\TransformationFailedException;
 
 class DocumentToUuidTransformer implements DataTransformerInterface
 {
+    /**
+     * @var DocumentManagerInterface
+     */
     private $documentManager;
 
-    public function __construct(DocumentManager $documentManager)
+    public function __construct(DocumentManagerInterface $documentManager)
     {
         $this->documentManager = $documentManager;
     }
--- a/src/Sulu/Component/DocumentManager/DocumentManagerInterface.php
+++ b/src/Sulu/Component/DocumentManager/DocumentManagerInterface.php
@@ -74,9 +74,10 @@ interface DocumentManagerInterface
 
     /**
      * Re-Order node before or after a specific node.
+     * If the destId is null, then the document will be sorted at the end.
      *
      * @param object $document
-     * @param string $destId
+     * @param string|null $destId
      */
     public function reorder($document, $destId);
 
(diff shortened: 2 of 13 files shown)


alexander-schranz:
This change is fine for me, can you rebase it to `2.2` and target the `2.2` branch?
You maybe need to update the phpstan baseline as some parameters only exist from bc case  in the documentmanager class and not the interface and phpstan don't like additional parameters. 

```
php8.0 vendor/bin/phpstan analyse --generate-baseline
```

mamazu:
I have rebased it on the correct branch and I will regenerate the baseline if the tests fail.

alexander-schranz:
@mamazu Thx you need to repush to trigger the CI after branch change.

src/Sulu/Component/Content/Form/DataTransformer/DocumentToUuidTransformer.php
```diff
@@ -56,7 +60,11 @@ public function reverseTransform($uuid)
             ));
         }
 
-        $document = $this->documentManager->find($uuid);
+        try {
+            $document = $this->documentManager->find($uuid);
+        } catch (DocumentNotFoundException $exception) {
+            $document = null;
+        }
```
mamazu: This is a thing I added because of the interface. The document manager always returns an object (according to the type hints) and the code. The original problem was that the following `if` statement would never hit because if the document exists then the `DocumentManager` would have thrown an exception. And adding null to the interface of the `DocumentManager` sounds a) incorrect and b) will spawn a lot of other type errors.

src/Sulu/Component/DocumentManager/DocumentManagerInterface.php
```diff
@@ -76,7 +76,7 @@ public function copy($document, $destPath);
      * Re-Order node before or after a specific node.
      *
      * @param object $document
-     * @param string $destId
+     * @param string|null $destId if the destination id is null, then it will be sorted at the end
```
mamazu: This needs to be verified. That's what I got from the code that was using it with null: `src/Sulu/Component/Content/Mapper/ContentMapper.php:580`

mamazu: Yes it does.

mamazu:
@alexander-schranz I have updated the merge request. It now also includes:
* Fixing the `2nd argument is null in DocumentManager::reorder`. Which is done by accepting null everywhere in the typehint and adding proper explanation what `null` means.
* Adding a `try`-`catch` block for the `DocumentManager::find` method because it doesn't return null when the document isn't found. If you want me remove it and replace it with a phpstan ignore just comment that.

src/Sulu/Component/Content/Form/DataTransformer/DocumentToUuidTransformer.php
alexander-schranz: /cc @wachterjohannes what you think about this change. Should throw `DocumentNotFoundException` or `TransformationFailedException` here in case of the document not exists? I think I would keep the DocumentNotFoundException in that case and not catch it what you think?

wachterjohannes: @alexander-schranz @mamazu i would also throw the `DocumentNotFoundException` the `TransformationFailedException` with the message that the document was not found makes no sense for me

mamazu: Yeah, I just noticed that I changed the logic, the class currently throws the `DocumentNotFoundException`. And adding this would make it throw a `TransformationFailedException`. Then I will revert this change. In my opinion then it doesn't make sense to keep the other exception throwing around as well. Because the `TransformationFailedException` will never be thrown because according to the interface this method always returns an object and never `null`.

mamazu:
* Also updated the tests to properly reflect what the entity manager is doing in case of a missing document.

alexander-schranz (approved):

alexander-schranz:
@mamazu Thank you!

PR #6299, opened 2021-10-15
Allow return ghostLocale always

### LINKED ISSUE #6298: Show GhostLocale only when different to current locale
| Q | A
| --- | ---
| Sulu Version | 2.3.6
| PHP Version | all
| DB Version | all
| Browser Version | all

# Actual Behavior

Currently if a `ghostLocale` is returned by the api a `ghostLocale` label is shown, also when it contains the same value as the selected locale.

# Expected Behavior

It should not shown a `ghostLocale` label when selected content locale is the same as the returned `ghostLocale` value.

It should be adjusted to something like:

```js
if (item.ghostLocale && !item.locale) {

}
```

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? | yes
| BC breaks? | no 
| Deprecations? | no  
| Fixed tickets | fixes #6298 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Allow return ghostLocale always.

#### Why?

Makes queries for the list simpler to return the ghostLocale and locale instead of filter out if ghostLocale should be set to null or not.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/adapters/AbstractTableAdapter.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/adapters/AbstractTableAdapter.js
@@ -45,7 +45,7 @@ export default class AbstractTableAdapter extends AbstractAdapter {
 
             const indicators = [];
             if (index === 0) {
-                if (item.ghostLocale) {
+                if (item.ghostLocale && item.ghostLocale !== item.locale) {
                     indicators.push(
                         <GhostIndicator
                             className={abstractTableAdapterStyles.ghostIndicator}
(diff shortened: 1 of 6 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/adapters/AbstractTableAdapter.js
```diff
@@ -45,7 +45,7 @@ export default class AbstractTableAdapter extends AbstractAdapter {
 
             const indicators = [];
             if (index === 0) {
-                if (item.ghostLocale) {
+                if (item.ghostLocale && item.ghostLocale !== item.locale) {
```
alexander-schranz, line 48: to support this correclty it should be `item.ghostLocale && (item.locale === null || typeof item.locale 'undefined'` will currently adjust the query in the content bundle

PR #6304, opened 2021-10-15
Fix accidently remove of category entity when one category translation is removed

### LINKED ISSUE #6300: Delete category translation removes category
| Q | A
| --- | ---
| Sulu Version | 2.3.4
| PHP Version | PHP 8

# Actual Behavior

If you delete a category translation in a different locale, the category is deleted completely.

# Expected Behavior

If you delete a category translation in a different locale, the category with the default locale and its original category translation will stay.

# Steps to Reproduce

1) In https://sulu.rocks/admin/#/categories/ add a new category "football" in locale `en` 
2) Switch to locale `de` you will see category  "[en] football"
3) Add a category translation "fussball" for this category
4) Delete the german category translation "fussball"
5) Now category is deleted in locale `en` as well

Expected behaviour would be to see the "[en] football" again

# Possible Solutions

Remove `ON DELETE CASCADE` on the foreign key in category_translation

```
  CONSTRAINT `FK_5DCED5E3B8075882` FOREIGN KEY (`idCategories`) REFERENCES `ca_categories` (`id`) ON DELETE CASCADE,
  ```
  
 and use orphanRemoval?

### PR DESCRIPTION
**Pull request by simsationX**
…ntity is removed

| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #6300 
| License | MIT

#### What's in this PR?

Fixing cascading of category translation and category meta entity related to category entity

#### Why?

Allow deleting single category translation programmatically without deleting main category

#### Example Usage

```php
       $translationEntity = $category->findTranslationByLocale($locale);

        if ($translationEntity) {
            $this->entityManager->remove($translationEntity);
        }
```

### DIFF
--- a/src/Sulu/Bundle/CategoryBundle/Resources/config/doctrine/Category.orm.xml
+++ b/src/Sulu/Bundle/CategoryBundle/Resources/config/doctrine/Category.orm.xml
@@ -29,13 +29,13 @@
         </many-to-one>
         <one-to-many field="meta" target-entity="Sulu\Bundle\CategoryBundle\Entity\CategoryMetaInterface" mapped-by="category">
             <cascade>
-                <cascade-all/>
+                <cascade-persist/>
             </cascade>
         </one-to-many>
         <one-to-many field="translations" target-entity="Sulu\Bundle\CategoryBundle\Entity\CategoryTranslationInterface"
                      mapped-by="category">
             <cascade>
-                <cascade-all/>
+                <cascade-persist/>
             </cascade>
         </one-to-many>
         <one-to-many field="children" target-entity="Sulu\Bundle\CategoryBundle\Entity\CategoryInterface" mapped-by="parent"/>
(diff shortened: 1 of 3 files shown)


alexander-schranz:
Thx for the pull request. As a bugfix we should target here the `2.2` Branch can you rebase it with:

```bash
git rebase --interactive --onto origin/2.2 origin/2.x
git push yourremote bugfix/cascading-category --force
```

alexander-schranz:
@ymc-sise can you please repush the CI seems not to have not started after changing the target branch. Thank you!

```bash
git commit --amend --date="now"
git push yourremote bugfix/cascading-category --force
```

alexander-schranz (approved):

alexander-schranz:
@ymc-sise Thx I merged it also into the `2.3` branch so you can already use it with `~2.3.6@dev`, else it will be part of the next release 2.3.7 and 2.2.17.

simsationX:
@alexander-schranz Wow very thx, that was fast!

simsationX:
@alexander-schranz In my opinion there is still something wrong here.

When we have multiple translations for one category, the main category won't be deleted, when a translation is removed. That is fixed.

But if we have only one translation left and we delete it, then the main category stays without content. But it should be removed as well, isn't it?

Happens only programmatically:

```
       $translationEntity = $category->findTranslationByLocale($locale);

        if ($translationEntity) {
            $this->entityManager->remove($translationEntity);
        }
```

Via interface the main category is deleted always and correct because deleting of single translation is not provided.

alexander-schranz:
> But if we have only one translation left and we delete it, then the main category stays without content. But it should be removed as well, isn't it?

No this is not automatically handled by doctrine in our entities. If you want to remove the main entity you need todo that manually.

Issue #6302, opened 2021-10-15
Support nginx cache configuration in SuluHttpCacheBundle


ampaze:
| Q | A
| --- | ---
| Sulu Version | 2.2.16
| PHP Version | 8.0.8

# Actual Behavior

`Unrecognized option "nginx" under "sulu_http_cache.proxy_client". Available options are "symfony", "varnish".`

# Expected Behavior

I should be able to use nginx as this is supported by FOSHttpCache

# Steps to Reproduce

Edit sulu_http_cache.yaml

```yaml
sulu_http_cache:
    cache:
        max_age: 10
        shared_max_age: 10
    proxy_client:
        nginx:
            enabled: true
```

alexander-schranz:
Thanks for creating this issue. I personally would more move away from forwarding this configuration from `sulu_http_cache` to `fos_http_cache` and instead directly configure the `fos_http_cache` bundle to avoid this kind of duplication.
Never tried it but maybe you could try as workaround just disable in the `sulu_http_cache` symfony and varnish and enable the `nginx` in the `fos_http_cache` configuration.

ampaze:
I will try it. 

Seems like the following code in `SuluHttpCacheExtension` would never be called then.

```php
if ($proxyClientAvailable) {
    $loader->load('cache-manager.xml');
    $loader->load('cache-lifetime-enhancer.xml');

    if (true === $config['tags']['enabled']) {
        $loader->load('tags.xml');
    }
}
```

alexander-schranz:
@ampaze I'm also happy to review a PR if changing the SuluHttpCache configuration make it easier for you to support nginx directly. Just in future I want that all things are configured over the fos http cache bundle, but in the current version its fine for me adding nginx if this make things easier.

ampaze:
I changed the files needed, but which branch should I use for the PR?

ampaze:
@alexander-schranz Which branch should I base the PR on?

alexander-schranz:
@ampaze As a feature you can target the `2.x` Branch for it.

Issue #6307, opened 2021-10-18
[Caching] Use maxage (and/or s-maxage) instead of X-Reverse-Proxy-TTL by default


ampaze:
# Problem description

Currently Caching in Sulu is opinionated, as it uses **X-Reverse-Proxy-TTL** by default which works fine with the currently supported cache backends (e.g. Symfony and Varnish) but can be confusing, limits the cache backends that could be used and makes caching more complex overall.

# Proposed solution

Make the use of X-Reverse-Proxy-TTL optional, as it is only needed if you really need / want to control three separate TTLs.
In a normal setup max-age is more than enough and if you want to have separate caching times for the browser and your own cache backend then max-age and s-max-age are sufficient.

This also means that max-age and s-maxage would become enabled and useable even without a caching backend and on a per template basis.

## Additional context:

X-Reverse-Proxy-TTL is not a default or industry standard by any means:

https://foshttpcachebundle.readthedocs.io/en/latest/reference/configuration/headers.html#reverse-proxy-ttl

> This keeps your s-maxage free for use with reverse proxies not under your control.

And this is really a very specific use case not worth thinking about for most users.

alexander-schranz:
@ampaze The `X-Reverse-Proxy-TTL` is a widely spread technic to differ cache lifetimes between the own cache server and third proxy cache lifetimes.
We will not make the usage of `s-maxage` default because we would lose then the advantage to define cache lifetimes for third party proxies, and you don't want that a third party proxy caches your page more then some minutes.
So the `X-Reverse-Proxy-TTL` will stay as default behaviour for specially symfony and the varnish cache, which are the widely used caches used by Sulu.
Still if we will target to add support for `nginx` (#6302) we will make it easier to set the cachelifetime over `s-maxage` or make it depending on the used adapter if `x-reverse-poxy-ttl` or `s-maxage` is used, but from our point of view it does not make sense as to make `s-maxage` when using the symfony or varnish adapter.
Keep in mind that nginx is a very basic http cache, it does not support tag based invalidation. Also Sulu Features like Segments, Audience Targeting or when using User Context Based caching are in caches like nginx or other third party caches not supported. So only for some basic sites the nginx cache make sense. Most stays there just with symfony cache for this sites.
I hope I could give you an overview why it is this way, as said the s-maxage issue is something which need to be target when tackling the nginx support (#6302). Still thank you for giving your view on this topic.

ampaze:
[...]

PR #6308, opened 2021-10-18
Add support for Nginx Cache

ampaze:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Related issues/PRs | #6302
| License | MIT

#### What's in this PR?

Add support for Nginx Cache


ampaze:
This code is working fine for me, not sure what PHP Lint / PHPStan wants?

alexander-schranz:
@ampaze Thank you for the Pull Request. I think it is missing an import (use statement) of the Exception class, which is used when tags are enabled. But did not yet have time to test it out.

ampaze:
The PHP Lint error is in files that where not modified with this pull request.

Is there something else that prevents this PR from being merged?

alexander-schranz:
@ampaze This part looks fine. We also need to add a nginx example configuration like we are doing in the docs for [varnish](https://docs.sulu.io/en/2.3/cookbook/caching-with-varnish.html ) currently, which will be based on the current [nginx](https://docs.sulu.io/en/2.3/cookbook/web-server/nginx.html) vhost file if possible, else its fine for me as a reverse proxy configuration. I currently found only https://foshttpcache.readthedocs.io/en/latest/nginx-configuration.html which configuration seems not longer work, atleast for the free nginx version.

Can you create a Pull Request with that also? Think I would like the current varnish configuration add it as a cookbook documentation `Caching with nginx`

ampaze:
I appreciate your desire to add better nginx support, but having an agreed upon example configuration ready before merging this will most likely take a long time. Smart caching with Nginx is unfortunately a bit more complex and not really copy and paste friendly. I am wiling to make my config available but I am not sure it is best suited for everyone out there.

> Can you create a Pull Request with that also? Think I would like the current varnish configuration add it as a cookbook documentation `Caching with nginx`

Yes I hope to work on that when I had time to finish the Tagged Cached Invalidation for Nginx.

Be aware: a complete and clean solution for this is currently not possible, as you don't want to change the `X-Reverse-Proxy-TTL` header (or make  it configurable). See #6307

alexander-schranz:
> Be aware: a complete and clean solution for this is currently not possible, as you don't want to change the X-Reverse-Proxy-TTL header (or make it configurable). See #6307

This is I think the easiest thing as you can set the `s-maxage` by adding a new subsriber here : https://github.com/sulu/skeleton/blob/07211a90df12f1a75c4bc40c782c25490857dc19/src/Kernel.php#L42-L46

```php
if ($response->headers->has(SuluHttpCache::HEADER_REVERSE_PROXY_TT)) {
    $response->setSharedMaxAge($response->headers->get(SuluHttpCache::HEADER_REVERSE_PROXY_TTL));
}
```

You could add that subscriber(`CustomTtlToShareMaxAgeListener`) to the SuluHttpCache so it just need to be added to the Kernel HttpCache when nginx is used. That is more a documentation thing then.


> Yes I hope to work on that when I had time to finish the Tagged Cached Invalidation for Nginx.

As nginx does not support tagged caching out of the box and nginx could be run on another server, the config in our case should just simple to allow caching and purging. So at current state it would be good to have a simple nginx config which take the `s-maxage` into account cache it by the defined cache lifetime and that the url is purged when a page is republished. I know that you want to implement cache tag invalidation on php side but this I think will only work in cases the php application runs on same system as your nginx server, so this I think is not required for a basic nginx support.

ampaze:
Fair enough. Keeping it simple is always good :)

> the config in our case should just simple to allow caching and purging. 

Unfortunately the free version does not support purging and you would need to compile `nginx` yourself to add support, like described in the link you posted (https://foshttpcache.readthedocs.io/en/latest/nginx-configuration.html). (This can be worked around with something like `proxy_cache_bypass` but I have not done so yet.)

This is one reason I think directly accessing the cache files is much more hassle free, but as you mentioned, this would be more difficult if `nginx` is used only for caching on a different server. But imho using nginx for caching dynamic pages is mostly done because nginx is already used for everything else.

alexander-schranz:
As it only seems to support with none default Nginx I currently will close this one as I think if somebody has that access to the server they could easier configure a varnish service which support is more feature rich.

PR #6310, opened 2021-10-21
Add PropertyMetadataMapper for number field type

luca-rath:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | sulu/sulu-docs#684

#### What's in this PR?

Add PropertyMetadataMapper for number field type

#### Why?

This enables validation of `min`, `max` and `multiple_of` params of `number` field type.

#### Example Usage

```xml
<property name="number" type="number">
    <meta>
        <title lang="en">Number</title>
    </meta>

    <params>
        <param name="min" value="2"/>
        <param name="max" value="5"/>
        <param name="step" value="1"/>
        <param name="multiple_of" value="1"/>
    </params>
</property>
```

#### To Do


src/Sulu/Component/Content/Tests/Unit/Types/NumberTest.php
```diff
@@ -112,4 +113,304 @@ public function testWriteNoValue(): void
...
+        ];
+    }
+
+    /**
+     * @return array<string, mixed>
+     */
+    private function getEmptyStringSchema(): array
+    {
+        return [
+            'type' => 'string',
```
alexander-schranz: when is this happening?

src/Sulu/Component/Content/Types/Number.php
```diff
@@ -57,4 +65,68 @@ public function read(NodeInterface $node, PropertyInterface $property, $webspace
...
+
+    public function mapPropertyMetadata(ContentPropertyMetadata $propertyMetadata): PropertyMetadata
+    {
+        /** @var string $propertyName */
+        $propertyName = $propertyMetadata->getName();
+        $mandatory = $propertyMetadata->isRequired();
+
+        $min = $this->getFloatParam($propertyMetadata, 'min');
+        $max = $this->getFloatParam($propertyMetadata, 'max');
+        $multipleOf = $this->getFloatParam($propertyMetadata, 'multiple_of');
```
alexander-schranz, line 76: I would not introduce a new parameter here I would reuse the `step` parameter. I would not support to have a different `step` and a different `multiple_of` parameter in the number type.

alexander-schranz (changes requested):

src/Sulu/Component/Content/Types/Number.php
luca-rath: I think it's a valid example to have `step=1` and `multiple_of=0.01` to allow numbers with two decimals (prices e.g.) but the number field should still use `1` as step

src/Sulu/Component/Content/Tests/Unit/Types/NumberTest.php
luca-rath: You're right, this is never the case ... Will remove it

src/Sulu/Component/Content/Types/Number.php
alexander-schranz: I know what you mean and wanted to support with it but I personally would not do it. It can be misconfigured step lower then multiple_of, is step fallbacking or is multple_of of fallbacking, will multiple_of overwrite step when lower, many questions I think is added for a usecase which is mostly not used. And if there is just one value its just clear.
But if the others want also to add it I'm fine with it. /cc @sulu/core-team

alexander-schranz: I would at current state go with `step` and wait for feedback. If this is then requested we still could add `multiple_of` but at current state I would just use the value from `step`.

luca-rath: We could just throw an exception if `step` is not a multiple of `multiple_of` - then there's no fallback "magic" ... But `step` shouldn't be used for validation IMO, because if you specify `step` on a html input element, you can still type in numbers between steps, there won't be an error. And in the docs we could clearly specify that `step` is just used for the ui while `multiple_of` is used for validation

alexander-schranz: I'm aware of how html is handling it but as said I would at current state not add an additional parameter for it.

niklasnatter: dont have a strong opinion here - both options are okay for me

alexander-schranz (approved):

src/Sulu/Component/Content/Types/Number.php
alexander-schranz: /cc @chirimoya

alexander-schranz: @luca-rath I did discuss it with @chirimoya and we keep step and multiple_of. So the PR is ready. Can you check the PHP Lint failings then we could also merge this.

alexander-schranz (approved):

PR #6311, opened 2021-10-21
Add trash handling for accounts

### LINKED ISSUE #5150: Trashcan functionality

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? | yes 
| BC breaks? | yes 
| Deprecations? | no  
| Fixed tickets | fixes #5150  
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Add trash handling for accounts.

#### Why?

See #5150 

#### Example Usage

Remove account it can be restored over the trash menu item.

#### To Do

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -120,7 +120,7 @@
         "microsoft/azure-storage-blob": "^1.2",
         "php-ffmpeg/php-ffmpeg": "0.13 - 0.17",
         "phpcr/phpcr-shell": "^1.0",
-        "phpspec/prophecy": "^1.8",
+        "phpspec/prophecy": "^1.14.0",
         "phpstan/phpstan": "^0.12",
         "phpstan/phpstan-doctrine": "^0.12.33",
         "phpstan/phpstan-phpunit": "^0.12.18",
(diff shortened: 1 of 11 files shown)


wachterjohannes (approved):

wachterjohannes:
@alexander-schranz we should rebase the branches (incl. this one) to see if the tests will go green before we merge this

PR #6312, opened 2021-10-21
Fix xliff import publish date in WebspaceImport

popoplanter:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no yes
| Deprecations? | no 
| Fixed tickets | I dont know
| Related issues/PRs | I dont know

#### What's in this PR?

This Pull Request fixes pages not keeping their correct workflowStage after and XLIFF import.

#### Why?
When exporting a webspace with unpublished pages and importing again the workflowStage is ignored and just set to published.

### DIFF
--- a/src/Sulu/Component/Content/Import/WebspaceImport.php
+++ b/src/Sulu/Component/Content/Import/WebspaceImport.php
@@ -235,7 +235,7 @@ class WebspaceImport extends Import implements WebspaceImportInterface
             // save document
             $this->documentManager->persist($document, $locale);
 
-            if (WorkflowStage::PUBLISHED === ((int) $data['workflowStage']['value'])) {
+            if (WorkflowStage::PUBLISHED === ((int) $this->getParser($format)->getPropertyData('workflowStage', $data))) {
                 $this->documentManager->publish($document, $locale);
             }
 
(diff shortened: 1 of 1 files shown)


src/Sulu/Component/Content/Import/WebspaceImport.php
```diff
@@ -233,7 +233,12 @@ protected function importDocument(array $parsedData, $format, $webspaceKey, $loc
 
             // save document
             $this->documentManager->persist($document, $locale);
-            $this->documentManager->publish($document, $locale);
+            
+            // Only publish when workflowStage is set to published.
+            if (2 == $data['workflowStage']['value']){
```
niklasnatter: i think we should use the parser to get the value of the property like here:
https://github.com/sulu/sulu/blob/cfaa36b94443df000c00014b86f23bb927925a5e/src/Sulu/Component/Content/Import/WebspaceImport.php#L199

niklasnatter: also, would be nice to use `WorkflowStage::PUBLISHED` instead of `2`

niklasnatter (commented):
Would be nice if you could rebase this onto `2.2` as the issue is already present in the `2.2.*` releases

PR #6321, opened 2021-10-27
Implement trash handling for the snippet entity

niklasnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| Related issues/PRs | #5150
| License | MIT

#### What's in this PR?

This PR integrates the `SuluTrashBundle` with the built-in snippet entity. With the changes in this PR, snippets will be moved into the trash when being deleted and can be restored from the trash at a later point in time.


src/Sulu/Bundle/PageBundle/EventListener/PageTrashSubscriber.php
```diff
@@ -70,7 +70,7 @@ public function storePageToTrash(RemoveEvent $event): void
             return;
         }
 
-        $this->trashManager->store(BasePageDocument::RESOURCE_KEY, $event->getDocument());
+        $this->trashManager->store(BasePageDocument::RESOURCE_KEY, $document);
```
alexander-schranz, line 73: Can you describe this change?

niklasnatter: `$event->getDocument()` is already called a few lines above, we dont need to call it a second time 🙂

alexander-schranz (approved):

PR #6325, opened 2021-10-28
Add trash handling for custom urls

niklasnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| Related issues/PRs | #5150
| License | MIT

#### What's in this PR?

This PR integrates the `SuluTrashBundle` with the built-in `CustomUrl` entity. With the changes in this PR, custom urls will be moved into the trash when being deleted and can be restored from the trash at a later point in time.


src/Sulu/Bundle/CustomUrlBundle/EventListener/CustomUrlSerializeEventSubscriber.php
```diff
@@ -85,6 +95,12 @@ public function onPostSerialize(ObjectEvent $event)
             $customUrlProperty
         );
 
+        $webspaceKey = $this->documentInspector->getWebspace($customUrl);
+        $visitor->visitProperty(
```
niklasnatter, line 99: i had to add the webspace in the serialized data to navigate to the custom url list of the webspace after restoring a custom url

alexander-schranz: Sounds good for me as it is just parse the webspace out of the phpcr path and doesn't trigger any lazy loading its fine.

alexander-schranz (approved):

PR #6326, opened 2021-10-28
Add restoreType and restoreOptions to TrashItem 

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? |  yes
| BC breaks? | yes 
| Deprecations? | no  
| Fixed tickets | part of #5150 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Add subResourceKey and options to TrashItem .

#### Why?

Make it possible to remove a single page translation and restore a single page translation.

### DIFF
--- a/src/Sulu/Bundle/TrashBundle/Domain/Model/TrashItem.php
+++ b/src/Sulu/Bundle/TrashBundle/Domain/Model/TrashItem.php
@@ -38,7 +38,25 @@ class TrashItem implements TrashItemInterface
     /**
      * @var mixed[]
      */
-    private $restoreData;
+    private $restoreData = [];
+
+    /**
+     * The restoreType can be used to indicate a sub entity.
+     *     e.g.: Store and Restore a single translation of a page.
+     *          -> "translation".
+     *
+     * @var string|null
+     */
+    private $restoreType;
+
+    /**
+     * The restoreOptions are used to change behaviour of store and restore handler.
+     *     e.g.: Store and Restore a single translation of a page.
+     *          -> ["locale" => "en"].
+     *
+     * @var mixed[]
+     */
+    private $restoreOptions = [];
 
     /**
      * @var string|null
@@ -78,6 +96,7 @@ class TrashItem implements TrashItemInterface
     public function __construct()
     {
         $this->translations = new ArrayCollection();
+        $this->storeTimestamp = new \DateTimeImmutable();
     }
 
     public function getId(): ?int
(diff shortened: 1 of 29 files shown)


src/Sulu/Bundle/TrashBundle/UserInterface/Controller/Admin/TrashItemController.php
```diff
@@ -190,8 +190,10 @@ public function cgetAction(Request $request): Response
 
         $trashItems = \array_map(
             function(array $trashItem) use ($hiddenFieldDescriptors) {
-                if (isset($trashItem['resourceType'])) {
-                    $trashItem['resourceType'] = $this->getResourceTranslation($trashItem['resourceType']);
+                if (isset($trashItem['subResourceKey'])) {
+                    $trashItem['resourceKey'] = $this->getResourceTranslation($trashItem['subResourceKey']);
```
alexander-schranz: The concept is to show in this case `Page (Translation)` or `Page (Translation) EN` and the subResourceKey is `page_translations`.

src/Sulu/Bundle/TrashBundle/Domain/Model/TrashItem.php
```diff
@@ -36,9 +36,22 @@ class TrashItem implements TrashItemInterface
     private $resourceId;
 
     /**
+     * @var string|null
+     */
+    private $subResourceKey;
```
niklasnatter: it does not feel great to me to call this `(sub)resourceKey` because we dont use that value as resource key anywhere else 😕 

i think i would prefer to call it `dataType`/`itemType`/`restoreType` and append that type to the resource key in the table like this: `ResourceKeyTranslation DataTypeTranslation`

niklasnatter: but i dont want to stir a big discussion. i think the two options are the following:
- `subResourceKey`: use value of `subResourceKey` in the frontend if set
- `dataType`: append `dataType` to resource key in frontend if set

what do you prefer? @sulu/core-team

alexander-schranz: > dataType/itemType/restoreType

I'm not sure about this names as they are for me a little too less specific that they tell somebody about what to put into it when creating a TrashItem instance.

niklasnatter: yeah thats right, they are not specific. but do you think its easier to understand what to pass as `subResourceKey`? especially because the value you pass is something that you cannot use as resource key anywhere else?

alexander-schranz: > especially because the value you pass is something that you cannot use as resource key anywhere else?

for me it make sense that I can not pass a subResourceKey as a resourceKey as this are 2 different things for me. I like to have resourceKey in it because of its relation to the resourceKey. And name like `sub` or `partial` provide information about the usage.

From your names I would if to choice one I would use `restoreType` but then go with `restoreOptions` as the naming for them in the `TrashItem` object (not in the handler store method there would keep options).

alexander-schranz: as discussed we go with restoreType and restoreOptions

niklasnatter (approved):

PR #6327, opened 2021-10-28
Skip if findAvailableLocale does not return any webspace

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? | no 
| BC breaks? | no 
| Deprecations? | no yes 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Add better error message when webspace is not found in reindex.

#### Why?

If you are doing a reindex it will reindex all documents from the database, if there still exists old webspace data it show error calling `getLocalization` on null. Instead we show here a clear message that the webspace could not found.


wachterjohannes (approved):

wachterjohannes:
@alexander-schranz linting failed

alexander-schranz:
@wachterjohannes fixed!

PR #6328, opened 2021-10-29
Add External link in admin navigation

### LINKED ISSUE #5207: Support NavigationItems with absolute url configuration
## Problem description

I want to add menu items that point to external services in the cms admin menu.

sebheitzmann:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | fixes #5207
| Related issues/PRs | #5207
| License | MIT

#### What's in this PR?

Permit the admin navitation to link to external url

#### Why?

Link to external tools ( google analytics, or other web companion )

#### To Do

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Admin/Navigation/NavigationExternalItem.php
+++ b/src/Sulu/Bundle/AdminBundle/Admin/Navigation/NavigationExternalItem.php
@@ -276,7 +276,7 @@ class NavigationExternalItem implements \Iterator, NavigationItemInterface
     public function copyChildless()
     {
         $new = $this->copyWithName();
-        $new->setView($this->getView());
+        $new->setUrl($this->getUrl());
         $new->setChildViews($this->getChildViews());
         $new->setIcon($this->getIcon());
         $new->setId($this->getId());
@@ -449,14 +449,5 @@ class NavigationExternalItem implements \Iterator, NavigationItemInterface
 
         return $array;
     }
-
-    public function hasLink(): bool
-    {
-        return $this->getUrl() !== null;
-    }
-
-    public function getView()
-    {
-        return null;
-    }
+    
 }
--- a/src/Sulu/Bundle/AdminBundle/Admin/Navigation/NavigationExternalItem.php
+++ b/src/Sulu/Bundle/AdminBundle/Admin/Navigation/NavigationExternalItem.php
@@ -11,7 +11,7 @@
 
 namespace Sulu\Bundle\AdminBundle\Admin\Navigation;
 
-class NavigationItem implements \Iterator, NavigationItemInterface
+class NavigationExternalItem implements \Iterator, NavigationItemInterface
 {
     /**
      * The id of the NavigationItem.
@@ -42,7 +42,7 @@ class NavigationItem implements \Iterator, NavigationItemInterface
     /**
      * @var string
      */
-    protected $view;
+    protected $url;
 
     /**
      * @var string[]
@@ -162,14 +162,14 @@ class NavigationItem implements \Iterator, NavigationItemInterface
         return $this->icon;
     }
 
-    public function setView(string $view = null): void
+    public function setUrl(string $url = null): void
     {
-        $this->view = $view;
+        $this->url = $url;
     }
 
-    public function getView(): ?string
+    public function getUrl(): ?string
     {
-        return $this->view;
+        return $this->url;
     }
 
     public function setChildViews(array $childViews): void
@@ -420,7 +420,7 @@ class NavigationItem implements \Iterator, NavigationItemInterface
             'title' => $this->getName(),
             'label' => $this->getLabel(),
             'icon' => $this->getIcon(),
-            'view' => $this->getView(),
+            'url' => $this->getUrl(),
             'disabled' => $this->getDisabled(),
             'visible' => $this->getVisible(),
             'id' => (null != $this->getId()) ? $this->getId() : \str_replace('.', '', \uniqid('', true)), //FIXME don't use uniqid()
@@ -449,9 +449,14 @@ class NavigationItem implements \Iterator, NavigationItemInterface
 
         return $array;
     }
-    
+
     public function hasLink(): bool
     {
-        return $this->getView() !== null;
+        return $this->getUrl() !== null;
+    }
+
+    public function getView()
+    {
+        return null;
     }
 }
--- a/src/Sulu/Bundle/AdminBundle/Admin/Navigation/NavigationItemInterface.php
+++ b/src/Sulu/Bundle/AdminBundle/Admin/Navigation/NavigationItemInterface.php
@@ -29,6 +29,4 @@ interface NavigationItemInterface
     public function setVisible($visible);
     public function getVisible();
     public function toArray();
-    public function hasLink();
-    public function getView();
 }
--- a/src/Sulu/Bundle/AdminBundle/Admin/Navigation/NavigationItemInterface.php
+++ b/src/Sulu/Bundle/AdminBundle/Admin/Navigation/NavigationItemInterface.php
@@ -1,0 +1,34 @@
+<?php
+
+/*
+ * This file is part of Sulu.
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\AdminBundle\Admin\Navigation;
+
+interface NavigationItemInterface
+{
+    public function setId($id);
+    public function getId();
+    public function setName($name);
+    public function getName();
+    public function setLabel(string $label = null): void;
+    public function getLabel(): ?string;
+    public function setIcon($icon);
+    public function getIcon();
+    public function setPosition($position);
+    public function getPosition();
+    public function hasChildren();
+    public function setDisabled($disabled);
+    public function getDisabled();
+    public function setVisible($visible);
+    public function getVisible();
+    public function toArray();
+    public function hasLink();
+    public function getView();
+}
--- a/src/Sulu/Bundle/AdminBundle/Admin/Navigation/NavigationRegistry.php
+++ b/src/Sulu/Bundle/AdminBundle/Admin/Navigation/NavigationRegistry.php
@@ -75,9 +75,7 @@ class NavigationRegistry
             $admin->configureNavigationItems($navigationItemCollection);
         }
 
-        $navigationItems = \array_filter($navigationItemCollection->all(), function($navigationItem) {
-            return $navigationItem->getChildren() || $navigationItem->hasLink();
-        });
+        $navigationItems = $navigationItemCollection->all();
 
         foreach ($navigationItems as $navigationItem) {
             $this->processNavigationItem($navigationItem);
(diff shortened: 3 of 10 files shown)


sebheitzmann:
That's the php part for the moment. I've implemented an NavigationItemInterface to handle types in the registry and adminController. Feel free to comment and propose an other approach.

sebheitzmann:
Moving on the front part.

src/Sulu/Bundle/AdminBundle/Admin/Navigation/NavigationItemInterface.php
```diff
@@ -0,0 +1,34 @@
...
+    public function setPosition($position);
+    public function getPosition();
+    public function hasChildren();
+    public function setDisabled($disabled);
+    public function getDisabled();
+    public function setVisible($visible);
+    public function getVisible();
+    public function toArray();
+    public function hasLink();
+    public function getView();
```
alexander-schranz: @sulu/core-team how we define a interface here. In the views we only have in the interface which really relevant for all types and use `options` array to provide type specific options. I'm thinking about here that the normal `NavigationItem` has only a setView and getView and will write this into an options array like we do it for views e.g.:

```php
public function setView(string $view):  void
{
    return $this->options['view'] = $view;
}
```

and the ExternalNavigationItem then with:

```php
public function setUrl(string $url):  void
{
    return $this->options['url'] = $url;
}
```

And so the NavigationItemInterface instead of having `getView` it will have a `getOptions` and `getType` method.

src/Sulu/Bundle/AdminBundle/Admin/Navigation/NavigationRegistry.php
```diff
@@ -76,7 +76,7 @@ private function loadNavigationItems(): void
         }
 
         $navigationItems = \array_filter($navigationItemCollection->all(), function($navigationItem) {
-            return $navigationItem->getChildren() || $navigationItem->getView();
+            return $navigationItem->getChildren() || $navigationItem->hasLink();
```
alexander-schranz: Interesting didn't know we filter this out here this way. Need to check if the whole `array_filter` maybe could be removed so we don't need to have something like `hasLink` here.

src/Sulu/Bundle/AdminBundle/Admin/Navigation/NavigationExternalItem.php
```diff
@@ -0,0 +1,462 @@
...
+
+    /**
+     * Returns a copy of this navigation item without its children.
+     *
+     * @return NavigationItem
+     */
+    public function copyChildless()
+    {
+        $new = $this->copyWithName();
+        $new->setView($this->getView());
```
alexander-schranz: Think this should be `$new->setUrl($this->getUrl());`

src/Sulu/Bundle/AdminBundle/Admin/Navigation/NavigationExternalItem.php
```diff
@@ -0,0 +1,462 @@
...
+
+        return $array;
+    }
+
+    public function hasLink(): bool
+    {
+        return $this->getUrl() !== null;
+    }
+
+    public function getView()
```
alexander-schranz: The external should know nothing about view, need to discuss the Interface with the other core team members.

src/Sulu/Bundle/AdminBundle/Admin/Navigation/NavigationExternalItem.php
```diff
@@ -0,0 +1,462 @@
...
+
+        foreach ($children as $key => $child) {
+            /* @var NavigationItem $child */
+            $array['items'][$key] = $child->toArray();
+        }
+
+        return $array;
+    }
+
+    public function hasLink(): bool
```
alexander-schranz: Not sure if required, need to check different use cases and check if the array_filter is really still needed or can be handled differently.

alexander-schranz (commented):
@sebheitzmann Thanks for working on this, I did a quick review and we maybe need to discuss the Interface, but at all it looks good to that you could start with the frontend part.

src/Sulu/Bundle/AdminBundle/Admin/Navigation/NavigationRegistry.php
sebheitzmann: Don't really understand why we sould filter the items. Is it possible to have items without view ? If they are hidden i think it's useless. I will correct this.

src/Sulu/Bundle/AdminBundle/Admin/Navigation/NavigationItemInterface.php
sebheitzmann: I have done this because they are used in the Registry. But maybe it should be avoided

niklasnatter: i am not sure if i would separate internal and external navigation items into separate classes. they share a lot of things (eg. `label`, `icon`, `position`), right?

maybe it would be better to add a `type` on the `NavigationItem` class? we could add a `ViewNavigationItemBuilder` and a `ExternalLinkNavigationItemBuilder` to create and validate different types of navigation items. this would be consistent to the existing [`View` concept](https://github.com/sulu/sulu/tree/2.x/src/Sulu/Bundle/AdminBundle/Admin/View).

src/Sulu/Bundle/AdminBundle/Admin/Navigation/NavigationRegistry.php
niklasnatter: i think the items are filtered to filter out empty parents. for example, if a user does not have permissions for accounts and for contacts, the `Contact` navigation item would be empty

src/Sulu/Bundle/AdminBundle/Admin/Navigation/NavigationItemInterface.php
alexander-schranz: Sounds good for me to have the same concept as for Views here. We would need a `NavigationBuilderFactory` then like the `ViewBuilderFactory`.

sebheitzmann:
image

src/Sulu/Bundle/AdminBundle/Admin/Navigation/NavigationExternalItem.php
sebheitzmann: The getView is used in the NavigationRegistry ( line 108 )

```php
        // add child views
        $mainView = $navigationItem->getView();
        if ($mainView) {
            $mainPath = $this->viewRegistry->findViewByName($mainView)->getPath();
            if ('/' !== $mainPath) {
                foreach ($this->viewRegistry->getViews() as $view) {
                    if (0 === \strpos($view->getPath(), $mainPath)) {
                        $navigationItem->addChildView($view->getName());
                    }
                }
            }
        }
```

Maybe this should be done in the Item and not in the registry.

src/Sulu/Bundle/AdminBundle/Admin/Navigation/NavigationExternalItem.php
sebheitzmann: Removed in the last commit

src/Sulu/Bundle/AdminBundle/Admin/Navigation/NavigationItemInterface.php
alexander-schranz: For backward compatibility the `NavigationItemCollection` would need to be changed to accept `NavigationItemBuilder` and `NavigationItem` then.

src/Sulu/Bundle/AdminBundle/Admin/Navigation/NavigationExternalItem.php
sebheitzmann: I thing that an external link should not have children. So maybe this function should be removed.  ( but corrected )

src/Sulu/Bundle/AdminBundle/Admin/Navigation/NavigationRegistry.php
sebheitzmann: ok, I will revert my change to keep the getChildren filter. But the getView is useless no ?

src/Sulu/Bundle/AdminBundle/Admin/Navigation/NavigationExternalItem.php
alexander-schranz: Maybe we introduce a speical `GroupNavigationItem` because if we name the other one `ViewNavigationItem` it also does not make sense that it can have children.

src/Sulu/Bundle/AdminBundle/Admin/Navigation/NavigationRegistry.php
sebheitzmann: No, the getView is needed if there is no children. Instead all the leaf link will be removed

sebheitzmann: maybe a shouldShow() function on the interface ?

src/Sulu/Bundle/AdminBundle/Admin/Navigation/NavigationExternalItem.php
sebheitzmann: I thinks that it's for the "active" feature.

src/Sulu/Bundle/AdminBundle/Admin/Navigation/NavigationRegistry.php
alexander-schranz: If we add a special GroupNavigationItem we would not need a `shouldShow` instead we could check here: `return !$navigationItem instanceOf GroupNavigatinItemBuilder || count(!$navigationItem->getChildren());`. Because a ViewNavigationItem without view or a ExternalNavigationItem without Url would never exists and so don't need to be filtered out.

src/Sulu/Bundle/AdminBundle/Admin/Navigation/NavigationItemInterface.php
sebheitzmann: I'm sorry but I don't really understand what i should do with the ViewBuilderFactory. I can merge the NavigationItem and the ExternalNavigation if you want.

sebheitzmann: How can I Help ?

src/Sulu/Bundle/AdminBundle/Admin/Navigation/NavigationExternalItem.php
```diff
@@ -0,0 +1,453 @@
...
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\AdminBundle\Admin\Navigation;
+
+class NavigationExternalItem implements \Iterator, NavigationItemInterface
```
mamazu, line 14: This class is missing proper type information. We should use the php typehinting:

```php
/**
* @var string
*/
protected $id;
```

will become
```php
protected string $id;
```

That way we have help from php checking the types. Same with arguments and return types.

PR #6337, opened 2021-11-02
Add trash for page translations

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? | yes 
| BC breaks? | no 
| Deprecations? | no  
| Fixed tickets | fixes # 
| Related issues/PRs | #5150 
| License | MIT

#### What's in this PR?

Add trash for page translations.

#### Why?

Add store and restore of a single page translation.

### DIFF
--- a/src/Sulu/Bundle/PageBundle/Resources/translations/admin.de.json
+++ b/src/Sulu/Bundle/PageBundle/Resources/translations/admin.de.json
@@ -64,6 +64,7 @@
     "sulu_page.schedules_label": "Zeitfenster zuweisen",
     "sulu_page.schedules_description": "Der Block wird nur angezeigt, wenn mindestens einer der Bedingungen erfüllt wird.",
     "sulu_activity.resource.pages": "Seite",
+    "sulu_activity.resource.pages.translation": "Übersetzung",
     "sulu_activity.description.pages.created": "{userFullName} hat die Seite \"{resourceTitle}\" erstellt",
     "sulu_activity.description.pages.modified": "{userFullName} hat die Seite \"{resourceTitle}\" geändert",
     "sulu_activity.description.pages.removed": "{userFullName} hat die Seite \"{resourceTitle}\" gelöscht",
--- a/src/Sulu/Bundle/TrashBundle/Resources/config/lists/trash_items.xml
+++ b/src/Sulu/Bundle/TrashBundle/Resources/config/lists/trash_items.xml
@@ -31,7 +31,7 @@
 
     <properties>
         <property
-            name="resourceKey"
+            name="resourceType"
             visibility="yes"
             sortable="true"
             searchability="no"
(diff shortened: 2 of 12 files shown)


src/Sulu/Bundle/PageBundle/Resources/translations/admin.de.json
```diff
@@ -64,6 +64,7 @@
     "sulu_page.schedules_label": "Zeitfenster zuweisen",
     "sulu_page.schedules_description": "Der Block wird nur angezeigt, wenn mindestens einer der Bedingungen erfüllt wird.",
     "sulu_activity.resource.pages": "Seite",
+    "sulu_activity.resource.translation": "Übersetzung",
```
niklasnatter: think we should use `sulu_activity.resource.pages.translation` or something like `sulu_trash.restore_type.translation`/`sulu_trash.restore_type.pages.translation` here

src/Sulu/Bundle/TrashBundle/Resources/config/lists/trash_items.xml
```diff
@@ -31,7 +31,7 @@
 
     <properties>
         <property
-            name="resourceKey"
+            name="resourceType"
```
niklasnatter, line 34: is there a specific reason for this change? 🙂

alexander-schranz: I accidently changed it in #6326. Thought it was an accident but it is required as the `resourceType` is the item shown in the list e.g.: `Page` and the resourceKey (hardcoded added in the template) is the untranslated `resourceKey`. Currently the RestoreForm is not working because of this change.

src/Sulu/Bundle/PageBundle/Resources/translations/admin.de.json
alexander-schranz: updated to `sulu_activity.resource.pages.translation`

src/Sulu/Bundle/TrashBundle/UserInterface/Controller/Admin/TrashItemController.php
```diff
@@ -190,12 +190,8 @@ public function cgetAction(Request $request): Response
 
         $trashItems = \array_map(
             function(array $trashItem) use ($hiddenFieldDescriptors) {
-                if (isset($trashItem['resourceKey'])) {
-                    $trashItem['resourceKey'] = $this->getResourceTranslation($trashItem['resourceKey']);
-
-                    if (!empty($trashItem['restoreType'])) {
-                        $trashItem['resourceKey'] .= ' (' . $this->getResourceTranslation($trashItem['restoreType']) . ')';
-                    }
+                if (isset($trashItem['resourceType'])) {
```
niklasnatter, line 193: not sure if we are testing this controller. if we do, it would be nice to test the `restoreType` too 🙂

alexander-schranz: updated the testCgetAction to test this behaviour.

niklasnatter (approved):

PR #6352, opened 2021-11-04
Update array-move package to new major version

niklasnatter:
#### What's in this PR?

this PR updates the `array-move` package that is used in the `SuluAdminBundle` to the newly released version 4.

#### Why?

Because we try to keep our dependencies up-to-date if possible 🙂

### DIFF
--- a/package.json
+++ b/package.json
@@ -110,7 +110,7 @@
         ],
         "clearMocks": true,
         "transformIgnorePatterns": [
-            "node_modules/(?!(@ckeditor|ckeditor5|lodash-es|@react-leaflet|react-leaflet)/)"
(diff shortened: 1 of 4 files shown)


alexander-schranz (approved):

alexander-schranz (changes requested):
Currently failing, but is not urgent for the RC so we will keep this PR open until we find time to fix it.

alexander-schranz:
Think `array-move` is currently incompatible because it uses now ESM: https://github.com/sindresorhus/array-move/issues/25

alexander-schranz:
As the package seems not todo many things we could also just copy the `array-move` file from it: https://github.com/sindresorhus/array-move/blob/main/index.js

/cc @sulu/core-team

PR #6364, opened 2021-11-14
Improve performance of access control query enhancer

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? | no 
| BC breaks? | no 
| Deprecations? | no  
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Improve performance of access control query enhancer.

The change removes the subquery and selects all IDs which are not accessible by the user and exclude them from the real result.

#### Why?

Currently the queries with a lot of medias are taken a lot of time:

This will improve the list query `7462.32 ms` -> `2.34 ms + 2.64 ms`. 
The count query will also improve from `4872.34 ms`-> `2.83 ms + 635.06 ms`.

So its an over all improvement from `12361.82 ms` -> `659.44 ms`.

To implement this performance improvement we use an own query to load the restricted (not accessible) ids. This is faster as embedding the query as subquery (as it was before). Also loading the "accessible" ids would be more performance intense because sulu has more "accessible" entities that not accessible. Optimized embedded queries are mostly not compatible for MySQL 5.7 because of restrictions.

 As long as we dont have thousands of not "accessible" ids this approach should be faster.

### DIFF
--- a/src/Sulu/Bundle/SecurityBundle/AccessControl/AccessControlQueryEnhancer.php
+++ b/src/Sulu/Bundle/SecurityBundle/AccessControl/AccessControlQueryEnhancer.php
@@ -57,35 +57,55 @@ class AccessControlQueryEnhancer
         QueryBuilder $queryBuilder,
         ?UserInterface $user,
         int $permission,
+        string $entityClass,
+        string $entityAlias,
         string $entityClassField,
-        string $entityIdField,
-        string $entityAlias
+        string $entityIdField
     ): void {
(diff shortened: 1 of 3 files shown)


alexander-schranz:
@wachterjohannes can you add the stats and differents here as comment on the 2 tested databases (before / after).

alexander-schranz:
ping @wachterjohannes

PR #6367, opened 2021-11-17
Add visible-condition to media buttons

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | none
| Related issues/PRs | none
| License | MIT
| Documentation PR | none

#### What's in this PR?

This PR uses permissions to hide buttons when permission not available for current user.

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/Controller/CollectionController.php
+++ b/src/Sulu/Bundle/MediaBundle/Controller/CollectionController.php
@@ -364,7 +364,7 @@ class CollectionController extends AbstractRestController implements ClassResour
     }
 
     /**
-     * @return Collection
+     * @return mixed[]
      */
     protected function getData(Request $request)
     {
--- a/src/Sulu/Bundle/MediaBundle/Resources/js/containers/MediaCollection/CollectionSection.js
+++ b/src/Sulu/Bundle/MediaBundle/Resources/js/containers/MediaCollection/CollectionSection.js
@@ -113,9 +113,10 @@ class CollectionSection extends React.Component<Props> {
     };
 
     handleSaveResponse = (resourceStore: ResourceStore) => {
+        const openedCollectionOperationOverlayType = this.openedCollectionOperationOverlayType;
         this.closeCollectionOperationOverlay();
 
-        if (this.openedCollectionOperationOverlayType === 'update') {
+        if (openedCollectionOperationOverlayType === 'update') {
             this.props.resourceStore.setMultiple(resourceStore.data);
         } else {
             this.props.onCollectionNavigate(resourceStore.id);
--- a/src/Sulu/Bundle/MediaBundle/Resources/js/containers/MediaCollection/CollectionSection.js
+++ b/src/Sulu/Bundle/MediaBundle/Resources/js/containers/MediaCollection/CollectionSection.js
@@ -58,7 +58,7 @@ class CollectionSection extends React.Component<Props> {
             return resourceStore.clone();
         }
 
-        return new ResourceStore(
+        const newResourceStore = new ResourceStore(
             COLLECTIONS_RESOURCE_KEY,
             null,
             {
@@ -70,6 +70,12 @@ class CollectionSection extends React.Component<Props> {
                 parent: data.parent,
             }
         );
+
+        if (this.collectionId && this.openedCollectionOperationOverlayType === 'create') {
+            newResourceStore.set('parent', this.collectionId);
+        }
+
+        return newResourceStore;
     }
 
     handleCollectionClick = (collectionId: string | number) => {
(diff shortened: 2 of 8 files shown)


alexander-schranz (approved):

src/Sulu/Bundle/MediaBundle/Serializer/Subscriber/MediaPermissionsSubscriber.php
```diff
@@ -0,0 +1,81 @@
...
+        ];
+    }
+
+    public function onPostSerialize(ObjectEvent $event)
+    {
+        $object = $event->getObject();
+        /** @var SerializationVisitorInterface $visitor */
+        $visitor = $event->getVisitor();
+
+        // FIXME This should be removed, once all entities are restructured not using the ApiWrapper, possible BC break
```
alexander-schranz, line 60: @wachterjohannes looks good but we should do this only for a specific group else we will maybe change accidently Website API responses.

wachterjohannes: the subscriber is only registered in admin ad we do it for the "normal" `Sulu\Component\Security\Serializer\Subscriber\SecuredEntitySubscriber`

alexander-schranz: We should not build context based services now, as the context will be removed in the future in favor of the single kernel setup.

wachterjohannes: @alexander-schranz there is currently no `group` in the MediaController therefor we should keep it like the "original" service. everything else would be a BC Break and i would like to rebase that on 2.2

wachterjohannes:
@alexander-schranz have added the tests and fixed php-cs

alexander-schranz (approved):

alexander-schranz:
PHPStan is failing the new subscriber and test  should be implemented with our coding standards.

src/Sulu/Bundle/MediaBundle/Resources/js/containers/MediaCollection/CollectionSection.js
```diff
@@ -113,9 +113,10 @@ class CollectionSection extends React.Component<Props> {
     };
 
     handleSaveResponse = (resourceStore: ResourceStore) => {
+        const openedCollectionOperationOverlayType = this.openedCollectionOperationOverlayType;
         this.closeCollectionOperationOverlay();
 
-        if (this.openedCollectionOperationOverlayType === 'update') {
+        if (openedCollectionOperationOverlayType === 'update') {
```
wachterjohannes, line 122: @alexander-schranz because of this bug the wrong resourceStore will be updated.

alexander-schranz: @wachterjohannes Can we add a test for this, else somebody will revert it accidently in the future.

src/Sulu/Bundle/MediaBundle/Resources/js/containers/MediaCollection/tests/MediaCollection.test.js
```diff
@@ -859,6 +860,7 @@ test('Should send a request to add a new collection via the overlay', () => {
         title: 'Title',
         _permissions: {},
     };
+    // collectionStore.setMultiple = jest.fn();
```
alexander-schranz: not required? then we should remove it.

wachterjohannes: 😂  your are to fast

src/Sulu/Bundle/MediaBundle/Controller/CollectionController.php
```diff
@@ -387,7 +387,7 @@ protected function getData(Request $request)
      */
     protected function saveEntity($id, Request $request)
     {
-        $parent = $request->get('parent');
+        $parent = $request->request->get('parent');
```
alexander-schranz: The `parent` is set as a query parameter and security is as I know checked against the query parameter and so this should be kept that way. As this would be a bc break.

alexander-schranz: As this would be a bc break.

alexander-schranz (changes requested):

src/Sulu/Bundle/MediaBundle/Controller/CollectionController.php
wachterjohannes: you are right have reverted it but the controller uses `$request->get('parent')` so the change in the JS code also works like this 👍

alexander-schranz (approved):

PR #6395, opened 2021-12-01
Fix preserving scroll-position in preview

mavuio:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | 
| Related issues/PRs | 
| License | MIT
| Documentation PR |

#### What's in this PR?

Scroll Position of Preview Pane is restored after each content-update.

#### Why?

When editing content or blocks on larger pages the editor needs to scroll down in the preview in order
to see the content one is working on.

Unfortunately on each consequent update, the preview-pane scrolls to the top, moving the interesing part of the preview out of sight.

### DIFF
--- a/src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/Preview.js
+++ b/src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/Preview.js
@@ -204,14 +204,16 @@ class Preview extends React.Component<Props> {
 
     setContent = (previewContent: string) => {
         const previewDocument = this.getPreviewDocument();
-
         if (!previewDocument) {
             return;
         }
 
+        const preservedScrollPos=this.getPreviewScrollPos();
         previewDocument.open(); // This will lose in Firefox the and safari previewDocument.location
         previewDocument.write(previewContent);
         previewDocument.close();
+        setTimeout(() => this.setPreviewScrollPos(preservedScrollPos), 100);
+
     };
 
     componentWillUnmount() {
@@ -250,6 +252,32 @@ class Preview extends React.Component<Props> {
 
         return this.iframeRef.contentDocument;
     };
+
+    getPreviewWindow= (): ?Window => {
+        if (this.previewWindow) {
+            return this.previewWindow;
+        }
+
+        if (!(this.iframeRef instanceof HTMLIFrameElement)) {
+            return;
+        }
+        return this.iframeRef.contentWindow;
+    };
+
+
+    getPreviewScrollPos = ()  => {
+        const win=this.getPreviewWindow();
+        if(win) {
+            return win.document.documentElement.scrollTop || win.pageYOffset || win.document.body.scrollTop;
+        }
+    };
+
+    setPreviewScrollPos = (pos)  => {
+        const win=this.getPreviewWindow();
+        if(win) {
+            win.scrollTo({ top: pos});
+        }
+    }
 
     @action setIframe = (iframeRef: ?Object) => {
         this.iframeRef = iframeRef;
--- a/src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/Preview.js
+++ b/src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/Preview.js
@@ -208,12 +208,11 @@ class Preview extends React.Component<Props> {
             return;
         }
 
-        const preservedScrollPos=this.getPreviewScrollPos();
+        const preservedScrollPosition = this.getPreviewScrollPosition();
         previewDocument.open(); // This will lose in Firefox the and safari previewDocument.location
         previewDocument.write(previewContent);
         previewDocument.close();
-        setTimeout(() => this.setPreviewScrollPos(preservedScrollPos), 100);
-
+        setTimeout(() => this.setPreviewScrollPosition(preservedScrollPosition), 0);
     };
 
     componentWillUnmount() {
@@ -253,7 +252,7 @@ class Preview extends React.Component<Props> {
         return this.iframeRef.contentDocument;
     };
 
-    getPreviewWindow= (): ?Window => {
+    getPreviewWindow = (): ?any => {
         if (this.previewWindow) {
             return this.previewWindow;
         }
@@ -261,23 +260,25 @@ class Preview extends React.Component<Props> {
         if (!(this.iframeRef instanceof HTMLIFrameElement)) {
             return;
         }
+
         return this.iframeRef.contentWindow;
     };
 
-
-    getPreviewScrollPos = ()  => {
-        const win=this.getPreviewWindow();
-        if(win) {
-            return win.document.documentElement.scrollTop || win.pageYOffset || win.document.body.scrollTop;
+    getPreviewScrollPosition = () => {
+        const previewWindow = this.getPreviewWindow();
+        if (previewWindow) {
+            return previewWindow.document.documentElement.scrollTop
+                || previewWindow.pageYOffset
+                || previewWindow.document.body.scrollTop;
         }
     };
 
-    setPreviewScrollPos = (pos)  => {
-        const win=this.getPreviewWindow();
-        if(win) {
-            win.scrollTo({ top: pos});
+    setPreviewScrollPosition = (pos) => {
+        const previewWindow = this.getPreviewWindow();
+        if (previewWindow) {
+            previewWindow.scrollTo({top: pos});
         }
-    }
+    };
 
     @action setIframe = (iframeRef: ?Object) => {
         this.iframeRef = iframeRef;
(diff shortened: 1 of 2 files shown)


src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/Preview.js
```diff
@@ -192,14 +192,16 @@ class Preview extends React.Component<Props> {
...
-
         if (!previewDocument) {
             return;
         }
 
+        const preservedScrollPos=this.getPreviewScrollPos();
         previewDocument.open(); // This will lose in Firefox the and safari previewDocument.location
         previewDocument.write(previewContent);
         previewDocument.close();
+        setTimeout(() => this.setPreviewScrollPos(preservedScrollPos), 100);
```
alexander-schranz: A timeout mostly means nothing good. What happens if we directly scroll to the position?

src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/Preview.js
```diff
@@ -235,6 +237,32 @@ class Preview extends React.Component<Props> {
         return this.iframeRef.contentDocument;
     };
 
+    getPreviewWindow= (): ?Window => {
+        if (this.previewWindow) {
+            return this.previewWindow;
+        }
+
+        if (!(this.iframeRef instanceof HTMLIFrameElement)) {
```
alexander-schranz, line 263: is this working as expected mostly `instanceof` makes problem if there is a build. Is there no other way to check if its a window or an iframe, which is already used somewhere /cc @wachterjohannes

src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/Preview.js
```diff
@@ -235,6 +237,32 @@ class Preview extends React.Component<Props> {
         return this.iframeRef.contentDocument;
     };
 
+    getPreviewWindow= (): ?Window => {
```
alexander-schranz: check the code style here.

src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/Preview.js
```diff
@@ -235,6 +237,32 @@ class Preview extends React.Component<Props> {
...
+        }
+
+        if (!(this.iframeRef instanceof HTMLIFrameElement)) {
+            return;
+        }
+        return this.iframeRef.contentWindow;
+    };
+
+
+    getPreviewScrollPos = ()  => {
```
alexander-schranz: we never used abbrevations for any functions or variable names. So this should be `getPreviewScrollPosition`

src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/Preview.js
```diff
@@ -235,6 +237,32 @@ class Preview extends React.Component<Props> {
...
+        const win=this.getPreviewWindow();
```
alexander-schranz: same as above we never use any kind of abbrevations so this should be window or better `previewWindow`.

src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/Preview.js
```diff
@@ -235,6 +237,32 @@ class Preview extends React.Component<Props> {
...
+            return win.document.documentElement.scrollTop || win.pageYOffset || win.document.body.scrollTop;
+        }
+    };
+
+    setPreviewScrollPos = (pos)  => {
+        const win=this.getPreviewWindow();
+        if(win) {
+            win.scrollTo({ top: pos});
+        }
+    }
```
alexander-schranz: same as above no abbrevations should be used.

src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/Preview.js
```diff
@@ -235,6 +237,32 @@ class Preview extends React.Component<Props> {
         return this.iframeRef.contentDocument;
     };
 
+    getPreviewWindow= (): ?Window => {
+        if (this.previewWindow) {
+            return this.previewWindow;
+        }
+
+        if (!(this.iframeRef instanceof HTMLIFrameElement)) {
+            return;
```
alexander-schranz, line 264: should here not the window returned and so in one case the iframe scroll position read and set and in other case the window scroll position?

alexander-schranz (commented):
@mavuio Thank you for your pull request. I added some minimal comments. /cc @wachterjohannes can you have a look at the `getPreviewWindow` method is there currently no other way to get the correct item iframe or window? to set or get the scroll position.

src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/Preview.js
wachterjohannes: @alexander-schranz we already used `instanceof` and i thought that there is no problem until now. what are you thinking about? (see https://github.com/sulu/sulu/blob/2.x/src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/Preview.js#L232)

src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/Preview.js
mavuio: unfortunately nothing happend w/o the timeout.
the iframe's content needs to render the updated html first, so we need to wait a little. 
instead of waiting 100ms we should maybe get triggered by some ready-event of the iframe, but that was beyond my time.

mavuio:
today i was checking that in https://sulu.rocks.

 when i edit content there, i nearly never lose scroll-positions in preview.
but when trying in my own project with vanilla sulu (but custom template) i nearly always lose scroll-position in preview.

i wouldn't need that pull request at all if it mostly works without re-scroll.


maybe its an issue  in my templates ?

➜  my question: are you aware of this problem in real-world conditions or was i the only one having issues with that.

src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/Preview.js
alexander-schranz: Sounds good then it works with native objects without problems. Transpiled classes make problems to be checkt with `instanceof`.

ghost:
> ➜ my question: are you aware of this problem in real-world conditions or was i the only one having issues with that.

We had the problem too. When we removed our cookie consent manager for the preview, it worked. Probably the script did something with the position of the body element.

mavuio:
> > ➜ my question: are you aware of this problem in real-world conditions or was i the only one having issues with that.
> 
> We had the problem too. When we removed our cookie consent manager for the preview, it worked. Probably the script did something with the position of the body element.

yep, so it's basically not needed if the frontend code is not messing things up.

but the way we handle frontend-stuff we might mess things up, so we might need this feature in some cases,
it's really annoying when your preview always scrolls away.

it does no great harm to re-scroll after preview, does it ? 

i could do the re-scroll only, if the scroll-pos after the update is 0 and was more before.
which prevents a re-scroll on an already scrolled page.

➜ so i will find some time to continue to work on this PR.

src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/Preview.js
alexander-schranz: Is a timeout 0 (which means just push the method behind the JS event loop stack) maybe enough? And sorry for the late response here.

alexander-schranz:
Currently I see this issue happing more and more. There was also some request about always scroll to current written text but that would need some "pattern matching" as we can not force it todo something here for the devs. But maybe currently a simple solution like this would be enough.

src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/Preview.js
mavuio: i think i tried with 0 first.

will try again,  building my variant of 2.5 now, and will bring this PR forward hopefully. keep u updated

Prokyonn:
FYI this is also browser related, in chrome Version 117.0.5938.149 (Official Build) (arm64) it does preserve the scroll-position but in Firefox 118.0.2 (64-bit) it does not work at all

src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/Preview.js
alexander-schranz: I tried it with 0 and had no problems here so will go with that one.

alexander-schranz (approved):

alexander-schranz:
@mavuio Thank you!

PR #6406, opened 2021-12-03
Pass original request server parameters to target group request

lgraubner:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #
| Related issues/PRs | #
| License | MIT

#### What's in this PR?

This PR passes the original request server headers to the audience targeting request.

#### Why?

Requests handled through `_sulu_target_group` result in a `UrlMatchNotFoundException There exists no portal for the URL "_sulu_target_group"`. As the `ForwardedUrlRequestProcessor` creates a new request without passing the original requests `host` header it falls back to `localhost` which does not seem to work. Setting the URL to `localhost` in the webspace config "solves" the problem as it matches the default of `localhost`.

### DIFF
--- a/src/Sulu/Bundle/AudienceTargetingBundle/Request/ForwardedUrlRequestProcessor.php
+++ b/src/Sulu/Bundle/AudienceTargetingBundle/Request/ForwardedUrlRequestProcessor.php
@@ -40,7 +40,7 @@ class ForwardedUrlRequestProcessor implements RequestProcessorInterface
             return new RequestAttributes();
         }
 
-        $originalRequest = Request::create($request->headers->get($this->urlHeader), 'GET', [], [], [], $request->server);
+        $originalRequest = Request::create($request->headers->get($this->urlHeader), 'GET', [], [], [], $request->server->getHeaders());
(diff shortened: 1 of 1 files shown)


wachterjohannes:
@lgraubner i cannot reproduce your problem but it does not seem to break anything - thanks for the contribution. Can you maybe add a comment about your thoughts in the class and a testcase to the TestClass `src/Sulu/Bundle/AudienceTargetingBundle/Tests/Unit/Request/ForwardedUrlRequestProcessorTest.php` to ensure that nobody remove the headers from the class.

wachterjohannes (changes requested):
see comment above

wachterjohannes:
@lgraubner OK i see my problem 🙈  i have configured the URL in the webspace. with the {host} parameter it really seems to break.

But my comment above is still valid

alexander-schranz:
As a bugfix we should target the `2.2` branch for this as I think the bug also exist there?

wachterjohannes (approved):

PR #6410, opened 2021-12-07
Fix php-cs and ci tests

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no yes
| New feature? | no yes
| BC breaks? | no yes
| Deprecations? | no yes 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### To Do

### DIFF
--- a/.github/workflows/test-application.yaml
+++ b/.github/workflows/test-application.yaml
@@ -138,6 +138,10 @@ jobs:
                   composer remove jackalope/jackalope-doctrine-dbal --dev --no-interaction --no-update
                   composer require jackalope/jackalope-jackrabbit:~1.2  --no-interaction --no-update
 
+            - name: Require Laminas Zend bridge
+              if: ${{ matrix.php-version == '7.4' }}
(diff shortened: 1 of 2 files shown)


.github/workflows/test-application.yaml
```diff
@@ -138,6 +138,10 @@ jobs:
                   composer remove jackalope/jackalope-doctrine-dbal --dev --no-interaction --no-update
                   composer require jackalope/jackalope-jackrabbit:~1.2  --no-interaction --no-update
 
+            - name: Require Laminas Zend bridge
+              if: ${{ matrix.php-version == '7.4' }}
+              run: composer require laminas/laminas-zendframework-bridge --no-update
```
alexander-schranz, line 143: This is required because of a bug in `symfony/proxy-manager-bridge` but the never fixed it for Symfony 3.4 as when it did appear 3.4 was not longer supported. Sadly we can not require it directly as the library does not support our PHP Version range.

PR #6415, opened 2021-12-07
Test mariadb in CI and postgre against jackalope doctrine dbal

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? |  yes
| New feature? | no 
| BC breaks? | no 
| Deprecations? | no  
| Fixed tickets |  # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Add mariadb to the test matrix.

#### Why?

Catching issues of mariadb.

### DIFF
--- a/.github/workflows/test-application.yaml
+++ b/.github/workflows/test-application.yaml
@@ -97,6 +97,18 @@ jobs:
                           DATABASE_CHARSET: UTF8
                           DATABASE_COLLATE:
 
+                    - php-version: '7.3'
+                      database: postgres
+                      phpcr-transport: doctrinedbal
+                      dependency-versions: 'highest'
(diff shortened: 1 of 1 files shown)


.github/workflows/test-application.yaml
```diff
@@ -131,6 +131,14 @@ jobs:
                     - 3306:3306
                 options: --health-cmd="mysqladmin ping" --health-interval=10s --health-timeout=5s --health-retries=5
 
+            mariadb:
+                image: mariadb:10.3
```
alexander-schranz: @chirimoya should we add mariadb to our test matrix? Then I would go by adding here a 10.7 or 10.6 version (they have a strange versioning 10.7.1 is only a preview and not a stable version https://en.wikipedia.org/wiki/MariaDB).

PR #6423, opened 2021-12-13
Rename queryBuilder "system" parameter to "roleSystem"

### LINKED ISSUE #6412: Crash on MariaDB because of reserved keyword in query
| Q | A
| --- | ---
| Sulu Version | 2.3.7
| PHP Version | 7.4.26
| DB Version | 10.3.32-MariaDB-0ubuntu0.20.04.1 - Ubuntu 20.04
| Browser Version | Brave

# Actual Behavior

Sulu crash on pages using the permission system (error 500) on some versions of MariaDB.
It seems to be related to the usage of the reserved keyword "system" in Sulu\Bundle\SecurityBundle\Entity\RoleRepository.
This page https://mariadb.com/kb/en/reserved-words/#oracle-mode says that "system" is a reserved keyword in the "Oracle mode" section. 
`SYSTEM: <= MariaDB 10.3.6 only. Note however that SYSTEM sometimes needs to be quoted to avoid confusion with System-versioned tables.`

I'm not sure what is this oracle mode, my db settings are as follow:
`DATABASE_URL=mysql://root:root@mysql:3306/xxx?serverVersion=mariadb-10.3.32`

# Expected Behavior

It should'nt crash :)

# Steps to Reproduce

Activate the permission system on some page.
Use an affected version of MariaDb
Try to access to these pages through the website or the admin

Note: my local install (docker stack) has this version of MariaDb:
`10.3.25-MariaDB-1:10.3.25+maria~focal - mariadb.org binary distribution`
I can't explain why but it works without problem here although it should be affected if the problem is related to the reserved keywords and the MariaDb version.

# Possible Solutions

No idea if this is a proper fix but I can workaround the crash by changing the queries in Sulu\Bundle\SecurityBundle\Entity\RoleRepository

Original code:

```php
public function findRoleIdsBySystem($system)
    {
        $result = $this->createQueryBuilder('role')
            ->select('role.id')
            ->where('role.system = :system')
            ->setParameter('system', $system)
            ->getQuery()
            ->getResult();

        return \array_map(function($role) {
            return $role['id'];
        }, $result);
    }

```

Updated code (I just changed the parameter name "system" to "systemRenamed"):

```php
public function findRoleIdsBySystem($system)
    {
        $result = $this->createQueryBuilder('role')
            ->select('role.id')
            ->where('role.system = :systemRenamed')
            ->setParameter('systemRenamed', $system)
            ->getQuery()
            ->getResult();

        return \array_map(function($role) {
            return $role['id'];
        }, $result);
    }
```

The findAllRoles() method has a "system" parameter too, I suppose it could lead to a crash but I've not changed it.
Note that I have no idea if other parts of the code are affected by this change.

danakil:
Rename querybuilder "system" parameter to "roleSystem" to avoid reserved keywords issues (see issue 6412)

| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| Fixed tickets | fixes #6412
| Related issues/PRs | #6412
| License | MIT

#### What's in this PR?

Rename the querybuilder "system" paramter to "roleSystem".

#### Why?

Workaround for reserved keywords issues. See issue #6412.

### DIFF
--- a/src/Sulu/Bundle/SecurityBundle/Entity/RoleRepository.php
+++ b/src/Sulu/Bundle/SecurityBundle/Entity/RoleRepository.php
@@ -77,8 +77,8 @@ class RoleRepository extends EntityRepository implements RoleRepositoryInterface
             }
 
             if (isset($filters['system'])) {
-                $queryBuilder->andWhere('role.system = :system')
-                    ->setParameter('system', $filters['system']);
+                $queryBuilder->andWhere('role.system = :roleSystem')
+                    ->setParameter('roleSystem', $filters['system']);
             }
 
             $query = $queryBuilder->getQuery();
@@ -107,8 +107,8 @@ class RoleRepository extends EntityRepository implements RoleRepositoryInterface
     {
         $result = $this->createQueryBuilder('role')
             ->select('role.id')
-            ->where('role.system = :system')
-            ->setParameter('system', $system)
+            ->where('role.system = :roleSystem')
+            ->setParameter('roleSystem', $system)
             ->getQuery()
             ->getResult();
 


alexander-schranz (approved):

alexander-schranz:
@danakil Thank you for your contribution!

PR #6434, opened 2021-12-20
Fix permission check for document-webspace instead of passed webspace

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes none
| Related issues/PRs | none
| License | MIT
| Documentation PR | none

#### What's in this PR?

As the PHPCRTeaserProvider does not pass any webspace - because it has no idea which webspace the selected documents have - the Permission check was deactivated.

But the place where the check is done has all the information which are needed todo the check. So we use the local information there instead of the passed webspace-key.

#### Why?

The teaser-selection should check the permissions.

### DIFF
--- a/src/Sulu/Bundle/PageBundle/Tests/Functional/SmartContent/QueryBuilderTest.php
+++ b/src/Sulu/Bundle/PageBundle/Tests/Functional/SmartContent/QueryBuilderTest.php
@@ -463,7 +463,7 @@ class QueryBuilderTest extends SuluTestCase
         $request = new Request([], [], ['_sulu' => new RequestAttributes(['webspace' => $webspace])]);
         $this->getContainer()->get('request_stack')->push($request);
 
-        $root = $this->sessionManager->getContentNode('test_io');
+        $root = $this->sessionManager->getContentNode('sulu_io');
 
         /** @var PageDocument $document */
         $document = $this->documentManager->create('page');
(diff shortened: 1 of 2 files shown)


src/Sulu/Bundle/PageBundle/Tests/Functional/SmartContent/QueryBuilderTest.php
```diff
@@ -463,15 +463,15 @@ public function testSecurityWithoutPermissionCheck()
         $request = new Request([], [], ['_sulu' => new RequestAttributes(['webspace' => $webspace])]);
         $this->getContainer()->get('request_stack')->push($request);
 
-        $root = $this->sessionManager->getContentNode('test_io');
```
wachterjohannes, line 466: @alex `test_io` has `<security permission-check="true">` as the name is **WITHOUT**PermissionCheck i have changed this to `sulu_io` which has `<security permission-check="false">`

alexander-schranz (approved):

src/Sulu/Bundle/PageBundle/Tests/Functional/SmartContent/QueryBuilderTest.php
alexander-schranz: Make sense. Thank you for explanation!

PR #6440, opened 2021-12-26
Add possibility for query parameters in sulu link

### LINKED ISSUE #6418: Query parameters in <sulu-link> tag
# The `<sulu-link>` tag only supports anchors

Currently the `<sulu-link>` tag only allows anchors using `#` in the `href` attribute. IMHO it should be possible to also add query parameters using a `?` here.

# Proposed solution

~~A workaround is to use an empty anchor followed by the query parameters: `123-123-123-123#?query=parameter`.~~

EDIT: I did not check if the resulting link actually works. This is not the case, as no new request is send to the server if the page is the same.

After taking a look at the code, it doesn't seem to be to difficult to add the possibility to support query parameters directly and I would be willing to create a pull request.

exastion:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #6418
| Related issues/PRs | #6418
| License | MIT
| Documentation PR | sulu/sulu-docs#692

#### What's in this PR?

This PR adds the possibility to use query parameters in the `sulu-link` tags.

#### To Do

### DIFF
--- a/src/Sulu/Bundle/MarkupBundle/Markup/LinkTag.php
+++ b/src/Sulu/Bundle/MarkupBundle/Markup/LinkTag.php
@@ -65,9 +65,10 @@ class LinkTag implements TagInterface
             $provider = $this->getValue($attributes, 'provider', self::DEFAULT_PROVIDER);
             $validationState = $attributes['sulu-validation-state'] ?? null;
 
-            $hrefParts = $this->getUuidAndAnchorFromHref($attributes['href'] ?? null);
+            $hrefParts = $this->getPartsFromHref($attributes['href'] ?? null);
             $uuid = $hrefParts['uuid'];
             $anchor = $hrefParts['anchor'];
+            $query = $hrefParts['query'];
 
             if ($uuid && \array_key_exists($provider . '-' . $uuid, $contents)) {
                 $item = $contents[$provider . '-' . $uuid];
@@ -79,6 +80,10 @@ class LinkTag implements TagInterface
 
                 if ($anchor) {
                     $url .= '#' . $anchor;
+                }
+
+                if ($query) {
+                    $url .= '?' . $query;
                 }
 
                 $title = $item->getTitle();
@@ -125,7 +130,7 @@ class LinkTag implements TagInterface
         $result = [];
         foreach ($attributesByTag as $tag => $attributes) {
             $provider = $this->getValue($attributes, 'provider', self::DEFAULT_PROVIDER);
-            $uuid = $this->getUuidAndAnchorFromHref($attributes['href'] ?? null)['uuid'];
+            $uuid = $this->getPartsFromHref($attributes['href'] ?? null)['uuid'];
 
             if (!$uuid || !\array_key_exists($provider . '-' . $uuid, $items)) {
                 $result[$tag] = self::VALIDATE_REMOVED;
@@ -155,7 +160,7 @@ class LinkTag implements TagInterface
                 $uuidsByType[$provider] = [];
             }
 
-            $uuid = $this->getUuidAndAnchorFromHref($attributes['href'] ?? null)['uuid'];
+            $uuid = $this->getPartsFromHref($attributes['href'] ?? null)['uuid'];
 
             if ($uuid) {
                 $uuidsByType[$provider][] = $uuid;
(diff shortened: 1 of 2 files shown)


src/Sulu/Bundle/MarkupBundle/Markup/LinkTag.php
```diff
@@ -209,21 +214,39 @@ private function getContent(array $attributes)
...
 
-        /** @var string[] $hrefParts */
-        $hrefParts = $href ? \explode('#', $href, 2) : [];
-
-        $uuid = $hrefParts[0] ?? null;
-        $anchor = $hrefParts[1] ?? null;
+        $query = null;
+        $anchor = null;
+
+        $hasQuery = strpos($href, '?');
```
alexander-schranz: @exastion  Something like `my.url/#hello?page=1` means that `hello?page=1` is the value of the anchor and it still a valid anchor. This is as example used in Single Page Application and are valid anchors. Sulu example use this kind of special anchor for frontend navigation.

That's why its important first to split the string by `#` and then go with split the first part by `?` and not the other way around. So there is no problem also from backward compatibility we are not allowed to change to seperate attributes and as said its not really required.

So the following examples should show you best how the end result should look like:

```bash
<uuid>?test=query#anchor

ID: <uuid>
query: test=query
anchor: anchor
```


```bash
<uuid>#anchor?test=query

ID: <uuid>
query: 
anchor: anchor?test=query
```

```bash
<uuid>?test=query

ID: <uuid>
query: test=query
anchor: 
```

See https://3v4l.org/j94IN#v7.2.34 how to correctly split the string. A `?` is a valid symbol as part of the anchor so we first need to split by `#` and then split the part before the anchor by `#`.

alexander-schranz (changes requested):
Thank you for working on this. We need to split the parts correctly first by `#` and then the first part by `?` I added a comment and an example how to achieve this.

src/Sulu/Bundle/MarkupBundle/Markup/LinkTag.php
exastion: I did not know about these specifics up to now, I even thought in the beginning, query and anchor would be the other way around.. Well you always learn :)
This made things also easier.

alexander-schranz: Also nice to know the `#anchor` is never send to a server its only available for the browser. So if your browser requests `https://sulu.io/services/support#chat` a request is only send to `https://sulu.io/services/support` and browsers is handling `#chat` to scroll to the correct part. Or a javascript library is handling the `#anchor` like its done in the sulu admin.

src/Sulu/Bundle/MarkupBundle/Markup/LinkTag.php
```diff
@@ -209,21 +214,24 @@ private function getContent(array $attributes)
...
-    private function getUuidAndAnchorFromHref($href): array
+    private function getPartsFromHref($href): array
     {
         $href = (string) $href ?: null;
 
         /** @var string[] $hrefParts */
         $hrefParts = $href ? \explode('#', $href, 2) : [];
+        $anchor = $hrefParts[1] ?? null;
 
+        $hrefParts = \explode('?', $hrefParts[0], 2);
```
alexander-schranz: Some tests seems to fail here not sure in which cases `$hrefParts[0]` can not exist.

https://github.com/sulu/sulu/runs/4657974922?check_suite_focus=true

alexander-schranz (changes requested):
To run the failing tests you can use:

```
bin/runtests -i -t MarkupBundle
```

The `-i` is only required at first run.

exastion:
What is the issue with the two Node tests?

alexander-schranz:
@exastion as you didn't change anything in the JS you can ignore them. Maybe some dependency changes did make them fail now. We will have a look at them in a separate pull request.

alexander-schranz:
Issue was found here: https://github.com/browserify/resolve/pull/269. Hopefully we get quick a response from them so we can continue here.

alexander-schranz:
@chirimoya should we add the possibility to add `query` to the link content type and text editor link overlay?

alexander-schranz (approved):

alexander-schranz:
@exastion Thank you for the contribution. I did merge this part as it can be used as it's own. Do you want to work also on the text editor integration for the `query`? There is currently possible to define a `anchor` but `query` should also be possible:

https://github.com/sulu/sulu/blob/2.4.0/src/Sulu/Bundle/AdminBundle/Resources/js/containers/CKEditor5/plugins/InternalLinkPlugin/InternalLinkPlugin.js

exastion:
@alexander-schranz I'm not too familiar with react, but I will give it a try.

alexander-schranz:
@exastion it is mostly the same as for the anchor was done some time ago. The following pull reqest should help you to find the files: https://github.com/sulu/sulu/pull/6090/files which are effected. Thank you for giving it a try :+1:

exastion:
@alexander-schranz I think I did all the necessary changes. However I am not able to test it.
I am not able to initialize the database to run the AdminBundle tests, as it always uses root with no password for the database, despite changing the .env file.
I also cannot test it in the browser, as I don't seem to be able to build the admin interface directly in the sulu repository.

alexander-schranz:
For the ckeditor the tests are in the frontend so you run them with npm:

```js
npm install

# fix code style
npm run lint:js -- --fix
npm run lint:scss -- --fix

# run typechecks
npm run flow

# run tests
npm run test
```

You can also just open the new pull request there we also will see what in the CI will fail.

Still for completeness the backend tests which I think you did not needed to adjust something for the ckeditor you can change the database in the `src/Sulu/Bundle/TestBundle/Resources/app/config/parameters.yml` file which was created. To run tests use the `bin/runtests -i -a` script. Or a specific bundle `bin/runtests -i -C -t AdminBundle`. There is also `--flags="--filters .."` to add phpunit filter option

Issue #6444, opened 2022-01-06
Make the admin interface more accessible


FlorianBoe:
# Problem description
In a project we are currently looking for a new CMS for a public university in Germany. SULU looks very promising. Unfortunately the editor interface is not accessible, one of our selection criteria (ATAG compliance).

A few problems I could find during a quick test of the admin interface are:

## Parts of the admin interface are not keyboard operable

- [X] #6446
- [x] #6942
- [x] #6949
- [x] #6951
- [x] #6952
- [ ] Dropdowns only expand on click. Options in the dropdown list cannot be reached through the keyboard. This is true for Dropdowns in the upper Toolbar and Popover components in the preview toolbar.
- [ ] Modals (Portals) can not be closed by keyboard. The 'x'-icon only triggers on click.
- [ ] Div-elements are used as buttons (search for `role="button"` in the AdminBundle), but only have a `onClick` handler. Use native button elements or add a `tabindex="0"` and handler for the Enter and Space keypresses.
- [ ] SmartContentItems
- [ ] PageList

## Not visible elements can receive keyboard focus

- [x] Elements in the hidden navigation can be focused
- [ ] Content behind an open modal is keyboard accessible, the focus is not trapped in the modal (look at this example of an [accessible modal pattern](https://scottaohara.github.io/accessible_modal_window/))

## The focus state is not always visible

- [ ] #6948
- [ ] All interactive elements with a hover state should have a focus state (eg. the Button component, the button in the MultiItemSelection component and the navigation toggle button). Search for `:hover` to find all occurrences.

## The correct semantic elements and aria roles are not utilized

- [x] #6943
- [x] #6953
- [ ] The Backdrop has a `role="button"`, but does not behave like a native button. In this case, it is sufficient to remove the role attribute, since the function is also available elsewhere via keyboard.
- [ ] The Icon component always renders an aria label, even for non-interactive elements like the SULU logo (more below).
- [ ] Missing relationship between inputs and help and error messages

## Elements do not have meaningful accessible names

- [ ] Remove the `aria-label` from the Icon component. This currently adds the icon name (e.g. "su-webspace") to the the computed accessible name (e.g. assistive technologies like screen readers will read "su-webspace Webspaces"). This change makes it visible that some interactive elements are missing an meaningful [accessible name](https://www.tpgi.com/what-is-an-accessible-name/).
[...]

PR #6446, opened 2022-01-06
Make main navigation keyboard operable

FlorianBoe:
Every option in the menu is now accessible by keyboard and present in tab order. A few minor changes to the styles were necessary.

| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets |
| Related issues/PRs | #6444
| License | MIT

#### What's in this PR?

- A little refactoring of the navigation. The navigation now uses native button elements for interactive elements.
- Child elements of the UserSection are removed from the tab order with the hidden attribute if they are not visible.

#### Why?

The navigation items could not be focused and triggered with the keyboard.

#### To Do

- [x] Update Snapshots

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Navigation/item.scss
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Navigation/item.scss
@@ -8,7 +8,6 @@ $hoverColor: $shakespeare;
 .item {
     font-size: 14px;
     line-height: 16px;
-    cursor: pointer;
     box-sizing: border-box;
     padding-left: 4px;
     position: relative;
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Navigation/item.scss
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Navigation/item.scss
@@ -31,6 +31,7 @@ $hoverColor: $shakespeare;
         background-color: transparent;
         text-align: left;
         font-weight: inherit;
+        cursor: pointer;
 
         &:hover,
         &:focus {
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Navigation/item.scss
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Navigation/item.scss
@@ -23,10 +23,17 @@ $hoverColor: $shakespeare;
     .title {
         display: flex;
         align-items: center;
+        justify-content: flex-start;
+        width: 100%;
         color: $color;
         padding: 20px;
+        border: 0;
+        background-color: transparent;
+        text-align: left;
+        font-weight: inherit;
 
-        &:hover {
+        &:hover,
+        &:focus {
             color: $hoverColor;
         }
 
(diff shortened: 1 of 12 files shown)


alexander-schranz (changes requested):
@FlorianBoe Thank you for providing a pull request. If you migrate the role="button" to a real button we should always define them as `<button type="button"` else it would in some cases think its `type="submit"` which is not the case here.

FlorianBoe:
I can add the type attribute, but I don't think it would be necessary in this case. The problem refers to buttons that appear within form elements, but the navigation is not nested in a native form element.

alexander-schranz:
@FlorianBoe Thank you, really appricate that you work on this issue. We always should define the type of a button, as it could be happen that an element will be nested in future in a `form` tag and to avoid that then a button is handled as submit we should define them as `type="button"` as we don't know which components maybe are nested into each other.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Application/global.scss
```diff
@@ -55,6 +55,7 @@ h3 {
 button {
     border-radius: 0;
     padding: 0;
+    cursor: pointer;
```
alexander-schranz: @FlorianBoe We are developing with a component based css styles and avoid to have global styling for elements. So this should not be here and the `cursor: pointer` if it is missing at some component should be defined there in the component css. Adding button { cursor: pointer } here would make some unclickable disabled button now cursor pointer which can confuse users, only the component know which button state and which cursor make sense there. So this should be moved to the component you see is missing cursor: pointer.

```suggestion
```

src/Sulu/Bundle/AdminBundle/Resources/js/components/Navigation/item.scss
```diff
@@ -31,6 +31,7 @@ $hoverColor: $shakespeare;
         background-color: transparent;
         text-align: left;
         font-weight: inherit;
+        cursor: pointer;
```
alexander-schranz, line 33: The `item` already has cursor: pointer. Think the title does then not need it?

FlorianBoe: The title is now an native button element and inherits styles from the user agent stylesheet. The default for button elements is `cursor: default`, which overwrites the inheritance from the cascade. Maybe the cursor styling can be removed from the item?

alexander-schranz: Make sense as the `button` is the one with the click listener and not the `item`.

alexander-schranz:

```bash
git checkout your-branch
git rebase origin/2.4
```

```
npm install
# fix code style
npm run lint:js -- --fix
npm run lint:scss -- --fix
# run typechecks
npm run flow
# run tests
npm run test
```

alexander-schranz:
@FlorianBoe Thank you! This was a great First-time contribution! 🚀

PR #6448, opened 2022-01-08
Remove the previous mainAccountContact from the contact entity when contact changes from mainAccount

### LINKED ISSUE #6447: Duplicate AccountContact entity when contact changes from account
| Q | A
| --- | ---
| Sulu Version | 2.4.0
| PHP Version | 8.0
| DB Version | MySQL 5.7.36

# Actual Behavior

When the relation between a Sulu contact and a Sulu account has been changed, the previous relation does not get removed. This happens because the AccountContact does get removed but not flushed so when the contact is persisted, the relation has not been removed.

# Expected Behavior

When you change the relation, the previous relation should be removed.

# Steps to Reproduce

First you go to the contact section in the Sulu admin panel. Go to the edit form from a contact and change his organization.
Schermafbeelding 2022-01-07 om 10 04 35
When you go to the database, you can see that the relation has not been removed.
Schermafbeelding 2022-01-07 om 10 03 31
When you go back to the contact section, there are now two contacts for the same contact.
Schermafbeelding 2022-01-07 om 10 06 01

# Possible Solutions

A possible solution is flushing right after the removal of the AccountContact. This means that the relation is deleted before persisting the contact. Here 'https://github.com/sulu/sulu/blob/2.x/src/Sulu/Bundle/ContactBundle/Contact/ContactManager.php#L613' it can look like this:
`$this->em->remove($mainAccountContact);`
`$this->em->flush();`

jasperPeeters98:
…ontact changes from mainAccount

| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #6447
| License | MIT

#### What's in this PR?

When a contact changes from mainAccount, the removeAccountContact function from the contact entity is called to remove the existing mainAccountContact.

#### Why?

This PR will resolve the issue #6447. 
Else when a contact changes from mainAccount, the previous mainAccountContact from the contact entity is not removed.

### DIFF
--- a/src/Sulu/Bundle/ContactBundle/Contact/ContactManager.php
+++ b/src/Sulu/Bundle/ContactBundle/Contact/ContactManager.php
@@ -610,6 +610,7 @@ class ContactManager extends AbstractContactManager implements DataProviderRepos
                     new AccountContactRemovedEvent($mainAccountContact->getAccount(), $mainAccountContact->getContact())
                 );
 
+                $contact->removeAccountContact($mainAccountContact);
                 $this->em->remove($mainAccountContact);
             }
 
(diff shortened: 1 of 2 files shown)


alexander-schranz (approved):
@jasperPeeters98 Thank you for the pull request. I also added a testcase to avoid same issue in the future.

PR #6449, opened 2022-01-09
Add Link Type Constants to Link Content Type

TheCadien:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? | no 
| BC breaks? | no 
| License | MIT

#### What's in this PR?

Add Link Type Constants to Link Content Type

#### Why?

To be able to use them in the SuluHeadlessBundle in the LinkContentTypeResolver. To keep the Code Clean

### DIFF
--- a/src/Sulu/Component/Content/Types/Link.php
+++ b/src/Sulu/Component/Content/Types/Link.php
@@ -23,6 +23,8 @@ use Sulu\Component\Content\SimpleContentType;
 class Link extends SimpleContentType
 {
     public const LINK_TYPE_EXTERNAL = 'external';
+    public const LINK_TYPE_PAGE = 'page';
+    public const LINK_TYPE_MEDIA = 'media';
 
     /**
      * @var LinkProviderPoolInterface


src/Sulu/Component/Content/Types/Link.php
```diff
@@ -23,6 +23,8 @@
 class Link extends SimpleContentType
 {
     public const LINK_TYPE_EXTERNAL = 'external';
+    public const LINK_TYPE_PAGE = 'page';
+    public const LINK_TYPE_MEDIA = 'media';
```
alexander-schranz, line 27: `page` and `media` are dynamic link provider types. They should not be added here as constant. As they come from a registry and more types are possible.
Only the external type is a internal type which we need handle here, specially. Where was it required for you to check for this constants?

alexander-schranz: Saw your description. The headlessbundle should also not care which provider exist it should just work the same way and use the `LinkProviderPool`.

TheCadien: okay, than lets close the PR :)

PR #6454, opened 2022-01-11
Create accept property for FileUploadButton

jasperPeeters98:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

When the FileUploadButton component is used, it is possible to give an accept property to the component.

#### Why?

When using the FileUploadButton component in a custom view. It was not possible to add an accept parameter to the FileUploadButton.

#### Example Usage
```
import { FileUploadButton } from "sulu-admin-bundle/components";

export default class TestToolbarAction extends AbstractListToolbarAction {
   getNode() {
    return (
             <FileUploadButton accept={"application/json"} onUpload={this.handleUpload}>
             Upload JSON file
             </FileUploadButton>
             );
           }
}
```

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/FileUploadButton/FileUploadButton.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/FileUploadButton/FileUploadButton.js
@@ -6,20 +6,20 @@ import type {Node} from 'react';
 import type {ButtonSkin} from '../Button';
 
 type Props = {|
+    accept?: string,
     children?: Node,
     disabled: boolean,
-    icon: string | typeof undefined,
+    icon?: string,
     onUpload: (file: File) => void,
-    skin: ButtonSkin | typeof undefined,
-    accept: string | typeof undefined,
+    skin?: ButtonSkin,
 |};
 
 export default class FileUploadButton extends React.Component<Props> {
     static defaultProps = {
+        accept: undefined,
         disabled: false,
         icon: undefined,
         skin: undefined,
-        accept: undefined,
     };
 
     handleDrop = (files: Array<File>) => {
@@ -30,12 +30,12 @@ export default class FileUploadButton extends React.Component<Props> {
 
     render() {
         const {children, disabled, icon, skin, accept} = this.props;
-        
+
         return (
             <Dropzone
+                accept={accept}
                 onDrop={this.handleDrop}
                 style={{}}
-                accept={accept}
             >
                 {({getInputProps, getRootProps}) => (
                     <div {...getRootProps()}>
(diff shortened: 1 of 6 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/components/FileUploadButton/FileUploadButton.js
```diff
@@ -39,7 +41,7 @@ export default class FileUploadButton extends React.Component<Props> {
                         <Button disabled={disabled} icon={icon} skin={skin}>
                             {children}
                         </Button>
-                        <input {...getInputProps()} />
+                        <input {...getInputProps()} accept={accept} />
```
alexander-schranz: This should  handled over the dropzone params so if you configure `accept` on `<Dropzone` the `getInputProps` will return it correctly.

If we add the `accept` here, we need to make sure this is also possible for all other dropzone components: Quick search:

 - FileUploadButton
 - UploadToolbarAction
 - SingleMediaDropzone
 - MultiMediaDropzone

The integration into sulu should then need to support to configure this new accept prop, as example the `sulu_admin.upload` toolbar action should be possible to be configured with the accept and the `single_media_upload` field type.

/cc @niklasnatter do you see other places where the accept should be kept in mind?

jasperPeeters98: @alexander-schranz Yes indeed, i believe the `Dropzone` from `UploadToolbarAction` already has an `accept` property.

alexander-schranz: You are right 👍 still would add accept to SingleMediaDropzone and MultiMediaDropzone also.

jasperPeeters98: I have added accept for SingleMediaDropzone and MultiMediaDropzone

alexander-schranz:
@jasperPeeters98 Thank you for working on this. Can you rebase your branch and run the node tests and linters and have a look at them. There should also be tests added for the new props.

Rebase:

```bash
git checkout your-branch
git rebase origin/2.4
```

Run npm node tests:

```bash
npm install
# fix code style
npm run lint:js -- --fix
npm run lint:scss -- --fix
# run typechecks
npm run flow
# run tests
npm run test
```

niklasnatter:
Thanks for implementing this! 🙂

niklasnatter (approved):

PR #6455, opened 2022-01-11
Add keypress handler for button icons

FlorianBoe:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets |
| Related issues/PRs | #6444
| License | MIT
| Documentation PR | 

#### What's in this PR?

Adds keyboard events to the icon when an onclick function is passed as prop into the component.

#### Why?

The icon component acts as a button in some cases. The icon can be focused, but cannot be activated with the keyboard, because no keyboard events are registered.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Icon/Icon.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Icon/Icon.js
@@ -29,6 +29,19 @@ export default class Icon extends React.PureComponent<Props> {
         onClick();
     };
 
+    handleKeypress = (event: SyntheticEvent<HTMLElement>) => {
+        const {onClick} = this.props;
+
+        if (!onClick) {
+            return;
+        }
+
+        if (event.key === 'Enter' || event.key === ' ') {
+            event.stopPropagation();
+            onClick();
+        }
+    };
+
     render() {
         const {className, name, onClick, style} = this.props;
         let fontClass = '';
@@ -64,6 +77,7 @@ export default class Icon extends React.PureComponent<Props> {
         const onClickProperties = onClick
             ? {
                 onClick: this.handleClick,
+                onKeyPress: this.handleKeypress,
                 role: 'button',
                 tabIndex: 0,
             }
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Icon/Icon.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Icon/Icon.js
@@ -29,7 +29,7 @@ export default class Icon extends React.PureComponent<Props> {
         onClick();
     };
 
-    handleKeypress = (event: SyntheticEvent<HTMLElement>) => {
+    handleKeypress = (event: SyntheticKeyboardEvent<HTMLElement>) => {
         const {onClick} = this.props;
 
         if (!onClick) {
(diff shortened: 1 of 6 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/components/Icon/Icon.js
```diff
@@ -69,6 +82,7 @@ export default class Icon extends React.PureComponent<Props> {
         const onClickProperties = onClick
             ? {
                 onClick: this.handleClick,
+                onKeyPress: this.handleKeypress,
                 role: 'button',
```
alexander-schranz, line 81: The change make sense: I think we should also don't set a `role` when there is nothing to click?

```suggestion
                role: onClick ? 'button' : null,
```

If a better solution if its possible is that we do:

```
if (onClick) {
    return <button type="button" {...attributes}></button>
}

return <span {...attributes}></span>;
```

We just need to make sure the css styling works for span and button correcly like before.

alexander-schranz: Maybe then we also don't need the `handleKeypress` as it is then supported natively?

FlorianBoe: This component currently tries to be both a button and a non-interactive icon. The icon therefore also needs hover and focus styles depending on its use. In the long term, the cleanest solution would be to use icons only for decorative purposes and always use them inside native buttons. I can take a look at how often the icon component is used with an onclick prop.

The role does not need to be provided with a ternary operator at this point, since the entire block is already just set when an onClick function is defined (two lines above).

To keep the style changes within this pull request small, I would suggest not changing the element type.

alexander-schranz: I agree here that the `icon` should not have a `onClick` and just should have been wrapped with button where required, but at current state we can not change that as that would be a backward compatibility break. But we could deprecate the `onClick` like [here](https://github.com/sulu/sulu/blob/04e987b42a80ede16c6279cc153df3b409622d6e/src/Sulu/Bundle/AdminBundle/Resources/js/views/List/toolbarActions/UploadToolbarAction.js#L36-L40)
and don't use <Icon onClick in our core components and go there with wrapping them in a button where they are clickable. But also I'm not aware where this is currently and we probably need to check all <Icon usages.

FlorianBoe: I added a deprecation warning and found the `onClick` prop in the following components:
`Block`, `Card`, `Chip`, `ColumnList/ItemButton`, `Input`, `Navigation/Item`, `Input`, `Overlay`, `Snackbar`, `Table`, `Table/Row` and `containers/List`. In all these cases, the icon component should now be keyboard operable.

In another pull request, I can look at restructuring the core components.

alexander-schranz:
Same as for the other PR.

```bash
git checkout your-branch
git rebase origin/2.4
```

```
npm install
# fix code style
npm run lint:js -- --fix
npm run lint:scss -- --fix
# run typechecks
npm run flow
# run tests
npm run test
```

src/Sulu/Bundle/AdminBundle/Resources/js/components/Icon/Icon.js
alexander-schranz: We should at current state remove then the `deprecation` of onClick. And add it after we did change it in all other cases. So we are sure not the sulu core itself is throwing a deprecation message and that it also work like we expected to remove the onClick here.

niklasnatter (approved):
Thanks a lot for working on this!

PR #6456, opened 2022-01-11
Remove duplicate event handler

FlorianBoe:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets |
| Related issues/PRs |  #6444
| License | MIT
| Documentation PR |

#### What's in this PR?

While testing, I noticed that an onKeyDown handler was defined on these buttons in addition to the onClick handler. This is unnecessary because native button elements are already triggered by the click event. This redundancy is interesting because a wrong event key string was used for the space button ("Space" instead of " "), so these buttons already fall back to the onClick event handler when the space key is pressed.

#### Why?

An unnecessary duplication with a bug.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Action.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Action.js
@@ -31,14 +31,6 @@ export default class Action<T> extends React.PureComponent<Props<T>> {
         this.triggerButton();
     };
 
-    handleButtonKeyDown = (event: KeyboardEvent) => {
-        if (event.key === 'Enter' || event.key === ' ') {
-            event.preventDefault();
-            event.stopPropagation();
-            this.triggerButton();
-        }
-    };
-
     setButtonRef = (ref: ?ElementRef<'button'>) => {
         const {buttonRef} = this.props;
 
@@ -59,7 +51,6 @@ export default class Action<T> extends React.PureComponent<Props<T>> {
                 <button
                     className={actionStyles.action}
                     onClick={this.handleButtonClick}
-                    onKeyDown={this.handleButtonKeyDown}
                     ref={this.setButtonRef}
                 >
                     {this.props.children}
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Action.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Action.js
@@ -52,6 +52,7 @@ export default class Action<T> extends React.PureComponent<Props<T>> {
                     className={actionStyles.action}
                     onClick={this.handleButtonClick}
                     ref={this.setButtonRef}
+                    type="button"
                 >
                     {this.props.children}
                 </button>
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Action.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Action.js
@@ -31,6 +31,14 @@ export default class Action<T> extends React.PureComponent<Props<T>> {
         this.triggerButton();
     };
 
+    handleButtonKeyDown = (event: KeyboardEvent) => {
+        if (event.key === 'Enter') {
+            event.preventDefault();
+            event.stopPropagation();
+            this.triggerButton();
+        }
+    };
+
     setButtonRef = (ref: ?ElementRef<'button'>) => {
         const {buttonRef} = this.props;
 
@@ -51,6 +59,7 @@ export default class Action<T> extends React.PureComponent<Props<T>> {
                 <button
                     className={actionStyles.action}
                     onClick={this.handleButtonClick}
+                    onKeyDown={this.handleButtonKeyDown}
                     ref={this.setButtonRef}
                     type="button"
                 >
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Option.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Option.js
@@ -40,14 +40,6 @@ export default class Option<T> extends React.PureComponent<Props<T>> {
         this.triggerButton();
     };
 
-    handleButtonKeyDown = (event: KeyboardEvent) => {
-        if (event.key === 'Enter' || event.key === ' ') {
-            event.preventDefault();
-            event.stopPropagation();
-            this.triggerButton();
-        }
-    };
-
     setItemRef = (ref: ?ElementRef<'li'>) => {
         const {
             optionRef,
@@ -110,7 +102,6 @@ export default class Option<T> extends React.PureComponent<Props<T>> {
                     className={optionClass}
                     disabled={disabled}
                     onClick={this.handleButtonClick}
-                    onKeyDown={this.handleButtonKeyDown}
                     ref={this.setButtonRef}
                     style={{minWidth: anchorWidth + ANCHOR_WIDTH_DIFFERENCE}}
                 >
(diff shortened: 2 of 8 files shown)


alexander-schranz:
Same as for the other PR.

```bash
git checkout your-branch
git rebase origin/2.4
```

```
npm install
# fix code style
npm run lint:js -- --fix
npm run lint:scss -- --fix
# run typechecks
npm run flow
# run tests
npm run test
```

src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/tests/Action.test.js
```diff
@@ -19,24 +19,6 @@ test('The component should call the callbacks after a click', () => {
     expect(afterAction).toBeCalled();
 });
 
-test('The component should call the callbacks after press enter', () => {
```
alexander-schranz, line 22: Can we adjust the tests that we are sure this still works?

src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/tests/Action.test.js
```diff
@@ -19,24 +19,6 @@ test('The component should call the callbacks after a click', () => {
...
-    const onClick = jest.fn();
-    const afterAction = jest.fn();
-    const action = shallow(<Action afterAction={afterAction} onClick={onClick} value="my-option">My action</Action>);
-    action.find('button').simulate('keydown', {key: 'Enter', preventDefault: jest.fn(), stopPropagation: jest.fn()});
-    expect(onClick).toBeCalled();
-    expect(afterAction).toBeCalled();
-});
-
-test('The component should call the callbacks after press space', () => {
```
alexander-schranz, line 31: Same for this can we adjust them so we test that the navigation works like expected?

src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/tests/Option.test.js
```diff
@@ -32,20 +32,6 @@ test('A click on the component should fire the callback', () => {
     expect(clickSpy).toBeCalled();
 });
 
-test('Pressing enter on the component should fire the callback', () => {
```
alexander-schranz, line 35: Same here.

src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/tests/Option.test.js
```diff
@@ -32,20 +32,6 @@ test('A click on the component should fire the callback', () => {
...
-    const clickSpy = jest.fn();
-    const option = shallow(<Option onClick={clickSpy}>My option</Option>);
-    option.find('button').simulate('keydown', {key: 'Enter', preventDefault: jest.fn(), stopPropagation: jest.fn()});
-    expect(clickSpy).toBeCalled();
-});
-
-test('Pressing space on the component should fire the callback', () => {
```
alexander-schranz, line 42: Same here.

alexander-schranz (changes requested):
The change make sense for me as a `button` already handles enter and space itself. 

@luca-rath could you test this out if still all cases which you had in mind in #5567 still works. Also we should make sure that it behaves the same in all major browsers (Chrome, Firefox, Safari, Edge).

@FlorianBoe as added if possible I would like to keep the testcases for the keydowns did you run here into problems that `simulate` keydown of jest does behave not the same here the same as `button` keydown in the browser? Thank you again for working on this issues.

src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Option.js
```diff
@@ -110,7 +102,6 @@ export default class Option<T> extends React.PureComponent<Props<T>> {
                     className={optionClass}
                     disabled={disabled}
                     onClick={this.handleButtonClick}
-                    onKeyDown={this.handleButtonKeyDown}
                     ref={this.setButtonRef}
                     style={{minWidth: anchorWidth + ANCHOR_WIDTH_DIFFERENCE}}
```
alexander-schranz, line 115:
```suggestion
                    style={{minWidth: anchorWidth + ANCHOR_WIDTH_DIFFERENCE}}
                    type="button"
```

Did see this only yet sorry, this button should also define a type property.

src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Action.js
```diff
@@ -59,7 +51,6 @@ export default class Action<T> extends React.PureComponent<Props<T>> {
                 <button
                     className={actionStyles.action}
                     onClick={this.handleButtonClick}
-                    onKeyDown={this.handleButtonKeyDown}
                     ref={this.setButtonRef}
```
alexander-schranz, line 63:
```suggestion
                    ref={this.setButtonRef}
                    type="button"
```

same as beneath

src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/tests/Action.test.js
FlorianBoe: The Enter and Space keys trigger the click event for buttons, so an `EventListener` listening for keypresses is not triggered. An explicit test for this scenario may not be necessary, as it ultimately only tests the function of the native button element.

alexander-schranz: Make sense!

alexander-schranz (changes requested):
I rebased this to 2.3, so it will also be fixed there.

While testing (used Chrome 97 and Firefox 96). I found out that `Enter` is not working for me. When Example go to `Contacts`, open a contact and change the `Form of Address` dropdown with the keyboard. `Space` work like expected to select an option but `Enter` does nothing. So maybe the even is not correctly trigger or some preventDefault in a parent container does maybe avoid it.

@FlorianBoe Can you confirm if it behave the same for you? Maybe you find a solution. Did not yet debug it who maybe avoid the `Enter` click event.

luca-rath:
> @luca-rath could you test this out if still all cases which you had in mind in #5567 still works. Also we should make sure that it behaves the same in all major browsers (Chrome, Firefox, Safari, Edge).

@alexander-schranz @FlorianBoe Interaction using `Space` is still possible, it's indeed not necessary to add an event listener for that ... But since you've also removed the event listener for the `Enter` key, selecting values using the `Enter` key isn't possible anymore, therefore you should only remove the following case from the clause: `event.key === 'Space'`

PS: I have tested this in chromium, firefox and safari

alexander-schranz:
@luca-rath do you know why we need to implement Enter ourself as for native buttons enter works. It looks like some component is catching this key and prevent default behaviour and it would be best we could change that instead of reimplementing here browsers behaviours again.

niklasnatter:
Thanks a lot for working on this! I rebased the branch onto the 2.3 branch, I hope that is okay for you.

I also tested the functionality - the `Space` key works great, but as already recognized by @luca-rath, the `Enter` key does not work anymore. I tried to find the reason for this, but unfortunately, I was not able to 😕

FlorianBoe:
If there is no solution to the enter key problem, I can also bring back the `handleButtonKeyDown` function and just keep the little bugfix.

niklasnatter:
@FlorianBoe I tried to debug this again, but havent found a solution. I think the pragmatic thing to do would be adding the `handleButtonKeyDown` for the `Enter` key again, even if it is no 100% correct 🙈  Would you like to do that?
Thanks again for working on this!

alexander-schranz:
I readded the listener for enter, so the behaviour should be like expeced now. @FlorianBoe thx for working on this.

alexander-schranz (approved):

PR #6461, opened 2022-01-14
Remove wrapping label element to improve accessibility

FlorianBoe:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | 
| Related issues/PRs | #6444
| License | MIT
| Documentation PR | 

#### What's in this PR?

Some input elements in the admin interface have two associated label elements. Besides the visible label (with a `for="/xyz"`), input elements are also nested within a label element.

two-labels

#### Why?

This enclosing label is problematic because icons and buttons within the wrapper are added to input name. This problem is particularly problematic for assistive technologies such as screen readers.

#### To Do

- [ ] Update snapshots
- [ ] Update the type of the `labelRef`

I am not quite sure what is a good solution to adjust the typing accordingly. Suggestions are very welcome!


src/Sulu/Bundle/AdminBundle/Resources/js/components/Input/Input.js
```diff
@@ -121,7 +121,7 @@ export default class Input<T: ?string | ?number> extends React.PureComponent<Inp
 
         return (
             <Fragment>
-                <label
+                <div
```
niklasnatter, line 124: as far as i see, the `labelRef` prop of this component is only used by the `ColorPicker` and `SingleAutoComplete` to position their popover. so it would be okay in the core to change this as the other components dont need a `<label>` element.

but if we change this, we should rename `labelClass` and `labelRef`. maybe to something like `inputContainerClass`?

niklasnatter: regarding to the type of the ref - i think i would just use `ElementRef<*>` then 🙂

alexander-schranz: @niklasnatter I'm not completely sure about your comment. Can we remove the `labelRef`?  Or what is it used for?

alexander-schranz:
Reminder for myself we need to check the checkbox and the checkbox toggler type work still like expected.

src/Sulu/Bundle/AdminBundle/Resources/js/components/Input/Input.js
niklasnatter: the `labelRef` is used by the `SingleAutoComplete` to display the popover at the correct position. so we should rename it to something like `inputContainerRef` and use that in the `SingleAutoComplete` 🙂

niklasnatter:
@alexander-schranz Is this ready for review? 🙂

alexander-schranz:
@niklasnatter look like something is still failing in the tests. But if you want to have a look at it or take it over its fine for me :)

niklasnatter (approved):
Thanks for bringing this up!

PR #6464, opened 2022-01-17
Fix PHP 8 deprecation notice

eekes:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Fixes these deprecation errors in PHP 8 (see https://wiki.php.net/rfc/stable_sorting):

```
Deprecated: usort(): Returning bool from comparison function is deprecated, return an integer less than, equal to, or greater than zero
```


alexander-schranz:
@eekes Thank you!

PR #6465, opened 2022-01-18
Remove unused unlocalized extensionsData from page trash item

niklasnatter:
I think this was a copy-paste error in the initial implementation 😇


alexander-schranz (approved):
Make sense from my point of view only localized extensionData need to be set and thats don on L116

PR #6466, opened 2022-01-20
Fix default operator for Tags and categories filters for smart content

matthieu2607:
| Q | A
| --- | ---
| Bug fix? |  yes
| New feature? | no 
| BC breaks? | no 
| Deprecations? | no 
| License | MIT

#### What's in this PR?

Add a default operator for tags and categories filters on smart content.

### DIFF
--- a/src/Sulu/Component/SmartContent/Orm/DataProviderRepositoryTrait.php
+++ b/src/Sulu/Component/SmartContent/Orm/DataProviderRepositoryTrait.php
@@ -136,7 +136,7 @@ trait DataProviderRepositoryTrait
                     $queryBuilder,
                     $tagRelation,
                     $filters['tags'],
-                    \strtolower($filters['tagOperator']),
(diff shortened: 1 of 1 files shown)


alexander-schranz:
@matthieu2607 Thank you for the pull request. How does this happen? Would be nice to have a test case also for this if we are able to reproduce this in a simple way.

The pull request should also be rebased to 2.3 branch as it is a bugfix e.g.:

```
git remote add upstream git@github.com:sulu/sulu.git

git checkout your-branch
git rebase upstream/2.x --onto upstream/2.3

```

niklasnatter:
Hey, thanks for your work! 
Can you describe in which case the error is happening for you so we can test the change?

matthieu2607:
Hi @alexander-schranz and @niklasnatter,

Sorry for my late response.
I can no longer reproduce the problem.

I close the PR

PR #6470, opened 2022-01-24
Fix permission check when clearing cache without permissions for all webspaces

jonland:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | 
| Related issues/PRs | 
| License | MIT

#### What's in this PR?

We have 2 webspaces WSA and WSB. The user has the live permission for WSA, but NOT for WSB. Now the user clicks in sulu admin "clear webspace cache" for the webspace WSA, the request is denied with http code 403. The reason is the guard at CacheController::clearAction. The first guard is fine, the user has access to to webspace WSA. Now the check goes to the elseif, but here it must also be checked if the webspace key is given, only with this combination the check makes sense.

#### Why?

Without this fix a user with which has only access to one webspace, can't clear his webspace cache.

### DIFF
--- a/src/Sulu/Bundle/WebsiteBundle/Controller/CacheController.php
+++ b/src/Sulu/Bundle/WebsiteBundle/Controller/CacheController.php
@@ -84,7 +84,7 @@ class CacheController
     private function checkLivePermissionForAllWebspaces()
     {
         foreach ($this->webspaceManager->getWebspaceCollection() as $webspace) {
-            $context = PageAdmin::SECURITY_CONTEXT_PREFIX . $webspace->getKey();
+            $context = PageAdmin::getPageSecurityContext($webspace->getKey());
             if (!$this->securityChecker->hasPermission($context, PermissionTypes::LIVE)) {
                 return false;
             }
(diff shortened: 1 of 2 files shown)


wachterjohannes:
@jonland the intention behind this check was:

1. If you pass a webspaceKey you want to clear the cache for this webspace - So you should have permissions for this webspace
2. If you pass no webspaceKey you will clear all the caches for all webspaces (that is what the code does then) - So you should have permissions for all webspaces

The Admin of sulu does always pass a webspace (from 2.4). Do you have a custom view which calls that api?

/cc @alexander-schranz

jonland:
@wachterjohannes there is as typo my comment.

"fix guard if **NO** webspace key is given" => correct "fix guard if webspace key is given"

As I wrote the user has access to only 1 webspace. If the user is in WSA and has the "live"-permission for WSA, now he pushes the button "Clear website cache" the in sulu admin, the request is denied with 403.

The first guard is fine. This is true && false => false.
`if ($webspaceKey && !$this->checkLivePermissionForWebspace($webspaceKey)) {`

The second guard is wrong, because it's not checked that `$webspaceKey` is given. It checks the permissions for all webspaces. The user has only permission for WSA and not WSB, so the check is true and 403 is returned.
`} elseif (!$this->checkLivePermissionForAllWebspaces()) {`

Please just try this issue with 2 webspaces and 1 "live"-permisson. You will see.

wachterjohannes:
@jonland ou yeah i see 🙈 you are totally right! we should create a test for that! at least a Unit Test. Are you able to add them?

niklasnatter (approved):
Thanks a lot for fixing this!

PR #6474, opened 2022-01-25
Dispatch PageTranslationRestoredEvent when restoring single translation only

niklasnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Related issues/PRs | #6337
| License | MIT

#### What's in this PR?

This PR adjusts the `PageTrashItemHandler` service to dispatch a `PageTranslationRestoredEvent` instead of a `PageRestoredEvent` if a single translation is restored.

#### Why?

This allows us to display a more intuitive message in the activity log. Also, we already dispatch specific events when managing translations in other places (eg. `PageTranslationCopiedEvent` and `PageTranslationRemovedEvent`).

### DIFF
--- a/src/Sulu/Bundle/PageBundle/Domain/Event/PageTranslationRestoredEvent.php
+++ b/src/Sulu/Bundle/PageBundle/Domain/Event/PageTranslationRestoredEvent.php
@@ -1,0 +1,107 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of Sulu.
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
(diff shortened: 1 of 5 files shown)


wachterjohannes (changes requested):
@niklasnatter is there nowhere a test which does a restore? If it would be one we should add a check there if the correct event is thrown

niklasnatter:
@wachterjohannes No, unfortunately the activites of the bundle are not tested anywhere at the moment. Should I add a test?

wachterjohannes:
@niklasnatter yes please

wachterjohannes:
@niklasnatter the tests are failing 🙈

niklasnatter:
@wachterjohannes Yes unfortunately, but the failing tests are not caused by the changes in this PR. The same tests are failing on the target branch: https://github.com/sulu/sulu/actions

wachterjohannes (approved):

PR #6477, opened 2022-02-01
Remove constructor declaration from PreviewLinkInterface

luca-rath:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

Remove constructor declaration from PreviewLinkInterface

#### Why?

Because otherwise this leads to a `Declaration incompatible` error with doctrine proxies

### DIFF
--- a/src/Sulu/Bundle/PreviewBundle/Domain/Model/PreviewLink.php
+++ b/src/Sulu/Bundle/PreviewBundle/Domain/Model/PreviewLink.php
@@ -53,6 +53,15 @@ class PreviewLink implements PreviewLinkInterface
      */
     private $lastVisit;
 
+    public function __construct(string $token, string $resourceKey, string $resourceId, string $locale, array $options)
+    {
+        $this->token = $token;
+        $this->resourceKey = $resourceKey;
+        $this->resourceId = $resourceId;
+        $this->locale = $locale;
+        $this->options = $options;
+    }
+
     public function getId(): int
     {
         return $this->id;
@@ -106,14 +115,12 @@ class PreviewLink implements PreviewLinkInterface
      */
     public static function create(string $token, string $resourceKey, string $resourceId, string $locale, array $options): PreviewLinkInterface
     {
-        $previewLink = new self();
-
-        $previewLink->token = $token;
-        $previewLink->resourceKey = $resourceKey;
-        $previewLink->resourceId = $resourceId;
-        $previewLink->locale = $locale;
-        $previewLink->options = $options;
-
-        return $previewLink;
+        return new self(
+            $token,
+            $resourceKey,
+            $resourceId,
+            $locale,
+            $options
+        );
     }
 }
--- a/src/Sulu/Bundle/PreviewBundle/Domain/Model/PreviewLink.php
+++ b/src/Sulu/Bundle/PreviewBundle/Domain/Model/PreviewLink.php
@@ -53,15 +53,6 @@ class PreviewLink implements PreviewLinkInterface
      */
     private $lastVisit;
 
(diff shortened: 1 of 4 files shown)


src/Sulu/Bundle/PreviewBundle/Domain/Model/PreviewLink.php
```diff
@@ -53,15 +53,6 @@ class PreviewLink implements PreviewLinkInterface
      */
     private $lastVisit;
```
alexander-schranz: Would still keep the constructor this way:

```suggestion

    public function __construct(string $token, string $resourceKey, string $resourceId, string $locale, array $options)
    {
        $this->token = $token;
        $this->resourceKey = $resourceKey;
        $this->resourceId = $resourceId;
        $this->locale = $locale;
        $this->options = $options;
    }

```

src/Sulu/Bundle/PreviewBundle/Domain/Model/PreviewLink.php
```diff
@@ -109,4 +100,20 @@ public function getLastVisit(): ?\DateTimeImmutable
...
+    {
+        $previewLink = new self();
+
+        $previewLink->token = $token;
+        $previewLink->resourceKey = $resourceKey;
+        $previewLink->resourceId = $resourceId;
+        $previewLink->locale = $locale;
+        $previewLink->options = $options;
+
+        return $previewLink;
```
alexander-schranz:
```suggestion
        return new self(
            $token,
            $resourceKey,
            $resourceId,
            $locale,
            $options
        );
```

alexander-schranz (changes requested):

src/Sulu/Bundle/PreviewBundle/Infrastructure/Doctrine/Repository/PreviewLinkRepository.php
```diff
@@ -39,10 +39,10 @@ public function create(string $resourceKey, string $resourceId, string $locale,
     {
         $token = $this->generateToken();
 
-        /** @var class-string<PreviewLinkInterface> $className */
+        /** @var class-string<PreviewLinkInterface>|PreviewLinkInterface $className */
```
alexander-schranz: This is a false indicator as it says its an object and not a string.

I could also not find why it is required when trying to report here something back to phpstan: https://phpstan.org/r/9095f567-476d-4834-951e-a6fe666a7584

And phpstan seems correctly understand class-string of an interface.

luca-rath:
```suggestion
        /** @var class-string<PreviewLinkInterface>$className */
```

src/Sulu/Bundle/PreviewBundle/Infrastructure/Doctrine/Repository/PreviewLinkRepository.php
```diff
@@ -39,10 +39,10 @@ public function create(string $resourceKey, string $resourceId, string $locale,
     {
         $token = $this->generateToken();
 
-        /** @var class-string<PreviewLinkInterface> $className */
+        /** @var class-string<PreviewLinkInterface>$className */
```
luca-rath:
```suggestion
        /** @var class-string<PreviewLinkInterface> $className */
```

alexander-schranz (approved):

PR #6493, opened 2022-03-02
Fix phpstan baseline, eslint config and caching tests

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? |  yes
| New feature? | no 
| BC breaks? | no 
| Deprecations? | no  
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Fix phpstan baseline.

#### Why?

Fix baseline.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -143,7 +143,11 @@
         "react/no-deprecated": "error",
         "react/jsx-curly-brace-presence": [
             "error",
-            "never"
+            {
+                "props": "never",
+                "children": "never",
+                "propElementValues": "always"
+            }
         ],
         "react/jsx-curly-spacing": [
             "error",
--- a/src/Sulu/Bundle/AudienceTargetingBundle/Tests/Functional/CachingTest.php
+++ b/src/Sulu/Bundle/AudienceTargetingBundle/Tests/Functional/CachingTest.php
@@ -38,6 +38,7 @@ class CachingTest extends SuluTestCase
         $cacheKernel = new AppCache(self::bootKernel());
         $cookieJar = new CookieJar();
         $client = new KernelBrowser($cacheKernel, [], null, $cookieJar);
+        $client->disableReboot();
 
         $client->request('PURGE', '/');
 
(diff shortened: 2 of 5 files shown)


.eslintrc.json
```diff
@@ -143,7 +143,11 @@
         "react/no-deprecated": "error",
         "react/jsx-curly-brace-presence": [
             "error",
-            "never"
+            {
+                "props": "never",
+                "children": "never",
+                "propElementValues": "always"
```
alexander-schranz, line 149: This is a new but it is as stated in the following issue / comment https://github.com/yannickcr/eslint-plugin-react/issues/3228#issuecomment-1054601164 not supported by all parser so we need to define it to always so our parsers do not have problems with it.

src/Sulu/Bundle/AudienceTargetingBundle/Tests/Functional/CachingTest.php
```diff
@@ -38,6 +38,7 @@ public function testFirstRequestIsACacheMiss()
         $cacheKernel = new AppCache(self::bootKernel());
         $cookieJar = new CookieJar();
         $client = new KernelBrowser($cacheKernel, [], null, $cookieJar);
+        $client->disableReboot();
```
alexander-schranz, line 41: this is now required else other tests are running into an issue about a not booted kernel

src/Sulu/Bundle/MediaBundle/Tests/Functional/Media/Storage/S3StorageTest.php
```diff
@@ -21,6 +21,11 @@ class S3StorageTest extends SuluTestCase
 {
     protected static $class = S3Kernel::class;
 
+    protected static function getKernelClass(): string
+    {
+        return S3Kernel::class;
+    }
```
alexander-schranz, line 30: required because of: https://github.com/symfony/symfony/pull/45414

We still need to define `protected static $class = S3Kernel::class;` for prefer lowest tests.

chirimoya (approved):

PR #6494, opened 2022-03-02
Fix page permissions when there is no array

### LINKED ISSUE #6482: Can't copy workspace if the pages within don't have document-level permissions set
| Q | A
| --- | ---
| Sulu Version | 2.3
| PHP Version | 8.0.15
| DB Version | Ver 15.1 Distrib 10.4.20-MariaDB
| Browser Version | Microsoft Edge 98.0.1108.43

# Actual Behavior

Upon executing the `sulu:webspaces:copy` command, the following exception is thrown:
```
Sulu\Bundle\PageBundle\Document\BasePageDocument::setPermissions(): Argument #1 ($permissions) must be of type array, null given, called in /var/www/vendor/sulu/sulu/src/Sulu/Bundle/PageBundle/Command/WebspaceCopyCommand.php on line 322
```

# Expected Behavior

The `sulu:webspaces:copy` command to execute succesfully without any exceptions.

# Steps to Reproduce

For now, investigation has resulted in the following hypothesis:
* This only occurs when the webspace has any page without set permissions.

# Possible Solutions

See the following slack thread for discussion about this bug:
https://sulu-io.slack.com/archives/C0430GGCV/p1644162170094529

As suggested, initializing the `$permissions` property in `PageBundle\Document\BasePageDocument` with an empty array should solve this issue.

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? |  yes
| New feature? |  no
| BC breaks? | no 
| Deprecations? | no yes 
| Fixed tickets | fixes #6482 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Fix page permissions when there is no array

#### Why?

There was a report here #6482 and also in the slack channel that this happens. That permission is null instead of an array. This change will initialize the BasePageDocument permissions always with an array to avoid that problem and will in restore avoid problem when permissions are there already null as it would be to hard to create a migration there.

### DIFF
--- a/src/Sulu/Bundle/PageBundle/Document/BasePageDocument.php
+++ b/src/Sulu/Bundle/PageBundle/Document/BasePageDocument.php
@@ -240,7 +240,7 @@ class BasePageDocument implements
      *
      * @var array
      */
-    protected $permissions;
+    protected $permissions = [];
 
     /**
      * List of versions.
--- a/src/Sulu/Bundle/PageBundle/Tests/Functional/Controller/PageControllerTest.php
+++ b/src/Sulu/Bundle/PageBundle/Tests/Functional/Controller/PageControllerTest.php
@@ -880,7 +880,6 @@ class PageControllerTest extends SuluTestCase
             ],
         ])[0];
 
-        $this->client->disableReboot();
         $this->client->jsonRequest('DELETE', '/api/pages/' . $page1['id'] . '?webspace=sulu_io&language=en');
 
         $response = $this->client->getResponse();
--- a/src/Sulu/Bundle/PageBundle/Tests/Functional/Controller/PageControllerTest.php
+++ b/src/Sulu/Bundle/PageBundle/Tests/Functional/Controller/PageControllerTest.php
@@ -69,6 +69,7 @@ class PageControllerTest extends SuluTestCase
     public function setUp(): void
     {
         $this->client = $this->createAuthenticatedClient();
+        $this->client->disableReboot(); // see https://github.com/symfony/symfony/issues/45580
         $this->purgeDatabase();
         $this->initPhpcr();
         $this->session = $this->getContainer()->get('sulu_document_manager.default_session');
--- a/src/Sulu/Bundle/PageBundle/Tests/Functional/Controller/PageControllerTest.php
+++ b/src/Sulu/Bundle/PageBundle/Tests/Functional/Controller/PageControllerTest.php
@@ -880,6 +880,7 @@ class PageControllerTest extends SuluTestCase
             ],
         ])[0];
 
+        $this->client->disableReboot();
         $this->client->jsonRequest('DELETE', '/api/pages/' . $page1['id'] . '?webspace=sulu_io&language=en');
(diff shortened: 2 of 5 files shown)


src/Sulu/Bundle/PageBundle/Document/BasePageDocument.php
```diff
@@ -240,7 +240,7 @@ class BasePageDocument implements
      *
      * @var array
      */
-    protected $permissions;
+    protected $permissions = [];
```
alexander-schranz, line 243: this make sure that also when no `setPermissions` is called the permissions are as defined an array and not null.

src/Sulu/Bundle/PageBundle/Trash/PageTrashItemHandler.php
```diff
@@ -183,7 +183,7 @@ public function restore(TrashItemInterface $trashItem, array $restoreFormData =
             $localizedPage->setStructureType($localeData['structureType']);
             $localizedPage->getStructure()->bind($localeData['structureData']);
             $localizedPage->setExtensionsData($localeData['extensionsData']);
-            $localizedPage->setPermissions($localeData['permissions']);
+            $localizedPage->setPermissions($localeData['permissions'] ?: []);
```
alexander-schranz, line 186: this will fix restore for elements which were stored with permissions = null. We currently have no way to reproduce how it happens but it seems like old sulu which did upgrade are sometimes running into this case as it was reported by two different sulu users.

chirimoya (approved):

src/Sulu/Bundle/PageBundle/Tests/Functional/Controller/PageControllerTest.php
```diff
@@ -69,6 +69,7 @@ class PageControllerTest extends SuluTestCase
     public function setUp(): void
     {
         $this->client = $this->createAuthenticatedClient();
+        $this->client->disableReboot(); // see https://github.com/symfony/symfony/issues/45580
```
alexander-schranz, line 72: this change is unrelated to the change in this PR. as the test already fails on 2.4 branch: https://github.com/sulu/sulu/commit/9561f75d81d89c8ced7733578525977457e63d2f

alexander-schranz: we already needed to add some `disableReboot` on 2.3 branch to fix this kind of issues.

src/Sulu/Bundle/TestBundle/Testing/TestUserProvider.php
```diff
@@ -160,4 +161,9 @@ public function supportsClass($class)
     {
         return $this->userProvider->supportsClass($class);
     }
+
+    public function reset(): void
+    {
+        $this->user = null;
+    }
```
alexander-schranz, line 168: Symfony now resets the kernel between requests. As this means the entity manager unit of work is being resetd we also need to reset the user here that it is loaded again into Unit of work next time it is requested.

PR #6503, opened 2022-03-09
Add cache control max age and immutable header

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no 
| BC breaks? | no 
| Deprecations? | no  
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Add the `max-age` and `immutable` flag for the media image stream action.

#### Why?

Currently image directly responded by php are not cached. If we only set `Cache-Control: public, max-age`, the browser is still doing a request as there was no `last-modified` header or something similar provided, so it waits for 304 request. After some research the way to force the browser to cache the resource is the `immutable` flag: https://developer.mozilla.org/de/docs/Web/HTTP/Headers/Cache-Control

This already is in use on our own website https://sulu.io. As we are versioning all our assets with `?v=1-0` and so have a unique identifier. It make sense to use the immutable cache control header to tell the browser not even do a lookup for a 304 response.

### DIFF
--- a/public/.htaccess
+++ b/public/.htaccess
@@ -3,7 +3,7 @@
 # mod_rewrite). Additionally, this reduces the matching process for the
 # start page (path "/") because otherwise Apache will apply the rewriting rules
 # to each configured DirectoryIndex file (e.g. index.php, index.html, index.pl).
-DirectoryIndex website.php
+DirectoryIndex index.php
 
 # By default, Apache does not evaluate symbolic links if you did not enable this
 # feature in your server configuration. Uncomment the following line if you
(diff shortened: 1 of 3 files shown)


public/.htaccess
```diff
@@ -3,16 +3,16 @@
 # mod_rewrite). Additionally, this reduces the matching process for the
 # start page (path "/") because otherwise Apache will apply the rewriting rules
 # to each configured DirectoryIndex file (e.g. index.php, index.html, index.pl).
-DirectoryIndex website.php
```
alexander-schranz, line 6: this is just an update of the `.htaccess` from the skeleton version: https://github.com/sulu/skeleton/blob/2.4.1/public/.htaccess

wachterjohannes:
@alexander-schranz tests failing

alexander-schranz:
@wachterjohannes fixed!

PR #6506, opened 2022-03-14
Add qrcode field type

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | none
| Related issues/PRs | none
| License | MIT
| Documentation PR | none

#### What's in this PR?

This PR integrates a QR-Code library into the form.

#### Why?

Can be used to generate QR-Codes in the form.


alexander-schranz (approved):
Code looks fine. Should we add a content type for it? And add some docs for our content type documentation. 

I will test it with the current project to check the design and how it looks like in different formats / grid sizes.

src/Sulu/Bundle/AdminBundle/Resources/js/package.json
```diff
@@ -39,6 +39,7 @@
         "react-datetime": "^3.0.0",
         "react-dropzone": "^11.2.0",
         "react-portal": "^4.1.0",
+        "react-qr-code": "^2.0.3",
```
alexander-schranz, line 42: I checked some alternatives, but this did look for me for the best match currently as it concentrates on the qr code usecase and seems good maintained.

src/Sulu/Bundle/AdminBundle/Resources/js/components/QRCode/qrcode.scss
```diff
@@ -0,0 +1,5 @@
+@import '../../containers/Application/colors.scss';
+
+.qrcode {
+    margin-top: 20px;
```
alexander-schranz, line 6: Should have a define `width` and a `max-width: 100%`.
It should also have the `background: $white` being set and have some `padding` white padding around this. Some qr code scanner requires a white border around the qr code.

PR #6508, opened 2022-03-14
Improved admin table cell text overflow handling

### LINKED ISSUE #320: Form data display bug
| Q | A
| --- | ---
| Sulu Version | 2.4.0
| Form Bundle Version | 2.2.1
| PHP Version | 8.1.0
| DB Version | MySQL 5.7
| Browser Version | Chrome 97.0.4692.99
| OS Version | macOS 12.2

# Actual Behavior

Longer texts are cutted in form data list view while text is there, see screenshot.

form-data

# Expected Behavior

Form data is fully readable in list view, without exporting, downloading and opening it in an app like Excel or Numbers. 

# Steps to Reproduce

Create a form with a textarea and add longer text to it. View form data in Sulu Admin of this submitted text.

# Possible Solutions

I would create a PR but I don't know which solution you guys prefer. 
IMHO there are multiple possibilities:

- display the text truncated with overflow ellipsis and add a title attrib to the text so text could be fully read in the Tooltip when hovering it
- display the full text so table rows will grow in its height
- Add a detail view or open action in front of each row. (Like the edit button on other lists)

ymc-thzi:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes https://github.com/sulu/SuluFormBundle/issues/320
| License | MIT

#### What's in this PR?

Improved handling of text overflows in admin table cells including a tooltip on hovering (added a title attribute). See screenshot of the solution:
Bildschirmfoto 2022-03-14 um 14 38 31 (3)

#### Why?

Longer texts are truncated in admin table list cells. Also, the css ellipsis is not displayed correctly. See problem screenshot:
152322124-ac1f5b05-16c3-4fad-baa4-66ec4e194608

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Table/Cell.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Table/Cell.js
@@ -46,7 +46,7 @@ export default class Cell extends React.PureComponent<Props> {
                 className={cellClass}
                 colSpan={colSpan}
             >
-                <div className={tableStyles.cellContent} style={style}>
+                <div className={tableStyles.cellContent} style={style} title={children[1]}>
                     {children}
                 </div>
             </td>
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Table/Cell.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Table/Cell.js
@@ -46,7 +46,7 @@ export default class Cell extends React.PureComponent<Props> {
                 className={cellClass}
                 colSpan={colSpan}
             >
-                <div className={tableStyles.cellContent} style={style} title={children[1]}>
+                <div className={tableStyles.cellContent} style={style}>
                     {children}
                 </div>
             </td>
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Table/table.scss
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Table/table.scss
@@ -96,7 +96,10 @@ $borderRadius: 3px;
 }
 
 .cell-content {
-    display: flex;
+    display: inline;
+    text-overflow: ellipsis;
+    overflow: hidden;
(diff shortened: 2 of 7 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/components/Table/Cell.js
```diff
@@ -46,7 +46,7 @@ export default class Cell extends React.PureComponent<Props> {
                 className={cellClass}
                 colSpan={colSpan}
             >
-                <div className={tableStyles.cellContent} style={style}>
+                <div className={tableStyles.cellContent} style={style} title={children[1]}>
```
niklasnatter: as this is the general `Cell` component of the table, the `children` property could includes things like a button or a select. i think the better place to do this would be the `StringFieldTransformer`. 

the `StringFieldTransformer` is called with the string value and returns a html node for it. i would apply the styling to the returned node like the `ColorFieldTransformer.js` already does 🙂

niklasnatter:
Thanks a lot for the PR. I think this would be a nice UX improvement, but would prefer to move the styling to the `StringFieldTransformer` like described in https://github.com/sulu/sulu/pull/6508#discussion_r907284275. I am happy to review and test it if anybody finds the time to do that!

src/Sulu/Bundle/AdminBundle/Resources/js/components/Table/Cell.js
ymc-thzi: Great, thanks. Will do it with `StringFieldTransformer` that makes way more sense 😃

ymc-thzi:
@niklasnatter I Have reverted the first changes and pushed the title attrib and overflow css to StringFieldTransformer as you have mentioned. Can be reviewed.

niklasnatter (approved):
Thank you! 🙂

PR #6520, opened 2022-04-01
Fix phpstan and php-cs-fixer issues

luca-rath:


luca-rath:
@alexander-schranz should we maybe disable the `single_line_comment_spacing` rule, because of the comments for optional function arguments?

alexander-schranz:
@luca-rath I would disable it for all cases to don't update that all code.

luca-rath:
@alexander-schranz fixed

PR #6526, opened 2022-04-06
Catch invalid path exception from phpcr in load by resource-locator

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no yes
| New feature? | no yes
| BC breaks? | no yes
| Deprecations? | no yes 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

This PR fixes a Problem when passing an invalid path to the `oadByResourceLocator` function of the `PHPCRMapper`

#### Why?

The exception passes and has to be handled in the follow up classes which should not be needed.

### DIFF
--- a/src/Sulu/Component/Content/Types/ResourceLocator/Mapper/PhpcrMapper.php
+++ b/src/Sulu/Component/Content/Types/ResourceLocator/Mapper/PhpcrMapper.php
@@ -15,6 +15,7 @@ use DateTime;
 use PHPCR\ItemExistsException;
 use PHPCR\NodeInterface;
 use PHPCR\PathNotFoundException;
+use PHPCR\RepositoryException;
 use Sulu\Bundle\DocumentManagerBundle\Bridge\DocumentInspector;
 use Sulu\Component\Content\Document\Behavior\ResourceSegmentBehavior;
 use Sulu\Component\Content\Exception\ResourceLocatorAlreadyExistsException;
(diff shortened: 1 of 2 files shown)


src/Sulu/Component/Content/Types/ResourceLocator/Mapper/PhpcrMapper.php
```diff
@@ -257,6 +258,8 @@ public function loadByResourceLocator($resourceLocator, $webspaceKey, $languageC
             }
         } catch (PathNotFoundException $e) {
             throw new ResourceLocatorNotFoundException(\sprintf('Path "%s" not found', $path), null, $e);
+        } catch (RepositoryException $e) {
```
alexander-schranz: I would not catch the repository exception here.

I think it would be better to check `$path` is a validabsolutepath via the PathHelper so we even do not hit the database here:

https://github.com/phpcr/phpcr-utils/blob/6b5db3321d6f9e7ceaf3880131a089e8e2505cfc/src/PHPCR/Util/PathHelper.php#L47

alexander-schranz (changes requested):
Check for valid path before trying to load the data.

alexander-schranz (approved):

alexander-schranz (commented):
@sabinebaer I remember a project that you did overwrite the path generation. Could you test if this change have any effects on that project. For chinese letter it seems to work: https://3v4l.org/A3KcO even a emoticon seems to work: https://3v4l.org/SFlOC

sabinebaer:
@alexander-schranz Exactly - we have overwritten path generation to allow Chineses characters. I did a quick test with your changes - it still works fine, so this change does not seem to have any impact.

alexander-schranz:
@sabinebaer thx for testing!

PR #6528, opened 2022-04-07
display the target page's title even if enableTitle is false

### LINKED ISSUE #6525: Link element // No title displayed in admin preview.
| Q | A
| --- | ---
| Sulu Version | 2.4.1
| PHP Version | 8.1
| DB Version | MariaDB 10.7.3
| Browser Version | all

# Actual Behavior

Given an element of content type "link" is used without any parameters or the parameter "enable_title" is set to false. 
When an editor selects a target of type "page" no title is displayed in the element's preview.

link

# Expected Behavior

Since this makes it hard for editors to see that a target is set, the target page's title should be displayed in the element's preview.

# Steps to Reproduce

1. Insert a content element with the following configuration to your page.

```

<property name="internal_link" type="link">
    <meta>
        <title lang="en">Link</title>
    </meta>
</property>
```

or

```
<property name="internal_link" type="link" >
    <meta>
       <title lang="en">Link</title>
    </meta>
    <params>
        <param name="enable_title" value="false"/>
    </params>
</property>
```

2. Select any internal page as the link target

# Possible Solutions

see expected behavior.

DavidFeller:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #6525 
| License | MIT

#### What's in this PR?

This PR makes sure that a content element of type "link" always displays the target's title in its admin preview even if the parameter "enableTitle" (which is only used to set a custom title) is false or undefined.

#### Why?
see #6525

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Link/Link.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Link/Link.js
@@ -115,7 +115,7 @@ class Link extends Component<Props> {
                 target: enableTarget ? target : undefined,
                 anchor: enableAnchor ? anchor : undefined,
                 href,
-                title: enableTitle || title.length ? title : undefined,
+                title: enableTitle || (title != undefined && title.length) ? title : undefined,
                 locale: toJS(locale),
             }
         );
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Link/Link.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Link/Link.js
@@ -115,7 +115,7 @@ class Link extends Component<Props> {
                 target: enableTarget ? target : undefined,
                 anchor: enableAnchor ? anchor : undefined,
                 href,
-                title: enableTitle ? title : undefined,
+                title: enableTitle || title.length ? title : undefined,
                 locale: toJS(locale),
             }
         );


src/Sulu/Bundle/AdminBundle/Resources/js/containers/Link/Link.js
```diff
@@ -115,7 +115,7 @@ class Link extends Component<Props> {
                 target: enableTarget ? target : undefined,
                 anchor: enableAnchor ? anchor : undefined,
                 href,
-                title: enableTitle ? title : undefined,
+                title: enableTitle || (title != undefined && title.length) ? title : undefined,
```
alexander-schranz, line 118: I think the logic before here is correctly the title should never be saved here when enableTitle is set to false (onChange is saving it). I think the problem is that it need to hold the title here internal something like:

```js
this.overlayTitle = title;
```

because if we are calling the title with `onChange` and later change the title of the page it will not show the new title of that page then.

alexander-schranz:
I did have a look at this but looks for me like there is currently no way to get the `title` by `href` in the frontend.
The current change will save the title, but that is not what we want in case of `enableTitle: false`, as we want always get the up2date title of the selected item.
Looking at it I think we would need an endpoint something like `/admin/api/link-items/{provider}-{href}` which returns us the `LinkItem` which we can use to display the title but also the published state here.

/cc @niklasnatter Do you see another way then providing an endpoint to get to the correct title?

niklasnatter:
Implementing an additional API endpoint for this is a bis of a hassle because we would also need to extend the `LinkProviderInterface` 😕 

But I think it would be possible to implement this in the frontend only. The Overlay component of the link type (`LinkTypeOverlay`/`MediaLinkTypeOverlay`) already needs to know how to load the title for the respective type. We could extract this functionality into a static function on the Overlay component class or extract `LinkTitleProvider`that is registered in the `linkTypeRegistry` alongside the Overlay component.

alexander-schranz:
@wachterjohannes as I know already works on it. As they had additional issue in their project.

niklasnatter:
Thanks for the pull request! We try to solve this from another angle in https://github.com/sulu/sulu/pull/6648. Hopefully we can merge that one soon 🙂

PR #6540, opened 2022-04-14
Fix cache clear when cache tags are disabled

Quehnie:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Now the CacheClearer also check if tags are enabled not only if they are supported.

#### Why?

The user was not able to clear the cache in the admin UI.

### DIFF
--- a/src/Sulu/Bundle/WebsiteBundle/Cache/CacheClearer.php
+++ b/src/Sulu/Bundle/WebsiteBundle/Cache/CacheClearer.php
@@ -65,7 +65,7 @@ class CacheClearer implements CacheClearerInterface
         EventDispatcherInterface $eventDispatcher,
         string $varDir,
         ?CacheManager $cacheManager,
-        bool $tagsEnabled
+        bool $tagsEnabled = true
     ) {
         $this->kernelEnvironment = $kernelEnvironment;
         $this->filesystem = $filesystem;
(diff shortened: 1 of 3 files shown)


src/Sulu/Bundle/WebsiteBundle/Cache/CacheClearer.php
```diff
@@ -53,20 +53,27 @@ class CacheClearer implements CacheClearerInterface
...
+
     public function __construct(
         Filesystem $filesystem,
         $kernelEnvironment,
         RequestStack $requestStack,
         EventDispatcherInterface $eventDispatcher,
         string $varDir,
-        ?CacheManager $cacheManager
+        ?CacheManager $cacheManager,
+        bool $tagsEnabled
```
alexander-schranz: add a default value here to avoid a bc break:

```suggestion
        bool $tagsEnabled = true
```

src/Sulu/Bundle/WebsiteBundle/Cache/CacheClearer.php
```diff
@@ -80,7 +87,10 @@ public function clear(/*?array $tags = null*/)
 
         $tags = \func_num_args() >= 1 ? \func_get_arg(0) : null;
 
-        if (null !== $tags && $this->cacheManager && $this->cacheManager->supportsTags()) {
+        dump($this->tagsEnabled);
+        die;
```
alexander-schranz: 😱

src/Sulu/Bundle/WebsiteBundle/Cache/CacheClearer.php
```diff
@@ -80,7 +87,10 @@ public function clear(/*?array $tags = null*/)
...
+
+        if (null !== $tags && $this->cacheManager && $this->cacheManager->supportsTags() && $this->tagsEnabled) {
```
alexander-schranz:
```suggestion
        if (null !== $tags && $this->tagsEnabled && $this->cacheManager && $this->cacheManager->supportsTags()) {
```

would check the variable before checking the service. so it avoids the service call.

alexander-schranz (changes requested):

alexander-schranz (approved):

alexander-schranz (approved):

PR #6558, opened 2022-04-26
show tag names and category names in contact list

jonland:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | 
| Related issues/PRs | 
| License | MIT

#### What's in this PR?

It's possible to filter tags or categories in the contact list, but the admin doesn't see which contact has which tags or categories. As admin it's faster to see assigned categories in the list and find where categories are missing.

#### Why?

If the admin wants to see e.g. which categories are assigned to contact he must open the detail view of the contact. With this PR the admin can display the column "Categories" and sees the assigned categories directly in the list.

These changes can also be done for the accounts list.

#### To Do

- [ ] Add changes to UPGRADE.md

### DIFF
--- a/src/Sulu/Bundle/ContactBundle/Resources/config/lists/contacts.xml
+++ b/src/Sulu/Bundle/ContactBundle/Resources/config/lists/contacts.xml
@@ -296,8 +296,8 @@
         </property>
 
         <group-concat-property
-            name="tagIds"
-            visibility="never"
+            name="tagNames"
+            visibility="yes"
             translation="sulu_tag.tags"
             glue=","
         >
@@ -351,6 +351,29 @@
                 </params>
             </filter>
         </property>
+
+        <group-concat-property
+            name="categoryNames"
+            visibility="yes"
+            translation="sulu_category.categories"
+            glue=","
+            distinct="true"
+        >
+            <field-name>translation</field-name>
+            <entity-name>SuluCategoryBundle:CategoryTranslation</entity-name>
+
+            <joins>
+                <join>
+                    <entity-name>SuluCategoryBundle:Category</entity-name>
+                    <field-name>%sulu.model.contact.class%.categories</field-name>
+                </join>
+                <join>
+                    <entity-name>SuluCategoryBundle:CategoryTranslation</entity-name>
+                    <field-name>SuluCategoryBundle:Category.translations</field-name>
+                    <condition>SuluCategoryBundle:CategoryTranslation.locale = '%default_locale%'</condition>
+                </join>
+            </joins>
+        </group-concat-property>
 
         <property name="user" visibility="no" searchability="yes" translation="sulu_security.user_name">
             <field-name>username</field-name>
--- a/src/Sulu/Bundle/ContactBundle/Resources/config/lists/contacts.xml
+++ b/src/Sulu/Bundle/ContactBundle/Resources/config/lists/contacts.xml
@@ -300,8 +300,9 @@
             visibility="yes"
             translation="sulu_tag.tags"
             glue=","
+            distinct="true"
         >
-            <field-name>id</field-name>
+            <field-name>name</field-name>
             <entity-name>SuluTagBundle:Tag</entity-name>
 
             <joins ref="tags"/>
--- a/src/Sulu/Bundle/ContactBundle/Resources/config/lists/contacts.xml
+++ b/src/Sulu/Bundle/ContactBundle/Resources/config/lists/contacts.xml
@@ -296,8 +296,20 @@
         </property>
 
         <group-concat-property
+            name="tagIds"
+            visibility="never"
+            translation="sulu_tag.tags"
+            glue=","
+        >
+            <field-name>id</field-name>
+            <entity-name>SuluTagBundle:Tag</entity-name>
+
+            <joins ref="tags"/>
+        </group-concat-property>
+
+        <group-concat-property
             name="tagNames"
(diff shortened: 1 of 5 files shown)


src/Sulu/Bundle/ContactBundle/Resources/config/lists/contacts.xml
```diff
@@ -296,12 +296,13 @@
         </property>
 
         <group-concat-property
-            name="tagIds"
-            visibility="never"
```
alexander-schranz: This tagIds should stay as the were else it would be a backward compatibility break.

Create an additional tagNames group concat property instead.

src/Sulu/Bundle/ContactBundle/Resources/config/lists/contacts.xml
```diff
@@ -296,12 +296,13 @@
...
+            name="tagNames"
+            visibility="yes"
```
alexander-schranz: should be not visible by default should be set to `no` and the user can then enable it when they use it.

src/Sulu/Bundle/ContactBundle/Resources/config/lists/contacts.xml
```diff
@@ -296,12 +296,13 @@
...
             translation="sulu_tag.tags"
             glue=","
+            distinct="true"
```
alexander-schranz, line 315: is this really required? I think distinct should not be needed here, but not sure 🤔

src/Sulu/Bundle/ContactBundle/Resources/config/lists/contacts.xml
```diff
@@ -352,6 +353,29 @@
             </filter>
         </property>
 
+        <group-concat-property
+            name="categoryNames"
+            visibility="yes"
```
alexander-schranz: should be `no` like the tagNames

src/Sulu/Bundle/ContactBundle/Resources/config/lists/contacts.xml
```diff
@@ -352,6 +353,29 @@
...
+            translation="sulu_category.categories"
+            glue=","
+            distinct="true"
```
alexander-schranz, line 373: same as above this should be is this distinct really requires?

alexander-schranz (changes requested):
Thank you for the pull request. Sounds good to add the category and tag names to the list. They should still be optional and not enabled by default so the visiblity should be `no` so a user which need them can enable them via the list settings.

As we are adding category and tag names to the contact we should do the same for the `accounts` and `media` list view to be consistent here. Can you add it also to there `xml` definitions?

jonland:
- tagIds are restored
- tagNames and categoryNames are set to visibility="no"
- tagNames and categoryNames are added to account list
- distinct="true" is necessary to show the names only once

alexander-schranz:
Thx for the quick update. Seems like this is not working as the `default_locale` seems to be set by the endpoint. Can you  have a look at failing test cases? PHP Lint and Windows can be ingored at current state.

alexander-schranz:
Hello @jonland,
do you have some time to finish this? Would be great to get this also into 2.5 release :)

jonland:
@alexander-schranz
I don't know how do you run your tests or how to get them working! Where can I find a description or tutorial?

alexander-schranz:
@jonland You can run the phpunit file in each bundle if you want. But you can also use our test wrapper for it, you need to start first a running database you can use the docker compose file in sulu/sulu also for it:

```bash
docker compose up
```

The tests running parameters are configured in `src/Sulu/Bundle/TestBundle/Resources/app/config/parameters.yml` which will be created on the first run via `[parameters.yml.dist](https://github.com/sulu/sulu/blob/2.5/src/Sulu/Bundle/TestBundle/Resources/app/config/parameters.yml.dist)`

To run all tests you can go with:

```bash
bin/runtests --initialize --all 
```

While initialize is only required for first run. To run a specific bundle you can use:

```bash
bin/runtests -t CategoryBundle 
```

And via following you can add phpunit flags:

```bash
bin/runtests -t CategoryBundle --flags="--filter=SomeTestCase"
```

But if the database is created you can also just run phpunit directly in the bundles, but as phpunit is not in the bundle path you need use .. to get path to phpunit:

```bash
cd src/Sulu/Bundle/CategoryBundle/
../../../../vendor/bin/phpunit
```

src/Sulu/Bundle/TestBundle/Resources/app/config/parameters.yml.dist
```diff
@@ -9,6 +9,7 @@ parameters:
     phpcr.password: '%env(PHPCR_PASSWORD)%'
     phpcr.workspace: '%env(PHPCR_WORKSPACE)%'
 
+    default_locale: en
```
alexander-schranz: No this should not be come from here. I see now what is wrong is that the category should use the default_locale from the category entity.

See as example the categories.xml join: https://github.com/sulu/sulu/blob/027b90c68353cebe6e267415ad1e0e11d9dee117/src/Sulu/Bundle/CategoryBundle/Resources/config/lists/categories.xml#L16

So replace the `%default_locale%` with getting the `.default_locale` property from the main category entity.

```suggestion
```

alexander-schranz:
Categories and Tags are now shown in contact and account list when activated by #6998.
Thx for your effort here!

Issue #6556, opened 2022-04-25
Add Support for Symfony 6


alexander-schranz:
# Problem description

Add Support for Symfony 6.

# Proposed solution

Upgrade dependencies and sulu core to be compatible to Symfony 5.4 / 6.0.

 - #6515
 
# External Dependencies Symfony 6 Support

The following packages need to be updated first before symfony 6 compatibility can be added:

 - [x] [massive/search-bundle](https://github.com/massiveart/MassiveSearchBundle)
   - [x] https://github.com/massiveart/MassiveSearchBundle/pull/167
 - [x] [massive/build-bundle](https://github.com/massiveart/MassiveBuildBundle)
   - [x] https://github.com/massiveart/MassiveBuildBundle/pull/23
 - [x] [handcraftedinthealps/rest-routing-bundle](https://github.com/handcraftedinthealps/RestRoutingBundle)
   - [x] https://github.com/handcraftedinthealps/RestRoutingBundle/pull/18
   - [x] https://github.com/FriendsOfSymfony/FOSRestBundle/pull/2368
 - [x] [jackalope/jackalope](https://github.com/jackalope/jackalope)
    - [x] https://github.com/jackalope/jackalope/pull/369
 - [x] [jackalope/jackalope/jackalope-doctrine-dbal](https://github.com/jackalope/jackalope-doctrine-dbal)
 - [x] [jackalope/jackalope-jackrabbit](https://github.com/jackalope/jackalope-jackrabbit)
    - [x] https://github.com/jackalope/jackalope-jackrabbit/pull/168
    - [x] https://github.com/jackalope/jackalope-jackrabbit/pull/169
 - [x] [phpcr/phpcr-shell](https://github.com/phpcr/phpcr-shell)
   - [x] https://github.com/jackalope/jackalope/pull/372
   - [x] https://github.com/phpcr/phpcr-utils/pull/207
   - [x] https://github.com/jackalope/jackalope-jackrabbit/pull/170 
   - [x] https://github.com/phpcr/phpcr-shell/pull/210
 - [x] [friendsofsymfony/http-cache-bundle](https://github.com/FriendsOfSymfony/FOSHttpCacheBundle)
    - [x] https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/pull/573
 - [x] [phpcr/phpcr-migrations](https://github.com/phpcr/phpcr-migrations)
    - [x] https://github.com/phpcr/phpcr-migrations/pull/14
[...]

PR #6571, opened 2022-04-28
Replace deprecated self::$container with self::getContainer() in tests

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? | no 
| BC breaks? | no 
| Deprecations? | no  
| Fixed tickets | fixes # 
| Related issues/PRs | #6556 
| License | MIT

#### What's in this PR?

Replace deprecated self:: with self::getContainer() in tests

#### Why?

The `self::$container` was removed and the container is now provided over symfony `getContainer` method.


src/Sulu/Bundle/TestBundle/Testing/ContainerTrait.php
```diff
@@ -1,31 +0,0 @@
...
-use Symfony\Component\DependencyInjection\ContainerInterface;
-
-trait ContainerTrait
-{
-    /**
-     * @internal The "getContainer" function is deprecated use self::$container variable instead.
-     *
-     * This function exist to keep compatibility to old tests.
-     */
-    protected static function getContainer(): ContainerInterface
```
alexander-schranz, line 23: This conflicts with the symfony getContainer method and so need to be removed. Symfony ships the same implementation now https://github.com/symfony/symfony/blob/d11e41be6cb23016894c564d5c2b9ef3357ecaa7/src/Symfony/Bundle/FrameworkBundle/Test/KernelTestCase.php#L100

PR #6591, opened 2022-05-04
Relay deletion errors to snackbar

R3dLantern:
Error messages as a result of deletion failure are not relayed to the snackbar in the media overview. This additional prop definition ensures that.

| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| License | MIT

#### What's in this PR?

Adds the error handler callback to the List rendering definition of the MediaCollection container.

#### Why?

When delete operations in the Media overview fail, no error message can be relayed to the snackbar.

#### Example Usage

Media controllers can now supply error messages as usual:

```php
return $this->handleView($this->view(['detail' => 'Media cannot be deleted'], Response::HTTP_INTERNAL_SERVER_ERROR));
```


niklasnatter (approved):
Good catch - thanks for the PR! 🙂

Issue #6436, opened 2021-12-22
Copy & Paste for Blocks


mavuio:
hi !

i'd like to have a copy&paste feature in the block-editor, and am willing to contribute such a feature to
sulu.

i have started to implement a basic working version of that feature in my own fork and want to discuss the details of how
the feature could work.

TOC:
- A) feature-description
- B) UX-design
- C) code implementation details

### A) the basic feature would be:

**copy:** to enable editors to select one or many blocks from a single List (top or nested "SortableBlockList") and copy them
to a clipboard attached to the editor's session

**paste:** the  block(s) on the clipboard can then be inserted into any other Block-List ("SortableBlockList") which the editor has access to.

blocks  can only be pasted if all of them could be created with a simple "add block" click at the target -location as well
(taking valid block-types and list-limits into account)

#### B1) **action-button:**

i would propose a dedicated kind of "action"-button for each block, which reveals a context menu with a list of available actions for that block:

![clipping-2021_12_22-10_25_12](https://user-images.githubusercontent.com/1092088/147069323-ff0b43d2-3a45-4ac3-a615-956b62783c03.png)

the placement and design of this needs to be discussed (i placed it at the bottom right of the block for now), also the popup-menu and its icons and wording.

#### B.2) **multi-select:**
 when clicking "multi-select" i'd suggest to start a multi-select mode:

we'd need to enable checkboxes for each block to be selected, and some buttons

- select all
- select none
- copy selected to clipboard
- move selected to clipboard
- cancel 

similar how that looks like in my other-php-CMS:
![clipping-2021_12_22-10_31_01](https://user-images.githubusercontent.com/1092088/147070394-f45eb409-1ea6-495f-b211-4309a2cda5fa.png)

or in my other-elixir-CMS:
![clipping-2021_12_22-10_35_02](https://user-images.githubusercontent.com/1092088/147070803-60b482dc-7d3c-4d5c-883b-a4f4a2616d7e.png)

for sulu-UX i'd suggest we just turn each action-button within the activated list into a checkbox if multi-select mode
ist started and add a somehow highlighted button-area on top.

as you see from my examples i also included some none copy-paste features in the multi-select mode, which may be of interest:

- delete selected
- hide selected
- unhide selected

#### B.3) paste:

pasting is also triggered  with the action-button e.g. like here:
[...]

PR #6599, opened 2022-05-09
Replace remove block button with dropdown that can display multiple actions

niklasnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | yes
| Related issues/PRs | https://github.com/sulu/sulu/issues/6436
| License | MIT

#### What's in this PR?

This PR replaces the remove block button that is displayed in the upper right corner of blocks with a dropdown that allows to display multiple possible actions. 

#### Why?

This is a preparation for implementing a copy feature for blocks like described in #6436


src/Sulu/Bundle/AdminBundle/Resources/js/components/Block/Block.js
```diff
@@ -49,13 +70,13 @@ export default class Block<T: string> extends React.Component<Props<T>> {
...
     render() {
         const {
+            actions,
             activeType,
             children,
             dragHandle,
             icons,
             onCollapse,
             onExpand,
-            onRemove,
```
niklasnatter, line 58: TODO: describe how to upgrade a custom component that used this property in the `UPGRADE.md`

src/Sulu/Bundle/AdminBundle/Resources/js/components/Block/actionPopover.scss
```diff
@@ -0,0 +1,24 @@
+@import 'sulu-admin-bundle/containers/Application/colors.scss';
+
+$actionTextColor: $black;
+$actionTextColorActive: $shakespeare;
+
+.icon {
+    margin-right: 15px;
+}
+
+.action {
```
niklasnatter: this matches the implementation of the `download media` popover and the `autocomplete suggestions` popover

src/Sulu/Bundle/AdminBundle/Resources/js/components/Block/Block.js
alexander-schranz: Should we just migrate it to actions if it is defined to avoid a bc break here.

Do we know if somebody did use the Block component standalone yet? /cc @sulu/core-team

niklasnatter: yeah we could do that 🙂 if we do that, we need to implement the logic in multiple components (`Block`, `SortableBlock`, `SortableBlockList`)

.eslintrc.json
```diff
@@ -203,7 +203,6 @@
                 "indentLogicalExpressions": true
             }
         ],
-        "react/jsx-no-bind": "error",
```
alexander-schranz: The rule sounds for me something important in case of performance: https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md

So not sure if I would remove it. I think I would go with keep it and add ignore where really required.

niklasnatter: i removed the rule because it makes list rendering a pain as you need to extract a component for the list item. in this case, i think the rule makes the code even slower because we need to instantiate a whole component for each item instead of just an arrow function 🙈 

but yeah, i guess there are cases where it improves performance too. i will readd it.

alexander-schranz: > we need to instantiate a whole component for each item instead of just an arrow function

I see, but theoretically the component is initialized only once. And the function is created for every render so it also can in this case improve the performance. I'm just not sure how much mobx maybe would effect here things and maybe avoid rerender here something. 
Atleast we also created in this cases also for [Select Option](https://github.com/sulu/sulu/blob/2.5/src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Option.js) own component to avoid this kind of problems. So I personally would go the save way of introducing also here a `MenuOption` or what the best name is here.

src/Sulu/Bundle/AdminBundle/Resources/js/components/Block/ActionPopover.js
```diff
@@ -0,0 +1,65 @@
...
+                        menuRef={setPopoverRef}
+                        style={popoverStyle}
+                    >
+                        {this.props.actions.map((action, index) => {
+                            if (action.type === 'divider') {
+                                return <Menu.Divider key={index} />;
+                            }
+
+                            return (
+                                <li key={index}>
```
alexander-schranz: To be consistent I would create a MenuOption and or MenuItem component or what best matching name is. So we also get rid of the `react/jsx-no-bind` ignore here.

src/Sulu/Bundle/AdminBundle/Resources/js/components/Block/ActionPopover.js
```diff
@@ -0,0 +1,65 @@
...
+                                    <button
+                                        className={actionPopoverStyles.action}
+                                        /* eslint-disable-next-line react/jsx-no-bind */
+                                        onClick={() => {
+                                            action.onClick();
+                                            this.props.onClose();
+                                        }}
+                                    >
```
alexander-schranz: a button should always define its `type` else it accidently gets a submit button when wrapped inside a form tag.

```suggestion
                                        }}
                                        type='button'
                                    >
```

alexander-schranz: We could add https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/button-has-type.md to avoid this in future.

src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/BlockCollection.js
```diff
@@ -154,18 +155,43 @@ class BlockCollection<T: string, U: {type: T}> extends React.Component<Props<T,
         onChange(newValue);
     };
 
-    hasMaximumReached() {
```
alexander-schranz, line 157: Not sure about htis rename. Think `hasMaximumReached` present it better as maxOccursReached but was never a fan of the `occurs` word 🙈

src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/SortableBlock.js
```diff
@@ -39,14 +85,6 @@ class SortableBlock<T: string, U: {type: T}> extends React.Component<Props<T, U>
         }
     };
 
-    handleRemove = () => {
```
alexander-schranz, line 42: do we need to keep this as methods are public in js? or can we safely remove it?

src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/BlockCollection.js
```diff
@@ -154,18 +155,43 @@ class BlockCollection<T: string, U: {type: T}> extends React.Component<Props<T,
...
+    @computed get maxOccursReached() {
         const {maxOccurs, value} = this.props;
 
         return !!maxOccurs && value.length >= maxOccurs;
     }
 
-    hasMinimumReached() {
```
alexander-schranz, line 163: Do we need to keep bc breaks here in mind?

alexander-schranz (changes requested):

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/tests/__snapshots__/FieldBlocks.test.js.snap
```diff
@@ -154,14 +154,14 @@ exports[`Render block with schema 1`] = `
             class="iconButtons"
           >
             <span
-              aria-label="su-trash-alt"
-              class="su-trash-alt clickable"
+              aria-label="su-more-circle"
```
alexander-schranz: Any easy way to make this better accessible?

E.g.:

```html
<ul class="iconButtons"
   <li>
       <button ...>
           <span class="su-more-circle" aria-hidden="true"/>
           <span class="sr-only">More</span>
       </button>
   </li>
   <li>
       <button ...>
           <span class="su-collapse-vertical" aria-hidden="true"/>
           <span class="sr-only">Collapse Block</span>
       </button>
   </li>
</ul>
```

src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/SortableBlock.js
niklasnatter: it is unusual to call a method of a component (you would need to get the ref) and react recommends to use composition over inheritance, so i would say its safe to remove it 🙂

src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/BlockCollection.js
niklasnatter: no, i would not build a bc layer here as it is unusual to call a method on a component class. but i will rename the methods back to the original name as you said you like the better 🙂

src/Sulu/Bundle/AdminBundle/Resources/js/components/Block/ActionPopover.js
niklasnatter: added the rule 🙂 i guess i will need to update a lot of snapshots

src/Sulu/Bundle/AdminBundle/Resources/js/components/Block/ActionPopover.js
niklasnatter: i extracted a `ActionPopoverItem`. but it really feels cumbersome to do this, i am not a fan of the `react/jsx-no-bind` rule 🙈

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/tests/__snapshots__/FieldBlocks.test.js.snap
niklasnatter: this is rendered using the `Icon` component:
https://github.com/sulu/sulu/blob/c298925a6815803b5f94ca80be3ffce1c939615f/src/Sulu/Bundle/AdminBundle/Resources/js/components/Block/Block.js#L100-L106

i can check if anything breaks if i use the `Button` component

alexander-schranz: I would not use the button component just us a button tag ;)

alexander-schranz (approved):

PR #6605, opened 2022-05-11
Fix account deletion by setting parent account references to null

### LINKED ISSUE #6600: Deleting multiple (interrelated) organizations fails
Deleting multiple (interrelated) organizations fails

| Q | A
| --- | ---
| Sulu Version | 2.4.2
| PHP Version | 8.1.3
| DB Version | MySQL 5.7.36
| Browser Version | Chrome 100.0.4896.127

# Actual Behavior

- Deleting multiple, interrelated organizations at the same time will fail with unexpected error (tested relation: "Parent Company")
- After failure, the list-view of organizations is not reloaded causing subsequent editorial errors (successfully deleted organizations are still visible, selection  of items in list is not updated)
- Not reproducible when deleting only one organization

# Expected Behavior

- Deleting interrelated organizations does not fail
- List view and selection are updated in case of an occuring error

# Steps to Reproduce

- Log into admin using a user that has sufficient rights to manage contacts/organizations
- Open account list view (/admin/#/accounts)
- Select multiple organizations (e.g. 10-20 items); Selection should contain organizations that have Parent Company relations (idAccountsParent) to other organizations within the selection
- Delete selected organizations via option "Delete" in the menu bar
- Sulu will trigger parallel deletion requests for every selected item in the background
- Deletion will fail with error message "Error - An unexpected error occured while deleting" in case that any of the deletion requests is failing; 2 types of errors will occur for the background requests:
1.) ```Doctrine\DBAL\Exception\DeadlockException An exception occurred while executing a query: SQLSTATE[40001]: Serialization failure: 1213 Deadlock found when trying to get lock; try restarting transaction```
2.) ```Doctrine\DBAL\Exception\ForeignKeyConstraintViolationException
An exception occurred while executing a query: SQLSTATE[23000]: Integrity constraint violation: 1451 Cannot delete or update a parent row: a foreign key constraint fails (`blk_sulu`.`co_accounts`, CONSTRAINT `FK_805CD14AC9171171` FOREIGN KEY (`idAccountsParent`) REFERENCES `co_accounts` (`id`))```
- List view is not updated and will show already deleted datasets leading to consequential errors when working with this list as an editor

# Possible Solutions

- ```sulu/src/Sulu/Bundle/ContactBundle/Resources/config/doctrine/Account.orm.xml```, line 69:
```on-delete="SET NULL"``` is missing; currently defaulting to ```on-delete="RESTRICT"``` in the database causing the failing foreign key contraints (successfully deleted datasets will leave a restricted reference)
- Scheduling or serial processing of deletion requests; Parallel deletion of related organizations will cause the observed deadlocks between datasets
- Reload list view of organizations in case that unexpected errors occur
- Possible quickfix for the time being: Disallow the deletion of multiple organizations at the same time

ymc-siwi:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #6600 (partially)
| Related issues/PRs | none
| License | MIT

#### What's in this PR?

Will set ```on-delete="SET NULL"``` for parent organization references in organizations; Current solution will default to ```on-delete=„PROTECT“```; This will cause failing foreign key constraints when deleting organizations with broken references - the references are protected and will not be deleted/cleaned up properly.

#### Why?

For detailed description see https://github.com/sulu/sulu/issues/6600

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,19 @@
 # Upgrade
 
+## 2.3.12
+
+### Add DELETE SET NULL to account parent relation
+
+There did exist a bug in sulu when trying to remove an account entity
+which was connected to another account entity it was not possible to
+remove it. The `DELETE SET NULL` on the parent connection will solve
+this issue. For this a database change is required:
+
+```sql
+ALTER TABLE co_accounts DROP FOREIGN KEY FK_805CD14AC9171171;
+ALTER TABLE co_accounts ADD CONSTRAINT FK_805CD14AC9171171 FOREIGN KEY (idAccountsParent) REFERENCES co_accounts (id) ON DELETE SET NULL
+```
+
 ## 2.3.7
 
 ### Add missing `kernel.reset` tag for document manager cache services
--- a/src/Sulu/Bundle/ContactBundle/Resources/config/doctrine/Account.orm.xml
+++ b/src/Sulu/Bundle/ContactBundle/Resources/config/doctrine/Account.orm.xml
@@ -58,7 +58,7 @@
 
         <many-to-one field="parent" target-entity="Sulu\Bundle\ContactBundle\Entity\AccountInterface" inversed-by="children">
             <join-columns>
-                <join-column name="idAccountsParent" referenced-column-name="id"/>
+                <join-column name="idAccountsParent" referenced-column-name="id" on-delete="SET NULL" nullable="true"/>
             </join-columns>
             <gedmo:tree-parent/>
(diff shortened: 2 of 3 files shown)


alexander-schranz (approved):

alexander-schranz:
@ymc-siwi Thank you!

ymc-siwi:
@alexander-schranz Cheers for the quick reaction and finalization :-)

PR #6606, opened 2022-05-11
Remove `send` for compatibility with phppm kernel

Deltachaos:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

Calling `send` is directly writing to stdout at a point in request/response flow execution where the outside component (e.g. the currently used kernel) is not expecting something to write to stdout.

If you run sulu using phppm or wrap the kernels `$app->handle()` into another application, it expects to get a response object returned, and have nothing written to stdout yet. Doing this might break the calling application as it does currently with phppm.

I also don't see the need to call `send` here, as it is called by the kernel later anyways.

### DIFF
--- a/src/Sulu/Component/Rest/Csv/CsvHandler.php
+++ b/src/Sulu/Component/Rest/Csv/CsvHandler.php
@@ -104,7 +104,6 @@ class CsvHandler
                 $exporter->export('php://output', $data);
             }
         );
-        $response->send();
 
         return $response;
     }
--- a/src/Sulu/Component/Rest/Tests/Unit/Csv/CsvHandlerTest.php
+++ b/src/Sulu/Component/Rest/Tests/Unit/Csv/CsvHandlerTest.php
@@ -65,6 +65,7 @@ class CsvHandlerTest extends TestCase
 
         \ob_start();
         $response = $handler->createResponse($viewHandler->reveal(), $view, $request->reveal(), $format);
+        $response->send();
         $content = \ob_get_contents();
         \ob_end_clean();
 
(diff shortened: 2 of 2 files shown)


alexander-schranz:
@Deltachaos Thx for the pull request. That is really strange that `send` is here called manually. Can you rebase the branch to target the `2.3` version here. So we can release this then as via a patch release?

Deltachaos:
@alexander-schranz i have updated the PR. Still streaming responses is somehow awkward in symfony if you are using phppm for example, but this is not relevant here.

I have seen that you have participated in this discussion as well, so maybe you could add your opinion on my approach for solving this in symfony here https://github.com/php-runtime/runtime/issues/115#issuecomment-1124370152

alexander-schranz:
The swoole php runtime implementation is currently handling StreamedResponses here: https://github.com/Ocramius/runtime/blob/5be78a2f76ce6d7e858d2e16a1fef4d1c7f4d4f2/src/swoole/src/SymfonyHttpBridge.php#L48-L57. It is just currently not supported by the roadrunner runtime https://github.com/php-runtime/runtime/issues/50. I'm not familiar with phppm not sure if there is maybe required to create its own runtime for it in the php runtime project: https://github.com/php-runtime/runtime which then handles streamedresponses correctly.

alexander-schranz (approved):

alexander-schranz:
@Deltachaos Thank you!

Issue #6613, opened 2022-05-17
SingleSelection field only changes its type when the name of the field updates


j-mahlke:
| Q | A
| --- | ---
| Sulu Version | 2.3.11
| PHP Version | 8.1
| Browser Version | Chrome 101

# Actual Behavior

When a field in a block changes its type from single_something_selection to single_other_selection without changing the name of the field, the overlay list shows the list of the last entity (in the example above, single_other_selection would list entities of type something)

# Expected Behavior

when the block type changes the type of a field without changing the name, it should also affect the type of entities in single-selection lists

# Steps to Reproduce

use this page-template:
note, that the name of the field is always "item", no matter what type of entity_selection
`
<?xml version="1.0" ?>
<template xmlns="http://schemas.sulu.io/template/template"
          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xsi:schemaLocation="http://schemas.sulu.io/template/template http://schemas.sulu.io/template/template-1.0.xsd">

    <key>default</key>

    <view>pages/default</view>
    <controller>Sulu\Bundle\WebsiteBundle\Controller\DefaultController::indexAction</controller>
    <cacheLifetime>604800</cacheLifetime>

    <meta>
        <title lang="en">Default</title>
    </meta>

    <properties>
        <property name="title" type="text_line" mandatory="true">
            <meta>
                <title lang="en">Title</title>
            </meta>
            <params>
                <param name="headline" value="true"/>
            </params>

            <tag name="sulu.rlp.part"/>
        </property>

        <property name="url" type="resource_locator" mandatory="true">
            <meta>
                <title lang="en">Resourcelocator</title>
            </meta>

            <tag name="sulu.rlp"/>
        </property>

        <block name="content" minOccurs="0">
            <meta>
                <title lang="de">Inhalt</title>
            </meta>

            <params>
                <param name="add_button_text">
                    <meta>
                        <title lang="en">Add feature</title>
                    </meta>
                </param>
            </params>

            <types>
                <type name="contact">
                    <meta>
                        <title lang="en">Contacts</title>
                    </meta>

                    <properties>
                        <property name="item" type="single_contact_selection">
                            <meta>
                                <title lang="en">Contacts</title>
                            </meta>
                        </property>
                    </properties>
                </type>
                <type name="snippet">
                    <meta>
                        <title lang="en">Snippet</title>
                    </meta>

                    <properties>
                        <property name="item" type="single_snippet_selection">
                            <meta>
[...]

PR #6615, opened 2022-05-19
Rerender form fields when type changes to prevent stale components

j-mahlke:
see issue #6613

PR #6620, opened 2022-05-23
Rerender form fields when type changes to prevent stale components

niklasnatter:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Fixed tickets | fixes https://github.com/sulu/sulu/issues/6613
| Related issues/PRs | replaces https://github.com/sulu/sulu/pull/6615
| License | MIT

#### What's in this PR?

The PR adjusts the `Rerender` component to rerender a form field if its type is changed.

#### Why?

Because we might end up with stale components if we do not do this. For example, at the moment, when switching between block types with selection properties, the selection property is not rerendered and therefore the type of the selection overlay does not match the new block typy (see #6613).

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/Renderer.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/Renderer.js
@@ -67,7 +67,7 @@ class Renderer extends React.Component<Props> {
                 dataPath={itemDataPath}
                 error={error}
                 formInspector={formInspector}
-                key={schemaKey}
+                key={schemaKey + '_' + schemaField.type}
                 name={schemaKey}
                 onChange={onChange}
                 onFinish={this.handleFieldFinish}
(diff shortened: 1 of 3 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/Renderer.js
```diff
@@ -67,7 +67,7 @@ class Renderer extends React.Component<Props> {
                 dataPath={itemDataPath}
                 error={error}
                 formInspector={formInspector}
-                key={schemaKey}
+                key={schemaKey + '_' + schemaField.type}
```
niklasnatter, line 70: I moved the key into the `Renderer` component because we had a key here already. I successfully tested the changes with the template provided in https://github.com/sulu/sulu/issues/6613#issue-1238789465

Issue #6627, opened 2022-05-25
Deleting referenced tag fails


ymc-siwi:
Deleting referenced tag fails

| Q | A
| --- | ---
| Sulu Version | 2.4.2
| PHP Version | 8.1.3
| DB Version | MySQL 5.7.36
| Browser Version | Chrome 100.0.4896.127

# Actual Behavior

- If tags are used in contacts (reference in table co_contact_tags) or accounts (reference in table co_account_tags), they cannot be deleted via backend interface (Settings -> Tags -> delete referenced tag)
- Deletion fails with an unspecific error message

# Expected Behavior

- Deleting a tag will also delete related references to contacts and accounts
- Deleting a contact will also delete related references to tags (entries in table co_contact_tags)
- Deleting an account will also delete related references to tags (entries in table co_account_tags)

# Steps to Reproduce

- Log into admin using a user that has sufficient rights to manage tags, contacts and organizations
- Open tag list view (/admin/#/tags)
- Add and save new tag
- Edit a contact, add the new tag, save contact
- Open tag list view (/admin/#/tags) again, try to delete tag
- Deletion will fail with error message ```An exception occurred while executing a query: SQLSTATE[23000]: Integrity constraint violation: 1451 Cannot delete or update a parent row: a foreign key constraint fails (`blk_sulu`.`co_contact_tags`, CONSTRAINT `FK_4CB525501C41CAB8` FOREIGN KEY (`idTags`) REFERENCES `ta_tags` (`id`))```
- The same will happen for organizations (table and index name in error message will of course vary).

# Possible Solutions

- Workaround for the time being: delete all references in contacts and accounts manually before deleting a tag
- Enable cascading between tags (table ta_tags) and tables co_contact_tags/co_account_tags
- Pull request currently in preparation.

alexander-schranz:
@ymc-siwi Thx for creating this issue. Looks like here is a on-delete cascade missing:

https://github.com/sulu/sulu/blob/6069d05f8efae4ad89c428adf4113dc326a0bf74/src/Sulu/Bundle/ContactBundle/Resources/config/doctrine/Contact.orm.xml#L182-L194

This should be added to all many-to-many relations on contacts and accounts seems like it is missed in other cases also.

As an example here it is correct:

[...]

PR #6628, opened 2022-05-25
Allow reference cascading when deleting tags

ymc-siwi:
See https://github.com/sulu/sulu/issues/6627

| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #6672
| Related issues/PRs | none
| License | MIT

#### What's in this PR?

Will set on-delete="CASCADE" for tag references to accounts and contacts

#### Why?

Previous solution prevents deletion of referenced tags via backend; relations are not cleaned up properly if accounts or organizations are deleted.

For detailed description see https://github.com/sulu/sulu/issues/6627

#### To Do

- [ ] Create testcase
- [x] Add documentation for database change to UPGRADE.md


alexander-schranz:
Thank you for the pull request can you update also the other many to many relations so we don't run there in the same problems:

- https://github.com/sulu/sulu/blob/08863981d9e607dd1432db68eec2dac710732f74/src/Sulu/Bundle/WebsiteBundle/Resources/config/doctrine/Analytics.orm.xml#L26-L33
- https://github.com/sulu/sulu/blob/08863981d9e607dd1432db68eec2dac710732f74/src/Sulu/Bundle/SecurityBundle/Resources/config/doctrine/Group.orm.xml#L32-L41
- https://github.com/sulu/sulu/blob/08863981d9e607dd1432db68eec2dac710732f74/src/Sulu/Bundle/ContactBundle/Resources/config/doctrine/Contact.orm.xml#L86-L210
- https://github.com/sulu/sulu/blob/08863981d9e607dd1432db68eec2dac710732f74/src/Sulu/Bundle/ContactBundle/Resources/config/doctrine/Account.orm.xml#L108-L234


For category keywords I'm not sure if needed but think it is also required there:

- https://github.com/sulu/sulu/blob/08863981d9e607dd1432db68eec2dac710732f74/src/Sulu/Bundle/CategoryBundle/Resources/config/doctrine/CategoryTranslation.orm.xml#L29-L30

ymc-siwi:
@alexander-schranz Yes, makes sense - already done for se_group_roles. Would the following change be equal / favorable to `on-delete="CASCADE"` if a cascading tag is already defined for the regarding many-to-many field?
`<cascade> 
     <cascade-persist/> 
     <cascade-delete/> 
</cascade>`
Not sure if this has a functional impact or performance impact though. UnitOfWork should be the same as for cascade-persist...

alexander-schranz:
The `on-delete="CASCADE"` has no impact on the PHP Performance as it is database only. And currently it would ended in a 500 error when you tried to remove a related entity and now it will just be savely removed.

The `<cascade>` itself I would currently not change in any entities as that would require a deeper discussion where it make sense and where not.

alexander-schranz:
The failing Lint task you can currently ignore as it is not related to your Pull Request.

ymc-siwi:
@alexander-schranz All done except category keywords. There is no blocking behavior when deleting categories/keywords/translations - all FK-relations already cascade or are set to null properly (no relation is defaulting to ON DELETE RESTRICT here). The change would only make sense for cleanup, but this could be addressed in another Pull Request.

alexander-schranz:
@ymc-siwi Thank you, great work 👍

PR #6630, opened 2022-05-30
Implement block action for copying and cutting a single block

### LINKED ISSUE #6436: Copy & Paste for Blocks

niklasnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no yes
| Deprecations? | no
| Fixed tickets | partly #6436
| License | MIT

#### What's in this PR?

This pull request adds a `copy` and a `cut` action for blocks. The implementation builds upon the HTML `localStorage` to allow for copying blocks from one browser window into another browser window.


src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/BlockCollection.js
```diff
@@ -104,6 +123,31 @@ class BlockCollection<T: string, U: {type: T}> extends React.Component<Props<T,
...
 
+    @action handlePasteBlocks = (insertionIndex: number) => {
+        const {onChange, value} = this.props;
+
+        if (this.hasMaximumReached) {
+            throw new Error('The maximum amount of blocks has already been reached!');
+        }
+
+        if (value) {
+            // TODO: gracefully handle clipboard data this is incompatible with available block types
```
niklasnatter: i have not tested this at all yet

src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/BlockCollection.js
```diff
@@ -37,14 +41,27 @@ class BlockCollection<T: string, U: {type: T}> extends React.Component<Props<T,
...
         super(props);
 
-        this.fillArrays();
-        reaction(() => this.props.value.length, this.fillArrays);
+        this.fillArraysDisposer = reaction(() => this.props.value.length, this.fillArrays, {fireImmediately: true});
+        this.setPasteableBlocksDisposer = clipboardStore.observe(BLOCKS_CLIPBOARD_KEY, action((blocks) => {
+            this.pasteableBlocks = (blocks: any) || [];
+        }));
+
+        this.pasteableBlocks = (clipboardStore.get(BLOCKS_CLIPBOARD_KEY): any) || [];
```
alexander-schranz: Why having a `get` here? I would in `observe` call the callback or was there any problem with `action` inside a constructor?

src/Sulu/Bundle/AdminBundle/Resources/js/stores/clipboardStore/clipboardStore.js
```diff
@@ -0,0 +1,66 @@
...
+
+    notifyObservers(key: string, value: mixed): void {
+        const observers = this.observers[key] || [];
+
+        for (const observer of observers) {
+            observer(value);
+        }
+    }
+
+    get(key: string): mixed {
```
alexander-schranz: I would like to avoid the `get` as it would make it possible to use then without a observer which should not be the case.

alexander-schranz (changes requested):
Did not yet have a look at all.

src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/BlockCollection.js
niklasnatter: i want to get the current value initially because it might have been copied into the clipboard before the component was mounted and the observer was registered

niklasnatter: we could change the semantics of observe to fire once when the observer is registered or add a option like `fireImmediately` of mobx 🙂

alexander-schranz: I would expect that the code of observe does something like this:

```diff
    observe(key: string, observer: Observer) {
        if (!this.observers[key]) {
            this.observers[key] = [];
        }
        this.observers[key].push(observer);
        this.updateStorageEventListener();
        
+        const serializedValue = window.localstorage.getItem(key);
+        if (serializedValue) {
+           observer(serializedValue ? JSON.parse(serializedValue) : undefined);
+        }

        // return disposer function that allows to remove the registered observer
        return () => {
            const index = this.observers[key]?.indexOf(observer);
            if (index > -1) {
                this.observers[key].splice(index, 1);
            }
            this.updateStorageEventListener();
        };
    }
```

To understand it correctly if we just do that mobx errors currently, because it doesn't like that a action is called inside constructor?

niklasnatter: no, i dont think it would cause an error. but it is inconsistent to the `observe` method of mobx. therefore i would prefer to add a `invokeImmediately` argument like mobx does on its `observe` method 🙂 
https://mobx.js.org/intercept-and-observe.html#observe

src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/BlockCollection.js
```diff
@@ -37,14 +40,26 @@ class BlockCollection<T: string, U: {type: T}> extends React.Component<Props<T,
...
+    fillArraysDisposer: ?() => *;
+    setPasteableBlocksDisposer: ?() => *;
+
     constructor(props: Props<T, U>) {
         super(props);
 
-        this.fillArrays();
-        reaction(() => this.props.value.length, this.fillArrays);
+        this.fillArraysDisposer = reaction(() => this.props.value.length, this.fillArrays, {fireImmediately: true});
+        this.fillArraysDisposer = reaction(() => this.props.value.length, this.fillArrays, {fireImmediately: true});
```
alexander-schranz: Is this expected to be duplicated?

alexander-schranz:
```suggestion
        this.fillArraysDisposer = reaction(() => this.props.value.length, this.fillArrays, {fireImmediately: true});
```

One can be removed?

alexander-schranz (approved):

PR #6651, opened 2022-06-15
Fix logout response for symfony 6

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? |  yes
| New feature? | no 
| BC breaks? | no 
| Deprecations? | no  
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Fix logout response for symfony 6.

#### Why?

A response is always set so we need to check if the request url starts with a admin url so we know we are in the admin context.

### DIFF
--- a/src/Sulu/Bundle/SecurityBundle/EventListener/LogoutEventSubscriber.php
+++ b/src/Sulu/Bundle/SecurityBundle/EventListener/LogoutEventSubscriber.php
@@ -36,14 +36,19 @@ final class LogoutEventSubscriber implements EventSubscriberInterface
 
     public function onLogout(LogoutEvent $logoutEvent): void
     {
-        if (null !== $logoutEvent->getResponse()) {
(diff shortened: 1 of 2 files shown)


src/Sulu/Bundle/SecurityBundle/EventListener/LogoutEventSubscriber.php
```diff
@@ -36,14 +36,19 @@ public function __construct(UrlGeneratorInterface $urlGenerator)
 
     public function onLogout(LogoutEvent $logoutEvent): void
     {
-        if (null !== $logoutEvent->getResponse()) {
```
alexander-schranz, line 39: Is always set. So to avoid conflicts with community bundle or similar we need to check if we are in the admin context. To avoid context based service this is done via a the pathInfo of the request.

PR #6660, opened 2022-06-22
Add excluded types to link field type

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | none
| Related issues/PRs | none
| License | MIT
| Documentation PR | https://github.com/sulu/sulu-docs/pull/719

#### What's in this PR?

This PR add the possibility to filter link types by excluded types.

#### Why?

Add possibility to filter types by excluded types.

#### Example Usage

```xml
            <meta>
                <title>app.link</title>
            </meta>
            <params>
                <param name="enable_title" value="true"/>
                <param name="excluded_types" type="collection">
                    <param name="media"/>
                </param>
            </params>
        </property>
```

#### To Do

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Link/Link.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Link/Link.js
@@ -204,13 +204,10 @@ class Link extends Component<Props> {
             }
         );
 
-        const allowedTypes = linkTypeRegistry.getKeys().filter((key) => {
-            if (types !== undefined && types.length > 0) {
-                return types.includes(key);
-            }
-
-            return !excludedTypes.includes(key);
-        });
+        let allowedTypes = linkTypeRegistry.getKeys().filter((key) => !excludedTypes.includes(key));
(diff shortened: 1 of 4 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/containers/Link/Link.js
```diff
@@ -143,11 +146,11 @@ class Link extends Component<Props> {
         );
 
         const allowedTypes = linkTypeRegistry.getKeys().filter((key) => {
-            if (types === undefined || types.length === 0) {
-                return true;
+            if (types !== undefined && types.length > 0) {
+                return types.includes(key);
```
niklasnatter: i think i would expect that both, `types` and `excludedTypes` are considered if both are set.
so if both are set, we should only display the types that are inside of `types` but not inside of `excludedTypes`.

niklasnatter (approved):

PR #6666, opened 2022-06-27
Disable collections without add permissions in overlay

### LINKED ISSUE #6662: Media upload from root collection provokes permission error
| Q | A
| --- | ---
| Sulu Version | 2.4.3
| PHP Version | 8.1.7
| DB Version | MariaDB 10.4.12
| Browser Version | Google Chrome 102.0.5005.115 (64-bit)

# Actual Behavior

When in Media overview ("root collection"): You can trigger the upload process and then proceed to select a collection you are not permitted to reach for the new upload. The result is a HTTP 403 response, however the error message is generic and does not provide any information on what exactly went wrong.

# Expected Behavior

It should not be possible to select a restricted collection for a new Media upload. The best case would be that restricted collections are not up for selection in the overlay column list.

# Steps to Reproduce

1. Create a collection and enable custom permissions for it.
2. Log in as a user that is not permitted to access the new collection, and go to the Media overview.
3. Initiate an upload process. You should see the collection you have no access to (albeit it with the "shield" icon).
4. Select the collection as the upload target and confirm.

# Possible Solutions

* When fetching the collection list via API, the system should already consider filtering out restricted collections, albeit it with the caveat that child collections without restriction would fall victim to such filtering.
* The UI could be visually and funcationally enhanced by preventing selection of restricted collections and indicate the restriction, e.g. via icons or distinct coloring.

niklasnatter:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Fixed tickets | fixes #6662
| License | MIT

#### What's in this PR?

Disable collections without add permissions in overlay that is displayed when uploading media in root collection.

#### Why?

See https://github.com/sulu/sulu/issues/6662

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/Resources/js/containers/MultiMediaDropzone/MultiMediaDropzone.js
+++ b/src/Sulu/Bundle/MediaBundle/Resources/js/containers/MultiMediaDropzone/MultiMediaDropzone.js
@@ -178,7 +178,7 @@ class MultiMediaDropzone extends React.Component<Props> {
                 <SingleListOverlay
                     adapter="column_list"
                     clearSelectionOnClose={true}
-                    itemDisabledCondition="!!locked"
+                    itemDisabledCondition="!!locked || (_permissions && !_permissions.add)"
                     listKey={COLLECTIONS_RESOURCE_KEY}
                     locale={locale}
                     onClose={this.handleSelectCollectionOverlayClose}


src/Sulu/Bundle/MediaBundle/Resources/js/containers/MultiMediaDropzone/MultiMediaDropzone.js
```diff
@@ -178,7 +178,7 @@ class MultiMediaDropzone extends React.Component<Props> {
                 <SingleListOverlay
                     adapter="column_list"
                     clearSelectionOnClose={true}
-                    itemDisabledCondition="!!locked"
+                    itemDisabledCondition="!!locked || (_permissions && !_permissions.add)"
```
niklasnatter, line 181: we check the permissions in a similar manner in some other places, including the toolbar actions created in the `PageAdmin`: https://github.com/sulu/sulu/blob/b6613b73d94580d0c5daf65dd6c9f719804458f3/src/Sulu/Bundle/PageBundle/Admin/PageAdmin.php#L122

alexander-schranz: Where does `!!locked` come from I don't see any definition of it?

niklasnatter: looks like the `CollectionController` does not use our `ListBuilder`. the `locked` property is defined here:  https://github.com/sulu/sulu/blob/e3244a8fe56379d02504e47c4c31b2681223e4e1/src/Sulu/Bundle/MediaBundle/Api/Collection.php#L510-L520

alexander-schranz: forget that this is a `string` and validate via jexl expression 🙈

alexander-schranz (approved):

PR #6669, opened 2022-06-27
Add BlockToolbar for copy, remove, cut, duplicate multiple selected blocks

### LINKED ISSUE #6436: Copy & Paste for Blocks

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? |  yes
| BC breaks? | no 
| Deprecations? | no  
| Fixed tickets | fixes #6436 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Add blocktoolbar for multi copy and paste functionality

#### Why?

Multiple blocks should be able to be copied and pasted.


UPGRADE.md
```diff
@@ -48,6 +48,15 @@ The Symfony 4.4 compatibility service
 `Sulu\Bundle\WebsiteBundle\Controller\ExceptionController` / `sulu_website.exception_controller`
 was removed. See also [UPGRADE 2.1.0-RC1](#210-rc1).
 
+### Block component draftHandle property was renamed
+
+The `draftHandle` property of the `Block.js` component was changed to `handle`:
+
+```diff
+-<Block draftHandle={handleComponent} />
++<Block handle={handleComponent} />
```
luca-rath:
```suggestion
### Block component dragHandle property was renamed

The `dragHandle` property of the `Block.js` component was changed to `handle`:

```diff
-<Block dragHandle={handleComponent} />
+<Block handle={handleComponent} />
```

src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/SortableBlockList.js
```diff
@@ -34,9 +38,21 @@ class SortableBlockList<T: string, U: {type: T}> extends React.Component<Props<T
...
+        mode: 'sortable',
+        movable: null,
     };
 
+    constructor(props: Props<T, U>) {
+        super(props);
+
+        if (props.movable === false) {
+            log.warn(
+                'The "mode" prop of the "SortableBlockList" component is deprecated since 2.5 and will ' +
```
luca-rath:
```suggestion
                'The "movable" prop of the "SortableBlockList" component is deprecated since 2.5 and will ' +
```

src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/types.js
```diff
@@ -8,3 +8,5 @@ export type BlockActionConfig = BlockButtonActionConfig | DividerActionConfig ;
 
 export type RenderBlockContentCallback<T: string, U: {type: T}>
     = (value: U, type: T, index: number, expanded: boolean) => Node;
+
+export type BlockMode = 'static' | 'sortable' | 'selection';
```
luca-rath: To unify those modes a bit, we should either go with (`sortable` and `selectable`) or (`sort` and `selection`), would prefer the first suggestion

luca-rath (changes requested):

src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/BlockCollection.js
```diff
@@ -331,35 +437,134 @@ class BlockCollection<T: string, U: {type: T}> extends React.Component<Props<T,
...
+                    onSelect={this.handleSelect}
                     onSettingsClick={onSettingsClick ? this.handleSettingsClick : undefined}
                     onSortEnd={this.handleSortEnd}
                     onTypeChange={this.handleTypeChange}
+                    onUnselect={this.handleUnselect}
                     renderBlockContent={renderBlockContent}
                     renderDivider={this.renderAddButton}
+                    selectedBlocks={this.selectedBlocks}
                     types={types}
                     useDragHandle={true}
```
luca-rath, line 569: Does this prop even exist?

luca-rath:
Another thing: Personally I would expect that the BlockToolbar disappears, when the user either unselect all items or finishes an action, but I guess a UX expert should decide that

src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/BlockCollection.js
alexander-schranz: it seems be used by the underlaying dragging library.

alexander-schranz:
@luca-rath yeah I'm also not sure about behaviour after some action. But @chirimoya said he will have a look at it then. So I think we will merge it with the current state and so @chirimoya can test it then.

src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/types.js
alexander-schranz: Used `selectable`

luca-rath (approved):

PR #6673, opened 2022-06-29
Fix import xliff link problem

floscom:
Added the importData function so that the link can be imported correctly.

| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | slack communication - https://sulu-io.slack.com/archives/C0430GGCV/p1656490331677969

#### What's in this PR? Why?

Fixes a problem while importing xliff files. ImportData function was missing for type link.

#### Example Usage
```sh
bin/console sulu:webspaces:import example.en.xliff main en
```

### DIFF
--- a/src/Sulu/Component/Content/Tests/Unit/Types/LinkTest.php
+++ b/src/Sulu/Component/Content/Tests/Unit/Types/LinkTest.php
@@ -11,6 +11,7 @@
 
 namespace Sulu\Component\Content\Tests\Unit\Types;
 
+use PHPCR\NodeInterface;
 use PHPUnit\Framework\TestCase;
 use Prophecy\Argument;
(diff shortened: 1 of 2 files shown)


alexander-schranz:
Thank you for the pull request. I will need to test it out, but already wanted to ask if you could add a testcase also to the [LinkTest.php](https://github.com/sulu/sulu/blob/2.4/src/Sulu/Component/Content/Tests/Unit/Types/LinkTest.php) for this method..

niklasnatter:
I have pushed a test case for you. Hope thats okay 🙂

floscom:
Yah.. thx

alexander-schranz (approved):

alexander-schranz:
@floscom thank you!

PR #6675, opened 2022-06-30
Add rel attribute to external links

### LINKED ISSUE #3668: Add nofollow option to links in ckeditor
| Q | A
| --- | ---
| Bug? | no
| New Feature? | yes
| Sulu Version | latest

#### Actual Behavior

Right now a user is able to add a link within a text block using the text_editor (ckeditor). When using the 'link' button to create a link the user does not have the option to add `rel="nofollow"` to the link. Which is very useful for certain links.

Source: https://support.google.com/webmasters/answer/96569 

#### Expected Behavior

I would like the user to be able to also have the option to add `rel="nofollow"` to the link so crawlers know how to read these links as the user intents them to. Without having to give users the rights to go into 'code mode' and add the nofollow by hand.

#### Possible Solutions

I do not have enough knowledge to extend the ckeditor within Sulu. However some else had a similar question which seems to go into the same direction i'm aiming at. https://stackoverflow.com/questions/39416655/how-can-i-add-rel-nofollow-to-a-link-in-ckeditor-if-its-an-external-link

luca-rath:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #3668
| Related issues/PRs | -
| License | MIT
| Documentation PR | sulu/sulu-docs#721

#### What's in this PR?

Add rel attribute to external links

#### Example Usage

```xml
    <meta>
        <title lang="en">Link</title>
        <title lang="de">Link</title>
    </meta>

    <params>
        <param name="enable_attributes" value="true"/>
    </params>
</property>
```


src/Sulu/Bundle/AdminBundle/Resources/js/containers/Link/overlays/ExternalLinkTypeOverlay.js
```diff
@@ -169,11 +203,21 @@ class ExternalLinkTypeOverlay extends React.Component<LinkTypeOverlayProps> {
...
                             <Input onChange={onTitleChange} value={title} />
                         </Form.Field>
                     }
+
+                    {onRelChange
+                        && <Form.Field label={translate('sulu_admin.link_rel')}>
+                            <MultiSelect onChange={this.handleRelChange} values={this.relValue}>
+                                <MultiSelect.Option value="nofollow">nofollow</MultiSelect.Option>
+                                <MultiSelect.Option value="noreferrer">noreferrer</MultiSelect.Option>
+                                <MultiSelect.Option value="noopener">noopener</MultiSelect.Option>
```
luca-rath: I implemented this as a `MultiSelect` for now, to be most flexible in the future, even though @alexander-schranz mentioned [here](https://github.com/sulu/sulu/issues/3668#issuecomment-1143579307), that he only wants to support `nofollow`. I'm open to change this to a checkbox to only allow `nofollow`, but the underlying logic should stay the same, that `rel` is passed around as `string` instead of a bool flag specificially for `nofollow`.

/cc @sulu/core-team Would you only allow `nofollow` or are there more options that would make sense? A list can be found here: https://www.w3schools.com/tags/att_a_rel.asp

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Link/Link.js
```diff
@@ -69,6 +72,16 @@ class Link extends Component<Props> {
...
+        if (value.provider === 'external') {
+            const {href} = value;
+
+            if (href) {
+                this.titleParts = [href];
+
+                return;
+            }
+        }
+
```
luca-rath, line 81: Currently the `link` content type doesn't show a value, if the `external` provider is used. I know, that the `Link` container should not know about `external`, but I'm not sure how we should fix that. Checking if `linkTypeRegistry.getOptions(value.provider)` returns `undefined` is a bit hacky and might break in the future.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/Link.js
```diff
@@ -25,6 +25,9 @@ export default class Link extends React.Component<FieldTypeProps<LinkValue>> {
                 enable_title: {
                     value: enableTitle,
                 } = {},
+                enable_rel: {
+                    value: enableRel,
+                } = {},
```
alexander-schranz: As written in the issue the `enable_target`, `enable_title` should be deprecated and be replaced under a `enable_attributes` option.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Link/overlays/ExternalLinkTypeOverlay.js
alexander-schranz: We should go with `Switch` here for usability and not a multi selected. Still it can be multiple switches but a switch avoids requirement for 2 clicks.

At current state as written in the issue we will not support something else as `nofollow`. The other 2 flags are to often misinterpreted and should not be managed by content managers.

wachterjohannes: @luca-rath not sure but the only one which could make sense is `noreferrer` /cc @alexander-schranz

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/Link.js
```diff
@@ -34,9 +38,52 @@ export default class Link extends React.Component<FieldTypeProps<LinkValue>> {
             },
         } = this.props;
 
+        const enabledAttributes = [];
+
+        if (enableAttributes !== undefined) {
+            if (!isArrayLike(enableAttributes)) {
```
alexander-schranz: Sorry miscommunicated. The `enable_attributes` is a `boolean` flag which just enables every attribute for the content manager. There is not longer a limitation here for specific attributes or not.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/Link.js
```diff
@@ -17,13 +18,16 @@ export default class Link extends React.Component<FieldTypeProps<LinkValue>> {
             value,
             schemaOptions: {
                 enable_anchor: {
-                    value: enableAnchor,
+                    value: deprecatedEnableAnchor,
```
alexander-schranz: enable_anchor is not effected as it is not an `attribute`

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Link/Link.js
niklasnatter: i think i would always use `value.href` as title if we dont have any `options` or `options.displayProperties`. then we dont need to check the type and i think its better to display anything than display nothing 🙂

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/Link.js
```diff
@@ -34,9 +38,53 @@ export default class Link extends React.Component<FieldTypeProps<LinkValue>> {
...
 
+        if (enableAnchor !== undefined && enableAnchor !== null && typeof enableAnchor !== 'boolean') {
+            throw new Error('The "enable_anchor" schema option must be a boolean if given!');
+        }
+
+        let enableTarget = false,
+            enableTitle = false,
+            enableRel = false;
+
+        if (enableAttributes !== undefined && enableAnchor !== null) {
```
niklasnatter:
```suggestion
        if (enableAttributes !== undefined && enableAttributes !== null) {
```

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/Link.js
```diff
@@ -34,9 +38,53 @@ export default class Link extends React.Component<FieldTypeProps<LinkValue>> {
...
+            if (typeof enableAttributes !== 'boolean') {
```
niklasnatter, line 50: the description of this pr passes a collection here. is the description outdated?

luca-rath: yes, I adjusted it

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Link/Link.js
alexander-schranz: Sounds good for me fallbacking always to `href` in that case.

niklasnatter (approved):

PR #6676, opened 2022-06-30
Add support for position sticky inside tabs container

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? | yes
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Add support for position sticky inside tabs container.

#### Why?

The current negative margin on the `tabs-container` does break the whole position sticky things.

Why it would be possible to use a top: - the same value. The field does not know if its inside a tabs-container or not even the tabs-container is not a parent class. And as fields are also used outside of tabs-container example in form overlay it is required that get ride of the minus margins in the views.

By moving the padding from the application to the `ViewRenderer` we can based on new added `ViewConfig` and the new `rootSpaceless` option there if the padding should be added or not on its rootelement.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/ViewRenderer/ViewRenderer.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/ViewRenderer/ViewRenderer.js
@@ -4,9 +4,9 @@ import {observer} from 'mobx-react';
 import {observable, reaction} from 'mobx';
 import Router, {getViewKeyFromRoute, Route} from '../../services/Router';
 import userStore from '../../stores/userStore';
+import View from '../../components/View';
 import viewRegistry from './registries/viewRegistry';
 import type {Element} from 'react';
-import type {View} from './types';
 
 type Props = {
     router: Router,
@@ -48,37 +48,36 @@ class ViewRenderer extends React.Component<Props> {
         }
     }
 
-    getView = (route: Route): View => {
-        const View = viewRegistry.get(route.type);
-
-        if (!View) {
-            throw new Error('View "' + route.type + '" has not been found');
-        }
-
-        return View;
-    };
-
     renderView(route: Route, child: Element<*> | null = null) {
         const {router} = this.props;
-        const View = this.getView(route);
+        const CurrentView = viewRegistry.get(route.type);
+        const viewConfig = viewRegistry.getConfig(route.type);
 
         let viewKey = getViewKeyFromRoute(route, router.attributes) || '';
-        if (View.remountViewOnLogin) {
+        if (CurrentView.remountViewOnLogin) {
             viewKey = viewKey + '__' + this.loginCount;
         }
 
         const element = (
-            <View
+            <CurrentView
                 isRootView={!route.parent}
                 key={viewKey}
                 route={route}
                 router={router}
             >
                 {(props) => child ? React.cloneElement(child, props) : null}
-            </View>
+            </CurrentView>
         );
 
         if (!route.parent) {
+            if (!viewConfig.rootSpaceless) {
+                return (
+                    <View>
+                        {element}
+                    </View>
+                );
+            }
+
             return element;
         }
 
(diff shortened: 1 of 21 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/containers/ViewRenderer/tests/ViewRenderer.test.js
```diff
@@ -30,16 +44,12 @@ test('Render view returned from ViewRegistry with passed router', () => {
...
+    viewRegistry.getConfig.mockReturnValue({});
     const view = render(<ViewRenderer router={router} />);
     expect(view).toMatchSnapshot();
     expect(viewRegistry.get).toBeCalledWith('test');
 });
 
-test('Render view should throw if view does not exist', () => {
-    viewRegistry.get.mockReturnValue(undefined);
-    expect(() => render(<ViewRenderer router={{route: {type: 'not_existing'}}} />)).toThrow(/not_existing/);
-});
```
alexander-schranz, line 41: This is unnecessary check which was removed the `viewRegistry.get` already check and throw an exception is view does not exist there was no need todo the same in the `viewRenderer`.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/ViewRenderer/ViewRenderer.js
```diff
@@ -48,37 +49,36 @@ class ViewRenderer extends React.Component<Props> {
         }
     }
 
-    getView = (route: Route): View => {
-        const View = viewRegistry.get(route.type);
-
-        if (!View) {
-            throw new Error('View "' + route.type + '" has not been found');
-        }
```
alexander-schranz, line 56: This is unnecessary check which was removed because the `viewRegistry.get` already check and throw an exception is view does not exist there was no need todo the same in the `viewRenderer`.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/ViewRenderer/ViewRenderer.js
```diff
@@ -48,37 +49,36 @@ class ViewRenderer extends React.Component<Props> {
...
                 route={route}
                 router={router}
             >
                 {(props) => child ? React.cloneElement(child, props) : null}
-            </View>
+            </CurrentView>
         );
 
         if (!route.parent) {
+            if (!viewConfig.rootSpaceless) {
```
alexander-schranz: The new configuration was required. There actually was no way to get it work without it specially in cases of nested tab views (version & activities) and Tabs inside WebspaceTabs. Funfact actually with this new option there is no bc break and so custom views still get the same padding like they did before.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/ViewRenderer/tests/ViewRenderer.test.js
```diff
@@ -144,48 +155,11 @@ test('Render view with parents should nest rendered views and correctly pass chi
...
-                        <div>
-                            <h2>Form</h2>
-                            {props.children}
-                        </div>
-                    );
-                };
-        }
-    });
-
-    expect(() => render(<ViewRenderer router={router} />)).toThrow(/app/);
```
alexander-schranz, line 186: Same as above the viewRegistry already handles this.

src/Sulu/Bundle/PageBundle/Resources/js/views/WebspaceTabs/variables.scss
```diff
@@ -1 +1 @@
-$webspaceTabsWebspaceSelectMargin: 60px;
```
alexander-schranz, line 1: 40px are now already given by the `<View>` component

src/Sulu/Bundle/AdminBundle/Resources/js/containers/ViewRenderer/registries/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+```js
+const CustomView = <div>Custom View </div>;
+
+viewRegistry.set('customView', CustomView, { rootSpaceless: true });
+```
+
+| Name          | Type    | DefaultValue |
+|---------------|---------|--------------|
+| rootSpaceless | boolean | false        |
```
niklasnatter, line 19: would call this parameter `disableDefaultSpacing`/`omitDefaultSpacing`

niklasnatter:
Looks good to me beside the parameter name 🙂

PR #6677, opened 2022-06-30
Add Sticky component

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? | yes
| BC breaks? | no 
| Deprecations? | no  
| Fixed tickets | fixes # 
| Related issues/PRs | https://github.com/sulu/sulu/pull/6676, https://github.com/sulu/sulu/pull/6669 
| License | MIT

#### What's in this PR?

Add a sticky component which childs allow to behave different between stick state.

#### Why?

This will be required for the BlockToolbar. 

#### Example Usage

```js
<Sticky>{
    (isSticky) => <span>{isSticky ? 'Stick' : 'Unsticky'}</span>
}</Sticky>
```

sticky

#### To Do

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Sticky/sticky.scss
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Sticky/sticky.scss
@@ -1,0 +1,14 @@
+.sticky {
+    position: sticky;
+    top: 0;
+    z-index: 1; /* push element in front of position: relative elements */
+}
+
+.stickySentinel {
+    position: absolute;
+    left: 0;
+    right: 0;
+    height: 1px;
+    visibility: hidden;
+    pointer-events: none;
+}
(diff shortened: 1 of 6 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/components/Sticky/sticky.scss
```diff
@@ -0,0 +1,14 @@
+.sticky {
+    position: sticky;
+    top: 0;
+    z-index: 1; /* push element in front of position: relative elements */
```
alexander-schranz, line 4: Else position relative elements are above the sticky toolbar:

position-sticky-issue

https://jsfiddle.net/oey4dq1x/

niklasnatter: 😱

niklasnatter: this will probably push the sticky element above our overlays/dialogs too 😕  i think we can prevent that from happening by setting `z-index: 0` in the parent component that uses this component, right?

niklasnatter (approved):

src/Sulu/Bundle/AdminBundle/Resources/js/components/Sticky/sticky.scss
niklasnatter: unfortunately, i think there is no way around this problem while keeping the needed functionality, so i am going to merge this 🙂

PR #6678, opened 2022-06-30
Add Tooltip component

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? | yes 
| BC breaks? | no 
| Deprecations? | no  
| Fixed tickets | fixes # 
| Related issues/PRs | https://github.com/sulu/sulu/pull/6669 
| License | MIT

#### What's in this PR?

Add Tooltip component.

#### Why?

For the BlockToolbar we need a Tooltip component.

#### Example Usage

```html
<Tooltip label="Copy">
    <-- Button is inside to bubble focus event to the Tooltip -->
    <button type="button" onClick={() => {}} aria-label="Copy" style={buttonStyle}>
        <Icon name="su-copy" />
    </button>
</Tooltip>
```

tooltip

#### To Do

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Popover/Popover.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Popover/Popover.js
@@ -22,6 +22,7 @@ type Props = {
         verticalPosition: string,
         horizontalPosition: string,
     ) => Node,
+    horizontalCenter?: boolean,
     horizontalOffset: number,
     onClose?: () => void,
     open: boolean,
@@ -35,6 +36,7 @@ const CLOSE_KEY = 'esc';
 class Popover extends React.Component<Props> {
     static defaultProps = {
         backdrop: true,
+        horizontalCenter: false,
         horizontalOffset: 0,
         open: false,
         verticalOffset: 0,
@@ -103,6 +105,7 @@ class Popover extends React.Component<Props> {
         const {
             anchorElement,
             verticalOffset,
+            horizontalCenter,
             horizontalOffset,
             centerChildElement,
         } = this.props;
@@ -115,6 +118,8 @@ class Popover extends React.Component<Props> {
         const centerChildOffsetTop = (centerChildElement) ? centerChildElement.offsetTop : 0;
         const alignOnVerticalAnchorEdges = !centerChildElement;
 
+        const horizontalCenterValue = horizontalCenter ? (width - this.popoverWidth) / 2 : 0;
+
         return PopoverPositioner.getCroppedDimensions(
             this.popoverWidth,
             this.popoverHeight,
@@ -122,7 +127,7 @@ class Popover extends React.Component<Props> {
             left,
             width,
             height,
-            horizontalOffset,
+            horizontalCenterValue + horizontalOffset,
             verticalOffset,
             centerChildOffsetTop,
             alignOnVerticalAnchorEdges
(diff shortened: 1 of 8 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/components/Popover/Popover.js
```diff
@@ -22,6 +22,7 @@ type Props = {
         verticalPosition: string,
         horizontalPosition: string,
     ) => Node,
+    horizontalCenter?: boolean,
```
niklasnatter: i think calling this `horizontalAnchorMode` or `horizontalAnchorPosition` would be a bit easier to understand. we can allow `left` and `center` for now as value

src/Sulu/Bundle/AdminBundle/Resources/js/components/Tooltip/README.md
```diff
@@ -0,0 +1,52 @@
+This component provide the Tooltip:
```
niklasnatter:
```suggestion
This component displays a tooltip if its content is hovered or focused:
```

src/Sulu/Bundle/AdminBundle/Resources/js/components/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,83 @@
...
+import React from 'react';
+import classNames from 'classnames';
+import {action, observable} from 'mobx';
+import {observer} from 'mobx-react';
+import Popover from '../Popover';
+import tooltipStyles from './tooltip.scss';
+import type {ChildrenArray, Element, ElementRef} from 'react';
+
+type Props = {|
+    children: ChildrenArray<Element<*> | false>,
```
niklasnatter: if we want to allow all valid react children, we should use `Node` like we do in the `Dialog`:
https://github.com/sulu/sulu/blob/62b24decee4b209a1e54590ae18a590a3bdee01a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Dialog/Dialog.js#L17

alexander-schranz: Need to check that. Node was failing strangely for me in flow when I had it right in my mind.

src/Sulu/Bundle/AdminBundle/Resources/js/components/Popover/Popover.js
alexander-schranz: Did go with `horizontalAnchorMode`

src/Sulu/Bundle/AdminBundle/Resources/js/components/Tooltip/Tooltip.js
alexander-schranz:
```
Cannot create Tooltip element because React.Element [1] is incompatible with Node [2] in property children.
[incompatible-type]

     src/Sulu/Bundle/AdminBundle/Resources/js/components/Tooltip/tests/Tooltip.test.js
      7│ test('The component should render in unfocused state', () => {
      8│     const component = render(
      9│         <Tooltip label="Copy">
 [1] 10│             <button aria-label="Copy" type="button">
     11│                 <Icon name="su-copy" />
     12│             </button>
     13│         </Tooltip>
     14│     );
     15│

     src/Sulu/Bundle/AdminBundle/Resources/js/components/Tooltip/Tooltip.js
 [2] 11│     children: Node,
 ```

alexander-schranz: Okay need to import the correct type 🙈

PR #6682, opened 2022-07-06
Add possibility to use translation keys in template meta information

### LINKED ISSUE #6661: Using translation keys for template meta information does not work
| Q | A
| --- | ---
| Sulu Version | 2.4.2
| PHP Version | 8.1.3
| DB Version | MySQL 5.7.36
| Browser Version | Firefox 101.0

# Actual Behavior

When defining an article template, title and info text can be defined as meta information. The same can be done for properties.

For properties it's possible to remove the `lang` attribute and only specify one title or info text. The content is then interpreted as a translation key and looked up in the admin translation json file.

However, this does not work for the title and info text of the template itself. If the localized versions are missing, the `key` of the template is taken as the fallback value.

# Expected Behavior

I can define a title or info text as meta information of a template without defining localized versions with the `lang` attribute. Instead I can add a translation key as the value, same as for the properties of the template.

# Steps to Reproduce

Change the meta information of an article template from

grafik

to 

grafik

The template key will be used instead of the translation key when the title is rendered in Sulu.

grafik

# Possible Solutions

Use the same logic as in the PropertiesXmlParser https://github.com/sulu/sulu/blob/2.5/src/Sulu/Component/Content/Metadata/Parser/PropertiesXmlParser.php#L242 in the StructureXmlLoader https://github.com/sulu/sulu/blob/2.5/src/Sulu/Component/Content/Metadata/Loader/StructureXmlLoader.php#L362

I checked this solution locally and it delivers the expected behaviour as far as I am aware. If you are okay with this approach, I could create a PR for it.

lribi:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | fixes #6661 
| Related issues/PRs | - 
| License | MIT

#### What's in this PR?

This PR adds support to define translation keys as meta information in template meta tags.

#### Why?

This streamlines the behaviour between the meta tags of the template itself and its properties as they already have this behaviour.

See #6661 for further details.

### DIFF
--- a/src/Sulu/Bundle/PageBundle/Resources/config/structure.xml
+++ b/src/Sulu/Bundle/PageBundle/Resources/config/structure.xml
@@ -11,8 +11,10 @@
             <argument type="service" id="sulu_page.structure.properties_xml_parser"/>
             <argument type="service" id="sulu_page.structure.schema_xml_parser"/>
             <argument type="service" id="sulu.content.type_manager" />
+            <argument type="service" id="translator" />
             <argument>%sulu.content.structure.required_properties%</argument>
             <argument>%sulu.content.structure.required_tags%</argument>
+            <argument>%sulu_core.translated_locales%</argument>
         </service>
 
         <service id="sulu_page.structure.properties_xml_parser"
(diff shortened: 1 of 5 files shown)


alexander-schranz (approved):

alexander-schranz:
@lribi Good work 👍 I just changed the constructor to avoid a bc break as we sadly did not make the class final and somebody could have overwritten it in there project.

lribi:
Thanks @alexander-schranz for your additions, didn't cross my mind at the time.

PR #6685, opened 2022-07-07
Add link provider as a data attribute to rendered <sulu-link>

rogamoore:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | sulu/sulu-docs#724

#### What's in this PR?

When adding internal links in CKEditor there are several link providers available ("page", "media", "article").
Right now the information which provider has been selected will be removed when converting the `<sulu-link>` tag to a real HTML `a` tag.
This simple change will forward the selected provider as an HTML data attribute.

#### Why?

Adding the selected provider as a data attribute allows targeting the links in CSS and in Javascript.

Examples:
- Add an icon to download (media) links
- Format internal links ("page", "article") differently
- Add custom analytics tracking by link type

#### Example Usage

```css
a[data-provider="media"] {
  font-color: green;
}
```

#### To Do

### DIFF
--- a/src/Sulu/Bundle/MarkupBundle/DependencyInjection/Configuration.php
+++ b/src/Sulu/Bundle/MarkupBundle/DependencyInjection/Configuration.php
@@ -1,0 +1,35 @@
+<?php
+
+/*
+ * This file is part of Sulu.
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\MarkupBundle\DependencyInjection;
+
+use Symfony\Component\Config\Definition\Builder\TreeBuilder;
+use Symfony\Component\Config\Definition\ConfigurationInterface;
+
+class Configuration implements ConfigurationInterface
+{
+    public function getConfigTreeBuilder()
+    {
+        $treeBuilder = new TreeBuilder('sulu_markup');
+
+        $treeBuilder->getRootNode()
+            ->children()
+                ->arrayNode('link_tag')
+                ->addDefaultsIfNotSet()
+                    ->children()
+                        ->scalarNode('provider_attribute')->defaultNull()->end()
+                    ->end()
+                ->end()
+            ->end();
+
+        return $treeBuilder;
+    }
+}
--- a/src/Sulu/Bundle/MarkupBundle/DependencyInjection/SuluMarkupExtension.php
+++ b/src/Sulu/Bundle/MarkupBundle/DependencyInjection/SuluMarkupExtension.php
@@ -23,6 +23,10 @@ class SuluMarkupExtension extends Extension
 {
     public function load(array $configs, ContainerBuilder $container)
     {
+        $config = $this->processConfiguration(new Configuration(), $configs);
+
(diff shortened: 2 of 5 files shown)


alexander-schranz:
@rogamoore  I like the idea. For backwards compatibility we need to make sure that a custom "sulu-link" added tag with a `data-provider` attribute does not getting overwritten:

```html
<sulu-link href="4f607788-3d76-48cc-80d8-89714c370c38" provider="page" data-provider="custom" data-other="other">Test</sulu-link>
```
Curently does render `data-provider="page"` instead of `data-provider="custom"` which should be the case.

For this cases we should add also test cases so we make sure here get nothing broken. Can you add a check for exist data-provider property and can you add test cases for both cases one sulu-link with a data-provider one without a data-provider property?

rogamoore:
@alexander-schranz Thanks a lot for your feedback! 🙂
I just updated the existing tests to take into account the new attribute and I also added another test case covering the situation where there is already an existing `data-provider` attribute.

I was also thinking about the naming - maybe `data-link-provider` or `data-sulu-link-provider` would be a bit more expressive?

alexander-schranz:
The `data-provider` is fine we should not expose `sulu` here and the `a` tag already defines that its a link, still you get me to the point that somebody want to use something different and maybe this feature should not be activated by default. So I would hide it behind a configuration:

```yaml
sulu_markup:
    link_tag:
        provider_attribute: 'data-provider'
```

By default it the `provider_attribute` is `null` and so not written. So we are not changing any existing projects html and make some html "snapshot" tests so failing. 

If you need any help creating the configuration tree let me know. The `LinkTag` service should then get the `providerAttribute` into the constructor:

```php
    public function __construct(
        LinkProviderPoolInterface $linkProviderPool,
        bool $isPreview = false,
        UrlHelper $urlHelper = null
        ?string $providerAttribute = null
    ) {
```

And when null nothing should be outputted. When set it should be outputted the behaviour with the `dummy` should stay as it is.

Sorry for the circumstances 🙈

alexander-schranz (approved):

rogamoore:
@alexander-schranz I like the idea using a config option 👍 
I just added some changes - not sure if more tests are needed. I tested it locally and it worked as expected. :)

alexander-schranz:
@rogamoore Thank you. Good work 👍

rogamoore:
Thank you @alexander-schranz for the quick response and review! 🙂

alexander-schranz:
@rogamoore could you add a section about this new configuration option to the markup bundle documentation? https://github.com/sulu/sulu-docs/blob/2.x/bundles/markup/index.rst

rogamoore:
@alexander-schranz Sure, will take care of it in the next few hours.

alexander-schranz:
Nice. For the skeleton we decided to configure it so new project can use it out of the box: https://github.com/sulu/skeleton/pull/180

rogamoore:
@alexander-schranz I just created the docs PR https://github.com/sulu/sulu-docs/pull/724
I moved it to `link.rst` because it only affects the internal `sulu-link` implementation.

PR #6688, opened 2022-07-10
Upgrade ckeditor to version 34.2.0 and postcss to version 8

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? |  yes
| BC breaks? | no 
| Deprecations? | no  
| Fixed tickets | parts of #6343 
| Related issues/PRs | # 
| License | MIT
| Documentation PR | sulu/sulu-docs# 

#### What's in this PR?

Upgrade ckeditor to version 34.2.0.

#### Why?

Supporting the latest ckeditor features and plugins.

#### TODO

 - [x] https://github.com/madeleineostoja/postcss-hexrgba/issues/31 all `rgba($` do not longer work

Issue #6343, opened 2021-11-02
Outdated javascript dependencies


niklasnatter:
Unfortunately, we cannot update some javascript packages because of various reasons. This issue tracks which packages are affected and why we need to use an outdated package version:

| Package | Used Version |  Available Version | Reason |
| --- | --- | --- | --- |
| ckeditor | 41 | 47 LTS | requires updates plugins see comment below
| mobx | 4 | 6 | contains breaking changes for project code, comeback of decorators (https://github.com/mobxjs/mobx/pull/3638) will make this easier [WIP PR](https://github.com/sulu/sulu/pull/7211) |
| jest / babel-jest | 29 | 30 | ~tests are failing because of comparison changes: #6361~ Update 2026-02: next step to 30 (Enzym should be merged before) |
| react | 17 | 18 | our testing library `enzyme` is not compatible with react@18: https://dev.to/wojtekmaj/enzyme-is-dead-now-what-ekl #6689
| @testing-library/react | 12 | - | we need to update react first
| @testing-library/jest | 5 | - | we need to update jest first
| flow-bin | 0.142 | 0.303 | requires big upgrade effort because flow-bin@0.143 dropped classic mode. it might be better to switch to typescript than making this effort |
| mobx-react | 5 | 7 |  |
| react-leaflet | 3 | 4 | depends on react@18: #6689

Webpack 5:

 - [x] When switching to webpack 5 we can avoid the hardcoded browserlist which was introduced in #6919 and sulu/skeleton#200

danrot:
I'd be interested to update webpack 5, so that we could maybe later fix sulu/skeleton#88 by introducing [module federation](https://webpack.js.org/concepts/module-federation/).

The good news is that ckeditor seems to have [fixed their issue with Webpack 5](https://github.com/ckeditor/ckeditor5/commit/d312ab630000f84f232ff2c372cdfc53e06b7f16). 

But the linked commit from ckeditor has not been released yet. And when thinking about it, since multiple ckeditor packages are in use, it might be hard to get all of that working when not being able to rely on releases, so it might be worthwhile waiting for the release to start working on this.

While this is nothing that you can have an influence on, there is a second question that I feel has to be asked: Would you consider upgrading to webpack 5 to be a breaking change? I think strictly speaking it is, since Sulu allows to extend the admin interface by adjusting its webpack config, and upgrading to webpack 5 might cause many custom admin build to fail. Is that something that should be done for 3.0 only?

EDIT:
[...]

PR #6690, opened 2022-07-10
Upgrade webpack to version 5

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? |  yes
| BC breaks? | no 
| Deprecations? | no  
| Fixed tickets | parts of #6343 
| Related issues/PRs | ~~requires #6688~~, https://github.com/sulu/skeleton/pull/201 
| License | MIT

#### What's in this PR?

Upgrade webpack to version 5.

#### Why?

Support latest version of webpack. 

#### TODO

 - [x] Remove dependencies to `util`, `process` packages via `isemail` package.
 - [x] Remove dependency to `Buffer` package see own js
 - [x] Undo changes of https://github.com/sulu/sulu/pull/6690/commits/822516faa66cff14abd75c1fec9a3b79c0b12b06
 - [x] Fix Build with skeleton
 - [x] Replace Email Validator

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -16,7 +16,7 @@ Sulu requires now new versions of the following JS packages:
 - `is-email` was removed and replaced by `sulu-admin-bundle/utils/Email/validateEmail` method 
 
 If you created custom admin components you need to upgrade also
-upgrade your components. The core js will automatically be update
+that components. The core js will automatically be update
 via the [update build](https://docs.sulu.io/en/latest/upgrades/upgrade-2.x.html) command.
 
 ### Deprecated urls variable in return value of sulu_content_load
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -2,6 +2,23 @@
 
 ## 2.6.0
 
+### Webpack 5 upgrade
+
+Sulu requires now new versions of the following JS packages:
+
+- `webpack`: `^5.75.0`
+- `webpack-cli`: `^5.0`
+- `webpack-manifest-plugin`: `^5.0.0`
+- `mini-css-extract-plugin`: `^2.7.1`
+- `optimize-css-assets-webpack-plugin` was removed replaced by `css-minimizer-webpack-plugin`: `^4.2.2`
+- `clean-webpack-plugin` was removed and replaced by `clean: true` webpack output option
+- `webpack-clean-obsolete-chunks` was removed and replaced by `clean: true` webpack output option
+- `is-email` was removed and replaced by `sulu-admin-bundle/utils/Email/validateEmail` method 
+
+If you created custom admin components you need to upgrade also
+upgrade your components. The core js will automatically be update
+via the [update build](https://docs.sulu.io/en/latest/upgrades/upgrade-2.x.html) command.
+
 ### Deprecated urls variable in return value of sulu_content_load
 
 The `urls` variable in the return value of the `sulu_content_load` function was deprecated.
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -4,7 +4,7 @@
 
 ### Webpack 5 upgrade
 
-Sulu requires now new versions of the following JS packages:
+Sulu now uses Webpack 5 to build the administration interface application. To enable this, the following JavaScript dependencies were updated/changed:
 
 - `webpack`: `^5.75.0`
 - `webpack-cli`: `^5.0`
@@ -15,9 +15,10 @@ Sulu requires now new versions of the following JS packages:
 - `webpack-clean-obsolete-chunks` was removed and replaced by `clean: true` webpack output option
 - `is-email` was removed and replaced by `sulu-admin-bundle/utils/Email/validateEmail` method 
 
-If you created custom admin components you need to upgrade also
-that components. The core js will automatically be update
-via the [update build](https://docs.sulu.io/en/latest/upgrades/upgrade-2.x.html) command.
+If you have integrated custom JavaScript components into the administration interface,
+you might need to adjust your components to be compatible with the updated dependencies.
+If you have not integrated custom JavaScript code, you project is adjusted automatically by the
+[update build](https://docs.sulu.io/en/latest/upgrades/upgrade-2.x.html) command.
 
 ### Deprecated urls variable in return value of sulu_content_load
 
--- a/package.json
+++ b/package.json
@@ -50,9 +50,9 @@
         "autoprefixer": "^10.4.7",
         "babel-jest": "^26.3.0",
         "babel-loader": "^8.0.6",
-        "clean-webpack-plugin": "^4.0.0",
         "core-js": "^3.18.0",
         "css-loader": "^5.2.4",
+        "css-minimizer-webpack-plugin": "^4.2.2",
         "dependency-cruiser": "^11.12.0",
         "empty": "^0.10.1",
         "enzyme": "^3.3.0",
@@ -71,11 +71,10 @@
         "identity-obj-proxy": "^3.0.0",
         "jest": "^26.4.2",
         "jest-canvas-mock": "^2.1.1",
-        "mini-css-extract-plugin": "^1.5.0",
+        "mini-css-extract-plugin": "^2.7.1",
         "mobx": "^4.0.0",
         "mobx-react": "^5.0.0",
         "moment-timezone": "^0.5.14",
-        "optimize-css-assets-webpack-plugin": "^6.0.1",
         "postcss": "^8.4.14",
         "postcss-calc": "^8.2.4",
         "postcss-hexrgba": "^2.1.0",
@@ -93,10 +92,9 @@
         "stylelint": "^14.9.1",
         "stylelint-config-standard-scss": "^5.0.0",
         "url-search-params-polyfill": "^8.0.0",
-        "webpack": "^4.27.0",
-        "webpack-clean-obsolete-chunks": "^0.4.0",
-        "webpack-cli": "^4.7.0",
-        "webpack-manifest-plugin": "^4.0.2"
+        "webpack": "^5.75.0",
+        "webpack-cli": "^5.0.0",
+        "webpack-manifest-plugin": "^5.0.0"
     },
     "jest": {
         "moduleNameMapper": {
--- a/package.json
+++ b/package.json
@@ -88,7 +88,6 @@
         "react-styleguidist": "^11.0.5",
         "react-test-renderer": "^17.0.0",
         "regenerator-runtime": "^0.13.3",
-        "style-loader": "^2.0.0",
         "stylelint": "^14.9.1",
         "stylelint-config-standard-scss": "^5.0.0",
         "url-search-params-polyfill": "^8.0.0",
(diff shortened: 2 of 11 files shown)


package.json
```diff
@@ -65,18 +65,17 @@
         "identity-obj-proxy": "^3.0.0",
         "jest": "^26.4.2",
         "jest-canvas-mock": "^2.1.1",
-        "mini-css-extract-plugin": "^1.5.0",
+        "mini-css-extract-plugin": "^2.6.1",
         "mobx": "^4.0.0",
         "mobx-react": "^5.0.0",
         "moment-timezone": "^0.5.14",
-        "optimize-css-assets-webpack-plugin": "^6.0.1",
```
alexander-schranz, line 78: replaced by `css-minimizer-webpack-plugin`: https://github.com/NMFR/optimize-css-assets-webpack-plugin

package.json
```diff
@@ -39,16 +39,16 @@
...
-        "@ckeditor/ckeditor5-dev-utils": "^26.1.0",
-        "@ckeditor/ckeditor5-theme-lark": "^31.0.0",
+        "@ckeditor/ckeditor5-dev-utils": "^30.3.2",
+        "@ckeditor/ckeditor5-theme-lark": "^34.2.0",
         "@wojtekmaj/enzyme-adapter-react-17": "^0.6.1",
         "autoprefixer": "^9.8.6",
         "babel-eslint": "^10.0.2",
         "babel-jest": "^26.3.0",
         "babel-loader": "^8.0.6",
-        "clean-webpack-plugin": "^4.0.0",
```
alexander-schranz, line 53: think we not longer need this with `clean: true`

package.json
```diff
@@ -87,10 +86,9 @@
         "stylelint": "^13.2.1",
         "stylelint-config-standard": "^22.0.0",
         "url-search-params-polyfill": "^8.0.0",
-        "webpack": "^4.27.0",
-        "webpack-clean-obsolete-chunks": "^0.4.0",
```
alexander-schranz, line 97: The `clean: true` is the new way for this: https://github.com/GProst/webpack-clean-obsolete-chunks

webpack.config.js
```diff
@@ -100,10 +102,6 @@ module.exports = (env, argv) => { // eslint-disable-line no-undef
                     use: [
                         {
                             loader: MiniCssExtractPlugin.loader,
-                            options: {
-                                // set path to public root from bundled css: https://github.com/sulu/sulu/pull/6225
-                                publicPath: path.relative(outputPath, '.'),
```
alexander-schranz, line 105: not longer required as all pathes in css/js are relative

alexander-schranz:
While it did first look good to build with webpack 5 there are issues with the `isemail` package. There are currently 3 node packages which we require `util`, `process`, `Buffer`. This is not longer supported. The workarounds via webpack and browserify didn't work in combination with skeleton. So we will currently keep on webpack 4 until resolving this. Issues.

niklasnatter:
I think I would be okay with replacing the `isemail` if this is a blocker in the future. Where did we use the `Buffer` class in out code? 🤔

src/Sulu/Bundle/AdminBundle/Resources/js/components/Url/Url.js
```diff
@@ -4,8 +4,8 @@ import {observer} from 'mobx-react';
 import {action, computed, observable} from 'mobx';
 import classNames from 'classnames';
 import log from 'loglevel';
-import IsEmail from 'isemail';
```
alexander-schranz, line 7: Was never planned by the package to support browsers. They only support node based applications.

src/Sulu/Bundle/AdminBundle/Resources/js/utils/Email/validateEmail.js
```diff
@@ -0,0 +1,8 @@
+// @flow
+export default function(value: ?string): boolean {
+    if (!value) {
+        return false;
+    }
+
+    return /^(.+)@(.+)$/.test(value);
```
alexander-schranz: This is very dump regex. Not sure how strict we should even implement things here. /cc @sulu/core-team 

We definitely should not be too strict in case of supporting also local and intranet domains `@localhost`, `@massive.local`, ...

package.json
```diff
@@ -89,14 +88,12 @@
         "react-styleguidist": "^11.0.5",
         "react-test-renderer": "^17.0.0",
         "regenerator-runtime": "^0.13.3",
-        "style-loader": "^2.0.0",
```
alexander-schranz, line 92: I removed this as it seems not longer be used by us, think the css extract is working differently for us. And in the skeleton it was not even installed.

src/Sulu/Bundle/AdminBundle/Resources/js/utils/Email/validateEmail.js
alexander-schranz: Looks like there is a official regex by whatwg:

 - https://html.spec.whatwg.org/#valid-e-mail-address
 
Its also used inside the safari webkit browser: https://github.com/WebKit/WebKit/blob/e205a844146bab749642d1b88974b706fe4d9e7e/Source/WebCore/html/EmailInputType.cpp#L38-L39

src/Sulu/Bundle/AdminBundle/Resources/js/utils/Ajv/formats/tests/idnEmailValidator.test.js
```diff
@@ -5,10 +5,6 @@ test('Normal email address should pass validation', () => {
     expect(idnEmailValidator('hello@example.com')).toBe(true);
 });
 
-test('Email address with emojis should pass validation', () => {
-    expect(idnEmailValidator('🤌@😎')).toBe(true);
-});
-
```
alexander-schranz, line 11: Test was introduced by @luca-rath in https://github.com/sulu/sulu/pull/5815 but even browsers yet does not support this. So I think we can safely remove it:

Bildschirmfoto 2022-11-30 um 01 21 48

UPGRADE.md
```diff
@@ -2,6 +2,23 @@
 
 ## 2.6.0
 
+### Webpack 5 upgrade
+
+Sulu requires now new versions of the following JS packages:
```
```suggestion
Sulu now uses Webpack 5 to build the administration interface application. To enable this, the following JavaScript dependencies were updated/changed:
```

UPGRADE.md
```diff
@@ -2,6 +2,23 @@
...
+- `webpack-manifest-plugin`: `^5.0.0`
+- `mini-css-extract-plugin`: `^2.7.1`
+- `optimize-css-assets-webpack-plugin` was removed replaced by `css-minimizer-webpack-plugin`: `^4.2.2`
+- `clean-webpack-plugin` was removed and replaced by `clean: true` webpack output option
+- `webpack-clean-obsolete-chunks` was removed and replaced by `clean: true` webpack output option
+- `is-email` was removed and replaced by `sulu-admin-bundle/utils/Email/validateEmail` method 
+
+If you created custom admin components you need to upgrade also
+that components. The core js will automatically be update
+via the [update build](https://docs.sulu.io/en/latest/upgrades/upgrade-2.x.html) command.
```
niklasnatter, line 20: If you have integrated custom JavaScript components into the administration interface,
you might need to adjust your components to be compatible with the updated dependencies.
If you have not integrated custom JavaScript code, you project is adjusted automatically by the
[update build](https://docs.sulu.io/en/latest/upgrades/upgrade-2.x.html) command.

```suggestion
If you have integrated custom JavaScript components into the administration interface,
you might need to adjust your components to be compatible with the updated dependencies.
If you have not integrated custom JavaScript code, you project is adjusted automatically by the
[update build](https://docs.sulu.io/en/latest/upgrades/upgrade-2.x.html) command.
```

package.json
niklasnatter: maybe it was used by `react-styleguidist`? 🤔

src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/blockCollection.scss
```diff
@@ -42,7 +42,7 @@
 
 .selectMultipleButtonContainer {
     display: flex;
-    justify-content: end;
```
niklasnatter, line 45: was this change intentional?

alexander-schranz: yes autoprefixer did complain that `flex-end` is supported better then `end` and has same effect and did output a warning message about this line.

webpack.config.js
```diff
@@ -31,43 +30,47 @@ module.exports = (env, argv) => { // eslint-disable-line no-undef
...
-            new CleanObsoleteChunksPlugin(),
             new webpack.DefinePlugin({
                 SULU_ADMIN_BUILD_VERSION: JSON.stringify(suluVersion),
             }),
         ],
+        optimization: {
+            minimizer: [
+                // For webpack@5 you can use the `...` syntax to extend existing minimizers
+                //      see: https://webpack.js.org/plugins/css-minimizer-webpack-plugin/
+                `...`, // eslint-disable-line quotes
```
niklasnatter:
```suggestion
                '...', // extend existing minimizers: https://webpack.js.org/plugins/css-minimizer-webpack-plugin/
```

package.json
niklasnatter: okay i think this was a leftover from a previous ckeditor configuration that was copied from their documentation: https://ckeditor.com/docs/ckeditor5/latest/installation/advanced/alternative-setups/integrating-from-source.html

webpack.config.js
niklasnatter: this is nice 🥳 i think the reason we dont need it anymore is because we configure the `MiniCssExtractPlugin` with a relative `filename` option

package.json
alexander-schranz: yes did see it in the ckeditor and I was thinking that I already once had that case. That I tried to use the style-loader but did then come to the conclution that it isn't compatible or require for our setup.

webpack.config.js
```diff
@@ -31,43 +30,47 @@ module.exports = (env, argv) => { // eslint-disable-line no-undef
...
-            publicPath,
+            clean: true,
+            path: path.resolve(projectRootPath, publicDir, outputPath),
+            filename: '[name].[chunkhash].js',
         },
         stats: 'minimal',
         performance: {
             hints: false,
         },
         devtool: argv.mode === 'development' ? 'eval-source-map' : 'source-map',
```
niklasnatter, line 53: the build process works for me and the result looked good during my testing 🥳 
unfortunately, it looks like our `npm run watch` task does not work anymore. right now, i see the changes only after restarting the watch task

niklasnatter: the output of my console indicates that the change was detected, but no changes are written to the filesystem if i change the sulu source code

niklasnatter: okay i suspect that this is somehow caused by the location of the code of Sulu 😢 if i change the code of the `assets/admin/app.js` file, the watch task works as expected. maybe files are not reloaded if they are located within the `node_modules` folder?

niklasnatter: okay it looks like we are running into this issue: https://github.com/webpack/webpack/issues/11612
adding the following to the configuration fixes the behaviour of the watch task:
```
snapshot: {
    // detect changes in "node_modules/@sulu": https://github.com/webpack/webpack/issues/11612
    managedPaths: [], 
},
```

niklasnatter (approved):
🥳

alexander-schranz:
🎉

PR #6692, opened 2022-07-10
Add configuration to force two factor authentication

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? | yes
| BC breaks? | no 
| Deprecations? | no  
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Add configuration to force two factor authentication.

#### Why?

This way a agency can force two factor authentication for there all users or there users.


config/packages/security.yaml
```diff
@@ -85,6 +85,9 @@ sulu_security:
         enabled: true
     password_policy:
         enabled: true
+    two_factor:
```
niklasnatter, line 88: would call this `two_factor_authentication`

config/packages/security.yaml
```diff
@@ -85,6 +85,9 @@ sulu_security:
...
+        force:
+            pattern: '/@sulu\.io$/'
```
niklasnatter, line 90: not sure if i would place this in the `force` section. maybe it makes sens to configure it even if 2fa is not mandatory?

config/packages/security.yaml
```diff
@@ -85,6 +85,9 @@ sulu_security:
         enabled: true
     password_policy:
         enabled: true
+    two_factor:
+        force:
```
niklasnatter, line 89: think i would call this `mandatory` and maybe make it a boolean?

config/packages/security.yaml
alexander-schranz: @niklasnatter it allows the folowing:

```yaml
security:
    two_factor:
        force:
            pattern: '/@sulu\.io$/'

when@dev:
    security:
        two_factor:
            force:
                enabled: false
```

config/packages/security.yaml
alexander-schranz: it can be used as a boolean to activate it for all via:

```yaml
security:
   two_factor:
       force: true
```

But to limit it to only some email adresses it can be done via:


```yaml
security:
   two_factor:
       force:
           pattern: '/@sulu\.io$/'
```

alexander-schranz: but I'm fine with go with `mandatory` instead of `force`. But think force is more representing the behaviour. The behaviour does not make any field mandotory, it forces that always a 2fa for this account is active.

src/Sulu/Bundle/SecurityBundle/Metadata/TwoFactorFormMetadataVisitor.php
```diff
@@ -57,9 +68,18 @@ public function visitFormMetadata(FormMetadata $formMetadata, string $locale, ar
                 if ($name && !\in_array($name, $this->twoFactorMethods)) {
                     unset($methods[$key]);
                 }
+
+                if (!$name && null !== $this->twoFactorForcePattern) {
```
niklasnatter, line 72: i am not able to understand what this does 🙈  can you explain it to me?

src/Sulu/Bundle/SecurityBundle/Metadata/TwoFactorFormMetadataVisitor.php
```diff
@@ -57,9 +68,18 @@ public function visitFormMetadata(FormMetadata $formMetadata, string $locale, ar
...
+                    /** @var User|null $user */
+                    $user = $this->security?->getUser();
```
niklasnatter, line 74: not sure if this breaks caching for ort metadata endpoint? 🤔

alexander-schranz: good point we already have a `$formMetadata->setCacheable(false);` for this usecase.

src/Sulu/Bundle/SecurityBundle/Metadata/TwoFactorFormMetadataVisitor.php
alexander-schranz: It removes the `empty` value from the select. So `None` is not longer available for users which are forced to use 2fa. 

Keep in mind the `select` value is called `name` that is why it is `!$name` and why above `$aneM` is used for in_array $this->twoFactorMethods.

config/packages/security.yaml
alexander-schranz: would keep `two_factor` to be consistent to the `scheb/2fa` bundle which is configured above also via `two_factor:`

config/packages/security.yaml
chirimoya: keep `force`

chirimoya (approved):

PR #6711, opened 2022-07-18
Setup @testing-library/react as replacement for enzyme

niklasnatter:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Related issues/PRs | #6343
| License | MIT

#### What's in this PR?

This PR sets up the `@testing-library/react` package which should be used instead of `enzyme` in the future. Additionally, the pull request adjusts the test cases for the `ArrowMenu` component to use `@testing-library/react` instead of `enzyme`.

I found the following resource helpful for getting familiar with the `@testing-library/react` package:
- https://chariotsolutions.com/blog/post/migrating-from-enzyme-to-react-testing-library/
- https://testing-library.com/docs/react-testing-library/migrate-from-enzyme
- https://kentcdodds.com/blog/common-mistakes-with-react-testing-library

#### Why?

Unfortunately, `enzyme` does not support `react@18` and it looks like it never will: https://dev.to/wojtekmaj/enzyme-is-dead-now-what-ekl
Because of this, we should start migrating away from `enzyme` and use `@testing-library/react` in the future.


src/Sulu/Bundle/AdminBundle/Resources/js/components/ArrowMenu/tests/Action.test.js
```diff
@@ -1,43 +1,53 @@
...
 });
 
-test('Clicking the Action should call the right handler', () => {
+test('Clicking the Action should call the right handler', async() => {
     const clickHandler = jest.fn();
+    render(<Action onClick={clickHandler}>My Action</Action>);
 
-    const action = mount(<Action onClick={clickHandler}>My Action</Action>);
+    const user = userEvent.setup();
+    await user.click(screen.getByText('My Action'));
```
niklasnatter, line 30: user interactions are asynchronous in `@testing-library/react`. using the `.then()` syntax will be verbose, especially if there are more user interactions in a test case. therefore i decided to use the `await` syntax here (it is used in the `@testing-library/react` documentation too).

alexander-schranz (approved):

PR #6712, opened 2022-07-18
Use if statement in index.php

szepeviktor:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? | no 
| BC breaks? | no 
| Deprecations? | no 

| License | MIT

#### What's in this PR?

Please see comments.

#### Why?

Clean code.

### DIFF
--- a/public/index.php
+++ b/public/index.php
@@ -15,7 +15,9 @@ use Symfony\Component\Dotenv\Dotenv;
 use Symfony\Component\ErrorHandler\Debug;
 use Symfony\Component\HttpFoundation\Request;
 
-\defined('SULU_MAINTENANCE') || \define('SULU_MAINTENANCE', \getenv('SULU_MAINTENANCE') ?: false);
+if (!\defined('SULU_MAINTENANCE')) {
+    \define('SULU_MAINTENANCE', \getenv('SULU_MAINTENANCE'));
+}
 
 // maintenance mode
 if (SULU_MAINTENANCE) {


public/index.php
```diff
@@ -15,7 +15,9 @@
 use Symfony\Component\ErrorHandler\Debug;
 use Symfony\Component\HttpFoundation\Request;
 
-\defined('SULU_MAINTENANCE') || \define('SULU_MAINTENANCE', \getenv('SULU_MAINTENANCE') ?: false);
+if (!\defined('SULU_MAINTENANCE')) {
```
szepeviktor, line 18: Logical operators are for booleans only.

public/index.php
```diff
@@ -15,7 +15,9 @@
...
+    \define('SULU_MAINTENANCE', \getenv('SULU_MAINTENANCE'));
```
szepeviktor, line 19: `getenv` returns false by default.
https://www.php.net/manual/en/function.getenv.php

szepeviktor: Should it be `(bool)\getenv('SULU_MAINTENANCE')` instead?

szepeviktor: Going lazy unfriendly/strict would be `\getenv('SULU_MAINTENANCE') === 'true'`

alexander-schranz: at the end currently it is set if not empty it will be in Maintainance mode. all other things would be currently a bc break.

public/index.php
alexander-schranz: I think at the end it results in the same and I don't see why it is important here to split this at current state into multiple lines. Would more discuss if envs are still the way here to go here.

alexander-schranz (commented):
Thx for your input, at current state I don't see any need to change this or split it into multiple lines.

PS: this file is actually not used for projects just for internal tests the real project `public/index.php` lives in the `sulu/skeleton` repository.

Still thank you for your input here.

public/index.php
szepeviktor: > why it is important here to split this at current state into multiple lines.

`||` is a operator for booleans. It is a _trick_ to use it this way.
Trick === Unsustainable.

PR #6714, opened 2022-07-19
Fix webspace name with special characters in it

### LINKED ISSUE #6706: Webspace name doesn't properly encode all characters
| Q | A
| --- | ---
| Sulu Version | 2.5.0
| PHP Version | 8.1.7
| DB Version | mysql-server:8.0
| Browser Version | 101.0.1

# Actual Behavior

When I configure a webspace name to contain special characters. Some of the special characters like the umlauts work but things like the `'` do not encode correctly.
webspace

# Expected Behavior
When I configure a webspace in the XML file I expect the name to be displayed correctly.

# Steps to Reproduce
Create an empty sulu project and configure the webspace's name like so.
```xml
<webspace>
    <name>Günter's Homepage</name>
    <key>homepage</key>
</webspace>
```

# Possible Solutions
I don't know if the problem is in the storing of the data but when I checked it it looks like the data was incorrect when loaded from the database so it is probably not a frontend issue.

mamazu:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #6706
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

This PR will fix the generated code for the webspace collection cache.

#### Why?

If you configure a name in the XML file you also want to have it be displayed the same on the website. For an example see #6706

You can't just use the raw filter because this will create an issue with the surrounding code resulting in something like this:
```php
$webspace->setName('Tim's Webspace');
```

#### Example Usage

Just like before and now you can also have apostrophies in your webspace name.

#### To Do

- [x] Create a documentation PR (not needed)
- [x] Add breaking changes to UPGRADE.md (not needed)


mamazu:
Codestyle is broken but it can't be my class since this isn't even a proper class. :D

src/Sulu/Component/Webspace/Resources/skeleton/WebspaceCollectionClass.php.twig
```diff
@@ -32,7 +32,7 @@ class {{ cache_class }} extends {{ base_class }}
         // new webspace
         $webspace = new Webspace();
         $webspace->setKey('{{ webspace.key }}');
-        $webspace->setName('{{ webspace.name }}');
+        $webspace->setName(html_entity_decode('{{ webspace.name }}'));
```
alexander-schranz: We should write the properties already correctly into this cache file.

The problem here currently is the "Twig" autoescaping. Which I think we should disable and go with a "json_encode" solution here:

https://twigfiddle.com/d3hoih

https://twigfiddle.com/d3hoih

alexander-schranz (changes requested):
Good catch. Some small comment.

src/Sulu/Component/Webspace/Resources/skeleton/WebspaceCollectionClass.php.twig
mamazu: Yes this makes it easier. I have changed it to JSON now.

src/Sulu/Component/Webspace/Resources/skeleton/WebspaceCollectionClass.php.twig
```diff
@@ -32,7 +32,7 @@ class {{ cache_class }} extends {{ base_class }}
         // new webspace
         $webspace = new Webspace();
         $webspace->setKey('{{ webspace.key }}');
-        $webspace->setName(html_entity_decode('{{ webspace.name }}'));
+        $webspace->setName(json_decode({{ webspace.name|json_encode }}));
```
alexander-schranz: json_decode is not required here:

```suggestion
        $webspace->setName({{ webspace.name|json_encode|raw }});
```

mamazu: You're right. :D

src/Sulu/Component/Webspace/Resources/skeleton/WebspaceCollectionClass.php.twig
```diff
@@ -32,7 +32,7 @@ class {{ cache_class }} extends {{ base_class }}
         // new webspace
         $webspace = new Webspace();
         $webspace->setKey('{{ webspace.key }}');
-        $webspace->setName('{{ webspace.name }}');
+        $webspace->setName({{ webspace.name|json_encode }});
```
alexander-schranz:
```suggestion
        $webspace->setName({{ webspace.name|json_encode|raw }});
```

raw is required or as written in the comment autoescape false ;)

alexander-schranz:
Can you add a testcase for your case?

mamazu:
Where would be the best way to put it?

mamazu:
And what about the other code that is failing the pipeline?

alexander-schranz:
We sadly did not yet found out why the PHP 8.0 tests in 90% run into a timeout but work without problems locally. Funfact it seems always be that there is atleast one CI whatever it is which just like to go into timeout instead of running finish a job. Still there are PR where all runs without any problems.

The `Test application / PHP Lint (pull_request)` can be ignored. If it doesn't show anything related to your pull request. We mostly get new versions of phpstan or php-cs-fixer which can make the CI fail. But that is a job we are doing and contributers shoud focus on the PR itself :)

alexander-schranz:
For the test I would create a unit test for the `PhpWebspaceCollectionDumper` class.

Somewhere here: https://github.com/sulu/sulu/tree/2.4/src/Sulu/Component/Webspace/Tests/Unit/Manager

By creating a `Dumper` namespace.

You can create the given `WebspaceCollection` by creating one via:

```php
$webspaceCollection = new WebspaceCollection();
// and add then different webspace
$webspace = new Webspace();
$webspaceCollection->add(...);
```

Call the dumper then and after it you `require` the generated collection and call the getter to test if you get the correct name.

alexander-schranz:
@mamazu looks good. Can we rebase this branch to the `2.4` and set the target branch to 2.4? 

Copy & Paste rebase script:

```bash
git clone git@github.com:mamazu/sulu.git
cd sulu
git checkout webspace_name_bug

git fetch sulu

git rebase -i sulu/2.5 --onto sulu/2.4
# close shown file

git push origin webspace_name_bug --force # force required in this case
```

mamazu:
Should be done.

src/Sulu/Component/Webspace/Tests/Unit/Manager/PhpWebspaceCollectionDumperTest.php
```diff
@@ -0,0 +1,213 @@
...
+        $this->setWebspaces($webspaceRefs);
+        $this->setPortals($portalRefs);
+        $this->setPortalInformations($portalInformationRefs);
+    }
+}
+
+
+';
+
+        $this->assertEquals($expected, $string);
```
alexander-schranz:
```suggestion
        $this->assertSame($expected, $string);
```

src/Sulu/Component/Webspace/Tests/Unit/Manager/PhpWebspaceCollectionDumperTest.php
```diff
@@ -0,0 +1,213 @@
...
```
alexander-schranz:
```suggestion
        $this->assertSame($expected, $string);
```

alexander-schranz:
Did quickly replace the assertEquals with assertSame as we are trying to use typesafe phpunit methods where possible.

mamazu:
Yeah, old habits hardly die with the old method. :D

src/Sulu/Component/Webspace/Tests/Unit/Manager/PhpWebspaceCollectionDumperTest.php
```diff
@@ -0,0 +1,213 @@
...
+use Sulu\Component\Webspace\Manager\WebspaceCollection;
+
+class PhpWebspaceCollectionDumperTest extends TestCase
+{
+    use ProphecyTrait;
+
+    /** @var WebspaceCollection|MockObject */
+    private $webspaceCollection;
+
+    private PhpWebspaceCollectionDumper $webspaceDumper;
```
alexander-schranz: looks like php 7.2 does not like this:

```suggestion
     /**
      * @var PhpWebspaceCollectionDumper
      */
    private $webspaceDumper;
```

src/Sulu/Component/Webspace/Tests/Unit/Manager/PhpWebspaceCollectionDumperTest.php
```diff
@@ -0,0 +1,213 @@
...
+use Prophecy\PhpUnit\ProphecyTrait;
+use Sulu\Component\Webspace\Manager\Dumper\PhpWebspaceCollectionDumper;
+use Sulu\Component\Webspace\Manager\WebspaceCollection;
+
+class PhpWebspaceCollectionDumperTest extends TestCase
+{
+    use ProphecyTrait;
+
+    /** @var WebspaceCollection|MockObject */
+    private $webspaceCollection;
```
alexander-schranz: Should be:

```suggestion
    /**
     * @var ObjectProphecy<WebspaceCollection>
     */
    private $webspaceCollection;
```

ObjectProphecy need to be imported.

src/Sulu/Component/Webspace/Tests/Unit/Manager/PhpWebspaceCollectionDumperTest.php
```diff
@@ -0,0 +1,213 @@
...
+namespace Sulu\Component\Webspace\Tests\Unit\Manager;
+
+use PHPUnit\Framework\TestCase;
+use Prophecy\PhpUnit\ProphecyTrait;
+use Sulu\Component\Webspace\Manager\Dumper\PhpWebspaceCollectionDumper;
+use Sulu\Component\Webspace\Manager\WebspaceCollection;
+
+class PhpWebspaceCollectionDumperTest extends TestCase
+{
+    use ProphecyTrait;
```
alexander-schranz: Think this is not required on the 2.4 branch / it does not exist on it. I will add it when merging it back to 2.5.

alexander-schranz (changes requested):
CI seems to fail I added comments to the line failing:

src/Sulu/Component/Webspace/Tests/Unit/Manager/PhpWebspaceCollectionDumperTest.php
mamazu: Good old php7.2 :D

alexander-schranz:
@mamazu Thank you!

PR #6715, opened 2022-07-20
Add better error handling to the block property for non array values

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | sulu/sulu-docs#729

#### What's in this PR?

It contains a better error message for handling setting the value for `BlockProperty`s. This also adds type hints to the function.

#### Why?

Let's say you have a document where the property abc contains a block with a custom type `button`. If you want to bind data to the page your array with the fixture data would look something like this.
```php
$document->bind([
   'abc' => ['type' => 'button', 'value' => 10], 
   // ...more properties
]);
```

But if you put a non-array value in there like this:
```php
$document->bind([
   'abc' => 10
   // ...more properties
]);
```
the following error message will occur:
```text
First argument to the array_keys function must be an array int given.
```
Which at first glance tells you nothing, just by inspecting the stack trace of the error you see that it is inside the block property's `doSetValue` Function trying to set a value.

The new error message reads something like this:
```text
Expected array containing a block configuration but got "10" at property: abc
```

#### To Do

### DIFF
--- a/src/Sulu/Component/Content/Compat/Block/BlockProperty.php
+++ b/src/Sulu/Component/Content/Compat/Block/BlockProperty.php
@@ -73,6 +73,8 @@ class BlockProperty extends Property implements BlockPropertyInterface
      * that the "real" property is updated.
      *
      * TODO: This is very tedious code. It is important to factor this out.
+     *
+     * @param array|PropertyValue|null $value
      */
     public function doSetValue($value)
     {
@@ -83,6 +85,17 @@ class BlockProperty extends Property implements BlockPropertyInterface
 
         if (null == $items) {
             return;
+        }
+
+        if (!is_array($items)) {
+            if (is_object($items)) {
+                $itemsString = 'object of class ' . get_class($items);
+            } else {
+                $itemsString = sprintf('"%s"', var_export($items, true));
+            }
+            throw new \InvalidArgumentException(
+                sprintf('Expected block configuration but got %s at property: %s', $itemsString, $this->getName())
+            );
         }
 
         // check value for single value
--- a/src/Sulu/Component/Content/Compat/Block/BlockProperty.php
+++ b/src/Sulu/Component/Content/Compat/Block/BlockProperty.php
@@ -87,14 +87,9 @@ class BlockProperty extends Property implements BlockPropertyInterface
             return;
         }
 
-        if (!is_array($items)) {
-            if (is_object($items)) {
-                $itemsString = 'object of class ' . get_class($items);
-            } else {
-                $itemsString = sprintf('"%s"', var_export($items, true));
-            }
+        if (!\is_array($items)) {
             throw new \InvalidArgumentException(
-                sprintf('Expected block configuration but got %s at property: %s', $itemsString, $this->getName())
+                \sprintf('Expected block configuration but got "%s" at property: "%s"', \get_debug_type($items), $this->getName())
             );
         }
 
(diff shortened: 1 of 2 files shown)


src/Sulu/Component/Content/Compat/Block/BlockProperty.php
```diff
@@ -85,6 +87,17 @@ public function doSetValue($value)
...
 
+        if (!\is_array($items)) {
+            if (\is_object($items)) {
+                $itemsString = 'object of class ' . \get_class($items);
+            } else {
+                $itemsString = \sprintf('"%s"', \var_export($items, true));
+            }
+            throw new \InvalidArgumentException(
+                \sprintf('Expected block configuration but got %s at property: %s', $itemsString, $this->getName())
+            );
```
alexander-schranz: We should keep this simple:

```suggestion
            throw new \InvalidArgumentException(
                \sprintf('Expected block configuration but got "%s" at property: "%s"', \get_debug_type($itemsString), $this->getName())
            );
```

src/Sulu/Bundle/SecurityBundle/EventListener/LogoutEventSubscriber.php
```diff
@@ -39,7 +39,7 @@ public function onLogout(LogoutEvent $logoutEvent): void
         $adminUrl = $this->urlGenerator->generate('sulu_admin');
         $request = $logoutEvent->getRequest();
 
-        if (!\str_starts_with($request->getPathInfo(), $adminUrl)) {
+        if (!str_starts_with($request->getPathInfo(), $adminUrl)) {
```
alexander-schranz: This `\` is expected in our code style:

```suggestion
        if (!\str_starts_with($request->getPathInfo(), $adminUrl)) {
```

alexander-schranz (changes requested):
Some small change request.

src/Sulu/Component/Content/Compat/Block/BlockProperty.php
mamazu: Does this also work in production since this is a debug function.

src/Sulu/Component/Content/Tests/Unit/Block/BlockPropertyTest.php
```diff
@@ -67,4 +68,24 @@ public function testGetIsMultiple($minOccurs, $maxOccurs, $result)
...
+
+        $blockProperty = new BlockProperty('block', [], 'test');
+        $blockProperty->doSetValue($value);
+    }
+
+    public function provideSetInvalidValue(): array
+    {
+        return [
+            'invalid int' => [10, 'Expected block configuration but got "10" at property: block'],
+            'invalid object' => [new Webspace(), 'Expected block configuration but got object of class Sulu\Component\Webspace\Webspace at property: block'],
```
alexander-schranz: Tests need to be adopted to the updated code.

src/Sulu/Component/Content/Compat/Block/BlockProperty.php
alexander-schranz: Yes it is also available on prod php version aslong it is PHP 8.0. But as this is more an improvement and not a bugfix its fine to be merged this way into the 2.5 branch.

mamazu: Ah a new php 8 feature. Good to know.

mamazu:
Should be fixed now.

alexander-schranz:
@mamazu Thank you!

PR #6717, opened 2022-07-21
Using LCID syntax for setting request locale

mamazu:
| Q | A
| --- | ---
| Bug fix? | yes?
| New feature? | no
| BC breaks? | yes?
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

In Symfony the standard for locales is the LCID format aka `en_US` or `it_IT`. This is now also the format that Sulu sets the request locale to.

#### Why?

If you want to render another controller inside of twig you can do this like so:
```twig
render(controller('controller_service_name'));
```
This then also starts the request analyser that processes the request. If you then proceed to check the locale after rendering said controller it will not be the same as going in. Something like `de_DE` will turn into `de_de`. This should be acceptable as well. But if you try to render translations with this non-standard locale Symfony might not find it and fall back to the application's default locale being `en_US` in most cases.

#### Example Usage

This processor isn't really used anywhere. If you want to recreate the problems just create a Webspace with a non English default locale render a controller inside twig and see the locale change.

#### To Do

I don't know if this a breaking change since people might rely that the format is that way now.

### DIFF
--- a/src/Sulu/Component/Webspace/Analyzer/Attributes/PortalInformationRequestProcessor.php
+++ b/src/Sulu/Component/Webspace/Analyzer/Attributes/PortalInformationRequestProcessor.php
@@ -11,6 +11,7 @@
 
 namespace Sulu\Component\Webspace\Analyzer\Attributes;
 
+use Sulu\Component\Localization\Localization;
 use Sulu\Component\Webspace\Analyzer\RequestAnalyzerInterface;
 use Sulu\Component\Webspace\PortalInformation;
 use Symfony\Component\HttpFoundation\Request;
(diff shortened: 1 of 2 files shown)


src/Sulu/Component/Webspace/Analyzer/Attributes/PortalInformationRequestProcessor.php
```diff
@@ -52,7 +53,7 @@ public function process(Request $request, RequestAttributes $requestAttributes)
         if ($attributes['localization']) {
             $localization = $attributes['localization'];
             $attributes['locale'] = $localization->getLocale();
-            $request->setLocale($localization->getLocale());
+            $request->setLocale($localization->getLocale(Localization::LCID));
```
alexander-schranz, line 56: Sadly we can not change this as it would be a bc break for all accessing it currently and doing checks on it.

alexander-schranz: If I understand you correctly I think at current state you need to use `messages.de_de.yaml` instead of `messages.de_DE.yaml` for translations. Because I think translations are currently the only services having a fallback locale?

mamazu: Correct.

mamazu:
We fixed the problem in our project by just not using the controller function. This is still an issue though.

PR #6720, opened 2022-07-26
Migrate enzyme Block tests to React Testing Library

b3h3m0th:
#### What's in this PR?

Block component migrations enzyme -> RTL


src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/tests/__snapshots__/SortableBlock.test.js.snap
```diff
@@ -66,10 +66,12 @@ exports[`Render expanded sortable, non-collapsable block with types 1`] = `
         class="types"
       >
         <div
+          aria-label="select"
```
alexander-schranz: We should not add here a aria-label. At current state we should keep the components as they are. And use `classes` and `>` selector to check for specific component. to exist.

```suggestion
```

src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/tests/__snapshots__/SortableBlock.test.js.snap
```diff
@@ -66,10 +66,12 @@ exports[`Render expanded sortable, non-collapsable block with types 1`] = `
...
           class="select"
           role="none"
         >
           <button
+            aria-label="button"
```
alexander-schranz: same as above.

```suggestion
```

src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/DisplayValue.js
```diff
@@ -51,6 +51,7 @@ export default class DisplayValue extends React.PureComponent<Props> {
 
         return (
             <button
+                aria-label="button"
```
alexander-schranz: same as above this will hurt our accessibility and should not be done.


```suggestion
```

src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/Select.js
```diff
@@ -283,6 +283,7 @@ class Select<T> extends React.Component<Props<T>> {
 
         return (
             <div
+                aria-label="select"
```
alexander-schranz: same this should be removed:

```suggestion
```

src/Sulu/Bundle/AdminBundle/Resources/js/components/Block/tests/Block.test.js
```diff
@@ -20,96 +20,100 @@ test('Render an expanded block with a multiple types', () => {
...
 test('Do not show type dropdown if only a single type is passed', () => {
-    const block = shallow(
+    render(
         <Block expanded={true} onCollapse={jest.fn()} onExpand={jest.fn()} types={{'type': 'Type'}}>
             Some block content
         </Block>
     );
 
-    expect(block.find('SingleSelect')).toHaveLength(0);
+    const element = screen.queryByLabelText('select');
```
alexander-schranz: We should go herre with `getByText` and check for the `displayValue` of the select.

src/Sulu/Bundle/AdminBundle/Resources/js/components/Block/tests/Block.test.js
```diff
@@ -133,7 +138,12 @@ test('Changing the type should call the onTypeChange callback', () => {
         </Block>
     );
 
-    block.find('SingleSelect').simulate('change', 'type2');
+    const selectButton = screen.queryByLabelText('button');
```
alexander-schranz: we should go here with `getByText` or `getByRole` as there are multiple buttons we can go with `getAllByRole` or first select the correct parent. we should not add a `aria-label="button"` as that would hurt accessibility.

alexander-schranz (changes requested):
Some small change request about the added aria-label. Which we should not doing. We should keep the components currently as they are and not add aria-* attributes here as that would hurt the accessibility of the components.

src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/tests/__snapshots__/SortableBlock.test.js.snap
luca-rath: @alexander-schranz I disagree here a bit ... It definitely doesn't make sense to add aria-label to lots of tags (including this specific example here), but if there are buttons without a button text with just an icon, we should add a useful `aria-label`, which describes, what the button does. That label should of course also be translated. Adding a useful aria-label definitely doesn't hurt and makes writing some tests a lot easier.

luca-rath: But yes, if the button already has a button text, it's definitely not necessary, because jest allows to select buttons by their text easily

alexander-schranz: @b3h3m0th if you run into any way where you can currently not select or a aria-label would make selection easier for the tests. Add a comment there so we can have a look at together and discuss how to forward with that specific tests. Aria-Label would require always come from the backend translated, also im some other cases Aria-role or tag selection (button, ...) could be used. Most of the selection but should be done via Text if possible as that would make it most future proof also when we are refractoring the html and accessibility.

b3h3m0th: Sure, make sense. Thanks!

src/Sulu/Bundle/AdminBundle/Resources/js/components/Block/tests/Block.test.js
b3h3m0th: This regards the discussion of the `aria-label` problem. Sometimes it's hard to query items because (as you mentioned) there is literally no relevant information about a DOM element for RTL. If that is the case one would have to navigate though the tree (e.g. using `.parentElement`, `.children[0]`) which is not allowed by ESlint. Even simply accessing elements by `.getElementsbyClassName()` is forbidden by the linter. The example below passes the test but not the linter. Any suggestions?
image

alexander-schranz: I'm personally fine to add in the test a `// eslint-disable-next-line the-violating-rule` if it make the test easier to understand by using this selector. /cc @luca-rath @niklasnatter

alexander-schranz: But as said you can also add a comment and we will check how we can improve it to make this things easier.

src/Sulu/Bundle/AdminBundle/Resources/js/components/Block/tests/Block.test.js
```diff
@@ -120,9 +126,10 @@ test('Changing the type should call the onTypeChange callback', () => {
         type1: 'Type 1',
         type2: 'Type 2',
     };
-    const block = shallow(
+
+    render(
         <Block
-            activeType="type1"
+            activeType={Object.keys(types)[0]}
```
alexander-schranz: Would keep here `"type1"` tests are fine to be repetitive that make them better readable and understandable. Else you need always first check okay what is inside `types` to know what here is given.

src/Sulu/Bundle/AdminBundle/Resources/js/components/Block/tests/Block.test.js
```diff
@@ -133,7 +140,12 @@ test('Changing the type should call the onTypeChange callback', () => {
         </Block>
     );
 
-    block.find('SingleSelect').simulate('change', 'type2');
+    const selectButton = screen.queryByText(types.type1);
```
alexander-schranz: same as above use here the a string make it better understandable

src/Sulu/Bundle/AdminBundle/Resources/js/components/Block/tests/Block.test.js
```diff
@@ -133,7 +140,12 @@ test('Changing the type should call the onTypeChange callback', () => {
...
+    fireEvent.click(selectButton);
+
+    const typeButton = screen.queryByText(types.type2);
```
alexander-schranz: same as above use here the a string make it better understandable

alexander-schranz (changes requested):

alexander-schranz (approved):

PR #6721, opened 2022-07-26
Migrate enzyme Button tests to React Testing Library

b3h3m0th:
#### What's in this PR?

Button test migrations from enzyme to RTL


src/Sulu/Bundle/AdminBundle/Resources/js/components/Button/tests/__snapshots__/Button.test.js.snap
```diff
@@ -1,129 +1,222 @@
...
+        type="button"
+      >
+        <span
+          aria-label="su-angle-down"
+          class="su-angle-down dropdownIcon"
+        />
+      </button>
+    </div>
+  </body>,
+  "container": <div>
```
alexander-schranz: This looks strange look like we are creating here a false snapshot. I think the snapshot is done on the false element.

src/Sulu/Bundle/AdminBundle/Resources/js/components/Button/tests/Button.test.js
```diff
@@ -1,72 +1,90 @@
...
 });
 
 test('Should render with skin secondary and dropdown icon', () => {
-    expect(render(<Button showDropdownIcon={true} skin="secondary" />)).toMatchSnapshot();
+    const {container} = render(<Button showDropdownIcon={true} skin="secondary" />);
+    expect(container).toMatchSnapshot();
 });
 
 test('Should render with skin link and dropdown icon', () => {
     expect(render(<Button showDropdownIcon={true} skin="link" />)).toMatchSnapshot();
```
alexander-schranz: the snapshot of this tests looks false. I think in react testing library we should not do this on test().toSnapshot instead on render()

alexander-schranz (changes requested):
@b3h3m0th  One small test need to be adopted. Else good work 👍

src/Sulu/Bundle/AdminBundle/Resources/js/components/Button/tests/__snapshots__/Button.test.js.snap
b3h3m0th: You're right. I forgot to migrate the [test you mentioned here](https://github.com/sulu/sulu/pull/6721#discussion_r931481107) which results in printing the whole `render()` return value to the snapshot file.

src/Sulu/Bundle/AdminBundle/Resources/js/components/Button/tests/Button.test.js
b3h3m0th: See [this previous comment](https://github.com/sulu/sulu/pull/6721#discussion_r931492355).

src/Sulu/Bundle/AdminBundle/Resources/js/components/Button/tests/__snapshots__/Button.test.js.snap
alexander-schranz: Yeah think moving the snapshot inside the `test` like its done in the other tests should fix it :)

alexander-schranz (approved):

alexander-schranz:
@b3h3m0th Thx good work 👍

PR #6730, opened 2022-07-28
Migrate enzyme CharacterCounter tests to React Testing Library

b3h3m0th:
#### What's in this PR?

CharacterCount component unit test migrations enzyme --> RTL

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/CharacterCounter/CharacterCounter.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/CharacterCounter/CharacterCounter.js
@@ -23,7 +23,7 @@ export default class CharacterCounter extends React.Component<Props> {
 
         return (
             <label className={charactersLeftLabelClass}>
-                {charactersLeft} {translate('sulu_admin.characters_left')}
+                {charactersLeft + ' ' + translate('sulu_admin.characters_left')}
             </label>
         );
     }
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/CharacterCounter/tests/__snapshots__/CharacterCounter.test.js.snap
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/CharacterCounter/tests/__snapshots__/CharacterCounter.test.js.snap
@@ -1,41 +1,61 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`Should show a negative count 1`] = `
-<label
-  class="characterCounter exceeded"
->
-  -8 sulu_admin.characters_left
-</label>
+<div>
+  <label
+    class="characterCounter exceeded"
+  >
+    -8
+     
+    sulu_admin.characters_left
+  </label>
+</div>
 `;
 
 exports[`Should show a negative count with numbers 1`] = `
-<label
-  class="characterCounter exceeded"
->
-  -1 sulu_admin.characters_left
-</label>
+<div>
+  <label
+    class="characterCounter exceeded"
+  >
+    -1
+     
+    sulu_admin.characters_left
+  </label>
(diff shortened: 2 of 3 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/components/CharacterCounter/tests/__snapshots__/CharacterCounter.test.js.snap
```diff
@@ -1,41 +1,61 @@
...
->
-  -8 sulu_admin.characters_left
-</label>
+<div>
+  <label
+    class="characterCounter exceeded"
+  >
+    -8
+     
+    sulu_admin.characters_left
```
alexander-schranz: @b3h3m0th Do we know why it adds 2 line breaks here instead of just the space? Can not understand it as there seems no breaks in the file itself: https://github.com/sulu/sulu/blob/544a6714d9ea4500e7cfdd5913d31c0193511858/src/Sulu/Bundle/AdminBundle/Resources/js/components/CharacterCounter/CharacterCounter.js#L26

Is this a default react behaviour?

b3h3m0th: That's weird, I don't know either. It's definitely not default react behaviour as you can see in [this sandbox](https://codesandbox.io/s/youthful-shtern-idsxis). Is it possible that RTL does something odd there?

alexander-schranz: Found this issue: https://github.com/testing-library/react-testing-library/issues/53 but the are not targeting how the snapshot is created they seems just try to fix that the selectors still work there.

Maybe we can workaround by using: `{charactersLeft + ' ' + translate('sulu_admin.characters_left')}` in the component currently.

b3h3m0th: Yes that fixed it. I updated the component and it's shapshot in [the following commit](https://github.com/sulu/sulu/pull/6730/commits/8c44a4678cdeb2ebe515209580dc28e2ba63370d).

src/Sulu/Bundle/AdminBundle/Resources/js/components/CharacterCounter/CharacterCounter.js
```diff
@@ -23,7 +23,7 @@ export default class CharacterCounter extends React.Component<Props> {
 
         return (
             <label className={charactersLeftLabelClass}>
```
alexander-schranz, line 25: Note for myself. Check why this is a `label` 🙈

alexander-schranz (approved):

src/Sulu/Bundle/AdminBundle/Resources/js/components/CharacterCounter/tests/__snapshots__/CharacterCounter.test.js.snap
```diff
@@ -5,9 +5,7 @@ exports[`Should show a negative count 1`] = `
   <label
     class="characterCounter exceeded"
   >
-    -8
-     
-    sulu_admin.characters_left
+    -8 sulu_admin.characters_left
```
b3h3m0th, line 8: What do you mean? Doesn't it look correct like this or am I missing something? 🤔

alexander-schranz: all fine just waiting for the CI :)

PR #6738, opened 2022-07-29
Migrate enzyme Email tests to React Testing Library

b3h3m0th:
#### What's in this PR?

Email component unit test migrations enzyme --> RTL


src/Sulu/Bundle/AdminBundle/Resources/js/components/Input/Input.js
```diff
@@ -128,6 +128,7 @@ export default class Input<T: ?string | ?number> extends React.PureComponent<Inp
                     {!loading && icon &&
                         <div className={prependContainerClass}>
                             <Icon
+                                // maybe add an aria-label here?
```
b3h3m0th: @alexander-schranz Is it possible to add an `aria-label` here?

alexander-schranz: There is already a `aria-label` on it rendered by the Icon component. Which should be more removed as it is totally false here. But at current state you can use `aria-label="su-envelope"` selector if there is no other way.

alexander-schranz: Just for your information `onClick` on the Icon was a false decision and ideal case the click would be handled not in a Icon component and so it would be something like this:

```
const iconComponent = <Icon name={icon} className={className} style={iconStyle} />

onIconClick
   ? <button aria-label={iconClickLabel} type="button" onClick={onIconClick}>
       {iconComponent}
   </button> 
   : iconComponent
```

but at current state we can not change it. And you can go on the existing: `aria-label="su-envelope"`

b3h3m0th: Ah, I forgot the Icon component itself already had an `aria-label` on the span.

src/Sulu/Bundle/AdminBundle/Resources/js/components/Email/tests/Email.test.js
```diff
@@ -1,46 +1,63 @@
...
+    expect(container).toMatchSnapshot();
 });
 
 test('Email should not set onIconClick when value is invalid', () => {
     const onChange = jest.fn();
     const onBlur = jest.fn();
-    const email = mount(<Email onBlur={onBlur} onChange={onChange} valid={false} value={null} />);
+    const onIconClickSpy = jest.fn();
+
+    render(<Email onBlur={onBlur} onChange={onChange} onIconClick={onIconClickSpy} valid={false} value={null} />);
```
alexander-schranz: This is not possible as `Email` does not have a `OnIconClick` and that is why `Node Lint` task is failing.

Maybe we can check that nothing is changed on window.location.assign when icon is clicked. Or we can check if const icon does not have a `onClick` registered on its dom element.

alexander-schranz (changes requested):

src/Sulu/Bundle/AdminBundle/Resources/js/components/Email/tests/Email.test.js
b3h3m0th: Yes, that's a good idea.

src/Sulu/Bundle/AdminBundle/Resources/js/components/Input/Input.js
niklasnatter:
```suggestion
```

niklasnatter (approved):

PR #6752, opened 2022-08-01
Migrate enzyme Masonry tests to React Testing Library

b3h3m0th:
#### What's in this PR?

Masonry component unit test migration enzyme --> RTL


src/Sulu/Bundle/AdminBundle/Resources/js/components/Masonry/tests/__snapshots__/Masonry.test.js.snap
```diff
@@ -1,41 +1,47 @@
...
-    style="margin-bottom:30px"
-  >
-    <div>
-      <img
-        src="http://lorempixel.com/300/200"
-      />
-    </div>
-  </li>
+    <li
+      style="margin-bottom: 30px; position: absolute; left: 0px; top: 0px;"
```
alexander-schranz, line 10: little bit strange this is added now but looks like react testing library create the snapshot at another timing of component state as previous enzym render method so this change looks fine for me.

alexander-schranz (approved):

PR #6753, opened 2022-08-01
Register list builder services in website context

### LINKED ISSUE #4946: FieldDescriptorFactoryInterface not available in webspace context
| Q | A
| --- | ---
| Bug? | yes
| New Feature? | no
| Sulu Version | 2.0.2

#### Actual Behavior

```
Cannot autowire service "App\Controller\Admin\PointController": argument "$fieldDescriptorFactory" of method "__construct()" references interface "Sulu\Component\Rest\ListBuilder\Metadata\FieldDescriptorFactoryInterface" but no such service exists. Did you create a class that implements this interface?
```
### Expected Behavior

working list in admin

#### Steps to Reproduce

http://docs.sulu.io/en/2.0/book/extend-admin.html

#### Possible Solutions

If you have already ideas how to solve the issue, add them here.
(remove this section if not needed)

niklasnatter:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Fixed tickets | fixes https://github.com/sulu/sulu/issues/4946
| Related issues/PRs | https://github.com/sulu/sulu/pull/5942
| License | MIT

#### What's in this PR?

This pull request adjusts the `SuluCoreExtension` to register our list builder services in the website context too.

#### Why?

Because it makes things cumbersome if these services are not available in the website context (you need to register dependant services only in the admin context too). Also, we already closed an issue about this (https://github.com/sulu/sulu/issues/4946) after merging https://github.com/sulu/sulu/pull/5942.

### DIFF
--- a/src/Sulu/Bundle/CoreBundle/DependencyInjection/SuluCoreExtension.php
+++ b/src/Sulu/Bundle/CoreBundle/DependencyInjection/SuluCoreExtension.php
@@ -317,8 +317,7 @@ class SuluCoreExtension extends Extension implements PrependExtensionInterface
             $this->initFields($config['fields_defaults'], $container);
         }
 
-        $this->initListBuilder($container, $loader);
-
+        $loader->load('list_builder.xml');
         $loader->load('expression_language.xml');
         $loader->load('phpcr.xml');
         $loader->load('rest.xml');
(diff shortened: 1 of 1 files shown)


src/Sulu/Bundle/CoreBundle/DependencyInjection/SuluCoreExtension.php
```diff
@@ -317,8 +317,7 @@ public function load(array $configs, ContainerBuilder $container)
             $this->initFields($config['fields_defaults'], $container);
         }
 
-        $this->initListBuilder($container, $loader);
-
+        $loader->load('list_builder.xml');
```
alexander-schranz: @niklasnatter all projects from 2.2 has no AdminBundle register so to make this not crash that projects upgrading from that version we should check if `kernel.bundles` parameter including SuluAdminBundle and then include the `list_builder.xml` only so we are sure the admin bundle parameters are set which are used by the `list_builder.xml`.

Example: https://github.com/sulu/sulu/blob/625b7c595392e109613631c6605f871e1b184918/src/Sulu/Bundle/WebsiteBundle/DependencyInjection/SuluWebsiteExtension.php#L138-L140

alexander-schranz (approved):

PR #6760, opened 2022-08-02
Migrate enzyme Dialog tests to React Testing Library

b3h3m0th:
#### What's in this PR?

Dialog component unit test migration enzyme --> RTL


src/Sulu/Bundle/AdminBundle/Resources/js/components/Dialog/tests/Dialog.test.js
```diff
@@ -8,24 +9,23 @@ jest.mock('../../../utils/Translator', () => ({
...
+        onConfirm={jest.fn()}
+        open={true}
+        title="My dialog title"
+    >
+        <div>My dialog content</div>
+    </Dialog>
     );
 
-    expect(view.find('Dialog > Portal').at(0).render()).toMatchSnapshot();
+    expect(screen.getByText('My dialog content')).toBeInTheDocument();
```
b3h3m0th, line 25: I don't really know how to match a snapshot here because the `<Dialog/>` is rendered inside of a React Portal. I assume that is the reason why the regular container snapshot only outputs `<div/>` instead of a nice component snapshot of the `<Dialog/>`. I have already posted a [question on SO](https://stackoverflow.com/questions/73205954/how-can-i-test-a-modal-inside-of-a-react-portal-using-react-testing-library).

In the meantime I added a test case that checks whether an element with the passed children text is in the document. Neither am I sure if that can replace the snapshot 100% nor if that is a bulletproof solution... probably not.

alexander-schranz: I'm not 100% sure but can you do a snapshot to `document.body`.

alexander-schranz: Maybe this discussion can help yout to find a solution here: https://github.com/testing-library/react-testing-library/issues/62

b3h3m0th: I've already checked out that discussion but probably wasn't reading it carefully enough the first time. Anyway I made it work by snapshotting the `baseElement` instead of the `container` (both are properties returned by `render()`).

PR #6762, opened 2022-08-02
Migrate enzyme Matrix tests to React Testing Library

b3h3m0th:
#### What's in this PR?

Matrix component unit test migration enzyme --> RTL


alexander-schranz:
@b3h3m0th the error in the CI is currently unrelated to your changes. It seems like a new version of `@testing-library/user-event`.

https://github.com/testing-library/user-event/issues/1014

At current state I think you can adopt the `package.json` to:

```json
"@testing-library/user-event": "^14.2.5, <14.4.0",
```

Until the bug is fixed in the `@testing-library/user-event` library. And so you can continue with this and the other components.

b3h3m0th:
@alexander-schranz Re: [`package.json`](https://github.com/sulu/sulu/pull/6762#issuecomment-1203172792). Should I update the `package.json` in the root folder as well or just the one in `sulu/src/Sulu/Bunle/AdminBundle/Resources/js/package.json` ?

alexander-schranz:
@b3h3m0th it would have been to be added as `"^14.2.5, <14.4.0"` on all which did define that package, but looks like a new version with a fix was already released so we can revert it to "^14.2.5"

package.json
```diff
@@ -1,126 +1,126 @@
...
-        "lint:js": "eslint . --cache",
-        "lint:scss": "stylelint src/Sulu/Bundle/*/Resources/js/**/*.scss",
-        "flow": "flow",
-        "test": "jest",
-        "depcruise": "depcruise src/Sulu/Bundle/*/Resources/js -c dependency-cruiser.json",
-        "styleguide": "styleguidist server",
-        "styleguide:build": "styleguidist build",
-        "build": "webpack --mode production",
-        "watch": "webpack --mode development --watch"
+  "name": "sulu-sulu",
```
alexander-schranz: @b3h3m0th please undo this changes of this file.

b3h3m0th: Sorry about that 😬

src/Sulu/Bundle/AdminBundle/Resources/js/components/Matrix/tests/Matrix.test.js
```diff
@@ -162,7 +167,9 @@ test('Changing a value should call onChange ', () => {
         },
     };
 
-    matrix.find(Item).at(3).simulate('click');
+    // eslint-disable-next-line testing-library/no-node-access
+    const item = screen.queryAllByLabelText('su-pen')[1].parentElement;
+    fireEvent.click(item);
```
alexander-schranz, line 172: Is fine at me at current state as we do not have good accessibilty for this component yet we need implement hack via parentElement here.

alexander-schranz (approved):

PR #6767, opened 2022-08-04
Migrate enzyme Number tests to React Testing Library

b3h3m0th:
#### What's in this PR?

Number component unit test migration enzyme --> RTL


src/Sulu/Bundle/AdminBundle/Resources/js/components/Number/tests/Number.test.js
```diff
@@ -1,40 +1,44 @@
...
-    const number = shallow(<Number onChange={onChange} value={undefined} />);
+    render(<Number onChange={onChange} value={0.5} />);
 
-    const event = {};
-    number.find('Input').simulate('change', undefined, event);
+    const input = screen.queryByDisplayValue(0.5);
+    fireEvent.change(input, {target: {value: undefined}});
 
-    expect(onChange).toBeCalledWith(undefined, event);
+    expect(onChange).toBeCalledWith(undefined, expect.anything());
```
b3h3m0th, line 43: Why does this test fail? I just don't understand.

```
    Expected: undefined, Anything

    Number of calls: 0

      41 |     fireEvent.change(input, {target: {value: undefined}});
      42 |
    > 43 |     expect(onChange).toBeCalledWith(undefined, expect.anything());
         |                      ^
      44 | });
      45 |
```

src/Sulu/Bundle/AdminBundle/Resources/js/components/Number/tests/Number.test.js
```diff
@@ -1,40 +1,44 @@
...
 
 test('Number should call onChange with undefined when value is undefined', () => {
     const onChange = jest.fn();
-    const number = shallow(<Number onChange={onChange} value={undefined} />);
+    render(<Number onChange={onChange} value={0.5} />);
 
-    const event = {};
-    number.find('Input').simulate('change', undefined, event);
+    const input = screen.queryByDisplayValue(0.5);
+    fireEvent.change(input, {target: {value: null}});
```
b3h3m0th: Somehow I wasn't able to fire a change event with `value: undefined`. It just doesn't call the callback. Although it definitely should do so because of the `handleChange` logic in the component itself. But the event fired in the test won't even reach the component, at least I couldn't detect it with a `console.log`. Maybe this just doesn't count as *change* by RTL (which would be really weird). I now decided to pass `value: null` to the event, this works as expected. Is this problematic? I would really appreciate any help :)

src/Sulu/Bundle/AdminBundle/Resources/js/components/Number/tests/Number.test.js
```diff
@@ -1,40 +1,44 @@
 // @flow
 import React from 'react';
-import {render, shallow} from 'enzyme';
+import {render, fireEvent, screen} from '@testing-library/react';
```
niklasnatter: would be nice to use the `@testing-library/user-event` package instead of `fireEvent` 🙂 see https://github.com/sulu/sulu/pull/6783 for some examples

src/Sulu/Bundle/AdminBundle/Resources/js/components/Number/tests/Number.test.js
niklasnatter: i think we should use the `clear` method of the `@testing-library/user-event` instead of firing the vent by ourselves 🙂

b3h3m0th: Thanks!

src/Sulu/Bundle/AdminBundle/Resources/js/components/Number/tests/Number.test.js
```diff
@@ -1,40 +1,45 @@
...
-test('Number should call onChange with parsed value', () => {
+test('Number should call onChange with parsed value', async() => {
     const onChange = jest.fn();
-    const number = shallow(<Number onChange={onChange} value={undefined} />);
+    render(<Number onChange={onChange} value={2} />);
 
-    const event = {};
-    number.find('Input').simulate('change', '10.2', event);
+    const input = screen.queryByDisplayValue(2);
+    fireEvent.keyDown(input, {target: {value: '10.2'}});
```
alexander-schranz: Think we have a problem here to go with `10.` to `10.2` /cc @niklasnatter any idea?

niklasnatter: https://github.com/testing-library/user-event/issues/549 recommends to add a wrapper component for these components. i will have a look at that 🙂  
another workaround would be using `paste` instead of `type` to change multiple characters in a single event

niklasnatter: i merged this PR using `paste` and added a helper function to wrap the component in https://github.com/sulu/sulu/pull/6795

Issue #4278, opened 2018-11-15
Ability to browse each snippet type independently


AcelisWeaven:
| Q | A
| --- | ---
| Bug? | no
| New Feature? | yes
| Sulu Version | --
| Browser Version | --

#### Actual Behavior

All snippets can be found on the "Snippets" tab, and can't be filtered.

#### Expected Behavior

Every snippet should be manageable independently. This has been mentioned in #3235, where @trickreich suggested to have each snippet in its own tab, like in SuluArticleBundle. (I don't know how it looks like, never had more than one article type)

In a project on Sulu 1.6, I created a small bundle to change the menu dynamically, depending on my snippet configuration. In my sample implementation, snippet can be grouped by category. Here's an example of a tag that made the snippet appear here:
`<tag name="separated" category="snippet.category.addresses" icon="home"/>`

![screenshot from 2018-11-15 10-25-29](https://user-images.githubusercontent.com/6735195/48560172-86b98b00-e8ed-11e8-8b08-3dca5c67775f.png)

Pros:
- manageable at the snippet level (in the xml file)
- not every snippet have to appear in the menu (if the tag is not here, it'll only be accessible via the snippet menu)

Cons:
- phpcr may have performance issues when filtering by type (slower back-end)
- defining the group icon at the snippet level is not a good idea

#### Possible Solutions

The ability to handle this problem at the snippet level has its limits when you have snippet groups, but it would be fine if you have one tab per snippet.

An hybrid solution could also work. Let's say you add a tag like `<tag name="sulu.snippet.isolate">`: the snippet would appear as its own menu entry.
Now, let's say I have two snippets and I want group them under the same menu entry. This could be done at the configuration level:
```yml
sulu_snippet:
    menu:
        group1:
            icon: 'puzzle'
            snippets: ['snippet1', 'snippet2']
```

Also, the xml tag could be completely left out, and everything could be done in the configuration file.
```yml
sulu_snippet:
    menu:
        snippet1:
            icon: 'home'
        group1:
            icon: 'puzzle'
            snippets: ['snippet2', 'snippet3']
```
This would look like that:
![screenshot from 2018-11-15 16-03-17](https://user-images.githubusercontent.com/6735195/48561349-0a747700-e8f0-11e8-85d0-904c13d1d86f.png)

This is a feature I'd really like to have in Sulu 2.

danrot:
[...]

PR #6773, opened 2022-08-04
Add option to toggle the snippet admin to a list per snippet type

wimvds:
Added a config toggle to switch between the single snippet list view (current) and a list view per snippet type (with a navigation entry per snippet type in the admin sidebar).

| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | 
| Related issues/PRs | #4278
| License | MIT

#### What's in this PR?

A config toggle to switch between the single snippet list view (current) and a list view per snippet type (with a navigation entry per snippet type in the admin sidebar) has been added to the Sulu snippet configuration.

#### Why?

The single list of snippets becomes a burden when you have a few snippet types with a lot of content, it's easier to manage the snippet types separately. 

#### Example Usage

The default admin view for the snippets is the current single list view. To toggle to the new view (with subnavigation per snippet type) you should have the following in your configuration :

```
sulu_snippet:
    admin_view: types_list
```

To revert back either remove the admin_view or set it's value to `single_list`.

A screenshot of the admin when the new view is active :
snippets-nav

#### To Do

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/StructureFormMetadataLoader.php
+++ b/src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/StructureFormMetadataLoader.php
@@ -110,6 +110,8 @@ class StructureFormMetadataLoader implements FormMetadataLoaderInterface, CacheW
             foreach ($webspace->getExcludedTemplates() as $excludedTemplate) {
                 $form->removeForm($excludedTemplate);
             }
+        } elseif (isset($metadataOptions['defaultType'])) {
+            $form->setDefaultType($metadataOptions['defaultType']);
         } elseif (isset($this->defaultTypes[$key])) {
             $form->setDefaultType($this->defaultTypes[$key]);
         }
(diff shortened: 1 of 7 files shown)


wimvds:
I just realized that this can also be done by decorating the StructureFormMetadataLoader and replacing the AdminSnippet service via Symfony configuration... So maybe I could just document how to do that instead (though it's not 100% future proof of course - as it depends on the StructureFormMetadataLoader not changing ;p).

alexander-schranz:
@wimvds Nice work, we need to discuss also with the UX team in which direction we want to go here. Thank you already for the work and implementation you provided.

mario-fehr:
How about a tabview like in the ArticleBundle?

mamazu:
Voting all for it, but I think the tabview could be a problem if you have a lot of snippet types.

alexander-schranz:
Think this is something we could discuss again when we now rewrite the articles and snippets based on Content Bundle. To think what is best or if we want to have possibility todo it by config or not.

alexander-schranz:
This will sure be something we discuss soon with the Article Bundle now merged into 3.0.

Still want to mention: https://github.com/perspeqtive/sulu-snippet-tabs-bundle as a third party bundle by @perspeqtive @markusholstein @alexanderbuerner which provides something similar to Sulu 2.6.

alexanderbuerner:
Thanks for mentioning! Just to point to the correct repository: https://github.com/perspeqtive/sulu-snippet-manager-bundle

PR #6774, opened 2022-08-05
Migrate enzyme Tabs tests to React Testing Library

b3h3m0th:
#### What's in this PR?

Tabs component unit test migration enzyme --> RTL


src/Sulu/Bundle/AdminBundle/Resources/js/components/Tabs/tests/Tabs.test.js
```diff
@@ -1,7 +1,6 @@
 /* eslint-disable flowtype/require-valid-file-annotation */
-import {mount, render, shallow} from 'enzyme';
+import {fireEvent, render, screen} from '@testing-library/react';
```
niklasnatter: would be nice to use the `@testing-library/user-event` package instead of `fireEvent` 🙂 see https://github.com/sulu/sulu/pull/6783 for some examples

src/Sulu/Bundle/AdminBundle/Resources/js/components/Tabs/tests/Tabs.enzyme.test.js
```diff
@@ -0,0 +1,203 @@
...
+});
+
+Object.defineProperty(window, 'getComputedStyle', {
+    value: () => ({
+        paddingLeft: 20.0,
+        paddingRight: 20.0,
+    }),
+});
+
+test('Render a Tabs component with type root', () => {
```
alexander-schranz: What is the issue with this tests?

b3h3m0th: I don't think there is an issue with this test, is there?

niklasnatter: i think @alexander-schranz meant if there there is a reason why we cannot migrate it to rtl it or if you just did not do it yet? 🙂

b3h3m0th: The test at the marked line is migrated already. But yes, the test about collapsing isn't yet because I don't really know how.

src/Sulu/Bundle/AdminBundle/Resources/js/components/Tabs/tests/Tabs.test.js
```diff
@@ -21,169 +21,80 @@ Object.defineProperty(window, 'getComputedStyle', {
...
-            <Tabs.Tab>Tab 3</Tabs.Tab>
-        </Tabs>
     );
 
-    tabs.find('.tab button').at(0).simulate('click');
-    expect(changeSpy).toHaveBeenCalledWith(selectedTabIndex);
+    expect(container).toMatchSnapshot();
 });
 
-test('Clicking on several non- and collapsed tabs', () => {
```
niklasnatter, line 87: we cannot migrate this test easily as it sets properties of the component instance. i think i would propose to remove it for now and investigate how to test this if there is a bug in this part in the code in the future 🙈

niklasnatter: i tried to implement a simpler version of this test as a replacement. unfortunately, the `Tabs` component uses `offsetWidth` to determine the collapsed tabs but `jsdom` does not support `offsetWidth` at all: https://github.com/jsdom/jsdom/issues/135. this means we cannot test if tabs are collapsed because they will never be collapsed in the test environment 😕

alexander-schranz: I think the target of the whole test is to test if at specific window size the dropdown with the rest of tabs is shown. Is that possible in a more simple test or is that already tested in other test case?

alexander-schranz: And it tested if I click one of the dropdown items that then that active element is shown and the other one goes into the dropdown. But If you say it is too much work at current state I'm fine with merging it.

niklasnatter: the problem is that there will never be a dropdown because the calculations of the components depend on `offsetWidth` and `offsetWidth` is not implemented in jsdom (https://github.com/jsdom/jsdom/issues/135) 😕

alexander-schranz: so testing library is a bad choice for testing something which is not server rendered and depends on browser features 🙈

niklasnatter: at least not if the component depends on size calculations. we decided to keep the test case in a `Tabs.enzyme.test.js` for now.

PR #6777, opened 2022-08-05
Migrate enzyme Radio tests to React Testing Library

b3h3m0th:
#### What's in this PR?

Radio component unit test migration enzyme --> RTL


src/Sulu/Bundle/AdminBundle/Resources/js/components/Radio/tests/Radio.test.js
```diff
@@ -1,25 +1,25 @@
 // @flow
-import {shallow, render} from 'enzyme';
+import {fireEvent, render, screen} from '@testing-library/react';
```
niklasnatter: would be nice to use the `@testing-library/user-event` package instead of `fireEvent` 🙂 see https://github.com/sulu/sulu/pull/6783 for some examples

src/Sulu/Bundle/AdminBundle/Resources/js/components/Radio/tests/RadioGroup.test.js
```diff
@@ -1,65 +1,79 @@
 // @flow
-import {mount, render} from 'enzyme';
+import {fireEvent, render, screen} from '@testing-library/react';
```
niklasnatter: would be nice to use the `@testing-library/user-event` package instead of `fireEvent` 🙂 see https://github.com/sulu/sulu/pull/6783 for some examples

alexander-schranz (approved):

PR #6782, opened 2022-08-08
Fix eslint errors in javascript test cases

niklasnatter:
#### What's in this PR?

This pull request fixes some eslint errors in our javascript test cases. Additionally, it disables the `testing-library/no-node-access` rule to keep the effort for migrating enzyme tests as low as possible.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -237,7 +237,8 @@
         "flowtype/sort-keys": "error",
         "babel/semi": "error",
         "jsx-a11y/no-static-element-interactions": "error",
-        "react/button-has-type": "error"
+        "react/button-has-type": "error",
+        "testing-library/no-node-access": "off"
     },
     "reportUnusedDisableDirectives": true
 }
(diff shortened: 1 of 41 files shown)


.eslintrc.json
```diff
@@ -237,7 +237,8 @@
         "flowtype/sort-keys": "error",
         "babel/semi": "error",
         "jsx-a11y/no-static-element-interactions": "error",
-        "react/button-has-type": "error"
+        "react/button-has-type": "error",
+        "testing-library/no-node-access": "off"
```
niklasnatter, line 241: the `testing-library/no-node-access` rule prevents accessing properties of html nodes (eg. `node.parentElement`):
https://github.com/testing-library/eslint-plugin-testing-library/blob/main/docs/rules/no-node-access.md

this might make sense in new projects as it forces you to add things like aria labels. but it makes things hard for us when migrating tests from enzyme to rtl. i think we should disable the rule to make the migration as easy as possible.

alexander-schranz (approved):

alexander-schranz:
Make sense for me. Can you check why flow is failing?

alexander-schranz (approved):

alexander-schranz (approved):

PR #6783, opened 2022-08-08
Use userEvent instead of fireEvent for simulating user interactions in javascript tests

niklasnatter:
#### What's in this PR?

This PR adjusts our javascript test cases to use the `@testing-library/user-event` package instead of the `fireEvent` method.

#### Why?

Because the `@testing-library/user-event` will simulate user interaction more realistically. Eg, the `userEvent.click()` will fire a `hover` event and a `mousedown` event before the `click` event is fired: https://stackoverflow.com/questions/66598305/fireevent-click-form-fireevent-vs-element-click-from-jsdom

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Backdrop/tests/Backdrop.test.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Backdrop/tests/Backdrop.test.js
@@ -1,5 +1,6 @@
 // @flow
-import {fireEvent, render, screen} from '@testing-library/react';
+import {render, screen} from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
 import React from 'react';
 import Backdrop from '../Backdrop';
 
@@ -8,12 +9,12 @@ test('The component should render', () => {
     expect(container).toMatchSnapshot();
 });
 
-test('The component should call a function when clicked', () => {
+test('The component should call a function when clicked', async() => {
     const onClickSpy = jest.fn();
     render(<Backdrop onClick={onClickSpy} />);
     const backdrop = screen.queryByTestId('backdrop');
 
     expect(onClickSpy).toHaveBeenCalledTimes(0);
-    fireEvent.click(backdrop);
+    await userEvent.click(backdrop);
     expect(onClickSpy).toHaveBeenCalledTimes(1);
 });
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/TextArea/tests/TextArea.test.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/TextArea/tests/TextArea.test.js
@@ -1,6 +1,7 @@
 // @flow
 import React from 'react';
-import {fireEvent, render, screen} from '@testing-library/react';
+import {render, screen} from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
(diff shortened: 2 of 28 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/components/Backdrop/tests/Backdrop.test.js
```diff
@@ -8,12 +9,12 @@ test('The component should render', () => {
...
 
-test('The component should call a function when clicked', () => {
+test('The component should call a function when clicked', async() => {
     const onClickSpy = jest.fn();
     render(<Backdrop onClick={onClickSpy} />);
     const backdrop = screen.queryByTestId('backdrop');
 
     expect(onClickSpy).toHaveBeenCalledTimes(0);
-    fireEvent.click(backdrop);
+    await userEvent.click(backdrop);
```
niklasnatter, line 18: the `@testing-library/user-event` package will simulate user interactions morea realistically than the `fireEvent` method. the `fireEvent` method will only fire a single `click` event while the `@testing-library/user-event` will trigger a `hover` and a `mousedown` event before firing the `click` event like a real user would do.

alexander-schranz (approved):

alexander-schranz (approved):

src/Sulu/Bundle/AdminBundle/Resources/js/components/TextArea/tests/TextArea.test.js
```diff
@@ -42,32 +43,35 @@ test('TextArea should render with value and character counter', () => {
...
 });
 
-test('TextArea should call onChange when the TextArea changes', () => {
+test('TextArea should call onChange when the TextArea changes', async() => {
     const changeSpy = jest.fn();
     render(<TextArea onChange={changeSpy} value="My value" />);
 
     const textarea = screen.queryByDisplayValue('My value');
-    fireEvent.change(textarea, {target: {value: 'my-value'}});
+    await userEvent.type(textarea, '!');
```
niklasnatter, line 64: when typing something to the component, the `changeSpy` is called after every character. we do not update the `value` that is passed to the component (`value="My value"`) when our `changeSpy` is called.
unfortunately, this means we cannot type more than one character without rewriting the tests 😕

alexander-schranz (approved):

src/Sulu/Bundle/AdminBundle/Resources/js/components/TextArea/tests/TextArea.test.js
alexander-schranz: okay lets see how this works. Else there are 2 solutions our Input work out of the box via internal state which would I think break our "source of truth" or our tests us a "test component".

b3h3m0th: The fact that only one character can be typed sometimes leads to problems or unexpected behavior (at least for me). E. g. when working with floating number inputs one necessarily needs to type 3 characters to test it. `1`->`1.`->`1.5`. (#6767)

niklasnatter: i have implemented a helper function for wrapping controlled components in https://github.com/sulu/sulu/pull/6795. this should hopefully solve such problems 🙂

PR #6790, opened 2022-08-09
Migrate enzyme Select tests to React Testing Library

b3h3m0th:
#### What's in this PR?

Select components unit test migration enzyme --> RTL

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/tests/Action.test.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/tests/Action.test.js
@@ -1,5 +1,5 @@
 // @flow
-import {fireEvent, render, screen} from '@testing-library/react';
+import {render, screen} from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import React from 'react';
 import Action from '../Action';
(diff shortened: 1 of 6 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/tests/Action.test.js
```diff
@@ -1,46 +1,63 @@
...
 
 test('The component should call the callbacks after a click', () => {
     const onClick = jest.fn();
     const afterAction = jest.fn();
-    const action = shallow(<Action afterAction={afterAction} onClick={onClick} value="my-option">My action</Action>);
-    action.find('button').simulate('click');
+    render(<Action afterAction={afterAction} onClick={onClick} value="my-option">My action</Action>);
+
+    const button = screen.queryByText('My action');
+    // eslint-disable-next-line testing-library/prefer-user-event
```
alexander-schranz: can you add why is user-event here not possible?

alexander-schranz (changes requested):
Looks like the tests are still failing.

src/Sulu/Bundle/AdminBundle/Resources/js/components/Select/tests/Action.test.js
b3h3m0th: I have no idea. But when doing the same thing with `userEvent` the test just fails.

niklasnatter: is it possible that you forgot to `await` the interaction? i forgot about that a few times already 🙈

niklasnatter (approved):

PR #6807, opened 2022-08-29
Fix query parameters for external links

ghost:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

The external link type has removed all query parameters before editing.

Video before:

https://user-images.githubusercontent.com/47974249/187246980-40eeae23-bade-41d8-93dd-2f3dabbd67ea.mov

Video after:

https://user-images.githubusercontent.com/47974249/187247051-d517d40e-e767-4526-a91c-35bc95a03533.mov

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Link/overlays/ExternalLinkTypeOverlay.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Link/overlays/ExternalLinkTypeOverlay.js
@@ -38,22 +38,22 @@ class ExternalLinkTypeOverlay extends React.Component<LinkTypeOverlayProps> {
             return;
         }
 
-        if (!href.startsWith('mailto:')) {
-            this.href = href;
+        if (href.startsWith('mailto:')) {
+            const urlParts = String(href).split('?');
+            const urlParameters = new URLSearchParams(urlParts[1]);
+            const mailSubject = urlParameters.get('subject');
+            const mailBody = urlParameters.get('body');
+
+            this.href = urlParts[0];
+            this.mailSubject = mailSubject ? mailSubject : undefined;
+            this.mailBody = mailBody ? mailBody : undefined;
 
             return;
         }
 
-        const urlParts = String(href).split('?');
-
-        this.href = urlParts[0];
-
-        const urlParameters = new URLSearchParams(urlParts[1]);
-        const mailSubject = urlParameters.get('subject');
-        const mailBody = urlParameters.get('body');
-
-        this.mailSubject = mailSubject ? mailSubject : undefined;
-        this.mailBody = mailBody ? mailBody : undefined;
+        this.href = href;
+        this.mailSubject = undefined;
+        this.mailBody = undefined;
     }
 
     callUrlChange = () => {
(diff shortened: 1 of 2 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/containers/Link/overlays/ExternalLinkTypeOverlay.js
```diff
@@ -42,6 +42,12 @@ class ExternalLinkTypeOverlay extends React.Component<LinkTypeOverlayProps> {
             return;
         }
 
+        if (this.protocol !== 'mailto:') {
```
luca-rath:
```suggestion
        if (!href.startsWith('mailto:')) {
```

luca-rath (changes requested):
Hi, thanks a lot for the pull request. Unfortunately your solution breaks the current behaviour of `mailto` links, because `this.protocol` is not yet set at that point. It will be set with a callback from the `Url` component, as soon as that renders, but the suggestion in my comment fixes that problem.

There are two things that prevent me from merging this right now:
- Please rebase this to the `2.4` branch. We follow SEMVER, so we fix bugs in the lowest supported version (2.4) and introduce new features in the next minor version (2.6)
- It would be great if you could test this fix. In the `ExternalLinkTypeOverlay.test.js` is already a `Render overlay with a URL` test, your solution could be easily tested, if you just change the url in that test from `http://www.sulu.io` to something like `http://www.sulu.io?foo=bar`. But be aware, that you also have to update the snapshot file then (you can run the tests for that file and update the snapshots using `npm run test -- -i ExternalLinkTypeOverlay.test.js -u`)

niklasnatter:
Hey, thanks for the pull request! 
I have fixed the handling of the `mailto:` specific parameters and added two respective test cases 🙂

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Link/tests/overlays/ExternalLinkTypeOverlay.test.js
```diff
@@ -84,6 +84,27 @@ test('Pass correct props to Dialog', () => {
     expect(externalLinkOverlay.find('Dialog').prop('open')).toEqual(false);
 });
 
+test('Display given URL with query parameters in href input', () => {
+    const targetChangeSpy = jest.fn();
+    const urlChangeSpy = jest.fn();
+
+    const externalLinkOverlay = shallow(
+        <ExternalLinkTypeOverlay
+            href="http://www.sulu.io?param=value-1"
```
alexander-schranz:
```suggestion
            href="http://www.sulu.io/contact-us?param=value-1"
```

use a valid url. Domain without `/` is not a valid url it always requires a path part and for homepage it is `/`.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Link/tests/overlays/ExternalLinkTypeOverlay.test.js
```diff
@@ -84,6 +84,27 @@ test('Pass correct props to Dialog', () => {
...
+            onHrefChange={urlChangeSpy}
+            onTargetChange={targetChangeSpy}
+            onTitleChange={jest.fn()}
+            open={true}
+            target="_blank"
+            title={undefined}
+        />
+    );
+
+    expect(externalLinkOverlay.find('Url').prop('value')).toEqual('http://www.sulu.io?param=value-1');
```
alexander-schranz:
```suggestion
    expect(externalLinkOverlay.find('Url').prop('value')).toEqual('http://www.sulu.io/contact-us?param=value-1');
```

alexander-schranz (changes requested):

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Link/tests/overlays/ExternalLinkTypeOverlay.test.js
niklasnatter: pushed the change together with linting fixes 🙂

alexander-schranz (approved):

alexander-schranz:
@flogado @niklasnatter Thank you!

PR #6814, opened 2022-09-14
Adjust CacheLifetimeEnhancer service to not check for PageInterface

luca-rath:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | sulu/SuluArticleBundle#621
| License | MIT
| Documentation PR | -

#### What's in this PR?

Adjust `CacheLifetimeEnhancer` service to not check for `PageInterface`, instead check if `getCacheLifeTime` method exists

#### Why?

To allow other structures to implement this method, so the `CacheLifetimeEnhancer` service works with those structures too. This PR is required for the integration of the SuluHeadlessBundle with the SuluArticleBundle


alexander-schranz (changes requested):
We should not introduce new interface on 2.4 instead 2.4 relevant software should currently use `method_exists` check instead rely on an interface. Still I think it make sense for 2.6 branch.

luca-rath:
Related PR for Sulu 2.6: #6815

niklasnatter (approved):

PR #6815, opened 2022-09-15
Extract PageInterface::getCacheLifeTime into own interface

luca-rath:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | #6814
| License | MIT
| Documentation PR | -

#### What's in this PR?

Extract `PageInterface::getCacheLifeTime()` into own interface

#### Why?

To allow other structures to implement this method, so the `CacheLifetimeEnhancer` service works with those structures too. This PR is required for the integration of the SuluHeadlessBundle with the SuluArticleBundle

### DIFF
--- a/src/Sulu/Component/Content/Compat/CacheLifetimeBehaviourInterface.php
+++ b/src/Sulu/Component/Content/Compat/CacheLifetimeBehaviourInterface.php
@@ -1,0 +1,25 @@
+<?php
+
+/*
+ * This file is part of Sulu.
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Component\Content\Compat;
+
+interface CacheLifetimeBehaviourInterface
+{
+    /**
+     * cacheLifeTime of template definition.
+     *
+     * @return array{
+     *     type: string,
+     *     value: string,
+     * }
+     */
+    public function getCacheLifeTime();
+}
--- a/src/Sulu/Component/Content/Compat/Structure/PageBridge.php
+++ b/src/Sulu/Component/Content/Compat/Structure/PageBridge.php
@@ -46,7 +46,10 @@ class PageBridge extends StructureBridge implements PageInterface
 
     public function getCacheLifeTime()
     {
-        return $this->structure->getCacheLifetime();
+        /** @var array{type: string, value: string} $cacheLifetime */
+        $cacheLifetime = $this->structure->getCacheLifetime();
+
+        return $cacheLifetime;
     }
 
     public function getOriginTemplate()
--- a/src/Sulu/Bundle/HttpCacheBundle/CacheLifetime/CacheLifetimeEnhancer.php
+++ b/src/Sulu/Bundle/HttpCacheBundle/CacheLifetime/CacheLifetimeEnhancer.php
@@ -12,7 +12,7 @@
 namespace Sulu\Bundle\HttpCacheBundle\CacheLifetime;
 
 use Sulu\Bundle\HttpCacheBundle\Cache\SuluHttpCache;
-use Sulu\Component\Content\Compat\CacheLifetimeStructureInterface;
+use Sulu\Component\Content\Compat\RoutableStructureInterface;
 use Sulu\Component\Content\Compat\StructureInterface;
 use Symfony\Component\HttpFoundation\Response;
 
@@ -52,7 +52,7 @@ class CacheLifetimeEnhancer implements CacheLifetimeEnhancerInterface
 
     public function enhance(Response $response, StructureInterface $structure)
     {
-        if (!$structure instanceof CacheLifetimeStructureInterface) {
+        if (!$structure instanceof RoutableStructureInterface) {
             return;
         }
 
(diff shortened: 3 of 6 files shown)


src/Sulu/Component/Content/Compat/CacheLifetimeBehaviourInterface.php
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Component\Content\Compat;
+
+interface CacheLifetimeBehaviourInterface
```
niklasnatter: i guess it makes sense to extract this, but i am not sure about the name/namespace. this is different to the behaviours in `Sulu\Component\Content\Document\Behavior` (eg `AuthorBehavior`), so i am not sure if i would use `Behaviour` here. but i also dont have a better idea, so if everybody else is okay with the name, its okay for me too 🙈

luca-rath: I don't wanna go with just `CacheLifetimeInterface`, but I guess `CacheLifetimeStructureInterface` would also work. wdyt? But this would imply, that the `CacheLifetimeStructureInterface` extends from `StructureInterface`, which it does not currently (but that wouldn't be a problem though)

niklasnatter: i would prefer `CacheLifetimeStructureInterface` to `CacheLifetimeBehaviourInterface` 🙂 but i would not extend the `StructureInterface`. i think its okay if it has just a single method

alexander-schranz: There was also another Pull Request in the past targetting the same issue: https://github.com/sulu/sulu/pull/4893

Which did include `getController`, `getView` and `getCacheLifeTime` maybe that make more sense. 🤔

luca-rath: @alexander-schranz What do you mean? Moving the `getCacheLifeTime` to the `StructureInterface` directly?

alexander-schranz: That comment was from `danrot` not from myself. At the opening it would have been possible to add the methods to structureinterface (pre 2.x) but we did not do it as snippets don't need them. And at current state it would be a bc break so we can not add it to the structureinterface and I also think it is not a good idea. I more want to add to the discussion if the other 2 things `getController` and `getView` should be in the same interface as they always will live together. When we have a `controller` we will have a `view` and that response will have a `cachelifetime`.

luca-rath: Okay but the ContentBundle does not have an interface yet defining the `getController()`, `getView()` and `getCacheLifeTime()` methods. So we could introduce that in sulu and implement it in the `ContentStructureBridge` too

alexander-schranz: > So we could introduce that in sulu and implement it in the ContentStructureBridge too.

Yeah the content bundle and article bundle should then use the new interface. So in future they maybe work over the same controller, but atleast currently can use the same service setting and resolving the cachelifetime from the structure.

Currently the cachelifetime in the content bundle is just set on `_cacheLifeTime` attribute in the route default provider (which is very inconsistent to the others (page / articles)) but I think currently not read because it does not implement the `PageInterface`. So would introduce the interface and then we can adopt the bundles to go with that interface and do not have to check for method_exists longer.

What do the other think about Interface containing the 3 methods? /cc @wachterjohannes @niklasnatter

niklasnatter: so you would create a `RoutableStructureInterface` that contains `getController`, `getView` and `getCacheLifeTime`, right?
that sound good to me 🙂

luca-rath: Did update the PR

src/Sulu/Component/Content/Compat/Structure/PageBridge.php
```diff
@@ -46,7 +46,10 @@ public function getLanguageCode()
 
     public function getCacheLifeTime()
     {
-        return $this->structure->getCacheLifetime();
+        /** @var array{type: string, value: string} $cacheLifetime */
+        $cacheLifetime = $this->structure->getCacheLifetime();
+
+        return $cacheLifetime;
```
alexander-schranz, line 52: Would be better to type this correclty in the [`StructureMetadata`](https://github.com/sulu/sulu/blob/668fd7cd8819407b546c1f376cae98deaf039077/src/Sulu/Component/Content/Metadata/StructureMetadata.php#L52-L63) class instead of have the need to set here `array{type: string, value: string}`: https://github.com/sulu/sulu/blob/668fd7cd8819407b546c1f376cae98deaf039077/src/Sulu/Component/Content/Metadata/StructureMetadata.php#L52-L63

alexander-schranz:
I'm closing this as I don' think we will soonish extract that into a interface. Else we can still cherry pick it from here.

PR #6825, opened 2022-09-26
Fix AccessControlQueryEnhancer to include entities with restricted permissions in unrelated system

niklasnatter:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Related issues/PRs | https://github.com/sulu/sulu/discussions/6804
| License | MIT

#### What's in this PR?

This pull requests fixes a bug in the `AccessControlQueryEnhancer` that excluded entities with restricted permissions in an unrelated system from the result set.

### DIFF
--- a/src/Sulu/Bundle/SecurityBundle/AccessControl/AccessControlQueryEnhancer.php
+++ b/src/Sulu/Bundle/SecurityBundle/AccessControl/AccessControlQueryEnhancer.php
@@ -16,7 +16,6 @@ use Doctrine\ORM\QueryBuilder;
 use Doctrine\Persistence\Mapping\MappingException;
 use Sulu\Bundle\SecurityBundle\Entity\AccessControl;
 use Sulu\Bundle\SecurityBundle\System\SystemStoreInterface;
-use Sulu\Component\Security\Authentication\RoleInterface;
 use Sulu\Component\Security\Authentication\UserInterface;
 
 class AccessControlQueryEnhancer
@@ -159,9 +158,15 @@ class AccessControlQueryEnhancer
     private function getUserRoleIds(?UserInterface $user): array
     {
         if ($user) {
-            return \array_map(function(RoleInterface $role) {
-                return $role->getId();
-            }, $user->getRoleObjects());
+            $roleIds = [];
+
+            foreach ($user->getRoleObjects() as $role) {
+                if ($role->getSystem() === $this->systemStore->getSystem()) {
+                    $roleIds[] = $role->getId();
+                }
+            }
+
+            return $roleIds;
         }
 
         $anonymousRole = $this->systemStore->getAnonymousRole();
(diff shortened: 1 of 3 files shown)


src/Sulu/Bundle/SecurityBundle/AccessControl/AccessControlQueryEnhancer.php
```diff
@@ -115,12 +115,12 @@ private function enhanceQueryWithAccessControl(
             'WITH',
             $accessClassCondition . ' AND ' . $this->getEntityIdCondition($entityClass, 'entity', $entityIdField)
         );
-        $subQueryBuilder->leftJoin('accessControl.role', 'role', 'WITH', 'role.system = :system');
+        $subQueryBuilder->leftJoin('accessControl.role', 'role');
```
niklasnatter: with the previous statement, if an access control entity existed in another system, the `leftJoin` joins an empty row here. the empty row will be interpreted as "entity cannot be accessed" by the `OR role.id IS NULL` below. 

this logic was most likely a leftover from the previous implementation that was changed in https://github.com/sulu/sulu/pull/6364

src/Sulu/Bundle/SecurityBundle/AccessControl/AccessControlQueryEnhancer.php
```diff
@@ -115,12 +114,12 @@ private function enhanceQueryWithAccessControl(
...
         $subQueryBuilder->andWhere(
             'BIT_AND(accessControl.permissions, :permission) <> :permission AND accessControl.permissions IS NOT NULL'
         );
 
-        $subQueryBuilder->andWhere('role.id IN(:roleIds) OR role.id IS NULL');
+        $subQueryBuilder->andWhere('role.id IN(:roleIds) AND role.system = :system');
```
alexander-schranz: As `:roleIds` is already filtered by `system` in `getUserRoleIds` the ` AND role.system = :system` is not required longer  here.

niklasnatter: changed 🙂

alexander-schranz:
Think the PR should target `2.4` ?

src/Sulu/Bundle/SecurityBundle/AccessControl/AccessControlQueryEnhancer.php
alexander-schranz: I'm missing something or can now the 2 joins be `innerJoin` instead of `leftJoin`?

niklasnatter: yes you are right, that should be a bit more performant 🙂

alexander-schranz (approved):

PR #6834, opened 2022-10-03
Add support for jackalope 2.0 and doctrine phpcr bundle 3.0

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? |  yes
| BC breaks? | yes (Row -> RowInterface) 
| Deprecations? | no  
| Fixed tickets | fixes # 
| Related issues/PRs | 
| License | MIT

#### What's in this PR?

Test CI against jackalope 2.0@dev version.

#### Why?

A new version is in work by @dbu which uses static typings and get ride of doctrine/cache.

#### Waiting

 - [x] https://github.com/jackalope/jackalope-doctrine-dbal/pull/386
 - [x] https://github.com/jackalope/jackalope-jackrabbit/pull/166
 - [x] https://github.com/jackalope/jackalope/pull/365
 - [x] https://github.com/doctrine/DoctrinePHPCRBundle/pull/369
 - [x] phpcr-shell https://github.com/phpcr/phpcr-shell/pull/214
 - [x] dantleech/phpcr-migrations?
 - [x] dantleech/phpcr-migrations-bundle https://github.com/dantleech/phpcr-migrations-bundle/pull/21


composer.json
```diff
@@ -33,7 +33,6 @@
         "contao/imagine-svg": "^1.0",
         "dantleech/phpcr-migrations-bundle": "^1.3",
         "doctrine/annotations": "^1.2",
-        "doctrine/cache": "^1.0",
```
alexander-schranz, line 35: The `doctrine/phpcr-bundle` still installs `doctrine/cache` `^1.0`. Which should be removed from that bundle as it does not even use that package.

composer.json
```diff
@@ -122,7 +121,6 @@
         "microsoft/azure-storage-blob": "^1.2",
         "monolog/monolog": "^1.26.1 || ^2.3",
         "php-ffmpeg/php-ffmpeg": "^0.17 || ^1.0",
-        "phpcr/phpcr-shell": "^1.4.0",
```
alexander-schranz: Requires: https://github.com/phpcr/phpcr-shell/pull/214

src/Sulu/Component/Content/Repository/Content.php
```diff
@@ -322,14 +322,14 @@ public function getChildren()
...
-     * @return Row
+     * @return RowInterface
      */
     public function getRow()
     {
         return $this->row;
     }
 
-    public function setRow(Row $row)
+    public function setRow(RowInterface $row)
```
alexander-schranz, line 330: BC Break

src/Sulu/Component/Content/Repository/ContentRepository.php
```diff
@@ -797,7 +798,7 @@ private function resolveAvailableLocales(Row $row)
      * @return Content|null
      */
     public function resolveInternalLinkContent(
-        Row $row,
+        RowInterface $row,
```
alexander-schranz, line 796: BC Break

alexander-schranz:
I rebased this branch to run against `^2.0@dev` currentl 2 tests are failing:

```
There were 2 errors:

1) Sulu\Component\DocumentManager\Tests\Unit\NodeHelperTest::testCopy
Prophecy\Exception\Call\UnexpectedCallException: Unexpected method call on Double\Jackalope\Workspace\P175:
  - copy(
        "/path/to/node",
        "uuid/node"
    )
expected calls were:
  - copy(
        exact("/path/to/node"),
        exact("/path/to/some/other/node/node")
    )

/home/runner/work/sulu/sulu/vendor/phpspec/prophecy/src/Prophecy/Call/CallCenter.php:203
/home/runner/work/sulu/sulu/vendor/phpspec/prophecy/src/Prophecy/Call/CallCenter.php:160
/home/runner/work/sulu/sulu/vendor/phpspec/prophecy/src/Prophecy/Prophecy/ObjectProphecy.php:218
/home/runner/work/sulu/sulu/vendor/phpspec/prophecy/src/Prophecy/Prophet.php:129
/home/runner/work/sulu/sulu/vendor/phpspec/prophecy-phpunit/src/ProphecyTrait.php:59

2) Sulu\Component\DocumentManager\Tests\Unit\NodeHelperTest::testCopyWithDestinationName
Prophecy\Exception\Call\UnexpectedCallException: Unexpected method call on Double\Jackalope\Workspace\P175:
  - copy(
        "/path/to/node",
        "uuid/new-node"
    )
expected calls were:
  - copy(
        exact("/path/to/node"),
        exact("/path/to/some/other/node/new-node")
    )

/home/runner/work/sulu/sulu/vendor/phpspec/prophecy/src/Prophecy/Call/CallCenter.php:203
/home/runner/work/sulu/sulu/vendor/phpspec/prophecy/src/Prophecy/Call/CallCenter.php:160
/home/runner/work/sulu/sulu/vendor/phpspec/prophecy/src/Prophecy/Prophecy/ObjectProphecy.php:218
/home/runner/work/sulu/sulu/vendor/phpspec/prophecy/src/Prophecy/Prophet.php:129
/home/runner/work/sulu/sulu/vendor/phpspec/prophecy-phpunit/src/ProphecyTrait.php:59
```

Not sure if this is an unexpect regression in PHPCR or an error on our side.

alexander-schranz:
I did do again a rebase the following errors currently appear:


```
In DefinitionErrorExceptionPass.php line 49:
                                                                               
  You have requested a non-existent parameter "doctrine_phpcr.odm.document_ma  
  nagers".                                                                                            
```

Strangely this parameter seems not be used from Sulu but should be defined by the phpcr bundle:

Bildschirmfoto 2023-08-30 um 17 42 12

Update: Parameter was defined with a default value previously here: https://github.com/doctrine/DoctrinePHPCRBundle/blob/c59d7449cd2d264af3ff2b510a30853e6f6f78ca/src/Resources/config/phpcr.xml#L9

src/Sulu/Bundle/DocumentManagerBundle/DependencyInjection/Compiler/InitializerPass.php
```diff
@@ -18,6 +18,10 @@ class InitializerPass implements CompilerPassInterface
 {
     public function process(ContainerBuilder $container)
     {
+        if (!$container->hasParameter('doctrine_phpcr.odm.document_managers')) { // TODO remove this see https://github.com/sulu/sulu/pull/6834#issuecomment-1699423757
+            $container->setParameter('doctrine_phpcr.odm.document_managers', []);
+        }
```
alexander-schranz: Temporary fix for: https://github.com/sulu/sulu/pull/6834#issuecomment-1699423757 should be fixed in the DoctrinePHPCRBundle

alexander-schranz:
Next Error seems related to the new cache configuration.

Bildschirmfoto 2023-08-30 um 18 29 12

Not sure if we or phpcrbundle need to require `psr/simple-cache` /cc @dbu?

Atleast `^3.0` of `psr/simple-cache` is not yet supported:

Bildschirmfoto 2023-08-30 um 18 30 57

composer.json
```diff
@@ -63,6 +68,7 @@
         "oro/doctrine-extensions": "^1.0.8 || ^2.0",
         "phpcr/phpcr-migrations": "^1.3.1",
         "phpcr/phpcr-utils": "^1.6.2",
+        "psr/simple-cache": "^1.0 || ^2.0 || ^3.0",
```
alexander-schranz: Not sure if sulu is really responsible to require this or if this should be required by the phpcr bundle 🤔

composer.json
```diff
@@ -24,16 +24,22 @@
...
         "ext-dom": "*",
         "ext-json": "*",
         "ext-simplexml": "*",
         "composer/package-versions-deprecated": "^1.8",
         "contao/imagine-svg": "^1.0",
-        "dantleech/phpcr-migrations-bundle": "^1.3",
+        "dantleech/phpcr-migrations-bundle": "dev-feature/test-phpcr-bundle-3-0 as 1.4.0",
         "doctrine/annotations": "^1.2 || ^2.0",
-        "doctrine/cache": "^1.0",
+        "doctrine/cache": "^1.0.1",
```
alexander-schranz, line 35: we should also remove this requirement from Sulu this would mostly mean small BC Breaks or create a copy of ArrayCache of Doctrine.

src/Sulu/Bundle/DocumentManagerBundle/DependencyInjection/Compiler/InitializerPass.php
dbu: fixing in https://github.com/doctrine/DoctrinePHPCRBundle/pull/386

src/Sulu/Component/Content/Repository/Content.php
dbu: if the bc break is an issue, you could still keep Row here, as the jackalope row does implement RowInterface.

but eventually it would be best if no Jackalope references need to be in sulu code.

composer.json
dbu: no longer by phpcr-bundle, but by a lot of other parts of doctrine (at least those would allow ^2.0 too): 
```
$ composer why doctrine/cache
sulu/sulu                3.0.x-dev requires doctrine/cache (^1.0.1)            
doctrine/dbal            3.6.6     requires doctrine/cache (^1.11|^2.0)        
doctrine/doctrine-bundle 2.10.2    requires doctrine/cache (^1.11 || ^2.0)     
doctrine/orm             2.15.5    requires doctrine/cache (^1.12.1 || ^2.1.1) 
doctrine/persistence     2.5.7     requires doctrine/cache (^1.11 || ^2.0) 
```

composer.json
dbu: jackalope is bringing this in already: https://github.com/jackalope/jackalope-doctrine-dbal/blob/2.x/composer.json#L26

(i did not change this in the last 2 days, not sure why this happened.)

dbu: i now allow simple-cache 2 and 3 as well in the jackalope-doctrine-dbal package, this should make things more flexible.

dbu:
i can reproduce the failing unit tests with NodeHelperTest

looking at the NodeHelper, i can't imagine why this has worked in the past. you use `'uuid'` as node id, instead of a valid uuid string. the NodeHelper then in `normalizePath` checks if this is a uuid and if not, returns the literal path that was passed.

i am also not quite sure why there is a call to a prophesized Workspace object to the copy method. i am not familiar with prophecy but it seems to me like you try to get it to actually copy the node, but without the object manager that will not do anything. apart from the `: void` return type, the `Workspace::copy` method has not changed.

alexander-schranz:
@dbu I rebased the branch to test against current version. There appears a new error which I maybe think we need to change something:

<details>

<summary>

> TypeError: Doctrine\Bundle\PHPCRBundle\ManagerRegistry::__construct(): Argument #5 ($defaultEntityManagerName) must be of type string, null given, called in /home/runner/work/sulu/sulu/src/Sulu/Bundle/SearchBundle/Tests/Application/var/cache/admin/test/ContainerOEjlAQ2/getDoctrinePhpcrService.php on line 21 and defined in /home/runner/work/sulu/sulu/vendor/doctrine/phpcr-bundle/src/ManagerRegistry.php:20

</summary>

Bildschirmfoto 2023-12-05 um 09 47 51

Another issue appears on prefer lowest but maybe it just add a conflict to `2.0.0-beta1` 🤔 

</details>

**Updated** Solved by: https://github.com/sulu/sulu/pull/6834/commits/cde0dae3bc00d9308e0dfcd6a67d14d32bfef599

> 37) Sulu\Bundle\PageBundle\Tests\Functional\Mapper\ContentMapperTest::testDelete
TypeError: Jackalope\ObjectManager::pathArrayToPropertiesIterator(): Argument #1 ($propertyPaths) must be of type string, array given, called in /home/runner/work/sulu/sulu/vendor/jackalope/jackalope/src/Jackalope/ObjectManager.php on line 689

composer.json
alexander-schranz: Nice I removed so the requirement here as sulu is using that namespace.

alexander-schranz:
@dbu I just had a relook at this. Is there some more devs we need to update here? Currently following error happens in the tests:

> PHP Fatal error:  During inheritance of Doctrine\Bundle\PHPCRBundle\Command\LoadFixtureCommand, while autoloading Doctrine\ODM\PHPCR\Tools\Console\Helper\DocumentManagerHelper: Uncaught ReflectionException: Class "Doctrine\ODM\PHPCR\Tools\Console\Helper\DocumentManagerHelper" not found while loading "Doctrine\Bundle\PHPCRBundle\Command\LoadFixtureCommand". in /home/runner/work/sulu/sulu/vendor/symfony/error-handler/DebugClassLoader.php:296

dbu:
i think the strict typing refactorings make visible what before would have been a fatal error when trying to run it, but did not break the container.

can you try with the branch in https://github.com/doctrine/DoctrinePHPCRBundle/pull/394 ?

dbu:
i fixed `You have requested a non-existent parameter "doctrine_phpcr.jackrabbit_jar".` in https://github.com/jackalope/jackalope-jackrabbit/pull/185 and https://github.com/doctrine/DoctrinePHPCRBundle/pull/395

are any of the other failures related to regressions in phpcr/jackalope?

alexander-schranz:
The 2.6 branch (jackalope 1) works without any problems https://github.com/sulu/sulu/actions/runs/7929229731/job/21735380686.

<details>
<summary>Dependency Diff</summary>

Bildschirmfoto 2024-02-19 um 18 05 03

</details>

I'm currently try to figure out what in the phpcr bundle has changed that could trigger:

> Doctrine\Persistence\Mapping\MappingException: Class 'Sulu\Bundle\TrashBundle\Domain\Model\TrashItemInterface' does not exist

Sulu as I know never used the phpcr `ODM` and so normally PHPCR bundle changes should not effect how or when Sulu loads the Doctrine Metadata 🤔 . But it seems something changed which is now triggering this error.

dbu:
that is weird indeed. what did change is that doctrine and symfony prefer attributes over annotations. 

the stack trace is indicating that this is about doctrine bundle (the orm thing) and doctrine orm rather than phpcr-odm. maybe the new version unlocks installation with newer doctrine that triggers the problem?

dbu:
there also seems to be a setup issue with `An exception occurred while executing a query: SQLSTATE[42P01]: Undefined table: 7 ERROR:  relation "phpcr_namespaces" does not exist`

i am not sure if something changed in jackalope-doctrine-dbal that makes us require the repository to be initialized that previously did not happen, or if the setup somehow changed and is not done correctly.

alexander-schranz:
Good hint PHP 8.1 seems to show only the previous errors: https://github.com/sulu/sulu/actions/runs/7962902795/job/21737385713?pr=6834.

I need to check what the dev diff to PHP 8.2 and 8.3 is.

alexander-schranz:
Looks like PHP 8.2 and 8.3 installs a newer version of doctrine-bridge:

Bildschirmfoto 2024-02-19 um 18 15 34

I'm not understanding why the doctrine-bridge fails here now but atleast I think we can say its not related to the phpcr changes. I did add a requirement to it let see if the CI works now.

composer.json
```diff
@@ -75,6 +81,7 @@
         "symfony/css-selector": "^5.4 || ^6.0",
         "symfony/dependency-injection": "^5.4 || ^6.0",
         "symfony/deprecation-contracts": "^2.1 || ^3.0",
+        "symfony/doctrine-bridge": "^5.4 || ^6.0",
```
alexander-schranz, line 78: Else Symfony 7 is installed which errors with:

Bildschirmfoto 2024-02-19 um 18 17 32

dbu: you will eventually want to figure out why that happens ;-) but agree to not mix it into this checks.

dbu:
> there also seems to be a setup issue with `An exception occurred while executing a query: SQLSTATE[42P01]: Undefined table: 7 ERROR: relation "phpcr_namespaces" does not exist`
> 
> i am not sure if something changed in jackalope-doctrine-dbal that makes us require the repository to be initialized that previously did not happen, or if the setup somehow changed and is not done correctly.

that seems to only happen with `lowest`. is the test setup in this repository itself or a dependency? if its a dependency, that probably needs to be bumped to a newer minimal version.

also something with doctrine dbal configuration vs doctrine data fixtures (that one seems to me like the data fixtures version you end up does not correctly define the dbal version it needs)

alexander-schranz:
lowest currently is using:

```
doctrine/phpcr-bundle (3.0.0-beta2)
jackalope/jackalope (2.0.0-beta2)
jackalope/jackalope-doctrine-dbal (2.0.0-beta2)
```

https://github.com/sulu/sulu/actions/runs/7963142702/job/21738134720?pr=6834 I will try to change `3.0@dev` to `dev-3.x as 3.0.0` hope that checkout the correct branch then.

alexander-schranz:
Lowest seems also use now `doctrine/dbal` `3` which errors in an unexpected way with the `doctrine/data-fixtures` package where the ORMPurger does not longer work like expected 🤔 . I'm not yet sure which of the doctrine dependency causing the new issue. data-fixtures are same on PHP 8.1 and PHP 8.2 now after I upgradedit, it triggers also now another error.

What is new is maybe `TypeError: Mock_Node_7cc39ccc::getNodes(): Return value must be of type Iterator, array returned` / `TypeError: Mock_Node_ac1e18ca::getProperties(): Return value must be of type Iterator, array returned`. Is this really now a `Iterator` everytime or can this be `iterable` return type to be also array?

dbu:
the typeerror: phpcr always had the phpdoc of `@return \Iterator`. phpcr-api-tests is not explicitly testing that, but [implicitly assuming it](https://github.com/phpcr/phpcr-api-tests/blob/f22ed0e882debf1c98da52a2b9262ec6dc8b459c/inc/BaseCase.php#L205).

i feel that relaxing the phpcr interface to allow any iterable is more of a BC break than jackalope enforcing the interface that has been documented. if we would be doing the very first version, i am also not sure how i would choose. iterable is very flexible, and most places do foreach anyways, so don't care if they get an Iterator or an array. but for some things like getting the first element, its very different if you cal `->first()` or do `[0]`. 

the reason we went with itherators is that they allow to handle large node collections with lazyness. afaik in some places, we actually use that to only load batches of nodes. and the Iterators also makes clear that one should not edit the structure returned.

if you can mock NodeInterface in this place, you will not see the problem (at least until we might do a strictly typed new major version of phpcr, which would be sorta nice but i don't think happening anytime soon).
if you need to mock the actual jackalope implementation, i't wrap the returns in `new \ArrayIterator`.

mamazu:
> i can reproduce the failing unit tests with NodeHelperTest
> 
> looking at the NodeHelper, i can't imagine why this has worked in the past. you use `'uuid'` as node id, instead of a valid uuid string. the NodeHelper then in `normalizePath` checks if this is a uuid and if not, returns the literal path that was passed.
> 
> i am also not quite sure why there is a call to a prophesized Workspace object to the copy method. i am not familiar with prophecy but it seems to me like you try to get it to actually copy the node, but without the object manager that will not do anything. apart from the `: void` return type, the `Workspace::copy` method has not changed.

This has never worked. It also doesn't work in the base branch. The prophecy thing for void methods is that you need to call  `shouldBeCalled` on it for it to trigger an assertion. Looks like an involuntary package update caused it to change the default behaviour to always include it now. So the test was broken before, it was just never asserted until now.

mamazu:
Well the last test that's wrong that's broken, doesn't sound like a problem with this PR directly. I have noticed that the "lowest version" of the dbal package is now updated.

alexander-schranz:
I did change the composer requirements so both versions are now supported tests currently failing because it does install phpcr bundle 2.5 which is not compatible with jackalope 2 but phpcr bundle does not define its requirements for jackaloe.

The installation of phpcr bundle 3.0 is blocked by https://github.com/phpcr/phpcr-migrations-bundle/pull/4

dbu:
> it does install phpcr bundle 2.5 which is not compatible with jackalope 2 but phpcr bundle does not define its requirements for jackalope.

i could add a conflict to latest phpcr-bundle, but i guess that does not help much as composer would just pick an older version to satisfy the solution.
otherwise i could try to make phpcr-bundle 2.x compatible with jackalope 2, but probably a bit wasted effort, lets rather get phpcr-bundle 3 working and use that.

alexander-schranz:
> i could add a conflict to latest phpcr-bundle, but i guess that does not help much as composer would just pick an older version to satisfy the solution.

We would need to add conflict to `^2.5` version and sulu then would require to require atleast that version that way it would work, as composer then can not just downgrade phpcr bundle.

dbu:
> We would need to add conflict to ^2.5 version and sulu then would require to require atleast that version that way it would work, as composer then can not just downgrade phpcr bundle.

you mean a new patch version of the phpcr-bundle that declares a conflict with jackalope 2? or a conflict in sulu (that could be `conflicts: {"doctrine/phpcr-bundle": "<3.0"}` if you don't require phpcr-bundle directly)

src/Sulu/Component/Content/Tests/Unit/Repository/ContentRepositoryTest.php
```diff
@@ -164,8 +169,8 @@ public function setUp(): void
         $this->structureManager->getStructures('page')->willReturn([$structure->reveal()]);
         $this->structureManager->getStructure('test')->willReturn($structure->reveal());
 
-        $this->query = $this->prophesize(QueryInterface::class);
-        $this->query->setLimit(Argument::any())->willReturn(null);
+        $this->query = $this->prophesize(QueryObjectModel::class);
+        $this->query->setLimit(Argument::any());
```
alexander-schranz, line 172: Looks like this change does make prefer lowest failing.

Bildschirmfoto 2024-04-04 um 16 22 54

@mamazu do you want to have a look at this?

alexander-schranz: Looks like something here https://github.com/sulu/sulu/blob/9fb83560219c38e19f92fa6ad773f39490c81c8e/src/Sulu/Component/Content/Tests/Unit/Repository/ContentRepositoryTest.php#L182 is missing.

mamazu: Yes, PR is open.

alexander-schranz: @mamazu Thank you works 👍

dbu:
yay, that looks pretty green now \0/

looks like whats left now is mainly about the added strict typing in upstream libraries that makes phpstan report missing checks for `null` and similar things.

alexander-schranz:
Ready for review. @chirimoya

chirimoya (approved):

UPGRADE.md
```diff
@@ -138,6 +138,61 @@ Use the newly added `getUrl` method:
...
+public function resolveInternalLinkContent(
+-    Row $row,
++    RowInterface $row,
+     $locale,
+```
+
+### PHPCR and Jackalope update
+
+An update of PHPCR and Jackalope to latest major version is optional but recommended.   
+The following new version are supported by Sulu:
```
```suggestion
The following new versions are supported by Sulu:
```

Prokyonn (approved):

PR #6837, opened 2022-10-04
Fix SecurityChecker service to check permissions for anonymous users

### LINKED ISSUE #6833: Anonymous roles always granted access to media when webspace security system is defined
| Q | A
| --- | ---
| Sulu Version | 2.3 or higher
| Symfony | 5.3 or higher
| PHP Version | 7.2 or higher
| DB Version | N/A
| Browser Version | N/A

# Actual Behavior

If there's a security system configured for a webspace and the anonymous role for that system is denied access to a media collection, access is still always granted if a user knows the download routes for the files in the collection.

# Expected Behavior

Access should be denied to a media file if an anonymous role doesn't have the view permission.

# Steps to Reproduce

(**Edit:** As per the discussion below, I've created a repository to reproduce this issue [here](https://github.com/NeuralClone/sulu-skeleton/tree/reproducer/anonymous-role-media-download).)

1. Create a new project using `sulu/skeleton` as outlined in the [Getting Started](https://docs.sulu.io/en/latest/book/getting-started.html) guide.
2. Add a security system to the example webspace:

```xml
    <system>Website</system>
</security>
```

3. Enable user context caching as [described here](https://docs.sulu.io/en/2.5/cookbook/user-context-caching.html).
4. Create a new media collection, activate the above security system for it, and deny access to the anonymous role.
5. Upload a media file to the new collection.
6. Create a download link for the newly uploaded media in a template:

```twig
{% set media = sulu_resolve_media(1, app.request.locale) %}
{% set url = sulu_get_media_url(media, 'attachment') %}

<a href="{{ url }}">Download</a>
```
7. Download the file without logging in.

# Additional Information

The problem seems to occur in the `SecurityChecker` service, Specifically, the first `if` statement:

https://github.com/sulu/sulu/blob/54a06ae9662d4944d51ebfcebafa07bbd68982fd/src/Sulu/Component/Security/Authorization/SecurityChecker.php#L40-L46

Anonymous roles are always granted access to media files regardless of permissions in Symfony 5.3 or later due the token always being `null`. This is fine when a security system isn't defined.

But it's problematic if you've configured one, want to deny media access to anonymous roles, and you're using a version of Symfony that doesn't create anonymous tokens.

# Possible Solutions

I've created a temporary workaround for this by overriding the `SecurityChecker` service and removing the anonymous check. This forces `isGranted()` to be called and access is correctly denied when appropriate.

I don't know if this is the best fix. I needed a quick way to prevent access to media collections while working out a more general solution and this seemed to break the least amount of stuff without knowing all of the inner workings of Sulu.

niklasnatter:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #6833
| License | MIT

#### What's in this PR?

This pull request fixes the `SecurityChecker` service for anonymous users (users that are not logged in). At the moment, anonymous users are always granted access. The bug occurs because the Symfony security bundle does not return any `Token` object for unauthenticated users anymore: https://symfony.com.ua/doc/current/security/experimental_authenticators.html


src/Sulu/Bundle/MediaBundle/Tests/Application/Kernel.php
```diff
@@ -13,10 +15,22 @@
...
+use Symfony\Bundle\SecurityBundle\SecurityBundle;
 use Symfony\Component\Config\Loader\LoaderInterface;
 
 class Kernel extends SuluTestKernel
 {
+    public function registerBundles(): iterable
+    {
+        $bundles = parent::registerBundles();
+
+        if (SuluTestKernel::CONTEXT_WEBSITE === $this->getContext()) {
```
niklasnatter: This enables the `SecurityBundle` for all functional test cases of the `MediaBundle` that run in the website context.

src/Sulu/Component/Security/Authorization/SecurityChecker.php
```diff
@@ -39,9 +39,8 @@ public function __construct(
 
     public function hasPermission($subject, $permission)
     {
-        if (!$subject || !$this->tokenStorage->getToken()) {
```
niklasnatter, line 42: `tokenStorage->getToken()` returned an `AnonymousToken` if no user was logged in previously. This was changed in the recently introduced authenticator based security model: https://symfony.com.ua/doc/current/security/experimental_authenticators.html

With the new security model, `tokenStorage->getToken()` will return `null` if no user is logged in.

src/Sulu/Component/Security/Authorization/SecurityChecker.php
```diff
@@ -39,9 +39,8 @@ public function __construct(
...
+        if (!$subject) {
             // if there is no subject the operation is allowed, since we have nothing to check against
-            // if there is no token we are not behind a firewall, so the action is also allowed (e.g. command execution)
```
niklasnatter, line 44: we decided that this is not the correct place to disable security checks during command execution. instead of disabling checks automatically, out services should allow to turn off check explicitly like the `MediaManager::delete()` method:
https://github.com/sulu/sulu/blob/b5e6afab84ea8c01d08be5890aa21c0d244a331b/src/Sulu/Bundle/MediaBundle/Media/Manager/MediaManagerInterface.php#L89

src/Sulu/Bundle/MediaBundle/Tests/Application/config/config_secured.yml
```diff
@@ -0,0 +1,24 @@
+security:
+    enable_authenticator_manager: true
```
alexander-schranz, line 2: @niklasnatter I had to add this line else the test did not fail in my case.

src/Sulu/Bundle/MediaBundle/Tests/Application/Kernel.php
alexander-schranz: Updates this to create a new `WebsiteSecured` Testcase

src/Sulu/Bundle/MediaBundle/Tests/Functional/Controller/MediaStreamControllerWebsiteSecuredTest.php
```diff
@@ -0,0 +1,179 @@
...
+    protected static function getKernelClass(): string
+    {
+        return SecuredKernel::class;
+    }
+
+    public function setUp(): void
+    {
+        if (\version_compare(SecuredKernel::VERSION, '5.4', '<')) { // @phpstan-ignore-line
+            $this->markTestSkipped('Test is written for Symfony 5.4 or newer.');
+        }
```
alexander-schranz, line 47: The test does not work on lowest because we need `enable_authenticator_manager`. In Sulu 2.5 this line can be removed when merging back.

alexander-schranz (approved):

src/Sulu/Bundle/MediaBundle/Tests/Application/config/config_secured.yml
niklasnatter: thats interesting, i think this should always be set to true on Symfony 6: https://github.com/symfony/symfony/blob/5.4/UPGRADE-6.0.md 
but shouldnt hurt to set it explicitly 🙂

alexander-schranz: The 2.4 branch did run on `4.4` and `5.4` the default there is `enable_authenticator_manager: false`.

PR #5924, opened 2021-04-07
Update locale of preview when switching locale of form

niklasnatter:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Fixed tickets | fixes #5084
| Related issues/PRs | #5665
| License | MIT

#### What's in this PR?

This PR adjusts the `Preview` container to pass the observable locale of the form to the `PreviewStore` instead of reading to locale from the router a single time.

This PR requires #5665 to handle the new locale on the server side the locale. Without #5665 the new locale will be sent to the server, but the server will ignore the locale of the request and [use the outdated locale from the cache](https://github.com/sulu/sulu/pull/5665/files#diff-4e58f16c1677e0b707c7193da31bfed39b335555747cf84d89ce00042a35e572R100). 

#### Why?

At the moment, if the user switches the locale of the form, the locale of the `PreviewStore` is not updated (see #5084).

PR #6861, opened 2022-10-21
Add restart of preview when locale of form is changed

### LINKED ISSUE #6831: Translation bug in preview (snippets and smart content)
| Q | A
| --- | ---
| Sulu Version | 2.4+
| PHP Version | 7.4.30
| DB Version | MySQL 5.7.39
| Browser Version | Chrome 105.0.5195.125 / Firefox 105.0.1

# Actual Behavior

When you edit a page in the admin and toggle the language via the dropdown in the edit window, the snippets and smart content in the preview will keep the original language you opened the page in, instead of switching to the new language you selected.

# Expected Behavior

All snippets and smart content should be shown in the correct language.

# Steps to Reproduce

Go to the admin on sulu.rocks and edit the blog page with `en` selected as the language. Then toggle the language to `de`via the dropdown. You'll see that the content in the preview window is updated, but the article teasers (smart content) will still be in English instead of German.

translation-issue

# Possible Solutions

This seems to be due to the fact that the /admin/preview/start call is only executed on the initial page load, so content will be cached for that locale, and subsequent update calls will not touch the snippets and smart content. Because when you toggle the language only the /admin/preview/update call is executed, and this will load the cached data from the previously selected locale. You can test this by reloading the edit page in the browser (using the browser "Reload this page" button), then everything will be ok again because the /admin/preview/start call will be triggered.

Not sure what the best fix is here, but based on what I described above a quick fix seems to be to call the previewStore.start(); method in JS whenever the language is toggled in the UI.

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #6831
| Related issues/PRs | none
| License | MIT
| Documentation PR | none

#### What's in this PR?

This PR add a restart of preview if the locale of the form has changed.

### DIFF
--- a/src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/Preview.js
+++ b/src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/Preview.js
@@ -46,6 +46,7 @@ class Preview extends React.Component<Props> {
 
     schemaDisposer: () => mixed;
     dataDisposer: () => mixed;
+    localeDisposer: () => mixed;
 
     @computed get webspaceKey() {
         const {
@@ -88,7 +89,6 @@ class Preview extends React.Component<Props> {
         }));
 
         this.createPreviewStore();
-
         if (Preview.mode === 'auto') {
             this.startPreview();
         }
@@ -182,6 +182,13 @@ class Preview extends React.Component<Props> {
                 }
             }
         );
+
+        this.localeDisposer = reaction(
+            () => toJS(formStore.locale),
+            () => {
+                this.previewStore.restart();
+            }
+        );
     };
 
     updatePreview = debounce((data: Object) => {
@@ -221,6 +228,10 @@ class Preview extends React.Component<Props> {
 
         if (this.dataDisposer) {
             this.dataDisposer();
+        }
+
+        if (this.localeDisposer) {
+            this.localeDisposer();
         }
     }
 
--- a/src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/stores/PreviewStore.js
+++ b/src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/stores/PreviewStore.js
@@ -89,6 +89,10 @@ export default class PreviewStore {
         });
     }
 
+    restart(): Promise<string> {
+        return this.stop().then(this.start);
+    }
+
     update(data: Object): Promise<string> {
         const route = generateRoute('update', {
             locale: this.locale,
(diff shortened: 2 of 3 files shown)


niklasnatter:
I thought we fixed this some time ago with #5924 🙈 Will check why the previous PR did not solve the problem

niklasnatter:
Okay I debugged this through and found the root cause. The problem is that the `TeaserContentType` reads the locale from the structure of the property using `$property->getStructure()->getLanguageCode()`:
https://github.com/sulu/sulu/blob/ea56fe23dca0699657d56f530c3a2d58451cce7a/src/Sulu/Bundle/PageBundle/Teaser/TeaserContentType.php#L99

The `PageObjectProvider` does not update the locale of the cached `PageDocument` in its `setValues` method and therefore the initial locale is used by the `TeaserContentType`. 

I think it would be nicer to fix this in the `PageObjectProvider` because the method already has all necessary information and this prevents an additional request from the frontend 🙂 The following diff (excluding tests) should be enough to fix it:

```diff
diff --git a/src/Sulu/Bundle/PageBundle/Preview/PageObjectProvider.php b/src/Sulu/Bundle/PageBundle/Preview/PageObjectProvider.php
index 14cd0eb0b3..6efce8eeab 100644
--- a/src/Sulu/Bundle/PageBundle/Preview/PageObjectProvider.php
+++ b/src/Sulu/Bundle/PageBundle/Preview/PageObjectProvider.php
@@ -70,6 +70,9 @@ class PageObjectProvider implements PreviewObjectProviderInterface
             ->enableMagicCall()
             ->getPropertyAccessor();
 
+        $object->setLocale($locale);
+        $object->setOriginalLocale($locale);
+
         $structure = $object->getStructure();
         foreach ($data as $property => $value) {
             try {

```

wachterjohannes:
@niklasnatter seems you are right - but the issue i think is that for all the other object provider it is also neccesary to add something like that - if we fix it on the JS side it would be fixed for all the providers. Additionally it seems strange to reuse a token < which is generated for a locale > for a different locale - what do you think?

niklasnatter:
I think it should only be necessary in the `ArticleObjectProvider` of the article bundle, right?
From a first look, I think the `SuluContentBundle` should not have the same problem.

> Additionally it seems strange to reuse a token < which is generated for a locale > for a different locale - what do you think?

I dont have a strong opinion on that. But if we dont allow to switch the locale, the `locale` argument of the `PreviewObjectProviderInterface::setValues` method is a bit confusing, right? 🙈

wachterjohannes:
@niklasnatter @alexander-schranz replaced by #6866

niklasnatter:
We figured that my solution above works for `smart_content` properties, but the content of the extensions like the SEO tab will not be reloaded (https://github.com/sulu/sulu/pull/6866#pullrequestreview-1154372756). Therefore we decided to restart the preview when the locale is switched 🙂

alexander-schranz (approved):

PR #6866, opened 2022-10-25
Add update of locale in previews PageObjectProvider

### LINKED ISSUE #6831: Translation bug in preview (snippets and smart content)

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #6831
| Related issues/PRs | replaces #6861 , linked with https://github.com/sulu/SuluArticleBundle/pull/630
| License | MIT
| Documentation PR | none

#### What's in this PR?

This PR add an update of locale and original locale on the preview document. To allow loading the data (e.g. teaser_selection) in the correct locale.

### DIFF
--- a/src/Sulu/Bundle/PageBundle/Preview/PageObjectProvider.php
+++ b/src/Sulu/Bundle/PageBundle/Preview/PageObjectProvider.php
@@ -70,6 +70,9 @@ class PageObjectProvider implements PreviewObjectProviderInterface
             ->enableMagicCall()
             ->getPropertyAccessor();
 
+        $object->setLocale($locale);
+        $object->setOriginalLocale($locale);
+
         $structure = $object->getStructure();
         foreach ($data as $property => $value) {
             try {


src/Sulu/Bundle/PageBundle/Preview/PageObjectProvider.php
```diff
@@ -70,6 +70,9 @@ public function setValues($object, $locale, array $data)
             ->enableMagicCall()
             ->getPropertyAccessor();
 
+        $object->setLocale($locale);
+        $object->setOriginalLocale($locale);
+
```
alexander-schranz, line 75: I think this will not work for the ContentBundle: https://github.com/sulu/SuluContentBundle/blob/8435b81422f708be0d7733c91526b981a9fe8a96/Content/Infrastructure/Sulu/Preview/ContentObjectProvider.php#L144

There we are loading the locale from the cached document (in this case still the old locale). And if we not restart the preview via the frontend still the old `locale` is in the `document`. A setter is bad as we need to load the document by the correct locale from the database.

Also I think with the current way things like `settings` e.g.: `author` is still false in the cached document as they are not from `setValues` given. So only data is given from the current tab not the other tabs.

Issue #6501, opened 2022-03-09
Search: Do not create a new Elasticsearch index field for every field in a Sulu XML template


sabinebaer:
# Problem description

A new field is created in the index for every field in the Sulu XML template if the Sulu search is used with the Elasticsearch adapter. This could lead to a huge amount of fields, above all when using the `sulu.search.field` tag for fields in block in block structures. Elasticsearch limits the number of fields by default to 1000 - so it could happen that the content manager adds one new block and is then running into an error when saving the page as the following Elasticsearch exception is thrown: `Limit of total fields [1000] has been exceeded`.

# Proposed solution

Best solution would be a new attribute in the `sulu.search.field` which gives the developer the possibility to choose if the content of this field should just be taken into account for searching or if it is really needed as a separate field that could be retrieved from the index (similar to the `Unstored` field type in Zend Lucene).

thomasduenser:
We ran into this problem for two projects in the last few weeks and therefore I would love to see an implementation of the proposed solution.

alexander-schranz:
fixed in #6868

PR #6868, opened 2022-10-27
Add condition to StructureProvider

thomasduenser:
Necessary for the MassiveSearchBundle so that too many fields are no longer generated for blocks.

| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | yes - The search index has lesser fields now as only the necessary ones are created (and grouped).
| Deprecations? | no
| Fixed tickets | #6501 
| Related issues/PRs | https://github.com/massiveart/MassiveSearchBundle/pull/168/files
| License | MIT
| Documentation PR | -

#### What's in this PR?

This PR adds the `$condition` to the `mapProperty` method. This is necessary for the MassiveSearchBundle to validate if the field should be indexed or not (when using the same property names across multiple block types).

#### Why?

To fix the ElasticSearch error:

```
Limit of total fields [1000] has been exceeded
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -58,7 +58,7 @@
         "jms/serializer": "^3.0",
         "jms/serializer-bundle": "^3.0 || ^4.0",
         "massive/build-bundle": "^0.3 || ^0.4 || ^0.5",
-        "massive/search-bundle": "^2.7",
+        "massive/search-bundle": "^2.0",
         "matomo/device-detector": "^3.7 || ^4.0.1",
         "nyholm/psr7": "^1.0",
         "oro/doctrine-extensions": "^1.0.8 || ^2.0",
--- a/composer.json
+++ b/composer.json
@@ -58,7 +58,7 @@
         "jms/serializer": "^3.0",
         "jms/serializer-bundle": "^3.0 || ^4.0",
         "massive/build-bundle": "^0.3 || ^0.4 || ^0.5",
-        "massive/search-bundle": "^2.0",
+        "massive/search-bundle": "^2.7",
         "matomo/device-detector": "^3.7 || ^4.0.1",
         "nyholm/psr7": "^1.0",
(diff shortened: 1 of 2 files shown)


wachterjohannes (approved):
@alexander-schranz WDYT?

composer.json
```diff
@@ -58,7 +58,7 @@
         "jms/serializer": "^3.0",
         "jms/serializer-bundle": "^3.0 || ^4.0",
         "massive/build-bundle": "^0.3 || ^0.4 || ^0.5",
-        "massive/search-bundle": "^2.0",
+        "massive/search-bundle": "^2.7",
```
wachterjohannes: @thomasduenser can you revert that line?

src/Sulu/Bundle/PageBundle/Search/Metadata/StructureProvider.php
```diff
@@ -314,7 +317,8 @@ private function mapProperty(ItemMetadata $property, $metadata, string $prefix =
                     $this->mapProperty(
                         $componentProperty,
                         $propertyMapping,
-                        $component->getName() . '_'
+                        $component->getName() . '_',
```
wachterjohannes: @thomasduenser change to `$componentProperty->getType()`

wachterjohannes:
@thomasduenser after changed the last comments can you add a screenshot of the result in elasticsearch

composer.json
alexander-schranz:
```suggestion
        "massive/search-bundle": "^2.0",
```

src/Sulu/Bundle/PageBundle/Search/Metadata/StructureProvider.php
alexander-schranz:
```suggestion
                        $componentProperty->getType() . '_',
```

alexander-schranz (changes requested):

PR #6906, opened 2022-11-11
Add page resource segment to PageRemovedEvent

### LINKED ISSUE #6198: Add resource locator of page to PageRemovedEvent
# Problem description

Currently you don't have access to the url when removing a page.

# Proposed solution

The url is a important information which you example use when push sulu content to a external library and is maybe used there as identifier, so it would be great that the PageRemovedEvent contains that url.

https://github.com/sulu/sulu/blob/8ce644508cf3d0f70f9cfe489aceab271298a492/src/Sulu/Bundle/PageBundle/EventListener/DomainEventSubscriber.php#L269-L276

Should be updated to something like:

```php
 $this->domainEventCollector->collect( 
     new PageRemovedEvent( 
         $document->getUuid(), 
         $document->getWebspaceName(), 
         $document->getTitle(), 
         $document->getLocale(),
         [
             'url' => $document->getResourceLocator(),
         ]
     ) 
 ); 
```

Think the url can then just be added to the payload.

vvasiloi:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #6198
| Related issues/PRs |
| License | MIT

#### To Do

- [x] ~~Use mocking in test (phpspec/prophecy)~~ Not necessary
- [x] ~~Create a documentation PR~~ There's no documentation for particular events

### DIFF
--- a/src/Sulu/Bundle/PageBundle/Domain/Event/PageRemovedEvent.php
+++ b/src/Sulu/Bundle/PageBundle/Domain/Event/PageRemovedEvent.php
@@ -40,11 +40,17 @@ class PageRemovedEvent extends DomainEvent
      */
     private $pageTitleLocale;
 
+    /**
+     * @var array
+     */
+    private $context;
+
     public function __construct(
         string $pageId,
         string $webspaceKey,
         ?string $pageTitle,
-        ?string $pageTitleLocale
+        ?string $pageTitleLocale,
+        array $context = []
     ) {
         parent::__construct();
 
@@ -52,11 +58,17 @@ class PageRemovedEvent extends DomainEvent
         $this->webspaceKey = $webspaceKey;
         $this->pageTitle = $pageTitle;
         $this->pageTitleLocale = $pageTitleLocale;
+        $this->context = $context;
     }
 
     public function getEventType(): string
     {
         return 'removed';
+    }
+
+    public function getEventContext(): array
+    {
+        return $this->context;
     }
 
     public function getResourceKey(): string
--- a/src/Sulu/Bundle/PageBundle/Domain/Event/PageRemovedEvent.php
+++ b/src/Sulu/Bundle/PageBundle/Domain/Event/PageRemovedEvent.php
@@ -41,10 +41,13 @@ class PageRemovedEvent extends DomainEvent
     private $pageTitleLocale;
 
     /**
-     * @var array
+     * @var array<string, mixed>
      */
     private $context;
 
+    /**
+     * @param array<string, mixed> $context
+     */
     public function __construct(
         string $pageId,
         string $webspaceKey,
--- a/src/Sulu/Bundle/PageBundle/Tests/Unit/EventListener/DomainEventSubscriberTest.php
+++ b/src/Sulu/Bundle/PageBundle/Tests/Unit/EventListener/DomainEventSubscriberTest.php
@@ -12,12 +12,13 @@
 namespace Sulu\Bundle\PageBundle\Tests\Unit\EventListener;
 
 use PHPUnit\Framework\TestCase;
-use Sulu\Bundle\ActivityBundle\Application\Collector\DomainEventCollector;
 use Sulu\Bundle\ActivityBundle\Application\Dispatcher\DomainEventDispatcher;
+use Sulu\Bundle\ActivityBundle\Domain\Event\DomainEvent;
 use Sulu\Bundle\DocumentManagerBundle\Collector\DocumentDomainEventCollector;
 use Sulu\Bundle\PageBundle\Document\PageDocument;
 use Sulu\Bundle\PageBundle\Domain\Event\PageRemovedEvent;
 use Sulu\Bundle\PageBundle\EventListener\DomainEventSubscriber;
+use Sulu\Bundle\TestBundle\Testing\SetGetPrivatePropertyTrait;
 use Sulu\Component\DocumentManager\DocumentManager;
 use Sulu\Component\DocumentManager\Event\RemoveEvent;
 use Sulu\Component\DocumentManager\NamespaceRegistry;
@@ -26,16 +27,16 @@ use Symfony\Component\EventDispatcher\EventDispatcher;
 
 class DomainEventSubscriberTest extends TestCase
 {
-    public function testAddResourceSegmentToPageRemovedEvent(): void
+    use SetGetPrivatePropertyTrait;
+
+    public function testAddContextToPageRemovedEvent(): void
     {
         $pageDocument = new PageDocument();
         $pageDocument->setTitle('Cool page');
         $pageDocument->setLocale('en_US');
         $pageDocument->setResourceSegment('/cool-page');
-        (function () {
-            $this->uuid = '6860560d-5f90-445b-b263-7e88ce03db4d';
-            $this->webspaceName = 'My Webspace';
-        })->bindTo($pageDocument, get_class($pageDocument))();
+        static::setPrivateProperty($pageDocument, 'uuid', '6860560d-5f90-445b-b263-7e88ce03db4d');
+        static::setPrivateProperty($pageDocument, 'webspaceName', 'My Webspace');
 
         $eventDispatcher = new EventDispatcher();
         $domainEventSubscriber = new DomainEventSubscriber(
@@ -48,14 +49,12 @@ class DomainEventSubscriberTest extends TestCase
 
         $domainEventSubscriber->handleRemove(new RemoveEvent($pageDocument));
 
-        $reflection = new \ReflectionClass(DomainEventCollector::class);
-        $property = $reflection->getProperty('eventsToBeDispatched');
-        $eventsToBeDispatched = $property->getValue($domainEventCollector);
-        $event = current($eventsToBeDispatched);
+        /** @var array<int, DomainEvent> $eventsToBeDispatched */
+        $eventsToBeDispatched = static::getPrivateProperty($domainEventCollector, 'eventsToBeDispatched');
+        $event = \current($eventsToBeDispatched);
 
         $this->assertCount(1, $eventsToBeDispatched);
         $this->assertInstanceOf(PageRemovedEvent::class, $event);
-        $this->assertArrayHasKey('url', $event->getEventContext());
-        $this->assertEquals('/cool-page', $event->getEventContext()['url']);
+        $this->assertEquals(['url' => '/cool-page'], $event->getEventContext());
     }
 }
--- a/src/Sulu/Bundle/PageBundle/Tests/Unit/EventListener/DomainEventSubscriberTest.php
+++ b/src/Sulu/Bundle/PageBundle/Tests/Unit/EventListener/DomainEventSubscriberTest.php
@@ -1,0 +1,61 @@
+<?php
+
+/*
+ * This file is part of Sulu.
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\PageBundle\Tests\Unit\EventListener;
+
+use PHPUnit\Framework\TestCase;
+use Sulu\Bundle\ActivityBundle\Application\Collector\DomainEventCollector;
+use Sulu\Bundle\ActivityBundle\Application\Dispatcher\DomainEventDispatcher;
+use Sulu\Bundle\DocumentManagerBundle\Collector\DocumentDomainEventCollector;
+use Sulu\Bundle\PageBundle\Document\PageDocument;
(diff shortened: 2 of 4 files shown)


src/Sulu/Bundle/PageBundle/Tests/Unit/EventListener/DomainEventSubscriberTest.php
```diff
@@ -0,0 +1,61 @@
...
+    public function testAddResourceSegmentToPageRemovedEvent(): void
+    {
+        $pageDocument = new PageDocument();
+        $pageDocument->setTitle('Cool page');
+        $pageDocument->setLocale('en_US');
+        $pageDocument->setResourceSegment('/cool-page');
+        (function () {
+            $this->uuid = '6860560d-5f90-445b-b263-7e88ce03db4d';
+            $this->webspaceName = 'My Webspace';
+        })->bindTo($pageDocument, get_class($pageDocument))();
```
alexander-schranz: We did not yet use `bindTo` in any tests. Would just call the setters or why is the bindTo required?

src/Sulu/Bundle/PageBundle/Tests/Unit/EventListener/DomainEventSubscriberTest.php
```diff
@@ -0,0 +1,61 @@
...
+            ),
+            new DocumentManager($eventDispatcher),
+            new PropertyEncoder(new NamespaceRegistry([]))
+        );
+
+        $domainEventSubscriber->handleRemove(new RemoveEvent($pageDocument));
+
+        $reflection = new \ReflectionClass(DomainEventCollector::class);
+        $property = $reflection->getProperty('eventsToBeDispatched');
+        $eventsToBeDispatched = $property->getValue($domainEventCollector);
```
alexander-schranz: We already have a trait for this: https://github.com/sulu/sulu/blob/2.5/src/Sulu/Bundle/TestBundle/Testing/SetGetPrivatePropertyTrait.php

src/Sulu/Bundle/PageBundle/Tests/Unit/EventListener/DomainEventSubscriberTest.php
```diff
@@ -0,0 +1,61 @@
...
+        $event = current($eventsToBeDispatched);
+
+        $this->assertCount(1, $eventsToBeDispatched);
+        $this->assertInstanceOf(PageRemovedEvent::class, $event);
+        $this->assertArrayHasKey('url', $event->getEventContext());
+        $this->assertEquals('/cool-page', $event->getEventContext()['url']);
```
alexander-schranz: Would do here:

```suggestion
        $this->assertSame(['url' => '/cool-page'], $event->getEventContext());
```

This would make the test fail if additional data is added and forces us that it is not forgot to adjust in that case also the test to add the new additional data then.

src/Sulu/Bundle/PageBundle/Domain/Event/PageRemovedEvent.php
```diff
@@ -40,25 +40,37 @@ class PageRemovedEvent extends DomainEvent
...
+     */
+    private $context;
+
     public function __construct(
         string $pageId,
         string $webspaceKey,
         ?string $pageTitle,
-        ?string $pageTitleLocale
+        ?string $pageTitleLocale,
+        array $context = []
```
alexander-schranz, line 56: If looking at the `DomainEvent` class I think this kind of data should be in the `payload` and `getEventPayload` instead of `getEventContext`.

/cc @wachterjohannes what do you think?

alexander-schranz (commented):
@vvasiloi Thank you for working on this. Just some small comments.

src/Sulu/Bundle/PageBundle/Domain/Event/PageRemovedEvent.php
vvasiloi: Since I never used Sulu I have no idea what's the difference between those two and coudn't find any explanation in the docs or code.
My decision of using `getEventContext` is based on the following observations from the current usage of both methods:
- `getEventPayload` is used in `PageCreatedEvent`, `PageModifiedEvent` and `PageRestoredEvent` events, where it holds the raw data that's used to create or update the `PageDocument` and is just carried on downstream. It's basically an input, the primary source
- `getEventContext` is used in `PageCopiedEvent` and `PageMovedEvent` events and it holds contextual data which is not an input and not a primary source, but rather data derived from the document and other sources

Howerver, even following that logic, I should've probably added an `$url` constructor argument and do:
```php
    public function getEventContext(): array
    {
        return [
            'url' => $this->pageUrl,
        ];
    }
```
This will be consistent with how the context is used in other events.

alexander-schranz: Let see what the other @sulu/core-team member think about it where to put this kind of data.

src/Sulu/Bundle/PageBundle/Tests/Unit/EventListener/DomainEventSubscriberTest.php
vvasiloi: Those properties do not have setters.
In some place they're mocked:
```php
$object = $this->prophesize(BasePageDocument::class);
$object->getUuid()->willReturn($id);
```
And in other places the documents are persisted, so the values are set by Doctrine using reflection.
Choose your poison.

vvasiloi: As you can see I didn't use mocking in this test and persistent felt like an overkill, so I used a closure, which is an alternative for using reflection api for such cases.

src/Sulu/Bundle/PageBundle/Tests/Unit/EventListener/DomainEventSubscriberTest.php
vvasiloi: Nice! I actually looked for it... Maybe that should be documented somewhere.

src/Sulu/Bundle/PageBundle/Tests/Unit/EventListener/DomainEventSubscriberTest.php
vvasiloi: I thought about doing it like this, but it feels like an unnecessary restriction.
The same context can easily accomodate multiple features without interfering with each other.
This test only cares about the url, not the whole context, as per it's name `testAddResourceSegmentToPageRemovedEvent`
Alternatively, I could rename the test and assert the whole context, then future extension of the context could reuse the same test.

src/Sulu/Bundle/PageBundle/Domain/Event/PageRemovedEvent.php
niklasnatter: IMO `payload` should contain the data that is used for updating the entity. So I would say `context` is the way to go here 🙂

wachterjohannes: @alexander-schranz i also agree with the opinion of niklas - we should go with context for this kind of data

vvasiloi: Should I keep the `$context` constructor argument or replace it with `$pageUrl`?

alexander-schranz: I would keep `$context`, this make it more flexible for the future if we maybe have additional data like a externalId or so on. /cc @sulu/core-team

vvasiloi: I prefer `$pageUrl`, because it's explicit, typehinted and easier to refactor.
The last time that constructor was modified was 2 years ago.

niklasnatter: I think there are valid points for both sides. I think I would try to keep it consistent with the other events that have a similar need, but I dont have a too strong opinion 🙂

vvasiloi: Only the `TagRemovedEvent` class uses an array context contructor argument.

alexander-schranz: We discussed it internal and we will keep the `$context` constructor parameter this way. So you can continue with the other comments. Again thank you for working on this 👍

src/Sulu/Bundle/PageBundle/Tests/Unit/EventListener/DomainEventSubscriberTest.php
alexander-schranz: > And in other places the documents are persisted, so the values are set by Doctrine using reflection.

We did more and more avoid mocks lately there are sadly still to many tests using mocks where they are not even required. So we did use in newer code mostly the Trait when there was no constructor or setter method: https://github.com/sulu/sulu/blob/2.5/src/Sulu/Bundle/TestBundle/Testing/SetGetPrivatePropertyTrait.php which make to code so more decoupled from mocking libraries. We still use mocks to mock things like Repository calls away. But where possible we try to avoid them. Personally I'm real fan of https://blog.frankdejonge.nl/testing-without-mocking-frameworks/ but we are definitely not there yet with the codebase.

src/Sulu/Bundle/PageBundle/Tests/Unit/EventListener/DomainEventSubscriberTest.php
alexander-schranz: Great Idea, it would be a great to have the TestBundle also documented. It is totally missing currently in the Documentation.

src/Sulu/Bundle/PageBundle/Tests/Unit/EventListener/DomainEventSubscriberTest.php
alexander-schranz: Make sense, would keep the test more general name so it can be extended in the future, instead of having a seperate test for every context value.

src/Sulu/Bundle/PageBundle/Tests/Unit/EventListener/DomainEventSubscriberTest.php
vvasiloi: I also prefer to avoid mocking, at least the excessive use.
That trait is exactly what's needed here.

src/Sulu/Bundle/PageBundle/Tests/Unit/EventListener/DomainEventSubscriberTest.php
vvasiloi: It didn't work, so I had to make some changes to the trait.

src/Sulu/Bundle/PageBundle/Tests/Unit/EventListener/DomainEventSubscriberTest.php
vvasiloi: Renamed it and made it assert the whole event context array.

src/Sulu/Bundle/PageBundle/Tests/Unit/EventListener/DomainEventSubscriberTest.php
vvasiloi: Replaced the closure with the trait methods.

src/Sulu/Bundle/TestBundle/Testing/SetGetPrivatePropertyTrait.php
```diff
@@ -30,10 +29,36 @@ protected static function getPrivateProperty(object $object, string $propertyNam
...
+        $reflection = new \ReflectionClass($object);
+
+        do {
+            if ($reflection->hasProperty($propertyName)) {
+                return $reflection->getProperty($propertyName);
+            }
+
+            // The property might be a private property declared in a parent class
+            $reflection = $reflection->getParentClass();
+        } while (false !== $reflection);
```
alexander-schranz, line 54: Nice addition to this helper function :+1:

vvasiloi:
@alexander-schranz I think it's ready. I looked to add some docs, but I didn't find any for spefic events, so I don't know where to start. 😅

alexander-schranz (approved):

alexander-schranz:
@vvasiloi thx for the pull request. great work 👍 

Yeah the events are really leaking. We shoul ddefinitly add a Events section to all the bundles to show and describe them.

luca-rath:
Hi @vvasiloi, just wanted to let you know how impressed I am about your methodical approach on this and the quality of the entire pull request, even if you have never worked with Sulu before.

Thanks for your contribution! 🙂

PR #6913, opened 2022-11-22
Add command to generate multiple database migrations

### LINKED ISSUE #4406: Add better way for upgrade for database changes
| Q | A
| --- | ---
| Bug? | no
| New Feature? | no
| Sulu Version | for 2.0.0
| Browser Version | -

#### Actual Behavior

Currently the we add all SQLs which need to be run to the UPGRADE.md this make it really hard for developers to upgrade there sulu.

#### Expected Behavior

We should add a easy way (command) to upgrade sulu to the current version.

#### Possible Solutions

Adding DoctrineMigrationBundle to the sulu skeleton (sulu-minimal). 

https://github.com/Sylius/Sylius-Standard/tree/master/src/Migrations

In this case we need to change the `sulu:build` process instead of the `doctrine:schema:update` which currently throw an a warning. Create new/destroy need to be handled seperatly in case of update.

In case of create new the following should be run:

 - `bin/console doctrine:schema:create`
 - `bin/console doctrine:migrations:version --add --all latest`

In case of update:

 - `bin/console doctrine:migration:migrate`

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? | yes
| BC breaks? | no 
| Deprecations? | no  
| Fixed tickets | fixes #4406 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Add command to generate multiple database migrations

#### Why?

To simplify the process of Sulu updates whatever database is used

After some discussion about the update process with @luca-rath I thought I hack together some prototype how we could easily generate doctrine migrations for different databases (postgresql, mysql, mariadb)..

#### Example Usage

```bash
cd src/Sulu/Bundle/MigrationBundle/Resources/docker/
docker compose up

bin/console sulu:migration:migration-diff
```

#### To Do

- [ ] Move parts into an own reusable bundle
- [ ] Create Migrations
  - [ ] 2.0 -> 2.1
  - [ ] 2.1 -> 2.2
  - [ ] 2.2 -> 2.3
  - [ ] 2.3 -> 2.4
  - [ ] 2.4 -> 2.5
  - [ ] 2.5 -> 2.6
- [ ] Supported Databases
   - [ ] PostgreSQL
      - [ ] Sequences vs Autoincrement IDs
      - [ ] JSON vs JSONB
   - [ ] MySQL
      - [ ] MySQL 8.0 (check if 8.0 requires another migration version)
      - [ ] MySQL 5.7 (check if 5.7 requires another migration version)
   - [ ] MariaDB
      - [ ] JSON vs TEXT
- [ ] Discuss dropping some databases (see also dbal 4 https://www.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/platforms.html#mysql vs dbal 3.5 https://www.doctrine-project.org/projects/doctrine-dbal/en/current/reference/platforms.html)
   - [ ] MySQL < 5.7
   - [ ] PostgreSQL < 10
- [ ] Use Create If Not Exist

Even this is a feature for `2.6` all migration files between minor version jumps should be there own `Version...php` file so from whaterver `2.x` version to `2.6` somebody is jumping they can directly use the doctrine migrations.

All migrations before `2.6` need first a check if they really need to be executed, as a project itself could already that field in a custom migration added.


src/Sulu/Bundle/MigrationBundle/Command/MigrationDiffCommand.php
```diff
@@ -0,0 +1,110 @@
...
+            $ui->section('Create Database: ' . $database);
+
+            $envs = ['DATABASE_URL' => $databaseUrl]; // TODO not working like expected
+
+            $createDatabaseCommand = new Process([
+                \PHP_BINARY,
+                $consolePath,
+                'doctrine:database:create',
+                '--if-not-exists',
+            ], null, $envs);
```
alexander-schranz: sadly this `envs` is yet not working like expected so we need to have a deeper look how workaround this issue.

alexander-schranz: This could be resolved via:

```php
            $envs = [
                'DATABASE_URL' => $databaseUrl,
                'SYMFONY_DOTENV_VARS' => false, // see https://symfony.com/doc/6.1/components/process.html#setting-environment-variables-for-processes
            ];
```

src/Sulu/Bundle/MigrationBundle/Command/MigrationDiffCommand.php
```diff
@@ -0,0 +1,137 @@
...
+        $finder = new Finder();
+        $finder->files()->in(\dirname(__DIR__) . '/Migrations')->name('Version*.php');
+
+        foreach ($finder as $file) {
+            \file_put_contents(
+                $file->getPathname(),
+                \str_replace('abortIf', 'skipIf', $file->getContents())
+            );
+        }
+    }
```
alexander-schranz, line 136: This should not be required as we should make sure that `PostgreSQL`, `MySQL` and `MariaDB` migrations life each in an own directory. And only the directory which is relevant is really loaded for the migrations. As `skipIf` migrations will trigger are a warning message in the doctrine migration command always, which we should avoid.

src/Sulu/Bundle/MigrationBundle/Migrations/Version20221122205423.php
```diff
@@ -0,0 +1,503 @@
...
+    }
+
+    public function up(Schema $schema): void
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->skipIf(
+            !$this->connection->getDatabasePlatform() instanceof \Doctrine\DBAL\Platforms\MySQL80Platform,
+            "Migration can only be executed safely on '\Doctrine\DBAL\Platforms\MySQL80Platform'."
+        );
+
```
alexander-schranz, line 27: Just example migration file which should be replaced in future.

src/Sulu/Bundle/MigrationBundle/Migrations/Version20221122205424.php
```diff
@@ -0,0 +1,824 @@
...
+        return '';
+    }
+
+    public function up(Schema $schema): void
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->skipIf(
+            !$this->connection->getDatabasePlatform() instanceof \Doctrine\DBAL\Platforms\PostgreSQL100Platform,
+            "Migration can only be executed safely on '\Doctrine\DBAL\Platforms\PostgreSQL100Platform'."
+        );
```
alexander-schranz, line 26: Just example migration file which should be replaced in future into more splitted files.

src/Sulu/Bundle/MigrationBundle/Migrations/Version20221122205425.php
```diff
@@ -0,0 +1,503 @@
...
+        return '';
+    }
+
+    public function up(Schema $schema): void
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->skipIf(
+            !$this->connection->getDatabasePlatform() instanceof \Doctrine\DBAL\Platforms\MariaDb1027Platform,
+            "Migration can only be executed safely on '\Doctrine\DBAL\Platforms\MariaDb1027Platform'."
+        );
```
alexander-schranz, line 26: Just example migration file which should be replaced in future into more splitted files.

stefantalen:
Any updates on this?  🙂

alexander-schranz:
We found another way todo this by using the dbal query builder inside the. migraitons: https://github.com/sulu/sulu/issues/4406#issuecomment-5251768379

PR #6916, opened 2022-11-26
Add emoji support for url slugger

### LINKED ISSUE #6882: Add emoji support for url slugger
# Problem description

Currently emojis are just removed and not replaced by the name when slugging a title with emoji for it. It would be nice to slug them also.

# Proposed solution

In Symfony 6.2 a new `withEmoji`: https://symfony.com/blog/new-in-symfony-6-2-better-emoji-support

For bc reasons it would require something like following that we run not into any error with Symfony 6.1 or lower:

```php
if (method_exists($slugger, 'withEmoji') {
    $slugger->withEmoji();
}
```

The class doing the slug can be found here: https://github.com/sulu/sulu/blob/2.5/src/Sulu/Component/PHPCR/PathCleanup.php and the slugger service definition here: https://github.com/sulu/sulu/blob/08863981d9e607dd1432db68eec2dac710732f74/src/Sulu/Bundle/CoreBundle/Resources/config/content.xml#L36

<br/>
<br/>

> **Note**
> This issue was specially created for beginners as it is a easy to implement. So it is perfectly for beginners and first time contributors. Experienced contributors should let this open for beginners.

nozarashi20:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #6882
| License | MIT

#### What's in this PR?

Add emoji support for url slugger

#### Why?

Currently emojis are just removed and not replaced by the name when slugging a title with emoji for it.

### DIFF
--- a/src/Sulu/Component/PHPCR/PathCleanup.php
+++ b/src/Sulu/Component/PHPCR/PathCleanup.php
@@ -29,12 +29,7 @@ class PathCleanup implements PathCleanupInterface
     /**
      * @var SluggerInterface
      */
-    private $sluggerWithEmoji;
-
-    /**
-     * @var SluggerInterface
-     */
-    private $sluggerWithoutEmoji;
+    private $slugger;
 
     /**
      * valid pattern for path
@@ -59,13 +54,11 @@ class PathCleanup implements PathCleanupInterface
         }
 
         if (\method_exists($slugger, 'withEmoji')) {
-            $this->sluggerWithEmoji = $slugger->withEmoji();
-        } else {
-            $this->sluggerWithEmoji = $slugger;
+            $slugger = $slugger->withEmoji();
         }
 
         $this->replacers = $replacers;
-        $this->sluggerWithoutEmoji = $slugger;
+        $this->slugger = $slugger;
     }
 
     /**
@@ -85,6 +78,7 @@ class PathCleanup implements PathCleanupInterface
                 $replacers,
                 isset($this->replacers[$languageCode]) ? $this->replacers[$languageCode] : []
             );
+            $languageCode = \str_replace('-', '_', $languageCode);
         }
 
         if (\count($replacers) > 0) {
@@ -113,11 +107,7 @@ class PathCleanup implements PathCleanupInterface
 
         $totalParts = \count($parts);
         foreach ($parts as $i => $part) {
-            try {
-                $slug = $this->sluggerWithEmoji->slug($part, '-', $languageCode);
-            } catch (\IntlException $e) { // handle unknown emoji transliterator id
-                $slug = $this->sluggerWithoutEmoji->slug($part, '-', $languageCode);
-            }
+            $slug = $this->slugger->slug($part, '-', $languageCode);
             $slug = $slug->lower();
             if (0 === $i || $i + 1 === $totalParts || !$slug->isEmpty()) {
                 $newParts[] = $slug->toString();
--- a/src/Sulu/Component/PHPCR/PathCleanup.php
+++ b/src/Sulu/Component/PHPCR/PathCleanup.php
@@ -29,7 +29,12 @@ class PathCleanup implements PathCleanupInterface
     /**
      * @var SluggerInterface
      */
-    private $slugger;
+    private $sluggerWithEmoji;
+
+    /**
+     * @var SluggerInterface
+     */
+    private $sluggerWithoutEmoji;
 
     /**
      * valid pattern for path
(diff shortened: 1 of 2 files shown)


src/Sulu/Component/PHPCR/PathCleanup.php
```diff
@@ -102,7 +106,16 @@ public function cleanup($dirty, $languageCode = null)
 
         $totalParts = \count($parts);
         foreach ($parts as $i => $part) {
-            $slug = $this->slugger->slug($part, '-', $languageCode);
+            try {
+                $slug = $this->slugger->slug($part, '-', $languageCode);
+            } catch (\IntlException $e) { // handle unknown emoji transliterator id
```
alexander-schranz: This sounds like something which should be reported to symfony itself. I think the slugger should just ignore if it does not know something.

alexander-schranz (changes requested):
@nozarashi20 Really thank you for working on this 👍 . I think the `IntlException` should not happen in case of the slugger. In which case it does happen? I think we should create an issue on [symfony side for this](https://github.com/symfony/symfony/issues). If it is resolved we could maybe move the whole `withEmoji` into the service definition if possible.

src/Sulu/Component/PHPCR/PathCleanup.php
nozarashi20: @alexander-schranz , this is [the condition](https://github.com/symfony/symfony/blob/6.2/src/Symfony/Component/Intl/Transliterator/EmojiTransliterator.php#L43) that triggers the `IntlException`. 

It happens when there is no [mapping file](https://github.com/symfony/symfony/tree/6.2/src/Symfony/Component/Intl/Resources/data/transliterator/emoji) for a locale. For instance, there is no `emoji-de_at.php` file; which leads to [this error](https://github.com/sulu/sulu/actions/runs/3554459366/jobs/5970575754#step:10:578)

alexander-schranz: Thank you for detailed debugging here. Let see if symfony will change here something. Created an issue here: https://github.com/symfony/symfony/issues/48364

Else I think we should just in the constructor created a `$this->sluggerWithEmoji` and `$this->sluggerWithoutEmoji` to avoid creating new objects always inside the loop.

nozarashi20: Done.

alexander-schranz: Think we not longer need the sluggerWithEmoji? Since https://github.com/symfony/symfony/pull/48395?

Also I think we should do `str_replace('-', '_', $languageCode)` to convert it into a valid intl locale. @nozarashi20 What do you think?

nozarashi20: Great @alexander-schranz , the code will be much simpler.
And I agree on replacing dashes with underscore. We'll then be able to handle codes like `en-US`. 
Have you already encounter this case ?

alexander-schranz: On other cases we did also just output the `language` this way `explode('-', $locale)[0]` but as it could be that the slugger support `en-us` and `en-uk` I would give the whole intl locale into it, if possible.

I think for general handling we should create an own issue how we maybe normalize the locale better in future. So maybe all things are handle by intl based format like `en_UK` problem is that most website don't want to have `domain.com/en_UK´ instead `domain.com/en-uk` in it. But as said that is an own issue by itself, so first concentrate on the slugger case.

nozarashi20: Updated code and tests

src/Sulu/Component/PHPCR/PathCleanup.php
```diff
@@ -70,6 +74,7 @@ public function cleanup($dirty, $languageCode = null)
         $replacers = $this->replacers['default'];
 
         if (null !== $languageCode) {
+            $languageCode = \str_replace('-', '_', $languageCode);
```
nozarashi20: @alexander-schranz , should I do this before or after the merge of `replacers` ? :thinking:

alexander-schranz: For BC reasons I would do it after the replacers.

nozarashi20: thanks, it's fixed

alexander-schranz (approved):

alexander-schranz:
@nozarashi20 Thank you! Good work 👍

PR #4119, opened 2018-08-22
Add SmartContent FieldType

danrot:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | ---
| Related issues/PRs | ---
| License | MIT
| Documentation PR | ---

#### What's in this PR?

This PR adds the field type for the smart content.

#### Why?

Because it is a very popular feature in Sulu, and we also need it in the new UI.

#### To Do

- [ ] Create a documentation PR
- [ ] Add breaking changes to UPGRADE.md
- [x] Add overlay to select DataSource
- [x] Add overlay to select Categories
- [x] Add proper translations with placeholders for the title of the `FilterOverlay`
- [x] Set correct values when opening `FilterOverlay`
- [x] Add `Dialog` type for `DatagridOverlay` (Dialog misses `confirmDisabled` and `size`)
- [x] Show selected data source in `FilterOverlay`
- [x] Show selected categories in `FilterOverlay`
- [x] Show selected tags in `FilterOverlay`
- [x] Go through `DatagridOverlay`s once more when #4116 is merged
- [x] Add `preselectedItems` to `DatagridOverlay`
- [x] Add `expandedIds` to Request for `categories` `DatagridOverlay`
- [x] Read provider configuration to only show valid options in `FilterOverlay`
- [x] Add `presentAs` to `FilterCriteria`
- [x] Check if form is shown immediately dirty after initial load with different `FilterOverlay` sections being enabled
- [x] Determine default values based on which `FilterOverlay` sections are enabled
- [x] Load data from server for preview (register as resourceKey although strange URL?)
- [x] Fix multiple requests send initially
- [x] Reset does not reset state of `MultiAutoComplete`
- [x] Defaults option for `provider` is not passed to 
- [x] Pass correct sortings to SmartContent component
- [x] Pass correct presentations to SmartContent component
- [x] Extract `ResourceRequester::buildQueryString` to `utils`?
- [x] Missing excluded response when assigning some values to content SmartContent in Page Add Form
- [x] Make smart content items not sortable
- [x] Fix loading sub collections in `MediaOverview`

PR #6921, opened 2022-11-30
Use the reload() method to trigger the reload in the ListStore

JakobTolkemit:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no 
| License | MIT

#### What's in this PR?

Use reload() instead of activate() to trigger the reload in the ListStore after changing the Structure strategy.

#### Why?

The reload was triggered by re setting the active item, but that doesn't work if there never was an active item.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/stores/ListStore.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/stores/ListStore.js
@@ -391,8 +391,8 @@ export default class ListStore {
         this.structureStrategy = structureStrategy;
 
         if (hadStructureStrategy) {
-            // force a reload with the currently active item to match new structure
-            this.activate(this.active.get());
+            // force a reload to match new structure
+            this.reload();
         }
     };
 
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/tests/stores/ListStore.test.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/tests/stores/ListStore.test.js
@@ -1792,9 +1792,10 @@ test('Should trigger a mobx autorun if activate is called with the same id', () 
     autorunDisposer();
 });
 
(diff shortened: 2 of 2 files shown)


alexander-schranz:
@JakobTolkemit Thank you for the pull request.

Can you show us how we can reproduce this bug? We would require some kind of tests so this bug does not appear again. And I think the bug also exist on `2.4`? Then we should target the `2.4` branch.

JakobTolkemit:
I actually based my branch of the 2.4 branch. Must have miss clicked when creating the PR.

As to how to reproduce this bug:
1. Go into the webspace page view
2. Select a webspace where you never selected a page
Screenshot 2022-11-30 at 12-38-13 SULU
3. Switch the list representation
Screenshot 2022-11-30 at 12-39-43 SULU

alexander-schranz:
@JakobTolkemit nice catch 👍 I think some tests need to be adopted so the new case is handled.

For reference the failing code was introduced in https://github.com/sulu/sulu/pull/4119/files#diff-cda55a5d95dea084f18b4912fbae1c79758cbe3f7d663216ee2a86691d42dd20. Sadly could not find out why there was used that mechanism over `reload`.

Is active item still working like expected?

JakobTolkemit:
Yes, with active items it still works like expected.

alexander-schranz (approved):

alexander-schranz:
@JakobTolkemit Thank you!

alexander-schranz:
@JakobTolkemit looks like auto merge didn't work as there are still some tests failing. Can you check it out what is going wrong here?

alexander-schranz:
Looks like the current CI is also failing on `2.4` branch because of a new version of `sanitze-html`: https://github.com/sulu/sulu/actions/runs/3677963358/jobs/6237159225

alexander-schranz:
@JakobTolkemit The issue with `sanitize-html` package is fixed. Can you rebase your branch?

PR #6928, opened 2022-12-06
Improving the role querying in user creation

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?
Using a more precise query to get a list of non-anonymous roles. Also adding a return type to the function.

#### Why?
Then we don't need to filter the roles with PHP and it technically also reduces the amount of data that needs to be handled.

#### Example Usage
There shouldn't be a change in usage since the method is private it doesn't break any backwards compatibility.

#### To Do

### DIFF
--- a/src/Sulu/Bundle/SecurityBundle/Command/CreateUserCommand.php
+++ b/src/Sulu/Bundle/SecurityBundle/Command/CreateUserCommand.php
@@ -350,25 +350,23 @@ class CreateUserCommand extends Command
     /**
      * Return the names of all the roles.
      *
-     * @return array
+     * @return array<string>
      *
      * @throws \RuntimeException If no roles exist
      */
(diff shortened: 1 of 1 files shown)


mamazu:
Does anyone know why the pipeline is broken?

alexander-schranz:
@mamazu just a new version of PHPStan, can be ignored need to be fixed on lowest supported version. :)

alexander-schranz:
Change look fine can you rebase your branch to `2.4`?

alexander-schranz (approved):

alexander-schranz:
@mamazu Thank you!

PR #6929, opened 2022-12-06
Update js dependencies babel-loader, postcss, styleguidist and stylelint packages

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? | yes 
| BC breaks? | no 
| Deprecations? | no  
| Fixed tickets | fixes parts of #6343   
| Related issues/PRs | https://github.com/sulu/skeleton/pull/202 
| License | MIT

#### What's in this PR?

Update js dependencies.

#### Why?

Using latest version of our dependencies to provide feature and bugfix support.


package.json
```diff
@@ -48,12 +48,12 @@
         "@testing-library/user-event": "^14.2.5",
         "@wojtekmaj/enzyme-adapter-react-17": "^0.6.1",
         "autoprefixer": "^10.4.7",
-        "babel-jest": "^26.3.0",
-        "babel-loader": "^8.0.6",
+        "babel-jest": "^29.3.0",
```
niklasnatter: i dont know if it makes a difference, but i think i would keep this on the same version as `jest`? unfortunately, we are stuck on `jest@26` for now

package.json
```diff
@@ -48,12 +48,12 @@
         "@testing-library/user-event": "^14.2.5",
         "@wojtekmaj/enzyme-adapter-react-17": "^0.6.1",
         "autoprefixer": "^10.4.7",
-        "babel-jest": "^26.3.0",
-        "babel-loader": "^8.0.6",
+        "babel-jest": "^26.6.3",
```
alexander-schranz, line 51: as discussed with @niklasnatter we need to keep babel-jest on ^26 as we are currently stuck with jest ^26 I require the latest version of ^26 here #6929

PR #6933, opened 2022-12-12
Fix phpstan baseline error and update jest snapshot for QRCode lib

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? | no 
| BC breaks? | no 
| Deprecations? | no  
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Fix phpstan baseline error.

#### Why?

Currently errors becauce MAILER_DSN is not defined.


.env
```diff
@@ -37,5 +37,5 @@ SULU_ADMIN_EMAIL=example@localhost
 ###< sulu/sulu ###
 
 ###> symfony/mailer ###
-# MAILER_DSN=null://null
+MAILER_DSN=null://null
```
alexander-schranz, line 40: phpstan did failed without it as it failed initial some command services but a mailer is required as swiftmailer is not longer registered

src/Sulu/Bundle/AdminBundle/Resources/js/components/QRCode/tests/__snapshots__/QRCode.test.js.snap
```diff
@@ -13,2214 +13,20 @@ Array [
   <svg
     class="qrcode"
     height="256"
-    viewBox="0 0 256 256"
```
alexander-schranz, line 16: Why this looks unexpected it still works like it should. We only added the viewBox as it did not work without it. Seems like they fixed the bug. For backward compatibility we still le our definition of `viewBox="0 0 256 256"` in it.

PR #6943, opened 2022-12-14
Change help and error messages from <label> to <div>

FlorianBoe:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | 
| Deprecations? | 
| Fixed tickets | 
| Related issues/PRs | #6444
| License | MIT
| Documentation PR | 

#### What's in this PR?

For help and error messages HTHL label elements are currently used, which is semantically incorrect and has an impact on the form accessibility. As a first step I replaced the labels with div elements.

#### Why?

To link help and error texts semantically with input elements the following structure could be used:

```html
<label for="/example">Example</label>
<input type="text" id="/example" aria-describedby="/example-hint /example-error">
<div id="/example-hint">Help message</div>
<div id="/example-error">Error message</div>
```

If interested, I can also make these additions.

#### To Do

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Form/Field.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Form/Field.js
@@ -141,13 +141,13 @@ class Field<T: string | number> extends React.Component<Props<T>> {
                     }
                     {children}
                     {description &&
-                        <label className={fieldStyles.descriptionLabel}>
+                        <div className={fieldStyles.descriptionLabel}>
                             {description}
-                        </label>
+                        </div>
                     }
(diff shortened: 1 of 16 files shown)


alexander-schranz:
@FlorianBoe Thank you! The `aria-describedby` sounds interesting not sure how easy it is to implement into the current components.

FlorianBoe:
I'll be happy to take a look at it, it should be pretty straightforward in the end. If the field has an ID, it should point to the help and error texts. When creating the field description, these are taken into account, if available.

The only limitation I see right now is that input elements don't know if the field has a description or a help text. This means that `aria-describedby` may point to nowhere. Not a best practice, but acceptable.

PR #6935, opened 2022-12-13
Fix icon button active hover state

Jupi007:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #6925
| License | MIT

### What's in this PR?

Add style for icon button active-hovered state

**Before:**

image
_normal, active, active hovered_

**After:**

Capture d’écran du 2022-12-03 19-05-00

PR #6944, opened 2022-12-15
Fix icon button active+hover state

### LINKED ISSUE #6925: Active hovered icon buttons are not correctly styled
| Q | A
| --- | ---
| Sulu Version | 2.5.4
| Browser Version | Firefox 107.0.1

# Actual Behavior

Hover an active icon button use_not active_ hover background color + _active_ text color:

image
_normal, active, active hovered_

# Expected Behavior

The button background color should be correctly styled.

# Steps to Reproduce

Go to the media page, and hover an active button (like the ones about grid/list view).

# Possible Solutions

The solution is to add a specific style for this case.
Use the `$scorpion: #585858` color looks good imo:

```scss
$iconBackgroundColorActiveHover: $scorpion;
// ...
.icon {
    // ...
    &:hover:not(:disabled) {
        /// ...
        &.active {
            background-color: $iconBackgroundColorActiveHover;
        }
    }
}
```

Capture d’écran du 2022-12-03 19-05-00

---

Also, off-topic but there are `$iconBackgroundColor***` vars which are not used on top of the `button.scss` file:

https://github.com/sulu/sulu/blob/8ed3425211e8b72b4ac7ecc5239842408927fc42/src/Sulu/Bundle/AdminBundle/Resources/js/components/Button/button.scss#L24-L25

Jupi007:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #6925
| License | MIT

### What's in this PR?

Add style for icon button active-hovered state

**Before:**

image
_normal, active, active hovered_

**After:**

Capture d’écran du 2022-12-03 19-05-00

_Redo of #6935, for Sulu `2.4`_

@alexander-schranz I am sorry for unnecessary noise cause to my error with the target branch.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Button/button.scss
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Button/button.scss
@@ -150,6 +150,10 @@ $iconColorActive: $white;
 
     &:hover:not(:disabled) {
         background-color: $silver;
+
+        &.active {
+            background-color: $scorpion;
+        }
     }
 
     &:disabled {


alexander-schranz:
@Jupi007 can you repush or check in your fork if Github disabled maybe some workflow triggers. (Had that yesterday that I need under https://github.com/Jupi007/sulu/actions enable workflow again) As it seems the CI don't get started.

Jupi007:
Done, the actions was indeed disabled :+1:

alexander-schranz:
@Jupi007 Thank you. Github seems to disable now all which has a `schedule` workflow in it after some time, strangely they not only disable the schedule instead just all workflows 🙈

Jupi007:
Yeah, this is a weird behaviour. Very annoying as a lot contributions comes from forks 😅

PR #6951, opened 2022-12-15
Make table actions reachable for keyboards

FlorianBoe:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | 
| Related issues/PRs | #6444
| License | MIT
| Documentation PR | 

#### What's in this PR?

I've restructured the CSS code a bit to allow the table actions to be visible when focussing using the keyboard. Since `visibility` also removes the element from the tab order, I had to solve the show and hide via the `opacity`.

#### Why?

Until now, entries in the list and tree list could only be edited using the mouse.


alexander-schranz (approved):

alexander-schranz:
@FlorianBoe Thank you! 👍

PR #6953, opened 2022-12-15
Wrap masonry item in unordered list element

FlorianBoe:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no 
| Deprecations? | no
| Fixed tickets | 
| Related issues/PRs | #6444
| License | MIT
| Documentation PR | 

#### What's in this PR?

The MasonryItem is defined as a list item element, so the container should be an list (ul or ol) to provide correct relations.
CSS already sets `list-style: none`, is there a reason why a div element was used here?


alexander-schranz (approved):

alexander-schranz:
@FlorianBoe Good catch, thx for the pull request 👍

PR #6958, opened 2022-12-19
Move admin build to dist directory

### LINKED ISSUE #5216: WebpackEncoreBundle conflict with public/build/admin assets
| Q | A
| --- | ---
| Sulu Version | 2.0.6
| Browser Version | any

# Actual Behavior

After install `symfony/webpack-encore-bundle` to manage my frontend assets, every time that I run `$ yarn encore dev` command my backend (admin) assets are broken.

# Expected Behavior

I need to keep both (/public/build/admin and /public/build) compiled assets folders to ensure that admin area is working well. 

# Steps to Reproduce

* install sulu/skeleton
* composer require symfony/webpack-encore-bundle
* yarn install
* check that webpack.config.js file is installed in project root dir
*  yarn encore dev

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? |yes
| New feature? | no
| BC breaks? | no 
| Deprecations? | no 
| Fixed tickets | fixes #5216
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Move admin build to dist directory.

#### Why?

Avoid conflict with webpack encore or asset mapper.

Webpack Encore uses: `public/build`
Assetmapper uses: `public/assets`

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -2,7 +2,7 @@
 **/vendor/
 **/flow-typed/
 styleguide/
-**/build/admin/
+**/dist/admin/
 
 .git/
 .github/
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@ phpunit.stdout
 /public/bundles
 /public/uploads
 /public/build
+/public/dist
 
 # composer
 /composer.phar
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -6,6 +6,11 @@
 
 The admin build directory was moved from `public/build/admin` to `public/dist/admin` to avoid conflicts with webpack encore.
 
+```bash
+mkdir public/dist/
+git mv public/build/admin public/dist/admin
+```
+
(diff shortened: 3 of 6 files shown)


alexander-schranz:
@chirimoya Did mention that he would maybe go with `/dist/build/admin`. @sulu/core-team what do you think?

wachterjohannes:
> @chirimoya Did mention that he would maybe go with `/dist/build/admin`. @sulu/core-team what do you think?

would go with `public/dist`

alexander-schranz:
@chirimoya Should we do this in `2.6`? Or only in the major 3.0 upgrade? 

----

I did lately setup a assetmapper project there a different directory is also used which we need avoid for our build:

Webpack Encore uses: `public/build`
Assetmapper uses: `public/assets`

One other way could be we use something like `public/cms` as a build directory.

chirimoya:
@alexander-schranz let's discuss on Monday.

PR #6959, opened 2022-12-19
Fix webpack build and slim dependency

alexander-schranz:
| Q | A
| --- | ---
| Bug fix?  yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | https://github.com/sulu/skeleton/pull/211
| License | MIT

#### What's in this PR?

Fix webpack build and slim dependency.

#### Why?

A lot of babel plugins where just required for webpack 4. We could remove them to as we upgraded to webpack 5. Even the latest change with `htmlparser2` actually does not work for webpack 5 just for 4. So we are not longer run that package and other via babel.

### DIFF
--- a/package.json
+++ b/package.json
@@ -112,7 +112,7 @@
         ],
         "clearMocks": true,
         "transformIgnorePatterns": [
-            "node_modules/(?!(@ckeditor|ckeditor5|array-move)/)"
+            "node_modules/(?!(@ckeditor|ckeditor5|array-move|lodash-es)/)"
         ],
         "testPathIgnorePatterns": [
             "vendor/friendsofsymfony"
--- a/package.json
+++ b/package.json
@@ -112,7 +112,7 @@
         ],
         "clearMocks": true,
         "transformIgnorePatterns": [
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -115,7 +112,7 @@
         ],
         "clearMocks": true,
         "transformIgnorePatterns": [
-            "node_modules/(?!(@ckeditor|ckeditor5|array-move|lodash-es|@react-leaflet|react-leaflet)/)"
+            "node_modules/(?!(@ckeditor|ckeditor5|array-move|lodash-es)/)"
```
alexander-schranz, line 115: `array-move` and `lodash-es` still required to be transformed for running `jest` tests. Link from maintainer of array-move about esm packages: https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c

Related Jest issue: https://github.com/facebook/jest/issues/9430

Prokyonn (approved):

PR #6960, opened 2022-12-20
Make form of address in contact form gender neutral

luca-rath:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

This PR adds a `FormOfAddressProvider` which provides the options for the respective values in the contact form. That service can be overridden to adjust the possible values for that field.

#### Why?

To make form of address in contact form gender neutral

### DIFF
--- a/src/Sulu/Bundle/ContactBundle/Resources/config/forms/contact_details.xml
+++ b/src/Sulu/Bundle/ContactBundle/Resources/config/forms/contact_details.xml
@@ -46,7 +46,7 @@
                     </params>
                 </property>
 
-                <property name="formOfAddress" type="single_select" colspan="3">
+                <property name="formOfAddress" type="single_select" mandatory="true" colspan="3">
                     <meta>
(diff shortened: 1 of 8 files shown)


src/Sulu/Bundle/ContactBundle/Resources/config/forms/contact_details.xml
```diff
@@ -46,24 +46,12 @@
                     </params>
                 </property>
 
-                <property name="formOfAddress" type="single_select" mandatory="true" colspan="3">
+                <property name="formOfAddress" type="single_select" colspan="3">
```
luca-rath, line 49: I don't see why this field should be mandatory, because the database column isn't mandatory

alexander-schranz (approved):

PR #6962, opened 2022-12-25
[DX] Adding types for list metadata configuration

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

Using proper typing for the sulu list metadata objects. Many of them can be converted to readonly objects in PHP 8.1

#### Why?

Two reasons:
* Typing makes it easier to see what values are actually expected from the config
* It makes it easier to provide an option to accept php as an alternative configuration language
* I found a bug by adding types

#### Example Usage

This should not change the usage of the value objects at all.

### DIFF
--- a/src/Sulu/Component/Rest/ListBuilder/Metadata/AbstractPropertyMetadata.php
+++ b/src/Sulu/Component/Rest/ListBuilder/Metadata/AbstractPropertyMetadata.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 /*
  * This file is part of Sulu.
  *
--- a/src/Sulu/Component/Rest/ListBuilder/Metadata/AbstractPropertyMetadata.php
+++ b/src/Sulu/Component/Rest/ListBuilder/Metadata/AbstractPropertyMetadata.php
@@ -18,195 +18,120 @@ use Sulu\Component\Rest\ListBuilder\FieldDescriptorInterface;
  */
 abstract class AbstractPropertyMetadata
 {
-    /**
-     * @var string
-     */
-    private $name;
+    private string $name;
 
-    /**
-     * @var string
-     */
-    private $translation;
+    private string $translation;
 
-    /**
-     * @var string
-     */
-    private $visibility = FieldDescriptorInterface::VISIBILITY_NEVER;
+    private string $visibility = FieldDescriptorInterface::VISIBILITY_NEVER;
 
-    /**
-     * @var string
-     */
-    private $searchability = FieldDescriptorInterface::SEARCHABILITY_NEVER;
+    private string $searchability = FieldDescriptorInterface::SEARCHABILITY_NEVER;
 
-    /**
-     * @var string
-     */
-    private $type = 'string';
+    private string $type = 'string';
 
-    /**
-     * @var bool
-     */
-    private $sortable = true;
+    private bool $sortable = true;
 
-    /**
-     * @var array
-     */
-    private $transformerTypeParameters;
+    /** @var array<mixed> */
+    private array $transformerTypeParameters;
 
-    /**
-     * @var string
-     */
-    private $filterType;
+    private string $filterType;
 
-    /**
-     * @var array
-     */
-    private $filterTypeParameters;
+    /** @var array<mixed> */
+    private array $filterTypeParameters;
 
-    /**
-     * @var string
-     */
-    private $width;
+    private string $width;
 
-    public function __construct($name)
+    public function __construct(string $name)
     {
         $this->name = $name;
         // default for translation can be overwritten by setter
         $this->translation = \ucfirst($name);
     }
 
-    /**
-     * @return string
-     */
-    public function getName()
+    public function getName(): string
     {
         return $this->name;
     }
 
-    /**
-     * @return string
-     */
-    public function getTranslation()
+    public function getTranslation(): string
     {
         return $this->translation;
     }
 
-    /**
-     * @param string $translation
-     */
-    public function setTranslation($translation)
+    public function setTranslation(string $translation): void
     {
         $this->translation = $translation;
     }
 
-    /**
-     * @return string
-     */
-    public function getVisibility()
+    public function getVisibility(): string
     {
         return $this->visibility;
     }
 
-    /**
-     * @param string $visibility
-     */
-    public function setVisibility($visibility)
+    public function setVisibility(string $visibility): void
     {
         $this->visibility = $visibility;
     }
 
-    /**
-     * @return string
-     */
-    public function getSearchability()
+    public function getSearchability(): string
     {
         return $this->searchability;
     }
 
-    /**
-     * @param string $searchability
-     */
-    public function setSearchability($searchability)
+    public function setSearchability(string $searchability): void
     {
         $this->searchability = $searchability;
     }
 
-    /**
-     * @return string
-     */
-    public function getType()
+    public function getType(): string
     {
         return $this->type;
     }
 
-    /**
-     * @param string $type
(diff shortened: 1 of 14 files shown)


src/Sulu/Component/Rest/ListBuilder/Metadata/FieldDescriptorFactory.php
```diff
@@ -223,7 +228,7 @@ private function getCountFieldDescriptor(
             $propertyMetadata->getSearchability(),
             $propertyMetadata->getType(),
             $propertyMetadata->isSortable(),
-            $this->resolveOptions($propertyMetadata->getDistinct(), $options),
+            $propertyMetadata->getDistinct(),
```
mamazu, line 236: This code previously only worked because of two things:
* either the options array is empty then `resolveOptions` will return the value as is (and keep it as a boolean
* sulu doesn't use strict types and empty strings are false, and filled strings are true.

src/Sulu/Component/Util/XmlUtil.php
```diff
@@ -45,14 +45,14 @@ public static function getValueFromXPath($path, \DOMXPath $xpath, \DOMNode $cont
...
+     * @return bool|null
      */
-    public static function getBooleanValueFromXPath($path, \DOMXPath $xpath, \DOMNode $context = null, $default = null)
+    public static function getBooleanValueFromXPath($path, \DOMXPath $xpath, \DOMNode $context = null, $default = null): ?bool
     {
         $value = self::getValueFromXPath($path, $xpath, $context, $default);
 
         if (null === $value) {
-            return;
+            return null;
```
mamazu, line 59: This doesn't change the behavior of the function. It only makes it explicit.

```php
function f() {
}

var_dump(f()); // NULL
```

src/Sulu/Component/Util/XmlUtil.php
```diff
@@ -45,14 +45,14 @@ public static function getValueFromXPath($path, \DOMXPath $xpath, \DOMNode $cont
      * @param string $path
      * @param \DOMNode $context
      *
-     * @return bool|null|string|mixed
+     * @return bool|null
      */
-    public static function getBooleanValueFromXPath($path, \DOMXPath $xpath, \DOMNode $context = null, $default = null)
+    public static function getBooleanValueFromXPath($path, \DOMXPath $xpath, \DOMNode $context = null, $default = null): ?bool
```
mamazu: Everywhere in the code the default value is provided and it would make sense that it is a boolean as well. I would suggest to have the `$default` parameter be typed and default to false (which would be null as a boolean under non-strict types).

src/Sulu/Component/Util/XmlUtil.php
```diff
@@ -21,10 +21,8 @@ class XmlUtil
      *
      * @param string $path
      * @param \DOMNode $context
-     *
-     * @return bool|null|string|mixed
      */
-    public static function getValueFromXPath($path, \DOMXPath $xpath, \DOMNode $context = null, $default = null)
+    public static function getValueFromXPath($path, \DOMXPath $xpath, \DOMNode $context = null, $default = null): mixed
     {
         $result = $xpath->query($path, $context);
```
mamazu, line 31: What if `$result` is `false` and the node couldn't be found? I think currently this should throw an error because calling `->length` on `false` is not allowed. There should probably be an assert here if we want to have a better error message or just return a the default?

src/Sulu/Component/Rest/ListBuilder/Metadata/AbstractPropertyMetadata.php
```diff
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
```
alexander-schranz: We are currently not adding this. As this would be a bc break.

alexander-schranz (changes requested):
Like to move this forward but we need convert public method changes to phpdoc. That should also fix the same reported phpstan issues.

src/Sulu/Component/Rest/ListBuilder/Metadata/AbstractPropertyMetadata.php
```diff
@@ -18,195 +20,120 @@
...
+    public function getTranslation(): string
     {
         return $this->translation;
     }
 
-    /**
-     * @param string $translation
-     */
-    public function setTranslation($translation)
+    public function setTranslation(string $translation): void
```
alexander-schranz: The return and method parameter types of public and protected methods need to be moved to phpdoc so we are not doing here any unneeded bc breaks.

mamazu:
Yeah, makes sense. I will revert all BC breaks. Are there any plans like maybe for the next major version to move to 7.4 typing or even 8.0 features to make the code shorter and more robust?

alexander-schranz:
Something I hope we will target with 3.0. But every correctly defined `PHPDoc` make this migration easier and phpstan already can help here every project. So it is sure nice to check every existing Entities and value objects like this Metadata for correct types. Specially I think a lot of `|null` are missing around entities.

mamazu:
Sounds great. If you need help with that for the 3.0 release I can help with that.

alexander-schranz:
@mamazu if you want to check current `phpdocs` of entities properties and entities return methods (based on doctrine.orm.xml files it sure something which would improve a lot into typesafety. I think there are still `string` defined where `nullable=true` is set in doctrine and so there it should be `string|null`. Updating and fixing the current PHPDoc already helps.

src/Sulu/Component/Rest/ListBuilder/Metadata/SinglePropertyMetadata.php
```diff
@@ -16,10 +16,7 @@
  */
 class SinglePropertyMetadata extends AbstractPropertyMetadata
 {
-    /**
-     * @var ?FieldMetadata
-     */
-    private $field;
+    private ?FieldMetadata $field = null;
```
mamazu, line 19: This can be fixed. Because the setter and getter both are typed and the property is private. So unless you are doing something very unorthodox this shouldn't be a bc break.

src/Sulu/Component/Rest/ListBuilder/Metadata/IdentityPropertyMetadata.php
```diff
@@ -16,11 +16,11 @@
...
 {
+    private ?FieldMetadata $field;
+
     /**
-     * @var ?FieldMetadata
+     * @return void
      */
-    private $field;
-
     public function setField(?FieldMetadata $field)
```
mamazu, line 24: That highlight is a little confusing, the doc comment now belongs to the method and the field is now doc comment free. (it's private with accesstor's that already had types in).

mamazu:
So I have removed all hard types that could be a bc break and all the `declare(strict_types=1)`

src/Sulu/Component/Util/XmlUtil.php
mamazu: I have solved it by adding an assertion that has a more helpful error message.

mamazu:
@alexander-schranz Anything else that needs changing?

mamazu:
Wouldn't it make sense to add them as typehints when rebasing the branch to 3.0?

alexander-schranz (approved):
Looks fine from my side @wachterjohannes can you have another look at it if you don't see something we should not cast here?

wachterjohannes (approved):

alexander-schranz:
@mamazu Thank you!

PR #6964, opened 2022-12-30
 Move config from inline js to data attr

### LINKED ISSUE #6354: Avoid inline Javascript in Admin main template
# Problem description

Currently the admin main template is setting the SULU_CONFIG via a inline javascript file: https://github.com/sulu/sulu/blob/2.3.6/src/Sulu/Bundle/AdminBundle/Resources/views/Admin/main.html.twig

# Proposed solution

Instead of setting this configuration VIA javascript the startAdmin component could read this configuration from the DOM itself e.g.:

```twig
<div id="application" data-config="{{ config|json_encode }}"></div>
```

This will make it in future possible to use a strict CSP headers for the admin when [lodash issue](https://github.com/lodash/lodash/issues/832) is solved.

vvasiloi:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | fixes #6354 
| Related issues/PRs | 
| License | MIT
| Documentation PR |

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/index.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/index.js
@@ -389,6 +389,20 @@ function startAdmin() {
         'font-family: monospace; font-size: 10px; text-decoration: none;'
     );
 
+    const id = 'application';
+    const applicationElement = document.getElementById(id);
+
+    if (!applicationElement) {
+        throw new Error(`DOM element with ID "${id}" was not found!`);
+    }
+
+    if (!('config' in applicationElement.dataset)) {
+        throw new Error('Configuration not found');
+    }
+
+    Object.assign(Config, JSON.parse(applicationElement.dataset.config));
+    Object.freeze(Config);
+
     if (Config.suluVersion !== SULU_ADMIN_BUILD_VERSION) {
         log.error(
             'Sulu administration interface: JavaScript build version mismatch' +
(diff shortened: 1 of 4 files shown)


alexander-schranz:
We will target the `2.6` branch as for this. Can you rebase it. The current PHP 8.1 test failing by doctrine can be ignore it is related to https://github.com/doctrine/dbal/issues/5837

vvasiloi:
Oh, I used the wrong base branch...

src/Sulu/Bundle/AdminBundle/Resources/js/index.js
```diff
@@ -389,6 +389,19 @@ function startAdmin() {
...
+
+    if (!applicationElement) {
+        throw new Error(`DOM element with ID "${id}" was not found!`);
+    }
+
+    if (!('config' in applicationElement.dataset)) {
+        throw new Error('Configuration not found');
+    }
+
+    Object.assign(Config, JSON.parse(applicationElement.dataset.config));
```
alexander-schranz, line 403: Looks like we freezed the Config object in the main.html.twig currently. We should do the same currrently here.

```suggestion
    Object.assign(Config, JSON.parse(applicationElement.dataset.config));
    Object.freeze(Config);
```

alexander-schranz (changes requested):
Some small change request. Else looks good and the AdminControllerTest looks a lot better and easier understandable now. 👍

alexander-schranz (approved):

alexander-schranz:
@vvasiloi Nice, thank you 👍

PR #6966, opened 2022-12-30
Adding the correct nullable types to entities

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

As discussed here https://github.com/sulu/sulu/pull/6962#issuecomment-1367611084 we needed to check that the XML Definition is in Line with the definition of the entity types. This PR fixes that.

Furthermore I have also attached the script that detected the inconsistencies. I don't know if we want to keep it or not.

#### Why?

see above

#### Example Usage

--

#### To Do

--


alexander-schranz:
@mamazu Thank you 🙏

PR #6973, opened 2023-01-14
Add "AsCommand" annotation for deprecated "$defaultName" for all commands

StaffNowa:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | yes
| Fixed tickets | fixes #
| Related issues/PRs | #
| License | MIT

Since symfony/console 6.1: Relying on the static property "$defaultName" for setting a command name is deprecated.

### DIFF
--- a/src/Sulu/Bundle/CategoryBundle/Command/RecoverCommand.php
+++ b/src/Sulu/Bundle/CategoryBundle/Command/RecoverCommand.php
@@ -17,15 +17,15 @@ use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Attribute\AsCommand;
 
 /**
  * Command for recovering categories.
  * This command is fixing wrong left/right and depths (see -d) assignments of the categories tree.
  */
+#[AsCommand(name: 'sulu:categories:recover')]
 class RecoverCommand extends Command
 {
-    protected static $defaultName = 'sulu:categories:recover';
-
     /**
      * @var EntityManagerInterface
      */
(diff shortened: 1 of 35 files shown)


src/Sulu/Bundle/CategoryBundle/Command/RecoverCommand.php
```diff
@@ -17,15 +17,15 @@
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Attribute\AsCommand;
 
 /**
  * Command for recovering categories.
  * This command is fixing wrong left/right and depths (see -d) assignments of the categories tree.
  */
+#[AsCommand(name: 'sulu:categories:recover')]
```
alexander-schranz, line 26: Is this now the only way to define this or can it be defined this also on service definition level, when there are other ways I would like to discuss it with the other core members which way we would like to go. 

If I read the docs about it, it maybe make also sense to move descriptions to it?

src/Sulu/Bundle/ContactBundle/Command/AccountRecoverCommand.php
```diff
@@ -17,15 +17,15 @@
...
 
 /**
  * Command for recovering nested tree of accounts.
  * This command is fixing wrong left/right and depths (see -d) assignments of the nested tree.
  */
+#[AsCommand(name: 'sulu:contacts:accounts:recover')]
 class AccountRecoverCommand extends Command
 {
-    protected static $defaultName = 'sulu:contacts:accounts:recover';
-
```
alexander-schranz, line 28: Would document this in the `UPGRADE.md` that `$defaultName` of the commands where removed. As it would be real edge case single line is enough, just if somebody really did access it which I think is not the case.

alexander-schranz (approved):

alexander-schranz:
@StaffNowa Thank you!

PR #6982, opened 2023-01-25
Add support for Accept Language in Geolocators

dev-newvisibility:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Both Nominatim and Google Geolocator support additional parameters to optimize the result of the query, this PR adds a new optional parameter to the GeolocatorInterface::locate method to pass additional options specific for a single query.

As an initial implementation this PR adds the ability to pass to the geolocator an "accept language" string.

#### Why?

Google Geolocator, if you don't pass the language query string parameter or an Accept-Language http header, tries to infer the desired results' language from the remote IP that it's calling its API, but the server in which runs the Sulu instance may not be located in the same country of the user, this gives a wrong output (wrong labels and/or wrong order).

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -206,8 +206,7 @@ Or the Merge request implementing this changes in Sulu [here](https://github.com
 
 ### GeolocatorInterface locate method GeolocatorOptions added
 
-To support to give the Accept-Language locale to the geolocator services a custom Geolocator
-requires now to support the new `GeolocatorOptions` parameter:
+To provide the Accept-Language locale to geolocator services, a custom Geolocator now requires support for the new ﻿`GeolocatorOptions` parameter:
 
 ```diff
 -    public function locate(string $query): GeolocatorResponse
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -204,6 +204,16 @@ Example:
 See also the documentation [official Doctrine Events documentation](https://symfony.com/doc/6.4/doctrine/events.html).  
 Or the Merge request implementing this changes in Sulu [here](https://github.com/sulu/sulu/pull/7374/files).
 
+### GeolocatorInterface locate method GeolocatorOptions added
+
+To support to give the Accept-Language locale to the geolocator services a custom Geolocator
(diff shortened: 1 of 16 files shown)


alexander-schranz:
I'm here a little bit splitted what the correct language to send to this controller should be. Is it really the accept language of the browser sending us or when used inside translateable form if we should send then the language to the controller of that content. That would mean on a german form they may need to input `Wien` and on the english `Vienna` for the same result. Or if we stay with the browser language it even could be if the browser language is `fr` even on german and english the best result would be the user then inputting `Vienne`.  🤔 

@sulu/core-developer what do you think?

Prokyonn:
> I'm here a little bit splitted what the correct language to send to this controller should be. Is it really the accept language of the browser sending us or when used inside translateable form if we should send then the language to the controller of that content. That would mean on a german form they may need to input `Wien` and on the english `Vienna` for the same result. Or if we stay with the browser language it even could be if the browser language is `fr` even on german and english the best result would be the user then inputting `Vienne`. 🤔
> 
> @sulu/core-developer what do you think?

By default, I would recommend using the locale of the request to avoid inconsistencies, such as displaying an English address in a German format. However, I would still provide the option to always use the browser language, in case the project requires it.

UPGRADE.md
```diff
@@ -204,6 +204,16 @@ Example:
 See also the documentation [official Doctrine Events documentation](https://symfony.com/doc/6.4/doctrine/events.html).  
 Or the Merge request implementing this changes in Sulu [here](https://github.com/sulu/sulu/pull/7374/files).
 
+### GeolocatorInterface locate method GeolocatorOptions added
+
+To support to give the Accept-Language locale to the geolocator services a custom Geolocator
+requires now to support the new `GeolocatorOptions` parameter:
```
```suggestion
To provide the Accept-Language locale to geolocator services, a custom Geolocator now requires support for the new ﻿`GeolocatorOptions` parameter:
```

Prokyonn (approved):

PR #6989, opened 2023-02-06
Check if security->getUser() is instance of UserInterface

0xJas0n:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets |
| Related issues/PRs |
| License | MIT
| Documentation PR |

#### What's in this PR?

Return null or UserInterface from security->getUser() instead of InMemoryUser.

#### Why?

While trying to delete articles via the documentManager we receive the following error message:
TrashItemRepository::getCurrentUser(): Return value must be of type ?Sulu\Component\Security\Authentication\UserInterface

### DIFF
--- a/src/Sulu/Bundle/TrashBundle/Infrastructure/Doctrine/Repository/TrashItemRepository.php
+++ b/src/Sulu/Bundle/TrashBundle/Infrastructure/Doctrine/Repository/TrashItemRepository.php
@@ -122,6 +122,10 @@ final class TrashItemRepository implements TrashItemRepositoryInterface
         /** @var UserInterface $user */
         $user = $this->security->getUser();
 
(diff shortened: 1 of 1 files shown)


src/Sulu/Bundle/TrashBundle/Infrastructure/Doctrine/Repository/TrashItemRepository.php
```diff
@@ -122,6 +122,10 @@ private function getCurrentUser(): ?UserInterface
         /** @var UserInterface $user */
```
alexander-schranz: The `@var` should be removed:

```suggestion
```

Issue #6842, opened 2022-10-06
Deprecation-Marking for page templates


picturestone:
# Problem description

With longer running projects, sometimes a redesign for page templates is done. Then the content can be migrated manually from old page templates to the new templates over some time. If its a big website, this migration can take some time. During this time both the old and new templates need to be available so visitors of the website can at least see the content in the old template until it's migrated.

To communicate to the content managers which page templates are new and which ones are old, the meta-title of the template can be changed. For example, the `detail page` might be called `detail page DEPRECATED` until all the content is migrated to another page. However, it would be nicer if the content manager gets a visual cue instead of a really long template name in the admin UI.

# Proposed solution

I propose a xml property which can be set by developers in the xml to mark a template as deprecated. Maybe it can be added inside the meta-tag. The second last line `<param name="isDeprecated" value="true" />` would mark the following detail-template as deprecated:

```
<template xmlns="http://schemas.sulu.io/template/template"
          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xmlns:xi="http://www.w3.org/2001/XInclude"
          xsi:schemaLocation="http://schemas.sulu.io/template/template http://schemas.sulu.io/template/template-1.0.xsd">

    <key>detail</key>

    <view>pages/detail</view>
    <controller>Sulu\Bundle\WebsiteBundle\Controller\DefaultController::indexAction</controller>
    <cacheLifetime>86400</cacheLifetime>

    <meta>
        <title lang="en">Detail Page</title>
        <title lang="de">Detailseite</title>
        <param name="isDeprecated" value="true" />
    </meta>
...
```

Not having a value at all would mean that the page is not deprecated. 

In the admin, the names of deprecated templates could have some icons in front when creating new templates. In the following screenshot, the "Detailseite" page template would be marked as deprecated. An icon is shown to the content manager which signals that this template should no longer be used:

<img width="870" alt="New_Page" src="https://user-images.githubusercontent.com/6418474/194333835-4f121fff-389f-42e5-b6bb-68347e555373.png">

In the page tree there could also be some icons so the content managers can see which pages still need to be migrated. The pages called "Old Details" and "More old details" still use a deprecated page template, whereas the page called "Migrated details" uses a non-deprecated page template.

<img width="965" alt="Page_Tree" src="https://user-images.githubusercontent.com/6418474/194333831-00bfd5cf-17c7-4c98-aed1-3608ecf942dc.png">

[...]

PR #6992, opened 2023-02-07
Cleaning up the StructureLoading

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | #6842
| License | MIT
| Documentation PR | -none-

#### What's in this PR?

* Adding types to some function in the StructureXmlLoader
* Merging the functions to find reserved properties
* Moving the template attributes out that are the same for every template
* Using php8 syntax to filter an array

#### Why?

Less code to maintain and easier to see what attributes are global for all templates and what aren't.

In this case adding deprecated attributes would affect all structure types in sulu.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -44856,17 +44856,7 @@ parameters:
 			path: src/Sulu/Component/Content/Metadata/Loader/StructureXmlLoader.php
 
 		-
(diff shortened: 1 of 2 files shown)


mamazu:
Since Sulu 3 already has different plans, this merge request will be closed.

PR #7004, opened 2023-02-13
Remove reference to LiipThemeBundle and dispatch event instead

rogamoore:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #
| Related issues/PRs | #
| License | MIT

#### What's in this PR?

The SuluThemeBundle switched to SyliusThemeBundle some time ago. See https://github.com/sulu/SuluThemeBundle/pull/19

~~This PR updates the ValidateWebspacesCommand to use the correct components to set the theme and use it when checking for the existence of templates.~~

This PR updates the ValidateWebspacesCommand to dispatch a PreRenderEvent instead which causes the SuluThemeBundle to correctly set the correct theme.

#### Why?

Without this fix, when using themes, the command will report missing templates even if the templates exist.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,11 @@
 # Upgrade
 
+## 2.5.7
+
+### Constructor of ValidateWebspacesCommand changed
+
+The constructor of `ValidateWebspacesCommand` requires now EventDispatcherInterface instead of activeTheme.
+
 ## 2.5.2
 
 ### Add indexes to route table
--- a/src/Sulu/Bundle/PageBundle/Command/ValidateWebspacesCommand.php
+++ b/src/Sulu/Bundle/PageBundle/Command/ValidateWebspacesCommand.php
@@ -11,8 +11,11 @@
 
 namespace Sulu\Bundle\PageBundle\Command;
 
+use Sulu\Bundle\PreviewBundle\Preview\Events;
+use Sulu\Bundle\PreviewBundle\Preview\Events\PreRenderEvent;
 use Sulu\Component\Content\Compat\StructureManagerInterface;
 use Sulu\Component\Content\Metadata\Factory\StructureMetadataFactoryInterface;
+use Sulu\Component\Webspace\Analyzer\Attributes\RequestAttributes;
 use Sulu\Component\Webspace\Manager\WebspaceManagerInterface;
 use Sulu\Component\Webspace\StructureProvider\WebspaceStructureProvider;
 use Sulu\Component\Webspace\Webspace;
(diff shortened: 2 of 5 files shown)


alexander-schranz:
@rogamoore the problem is we should not have any requirement to SyliusThemeBundle inside Sulu Core. So it would be more abstract and extendable if we would inject `eventDispatcher` and dispatch am event which will be thrown and SuluThemeBundle could listen to it. This way even custom theming and custom template loaders can be used and they can just listen to that event to make the validator pass.

alexander-schranz:
We also could use existing event from the Preview the `PreRenderEvent`: https://github.com/sulu/SuluThemeBundle/blob/d08918569fe5cb261a16efa5f816fbe9f0153ee7/EventListener/SetThemeEventListener.php#L67 to be thrown here as that event already used by the SuluThemeBundle to set the correct theme.

rogamoore:
@alexander-schranz I was thinking the same. Actually the whole bug is based on the hard coded reference to `LiipThemeBundle`. This PR was meant as a quick fix because I stumbled on this problem when starting to use themes.

If you think we could re-use the `PreRenderEvent`, I could change the PR to dispatch the event as it has been done here:
https://github.com/sulu/sulu/blob/2.5/src/Sulu/Bundle/PreviewBundle/Preview/Renderer/PreviewRenderer.php#L187
PS:  I'm wondering now what the TODO comment is about :)

Regarding tests I guess we could check somehow if the event has been dispatched here in `testExecute`: https://github.com/sulu/sulu/blob/2.5/src/Sulu/Bundle/PageBundle/Tests/Functional/Command/ValidateWebspacesCommandTest.php#L54

alexander-schranz:
We did keep the LiipthemeBundle in it because it was that case in the past. But with the Min Requirement to Symfony 5.4 it can completely removed as it can not longer exist in Symfony 5.4. And for SyliusTheme we should find a more general solution so we have not a dependency to a service which is actually not required by the composer.json of sulu.

> PS: I'm wondering now what the TODO comment is about :)

I'm even not sure about the comment. I even think that it is with the change to SyliusThemeBundle again required to dispatch that event. But I would need to test it if we remove that event if theme is still correctly be used. But as it is still there I'm also fine to keep it if we have a usecase for it and remove the comment and deprecation.

> Regarding tests I guess we could check somehow if the event has been dispatched here in testExecute: 

From test point of view I think it make sense to check if the event was dispatched with correct webspace attribute.

alexander-schranz:
The CI errors seems unrelated to the changes here. I will have a look at it.

alexander-schranz:
I tested it out locally all works like expected for current stable and current `6.2@dev` release. So the issue is somewhere in Symfony `6.3@dev` branch, I created an issue in Symfony for it https://github.com/symfony/symfony/issues/49387.

I also updated the UPGRADE.md file, it is a small bc break in the constructor but not worth investing here as I don't think somebody changed here anything.

rogamoore:
Thanks a lot @alexander-schranz !

PR #7005, opened 2023-02-14
Catch MediaNotFoundException in MediaRedirectController

### LINKED ISSUE #6924: MediaNotFoundException is not catched in media redirect
| Q | A
| --- | ---
| Sulu Version | 2.5.4
| PHP Version | 8.1.11

# Actual Behavior

Trying to access to the `sulu_media.redirect` route with a non existent media id, throw a `MediaNotFoundException` which is not caught by the controller.

# Expected Behavior

I could be wrong, but I think this error should be caught to throw a `NotFoundHttpException` instead.
This mean that the http code should be 404 and not 500.

# Steps to Reproduce

In a raw way: access to `/admin/media/redirect/media/{id}?locale={locale}`, with a non existent media id.

In fact, I'm using a monolog email sender to report errors, and sometimes, when someone remove a media, this exception is throw shortly after, and I'm receiving an email (I received many ones each time for a different media which was recently removed).
So I'm sorry, but I don't know the exact way to reproduce this error _in Sulu admin itself_. I'm investigating, but still not very familiar with Sulu, so I preferred to report this problem as it in priority.

<del>
# Possible Solutions

With this simple patch, the 500 error is converted in a 404 one:

```patch
-        /** @var Media $media */
-        $media = $this->mediaManager->getById($id, $locale);
+        try {
+            /** @var Media $media */
+            $media = $this->mediaManager->getById($id, $locale);
+        } catch (MediaNotFoundException $e) {
+            throw new NotFoundHttpException($e->getMessage());
+        }
```

Here: https://github.com/sulu/sulu/blob/8ed3425211e8b72b4ac7ecc5239842408927fc42/src/Sulu/Bundle/MediaBundle/Controller/MediaRedirectController.php#L51-L52

But this is surly not the right way to fix that, maybe the same thing in the manager?
</del>

Jupi007:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #6924
| License | MIT

#### What's in this PR?

Catch `MediaNotFoundException` in `MediaRedirectController`.

#### Why?

To avoid 500 error, and return the correct 404 http code.

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/Controller/MediaRedirectController.php
+++ b/src/Sulu/Bundle/MediaBundle/Controller/MediaRedirectController.php
@@ -53,7 +53,7 @@ class MediaRedirectController
             /** @var Media $media */
             $media = $this->mediaManager->getById($id, $locale);
         } catch (MediaNotFoundException $e) {
-            throw new NotFoundHttpException($e->getMessage());
+            throw new NotFoundHttpException($e->getMessage(), $e);
         }
 
         if (null === $format) {
--- a/src/Sulu/Bundle/MediaBundle/Controller/MediaRedirectController.php
+++ b/src/Sulu/Bundle/MediaBundle/Controller/MediaRedirectController.php
@@ -12,6 +12,7 @@
 namespace Sulu\Bundle\MediaBundle\Controller;
 
 use Sulu\Bundle\MediaBundle\Api\Media;
+use Sulu\Bundle\MediaBundle\Media\Exception\MediaNotFoundException;
 use Sulu\Bundle\MediaBundle\Media\Manager\MediaManagerInterface;
 use Sulu\Component\Rest\RequestParametersTrait;
 use Symfony\Component\HttpFoundation\RedirectResponse;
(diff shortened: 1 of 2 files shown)


alexander-schranz (approved):

src/Sulu/Bundle/MediaBundle/Controller/MediaRedirectController.php
```diff
@@ -48,8 +49,12 @@ public function redirectAction(Request $request, $id)
         $locale = $this->getRequestParameter($request, 'locale', true);
         $format = $this->getRequestParameter($request, 'format');
 
-        /** @var Media $media */
-        $media = $this->mediaManager->getById($id, $locale);
+        try {
+            /** @var Media $media */
+            $media = $this->mediaManager->getById($id, $locale);
+        } catch (MediaNotFoundException $e) {
+            throw new NotFoundHttpException($e->getMessage());
```
alexander-schranz: @Jupi007 the $previous exception should be given to the new exception.

```suggestion
            throw new NotFoundHttpException($e->getMessage(), $e);
```

Jupi007: Oh, indeed. Sorry.
Thanks for the fix.

PR #7027, opened 2023-03-02
Refactored usages of short namespace aliases of entities to use of FQCN

stollr:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no (but not really sure)
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

This is a refactoring only PR. All usages of short namespace aliases for doctrine entities are changed to FQCN.

#### Why?

This is required to allow doctrine/persistence 3.0, because it has removed support for those aliases (see [here](https://github.com/doctrine/persistence/blob/8bf8ab15960787f1a49d405f6eb8c787b4841119/UPGRADE.md)).

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -47,6 +47,45 @@ This update is also handled normally by the update build command automatically.
 As deprecated in Symfony 6.1 the `$defaultName` of Sulu Commands where replaced with the new
 `Symfony\Component\Console\Attribute\AsCommand` annotation.
 
+### ListBuilder Doctrine Changes
+
+Bundle aliases where deprecated by [`doctrine/persistence` 3.0](https://github.com/doctrine/persistence/blob/3.2.0/UPGRADE.md#bc-break-removed-support-for-short-namespace-aliases) the full FQCN need to be used when upgrading to `doctrine/persistence:^3.0`:
+
+```diff
+    <joins name="address">
+        <join>
+-            <entity-name>SuluContactBundle:AccountAddress</entity-name>
++            <entity-name>Sulu\Bundle\ContactBundle\Entity\AccountAddress</entity-name>
+            <field-name>%sulu.model.account.class%.accountAddresses</field-name>
+            <method>LEFT</method>
+-            <condition>SuluContactBundle:AccountAddress.main = TRUE</condition>
++            <condition>Sulu\Bundle\ContactBundle\Entity\AccountAddress.main = TRUE</condition>
+        </join>
+
+        <join>
+-            <entity-name>SuluContactBundle:Address</entity-name>
++            <entity-name>Sulu\Bundle\ContactBundle\Entity\Address</entity-name>
+-            <field-name>SuluContactBundle:AccountAddress.address</field-name>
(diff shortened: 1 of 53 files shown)


stollr:
I have targeted the 2.5 branch. Not sure if this is correct. If you like the target can be changed to 2.6

stollr:
Looks like the test environment is broken, because those `SQLSTATE[HY000] [1049] Unknown database 'sulu_test'` errors cannot be a result of my changes.

alexander-schranz:
@naitsirch thx for the pull request. It would be great to target here 2.6 branch can you rebase it to 2.6?

Curently the CI is failing because of 2 open external issues for `@dev`:

 - [x] https://github.com/symfony/symfony/issues/49387
 - [x] https://github.com/doctrine/orm/issues/10552

As this are currently issues for the `@dev` versions locally the tests should run. 

> SQLSTATE[HY000] [1049] Unknown database 'sulu_test'

Sounds more like a not available and running DB if you have run the tests locally?

The `PHP Lint` can be ignored there is already an open pull request fixing it on the lowest branch.

stollr:
@alexander-schranz I have rebased to 2.6 ;-)
On my local system `bin/runtests` doesn't report any errors.

alexander-schranz (approved):

wachterjohannes (approved):

alexander-schranz (changes requested):
What did come to my mind, that we need to check if still all works correctly when the Entities get extended: 

https://docs.sulu.io/en/2.5/cookbook/extend-entities.html

Because for then for doctrine point of view: `Sulu\Bundle\TagBundle\Entity\Tag` does not longer exist and `Sulu\Bundle\TagBundle\Entity\TagInterface` exist. So we maybe need to use the `Interface` in the cases where it is possible to extend the entity by a custom `App` class.

@stollr do you want to check that cases?

stollr:
In my project I am extending `Sulu\Bundle\ContactBundle\Entity\Account` with my custom `App\Entity\Account` like it is explained in the documentation and I am not facing any issues (I have checked out this branch).

The following screenshot shows it:

grafik

When I downgrade to the latest 2.5.x branch and use `SuluContactBundle:Account` the result is the same.

As far as I know shortcuts like `SuluContactBundle:Account` has always been normalized to the fqcn before doctrine used it. So I'd say this is safe.

Or do you have any special use case in mind that I should check?

alexander-schranz:
@stollr good would be if you overwrite once all available:

```yaml
<bundle>:
    objects:
        ...
```

configurations which exist https://docs.sulu.io/en/2.5/cookbook/extend-entities.html

And check if the list e.g. tags and categories in contacts and accounts, and other places where the `Bundle:Model` was replaced with FQCN still works.

Do that on a current skeleton and push that for testing into a seperate Repo, so we could check that out.

stollr:
I have created a test project (see https://github.com/stollr/sulu-test-pr-7027) where I have overridden all extendable entities.
As far as I see this patch works for all forms where categories and tags apply (account, contact and media).

alexander-schranz:
@stollr Thank you!

PR #7030, opened 2023-03-06
Migrate deprecations to symfony/deprecation-contracts

YetiCGN:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #6977
| Related issues/PRs | none
| License | MIT
| Documentation PR | not necessary

#### What's in this PR?

This PR adds a dependency to symfony/deprecation-contracts and utilizes the method `trigger_deprecations` from that package to trigger deprecations all throughout Sulu.

I have tried to guess the versions when a feature was deprecated as indicated by the timestamp of the commit when a version was not provided in the message.

#### Why?

Deprecations can be silenced or funneled at a central entry point in the `trigger_deprecations` function instead of the lower-level `trigger_error`. And to quote @alexander-schranz: "As this is the new way in Symfony we should switch to that one."

#### Example Usage

Transparent user experience, deprecation messages will change slightly in their wording, but not much else to do.

#### To Do

none

PR #7035, opened 2023-03-15
Migrate one remaining deprecation on 2.5 (#6977)

### LINKED ISSUE #6977: Migrate deprecations to symfony/deprecation-contracts
# Problem description

While Symfony now uses a central function `trigger_deprecation` from the package `symfony/deprecation-contracts` which can be overriden to do nothing in production, Sulu triggers `E_USER_DEPRECATED` itself in several places. This makes silencing deprecation INFO messages hard.

# Proposed solution

Require `symfony/deprecation-contracts` and refactor all occurences of `@\trigger_error(\sprintf('The "%s" method is deprecated on "%s" use "..." instead.', __METHOD__, __CLASS__), \E_USER_DEPRECATED);` to use `trigger_deprecation()`.

YetiCGN:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #6977
| Related issues/PRs | #7030
| License | MIT
| Documentation PR | not necessary

#### What's in this PR?

This PR migrates one deprecation warning on Sulu 2.5 to symfony/deprecation-contracts

#### Why?

Because [alex said so](https://github.com/sulu/sulu/pull/7030#issuecomment-1461762528) 😉 

#### Example Usage

Transparent user experience, deprecation messages will change slightly in their wording, but not much else to do.

#### To Do

none

### DIFF
--- a/src/Sulu/Bundle/SecurityBundle/Security/LogoutSuccessHandler.php
+++ b/src/Sulu/Bundle/SecurityBundle/Security/LogoutSuccessHandler.php
@@ -29,7 +29,7 @@ class LogoutSuccessHandler implements LogoutSuccessHandlerInterface
 
     public function __construct(RouterInterface $router)
     {
-        @\trigger_deprecation('sulu/sulu', '2.5', __CLASS__ . '() is deprecated and will be removed in 3.0. Use LogoutEventSubscriber instead.');
+        @trigger_deprecation('sulu/sulu', '2.5', __CLASS__ . '() is deprecated and will be removed in 3.0. Use LogoutEventSubscriber instead.');
 
         $this->router = $router;
     }
--- a/src/Sulu/Bundle/SecurityBundle/Security/LogoutSuccessHandler.php
+++ b/src/Sulu/Bundle/SecurityBundle/Security/LogoutSuccessHandler.php
@@ -29,7 +29,7 @@ class LogoutSuccessHandler implements LogoutSuccessHandlerInterface
 
     public function __construct(RouterInterface $router)
     {
-        @\trigger_error(__CLASS__ . '() is deprecated since version sulu/sulu 2.5 and will be removed in 3.0. Use LogoutEventSubscriber instead.', \E_USER_DEPRECATED);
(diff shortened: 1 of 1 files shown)


src/Sulu/Bundle/SecurityBundle/Security/LogoutSuccessHandler.php
```diff
@@ -29,7 +29,7 @@ class LogoutSuccessHandler implements LogoutSuccessHandlerInterface
 
     public function __construct(RouterInterface $router)
     {
-        @\trigger_error(__CLASS__ . '() is deprecated since version sulu/sulu 2.5 and will be removed in 3.0. Use LogoutEventSubscriber instead.', \E_USER_DEPRECATED);
+        @\trigger_deprecation('sulu/sulu', '2.5', __CLASS__ . '() is deprecated and will be removed in 3.0. Use LogoutEventSubscriber instead.');
```
alexander-schranz:
```suggestion
        @trigger_deprecation('sulu/sulu', '2.5', __CLASS__ . '() is deprecated and will be removed in 3.0. Use LogoutEventSubscriber instead.');
```

alexander-schranz:
> Because alex said so 😉

:laughing: Thank you!

src/Sulu/Bundle/SecurityBundle/Security/LogoutSuccessHandler.php
YetiCGN: Why not with the backslash? In #7030 I did every call with it and to my knowledge this makes it slightly faster because PHP checks the root namespace for the function immediately and not the local namespace first. But it works either way, so not a real problem. Just a bit of an inconsistency.

alexander-schranz: The current `php-cs-fixer` config does only prefix `@internal` php methods: https://github.com/sulu/sulu/blob/ebaab29cb35155e44005305f3973764562da6eac/.php-cs-fixer.dist.php#L27. Didn't want to change this.

PR #7048, opened 2023-03-30
Using typed properties

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | maybe
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?
Using typed properties for all classes (with private properties).

#### Why?

Php has added typed properties for some time now. Seeing that Sulu wants to become more type safe we should have this.

#### Example Usage
None

#### To Do


alexander-schranz:
I would see this kind of typing more as an automated task via `rector` / `php-cs-fixer`. We should check what is not working here in this pull request i the CI. But the fixes should be done on the PHPDocs in the lower 2.6 branch. As merging this currently would make to much merge conflicts for things we are currently implementing. But think something like we could automate here via task:

 - get current 3.0 branch
 - install rector
 - install php-cs-fixer
 - do type update
 - push it here
 - run tests against

And everytime we backmerge into 3.0 we could trigger this pull request todo things.

By fixing the PHPDocs we also improve the code quality of 2.6 branch which we have to maintain. And by automating the task we are sure nothing is forgotten. And we can do the merging of this branch the time where we know there will not come a lot of issues for us doing this change.

mamazu:
That makes a lot of sense to reduce the possible amount of conflicts between versions and I have been thinking on how to improve the type situation in 2.6 right now as well and came to the following conclusion:

There are two kinds of properties: read-write properties and what php 8.1 introduced as a keyword: readonly properties. The classic example of a readonly property would be one service having a reference to another like this:
```php
class OneService
{
    /** @var OtherService $otherService */
    private $otherService;
    
    public function __construct(OtherService $otherService) {
        $this->otherService = $otherService;
    }
}
```

Since there isn't any way to set the `$otherService` property to anything else since the property is private and the constuctor has a typehint, we can just promote the property here right now without any BC break.

```php
class OneService
{   
    public function __construct(private OtherService $otherService) {
    }
}
```

This is something that we can do in 2.6 as well since as I said the service is only written to the property in one place and this place (constructor) is already typesafe. I think this would also massively reduce the amount of changes in this PR which aims to bolt down all other types. What do you think?

PR #6845, opened 2022-10-08
Fix ObjectProphecy types to reduce phpstan baseline

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? | no 
| BC breaks? | no 
| Deprecations? | no  
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT
| Documentation PR | sulu/sulu-docs# 

#### What's in this PR?

Fix ObjectProphecy types in Sulu Component tests

#### Why?

Reduce the phpstan baseline errors.

#### Rector Config:

rector.php

```php
<?php

use Rector\Config\RectorConfig;
use Rector\PHPUnit\Rector\Property\ProphecyPHPDocRector;

return static function(RectorConfig $rectorConfig): void {
    $rectorConfig->paths([
        __DIR__ . '/src/Sulu/Bundle/ActivityBundle/Tests',
        __DIR__ . '/src/Sulu/Bundle/AdminBundle/Tests',
        __DIR__ . '/src/Sulu/Bundle/AudienceTargetingBundle/Tests',
        __DIR__ . '/src/Sulu/Bundle/CategoryBundle/Tests',
        __DIR__ . '/src/Sulu/Bundle/ContactBundle/Tests',
        __DIR__ . '/src/Sulu/Bundle/CoreBundle/Tests',
        __DIR__ . '/src/Sulu/Bundle/CustomUrlBundle/Tests',
        __DIR__ . '/src/Sulu/Bundle/DocumentManagerBundle/Tests',
        __DIR__ . '/src/Sulu/Bundle/HttpCacheBundle/Tests',
        __DIR__ . '/src/Sulu/Bundle/LocationBundle/Tests',
        __DIR__ . '/src/Sulu/Bundle/MarkupBundle/Tests',
        __DIR__ . '/src/Sulu/Bundle/MediaBundle/Tests',
        __DIR__ . '/src/Sulu/Bundle/PageBundle/Tests',
        __DIR__ . '/src/Sulu/Bundle/PersistenceBundle/Tests',
        __DIR__ . '/src/Sulu/Bundle/PreviewBundle/Tests',
        __DIR__ . '/src/Sulu/Bundle/RouteBundle/Tests',
        __DIR__ . '/src/Sulu/Bundle/SearchBundle/Tests',
        __DIR__ . '/src/Sulu/Bundle/SecurityBundle/Tests',
        __DIR__ . '/src/Sulu/Bundle/SnippetBundle/Tests',
[...]

PR #7049, opened 2023-03-31
Better doc typing

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | don't write untyped code :) 
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

* Adding some types in doc comments
* Improving types in some tests
* Adding explicit null return for non void functions (not a bc break)
* Updated some incorrect doctypes

#### Why?

As preparation for 3.0 we want the doc types to be as correct as possible for tools to rely on them when refactoring.

### DIFF
--- a/src/Sulu/Bundle/ContactBundle/Tests/Unit/CustomerIdConverterTest.php
+++ b/src/Sulu/Bundle/ContactBundle/Tests/Unit/CustomerIdConverterTest.php
@@ -32,7 +32,7 @@ class CustomerIdConverterTest extends TestCase
     /**
      * @dataProvider convertIdsToGroupedIdsProvider
      */
-    public function testConvertIdsToGroupedIds($ids, $default, $expected): void
+    public function testConvertIdsToGroupedIds(array $ids, array $default, $expected): void
     {
         $converter = new CustomerIdConverter();
         $result = $converter->convertIdsToGroupedIds($ids, $default);
@@ -56,7 +56,7 @@ class CustomerIdConverterTest extends TestCase
     /**
      * @dataProvider convertGroupedIdsToIdsProvider
      */
-    public function testConvertGroupedIdsToIds($groupedIds, $expected): void
+    public function testConvertGroupedIdsToIds(array $groupedIds, array $expected): void
     {
         $converter = new CustomerIdConverter();
         $result = $converter->convertGroupedIdsToIds($groupedIds);
--- a/src/Sulu/Bundle/ContactBundle/Tests/Unit/IndexComparatorTest.php
+++ b/src/Sulu/Bundle/ContactBundle/Tests/Unit/IndexComparatorTest.php
@@ -15,6 +15,9 @@ use PHPUnit\Framework\TestCase;
 
 class IndexComparatorTest extends TestCase
 {
+    /**
+     * @return array
+     */
     public function usortProvider()
     {
         return [
@@ -36,7 +39,7 @@ class IndexComparatorTest extends TestCase
     /**
      * @dataProvider usortProvider
      */
-    public function testUsortCallback($array, $ids, $startsWith, $contains): void
+    public function testUsortCallback(array $array, array $ids, $startsWith, $contains): void
     {
         $comparator = new IndexComparator();
         // the @ is necessary in case of a PHP bug https://bugs.php.net/bug.php?id=50688
--- a/src/Sulu/Component/Webspace/Tests/Unit/Manager/WebspaceManagerTest.php
+++ b/src/Sulu/Component/Webspace/Tests/Unit/Manager/WebspaceManagerTest.php
@@ -22,6 +22,7 @@ use Sulu\Component\Webspace\Loader\XmlFileLoader10;
 use Sulu\Component\Webspace\Loader\XmlFileLoader11;
 use Sulu\Component\Webspace\Manager\WebspaceManager;
 use Sulu\Component\Webspace\Portal;
+use Sulu\Component\Webspace\PortalInformation;
 use Sulu\Component\Webspace\Url\Replacer;
 use Sulu\Component\Webspace\Webspace;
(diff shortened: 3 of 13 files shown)


src/Sulu/Bundle/ContactBundle/Tests/Unit/CustomerIdConverterTest.php
```diff
@@ -32,7 +32,7 @@ public function convertIdsToGroupedIdsProvider()
     /**
      * @dataProvider convertIdsToGroupedIdsProvider
      */
-    public function testConvertIdsToGroupedIds($ids, $default, $expected): void
+    public function testConvertIdsToGroupedIds(array $ids, array $default, $expected): void
```
mamazu, line 35: This is not a bc break because the function where those variables are used already have a return type of array. So the only thing is we fail the code faster on broken types.

src/Sulu/Bundle/ContactBundle/Tests/Unit/CustomerIdConverterTest.php
```diff
@@ -56,7 +56,7 @@ public function convertGroupedIdsToIdsProvider()
     /**
      * @dataProvider convertGroupedIdsToIdsProvider
      */
-    public function testConvertGroupedIdsToIds($groupedIds, $expected): void
+    public function testConvertGroupedIdsToIds(array $groupedIds, array $expected): void
```
mamazu, line 59: This is not a bc break because the function where those variables are used already have a return type of array. So the only thing is we fail the code faster on broken types.

src/Sulu/Bundle/ContactBundle/Tests/Unit/IndexComparatorTest.php
```diff
@@ -36,7 +39,7 @@ public function usortProvider()
     /**
      * @dataProvider usortProvider
      */
-    public function testUsortCallback($array, $ids, $startsWith, $contains): void
+    public function testUsortCallback(array $array, array $ids, $startsWith, $contains): void
```
mamazu, line 42: This is not a bc break because the function where those variables are used already have a return type of array. So the only thing is we fail the code faster on broken types.

src/Sulu/Component/Webspace/Tests/Unit/Manager/WebspaceManagerTest.php
```diff
@@ -501,8 +508,10 @@ public function provideFindPortalInformationByUrl()
 
     /**
      * @dataProvider provideFindPortalInformationByUrl
+     *
+     * @param bool $shouldFind
      */
-    public function testFindPortalInformationByUrlWithInvalidSuffix($url, $shouldFind): void
+    public function testFindPortalInformationByUrlWithInvalidSuffix(string $url, $shouldFind): void
```
mamazu, line 514: This is not a bc break because the function where those variables are used already have a return type of array. So the only thing is we fail the code faster on broken types.

alexander-schranz:
PHP 8.2 is unrelated not sure yet whats wrong with gedmo `@dev` something or doctrine/orm `2.15.x`.

src/Sulu/Component/Webspace/Tests/Unit/PortalInformationTest.php
```diff
@@ -24,15 +28,33 @@ class PortalInformationTest extends TestCase
...
+    /**
+     * @var ObjectProphecy<Portal>
+     */
+    private ObjectProphecy $portal;
+
+    /**
+     * @var ObjectProphecy<Localization>
+     */
+    private ObjectProphecy $localization;
+
```
mamazu, line 45: These properties have been added, because they were implicit properties before which is deprecated in php.

src/Sulu/Component/Webspace/Tests/Unit/Manager/WebspaceManagerTest.php
alexander-schranz: yeah normally parameters types are bc breaks when class is not final but in tests we don't need to take care about bc and can set them.

alexander-schranz:
PHP 8.1 should not fail there seems to be something wrong.

src/Sulu/Component/Webspace/Tests/Unit/Manager/WebspaceManagerTest.php
mamazu: So in tests "BC breaking behavior" (like adding parameters and return types) is okay in test methods?

mamazu:
Fixed and asked a question on the last remaining thread.

src/Sulu/Component/Webspace/Tests/Unit/Manager/WebspaceManagerTest.php
alexander-schranz: @mamazu yes totally in tests we don't need to take care of the bc breaks, only the `Testing` namespace of the SuluTestBundle is the only one which part of the public namespace.

We already added and fixed some types in tests via rector here:

 - https://github.com/sulu/sulu/pull/6845
 - https://github.com/sulu/sulu/pull/6839
 
If we do that kind of massive manipulation I would prefer todo it on 2.4 branch or for PHP 8.0 changes on 2.5 avoids that we have problems in backmerge a bugfix with a test.

mamazu: Okay, I will keep that in mind for further type changes.

mamazu:
Then this MR is ready for review

alexander-schranz (approved):

alexander-schranz:
@mamazu Thank you! ~400 less lines in phpstan baseline 💪

PR #7050, opened 2023-03-31
Replace string class with Object::class

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | in the tests (but that's okay)
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?
```php
<?php

use Rector\Config\RectorConfig;
use Rector\Php55\Rector\String_\StringClassNameToClassConstantRector;
use Rector\Set\ValueObject\LevelSetList;

return static function(RectorConfig $rectorConfig): void {
    $rectorConfig->paths([
        __DIR__ . '/config',
        __DIR__ . '/public',
        __DIR__ . '/src',
        __DIR__ . '/tests',
    ]);

    // register a single rule
    $rectorConfig->rule(StringClassNameToClassConstantRector::class);
};
```

* Running the `StringClassNameToClassConstantRector`
* fixing some more type issues
* Fixing test assertions since `expectException` can only take one argument.

#### Why?

Class References like `Something::class` is better than having a string to the class because of two reasons:
* Supported by tools (eg. go do definition and usages)
* If Class gets renamed, the rename will also change the reference but not the string version

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -3781,11 +3781,6 @@ parameters:
 			path: src/Sulu/Bundle/AudienceTargetingBundle/Tests/Unit/Rule/ReferrerRuleTest.php
 
 		-
-			message: "#^Method PHPUnit\\\\Framework\\\\TestCase\\:\\:expectException\\(\\) invoked with 2 parameters, 1 required\\.$#"
-			count: 1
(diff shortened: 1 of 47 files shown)


alexander-schranz (approved):

alexander-schranz:
@mamazu thank you!

PR #7052, opened 2023-04-01
Adding more types

mamazu:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | yes
| Deprecations? | yes
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

* Improving typing
* Fixing a bug in the src/Sulu/Component/Rest/RequestParametersTrait.php

#### Why?

Improving the types to mirror the actual return and parameter types is always good.

#### To Do

### DIFF
--- a/src/Sulu/Bundle/RouteBundle/Controller/RouteController.php
+++ b/src/Sulu/Bundle/RouteBundle/Controller/RouteController.php
@@ -141,14 +141,13 @@ class RouteController extends AbstractRestController implements ClassResourceInt
     /**
      * Find routes with given parameters.
      *
-     * @param bool $history
      * @param string $entityClass
      * @param string $entityId
      * @param string $locale
      *
      * @return RouteInterface|RouteInterface[]
      */
-    private function findRoutes($history, $entityClass, $entityId, $locale)
+    private function findRoutes(bool $history, $entityClass, $entityId, $locale)
     {
         if ($history) {
             return $this->routeRepository->findHistoryByEntity($entityClass, $entityId, $locale);
--- a/src/Sulu/Bundle/SnippetBundle/Controller/SnippetController.php
+++ b/src/Sulu/Bundle/SnippetBundle/Controller/SnippetController.php
@@ -175,6 +175,7 @@ class SnippetController implements SecuredControllerInterface, ClassResourceInte
             $snippets = $this->snippetRepository->getSnippetsByUuids($ids, $locale, true);
             $total = \count($snippets);
         } else {
+            $locale = $locale ?? 'de_de';
             $snippets = $this->snippetRepository->getSnippets(
                 $locale,
                 $types ? $types[0] : null,
@@ -274,7 +275,7 @@ class SnippetController implements SecuredControllerInterface, ClassResourceInte
      */
     public function deleteAction(Request $request, $id)
     {
-        $locale = $this->getLocale($request);
+        $locale = $this->getLocale($request) ?? 'de_de';
         $webspaceKey = (string) $request->query->get('webspace');
 
         $references = $this->snippetRepository->getReferences($id);
@@ -398,7 +399,7 @@ class SnippetController implements SecuredControllerInterface, ClassResourceInte
     }
 
     /**
-     * @return string
+     * @return string|null
      */
     public function getLocale(Request $request)
     {
@@ -413,6 +414,9 @@ class SnippetController implements SecuredControllerInterface, ClassResourceInte
         return $request->query->get('language', null);
     }
 
+    /**
+     * @return string
+     */
     public function getSecurityContext()
     {
         return 'sulu.global.snippets';
--- a/src/Sulu/Bundle/SnippetBundle/Snippet/SnippetRepository.php
+++ b/src/Sulu/Bundle/SnippetBundle/Snippet/SnippetRepository.php
@@ -77,7 +77,7 @@ class SnippetRepository
      *
      * UUIDs which fail to resolve to a snippet will be ignored.
      *
-     * @param string $locale
+     * @param ?string $locale
      * @param bool $loadGhostContent
      *
      * @return array<SnippetDocument>
@@ -107,12 +107,12 @@ class SnippetRepository
      * If $type is given then only return the snippets of that type.
      *
      * @param string $locale
-     * @param string $type Optional snippet type
-     * @param int $offset Optional offset
-     * @param int $max Optional max
-     * @param string $search
-     * @param string $sortBy
-     * @param string $sortOrder
+     * @param ?string $type Optional snippet type
+     * @param ?int $offset Optional offset
+     * @param ?int $max Optional max
+     * @param ?string $search
+     * @param ?string $sortBy
+     * @param ?string $sortOrder
      *
      * @return SnippetDocument[]
      *
@@ -141,10 +141,10 @@ class SnippetRepository
      * If $type is given then only return the snippets of that type.
      *
      * @param string $locale
-     * @param string $type Optional snippet type
-     * @param string $search
-     * @param string $sortBy
-     * @param string $sortOrder
+     * @param ?string $type Optional snippet type
+     * @param ?string $search
+     * @param ?string $sortBy
+     * @param ?string $sortOrder
      *
      * @return int
      *
--- a/src/Sulu/Component/Content/Mapper/ContentMapper.php
+++ b/src/Sulu/Component/Content/Mapper/ContentMapper.php
@@ -618,6 +618,7 @@ class ContentMapper implements ContentMapperInterface
      * Return the content document (aka the home page).
      *
      * @param string $webspaceKey
+     * @param string|string $locale
      *
      * @return BasePageDocument|SnippetDocument
      */
@@ -679,6 +680,14 @@ class ContentMapper implements ContentMapperInterface
 
     /**
      * converts a query row to an array.
+     *
+     * @param string $locale
+     * @param string $webspaceKey
+     * @param array $fields
+     * @param bool $onlyPublished
+     * @param array|null $permission
+     *
+     * @return array|false
      */
     private function rowToArray(
         Row $row,
--- a/src/Sulu/Component/Rest/RequestParametersTrait.php
+++ b/src/Sulu/Component/Rest/RequestParametersTrait.php
@@ -44,12 +44,14 @@ trait RequestParametersTrait
     /**
      * returns request parameter as boolean 'true' => true , 'false' => false.
      *
+     * @template T of bool|null
+     *
      * @param Request $request
      * @param string $name
      * @param bool $force TRUE if value is mandatory
-     * @param bool $default value if parameter not exists
+     * @param T $default value if parameter not exists
      *
-     * @return bool
+     * @return bool|T
      *
      * @throws MissingParameterException parameter is mandatory but does not exists
      * @throws ParameterDataTypeException parameter hast the wrong data type
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -51,11 +51,6 @@ parameters:
 			path: src/Sulu/Bundle/ActivityBundle/UserInterface/Controller/ActivityController.php
 
 		-
-			message: "#^Method Sulu\\\\Bundle\\\\ActivityBundle\\\\UserInterface\\\\Controller\\\\ActivityController\\:\\:getBooleanRequestParameter\\(\\) should return bool but returns bool\\|null\\.$#"
-			count: 1
-			path: src/Sulu/Bundle/ActivityBundle/UserInterface/Controller/ActivityController.php
-
-		-
 			message: "#^Method Sulu\\\\Bundle\\\\ActivityBundle\\\\UserInterface\\\\Controller\\\\ActivityController\\:\\:getRequestParameter\\(\\) should return string but returns mixed\\.$#"
 			count: 1
 			path: src/Sulu/Bundle/ActivityBundle/UserInterface/Controller/ActivityController.php
@@ -3917,11 +3912,6 @@ parameters:
 
 		-
 			message: "#^Method Sulu\\\\Bundle\\\\CategoryBundle\\\\Controller\\\\CategoryController\\:\\:getAction\\(\\) has parameter \\$id with no type specified\\.$#"
-			count: 1
-			path: src/Sulu/Bundle/CategoryBundle/Controller/CategoryController.php
-
-		-
-			message: "#^Method Sulu\\\\Bundle\\\\CategoryBundle\\\\Controller\\\\CategoryController\\:\\:getBooleanRequestParameter\\(\\) should return bool but returns bool\\|null\\.$#"
 			count: 1
 			path: src/Sulu/Bundle/CategoryBundle/Controller/CategoryController.php
 
@@ -10941,11 +10931,6 @@ parameters:
 			path: src/Sulu/Bundle/CustomUrlBundle/Controller/CustomUrlController.php
 
 		-
-			message: "#^Method Sulu\\\\Bundle\\\\CustomUrlBundle\\\\Controller\\\\CustomUrlController\\:\\:getBooleanRequestParameter\\(\\) should return bool but returns bool\\|null\\.$#"
-			count: 1
-			path: src/Sulu/Bundle/CustomUrlBundle/Controller/CustomUrlController.php
-
-		-
 			message: "#^Method Sulu\\\\Bundle\\\\CustomUrlBundle\\\\Controller\\\\CustomUrlController\\:\\:getRequestParameter\\(\\) should return string but returns mixed\\.$#"
 			count: 1
 			path: src/Sulu/Bundle/CustomUrlBundle/Controller/CustomUrlController.php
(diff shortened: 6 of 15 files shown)


src/Sulu/Bundle/MediaBundle/Controller/CollectionController.php
```diff
@@ -156,8 +156,8 @@ public function getAction($id, Request $request)
         try {
             $locale = $this->getRequestParameter($request, 'locale', true);
             $depth = \intval($request->get('depth', 0));
-            $breadcrumb = $this->getBooleanRequestParameter($request, 'breadcrumb', false, false);
-            $children = $this->getBooleanRequestParameter($request, 'children', false, false);
+            $breadcrumb = $this->getBooleanRequestParameter($request, 'breadcrumb', false) ?? false;
```
alexander-schranz: this `??` ar enot required as false is the default value set and returned. phpdoc of that method can be improved so phpstan understands it.

src/Sulu/Bundle/MediaBundle/Controller/CollectionController.php
```diff
@@ -156,8 +156,8 @@ public function getAction($id, Request $request)
...
+            $children = $this->getBooleanRequestParameter($request, 'children', false) ?? false;
```
alexander-schranz: same as above

src/Sulu/Bundle/MediaBundle/Controller/CollectionController.php
```diff
@@ -393,7 +393,7 @@ protected function getData(Request $request)
     protected function saveEntity($id, Request $request)
     {
         $parent = $request->get('parent');
-        $breadcrumb = $this->getBooleanRequestParameter($request, 'breadcrumb', false, false);
+        $breadcrumb = $this->getBooleanRequestParameter($request, 'breadcrumb', false) ?? false;
```
alexander-schranz: same as above

src/Sulu/Bundle/MediaBundle/Entity/Collection.php
```diff
@@ -88,12 +88,12 @@ class Collection implements CollectionInterface, PermissionInheritanceInterface
     private $key;
 
     /**
-     * @var DoctrineCollection<int, CollectionMeta>
+     * @var DoctrineCollection<array-key, CollectionMeta>
```
alexander-schranz: this are always `int` as it is not indexed by something else in the mapping. in case all our collections should be not indexed by something else as that would be a performance issue.

src/Sulu/Bundle/MediaBundle/Entity/Collection.php
```diff
@@ -88,12 +88,12 @@ class Collection implements CollectionInterface, PermissionInheritanceInterface
...
      */
     private $meta;
 
     /**
-     * @var DoctrineCollection<int, MediaInterface>
+     * @var DoctrineCollection<array-key, MediaInterface>
```
alexander-schranz: same as above

src/Sulu/Bundle/RouteBundle/Controller/RouteController.php
```diff
@@ -108,7 +108,7 @@ public function cgetAction(Request $request)
         $id = $this->getRequestParameter($request, 'id', true);
 
         // optional parameter
-        $history = $this->getBooleanRequestParameter($request, 'history', false, false);
+        $history = $this->getBooleanRequestParameter($request, 'history', false) ?? false;
```
alexander-schranz: same this `??` is not required.

src/Sulu/Bundle/SnippetBundle/Controller/SnippetController.php
```diff
@@ -169,12 +168,14 @@ public function cgetAction(Request $request)
...
 
         $idsString = $request->get('ids');
+        $locale = $this->getLocale($request);
 
         if ($idsString) {
             $ids = \explode(',', $idsString);
             $snippets = $this->snippetRepository->getSnippetsByUuids($ids, $locale, true);
             $total = \count($snippets);
         } else {
+            $locale = $locale ?? 'de_de';
```
alexander-schranz, line 178: this should not be added here.

src/Sulu/Bundle/SnippetBundle/Controller/SnippetController.php
```diff
@@ -169,12 +168,14 @@ public function cgetAction(Request $request)
         }
 
         $idsString = $request->get('ids');
+        $locale = $this->getLocale($request);
```
alexander-schranz: would keep it above.

src/Sulu/Bundle/SnippetBundle/Controller/SnippetController.php
```diff
@@ -427,7 +431,7 @@ public function getSecurityContext()
      * @param string $webspace
      * @param NodeInterface[] $references
      * @param string $id
-     * @param string $locale
+     * @param ?string $locale
```
alexander-schranz: think the locale is required here when calling the contentmapper

src/Sulu/Bundle/SnippetBundle/Snippet/SnippetRepository.php
```diff
@@ -77,7 +77,7 @@ public function getReferences($uuid)
      *
      * UUIDs which fail to resolve to a snippet will be ignored.
      *
-     * @param string $locale
+     * @param ?string $locale
```
alexander-schranz, line 80: the locale is required here. so would not add any `null` here.

UPGRADE.md
```diff
@@ -13,7 +13,7 @@ Sulu now uses Webpack 5 to build the administration interface application. To en
 - `optimize-css-assets-webpack-plugin` was removed replaced by `css-minimizer-webpack-plugin`: `^4.2.2`
 - `clean-webpack-plugin` was removed and replaced by `clean: true` webpack output option
 - `webpack-clean-obsolete-chunks` was removed and replaced by `clean: true` webpack output option
-- `is-email` was removed and replaced by `sulu-admin-bundle/utils/Email/validateEmail` method 
+- `is-email` was removed and replaced by `sulu-admin-bundle/utils/Email/validateEmail` method
```
alexander-schranz: please revert the unnecessary changes to keep here the git histories correctly of this lines

src/Sulu/Component/Content/Mapper/ContentMapper.php
```diff
@@ -292,6 +292,11 @@ public function loadByParent(
         return $children;
     }
 
+    /**
+     * @param string $uuid Path or UUID
+     * @param string|null $locale Locale
```
alexander-schranz: locale should stay required here

src/Sulu/Component/Rest/RequestParametersTrait.php
```diff
@@ -25,14 +25,18 @@ trait RequestParametersTrait
      *
      * @param string $name
      * @param bool $force TRUE if value is mandatory
-     * @param mixed $default value if parameter not exists
+     * @param mixed $default value if parameter not exists, this will be ignored if $force is set to true
```
alexander-schranz: would currently keep it as it was, this would more make problems and could crash existing applications.

src/Sulu/Component/Rest/RequestParametersTrait.php
```diff
@@ -25,14 +25,18 @@ trait RequestParametersTrait
...
      *
      * @return string
      *
      * @throws MissingParameterException parameter is mandatory but does not exists
      */
     protected function getRequestParameter(Request $request, $name, $force = false, $default = null)
     {
+        if ($force) {
+            $default = null;
+        }
```
alexander-schranz: would not add this, current state I see it more that it could crash existing applications if we change this one.

src/Sulu/Component/Rest/RequestParametersTrait.php
```diff
@@ -49,7 +53,7 @@ protected function getRequestParameter(Request $request, $name, $force = false,
      * @param bool $force TRUE if value is mandatory
      * @param bool $default value if parameter not exists
      *
-     * @return bool
+     * @return bool|null
```
alexander-schranz: This could be improved to let phpstan if default is bool it always return a bool:

```
	/**
     * @template T of bool|null
     *
	 * @param T $default
     *
	 * @return bool|T
	 */
```

https://phpstan.org/r/01d59854-daa7-4cb7-b003-8fe8ac17101c

alexander-schranz (changes requested):

src/Sulu/Component/Rest/RequestParametersTrait.php
mamazu: Yeah that looks better.

src/Sulu/Component/Rest/RequestParametersTrait.php
mamazu: Removed it.

UPGRADE.md
mamazu: Removed the BC break so that changing the `UPGRADE.md` wasn't necessary anymore.

src/Sulu/Bundle/MediaBundle/Entity/Collection.php
mamazu: Yes, by default doctrine collections are indexed by ints but in general it could be anything (according to doctrine). Reverting it back to ints upsets the phpstan again, since it can be `array-key` which is just an alias to `string|int`.

alexander-schranz: yeah but would go with what it really is `int` here and not let phpstan decide what it isnt ;)

mamazu:
Should be fixed.

src/Sulu/Bundle/MediaBundle/Entity/Collection.php
alexander-schranz:
```suggestion
     * @var DoctrineCollection<int, CollectionMeta>
```

Still go with int here, phpstan can currently not detect it, but it is always an int.

src/Sulu/Bundle/MediaBundle/Entity/Collection.php
alexander-schranz:
```suggestion
     * @var DoctrineCollection<int, MediaInterface>
```

Still go with int here, phpstan can currently not detect it, but it is always an int.

src/Sulu/Bundle/RouteBundle/Controller/RouteController.php
alexander-schranz:
```suggestion
        $history = $this->getBooleanRequestParameter($request, 'history', false);
```

Should not be required.

src/Sulu/Bundle/SnippetBundle/Controller/SnippetController.php
alexander-schranz: still open

src/Sulu/Bundle/SnippetBundle/Controller/SnippetController.php
alexander-schranz: still open this should never be the case that we set here locale to hardcoded value. just remove this line at current state.

```suggestion
```

src/Sulu/Bundle/SnippetBundle/Snippet/SnippetRepository.php
alexander-schranz: still open, we should keep this also when currently the locale is not handled on all cases correctly:

```suggestion
     * @param string $locale
```

src/Sulu/Component/Content/Mapper/ContentMapper.php
alexander-schranz: would keep the locale here required and not open the interface also when not handle currently in all cases correctly:

```suggestion
     * @param string $locale
```

src/Sulu/Component/Content/Mapper/ContentMapper.php
```diff
@@ -292,6 +292,11 @@ public function loadByParent(
         return $children;
     }
 
+    /**
+     * @param string $uuid Path or UUID
```
alexander-schranz: while looking again on this. this phpdoc should be removed as the phpdoc is defined on the `ContentMapperInterface` and as said the locale should be kept here required. would not open the interface also when not handled currently correclty in all cases.

alexander-schranz (approved):

src/Sulu/Bundle/SnippetBundle/Controller/SnippetController.php
```diff
@@ -274,7 +273,7 @@ public function putAction(Request $request, $id)
      */
     public function deleteAction(Request $request, $id)
     {
-        $locale = $this->getLocale($request);
+        $locale = $this->getLocale($request) ?? 'de_de';
```
alexander-schranz, line 278: we should also not default here to anything hardcoded

src/Sulu/Component/Content/ContentTypeManager.php
```diff
@@ -60,6 +60,11 @@ public function get($alias)
         return $this->container->get($serviceId);
     }
 
+    /**
+     * @param string $alias
+     *
+     * @return bool
+     */
```
alexander-schranz: This is defined on the interface and so should not be required to be defined here: https://github.com/sulu/sulu/blob/08a40b7de9c1b3f9a6c42edc928ee8e4ff9ea97c/src/Sulu/Component/Content/ContentTypeManagerInterface.php#L33

src/Sulu/Bundle/SnippetBundle/Controller/SnippetController.php
mamazu: Well the getLocale function shouldn't return null or else it will cause other problemes. But for now we'll just need to expand the baseline for those errors. 'Cause we can't have function to return something nullable but the places where the value is used not accept null. Would that work for you?

src/Sulu/Component/Content/Mapper/ContentMapper.php
mamazu: Removed it.

src/Sulu/Component/Content/ContentTypeManager.php
mamazu: Removed it and fixed the parameter name to match the one documented in the interface.

src/Sulu/Bundle/SnippetBundle/Controller/SnippetController.php
alexander-schranz: Need to check the Permission Request Listener. If we could fallback to `$request->getLocale` in `$this->getLocale()` or if we add a `bool $force` to `getLocale()` method for the cases where it is required.

alexander-schranz: Okay checked it I would go with `getLocale(Request $request, bool $force = false)` and go with `@return ($force is true ? string : string|null)`

mamazu: I don't like the adding a `$force` parameter idea, since you basically completely changes the function.
 
What should happen when the locale is empty? Throw an exception?

alexander-schranz: > I don't like the adding a $force parameter idea, since you basically completely changes the function.

The `$force = false` is required as if not forcing the `getLocale` is required to return `null` for the SecurityListener. With the  `bool $force = false` we make sure the method behave the same as befor and only call with with `, true` where we 100% require a locale.

The exception should be this: https://github.com/sulu/sulu/blob/08a40b7de9c1b3f9a6c42edc928ee8e4ff9ea97c/src/Sulu/Component/Rest/RequestParametersTrait.php#L38

mamazu: Makes sense. How about we make a new function for that and deprecate the old one? So that we can have a clean cut between the old and the new behaviour.

alexander-schranz: I need to think about that. At current state I would concentrate in this PR on the PHPDoc and keep the other code as it is so we have code changes itself then in own pull request make searching for issues later easier.

src/Sulu/Bundle/AdminBundle/Admin/AdminPool.php
```diff
@@ -58,6 +58,11 @@ public function getSecurityContexts()
         return $contexts;
     }
 
+    /**
+     * Returns the combined security contexts with placeholders.
+     *
+     * @return array
```
alexander-schranz: Would like to move this into a seperate pull request where we maybe target the `getSecurityContextsWithPlaceholder()` and `getSecurityContexts` correctly on the `Admin` and `AdminPool` class.

The structure is nearly always the same for getSecurityContexts and so should be for getSecurityContextsWithPlaceholder

System (string) -> Section (string) -> Context (string) -> Permissions (string).

Which ends in:

```php
array<string, array<string, array<string, string[]>>>
```

mamazu: Wouldn't it maybe also make sense to refactor out an interface as the `Admin` and the `AdminPool` are using the same methods.

mamazu: I think we should split this merge request up.

mamazu: Has been split up in #7181 and will be removed from here.

src/Sulu/Bundle/AdminBundle/Admin/View/ListItemAction.php
```diff
@@ -23,7 +23,7 @@ class ListItemAction
     private $type;
 
     /**
-     * @var array
+     * @var array<string, string>
```
alexander-schranz: I'm not sure if this can be even mixed as the JS components decide the options and it could there be also nested things.

```suggestion
     * @var array<string, mixed>
```

src/Sulu/Bundle/AdminBundle/Admin/View/ListItemAction.php
```diff
@@ -35,11 +35,17 @@ public function __construct(string $type, array $options = [])
...
+    /**
+     * @return string
+     */
     public function getType()
     {
         return $this->type;
     }
 
+    /**
+     * @return array<string, string>
```
alexander-schranz:
```suggestion
     * @return array<string, mixed>
```

alexander-schranz (changes requested):
Need to check what we do with ther other comments but think here 2 types we need to change.

mamazu:
This merge request was split up into its parts. That way we can address the lowest version possible for all changes and makes this MR obsolete.

PR #7054, opened 2023-04-03
Using constructor property promotion for better types

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Using constructor property promotion for exceptions and fixing the type issues.

#### Why?

* Less Code duplication
    * Types only need to be defined once
    * Less code for assigning properties

**Isn't this a BC break?**

No, because the properties are not accessible outside or are private and their only write access is already type protected.

#### Example Usage

Before:
```php
/** @var string */
private $test;

/** @param string $test */
public function __construct($test) {
    $this->test = $test;
}
```

After:
```php
/** @param string $test */
public function __construct(private $test) {
}
```


src/Sulu/Component/Rest/Exception/ConstraintViolationException.php
```diff
@@ -18,36 +18,27 @@ class ConstraintViolationException extends RestException
...
-    {
-        $this->field = $field;
-        $this->entity = $entity;
-        $this->type = $type;
-
+    public function __construct(
+        protected $field,
+        protected $entity,
+        protected $type,
+        protected $code
```
alexander-schranz:
```suggestion
       $code
```

Code should just given to the parent

alexander-schranz (changes requested):

src/Sulu/Component/Rest/Exception/ConstraintViolationException.php
```diff
@@ -18,36 +18,27 @@ class ConstraintViolationException extends RestException
...
+    ) {
         parent::__construct(
-            \sprintf('%s constraint for field "%s" of entity "%s" violated', \ucfirst($type), $field, $entity),
-            $code
+            \sprintf('%s constraint for field "%s" of entity "%s" violated', \ucfirst($this->type), $this->field, $this->entity),
+            $this->code
```
alexander-schranz:
```suggestion
            $code
```

alexander-schranz (approved):

src/Sulu/Bundle/MarkupBundle/Tag/TagNotFoundException.php
```diff
@@ -16,33 +16,19 @@
...
-
-    /**
-     * @var string
-     */
-    private $type;
-
     /**
      * @param string $namespace
      * @param string $tagName
      * @param int $type
```
alexander-schranz: @mamazu not sure but I think this should be `string` as it comes from this interface: https://github.com/sulu/sulu/blob/08a40b7de9c1b3f9a6c42edc928ee8e4ff9ea97c/src/Sulu/Bundle/MarkupBundle/Tag/TagRegistryInterface.php#L23

alexander-schranz:
```suggestion
     * @param string $type
```

src/Sulu/Bundle/MarkupBundle/Tag/TagNotFoundException.php
```diff
@@ -58,7 +44,7 @@ public function getTagName()
     /**
      * Returns type of content.
      *
-     * @return string
+     * @return int
```
alexander-schranz:
```suggestion
     * @return string
```

mamazu:
Your changes all make sense. I'd say go ahead and merge it if there isn't anything else wrong.

mamazu:
@alexander-schranz Can you have a look at this again? I have updated the branch and now it only fails in the same places the 2.5 Branch also fails.

src/Sulu/Component/Content/Exception/UnexpectedPropertyType.php
```diff
@@ -16,35 +16,19 @@
...
+        parent::__construct(
+            \sprintf('Property "%s" is unexpected in content type "%s"', $property->getName(), \get_class($contentType))
+        );
     }
 
-    /**
-     * @return PropertyInterface
-     */
-    public function getProperty()
+    public function getProperty(): PropertyInterface
```
alexander-schranz: For BC need to stay as PHPDoc as this method and this class is not final

src/Sulu/Component/Content/Exception/UnexpectedPropertyType.php
```diff
@@ -16,35 +16,19 @@
...
     {
         return $this->property;
     }
 
-    /**
-     * @return ContentTypeInterface
-     */
-    public function getContentType()
+    public function getContentType(): ContentTypeInterface
```
alexander-schranz: For BC need to stay as PHPDoc as this method and this class is not final

alexander-schranz (changes requested):

mamazu:
Yes you are correct. I have removed them.

alexander-schranz (approved):

alexander-schranz:
@mamazu thank you!

PR #7055, opened 2023-04-06
Adding an `OnlyKeysVisitor` to reduce the config size when serializing

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no 
| License | MIT

#### What's in this PR?

Adding an option to only get a list of page types without having to also get the configuration of the page types themselves.

#### Why?

**Before:** In the current version the React application requests the entire meta data configuration for all pages just to get a list of possible keys. This causes very long load times for complex structures.

```text
Stats*
> Time: 13.394 ms (aka 13 s)
> Peak Memory: 1089.21 MiB (aka 1.06 GiB)
> Time spend serializing: 11,561.1 ms (aka 11 s)

> Size of the output: 3,250 KB (aka 3.5 MB)
```

**After:** Now we only load what we really need and don't load stuff that is going to be thrown away.

```text
Stats*
> Time: 765 ms
> Peak Memory: 216.28 MiB
> Time spend serializing: 3ms

> Size of the output: 1.65 KB
```

* Those values are the performance values in dev. In prod it might even get a better result.

It does not affect the performance of the cache since we don't change the loading but we change the data after loading that gets serialized.

#### Example Usage
Open `/admin/metadata/form/page?only_keys=true&webspace=<webspace>` and see that the response doesn't contain the whole configuration

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/OnlyKeysVisistor.php
+++ b/src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/OnlyKeysVisistor.php
@@ -1,0 +1,34 @@
+<?php
+
+/*
+ * This file is part of Sulu.
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\AdminBundle\Metadata\FormMetadata;
+
+use Sulu\Bundle\AdminBundle\Metadata\SchemaMetadata\SchemaMetadata;
+
+class OnlyKeysVisistor implements TypedFormMetadataVisitorInterface
+{
+    public function visitTypedFormMetadata(
+        TypedFormMetadata $formMetadata,
+        string $key,
+        string $locale,
+        array $metadataOptions = []
+    ): void {
+        $onlyKeys = $metadataOptions['onlyKeys'] ?? false;
+        if (!$onlyKeys) {
+            return;
+        }
+
+        foreach ($formMetadata->getForms() as $metaData) {
+            $metaData->setItems([]);
+            $metaData->setSchema(new SchemaMetadata([], [], []));
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/OnlyKeysVisistor.php
```diff
@@ -0,0 +1,34 @@
...
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\AdminBundle\Metadata\FormMetadata;
+
+use Sulu\Bundle\AdminBundle\Metadata\SchemaMetadata\SchemaMetadata;
+
+class OnlyKeysVisistor implements TypedFormMetadataVisitorInterface
+{
```
alexander-schranz, line 17: This feels a little bit strange for me doing this as a Visitor. But I'm not familiar enough with the metadata process here, but for me it more would feel like a new serialization group instead of a visitor. But I not even know how the seralization process works of the metadata.

/cc @wachterjohannes what do you think is best to achieve this kind of thing?

mamazu: I completely agree that doing it with a serialization group would probably be the best idea. The optimal solution would just to add another endpoint and only load the information needed. (Right now it still loads the entire structure but only outputs a subsection.)

However, I don't know how Sulu handles Serialization groups. That's why I chose this approach.

wachterjohannes: @alexander-schranz @mamazu i would also have to take a deeper look at that. But in the first look we use the "normal" serialization process (in our case JMS) to output the metadata (see https://github.com/sulu/sulu/blob/2.5/src/Sulu/Bundle/AdminBundle/Controller/AdminController.php#L335) so groups would be quite easy to add.

alexander-schranz: Yeah then I think the new query parameter (?keys-only=true) should add a specific serialization group (`admin_form_metadata_keys_only`) which gets the the `@Exclude` on the items so only the keys are returned.

mamazu:
Closed in favour of #7070

PR #7070, opened 2023-05-08
Add a seralization group to reduce the config when serializing

maikrosenthal:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Adding an option to only get a list of page types without having to also get the configuration of the page types themselves.

#### Why?

**Before:** In the current version the React application requests the entire meta data configuration for all pages just to get a list of possible keys. This causes very long load times for complex structures.

```text
Stats*
> Time: 13.394 ms (aka 13 s)
> Peak Memory: 1089.21 MiB (aka 1.06 GiB)
> Time spend serializing: 11,561.1 ms (aka 11 s)

> Size of the output: 3,250 KB (aka 3.5 MB)
```

**After:** Now we only load what we really need and don't load stuff that is going to be thrown away.

```text
Stats*
> Time: 765 ms
> Peak Memory: 216.28 MiB
> Time spend serializing: 3ms

> Size of the output: 1.65 KB
```

* Those values are the performance values in dev. In prod it might even get a better result.

It does not affect the performance of the cache since we don't change the loading but we change the data after loading that gets serialized.

#### Example Usage
Open `/admin/metadata/form/page?only_keys=true&webspace=<webspace>` and see that the response doesn't contain the whole configuration

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Controller/AdminController.php
+++ b/src/Sulu/Bundle/AdminBundle/Controller/AdminController.php
@@ -340,8 +340,15 @@ class AdminController
         $metadata = $this->metadataProviderRegistry->getMetadataProvider($type)
             ->getMetadata($key, $user->getLocale(), $metadataOptions);
 
+        $context = new Context();
+        $context->addGroup('Default');
+        if (filter_var($metadataOptions['onlyKeys'] ?? 'false', FILTER_VALIDATE_BOOLEAN) === true) {
+            $context->addGroup('admin_form_metadata_keys_only');
+        }
+
         $view = View::create($metadata);
         $view->setFormat('json');
+        $view->setContext($context);
 
         $response = $this->viewHandler->handle($view);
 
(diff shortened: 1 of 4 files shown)


mamazu:
@alexander-schranz Effects in the live system aren't as noticable but I still think this could be beneficial.

mamazu:
@maikrosenthal Please rebase this feature as the base has now been fixed.

src/Sulu/Bundle/AdminBundle/Controller/AdminController.php
```diff
@@ -340,8 +340,15 @@ public function metadataAction(string $type, string $key, Request $request): Res
         $metadata = $this->metadataProviderRegistry->getMetadataProvider($type)
             ->getMetadata($key, $user->getLocale(), $metadataOptions);
 
+        $context = new Context();
+        $context->addGroup('Default');
+        if (true === \filter_var($metadataOptions['onlyKeys'] ?? 'false', \FILTER_VALIDATE_BOOLEAN)) {
+            $context->addGroup('admin_form_metadata_keys_only');
```
alexander-schranz, line 346: A functional test would be good for this part of the AdminController. Else the code looks fine.

alexander-schranz:
@maikrosenthal Thank you!

PR #6998, opened 2023-02-10
Include categories and tags in the account/contact export and fix GroupBy ListBuilder problem

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no yes
| New feature? | no yes
| BC breaks? | no yes
| Deprecations? | no
| Fixed tickets | none
| Related issues/PRs | replaces https://github.com/sulu/sulu/pull/6976
| License | MIT
| Documentation PR | 

#### What's in this PR?

This PR introduces own queries for the grouping selects (Count, Group-Concat).

#### Why?

This will allow to use strict mode in SQL.

PR #7072, opened 2023-05-10
Fix group by statement for sorted queries

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | none
| Related issues/PRs | #6998 
| License | MIT
| Documentation PR | none

#### What's in this PR?

This PR fixes an issue with the query-builder goup-by queries in combination with sorting.

First the Group by has to be extended by the sortfields to be compatible with only_full_group_by and also the joins has to be added from the sort fields. else there an issue happens.

### DIFF
--- a/src/Sulu/Component/Rest/ListBuilder/Doctrine/DoctrineListBuilder.php
+++ b/src/Sulu/Component/Rest/ListBuilder/Doctrine/DoctrineListBuilder.php
@@ -462,9 +462,13 @@ class DoctrineListBuilder extends AbstractListBuilder
      */
     protected function assignGroupBy($queryBuilder)
     {
-        if (!empty($this->groupByFields)) {
-            foreach ($this->groupByFields as $fields) {
-                $queryBuilder->groupBy($fields->getSelect());
+        $groupByFields = array_merge($this->groupByFields, $this->sortFields);
+
+        if (!empty($groupByFields)) {
+            foreach ($groupByFields as $field) {
+                if ($field instanceof DoctrineFieldDescriptor) {
+                    $queryBuilder->addGroupBy($field->getSelect());
+                }
             }
         }
     }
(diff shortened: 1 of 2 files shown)


src/Sulu/Component/Rest/ListBuilder/Doctrine/DoctrineListBuilder.php
```diff
@@ -462,9 +462,11 @@ protected function hasSelectStatement(QueryBuilder $queryBuilder, $statement)
      */
     protected function assignGroupBy($queryBuilder)
     {
-        if (!empty($this->groupByFields)) {
-            foreach ($this->groupByFields as $fields) {
-                $queryBuilder->groupBy($fields->getSelect());
+        $groupByFields = array_merge($this->groupByFields, $this->sortFields);
```
alexander-schranz: At current state we have 3 queries:

1. id query
2. select query
3. group query

Think from performance point of view, I think the only one which requires to sort by `id` is the `id query`. And the others one don't require sorting and or filters as they just require `WHERE id IN (...)`. And the rest can be handled on the PHP side.

So the current problem I think we applying sort by queries where it currently is not necessary to add sort by.

wachterjohannes: @alexander-schranz i dont think that this is a good idea - would keep it like that because handling that in PHP is in my opinion not the best option.

alexander-schranz: Why its not a good idea? Currently doing it in 3 queries the sorting feels like unnecessary overhead 🤔.

wachterjohannes: because i am quite sure that we will introduce another issue with that 😂  and i am also not really sure if sorting in php is faster that doing it in the queries :)

What i have done is removing the sort by from the grouping query. It is really unnecessary because we getting the results index by the id so the sort itself is made for nothing

alexander-schranz:
@wachterjohannes Tests failing and we should add a test case for this adoption.

alexander-schranz:
ping @wachterjohannes :)

AdamReece-WebBox:
Thank you for this patch. This appears to resolve #7166 when running the MySQL 5.7.44 database engine.

AdamReece-WebBox (commented):
Resolves #7166 when run with MySQL 5.7.44.

alexander-schranz:
ping @wachterjohannes

PR #7077, opened 2023-05-23
Fix trash action failure with null security context

### LINKED ISSUE #6969: Restore trash item without security context fails
| Q | A
| --- | ---
| Sulu Version | 2.5.5
| PHP Version | 8.1.11

# Actual Behavior

Trying to restore a trash item without a security context always fails at security check.
I can provide a minimal reproduction repository is needed.

# Expected Behavior

The item should be allowed to be restored, even if the security context is null.

# Possible Solutions

In `TrashItemController.php`, the security check should be conditionally call, only if the item security context is not null:

https://github.com/sulu/sulu/blob/0c564eb9b8df8a163dc6a4381c710b003f80641a/src/Sulu/Bundle/TrashBundle/UserInterface/Controller/Admin/TrashItemController.php#L290-L298

```php
        if (null !== $trashItem->getResourceSecurityContext()) {
            $this->securityChecker->checkPermission(
                // ...
            );
        }
```

Jupi007:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #6969

#### What's in this PR?

Check for null trash item security context before perform a permissions check.

#### Why?

Currently, trashing or restore an item without a security context throw an exception.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -36131,11 +36131,6 @@ parameters:
 			path: src/Sulu/Bundle/TrashBundle/Tests/Functional/UserInterface/Controller/Admin/TrashItemControllerTest.php
 
 		-
-			message: "#^Unsafe call to private method Sulu\\\\Bundle\\\\TrashBundle\\\\Tests\\\\Functional\\\\UserInterface\\\\Controller\\\\Admin\\\\TrashItemControllerTest\\:\\:setUpUserRole\\(\\) through static\\:\\:\\.$#"
-			count: 2
-			path: src/Sulu/Bundle/TrashBundle/Tests/Functional/UserInterface/Controller/Admin/TrashItemControllerTest.php
-
-		-
 			message: "#^Method Sulu\\\\Bundle\\\\TrashBundle\\\\UserInterface\\\\Controller\\\\Admin\\\\TrashItemController\\:\\:getBooleanRequestParameter\\(\\) should return bool but returns bool\\|null\\.$#"
 			count: 1
 			path: src/Sulu/Bundle/TrashBundle/UserInterface/Controller/Admin/TrashItemController.php
--- a/src/Sulu/Bundle/TrashBundle/Tests/Functional/UserInterface/Controller/Admin/TrashItemControllerTest.php
+++ b/src/Sulu/Bundle/TrashBundle/Tests/Functional/UserInterface/Controller/Admin/TrashItemControllerTest.php
@@ -21,6 +21,7 @@ use Sulu\Bundle\SecurityBundle\Entity\UserRole;
 use Sulu\Bundle\TestBundle\Kernel\SuluKernelBrowser;
 use Sulu\Bundle\TestBundle\Testing\SuluTestCase;
 use Sulu\Bundle\TrashBundle\Domain\Repository\TrashItemRepositoryInterface;
+use Sulu\Bundle\TrashBundle\Infrastructure\Sulu\Admin\TrashAdmin;
 use Sulu\Bundle\TrashBundle\Tests\Functional\Traits\CreateTrashItemTrait;
 use Sulu\Component\Security\Authorization\AccessControl\SecuredEntityInterface;
 use Sulu\Component\Security\Authorization\PermissionTypes;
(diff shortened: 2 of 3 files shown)


alexander-schranz:
@Jupi007 can you add a test to the TrashItemControllerTest for both actions changes? That this bug not accidentally is reintroduced in the futre.

Jupi007:
@alexander-schranz I initially wanted to add one, but it looks like the current test suite is already use a `null` security context: 
https://github.com/sulu/sulu/blob/e16f526bafec3e0bd99904a38b027b7ac5ed920d/src/Sulu/Bundle/TrashBundle/Tests/Functional/Traits/CreateTrashItemTrait.php#L35
https://github.com/sulu/sulu/blob/e16f526bafec3e0bd99904a38b027b7ac5ed920d/src/Sulu/Bundle/TrashBundle/Tests/Functional/UserInterface/Controller/Admin/TrashItemControllerTest.php#L180

So I don't know what I should do to test this case.

What I don't understand is that the test didn't fail before, did it?
But I'm quite sure that an exception was thrown in my test project.

alexander-schranz:
What is the Stack Trace of the Exception inside your Test Project?

Jupi007:
I just retested my minimal repo, and it is simply denying access with an `AccessDeniedException`:

```
Symfony\Component\Security\Core\Exception\AccessDeniedException:
Permission "add" in localization "" for context "" and object with id "" and type "" not granted

  at vendor/sulu/sulu/src/Sulu/Component/Security/Authorization/AbstractSecurityChecker.php:39
  at Sulu\Component\Security\Authorization\AbstractSecurityChecker->checkPermission()
     (vendor/sulu/sulu/src/Sulu/Bundle/TrashBundle/UserInterface/Controller/Admin/TrashItemController.php:290)
  at Sulu\Bundle\TrashBundle\UserInterface\Controller\Admin\TrashItemController->restoreTrashItem()
     (vendor/sulu/sulu/src/Sulu/Bundle/TrashBundle/UserInterface/Controller/Admin/TrashItemController.php:272)
  at Sulu\Bundle\TrashBundle\UserInterface\Controller\Admin\TrashItemController->postTriggerAction()
     (vendor/symfony/http-kernel/HttpKernel.php:163)
  at Symfony\Component\HttpKernel\HttpKernel->handleRaw()
     (vendor/symfony/http-kernel/HttpKernel.php:74)
  at Symfony\Component\HttpKernel\HttpKernel->handle()
     (vendor/symfony/http-kernel/Kernel.php:184)
  at Symfony\Component\HttpKernel\Kernel->handle()
     (public/index.php:68)      
```

Just this isn't the case in the test suite, and I don't understand why.

alexander-schranz:
For the `test` user the SecurityChecker will not fail. But if we create another user and use that one it should also fail:

```php
        $otherUserClient = static::createClient(
            $options,
            \array_merge(
                [
                    'PHP_AUTH_USER' => 'otheruser',
                    'PHP_AUTH_PW' => 'otheruser',
                ],
                $server
            )
        );
```

This is because of: https://github.com/sulu/sulu/blob/e16f526bafec3e0bd99904a38b027b7ac5ed920d/src/Sulu/Bundle/TestBundle/Testing/TestVoter.php#LL64C34-L64C52

Jupi007:
Oh, I didn't know this fact, thanks.

What would you prefer? Another test case with a different user or reuse the current ones? I planned to do the first one but I prefer to ask before.

alexander-schranz:
Just what is easier, if easy possible without lot of refactoring in the same test class new methods, else an own test case like `TrashItemControllerWithPermissionCheckTest` is also fine.

Jupi007:
After having a hard time with the execution of the test suite due to the database, I'm currently looking at how to test this.

Jupi007:
@alexander-schranz I tried to do something with as little modification as possible.
The two new test pass with the fix of this PR, but fail without.

alexander-schranz (approved):

alexander-schranz (approved):

alexander-schranz:
@Jupi007 Thank you!

Jupi007:
You're welcome.
Thanks for the linting fix!

PR #7078, opened 2023-05-23
CannotEvaluateTokenException exception with array

### LINKED ISSUE #7013: CannotEvaluateTokenException throw exception with array $entity
| Q | A
| --- | ---
| Sulu Version | 2.4 and 2.5
| PHP Version | Unrelated
| DB Version | Unrelated
| Browser Version | Unrelated

# Actual Behavior

The `Sulu\Bundle\RouteBundle\Generator\SymfonyExpressionTokenProvider::provide()` method can handle a non-object `$entity` parameter, and it is correctly managed in the method:

https://github.com/sulu/sulu/blob/0d92f0a3e837809333077fcedc2bd43f2752d7e2/src/Sulu/Bundle/RouteBundle/Generator/SymfonyExpressionTokenProvider.php#L59-L63

If something went wrong during the expression language evaluation, a `CannotEvaluateTokenException` is throw, but its constructor can only handle object `$entity`, cause to `\get_class($entity)`:

https://github.com/sulu/sulu/blob/0d92f0a3e837809333077fcedc2bd43f2752d7e2/src/Sulu/Bundle/RouteBundle/Generator/CannotEvaluateTokenException.php#L34-L38

With php versions from 8.0, this throw a `TypeError`.

# Expected Behavior

The `CannotEvaluateTokenException` should be instantiated without problem.

# Steps to Reproduce

By define a sulu `route_schema` with a custom entity, and by trying to use an inexistent property.

# Possible Solutions

Check if `$entity` is an object before call `get_class()`, and call `gettype()` if false:

```diff
    parent::__construct(
-       \sprintf('Cannot evaluate token "%s" for entity with type "%s"', $token, \get_class($entity)),
+       \sprintf('Cannot evaluate token "%s" for entity with type "%s"', $token, \is_object($entity) ? \get_class($entity) : \gettype($entity)),
        0,
        $previous
    );
```

Jupi007:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #7013

#### What's in this PR?

Handle `array` typed `$entity` of `CannotEvaluateTokenException`.

### DIFF
--- a/src/Sulu/Bundle/RouteBundle/Generator/CannotEvaluateTokenException.php
+++ b/src/Sulu/Bundle/RouteBundle/Generator/CannotEvaluateTokenException.php
@@ -35,7 +35,7 @@ class CannotEvaluateTokenException extends \Exception
             \sprintf(
                 'Cannot evaluate token "%s" for entity with type "%s"',
                 $token,
-                \is_object($entity) ? \get_class($entity) : \gettype($entity),
+                \is_object($entity) ? \get_class($entity) : \gettype($entity)
             ),
             0,
             $previous
--- a/src/Sulu/Bundle/RouteBundle/Generator/CannotEvaluateTokenException.php
+++ b/src/Sulu/Bundle/RouteBundle/Generator/CannotEvaluateTokenException.php
@@ -32,7 +32,11 @@ class CannotEvaluateTokenException extends \Exception
     public function __construct($token, $entity, \Exception $previous)
     {
         parent::__construct(
-            \sprintf('Cannot evaluate token "%s" for entity with type "%s"', $token, \get_class($entity)),
+            \sprintf(
+                'Cannot evaluate token "%s" for entity with type "%s"',
+                $token,
+                \is_object($entity) ? \get_class($entity) : \gettype($entity),
+            ),
             0,
             $previous
         );


src/Sulu/Bundle/RouteBundle/Generator/CannotEvaluateTokenException.php
```diff
@@ -32,7 +32,11 @@ class CannotEvaluateTokenException extends \Exception
     public function __construct($token, $entity, \Exception $previous)
     {
         parent::__construct(
-            \sprintf('Cannot evaluate token "%s" for entity with type "%s"', $token, \get_class($entity)),
+            \sprintf(
+                'Cannot evaluate token "%s" for entity with type "%s"',
+                $token,
+                \is_object($entity) ? \get_class($entity) : \gettype($entity),
```
alexander-schranz:
```suggestion
                \is_object($entity) ? \get_class($entity) : \gettype($entity)
```

Jupi007: <del>Hmm, why is it required to remove the comma?
In this situation, my formatter always adds one.</del>

I looked at the Sulu code, and it looks like you don't use a comma at that point.
I'm so used to adding one that I didn't notice.

Jupi007:
Oh, I targeted again `2.5`...
It is annoying that, even if my branch is based on `2.4`, github automatically targets another one.

alexander-schranz:
It always targets the default branch on github. But no problem. Thank you for the fix!

PR #7089, opened 2023-06-05
Fix Symfony deprecation for not declaring getSupportedTypes on FlattenExceptionNormalizer

stollr:
| Q | A
| --- | ---
| Bug fix? | deprecation fix for Symfony 7 support
| New feature? | no 
| BC breaks? | no 
| Deprecations? | no 
| Fixed tickets | 
| Related issues/PRs | 
| License | MIT

#### What's in this PR?

I have added a `getSupportedTypes` method to the `FlattenExceptionNormalizer`.

#### Why?

Classes that implement the `Symfony\Component\Serializer\NormalizerInterface` will require the `getSupportedTypes` method for Symfony >= 7.0.

Currently Symfony's phpunit-bridge reports a deprecation message.

***Update***:
In the Symfony docs you'll find the details about that method (see [here](https://symfony.com/doc/current/serializer/custom_normalizer.html#improving-performance-of-normalizers-denormalizers)).

### DIFF
--- a/src/Sulu/Component/Rest/FlattenExceptionNormalizer.php
+++ b/src/Sulu/Component/Rest/FlattenExceptionNormalizer.php
@@ -42,6 +42,9 @@ class FlattenExceptionNormalizer implements ContextAwareNormalizerInterface
         $this->translator = $translator;
     }
 
+    /**
+     * @return array<class-string, bool>  
+     */
     public function getSupportedTypes(?string $format): array
     {
         return [
--- a/src/Sulu/Component/Rest/FlattenExceptionNormalizer.php
+++ b/src/Sulu/Component/Rest/FlattenExceptionNormalizer.php
@@ -43,7 +43,7 @@ class FlattenExceptionNormalizer implements ContextAwareNormalizerInterface
     }
 
     /**
-     * @return array<class-string, bool>  
+     * @return array<class-string, bool>
      */
     public function getSupportedTypes(?string $format): array
     {
--- a/src/Sulu/Component/Rest/FlattenExceptionNormalizer.php
+++ b/src/Sulu/Component/Rest/FlattenExceptionNormalizer.php
@@ -42,6 +42,13 @@ class FlattenExceptionNormalizer implements ContextAwareNormalizerInterface
         $this->translator = $translator;
     }
 
+    public function getSupportedTypes(?string $format): array
+    {
+        return [
+            FlattenException::class => false,
+        ];
+    }
+
     /**
      * @return array<int|string, mixed>
      */


src/Sulu/Component/Rest/FlattenExceptionNormalizer.php
```diff
@@ -42,6 +42,13 @@ public function __construct(
         $this->translator = $translator;
     }
 
+    public function getSupportedTypes(?string $format): array
+    {
+        return [
+            FlattenException::class => false,
```
alexander-schranz, line 51: Thx for the pull request. Can you link the documentation in the Pull Request description. I'm not sure what the boolean flag does here.

stollr: I have added a link. In short the flag will be used as an indicator for the cachability of the result of the supports* method. It will replace the `CacheableSupportsMethodInterface::hasCacheableSupportsMethod`  in Symfony 7.

stollr: The reason why I used `false` for the `FlattenException` is that the support depends on the value of `$context['messenger_serialization']` ;-)

src/Sulu/Component/Rest/FlattenExceptionNormalizer.php
```diff
@@ -42,6 +42,13 @@ public function __construct(
...
+        ];
```
alexander-schranz, line 52: I see but as this is a decorated service I think we should in this case just call:

```suggestion
        return $this->decoratedNormalizer->getSupportedTypes($format);
```

alexander-schranz (changes requested):

src/Sulu/Component/Rest/FlattenExceptionNormalizer.php
stollr: I think the result would currently be the same, but the `supportsNormalization` method does not call the decorated service, so in my opinion it shouldn't do that in the `getSupportedTypes` method either.

getSupportedTypes specifies the cacheability of the result of its own supportsNormalization.

If one day `FOS\RestBundle\Serializer\Normalizer\FlattenExceptionNormalizer::getSupportedTypes` will return `true`, Sulus' `getSupportedTypes` still should return `false`, because the result of its own `supportsNormalization` still must not be cached.

I hope you get my point :-)

alexander-schranz: but its not a `normalizer` by its own its a decorated service and should behave like the decorate normalizer and just extend it with the data we require.

stollr: Edit:
Okay, I got it now ;-) but why does `supportsNormalization` not call the decorated service?

alexander-schranz: I remember now because the older supported symfony version didn't have that parent method. So the PR is fine this way maybe in future we can update it to use `parent::` for both methods.

src/Sulu/Component/Rest/FlattenExceptionNormalizer.php
```diff
@@ -42,6 +42,13 @@ public function __construct(
         $this->translator = $translator;
     }
 
+    public function getSupportedTypes(?string $format): array
```
alexander-schranz, line 48:
```suggestion
    /**
     * @return array<class-string, bool>  
     */
    public function getSupportedTypes(?string $format): array
```

src/Sulu/Component/Rest/FlattenExceptionNormalizer.php
```diff
@@ -42,6 +42,16 @@ public function __construct(
         $this->translator = $translator;
     }
 
+    /**
+     * @return array<class-string, bool>
```
alexander-schranz:
```suggestion
     * @return array<class-string, bool>
```

alexander-schranz (approved):

alexander-schranz:
@stollr Thank you!

stollr:
You're welcome

PR #7091, opened 2023-06-07
Fixed locale for smart content shadow page

martinlagler:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Fix for location smart content shadow pages.

### DIFF
--- a/src/Sulu/Component/SmartContent/ContentType.php
+++ b/src/Sulu/Component/SmartContent/ContentType.php
@@ -279,7 +279,7 @@ class ContentType extends ComplexContentType implements ContentTypeExportInterfa
             $filters['limitResult'] : null;
         $options = [
             'webspaceKey' => $this->requestAnalyzer->getWebspace()->getKey(),
-            'locale' => $property->getStructure()->getLanguageCode(),
+            'locale' => $property->getStructure()->getDocument()->getOriginalLocale() ?? $property->getStructure()->getLanguageCode(),
         ];
 
         if (isset($params['max_per_page']) && $configuration->hasPagination()) {


src/Sulu/Component/SmartContent/ContentType.php
```diff
@@ -279,7 +279,7 @@ public function getContentData(PropertyInterface $property)
             $filters['limitResult'] : null;
         $options = [
             'webspaceKey' => $this->requestAnalyzer->getWebspace()->getKey(),
-            'locale' => $property->getStructure()->getLanguageCode(),
+            'locale' => $property->getStructure()->getDocument()->getOriginalLocale() ?? $property->getStructure()->getLanguageCode(),
```
alexander-schranz, line 282: @sulu/core-team we should discuss this behaviour. I think currently `Smart Content` and a lot of `Selections` (Pages, Snippets, ...). are targeting the false local and eventually load content which is not available in the language I'm currently loading. 

In @martinlagler specific case they are on a page `zh` which load `en` via shadow feature `en` = `$property->getStructure()->getLanguageCode()` but when clicked on an `Article` a `404` is shown as the rendered url `/zh/some-english-url` the article itself does not exist. Only article which exist in `/zh` should be rendered and not `en` articles be loaded here or in other selection.

First look this is issue with:

 - Smart Content (all)
 - Page Selection


The handling inside snippets I'm not sure, looks also not like expected:

https://github.com/sulu/sulu/blob/5006f0bccfaf4a4f41ba3275e687263b4636e58b/src/Sulu/Bundle/SnippetBundle/Content/SingleSnippetSelection.php#L93-L96

But the problem is snippets itself do not have a shadow feature and so it looks like here first en is loaded with fallback to zh 🙈

alexander-schranz: @martinlagler managed that this solution currently run in the project where it was used into other errors where first a deeper dive into this topic is required why there 500 errors did appear then.

alexander-schranz: There was also an inconsistent between pages `getLanguageCode` vs article `getLanguageCode` which could end into this kind of error.

This was fixed by @wachterjohannes in: https://github.com/sulu/SuluArticleBundle/pull/644

So maybe this PR get now obsolete.

wachterjohannes: @alexander-schranz yes that could be possible - if it only happens for articles! if not we have to takkle that

alexander-schranz: @Prokyonn you did have a look as I remember at the shadow implementation. Can you confirm that this issue was already fixed? On language `zh` which was a shadow to `en` all `en` smart items where returned, instead of the `zh` available items.

See comment at: https://github.com/sulu/sulu/pull/7091#discussion_r1221792048

alexander-schranz:
I'm closing this as I think @Prokyonn did fix the issue which appears on @martinlagler project.

PR #6938, opened 2022-12-14
Add reference bundle

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no yes
| Issue | #6507
| Releated PRs | #7098
| License | MIT
| Documentation PR | sulu/sulu-docs# 

#### What's in this PR?

Adds the basic setup for the reference-bundle

#### To Do

Open Points:

 - [x] Fix Item Actions (Deep Link) for `table_tree` adapter
 - [x] Fix Publish should write into `website` context always
 - [x] Add block index to referenceProperty
 - [x] Test imageMap blocks
 - [x] Add imageMap reference unit test
 - [x] Add snippet references
     - [x] Admin Insights Tab (Activity & References)
     - [x] SingleSnippetSelection
     - [x] SnippetSelection
 - [x] Add and Adjust Tests
     - [x] ReferenceController
        - [x] New Root Response
        - [x] New Child Response
     - [x] ReferenceRepository new filters no tested
        - [x] New Filters not tested
        - [x] findFlatBy not tested
        - [x] count not tested

---------

 - [ ] Create a documentation PR
 - [ ] Add breaking changes to UPGRADE.md

PR #7098, opened 2023-06-20
Add DetailViewItemAction

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | yes 
| Fixed tickets | -
| Related issues/PRs | implemented for #6938
| License | MIT

#### What's in this PR?

Add ResourceDetailViewItemAction.

#### Why?

The ResourceDetailViewItemAction allow to deeplink from a list with different recources in it to a detail view.

#### Usage

```php
           ->addItemActions([
                new ListItemAction(
                    'detail_link',
                    [
                        'resource_key_property' => 'referenceResourceKey',
                        'resource_id_property' => 'referenceResourceId',
                        'resource_view_attributes_property' => 'referenceViewAttributes',
                    ],
                ),
```


alexander-schranz:
@Prokyonn keep an focus on the newly added methods and refactoring not 100% sure about the naming and the added new services and service methods.

src/Sulu/Bundle/AdminBundle/Resources/js/services/Router/registries/resourceViewRegistry.js
```diff
@@ -0,0 +1,46 @@
...
+        this.clear();
+    }
+
+    clear() {
+        this.resourceViews = {};
+    }
+
+    addResourceViews(resourceViews: ResourceViewsMap) {
+        Object.keys(resourceViews).forEach((resourceKey) => {
+            if (this.resourceViews[resourceKey]) {
```
Prokyonn, line 17: When we already have a `has` method, I would use it here

src/Sulu/Bundle/AdminBundle/Resources/js/services/Router/registries/resourceViewRegistry.js
```diff
@@ -0,0 +1,46 @@
...
+        return this.resourceViews[resourceKey].views[view] !== undefined;
+    }
+
+    get(view: string, resourceKey: string): string {
+        if (typeof this.resourceViews[resourceKey] !== 'object') {
+            throw new Error('The resource "' + resourceKey + '" was not found.');
+        }
+
+        if (typeof this.resourceViews[resourceKey].views[view] !== 'string') {
+            throw new Error('The resource view "' + view + '" for resource "' + resourceKey + '" was not found.');
```
Prokyonn, line 39: Do we really have to distinguish in the error between `'The resource "' + resourceKey + '" was not found.` and `'The resource view "' + view + '" for resource "' + resourceKey + '" was not found.'` ?

I think the second one should be sufficient and then we could here also use the `has` method

src/Sulu/Bundle/AdminBundle/Resources/js/views/List/itemActions/ResourceDetailLinkItemAction.js
```diff
@@ -0,0 +1,69 @@
+// @flow
+import AbstractListItemAction from './AbstractListItemAction';
+
+export default class ResourceDetailLinkItemAction extends AbstractListItemAction {
```
Prokyonn: Maybe just name it `DetailLinkItemAction` ? Would be a bit shorter 🤔

Prokyonn:
> @Prokyonn keep an focus on the newly added methods and refactoring not 100% sure about the naming and the added new services and service methods.

I think the naming makes sense, with `view` in it, it should be quite clear what routes are registered here

src/Sulu/Bundle/AdminBundle/Resources/js/services/Router/registries/resourceViewRegistry.js
alexander-schranz: The `has` is specific for the `view` of a resource key not if general the resourceKey exist.

src/Sulu/Bundle/AdminBundle/Resources/js/services/Router/registries/resourceViewRegistry.js
alexander-schranz: Think if the error happens it is good to know if the `resourceKey` was not provided or if the `view` is just missing, think for better DX it make sense to have different specific error messages here.

src/Sulu/Bundle/AdminBundle/Resources/js/views/List/itemActions/ResourceDetailLinkItemAction.js
alexander-schranz: Okay then we name it also `detail_link` in the PHP Code?

Prokyonn: yes I think it is a bit shorter and cleaner, and we don't lose important information

Prokyonn (approved):

PR #7104, opened 2023-06-25
Remove sulu-link tags completely if they are invalid

### LINKED ISSUE #6956: Add possibility to remove a sulu-link if not exist
# Problem description

If you have a `sulu-link` and the target page does not exist the whole tag is removed but the text stay as it is. This is expected behaviour when the `sulu-link` is used inside a text editor so nothing is removed from the text.

But when the sulu-link is used via the `link` content type it would be nice if there is a attribute to tell if the target link does not exist to remove also the text.

# Proposed solution

Something like:

```xml
<sulu-link remove-if-not-exist="true" ... >
```

The service handling this is the [HtmlMarkupParser](https://github.com/sulu/sulu/blob/2.5/src/Sulu/Bundle/MarkupBundle/Markup/HtmlMarkupParser.php) and its child services. Which would needed to add the new support for this attribute.

mamazu:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #6956  
| Related issues/PRs | #6956 
| License | MIT
| Documentation PR | sulu/sulu-docs#773

#### What's in this PR?
When specifying the 'remove-if-not-exist' property on a link to be true, then the entire tag (including it's contents will be removed)

#### Why?
To prevent links that are invalid and text that (when it's not a link) looks weird. Like if you have a link list and one of the links is broken, it would be better to remove the text as well.

#### Example Usage
```html
<sulu-link href="123-123-123" title="Test-Title" provider="article" remove-if-not-exists="true">Test-Content</sulu-link>
```

#### To Do

### DIFF
--- a/src/Sulu/Bundle/MarkupBundle/Markup/LinkTag.php
+++ b/src/Sulu/Bundle/MarkupBundle/Markup/LinkTag.php
@@ -101,15 +101,20 @@ class LinkTag implements TagInterface
                 $title = $this->getContent($attributes);
                 $attributes['href'] = null;
             } else {
-                // only render text instead of anchor to prevent dead links on website
-                $result[$tag] = $this->getContent($attributes);
+                // Completely remove the tag if this attribute is set
+                if ($attributes['remove-if-not-exists'] ?? false) {
+                    $result[$tag] = '';
+                } else {
+                    // only render text instead of anchor to prevent dead links on website
+                    $result[$tag] = $this->getContent($attributes);
+                }
 
                 continue;
             }
 
             $htmlAttributes = \array_map(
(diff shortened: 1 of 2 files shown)


alexander-schranz (approved):

alexander-schranz:
Looks good from my side. /cc @sulu/core-team do you agree with this new added feature and the name `remove-if-not-exists`?

wachterjohannes:
@mamazu @alexander-schranz this makes totally sense

wachterjohannes (approved):

alexander-schranz:
@mamazu Thank you!

PR #7118, opened 2023-07-09
Add AsSuluContext attribute for defining sulu context

eKsiK:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Add ability to define sulu context for service by using PHP Attribute.

#### Why?

Since symfony 5.4, it is possible to register attributes for the autoconfigure service. In default configuration for sulu project, `src` directory is marked with `autoconfigure` flag for DI. Actually it is possible by adding `AutoconfigureTag` for class.

```php
#[AutoconfigureTag('sulu.context', ['context' => 'admin'])]
class EventController extends AbstractRestController
```
This change will simplify the process and improves DX.

#### Example Usage

For class which should only available in sulu admin context 

```php

#[AsSuluContext('admin')]
class EventController extends AbstractRestController

```

For class which should only available in sulu website context 

```php

#[AsSuluContext('website')]
class EventController extends WebsiteController

```

#### To Do

- [ ] Consider add validation for provided argument (only `admin` or `website` value)
- [ ] Consider add more attributes for sulu specific elements (e.g. `sulu.content.type`)

### DIFF
--- a/src/Sulu/Bundle/CoreBundle/Attribute/AsSuluContext.php
+++ b/src/Sulu/Bundle/CoreBundle/Attribute/AsSuluContext.php
@@ -1,0 +1,20 @@
+<?php
+
+/*
+ * This file is part of Sulu.
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\CoreBundle\Attribute;
+
+#[\Attribute(\Attribute::TARGET_CLASS)]
+class AsSuluContext
+{
+    public function __construct(public string $context)
+    {
+    }
+}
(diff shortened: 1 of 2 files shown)


TheCadien:
Cool! Talked to someone the other day about how it would be super cool to offer more configurations in Sulu, also via PHP annotations, in order to align more with Symfony. 
Like that way !

alexander-schranz:
Why its a great idea. I would think that we should not add attributes for `SuluContext` the Sulu context get removed as soon as we not require MultiKernel Setup, that is currently the case because of PHPCR, which target is remove for 3.0. Adding a new feature on something which will get removed I think is not a good idea.

It is recommended that a project only make usage of `sulu.context` when they access services which maybe exist only in one context in all other cases you should not care about the `sulu.context` and your services should build that way that they don't require this.

The typcial rest controllers are not longer requires to be in the admin context as the admin bundle services are since Sulu 2.3.0 available in all contexts, so that tag is not required longer for that type of services.

PR #7121, opened 2023-07-18
Use structure language instead of property value as locale for Link content type

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Related issues/PRs | https://github.com/sulu/SuluHeadlessBundle/pull/129
| License | MIT

#### What's in this PR?

Fixes bug for shadow pages and multi locale setups

### DIFF
--- a/src/Sulu/Component/Content/Types/Link.php
+++ b/src/Sulu/Component/Content/Types/Link.php
@@ -89,6 +89,7 @@ class Link extends SimpleContentType
     public function getContentData(PropertyInterface $property): ?string
     {
         $value = $property->getValue();
+        $locale = $property->getStructure()->getLanguageCode();
 
         if (!$value || !isset($value['provider'])) {
             return null;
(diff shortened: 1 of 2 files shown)


src/Sulu/Component/Content/Types/Link.php
```diff
@@ -93,6 +93,7 @@ public function getViewData(PropertyInterface $property): array
     public function getContentData(PropertyInterface $property): ?string
     {
         $value = $property->getValue();
+        $locale = $property->getStructure()->getLanguageCode();
```
alexander-schranz, line 92: Need to have a deeper look at this we have a similar issue with this here: https://github.com/sulu/sulu/pull/7091.

Expected should be if I'm on `/de-ch` which is the shadow to `/de` and the linked page is `/test` it should render `/de-ch/test` and not `/de/test` but only when `/de-ch/test` exists (is itself a shadow to de or its own page).

But thought in https://github.com/sulu/sulu/pull/7091 we needed for this access the getOriginalLocale and then fallback to getLanguageCode. Else if `/de-ch/test` is not published or does not exist it links to 404 page.

alexander-schranz: For the Link content type this looks correctly for smart content (#7091) we need to check why the getLanguageCode seems to return something else.

alexander-schranz (approved):

PR #7122, opened 2023-07-21
Using interfaces

mamazu:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?
Replacing the type hint for the `StructureMetadataFactory` with its interface.

#### Why?
If the user wants to replace the implementation of the `StructureMetadataFactory`, we should use the interface instead of the class.

#### Example Usage
Same as before

### DIFF
--- a/.env
+++ b/.env
@@ -2,6 +2,7 @@
 # Copy this file to .env file for development, create environment variables when deploying to production
 # https://symfony.com/doc/current/best_practices/configuration.html#infrastructure-related-configuration
 
+
 ###> symfony/framework-bundle ###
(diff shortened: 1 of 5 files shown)


alexander-schranz:
Looks good, can you repush so the 2.4 CI is correctly triggered :)

alexander-schranz (approved):

alexander-schranz:
@mamazu thank you!

PR #7125, opened 2023-07-28
Add support of noop proxy client in sulu http cache

vntw:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no 
| BC breaks? | no 
| Deprecations? | no  
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Right now, to use http cache tagging and its features (especially **twig functions like `fos_httpcache_tag`**) in all environments, you need to have an enabled `proxy_client`. If you're in a different env that has no `proxy_client` enabled, **these functions will be missing** and need to be "polyfilled" in some way, in order not to throw an error, which seems unnecessary.

So now it's possible to set this in your `sulu_http_cache.yaml` and all functionality will be available:

```yaml
sulu_http_cache:
  proxy_client:
    noop: ~
```

https://foshttpcachebundle.readthedocs.io/en/latest/reference/configuration/proxy-client.html?highlight=noop#noop

Tested in a 2.5 project without issues but happy to hear your input :)

#### Why?

For better DX and less user code

#### Example Usage

No changes to the config

#### To Do

Nothing

### DIFF
--- a/src/Sulu/Bundle/HttpCacheBundle/DependencyInjection/SuluHttpCacheExtension.php
+++ b/src/Sulu/Bundle/HttpCacheBundle/DependencyInjection/SuluHttpCacheExtension.php
@@ -44,6 +44,10 @@ class SuluHttpCacheExtension extends Extension implements PrependExtensionInterf
             ];
         }
 
+        if (\array_key_exists('noop', $config['proxy_client'])) {
+            $fosHttpCacheConfig['proxy_client']['noop'] = $config['proxy_client']['noop'];
+        }
+
         if ($config['proxy_client']['symfony']['enabled']) {
             $symfonyProxyClient = $config['proxy_client']['symfony'];
             $fosHttpCacheConfig['proxy_client']['symfony']['http']['servers'] =
@@ -90,7 +94,7 @@ class SuluHttpCacheExtension extends Extension implements PrependExtensionInterf
         $proxyClientAvailable = false;
         if (\array_key_exists('proxy_client', $config)) {
             foreach ($config['proxy_client'] as $proxyClient) {
-                if (true === $proxyClient['enabled']) {
+                if ((\is_bool($proxyClient) && $proxyClient) || true === $proxyClient['enabled']) {
                     $proxyClientAvailable = true;
                     break;
                 }
--- a/src/Sulu/Bundle/HttpCacheBundle/DependencyInjection/Configuration.php
+++ b/src/Sulu/Bundle/HttpCacheBundle/DependencyInjection/Configuration.php
@@ -49,6 +49,7 @@ class Configuration implements ConfigurationInterface
                 ->arrayNode('proxy_client')
                     ->addDefaultsIfNotSet()
                     ->children()
+                        ->booleanNode('noop')->end()
                         ->arrayNode('symfony')
                             ->canBeEnabled()
                             ->addDefaultsIfNotSet()


alexander-schranz:
I'm not sure why this check was added. /cc @wachterjohannes do you know?

src/Sulu/Bundle/HttpCacheBundle/DependencyInjection/SuluHttpCacheExtension.php
```diff
@@ -62,9 +62,7 @@ public function prepend(ContainerBuilder $container)
             }
         }
 
-        if (\array_key_exists('proxy_client', $fosHttpCacheConfig)) {
-            $fosHttpCacheConfig['tags']['enabled'] = $config['tags']['enabled'];
-        }
+        $fosHttpCacheConfig['tags']['enabled'] = $config['tags']['enabled'];
```
wachterjohannes: @vntw at the end it makes sense to have that - but the tests failing ... seems also to be checked in the foshttpcachebundle

wachterjohannes (changes requested):

src/Sulu/Bundle/HttpCacheBundle/DependencyInjection/SuluHttpCacheExtension.php
vntw: Oh sorry didn't see them, that's weird as it worked for me, I'll take a look and get back to you! Thanks for the reviews

alexander-schranz:
Looks like we are seeing the problem in the tests:

```bash
Symfony\Component\Config\Definition\Exception\InvalidConfigurationException: You need to configure a proxy_client to get the cache_manager needed for tag handling.
```

vntw:
@wachterjohannes @alexander-schranz thanks for the review, I took a deeper look and found (in my opinion) a much better and probably the intended solution: the [`noop` proxy client](https://foshttpcachebundle.readthedocs.io/en/latest/reference/configuration/proxy-client.html?highlight=noop#noop). I updated the PR description and code, so now the sulu http cache bundle supports configuring this. I tested this with the following config in dev and prod:

```yaml
sulu_http_cache:
    proxy_client:
        noop: ~

when@prod: &prod
    sulu_http_cache:
        proxy_client:
            symfony:
                enabled: true

when@stage: *prod
```

If the you and the tests are happy, I'll be happy to submit a docs PR to make a little note of it

alexander-schranz:
@vntw can you rebase this PR to 2.6 as adding configuration options this is kind of a feature and so should be part of 2.6. Also it seems like adoption of the `skeleton` is required so the new noop is used in case of `dev` and `test` env.

In this case I want also check this change with @dbu if this make sense to be used this way.

What I'm not sure if we need todo `noop: false` in case the other proxy clients are activated?

vntw:
Rebase done although this _should_ still be BC compatible IMO (but I'm not familiar with the sulu release process)

>What I'm not sure if we need todo noop: false in case the other proxy clients are activated?

Since we don't add any configuration that the user didn't do themselves in this regard, I don't think there's anything to do? If the config is invalid, fos http cache bundle should be the one to complain IMO (as it did with my previous change when the proxy_client was missing)

dbu:
seems reasonable to me. that  was the main intention of the noop client.

> What I'm not sure if we need todo `noop: false` in case the other proxy clients are activated?

i am not entirely sure but i would have thought the configuration overwrites the other proxy_client. to verify: what happens if you set more than 1 client at the same time? i'd expect that to give an error.

alexander-schranz:
In general it looks good from my side the only thing we need to check what happens if we now have multiple clients enabled should we set noop to false if other clients are enabled by default. See comment by @dbu https://github.com/sulu/sulu/pull/7125#issuecomment-1667627326  

I rebased to check what the current CI state is.

alexander-schranz:
@vntw Thank you!

PR #7131, opened 2023-08-10
Fixed tested methods in ForceTwoFactorSubscriberTest unit tests

stollr:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | 
| Related issues/PRs | 
| License | MIT
| Documentation PR | 

#### What's in this PR?

I have fixed the methods that were tested in the unit tests.

#### Why?

I think there is a typo, because the tested methods do not match the test name.
In fact this is not really a problem because both methods do the same, but theoretically this may change in the future..


alexander-schranz (approved):

alexander-schranz:
@stollr Thank you!

PR #7134, opened 2023-08-10
Fixed deprecated calls of getEntity and getEntityManager on doctrine lifecycle events

stollr:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | yes
| Deprecations? | no
| Fixed tickets | 
| Related issues/PRs | 
| License | MIT
| Documentation PR |

#### What's in this PR?

I have changed almost all calls of `getEntity` and `getEntityManager` to `getObject` and `getObjectManager` on the `LifecycleEventArgs`. 

There is one place, where an `EntityManager` is needed (have a look at `Sulu\Component\Persistence\EventSubscriber\ORM\UserBlameSubscriber`) and because it is not a bundle I am unsure if it is okay to use dependency injection in the constructor and inject the `EntityManager` per lazy service subscriber.

Additionally I have changed `Doctrine\ORM\Event\LifecycleEventArgs` to `Doctrine\Persistence\Event\LifecycleEventArgs`, because the use of the former is deprecated. But I assume that this is a bc break in case people have extended the event handlers.

Please give me feedback to the two issues and I'll adjust the PR accordingly.

#### Why?

Because of deprecations in doctrine/orm.

#### To Do

- [x] Fixing `getEntityManager` in `UserBlameSubscriber`
- [ ] Eventually revert the change of the `LifecycleEventArgs` class

### DIFF
--- a/src/Sulu/Component/Persistence/EventSubscriber/ORM/UserBlameSubscriber.php
+++ b/src/Sulu/Component/Persistence/EventSubscriber/ORM/UserBlameSubscriber.php
@@ -120,6 +120,8 @@ class UserBlameSubscriber implements EventSubscriber
 
     private function handleUserBlame(OnFlushEventArgs $event, UserInterface $user, bool $insertions)
     {
+        // TODO: calling getEntityManager on the event is deprecated, but access to the UnitOfWork is not
+        // guaranteed for an ObjectManager. So we need to get an EntityManager somehow different.
         $manager = $event->getEntityManager();
         $unitOfWork = $manager->getUnitOfWork();
 
--- a/src/Sulu/Component/Persistence/EventSubscriber/ORM/UserBlameSubscriber.php
+++ b/src/Sulu/Component/Persistence/EventSubscriber/ORM/UserBlameSubscriber.php
@@ -12,6 +12,7 @@
 namespace Sulu\Component\Persistence\EventSubscriber\ORM;
 
 use Doctrine\Common\EventSubscriber;
+use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Event\OnFlushEventArgs;
 use Doctrine\ORM\Events;
 use Doctrine\Persistence\Event\LoadClassMetadataEventArgs;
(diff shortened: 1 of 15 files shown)


src/Sulu/Component/Persistence/EventSubscriber/ORM/UserBlameSubscriber.php
```diff
@@ -123,6 +123,8 @@ public function onFlush(OnFlushEventArgs $event)
 
     private function handleUserBlame(OnFlushEventArgs $event, UserInterface $user, bool $insertions)
     {
+        // TODO: calling getEntityManager on the event is deprecated, but access to the UnitOfWork is not
+        // guaranteed for an ObjectManager. So we need to get an EntityManager somehow different.
```
stof: I suggest using `getObjectManager` and then making a check about the type of manager you receive

src/Sulu/Component/Persistence/EventSubscriber/ORM/UserBlameSubscriber.php
```diff
@@ -123,6 +123,8 @@ public function onFlush(OnFlushEventArgs $event)
...
         $manager = $event->getEntityManager();
         $unitOfWork = $manager->getUnitOfWork();
```
alexander-schranz: Think the suggestion of @stof make sense we just check if its an entity manager event if not early return.

```suggestion
        $manager = $event->getObjectManager();
        if (!$manager instanceof EntityManagerInterface {
            return;
        }
        
        $unitOfWork = $manager->getUnitOfWork();
```

src/Sulu/Component/Persistence/EventSubscriber/ORM/UserBlameSubscriber.php
stollr: Thanks. This is done.

src/Sulu/Component/Persistence/EventSubscriber/ORM/UserBlameSubscriber.php
stollr: Thank you, I have changed the PR according to your suggestion.

stollr:
@stof and @alexander-schranz I have adjusted the PR as suggested ;-)

alexander-schranz:
Thx some tests need to be adjusted (see 8.1, 8.2 failures) and also we need to define the correct `composer` requirements where this method exists (see 8.0 lowest currently failing). If I understand correctly we need to require `doctrine/orm: ^2.13`: https://github.com/doctrine/orm/commit/eea53397c5a4ed8191d0d84468062f3677fca96f

stollr:
@alexander-schranz I have increased the doctrine/orm version constraint, fixed tests and phpstan linter errors.

alexander-schranz (approved):

PR #7146, opened 2023-08-28
Fix dynamic property issues in phpstan via rector

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

First run rector with the following rule:

```php
<?php

declare(strict_types=1);

use Rector\CodeQuality\Rector\Class_\CompleteDynamicPropertiesRector;
use Rector\Config\RectorConfig;

return static function (RectorConfig $rectorConfig): void {
    $rectorConfig->paths([
        __DIR__ . '/src',
    ]);
};
```

Then go over all newly creatded `public $var` and change them to `private` and set the correct typehint. Mostly it are tests and mock which would need a fix here.

#### Why?

Make the PHPStan baseline smaller and activate better static analyzer by not build on top of any dynamic properties.

### DIFF
--- a/src/Sulu/Bundle/ContactBundle/Tests/Functional/Controller/AccountControllerTest.php
+++ b/src/Sulu/Bundle/ContactBundle/Tests/Functional/Controller/AccountControllerTest.php
@@ -2417,8 +2417,6 @@ class AccountControllerTest extends SuluTestCase
         $category->setKey($name);
         $category->setDefaultLocale($locale);
 
-        $this->category = $category;
-
         // name for first category
         $categoryTrans = $this->getContainer()->get('sulu.repository.category_translation')->createNew();
         $categoryTrans->setLocale($locale);
--- a/src/Sulu/Bundle/ContactBundle/Tests/Functional/Controller/ContactControllerTest.php
+++ b/src/Sulu/Bundle/ContactBundle/Tests/Functional/Controller/ContactControllerTest.php
@@ -2803,8 +2803,6 @@ class ContactControllerTest extends SuluTestCase
         $category->setKey($name);
         $category->setDefaultLocale($locale);
 
-        $this->category = $category;
-
         // name for first category
         $categoryTrans = $this->getContainer()->get('sulu.repository.category_translation')->createNew();
         $categoryTrans->setLocale($locale);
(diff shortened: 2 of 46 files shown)


src/Sulu/Bundle/ContactBundle/Tests/Functional/Controller/AccountControllerTest.php
```diff
@@ -2417,8 +2418,6 @@ private function createCategory(string $key, string $locale, string $name, strin
         $category->setKey($name);
         $category->setDefaultLocale($locale);
 
-        $this->category = $category;
-
```
alexander-schranz, line 2421: unused $this->category

src/Sulu/Bundle/ContactBundle/Tests/Functional/Controller/ContactControllerTest.php
```diff
@@ -2803,8 +2804,6 @@ private function createCategory(string $key, string $locale, string $name, strin
         $category->setKey($name);
         $category->setDefaultLocale($locale);
 
-        $this->category = $category;
-
```
alexander-schranz, line 2807: same as abobe was unused

src/Sulu/Bundle/CoreBundle/Tests/Unit/Cache/StructureWarmerTest.php
```diff
@@ -12,14 +12,25 @@
...
 use PHPUnit\Framework\TestCase;
+use Prophecy\Prophecy\ObjectProphecy;
 use Sulu\Bundle\CoreBundle\Cache\StructureWarmer;
+use Sulu\Component\Content\Compat\StructureManagerInterface;
 
 class StructureWarmerTest extends TestCase
 {
+    /**
+     * @var ObjectProphecy<StructureManagerInterface>
+     */
```
alexander-schranz, line 23: this where the typical issues that the mocked properties where not defined

src/Sulu/Bundle/WebsiteBundle/Tests/Application/Kernel.php
```diff
@@ -24,8 +24,6 @@ class Kernel extends SuluTestKernel
     public function __construct($environment, $debug, $suluContext = self::CONTEXT_ADMIN)
     {
         parent::__construct($environment, $debug, $suluContext);
-
-        $this->name = $suluContext;
```
alexander-schranz, line 28: this should already be handled by the SuluKernel here:

https://github.com/sulu/sulu/blob/337c078aef79e860ef955f5994c5e1adb554aa6f/src/Sulu/Component/HttpKernel/SuluKernel.php#L65-L67

Prokyonn (approved):

PR #7148, opened 2023-08-30
Fix incompatibility to Symfony 6.4 renderBlock method

alexander-schranz:
| Q | A
| --- | ---
| Bug fix?  yes
| New feature? | no
| BC breaks? | yes
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | https://github.com/sulu/SuluArticleBundle/pull/647
| License | MIT

#### What's in this PR?

Fix incompatibility to Symfony 6.4 renderBlock method.

#### Why?

Symfony introduced its own `renderBlock` method and we need to migrate to the compatible `renderBlockView` method.

#### Upgrade

Actually the upgrade here is hard, from Sulu side it would be.

```php
// before
$this->renderBlock(...);

// after
$this->renderBlockView(...);
```

but this would manipulate all `renderBlock` methods and rector rules should work always. So in this case we would need a custom rule which checks if the result of renderBlock/renderBlockView is used as a string or response.

```php
// before
$response = $this->renderBlock(...);
$string = $this->renderBlock(...);

new Response($string);

// after
$response = $this->renderBlock(...);
$string = $this->renderBlockView(...);

new Response($string);
```

So that is the hard case here to create automatic upgrade 🤔 . Inheritance make things always harder then composition.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,15 @@
 # Upgrade
 
+## 2.5.11
+
+### Rename WebsiteController::renderBlock to WebsiteController::renderBlockView
+
+Symfony 6.4 did add an own `renderBlock` method to its `AbstractController`.
+This would actually throw an error when projects upgrading to Symfony 6.4
+as the `renderBlock` method of Sulu is incompatible to Symfony `renderBlock`
+method. For these cases we rename `renderBlock` to `renderBlockView`.
+
+>>>>>>> adff3439d0 (Fix incompatibility to Symfony 6.4 renderBlock method)
 ## 2.5.7
 
 ### Constructor of ValidateWebspacesCommand changed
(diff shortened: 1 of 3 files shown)


UPGRADE.md
```diff
@@ -1,5 +1,14 @@
 # Upgrade
 
+## 2.5.11
+
+### Rename WebsiteController::renderBlock to WebsiteController::renderBlockView
+
+Symfony 6.4 did add an own `renderBlock` method to its `AbstractController`.
+This would actually throw an error when projects upgrading to Symfony 6.4
+as the `renderBlock` method of Sulu is incompatible to Symfony `renderBlock`
+method. For these cases we rename `renderBlock` to `renderBlockView`.
```
```suggestion
In Symfony 6.4, an independent `renderBlock` method was introduced to its `AbstractController`. 
This change poses issues for projects upgrading to Symfony 6.4, as the `renderBlock` method in Sulu is incompatible with Symfony's `renderBlock` method.
To address this issue, we have to rename the Sulu `renderBlock` method to `renderBlockView`.
```

Prokyonn (approved):

PR #7163, opened 2023-09-20
Fix categories menu permissions

### LINKED ISSUE #7088: VIEW permissions for User Roles are not working like expected
| Q | A
| --- | ---
| Sulu Version | 2.5.x 
| PHP Version | 8.1

# Actual Behavior

When using the User Roles for Sulu features like snippets, contacts, categories, tags, ... I can choose to enable only the VIEW permission. 
image

After logging in with an admin user with the related user role, I cannot see the related navigation items or access the list views/pages. Now those navigation items and list views require the EDIT permission.  

The same behavior you can see in the ArticleBundle and FormBundle.

# Expected Behavior

As an admin user with only VIEW-permissions I should see:
- the corresponding navigation item
- the list view/page without add or delete buttons
- an edit or view action to open the form without a save button (see screenshot below)
image

# Steps to Reproduce

- Create a new User Role
- Enable only VIEW permission for Setting > categories
- Create an new admin account (Contact > People) and assign the new created User Role
- Login with the admin account
- You will see there is no "Settings > Categories" navigation item in the sidebar

zanturik:
| Q | A
| --- | ---
| Bug fix? |  yes
| New feature? | no 
| BC breaks? | no 
| Deprecations? | no 
| Fixed tickets | fixes #7088 
| Related issues/PRs |
| License | MIT
| Documentation PR |

#### What's in this PR?

Bugfix category bundle menu/toolbar permissions issue

#### Why?

https://github.com/sulu/sulu/issues/7088
Users with VIEW role were not able to see anything, because the role was ignored.
Now they can see the list/category.
Toolbar actions are appearing according to EDIT/ADD/DELETE role

#### Example Usage

Now you can run tests on CategoryAdmin :) 

#### To Do
[ ] add some tests on other CategoryAdmin functions

### DIFF
--- a/src/Sulu/Bundle/CategoryBundle/Admin/CategoryAdmin.php
+++ b/src/Sulu/Bundle/CategoryBundle/Admin/CategoryAdmin.php
@@ -74,26 +74,33 @@ class CategoryAdmin extends Admin
 
         $formToolbarActions = [];
         $listToolbarActions = [];
+        $keywordsToolbarActions = [];
 
         if ($this->securityChecker->hasPermission(self::SECURITY_CONTEXT, PermissionTypes::ADD)) {
             $listToolbarActions[] = new ToolbarAction('sulu_admin.add');
+            $keywordsToolbarActions[] = new ToolbarAction('sulu_admin.add');
         }
 
         if ($this->securityChecker->hasPermission(self::SECURITY_CONTEXT, PermissionTypes::EDIT)) {
-            $formToolbarActions[] = new ToolbarAction('sulu_admin.save');
             $listToolbarActions[] = new ToolbarAction('sulu_admin.move');
+        }
+
+        if ($this->securityChecker->hasPermission(self::SECURITY_CONTEXT, PermissionTypes::EDIT)
+           || $this->securityChecker->hasPermission(self::SECURITY_CONTEXT, PermissionTypes::ADD)) {
+            $formToolbarActions[] = new ToolbarAction('sulu_admin.save');
         }
 
         if ($this->securityChecker->hasPermission(self::SECURITY_CONTEXT, PermissionTypes::DELETE)) {
             $formToolbarActions[] = new ToolbarAction('sulu_admin.delete');
             $listToolbarActions[] = new ToolbarAction('sulu_admin.delete');
+            $keywordsToolbarActions[] = new ToolbarAction('sulu_admin.delete');
         }
 
         if ($this->securityChecker->hasPermission(self::SECURITY_CONTEXT, PermissionTypes::VIEW)) {
             $listToolbarActions[] = new ToolbarAction('sulu_admin.export');
         }
 
-        if ($this->securityChecker->hasPermission(self::SECURITY_CONTEXT, PermissionTypes::EDIT)) {
+        if ($this->securityChecker->hasPermission(self::SECURITY_CONTEXT, PermissionTypes::VIEW)) {
             $listViewBuilder = $this->viewBuilderFactory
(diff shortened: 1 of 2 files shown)


src/Sulu/Bundle/CategoryBundle/Admin/CategoryAdmin.php
```diff
@@ -59,7 +59,7 @@ public function __construct(
 
     public function configureNavigationItems(NavigationItemCollection $navigationItemCollection): void
     {
-        if ($this->securityChecker->hasPermission(static::SECURITY_CONTEXT, PermissionTypes::EDIT)) {
+        if ($this->securityChecker->hasPermission(static::SECURITY_CONTEXT, PermissionTypes::VIEW)) {
```
alexander-schranz: This is expected behaviour of Sulu we are showing navigation points only for EDIT permissions.

This was decided some time ago. To have a difference between I can select `categories` (VIEW) inside pages and I can see the main menu point (EDIT).

This behaviour is use din all Admins you only see a navigation item when you have Edit permissions. https://github.com/sulu/sulu/blob/9361c365670e4b540897354b1c104ced1b0d0368/src/Sulu/Bundle/ContactBundle/Admin/ContactAdmin.php#L78

VIEW permission in this case is only for `selections`.

zanturik: mmm.... ok.
So, user with VIEW and ADD shouldn't see navigation point? As the user with VIEW and DELETE? 
It's a little bit confusing for users, as for me... but ok, I can revert to this  logic and just keep the fix for toolbar.

zanturik: @alexander-schranz fixed. 
I'm still a little bit puzzled about applications logic.  I can understand the business decision of hiding menu elements if user cannot change anything (edit), but as for me application is not aligned with this logic. For real we have a hierarchy of roles - EDIT requires VIEW to be able to edit something (user with EDIT without VIEW can see the menu item, but cannot see the list), DELETE/ADD requires EDIT to be able to delete/add something (otherwise they don't see  menu item to perform their  actions). On the other side, application allows to set roles separately, without indicating these requirements or automatically setting all required roles. It leads to confusion, so I think  bug reports are going to  appear again.

Anyway, the toolbar actions are fixed, buttons appear according to users  roles, there is no situation when user see the action button and gets "error" after. Please, review.

alexander-schranz: > VIEW and ADD shouldn't see navigation point.

If you want to `ADD` you also need `EDIT` permissions in case of the UI. There are more usecases in roles that People are allowed to `EDIT` things but not allowed to `ADD` new things then the other way around. And that is why we did choose `EDIT` over `ADD`.

> VIEW

This is more a case because Sulu is API based you require a lot of View Permissions even if you should not see that Menu Items. So to not confused people with limited access with a lot of menu points they not require because they require VIEW permissions to use any kind of selections in pages, ... we decided that you atleast need `EDIT` permissions to see the Navigation Item. There where lot of discussion about it but we decided to go that way. Even we decided to seperate menu items from the view / edit permissions, at current state we can not change it as that would be a major BC break for all application with roles building on top of that logic.

alexander-schranz:
Thx for your contribution. Can you for the current fixes rebase to `2.4` branch as bugfixes should go on the lowest branch. You can rebase it via:

```
# and then --force push
```

zanturik:
done

alexander-schranz:
@zanturik thank you!

PR #7164, opened 2023-09-21
Fix link content type return old file version

### LINKED ISSUE #7119: Media of "link" content type returns old file version (MediaLinkProvider)
| Q | A
| --- | ---
| Sulu Version | 2.5.9
| PHP Version | 8.1.20
| DB Version | 8.0.31

# Actual Behavior

When using the [`link` content type](https://docs.sulu.io/en/2.5/reference/content-types/link.html) to select a **media**, the stored URL contains the latest file version at that time.

When uploading a new file version (of the linked media), the page still has the old version in the download URL, e.g: `?v=1`, although the newest version is 2.

# Expected Behavior

The url of the `link` content type should return the media URL of the latest file version.

# Steps to Reproduce

Add a link content type to a page.
```xml
    <meta>
        <title lang="en">Link</title>
        <title lang="de">Link</title>
    </meta>

    <params>
        <param name="enable_target" value="true"/>
        <param name="enable_title" value="true"/>
        <param name="types" type="collection">
            <param name="page"/>
            <param name="external"/>
            <param name="media"/>
        </param>
    </params>
</property>
```

Select a **media** and print the the link `{{ dump(content.link) }}`.

Then, upload a new version of the selected media and dump the link again. It still shows the link with `?v=1` then.

zanturik:
Link type was returning first version of the link, no matter how many versions it has. Fixed, now  it returns the latest version.

| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no 
| BC breaks? | no 
| Deprecations? | no
| Fixed tickets | fixes #7119 
| Related issues/PRs | 
| License | MIT
| Documentation PR | 

#### What's in this PR?

Now Link type returns the latest  link

#### Why?

[](https://github.com/sulu/sulu/issues/7119)

### DIFF
--- a/src/Sulu/Component/Content/Types/Link.php
+++ b/src/Sulu/Component/Content/Types/Link.php
@@ -107,7 +107,8 @@ class Link extends SimpleContentType
             return null;
         }
 
-        $url = \reset($linkItems)->getUrl();
+        $url = \end($linkItems)->getUrl();
+
         if (isset($value['anchor'])) {
             $url = \sprintf('%s#%s', $url, $value['anchor']);
         }
--- a/src/Sulu/Component/Content/Types/Link.php
+++ b/src/Sulu/Component/Content/Types/Link.php
@@ -107,8 +107,7 @@ class Link extends SimpleContentType
             return null;
         }
 
-        $url = \end($linkItems)->getUrl();
-
+        $url = \reset($linkItems)->getUrl();
         if (isset($value['anchor'])) {
             $url = \sprintf('%s#%s', $url, $value['anchor']);
         }
--- a/src/Sulu/Bundle/MediaBundle/Entity/MediaRepository.php
+++ b/src/Sulu/Bundle/MediaBundle/Entity/MediaRepository.php
@@ -216,7 +216,7 @@ class MediaRepository extends EntityRepository implements MediaRepositoryInterfa
     {
         $queryBuilder = $this->createQueryBuilder('media')
             ->leftJoin('media.files', 'file')
-            ->leftJoin('file.fileVersions', 'fileVersion', Join::WITH, 'fileVersion.version = file.version')            
+            ->leftJoin('file.fileVersions', 'fileVersion', Join::WITH, 'fileVersion.version = file.version')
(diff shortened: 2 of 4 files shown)


src/Sulu/Component/Content/Types/Link.php
```diff
@@ -111,7 +111,7 @@ public function getContentData(PropertyInterface $property): ?string
             return null;
         }
 
-        $url = \reset($linkItems)->getUrl();
+        $url = \end($linkItems)->getUrl();
```
alexander-schranz: The array here is based on on the call in Line `108` so we are giving one `href` in it and we should get one href out of it.

So it sounds like an issue with one of our LinkProviders if it returns for a single `href` multiple results, that should never be and is then a bug of a LinkProvider. Which one is failing in which cases?

alexander-schranz: I see its the Media provider it looks like here is missing something:

https://github.com/sulu/sulu/blob/9361c365670e4b540897354b1c104ced1b0d0368/src/Sulu/Bundle/MediaBundle/Entity/MediaRepository.php#L219

This should join the current version `file.version = fileVersion.version` and not return all versions.

zanturik: @alexander-schranz I thought about it, but the problem is that Interface declares that function  return "information for all medias", not just the latest version. And I'm not sure who can  already use it this way. To "show all versions", for example. 

https://github.com/sulu/sulu/blob/9361c365670e4b540897354b1c104ced1b0d0368/src/Sulu/Bundle/MediaBundle/Entity/MediaRepositoryInterface.php#L67

alexander-schranz: All media != all file versions we are returning beside the version history in all cases only the last active version. So it fine to filter here also by the actual version.

zanturik: ok, fixed, please review.

alexander-schranz (approved):

alexander-schranz:
@zanturik thank you!

PR #7168, opened 2023-09-29
Improve error handling of loadByResourceLocator

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes nothing
| License | MIT

#### What's in this PR?
Better types

#### Why?
I am trying to understand how this works and right typing would help.

### DIFF
--- a/src/Sulu/Component/Content/Types/ResourceLocator/Mapper/PhpcrMapper.php
+++ b/src/Sulu/Component/Content/Types/ResourceLocator/Mapper/PhpcrMapper.php
@@ -131,7 +131,7 @@ class PhpcrMapper implements ResourceLocatorMapperInterface
      * @param string $languageCode
      * @param string $segmentKey
      *
-     * @return \PHPCR\NodeInterface
+     * @return NodeInterface|null
      */
     private function iterateRouteNodes(
         NodeInterface $node,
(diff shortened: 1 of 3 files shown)


alexander-schranz:
We should target `2.4` for this. @wachterjohannes can you confirm that the new exception also makes from your point of view sense?

src/Sulu/Component/Content/Types/ResourceLocator/Mapper/PhpcrMapper.php
```diff
@@ -276,10 +276,15 @@ public function loadByResourceLocator($resourceLocator, $webspaceKey, $languageC
...
-                throw new ResourceLocatorMovedException(
-                    $this->getResourceLocator($realPath->getPath(), $webspaceKey, $languageCode, $segmentKey),
-                    $realPath->getIdentifier()
-                );
+                $locator = $this->getResourceLocator($realPath->getPath(), $webspaceKey, $languageCode, $segmentKey);
+                if (false === $locator) {
+                    throw new ResourceLocatorNotFoundException(\sprintf(
+                        'Couldn\'t generate resource locator for path: %s',
+                        $realPath->getPath()
+                    ));
```
alexander-schranz, line 283: @wachterjohannes are you fine with this change? See https://github.com/sulu/sulu/pull/7168/commits/d264cc1bfc910ef0e92372302596e198603cd485

mamazu: Before it would throw a type error, so it's a better error message.

wachterjohannes: @mamazu @alexander-schranz yes it makes sense to have that

PR #7169, opened 2023-10-02
Removing unused service argument

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | yes
| License | MIT

#### Why?
The property is not used anywhere and instanciating the service is quite expensive so this should also improve the container speed a little.

#### To Do

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -10,6 +10,9 @@ Removing deprecated functions and properties:
 - `Sulu\Component\Localization\Localization::isXDefault`
 - `Sulu\Bundle\MediaBundle\Controller\MediaController::$entityKey`
 
+Removing unused properties from services
+- Sulu\Component\Webspace\Analyzer\Attributes\WebsiteRequestProcessor::contentMapper
+
 ## 2.6.0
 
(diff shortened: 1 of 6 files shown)


UPGRADE.md
```diff
@@ -10,6 +10,9 @@ Removing deprecated functions and properties:
 - `Sulu\Component\Localization\Localization::isXDefault`
 - `Sulu\Bundle\MediaBundle\Controller\MediaController::$entityKey`
 
+Removing unused properties from services
+- Sulu\Component\Webspace\Analyzer\Attributes\WebsiteRequestProcessor::contentMapper
```
```suggestion
Removing unused arguments:

- `Sulu\Component\Webspace\Analyzer\Attributes\WebsiteRequestProcessor::__construct` `$contentMapper` (2nd argument) removed
```

alexander-schranz (approved):

mamazu:
Should we also make a fix for the lower versions and mark the content manager as lazy so that it get only instantiated when it's needed?

alexander-schranz:
@mamazu think marking the `sulu.content.mapper` service as `lazy` should not hurt 🤔

PR #7172, opened 2023-10-13
Fix nominatim location

martinlagler:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT
| Documentation PR | sulu/sulu-docs#781

#### What's in this PR?

Adaptions for the Mapquest Api to fix geolocation in the LocationBundle.

#### Why?

Because the Mapquest Api has changed and isn't working anymore.

### DIFF
--- a/src/Sulu/Bundle/LocationBundle/DependencyInjection/Configuration.php
+++ b/src/Sulu/Bundle/LocationBundle/DependencyInjection/Configuration.php
@@ -32,7 +32,7 @@ class Configuration implements ConfigurationInterface
                             ->addDefaultsIfNotSet()
                             ->children()
                                 ->scalarNode('api_key')->defaultValue('')->end()
-                                ->scalarNode('endpoint')->defaultValue('http://open.mapquestapi.com/nominatim/v1/search.php')->end()
+                                ->scalarNode('endpoint')->defaultValue('https://nominatim.openstreetmap.org/search')->end()
                             ->end()
                         ->end()
                         ->arrayNode('google')


alexander-schranz:
@martinlagler one of the tests is failing:

Bildschirmfoto 2023-10-16 um 13 55 04

Run:

```bash
bin/runtests -i -t LocationBundle --flags="--filter=NominatimGeolocatorTest"
```

to debug this test.

@martinlagler also rebase to 2.4:

```bash
git fetch origin
```

src/Sulu/Bundle/LocationBundle/Geolocator/Service/NominatimGeolocator.php
```diff
@@ -94,33 +95,51 @@ public function locate(string $query): GeolocatorResponse
...
-                'setStreet' => 'road',
-                'setNumber' => 'house_number',
-                'setCode' => 'postcode',
-                'setTown' => 'city',
-                'setCountry' => 'country_code',
-            ] as $method => $key) {
-                if (isset($result['address'][$key])) {
-                    $location->$method($result['address'][$key]);
+
+        if (0 === $responseBody['info']['statuscode']) {
```
alexander-schranz: use early returns instead of nested if else statement:

```php
if (0 !== ($responseBody['info']['statuscode'] ?? null)) {
    throw new Exception($responseBody['info']['messages'][0] ?? 'Unexpected error');
}

if (!\array_key_exists('results', (array) $responseBody) and 0 < \count($responseBody['results'])) {
    throw new Exception('No results found.');
}
```

We also can not be 100% if the keys exists or go with `??` a default value.

src/Sulu/Bundle/LocationBundle/Geolocator/Service/NominatimGeolocator.php
```diff
@@ -94,33 +95,51 @@ public function locate(string $query): GeolocatorResponse
...
+
+                        $geoLocation->setId($locationId);
+                        $displayTitle = \trim($geoLocation->getStreet() . ', ' . $geoLocation->getTown() . ', ' . $geoLocation->getCountry(), ',');
+                        $geoLocation->setDisplayTitle($displayTitle);
+
+                        $response->addLocation($geoLocation);
+                    }
                 }
+            } else {
+                throw new Exception('No results found.');
```
alexander-schranz: We should avoid general exception and go with a specific exception. Else we could directly use `Assert::` and and avoid the if statements.

src/Sulu/Bundle/LocationBundle/Geolocator/Service/NominatimGeolocator.php
```diff
@@ -94,33 +95,51 @@ public function locate(string $query): GeolocatorResponse
...
-            ] as $method => $key) {
-                if (isset($result['address'][$key])) {
-                    $location->$method($result['address'][$key]);
+
+        if (0 === $responseBody['info']['statuscode']) {
+            if (\array_key_exists('results', (array) $responseBody) and 0 < \count($responseBody['results'])) {
+                $results = $responseBody['results'];
+
+                foreach ($results as $result) {
+                    $locations = $result['locations'];
```
alexander-schranz: what happens if the `locations` is empty? we should maybe go with `?? [];`

alexander-schranz (changes requested):
@martinlagler lets have quick call about it how we maybe can make this more readable.

src/Sulu/Bundle/LocationBundle/DependencyInjection/Configuration.php
```diff
@@ -32,7 +32,7 @@ public function getConfigTreeBuilder()
                             ->addDefaultsIfNotSet()
                             ->children()
                                 ->scalarNode('api_key')->defaultValue('')->end()
-                                ->scalarNode('endpoint')->defaultValue('http://open.mapquestapi.com/nominatim/v1/search.php')->end()
+                                ->scalarNode('endpoint')->defaultValue('https://www.mapquestapi.com/geocoding/v1/address')->end()
```
alexander-schranz: Should we change to the nominate endpoint suggested here by 
@remdan: https://github.com/sulu/sulu-docs/pull/781#issuecomment-1816560636

Does somebody know the history why? Not longer works? http://open.mapquestapi.com/nominatim/v1/search.php

Was the whole endpoint moved to https://nominatim.openstreetmap.org/search

PR #7174, opened 2023-10-16
Marking the ContentMapper as lazy

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Related issues/PRs | backport of #7169
| License | MIT
| Documentation PR | -

#### What's in this PR?
Making the service definition of the ContentMapper.

#### Why?
The `ContentMapper` is a heavy service (lots of dependencies) so we only want to instantiate the service when it's really needed.

#### Example Usage
Same as before Symfony does the magic under the hood.

### DIFF
--- a/src/Sulu/Bundle/CoreBundle/Resources/config/content.xml
+++ b/src/Sulu/Bundle/CoreBundle/Resources/config/content.xml
@@ -52,7 +52,7 @@
         <service id="sulu.content.template_resolver" public="false" class="%sulu.content.template_resolver.class%"/>
 
         <!-- content mapper -->
-        <service id="sulu.content.mapper" class="%sulu.content.mapper.class%" public="true">
+        <service id="sulu.content.mapper" class="%sulu.content.mapper.class%" public="true" lazy="true">
             <argument type="service" id="sulu_document_manager.document_manager" />
             <argument type="service" id="sulu_core.webspace.webspace_manager"/>
             <argument type="service" id="form.factory" />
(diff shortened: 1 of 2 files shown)


alexander-schranz:
Can you target the `2.4` branch for this?

```
```

mamazu:
Sure.

mamazu:
@alexander-schranz The failures aren't related to my service change though, right?

alexander-schranz:
Looks related to the changes as the 2.4 itself runs without problems.

mamazu:
Yes, you are right. I found the issue: The `delete` function of the `ContentMapper` has a function signature that allows for either 2 or 3 arguments (for BC reasons). The problem here is that Symfony when proxying strips all excess parameters from the parameter list. 

So if we want to keep the BC behavior we can't make this service lazy. I updated the merge request to use the new parameter order. So the test would work with the lazy service now.

alexander-schranz:
This sound interesting as that pattern is very common in backwards compatibility it maybe make sense to report that symfony.

mamazu:
I think the feature of force removal has been completely removed. So the test would not work with the "new" version of the signature.

alexander-schranz:
> I think the feature of force removal has been completely removed.

The feature forceRemoveChildren still exists: https://github.com/sulu/sulu/blob/60155ff1046247d3316f4bba0c2df1ee8b66b4c3/src/Sulu/Component/Content/Mapper/ContentMapper.php#L469

It is just written that way for BC compatibility as adding parameter would be a bc break when somebody decorates/extends the service. So that is why it was added this way in 2.4.

mamazu:
Ah okay that makes sense. Well the `ContentMapper` is going to be removed in 3.0 anyways. So I guess closing this is probably the best move.

PR #7176, opened 2023-10-17
Improve type information in the NavigationItem

mamazu:
Documenting the actual types a little better

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Admin/Navigation/NavigationItem.php
+++ b/src/Sulu/Bundle/AdminBundle/Admin/Navigation/NavigationItem.php
@@ -11,6 +11,9 @@
 
 namespace Sulu\Bundle\AdminBundle\Admin\Navigation;
 
+/**
+ * @implements \Iterator<int|null, self>
+ */
 class NavigationItem implements \Iterator
 {
     /**
@@ -28,7 +31,7 @@ class NavigationItem implements \Iterator
     protected $name;
 
     /**
-     * @var string
+     * @var string|null
      */
     protected $label;
 
@@ -40,7 +43,7 @@ class NavigationItem implements \Iterator
     protected $icon;
 
     /**
-     * @var string
+     * @var string|null
      */
     protected $view;
 
@@ -52,14 +55,14 @@ class NavigationItem implements \Iterator
     /**
      * Contains the children of this item, which are other NavigationItems.
      *
-     * @var array
+     * @var array<self>
      */
     protected $children = [];
 
     /**
      * The current position of the iterator.
      *
-     * @var int
+     * @var int|null
      */
     protected $position;
 
@@ -96,6 +99,8 @@ class NavigationItem implements \Iterator
      * Sets the id of the NavigationItem.
      *
      * @param string $id
+     *
+     * @return void
      */
     public function setId($id)
     {
@@ -116,6 +121,8 @@ class NavigationItem implements \Iterator
      * Sets the name being displayed in the navigation.
      *
      * @param string $name The name being displayed in the navigation
+     *
+     * @return void
      */
     public function setName($name)
     {
@@ -146,6 +153,8 @@ class NavigationItem implements \Iterator
      * Set the icon of the NavigaitonItem.
      *
      * @param string $icon The icon of the NavigationItem
+     *
+     * @return void
      */
     public function setIcon($icon)
     {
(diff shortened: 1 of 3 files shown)


src/Sulu/Bundle/AdminBundle/Admin/Navigation/NavigationItem.php
```diff
@@ -11,6 +11,9 @@
 
 namespace Sulu\Bundle\AdminBundle\Admin\Navigation;
 
+/**
+ * @implements \Iterator<int|null, self>
```
alexander-schranz: I think there is no case a iterator can has a `null` as key so this would be:

```suggestion
 * @implements \Iterator<int, self>
```

mamazu: If you don't set the position and just call $this->key(), then it will return null. Maybe initalizing it to 0 would make more sense? The we don't need the null in the type definition of `position`.

alexander-schranz: Got point! I think it would make sense to set `position = 0;` because `var_dump([null => 'hello']);` is converted to `['' => 'hello']` which is neither null or int. https://3v4l.org/WImgU

src/Sulu/Bundle/AdminBundle/Admin/Navigation/NavigationItem.php
```diff
@@ -52,14 +55,14 @@ class NavigationItem implements \Iterator
...
-     * @var array
+     * @var array<self>
      */
     protected $children = [];
 
     /**
      * The current position of the iterator.
      *
-     * @var int
+     * @var int|null
```
alexander-schranz: as discussed above let set a default value 0 so we can make sure position is always a int:

```suggestion
     * @var int
```

src/Sulu/Bundle/AdminBundle/Admin/Navigation/NavigationItem.php
```diff
@@ -209,15 +223,17 @@ public function getChildren()
...
+     * @return void
      */
     public function setPosition($position)
     {
         $this->position = $position;
     }
 
     /**
-     * @return int
+     * @return int|null
```
alexander-schranz:
```suggestion
     * @return int```

File: src/Sulu/Bundle/AdminBundle/Admin/Navigation/NavigationItem.php [hunk 4]
```diff
@@ -209,15 +223,17 @@ public function getChildren()
     }
 
     /**
-     * @param int $position
+     * @param int|null $position
```
alexander-schranz:
```suggestion
     * @param int $position
```

src/Sulu/Bundle/AdminBundle/Admin/Navigation/NavigationItem.php
```diff
@@ -52,14 +55,14 @@ class NavigationItem implements \Iterator
...
      */
     protected $children = [];
 
     /**
      * The current position of the iterator.
      *
-     * @var int
+     * @var int|null
      */
     protected $position;
```
alexander-schranz:
```suggestion
    protected $position = 0;
```

mamazu:
Technically this should break the sorting tests as the sorting before assumed position = null -> PHP_MAX_INT which means it would be placed at the end.

alexander-schranz:
@mamazu did it change something on the order of the current navigation items in example the demo which has some additiona bundles?

mamazu:
I didn't change any tests only the code.

mamazu:
I haven't got a dev shop installed but this is only an issue if you don't set the position.

mamazu:
image
The Sulu demo has still the same ordering after the change.

alexander-schranz:
@mamazu the screenshot shows the Website the NavigationItem are for the Admin :)

mamazu:
Where would I find that?

alexander-schranz:
On the left side in the admin there is the admin navigation 😄 which should be listed still in the same order as here :

Bildschirmfoto 2023-10-19 um 17 43 45

mamazu:
image
Yeah of course, totally lost track of what I was working on with all those types in my head. But looks the same to me.

alexander-schranz:
No problem, I clicked rebase to check if the phpstan baseline is uptodate

alexander-schranz (approved):

alexander-schranz:
@mamazu Thank you!

PR #7177, opened 2023-10-17
Add cache invalidation for snippet areas

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Invalidate snippet area cache, when the area snippet is modified/removed

#### Why?

Cache invalidation does not work properly for default area snippets that were updated

### DIFF
--- a/src/Sulu/Bundle/SnippetBundle/Content/SingleSnippetSelection.php
+++ b/src/Sulu/Bundle/SnippetBundle/Content/SingleSnippetSelection.php
@@ -37,14 +37,21 @@ class SingleSnippetSelection extends SimpleContentType implements PreResolvableC
      */
     private $snippetReferenceStore;
 
+    /**
+     * @var ReferenceStoreInterface
+     */
+    private $snippetAreaReferenceStore;
+
     public function __construct(
         SnippetResolverInterface $snippetResolver,
         DefaultSnippetManagerInterface $defaultSnippetManager,
-        ReferenceStoreInterface $snippetReferenceStore
+        ReferenceStoreInterface $snippetReferenceStore,
+        ReferenceStoreInterface $snippetAreaReferenceStore
     ) {
         $this->snippetResolver = $snippetResolver;
         $this->defaultSnippetManager = $defaultSnippetManager;
         $this->snippetReferenceStore = $snippetReferenceStore;
+        $this->snippetAreaReferenceStore = $snippetAreaReferenceStore;
 
         parent::__construct('SingleSnippetSelection', null);
     }
@@ -123,6 +130,7 @@ class SingleSnippetSelection extends SimpleContentType implements PreResolvableC
     {
         try {
             $snippet = $this->defaultSnippetManager->load($webspaceKey, $snippetArea, $locale);
+            $this->snippetAreaReferenceStore->add($snippetArea);
         } catch (WrongSnippetTypeException $exception) {
             return null;
         }
--- a/src/Sulu/Bundle/SnippetBundle/Content/SingleSnippetSelection.php
+++ b/src/Sulu/Bundle/SnippetBundle/Content/SingleSnippetSelection.php
@@ -38,7 +38,7 @@ class SingleSnippetSelection extends SimpleContentType implements PreResolvableC
     private $snippetReferenceStore;
 
     /**
-     * @var ReferenceStoreInterface
+     * @var ReferenceStoreInterface|null
      */
     private $snippetAreaReferenceStore;
 
--- a/src/Sulu/Bundle/SnippetBundle/Content/SingleSnippetSelection.php
+++ b/src/Sulu/Bundle/SnippetBundle/Content/SingleSnippetSelection.php
@@ -46,7 +46,7 @@ class SingleSnippetSelection extends SimpleContentType implements PreResolvableC
         SnippetResolverInterface $snippetResolver,
         DefaultSnippetManagerInterface $defaultSnippetManager,
         ReferenceStoreInterface $snippetReferenceStore,
-        ReferenceStoreInterface $snippetAreaReferenceStore
+        ?ReferenceStoreInterface $snippetAreaReferenceStore = null
     ) {
         $this->snippetResolver = $snippetResolver;
         $this->defaultSnippetManager = $defaultSnippetManager;
--- a/src/Sulu/Bundle/SnippetBundle/Content/SingleSnippetSelection.php
+++ b/src/Sulu/Bundle/SnippetBundle/Content/SingleSnippetSelection.php
@@ -53,6 +53,14 @@ class SingleSnippetSelection extends SimpleContentType implements PreResolvableC
         $this->snippetReferenceStore = $snippetReferenceStore;
         $this->snippetAreaReferenceStore = $snippetAreaReferenceStore;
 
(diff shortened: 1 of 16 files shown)


src/Sulu/Bundle/SnippetBundle/EventListener/CacheInvalidationSubscriber.php
```diff
@@ -0,0 +1,91 @@
...
+     */
+    private $defaultSnippetManager;
+
+    public function __construct(
+        ?CacheManager $cacheManager,
+        DefaultSnippetManagerInterface $defaultSnippetManager
+    ) {
+        $this->cacheManager = $cacheManager;
+        $this->defaultSnippetManager = $defaultSnippetManager;
+    }
```
mamazu: Maybe you could use constructor property promotion for this. Then the properties also have the right type and you don't need the doc-comments like so:
```suggestion
    public function __construct(
        private ?CacheManager $cacheManager,
        private DefaultSnippetManagerInterface $defaultSnippetManager
    ) {
    }
```

Prokyonn: Depending if we also want this in 2.4, then we would have to support also php 7.4 😅

mamazu: Yeah, you mean 7.2 because in 7.4 we would at least have typed properties. The php language has grown quite a bit in a few versions. :tada:

src/Sulu/Bundle/SnippetBundle/Twig/SnippetAreaTwigExtension.php
```diff
@@ -74,6 +83,8 @@ public function loadByArea($area, $webspaceKey = null, $locale = null)
             $locale = $this->requestAnalyzer->getCurrentLocalization()->getLocale();
         }
 
+        $this->snippetAreaReferenceStore->add($area);
```
alexander-schranz, line 85: This could let to invalidation of the whole website when a global area snippet is used. Lets discuss this with @chirimoya on thursday.

alexander-schranz (changes requested):
~~We should target 2.4 for this.~~ as discussed we add a new feature flags for snippets here

src/Sulu/Bundle/SnippetBundle/Tests/Unit/Twig/SnippetAreaTwigExtensionTest.php
```diff
@@ -64,6 +70,7 @@ public function testLoadByArea(): void
         $this->requestAnalyzer->getWebspace()->willReturn($this->webspace->reveal());
         $this->requestAnalyzer->getCurrentLocalization()->willReturn($this->localization);
         $this->snippetResolver = $this->prophesize(SnippetResolverInterface::class);
+        $this->snippetAreaReferenceStore = $this->prophesize(ReferenceStoreInterface::class);
```
alexander-schranz: Would not mock this, makes tests easier.

src/Sulu/Bundle/SnippetBundle/Twig/SnippetAreaTwigExtension.php
Prokyonn: In summary of our discussion, we will introduce a new parameter to the area configuration in the snippet template, which will determine whether the snippet-area should be excluded from the cache invalidation. By default, the snippet-area will always be invalidated. The only exception is for the automatically created `area`, they should not be invalidated, otherwise the old behaviour would be changed.

The decision of whether the invalidation should take place will be integrated into the CacheInvalidationSubscriber. Moreover, the snippet-area will always be included in the reference store. The benefit of this is, that we only have to check in one class if the snippet-area should be invalidated.

wachterjohannes (approved):

wachterjohannes:
@Prokyonn please rebase to see if the tests turn green and @alexander-schranz really target that to 2.4?

alexander-schranz:
2.6 is correct target branch for this.

Prokyonn:
> @Prokyonn please rebase to see if the tests turn green and @alexander-schranz really target that to 2.4?

@wachterjohannes php 8.0 did fail because of the `readonly` keyword 😅 

last week we discussed it and decided that it will be part of 2.6 and not backported to 2.4

src/Sulu/Bundle/SnippetBundle/Content/SnippetContent.php
```diff
@@ -43,23 +43,30 @@ class SnippetContent extends ComplexContentType implements ContentTypeExportInte
...
     /**
-     * @param true $defaultEnabled
+     * @param bool $defaultEnabled
      */
     public function __construct(
         DefaultSnippetManagerInterface $defaultSnippetManager,
         SnippetResolverInterface $snippetResolver,
         ReferenceStoreInterface $referenceStore,
+        ReferenceStoreInterface $snippetAreaReferenceStore,
         $defaultEnabled
```
alexander-schranz: This looks like a bc break think we require here some bc handling.

```suggestion
        $defaultEnabled,
        ?ReferenceStoreInterface $snippetAreaReferenceStore = null,
```

src/Sulu/Bundle/SnippetBundle/Content/SingleSnippetSelection.php
```diff
@@ -37,14 +37,21 @@ class SingleSnippetSelection extends SimpleContentType implements PreResolvableC
...
+     * @var ReferenceStoreInterface
+     */
+    private $snippetAreaReferenceStore;
+
     public function __construct(
         SnippetResolverInterface $snippetResolver,
         DefaultSnippetManagerInterface $defaultSnippetManager,
-        ReferenceStoreInterface $snippetReferenceStore
+        ReferenceStoreInterface $snippetReferenceStore,
+        ReferenceStoreInterface $snippetAreaReferenceStore
```
alexander-schranz: We require here some bc handling:

```suggestion
        ReferenceStoreInterface $snippetReferenceStore,
        ?ReferenceStoreInterface $snippetAreaReferenceStore = null
```

src/Sulu/Bundle/SnippetBundle/Content/SingleSnippetSelection.php
```diff
@@ -37,14 +37,29 @@ class SingleSnippetSelection extends SimpleContentType implements PreResolvableC
      */
     private $snippetReferenceStore;
 
+    /**
+     * @var ReferenceStoreInterface
```
alexander-schranz:
```suggestion
     * @var ReferenceStoreInterface|null
```

src/Sulu/Bundle/SnippetBundle/Content/SnippetContent.php
```diff
@@ -43,24 +43,39 @@ class SnippetContent extends ComplexContentType implements ContentTypeExportInte
      */
     private $referenceStore;
 
+    /**
+     * @var ReferenceStoreInterface
```
alexander-schranz:
```suggestion
     * @var ReferenceStoreInterface|null
```

alexander-schranz (changes requested):
Should fix the phpstan issues

PR #7178, opened 2023-10-17
Add type improvements to ListBuilder

mamazu:
More types are better types.


src/Sulu/Bundle/CategoryBundle/Controller/CategoryController.php
```diff
@@ -269,8 +274,11 @@ protected function getListRepresentation(
...
         foreach ($idsToExpand as $idToExpand) {
+            $fieldDescriptor = $listBuilder->getFieldDescriptor('parent');
+            Assert::notNull($fieldDescriptor, 'There is no field descriptor for field "parent"');
+
             $parentExpressions[] = $listBuilder->createWhereExpression(
-                $listBuilder->getFieldDescriptor('parent'),
+                $fieldDescriptor,
                 $idToExpand,
                 ListBuilderInterface::WHERE_COMPARATOR_EQUAL
             );
```
mamazu: No change in behavior, only making the error message more helpful. Error message before: Expected FieldDescriptorInterface got null

src/Sulu/Bundle/CategoryBundle/Controller/CategoryController.php
```diff
@@ -342,7 +350,7 @@ protected function getListRepresentation(
         );
     }
 
-    private function initializeListBuilder($locale, bool $defaultSort = false)
+    private function initializeListBuilder($locale, bool $defaultSort = false): ListBuilderInterface
```
mamazu, line 352: Not a bc break since this is private, even if the factory would return a non doctrine list builder.

src/Sulu/Component/Rest/ListBuilder/AbstractListBuilder.php
```diff
@@ -62,7 +62,7 @@ abstract class AbstractListBuilder implements ListBuilderInterface
     /**
      * The limit for this query.
      *
-     * @var int
+     * @var int|null
```
mamazu, line 65: Limit null means no limit. We handle it in the `DoctrineListBuilder` for all other use cases, I don't know. But since it's explicitly initialized with `null`, it should be handled.

src/Sulu/Bundle/CategoryBundle/Controller/CategoryController.php
```diff
@@ -269,8 +274,11 @@ protected function getListRepresentation(
         // generate expressions for collected parent-categories
         $parentExpressions = [];
         foreach ($idsToExpand as $idToExpand) {
+            $fieldDescriptor = $listBuilder->getFieldDescriptor('parent');
+            Assert::notNull($fieldDescriptor, 'There is no field descriptor for field "parent"');
+
```
alexander-schranz: Looks like we could do this check just once outside the `foreach` loop.

I'm also not sure about the `Assert::notNull` here 🤔 not a fan of runtime check for cases which never really can appear (parent is hardcoded fieldDescriptor). Sure best would be `getFieldDescriptor(): FieldDescriptor;` and `findFieldDescriptor(): ?FieldDescriptor;` but at current state we can not change that behaviour would more go with:

```suggestion
        /** @var FieldDescriptorInterface $parentFieldDescriptor */
        $parentFieldDescriptor = $listBuilder->getFieldDescriptor('parent');
        foreach ($idsToExpand as $idToExpand) {
```

or `assert($parentFieldDescriptor instanceof FieldDescriptorInterface);` which is optimized by the compiler to run only on dev / test envs.

mamazu: Yeah this would totally make sense to have a version of `getFieldDescriptor` which doesn't return a nullable result.

alexander-schranz (approved):

PR #7181, opened 2023-10-19
Types for security system

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | yes (kinda)
| Fixed tickets | -
| Related issues/PRs | partial #7052
| License | MIT
| Documentation PR | -

#### What's in this PR?
As Discussed [here](https://github.com/sulu/sulu/pull/7052#discussion_r1365467776) it makes sense to add typing to the rather complex structure of permission systems Sulu has.

#### Why?
I extracted out an interface to make it easier to test (interfaces can be mocked individually) and to have a central place where the type definition is located.


src/Sulu/Bundle/AdminBundle/Admin/SecurityContextAwareInterface.php
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+namespace Sulu\Bundle\AdminBundle\Admin;
+
+/**
+ * System (string) -> Section (string) -> Context (string) -> Permissions (list of strings).
+ *
+ * @phpstan-type SecurityContextInfo array<string, array<string, array<string, string[]>>>
+ */
+interface SecurityContextAwareInterface
```
alexander-schranz: We don't require a `phpstan-type` here as all our admin classes extends from `Admin` and so a type is not necessary here. We just need define it correctly in the `Admin` class.

src/Sulu/Bundle/AdminBundle/Admin/SecurityContextAwareInterface.php
```diff
@@ -0,0 +1,36 @@
...
+{
+    /**
+     * Returns all the security contexts, which are available in the concrete bundle.
+     *
+     * @return SecurityContextInfo
```
alexander-schranz: as written above this type is not required

src/Sulu/Bundle/AdminBundle/Admin/SecurityContextAwareInterface.php
```diff
@@ -0,0 +1,36 @@
...
+     */
+    public function getSecurityContexts();
+
+    /**
+     * Returns all the security contexts, which are available in the concrete bundle.
+     *
+     * @return SecurityContextInfo
```
alexander-schranz: as written above this type is not required

src/Sulu/Bundle/ActivityBundle/Infrastructure/Sulu/Admin/ActivityAdmin.php
```diff
@@ -22,6 +22,7 @@
 
 class ActivityAdmin extends Admin
 {
+    /** @var string */
```
alexander-schranz: This is not required PHPStan knows by the value that it is a string because it knows always the type of a constant: https://phpstan.org/r/ca9ee73c-fe6b-4e7d-b3cd-15f26ad26504

src/Sulu/Bundle/CategoryBundle/Admin/CategoryAdmin.php
```diff
@@ -24,6 +24,7 @@
 
 class CategoryAdmin extends Admin
 {
+    /** @var string */
```
alexander-schranz: This is not required PHPStan can analyse cosntants

alexander-schranz (changes requested):

src/Sulu/Bundle/AdminBundle/Admin/Admin.php
```diff
@@ -19,9 +19,12 @@
 /**
  * Defines all the required information from a bundle's admin class.
  */
-abstract class Admin implements ViewProviderInterface, NavigationProviderInterface
+abstract class Admin implements ViewProviderInterface, NavigationProviderInterface, SecurityContextAwareInterface
```
alexander-schranz: I currently would not add a new interface here we should just tackle the PHPDocs

src/Sulu/Bundle/AdminBundle/Admin/Admin.php
```diff
@@ -37,21 +40,11 @@ public function configureNavigationItems(NavigationItemCollection $navigationIte
     {
     }
 
-    /**
-     * Returns all the security contexts, which are available in the concrete bundle.
-     *
-     * @return mixed[]
```
alexander-schranz, line 43: we only need define the docs here

src/Sulu/Bundle/AdminBundle/Admin/Admin.php
```diff
@@ -37,21 +40,11 @@ public function configureNavigationItems(NavigationItemCollection $navigationIte
...
-     */
     public function getSecurityContexts()
     {
         return [];
     }
 
-    /**
-     * Returns all the security contexts, which are available in the concrete bundle.
-     *
-     * @return array
```
alexander-schranz, line 53: and here all other classes extends and so have the correct type  no need for a phpstan-type, we should keep this simple.

alexander-schranz (changes requested):

src/Sulu/Bundle/AdminBundle/Admin/SecurityContextAwareInterface.php
mamazu: I used the phpstan type because the functions `getSecurityContext` and `getSecurityContextWithPlaceholders` have the same complex array type and they should have the same structure. Otherwise we would have to define this array shape twice.

src/Sulu/Bundle/ActivityBundle/Infrastructure/Sulu/Admin/ActivityAdmin.php
mamazu: Well kinda, when using it with `self::PROPERTY` then phpstan can figure it out easily. But when using `static::PROPERTY` which we're using in the definition it struggles and thinks the type of the property is `int|string` and explicitly typing it here fixes that.

src/Sulu/Bundle/AdminBundle/Admin/Admin.php
mamazu: That's also fine. The AdminPool also needs those things should I just duplicate the doc comments to there as well?

alexander-schranz: Yeah fine in this case to duplicate.

mamazu:
Okay, I moved it to the `Admin` class and the `AdminPool` got the same typehints via copy and paste now.

src/Sulu/Bundle/PageBundle/Admin/PageAdmin.php
```diff
@@ -409,6 +410,7 @@ public function configureViews(ViewCollection $viewCollection): void
 
     public function getSecurityContexts()
     {
+        /** @var array<string, array<string, array<string, array<string>>>> */
```
alexander-schranz: What error is shown when we are not setting this? Normally I think phpstan should understand the code so we don't need this.

src/Sulu/Bundle/TagBundle/Admin/TagAdmin.php
```diff
@@ -23,6 +23,7 @@
 
 class TagAdmin extends Admin
 {
+    /** @var string */
```
alexander-schranz: phpstan should understand the const and not error, so we should be possible to avoid this const `@var` phpdocs annotations

src/Sulu/Bundle/PageBundle/Admin/PageAdmin.php
mamazu: Yeah, removed this was from an earlier attempt.

src/Sulu/Bundle/TagBundle/Admin/TagAdmin.php
mamazu: When using it with self::PROPERTY then phpstan can figure it out easily. But when using static::PROPERTY which we're using in the definition it struggles and thinks the type of the property is int|string and explicitly typing it here fixes that.

So we could use self::PROPERTY everywhere but that might cause a BC break.

alexander-schranz: There should in our classes `self::` or `SpecificAdmin::` used not `static::` can we change it to `self::` where this is done.

alexander-schranz: Also it would make sense to mark all `Admin` classes (MediaAdmin, ContactAdmin, ...) as final. As instead of extending just an own Admin instance should be created, when somebody overwrites:

```php
/**
 * @final
 */
```

But that should be an own Pull Request. With using the PHPDoc it would not be a bc break.

mamazu: Yeah making the Admin classes final would be a great idea. But I will make a separate pull request.

alexander-schranz (approved):

mamazu:
@alexander-schranz The pipeline was canceled for some reason. Can you restart it?

alexander-schranz:
Looks like it got stuck, interesting Github only cancel it after 6 hours.

PR #7182, opened 2023-10-19
Removing unused constructor arguments

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | yes
| Deprecations? | yes 

#### What's in this PR?
More services that have private properties that are unused

#### Why?
Unused properties make the application slightly slower.


mamazu:
Generally there are many more service arguments that could be removed. But I only did a few to not increase the size of the PR too much.

Furthermore the structure of Sulu's extension policy prevents removing any service arguments from services who's class can be overwritten. For example the `AdminController`. In those cases the only way to mitigate that would be to split up the service into multiple pieces (also outside of the scope of this PR).

mamazu:
Closed as this will be part of a bigger refactoring when moving to a better architecture.

PR #7183, opened 2023-10-19
More types for medias

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | in the tests but that's okay
| Deprecations? | no

#### What's in this PR?
Improving the types of the tests a little.

#### Why?
More types = Less errors.

### DIFF
--- a/src/Sulu/Bundle/CategoryBundle/Twig/CategoryTwigExtension.php
+++ b/src/Sulu/Bundle/CategoryBundle/Twig/CategoryTwigExtension.php
@@ -74,7 +74,7 @@ class CategoryTwigExtension extends AbstractExtension
      * If parentKey is set, only the children of the category which is assigned to the given key are returned.
      *
      * @param string $locale
-     * @param string $parentKey key of parent category
+     * @param ?string $parentKey key of parent category
      *
      * @return array
      */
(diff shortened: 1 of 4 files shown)


src/Sulu/Bundle/CategoryBundle/Twig/CategoryTwigExtension.php
```diff
@@ -74,7 +74,7 @@ public function getFunctions()
      * If parentKey is set, only the children of the category which is assigned to the given key are returned.
      *
      * @param string $locale
-     * @param string $parentKey key of parent category
+     * @param int|string $parentKey key of parent category
```
alexander-schranz: this looks unexpected as a key should be a string. also findChilrenByParentKey accept string|null not int. only findChildrenByParentId accept int but that is not used here.

alexander-schranz (changes requested):

src/Sulu/Bundle/CategoryBundle/Twig/CategoryTwigExtension.php
mamazu: In the tests the "key" is the id of the entity. But then I'll just fix the tests.

alexander-schranz: yeah would adopt the tests then that they use a proper string key instead of int.

alexander-schranz (approved):

PR #7185, opened 2023-10-19
404 on missing portal

### LINKED ISSUE #7045: No portal for this url found throws error 500
| Q | A
| --- | ---
| Sulu Version | 2.3.4
| PHP Version | 7.4
| DB Version | MySql 5.7

# Actual Behavior

Following webspace configuration is given:

<environment type="prod">
        <urls>
                <url language="de">www.mywebsite.com/ch-{localization}</url>
                <url language="fr">www.mywebsite.com/ch-{localization}</url>
                <url language="it">www.mywebsite.com/ch-{localization}</url>
        </urls>
</environment>

So if i call `www.mywebsite.com/test` it does not match any portal. But no portal found for this url Error 500 could we get an 404 error instead?

# Expected Behavior

Return a 404 error.

# Steps to Reproduce

Use the configuration above and call `www.mywebsite.com/test`

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #7045
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?
When the user calls a url where no portal matches, we don't return a 500 anymore.

#### Why?
The 500 implies that the server couldn't process it properly, which is incorrect.

### DIFF
--- a/src/Sulu/Component/Webspace/Analyzer/Attributes/WebsiteRequestProcessor.php
+++ b/src/Sulu/Component/Webspace/Analyzer/Attributes/WebsiteRequestProcessor.php
@@ -92,10 +92,7 @@ class WebsiteRequestProcessor implements RequestProcessorInterface
                 . (!\in_array($attributes->getAttribute('port'), ['80', '443'], true) ? ':' . $attributes->getAttribute('port') : '')
                     . $attributes->getAttribute('path');
 
-            throw new UrlMatchNotFoundException(
-                $fullUrl,
-                $portalUrls
-            );
+            throw new UrlMatchNotFoundException($fullUrl, $portalUrls);
         }
 
         return true;
--- a/src/Sulu/Component/Webspace/Analyzer/Exception/UrlMatchNotFoundException.php
+++ b/src/Sulu/Component/Webspace/Analyzer/Exception/UrlMatchNotFoundException.php
@@ -11,22 +11,18 @@
 
 namespace Sulu\Component\Webspace\Analyzer\Exception;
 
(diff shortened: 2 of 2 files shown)


mamazu:
Maybe that's not a good idea as this exposes the server configuration as the error message to the user.

alexander-schranz:
Yeah the initial case why the UrlMatchNotFoundException was added is to hint directly if there is an issue with the webspace / VHost setup, convert it to a HttpException would make it useless by itself.

I personally conflicted, because the validation of it sometimes really hurt when connect Sulu with existing Symfony application which is expected to run under an own Domain and then Sulu behaves differently then a normal Symfony application and most devs don't know how to disable the request analyzer for specific endpoint. So I'm not sure if we even should remove in 3.0 the validation 🤔

mamazu:
I think there might be a middle ground as for this. We could add logging instead. Log an error or a critical and return a 404. Then the developer can still look into the problem if they want but the user sees something sensible like a 404.

But yeah, the `RequestAnalyzer` is certainly a unique concept for Symfony devs. But I don't think that there is an easier way to handle multi site setups that is more Symfony like.

alexander-schranz:
Middle ground would not solve the problem which the exception did try to solve as nobody never really checks the application logs in this state. Most devs just don't like to look at logs when there is not a real error in the front.

mamazu:
Yeah, fair enough. Well then the only solution is to crash ugly and if the application devs want a better error message in Symfony 6 you can just map the exception to a 404 in the config.

PR #7186, opened 2023-10-20
Fixed deprecation happening when content type is checked in AppendAnalyticsListener

stollr:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

Setting default value in AppendAnalyticsListener when checking content type.

#### Why?

This is required, because the content type can be null (for example for responses with status code `204`), which results in the following deprecation message:

> strpos(): Passing null to parameter #1 () of type string is deprecated

in PHP 8.2

#### Example Usage

Add the following return statement to a controller action to reproduce the issue:

```php
return $this->json(null, Response::HTTP_NO_CONTENT);
```

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -34801,11 +34801,6 @@ parameters:
 			path: src/Sulu/Bundle/WebsiteBundle/EventListener/AppendAnalyticsListener.php
 
 		-
-			message: "#^Parameter \\#1 \\$haystack of function strpos expects string, string\\|null given\\.$#"
-			count: 1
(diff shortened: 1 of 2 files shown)


alexander-schranz:
@stollr nice catch, can you rebase it to the 2.4 as there the deprecation should also be removed.

```
git rebase -i origin/2.6 --onto origin/2.4
```

stollr:
@alexander-schranz done ;-)

alexander-schranz:
@stollr Thank you!

PR #7188, opened 2023-10-20
Add support for translation keys in snippet areas meta

### LINKED ISSUE #7120: Snippet areas - not possible to use translation keys
| Q | A
| --- | ---
| Sulu Version | 2.5.9
| PHP Version | 8.2

# Actual Behavior

When trying to use Symfony translation keys in the XML snippet, specifically for area meta titles, they are not being translated.

# Expected Behavior

The Symfony translation keys should be correctly translated when used in the XML snippet for area meta titles.

# Steps to Reproduce

The issue can be reproduced by using a Symfony translation key in the XML snippet for an area meta title. Here is an example:

```xml
    <areas>
        <area key="top">
            <meta>
                <title>some.translation.key</title>
            </meta>
        </area>
    </areas>
```

The problem seems to be in the `getArea` function in [SnippetAreaCompilerPass.php](https://github.com/sulu/sulu/blob/2.5/src/Sulu/Bundle/SnippetBundle/DependencyInjection/Compiler/SnippetAreaCompilerPass.php). This issue is similar to the one addressed in [PR #6682](https://github.com/sulu/sulu/pull/6682).

# Possible Solutions

I can offer to open a PR to address this issue, but we need to decide where to put the translation logic.

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | fixes #7120 
| License | MIT
| Documentation PR | -

#### What's in this PR?
Adding the option to use a translation key in the snippet area configuration.

#### Why?
Separating the content configuration from the translation to make it easier for external tools to help with translations. 

#### Example Usage
```xml
 <areas>
        <area key="top">
            <meta>
                <title>some.translation.key</title>
            </meta>
        </area>
    </areas>
```

### DIFF
--- a/src/Sulu/Bundle/SnippetBundle/DependencyInjection/Compiler/SnippetAreaCompilerPass.php
+++ b/src/Sulu/Bundle/SnippetBundle/DependencyInjection/Compiler/SnippetAreaCompilerPass.php
@@ -76,7 +76,7 @@ class SnippetAreaCompilerPass implements CompilerPassInterface
     }
 
     /**
-     * @param array{key: string, title: array<string>} $area
+     * @param array{key: string, title: array<string>, cache-invalidation: string} $area
      * @param array<string> $locales
      * @param array<string, string> $templateTitles
      *
--- a/src/Sulu/Bundle/SnippetBundle/DependencyInjection/Compiler/SnippetAreaCompilerPass.php
+++ b/src/Sulu/Bundle/SnippetBundle/DependencyInjection/Compiler/SnippetAreaCompilerPass.php
@@ -21,6 +21,9 @@ use Symfony\Component\DependencyInjection\ContainerBuilder;
  */
 class SnippetAreaCompilerPass implements CompilerPassInterface
 {
+    /**
+     * @return void
+     */
     public function process(ContainerBuilder $container)
     {
         $structureFactory = $container->get('sulu_page.structure.factory');
@@ -48,7 +51,7 @@ class SnippetAreaCompilerPass implements CompilerPassInterface
             ];
 
             foreach ($structure->getAreas() as $area) {
-                $area = $this->getArea($template, $area, $locales, $templateTitles);
+                $area = $this->getArea($container, (string) $template, $area, $locales, $templateTitles);
 
                 if (isset($areas[$area['key']])) {
                     throw new \InvalidArgumentException(\sprintf(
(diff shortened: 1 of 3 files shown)


mamazu:
@alexander-schranz That should be a quick review as well.

TheCadien:
looks good to me. 

But I would possibly write one more test. 
A test with add a value that should actually come from the transformation. But there is no translation for this.

TheCadien (approved):
Looks good ! :)

src/Sulu/Bundle/SnippetBundle/DependencyInjection/Compiler/SnippetAreaCompilerPass.php
```diff
@@ -72,23 +75,36 @@ public function process(ContainerBuilder $container)
...
-                $title = $area['title'][$locale];
+        // If we only have one title, it is probably a translation key
+        if (1 === \count($areaTitles)) {
+            $translator = $container->get('translator');
+            $titleToTranslate = \reset($areaTitles);
+            foreach ($locales as $locale) {
+                $titles[$locale] = $translator->trans($titleToTranslate, [], 'admin', $locale);
             }
+        } else {
+            \trigger_error('Translating snippet areas in the XML is deprecated use a translation instead.', \E_USER_DEPRECATED);
```
alexander-schranz: this will not be deprecated by sulu. It is more common that translations for the admin are done in the xml not in translations files.

src/Sulu/Bundle/SnippetBundle/DependencyInjection/Compiler/SnippetAreaCompilerPass.php
```diff
@@ -72,23 +75,36 @@ public function process(ContainerBuilder $container)
...
 
         $titles = [];
+        $areaTitles = $area['title'];
 
-        foreach ($locales as $locale) {
-            $title = $templateTitles[$locale] . ' ' . \ucfirst($key);
-            if (isset($area['title'][$locale])) {
-                $title = $area['title'][$locale];
+        // If we only have one title, it is probably a translation key
+        if (1 === \count($areaTitles)) {
```
alexander-schranz: this looks strange. It should get translated if there is no locale defined for it e.g.:

```xml
<meta>
   <title>app.title</title>
</meta>
```

but following should not be translated:

```xml
<meta>
</meta>
```

How is this handled for the other xml templates like snippets, pages, articles, ..?

alexander-schranz (changes requested):

src/Sulu/Bundle/SnippetBundle/DependencyInjection/Compiler/SnippetAreaCompilerPass.php
mamazu: Removed the deprecation message.

src/Sulu/Bundle/SnippetBundle/DependencyInjection/Compiler/SnippetAreaCompilerPass.php
mamazu: I was thinking about the same thing. [This PR added translatable metadata to the content structure.](https://github.com/sulu/sulu/pull/6682/files) and in there it's using the translated version if the tag has a locale attribute. So I updated the implementation to do the same.

mamazu:
@alexander-schranz Can you have a look at this again. It should be done.

src/Sulu/Bundle/SnippetBundle/DependencyInjection/Compiler/SnippetAreaCompilerPass.php
```diff
@@ -72,23 +75,35 @@ public function process(ContainerBuilder $container)
...
         $titles = [];
+        $areaTitles = $area['title'];
 
-        foreach ($locales as $locale) {
-            $title = $templateTitles[$locale] . ' ' . \ucfirst($key);
-            if (isset($area['title'][$locale])) {
-                $title = $area['title'][$locale];
+        // If we only have one title and no locale (indexed 0) then it's a translation key
+        if (1 === \count($areaTitles) && \array_key_exists(0, $areaTitles)) {
+            $translator = $container->get('translator');
```
alexander-schranz, line 95: @sulu/contributors normally because of how symfony recommends we avoid accessing services in compilerpasses. Does somebody have any concerns about this.

mamazu: I don't see any other way to do this. We could add a check if the container if the translator actually exists, but that's it.

alexander-schranz (approved):

PR #7194, opened 2023-11-01
Replace EntityManager::merge with EntityManager::persist

### LINKED ISSUE #7184: Replace EntityManager::merge with EntityManager::persist in Doctrine DataFixtures
# Problem description

The EntityManagerInterface::merge is deprecated and we should instead of `merge` use `persist`.

Following fixtures should be updated:

 - [x] https://github.com/sulu/sulu/blob/2.4/src/Sulu/Bundle/MediaBundle/DataFixtures/ORM/LoadMediaTypes.php
 - [x] https://github.com/sulu/sulu/blob/2.4/src/Sulu/Bundle/ContactBundle/DataFixtures/ORM/LoadDefaultTypes.php  
 - [x] https://github.com/sulu/sulu/blob/2.5/src/Sulu/Bundle/SecurityBundle/DataFixtures/ORM/LoadSecurityTypes.php
 - [x] https://github.com/sulu/sulu/blob/2.5/src/Sulu/Bundle/MediaBundle/DataFixtures/ORM/LoadCollectionTypes.php

This change can also target the 2.4 branch.

# Proposed solution

## Why was `merge` over `persist` used

Fixtures should only be loaded once so if the MediaType image already exists it should not created again. Also we are forcing the ids as there are some logic based on them, without merge currently a unique key constraint error would appear.

Like merge so we need first check if the entity already exists, if it does we just do nothing.

## How to use persist

The DataFixtures gives us a ObjectManager not a EntityManager so we only have `find` available but we can use that to check if the fixture was already loaded:

```diff
+        if (!$manager->find(MediaType::class, 1)) {
             $mediaDocument = new MediaType();
             $mediaDocument->setId(1);
-            $mediaDocument = $manager->merge($mediaDocument);
+            $mediaDocument = $manager->persist($mediaDocument);
             $mediaDocument->setName('document');
+        }
```

Testing this running `bin/console sulu:build dev` multiple times should not error.

## How to get started (for beginners)

1. Clone the Sulu Repository

```bash
git clone git@github.com:sulu/sulu.git
```

2. Checkout the 2.6 branch and create a new branch:

```bash
cd sulu
git checkout 2.6
git checkout -b feature/replace-merge-fixtures
```

3. Start the database

Configure the used database in the `.env.local` file. Supported are MySQL or PostgreSQL a docker-compose.yml is available in:  https://github.com/sulu/skeleton

```env
DATABASE_URL="mysql://root:ChangeMe@127.0.0.1:3306/su_myapp?serverVersion=8.0.27&charset=utf8mb4"
```

```bash
docker compose up
```

4. Implement the issue

See above (How to use persist)

5. Test changes

For this specific issue `bin/console sulu:build dev` is the easist way to test if the change works. Functional tests also use that fixtures.

6. Fork sulu/sulu repository

```bash
git remote add your git@github.com:<your_github_username>/sulu.git
```

7. Create pull request

```bash
git add -A
git commit -m "Replace ..."
git push your <your_branch>
```

shochdoerfer:
| Q | A
| --- | ---
| Fixed tickets | fixes #7184
| License | MIT

#### What's in this PR?

EntityManagerInterface::merge() is deprecated, replace the calls with EntityManagerInterface::persist()

### DIFF
--- a/src/Sulu/Bundle/ContactBundle/DataFixtures/ORM/LoadDefaultTypes.php
+++ b/src/Sulu/Bundle/ContactBundle/DataFixtures/ORM/LoadDefaultTypes.php
@@ -34,17 +34,17 @@ class LoadDefaultTypes extends AbstractFixture implements OrderedFixtureInterfac
 
         $phoneType1 = new PhoneType();
         $phoneType1->setId(1);
-        $phoneType1 = $manager->merge($phoneType1);
+        $manager->persist($phoneType1);
         $phoneType1->setName('sulu_contact.work');
 
         $phoneType2 = new PhoneType();
         $phoneType2->setId(2);
-        $phoneType2 = $manager->merge($phoneType2);
+        $manager->persist($phoneType2);
         $phoneType2->setName('sulu_contact.private');
 
         $phoneType3 = new PhoneType();
         $phoneType3->setId(3);
-        $phoneType3 = $manager->merge($phoneType3);
+        $manager->persist($phoneType3);
         $phoneType3->setName('sulu_contact.mobile');
 
         // Email types.
@@ -54,14 +54,14 @@ class LoadDefaultTypes extends AbstractFixture implements OrderedFixtureInterfac
 
         $emailType1 = new EmailType();
         $emailType1->setId(1);
-        $emailType1 = $manager->merge($emailType1);
+        $manager->persist($emailType1);
         $emailType1->setName('sulu_contact.work');
 
         $this->addReference('email.type.work', $emailType1);
 
         $emailType2 = new EmailType();
         $emailType2->setId(2);
-        $emailType2 = $manager->merge($emailType2);
+        $manager->persist($emailType2);
         $emailType2->setName('sulu_contact.private');
 
         $this->addReference('email.type.private', $emailType2);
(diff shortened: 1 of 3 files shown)


src/Sulu/Bundle/ContactBundle/DataFixtures/ORM/LoadDefaultTypes.php
```diff
@@ -34,17 +34,17 @@ public function load(ObjectManager $manager)
 
         $phoneType1 = new PhoneType();
         $phoneType1->setId(1);
-        $phoneType1 = $manager->merge($phoneType1);
+        $manager->persist($phoneType1);
```
alexander-schranz: A check need to be added to check if the fixture already exists. https://github.com/sulu/sulu/issues/7184#issue-1952697265.

Else it will fail if when `sulu:build dev` is run multiple times which should still be allowed.

shochdoerfer: Because of the hardcoded id, you mean?

alexander-schranz: Yes and that the fixtures not get duplicated.

shochdoerfer: So the logic would need to be changed like this?
```php
$phoneType1 = $manager->find(PhoneType::class, 1);
if ($phoneType1 === null) {
     $phoneType1 = new PhoneType();
     $phoneType1->setId(1);
     $manager->persist($phoneType1);
}
$phoneType1->setName('sulu_contact.work');
```

alexander-schranz: all can be inside the if statement.

shochdoerfer:
If anything else needs to be done, let me know.

shochdoerfer:
I ran `bin/console sulu:build dev` 5 times in a row and got not errors, so I would assume everything is working as expected.

alexander-schranz:
@shochdoerfer Thank you!

PR #7196, opened 2023-11-05
Validate Texteditor content only for string type

### LINKED ISSUE #7140: [MarkupBundle] preg_match_all(): Argument #2 ($subject) must be of type string, array given
| Q | A
| --- | ---
| Sulu Version | 2.5.8
| PHP Version | PHP 8.1

# Actual Behavior

An admin user tries to delete a page but the deletion process fails with the following error in the logs:
```
[2023-08-17T21:24:44.183109+02:00] php.CRITICAL: Uncaught Error: preg_match_all(): Argument #2 ($subject) must be of type string, array given {"exception":"[object] (TypeError(code: 0): preg_match_all(): Argument #2 ($subject) must be of type string, array given at /var/www/vendor/sulu/sulu/src/Sulu/Bundle/MarkupBundle/Markup/HtmlTagExtractor.php:40)"} []
```

# Expected Behavior

Page deletion should work.

# Steps to Reproduce

Since the deletion of other pages works fine, I assume something is odd with that specific page. So it's not a general bug in Sulu. Still, it should be fixed somehow to allow "defect" pages to be deleted.

# Possible Solutions

My current "quick fix" works as follows:

1. In `Sulu\Bundle\MarkupBundle\Markup\HtmlTagExtractor::count()` I added the following check:
```
if (!is_string($html)) {
    return 0;
}
```

2. In `Sulu\Bundle\MarkupBundle\Markup\HtmlTagExtractor::extract()` I added the following check:
```
if (!is_string($html)) {
    return [];
}
```

After applying the patch, the page deletion for the specific page is working again. If you feel the patch outlined above makes sense, let me know and I prepare a PR.

shochdoerfer:
| Q | A
| --- | ---
| Bug fix? | yes
| BC breaks? | yes (no exception is thrown anymore)
| Fixed tickets | fixes #7140
| License | MIT

#### What's in this PR?

See the discussion in #7140: Due to some template changes (and not having written proper migrations), a page could not be deleted because the MarkupBundle could not parse the content anymore as it was not of type string. Thanks to Alex' suggestion, this PR will only validate Texteditor content if the content is of type string.

#### Why?

Because uncaught errors are not nice :) And in our specific case, the uncaught error did not allow us to delete the page in question.

### DIFF
--- a/src/Sulu/Component/Content/Types/TextEditor.php
+++ b/src/Sulu/Component/Content/Types/TextEditor.php
@@ -45,7 +45,7 @@ class TextEditor extends SimpleContentType
     public function read(NodeInterface $node, PropertyInterface $property, $webspaceKey, $languageCode, $segmentKey)
     {
         $value = $node->getPropertyValueWithDefault($property->getName(), $this->defaultValue);
-        $property->setValue($this->validate($value, $languageCode));
+        $property->setValue(\is_string($value) ? $this->validate($value, $languageCode) : null);
 
         return $value;
     }


shochdoerfer:
Not sure how I could provide a test for this. Any suggestions?

alexander-schranz:
Was not able to push into this branch as it seems disabled. I added the missing test case here: https://github.com/sulu/sulu/pull/7216.

Thx for the fix @shochdoerfer

PR #7200, opened 2023-11-08
Add types to Symfony Configuration classes

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | (fixing some of them yes)
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

Adding return types to the configuration builder classes.

#### Why?

In Symfony 6 not having this type is deprecated and it can't hurt in 5.4 either. This is not a BC break as Return Type narrowing (the implementation of an interface can have a narrower return type than the interface implementing it) was introduced in 7.4 and not returning a `TreeBuilder` crashes the application anyways.

image

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -3346,16 +3346,6 @@ parameters:
 			path: src/Sulu/Bundle/AudienceTargetingBundle/Tests/Unit/EventListener/AudienceTargetingCacheListenerTest.php
 
 		-
-			message: "#^Offset 0 does not exist on array\\{\\}\\.$#"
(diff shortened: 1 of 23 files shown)


alexander-schranz:
Reminder for myself update `src/Sulu/Bundle/MarkupBundle/DependencyInjection/Configuration.php` after merging 2.4 -> 2.5 as 2.4 does not have that class.

alexander-schranz:
@mamazu thank you!

PR #7204, opened 2023-11-17
Add sorting for snippet area / default snippets by title instead of key

stollr:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | in a way, but only very low break
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?
Changed the sorting of the snippet areas. They are sorted by title instead of their key (which is not visible in the frontend).

#### Why?

Currently the snippet areas are sorted by their key. But this makes it harder to scan the list for the user because titles may have the wrong order.

#### Note

I am using PHP's `Collate` class for comparing the string. But this is only available if the intl extension is loaded. So I have added a fallback to simple `strcmp` function, because the intl extension is not a requirement by Sulu. 
But I am unsure how to handle this in the unit tests. Is it okay how it is done?

### DIFF
--- a/src/Sulu/Bundle/SnippetBundle/Controller/SnippetAreaController.php
+++ b/src/Sulu/Bundle/SnippetBundle/Controller/SnippetAreaController.php
@@ -123,8 +123,8 @@ class SnippetAreaController implements ClassResourceInterface
 
                 if ($uuid) {
                     $snippet = $this->documentManager->find($uuid, $this->getUser()->getLocale());
-                    $areaData['defaultUuid'] = $snippet ? $snippet->getUuid() : null;
-                    $areaData['defaultTitle'] = $snippet ? $snippet->getTitle() : null;
+                    $areaData['defaultUuid'] = null;
+                    $areaData['defaultTitle'] = null;
                 }
             }
 
@@ -175,8 +175,8 @@ class SnippetAreaController implements ClassResourceInterface
                 'key' => $key,
                 'template' => $area['template'],
                 'title' => $area['title'],
-                'defaultUuid' => $defaultSnippet ? $defaultSnippet->getUuid() : null,
-                'defaultTitle' => $defaultSnippet ? $defaultSnippet->getTitle() : null,
+                'defaultUuid' => $defaultSnippet->getUuid(),
+                'defaultTitle' => $defaultSnippet->getTitle(),
                 'valid' => true,
             ]
         );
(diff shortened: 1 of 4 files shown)


alexander-schranz:
I like the idea. What I think we need to test here if the template uses `translation-keys` if that is correctly handled already.

As Sulu will always support 2 ways of define a title:

Have every meta inlined this way:

```
    <meta>
        <title lang="en">Example</title>
        <title lang="de">Beispiel</title>
    </meta>
```

or using a translation key:

```
    <meta>
        <title>app.example</title>
    </meta>
```

I think the second one sadly is not everywhere correctly target in the snippet area. See https://github.com/sulu/sulu/pull/7188. Maybe @mamazu can have a look if this would break his pull request or if we fine to go with it.

mamazu:
Well, in the current version of Sulu using translations for the snippet areas is not supported. However if you merge my PR this PR would still work as the sorting happens in the controller and the translations are resolved in a compiler pass. So feel free to merge this.

src/Sulu/Component/Util/SortUtils.php
```diff
@@ -96,4 +96,49 @@ public static function multisort($values, $paths, $direction = 'ASC')
...
+        $array = \is_array($list) ? $list : \iterator_to_array($list);
+
+        if (null === $getComparableValue) {
+            $getComparableValue = fn ($item) => $item;
+        }
+
+        // Collator class requires intl extension
+        $collator = \class_exists(\Collator::class) ? new \Collator($locale) : null;
+
+        \usort($array, function(mixed $itemA, mixed $itemB) use ($collator, $getComparableValue) {
```
alexander-schranz: I think the util function should keep the index:

```suggestion
        \uasort($array, function(mixed $itemA, mixed $itemB) use ($collator, $getComparableValue) {
```

src/Sulu/Bundle/SnippetBundle/Controller/SnippetAreaController.php
```diff
@@ -130,11 +131,11 @@ public function cgetAction(Request $request)
             $dataList[$key] = $areaData;
         }
 
-        \ksort($dataList);
+        $dataList = SortUtils::sortLocaleAware($dataList, $this->getUser()->getLocale(), fn ($a) => $a['title']);
 
         $data = [
             '_embedded' => [
-                'areas' => \array_values($dataList),
+                'areas' => $dataList,
```
alexander-schranz: As the util function keep the index we should reintroduce the array_values part:

```suggestion
                'areas' => \array_values($dataList),
```

alexander-schranz (changes requested):

alexander-schranz (approved):

PR #7209, opened 2023-11-23
Fix parameter type of portal

mamazu:
| Q | A
| --- | ---
| Bug fix? | documentation
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?
The list of portals in the `WebspaceCollection` is not a Hash therefore using string as a typehint is wrong as no string will yield any results.

#### Why?
Better type annotations are always good.

#### Example Usage
```php
$this->webspaceCollection->getPortal(0); // works
$this->webspaceCollection->getPortal('somePortal'); // doesn't work
```

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -1046,7 +1046,7 @@ parameters:
 			path: src/Sulu/Bundle/AdminBundle/Entity/CollaborationRepository.php
 
 		-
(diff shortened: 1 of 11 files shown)


alexander-schranz (approved):

alexander-schranz:
@mamazu thank you!

PR #7211, opened 2023-11-25
Update mobx to 6.9 and mobx-react to 9.1

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Update mobx to 6.9 and mobx-react to 9.1

#### Why?

With the support for decorators again in mobx we could smoothly upgrade to latest version.

#### To Do

- [x] Replace `isArrayLike` with `isArray` and `isObservableArray` see https://github.com/mobxjs/mobx/blob/main/packages/mobx/CHANGELOG.md
- [ ] FlowJS Types (not longer provided by mobx 6)
- [ ] Use new instead of legacy decorators?

#### Update 2026-02

Bigger review is needed for:

- src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/tests/BlockCollection.test.js
- src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/ResourceFormStore.js
- src/Sulu/Bundle/AdminBundle/Resources/js/containers/Toolbar/tests/withToolbar.test.js
- src/Sulu/Bundle/AdminBundle/Resources/js/containers/WritingAssistant/tests/PromptInput.test.js
- src/Sulu/Bundle/AdminBundle/Resources/js/views/List/toolbarActions/UploadToolbarAction.js
- src/Sulu/Bundle/MediaBundle/Resources/js/containers/MediaCollection/CollectionSection.js
- src/Sulu/Bundle/MediaBundle/Resources/js/views/MediaOverview/tests/MediaOverview.test.js
- src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/Preview.js

Untested.


src/Sulu/Bundle/AdminBundle/Resources/js/services/Router/Router.js
```diff
@@ -132,6 +132,8 @@ export default class Router {
     redirectFlag: boolean = false;
 
     constructor(history: Object) {
+        makeObservable(this); // this should be enough that we still should be able to use legacy decorators
+
```
alexander-schranz: this should make the above variables observable, but the beneath test fails `expect(isObservable(router.route)).toBe(true)` as router.route is still not observable based on isObservable:

 
 Bildschirmfoto 2023-11-29 um 00 10 47

alexander-schranz: Looks like this error maybe related to some babel issues as I tried now define the observable via makeObservable directly and run into the following error:

Bildschirmfoto 2023-11-29 um 01 00 40

alexander-schranz: Strange is that it still output the route is defined:

but as expected it is undefined:

Bildschirmfoto 2023-11-29 um 01 19 15

mweststrate: there are some subtleties in how Babel defines its class fields. From top of my head in 'loose' mode, it only allocates the field on first assignment, so you might want to set this to 'strict' (or the other way around, it has been a while). Explicitly initializing it might also help, e.g. `route: ?Route = undefined`.

mweststrate: Also you probably want to check `expect(isObservableProp(router, "route"))`, otherwise it checks if the value stored in the route field is observable, not whether the property itself is.

alexander-schranz: Thx @mweststrate for your response. It really seems to be that in this case `isObservable` is failing but your suggested `expect(isObservableProp(router, "route"))` works. So in Mobx 6 (or 5) only the class itself knows if its observable property not the property itself?

It look like / I hope it is only used in tests and so I can maybe just do a replace via regex to convert existing `isObservable` to `isObservableProp`:

```
\(isObservable\((.*)\.(.*)\)\)\.
(isObservableProp($1, '$2')).
```

Bildschirmfoto 2023-11-30 um 10 24 06

Following changes seems to make the `Router.test.js` work: https://github.com/sulu/sulu/pull/7211/commits/b8cd4860eb2618dff5279da5abbf2ee98238d2d8. Seems I need to check this on a lot of other tests.


If props and params are never itself observable is the replacement of `isArrayLike` not required:

```diff
-if (!isArrayLike(values)) {
+if (!(Array.isArray(values) || isObservableArray(values))) {
```

and could it just be:

```diff
+if (!(Array.isArray(values)) {
```

mweststrate: I suspect your old tests where just testing the wrong thing, and that somehow the value got observable or not. Whether that should be the case probably differs per case. E.g. what you were testing previously was `const x = a.b; isObservable(x)`. Sometimes `x` was I guess supposed to be observable, sometimes maybe not (e.g. built in objects not managed by MobX etc.

mweststrate: `Array.isArray` should now work on MobX observable arrays indeed as well

PR #7212, opened 2023-11-26
Fix hide of default snippet tab when no permissions

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no 
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Fix hide of default snippet tab when no permissions.

#### Why?

Even having no permissions for the default snippets the tab is still visible and load nothing because of the controller not allowing it.

#### TODO

 - [x] Rebase to 2.4
 - [x] Adopt tests
 - [x] Check other permissions here (delete vs edit)
 - [x] When should the Tab be shown on VIEW or EDIT permissions?


src/Sulu/Bundle/SnippetBundle/Admin/SnippetAdmin.php
```diff
@@ -206,6 +209,20 @@ public function getSecurityContexts()
         return $contexts;
     }
 
+    private function hasSomeDefaultSnippetPermission(): bool
```
alexander-schranz, line 212: This is inspired by PageAdmin: https://github.com/sulu/sulu/blob/8191cc53629000fd4e3f25df7593d6ba189090f7/src/Sulu/Bundle/PageBundle/Admin/PageAdmin.php#L520

PR #7218, opened 2023-12-01
Using Symfony configuration for webspaces

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | kinda
| BC breaks? | yes
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | https://github.com/sulu/sulu-docs/pull/790

#### What's in this PR?
Creating a way to configure webspaces with Symfony configuration.

#### Why?
Currently Sulu has it's own XML parser for webspace configs. This is a lot of maintenance and hard to extend. Using the Symfony config solves both 

#### Example Usage
Just write your configs like before. But now you can also use YAML and PHP.

#### To Do

- [x] Make sublocales possible in the config (at most two)
- [x] Document this ^
- [x] Document how to extend the configuration
- [x] It should work with the same configuration as before
- [x] Making the `WebspaceCollection` read only

- [x] Checking all Exceptions have been turned into validation
    - [x]  ExpectedDefaultTemplatesNotFound.php (moved to WebspaceConfiguration)
    - [x]  InvalidAmountOfDefaultErrorTemplateException.php (still there)
    - [x]  InvalidCustomUrlException.php (moved to WebspaceConfiguration)
    - [x]  InvalidDefaultErrorTemplateException.php (moved to WebspaceConfiguration)
    - [x]  InvalidDefaultLocalizationException.php (moved to WebspaceConfiguration)
    - [x]  InvalidErrorTemplateException.php  (moved to WebspaceConfiguration)
    - [x]  InvalidPortalDefaultLocalizationException.php (moved to WebspaceConfiguration)
    - [x]  InvalidUrlDefinitionException.php (still exists) 
    - [x]  InvalidWebspaceDefaultLocalizationException.php (moved to WebspaceConfiguration)
    - [x]  InvalidWebspaceDefaultSegmentException.php (moved to WebspaceConfiguration)
    - [x]  PortalDefaultLocalizationNotFoundException.php (still exists)
    - [x]  WebspaceDefaultLocalizationNotFoundException.php (moved to WebspaceConfiguration)
    - [x]  WebspaceDefaultSegmentNotFoundException.php (moved to WebspaceConfiguration)
    - [x]  WebspaceException.php (base exception is still there)


src/Sulu/Bundle/CoreBundle/DependencyInjection/Configuration.php
```diff
@@ -60,18 +59,6 @@ private function getLocaleConfiguration(NodeBuilder $rootNode)
             ->scalarNode('fallback_locale')->defaultValue('en')->end();
     }
 
-    private function getWebspaceConfiguration(NodeBuilder $rootNode)
-    {
-        $rootNode->arrayNode('webspace')
-            ->addDefaultsIfNotSet()
-            ->children()
-                ->scalarNode('config_dir')
-                    ->defaultValue('%kernel.project_dir%/config/webspaces')
```
mamazu, line 69: Add this directory to the kernel maybe to also load webspaces from there?

config/packages/dev/webspaces.xml
```diff
@@ -0,0 +1,76 @@
+<services:container xmlns="http://example.org/schema/dic/sulu_website"
```
mamazu: This is only for testing purposes

src/Sulu/Component/Webspace/Manager/WebspaceManager.php
```diff
@@ -599,4 +503,9 @@ private function createResourceLocatorUrl($portalUrl, $resourceLocator, $scheme
 
         return $url;
     }
+
+    public function getWebspaceCollection(): WebspaceCollection
+    {
+        return $this->webspaceCollection;
+    }
```
mamazu: Remove, since it's deprecated in 2.6

tests/Resources/DataFixtures/Webspace/multiple/sulu.io.xml
```diff
@@ -36,7 +40,7 @@
     </navigation>
 
     <resource-locator>
-        <strategy>short</strategy>
```
mamazu, line 39: Is that a real strategy? If so, then I couldn't find any classes implementing the strategy.

src/Sulu/Bundle/CoreBundle/DependencyInjection/Configuration.php
mamazu: I don't think we can do that in Sulu code?

src/Sulu/Component/Webspace/Manager/PortalInformationBuilder.php
```diff
@@ -0,0 +1,222 @@
...
+        return $pattern;
+    }
+
+    private function validateUrlPartialMatch(string $urlResult, Environment $environment): bool
+    {
+        return
+            // only valid if there is no full match already
+            !\array_key_exists($urlResult, $this->portalInformations[$environment->getType()])
+            // check if last character is no dot
+            && '.' != \substr($urlResult, -1);
```
hacfi, line 208: alternatively you could use `!str_ends_with($urlResult, '.')`

mamazu:
The order of webspaces is now dependent on the order the files are loaded. So if the migration in the user project changes the order the files are loaded the `/api/webspaces` end point might return webspaces in a different order.

mamazu:
BC Breaks in the return value of the WebspaceController `cgetAction`:
In the localization json:
OLD:
```json
          "localizations": [
            {
              "country": "",
              "language": "en",
              "localization": "en",
!             "default": "1",
              "children": [],
!             "shadow": null
            },
            {
              "country": "",
              "language": "de",
              "localization": "de",
!             "default": "",
              "children": [],
!             "shadow": null
            }
          ],
```
NEW:
```json
          "localizations": [
            {
              "country": "",
              "language": "en",
              "localization": "en",
!             "default": true,
              "children": [],
!             "shadow": ""
            },
            {
              "country": "",
              "language": "de",
              "localization": "de",
!             "default": false,
              "children": [],
!             "shadow": ""
            }
          ],
```

* `default` is now a boolean and not a string that might be empty
* `shadow` is either a string or null. (This is already the behaviour in the localizations in the portal)

src/Sulu/Component/Webspace/Tests/Unit/Manager/WebspaceCollectionCache.php
```diff
@@ -0,0 +1,1118 @@
...
+use Sulu\Component\Webspace\Segment;
+use Sulu\Component\Webspace\Url;
+use Sulu\Component\Webspace\Webspace;
+
+/**
+ * WebspaceCollectionCache65db58eff00cf.
+ *
+ * DO NOT EDIT
+ * This file is autogenerated by the Sulu Webspace component
+ */
```
mamazu, line 31: This is a cache file from the last version to check that the new processes are compatible with the old behaviour.

src/Sulu/Bundle/SecurityBundle/System/SystemStoreInterface.php
```diff
@@ -17,7 +17,7 @@ interface SystemStoreInterface
 {
     public function getSystem(): ?string;
 
-    public function setSystem(string $system): void;
+    public function setSystem(?string $system): void;
```
mamazu, line 20: BC break

mamazu:
## Migration guide

1. Make sure that all webspace configuration files are under `config/webspaces/` directory. 
2. Select the starting `webspace`-tag in each file  
```xml
<webspace xmlns="http://schemas.sulu.io/webspace/webspace"
          xsi:schemaLocation="http://schemas.sulu.io/webspace/webspace http://schemas.sulu.io/webspace/webspace-1.1.xsd">
```
and replace it with this:
```xml
<services:container xmlns="http://example.org/schema/dic/sulu_website"
    xmlns:services="http://symfony.com/schema/dic/services"
    xsi:schemaLocation="http://symfony.com/schema/dic/services
        https://symfony.com/schema/dic/services/services-1.0.xsd"
    >
<config>
<webspace>
```
3. And at the end of each file after the closing webspace tag add those lines:
```xml 
</config>
</services:container>
```

alexander-schranz:
Not sure in which direction https://github.com/symfony/symfony/issues/60200 will go. If Symfony removes support for XML completely it would be bad way if we build ontop of symfony services for webspace and maybe templates.

mamazu:
I have to agree it's sad to see that Symfony is dropping xml for this usecase. However, I would still differentiate between configs. For templates and snippets I agree xml is still a good way to configure it and moving that to Symfony config is going to be tricky anyways. However for the webspace config I do think that it makes sense to move to Symfony configs because it's already quite a technical topic so I don't think xml for comfort is going to be that much of an issue.

And we're missing out on the following features:
- environment based config merging feature for the urls (makes no sense for templates)
- Symfony can show example config with comments (see `config:dump-reference`)

In the end this PR has to be reworked anyways but I still think that this is a good idea.

mamazu:
Symfony is removing the support for XML in bundle configurations and since Sulu wants to keep the XML config and I mean who wants to rebase this giant PR anyways. :D

PR #7220, opened 2023-12-02
Add more types to media controllers

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | partial #7052 
| License | MIT
| Documentation PR | -

#### What's in this PR?
Splitting up the Merge Request above and only add types to the Controllers.

#### Why?
For 3.0 we want to fix the types in the method signature. Better start with checking them earlier.

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/Controller/AbstractMediaController.php
+++ b/src/Sulu/Bundle/MediaBundle/Controller/AbstractMediaController.php
@@ -41,7 +41,7 @@ abstract class AbstractMediaController extends AbstractRestController
     /**
      * @param Request $request
      *
-     * @return string
+     * @return string|null
      */
     protected function getTitleFromUpload($request)
     {
@@ -54,15 +54,20 @@ abstract class AbstractMediaController extends AbstractRestController
 
             return \implode('.', \explode('.', $uploadedFile->getClientOriginalName(), -1));
         }
+
+        return null;
     }
 
     /**
      * @param string $name
      *
-     * @return UploadedFile
+     * @return UploadedFile|null
      */
     protected function getUploadedFile(Request $request, $name)
     {
-        return $request->files->get($name);
+        /** @var UploadedFile|null $file */
+        $file = $request->files->get($name);
+
+        return $file;
     }
 }
--- a/src/Sulu/Bundle/MediaBundle/Controller/CollectionController.php
+++ b/src/Sulu/Bundle/MediaBundle/Controller/CollectionController.php
@@ -40,7 +40,10 @@ use Symfony\Contracts\Translation\TranslatorInterface;
 /**
  * Makes collections available through a REST API.
  */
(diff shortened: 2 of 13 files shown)


src/Sulu/Bundle/MediaBundle/Controller/AbstractMediaController.php
```diff
@@ -54,14 +54,10 @@ protected function getTitleFromUpload($request)
 
             return \implode('.', \explode('.', $uploadedFile->getClientOriginalName(), -1));
         }
+        return null;
```
mamazu, line 58: Not a bc break. No return at end of function, it will return null. So this is making implicit behavior now explicit.

src/Sulu/Bundle/MediaBundle/Controller/CollectionController.php
```diff
@@ -154,11 +158,12 @@ public function getAction($id, Request $request)
 
         try {
             $locale = $this->getRequestParameter($request, 'locale', true);
-            $depth = \intval($request->get('depth', 0));
+            $depth = \intval($this->getRequestParameter($request, 'depth', false, 0));
```
mamazu, line 159: `getRequestParameter` does the same as `$request->get` just with a fake string return type. In my opinion we should move away from the `RequestParameterTrait` and just use the Symfony functions `->getInt` on the correct `ParameterBag`.

src/Sulu/Bundle/MediaBundle/Controller/MediaController.php
```diff
@@ -322,11 +322,10 @@ private function getListRepresentation(Request $request)
      * @deprecated
      *
      * @param FieldDescriptorInterface[] $fieldDescriptors
-     * @param array $types
      *
      * @return DoctrineListBuilder
      */
-    private function getListBuilder(Request $request, array $fieldDescriptors, $types)
+    private function getListBuilder(Request $request, array $fieldDescriptors, array $types)
```
mamazu, line 328: Would crash later in line 360 when using `$listbuilder->in` as the second parameter has array parameter type.

src/Sulu/Bundle/MediaBundle/Controller/CollectionController.php
```diff
@@ -40,7 +40,10 @@
 /**
  * Makes collections available through a REST API.
  */
-class CollectionController extends AbstractRestController implements ClassResourceInterface, SecuredControllerInterface, SecuredObjectControllerInterface
+class CollectionController extends AbstractRestController implements
+    ClassResourceInterface,
+    SecuredControllerInterface,
+    SecuredObjectControllerInterface
```
alexander-schranz:
```suggestion
class CollectionController extends AbstractRestController implements ClassResourceInterface, SecuredControllerInterface, SecuredObjectControllerInterface
```

Would keep this as it is as we currently not use this code style somewhere else.

alexander-schranz (approved):

PR #7222, opened 2023-12-03
Deprecating using the `WebspaceManger` for getting webspaces

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | yes 
| Fixed tickets | -
| Related issues/PRs | #7218 
| License | MIT
| Documentation PR | -

#### What's in this PR?
* Making the `WebspaceCollection` it's own service.
* Deprecating repository like functions on the `WebspaceManager`
* Adding constructor arguments to Webspace objects.

#### Why?
We want to split those huge services up. That way we reduce the amount of dependencies a class has.

#### Example Usage
OLD:
```php
class SomeClass {
    public function __construct(
        private WebspaceManager $webspaceManager
    ) {}

    public function doSomething(): ?string 
    {
        return $this->webspaceManager->findWebspaceByKey('someKey')?->getName();
    }
}
```

NEW:
```php
class SomeClass {
    public function __construct(
        private WebspaceCollection $webspaceCollection
    ) {}

    public function doSomething(): ?string 
    {
        return $this->webspaceCollection->getWebspace('someKey')?->getName();
    }
}
```

#### To Do


src/Sulu/Component/Webspace/Webspace.php
```diff
@@ -122,9 +108,21 @@ class Webspace implements ArrayableInterface
      */
     private $resourceLocatorStrategy;
 
+    /**
+     * @param string $key
+     * @param string $name
+     */
+    public function __construct(
+        private $key = null,
+        private $name = null,
```
mamazu, line 117: Technically passing null to those parameters will be also removed in 3.0.

mamazu:
@alexander-schranz Can you have a look at this and merge if possible?

mamazu:
This should probably be split up into smaller pieces.

PR #7225, opened 2023-12-04
Fix static code analyzer error on void methods on Sulu PHPCR Session and Sulu PHPCR Node classes

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no (removing the return doesn't change the result of the method as it still null)
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Fix static code analyzer error on void methods on Sulu PHPCR Session and Sulu PHPCR Node classes.

#### Why?

New PHPStan errors for return a decorated method which is void. As new typehints where added to jackalope and phpcr by @dbu.  And so new errors did appear in our baseline:

Bildschirmfoto 2023-12-04 um 16 03 51


src/Sulu/Bundle/DocumentManagerBundle/Session/Session.php
```diff
@@ -116,22 +116,22 @@ public function propertyExists($absPath)
 
     public function move($srcAbsPath, $destAbsPath)
     {
-        return $this->inner->move($srcAbsPath, $destAbsPath);
+        $this->inner->move($srcAbsPath, $destAbsPath);
```
alexander-schranz, line 119: Same static code analyzer errors mostly will happen in the PHPCR Shell here: https://github.com/phpcr/phpcr-shell/blob/c5e62c9cbc15e956f17b908ba4476770b992d3ca/src/PHPCR/Shell/Phpcr/PhpcrSession.php#L285

@dbu think we need to update that also there.

Prokyonn (approved):

src/Sulu/Component/Content/Tests/Unit/Document/Subscriber/PHPCR/SuluNodeTest.php
```diff
@@ -34,12 +34,12 @@ public function provideDelegateData()
             ['isNew', true],
             ['isModified', true],
             ['isSame', false, [$this->prophesize(NodeInterface::class)->reveal()]],
-            ['accept', true, [$this->prophesize(ItemVisitorInterface::class)->reveal()]],
-            ['revert', true],
+            ['accept', null, [$this->prophesize(ItemVisitorInterface::class)->reveal()]],
```
alexander-schranz, line 37: The unit tests were build on mocks and so we needed do change here the accepted result as the decoratedService result always is null.

PR #7226, opened 2023-12-04
Fix running UnitTests on a machine in a different timezone

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no 
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Fix running UnitTests in a different timezone.

#### Why?

Running tests in e.g. Europe/Berlin timezone will currently make some tests failing. So we define that the tests run in UTC like it already in the CI the case.

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -19,9 +19,9 @@
     </filter>
 
     <php>
-        <env name="SYMFONY_PHPUNIT_VERSION" value="8.0"/>
         <server name="APP_ENV" value="test" force="true"/>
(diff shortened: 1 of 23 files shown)


phpunit.xml.dist
```diff
@@ -19,9 +19,9 @@
     </filter>
 
     <php>
-        <env name="SYMFONY_PHPUNIT_VERSION" value="8.0"/>
```
alexander-schranz, line 22: This is unnecessary configuration in our case.

Prokyonn (approved):

PR #7228, opened 2023-12-06
Add pagination to admin search

### LINKED ISSUE #7109: search function in the admin backend only shows 10 results and has no pagination
| Q | A
| --- | ---
| Sulu Version |  2.5.9
| PHP Version |  8.2
| DB Version | 
| Browser Version |  Firefox 114.0.1 64-bit Linux

# Actual Behavior

The search function in the admin backend only shows a maximum of 10 results and no pagination is available to get further results.

# Expected Behavior

It should be possible to paginate through all results.

# Steps to Reproduce

Just enter a wildcard (*) in the search function for example in the demo installation under https://sulu.rocks/admin/#/

# Possible Solutions

zanturik:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no 
| BC breaks? | no 
| Deprecations? | no 
| Fixed tickets | fixes #7109
| Related issues/PRs | #7109
| License | MIT
| Documentation PR |

#### What's in this PR?

Add pagination to admin search page

#### Why?

Because good search pages contains pagination.

#### Example Usage

Search * in admin search page. Play with pagination.


alexander-schranz:
@zanturik Thx for working on this and sorry for the late response.

I did give this a quick try but it looks like this isn't working. Maybe related to the removed zend lucene part or fasl? But it even not triggers a request so maybe related more to the JS part.

search

The pagination component itself is maybe a little bit small on small screens. So maybe we should make that atleast the same width as the search field.

zanturik:
> I did give this a quick try but it looks like this isn't working. 

You are right, my bad, one line was not pushed. Fixed, Please test again.

> The pagination component itself is maybe a little bit small on small screens. So maybe we should make that atleast the same width as the search field.

I agree. The problem it that it's a Pagination component displaying results/pagination this way, I'm just reusing it. The problem itself comes from "max-width" of .search-result class. I think it should be fixed in a separate task, since it's not the only place where search is used and design in other places may differ.

alexander-schranz (changes requested):
I checked this with the Team and the only required change is that the `Pagination` component should go over the same width as the search field itself. Can you tackle that?

zanturik:
> I checked this with the Team and the only required change is that the `Pagination` component should go over the same width as the search field itself. Can you tackle that?

fixed, please review.

alexander-schranz (approved):

alexander-schranz:
@zanturik Looks good for me, Thanks you for the contribution! Auto merged when the CI finished.

PR #7229, opened 2023-12-09
Add global block types

### LINKED ISSUE #6381: Make block types an own form type (Reusable blocks)
# Problem description

When managing content it sometimes is the case that a content manager would like to reuse a block with the same content on another page, without having to duplicate it and so to change it on every page again. Currently the only way to achieve this is to tell the developer to create a new block type which uses reusable snippets. With this new global block reusable developer not longer need to ask if something need to be a snippet or not as the control about reusable things are given to content manager.

Another issue with the current implementation is that every block types is injected into every template currently using `xi:includes` this make our metadata a lot bigger as when we make the block types something more central so the frontend only need to load a block types metadata once.

# Proposed solution

Currently blocks types are not a global thing every template can define there own blocks. But from content manager perspective it would be great that a block type is something global.

So blocks can example be marked as `reusable` block and so can be embedded into another page and any change in the reusable block will be update the block in central storage. This would be in my opinion replace the current snippet feature.  

This new blocks could look like the following:

```xml
<!-- config/templates/pages/default.xml

<!-- ... -->
<block name="blocks" default-type="text-image" minOccurs="0">
    <meta>
    </meta>

    <params>
        <param name="types" type="collection">
            <param name="text-image"/>
            <param name="gallery"/>
            <param name="quote"/>
        </param>
    </params>
</block>

<!-- or property syntax -->
<property name="blocks" type="block">
    <meta>
    </meta>

    <params>
        <param name="default-type" value="text-image">
        <param name="minOccurs" value="0">
        <param name="types" type="collection">
            <param name="text-image"/>
            <param name="gallery"/>
            <param name="quote"/>
        </param>
    </params>
</block>

<!-- ... -->
```

```xml
<!-- config/templates/blocks/text-image.xml -->
<?xml version="1.0" ?>
<form xmlns="http://schemas.sulu.io/template/template"
    xsi:schemaLocation="http://schemas.sulu.io/template/template http://schemas.sulu.io/template/form-1.0.xsd"
>
    <key>text-image</key>

    <properties>
         <!-- ... -->
    </properties>
</form>
```

### Content vs Entity Blocks

Where from content perspective it always make sense that the root block types are a central blocks. There are still valid use cases where the `types` syntax is still valid. The first one is when a block is used inside an entity or a custom model. In this case the data is not saved by sulu and so this can not be reusable blocks or even you don't want to split there the block types into seperate files. So the `<types>` syntax should still be a valid option for data management. Another valid option where the `<types>` syntax still make sense is as example when used inside `nested` blocks as an example we have a block type called `services` which is a block which has against block into it. In this case a `service-item` block type would be an overhead from DX perspective and would make the services.xml block type not really understandable. So a something like this is still valid:

```xml
<!-- config/templates/blocks/text-image.xml -->
<?xml version="1.0" ?>
<form xmlns="http://schemas.sulu.io/template/template"
    xsi:schemaLocation="http://schemas.sulu.io/template/template http://schemas.sulu.io/template/form-1.0.xsd"
>
    <key>services</key>

    <properties>
         <property name="article" type="text_line">
              <meta>
                  <title lang="de">Artikel</title>
              </meta>
         </property>

        <block name="services" default-type="service-item" minOccurs="0">
            <meta>
            </meta>
        
            <params>
                <param name="add_button_text">
                    <meta>
                        <title lang="en">Add service</title>
                    </meta>
                </param>
            </params>
        
            <types>
                <type name="service-item">
                    <properties>
                        <property name="title" type="text_line" colspan="9">
                            <meta>
                                <title lang="de">Artikel</title>
                            </meta>
        
                            <tag name="sulu.block_preview" priority="1024"/>
                        </property>
                        
                        <property name="images" type="single_media_selection">
                            <meta>
                                <title lang="de">Bilder</title>
                                <title lang="en">Images</title>
                            </meta>
        
                            <tag name="sulu.block_preview" priority="512"/>
                        </property>
                    </properties>
                </type>
            </types>
        </block>
    </properties>
</form>
```

chirimoya:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #6381
| Related issues/PRs | none
| License | MIT
| Documentation PR | https://github.com/sulu/sulu-docs/pull/787

#### What's in this PR?

Global block types handling

#### Why?

Reduces size of block structure metadata and make it easier to reuse block structures.

#### Example Usage

```xml
<block name="blocks" default-type="text_block" minOccurs="0">
    <types>
        <type ref="text_block"/>
        <type name="gallery">
            <properties>
                <property name="images" type="media_selection"/>
            </properties>
        </type>
    </types>
</block>
```

#### To Do

- [x] Functionality
  - [x] Read Blocks
  - [x] Write Blocks
  - [x] Block Title (getMeta?)
  - [x] JSON Validation (mandatory fields)
- [x] Get feedback (@alexander-schranz, @mamazu)
- [x] Finalize implementation
- [x] Add tests

#### To Disucuss

- [x] Naming "global-block" => "global-type"
- [x] Configuration in the XML

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/metadataStore.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/metadataStore.js
@@ -47,6 +47,13 @@ class MetadataStore {
                     throw new Error(errorMessage);
                 }
 
+                if (!this.hasGlobalBlock(typeConfiguration)) {
+                    return {
+                        typeConfiguration,
+                        blockSchema: null,
+                    };
+                }
+
                 return metadataStore.loadMetadata(FORM_TYPE, 'block', {}).then((blockSchema) => {
                     return {
                         typeConfiguration,
@@ -54,21 +61,48 @@ class MetadataStore {
                     };
                 });
             }).then(({typeConfiguration, blockSchema}) => {
+                if (!blockSchema) {
+                    return typeConfiguration.form;
+                }
+
                 Object.keys(typeConfiguration.form).forEach((schemaFieldKey) => {
-                    if (typeConfiguration.form[schemaFieldKey].types) {
-                        Object.keys(typeConfiguration.form[schemaFieldKey].types).forEach((key) => {
-                            // need to be checked in another way, ref or global flag property ???
-                            if (typeConfiguration.form[schemaFieldKey].types && toJS(typeConfiguration.form[schemaFieldKey].types[key].form).length === 0) {
-                                if (typeConfiguration.form[schemaFieldKey].types && blockSchema.types[key]) {
-                                    typeConfiguration.form[schemaFieldKey].types[key].form = blockSchema.types[key].form;
-                                }
-                            }
-                        });
+                    if (!typeConfiguration.form[schemaFieldKey].types) {
+                        return;
                     }
+
+                    Object.keys(typeConfiguration.form[schemaFieldKey].types).forEach((key) => {
+                        if (!typeConfiguration.form[schemaFieldKey].types
+                            || toJS(typeConfiguration.form[schemaFieldKey].types[key].form).length > 0
+                        ) {
+                            return;
+                        }
+
+                        if (typeConfiguration.form[schemaFieldKey].types && blockSchema.types[key]) {
+                            typeConfiguration.form[schemaFieldKey].types[key].form = blockSchema.types[key].form;
+                        }
+                    });
                 });
 
                 return typeConfiguration.form;
             });
+    }
+
+    hasGlobalBlock(typeConfiguration: {form: Object}) {
+        return undefined !== Object.keys(typeConfiguration.form).find((schemaFieldKey) => {
+            if (!typeConfiguration.form[schemaFieldKey].types) {
+                return false;
+            }
+
+            return undefined !== Object.keys(typeConfiguration.form[schemaFieldKey].types).find((key) => {
+                if (!typeConfiguration.form[schemaFieldKey].types
+                    || toJS(typeConfiguration.form[schemaFieldKey].types[key].form).length > 0
+                ) {
+                    return false;
+                }
+
+                return !!typeConfiguration.form[schemaFieldKey].types;
+            });
+        });
     }
 
     getJsonSchema(formKey: string, type: ?string, metadataOptions: ?Object): Promise<Object> {
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/metadataStore.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/metadataStore.js
@@ -1,8 +1,9 @@
 // @flow
+import {toJS} from 'mobx';
 import metadataStore from '../../../stores/metadataStore';
 import type {Schema, SchemaTypes} from '../types';
 
-const FORM_TYPE = 'form';
+export const FORM_TYPE = 'form';
 
 class MetadataStore {
     getSchemaTypes(formKey: string, metadataOptions: ?Object): Promise<?SchemaTypes> {
@@ -45,6 +46,26 @@ class MetadataStore {
 
                     throw new Error(errorMessage);
                 }
+
+                return metadataStore.loadMetadata(FORM_TYPE, 'block', {}).then((blockSchema) => {
+                    return {
+                        typeConfiguration,
+                        blockSchema,
+                    };
+                });
+            }).then(({typeConfiguration, blockSchema}) => {
+                Object.keys(typeConfiguration.form).forEach((schemaFieldKey) => {
+                    if (typeConfiguration.form[schemaFieldKey].types) {
+                        Object.keys(typeConfiguration.form[schemaFieldKey].types).forEach((key) => {
+                            // need to be checked in another way, ref or global flag property ???
+                            if (typeConfiguration.form[schemaFieldKey].types && toJS(typeConfiguration.form[schemaFieldKey].types[key].form).length === 0) {
+                                if (typeConfiguration.form[schemaFieldKey].types && blockSchema.types[key]) {
+                                    typeConfiguration.form[schemaFieldKey].types[key].form = blockSchema.types[key].form;
+                                }
+                            }
+                        });
+                    }
+                });
 
                 return typeConfiguration.form;
             });
--- a/src/Sulu/Bundle/CoreBundle/DependencyInjection/SuluCoreExtension.php
+++ b/src/Sulu/Bundle/CoreBundle/DependencyInjection/SuluCoreExtension.php
@@ -62,16 +62,15 @@ class SuluCoreExtension extends Extension implements PrependExtensionInterface
             );
         }
 
-        $templatesPath = __DIR__ . '/../Content/templates';
         $container->prependExtensionConfig(
             'sulu_core',
             [
                 'content' => [
                     'structure' => [
                         'paths' => [
-                            'sulu' => [
-                                'path' => $templatesPath,
-                                'type' => 'page',
+                            'blocks' => [
+                                'path' => '%kernel.project_dir%/config/templates/blocks',
+                                'type' => 'block',
                             ],
                         ],
                     ],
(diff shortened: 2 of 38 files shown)


mamazu:
It was that easy? :D

But as for the implemetation, I'd suggest using a different tag name for this. This will then also allow us to expand this to other stuff as well.

mamazu:
Just tried this out in the dev environment only globalizing one block (or biggest content block) and these are the results:

| Metric |Before | After | 
|------------|----------|---------|
| File Size | 61.6MB | 13.49MB | 
| Compressed | 1.51 MB | 343KB |
| Time for getting the metadata | 5sec | 1sec|

PS: The measurements are done with beautified json so most of the config is tabs or spaces.

wachterjohannes:
i would suggest as the configuration:

```
<block name="blocks" default-type="text_block" minOccurs="0">
    <types>
        <type ref="text_block"/>
        <type name="gallery">
            <properties>
                <property name="images" type="media_selection"/>
            </properties>
        </type>
    </types>
</block>
```

which means that the reference is always a global block and without a name the name will be used by the global block but this name can be overwritten in the tag by adding a name attribute.

wachterjohannes:
Additionally is this not only a thing of of the block content type. We could also use it for other types which uses the types/type definition (image_map) for example.

But then the name "global-block" does not match anymore. I would suggest to rename this feature to "global-type". which is nearer a naming where it is used

wachterjohannes:
@alexander-schranz @chirimoya we have to discuss the open points next week

wachterjohannes:
@mamazu i have finalized the implementation (added the validation) and also changed the configuration a bit! can you maybe test again this branch?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/metadataStore.js
```diff
@@ -46,6 +47,26 @@ class MetadataStore {
                     throw new Error(errorMessage);
                 }
 
+                return metadataStore.loadMetadata(FORM_TYPE, 'block', {}).then((blockSchema) => {
```
wachterjohannes, line 57: only do this when a global block is needed

mamazu:
Well, one thing I found out during the testing is that you can not have `xi:includes` in your new block definition:
```text
[ERROR 1871] Element 'xi:include': This element is not expected. (in /var/www/htdocs/config/templates/blocks/flexbox_row.xml - line 15, column 0)
```

Code:
```xml
<?xml version="1.0" ?>
<properties xmlns="http://schemas.sulu.io/template/template"
            xmlns:xi="http://www.w3.org/2001/XInclude"
>
    <block name="blocks" default-type="flexbox_row">
        <types>
            <type name="flexbox_row">
                <meta>
                    <title>Flexbox row</title>
                </meta>
                <properties>
                    <xi:include href="../blocks/fragments/padding_responsive.xml" xpointer="xmlns(sulu=http://schemas.sulu.io/template/template)xpointer(/sulu:properties/sulu:section)"/>
                    <property name="full_container" type="checkbox" colspan="4">
                        <meta>
                            <title>Full flexbox row container</title>
                        </meta>

                        <params>
                            <param name="type" value="checkbox"/>
                        </params>
                    </property>

                    <property name="collapse_margin" type="checkbox" colspan="4">
                        <meta>
                            <title>Collapse margin</title>
                        </meta>

                        <params>
                            <param name="type" value="checkbox"/>
                        </params>
                    </property>

                    <property name="color" type="color" colspan="4">
                        <meta>
                            <title>ui.color</title>
                        </meta>
                    </property>

                    <block name="flex_box" default-type="flex_box_block">
                        <xi:include href="../blocks/misc/flex_box_types.xml" xpointer="xmlns(sulu=http://schemas.sulu.io/template/template)xpointer(/sulu:properties/sulu:block/sulu:types)"/>
                    </block>
                </properties>
            </type>
        </types>
    </block>
</properties>
```

alexander-schranz:
@mamazu The example looks unexpected from my point of view. I did not have a deeper look yet at this implementation but my expection is that a new block type inside `config/templates/blocks` looks like this with `<template>` root key:

```xml
<?xml version="1.0" ?>

    <key>flexbox_row</key>

    <meta>
        <title lang="en">Flexbox Row</title>
        <title lang="de">Flexbox Reihe</title>
    </meta>

    <properties>
                    <xi:include href="../blocks/fragments/padding_responsive.xml" xpointer="xmlns(sulu=http://schemas.sulu.io/template/template)xpointer(/sulu:properties/sulu:section)"/>
                    <property name="full_container" type="checkbox" colspan="4">
                        <meta>
                            <title>Full flexbox row container</title>
                        </meta>

                        <params>
                            <param name="type" value="checkbox"/>
                        </params>
                    </property>

                    <property name="collapse_margin" type="checkbox" colspan="4">
                        <meta>
                            <title>Collapse margin</title>
                        </meta>

                        <params>
                            <param name="type" value="checkbox"/>
                        </params>
                    </property>

                    <property name="color" type="color" colspan="4">
                        <meta>
                            <title>ui.color</title>
                        </meta>
                    </property>

                    <block name="flex_box" default-type="flex_box_block">
                        <xi:include href="../blocks/misc/flex_box_types.xml" xpointer="xmlns(sulu=http://schemas.sulu.io/template/template)xpointer(/sulu:properties/sulu:block/sulu:types)"/>
                    </block>
    </properties>
</template>
```

Where xincludes should work the same as for articles and snippets already as metadata should still be loaded the same way.

mamazu:
But with @alexander-schranz code I get a different error: 
Attempted to call an undefined method named "getTypes" of class "Sulu\Bundle\AdminBundle\Metadata\FormMetadata\SectionMetadata".
Did you mean to call "getType"?
> Same error message as in the pipelines.

image

alexander-schranz:
The example looks like that your `xincludes` xmls live inside the `config/templates/blocks` directory but they should not be in that directory. It is expected that there now only live block templates. Like the pages only have page templates and articles only articles templates.

Like in the sulu demo I recommend you to move the xincluded xmls to something like `config/templates/includes` so they are not part of the templates blocks directory.

https://github.com/sulu/sulu-demo/tree/master/config/templates

Still there could be an error with that visitor.

wachterjohannes:
@mamazu yes there was an issue with the visitor. I think you have a section somewhere in the template! I have not tested it with it. But the last push should fix that. // cc  @alexander-schranz

wachterjohannes:
@alexander-schranz this is ready now :)

alexander-schranz:
@chirimoya will do the review here.

mamazu:
Looks good from my side too.

src/Sulu/Bundle/CoreBundle/DependencyInjection/SuluCoreExtension.php
```diff
@@ -62,16 +62,15 @@ public function prepend(ContainerBuilder $container)
...
         }
 
-        $templatesPath = __DIR__ . '/../Content/templates';
         $container->prependExtensionConfig(
             'sulu_core',
             [
                 'content' => [
                     'structure' => [
                         'paths' => [
-                            'sulu' => [
```
Prokyonn, line 72: @wachterjohannes shouldn't we still have the templates path config?

src/Sulu/Component/Content/Compat/Structure/LegacyPropertyFactory.php
```diff
@@ -34,11 +36,20 @@
...
+     */
     private $namespaceRegistry;
 
-    public function __construct(NamespaceRegistry $namespaceRegistry)
+    /**
+     * @var StructureMetadataFactoryInterface
+     */
+    private $structureFactory;
+
+    public function __construct(NamespaceRegistry $namespaceRegistry, StructureMetadataFactoryInterface $structureFactory)
```
Prokyonn: bc break, although I doubt that anyone has ever overridden this class 😅

src/Sulu/Bundle/CoreBundle/DependencyInjection/SuluCoreExtension.php
wachterjohannes: this directory was inside of the bunde and never exists 😂  so we can remove it

Prokyonn (approved):

PR #7231, opened 2023-12-13
Add anchor-field to MediaLinkTypeOverlay

### LINKED ISSUE #7082: Feature request: Allow fragments (and query string) for media-links
# Problem description

At the moment (Sulu 2.5) links to files from the media management cannot have an additional query string nor fragments. My usecase is a big PDF file with legal information, where we linked to certain pages by adding a `#page=72` fragment to our links in the past. Now with Sulu that is only possible by linking this PDF as external resource with an absolute URL (due to #5875), which is a hacky workaround.

# Proposed solution

I would prefer an input for additional query string and fragments for the created URL of the selected media element. As there may be already a query string in place, that field could take inputs like `query=whatever&anotherquery=anyhow#page=72` or `#page=72` and decide on its own, if it renders the content with `?` or `&`. That would be the most flexible solution in my opinion.

And having #6718 in mind, that would also allow setting `inline=1` manually instead of having an extra select for that.

What do others think?

spackmat:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | sort of
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #7082
| Related issues/PRs | #7082
| License | MIT

#### What's in this PR?

This PR adds an anchor field to the MediaLinkTypeOverlay (as a copy from the internal link overlay, that already has one).

#### Why?

As discussed in https://github.com/sulu/sulu/issues/7082 at least for PDF-media an URL fragment can be very useful (to jump to a specific page with `#page=12`). 

#### Example Usage

Links to a media files (mainly PDF files) can now get an anchor like `page=12` within the overlay form (just like links to pages) that is appended to the link's URL like `...longdocument.pdf#page=12`.

I didn't find any documentation for the link overlays, so there is no documentation PR,

#### To Do

Nothing from my point of view. Update: Added a test to cover that field.

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/Resources/js/containers/Link/overlays/MediaLinkTypeOverlay.js
+++ b/src/Sulu/Bundle/MediaBundle/Resources/js/containers/Link/overlays/MediaLinkTypeOverlay.js
@@ -29,7 +29,7 @@ export default class MediaLinkTypeOverlay extends React.Component<LinkTypeOverla
             open,
             title,
             target,
-            anchor
+            anchor,
         } = this.props;
 
         if (typeof href === 'string') {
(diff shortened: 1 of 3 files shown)


brabli:
Would be great to see this merged!

spackmat (commented):
Fixed the linting errors.

spackmat (commented):
Added trailing comma to match more linting rules.

alexander-schranz:
Thx, I will update the snapshot tests, mostly a little bit tricky if your system may run on none english system.

alexander-schranz:
@spackmat Thank you!

PR #7233, opened 2023-12-14
Fix deprecations on bundle classes methods and mark them as `@internal`

stollr:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? | no 
| BC breaks? | yes, but usually irrelevant
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs |-
| License | MIT
| Documentation PR | -

#### What's in this PR?

I have added method's return type declarations on all bundle classes.

#### Why?

The overridden methods of Symfony's base class may add the return type declarations in the future, which will lead to compile errors if not fixed.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -2611,11 +2611,6 @@ parameters:
 			path: src/Sulu/Bundle/AudienceTargetingBundle/Serializer/Subscriber/TargetGroupSerializeSubscriber.php
 
 		-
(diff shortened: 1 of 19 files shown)


alexander-schranz:
Thx for the pull request.

I also marked the methods as `@internal` as they never should really be used by an application as commented in #7261.

> The methods should never be used by an application, they are symfony `@internal` and used by symfony dependency injection and never directly called. Currently symfony outputs lot of deprecations when bundle not yet uses : void. Adding : void is a technical bc break but don't think that somebody ever extended from a bundle class of our side that so think we can still do it.

alexander-schranz:
@stollr Thx

PR #7234, opened 2023-12-15
Fix LinkType warning when provider is not defined

hual7:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Fix warning on undefined array key "provider"

#### Why?

The use of links in blocks within blocks sometimes causes a warning when the link is empty.

### DIFF
--- a/src/Sulu/Component/Content/Types/Link.php
+++ b/src/Sulu/Component/Content/Types/Link.php
@@ -71,7 +71,7 @@ class Link extends SimpleContentType
         }
 
         $result = [
-            'provider' => $value['provider'] ?? false,
+            'provider' => $value['provider'] ?? null,
             'locale' => $value['locale'],
         ];
 
--- a/src/Sulu/Component/Content/Types/Link.php
+++ b/src/Sulu/Component/Content/Types/Link.php
@@ -71,7 +71,7 @@ class Link extends SimpleContentType
         }
 
         $result = [
-            'provider' => $value['provider'],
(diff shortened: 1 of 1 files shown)


src/Sulu/Component/Content/Types/Link.php
```diff
@@ -71,7 +71,7 @@ public function getViewData(PropertyInterface $property): array
         }
 
         $result = [
-            'provider' => $value['provider'],
+            'provider' => $value['provider'] ?? false,
```
mamazu:
```suggestion
            'provider' => $value['provider'] ?? null,
```
This would be consitent to the php default behavior when the key is not set. Otherwise it could be a BC break if stuff is relying on it being null.

alexander-schranz:
@hual7 thx for the pull request, do you remember how this did happen? Switching block types?

A test case would be nice. And as bugfix we need to target the 2.4 branch, I already changed the branch here but rebasing is required:

```bash
git checkout <your-branch>
git fetch sulu
git rebase -i sulu/2.5 --onto sulu/2.4
git push --force
```

hual7:
> @hual7 thx for the pull request, do you remember how this did happen? Switching block types?

Using empty links within blocks causes the error.
Bildschirmfoto 2024-01-20 um 11 49 42
Checking if the value exists, didn't fix the warning.

alexander-schranz:
@hual7 Thank you!

PR #7235, opened 2023-12-18
Replace `merge` with `persist` in `TargetGroupRepository` and `TargetGroupRuleRepository`

### LINKED ISSUE #7198: Replace usage of EntityManager::merge with EntityManager::persist in TargetGroup and TargetGroupRule
# Problem description

Like in #7184 the EntityManager::merge is deprecated and we should not longer build logic on top of this. 

Currently there are still two places where we need to load the Entity instead of using the `merge`:

Replace `merge` with `persist`:

- [x] https://github.com/sulu/sulu/blob/633f23052ab2e87dd5ff4cb18df36577701485f9/src/Sulu/Bundle/AudienceTargetingBundle/Entity/TargetGroupRepository.php#L39
- [x] https://github.com/sulu/sulu/blob/633f23052ab2e87dd5ff4cb18df36577701485f9/src/Sulu/Bundle/AudienceTargetingBundle/Entity/TargetGroupRuleRepository.php#L32

Remove  `merge`: when the main entity is alrady correctly loaded following merges can be removed as the already are cascade-persist:

- [x] https://github.com/sulu/sulu/blob/633f23052ab2e87dd5ff4cb18df36577701485f9/src/Sulu/Bundle/AudienceTargetingBundle/Entity/TargetGroupRuleRepository.php#L27
- [x] https://github.com/sulu/sulu/blob/633f23052ab2e87dd5ff4cb18df36577701485f9/src/Sulu/Bundle/AudienceTargetingBundle/Entity/TargetGroupRepository.php#L33

# Proposed solution

We should load the Entity else create and persist the Entity. I'm not yet what the best solution here is as I first need to have a deeper look into it. As the TargetGroup works with deserialization which we maybe need to remove to avoid the  `merge` and migrate to `new`/`persist` handling. Maybe `cascade-persist` and `orphan-removal` would help here.

The hardest part is mostly how we avoid BC breaks with the current implementation.

redjanym:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #7198
| Related issues/PRs 
| License | MIT

#### What's in this PR?

Replace `merge` with `persist` in `TargetGroupRepository` and `TargetGroupRuleRepository`.

#### Why?

It removed the deprecated `merge` method from EntityManager. 

#### Example Usage

```php
// If you added new features, show examples of how to use them here

$this->getEntityManager()->merge($webspace);

// Now we can do
$this->getEntityManager()->persist($webspace);
```

### DIFF
--- a/src/Sulu/Bundle/AudienceTargetingBundle/Entity/TargetGroupRepository.php
+++ b/src/Sulu/Bundle/AudienceTargetingBundle/Entity/TargetGroupRepository.php
@@ -20,6 +20,18 @@ use Sulu\Component\Persistence\Repository\ORM\EntityRepository;
  */
 class TargetGroupRepository extends EntityRepository implements TargetGroupRepositoryInterface
 {
+    /**
+     * Saves the given target group to the repository.
+     *
+     * @return TargetGroupInterface
+     */
+    public function save(TargetGroupInterface $targetGroup)
+    {
+        $this->getEntityManager()->persist($targetGroup);
+
+        return $targetGroup;
+    }
+
     public function findByIds($ids)
     {
         $queryBuilder = $this->createQueryBuilder('targetGroup')
--- a/src/Sulu/Bundle/AudienceTargetingBundle/Entity/TargetGroupRepository.php
+++ b/src/Sulu/Bundle/AudienceTargetingBundle/Entity/TargetGroupRepository.php
@@ -20,55 +20,6 @@ use Sulu\Component\Persistence\Repository\ORM\EntityRepository;
  */
 class TargetGroupRepository extends EntityRepository implements TargetGroupRepositoryInterface
 {
-    public function save(TargetGroupInterface $targetGroup)
-    {
-        $newRules = [];
-        foreach ($targetGroup->getRules()->toArray() as $rule) {
-            $rule = $this->getEntityManager()->getRepository(TargetGroupRuleInterface::class)->save($rule);
-            $newRules[] = $rule;
-        }
-
-        $newWebspaces = [];
-        foreach ($targetGroup->getWebspaces()->toArray() as $webspace) {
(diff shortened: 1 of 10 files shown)


chirimoya:
@redjanym I've tested the PR. Sadly replacing `merge` with `persist` is not enough. Have you tested it in the admin interface? For testing locally you need to activate the Audience Targeting Bundle like described in the documentation ([AudienceTargetingBundle](https://docs.sulu.io/en/2.5/bundles/audience_targeting.html))

src/Sulu/Bundle/AudienceTargetingBundle/Entity/TargetGroupRepository.php
```diff
@@ -20,55 +20,6 @@
  */
 class TargetGroupRepository extends EntityRepository implements TargetGroupRepositoryInterface
 {
-    public function save(TargetGroupInterface $targetGroup)
```
alexander-schranz, line 23: The `save` need to be keep. Atleast it is responsible for `$this->entityManager->persist($targetGroup);`

src/Sulu/Bundle/AudienceTargetingBundle/Entity/TargetGroupRuleRepository.php
```diff
@@ -20,31 +20,4 @@
  */
 class TargetGroupRuleRepository extends EntityRepository implements TargetGroupRuleRepositoryInterface
 {
-    public function save(TargetGroupRuleInterface $targetGroupRule)
```
alexander-schranz, line 23: The `save` need to be keep. Atleast it is responsible for `$this->entityManager->persist($targetGroupRule);`

PR #7238, opened 2023-12-20
Add lastModified field to page settings

martinlagler:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Related issues/PRs | https://github.com/sulu/SuluArticleBundle/pull/655
| License | MIT

#### What's in this PR?

Add new updated date field to the settings tab, with a toggler to enable and disable it.

#### Why?

So pages can have a editable update date.

### DIFF
--- a/src/Sulu/Bundle/PageBundle/Document/BasePageDocument.php
+++ b/src/Sulu/Bundle/PageBundle/Document/BasePageDocument.php
@@ -14,6 +14,7 @@ namespace Sulu\Bundle\PageBundle\Document;
 use Sulu\Component\Content\Document\Behavior\ExtensionBehavior;
 use Sulu\Component\Content\Document\Behavior\LocalizedAuditableBehavior;
 use Sulu\Component\Content\Document\Behavior\LocalizedAuthorBehavior;
+use Sulu\Component\Content\Document\Behavior\LocalizedLastModifiedBehavior;
 use Sulu\Component\Content\Document\Behavior\LocalizedStructureBehavior;
 use Sulu\Component\Content\Document\Behavior\NavigationContextBehavior;
 use Sulu\Component\Content\Document\Behavior\OrderBehavior;
@@ -58,7 +59,8 @@ class BasePageDocument implements
     LocalizedAuditableBehavior,
     LocalizedTitleBehavior,
     VersionBehavior,
-    LocalizedAuthorBehavior
+    LocalizedAuthorBehavior,
+    LocalizedLastModifiedBehavior
 {
     public const RESOURCE_KEY = 'pages';
 
@@ -248,6 +250,13 @@ class BasePageDocument implements
      * @var Version[]
      */
     protected $versions = [];
+
+    /**
+     * Date of lastModified.
+     *
+     * @var \DateTime|null
+     */
+    protected $lastModified;
 
     /**
      * Date of authoring.
@@ -510,6 +519,29 @@ class BasePageDocument implements
     public function setVersions($versions)
     {
         $this->versions = $versions;
+    }
+
+    public function getLastModifiedEnabled()
+    {
+        return null !== $this->lastModified;
+    }
+
+    /**
+     * @return \DateTime|null
+     */
+    public function getLastModified()
+    {
+        return $this->lastModified;
+    }
+
+    /**
+     * @param \DateTime|null $lastModified
+     *
+     * @return void
+     */
+    public function setLastModified($lastModified)
+    {
+        $this->lastModified = $lastModified;
     }
 
     public function getAuthored()
--- a/src/Sulu/Bundle/PageBundle/Form/Type/BasePageDocumentType.php
+++ b/src/Sulu/Bundle/PageBundle/Form/Type/BasePageDocumentType.php
@@ -55,6 +55,13 @@ abstract class BasePageDocumentType extends AbstractStructureBehaviorType
         $builder->add('shadowLocaleEnabled', CheckboxType::class);
         $builder->add('shadowLocale', TextType::class); // TODO: Should be choice of available shadow locales
         $builder->add(
+            'lastModified',
+            DateTimeType::class,
+            [
+                'widget' => 'single_text',
+            ]
+        );
+        $builder->add(
             'authored',
             DateTimeType::class,
             [
--- a/src/Sulu/Bundle/PageBundle/Resources/config/serializer/Document.BasePageDocument.xml
+++ b/src/Sulu/Bundle/PageBundle/Resources/config/serializer/Document.BasePageDocument.xml
@@ -6,6 +6,8 @@
         <property name="creator" type="integer" groups="defaultPage,preview"/>
         <property name="changer" type="integer" groups="defaultPage,preview"/>
         <property name="created" type="DateTime" groups="defaultPage,preview"/>
+        <virtual-property method="getLastModifiedEnabled" name="lastModifiedEnabled" serialized-name="lastModifiedEnabled" groups="defaultPage"/>
+        <property name="lastModified" type="DateTime" groups="defaultPage,preview"/>
         <property name="authored" type="DateTime" groups="defaultPage,preview"/>
         <property name="author" type="integer" groups="defaultPage,preview"/>
         <property name="published" type="DateTime" groups="defaultPage,smallPage,preview"/>
--- a/src/Sulu/Bundle/PageBundle/Trash/PageTrashItemHandler.php
+++ b/src/Sulu/Bundle/PageBundle/Trash/PageTrashItemHandler.php
@@ -63,7 +63,7 @@ final class PageTrashItemHandler implements
         TrashItemRepositoryInterface $trashItemRepository,
         DocumentManagerInterface $documentManager,
         DocumentInspector $documentInspector,
-        DocumentDomainEventCollectorInterface $documentDomainEventCollector
+        DocumentDomainEventCollectorInterface $documentDomainEventCollector,
     ) {
         $this->trashItemRepository = $trashItemRepository;
         $this->documentManager = $documentManager;
@@ -104,6 +104,7 @@ final class PageTrashItemHandler implements
                 : $localizedPage->getExtensionsData();
 
             $pageTitles[$locale] = $localizedPage->getTitle();
+            $lastModified = $localizedPage->getLastModified() ? $localizedPage->getLastModified()->format('c') : null;
 
             $data['locales'][] = [
                 'title' => $localizedPage->getTitle(),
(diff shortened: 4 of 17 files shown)


alexander-schranz:
I'm not sure about naming it `updated` I would more go with `lastModified`. Personally I like the `At` prefix for datetimes: `lastModifiedAt` but as we are not yet using that prefix for existsing`created`, `changed`, `published`, `authored`, ... it would be strange to include it now in one and not for the others.

alexander-schranz:
As discussed we will name it `lastModified`

src/Sulu/Bundle/PageBundle/Document/BasePageDocument.php
```diff
@@ -249,6 +249,18 @@ class BasePageDocument implements
      */
     protected $versions = [];
 
+    /**
+     * @var bool
+     */
+    protected $lastModifiedEnabled;
```
alexander-schranz: Do we need this here?

Is it not enough if we do:

```
    public function getLastModifiedEnabled()
    {
        return $this->lastModified !== null;
    }
```

src/Sulu/Bundle/PageBundle/Form/Type/BasePageDocumentType.php
```diff
@@ -54,6 +54,14 @@ public function buildForm(FormBuilderInterface $builder, array $options)
         $builder->add('workflowStage', IntegerType::class);
         $builder->add('shadowLocaleEnabled', CheckboxType::class);
         $builder->add('shadowLocale', TextType::class); // TODO: Should be choice of available shadow locales
+        $builder->add('lastModifiedEnabled', CheckboxType::class);
```
alexander-schranz: Do we need this for serializiation?

src/Sulu/Bundle/PageBundle/Trash/PageTrashItemHandler.php
```diff
@@ -104,13 +104,16 @@ public function store(object $page, array $options = []): TrashItemInterface
...
             $pageTitles[$locale] = $localizedPage->getTitle();
+            $lastModified = $localizedPage->getLastModified() ? $localizedPage->getLastModified()->format('c') : null;
 
             $data['locales'][] = [
                 'title' => $localizedPage->getTitle(),
                 'locale' => $locale,
                 'creator' => $localizedPage->getCreator(),
                 'created' => $localizedPage->getCreated()->format('c'),
                 'author' => $localizedPage->getAuthor(),
+                'lastModifiedEnabled' => $localizedPage->getLastModifiedEnabled(),
```
alexander-schranz: This should not be required to be part of trash. As the value should always be calculated based on the `lastModified` and not be saved by itself somewhere:

```suggestion
```

src/Sulu/Bundle/PageBundle/Trash/PageTrashItemHandler.php
```diff
@@ -172,12 +199,16 @@ public function restore(TrashItemInterface $trashItem, array $restoreFormData =
...
 
+            $lastModified = $localeData['lastModified'] ? new \DateTime($localeData['lastModified']) : null;
+
             $localizedPage->setTitle($localeData['title']);
             $localizedPage->setResourceSegment($localeData['resourceSegment']);
             $localizedPage->setSuluOrder($localeData['suluOrder']);
             $localizedPage->setLocale($locale);
             $localizedPage->setCreator($localeData['creator']);
             $localizedPage->setCreated(new \DateTime($localeData['created']));
+            $localizedPage->setLastModifiedEnabled($localeData['lastModifiedEnabled']);
```
alexander-schranz: This should not be set, as we should not save it. Just have a getter for the serialzer and to give it to the frontend:

```suggestion
```

src/Sulu/Bundle/PageBundle/Resources/config/serializer/Document.BasePageDocument.xml
```diff
@@ -6,6 +6,8 @@
         <property name="creator" type="integer" groups="defaultPage,preview"/>
         <property name="changer" type="integer" groups="defaultPage,preview"/>
         <property name="created" type="DateTime" groups="defaultPage,preview"/>
+        <property name="lastModifiedEnabled" type="boolean" groups="defaultPage"/>
```
alexander-schranz: We are not writing a property instead return a virtual property based on our new `getLastModifiedEnabled` method:

```
        <virtual-property method="getLastModifiedEnabled" name="lastModifiedEnabled" serialized-name="lastModifiedEnabled" groups="defaultPage"/>
 ```

File: src/Sulu/Component/Content/Document/Behavior/LocalizedAuthorBehavior.php [hunk 6]
```diff
@@ -16,6 +16,38 @@
...
+    public function getLastModifiedEnabled();
+
+    /**
+     * Set lastModifiedEnabled.
+     *
+     * @param bool $isLastModified
+     *
+     * @return void
+     */
+    public function setLastModifiedEnabled($isLastModified);
```
alexander-schranz: No setter for this.

src/Sulu/Component/Content/Document/Subscriber/LastModifiedSubscriber.php
```diff
@@ -0,0 +1,111 @@
...
+
+        $encoding = 'system_localized';
+        if ($document instanceof AuthorBehavior) {
+            $encoding = 'system';
+        } elseif (!$event->getLocale()) {
+            return;
+        }
+
+        $node = $event->getNode();
+        $document->setLastModifiedEnabled((bool)
```
alexander-schranz: we never set this value, we are handling this via `getLastModifiedEnabled`

src/Sulu/Component/Content/Document/Subscriber/LastModifiedSubscriber.php
```diff
@@ -0,0 +1,111 @@
...
+use Sulu\Component\DocumentManager\PropertyEncoder;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+/**
+ * Handles lastModifiedEnabled and lastModified.
+ */
+class LastModifiedSubscriber implements EventSubscriberInterface
+{
+    public const LAST_MODIFIED_PROPERTY_NAME = 'lastModified';
+    public const IS_LAST_MODIFIED_PROPERTY_NAME = 'isLastModified';
```
alexander-schranz: This will be not used and can be removed, as we just have a getter for the lastModifiedEnabled

src/Sulu/Component/Content/Document/Subscriber/LastModifiedSubscriber.php
```diff
@@ -0,0 +1,111 @@
...
+            $encoding = 'system';
+        } elseif (!$event->getLocale()) {
+            return;
+        }
+
+        $node = $event->getNode();
+        $node->setProperty(
+            $this->propertyEncoder->encode($encoding, self::IS_LAST_MODIFIED_PROPERTY_NAME, $event->getLocale()),
+            $document->getLastModifiedEnabled()
+        );
```
alexander-schranz: same as above we are not writing the enabled flag it is just a getter for the UI:

```suggestion
```

src/Sulu/Component/Content/Document/Subscriber/LastModifiedSubscriber.php
```diff
@@ -0,0 +1,111 @@
...
+            return;
+        }
+
+        $node = $event->getNode();
+        $document->setLastModifiedEnabled((bool)
+            $node->getPropertyValueWithDefault(
+                $this->propertyEncoder->encode($encoding, self::IS_LAST_MODIFIED_PROPERTY_NAME, $event->getLocale()),
+                null
+            )
+        );
```
alexander-schranz:
```suggestion
```

src/Sulu/Component/Content/Mapper/ContentMapper.php
```diff
@@ -809,6 +809,8 @@ private function rowToArray(
         }
 
         if ($document instanceof LocalizedAuthorBehavior) {
+            $documentData['isUpdated'] = $document->getLastModifiedEnabled();
+            $documentData['updated'] = $document->getLastModified();
```
alexander-schranz: I think we do not require the Enabled here so we can remove it and rename `updated`

```suggestion
            $documentData['lastModified'] = $document->getLastModified();
```

src/Sulu/Component/Content/Metadata/Loader/StructureXmlLoader.php
```diff
@@ -62,6 +62,8 @@ class StructureXmlLoader extends AbstractLoader
         'navContexts',
         'shadow-on',
         'shadow-base',
+        'isUpdated',
+        'updated',
```
alexander-schranz: The website/frontend should not get the is / enabled flag it just for the admin. So if possible we don't write it here.

```suggestion
        'lastModified',
```

src/Sulu/Component/Content/Compat/DataNormalizer.php
```diff
@@ -66,6 +66,8 @@ public static function normalize(FormEvent $event)
             'shadowLocaleEnabled' => self::getAndUnsetValue($data, 'shadowOn') ? true : false,
             'parent' => self::getAndUnsetValue($data, 'parent'),
             'workflowStage' => self::getAndUnsetValue($data, 'workflowStage'),
+            'lastModifiedEnabled' => self::getAndUnsetValue($data, 'lastModifiedEnabled'),
+            'lastModified' => self::getAndUnsetValue($data, 'lastModified'),
```
alexander-schranz: Is `lastModifiedEnabled` required here? What das toe DataNormalizer do is it responsible for writing the Admin API response? where data `$data` come from?

alexander-schranz (changes requested):
Where not required we should not return the `lastModifiedEnabled` that is just a flag for the Admin UI which we handle via virtual property in the serializer. In the other cases we should not write any props called like that to documents or somewhere else. Only return where we need it to return it for the pages / articles APIs

alexander-schranz (approved):
Need rebase to 2.6, we should wait with the merge until Article and Content Bundle is also implemented and fine.

src/Sulu/Component/Content/Document/Behavior/LocalizedAuthorBehavior.php
```diff
@@ -16,6 +16,29 @@
...
+
+    /**
+     * Set lastModified-date.
+     *
+     * @param \DateTime|null $lastModified
+     *
+     * @return void
+     */
+    public function setLastModified($lastModified);
+
```
alexander-schranz: This is a bc break, We need to create a new interface for this `LocalizedLastModifiedBehavior`.

alexander-schranz (approved):

PR #7239, opened 2023-12-20
Adding types around the `DocumentManager`

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | yes (in debug)
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?
This PR adds a service to the container that if the application is running in debug mode will add profiling info about the `DocumentManager` class.

> :warning: This assumes the interface definition is met. 
> If you have decorated the DocumentManager with custom logic that requires the `void` methods of the `DocumentManager` to return stuff, then in debug mode they don't return anything.

#### Why?
It's good to see what parts of the application are slow and if the `DocumentManager` is at fault (eg. routing is slow because the PHPCR takes long to find a given document).

#### Example Usage
Same as before.

### DIFF
--- a/src/Sulu/Component/DocumentManager/DocumentManager.php
+++ b/src/Sulu/Component/DocumentManager/DocumentManager.php
@@ -23,7 +23,7 @@ class DocumentManager implements DocumentManagerInterface, ResetInterface
     private $eventDispatcher;
 
     /**
-     * @var array Cached options resolver instances
+     * @var array<string, OptionsResolver> Cached options resolver instances
      */
     private $optionsResolvers = [];
 
@@ -161,7 +161,7 @@ class DocumentManager implements DocumentManagerInterface, ResetInterface
         return $event->getQuery();
     }
 
-    private function getOptionsResolver($eventName)
+    private function getOptionsResolver(string $eventName): OptionsResolver
     {
         if (isset($this->optionsResolvers[$eventName])) {
             return $this->optionsResolvers[$eventName];
--- a/src/Sulu/Component/DocumentManager/DocumentManagerInterface.php
+++ b/src/Sulu/Component/DocumentManager/DocumentManagerInterface.php
@@ -11,6 +11,7 @@
 
 namespace Sulu\Component\DocumentManager;
 
+use PHPCR\Query\QueryInterface;
 use Sulu\Component\DocumentManager\Query\Query;
 
 interface DocumentManagerInterface
@@ -20,6 +21,7 @@ interface DocumentManagerInterface
      *
      * @param string $identifier Path or UUID
      * @param string $locale Locale
+     * @param array<string, mixed> $options
      *
      * @return object
      *
@@ -43,6 +45,9 @@ interface DocumentManagerInterface
      *
      * @param object $document
      * @param string $locale
+     * @param array<string, mixed> $options
+     *
+     * @return void
      */
     public function persist($document, $locale = null, array $options = []);
 
@@ -51,6 +56,8 @@ interface DocumentManagerInterface
      * session.
      *
      * @param object $document
+     *
+     * @return void
      */
     public function remove($document/*, array $options = []*/);
 
@@ -59,6 +66,8 @@ interface DocumentManagerInterface
      *
      * @param object $document
      * @param string $locale
+     *
+     * @return void
      */
     public function removeLocale($document, $locale);
 
@@ -67,6 +76,8 @@ interface DocumentManagerInterface
      *
      * @param object $document
      * @param string $destId The path of the new parent
+     *
+     * @return void
      */
     public function move($document, $destId);
 
@@ -76,7 +87,7 @@ interface DocumentManagerInterface
      * @param object $document
      * @param string $destPath
      *
-     * @return string
+     * @return string|null
      */
     public function copy($document, $destPath);
 
@@ -97,6 +108,8 @@ interface DocumentManagerInterface
      *
      * @param object $document
      * @param string|null $destId
+     *
+     * @return void
      */
     public function reorder($document, $destId);
 
@@ -105,6 +118,9 @@ interface DocumentManagerInterface
      *
      * @param object $document
      * @param string $locale
+     * @param array<string, mixed> $options
+     *
+     * @return void
      */
     public function publish($document, $locale = null, array $options = []);
 
@@ -113,6 +129,8 @@ interface DocumentManagerInterface
      *
      * @param object $document
      * @param string $locale
+     *
+     * @return void
      */
     public function unpublish($document, $locale);
 
@@ -121,6 +139,8 @@ interface DocumentManagerInterface
      *
      * @param object $document
      * @param string $locale
+     *
+     * @return void
      */
     public function removeDraft($document, $locale);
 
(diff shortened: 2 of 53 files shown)


src/Sulu/Component/DocumentManager/DocumentManager.php
```diff
@@ -161,7 +161,7 @@ public function createQuery($query, $locale = null, array $options = [])
         return $event->getQuery();
     }
 
-    private function getOptionsResolver($eventName)
+    private function getOptionsResolver(string $eventName): OptionsResolver
```
mamazu, line 164: Not a BC break. It is always called with `string` and always returning OptionsResolver.

src/Sulu/Component/DocumentManager/TimingDocumentManager.php
```diff
@@ -0,0 +1,190 @@
...
+
+namespace Sulu\Component\DocumentManager;
+
+use Metadata\Driver\AdvancedFileLocatorInterface;
+use PHPCR\Query\QueryInterface;
+use Sulu\Component\DocumentManager\Query\Query;
+use Symfony\Component\Stopwatch\Stopwatch;
+use Symfony\Contracts\Service\ResetInterface;
+
+class TimingDocumentManager implements DocumentManagerInterface, ResetInterface
```
alexander-schranz: I just stumbled over https://github.com/sulu/sulu/blob/2.5/src/Sulu/Component/DocumentManager/EventDispatcher/DebugEventDispatcher.php which was build todo such things. As reported in the Slack Channel it is currently broken, so maybe we should just fix the `DebugEventDispatcher` and so avoid the BC Break in the fixtures you mentioned.

mamazu: That sounds like a good idea. It might still be worth using the interface in the fixtures though but that could also be part of 3.0

mamazu:
I have removed all the new and problematic code and it's now just type hints in this merge request.

src/Sulu/Component/DocumentManager/DocumentManagerInterface.php
```diff
@@ -20,6 +21,7 @@ interface DocumentManagerInterface
      *
      * @param string $identifier Path or UUID
      * @param string $locale Locale
+     * @param array<mixed> $options
```
alexander-schranz: I would even go with `array<string, mixed>` here

src/Sulu/Component/DocumentManager/Event/QueryCreateEvent.php
```diff
@@ -39,7 +40,7 @@ class QueryCreateEvent extends AbstractEvent
     private $primarySelector;
 
     /**
-     * @param string $innerQuery
+     * @param string|QueryInterface $innerQuery
```
alexander-schranz, line 43: when does this happen? Can it instead be casted? Or do we also need to change the `getInnerQuery`  return type

src/Sulu/Component/DocumentManager/DocumentManagerInterface.php
```diff
@@ -158,8 +187,9 @@ public function clear();
      * NOTE: This should not be used generally as it exposes the database structure and breaks abstraction. Use the
      *       domain-aware query builder instead.
      *
-     * @param mixed $query Either a JCR-SQL2 string, or a PHPCR query object
+     * @param string|QueryInterface $query Either a JCR-SQL2 string, or a PHPCR query object
```
alexander-schranz, line 190: when does this happen? is it maybe always QueryInterface and the string was just false?

In which cases it is a string and in which cases a QueryInterface?

src/Sulu/Component/DocumentManager/DocumentManagerInterface.php
```diff
@@ -130,6 +150,9 @@ public function removeDraft($document, $locale);
      * @param object $document
      * @param string $locale
      * @param string $version The UUID of the version to restore
+     * @param array<mixed> $options
```
alexander-schranz:
```suggestion
     * @param array<string, mixed> $options
```

src/Sulu/Component/DocumentManager/DocumentManagerInterface.php
```diff
@@ -105,6 +118,9 @@ public function reorder($document, $destId);
      *
      * @param object $document
      * @param string $locale
+     * @param array<mixed> $options
```
alexander-schranz:
```suggestion
     * @param array<string, mixed> $options
```

src/Sulu/Component/DocumentManager/DocumentManagerInterface.php
```diff
@@ -43,6 +45,9 @@ public function create($alias);
...
```
alexander-schranz:
```suggestion
     * @param array<string, mixed> $options
```

alexander-schranz (changes requested):
As it are just phpdocs changes we should target 2.4 for this.

src/Sulu/Component/DocumentManager/DocumentManagerInterface.php
mamazu: I think I have tried this in our implementation and there are still places that use this as a string, sadly.

mamazu:
Changed base to 2.4

src/Sulu/Component/DocumentManager/DocumentManagerInterface.php
alexander-schranz: I did have a look at the existing methods call in Sulu at current case I would not add the QueryInterface here as it indeed not designed to have a `PHPCR\Query\QueryInterface` reference here and is never used in sulu core that way or did I miss something?

alexander-schranz: Oh I see its here handled, then its fine: https://github.com/sulu/sulu/blob/347f0957917f9ac5e91b96fa4d668abe1e12fb6d/src/Sulu/Component/DocumentManager/Subscriber/Phpcr/QuerySubscriber.php#L64-L65

alexander-schranz: but the `getInnerQuery` should then also be `@return string|QueryInterface` on the `QueryCreateEvent`. @mamazu can you add that.

mamazu: Yeah both types are handled in this thing. That's why I extracted them as arguments like that.

src/Sulu/Component/DocumentManager/Event/QueryCreateEvent.php
mamazu: We could do that. I don't know the implications.

alexander-schranz: lets keep it and change the `getInnerQuery` to `@return string|QueryInterface`

mamazu: Updated it

alexander-schranz (approved):

alexander-schranz:
@mamazu Thank you!

PR #7247, opened 2024-01-15
Fix DebugEventDispatcher - Fix method naming and add a check if an event should be stoppable

claudissimus:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no 
| BC breaks? | no 
| Deprecations? | no 
| License | MIT

#### What's in this PR?

- The method doDispatch() is renamed to callListeners()
- A check if an event should be stoppable (as done in the base class) has been added
- the isPropagationStopped check is done before calling the listener (as done in the base class) to prevent behavioural differences between the original and this debug dispatcher.

#### Why?

The DebugEventDispatcher relies on overriding doDispatch() which has been removed from the symfony base class EventDispatcher, thus it never got called anymore anyway.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -46761,22 +46761,12 @@ parameters:
 			path: src/Sulu/Component/DocumentManager/Event/UnpublishEvent.php
 
 		-
-			message: "#^Method Sulu\\\\Component\\\\DocumentManager\\\\EventDispatcher\\\\DebugEventDispatcher\\:\\:doDispatch\\(\\) has no return type specified\\.$#"
+			message: "#^Call to an undefined method object\\:\\:getDebugMessage\\(\\)\\.$#"
 			count: 1
 			path: src/Sulu/Component/DocumentManager/EventDispatcher/DebugEventDispatcher.php
 
 		-
-			message: "#^Method Sulu\\\\Component\\\\DocumentManager\\\\EventDispatcher\\\\DebugEventDispatcher\\:\\:doDispatch\\(\\) has parameter \\$event with no type specified\\.$#"
-			count: 1
-			path: src/Sulu/Component/DocumentManager/EventDispatcher/DebugEventDispatcher.php
-
-		-
-			message: "#^Method Sulu\\\\Component\\\\DocumentManager\\\\EventDispatcher\\\\DebugEventDispatcher\\:\\:doDispatch\\(\\) has parameter \\$eventName with no type specified\\.$#"
-			count: 1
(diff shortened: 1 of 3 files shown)


mamazu:
Thanks for your contribution. Your change looks good. Would you mind also adding a test so that we can track that this doesn't happen again in the future? If you need help with that, just feel free to ask. ;)

And one thing I am curious about, you only copied over the right name and not the signature of the function. Wouldn't that also be a difference between dev and prod, if the dev dispatcher is more lenient?

claudissimus:
Hi @mamazu ,
indeed. Let me check if I can put together a test by myself.
As for the signature, that was lazyness on my side. I'll fix that, thanks for pointing that out.

mamazu (approved):

alexander-schranz:
Thx for your contribution @claudissimus, as a bugfix this should target the `2.4` branch as I think the issue also appear there. You can rebase on the 2.4 branch this way:

```bash
git checkout <your_branch>
git fetch sulu
git push --force
```

I already changed the target branch, so don't get confused by the current github view.

alexander-schranz:
@claudissimus Thank you!

alexander-schranz:
Looks like with the backmerge to 2.4 we run into an issue on PHP 7.2:

```
Declaration of Sulu\Component\DocumentManager\EventDispatcher\DebugEventDispatcher::callListeners(iterable $listeners, string $eventName, object $event): void should be compatible with Symfony\Component\EventDispatcher\EventDispatcher::callListeners(iterable $listeners, string $eventName, $event
```

We mostly would require using a Trait to support both PHP 7.2 and newer versions something like:

```php
if (PHP_VERSION_ID < 730000) { // we may alternatively need to check for the SymfonyVersion Kernel::VERSION_ID
    // PHP 7.2
    trait CallListenersTrait {
        public function callListeners(iterable $listeners, string $eventName, object $event)
        {
              $this->doCallListeners($listeners, $eventName, $event);
        }
    }
} else {
    trait CallListenersTrait {
        public function callListeners(iterable $listeners, string $eventName, object $event): void
        {
              $this->doCallListeners($listeners, $eventName, $event);
        }
    }
}

// use then inside the class the `CallListenersTrait` with
private function doCallListeners(iterable $listeners, string $eventName, object $event): void;
```

Somebody want to tackle this? /cc @claudissimus @mamazu

mamazu:
I don't have an easy way to test the 7.2 version.

alexander-schranz:
Fixing the PHP 7.2 compatibility in: https://github.com/sulu/sulu/pull/7275

PR #7248, opened 2024-01-16
Fix php-cs issues and update phpstan baseline

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no yes
| New feature? | no yes
| BC breaks? | no yes
| Deprecations? | no yes 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### To Do


alexander-schranz (approved):
@Prokyonn I fixed some of the phpstan issues. please check the last commit: https://github.com/sulu/sulu/pull/7248/commits/ce8713b8e51ddb5022376deaf76bc0f16b933b07

src/Sulu/Bundle/MediaBundle/Media/Manager/MediaManager.php
```diff
@@ -643,7 +643,7 @@ protected function setDataToMedia(Media $media, $data, $user)
                         $media->removeTags();
                         if (\count($value)) {
                             foreach ($value as $tag) {
-                                $tagEntity = $this->tagManager->findOrCreateByName($tag, $user->getId());
```
Prokyonn, line 646: @alexander-schranz I think the `$user` is not even needed anymore in this method, can we completely remove it?

Or do you want to preserve it because of BC break and the method is `protected` ?

alexander-schranz: would keep it, removing required parameters from protected methods is indeed a bc break, only none required params can be removed from a class or interface without breaking existing code.

PR #7257, opened 2024-01-28
Add a logger in the ResettingController

### LINKED ISSUE #7243: Exceptions in ResettingController:sendEmailAction
| Q | A
| --- | ---
| Sulu Version | 2.5.11 (but any version I think)
| PHP Version | 8.1
| DB Version | MySQL 5.7

# Actual Behavior

If something goes wrong when using the password reset form, a 204 is returned.

# Expected Behavior

An exception should be throw

# Steps to Reproduce

In ResettingController::sendEmailAction, all exceptions are ignored. I've had a couple of issues such as token send limit exceptions or Postmark exceptions, but I had to manually edit this file in production to be able to figure out what was going on.

Is there a specific reason why all exceptions are ignored?

# Possible Solutions

Just throw the exceptions?

mamazu:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? |  yes
| BC breaks? |  yes
| Deprecations? |  yes 
| Fixed tickets | fixes #7243 
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?
Adding the logger to the ResettingController

#### Why?
Swallowing the error and not showing any errors is not very helpful for debugging purposes adding a Logger to at least log the error will help.

### DIFF
--- a/src/Sulu/Bundle/SecurityBundle/Controller/ResettingController.php
+++ b/src/Sulu/Bundle/SecurityBundle/Controller/ResettingController.php
@@ -13,6 +13,8 @@ namespace Sulu\Bundle\SecurityBundle\Controller;
 
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\NoResultException;
+use Psr\Log\LoggerInterface;
+use Psr\Log\NullLogger;
 use Sulu\Bundle\ActivityBundle\Application\Collector\DomainEventCollectorInterface;
 use Sulu\Bundle\SecurityBundle\Domain\Event\UserPasswordResettedEvent;
 use Sulu\Bundle\SecurityBundle\Entity\User;
@@ -152,6 +154,11 @@ class ResettingController
     protected $secret;
 
     /**
+     * @var LoggerInterface
+     */
+    private $logger;
+
+    /**
      * @param PasswordHasherFactoryInterface|EncoderFactoryInterface $passwordHasherFactory
      */
     public function __construct(
@@ -174,7 +181,8 @@ class ResettingController
         string $mailTemplate,
         string $tokenSendLimit,
         string $adminMail,
-        string $secret
+        string $secret,
+        ?LoggerInterface $logger = null
     ) {
         $this->validator = $validator;
         $this->translator = $translator;
@@ -197,6 +205,7 @@ class ResettingController
         $this->tokenSendLimit = $tokenSendLimit;
         $this->adminMail = $adminMail;
         $this->secret = $secret;
+        $this->logger = $logger ?? new NullLogger();
     }
 
     /**
@@ -234,7 +243,7 @@ class ResettingController
             $email = $this->getEmail($user);
             $this->sendTokenEmail($user, $this->getSenderAddress($request), $email, $token);
         } catch (\Exception $ex) {
-            // do nothing
+            $this->logger->debug($ex->getMessage(), ['exception' => $ex]);
         }
 
         return new JsonResponse(null, Response::HTTP_NO_CONTENT);
--- a/src/Sulu/Bundle/SecurityBundle/Resources/config/services.xml
+++ b/src/Sulu/Bundle/SecurityBundle/Resources/config/services.xml
@@ -62,6 +62,7 @@
             <argument>%sulu_security.reset_password.mail.token_send_limit%</argument>
             <argument>%sulu_admin.email%</argument>
             <argument>%kernel.secret%%</argument>
+            <argument type="service" id="logger" on-invalid="null"/>
 
             <tag name="sulu.context" context="admin"/>
         </service>
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -27566,11 +27566,6 @@ parameters:
 			path: src/Sulu/Bundle/SecurityBundle/Controller/ResettingController.php
 
 		-
-			message: "#^Empty catch block\\. If you are sure this is meant to be empty, please add a \"// @ignoreException\" comment in the catch block\\.$#"
-			count: 1
-			path: src/Sulu/Bundle/SecurityBundle/Controller/ResettingController.php
-
-		-
 			message: "#^In method \"Sulu\\\\Bundle\\\\SecurityBundle\\\\Controller\\\\ResettingController\\:\\:sendEmailAction\", caught \"Exception\" must be rethrown\\. Either catch a more specific exception or add a \"throw\" clause in the \"catch\" block to propagate the exception\\. More info\\: http\\://bit\\.ly/failloud$#"
 			count: 1
 			path: src/Sulu/Bundle/SecurityBundle/Controller/ResettingController.php


src/Sulu/Bundle/SecurityBundle/Controller/ResettingController.php
```diff
@@ -224,11 +236,25 @@ public function sendEmailAction(Request $request)
...
+            // do not report when there is no user, because this could be used for username enumeration attacks
+            return new JsonResponse(null, Response::HTTP_NO_CONTENT);
+        } catch (UserNotInSystemException $ex) {
+            if ($this->logger instanceof LoggerInterface) {
+                $this->logger->error($ex->getMessage(), ['exception' => $ex]);
+            }
+            $response = new JsonResponse(null, Response::HTTP_BAD_REQUEST);
+        }
+
+        try {
```
alexander-schranz: This is expected in all cases the same http status code and content need to be returned for security reasons: 

```suggestion
```

https://github.com/sulu/sulu/security/advisories/GHSA-wfm4-pq59-wg6r

src/Sulu/Bundle/SecurityBundle/Controller/ResettingController.php
```diff
@@ -224,11 +236,25 @@ public function sendEmailAction(Request $request)
...
             $token = $this->generateTokenForUser($user);
             $email = $this->getEmail($user);
             $this->sendTokenEmail($user, $this->getSenderAddress($request), $email, $token);
         } catch (\Exception $ex) {
-            // do nothing
+            if ($this->logger instanceof LoggerInterface) {
+                $this->logger->error($ex->getMessage(), ['exception' => $ex]);
+            }
+            $response = new JsonResponse(null, Response::HTTP_INTERNAL_SERVER_ERROR);
```
alexander-schranz: This is expected in all cases the same http status code and content need to return for security reasons:

https://github.com/sulu/sulu/security/advisories/GHSA-wfm4-pq59-wg6r

```suggestion
```

src/Sulu/Bundle/SecurityBundle/Controller/ResettingController.php
```diff
@@ -224,11 +236,25 @@ public function sendEmailAction(Request $request)
...
+        }
+
+        try {
             $token = $this->generateTokenForUser($user);
             $email = $this->getEmail($user);
             $this->sendTokenEmail($user, $this->getSenderAddress($request), $email, $token);
         } catch (\Exception $ex) {
-            // do nothing
+            if ($this->logger instanceof LoggerInterface) {
+                $this->logger->error($ex->getMessage(), ['exception' => $ex]);
```
alexander-schranz: Errors should only be logged for unexpected exceptions. here a lot of expected exception happens. So either change the logger to debug.

```suggestion
                $this->logger->debug($ex->getMessage(), ['exception' => $ex]);
```

src/Sulu/Bundle/SecurityBundle/Controller/ResettingController.php
```diff
@@ -224,11 +236,25 @@ public function sendEmailAction(Request $request)
         try {
             /** @var UserInterface $user */
             $user = $this->findUser($request->get('user'));
+        } catch (EntityNotFoundException $ex) {
+            // do not report when there is no user, because this could be used for username enumeration attacks
+            return new JsonResponse(null, Response::HTTP_NO_CONTENT);
+        } catch (UserNotInSystemException $ex) {
+            if ($this->logger instanceof LoggerInterface) {
+                $this->logger->error($ex->getMessage(), ['exception' => $ex]);
```
alexander-schranz: This is an expected exception this should not end in an error log.

alexander-schranz (changes requested):

src/Sulu/Bundle/SecurityBundle/Controller/ResettingController.php
mamazu: Yeah, we should probably handle different exceptions differently. Like if the mailer can't reach the mail server. This could be considered an error. But having this as logging is already better than nothing.

src/Sulu/Bundle/SecurityBundle/Controller/ResettingController.php
```diff
@@ -228,7 +240,9 @@ public function sendEmailAction(Request $request)
             $email = $this->getEmail($user);
             $this->sendTokenEmail($user, $this->getSenderAddress($request), $email, $token);
         } catch (\Exception $ex) {
-            // do nothing
+            if ($this->logger instanceof LoggerInterface) {
```
alexander-schranz: This check can be avoided when fallback on __construct to `new NullLogger`.

src/Sulu/Bundle/SecurityBundle/Resources/config/services.xml
```diff
@@ -62,6 +62,7 @@
             <argument>%sulu_security.reset_password.mail.token_send_limit%</argument>
             <argument>%sulu_admin.email%</argument>
             <argument>%kernel.secret%%</argument>
```
mamazu, line 64: Btw that looks like a typo is it supposed to have two percent signs?

alexander-schranz: Looks wrong to me. Think this should be only one `%`. Currently it may just added always a `%` to the secret.

mamazu: I've opened an issue for that. I wouldn't want to fix that here.

src/Sulu/Bundle/SecurityBundle/Controller/ResettingController.php
mamazu: Done

src/Sulu/Bundle/SecurityBundle/Resources/config/services.xml
```diff
@@ -62,6 +62,7 @@
...
+            <argument type="service" id="monolog.logger"/>
```
alexander-schranz: I see that in most services the logger is optional so we can keep it optional also here and we use the `logger` as reference:

```suggestion
            <argument type="service" id="logger" on-invalid="null"/>
```

src/Sulu/Bundle/SecurityBundle/Controller/ResettingController.php
```diff
@@ -191,6 +199,11 @@ public function __construct(
         $this->tokenSendLimit = $tokenSendLimit;
         $this->adminMail = $adminMail;
         $this->secret = $secret;
+
+        if (null === $logger) {
+            trigger_deprecation('sulu/sulu', '2.4', 'The usage of the "ResettingController" without setting the Logger is deprecated. Please inject the Logger.');
+        }
+        $this->logger = $logger ?? new NullLogger();
```
alexander-schranz:
```suggestion
        $this->logger = $logger ?? new NullLogger();
```

alexander-schranz (changes requested):

alexander-schranz (approved):

PR #7261, opened 2024-02-02
Avoid bundle deprecation messages by add void return types and mark them as `@internal`

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no 
| BC breaks? | no
| Deprecations? | no  
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Avoid bundle deprecation messages by add void return types and mark them as `@internal`

#### Why?

The methods should never be used by an application, they are symfony `@internal` and used by symfony dependency injection and never directly called. Currently symfony outputs lot of deprecations when bundle not yet uses `: void`. Adding `: void` is a technical bc break but don't think that somebody ever extended from a bundle class of our side that so think we can still do it.


mamazu:
Possibily related to: #7233

Prokyonn (approved):

Prokyonn:
@alexander-schranz phpstan fails due to unnecessary ignored errors in the baseline 😀

alexander-schranz:
Thx will update it in #7233 and merge the community pull request.

PR #7263, opened 2024-02-05
Fix Article restore bug with pageTreeRoute if parent page is null by adding a fallback

martinlagler:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Related issues/PRs | https://github.com/sulu/SuluArticleBundle/pull/658
| License | MIT

#### What's in this PR?

Add fallback for PageTreeRoute, when there is no parent page (especially for articles)

#### Why?

When an Article with an empty parent page is restored from the trash, the parent page is set to null. But an array with path and uuid is expected.


alexander-schranz:
As discussed a unit test should be added so this get not accidently removed. and add a comment to the code.

alexander-schranz (approved):

PR #7268, opened 2024-02-07
Add focus for search fields

martinlagler:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Focus for Search fields. For example to focus when clicking on the Search icon on the list views. And auto focus search input on search page.

#### Why?

For better user experience.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/Search.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/Search.js
@@ -76,18 +76,20 @@ class Search extends React.Component<Props> {
 
     render() {
         return (
-            <Input
-                collapsed={this.collapsed}
-                icon="su-search"
-                onBlur={this.handleBlur}
-                onChange={this.handleChange}
-                onClearClick={this.handleClearClick}
-                onIconClick={this.handleIconClick}
-                onKeyPress={this.handleKeyPress}
-                placeholder={translate('sulu_admin.list_search_placeholder')}
-                skin="dark"
-                value={this.value}
-            />
+            <label aria-label={translate('sulu_admin.list_search_placeholder')}>
+                <Input
+                    collapsed={this.collapsed}
+                    icon="su-search"
+                    onBlur={this.handleBlur}
+                    onChange={this.handleChange}
+                    onClearClick={this.handleClearClick}
+                    onIconClick={this.handleIconClick}
+                    onKeyPress={this.handleKeyPress}
+                    placeholder={translate('sulu_admin.list_search_placeholder')}
+                    skin="dark"
+                    value={this.value}
+                />
+            </label>
         );
     }
 }
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/tests/__snapshots__/List.test.js.snap
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/tests/__snapshots__/List.test.js.snap
@@ -10,25 +10,29 @@ exports[`Do not render Loader instead of Adapter if no page count is given 1`] =
     <div
       class="toolbarLeft"
     >
-      <div
-        class="input dark left collapsed hasAppendIcon"
+      <label
+        aria-label="sulu_admin.list_search_placeholder"
       >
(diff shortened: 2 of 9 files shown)


alexander-schranz (approved):

alexander-schranz:
Seems to be this problem appears since: https://github.com/sulu/sulu/pull/6461.

So @FlorianBoe FYI if you think there is a better fix for this, let us know.

FlorianBoe:
Hi, since the changes in #6461, the search field no longer had a label, which is an problem.

However, it still does not make sense to place the label around the entire input component, as the input component also [contains interactive elements in this case](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#interactive_content).

In this case, the accessible name of the input is set uniquely by `aria-label`, but it should not contain the accessible name of the contained button as in the other cases, which is good.

From my experience, it is usually better to place the label element above the input element and create the relation using the `for` attribute.

```html
<label for="search">Search</label>
<div>
  <input type="search" name="search" id="search">
  <button>Search</button>
</div>
```

As there is no visible labelling in this case, the name of the entrance can also be assigned differently.

Using a visually-hidden label:

```html
<label for="search" class="visually-hidden">Search</label>
<div>
  <input type="search" id="search" inputmode="search">
  <button>Search</button>
</div>
```

Or by using `aria-labelledby`:

```html
<input type="search" aria-labelledby="search_button" inputmode="search">
<button type="submit" id="search_button">Search</button>
```

Attention! The accessible name of the search button is currently su-search, not search!

Using `aria-label` is generally not always the best solution to assign an accessible name for inputs, as it makes sense for most users to see the label of the input and there [are also problems to translate the provided value](https://adrianroselli.com/2019/11/aria-label-does-not-translate.html).

alexander-schranz:
@FlorianBoe Really thank you for your response and insight here. Did not know about the problem with the `aria-label`. So ma in our case `sr-only` would be good, the problem why we currently wrapping the `input` via `label` is that the way the browser autofocus the Input element. Without this seems not happening. My first indention was maybe use `autofocus` but that would mean that we have `label` pointing to a not available input:

```js
// copy to: https://playcode.io/react

import React, {useState} from 'react';

export function App(props) {
const [open, setOpen] = useState(false);

  return (
    <div>
      <label htmlFor="search" className="sr-only" style={{opacity: 0, position: 'absolute'}}>
         Search
      </label>
      <button type="button" onClick={() => { setOpen(true) }}>
         Search
      </button>
      {
        open
           ? <input id="search" type="name" autoFocus={true} />
           : null
      }
      </div>
  );
}
```

It may really would be better then that the button contains the label something like:

```js
// copy to: https://playcode.io/react

import React, {useState} from 'react';

export function App(props) {
const [open, setOpen] = useState(false);

  return (
    <div>
      <button id="search-button" type="button" onClick={() => { setOpen(true) }}>
          <span className="sr-only" style={{opacity: 0, position: 'absolute'}}>
             Search
          </span>

          <span>Search</span>{/* search icon */}
      </button>
      {
        open
           ? <input id="search" aria-labelledby="search-button" type="name" autoFocus={true} />
           : null
      }
      </div>
  );
}
```

What do you think?

FlorianBoe:
In your first example, you would hide the input, so you end up with a label pointing to nothing. As you said, not good. So it is better to hide and show the whole form element. Like this...

```js
return (
  <div>
    <button type="button" onClick={() => { setOpen(true) }}>
      Search
    </button>
    {
      open
        ? <div>
            <label htmlFor="search" className="sr-only" style={{opacity: 0, position: 'absolute'}}>Search</label>
            <input id="search" type="search" autoFocus={true} />
          </div>
        : null
    }
  </div>
);
```

In your second example you again have a problem with the accessible name. Because the icon has the aria-label “su-search”, the label now reads “Search su-search”. So you might want to remove the aria label from the icon or add `aria-hidden="true"` to it.

If you are having problems with autofocus, why not move the focus on toggle with the [focus method](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/focus)?

PR #7273, opened 2024-02-09
Fix Postgres DoctrineDbal tests and Refactor CI tasks 

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Refactor CI tasks.

#### Why?

Currently `prefer-lowest` use `postgres` and `jackrabbit` which make the most of the time to fail the tests. But lowest is one of the most important tests we really should take care of that it will not fail. Instead of using postgres and jackrabbit migrate lowest to use a more default setup with mysql and doctrinedbal which should make this tests better to run on the CI.

Using `docker-compose` file instead of github services allow to have more resources as postgres, mysql, jackrabbit are not always started / running. Only services which the specific run requires.

The cood thing is this make it also test locally easier with using:

```
# start services
docker compose -f tests/docker/docker-compose.postgres-13.yml up
# run postgres tests
DATABASE_URL=postgres://symfony:ChangeMe@127.0.0.1:5432/sulu_test?serverVersion=12.5 DATABASE_CHARSET=UTF8 DATABASE_COLLATE= bin/runtests -i -a
```

#### TODO

 - [x] Fix PHP 7.2 errors
 - [x] Fix Postgres DoctrineDBAL combination errors (CI never did run on postgres with doctrinedbal yet)
 - [x] Redefined required Github CI Tasks for 2.4 branch


alexander-schranz:
@chirimoya @Prokyonn @wachterjohannes please have a look at this. It should improve the CI and even make with docker compose file local running easier.

Most improtant changes ist postgres and jackrabbit are not longer used for `lowest` so that should in future not longer fail and avoids regression that PHP 7.2 or the lowest supported PHP version may not longer work.

src/Sulu/Component/Content/Export/WebspaceExport.php
```diff
@@ -299,6 +299,8 @@ protected function getDocumentsQueryString($webspaceKey, $uuid = null, $nodes =
 
         $queryString = 'SELECT * FROM [nt:unstructured] AS a WHERE ' . \implode(' AND ', $where);
 
+        $queryString .= ' ORDER BY [jcr:path] ASC';
+
```
alexander-schranz, line 303: This avoids error on postgres as it not returning the result in the same order as mysql when using doctrine dbal adapter

src/Sulu/Bundle/PageBundle/Tests/Functional/Repository/ContentRepositoryTest.php
```diff
@@ -209,9 +209,9 @@ public function testFindDescendantIdsById(): void
 
         $this->assertCount(3, $result);
 
-        $this->assertEquals($page1->getUuid(), $result[0]);
-        $this->assertEquals($page2->getUuid(), $result[1]);
-        $this->assertEquals($page3->getUuid(), $result[2]);
+        $this->assertContains($page1->getUuid(), $result);
+        $this->assertContains($page2->getUuid(), $result);
+        $this->assertContains($page3->getUuid(), $result);
```
alexander-schranz, line 214: Test should not care in which order the result is returned.

chirimoya (approved):
🎉

PR #7282, opened 2024-02-13
Fix removing of defaultMeta of a FileVersion removes also the whole FileVersion entity

### LINKED ISSUE #7266: Removing a FileVersionMeta entity also removes the FileVersion
| Q | A
| --- | ---
| Sulu Version | 2.5.11
| PHP Version | 8.2.14
| DB Version | MySQL 8.0.23

# Actual Behavior

Removing a `FileVersionMeta` (me_file_version_meta) entity from the database automatically removes the `FileVersion` (me_file_version) entity as well. This results in missing `FileVersions`.

# Expected Behavior

Removing `FileVersionMeta` only removes this entity and as long as there is still another `FileVersionMeta` (e.g. different locale) that references the `FileVersion`, the `FileVersion` should not be removed.

# Steps to Reproduce

Take a Sulu installation with multiple locales.
Save the file metadata in both languages, so that there are two `FileVersionMeta` entries for the same `FileVersion`.
Remove one of the `FileVersionMeta` entries and the `FileVersion` should still exist.

# Possible Solutions

Adjust the foreign key constraints

redjanym:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #7266 
| Related issues/PRs | NO
| License | MIT

#### What's in this PR?

Don't cascade to delete FireVersion when FileVersionMeta is deleted.

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/Resources/config/doctrine/FileVersion.orm.xml
+++ b/src/Sulu/Bundle/MediaBundle/Resources/config/doctrine/FileVersion.orm.xml
@@ -35,7 +35,7 @@
 
         <one-to-one field="defaultMeta" target-entity="Sulu\Bundle\MediaBundle\Entity\FileVersionMeta">
             <cascade>
-                <cascade-all />
+                <cascade-persist />
             </cascade>
 
             <join-column name="idFileVersionsMetaDefault" referenced-column-name="id" nullable="true" on-delete="CASCADE" />
--- a/src/Sulu/Bundle/MediaBundle/Resources/config/doctrine/FileVersion.orm.xml
+++ b/src/Sulu/Bundle/MediaBundle/Resources/config/doctrine/FileVersion.orm.xml
@@ -38,7 +38,7 @@
                 <cascade-persist />
             </cascade>
 
-            <join-column name="idFileVersionsMetaDefault" referenced-column-name="id" nullable="true" on-delete="CASCADE" />
+            <join-column name="idFileVersionsMetaDefault" referenced-column-name="id" nullable="true" on-delete="SET NULL" /><!-- why this should on-delete="RESTRICT" to avoid remove defaultMeta because of the nature of databases we need to use here SET NULL else the FileVersion is never deleteable -->
         </one-to-one>
         <one-to-many field="contentLanguages" target-entity="Sulu\Bundle\MediaBundle\Entity\FileVersionContentLanguage" mapped-by="fileVersion">
             <cascade>


src/Sulu/Bundle/MediaBundle/Resources/config/doctrine/FileVersion.orm.xml
```diff
@@ -35,7 +35,7 @@
 
         <one-to-one field="defaultMeta" target-entity="Sulu\Bundle\MediaBundle\Entity\FileVersionMeta">
             <cascade>
-                <cascade-all />
+                <cascade-persist />
             </cascade>
 
             <join-column name="idFileVersionsMetaDefault" referenced-column-name="id" nullable="true" on-delete="CASCADE" />
```
alexander-schranz:
```suggestion
            <join-column name="idFileVersionsMetaDefault" referenced-column-name="id" nullable="true" />
```

We want to remove this CASCADE so when somebody tries to remove the `defaultMeta` without setting a new default meta to the `FileVersion` they should get a SQL error.

Currently removing the DefaultMeta ends in trigger Cascade on the FileVersion which removes the FileVersion. While currently not reproducable in Sulu as there no language remove exists it happens in custom projects where removing of a file version did end in remove the fileversion object.

src/Sulu/Bundle/MediaBundle/Resources/config/doctrine/FileVersion.orm.xml
```diff
@@ -35,10 +35,10 @@
 
         <one-to-one field="defaultMeta" target-entity="Sulu\Bundle\MediaBundle\Entity\FileVersionMeta">
             <cascade>
-                <cascade-all />
+                <cascade-persist />
             </cascade>
 
-            <join-column name="idFileVersionsMetaDefault" referenced-column-name="id" nullable="true" on-delete="CASCADE" />
+            <join-column name="idFileVersionsMetaDefault" referenced-column-name="id" nullable="true" on-delete="SET NULL" /><!-- why this should on-delete="RESTRICT" to avoid remove defaultMeta because of the nature of databases we need to use here SET NULL else the FileVersion is never deleteable -->
```
alexander-schranz, line 41: @Prokyonn currently we can use not `RESTRICT` here because of nature how the deletion inside the database work it currently always fails to remove if we restrict it. Best solution for future would be not using relations for `default...` connection instead we use a `defaultLocale` and getDefaultMeta returns via doctrine critiria the correct meta.

src/Sulu/Bundle/MediaBundle/Resources/config/doctrine/FileVersion.orm.xml
alexander-schranz: We are not able todo that because of: https://github.com/sulu/sulu/pull/7282#discussion_r1508878235

PR #7287, opened 2024-02-15
Add field exclusion strategy for SnippetController

### LINKED ISSUE #7023: `fields` query parameter is being ignored on entity listings in the admin panel
| Q | A
| --- | ---
| Sulu Version | >=2.4.6
| PHP Version | 8.1
| DB Version | Mysql 8.0.23
| Browser Version | Firefox 109.0.1

# Actual Behavior

When viewing the listing or overview page of a content (like webspaces or snippets) we only need a subset of data that should be displayed in that view. For this use case Sulu has added a `field` query parameter to the url to specify which fields should be loaded from the database.

Currently this fields filter is either being ignored or not working which causes a list of 100 snippets with a lot of content in them to be very very slow since sulu is also rendering all the content on the snippets.

I only have checked this bug in the webspaces overview and the snippet list but I can imagine more endpoints are affected as well.

# Expected Behavior

When adding a filter parameter for the api like `fields=title,template,id` the response should only contain those fields.

# Steps to Reproduce

* Go to [sulu.rocks](https://sulu.rocks/admin) in the admin panel and login
* [Look at the following API Endpoint](https://sulu.rocks/admin/api/snippets?page=1&locale=en&limit=10&fields=title,template,id&flat=true)
* [Compare it to the same end point without field filters](https://sulu.rocks/admin/api/snippets?page=1&locale=en&limit=10&flat=true)

See that they are exactly the same and both have information on when the snippet was created in them. Which is okay when the `fields` parameter is missing but not when it is provided.

# Possible Solutions

Check to see why the filter has been removed or been deactivated or if it was a bug, please fix it.

mamazu:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | yes
| Deprecations? | no
| Fixed tickets | fixes #7023
| Related issues/PRs | #7023
| License | MIT
| Documentation PR | -

#### What's in this PR?
Adding code to only serialize fields specified in the request of the snippet controller.

#### Why?
When using the snippet listing by default it also serializes the content which is very slow. This PR adds an exclusion strategy that will not serialize the content anymore.

#### To Do

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -101,6 +101,11 @@ Additionally, the following packages where upgraded:
 
 This update is also handled normally by the update build command automatically.
 
+### Query parameters in the snippet controller work again
+
+In the previous version the `SnippetController` would return the entire content of the snippet in the `cgetAction`. Now
+it respects the list of fields provided in the query parameter and only returns those.
+
 ### DocumentToUuidTransformer return type changed
 
 For compatibility to Symfony 7 the `DocumentToUuidTransformer` methods return types changed:
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -101,11 +101,6 @@ Additionally, the following packages where upgraded:
 
 This update is also handled normally by the update build command automatically.
 
-### Query parameters in the snippet controller work again
-
-In the previous version the `SnippetController` would return the entire content of the snippet in the `cgetAction`. Now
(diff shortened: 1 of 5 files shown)


mamazu:
@alexander-schranz Do you think we can also get this shipped in 2.6 maybe?

alexander-schranz (approved):

PR #7043, opened 2023-03-28
Phpcr cleaner command

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

This PR adds a command to clean dead nodes from the phpCR.

#### Why?

Sulu doesn't delete properties when the data is no longer needed in the documents. This can pose a problem if you want to do a migration on phpcr nodes. You get a lot of unused properties that make it hard to distinguish if migrations are correct and it was just dead properties.

It also reduces the storage size.

#### Example Usage

```bash
bin/console sulu:document:clean
```

#### Known problems

With a lot of nodes and a lot of properties this command can run for a long time and even run into memory issues. For more info see this MR: https://github.com/jackalope/jackalope-doctrine-dbal/pull/421

#### To Do

- [x] Create a documentation PR
- [x] Add breaking changes to UPGRADE.md

#### Open Issues:

1. Shadowed Property

* Change a Block property to a text_line
* Block content will be the content of the new property
* The text_line content is a shadowedProperty value

2. Root properties

* When a root property was removed
* This data will not be removed

PR #7291, opened 2024-02-21
Add phpcr cleanup command

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no yes
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | none
| Related issues/PRs | replaces #7043 
| License | MIT
| Documentation PR | sulu/sulu-docs#791

#### What's in this PR?

This PR add a new command `sulu:phpcr:cleanup` to cleanup the phpcr repository.

#### Why?

This is needed to keep the repository clean.

#### Example Usage

```bash
bin/console sulu:phpcr:cleanup --force
```

#### To Do

- [x] Use also publish event and the live session to cleanup the live node
- [x] Test with different languages, blocks and image_maps
- [x] Test with audience-targeting and segmentation
- [x] Refactor to use StreamOutput with a file handle
- [x] Guard the node with:
  * Check the alias and if the structure manager has metadata for the alias (see https://github.com/mamazu/sulu/pull/2/files#diff-fd6471db309389445b7798ca7915fa5cd47010b1d91336a145f60e36276abb93R77-R80)
  * Check if language has a template 
  * Ignore nodes which could not be persisted (and or published)
- [x] Add summary of the cleanup (stats and ignored nodes) to the debug-file at the top
- [x] Feedback from @mamazu

### DIFF
--- a/src/Sulu/Bundle/DocumentManagerBundle/Command/PHPCRCleanupCommand.php
+++ b/src/Sulu/Bundle/DocumentManagerBundle/Command/PHPCRCleanupCommand.php
@@ -23,6 +23,7 @@ use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Output\StreamOutput;
 use Symfony\Component\Console\Style\SymfonyStyle;
 use Symfony\Component\HttpKernel\DependencyInjection\ServicesResetter;
+use Symfony\Component\Process\PhpExecutableFinder;
 use Symfony\Component\Process\PhpProcess;
 use Symfony\Component\Process\Process;
 
@@ -228,13 +229,17 @@ class PHPCRCleanupCommand extends Command
 
     protected function createProcess(string $uuid, bool $dryRun, bool $debug): Process
     {
-        $process = new PhpProcess(\implode(' ', \array_filter([
+        $executableFinder = new PhpExecutableFinder();
+        $php = $executableFinder->find(false);
+
+        $process = new Process(\array_filter([
+            $php,
             $_SERVER['argv'][0],
             'sulu:phpcr:cleanup:single-node',
             $uuid,
             $dryRun ? '--dry-run' : null,
             $debug ? '--debug' : null,
-        ])));
+        ]));
 
         $process->setTimeout(120);
 
--- a/src/Sulu/Bundle/DocumentManagerBundle/Command/PHPCRCleanupCommand.php
+++ b/src/Sulu/Bundle/DocumentManagerBundle/Command/PHPCRCleanupCommand.php
@@ -24,18 +24,10 @@ use Symfony\Component\Console\Output\StreamOutput;
 use Symfony\Component\Console\Style\SymfonyStyle;
 use Symfony\Component\HttpKernel\DependencyInjection\ServicesResetter;
 use Symfony\Component\Process\PhpExecutableFinder;
-use Symfony\Component\Process\PhpProcess;
 use Symfony\Component\Process\Process;
 
-/**
- * TODO:
- * * Snippets
- * * Articles
- */
 class PHPCRCleanupCommand extends Command
 {
-    protected static $defaultName = 'sulu:phpcr:cleanup';
-
     private OutputInterface $logger;
 
     public function __construct(
@@ -44,7 +36,7 @@ class PHPCRCleanupCommand extends Command
         private ServicesResetter $servicesResetter,
         private string $projectDirectory,
     ) {
-        parent::__construct();
+        parent::__construct('sulu:phpcr:cleanup');
     }
 
     protected function configure(): void
--- a/src/Sulu/Bundle/DocumentManagerBundle/Command/PHPCRCleanupCommand.php
+++ b/src/Sulu/Bundle/DocumentManagerBundle/Command/PHPCRCleanupCommand.php
@@ -125,6 +125,7 @@ class PHPCRCleanupCommand extends Command
         }
 
         $wheres[] = 'page.[jcr:path] LIKE "/cmf/snippets/%/%"';
+        $wheres[] = 'page.[jcr:path] LIKE "/cmf/articles/%/%/%"';
 
         $sql2 = \sprintf(
             'SELECT [jcr:uuid] FROM [nt:unstructured] AS page WHERE %s',
--- a/src/Sulu/Bundle/DocumentManagerBundle/Command/PHPCRCleanupCommand.php
+++ b/src/Sulu/Bundle/DocumentManagerBundle/Command/PHPCRCleanupCommand.php
@@ -127,6 +127,9 @@ class PHPCRCleanupCommand extends Command
             'removedProperties' => 0,
         ];
 
+        $ignoredUuids = [];
+        $erroredUuids = [];
+
         $io->section('Running cleanup process ...');
         $progressBar = $io->createProgressBar(\count($uuids));
         $progressBar->setFormat("Nodes: %nodes%\nIgnored: %ignoredNodes%\nErrored: %erroredNodes%\nDocuments: %documents%\nProperties: %properties%\nRemoved properties: %removedProperties%\n\n%current%/%max% [%bar%] %percent:3s%% %elapsed:6s%/%estimated:-6s% %memory:6s%\n\n");
@@ -153,11 +156,13 @@ class PHPCRCleanupCommand extends Command
                 $status = $process->wait();
                 if (PHPCRCleanupSingleNodeCommand::IGNORED === $status) {
                     ++$stats['ignoredNodes'];
+                    $ignoredUuids[] = $uuid;
 
                     continue;
                 }
                 if (0 !== $status) {
                     ++$stats['erroredNodes'];
+                    $erroredUuids[] = $uuid;
 
                     continue;
                 }
@@ -187,12 +192,18 @@ class PHPCRCleanupCommand extends Command
         $progressBar->finish();
         $io->success('Cleanup process finished');
 
+        $io->section('Following nodes are errored');
+        $io->listing($erroredUuids);
+
+        $io->section('Following nodes are ignored');
+        $io->listing($ignoredUuids);
+
         return self::SUCCESS;
     }
 
     protected function createProcess(string $uuid, bool $dryRun, bool $debug): Process
     {
-        return new Process(\array_filter([
+        $process = new Process(\array_filter([
             $_SERVER['PHP_BINARY'],
             $_SERVER['argv'][0],
             'sulu:phpcr:cleanup:single-node',
@@ -200,5 +211,9 @@ class PHPCRCleanupCommand extends Command
             $dryRun ? '--dry-run' : null,
             $debug ? '--debug' : null,
         ]));
+
+        $process->setTimeout(120);
+
+        return $process;
     }
 }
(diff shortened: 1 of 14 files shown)


src/Sulu/Bundle/DocumentManagerBundle/Command/PHPCRCleanupCommand.php
```diff
@@ -0,0 +1,251 @@
...
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\NullOutput;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Output\StreamOutput;
+use Symfony\Component\Console\Style\SymfonyStyle;
+use Symfony\Component\Process\Process;
+
+class PHPCRCleanupCommand extends Command
+{
+    protected static $defaultName = 'sulu:phpcr:cleanup';
```
mamazu: `defaultName` is deprecated we should either move to attributes or put the name into the configure function or the constructor of the base class.

src/Sulu/Bundle/DocumentManagerBundle/Command/PHPCRCleanupCommand.php
```diff
@@ -0,0 +1,251 @@
...
+        $io = new SymfonyStyle($input, $output);
+
+        $io->title('PHPCR Cleanup');
+
+        $dryRun = $input->getOption('dry-run');
+
+        if (!$dryRun) {
+            $io->warning('This command will remove properties from the PHPCR repository. Make sure to have a backup before running this command.');
+            if (!$input->getOption('force')) {
+                $answer = $io->ask('Do you want to continue [y/n]', null, function (?string $value) {
```
mamazu: Maybe it would be cleaner to use the [QuestionHelper](https://symfony.com/doc/current/components/console/helpers/questionhelper.html#asking-the-user-for-confirmation)?

src/Sulu/Bundle/DocumentManagerBundle/Command/PHPCRCleanupCommand.php
```diff
@@ -0,0 +1,251 @@
...
+                $property->getName(),
+                $matches,
+            );
+
+            if ($matches) {
+                $locales[$matches[1]] = $matches[1];
+            }
+        }
+
+        return \array_values(\array_unique($locales));
```
mamazu: Keys are always unique why not use those?

src/Sulu/Bundle/HttpCacheBundle/EventSubscriber/InvalidationSubscriber.php
```diff
@@ -80,6 +80,11 @@ class InvalidationSubscriber implements EventSubscriberInterface
      */
     private $environment;
 
+    /**
+     * @var bool
+     */
+    private $activated = true;
+
```
mamazu: That doesn't look like a good idea. Maybe we need some other sort of mechanism for disabling certain parts of the subscriber list. Like what if users had other subscribers that could be deactivated for performance?

src/Sulu/Bundle/DocumentManagerBundle/Command/PHPCRCleanupSingleNodeCommand.php
```diff
@@ -0,0 +1,280 @@
...
+            $writtenProperties = $liveCleanupNode->getWrittenPropertyKeys();
+            foreach ($this->cleanupNode($liveNode, $locale, $writtenProperties, $dryRun) as $result) {
+                ++$stats['properties'];
+                $stats['removedProperties'] += $result ? 1 : 0;
+            }
+
+            if (!$dryRun) {
+                $this->session->save();
+                $this->liveSession->save();
+            }
```
mamazu: Maybe batching could help performance?

wachterjohannes: the single command is only used for a single node 😂  so batching does not really make sense! the latest version uses multiple parallel processes to make it quite fast

src/Sulu/Bundle/DocumentManagerBundle/Command/PHPCRCleanupCommand.php
wachterjohannes: @mamazu why do you think it makes it cleaner? the SymfonyStyle is more or less the standard right?

src/Sulu/Bundle/DocumentManagerBundle/Command/PHPCRCleanupCommand.php
wachterjohannes: @mamazu you are right :) changed it

src/Sulu/Bundle/HttpCacheBundle/EventSubscriber/InvalidationSubscriber.php
wachterjohannes: hmm the issue here is not really for performance. the issue here is that when this subscriber is called all the pages in the varish will be purged. This is a performance risk on the website so i would really keep this as a guard that this is not happening // @alexander-schranz what do you think?

src/Sulu/Bundle/HttpCacheBundle/EventSubscriber/InvalidationSubscriber.php
```diff
@@ -103,6 +108,16 @@ public function __construct(
         $this->environment = $environment;
     }
 
+    public function deactivate(): void
+    {
+        $this->activated = false;
```
wachterjohannes: after a discusssion with alex we move this to an option

src/Sulu/Bundle/DocumentManagerBundle/Command/PHPCRCleanupSingleNodeCommand.php
mamazu: Fair enough. :D

src/Sulu/Bundle/DocumentManagerBundle/Command/PHPCRCleanupCommand.php
```diff
@@ -0,0 +1,251 @@
...
+use Symfony\Component\Console\Output\NullOutput;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Output\StreamOutput;
+use Symfony\Component\Console\Style\SymfonyStyle;
+use Symfony\Component\HttpKernel\DependencyInjection\ServicesResetter;
+use Symfony\Component\Process\PhpExecutableFinder;
+use Symfony\Component\Process\Process;
+use Webmozart\Assert\Assert;
+
+class PHPCRCleanupCommand extends Command
```
alexander-schranz:
```suggestion

/**
 * @internal
 */
class PHPCRCleanupCommand extends Command
```

src/Sulu/Bundle/DocumentManagerBundle/Command/PHPCRCleanupSingleNodeCommand.php
```diff
@@ -0,0 +1,291 @@
...
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\NullOutput;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+use Webmozart\Assert\Assert;
+
+class PHPCRCleanupSingleNodeCommand extends Command
```
alexander-schranz, line 42:
```suggestion
/**
 * @internal
 */
class PHPCRCleanupSingleNodeCommand extends Command
```

src/Sulu/Component/Content/Document/Subscriber/ShadowCopyPropertiesSubscriber.php
```diff
@@ -79,8 +81,11 @@ public function copyToShadows($document, NodeInterface $node)
         $categories = $this->getCategories($node, $document->getLocale());
         $navigationContext = $this->getNavigationContext($node, $document->getLocale());
 
-        foreach ($node->getProperties(self::SHADOW_BASE_PROPERTY) as $property) {
-            if ($property->getValue() === $document->getLocale()) {
+        foreach ($node->getProperties(self::SHADOW_BASE_PROPERTY) as $name => $property) {
+            $locale = $this->getLocale($name);
+            if ($node->getPropertyValueWithDefault(\sprintf(self::SHADOW_ON_PROPERTY, $locale), false)
+                && $property->getValue() === $document->getLocale()
+            ) {
```
alexander-schranz, line 88: Doe we need to backport this to 2.4 / 2.5?

src/Sulu/Bundle/PageBundle/Document/Subscriber/PublishSubscriber.php
```diff
@@ -170,6 +170,10 @@ public function reorderNodeInPublicWorkspace(ReorderEvent $event)
      */
     public function setNodeFromPublicWorkspaceForPublishing(PublishEvent $event)
     {
+        if ($event->hasNode()) {
+            return;
+        }
```
alexander-schranz, line 177: Can we add a comment here, why we need todo this.

alexander-schranz (changes requested):

src/Sulu/Component/Content/Document/Subscriber/ShadowCopyPropertiesSubscriber.php
wachterjohannes: done in https://github.com/sulu/sulu/pull/7310

alexander-schranz (approved):

mamazu:
:tada:

src/Sulu/Bundle/DocumentManagerBundle/Command/PHPCRCleanupCommand.php
```diff
@@ -0,0 +1,254 @@
...
+    }
+
+    protected function createProcess(string $uuid, bool $dryRun, bool $debug): Process
+    {
+        $executableFinder = new PhpExecutableFinder();
+        $php = $executableFinder->find(false);
+
+        $process = new Process(\array_filter([
+            $php,
+            $_SERVER['argv'][0],
```
alexander-schranz, line 243: @wachterjohannes @mamazu `$_SERVER['argv']` is risky and should not be accessed. This should be handled differently. Symfony resets for security reasons the values: https://github.com/symfony/symfony/pull/58785/files

mamazu: Isn't there a PhpProcess class in Symfony?

alexander-schranz: yeah but that is not the issue `$_SERVER` should not be accessed.

mamazu: I agree, however I don't know if there is an easy way of getting the path of the script that is currently executing in PHP otherwise.

alexander-schranz: `$GLOBALS['argv']` could be a workaround. or `$_SERVER['SCRIPT_NAME']`, `$_SERVER['PHP_SELF']`

mamazu: True, I mean phpcr is on the way out anyways. :D

Issue #7293, opened 2024-02-26
CDN url for storage providers


hual7:
# Problem description
When using storage providers https://docs.sulu.io/en/2.5/cookbook/external-media-storage.html via S3 like AWS, Digitalocean or Cloudflare the redirect path from `/media/ID/download/FILENAME` to the Cloudstorage url will be composed with `$this->endpoint . '/' . $this->bucketName . '/' . \ltrim($path, '/');`

It is not possible to set a CDN url and in case of Cloudflare R2 the redirect to the public files does not work without specifying a CDN url.

# Proposed solution

A separation of API url (endpoint) and CDN url (public) + url check in https://github.com/sulu/sulu/blob/e160828f94181a8eab539afba5ae7eca40dedba2/src/Sulu/Bundle/MediaBundle/Media/Storage/S3Storage.php#L53 would be great.

THX

alexander-schranz:
My prefered solution for this would be that we have a additonal `streamUrl`. Which is just `$path` and we could go over the `Symfony` way of configuring CDNs by providing a asset package.

```twig
{{ asset(media.streamUrl, 'sulu_media') }}
```

And let configure a CDN for it via:

```yaml
framework:
    assets:
        packages:
            sulu_media:
                base_urls: 'http://static_cdn.example.com/%bucketName%/'
```

PR #7294, opened 2024-02-28
Add public url to S3 storage for cdn support

hual7:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Related issues/PRs | #7293
| License | MIT

#### What's in this PR?

Add the possibility to define a S3 url for public redirect.

#### Why?

Some S3 providers like DigitalOcean or Cloudflare are providing a CDN url and with this update the url can be set in the S3 config.

#### Example Usage

```yaml
sulu_media:
    storage: s3
    storages:
        s3:
            key: 'xxx'
            secret: 'xxx'
            bucket_name: 'bucket-name'
            path_prefix: 'website/sulu'
            region: 'auto'
            endpoint: 'api-endpoint'
            url: 'cdn-url'
```

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/DependencyInjection/Configuration.php
+++ b/src/Sulu/Bundle/MediaBundle/DependencyInjection/Configuration.php
@@ -232,6 +232,7 @@ class Configuration implements ConfigurationInterface
                         ->scalarNode('path_prefix')->defaultNull()->end()
                         ->scalarNode('version')->defaultValue('latest')->end()
                         ->scalarNode('endpoint')->defaultNull()->end()
+                        ->scalarNode('url')->defaultNull()->end()
                         ->scalarNode('segments')->defaultValue(10)->end()
                         ->arrayNode('arguments')
                             ->scalarPrototype()->end()
--- a/src/Sulu/Bundle/MediaBundle/DependencyInjection/Configuration.php
+++ b/src/Sulu/Bundle/MediaBundle/DependencyInjection/Configuration.php
@@ -232,7 +232,7 @@ class Configuration implements ConfigurationInterface
                         ->scalarNode('path_prefix')->defaultNull()->end()
                         ->scalarNode('version')->defaultValue('latest')->end()
                         ->scalarNode('endpoint')->defaultNull()->end()
-                        ->scalarNode('url')->defaultNull()->end()
+                        ->scalarNode('public_url')->defaultNull()->end()
                         ->scalarNode('segments')->defaultValue(10)->end()
                         ->arrayNode('arguments')
                             ->scalarPrototype()->end()
--- a/src/Sulu/Bundle/MediaBundle/Media/Storage/S3Storage.php
+++ b/src/Sulu/Bundle/MediaBundle/Media/Storage/S3Storage.php
@@ -35,9 +35,9 @@ class S3Storage extends FlysystemStorage
     /**
      * @var string|null
      */
-    private $url;
+    private $publicUrl;
 
-    public function __construct(FilesystemInterface $filesystem, int $segments, ?string $url = null)
+    public function __construct(FilesystemInterface $filesystem, int $segments, ?string $publicUrl = null)
     {
         parent::__construct($filesystem, $segments);
 
@@ -50,7 +50,7 @@ class S3Storage extends FlysystemStorage
         $this->endpoint = (string) $this->adapter->getClient()->getEndpoint();
         $this->bucketName = $this->adapter->getBucket();
 
-        $this->url = null !== $url ? $url : ($this->endpoint . '/' . $this->bucketName);
+        $this->publicUrl = null !== $publicUrl ? $publicUrl : ($this->endpoint . '/' . $this->bucketName);
     }
 
     public function getPath(array $storageOptions): string
@@ -58,7 +58,7 @@ class S3Storage extends FlysystemStorage
         $filePath = $this->getFilePath($storageOptions);
         $path = $this->adapter->applyPathPrefix($filePath);
 
-        return $this->url . '/' . \ltrim($path, '/');
+        return $this->publicUrl . '/' . \ltrim($path, '/');
     }
 
     public function getType(array $storageOptions): string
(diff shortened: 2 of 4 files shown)


src/Sulu/Bundle/MediaBundle/Media/Storage/S3Storage.php
```diff
@@ -44,14 +46,16 @@ public function __construct(FilesystemInterface $filesystem, int $segments)
 
         $this->endpoint = (string) $this->adapter->getClient()->getEndpoint();
         $this->bucketName = $this->adapter->getBucket();
+
+        $this->url = $url;
```
alexander-schranz: We can do the fallback already here:

```suggestion
        $this->url = null !== $url ? $url : $this->endpoint;
```

src/Sulu/Bundle/MediaBundle/Media/Storage/S3Storage.php
```diff
@@ -44,14 +46,16 @@ public function __construct(FilesystemInterface $filesystem, int $segments)
...
     }
 
     public function getPath(array $storageOptions): string
     {
         $filePath = $this->getFilePath($storageOptions);
         $path = $this->adapter->applyPathPrefix($filePath);
 
-        return $this->endpoint . '/' . $this->bucketName . '/' . \ltrim($path, '/');
+        return ($this->url ?? $this->endpoint . '/' . $this->bucketName) . '/' . \ltrim($path, '/');
```
alexander-schranz: Can be simplified as arleady handled in the construct:

```suggestion
        return $this->url . '/' . $this->bucketName . '/' . \ltrim($path, '/');
```

src/Sulu/Bundle/MediaBundle/Media/Storage/S3Storage.php
```diff
@@ -32,9 +32,11 @@ class S3Storage extends FlysystemStorage
      */
     private $bucketName;
 
-    public function __construct(FilesystemInterface $filesystem, int $segments)
+    private $url;
+
+    public function __construct(FilesystemInterface $filesystem, int $segments, $url)
     {
-        parent::__construct($filesystem, $segments);
+        parent::__construct($filesystem, $segments, $url);
```
alexander-schranz: Parent call here not required:

```suggestion
        parent::__construct($filesystem, $segments);
```

src/Sulu/Bundle/MediaBundle/Media/Storage/S3Storage.php
```diff
@@ -32,9 +32,11 @@ class S3Storage extends FlysystemStorage
      */
     private $bucketName;
 
-    public function __construct(FilesystemInterface $filesystem, int $segments)
+    private $url;
+
+    public function __construct(FilesystemInterface $filesystem, int $segments, $url)
```
alexander-schranz: For BC reasons new params require to be optional:

```suggestion
    /**
     * @var string|null
     */
    private $url;

    public function __construct(FilesystemInterface $filesystem, int $segments, ?string $url = null)
```

src/Sulu/Bundle/MediaBundle/DependencyInjection/Configuration.php
```diff
@@ -232,6 +232,7 @@ private function addStorageSection(ArrayNodeDefinition $node)
                         ->scalarNode('path_prefix')->defaultNull()->end()
                         ->scalarNode('version')->defaultValue('latest')->end()
                         ->scalarNode('endpoint')->defaultNull()->end()
+                        ->scalarNode('url')->defaultNull()->end()
```
alexander-schranz: Not sure yet about naming it url maybe we should name it `cdn` or `cdn_url`. @wachterjohannes what do you think?

alexander-schranz (changes requested):
@chirimoya lets discuss this next week.

src/Sulu/Bundle/MediaBundle/Media/Storage/S3Storage.php
```diff
@@ -44,14 +49,16 @@ public function __construct(FilesystemInterface $filesystem, int $segments)
 
         $this->endpoint = (string) $this->adapter->getClient()->getEndpoint();
         $this->bucketName = $this->adapter->getBucket();
+
+        $this->url = null !== $url ? $url : $this->endpoint . '/' . $this->bucketName;
```
alexander-schranz: For better readability I would put the after part in ()

```suggestion
        $this->url = null !== $url ? $url : ($this->endpoint . '/' . $this->bucketName);
```

alexander-schranz: is the bucketName really not part of the url? https://3v4l.org/8JNSO

Else it would be need to be:

```suggestion
        $this->url = (null !== $url ? $url : $this->endpoint) . '/' . $this->bucketName;

hual7: That is correct, the CDN doesn't have the bucket name in the folder structure.

Digitalocean CDN: https://bucketname.fra1.cdn.digitaloceanspaces.com/file.example
Cloudflare CDN: https://customdomain.com/file.example

It is important to not automatically add the bucketName to the url.

src/Sulu/Bundle/MediaBundle/DependencyInjection/Configuration.php
alexander-schranz: /cc @chirimoya

chirimoya: `public_url`

alexander-schranz:
With the usage of flysystem 3 and the flysystem bundle this configuration will in 3.0 be moved to the new `flysystem.yaml` file.

A few examples can be found here: https://github.com/thephpleague/flysystem-bundle/blob/26d3794f7eb90acff0c5a51ad4709dd8e2ae8d8a/tests/Kernel/config/flysystem.yaml#L84

@mamazu think we should add the public_url also to the example upgrade: https://github.com/sulu/sulu/blob/3.0/UPGRADE.md#updating-to-flysystem-3

PR #7304, opened 2024-03-08
Fix preview issues with null token and slow requests

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no 
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Adds adjustments to the preview react component to prevent generating urls with the token being `null`.

#### Why?

This problem occures in a customer project when switching locales.

### DIFF
--- a/src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/Preview.js
+++ b/src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/Preview.js
@@ -170,6 +170,10 @@ class Preview extends React.Component<Props> {
         this.dataDisposer = reaction(
             () => toJS(formStore.data),
             (data) => {
+                if (this.iframeRef === null && !this.previewWindow) {
+                    return;
+                }
+
                 this.updatePreview(data);
             }
         );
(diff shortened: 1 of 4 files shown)


src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/Preview.js
```diff
@@ -170,6 +170,9 @@ class Preview extends React.Component<Props> {
         this.dataDisposer = reaction(
             () => toJS(formStore.data),
             (data) => {
+                if (this.iframeRef === null) {
```
Prokyonn: @wachterjohannes this is necessary to prevent sending an "update" request while the iframe is not yet rendered after a locale change

src/Sulu/Bundle/PreviewBundle/Resources/js/containers/Preview/Preview.js
```diff
@@ -186,13 +189,13 @@ class Preview extends React.Component<Props> {
         this.localeDisposer = reaction(
             () => toJS(formStore.locale),
             () => {
-                this.previewStore.restart();
+                this.previewStore.restart(toJS(formStore.locale));
```
wachterjohannes: move the toJS into the previewStore class. We also unpack the locale in the constructor there

wachterjohannes (approved):

PR #662, opened 2014-12-16
Added missing security checks

danrot:
Added the missing security checks for controller and navigation items in admin to Categories, Tags and Snippets.

__Tasks:__

- [x] test coverage
- [ ] gather feedback for my changes

__Informations:__

| Q             | A
| ------------- | ---
| Tests pass?   | yes
| Fixed tickets | fixes #467 
| BC Breaks     | Security have to be set correctly, otherwise users won't be able to work on categories, tags and snippets
| Doc           | none

PR #7309, opened 2024-03-13
Fix page_tree_route suffix

martinlagler:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #662
| License | MIT

#### What's in this PR?

Content type page_tree_route, saves wrong "routepath-suffix" if there are multiple pages/articles with the same route.

#### Why?

The routePath was updated correctly but not the routepath-suffix which is displayed in the sulu admin was wrong.

#### Example

If you have a page "overview" and create an page with "page_tree_route" and select "overview"  as parent page and name the page "test". The routePath is "/overview/test" and the routepath-suffix is "/test". If you do that again the routePath is "/overview/test-1" and the routepath-suffix was still "/test". Now the routepath-suffix is "/test-1" and displayed correctly in the sulu admin.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -26576,11 +26576,6 @@ parameters:
 			path: src/Sulu/Bundle/RouteBundle/Document/Subscriber/RoutableSubscriber.php
 
 		-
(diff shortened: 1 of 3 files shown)


wachterjohannes (approved):

wachterjohannes:
@Quehnie can you validate the correctness of the fix in your project?

PR #7313, opened 2024-03-18
Add last modified field value to sitemap when exists

martinlagler:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Related issues/PRs | #7238, #7313
| License | MIT

#### What's in this PR?

Adding the last modified field to the sitemap if it's defined.

#### Why?

For better content in the sitemap.

### DIFF
--- a/src/Sulu/Bundle/PageBundle/Sitemap/PagesSitemapProvider.php
+++ b/src/Sulu/Bundle/PageBundle/Sitemap/PagesSitemapProvider.php
@@ -80,7 +80,7 @@ class PagesSitemapProvider extends AbstractSitemapProvider
                 $portalInformation->getLocalization()->getLocale(),
                 $portalInformation->getPortalKey(),
                 MappingBuilder::create()
-                    ->addProperties(['changed', 'seo-hideInSitemap'])
+                    ->addProperties(['changed', 'seo-hideInSitemap', 'lastModified'])
                     ->setResolveUrl(true)
                     ->setHydrateGhost(false)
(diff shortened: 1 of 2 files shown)


src/Sulu/Bundle/PageBundle/Sitemap/PagesSitemapProvider.php
```diff
@@ -126,7 +126,7 @@ private function generateSitemapUrl(
         string $host,
         string $scheme
     ) {
-        $changed = $contentPage['changed'];
+        $changed = $contentPage['lastModified'] ?? $contentPage['changed'];
```
alexander-schranz, line 129: @chirimoya is the new expected behaviour not returning the created instead of changed date of the node? Should we remove the fallback to changed? It is a little bit strange here now. As soon as once the lastModified is defined the changed is not longer automatically provided to the sitemap, is that expected?

alexander-schranz: As discussed with @chirimoya to avoid a backwards compatibility break we will keep here the fallback to `changed` like currently ipmlemented. We maybe will change this in future when it is maybe possible in the UI of the publishing to set `update lastmodified timestamp` via a toggler.

alexander-schranz (changes requested):
We should add a test where the lastModified field is filled.

alexander-schranz (approved):

PR #7314, opened 2024-03-18
Allow to substitute the Contact entity

frickelbruder:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | no
| Related issues/PRs | 
| License | MIT
| Documentation PR | 

#### What's in this PR?

When applying this PR, it would be possible, to completely substitute the Contact entity in projects

#### Why?

I tried to substitute the contact entity following this guide: https://docs.sulu.io/en/2.5/cookbook/extend-entities.html

Instead of extending the contact entity, I wanted my own implementation, as I don't want to ask the user about his first and last name because of GDPR issues, when using the SuluCommunityBundle.
Extending the Contact-object leads to validating the first and last name, as all child objects automatically inherit the validation rules of the parent because of LSP.


alexander-schranz:
> Instead of extending the contact entity, I wanted my own implementation, as I don't want to ask the user about his first and last name because of GDPR issues, when using the SuluCommunityBundle.

The `firstName` and `lastName` are not required to use the SuluCommunityBundle, you need just define your defaults inside the `empty_data` in your custom registration type: 

https://github.com/sulu/SuluCommunityBundle/blob/64db8bd5b102489e2b7ef8c601c0aaa10105cbc6/Resources/doc/5-registration.md#type

And instead of using the shipped `registration` validation you use your own `app_registration` validation group and so you can define your validations and required fields yourself:

```php
    public function configureOptions(OptionsResolver $resolver)
    {
        $resolver->setDefaults(
            [
                'data_class' => User::class,
                'validation_groups' => ['app_registration'],
                'empty_data' => function(FormInterface $form) {
                    $user = new User();
                    $contact = new Contact();
                    $contact->setFirstName('');
                    $contact->setLastName('');
                    $user->setContact($contact);

                    return $user;
                }
            ]
        );
    }
```

I can not recommend replace the `Contact` with your own Entity without extending as it maybe will break several other cases where somebody want to create a relation to a sulu core contact entity and will then get your custom entity. So I really recommend to keep the contact that way. Also new features added to contact may then not work if you not correctly extend. The interface currently is just there because we require it for ORM mapping.

Keep still in mind you can extend your `User` or `Contact` entity with a one-to-one relation to your own custom entity, which we do ourselve also in projects. We add then additional tab to the contact to [display additional informations](https://github.com/sulu/sulu-demo/pull/91/files#diff-412607e34d7b73efaab548fb22726f4cf4b28e7ba92c4d916fe46f34ab9e68a4) we have in our custom entity, in case of DDD you could. If the relation is on the contact or user is up to you.

alexander-schranz:
rebased to 2.4: #7335 why the interface make sense to use where possible I still recommend you to extend from the normal entity and not overwrite it completely.

PR #7341, opened 2024-04-02
Add test case for page and snippet reference refresher

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Add test case for page reference refresher.

#### Why?

Currently this should fail on jackrabbit as the used syntax does not work there.


alexander-schranz:
@Prokyonn thank you

PR #7342, opened 2024-04-03
Tests should crash on non zero exit code

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | #6834
| License | MIT
| Documentation PR | -

#### What's in this PR?
This adds the function that the tests crash when one of the steps exists with a non-zero exit code.

#### Why?
Currently the tests just ignore the exit code and continue running which then causes issues further down the suite. This should make the tests easier to debug as the errors are reported earlier and closer to the source.

### DIFF
--- a/bin/runtests
+++ b/bin/runtests
@@ -135,13 +135,10 @@ function exec_cmd($cmdLine, $echo = true, $checkForError = true)
         write($out);
     });
 
-    if ($checkForError && $process->getExitCode() !== 0) {
-        write_error(sprintf(
-            'Could not execute command "%s", got exit code "%s": %s',
-            $cmdLine,
-            $process->getExitCode(),
-            $process->getErrorOutput()
-        ));
+    if ($checkForError) {
+        $process->mustRun();
+    } else {
+        $process->run();
     }
 
     return $process;
--- a/bin/runtests
+++ b/bin/runtests
@@ -284,16 +284,7 @@ function get_phpunit_path()
 
 function init_bundle()
 {
-    if (!$console = get_test_console_path()) {
-        return;
-    }
-
-    $process = exec_cmd(sprintf(
-        PHP_BINARY . ' %s %s',
-        $console,
-        'debug:container doctrine.orm.entity_manager'
-    ), false, false);
-
+    $process = exec_sf_cmd('debug:container doctrine.orm.entity_manager', false, false);
     // bundle does not use Doctrine ORM
     if ($process->getExitCode() !== 0) {
         return;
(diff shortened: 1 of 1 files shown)


alexander-schranz:
> Currently the tests just ignore the exit code and continue running which then causes issues further down the suite.

This is the expected behaviour of the runtests script. We run always all scripts and not exit directly when there is an error in one of the scripts. Else we don't have all errors in the CI for all bundles.

mamazu:
Okay the problem I was encountering in this case is the test shouldn't run if the init step fails (see the jackalope 2 merge request).

mamazu:
And the test cases in this case aren't affected because they are allowed to fail, their failure is recorded and just passed on.

alexander-schranz:
I see I would update the `init_dbal` part to throw exception there if the process for schema update finish with a none zero exit code.

alexander-schranz:
Something like:

```php
    $process = exec_sf_cmd(
        'doctrine:schema:update --force'
    );
    
    if (0 !== $process->getExitCode()) {
        // error already outputted before via write_error

        exit($process->getExitCode());
    }
```

mamazu:
That's also acceptable.

alexander-schranz:
We should target the 2.4 for this change.

mamazu:
Changed the base.

bin/runtests
```diff
@@ -281,34 +284,23 @@ function get_phpunit_path()
...
     write_info('Doctrine ORM detected, updating the schema for the bundle.');
 
-    exec_cmd(sprintf(
-        PHP_BINARY . ' %s %s',
-        $console,
-        ' doctrine:schema:update --force'
-    ));
+    $process = exec_sf_cmd('doctrine:schema:update --force');
+    if ($process->getExitCode() !== 0) {
+        exit($process->getExitCode());
```
alexander-schranz: we should not exit when one bundle fails to init.

the other bundle tests should still be executed.

bin/runtests
```diff
@@ -281,34 +284,23 @@ function get_phpunit_path()
...
+    }
 
-    exec_cmd(sprintf(
-        PHP_BINARY . ' %s %s',
-        $console,
-        ' sulu:document:initialize --purge --force --ansi'
```
alexander-schranz, line 310: I think we don't need to catch this. Aslong as the tests run as expected it is not required that a bundle uses the document manager bundle.

mamazu:
I wanted to catch this: https://github.com/sulu/sulu/actions/runs/8525431456/job/23352362111?pr=6834

The Init Task is successful although the update of the schema failed. Without this none of the tests will work.

bin/runtests
```diff
@@ -281,34 +284,23 @@ function get_phpunit_path()
...
 
-    exec_cmd(sprintf(
-        PHP_BINARY . ' %s %s',
-        $console,
-        ' doctrine:schema:update --force'
-    ));
+    $process = exec_sf_cmd('doctrine:schema:update --force');
+    if ($process->getExitCode() !== 0) {
+        exit($process->getExitCode());
+    }
```
alexander-schranz:
```suggestion
```

bin/runtests
```diff
@@ -281,34 +284,23 @@ function get_phpunit_path()
...
 
-    exec_cmd(sprintf(
-        PHP_BINARY . ' %s %s',
-        $console,
-        ' sulu:document:initialize --purge --force --ansi'
-    ));
+    $process = exec_sf_cmd('sulu:document:initialize --purge --force --ansi');
+    if ($process->getExitCode() !== 0) {
+        exit($process->getExitCode());
+    }
```
alexander-schranz:
```suggestion
```

alexander-schranz (changes requested):

alexander-schranz:
That error would still be catched we just don't should exit on the `init_bundle` part.

mamazu:
Looks good can you apply those and merge because the app doesn't support that.

alexander-schranz (approved):

PR #7343, opened 2024-04-03
Fix copy shadow properties

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | yes - Template will be overwritten for the locale which is a shadow
| Deprecations? | no
| Fixed tickets | fixes none
| Related issues/PRs | none
| License | MIT
| Documentation PR | none

#### What's in this PR?

This PR fixes the handling of shadow data. Some fields are copied to the shadow but until now not all which are needed and also not in the right priority

### DIFF
--- a/src/Sulu/Bundle/PageBundle/Resources/phpcr-migrations/Version202404051700.php
+++ b/src/Sulu/Bundle/PageBundle/Resources/phpcr-migrations/Version202404051700.php
@@ -13,10 +13,10 @@ namespace Sulu\Bundle\PageBundle;
 
 use PHPCR\Migrations\VersionInterface;
 use PHPCR\NodeInterface;
+use PHPCR\PhpcrMigrationsBundle\ContainerAwareInterface;
 use PHPCR\SessionInterface;
 use Sulu\Component\Content\Document\Subscriber\ShadowCopyPropertiesSubscriber;
 use Sulu\Component\Webspace\Webspace;
-use Symfony\Component\DependencyInjection\ContainerAwareInterface;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 
 class Version202404051700 implements VersionInterface, ContainerAwareInterface
@@ -26,8 +26,12 @@ class Version202404051700 implements VersionInterface, ContainerAwareInterface
      */
     private $container;
 
-    public function setContainer(?ContainerInterface $container = null)
+    public function setContainer(?ContainerInterface $container = null): void
     {
+        if (null === $container) {
+            throw new \RuntimeException('Container is required to run this migration.');
+        }
+
         $this->container = $container;
     }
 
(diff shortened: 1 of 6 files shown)


src/Sulu/Bundle/PageBundle/Resources/phpcr-migrations/Version202404051700.php
```diff
@@ -0,0 +1,190 @@
...
+class Version202404051700 implements VersionInterface, ContainerAwareInterface
+{
+    /**
+     * @var ContainerInterface
+     */
+    private $container;
+
+    public function setContainer(?ContainerInterface $container = null)
+    {
+        $this->container = $container;
```
alexander-schranz:
```suggestion
    public function setContainer(?ContainerInterface $container = null): void
    {
        if (null === $container) {
            throw new \RuntimeException('Container is required to run this migration.');
        }

    
        $this->container = $container;
```

src/Sulu/Bundle/PageBundle/Resources/phpcr-migrations/Version202404051700.php
```diff
@@ -0,0 +1,190 @@
...
+ */
+
+namespace Sulu\Bundle\PageBundle;
+
+use PHPCR\Migrations\VersionInterface;
+use PHPCR\NodeInterface;
+use PHPCR\SessionInterface;
+use Sulu\Component\Content\Document\Subscriber\ShadowCopyPropertiesSubscriber;
+use Sulu\Component\Webspace\Webspace;
+use Symfony\Component\DependencyInjection\ContainerAwareInterface;
```
alexander-schranz: This is deprecated and need to be replace on 2.6 with:

```suggestion
use PHPCR\PhpcrMigrationsBundle\ContainerAwareInterface;
```

src/Sulu/Bundle/PageBundle/Resources/phpcr-migrations/Version202404051700.php
```diff
@@ -0,0 +1,190 @@
...
+
+        /** @var string $shadowLocale */
+        $shadowLocale = $node->getPropertyValue($this->getPropertyName(ShadowCopyPropertiesSubscriber::SHADOW_BASE_PROPERTY, $locale));
+
+        $tags = $this->getTags($node, $shadowLocale);
+        $categories = $this->getCategories($node, $shadowLocale);
+        $navigationContext = $this->getNavigationContext($node, $shadowLocale);
+        $author = $this->getAuthor($node, $shadowLocale);
+        $authored = $this->getAuthored($node, $shadowLocale);
+        $lastModified = $this->getLastModified($node, $shadowLocale);
```
alexander-schranz, line 87: own migration for the new field this should not be part of the 2.4 migration.

alexander-schranz (changes requested):

alexander-schranz (approved):

PR #7339, opened 2024-03-31
Updating to phpunit 10

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | closes #6988 
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?
Updating phpunit to version 10 and fixing the tests

* Updating the dependency in the `composer.json`
* Running rector to upgrade the tests (it's using attributes instead of annotations now)
* Fixing the problems with the dataprovider functions being static now (mostly)

#### Why?
It's always a good idea to have an up to date testing library.

#### To Do

- [ ] Fix the tests

PR #7344, opened 2024-04-03
Upgrade Phpunit (part 1)

mamazu:
#### What's in this PR?

* replacing `assertObjectHasAttribute`
* Fixing some tests

#### Why?
Having it in the bigger #7339 PR makes it hard.

### DIFF
--- a/src/Sulu/Bundle/ContactBundle/Tests/Functional/Controller/AdminControllerTest.php
+++ b/src/Sulu/Bundle/ContactBundle/Tests/Functional/Controller/AdminControllerTest.php
@@ -147,12 +147,13 @@ class AdminControllerTest extends SuluTestCase
         $this->client->jsonRequest('GET', '/admin/metadata/list/accounts');
 
         $this->assertHttpStatusCode(200, $this->client->getResponse());
-        $response = \json_decode($this->client->getResponse()->getContent());
+        $response = \json_decode($this->client->getResponse()->getContent(), true);
 
-        $this->assertObjectHasAttribute('id', $response);
-        $this->assertObjectHasAttribute('name', $response);
-        $this->assertObjectHasAttribute('zip', $response);
-        $this->assertObjectHasAttribute('city', $response);
+        $this->assertIsArray($response);
+        $this->assertArrayHasKey('id', $response);
+        $this->assertArrayHasKey('name', $response);
+        $this->assertArrayHasKey('zip', $response);
+        $this->assertArrayHasKey('city', $response);
     }
 
     public function testContactFormMetadataAction(): void
@@ -162,16 +163,12 @@ class AdminControllerTest extends SuluTestCase
         $this->client->jsonRequest('GET', '/admin/metadata/form/contact_details');
 
         $this->assertHttpStatusCode(200, $this->client->getResponse());
-        $response = \json_decode($this->client->getResponse()->getContent());
+        $response = \json_decode($this->client->getResponse()->getContent(), true);
 
-        $form = $response->form;
-
-        $this->assertObjectHasAttribute('avatar', $form);
-        $this->assertObjectHasAttribute('contact', $form);
-
-        $schema = $response->schema;
-
-        $this->assertEquals(['firstName', 'lastName', 'formOfAddress'], $schema->required);
+        $this->assertIsArray($response);
+        $this->assertArrayHasKey('avatar', $response['form']);
+        $this->assertArrayHasKey('contact', $response['form']);
+        $this->assertEquals(['firstName', 'lastName', 'formOfAddress'], $response['schema']['required']);
     }
 
     public function testAccountFormMetadataAction(): void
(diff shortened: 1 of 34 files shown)


src/Sulu/Bundle/ContactBundle/Tests/Functional/Controller/AdminControllerTest.php
```diff
@@ -181,16 +178,15 @@ public function testAccountFormMetadataAction(): void
         $this->client->jsonRequest('GET', '/admin/metadata/form/account_details');
 
         $this->assertHttpStatusCode(200, $this->client->getResponse());
-        $response = \json_decode($this->client->getResponse()->getContent());
-
-        $form = $response->form;
-
-        $this->assertObjectHasAttribute('logo', $form);
-        $this->assertObjectHasAttribute('account', $form);
+        $response = \json_decode($this->client->getResponse()->getContent(), true);
```
mamazu, line 181: Going for array deserialization would make it easier in some places, as we can just compare arrays.

src/Sulu/Bundle/MediaBundle/Tests/Functional/Entity/MediaDataProviderRepositoryTest.php
```diff
@@ -59,7 +59,7 @@ class MediaDataProviderRepositoryTest extends SuluTestCase
     /**
      * @var array
      */
-    private $mediaData = [
+    private static $mediaData = [
```
mamazu, line 62: This data is static and doesn't need to live on the test. Helpful for making the dataproviders static later.

src/Sulu/Bundle/MediaBundle/Tests/Unit/Media/ImageConverter/Transformations/AbstractTransformation.php
```diff
@@ -17,10 +17,7 @@
 use Sulu\Bundle\MediaBundle\Media\ImageConverter\Transformation\TransformationInterface;
 use Sulu\Bundle\TestBundle\Testing\SuluTestCase;
 
-/**
- * Class AbstractScaleTest.
- */
-abstract class AbstractTransformationTest extends SuluTestCase
+abstract class AbstractTransformation extends SuluTestCase
```
mamazu, line 20: Renaming the test as it otherwise will be picked up as a test but can't be instantiated as a test.

src/Sulu/Bundle/TagBundle/Tests/Functional/Controller/TagControllerTest.php
```diff
@@ -310,13 +310,9 @@ public function testDeleteById(): void
 
         $tag1Id = $tag1->getId();
 
-        $mockedEventListener = $this->getMockBuilder('Mock')->setMethods(['onDelete'])->getMock();
-        $mockedEventListener->expects($this->once())->method('onDelete');
-
-        $this->client->getContainer()->get('event_dispatcher')->addListener(
-            'sulu.tag.delete',
-            [$mockedEventListener, 'onDelete']
-        );
```
mamazu, line 319: The way to check if this method is called can be simplified by just passing a closure and setting a variable inside to true. That way we don't need to create a mock object.

src/Sulu/Component/Rest/Tests/Unit/MockInterface.php
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Component\Rest\Tests\Unit;
+
+interface MockInterface
```
mamazu, line 16: This class was used in the RestHelperTest. Phpunit 10 will not allow to Mock classes that don't exist so, just created this interface to use it in mocking.

src/Sulu/Component/Rest/Tests/Unit/RestHelperTest.php
```diff
@@ -353,14 +353,15 @@ public function testprocessSubEntitiesWithAdd(): void
 
     public function testCompareEntitiesWithData(): void
     {
-        $mockedObject = $this->getMockBuilder('Mock')->setMethods(['getId', 'getValue'])->getMock();
-        $mockedObject->expects($this->any())->method('getId')->will($this->returnValue(1));
-        $mockedObject->expects($this->any())->method('getValue')->will($this->returnValue(2));
+        $mockedObject = $this->createConfiguredMock(MockInterface::class, [
+            'getId' => 1,
+            'getValue' => 2,
+        ]);
```
mamazu, line 359: Could be reverted but it's the same as above just shorter.

src/Sulu/Component/Security/Tests/Unit/Authorization/AccessControl/DoctrineAccessControlProviderTest.php
```diff
@@ -317,11 +317,9 @@ public function testSupport($type, $supported): void
 
     public function provideSupport()
     {
-        $securedEntity = $this->prophesize(SecuredEntityInterface::class);
-
         return [
             [\stdClass::class, false],
-            [\get_class($securedEntity->reveal()), true],
+            [SecuredEntityInterface::class, true],
```
mamazu, line 322: Why create a mock when we just need the class of it?

alexander-schranz (approved):

alexander-schranz:
@mamazu Thank you!

PR #7349, opened 2024-04-06
[DX] Using the `DocumentManagerInterface` everywhere

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | yes
| Deprecations? | yes
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?
Using the `DocumentManagerInterface` instead of the implementation.

#### Why?
* Making the `DocumentManager` implementation extendable
* Interface is used everywhere else in the code

#### To Do

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -14,6 +14,12 @@ Removing unused arguments:
 
 - `Sulu\Component\Webspace\Analyzer\Attributes\WebsiteRequestProcessor::__construct` `$contentMapper` (2nd argument) removed
 
+Classes implementing the `DocumentFixtureInterface` have to update the load method signature:
+```diff
+-    public function load(DocumentManager $documentManager): void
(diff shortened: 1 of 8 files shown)


alexander-schranz:
The document manager will be removed in 3.0 so there will be no changes for 3.0 branch in PHPCR / Document Manager namespaces.

alexander-schranz:
Why I think it would be good it would be always the interface. I think it is not worth a BC Break for 2.6 where every project need to adopt its fixtures loader for just a very few devs who tried to decorate the document manager.

/cc @chirimoya what do you think?

mamazu:
I agree that for 2.6 this would be too big of an ask but if it's already planned for removal then we might as well close it here.

Issue #7363, opened 2024-04-11
CKEditor: Activate abbr-Tag and lang-Attribute for more accessible content creation


datenwerk-at:
# Problem description

By default there is no possibility to explain abbreviations in RTE content. Further there is no possibility to mark parts of the RTE content written in another language. 

See:
https://www.w3.org/WAI/WCAG22/Understanding/language-of-parts.html (WCAG 2.2. AA)
https://www.w3.org/WAI/WCAG22/Understanding/abbreviations.html (WCAG 2.2. AAA)

# Proposed solution

Enable these features by default in the CKEditor of Sulu CMS.

alexander-schranz:
Think we could add:

 - https://ckeditor.com/docs/ckeditor5/latest/features/language.html

For the `abbr` I could not yet find a plugin for CKEditor 5 🤔

datenwerk-at:
There is a tutorial for a advanced CKEditor 5 Plugin for abbreviations https://ckeditor.com/docs/ckeditor5/latest/tutorials/abbreviation-plugin/abbreviation-plugin-level-1.html (all 3 steps of creating an advanced plugin)

alexander-schranz:
We already installed now the language plugin but it is not enabled by default. It requires some configuration like languages and so on. At currentl `2.6.*@dev` state it can be configured via our ckeditor registries: https://github.com/sulu/sulu/blob/2.5/src/Sulu/Bundle/AdminBundle/Resources/js/containers/CKEditor5/README.md

alexander-schranz:
Had a meeting today with the UX/UI team and briefed them about how we can make `<abbr>` and `[lang]` better visible for the Content Manager as the implementation of CKEditor is specially for the [lang] not visible in the text itself.

Creating an own plugin for abbrevation I would like to avoid and more then maybe check if CKEditor will have a abbr plugin core plugin itself. As maintaining a CKEditor plugin would require it to keep uptodate with the Major versions CKEditor always releases and it where about 37 major releases since we started using CKEditor 5 in Sulu 2.0.

datenwerk-at:
More important is the "language of part" thing because Level AA is required for more websites with the accessibility act from 2025. Sites of public institution already need it.
I'm happy that you startet with this so fast. :-)

alexander-schranz:
Our UI/UX Team created a design for both features. What do you think about the usability of this one:

![Image](https://github.com/user-attachments/assets/66e6757a-8357-4f11-a685-8f3bf67ca6bb)

PR #7367, opened 2024-04-15
Add language package for ckeditor

martinlagler:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Related issues/PRs | #7363
| License | MIT

#### What's in this PR?

Add ckeditor language plugin package. https://ckeditor.com/docs/ckeditor5/latest/features/language.html

#### Why?

So it's possible to add the language plugin to your editor toolbar, for better accessability.


alexander-schranz:
Installing the plugin does currently not hurt. We still need to check how we show the `[lang=de]` attribute for the user the CKEDitor solution I don't find has a good UX here as you need to click in the text and have that menu open to edit it.

PR #7156, opened 2023-09-11
Allow Symfony 7 versions of packages

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | part of #7413 
| License | MIT
| Documentation PR | sulu/sulu-docs# 

### What's in this PR?

Allow Symfony 7 versions of packages.

### Why?

The hardest part in Sulu 1.6 -> 2.0 was that it was not only a Sulu upgrade but also a Symfony upgrade (3.4 -> 4.4). We should avoid this when we go from 2.x -> 3.0 so we should sure the last version of 2.x will be compatible with Symfony 7, this would make upgrades a lot easier as Sulu projects can first upgrade PHP, then Symfony and then jump to the next Sulu version.

#### Symfony 6.4 errors

 - [x] doctrine-bridge 6.3 with validator 7.0: `Testing PHP Fatal error:  Declaration of Symfony\Bridge\Doctrine\Validator\DoctrineInitializer::initialize(object $object) must be compatible with Symfony\Component\Validator\ObjectInitializerInterface::initialize(object $object): void in /home/runner/work/sulu/sulu/vendor/symfony/doctrine-bridge/Validator/DoctrineInitializer.php on line 34` ( https://github.com/symfony/symfony/pull/51630 )
 - [x] sulu SecurityBundle and security-http 7.0 `PHP Fatal error:  Declaration of Sulu\Bundle\SecurityBundle\Security\AuthenticationEntryPoint::start(Symfony\Component\HttpFoundation\Request $request, ?Symfony\Component\Security\Core\Exception\AuthenticationException $authException = null) must be compatible with Symfony\Component\Security\Http\EntryPoint\AuthenticationEntryPointInterface::start(Symfony\Component\HttpFoundation\Request $request, ?Symfony\Component\Security\Core\Exception\AuthenticationException $authException = null): Symfony\Component\HttpFoundation\Response in /home/runner/work/sulu/sulu/src/Sulu/Bundle/SecurityBundle/Security/AuthenticationEntryPoint.php on line 25` 
 - [x] doctrine/persistence 3.0 support
    - [x] Fix deprecations https://github.com/sulu/sulu/pull/7134
    - [x] Fix merge deprecations in fixtures https://github.com/sulu/sulu/issues/7184
[...]

PR #7369, opened 2024-04-16
Fix PreviewKernel and RefreshCommand compatibility

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | yes
| Deprecations? | no 
| Fixed tickets | fixes #7156  
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Fix PreviewKernel and RefreshCommand compatibility.

#### Why?

On Symfony 7 this two classes currently errors.

### DIFF
--- a/src/Sulu/Bundle/ReferenceBundle/UserInterface/Command/RefreshCommand.php
+++ b/src/Sulu/Bundle/ReferenceBundle/UserInterface/Command/RefreshCommand.php
@@ -13,6 +13,7 @@ namespace Sulu\Bundle\ReferenceBundle\UserInterface\Command;
 
 use Sulu\Bundle\ReferenceBundle\Application\Refresh\ReferenceRefresherInterface;
 use Sulu\Bundle\ReferenceBundle\Domain\Repository\ReferenceRepositoryInterface;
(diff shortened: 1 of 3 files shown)


src/Sulu/Bundle/ReferenceBundle/UserInterface/Command/RefreshCommand.php
```diff
@@ -24,6 +25,7 @@
  *
  * @final
  */
+#[AsCommand(name: 'sulu:reference:refresh')]
 class RefreshCommand extends Command
 {
     protected static $defaultName = 'sulu:reference:refresh';
```
Prokyonn, line 31: @alexander-schranz can we remove this line then?

alexander-schranz: Not sure if its required for backwards compatibility. At current state when we remove it I would do it in own merge request in all commands.

Prokyonn: 👍  makes sense

Prokyonn (approved):

PR #7356, opened 2024-04-09
Fix prepend website cache class with preview

sauls:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no 
| BC breaks? | no 
| Deprecations? | no  
| Fixed tickets | fixes #7332
| Related issues/PRs | # 
| License | MIT
| Documentation PR | sulu/sulu-docs# 

#### What's in this PR?

Fix for https://github.com/sulu/sulu/issues/7332 , don't know if this is the right way. Additionally tried to run tests locally but had issues with db connection, did not looked deeper, did only unit test, but I guess functional test would make more sense. Additionally this bundle has dependency from core_bundle, so in tests somehow we need to put that, so I don't know if in monorepo cross dependency declaration is valid. @alexander-schranz

Issue #7332, opened 2024-03-27
FrankenPHP Preview cache exception


sauls:
| Q | A
| --- | ---
| Sulu Version | 2.5.13
| PHP Version | 8.3.3
| DB Version | MariaDB 11.0.2
| Browser Version | Google Chrome Version 123.0.6312.58 (Official Build) (64-bit)

# Actual Behavior
I followed https://sulu.io/blog/running-sulu-with-frankenphp article, and created dual Kernel class. Locally when running on `dev` environment everyting works. But running `prod` environment the error is generated.
```
{"level":"error","ts":1711489705.64349,"msg":"PHP Fatal error:  Cannot declare class websiteWebspaceCollectionCache, because the name is already in use in /app/var/cache/preview/prod/sulu/websiteWebspaceCollectionCache.php on line 21","syslog_level":"err"}
```

# Expected Behavior

Error should not be generated.

# Steps to Reproduce

Running on [FrankenPHP](https://hub.docker.com/r/dunglas/frankenphp/tags) docker image with `prod` environemnt, if you clear cache and go to view published article. Then if you try to view same article in the admin preview you get the error and wise versa if after clearing cache first you open admin with preview, then navigating to article page will throw this error.

# Possible Solutions

Maybe this is from [FrankenPHP](https://frankenphp.dev/) also, how the cache is loaded into memory and generated, or not generated or differently generated per workers. :man_shrugging: 

Running same configuration with nginx + php there is no issues at all.

alexander-schranz:
@sauls the issue seems to be here: https://github.com/sulu/sulu/blob/d635d8dcaa8679c4ff5ab3f1f52a6fa7ea3bc53a/src/Sulu/Bundle/CoreBundle/Resources/config/webspace.xml#L6

To avoid a conflict between Admin and Website Kernel we are using the `%sulu.context%`. But there is a third additional internal Kernel which is the PreviewKernel.

We require a compilerpass which check if it is the preview kernel `if (true === $container->getParameter('sulu.preview')) {` and prefix the `sulu_core.webspace.cache_class` with `preview`:

```
if (true === $container->getParameter('sulu.preview')) {
    $container->setParameter('sulu_core.webspace.cache_class', 'preview' + $container->getParameter('sulu_core.webspace.cache_class'));
}
```

@sauls do you want to give this fix a try?

alexander-schranz:
Closing this as fixed in #7356

PR #7370, opened 2024-04-17
Fix PreviewKernel compatibility for Projects overwritten the build method

tstauer:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | #7332 / #7356
| License | MIT
| Documentation PR | -

#### What's in this PR?

Call the parent build function in the PreviewKernel.

#### Why?

We have our own build function in our project Kernel and use Sulu.
With the advent of the PR 7356 the PreviewKernel overwrites our build function in App/Kernel of our project.
This is the place we registered our own Extension.

However because the PreviewKernel now overwrites this function using composer install does no longer work in our project. The reason is that the cache:clear steps fails because a configuration key in our project can no longer be identified due to the not registered Extension.

Please note that I don't have a FrankenPHP setup and therefore cannot verify how this PR affects the issue #7332.

### DIFF
--- a/src/Sulu/Bundle/PreviewBundle/Preview/Renderer/PreviewKernel.php
+++ b/src/Sulu/Bundle/PreviewBundle/Preview/Renderer/PreviewKernel.php
@@ -130,5 +130,6 @@ class PreviewKernel extends Kernel
     protected function build(ContainerBuilder $container)
     {
         $container->addCompilerPass(new RegisterPreviewWebspaceClassPass());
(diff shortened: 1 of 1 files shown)


alexander-schranz:
I think this should have no effects on FrankenPHP setup still @sauls can maybe check.

alexander-schranz:
@tstauer thank you!

PR #7378, opened 2024-04-23
Fix webspace view if no permissions for first webspace

### LINKED ISSUE #5359: False Webspace loaded when having no permission
| Q | A
| --- | ---
| Sulu Version | 2.1.0-RC1

# Actual Behavior

False Webspace loaded when not having no permission.

# Expected Behavior

The webspace which I have permissions for should be loaded

# Steps to Reproduce

1. Create a installation with 2 Webspace
2. Go to User Roles and give Permission only to second webspace (base.sulu.io, sulu.io (have permisson for this) )
3. Remove maybe exist user settings
> DELETE FROM se_user_settings WHERE settingsKey = 'sulu_page.webspace_tabs.webspace';
3. Reload Page so new Permission and user settings loaded
4. Click on Webspaces

Screenshot_2020-06-23 SULU

Now you will see the false webspace. As I have no permissions there I can't do anything.

# Possible Solutions

Load the first webspace which I have permission for instead of the first webspace which is available.

mamazu:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #5359 
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?
Getting the first webspace the user has access to instead of the first webspace in the collection.

#### Why?
The user might not have access to the webspace in the collection, so taking the first webspace could result in the user not seeing anything.

### DIFF
--- a/src/Sulu/Bundle/PageBundle/Admin/PageAdmin.php
+++ b/src/Sulu/Bundle/PageBundle/Admin/PageAdmin.php
@@ -101,7 +101,7 @@ class PageAdmin extends Admin
 
     public function configureNavigationItems(NavigationItemCollection $navigationItemCollection): void
     {
-        if ($this->hasSomeWebspacePermission()) {
+        if ($this->getFirstWebspaceWithPermissions() instanceof Webspace) {
             $webspaceItem = new NavigationItem('sulu_page.webspaces');
             $webspaceItem->setPosition(10);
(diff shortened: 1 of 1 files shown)


alexander-schranz:
@mamazu Thank you!

PR #7379, opened 2024-04-24
Fix condition provider crashing of the admin when no parent is found

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| License | MIT

#### What's in this PR?

When adding a new block the `settings` array is undefined, this MR prevents the admin from completely crashing in this case.

#### Why?

Admin ui should never crash completely due to JS errors

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/conditionDataProviders/parentConditionDataProvider.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/conditionDataProviders/parentConditionDataProvider.js
@@ -13,6 +13,12 @@ export default function(data: Object, dataPath: ?string): {[string]: any} {
 
     do {
         parentDataPath = parentDataPath.substring(0, parentDataPath.lastIndexOf('/'));
+
+        if (!jsonpointer.has(data, parentDataPath)) {
+            currentConditionData.__parent = null;
+            break;
+        }
+
         const evaluatedData = jsonpointer.get(data, parentDataPath);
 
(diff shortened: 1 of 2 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/conditionDataProviders/parentConditionDataProvider.js
```diff
@@ -13,6 +13,12 @@ export default function(data: Object, dataPath: ?string): {[string]: any} {
 
     do {
         parentDataPath = parentDataPath.substring(0, parentDataPath.lastIndexOf('/'));
+
+        if (!jsonpointer.has(data, parentDataPath)) {
+            currentConditionData.__parent = null;
```
Prokyonn, line 18: @alexander-schranz not sure if we should return here `null` or `data` like in line 7 when no `dataPath` is set

alexander-schranz: Lets keep it currently with `null`.

alexander-schranz (changes requested):
Do we need to target the 2.4 for this fix as it seems unrelated to 2.6 changes?

alexander-schranz:
A test would be nice to be added to avoid regression in the future.

alexander-schranz (approved):

Prokyonn:
@alexander-schranz added a test and rebased to 2.4 :)

PR #7381, opened 2024-04-24
Fix secret parameter in ResettingController

### LINKED ISSUE #7303: Incorrect argument for secret in ResetController
https://github.com/sulu/sulu/blob/62d7b1b26a5927c4c063fd00ab466bf8f16b48f7/src/Sulu/Bundle/SecurityBundle/Resources/config/services.xml#L64

Looks wrong to me. Think this should be only one `%`. Currently it may just added always a `%` to the secret.

_Originally posted by @alexander-schranz in https://github.com/sulu/sulu/pull/7257#discussion_r1514286982_

mamazu:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #7303
| Related issues/PRs | #7303
| License | MIT
| Documentation PR | -

#### What's in this PR?
Fixing the argument for the resetting controller.

#### Why?
The contents of this variable is not really used anywhere, it's used to generate a unique token for the user for resetting the password. Maybe since the secret is not checked anywhere any random information would be good enough.

### DIFF
--- a/src/Sulu/Bundle/SecurityBundle/Controller/ResettingController.php
+++ b/src/Sulu/Bundle/SecurityBundle/Controller/ResettingController.php
@@ -491,7 +491,7 @@ class ResettingController
      */
     private function generateTokenHash(string $token): string
     {
-        return \hash('sha1', $this->secret . $token);
+        return \hash('sha1', $this->secret . '%' . $token);
     }
 
     /**
--- a/src/Sulu/Bundle/SecurityBundle/Resources/config/services.xml
+++ b/src/Sulu/Bundle/SecurityBundle/Resources/config/services.xml
@@ -61,7 +61,7 @@
             <argument>%sulu_security.reset_password.mail.template%</argument>
             <argument>%sulu_security.reset_password.mail.token_send_limit%</argument>
             <argument>%sulu_admin.email%</argument>
-            <argument>%kernel.secret%%</argument>
+            <argument>%kernel.secret%</argument>
 
             <tag name="sulu.context" context="admin"/>
         </service>


src/Sulu/Bundle/SecurityBundle/Controller/ResettingController.php
```diff
@@ -491,7 +491,7 @@ private function generateTokenForUser(UserInterface $user)
      */
     private function generateTokenHash(string $token): string
     {
-        return \hash('sha1', $this->secret . $token);
+        return \hash('sha1', $this->secret .'%'. $token);
```
mamazu: The generated hash still has to stay the same otherwise all password reset tokens would be invalidated.

alexander-schranz:
> Maybe since the secret is not checked anywhere any random information would be good enough.

The generated part is not unused, the secret is used to load the user here: https://github.com/sulu/sulu/blob/3ad27aa96cc76398bb7058ecbdefe43c24c5e49a/src/Sulu/Bundle/SecurityBundle/Controller/ResettingController.php#L252

src/Sulu/Bundle/SecurityBundle/Controller/ResettingController.php
```diff
@@ -491,7 +491,7 @@ private function generateTokenForUser(UserInterface $user)
      */
     private function generateTokenHash(string $token): string
     {
-        return \hash('sha1', $this->secret . $token);
+        return \hash('sha1', $this->secret . '%' . $token);
```
alexander-schranz, line 494: sadly this will change the hash as the additional `%` was behind the token not in front of it. I will discuss this with @chirimoya if we do a bc break here or not.

mamazu: This shouldn't change the behavior. Before the change the percent sign was at the end of $this->secret now it's here. If we do a BC Break then we should not rely on the secret here.

mamazu:
> The generated part is not unused, the secret is used to load the user here:

Yes the generated token is used there but it doesn't validate the contents of the token. We could just as well put a random string in that column and it would also be valid.

mamazu:
fixes #7303

alexander-schranz:
@mamazu its validated by write = read token and if the token is changed between write and read its invalid and so a bc break. As all already generated tokens are not longer valid. That is what I mean with it is still validated and used. But as said will discuss with the team if we do this bc break or not.

mamazu:
I know, but the current merge request just moves where the "%" is added. Before it was added in the XML, now it's added in the generation process of the token. This shouldn't be a BC break as the logic hasn't changed.

PR #7382, opened 2024-04-24
Refactor ParameterBag in InputBag in Core Tests

TheCadien:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no 
| Related Tickets | #7156
| License | MIT

**Why?**

[ Symfony 6.4](https://github.com/symfony/http-foundation/blob/6.4/Request.php#L103) to  [Symfony 7](https://github.com/symfony/http-foundation/blob/7.0/Request.php#L97) change the `attribute` `request` and `querie` Types.

### DIFF
--- a/src/Sulu/Component/Webspace/Tests/Functional/Analyzer/RequestAnalyzerTest.php
+++ b/src/Sulu/Component/Webspace/Tests/Functional/Analyzer/RequestAnalyzerTest.php
@@ -28,6 +28,7 @@ use Sulu\Component\Webspace\Portal;
 use Sulu\Component\Webspace\PortalInformation;
 use Sulu\Component\Webspace\Webspace;
 use Symfony\Component\HttpFoundation\InputBag;
+use Symfony\Component\HttpFoundation\ParameterBag;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\RequestStack;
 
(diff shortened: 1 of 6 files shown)


src/Sulu/Component/Webspace/Tests/Functional/Analyzer/RequestAnalyzerTest.php
```diff
@@ -273,9 +273,9 @@ public function testAnalyzeNotExisting(): void
         $this->webspaceManager->getPortalInformations(Argument::any())->willReturn([]);
 
         $request = $this->getMockBuilder(Request::class)->getMock();
-        $request->request = new ParameterBag(['post' => 1]);
-        $request->query = new ParameterBag(['get' => 1]);
-        $request->attributes = new ParameterBag();
+        $request->request = new InputBag(['post' => 1]);
+        $request->query = new InputBag(['get' => 1]);
+        $request->attributes = new InputBag();
```
mamazu: This is still a `ParameterBag`. Then the tests should be fixed. The difference is the `InputBag` is stricter on what kind of values are set. You can only set scalars in it (see test) but in this case the attributes can also handle arrays.

TheCadien: oh yeah true !

mamazu (approved):

alexander-schranz (approved):

PR #7383, opened 2024-04-25
Improve command name consistency

Prokyonn:
#### What's in this PR?

This MR improves the command names to be more consistent with the other sulu commands.
Sulu commands should follow this naming convention `sulu:<bundle>:command-name`

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Command/ViewDebugCommand.php
+++ b/src/Sulu/Bundle/AdminBundle/Command/ViewDebugCommand.php
@@ -33,7 +33,7 @@ use Symfony\Component\Console\Style\SymfonyStyle;
 /**
  * @internal
  */
-#[AsCommand(name: 'sulu:admin:debug-views', description: 'Display current Views for the admin')]
+#[AsCommand(name: 'sulu:admin:debug-view', description: 'Display current Views for the admin')]
 class ViewDebugCommand extends Command
 {
     public function __construct(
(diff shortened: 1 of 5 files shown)


alexander-schranz (approved):

alexander-schranz (approved):

alexander-schranz:
Something looks strange in the container build / lint for the command:

Bildschirmfoto 2024-04-25 um 15 26 44

Prokyonn:
> Something looks strange in the container build / lint for the command:
> 
> Bildschirmfoto 2024-04-25 um 15 26 44

yes I removed an unused service from the constructor, but forgot to remove the entry in the xml configuration

src/Sulu/Bundle/AdminBundle/Command/ViewDebugCommand.php
```diff
@@ -33,7 +33,7 @@
 /**
  * @internal
  */
-#[AsCommand(name: 'sulu:debug:admin:views', description: 'Display current Views for the admin')]
+#[AsCommand(name: 'sulu:admin:debug-views', description: 'Display current Views for the admin')]
```
alexander-schranz: In the doc it is called `debug-view` shold we go with that?

https://github.com/sulu/sulu-docs/pull/793/files#diff-81b84d4fcee0a41917f5da6e64293

Other Symfony commands:

Bildschirmfoto 2024-04-25 um 18 30 49
99ea991108e2525d7b3c1ed4592031a656fR424-R431

Prokyonn: Oh, good catch I've completely overlooked that 🙈  personally I would go with singular, but I don't have a strong opinion

alexander-schranz: Lets go with the singular then

alexander-schranz (approved):

PR #7397, opened 2024-05-03
Add typehint to S3StorageTest class for Symfony 7 compatibility

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | #7156 
| License | MIT

#### What's in this PR?

Add typehint to S3StorageTest class for Symfony 7 compatibility

#### Why?

Symfony 7 added the same typehint.

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/Tests/Functional/Media/Storage/S3StorageTest.php
+++ b/src/Sulu/Bundle/MediaBundle/Tests/Functional/Media/Storage/S3StorageTest.php
@@ -24,6 +24,12 @@ class S3StorageTest extends SuluTestCase
         static::$class = S3Kernel::class;
     }
 
+    protected function tearDown(): void
+    {
+        static::$class = null;
+        parent::tearDown();
+    }
+
     public function testSave(): void
     {
         $kernel = self::bootKernel();
--- a/src/Sulu/Bundle/MediaBundle/Tests/Functional/Media/Storage/S3StorageTest.php
+++ b/src/Sulu/Bundle/MediaBundle/Tests/Functional/Media/Storage/S3StorageTest.php
@@ -19,7 +19,7 @@ use Sulu\Bundle\TestBundle\Testing\SuluTestCase;
 
 class S3StorageTest extends SuluTestCase
 {
-    protected static function setUp(): void
+    protected function setUp(): void
     {
         static::$class = S3Kernel::class;
     }
--- a/src/Sulu/Bundle/MediaBundle/Tests/Functional/Media/Storage/S3StorageTest.php
+++ b/src/Sulu/Bundle/MediaBundle/Tests/Functional/Media/Storage/S3StorageTest.php
@@ -19,7 +19,7 @@ use Sulu\Bundle\TestBundle\Testing\SuluTestCase;
 
 class S3StorageTest extends SuluTestCase
 {
-    protected static function setUp(): svoid
+    protected static function setUp(): void
     {
         static::$class = S3Kernel::class;
     }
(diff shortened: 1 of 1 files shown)


mamazu:
Why do we have to declare the variable in the test case? Can't we just set the value of it to our kernel?

alexander-schranz:
@mamazu because this test case requires an owm kernel because of using not a default media bundle configuration instead s3 configured container

mamazu:
But why can't we just use this to just set the value instead:
```php
self::$class = S3Kernel::class;
```

alexander-schranz:
Should work if we do it in setup. I will try it.

src/Sulu/Bundle/MediaBundle/Tests/Functional/Media/Storage/S3StorageTest.php
```diff
@@ -19,14 +19,9 @@
...
-    /**
-     * @var string
-     */
-    protected static $class = S3Kernel::class;
-
-    protected static function getKernelClass(): string
+    protected function setUp(): void
     {
-        return S3Kernel::class;
+        static::$class = S3Kernel::class;
```
mamazu, line 24: Are you missing a call to `parent::setUp()`?

alexander-schranz: there is no parent setUp functionality but setting the static values has effects to the next class so we need in teardown reset it.

src/Sulu/Bundle/MediaBundle/Tests/Functional/Media/Storage/S3StorageTest.php
```diff
@@ -19,14 +19,15 @@
 
 class S3StorageTest extends SuluTestCase
 {
-    /**
-     * @var string
-     */
-    protected static $class = S3Kernel::class;
```
alexander-schranz, line 25: Symfony 7 has this typehinted with `string` but to support Symfony 5 and 6 we can not just add a typehint as that is not allowed to change a variable from `mixed` -> `string`. So we need to change how we set the kernel class and how we reset it.

mamazu:
It's I think the better idea. That way we're a little more independent from the types of the Symfony testcases.

Prokyonn (approved):

src/Sulu/Bundle/MediaBundle/Tests/Functional/Media/Storage/S3StorageTest.php
```diff
@@ -19,14 +19,15 @@
...
+    protected function setUp(): void
+    {
+        static::$class = S3Kernel::class;
+    }
 
-    protected static function getKernelClass(): string
+    protected function tearDown(): void
     {
-        return S3Kernel::class;
+        static::$class = null;
```
Prokyonn, line 29: @alexander-schranz this is already done in the [KernelTestCase](https://github.com/symfony/symfony/blob/7.1/src/Symfony/Bundle/FrameworkBundle/Test/KernelTestCase.php#L37)

alexander-schranz: Class is missed on lowest version: https://github.com/symfony/symfony/blob/v5.4.0/src/Symfony/Bundle/FrameworkBundle/Test/KernelTestCase.php#L49 

Which did make the test fail here: https://github.com/sulu/sulu/actions/runs/8959134992/job/24604192505

Issue #6988, opened 2023-02-04
PHPUnit 10 Upgrade


alexander-schranz:
# Problem description

We can Upgrade Sulu 2.5 to PHPUnit 10 as Sulu 2.5 already requires atleat PHP 8.1.

# Proposed solution

Upgrade to PHPUnit 10 with `rector`:

 - https://getrector.com/blog/how-to-upgrade-to-phpunit-10-in-diffs
 - https://phpunit.de/announcements/phpunit-10.html

mamazu:
In the current version of the `composer.json` in the 2.5 branch 8.0 is still allowed.

alexander-schranz:
@mamazu good catch. Then maybe target 2.6 😄

mamazu:
Currently this isn't possible. The `phpspec/prophecy` package doesn't yet support phpunit 10, I just checked.

alexander-schranz:
For preparing maybe we could test out the current PHPUnit 10 branch/pull request of prophecy to give maybe prophecy already feedback if it works with PHPUnit 10 or not.

What already could be targetted are the deprecations like:

 - assertObjectHasAttribute() is deprecated and will be removed in PHPUnit 10.
 - Expecting E_WARNING and E_USER_WARNING is deprecated and will no longer be possible in PHPUnit 10.
 
Currently we can test PHPUnit 10 with:

```
    "require-dev": {
        "phpspec/prophecy": "^1.18@dev",
        "phpspec/prophecy-phpunit": "^2.1@dev",
        "phpunit/phpunit": "^10.2"
    }
```

alexander-schranz:
@TheCadien said he will tackle this :)

PR #7401, opened 2024-05-06
Remove unnecessary test that checks if resource is seekable

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | #6988 
| License | MIT
| Documentation PR | -

#### What's in this PR?
Removing a surperflous test.

#### Why?
In private I discussed with @alexander-schranz how we proceed with this test case seeing that it breaks in PHPUnit 10 and we agreed to remove it. 

The createUnseekableResource function already checks if the resource is unseekable


src/Sulu/Bundle/MediaBundle/Tests/Unit/Media/ImageConverter/MediaImageExtractorTest.php
```diff
@@ -50,24 +50,6 @@ public function setUp(): void
...
-        $this->expectWarning();
-
-        if (\PHP_VERSION_ID >= 80000) {
-            $this->expectWarningMessage('mime_content_type(): Stream does not support seeking');
-        } else {
-            $this->expectWarningMessage('mime_content_type(): stream does not support seeking');
-        }
-
-        $this->mediaImageExtractor->extract($resource);
-    }
```
alexander-schranz, line 69: The `createUnseekableResource` make sure its unseekable resource so we need not test if the warning is thrown. This make the PHPUnit 10 update easier.

alexander-schranz (approved):

Issue #7402, opened 2024-05-06
Using Symfony tagged_iterator instead of compiler passes


mamazu:
# Problem description
Currently Sulu reimplements the [Symfony DI `tagged_iterator` functionality](https://gitlab.dev-b24.de/sylius/sylius/-/merge_requests/2094) for many services. While this is a good idea if you want to have custom logic, this makes no sense in this case. Affected services would be:
- [x] src/Sulu/Component/Symfony/CompilerPass/TaggedServiceCollectorCompilerPass.php (#7424)
- [x] MarkupBundle/DependencyInjection/CompilerPass/TagCompilerPass.php
- [x] MarkupBundle/DependencyInjection/CompilerPass/ParserCompilerPass.php (#7403)
- [x] RouteBundle/DependencyInjection/RouteGeneratorCompilerPass.php
- [x] MediaBundle/DependencyInjection/FormatCacheClearerCompilerPass.php (#8037)
- [x] AdminBundle/DependencyInjection/Compiler/AddAdminPass.php (#8034)
- [x] MediaBundle/DependencyInjection/ImageTransformationCompilerPass.php (#8306)
- [x] PageBundle/DependencyInjection/Compiler/StructureExtensionCompilerPass.php
- [x] CoreBundle/DependencyInjection/Compiler/RegisterContentTypesCompilerPass.php
- [x] AdminBundle/DependencyInjection/Compiler/AddMetadataProviderPass.php (#8038 )
- [x] AudienceTargetingBundle/DependencyInjection/Compiler/AddRulesPass.php (#7405)
- [x] ~DocumentManagerBundle/DependencyInjection/Compiler/InitializerPass.php (#7724)~
- [x] PageBundle/DependencyInjection/Compiler/SmartContentDataProviderCompilerPass.php
- [x] CoreBundle/DependencyInjection/Compiler/ListBuilderMetadataProviderCompilerPass.php
- [x] SecurityBundle/DependencyInjection/Compiler/AccessControlProviderPass.php (#8039 )
- [x] CoreBundle/DependencyInjection/Compiler/RegisterLocalizationProvidersPass.php (#8035)

# Proposed solution
Use the constructor and call the function from there or make it immutable. So that you can't add services dynamically.

# Future ideas
Maybe in the future we could replace this tagging logic with attributes like this:
```php

#[AsLocalizationProvider]
class CustomLocalizationProvider {}
```

alexander-schranz:
[...]

PR #7403, opened 2024-05-06
Using `tagged_iterator`s for the markup listener

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | yes 
| Fixed tickets | -
| Related issues/PRs | #7402 
| License | MIT
| Documentation PR | -

#### What's in this PR?
Using the tagged iterator concept to reduce the amount of code that sulu has to create.

#### Why?
See linked issue.

#### Example Usage
```xml
<argument type="tagged_iterator" tag="testing" />
```

More info on [Symfony Tagged Services](https://symfony.com/doc/current/service_container/tags.html)

#### To Do

### DIFF
--- a/src/Sulu/Bundle/MarkupBundle/DependencyInjection/CompilerPass/ParserCompilerPass.php
+++ b/src/Sulu/Bundle/MarkupBundle/DependencyInjection/CompilerPass/ParserCompilerPass.php
@@ -17,6 +17,10 @@ use Symfony\Component\DependencyInjection\Reference;
 
 /**
  * Collects all parsers for markup.
+ *
+ * @internal
+ *
+ * @deprecated since version 2.6 and will be removed in 3.0. Use tagged_iterator instead.
  */
 class ParserCompilerPass implements CompilerPassInterface
 {
--- a/src/Sulu/Bundle/MarkupBundle/Listener/MarkupListener.php
+++ b/src/Sulu/Bundle/MarkupBundle/Listener/MarkupListener.php
@@ -22,10 +22,16 @@ use Symfony\Component\HttpKernel\KernelEvents;
 class MarkupListener implements EventSubscriberInterface
 {
     /**
-     * @param array<string, MarkupParserInterface> $markupParser
+     * @var array<string, MarkupParserInterface>
      */
-    public function __construct(private array $markupParser)
+    private array $markupParser;
+
+    /**
+     * @param iterable<string, MarkupParserInterface> $markupParser
+     */
+    public function __construct(iterable $markupParser)
     {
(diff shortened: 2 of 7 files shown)


src/Sulu/Bundle/MarkupBundle/DependencyInjection/CompilerPass/ParserCompilerPass.php
```diff
@@ -13,36 +13,28 @@
...
 {
+    /**
+    * @deprecated Don't use this constant anymore use tagged_iterator
+    */
     public const SERVICE_ID = 'sulu_markup.response_listener';
 
+    /**
+    * @deprecated Use the value instead
+    */
     public const TAG_NAME = 'sulu_markup.parser';
```
mamazu, line 29: Maybe we could move this to the `ParserInterface` as a constant. Could be helpful for people who use PHP to Configure their app?

mamazu:
@alexander-schranz This should be BC break free now. We can still remove those classes in 3.0 but that's a different MR.

mamazu:
@mamazu  This should be using a tagged_locator instead because we do not need all services here.

src/Sulu/Bundle/MarkupBundle/Tests/Unit/Listener/MarkupListenerTest.php
```diff
@@ -72,7 +73,25 @@ protected function setUp(): void
...
+            public function get($id)
+            {
+                return $this->services[$id];
+            }
+
+            public function has($id): bool
+            {
+                return \array_key_exists($id, this->services);
+            }
+        };
```
mamazu: Maybe this should be generalized or we could pull this from the container if it's a `KernelTestCase`

src/Sulu/Bundle/MarkupBundle/Listener/MarkupListener.php
```diff
@@ -21,10 +22,7 @@
  */
 class MarkupListener implements EventSubscriberInterface
 {
-    /**
-     * @param array<string, MarkupParserInterface> $markupParser
-     */
-    public function __construct(private array $markupParser)
+    public function __construct(private ContainerInterface $markupParser)
```
alexander-schranz: this a bc break on 2.6 would need to be a tagged iterator instead of tagged locator

mamazu:
This is the last one. 🎉

alexander-schranz (approved):

alexander-schranz:
@mamazu Thank you, merged also to 3.0 so you can tackle the removal.

PR #7362, opened 2024-04-11
Fix shadow migration with multiple locales

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? |  yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT
| Documentation PR | sulu/sulu-docs# 

#### What's in this PR?

Fix shadow migration with multiple locales. See https://github.com/sulu/sulu/blob/601f8d359036adf165247f55cd2d960f279376f0/src/Sulu/Component/Content/Document/Subscriber/ShadowCopyPropertiesSubscriber.php#L27

#### Why?

It currently errors with PathNotFoundException.

PR #7408, opened 2024-05-07
Fix shadow migration with multiple locales

ghost:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| License | MIT

#### What's in this PR?

Same fix as in https://github.com/sulu/sulu/pull/7362 but for Version202404051700


alexander-schranz (approved):

alexander-schranz:
@flogado good catch. Thank you!

PR #7410, opened 2024-05-07
Fix WebsiteController::renderPreview to call own renderView method

ghost:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

We have overridden the `render` method in our project. However, when the `PreviewController` calls `parent::render()` directly, our method is skipped in the preview. I see no reason why the parent should be called directly here and have changed it accordingly.


alexander-schranz (approved):

alexander-schranz:
@flogado thank you!

PR #7411, opened 2024-05-07
Add section to upgrade about new reserved directory

### LINKED ISSUE #7409: [BC Break] Sulu 2.6: Possible bug and/or documentation needed? How to migrate to new global blocks?
| Q | A
| --- | ---
| PHP Version | 8.2 (several versions)
| DB Version | MariaDB 10.3.39
| Browser Version | Chrome 124.0.6367.91

# Actual Behavior

After upgrading to Sulu 2.6 my included block types lead to different errors (on cache clear or in browser-console):

I have nested blocks that are all on their own xml-file. Like `blocks/block_text.xml`. Those are `type` elements with a `name` attribute like `name="block_text"` and they are included within the `types` node of other templates via `<xi:include href="blocks/block_text.xml"/>`. That worked well until Sulu 2.5:

```` xml
<?xml version="1.0" ?>
<type xmlns="http://schemas.sulu.io/template/template"
      name="block_text">
  <meta>
    <title lang="en">Textblock</title>
  </meta>
  <properties>
    <property name="text" type="text_area">
      <meta>
      </meta>
    </property>
  </properties>
</type>
````

```` xml
<!-- … -->
  <types>
    <xi:include href="blocks/block_text.xml"/>
  </types>
<!-- … -->
````

After upgrading to Sulu 2.6, I get an error on cache clear with that:

>  [ERROR 1845] Element '{http://schemas.sulu.io/template/template}type': No matching global declaration available for the validation root. (in D:\Projects\sulu.dev.lan/config/templates/blocks\block_text.xml - line 5, column 0)

Interestingly, that error does not raise for elements (instead of blocks) that are defined and included the exact same way and it also does not raise, when I copy the type definition directly into its parent instead of including it via XInclude. **This is my workaround for now**.

Since the blocks have changed in Sulu 2.6 to global blocks, I then changed their definition according to the current documentation in https://docs.sulu.io/en/2.6/book/templates.html#using-global-blocks to the following code:

```` xml
<?xml version="1.0" ?>
  <key>block_text</key>
  <meta>
    <title lang="en">Textblock</title>
  </meta>
  <properties>
    <property name="text" type="text_area">
      <meta>
      </meta>
    </property>
  </properties>
</template>
````

```` xml
<!-- … -->
  <types>
    <type ref="block_text"/>
  </types>
<!-- … -->
````

That removes the error on cache clear, but now, I cannot edit pages containing that block type any more. Instead I get the following error inside the browser console:

> uncaught (in promise) Error: can't resolve reference #/definitions/block_text from id #

Adding an name-attribute to the ´type´ node with the `ref` leads to an xml error, as that `ref` and `name` are not allowed together.

# Expected Behavior

I expected either the old `xi:include` way to still work for included block types or the new global block handling to work with my existing content. Or at least a documented way to migrate my existing content to work with the new global blocks.

# Steps to Reproduce

Have a Sulu 2.5 instance with nested block based content using xi:include and upgrade to Sulu 2.6. We have two problems here: The first is the one with the new `xi:include` error for included block types and the other one that the needed migration for existing content that is not documented.

# Possible Solutions

I don't know how to solve that, but I guess that the existing content has to be migrated to be compatible to the new global block handling. Some documentation on how to do that would be helpful. Better would be to change the implementation in a way that no migration would be needed at all when the blocks keep their name.

For now, my workaround is to not include my block types via `xi:include`, but to copy their definitions directly into the templates using them.

### PR DESCRIPTION
**Pull request by alexander-schranz**
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | fixes #7409 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Add section to upgrade about new reserved directory.

#### Why?

See #7409

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -62,6 +62,15 @@ return [
 ];
 ```
 
+### New Reserved Templates directory for Global Blocks
+
+With the introduction of the new [Global Blocks](https://docs.sulu.io/en/2.6/book/templates.html#templates-global-blocks) there
+is a new reserved directory `config/templates/blocks`.
+
+If you already did use that directory for `xi:includes` your existing blocks to `config/templates/includes/blocks`.
+Migrating to Global Blocks is not required and can be done step by step if you want to use the new Global Blocks
+feature. Have a look at the Global Blocks [Documentation](https://docs.sulu.io/en/2.6/book/templates.html#templates-global-blocks).
+
 ### Custom Admin Builds npm version changed
 
 Sulu 2.6 now supports [npm 8, 9, and 10](https://nodejs.org/en/download), 
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -67,7 +67,7 @@ return [
 With the introduction of the new [Global Blocks](https://docs.sulu.io/en/2.6/book/templates.html#templates-global-blocks) there
 is a new reserved directory `config/templates/blocks`.
 
-If you already did use that directory for `xi:includes` your existing blocks to `config/templates/includes/blocks`.
+If you already did use that directory for `xi:includes` you should move the existing blocks to `config/templates/includes/blocks`.
 Migrating to Global Blocks is not required and can be done step by step if you want to use the new Global Blocks
 feature. Have a look at the Global Blocks [Documentation](https://docs.sulu.io/en/2.6/book/templates.html#templates-global-blocks).
 

### CODE REVIEW

File: UPGRADE.md [hunk 1]
```diff
@@ -62,6 +62,15 @@ return [
 ];
 ```
 
+### New Reserved Templates directory for Global Blocks
+
+With the introduction of the new [Global Blocks](https://docs.sulu.io/en/2.6/book/templates.html#templates-global-blocks) there
+is a new reserved directory `config/templates/blocks`.
+
+If you already did use that directory for `xi:includes` your existing blocks to `config/templates/includes/blocks`.
```
```suggestion
If you already did use that directory for `xi:includes` you should move the existing blocks to `config/templates/includes/blocks`.
```

Prokyonn (approved):

PR #7412, opened 2024-05-07
Replacing the mocks in phpunit dataproviders with actual objects

### LINKED ISSUE #7406: Convert existing PHPUnit DataProvider to not longer use mocks
# Problem description

Currently some data providers use mocks which is not supported or even suggested by PHPUnit / Prophecy: https://github.com/phpspec/prophecy-phpunit/pull/56#issuecomment-1693155549

To avoid problem with the PHPUnit 10 upgrade we should convert them to use not longer any mocks. And create the models / entity ourself and avoid mocks in this data providers.

# State of mocking

Currently we will keep prophecy as mocking library but will try to use mocks only when really necessary (e.g. RepositoryInterface mocks), which should not be any problem as such service calls should not be part of data providers.

# Proposed solution

Some refactoring like this can be used to avoid mocks:

```diff
    public function provideTargetGroup()
    {
-        $targetGroup1 = $this->prophesize(TargetGroupInterface::class);
+        $targetGroup1 = new TargetGroup();

-        $targetGroup2 = $this->prophesize(TargetGroupInterface::class);
-        $targetGroup2->getId()->willReturn(2);
+        $targetGroup2 = new TargetGroup();
+        static::setPrivateProperty($targetGroup2, 'id', 2);

-        $targetGroup3 = $this->prophesize(TargetGroupInterface::class);
-        $targetGroup3->getId()->willReturn(3);
+        $targetGroup3 = new TargetGroup();
+        static::setPrivateProperty($targetGroup3, 'id', 3);

-        $targetGroup4 = $this->prophesize(TargetGroupInterface::class);
-        $targetGroup4->getId()->willReturn(4);
+        $targetGroup4 = new TargetGroup();
+        static::setPrivateProperty($targetGroup4, 'id', 4);

        return [
-            ['X-Sulu-Target-Group-Hash', null, $targetGroup1->reveal(), null],
+            ['X-Sulu-Target-Group-Hash', null, $targetGroup1, null],
-            ['X-Sulu-Target-Group-Hash', null, $targetGroup2->reveal(), 2],
+            ['X-Sulu-Target-Group-Hash', null, $targetGroup2, 2],
-            ['X-Sulu-Target-Group', null, null, 0],
+            ['X-Sulu-Target-Group', null, null, 0],
-            ['X-Sulu-Target-Group', $targetGroup3->reveal(), $targetGroup4->reveal(), 4],
+            ['X-Sulu-Target-Group', $targetGroup3, $targetGroup4, 4],
        ];
    }
```

The target branch for this refactoring should be 2.5 to avoid merge conflicts with backmerge 2.5 -> 2.6.

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | fixes #7406 
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?
Replacing the mock object with real objects in the dataproviders

#### Why?
For the phpunit 10 upgrade and mocking hides implementation problems.


src/Sulu/Bundle/PageBundle/Content/Structure/TestExtension.php
```diff
@@ -0,0 +1,33 @@
...
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\PageBundle\Content\Structure;
+
+use PHPCR\NodeInterface;
+use Sulu\Component\Content\Extension\AbstractExtension;
+
+class TestExtension extends AbstractExtension
```
alexander-schranz: would not create a `TestExtension` class here instead use anonymous class inside the specific test case: e.g.:

```php
private static function createTestExtension(): AbstractExtension
{
     return new class extends AbstractExtension() {
         // ..
     };
}
```

alexander-schranz:
We should target the 2.5 to make backmerge of bugfixes easier.

src/Sulu/Component/Contact/Tests/Unit/SmartContent/AccountDataProviderTest.php
```diff
@@ -234,21 +238,21 @@ private function getRepository(
...
+        self::setPrivateProperty($entity, 'id', $id);
+        self::setPrivateProperty($entity, 'number', $id);
+        self::setPrivateProperty($entity, 'name', $name);
+        self::setPrivateProperty($entity, 'tags', new ArrayCollection($tags));
+        self::setPrivateProperty($entity, 'placeOfJurisdiction', '');
+        self::setPrivateProperty($entity, 'uid', '');
+        self::setPrivateProperty($entity, 'corporation', '');
+        self::setPrivateProperty($entity, 'created', new \DateTime());
+        self::setPrivateProperty($entity, 'changed', new \DateTime());
+        self::setPrivateProperty($entity, 'medias', new ArrayCollection([]));
```
mamazu: This is highly in efficient because we are reflecting on this class with every method call. But it's just the tests.

alexander-schranz: @mamazu for properties providing a setter I would directly call the setter method and not go over reflection. So in our case the reflection mostly only used for the primary keys.

mamazu: Fixed. The only place I have also left with the function is the cache in the `MediaDataProvider`.

alexander-schranz (approved):

alexander-schranz:
@mamazu Thank you!

PR #7423, opened 2024-05-14
Fix nested global blocks

### LINKED ISSUE #7409: [BC Break] Sulu 2.6: Possible bug and/or documentation needed? How to migrate to new global blocks?

### LINKED ISSUE #7414: A global block in a global block leads to 500 errors
| Q | A
| --- | ---
| Sulu Version | 2.6.0
| PHP Version | 8.3

# Actual Behavior

If a global block (e.g. `foo`) is loaded into another global block (e.g. `bar`), this results in a 500 error for `/admin/config` due to a recursion. This was not a problem with includes.

# Steps to Reproduce

I have created a small example (with text_block and text_block**2**):
https://github.com/flogado/sulu-global-blocks/commit/adaab4a398e975de089e9265a019f0f1f74ff26c

# Possible Solutions

The recursion seems to happen in the `GlobalBlocksTypedFormMetadataVisitor::getGlobalBlockMetadata`.

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #7414 , fixes #7409
| Related issues/PRs | #7229 
| License | MIT
| Documentation PR | none

#### What's in this PR?

This PR fixed following issues:

* #7414
* #7409

#### Why?

This issues where introduced by #7229

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Tests/Functional/Controller/AdminControllerTest.php
+++ b/src/Sulu/Bundle/AdminBundle/Tests/Functional/Controller/AdminControllerTest.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 /*
  * This file is part of Sulu.
  *
@@ -22,7 +24,7 @@ class AdminControllerTest extends SuluTestCase
      */
     private $client;
 
-    public function setUp(): void
+    protected function setUp(): void
     {
         $this->client = $this->createAuthenticatedClient();
         $this->purgeDatabase();
@@ -50,10 +52,10 @@ class AdminControllerTest extends SuluTestCase
         $this->assertIsObject($response->sulu_admin->resources);
         $this->assertTrue(\property_exists($response, 'sulu_preview'));
 
-        $this->assertEquals('en', $response->sulu_admin->localizations[0]->localization);
-        $this->assertEquals('en_us', $response->sulu_admin->localizations[1]->localization);
-        $this->assertEquals('de', $response->sulu_admin->localizations[2]->localization);
-        $this->assertEquals('de_at', $response->sulu_admin->localizations[3]->localization);
+        $this->assertSame('en', $response->sulu_admin->localizations[0]->localization);
+        $this->assertSame('en_us', $response->sulu_admin->localizations[1]->localization);
+        $this->assertSame('de', $response->sulu_admin->localizations[2]->localization);
+        $this->assertSame('de_at', $response->sulu_admin->localizations[3]->localization);
     }
 
     public function testGetConfigWithFallbackNonExistUserLocale(): void
@@ -81,10 +83,10 @@ class AdminControllerTest extends SuluTestCase
         $this->assertIsObject($response->sulu_admin->resources);
         $this->assertTrue(\property_exists($response, 'sulu_preview'));
 
-        $this->assertEquals('en', $response->sulu_admin->localizations[0]->localization);
-        $this->assertEquals('en_us', $response->sulu_admin->localizations[1]->localization);
-        $this->assertEquals('de', $response->sulu_admin->localizations[2]->localization);
-        $this->assertEquals('de_at', $response->sulu_admin->localizations[3]->localization);
+        $this->assertSame('en', $response->sulu_admin->localizations[0]->localization);
+        $this->assertSame('en_us', $response->sulu_admin->localizations[1]->localization);
+        $this->assertSame('de', $response->sulu_admin->localizations[2]->localization);
+        $this->assertSame('de_at', $response->sulu_admin->localizations[3]->localization);
     }
 
     public function testTemplateConfig(): void
@@ -102,7 +104,7 @@ class AdminControllerTest extends SuluTestCase
 
         // extract json from html
         $applicationElement = $crawler->filter('#application')->first();
-        $this->assertEquals(1, $applicationElement->count(), 'Failed getting the application element');
+        $this->assertCount(1, $applicationElement, 'Failed getting the application element');
         $configJson = $applicationElement->attr('data-config');
         $this->assertIsString($configJson, 'Failed to retrieve the configuration json');
         $config = \json_decode($configJson, true);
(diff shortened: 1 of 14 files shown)


mamazu:
Now when I expand a global block type the list of properties is empty.

wachterjohannes:
@mamazu seems to work for me pretty well:

image

Have you rebuild the admin dist files?

wachterjohannes:
@mamazu see comment in the issue https://github.com/sulu/sulu/issues/7409#issuecomment-2109840871

mamazu:
Yes. Oddly there are also no console errors this time when I expand a block which is empty. So no clue what's happening.

wachterjohannes:
@mamazu does it work now?

spackmat:
@wachterjohannes I also get no console errors now, bus still does not work.

wachterjohannes:
@spackmat only to be clear - after updating the vendor to my branch - you have to remove the node_modules folder and run `npm install && npm run build` to be sure that my new files are used

mamazu:
@wachterjohannes Still no change. Even ran the Sulu admin build update command.

ghost:
Works for me, but there is an error with local and global blocks combined:
https://github.com/sulu/sulu/issues/7414#issuecomment-2109979442

spackmat:
@wachterjohannes did it all again for you, still the same: When the blocks are within a section, there are no fields displayed. Outside of a section it works well for normal blocks, nested blocks do not work at all, showing the wrong name.

Looks like this (opened), when within a section:
image

And should look like this (when outside of a section:
image

And looks like this (without section) for nested blocks:
image
(outside of the nested block, the block is correctly named "Studienverlaufsplan", when not nested)

wachterjohannes:
@spackmat me and @mamazu found the issue and pushed it to the branch - can you recheck?

spackmat:
@wachterjohannes @mamazu Hey, the non nested blocks within sections do work now 👏

The nested blocks still fail the same way as before.

spackmat:
@wachterjohannes with your last commit, the nested blocks also work. Nice! 😎

wachterjohannes:
@spackmat thanks for the testing :)

@mamazu will you give it a try again?

FlorianKoerner:
@wachterjohannes Looks good - even with nested blocks. Thank you! :)

src/Sulu/Bundle/AdminBundle/Tests/Functional/Controller/AdminControllerTest.php
```diff
@@ -167,4 +167,48 @@ public function testGetMetaDataKeysOnly(): void
...
+        $response = $this->client->getResponse();
+        $this->assertHttpStatusCode(200, $response);
+        $json = $response->getContent();
+        $this->assertIsString($json);
+
+        $metaData = \json_decode($json, true, 512, \JSON_THROW_ON_ERROR);
+
+        $defaultMetadata = $metaData['types']['default'];
+        $defaultSchema = $defaultMetadata['schema'];
+        $this->assertArrayHasKey('text_block', $defaultSchema['definitions']);
```
alexander-schranz: Why not do:

```
$this->assertSame($expectedSchema, $defaultMetadata['schema']);
```

Would show us unexpected changes in the schema and you don't have the new the phpstan issues.

phpstan-baseline.neon
```diff
@@ -1455,6 +1455,81 @@ parameters:
 			count: 1
 			path: src/Sulu/Bundle/AdminBundle/Serializer/Subscriber/TogglerToolbarActionSubscriber.php
 
+		-
+			message: "#^Cannot access offset 'block' on mixed\\.$#"
+			count: 2
+			path: src/Sulu/Bundle/AdminBundle/Tests/Functional/Controller/AdminControllerTest.php
```
alexander-schranz: we should avoid any new code to be added to the baseline

mamazu:
Works for me.

src/Sulu/Bundle/AdminBundle/Tests/Functional/Controller/AdminControllerTest.php
mamazu: I think at this point we might as well put it into a json file and compare results that way. This test is super unreadable. :D

alexander-schranz (approved):

src/Sulu/Bundle/AdminBundle/Tests/Functional/Controller/AdminControllerTest.php
```diff
@@ -102,7 +104,7 @@ public function testTemplateConfig(): void
 
         // extract json from html
         $applicationElement = $crawler->filter('#application')->first();
-        $this->assertEquals(1, $applicationElement->count(), 'Failed getting the application element');
+        $this->assertCount(2, $applicationElement, 'Failed getting the application element');
```
alexander-schranz: This line errors currently 🤔

alexander-schranz:
Thank you all for testing!

PR #7432, opened 2024-05-16
Remove SymfonyTestsListener configuration

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no 
| BC breaks? | no 
| Deprecations? | no
| Fixed tickets | fixes 
| Related issues/PRs | related #7339 
| License | MIT
| Documentation PR | 

#### What's in this PR?
Removing SymfonyTestsLister since it's being removed in PhpUnit 10 anyways.

#### Why?
see above

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -31,7 +31,4 @@
         <ini name="date.timezone" value="UTC"/>
     </php>
 
-    <listeners>
-        <listener class="Symfony\Bridge\PhpUnit\SymfonyTestsListener"/>
-    </listeners>
 </phpunit>
--- a/src/Sulu/Bundle/ActivityBundle/phpunit.xml.dist
+++ b/src/Sulu/Bundle/ActivityBundle/phpunit.xml.dist
@@ -27,7 +27,4 @@
         <ini name="date.timezone" value="UTC"/>
     </php>
 
-    <listeners>
-        <listener class="Symfony\Bridge\PhpUnit\SymfonyTestsListener"/>
-    </listeners>
 </phpunit>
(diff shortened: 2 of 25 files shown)


mamazu:
Why do the tests break when I change the XML?

alexander-schranz:
@mamazu I don't know maybe some tests relay on some functionality of the bridge: https://symfony.com/doc/current/components/phpunit_bridge.html.

But as a I know there should be for all things a replacement in PHPUnit itself or a way to workaround it https://thephp.cc/presentations/phpunit-10-for-symfony-developers

Lets check if we can change the two tests to work again on 2.5 without the listener.

mamazu:
It looks like the problem is "perserveGlobalState". This should be off for isolated processes anyways (see: https://github.com/sebastianbergmann/phpunit/issues/3954) which is probably why it runs in phpunit 10.

So what to do: 
* But for the old phpunit 9 we would have to find a way to generate a listener that disabled this feature for all tests
* Introduce a base `setUp` method in the SuluTestcase (which could be removed when migrating to phpunit 10 anways).

But I'm open for suggestions. The line that is currently fixing that is here:
https://github.com/symfony/symfony/blob/c1bca297cf15145fff29ee169e9820ffacc5d37d/src/Symfony/Bridge/PhpUnit/Legacy/SymfonyTestsListenerTrait.php#L121-L123

alexander-schranz:
@mamazu if I understand correctly adding `@preserveGlobalState disabled` to the two tests should fix the issue. So I would go with that one and remove `@preserveGlobalState disabled` when we upgrade to PHPUnit 10 / 11.

mamazu:
Oh yeah that would be too easy. :D Did it like that.

alexander-schranz:
@mamazu Thank you!

Issue #7325, opened 2024-03-23
Updating the codebase to php 8


mamazu:
# Problem description
We want to use all the new and good stuff from PHP8. But not make any BC breaking changes.

- [x] Using match
- [x] Constructor Property Promotion
- [ ] New string function (str_starts_with, str_ends_with, str_contains)
- [ ] Using the new `fn` callback syntax

# Proposed solution
To reduce amount of errors most of the refactoring should be done with rector.

mamazu:
The rest of the rules don't make sense for Sulu as it wouldn't improve the code at all.

PR #7435, opened 2024-05-18
Use constructor property promotion for ContactBundle classes

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | #7325 
| License | MIT
| Documentation PR | -

#### What's in this PR?
Using constructor property promotion for the Contact Bundle.

#### Why?
Less code = Better

### DIFF
--- a/src/Sulu/Bundle/ContactBundle/Contact/ContactManager.php
+++ b/src/Sulu/Bundle/ContactBundle/Contact/ContactManager.php
@@ -52,21 +52,6 @@ use Sulu\Component\SmartContent\Orm\DataProviderRepositoryInterface;
 class ContactManager extends AbstractContactManager implements DataProviderRepositoryInterface
 {
     /**
-     * @var AccountRepositoryInterface
-     */
-    private $accountRepository;
-
-    /**
-     * @var ContactTitleRepository
-     */
-    private $contactTitleRepository;
-
-    /**
-     * @var ContactRepository
-     */
-    private $contactRepository;
-
-    /**
      * @var MediaRepositoryInterface
      */
     protected $mediaRepository;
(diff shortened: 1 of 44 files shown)


src/Sulu/Bundle/ContactBundle/Contact/ContactManager.php
```diff
@@ -81,31 +66,22 @@ class ContactManager extends AbstractContactManager implements DataProviderRepos
      */
     protected $userRepository;
```
mamazu: Sadly we can't promote protected properties as this would enforce the types on the properties (and be a bc break). I don't think they should have been protected in the first place though.

alexander-schranz: We could still promote them just with phpdoc instead of hardcoded type.

mamazu: This would remove the typehint for the constructor but once all types and the doc comments are gone, this should be a good solution.

mamazu: Alternatively, I don't think this is a huge bc break. Because if you put any other classes into that property the class is going to crash anyways. So I think promoting that and keeping the typehint should be fine.

alexander-schranz: We should keep a eye if everywhere the Interfaces if exists of the service was used. Else I think its not a problem as then service decoration and overwrite by inheritance works.

mamazu: Will do, it's probably not the last time that I'll go over classes. And specific types should be replaced with interfaces whereever possible anyways.

src/Sulu/Bundle/ContactBundle/Controller/AccountController.php
```diff
@@ -213,9 +140,11 @@ public function getContactsAction($id, Request $request)
             $this->restHelper->initializeListBuilder($listBuilder, $fieldDescriptors);
 
             $listBuilder->addSelectField($fieldDescriptors['contactId']);
-            $listBuilder->setIdField($fieldDescriptors['id']);
-            $listBuilder->where($fieldDescriptors['account'], $id);
-            $listBuilder->sort($fieldDescriptors['lastName'], $listBuilder::SORTORDER_ASC);
+            if ($listBuilder instanceof DoctrineListBuilder) {
```
alexander-schranz: We should avoid such changes in this kind of refactors. Also in this case we should just assert that `$listBuilder` is `DoctrineListBuilder` so `\assert($listBuilder instanceof DoctrineListBuilder);` after `$this->listBuilderFactory->create` would be better. but also `@var` is in this case enough as the `DoctrineListBuilderFactoryInterface` always returns a `DoctrineListBuilder` and the controller should so not care and would require to fix it with generics in the list builder interfaces classes itself. But as said that kind of tasks should be in seperate pull request to keep things concentrated on one single refactors.

src/Sulu/Bundle/ContactBundle/Controller/AccountController.php
```diff
@@ -213,9 +140,11 @@ public function getContactsAction($id, Request $request)
...
+                $listBuilder->setIdField($fieldDescriptors['id']);
+                $listBuilder->where($fieldDescriptors['account'], $id);
+                $listBuilder->sort($fieldDescriptors['lastName'], $listBuilder::SORTORDER_ASC);
+            }
```
alexander-schranz:
```suggestion
            $listBuilder->setIdField($fieldDescriptors['id']);
            $listBuilder->where($fieldDescriptors['account'], $id);
            $listBuilder->sort($fieldDescriptors['lastName'], $listBuilder::SORTORDER_ASC);
```

alexander-schranz (approved):

alexander-schranz (approved):

alexander-schranz:
@mamazu Thank you!

PR #7437, opened 2024-05-18
Make EnvironmentTest for Webspace more explicit

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no  
| BC breaks? | no  
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | #7222 #7218
| License | MIT
| Documentation PR | -

#### What's in this PR?
Making the `EnvironmentTest` more explicit on what it does.

#### Why?
When fixing the tests for the Webspace configuration migration I noticed that the second parameter of the Url wasn't used in this test. It turns out that it is overriden by the environment configuration. However, the type of the environment config is set after the url is added so the `environment` property in the expected result is empty (which should not happen). Seeing that the type is an essential property of the environment and should be configured in the constructor (see #7222 for that) We should move the setter call (that is part of the object setup) infront of the `addUrl` call.

### DIFF
--- a/src/Sulu/Bundle/PreviewBundle/Preview/Renderer/PreviewRenderer.php
+++ b/src/Sulu/Bundle/PreviewBundle/Preview/Renderer/PreviewRenderer.php
@@ -315,6 +315,7 @@ class PreviewRenderer implements PreviewRendererInterface
         $portal->setXDefaultLocalization($localization);
         $portal->setLocalizations([$localization]);
         $portal->setDefaultLocalization($localization);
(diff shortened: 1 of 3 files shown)


alexander-schranz:
@mamazu can you rebase this to 2.5?

mamazu:
Done

mamazu:
@alexander-schranz I think this is ready for merging now.

alexander-schranz (approved):

PR #7438, opened 2024-05-22
Fix resetting of PasswordResetTokenEmailsSent Value after Expiring

### LINKED ISSUE #7037: Reset password not working properly
| Q | A
| --- | ---
| Sulu Version | 2.5.7
| PHP Version | 8.2.2 
| DB Version | MariaDB 10.3.36
| Browser Version | Chrome 109.0.5414.74

# Actual Behavior

Reset password does not work properly.
If someone makes a reset password request, request works and "passwordResetTokenEmailsSent" goes up to 3.
Even 48h after, passwordResetTokenEmailsSent stays at 3, so reset emails are not sent.
BUT, this number is never reset to 0, even if time interval (24h) is done.

# Expected Behavior

passwordResetTokenEmailsSent should reset to 0 if a new request is made after 24h interval.

# Steps to Reproduce

In admin login form : 
- Forgot password
- Fill email/user
- request password x>3 times

You should receive max 3 emails.

In database : 
- Change value of passwordResetTokenExpiresAt :  substract some days/months so expiresAt < Now DateTime

In admin login form : 
- Request password

You should not receive emails, because, when requesting, it generates a new token that expires in 24h BEFORE checking if your passwordResetTokenEmailsSent and passwordResetTokenExpiresAt. Then it checks theses 2 values.
Of course, your newly generated token expires in 24h and your passwordResetTokenEmailsSent is still at 3, so you will not receive any email. 

# Possible Solutions
 In \Sulu\Bundle\SecurityBundle\Controller\ResettingController::sendEmailAction , calling 
$token = $this->generateTokenForUser($user);
then
$this->sendTokenEmail($user, $this->getSenderAddress($request), $email, $token);

Conditions are verified in \Sulu\Bundle\SecurityBundle\Controller\ResettingController::sendTokenEmail.

These conditions should be verified in generateTokenForUser and not in sendTokenEmail as generateTokenForUser updates values of the conditions and makes it irrelevant.

TimonStadelmann:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #7037
| License | MIT

#### What's in this PR?

Fix check of `passwordResetTokenExpiresAt` by moving it to the calling function and dividing it into two checks. So that if the token is expired the `passwordResetTokenEmailsSent` is reset to 0. Otherwise it's going to be checked whether the `maxNumberEmails` was reached or not, and acted upon that accordingly.

#### Why?

"Reset password does not work properly.
If someone makes a reset password request, request works and "passwordResetTokenEmailsSent" goes up to 3.
Even 48h after, passwordResetTokenEmailsSent stays at 3, so reset emails are not sent.
BUT, this number is never reset to 0, even if time interval (24h) is done."
- https://github.com/sulu/sulu/issues/7037

#### To Do

- [x] Gather feedback for my changes
- [x] Change target branch to 2.5
- [x] Fix lint
- [x] Add test case

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -26467,7 +26467,7 @@ parameters:
 
 		-
 			message: "#^Call to an undefined method Symfony\\\\Component\\\\Security\\\\Core\\\\User\\\\UserInterface\\:\\:setPasswordResetTokenEmailsSent\\(\\)\\.$#"
-			count: 2
+			count: 3
 			path: src/Sulu/Bundle/SecurityBundle/Controller/ResettingController.php
 
 		-
@@ -27667,7 +27667,7 @@ parameters:
 
 		-
 			message: "#^Cannot call method getCollector\\(\\) on Symfony\\\\Component\\\\HttpKernel\\\\Profiler\\\\Profile\\|false\\|null\\.$#"
-			count: 9
+			count: 11
 			path: src/Sulu/Bundle/SecurityBundle/Tests/Functional/Controller/ResettingControllerTest.php
 
 		-
@@ -27697,7 +27697,7 @@ parameters:
 
 		-
 			message: "#^Parameter \\#1 \\$json of function json_decode expects string, string\\|false given\\.$#"
-			count: 13
+			count: 15
 			path: src/Sulu/Bundle/SecurityBundle/Tests/Functional/Controller/ResettingControllerTest.php
 
 		-
--- a/src/Sulu/Bundle/SecurityBundle/Controller/ResettingController.php
+++ b/src/Sulu/Bundle/SecurityBundle/Controller/ResettingController.php
@@ -110,9 +110,21 @@ class ResettingController
         try {
             /** @var UserInterface $user */
             $user = $this->findUser($request->get('user'));
+            $maxNumberEmails = $this->tokenSendLimit;
+
+            if (new \DateTime() >= $user->getPasswordResetTokenExpiresAt()) {
+                $user->setPasswordResetTokenEmailsSent(0);
+            } elseif ($user->getPasswordResetTokenEmailsSent() === \intval($maxNumberEmails)) {
+                throw new TokenEmailsLimitReachedException($maxNumberEmails, $user);
+            }
+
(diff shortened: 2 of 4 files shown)


alexander-schranz:
Thx for the pull request. There should be a test case added which uses a expired token with a count limit to check if it get correctly reseted. Also as a bugfix this merge request should target the `2.5` branch, you can do that via:

```bash
git rebase -i origin/2.6 --onto origin/2.5

git push yourremote yourbranch --force
```

if origin is your fork you can rebase to sulu 2.5 via:

```bash
git fetch sulu

git rebase -i sulu/2.6 --onto sulu/2.5

git push yourremote yourbranch --force
```

And then change the target branch here.

alexander-schranz:
Let us know if you need any help here or have questions.

TimonStadelmann:
> Let us know if you need any help here or have questions.

Yeah please, I don't really understand why the tests never work. Am I missing something?

alexander-schranz:
> 1) Sulu\Bundle\SecurityBundle\Tests\Functional\Controller\ResettingControllerTest::testResetCounterAfterTokenExpiration
InvalidArgumentException: Collector "swiftmailer" does not exist.

Swiftmailer is only used on lowest versions newer version using the Symfony Mailer so you need bridge that inside the test. Normally the LegacyResettingControllerTest runs for swiftmailer I think and the other for Symfony Mailer which has its own collector.

> octrine\ORM\ORMInvalidArgumentException: A new entity was found through the relationship 'Sulu\Bundle\SecurityBundle\Entity\User#contact' that was not configured to cascade persist operations for entity: Sulu\Bundle\ContactBundle\Entity\Contact@00000000255f63700000000072e9132c. To solve this issue: Either explicitly call EntityManager#persist() on this unknown entity or configure cascade persist this association in the mapping for example @ManyToOne(..,cascade={"persist"}). If you cannot find out which entity causes the problem implement 'Sulu\Bundle\ContactBundle\Entity\Contact#__toString()' to get a clue.

Is mostly an issue of test fixtures vs test client data. Try to clear the entityManager after you created the fixtures before doing the client request.

PR #7439, opened 2024-05-23
Update RouteRepository.php

tkuska:
I belive persisting route should also flush it in database

| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs |  -
| License | MIT
| Documentation PR | -

#### What's in this PR?

i have added flushing entity manager in `persist` and `remove` methods in `Sulu\Bundle\RouteBundle\Entity\RouteRepository`

#### Why?

Using  `Sulu\Bundle\RouteBundle\Entity\RouteRepository:persist`  method you had to manually call `flush()` on entityManager. It means if you wanted to use `RouteManager` you had to inject not only `RouteManager` itself but `EntityManager` also to flush routes.

### DIFF
--- a/src/Sulu/Bundle/RouteBundle/Entity/RouteRepository.php
+++ b/src/Sulu/Bundle/RouteBundle/Entity/RouteRepository.php
@@ -89,10 +89,13 @@ class RouteRepository extends EntityRepository implements RouteRepositoryInterfa
     public function persist(RouteInterface $route)
     {
         $this->getEntityManager()->persist($route);
+        $this->getEntityManager()->flush();
+        
     }
 
     public function remove(RouteInterface $route)
     {
         $this->getEntityManager()->remove($route);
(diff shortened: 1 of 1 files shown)


alexander-schranz:
@tkuska our repositories should not do the flush in the add method, this is expected as the flush is done by the entity which is being created or should be done by the controller. In future flushing will happening in the messagebus over the FlushMiddleware ( https://github.com/sulu/messenger?tab=readme-ov-file#doctrineflushmiddleware).

Flushing inside the repository will should not be done as when you are importing a list of entities you don't want automatically trigger a flush after every route is created for performance reasons, flushing should so be done by the most outside service possible (controllers or commands), the which knows when the right time for the flush is. This is mostly in old repositories differently but will be the way how we handling flush in all repositories in the future major versions.

PR #7442, opened 2024-05-28
Use php-cs-fixer shim to avoid conflicts

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no 
| BC breaks? | no 
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Use php-cs-fixer shim to avoid conflicts.

#### Why?

The php cs fixer added some more dependency for parallel runs to avoid future dependeny conflicts we should use the `shim` verson instead.

 - https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/blob/master/composer.json
 - https://github.com/PHP-CS-Fixer/shim/blob/master/composer.json

### DIFF
--- a/.github/workflows/test-application.yaml
+++ b/.github/workflows/test-application.yaml
@@ -205,8 +205,8 @@ jobs:
                   coverage: none
 
             - name: Remove Lint Tools
-              # This tools are not required for running tests and are so removed to improve dependency resolving and
-              # testing lowest versions
+              # These tools are not required to run tests, so we are removing them to improve dependency resolving and
+              # testing lowest versions.
               run: composer remove "*php-cs-fixer*" "*phpstan*" --dev --no-update
 
             - name: Require jackrabbit dependencies
--- a/.github/workflows/test-application.yaml
+++ b/.github/workflows/test-application.yaml
@@ -204,6 +204,11 @@ jobs:
                   ini-values: memory_limit=-1
                   coverage: none
 
+            - name: Remove Lint Tools
+              # This tools are not required for running tests and are so removed to improve dependency resolving and
+              # testing lowest versions
+              run: composer remove "*php-cs-fixer*" "*phpstan*" --dev --no-update
+
             - name: Require jackrabbit dependencies
               if: ${{ matrix.phpcr-transport == 'jackrabbit' }}
               run: |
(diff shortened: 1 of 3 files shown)


.php-cs-fixer.dist.php
```diff
@@ -15,6 +15,7 @@
 
 $config = new PhpCsFixer\Config();
 $config->setRiskyAllowed(true)
+    ->setParallelConfig(PhpCsFixer\Runner\Parallel\ParallelConfigFactory::detect())
```
alexander-schranz, line 18: is experimental but would improve the performance locally alot: https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/pull/7777#issuecomment-2134977471 @sulu/core-developer

.github/workflows/test-application.yaml
```diff
@@ -204,6 +204,11 @@ jobs:
                   ini-values: memory_limit=-1
                   coverage: none
 
+            - name: Remove Lint Tools
+              # This tools are not required for running tests and are so removed to improve dependency resolving and
+              # testing lowest versions
+              run: composer remove "*php-cs-fixer*" "*phpstan*" --dev --no-update
+
```
alexander-schranz: This improves the CI dependency resolving as this dependencies are not required for tests and so composer has less todo.

Prokyonn (approved):

.github/workflows/test-application.yaml
```diff
@@ -204,6 +204,11 @@ jobs:
                   ini-values: memory_limit=-1
                   coverage: none
 
+            - name: Remove Lint Tools
+              # This tools are not required for running tests and are so removed to improve dependency resolving and
+              # testing lowest versions
```
Prokyonn:
```suggestion
              # These tools are not required to run tests, so we are removing them to improve dependency resolving and
              # testing lowest versions.
```

.php-cs-fixer.dist.php
Wirone: FYI: it's experimental mostly because we wanted to make "canary" adoption, as an opt-in, and to not feel the pressure about potential bugs. But it's well tested and stable enough to just use it 😁.

wachterjohannes (approved):

PR #7311, opened 2024-03-15
Fix routepath on language copy if the parent page only exist in one l…

martinlagler:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes https://github.com/sulu/SuluArticleBundle/issues/661
| License | MIT

#### What's in this PR?

Only add the parent page (page tree route) if it exists in the target language.

#### Why?

Otherwise there will be wrong routes.

PR #7450, opened 2024-06-05
Do not override existing route on copy locale subscriber

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes # 
| Related issues/PRs | #7311, #7309
| License | MIT

#### What's in this PR?

Fixes the copy locale subscriber to work properly for articles

#### Why?

The existing code fails in two different ways for both cases:

Original locale is already published:
copy locale runs through without an exception, but the original route is modified and changed to the new locale due to these two lines:

```php
$destDocument->getRoute()->setPath($routePathProp);
$destDocument->getRoute()->setLocale($destLocale);
```

Original locale is not published:
the copy locale request fails with an exception, because the `destDocument->getRoute()` returns null in the same two lines
```php
$destDocument->getRoute()->setPath($routePathProp);
$destDocument->getRoute()->setLocale($destLocale);
```

Additionally the parent page was lost in case that the page is not yet published, in my opinion this is not necessary, the new copied locale can still reference an unpublished page for the route.

### DIFF
--- a/src/Sulu/Component/Content/Document/Subscriber/CopyLocaleSubscriber.php
+++ b/src/Sulu/Component/Content/Document/Subscriber/CopyLocaleSubscriber.php
@@ -167,7 +167,7 @@ class CopyLocaleSubscriber implements EventSubscriberInterface
             return $documentStructure;
         }
 
-        $parentPageUuid = $routePath['page']['uuid'];
+        $parentPageUuid = $routePath['page']['uuid'] ?? null;
 
         /** @var ?PageDocument $destParentDocument */
         $destParentDocument = $this->documentManager->find(
--- a/src/Sulu/Component/Content/Document/Subscriber/CopyLocaleSubscriber.php
+++ b/src/Sulu/Component/Content/Document/Subscriber/CopyLocaleSubscriber.php
@@ -13,6 +13,7 @@ namespace Sulu\Component\Content\Document\Subscriber;
 
 use Sulu\Bundle\DocumentManagerBundle\Bridge\DocumentInspector;
 use Sulu\Bundle\PageBundle\Document\PageDocument;
+use Sulu\Bundle\RouteBundle\Model\RouteInterface;
 use Sulu\Component\Content\Document\Behavior\ExtensionBehavior;
 use Sulu\Component\Content\Document\Behavior\ResourceSegmentBehavior;
 use Sulu\Component\Content\Document\Behavior\StructureBehavior;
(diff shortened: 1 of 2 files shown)


src/Sulu/Component/Content/Document/Subscriber/CopyLocaleSubscriber.php
```diff
@@ -192,8 +191,6 @@ public function checkPageTreeRoute(RoutableBehavior $destDocument, StructureBeha
             'suffix' => $routePath['suffix'],
         ];
         $destDocument->setRoutePath($routePathProp);
-        $destDocument->getRoute()->setPath($routePathProp);
```
alexander-schranz, line 195: do we need to call `$destDocument->setRoute(null);`

alexander-schranz (commented):
We need test two cases:

1. The locale does not yet exist and I use the copy language overlay

2. The locale does already exist and I use the copy language toolbar action (maybe here two cases need to be tested published and only draft)

alexander-schranz:
@Prokyonn I'm running into another issue with Sulu Article Bundle tested with Sulu Demo (no page_tree_route used):

500 error with:

Bildschirmfoto 2024-06-07 um 13 39 04

src/Sulu/Component/Content/Document/Subscriber/CopyLocaleSubscriber.php
```diff
@@ -163,7 +163,7 @@ public function checkPageTreeRoute(RoutableBehavior $destDocument, StructureBeha
         $pageTreeRoutePropertyName = $this->getPageTreeRoutePropertyName($document);
         $routePath = $documentStructure[$pageTreeRoutePropertyName] ?? null;
 
-        if (!$routePath || !\array_key_exists('page', $routePath)) {
+        if (!$routePath || !\is_array($routePath) || !\array_key_exists('page', $routePath)) {
```
alexander-schranz, line 166: @Prokyonn added this to avoid: https://github.com/sulu/sulu/pull/7450#issuecomment-2154662081

src/Sulu/Component/Content/Document/Subscriber/CopyLocaleSubscriber.php
Prokyonn: `setRoute` does not accept `null` so we cannot call it

src/Sulu/Component/Content/Document/Subscriber/CopyLocaleSubscriber.php
Prokyonn: added a test for it, although I'm not exactly sure how it can happen in production 🤔

alexander-schranz: I just created an article in the sulu demo and did a language copy.

alexander-schranz (approved):

PR #7456, opened 2024-06-14
Fix empty phone/email/url/fax error on contact and accounts

domPatera:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no

#### What's in this PR?

Fixes the error when creating a new contact.

#### Why?

When you try to create a new contact, fill something in a field (e.g. phone), and then delete it, you cannot save the contact because of the following error:
`An exception occurred while executing a query: SQLSTATE[23000]: Integrity constraint violation: 1048 Column 'phone' cannot be null`

### DIFF
--- a/src/Sulu/Bundle/ContactBundle/Contact/AbstractContactManager.php
+++ b/src/Sulu/Bundle/ContactBundle/Contact/AbstractContactManager.php
@@ -577,7 +577,7 @@ abstract class AbstractContactManager implements ContactManagerInterface
         $urls = $this->getProperty($contactDetailsData, 'websites');
         if (!empty($urls)) {
             foreach ($urls as $urlData) {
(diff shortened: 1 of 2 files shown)


alexander-schranz:
Thx for your pull request but I could not reproduce the issue can you record a screencast using https://sulu.rocks/admin how to reproduce it.

domPatera:
> Thx for your pull request but I could not reproduce the issue can you record a screencast using https://sulu.rocks/admin how to reproduce it.

https://imgur.com/fIh053Y

alexander-schranz:
@domPatera Thank you. I rebased this as a bugfix to 2.5 and adopted the test to catch this error.

alexander-schranz (approved):

PR #7459, opened 2024-06-18
Reindex moved documents so that the URL is updated

FlorianKoerner:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

The document indexed by MassiveSearch also contains a field for the URL (__url). This field was not updated when the document was moved. As a result, the search results pointed to the old URL until the document was published again.


alexander-schranz (approved):

alexander-schranz:
Sounds reasonable @FlorianKoerner Thx for the contribution.

Just fixed the PHPStan issues in this old file.

PR #7463, opened 2024-06-19
Fix unexpected single sign on definitions from SecurityBundle services.xml

### LINKED ISSUE #7447: [BC break v2.6] SingleSignOnLoginRequestSubscriber breaks Password reset
| Q | A
| --- | ---
| PHP Version | all
| DB Version | all
| Browser Version | all

# Actual Behavior

After upgrading to Sulu 2.6 the password reset stopped working. I investigated and found out that the new `SingleSignOnLoginRequestSubscriber` from the `SecurityBundle` prevents the `sendEmailAction` of the `ResettingController` to be called, when no `singleSignOnAdapter` is found. When I comment out that two lines setting another Response and call `stopPropagation()`, the reset controller is called again.

```` php
        $adapter = $this->singleSignOnAdapterProvider->getAdapterByDomain($domain);
        if (!$adapter) {
            if ($password) {
                return;
            }

            $event->setResponse(new JsonResponse(['method' => 'json_login'], 200));
            $event->stopPropagation();

            return;
        }
````

# Expected Behavior

The password reset should work when the SSO is not enabled. Edit: And also for a local account with an e-mail address for which no SSO adapter is configured in a mixed SSO/local-user scenario.

# Steps to Reproduce

Have Sulu 2.6 installed and try to send a password reset email and observe that its controller code is not executed at all and instead of an empty response, get `{method: "json_login"}` that comes from the ``SingleSignOnLoginRequestSubscriber` aborting the process.

# Possible Solutions

~~Only register that `SingleSignOnLoginRequestSubscriber` if the SSO is used or~~ fix the `SingleSignOnLoginRequestSubscriber` so it does allow the `sendEmailAction` to be executed, when the SSO is not enabled or a  reset for a local account is requested.

rogamoore:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #7447
| Related issues/PRs | #
| License | MIT

#### What's in this PR?

Remove SSO service definitions from main services.xml

#### Why?

They are already defined in single_sign_on.xml and should be loaded conditionally (see https://github.com/sulu/sulu/blob/2.6/src/Sulu/Bundle/SecurityBundle/DependencyInjection/SuluSecurityExtension.php#L115)

### DIFF
--- a/src/Sulu/Bundle/SecurityBundle/Resources/config/services.xml
+++ b/src/Sulu/Bundle/SecurityBundle/Resources/config/services.xml
@@ -388,41 +388,5 @@
 
             <tag name="sulu_admin.form_metadata_visitor" />
         </service>
-
-        <service id="sulu_security.open_id_login_subscriber" class="Sulu\Bundle\SecurityBundle\SingleSignOn\SingleSignOnLoginRequestSubscriber">
-            <argument type="service" id="sulu_security.single_sign_provider"/>
-            <argument type="service" id="router"/>
-            <argument type="service" id="sulu.repository.user"/>
-            <tag name="kernel.event_subscriber"/>
-        </service>
-
-        <service id="sulu_security.single_sign_on_adapter_factory_open_id" class="Sulu\Bundle\SecurityBundle\SingleSignOn\Adapter\OpenId\OpenIdSingleSignOnAdapterFactory">
-            <argument type="service" id="http_client"/>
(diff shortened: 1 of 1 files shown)


alexander-schranz (approved):

alexander-schranz:
@rogamoore thank you!

Issue #7207, opened 2023-11-22
Ignore not existing columns in the ListBuilder


Prokyonn:
| Q | A
| --- | ---
| Sulu Version | 2.5

# Actual Behavior

When you have a page with multiple tabs with listviews, e.g. the `Automation` tab from the AutomationBundle and the `Data` tab from the FormBundle, the query parameters for sorting/filtering the lists are copied from one tab to the other. 
Therefore it can happen, that you are on the `Automation` tab, sort after `schedule` and then move to the `Data` tab, the request sent to the server has then `sortBy=schedule` which ends in a 500, because the `Data` listview configuration does not have a column `schedule`.

# Expected Behavior

The request doesn't end in a 500 and returns the data with default sorting, ignoring the query parameters that define non-existing columns.

# Steps to Reproduce

see above

# Possible Solutions

Ignore query parameters that reference not existing columns

MrSrsen:
It looks like bug in the `vendor/sulu/sulu/src/Sulu/Component/Rest/RestHelper.php` class in the `initializeListBuilder` method. This condition:

```php
        $sortBy = $this->listRestHelper->getSortColumn();
        if (null != $sortBy) {
            $listBuilder->sort($fieldDescriptors[$sortBy], $this->listRestHelper->getSortOrder());
        }
```

should look like this:

```php
        $sortBy = $this->listRestHelper->getSortColumn();
        if (null !== $sortBy && \array_key_exists($sortBy, $fieldDescriptors)) {
            $listBuilder->sort($fieldDescriptors[$sortBy], $this->listRestHelper->getSortOrder());
        }
```

I copied the original code and fixed it in my own class and decorated the original like this:

```
    App\Sulu\RestHelper:
        decorates: sulu_core.rest_helper
        arguments:
            - '@sulu_core.list_rest_helper'
```

and it fixed my problem. Good-enough hotfix for now.

alexander-schranz:
@MrSrsen Sounds like a good solution. Do you want to create a pull request for this?

mamazu:
Alternatively we could also allow the `sort` to allow `null` as sorting then we could remove the if statement entirely.

MrSrsen:
> @MrSrsen Sounds like a good solution. Do you want to create a pull request for this?

@alexander-schranz to which repository should I make the PR? `sulu/sulu`? And shouldn't this rather be fixed in the JavaScript?

mamazu:
[...]

PR #7469, opened 2024-06-21
Only allow sorting references by existing fields

mamazu:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | #7207
| License | MIT
| Documentation PR | -

#### What's in this PR?
Only allow sorting by columns that the user can see (because we don't use the `DoctrineListBuilder` here).

#### Why?
If the user gets a `sortBy` inserted into the query parameter and can't get rid of it (since it's saved in the user settings) the controller returns a 500 and the user sees nothing.

The downside is that now you can't sort by fields that you can't see. But I don't think that this is such a downside.

### DIFF
--- a/src/Sulu/Bundle/ReferenceBundle/UserInterface/Controller/Admin/ReferenceController.php
+++ b/src/Sulu/Bundle/ReferenceBundle/UserInterface/Controller/Admin/ReferenceController.php
@@ -85,14 +85,6 @@ class ReferenceController extends AbstractRestController implements ClassResourc
             'offset' => $offset,
         ]);
 
-        /** @var string|null $sortBy */
-        $sortBy = $request->query->get('sortBy');
-        /** @var 'asc'|'desc' $sortOrder */
-        $sortOrder = $request->query->get('sortOrder', 'asc');
-        $sortBys = $sortBy ? [
-            $sortBy => $sortOrder,
-        ] : [];
-
         $fields = \explode(',', $request->query->get('fields', ''));
         $removeFields = ['id']; // the frontend always add the id field, but we don't need it in this case as we group by other fields
         $fields = [
@@ -102,6 +94,15 @@ class ReferenceController extends AbstractRestController implements ClassResourc
             'referenceLocale',
             'referenceRouterAttributes',
         ];
+
+        $sortBys = [];
+        /** @var string|null $sortBy */
+        $sortBy = $request->query->get('sortBy');
+        if (\in_array($sortBy, $fields, true)) {
+            /** @var 'asc'|'desc' $sortOrder */
+            $sortOrder = $request->query->get('sortOrder', 'asc');
+            $sortBys = [$sortBy => $sortOrder];
+        }
 
         if ($rootLevel) {
             $removeFields[] = 'referenceContext';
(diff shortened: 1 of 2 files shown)


src/Sulu/Bundle/ReferenceBundle/UserInterface/Controller/Admin/ReferenceController.php
```diff
@@ -85,14 +85,6 @@ public function cgetAction(Request $request): Response
...
-        $sortBy = $request->query->get('sortBy');
-        /** @var 'asc'|'desc' $sortOrder */
-        $sortOrder = $request->query->get('sortOrder', 'asc');
-        $sortBys = $sortBy ? [
-            $sortBy => $sortOrder,
-        ] : [];
-
         $fields = \explode(',', $request->query->get('fields', ''));
         $removeFields = ['id']; // the frontend always add the id field, but we don't need it in this case as we group by other fields
         $fields = [
```
Prokyonn, line 90: @mamazu We should also validate the `fields` with the list metadata, otherwise it is still possible to crash, when the FE sends invalid fields.
Both `fields` and `sortBy` are directly used from the request and unfortunately it is also not validated in the FE 🙈

mamazu: Yes, I don't know why we can't use some parts of the `DoctrineListBuilder` here as that should have all the necessary information.

Prokyonn: @mamazu I asked the same thing @alexander-schranz 😂 

I don’t recall the details exactly, but I believe we were considering an alternative approach at that time. This approach would have utilized the ⁠DoctrineListBuilder in the repository. Unfortunately, time slipped away, and we didn’t have the opportunity to thoroughly discuss or refactor it before the release 🙈.

I’m still open to refactoring it to use the DoctrineListBuilder, so we can maintain some consistency between the controllers.

@alexander-schranz WDYT?

mamazu: I mean that would be the cherry on top. Seeing that together with the automation bundle you can't use this feature anymore. (We fixed it in our instance with a vendor patch so no hurry).

rogamoore:
I just ran into the same problem. After telling everybody about the cool new references features, nobody was able to see it. 🙈  A quick workaround is to tell the users to click on the title column.

Is there anything I can help with?

mamazu:
Yep same here. We're about to replace our own implementation with Sulu's. If you need a quick fix of this. Apply this patch in your vendor folder.

Prokyonn:
> Yep same here. We're about to replace our own implementation with Sulu's. If you need a quick fix of this. Apply this patch in your vendor folder.

@mamazu if this works for you, I would say we merge this and open an issue to discuss/refactor this, then we do not have time pressure and it will at least more or less work as expected

Prokyonn (approved):

PR #7472, opened 2024-06-23
Add Global blocks definitions for the FormMetadata

eKsiK:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | #7229 #7423
| License | MIT
| Documentation PR | -

#### What's in this PR?

Add missing definitions of used global blocks in the FormData in custom entities.

more info: https://sulu-io.slack.com/archives/C0430GGCV/p1718908503052599

#### Why?

Actually when using global blocks in the form in custom entities will throw error `Error: can't resolve reference #/definitions/[global_block] from id #` because GlobalBlockVisitor has missing interface for parsing `FormMetadata`.

### DIFF
--- a/src/Sulu/Bundle/CoreBundle/Resources/config/content.xml
+++ b/src/Sulu/Bundle/CoreBundle/Resources/config/content.xml
@@ -211,6 +211,7 @@
             <argument type="service" id="sulu_admin.metadata_provider_registry" />
 
             <tag name="sulu_admin.typed_form_metadata_visitor"/>
+            <tag name="sulu_admin.form_metadata_visitor" />
         </service>
 
         <!-- content query -->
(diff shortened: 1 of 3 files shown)


alexander-schranz:
Looking at the code I think there is an additional tag missing for the new visitor to get it loaded: https://github.com/sulu/sulu/blob/b58abfc5e3c947d03547a70b59a7cf53c1814985/src/Sulu/Bundle/CoreBundle/Resources/config/content.xml#L213. `sulu_admin.form_metadata_visitor` should I think be added.

eKsiK:
@alexander-schranz You are right! Sorry... I added the missing tag :)

wachterjohannes (approved):

PR #7481, opened 2024-06-26
Fix locale length for languages like zh-Hans

martinlagler:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Increase the length of the "locale" fields.

#### Why?

To allow locales like "zh-Hans".

So the biggest official locale is `zh-Hans_CN` which 10 chars.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,24 @@
 # Upgrade
 
+## 2.6.3
+
+### Change locale length
+
+Change length of the locale fields to support longer locales.
+
+```sql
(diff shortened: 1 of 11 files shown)


alexander-schranz (changes requested):
UPGRADE md with the SQLs missing

alexander-schranz (commented):
As discussed we will go with `15` as a buffer and it also matches binary `1111`

Issue #7319, opened 2024-03-20
Old media versions throw 404 error


Quehnie:
| Q | A
| --- | ---
| Sulu Version | 2.5.12 introduced mostly with #7259

# Actual Behavior

If an old media version gets requested the MediaStreamController throws a 404 Error.

# Expected Behavior

src/Sulu/Bundle/MediaBundle/Controller/MediaStreamController.php Line 257  
Should use getMediaById so we get all Versions and not only the current one.

# Steps to Reproduce

Upload 2 File Versions.

# Possible Solutions

src/Sulu/Bundle/MediaBundle/Controller/MediaStreamController.php Line 257  
Should use getMediaById so we get all Versions and not only the current one.

alexander-schranz:
I think the issue is here https://github.com/sulu/sulu/pull/7259/files#r1625969195
@Quehnie or @harunbuljubasic do you want to create a pull request for it?

alexander-schranz:
fixed in #7480

PR #7484, opened 2024-06-26
Fix media version file not found error

### LINKED ISSUE #7480: Media item > History | Previous versions not found error
| Q | A
| --- | ---
| Sulu Version | 2.5.18 (Also 2.6)
| PHP Version | 8.2
| DB Version | MySQL 8.0
| Browser Version | Edge/Chrome...

# Actual Behavior

Changing the file in the media menu creates a new version of the file that you can see in the History tab. The issue arises when you want to view a previous version by clicking on the eye icon.
It throws a **not found** exception.

# Expected Behavior

The same as clicking the current version. Show the selected file, in this case, a previous version of the Media item.

# Steps to Reproduce

Optional: Create a new media file from the Admin menu **Media**.
Do what is indicated in **Actual Behavior**, that is, modify the media by **Uploading a new version**, then try to view any previous version.

# Possible Solutions

I see here that `MediaRepository::findMediaByIdForRendering` is only retrieving one version. This is called from the download action in `MediaStreamController`, inside the `getFileVersion` method.

`->leftJoin('file.fileVersions', 'fileVersion', Join::WITH, 'file.version = fileVersion.version')`

I just removed `Join::WITH, 'file.version = fileVersion.version` and everything is working fine.

I am not sure if I should upload a fix because I do not have much context on Sulu/Media core, so I will leave this here for someone else to make that decision.

tapasa30:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? |yes
| Deprecations? | no
| Fixed tickets | fixes #7480 and #7319
| Related issues/PRs | #7480 and #7319
| License | MIT

#### What's in this PR?

Adds new parameter to add specific version into Media.fileVersions results in addition to the current version

#### Why?

As explained here (https://github.com/sulu/sulu/issues/7480) and here (https://github.com/sulu/sulu/issues/7319), trying to view old versions of a media file results in a 404 not found error.
This is because the query that retrieves the media item only fetches the current version of the file. Therefore, when attempting to find a specific version that is not the current one, the file version cannot be found.

#### Example Usage

Optional: Create a new media file from the Admin menu Media.
Changing the file in the media menu creates a new version of the file that you can see in the History tab. Click on the eye icon in any previous version and the file should show without any problem.

#### To Do

- [ ] Integration Tests
- [ ] Unit Tests

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/Controller/MediaStreamController.php
+++ b/src/Sulu/Bundle/MediaBundle/Controller/MediaStreamController.php
@@ -214,8 +214,7 @@ class MediaStreamController
             return null;
         }
 
-        /** @var File|null $file */
-        $file = $mediaEntity->getFiles()[0];
+        $file = $mediaEntity->getFiles()[0] ?? null;
 
         if (!$file) {
             return null;
--- a/src/Sulu/Bundle/MediaBundle/Controller/MediaStreamController.php
+++ b/src/Sulu/Bundle/MediaBundle/Controller/MediaStreamController.php
@@ -205,26 +205,29 @@ class MediaStreamController
      */
     protected function getFileVersion($id, $version)
     {
-        /** @var MediaInterface $mediaEntity */
-        $mediaEntity = $this->mediaRepository->findMediaByIdForRendering($id, null);
+        $version = empty($version) ? null : (int)$version;
 
-        if (!$mediaEntity) {
+        /** @var MediaInterface|null $mediaEntity */
+        $mediaEntity = $this->mediaRepository->findMediaByIdForRendering($id, null, $version);
+
+        if ($mediaEntity === null) {
             return null;
         }
 
-        $file = $mediaEntity->getFiles()[0];
+        /** @var File|null $file */
+        $file = $mediaEntity->getFiles()->get(0);
 
-        if (!$file) {
+        if ($file === null) {
             return null;
         }
 
-        if (!$version) {
-            $version = $mediaEntity->getFiles()[0]->getVersion();
+        if ($version === null) {
+            $version = $file->getVersion();
         }
 
-        $fileVersion = $file->getFileVersion((int) $version);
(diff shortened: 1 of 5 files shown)


alexander-schranz:
I did some updates on this to keep bc and fix some stan issues. @Prokyonn can you also have a look at this?

src/Sulu/Bundle/MediaBundle/Controller/MediaStreamController.php
```diff
@@ -197,32 +199,33 @@ private function createBinaryFileResponse(
...
-        $mediaEntity = $this->mediaRepository->findMediaByIdForRendering($id, null);
+        /** @var MediaInterface|null $mediaEntity */
+        $mediaEntity = $this->mediaRepository->findMediaByIdForRendering($id, null, $version);
 
         if (!$mediaEntity) {
             return null;
         }
 
+        /** @var File|null $file */
         $file = $mediaEntity->getFiles()[0];
```
Prokyonn: this breaks when `getFiles()` returns an empty array

```suggestion
        $file = $mediaEntity->getFiles()[0] ?? null;
```

Prokyonn (changes requested):

src/Sulu/Bundle/MediaBundle/Controller/MediaStreamController.php
tapasa30: `getFiles` is an **ArrayCollection**, the index 0 can be retrieved with `getFiles()->get(0)`, so in the case it does not exists, it returns `null`

alexander-schranz: This is expected we avoid coupling to Doctrine Collection specific classes in case as we move step by step to a Hexagonial Architecture.

alexander-schranz:
Thx @tapasa30

PR #7490, opened 2024-06-27
Add constructor property promotion in RouteBundle classes

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | #7435 #7325 
| License | MIT
| Documentation PR | -

#### What's in this PR?
Using constructor property promotion in the RouteBundle

#### Why?
There isn't much properties to promote here so we do type fixing. :D

### DIFF
--- a/src/Sulu/Bundle/RouteBundle/Command/UpdateRouteCommand.php
+++ b/src/Sulu/Bundle/RouteBundle/Command/UpdateRouteCommand.php
@@ -21,46 +21,17 @@ use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Contracts\Translation\LocaleAwareInterface;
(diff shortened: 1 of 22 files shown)


src/Sulu/Bundle/RouteBundle/Command/UpdateRouteCommand.php
```diff
@@ -21,46 +21,17 @@
...
-    /**
-     * @var RouteManagerInterface
-     */
-    private $routeManager;
-
     public function __construct(
-        TranslatorInterface $translator,
-        EntityManagerInterface $entityManager,
-        RouteManagerInterface $routeManager
+        private LocaleAwareInterface $translator,
```
alexander-schranz, line 30: I think this still shold be both?

alexander-schranz: okay looks like trans is not used here?

alexander-schranz (approved):

PR #7496, opened 2024-06-27
Use constructor property promotion in TagBundle

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | #7435 #7325 
| License | MIT
| Documentation PR | -

#### What's in this PR?
Using constructor property promotion in the TagBundle

#### Why?
Less code = Better

### DIFF
--- a/src/Sulu/Bundle/TagBundle/Controller/Exception/ConstraintViolationException.php
+++ b/src/Sulu/Bundle/TagBundle/Controller/Exception/ConstraintViolationException.php
@@ -27,20 +27,13 @@ class ConstraintViolationException extends RestException
     public const EXCEPTION_CODE_NON_UNIQUE_NAME = 1101;
 
     /**
-     * The field of the tag which is not unique.
-     *
-     * @var string
-     */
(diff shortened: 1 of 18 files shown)


src/Sulu/Bundle/TagBundle/Controller/Exception/ConstraintViolationException.php
```diff
@@ -27,20 +27,13 @@ class ConstraintViolationException extends RestException
...
      * @param string $field The field which is not
-     * @param int $code
      */
-    public function __construct($message, $field, $code = 0)
-    {
-        $this->field = $field;
+    public function __construct(
+        string $message,
+        protected $field,
+        int $code = 0,
```
mamazu, line 35: Hoisting type for message and code from the base exception. In PHP8 message must be a string and code must be an integer.

alexander-schranz (approved):

PR #7510, opened 2024-07-06
Replacing `will` method with more specific methods

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | #7507 
| License | MIT
| Documentation PR | -

#### What's in this PR?
Replacing outdated phpunit methods with the specific methods. Thanks to Rector for https://github.com/rectorphp/rector-phpunit/blob/main/rules/CodeQuality/Rector/MethodCall/UseSpecificWillMethodRector.php

#### Why?
PHPUnit 11


mamazu:
I am just looking through the code of the tests and found a lot of `$this->expect($this->any())` which doesn't do anything. Expecting that this method has been called would require `$this->expect($this->atLeastOnce())` or replace it with a more generic version to mock methods like `$this->createConfiguredMock()`.

alexander-schranz (approved):

PR #7525, opened 2024-07-12
Add noindex switch to link

greatmaxix:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets |
| Related issues/PRs | #6675
| License | MIT
| Documentation PR | 

#### What's in this PR?

This PR allows setting noindex rel to external links

#### Why?

Our SEO engineers asked us to be able to set noindex attribute for external links

#### To Do


alexander-schranz:
@greatmaxix Thx for your work but this does not make sense as the `noindex` does not exist for `rel` of links: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel

You can not tell a search engine if the target of a link should be indexed or not. The site / page itself decides via meta tags (or headers or robots.txt) to be indexed or not (in sulu via the SEO Tab the noindex can be defined).

If you want to tell a search engine to ignore that link you need to set `nofollow`. But the external site itself need to tell the search engine if it want to get indexed or not, that something not a link can do.

greatmaxix:
> You can not tell a search engine if the target of a link should be indexed or not. The site / page itself decides via meta tags (or headers or robots.txt) to be indexed or not (in sulu via the SEO Tab the noindex can be defined).

Oh, I see. Let me discuss this with our seo engineers, and ask from them more info.

greatmaxix:
Will close it for now

alexander-schranz:
@greatmaxix Thx still for your try to contribute something to Sulu, there can be different links to the same site with different rel and a search engine would then not know what really todo thats why noindex only can only defined by the site itself, and doesnt exist in the specification for links and has there no effects.

The `nofollow` is something which is common in this case to tell not following that external link.

PR #7534, opened 2024-07-22
Add forcing direct matching for media image format urls

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no 
| Deprecations? | yes 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Add direct matching for media image format urls.

#### Why?

Old Version Urls will rerendering the image format which can end up in high resources usage. So instead we are redirecting old version to new versions which should normally end in a cache hit.

As reported by one of our partners to improve traffic on their projects.

#### To Do

- [x] Test for redirect to new version
- [x] Test for redirect to new version on webp
- [x] Test for 404 for not matching file  name

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -4,8 +4,8 @@
 
 ### Stricter Image Format Url Handling
 
-The image formats url need a direct match of filename to get the image format returned.
-Old versions will be redirected to the new version and none direct matches return 404 now.
+The image formats URL requires an exact filename match to retrieve the correct image format. 
+Old versions will be redirected to the new version and any non-matching filenames will now return a 404 error.
 
 ## 2.5.15
 
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,12 @@
 # Upgrade
 
+## 2.5.20
+
+### Stricter Image Format Url Handling
+
+The image formats url need a direct match of filename to get the image format returned.
+Old versions will be redirected to the new version and none direct matches return 404 now.
+
 ## 2.5.15
 
 ### Run Shadow migrations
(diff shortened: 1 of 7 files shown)


UPGRADE.md
```diff
@@ -1,5 +1,12 @@
 # Upgrade
 
+## 2.5.20
+
+### Stricter Image Format Url Handling
+
+The image formats url need a direct match of filename to get the image format returned.
+Old versions will be redirected to the new version and none direct matches return 404 now.
```
```suggestion
The image formats URL requires an exact filename match to retrieve the correct image format. 
Old versions will be redirected to the new version and any non-matching filenames will now return a 404 error.
```

src/Sulu/Bundle/MediaBundle/Controller/MediaStreamController.php
```diff
@@ -61,13 +61,18 @@ public function getImageAction(Request $request)
             }
 
             $url = $request->getPathInfo();
+            // there are some projects which do not call this action with ?v=1-0 because of they didn't wanted query strings in the image urls ( unnecessary SEO mystic reasons )
+            // to not break this projects we will fallback to 1-0 if no version is given
```
Prokyonn:
```suggestion
            // To maintain compatibility with these projects, we will fallback to version 1-0 if no version is specified.
```

src/Sulu/Bundle/MediaBundle/Controller/MediaStreamController.php
```diff
@@ -61,13 +61,18 @@ public function getImageAction(Request $request)
             }
 
             $url = $request->getPathInfo();
+            // there are some projects which do not call this action with ?v=1-0 because of they didn't wanted query strings in the image urls ( unnecessary SEO mystic reasons )
```
Prokyonn:
```suggestion
            // Some projects do not call this action with ?v=1-0 because they don't want query strings in the image urls ( unnecessary SEO mystic reasons )
```

Prokyonn (approved):

PR #7543, opened 2024-07-25
Fix whitespace handling in URLs of media file names

stollr:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | 
| Related issues/PRs | 
| License | MIT
| Documentation PR | 

#### What's in this PR?

The file name in the URL of media requests is url-decoded in the [LocalFormatCache](https://github.com/sulu/sulu/blob/9354c5e8ff2eb40154eac1122b6ddc720eb785a1/src/Sulu/Bundle/MediaBundle/Media/FormatCache/LocalFormatCache.php#L207) to normalize special characters like white spaces.

The URL to a file may look like this: `https://www.example.com/uploads/media/950x/01/1-my%20image.png?v=1-0`. The filename "my%20image.png" will be url-decoded to "my image.png", now.

#### Why?

Currently there is a bug when media files are used that have white spaces in their file name. It leads to a `Sulu\Bundle\MediaBundle\Media\Exception\ImageProxyMediaNotFoundException` in [src/Sulu/Bundle/MediaBundle/Media/FormatManager/FormatManager.php at line 117](https://github.com/sulu/sulu/blob/9354c5e8ff2eb40154eac1122b6ddc720eb785a1/src/Sulu/Bundle/MediaBundle/Media/FormatManager/FormatManager.php#L117) when they are requested from the frontend, because in the database the filename is stored with a white space and in the HTTP request the white space is url-encoded. Thus the file name in the request won't match the file name of the file version in the database.


src/Sulu/Bundle/MediaBundle/Media/FormatCache/LocalFormatCache.php
```diff
@@ -198,13 +198,11 @@ protected function getFormatFromUrl($url)
     /**
      * Return the requested image format from url.
      *
-     * @param string $url
-     *
-     * @return int
+     * @return string
```
alexander-schranz, line 202:
```suggestion
     * @param string $url
     *
     * @return string
```

src/Sulu/Bundle/MediaBundle/Media/FormatCache/LocalFormatCache.php
```diff
@@ -198,13 +198,11 @@ protected function getFormatFromUrl($url)
...
      *
      * @throws ImageProxyInvalidUrl
      */
-    protected function getFileNameFromUrl($url)
+    protected function getFileNameFromUrl(string $url)
```
alexander-schranz, line 206: Adding typehints is for bc reasons not allowed for protected or public methods:

```suggestion
    protected function getFileNameFromUrl($url)
```

src/Sulu/Bundle/MediaBundle/Media/FormatCache/LocalFormatCache.php
```diff
@@ -213,6 +211,6 @@ protected function getFileNameFromUrl($url)
             throw new ImageProxyInvalidUrl('No `filename` was found in the url');
         }
 
-        return $fileName;
+        return \urldecode($fileName);
```
alexander-schranz: Looks like we are using `rawurlencode` here: https://github.com/sulu/sulu/blob/7b1490639e12d21d00a0a743b12f8aec833e84d9/src/Sulu/Bundle/MediaBundle/Media/FormatCache/LocalFormatCache.php#L119

Then we should also use `rawurldecode`:

```suggestion
        return \rawurldecode($fileName);
```

alexander-schranz (changes requested):
Good catch 👍  Can you also add a test case to the `LocalFormatCacheTest` test class

src/Sulu/Bundle/MediaBundle/Media/FormatCache/LocalFormatCache.php
stollr: Please correct my if I am wrong, but as far as I know, contravariant parameter types are safe. If the overriding class did not specify a parameter type there is no issue (see [example](https://3v4l.org/4Lj2s)), because then its type is wider than the string type.

Adding a return type is a BC problem of course.

alexander-schranz: Oh that should work. Remember there were some bugs in some PHP 7.x versions but as we are on 8.0 for Sulu 2.5 nobody should get into that bug longer.

alexander-schranz:
@stollr do you also have time to add the test case now? As we plan a release for today which I want to put this also in it.

stollr:
@alexander-schranz I am working on it. Is a unit test enough or do you also require functional/application tests?

alexander-schranz:
@stollr Unit test in the localformatcachetest is enough.

Same as the encoded test: https://github.com/sulu/sulu/blob/7b1490639e12d21d00a0a743b12f8aec833e84d9/src/Sulu/Bundle/MediaBundle/Tests/Unit/LocalFormatCacheTest.php#L62

alexander-schranz (approved):
Thx did a little refactoring of the tests.

stollr:
Thank your for your support and merging ;-)

PR #7544, opened 2024-07-25
Removing trailing whitespaces from UPGRADE.md

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?
Formatting the UPGRADE.md

#### Why?
Most editors format the file anyways making it harder to keep the diffs clean. Furthermore git also complains about trailing whitespaces when comitting

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -4,7 +4,7 @@
 
 ### Stricter Image Format Url Handling
 
-The image formats URL requires an exact filename match to retrieve the correct image format. 
+The image formats URL requires an exact filename match to retrieve the correct image format.
 Old versions will be redirected to the new version and any non-matching filenames will now return a 404 error.
 
 ## 2.6.3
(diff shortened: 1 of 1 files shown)


alexander-schranz:
That makes more problem with backmerges conflicts and make `git blame`  not work to find fast referenced merger request so I would not adjust here something.

PS: markdown ending double spaces = line breaks and so your IDE should never strip markdown ending spaces.

See also: https://github.com/sulu/sulu/blob/7b1490639e12d21d00a0a743b12f8aec833e84d9/.editorconfig#L19-L20

mamazu:
That's a valid point. I'll just disable it for markdown.

PR #7550, opened 2024-07-29
Fix block property call

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no 
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

This fix block property call which was accidently introduced when merging 2.5 -> 2.6 with https://github.com/sulu/sulu/pull/7548

#### Why?

False arguments given to constructor call.

### DIFF
--- a/src/Sulu/Component/Content/Compat/Block/BlockProperty.php
+++ b/src/Sulu/Component/Content/Compat/Block/BlockProperty.php
@@ -45,7 +45,6 @@ class BlockProperty extends Property implements BlockPropertyInterface
             $mandatory,
             $multilingual,
             $maxOccurs,
-            $maxOccurs,
             $minOccurs,
             $params,
             $tags,
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -33916,26 +33916,6 @@ parameters:
 			path: src/Sulu/Component/Content/Compat/Block/BlockProperty.php
 
 		-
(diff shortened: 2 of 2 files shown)


src/Sulu/Component/Content/Compat/Block/BlockProperty.php
```diff
@@ -45,7 +45,6 @@ public function __construct(
             $mandatory,
             $multilingual,
             $maxOccurs,
-            $maxOccurs,
```
mamazu, line 48: Maybe we could use named arguments here. That would make errors easier to spot and prevent this from happening in the future.

alexander-schranz: phpstan would also spot it so the error was just the big phpstan baseline update.

mamazu: Ah okay.

alexander-schranz: its just that the CI runs after we merged back to 2.6 and not in a merge request. Some make 2.5-2.6 backmerge branches but think that would be overhead. so typed properties would in this case not catched the error and the current CI did already catch it ;)

PR #7557, opened 2024-08-02
Fix portalInformaton handling in RequestListener when UserBasedContext is active

### LINKED ISSUE #7551: RequestListener portal information not work with User Context Based Caching
| Q | A
| --- | ---
| Sulu Version | 2.5.10
| PHP Version | -
| DB Version | -
| Browser Version | -

# Actual Behavior

RequestListener portal information not work with UserContextBased caching.

Even the reuqest analyzer is disabled the analyzer still runs and set a prefix here:

https://github.com/sulu/sulu/blob/43fa5d35eb0b97ae48470846289d33c1104d8a9c/src/Sulu/Bundle/WebsiteBundle/Routing/RequestListener.php#L43-L56

As a prefix was already set for `/_fos_http_cache_user_context` which ends in locale `en`.
The setting is skipped for the real call of `/zh/login`.

# Expected Behavior

It should still work.

# Steps to Reproduce

Webspace:

```xml
<url language="en">{host}</url>
<url language="zh">{host}/zh</url>
```

`/zh` renderens `sulu_community.login` as `/login` instead of `/zh/login`

As reported by @Quehnie

# Possible Solutions

User context controller should may not call the request analyzer.

A. Solution a do not call it if RequestAnalyzer is set to false (requires documentation)
B. Always set the `prefix` (is may an issue with subrequest ESI, ...)
C. Always set it if the request is a `MainRequest` this may the best solution as it fixes for existing projects without adoption.

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no  
| Fixed tickets | fixes #7551 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Only set the portal information if its a main request and avoid for sub requests like `/_fos_user_context_hash`.

#### Why?

The `/_fos_user_context_hash` does not contain a locale prefix or portal prefix and as previously set prefix and host currently wins the `/_fos_user_context_hash` sets the default locale and the main request will not overwrite it correctly. By check for `isMainRequest` we can avoid this.

This bug so only appeared when a SubRequest was called before the MainRequest. In most cases MainRequest is the first and set the prefix and host correctly. But in case of HttpCache it wasn't.

### DIFF
--- a/src/Sulu/Bundle/WebsiteBundle/Routing/RequestListener.php
+++ b/src/Sulu/Bundle/WebsiteBundle/Routing/RequestListener.php
@@ -15,12 +15,19 @@ use Sulu\Component\Webspace\Analyzer\RequestAnalyzerInterface;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 use Symfony\Component\HttpKernel\Event\RequestEvent;
 use Symfony\Component\HttpKernel\KernelEvents;
-use Symfony\Component\Routing\RouterInterface;
+use Symfony\Component\Routing\RequestContextAwareInterface;
 
+/**
+ * Create your own RequestListener with lower priority if you need to change the behaviour of this one.
+ *
+ * @internal
+ *
+ * @final
+ */
 class RequestListener implements EventSubscriberInterface
 {
     /**
-     * @var RouterInterface
+     * @var RequestContextAwareInterface
      */
     private $router;
 
@@ -29,7 +36,7 @@ class RequestListener implements EventSubscriberInterface
      */
     private $requestAnalyzer;
 
-    public function __construct(RouterInterface $router, RequestAnalyzerInterface $requestAnalyzer)
+    public function __construct(RequestContextAwareInterface $router, RequestAnalyzerInterface $requestAnalyzer)
     {
         $this->router = $router;
         $this->requestAnalyzer = $requestAnalyzer;
@@ -37,13 +44,19 @@ class RequestListener implements EventSubscriberInterface
 
     public static function getSubscribedEvents(): array
     {
-        return [KernelEvents::REQUEST => ['onRequest', 31]];
+        return [KernelEvents::REQUEST => ['onRequest', 31]]; // directly after the RouterListener where the portal information is set
     }
 
-    public function onRequest(RequestEvent $event)
+    public function onRequest(RequestEvent $event): void
     {
         $context = $this->router->getContext();
         $portalInformation = $this->requestAnalyzer->getPortalInformation();
+
+        if (!$event->isMainRequest()) {
+            // sub requests like /_fos_user_context_hash ends in false prefix and should not set anything here
+            // else it can happen that the main request get the false prefix and generate urls incorrect.
+            return;
+        }
 
         if ($portalInformation) {
             if (!$context->hasParameter('prefix')) {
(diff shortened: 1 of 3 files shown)


src/Sulu/Bundle/WebsiteBundle/Routing/RequestListener.php
```diff
@@ -15,12 +15,19 @@
...
+ *
+ * @internal
+ *
+ * @final
+ */
 class RequestListener implements EventSubscriberInterface
 {
     /**
-     * @var RouterInterface
+     * @var RequestContextAwareInterface
```
alexander-schranz, line 30: This Interface does have the `getContext()` part which we require here.

alexander-schranz:
Damn looks like I forgot to enable the context route:

```yaml
user_context_hash:
    path: /_fos_user_context_hash
```

If that correctly enabled the request is a `mainRequest` so my fix doesnt seem to work when User Based Context is enabled / configured correctly.

@dbu What is your opinion here should the context hash request be a sub request instead of main request here: https://github.com/FriendsOfSymfony/FOSHttpCache/blob/a582deb3f55f8a7efdae8ac916ef4adc285543a0/src/SymfonyCache/UserContextListener.php#L170

alexander-schranz:
@dbu okay think it need to be a main request as in case of varnish it would be a main request so we need another way to decide skipt this.

alexander-schranz:
Think I found a solution now with `if ($request->attributes->get('internalRequest', false)) {` which should also fix the case with our own audience targeting call.

alexander-schranz:
Some little change as 404 ( route not found ) the request listener is called as `SUB_REQUEST` and to render the 404 page correctly we still need to keep existing logic for it by set prefix and host correctly to generate urls.

So the only change is that we not longer call the request listener when its a internalRequest (foshttpcache and audience targeting requests).

Added tests for both cases.

dbu:
> @dbu okay think it need to be a main request as in case of varnish it would be a main request so we need another way to decide skipt this.

yep, with varnish it will always be a request coming from outside symfony. i even think the CacheKernel does not do sub requests. are you sure the request is considered internal by symfony even when using varnish?

dbu:
then again, with varnish the request context would be lost i guess, because the actual request with hash will be a completely separate web request.

alexander-schranz:
> are you sure the request is considered internal by symfony even when using varnish?

Our issue is only a problem in none varnish mode / when using the symfony http cache where the /_fos_user_context_hash runs in the same request stack as the main request. In case of varnish it doesn't matter what this listener sets as in that case the main request is a real http request and so fos user context hash call does not have any sideeffects to it.

alexander-schranz:
@dbu PS: first I thought I can change sulu/skeleton from:

```diff
fos_http_cache:
    proxy_client:
        symfony:
-            use_kernel_dispatcher: true
+            use_kernel_dispatcher: false
```

But looks like the `use_kernel_dispatcher` has no effects for the UserListener. Which makes also sense as the kernel is not booted at the time the `UserListener` is called.

dbu:
yep, kernel dispatcher is used to make the invalidation requests from the application to the cache kernel as PHP calls instead of going through web requests (to avoid overhead when only a single instance of the application is running with the Symfony HttpCache). 

the hash lookup can not be shortcut i think. the concept is that we do the request with the specific request information to trigger symfony security to obtain the correct user, so that we can calculate the hash on that users roles (and/or possibly other parameters of the request)

wachterjohannes (approved):

PR #7558, opened 2024-08-03
Add some more type improvements for Admin and some more classes

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?
More type improvements

#### Why?
We want to have stricter types in 3.0 so better get starting now.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Admin/View/ToolbarAction.php
+++ b/src/Sulu/Bundle/AdminBundle/Admin/View/ToolbarAction.php
@@ -35,11 +35,17 @@ class ToolbarAction
         $this->options = $options;
     }
 
+    /**
+     * @return string
+     */
     public function getType()
     {
         return $this->type;
     }
 
+    /**
+     * @return array
+     */
     public function getOptions()
     {
         return $this->options;
--- a/src/Sulu/Bundle/AdminBundle/Admin/View/ToolbarAction.php
+++ b/src/Sulu/Bundle/AdminBundle/Admin/View/ToolbarAction.php
@@ -44,7 +44,7 @@ class ToolbarAction
     }
 
     /**
-     * @return array
+     * @return array<string, mixed>
      */
     public function getOptions()
     {
--- a/src/Sulu/Bundle/AdminBundle/Command/DownloadLanguageCommand.php
+++ b/src/Sulu/Bundle/AdminBundle/Command/DownloadLanguageCommand.php
@@ -19,6 +19,7 @@ use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Style\SymfonyStyle;
 use Symfony\Component\Filesystem\Filesystem;
 use Symfony\Contracts\HttpClient\HttpClientInterface;
+use Webmozart\Assert\Assert;
 
 class DownloadLanguageCommand extends Command
 {
@@ -83,7 +84,12 @@ class DownloadLanguageCommand extends Command
             return;
         }
 
-        $composerJson = \json_decode(\file_get_contents($this->projectDir . \DIRECTORY_SEPARATOR . 'composer.json'), true);
+        /** @var array{require: array<string, string>} $composerJson */
+        $composerJson = \json_decode(
+            \file_get_contents($this->projectDir . \DIRECTORY_SEPARATOR . 'composer.json'),
+            true,
+            flags: \JSON_THROW_ON_ERROR
+        );
         $packages = \array_keys($composerJson['require']);
         $suluPackages = \array_filter($packages, function($package) {
             return 0 === \strpos($package, 'sulu/');
@@ -95,10 +101,10 @@ class DownloadLanguageCommand extends Command
             . \DIRECTORY_SEPARATOR
             . 'admin.' . $language . '.json';
 
+        $translations = [];
         if (\file_exists($translationsFile)) {
             $translations = \json_decode(\file_get_contents($translationsFile), true);
-        } else {
-            $translations = [];
+            Assert::isArray($translations, 'Unable to read translations file into array: ' . $translationsFile);
         }
 
         $packages = [];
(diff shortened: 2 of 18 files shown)


src/Sulu/Bundle/AdminBundle/Command/DownloadLanguageCommand.php
```diff
@@ -83,29 +84,34 @@ protected function executeLanguage(InputInterface $input, OutputInterface $outpu
...
+            if (!str_starts_with($packageName, 'sulu/')) {
+                continue;
+            }
+
+            $packageTranslations = $this->downloadLanguage($output, $language, $packageName, $translationBaseUrl);
+            $packages[$packageName] = [
+                'Package' => $packageName,
                 'Translations' => \count($packageTranslations),
             ];
             $translations = \array_merge($translations, $packageTranslations);
```
mamazu, line 117: Reading the composer list of packages and iterating only once over it, seems like the smarter move that way we could also display the package version number in the table output.

And this also crashes better if the composer file contains invalid json.

src/Sulu/Bundle/AdminBundle/Command/UpdateBuildCommand.php
```diff
@@ -252,16 +254,17 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
     {
-        if (!\file_exists($this->projectDir . $path)) {
+        $content = \file_get_contents($this->projectDir . $path);
+        if (!\is_string($content)) {
             return '';
         }
 
-        return \file_get_contents($this->projectDir . $path);
+        return $content;
     }
```
mamazu, line 262: Try reading the file and when it fails return the default `''` technically a BC break because the file could exist and you just don't have read permissons then it would return an empty string which used to be false.

src/Sulu/Bundle/AdminBundle/Command/UpdateBuildCommand.php
```diff
@@ -274,10 +277,13 @@ private function getRemoteFile(string $remoteRepository, string $path)
         }
     }
 
-    private function hash($content)
+    private function hash(string $content): string
```
mamazu, line 277: It's only used with a string, so we can type it like that.

src/Sulu/Bundle/AdminBundle/Controller/AdminController.php
```diff
@@ -72,7 +74,7 @@ public function __construct(
         private array $locales,
         private array $translations,
         private string $fallbackLocale,
-        private string $collaborationInterval,
+        private int $collaborationInterval,
```
mamazu, line 75: This is not a break, because strict types = 0. We allowed numeric strings (we use it to multiply in the class) so php automatically casts it to an int. Now it's done in the constructor.

mamazu (commented):
Let me know if any of the changes need to be mentioned in the UPGRADE.md guide.

src/Sulu/Bundle/AdminBundle/Command/DownloadLanguageCommand.php
mamazu: It might even make sense to extract composer handling into it's own class for better typing. (This is not the only class that needs it.

src/Sulu/Bundle/AdminBundle/Command/UpdateBuildCommand.php
alexander-schranz: No this behaviour should not be changed. The implementation with return empty string if not exist is expected in the upgrade process.

alexander-schranz: PS: type change pull requests should not change any logic ;) We should avoid hidden logic changes in such big pull requests.

mamazu: Yeah this escalated a little. But I'll revert it.

src/Sulu/Bundle/AdminBundle/Command/DownloadLanguageCommand.php
mamazu: Reverted.

alexander-schranz (approved):

src/Sulu/Component/Content/Document/Subscriber/ShadowCopyPropertiesSubscriber.php
```diff
@@ -216,7 +216,7 @@ private function getLocale(string $propertyName): string
     {
         \preg_match('/i18n:(?P<locale>.+)-shadow-base/', $propertyName, $match);
 
-        return $match['locale'];
+        return $match['locale'] ?? '';
```
mamazu: Here a better error handling would be nice. Or just asserting that this property exists.

mamazu:
We'll have to see how #7733 impacts this when merged up.

src/Sulu/Bundle/AdminBundle/Command/UpdateBuildCommand.php
```diff
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
```
alexander-schranz: this should not be added

mamazu: Removed

src/Sulu/Bundle/AdminBundle/Admin/View/ToolbarAction.php
```diff
@@ -35,11 +35,17 @@ public function __construct(string $type, array $options = [])
...
+    /**
+     * @return string
+     */
     public function getType()
     {
         return $this->type;
     }
 
+    /**
+     * @return array
```
alexander-schranz: atleast newer version of phpstan forces generic definitions

```suggestion
     * @return array<string, mixed>
```

alexander-schranz (changes requested):

alexander-schranz (approved):

alexander-schranz:
@mamazu Following classes may are not updated on backmerge to 3.0 as they were moved or not longer exists:

```
	von uns gelöscht:       src/Sulu/Bundle/AdminBundle/FormMetadata/FormMetadata.php
	von uns gelöscht:       src/Sulu/Component/Content/Metadata/ItemMetadata.php
```

https://github.com/sulu/sulu/blob/3.0/src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/FormMetadata.php

https://github.com/sulu/sulu/blob/3.0/src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/ItemMetadata.php

Do you want to check the 3.0 branch if there are also changes required?

mamazu:
@alexander-schranz If they're deleted in 3.0 it should be safe to keep them deleted.

alexander-schranz:
> If they're deleted in 3.0 it should be safe to keep them deleted.

@mamazu see the files links in previous comment if they have already this changes or not.

PR #7564, opened 2024-08-09
Remove deprecated optional things in constructor

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | yes 
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?
Removing optional services in the parameters. They're now required.

#### Why?
Getting rid of some depredations for 3.0 would be nice.

### DIFF
--- a/src/Sulu/Bundle/WebsiteBundle/Tests/Unit/Twig/ContentTwigExtensionTest.php
+++ b/src/Sulu/Bundle/WebsiteBundle/Tests/Unit/Twig/ContentTwigExtensionTest.php
@@ -150,6 +150,10 @@ class ContentTwigExtensionTest extends TestCase
 
         $this->startPageNode->getDepth()->willReturn(3);
 
+        $requestStack = new RequestStack([
+            new Request(),
+        ]);
+
         $this->extension = new ContentTwigExtension(
             $this->contentMapper->reveal(),
             $this->structureResolver->reveal(),
@@ -158,7 +162,7 @@ class ContentTwigExtensionTest extends TestCase
             $this->logger->reveal(),
             $this->securityChecker->reveal(),
             $this->webspaceManager->reveal(),
-            $this->prophesize(RequestStack::class)->reveal(),
+            $requestStack,
             []
         );
     }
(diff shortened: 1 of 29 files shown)


src/Sulu/Component/PHPCR/PathCleanup.php
```diff
@@ -43,35 +30,17 @@ class PathCleanup implements PathCleanupInterface
...
-        }
-
-        if (\method_exists($slugger, 'withEmoji')) { // BC Layer <= Symfony 6.3
-            if (
-                !\method_exists(\Symfony\Component\String\AbstractUnicodeString::class, 'localeUpper') // BC Layer <= Symfony 7.0
-                || \class_exists(\Symfony\Component\Emoji\EmojiTransliterator::class) // Symfony >= 7.1 requires symfony/emoji
-            ) {
-                $slugger = $slugger->withEmoji();
-            }
-        }
```
alexander-schranz: This is still required.

src/Sulu/Bundle/WebsiteBundle/Tests/Unit/Twig/ContentTwigExtensionTest.php
```diff
@@ -158,7 +158,7 @@ protected function setUp(): void
             $this->logger->reveal(),
             $this->securityChecker->reveal(),
             $this->webspaceManager->reveal(),
-            null,
+            $this->prophesize(RequestStack::class)->reveal(),
```
alexander-schranz: better directlydo `new RequestStack()` no need to mock such objects

src/Sulu/Component/PHPCR/PathCleanup.php
mamazu: I've readded it .

src/Sulu/Bundle/WebsiteBundle/Tests/Unit/Twig/ContentTwigExtensionTest.php
mamazu: If we're not mocking the object though, we can not assert that calls on it have been made. Is that okay?

alexander-schranz: What is happening should normally not be tested what result of a method is should be tested. So we can test all what is important. Put different request objects into it to test different requests: `$this->requestStack->push($request)` and we can test state of current request after method was called `$this->requestStack->getCurrentRequest()`

PR #7565, opened 2024-08-09
Marking twig function as deprecated

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | yes (but they aren't new)
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?
Marking the twig function as deprecated.

#### Why?
Some Editors (PHPStorm) then show it in the twig files as strikethrough.


mamazu:
Should I fix the codestyle in this one as well? That would be an unrelated change and make the PR much bigger than it needs to be.

alexander-schranz:
@mamazu I would prefer to configure `new_with_parentheses` to `true` so we keep the `()` in the php cs fixer.

PR #7571, opened 2024-08-25
Make `Environment` type part of the constructor

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | yes
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?
* Adding a constructor to the `Environment` class (an environment without a type doesn't make sense)
* Deprecating the `$environment` property on the `Url` class (Use the portal's environment)

#### Why?
Instantiating an object of that class without setting the property makes it useless. Because when adding this to a portal (the main use of this class) the `getType` function is getting called and in 8.0 and higher causes an uninitialized property exception.

#### Example Usage
Old: 
```php
$env = new Environment();
$env->setType('prod');
```

New:
```php
$env = new Environment('prod');
```

#### To Do

### DIFF
--- a/src/Sulu/Bundle/PreviewBundle/Preview/Renderer/PreviewRenderer.php
+++ b/src/Sulu/Bundle/PreviewBundle/Preview/Renderer/PreviewRenderer.php
@@ -255,10 +255,10 @@ class PreviewRenderer implements PreviewRendererInterface
         $portal->setLocalizations([$localization]);
         $portal->setDefaultLocalization($localization);
 
-        $environment = new Environment();
-        $url = new Url($domain, $this->environment);
-        $environment->setUrls([$url]);
-        $portal->setEnvironments([$environment]);
+        $environment = new Environment($this->environment);
+        $environment->addUrl(new Url($domain));
+
+        $portal->addEnvironment($environment);
         $webspace->setPortals([$portal]);
 
         return new PortalInformation(RequestAnalyzer::MATCH_TYPE_FULL, $webspace, $portal, $localization, $domain);
--- a/src/Sulu/Component/Webspace/Environment.php
+++ b/src/Sulu/Component/Webspace/Environment.php
@@ -45,7 +45,7 @@ class Environment
     public function __construct(?string $type = null)
     {
         if (null === $type) {
-            @trigger_deprecation('sulu/sulu', '2.5', 'Not passing the type of the environment is deprecated.');
+            @trigger_deprecation('sulu/sulu', '3.0', 'Not passing the type of the environment is deprecated.');
         } else {
             $this->type = $type;
         }
@@ -60,7 +60,7 @@ class Environment
      */
     public function setType($type)
     {
-        @trigger_deprecation('sulu/sulu', '2.5', 'Using the setter to set the type of the environment is deprecated. Use the constructor instead.');
+        @trigger_deprecation('sulu/sulu', '3.0', 'Environment::setType is deprecated. Use the constructor instead.');
         $this->type = $type;
     }
 
--- a/src/Sulu/Component/Webspace/Environment.php
+++ b/src/Sulu/Component/Webspace/Environment.php
@@ -42,13 +42,25 @@ class Environment
      */
     private $mainUrl;
 
+    public function __construct(?string $type = null)
+    {
+        if (null === $type) {
+            @trigger_deprecation('sulu/sulu', '2.5', 'Not passing the type of the environment is deprecated.');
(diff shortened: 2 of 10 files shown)


src/Sulu/Component/Webspace/Loader/XmlFileLoader10.php
```diff
@@ -492,8 +492,7 @@ protected function generateEnvironments(\DOMNode $portalNode, Portal $portal)
     {
         foreach ($this->xpath->query('x:environments/x:environment', $portalNode) as $environmentNode) {
             /** @var \DOMNode $environmentNode */
-            $environment = new Environment();
-            $environment->setType($environmentNode->attributes->getNamedItem('type')->nodeValue);
+            $environment = new Environment($environmentNode->attributes->getNamedItem('type')->nodeValue);
```
mamazu, line 484: The ->nodeValue of the XML Node actually has type `string|null` So there is no BC break.

mamazu:
Can someone restart the pipeline? It should be green.

src/Sulu/Bundle/PreviewBundle/Preview/Renderer/PreviewRenderer.php
```diff
@@ -264,10 +264,10 @@ private function createPortalInformation($object, $id, $webspaceKey, $locale)
...
         $portal->setDefaultLocalization($localization);
 
-        $environment = new Environment();
-        $url = new Url($domain, $this->environment);
-        $environment->setUrls([$url]);
-        $portal->setEnvironments([$environment]);
+        $environment = new Environment($this->environment);
+        $environment->addUrl(new Url($domain));
+
+        $portal->addEnvironment($environment);
```
mamazu, line 261: Using the `addUrl` and `addEnvironment` calls here propagates the environment through to the url and the portal. It's still quite a BC layer. But maybe we can reduce this in 3.0

mamazu:
@Prokyonn This should also be an easy merge for 2.6 then I can start working on removing it in 3.0

mamazu:
@alexander-schranz This is a simmilar to #7760 or should this target 3.0 it's also mostly DX since creating an environment without the env string doesn't make sense (and can cause errors)

src/Sulu/Component/Webspace/Environment.php
```diff
@@ -42,13 +42,25 @@ class Environment
      */
     private $mainUrl;
 
+    public function __construct(?string $type = null)
+    {
+        if (null === $type) {
+            @trigger_deprecation('sulu/sulu', '2.5', 'Not passing the type of the environment is deprecated.');
```
Prokyonn: this would then also be 3.0 right?

src/Sulu/Component/Webspace/Environment.php
```diff
@@ -42,13 +42,25 @@ class Environment
...
-     * Sets the tye of this environment.
+     * Sets the type of the environment.
      *
      * @param string $type
+     *
+     * @deprecated
      */
     public function setType($type)
     {
+        @trigger_deprecation('sulu/sulu', '2.5', 'Using the setter to set the type of the environment is deprecated. Use the constructor instead.');
```
Prokyonn: same

src/Sulu/Component/Webspace/Url.php
```diff
@@ -233,6 +233,8 @@ public function toArray($depth = null)
         $res['segment'] = $this->getSegment();
         $res['redirect'] = $this->getRedirect();
         $res['main'] = $this->isMain();
+
+        // @deprecated: We should not rely on this anymore
         $res['environment'] = $this->getEnvironment();
```
mamazu, line 243: @Prokyonn I don't know how we want to handle changes in the API output, this property should always be null from now on. And later be removed.

mamazu:
@Prokyonn Is it worth keeping this around or just we close this until Sulu 4?

Prokyonn:
> @Prokyonn Is it worth keeping this around or just we close this until Sulu 4?

@mamazu hmm if we really have some value out of it, I think we can discuss about adding this in Sulu 3, as those classes are most likely not used by projects or third party bundles. WDYT @alexander-schranz ?

mamazu:
Well not really the goal would be to remove the environment structure completely and only load the clues for the current environment in the config. Because I don't see why the prod env should have access to the dev values in the first place. But that would probably be a bc break.

PR #7586, opened 2024-09-13
Add raw href value to view data of Link Content Type

manuxi:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no

#### What's in this PR?

Added uuid and linkitems to the view of link content type.

#### Why?

When using the link content type it was not possible to me to get title/contents of the page easily since only the url was provided.

### DIFF
--- a/src/Sulu/Component/Content/Types/Link.php
+++ b/src/Sulu/Component/Content/Types/Link.php
@@ -124,7 +124,7 @@ class Link extends SimpleContentType
         $userId,
         $webspaceKey,
         $languageCode,
-        $segmentKey = null
+        $segmentKey = null,
     ): void {
         $property->setValue(\json_decode($value, true));
         $this->write($node, $property, $userId, $webspaceKey, $languageCode, $segmentKey);
--- a/src/Sulu/Component/Content/Types/Link.php
+++ b/src/Sulu/Component/Content/Types/Link.php
@@ -80,6 +80,13 @@ class Link extends SimpleContentType
             $result['rel'] = $value['rel'];
         }
 
+        if (null !== $value['provider'] && self::LINK_TYPE_EXTERNAL !== $value['provider'] && null !== $value['href']) {
+            $provider = $this->providerPool->getProvider($value['provider']);
+            $locale = $property->getStructure()->getLanguageCode();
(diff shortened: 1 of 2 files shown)


src/Sulu/Component/Content/Types/Link.php
```diff
@@ -80,6 +80,13 @@ public function getViewData(PropertyInterface $property): array
             $result['rel'] = $value['rel'];
         }
 
+        if (null !== $value['provider'] && self::LINK_TYPE_EXTERNAL !== $value['provider'] && null !== $value['href']) {
+            $provider = $this->providerPool->getProvider($value['provider']);
+            $locale = $property->getStructure()->getLanguageCode();
+            $result['uuid'] = $value['href'];
+            $result['linkitems'] = $provider->preload([$value['href']], $locale, true);
+        }
+
```
alexander-schranz: `view` should not resolve anything it should mostly just provide raw informations.

We should inside the `view` just return `href` and `provider` when that is the one you want to load more data and not recalling a resolver here:

```suggestion
        $result['href'] = $value['href'] ?? null;
```

Then it up to the user itself if the want to call something via a twig extension or not.

alexander-schranz (changes requested):
@manuxi thx for the pull request. I know that for your usecase you might want to get the linkItems but that would in most cases an overhead. So instead of returning the link items the view should just give you all information that you can call the resolver yourself via a custom twig extension. So return `href` you require for your usecase. The `provider` should already be given as I see the lines above, so this might enough to get all information you can process then:

alexander-schranz:
@manuxi Can you adjust the Link Content Type Test to test on the new value and run `composer php-cs-fix -- src/Sulu/Component/Content/Types/Link.php" to fix the code style.

src/Sulu/Component/Content/Types/Link.php
```diff
@@ -124,7 +124,7 @@ public function importData(
         $userId,
         $webspaceKey,
         $languageCode,
-        $segmentKey = null,
+        $segmentKey = null
```
manuxi: Seems that the cs-fixer added a comma in line 127 after $segementKey = null

PR #7590, opened 2024-09-19
Fix default locale in SSO when no 'locale' attribute is provided

### LINKED ISSUE #7444: OpenId SSO default locale user
| Q | A
| --- | ---
| Sulu Version | 2.6.2
| PHP Version | 8.3.7
| DB Version | PGSQL
| Browser Version | Brave 1.66

# Actual Behavior

Hello,

I did the upgrade to sulu 2.6 and also the OpenId Integration with our Azure active directory for authentification.
I just have a problem with the default locale ("en" or "de") for each connection.

In the class OpenIdSingleSignOnAdapter, at line 277, at each connection the locale is set to the default value "en". But in my project i have define the default value to "fr" because we maintain our website only with "fr" locale.
I have to set "fr" again in DB to give the right access to my user.
Currently it is impossible to use for me, because at each connection the bdd field is set to "en" and erase my "fr".

Also when a new user connect, the default role locale is set to ["en", "de"] (line 269) instead of the global variable defaultLocale "fr". So i have to manualy fix this every time.

# Expected Behavior

Use the global variable defaultLocale for default locale when user connect.

Or can i have a workaround to force default locale to fr with Azure. The attribute "locale" doesn't exist in the call getUserInfo at Azure.

steeven-th:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #7444
| Related issues/PRs | ❌
| License | MIT
| Documentation PR | ❌

#### What's in this PR?

This PR fixes the behavior in the SSO system where the user's locale is always reset to `'en'` when no `locale` attribute is provided. Now, if the user already has a locale stored in the database, it will not be overwritten.

#### Why?

Currently, when a user logs in via SSO and the `locale` attribute is missing, their locale is reset to `'en'` even if they have a locale already set in the database. This PR ensures that the user's locale remains unchanged unless it's not set in the database and no valid locale is provided.

#### Example Usage

```php
// Example of current behavior
$locale = (isset($attributes['locale']) && \in_array($attributes['locale'], $this->translations, true)) ? $attributes['locale'] : 'en';
$user->setLocale($locale);

// The new behavior will ensure the locale is not reset if the user already has one
$locale = (isset($attributes['locale']) && \in_array($attributes['locale'], $this->translations, true)) ? $attributes['locale'] : ($user instanceof User && $user->getLocale() ? $user->getLocale() : 'en');
$user->setLocale($locale);
```

### DIFF
--- a/src/Sulu/Bundle/SecurityBundle/SingleSignOn/Adapter/OpenId/OpenIdSingleSignOnAdapter.php
+++ b/src/Sulu/Bundle/SecurityBundle/SingleSignOn/Adapter/OpenId/OpenIdSingleSignOnAdapter.php
@@ -272,7 +272,7 @@ class OpenIdSingleSignOnAdapter implements SingleSignOnAdapterInterface
         $contact = $user->getContact();
         $contact->setFirstName($attributes['given_name'] ?? '');
         $contact->setLastName($attributes['family_name'] ?? '');
-        $locale = (isset($attributes['locale']) && \in_array($attributes['locale'], $this->translations, true)) ? $attributes['locale'] : 'en';
+        $locale = (isset($attributes['locale']) && \in_array($attributes['locale'], $this->translations, true)) ? $attributes['locale'] : ($user instanceof User && $user->getLocale() ? $user->getLocale() : 'en');
         $user->setLocale($locale);
 
         $this->entityManager->flush();


alexander-schranz:
@steeven-th thank you!

steeven-th:
@alexander-schranz
Hello
How can I be listed as a contributor?
Do I need more PR and more contributions?
thx

alexander-schranz:
Github takes care of the contributers view. I'm not sure in which interval they  update it: https://github.com/sulu/sulu/graphs/contributors

The only thing you need make sure that your commit email address is connected with your github account you can add multiple email adressesin your settings.

PR #7599, opened 2024-10-05
Improve doc types for interfaces

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

More types

#### Why?
Preparation for stricter typing

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -1165,30 +1165,6 @@ parameters:
 			path: src/Sulu/Bundle/AdminBundle/Controller/CollaborationController.php
 
 		-
-			message: '#^Method Sulu\\Bundle\\AdminBundle\\Controller\\CollaborationController\:\:cdeleteAction\(\) has no return type specified\.$#'
-			identifier: missingType.return
-			count: 1
-			path: src/Sulu/Bundle/AdminBundle/Controller/CollaborationController.php
-
-		-
-			message: '#^Method Sulu\\Bundle\\AdminBundle\\Controller\\CollaborationController\:\:cputAction\(\) has no return type specified\.$#'
-			identifier: missingType.return
-			count: 1
-			path: src/Sulu/Bundle/AdminBundle/Controller/CollaborationController.php
-
-		-
-			message: '#^Method Sulu\\Bundle\\AdminBundle\\Controller\\CollaborationController\:\:createCollaboration\(\) has no return type specified\.$#'
-			identifier: missingType.return
-			count: 1
-			path: src/Sulu/Bundle/AdminBundle/Controller/CollaborationController.php
-
-		-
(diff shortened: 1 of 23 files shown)


mamazu:
@alexander-schranz This should also be an easy merge. Because it only (except for one) is just doctypes.

src/Sulu/Component/Webspace/Analyzer/Attributes/RequestProcessorInterface.php
```diff
@@ -29,7 +29,7 @@ public function process(Request $request, RequestAttributes $requestAttributes);
      * Validate the given attributes and return true if it is valid,
      * if not throw a specific exception.
      *
-     * @return bool
+     * @return true
```
alexander-schranz: should be `bool` not hardcoded true

mamazu: All implementations always return true or throw an exception. Maybe adding a throws annotation to that.

src/Sulu/Component/Webspace/Analyzer/Attributes/RequestProcessorInterface.php
```diff
@@ -29,7 +29,9 @@ public function process(Request $request, RequestAttributes $requestAttributes);
...
```
alexander-schranz:
```suggestion
     * @return bool
```

src/Sulu/Component/Webspace/Analyzer/Attributes/RequestProcessorInterface.php
alexander-schranz: still doesnt mean it can return false in a app implementation

alexander-schranz (approved):

alexander-schranz (approved):

phpstan.neon
```diff
@@ -28,6 +28,7 @@ parameters:
         - %currentWorkingDirectory%/packages/content/tests/Unit/Mocks/*
         - %currentWorkingDirectory%/*/Tests/app/cache/*
         - %currentWorkingDirectory%/*/Tests/app/var/*
+        - %currentWorkingDirectory%/packages/*/tests*
```
alexander-schranz:
```suggestion
```

src/Sulu/Component/Webspace/Analyzer/Attributes/RequestProcessorInterface.php
mamazu: Reverted, but no matter what this function returns as long as it isn't an exception then it's valid.

Prokyonn (approved):

PR #7600, opened 2024-10-05
Remove deprecated `Urlizer`

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | yes
| Deprecations? | no (already deprecated)
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?
Replacing the Urlizer with the service that the deprecation already states.

#### Why?
No need to keep the old code.

### Long term
We're already using the `symfony/string` package which has a slugger function. Maybe we could use that.


alexander-schranz:
@mamazu Thank you!

PR #7602, opened 2024-10-07
Improve performance with reevaluation of only changed blocks

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| License | MIT

#### What's in this PR?

This PR prevents reevaluating all icon conditions of all blocks, when one block is changed. Only the condition for the changed block will be reevaluated.

#### Why?

Projects that have 50+ blocks and multiple conditions regarding icons from the blocksettings run into performance issues, due to the mass reevaluation of the conditions.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/tests/__snapshots__/FieldBlocks.test.js.snap
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/tests/__snapshots__/FieldBlocks.test.js.snap
@@ -47,9 +47,6 @@ exports[`Render block with schema 1`] = `
           class="header"
         >
           <div
-            class="icons"
-          />
-          <div
             class="iconButtons"
           >
             <button
@@ -174,9 +171,6 @@ exports[`Render block with schema 1`] = `
         <header
           class="header"
         >
-          <div
-            class="icons"
-          />
           <div
             class="iconButtons"
           >
@@ -335,9 +329,6 @@ exports[`Render block with schema and error on fields already being modified 1`]
           class="header"
         >
           <div
-            class="icons"
-          />
-          <div
             class="iconButtons"
           >
             <button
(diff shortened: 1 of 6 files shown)


Prokyonn:
@alexander-schranz Please take a look at the changes and tell me if this would be a viable approach for you

alexander-schranz (commented):
as discussed maybe try out fast-deep-equal to call for each block. We need sure some describing tests else this logic is too complex to just be added.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/tests/__snapshots__/FieldBlocks.test.js.snap
```diff
@@ -46,9 +46,6 @@ exports[`Render block with schema 1`] = `
         <header
           class="header"
         >
-          <div
-            class="icons"
-          />
```
alexander-schranz, line 51: are the icons now skipped on initial rendering? Or why is this icons div now away?

Prokyonn: > are the icons now skipped on initial rendering? Or why is this icons div now away?

I think this is because we skip the rendering now completely on this line `Object.keys(this.iconsMapping).length === 0` when the iconsMapping is empty. Previously we did return an empty array which did render the icons div but without icons.

alexander-schranz: I'm confused why would the before method with `this.value.map((value) => Object.keys(this.iconsMapping).reduce((icons, pointer) => {` not return `[]`? What did it return before?

Because the div isn't created in the early return method.

Prokyonn: Ah sorry I did not explain it correctly, it returned empty arrays in an empty array, like `[ [], [] ]`

PR #7607, opened 2024-10-09
Fix nested file inspection

### LINKED ISSUE #7604: UploadFileSubscriber causes problem for multi-dimensional file array
| Q | A
| --- | ---
| Sulu Version | 2.6.5
| PHP Version | 8.2
| DB Version | irrelevant

# Actual Behavior

Our application uses a file upload function within EasyAdmin (which we use besides Sulu), where file upload fields are used in array shape. This is how the `files` array looks like when debugged in PhpStorm:

grafik

As you can see, the structure is a multi-dimensional array, and not only one with simple keys and files as values. As the `UploadFileSubscriber` can't handle such a structure, the request fails pretty hard. And as Sulu binds this pretty close on **any** incoming request, there's no nice way to get around this (for example, as we don't need it in our application, or as any form request from the frontend should not trigger this)

# Expected Behavior

The `UploadFileSubscriber` would either try to work on a multi-dimensional structure, or skip checking the content of this array.

# Steps to Reproduce

Use any kind of form with upload fields with multi-dimensional structure, like 

```
<input type="file" name="Product[thumbnail][file]">
```

and trigger an upload.

# Possible Solutions

- put this more closely into Sulu's own responsibilties (such that this check is only triggered when run from within Sulu code)
- detect whether the FILES array is multi-dimensional, and handle this structure properly in the subscriber

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #7604 
| Related issues/PRs | none
| License | MIT
| Documentation PR | none

#### What's in this PR?

This PR fixes the issue #7604 which was introduced in 2.6.5 and the SVG validation.

#### Why?

There was not encountered that the files structure in the request can be nested.

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/FileInspector/UploadFileSubscriber.php
+++ b/src/Sulu/Bundle/MediaBundle/FileInspector/UploadFileSubscriber.php
@@ -75,7 +75,7 @@ final class UploadFileSubscriber implements EventSubscriberInterface
                 continue;
             }
 
-            $mimeType = $file->getClientMimeType();
+            $mimeType = $file->getMimeType() ?: $file->getClientMimeType();
             foreach ($this->fileInspectors as $fileInspector) {
                 if (null !== $mimeType && $fileInspector->supports($mimeType)) {
                     try {
--- a/src/Sulu/Bundle/MediaBundle/FileInspector/UploadFileSubscriber.php
+++ b/src/Sulu/Bundle/MediaBundle/FileInspector/UploadFileSubscriber.php
@@ -51,21 +51,34 @@ final class UploadFileSubscriber implements EventSubscriberInterface
     {
         $request = $event->getRequest();
 
-        /**
-         * @var string $key
-         * @var UploadedFile $file
-         */
-        foreach ($request->files as $key => $file) {
+        $request->files->replace($this->inspectFiles($request->files->all()));
+    }
+
+    /**
+     * @param array<string, mixed> $files
+     *
+     * @return array<string, mixed>
+     */
+    private function inspectFiles(array $files): array
+    {
+        foreach ($files as $key => $file) {
+            if (\is_array($file)) {
+                $files[$key] = $this->inspectFiles($file);
+                continue;
+            }
+
+            $mimeType = $file->getClientMimeType();
             foreach ($this->fileInspectors as $fileInspector) {
-                $mimeType = $file->getClientMimeType();
                 if (null !== $mimeType && $fileInspector->supports($mimeType)) {
                     try {
-                        $request->files->set($key, $fileInspector->inspect($file));
+                        $files[$key] = $fileInspector->inspect($file);
                     } catch (UnsafeFileException $exception) {
                         throw new BadRequestHttpException($exception->getMessage(), $exception);
                     }
                 }
             }
         }
+
+        return $files;
     }
 }
--- a/src/Sulu/Bundle/MediaBundle/FileInspector/UploadFileSubscriber.php
+++ b/src/Sulu/Bundle/MediaBundle/FileInspector/UploadFileSubscriber.php
@@ -61,9 +61,17 @@ final class UploadFileSubscriber implements EventSubscriberInterface
      */
     private function inspectFiles(array $files): array
     {
+        /**
+         * @var string $key
+         * @var UploadedFile|array<string, mixed>|null $file
+         */
         foreach ($files as $key => $file) {
             if (\is_array($file)) {
                 $files[$key] = $this->inspectFiles($file);
+                continue;
+            }
+
+            if (null === $file) {
                 continue;
             }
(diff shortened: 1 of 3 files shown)


chirimoya (approved):

chirimoya:
@wachterjohannes Linting failed. Please check.

src/Sulu/Bundle/MediaBundle/FileInspector/UploadFileSubscriber.php
```diff
@@ -51,21 +51,34 @@ public function onKernelRequest(RequestEvent $event): void
...
+     */
+    private function inspectFiles(array $files): array
+    {
+        foreach ($files as $key => $file) {
+            if (\is_array($file)) {
+                $files[$key] = $this->inspectFiles($file);
+                continue;
+            }
+
+            $mimeType = $file->getClientMimeType();
```
mamazu, line 70: We're using this in production and it crashes with "Uncaught Error: Call to a member function getClientMimeType() on null"

mamazu:
```suggestion
            $mimeType = $file?->getClientMimeType();
```

We're using this in production and it crashes with "Uncaught Error: Call to a member function getClientMimeType() on null"

mamazu:
In my opinion this should only be run in the admin section of the page and not also effect user content, but that's a different issue.

src/Sulu/Bundle/MediaBundle/FileInspector/UploadFileSubscriber.php
mhnrm: Yes, the original line will surely crash if $file is null, so your change makes sense.

wachterjohannes: @mamazu @mhnrm done :)

wachterjohannes:
> In my opinion this should only be run in the admin section of the page and not also effect user content, but that's a different issue.

@alexander-schranz @chirimoya what do you think about that?

alexander-schranz:
> In my opinion this should only be run in the admin section of the page and not also effect user content, but that's a different issue.

No this should run in always, if somebody wants that they can decorate the service but as it is security related this should run for all Admin and Website uploads, specially more for Website uploads as Admin users mostly don't hack there systems :)

.github/workflows/test-application.yaml
```diff
@@ -15,7 +15,7 @@ concurrency:
 jobs:
     js-css:
         name: "Node ${{ matrix.node-version }}"
-        runs-on: ubuntu-latest
+        runs-on: ubuntu-20.04
```
alexander-schranz: this should be done on 2.5 branch

wachterjohannes: done https://github.com/sulu/sulu/pull/7615

src/Sulu/Bundle/MediaBundle/Tests/Unit/SvgInspector/UploadFileSubscriberTest.php
```diff
@@ -67,6 +78,54 @@ public function testOnKernelRequestWithSafeSvg(): void
...
+    public function testOnKernelRequestWithNestedFiles(): void
+    {
+        $mimeType = 'image/svg+xml';
+        $uploadedFile = new UploadedFile('test.svg', 'test.svg', $mimeType, 0, true);
+
+        $request = new Request([], [], [], [], ['Product' => ['thumbnail' => $uploadedFile]]);
+        $event = $this->createRequestEvent($request);
+
+        $this->svgInspector->supports($mimeType)->willReturn(true);
+        $this->svgInspector->inspect($uploadedFile)->willReturn($uploadedFile);
```
mamazu: As far as I know this this is missing `shouldBeCalled` calls. Because just specifying the return type of the mock doesn't also assert that this function is actually used.

stollr:
@alexander-schranz Sorry for being off-topic but is this security fix really helpful? It relies on the mime-type sent by the client. As this could be faked easily, is this check really safe?

Apart from that the bugfix works like expected. Thanks @wachterjohannes.

alexander-schranz:
@stollr why I think if you send false mimetype then the mimetype will in our case also be used as return mime type which means SVG will not be rendered as SVG and so no JS code executed. Still I think it make sense to use `getMimeType` instead of `getClientMimeType` if that works. If getMimeType may can not detect probably we could fallback to getClientMimeType. Thx for your thoughts here.

/cc @wachterjohannes

alexander-schranz:
is this already handled the same was in 2.5 branch? (mime type, nested objects, ..)

alexander-schranz (approved):

PR #7609, opened 2024-10-10
Fix OpenIdSingleSignOnAdapter for custom class overrides using userRepository

Cephra:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

I've changed a line in `src/Sulu/Bundle/SecurityBundle/SingleSignOn/Adapter/OpenId/OpenIdSingleSignOnAdapter.php` so it uses the `UserRepository` to create a new Instance of a User instead of just constructing a new Sulu User object.

#### Why?

Without doing that, it'll always be the Sulu User and any custom overrides you may have done will cause an error due to the classes not matching.

#### Example Usage

When defining a custom User class like so:

```yaml
sulu_security:
    # ...
    objects:
        user:
            model: App\Entity\User
```

The old code would not create an instance of this class, but rather the Sulu User class, causing an Exception further down the line. 

Using the userRespository circumvents that by ensuring the custom class is used instead.

### DIFF
--- a/src/Sulu/Bundle/SecurityBundle/SingleSignOn/Adapter/OpenId/OpenIdSingleSignOnAdapter.php
+++ b/src/Sulu/Bundle/SecurityBundle/SingleSignOn/Adapter/OpenId/OpenIdSingleSignOnAdapter.php
@@ -19,6 +19,7 @@ use Sulu\Bundle\SecurityBundle\Entity\User;
 use Sulu\Bundle\SecurityBundle\Entity\UserRole;
 use Sulu\Bundle\SecurityBundle\SingleSignOn\SingleSignOnAdapterInterface;
 use Sulu\Component\Security\Authentication\RoleRepositoryInterface;
+use Sulu\Component\Security\Authentication\UserInterface;
 use Sulu\Component\Security\Authentication\UserRepositoryInterface;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpKernel\Exception\HttpException;
--- a/src/Sulu/Bundle/SecurityBundle/SingleSignOn/Adapter/OpenId/OpenIdSingleSignOnAdapter.php
+++ b/src/Sulu/Bundle/SecurityBundle/SingleSignOn/Adapter/OpenId/OpenIdSingleSignOnAdapter.php
@@ -238,7 +238,7 @@ class OpenIdSingleSignOnAdapter implements SingleSignOnAdapterInterface
         $user = $this->userRepository->findOneBy(['email' => $email]);
 
         if (!$user instanceof User) {
-            /** @var UserInterface $user */
+            /** @var User $user */
             $user = $this->userRepository->createNew();
             $user->setEmail($email);
             $user->setUsername($email);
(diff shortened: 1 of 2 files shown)


Cephra:
@alexander-schranz Hey, turns out I forgot to add the import for UserInterface, sorry about that.

Cephra:
@alexander-schranz I will change this back to draft as I'm not sure why I get the errors that I'm getting. We are using these fixes already in our sulu stack without any issues. I'll investigate further, update my PR, and set it back to ready once I solved the linting issues.

alexander-schranz (approved):

src/Sulu/Bundle/SecurityBundle/SingleSignOn/Adapter/OpenId/OpenIdSingleSignOnAdapter.php
```diff
@@ -261,7 +262,8 @@ private function createOrUpdateAdminUser(string $email, array $attributes): void
         }
 
         if (!\in_array($role->getIdentifier(), $roleNames, true)) {
-            $defaultRoleKey = new UserRole();
+            /** @var UserRole $defaultRoleKey */
+            $defaultRoleKey = $this->roleRepository->createNew();
```
alexander-schranz, line 266: this looks wrong as we try to create a `UserRole` via `RoleRepository` which returns a `Role` not a `UserRole` which is wrong. This change I need be reverted:

```suggestion
            $defaultRoleKey = new UserRole();
```

The `UserRole` is not overwriteable and so not require the repository to be created.

src/Sulu/Bundle/SecurityBundle/Tests/Unit/SingleSignOn/Adapter/OpenId/OpenIdSingleSignOnAdapterTest.php
```diff
@@ -180,11 +181,30 @@ function($method, $url, $options): MockResponse {
...
+        $user->getContact()->willReturn($contact->reveal());
+
+        $this->userRepository->createNew()->willReturn($user->reveal());
+
         $this->contactRepository->createNew()->willReturn($this->prophesize(Contact::class)->reveal());
         $this->entityManager->persist(Argument::any())->shouldBeCalled();
         $role = $this->prophesize(Role::class);
         $role->getAnonymous()->shouldBeCalled()->willReturn(false);
         $role->getIdentifier()->willReturn('hello@sulu.io');
+        $this->roleRepository->createNew()->willReturn($role->reveal());
```
alexander-schranz, line 207:
```suggestion
```

src/Sulu/Bundle/SecurityBundle/Tests/Unit/SingleSignOn/Adapter/OpenId/OpenIdSingleSignOnAdapterTest.php
```diff
@@ -180,11 +181,30 @@ function($method, $url, $options): MockResponse {
...
+        $user->setEnabled(Argument::any())->shouldBeCalled();
+        $user->setContact(Argument::any())->shouldBeCalled();
+        $user->addUserRole(Argument::any())->shouldBeCalled();
+        $user->getLocale()->shouldBeCalled();
+        $user->setLocale(Argument::any())->shouldBeCalled();
+
+        $contact = $this->prophesize(Contact::class);
+        $user->getContact()->willReturn($contact->reveal());
+
+        $this->userRepository->createNew()->willReturn($user->reveal());
```
alexander-schranz, line 200: There is no need here for such complexity of mockings just return a real user entity here:

```suggestion
        $user = new User();
        $this->userRepository->createNew()->willReturn($user);
```

alexander-schranz (changes requested):
The UserRole can not be created via RoleRepository so instead it should be directly created.

alexander-schranz (changes requested):
.

alexander-schranz:
I can not apply the patches to the your fork so I updated the changes in an own PR here #7633

PR #7611, opened 2024-10-14
Add content resolver for the new ContentBundle architecture

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Adds new `PropertyResolver` and `ResourceLoader` to use for the `ContentBundle`

#### Why?

ContentBundle will introduce a new architecture to more efficiently resolve content.

#### To Do

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -37,6 +37,7 @@
         "ext-simplexml": "*",
         "composer-runtime-api": "^2.0",
         "composer/package-versions-deprecated": "^1.8",
+        "composer/semver": "^3.4",
         "contao/imagine-svg": "^1.0",
         "doctrine/annotations": "^1.14 || ^2.0",
         "doctrine/cache": "^1.12.1",
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -7281,7 +7281,7 @@ parameters:
 			path: src/Sulu/Bundle/ContactBundle/Contact/ContactManager.php
 
 		-
-			message: "#^Offset 'id' does not exist on non\\-falsy\\-string\\.$#"
+			message: "#^Offset 'id' does not exist on non\\-empty\\-string\\.$#"
 			count: 1
 			path: src/Sulu/Bundle/ContactBundle/Contact/ContactManager.php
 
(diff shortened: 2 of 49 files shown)


alexander-schranz:
As discussed with @Prokyonn teaser and link content type are a bit differently. But we said that we will go with a `LinkLoader` and `TeaserLoader` and so a special "link" and "teaser" resource, that way we don't require a lot of changes for teaser provider and link providers and can keep them currently as they are. But all teasers and links should be loaded via one query per type that way.

alexander-schranz:
@Prokyonn for the PropertyResolvers inside bundle I would remove the `Resolver` namespace think `PropertyResolver` is enough. For ContentBundle it make sense to have that namespace for implementation I would skip it. What do you think?

```diff
-Sulu\Bundle\ContactBundle\Infrastructure\Sulu\Content\PropertyResolver\Resolver\AccountSelectionPropertyResolver
+Sulu\Bundle\ContactBundle\Infrastructure\Sulu\Content\PropertyResolver\AccountSelectionPropertyResolver
```

Prokyonn:
> @Prokyonn for the PropertyResolvers inside bundle I would remove the `Resolver` namespace think `PropertyResolver` is enough. For ContentBundle it make sense to have that namespace for implementation I would skip it. What do you think?
> 
> ```diff
> -Sulu\Bundle\ContactBundle\Infrastructure\Sulu\Content\PropertyResolver\Resolver\AccountSelectionPropertyResolver
> +Sulu\Bundle\ContactBundle\Infrastructure\Sulu\Content\PropertyResolver\AccountSelectionPropertyResolver
> ```

Yes that totally makes sense 👍

mamazu:
Just as a random thought. Could you also design the system so that it is easy to profile and debug these functions simmilar to the `DocumentManager/DebugEventDispatcher`?

alexander-schranz:
@mamazu as now all are service tags it is straight forward over the `debug:container` now:

 - Get All property resolvers: `bin/console debug:container --tag sulu_content.property_resolver`
 - Get All resource loaders: `bin/console debug:container --tag sulu_content.resource_loader`

alexander-schranz:
As discussed we will target the 3.0 for the new resolver and merge other bundles into that branch for easier maintainance, but that will be its own Pull requests.

alexander-schranz:
Moved TODO list into own issue: https://github.com/sulu/sulu/issues/7672

PR #7616, opened 2024-10-17
Log unexpected exceptions as error instead of debug in password reset request

stollr:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | yes, but very tiny
| Deprecations? | no 
| License | MIT

#### What's in this PR?

In case of any exceptions in the `sendEmailAction` of the `ResettingController ` of the SecurityBundle, only a debug log was added. This PR adds a distinction between "expected" exceptions (like user not found or too many tokens sent) and unexpected exception. The latter ones will trigger an *error* log.

The added unit tests should cover the new behaviour.

#### Why?

Debug logs have a much lower level as an error log. But unexpected exceptions should raise more attention, because there may be a bug behind it.

In our project it has hidden an error in an event listener which was called during flush event.


alexander-schranz:
@stollr Thank you!

PR #7617, opened 2024-10-17
Fix password reset if SSO is enabled but not matched domain

Quehnie:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Also check for reset route in the single sign on adapter.

#### Why?

Currently the password reset does not work.

### DIFF
--- a/src/Sulu/Bundle/SecurityBundle/SingleSignOn/SingleSignOnLoginRequestSubscriber.php
+++ b/src/Sulu/Bundle/SecurityBundle/SingleSignOn/SingleSignOnLoginRequestSubscriber.php
@@ -92,7 +92,7 @@ class SingleSignOnLoginRequestSubscriber implements EventSubscriberInterface
 
         $adapter = $this->singleSignOnAdapterProvider->getAdapterByDomain($domain);
         if (!$adapter) {
-            if ($password) {
+            if ($password || $isResetPassword) {
                 return;
             }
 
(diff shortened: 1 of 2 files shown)


src/Sulu/Bundle/SecurityBundle/SingleSignOn/SingleSignOnLoginRequestSubscriber.php
```diff
@@ -92,7 +92,7 @@ public function onKernelRequest(RequestEvent $event): void
 
         $adapter = $this->singleSignOnAdapterProvider->getAdapterByDomain($domain);
         if (!$adapter) {
-            if ($password) {
+            if ($password || $isResetPassword) {
```
alexander-schranz, line 95: Looks correct as we do the same on line 83. @martinlagler can you also check if all make sense here. I needed adopt some tests so not 100% sure if I did miss something.

alexander-schranz (approved):

martinlagler (approved):

src/Sulu/Bundle/SecurityBundle/SingleSignOn/SingleSignOnLoginRequestSubscriber.php
martinlagler: @alexander-schranz Also looks good to me, approved.

PR #7623, opened 2024-10-24
Improve error messages for regenerate-format

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?
Adding a better error message.

#### Why?
I was trying to run this command on a fresh instance where the public upload directory does not exist, so the `realpath` call returned false which due to lax typing became an empty string and produced this error message:
image

Now the error message is a little better:
image


alexander-schranz:
@mamazu thank you!

PR #7626, opened 2024-10-28
Fix translations locale issue

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

As we did load the translation in parallel to the config request, in some cases it happened that the UI was translated in two different languages, the default client language and the default backend language.

This MR fixes it, as it does not send translation/config request in parallel but in sequence to directly use the user locale from the config request.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/services/initializer/initializer.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/services/initializer/initializer.js
@@ -78,15 +78,18 @@ class Initializer {
         });
     }
 
-    initializeTranslations() {
+    initializeTranslations(userIsLoggedIn: boolean) {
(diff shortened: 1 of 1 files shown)


src/Sulu/Bundle/AdminBundle/Controller/AdminController.php
```diff
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
```
wachterjohannes: @alexander-schranz i think this would be a BC break?

Prokyonn: reverted it anyway, as we do not need backend adjustments with the new solution

PR #7632, opened 2024-10-31
Fix admin frontend initialization after login with 2FA

### LINKED ISSUE #7560: Admin frontend config and navigation is not reloaded when logging in as different user
| Q | A
| --- | ---
| Sulu Version | 2.5 branch
| PHP Version | 8.2
| Browser Version | Firefox 115.13.0esr

# Actual Behavior

When I log in to the admin area, logout again and relogin with another user (**without refreshing the site**) the `/admin/config` is not requested again and thus the navigation shows the name of the previous user and the same navigation items - even if the user should see different navigation items due to more or less privileges.

When I reload the page, the UI is correctly refreshed. So it is just a frontend issue ;-)

# Expected Behavior

The config has to be reloaded after each login and navigation has to be refreshed.

stollr:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #7560
| License | MIT

#### Why?

This PR addresses #7560.

After successfull verification of the admin login with a second factor the `handleLogin` method of the `userStore` gets a data object where the username is undefined, which makes the check `data.username === this.user.username` to be always true, because the username property of both objects are undefined in this moment.

The result is that the frontend is never reinitialized after a login with 2FA and the navigation and username is used from the previous logged in user.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/stores/userStore/userStore.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/stores/userStore/userStore.js
@@ -120,7 +120,7 @@ class UserStore {
             // when the user was logged in already and comes again with the same user
             // we don't need to initialize again
 
-            if (data.username === this.user.username) {
+            if (this.user.username && data.username === this.user.username) {
                 this.setLoggedIn(true);
                 this.setLoading(false);
 
--- a/src/Sulu/Bundle/SecurityBundle/Entity/User.php
+++ b/src/Sulu/Bundle/SecurityBundle/Entity/User.php
@@ -49,7 +49,7 @@ class User extends ApiEntity implements UserInterface, EquatableInterface, Audit
      * @var string
      *
      * @Expose
-     * @Groups({"fullUser", "profile"})
+     * @Groups({"frontend", "fullUser", "profile"})
      */
     protected $username;
 
--- a/test.svg
+++ b/test.svg
@@ -1,0 +1,1 @@
+<svg><a href="http://phishing-site.com" target="_blank"><text x="10" y="50">Click here for a prize!</text></a></svg>


src/Sulu/Bundle/AdminBundle/Resources/js/stores/userStore/userStore.js
```diff
@@ -142,10 +142,12 @@ class UserStore {
...
         if (this.user) {
-            // when the user was logged in already and comes again with the same user
-            // we don't need to initialize again
+            // When the user was logged in already and comes again with the same user
+            // we don't need to initialize again.
+            // But after 2FA verification data.username is undefined, so at this point
+            // we do not know who the new user is. So we have to initialize nevertheless.
 
-            if (data.username === this.user.username) {
+            if (data.username && data.username === this.user.username) {
```
alexander-schranz: I find this a little bit confusing as if data.username is undefined then `if (data.username === this.user.username) {` should not match.

So is it maybe more that `this.user` is already undefined that would be in my case a more logical if statement if we go this way:

```suggestion
            if (this.user && this.user.username === data.username) {
```

Can you check if that works?

stollr: `this.user` is defined. It is an object, but it does not have a `username` property. Here's a screenshot of the debugger:
grafik

stollr: Looks like the `/admin/config` route does return this user data without the username. I'll try to find out, why this happens.

**Update**: In the `AdminController::configAction` the serialization groups `['frontend', 'partialContact', 'fullView']` are used. But the `username` property of the `User`  has no matching groups defined (see [here](https://github.com/sulu/sulu/blob/3754b5da12c1a724bf06fe2ca021d3cc701b0b7f/src/Sulu/Bundle/SecurityBundle/Entity/User.php#L51)).
I suggest to add the `frontend` group to the property. WDYT?

alexander-schranz: Sounds strange is that only happening on SSO?

stollr: I am not using SSO. Or did you mean 2FA? Then yes - it only happens for 2FA. Because in normal login `data.username` is defined.

alexander-schranz: So I think to be safe we need then:

```js
if (this.user.username && this.user.username === data.username) {
```

stollr: I have changed the condition, now and added the serialization group 'frontend' to the `username` property.

alexander-schranz:
@stollr Thank you! Merged also into 2.6

PR #7633, opened 2024-10-31
Fix OpenIdSingleSignOnAdapter for custom class overrides using userRepository

alexander-schranz:
> replaces @Cephra https://github.com/sulu/sulu/pull/7609

| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Related PR | replaces #7609
| License | MIT

#### What's in this PR?

I've changed a line in `src/Sulu/Bundle/SecurityBundle/SingleSignOn/Adapter/OpenId/OpenIdSingleSignOnAdapter.php` so it uses the `UserRepository` to create a new Instance of a User instead of just constructing a new Sulu User object.

#### Why?

Without doing that, it'll always be the Sulu User and any custom overrides you may have done will cause an error due to the classes not matching.

#### Example Usage

When defining a custom User class like so:

```yaml
sulu_security:
    # ...
    objects:
        user:
            model: App\Entity\User
```

The old code would not create an instance of this class, but rather the Sulu User class, causing an Exception further down the line. 

Using the userRespository circumvents that by ensuring the custom class is used instead.

### DIFF
--- a/src/Sulu/Bundle/SecurityBundle/Tests/Unit/SingleSignOn/Adapter/OpenId/OpenIdSingleSignOnAdapterTest.php
+++ b/src/Sulu/Bundle/SecurityBundle/Tests/Unit/SingleSignOn/Adapter/OpenId/OpenIdSingleSignOnAdapterTest.php
@@ -226,12 +226,12 @@ class OpenIdSingleSignOnAdapterTest extends TestCase
         $this->assertEquals($expectedUserBadge, $result);
 
         $this->assertNotNull($persistedUser);
-        $this->assertSame($persistedUser, $persistedUserRole->getUser());
+        $this->assertSame('hello@sulu.io', $persistedUser->getEmail());
         $this->assertNotNull($persistedContact);
         $this->assertSame('Hikaru', $persistedContact->getFirstName());
         $this->assertSame('Sulu', $persistedContact->getLastName());
-        $this->assertSame('hello@sulu.io', $persistedUser->getEmail());
         $this->assertNotNull($persistedUserRole);
+        $this->assertSame($persistedUser, $persistedUserRole->getUser());
         $this->assertSame($role, $persistedUserRole->getRole());
     }
 }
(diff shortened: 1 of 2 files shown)


src/Sulu/Bundle/SecurityBundle/Tests/Unit/SingleSignOn/Adapter/OpenId/OpenIdSingleSignOnAdapterTest.php
```diff
@@ -180,20 +184,54 @@ function($method, $url, $options): MockResponse {
             ->willReturn('https://sulu.io/admin');
 
         $this->userRepository->findOneBy(Argument::any())->willReturn(null);
-        $this->contactRepository->createNew()->willReturn($this->prophesize(Contact::class)->reveal());
-        $this->entityManager->persist(Argument::any())->shouldBeCalled();
-        $role = $this->prophesize(Role::class);
```
alexander-schranz, line 185: removed the mocks we should never use mocks for entities, its unnecessary overhead

src/Sulu/Bundle/SecurityBundle/Tests/Unit/SingleSignOn/Adapter/OpenId/OpenIdSingleSignOnAdapterTest.php
```diff
@@ -180,20 +184,54 @@ function($method, $url, $options): MockResponse {
...
         $this->assertEquals($expectedUserBadge, $result);
+
+        $this->assertNotNull($persistedUser);
+        $this->assertSame($persistedUser, $persistedUserRole->getUser());
+        $this->assertNotNull($persistedContact);
+        $this->assertSame('Hikaru', $persistedContact->getFirstName());
+        $this->assertSame('Sulu', $persistedContact->getLastName());
+        $this->assertSame('hello@sulu.io', $persistedUser->getEmail());
+        $this->assertNotNull($persistedUserRole);
+        $this->assertSame($role, $persistedUserRole->getRole());
```
alexander-schranz: this was not yet tested so I added tests to check what we did persist

gitlost808: Thank you!

Cephra:
@alexander-schranz I just saw your changes to my PR, thanks a lot!

martinlagler (approved):

alexander-schranz:
@Cephra thank you.

PR #7646, opened 2024-11-09
New parameter "form_options_to_request" for the Selection form field

dev-newvisibility:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This PR adds a new parameter "form_options_to_request" to the Selection form field. This parameter allows a form option to be passed to the resource controller like the "resource_store_properties_to_request" does.

#### Why?

Our use case is to pass the current webspace to the resource controller for filtering. At the moment we're using the "resource_store_properties_to_request" parameter, but this only works in an "edit form". The "form_options_to_request" parameter can also be used in an "add form".

The SingleSelection form field already has the "form_options_to_list_options" parameter, which does something similar.

#### Example Usage

```xml
<property name="related" type="custom_selection">
  <params>
    <param name="form_options_to_request" type="collection">
      <param name="webspace" value="webspace"/>
    </param>
  </params>
</property>
```

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/Selection.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/Selection.js
@@ -56,6 +56,9 @@ class Selection extends React.Component<Props> {
                 resource_store_properties_to_request: {
                     value: unvalidatedResourceStorePropertiesToRequest = [],
                 } = {},
+                form_options_to_request: {
+                    value: unvalidatedFormOptionsToRequest = [],
+                } = {},
             },
         } = this.props;
 
@@ -75,9 +78,16 @@ class Selection extends React.Component<Props> {
         // $FlowFixMe: flow does not recognize that isArrayLike(value) means that value is an array
         const resourceStorePropertiesToRequest: Array | IObservableArray = unvalidatedResourceStorePropertiesToRequest;
 
+        if (!isArrayLike(unvalidatedFormOptionsToRequest)) {
+            throw new Error('The "form_options_to_request" schemaOption must be an array!');
+        }
+        // $FlowFixMe: flow does not recognize that isArrayLike(value) means that value is an array
+        const formOptionsToRequest: Array | IObservableArray = unvalidatedFormOptionsToRequest;
+
         this.requestOptions = this.buildRequestOptions(
             requestParameters,
             resourceStorePropertiesToRequest,
+            formOptionsToRequest,
             formInspector
         );
 
@@ -91,6 +101,7 @@ class Selection extends React.Component<Props> {
                 const newRequestOptions = this.buildRequestOptions(
                     requestParameters,
                     resourceStorePropertiesToRequest,
+                    formOptionsToRequest,
                     formInspector
                 );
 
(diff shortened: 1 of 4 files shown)


alexander-schranz:
@dev-newvisibility Thx for giving this a try. I did not yet have a deep look at the code.

I think it would make sense to add something we already have on the FormBuilder `addRouterAttributesToFormRequest` https://github.com/sulu/sulu/blob/2.6.5/src/Sulu/Bundle/AdminBundle/Admin/View/FormViewBuilderInterface.php#L41.

So I think the correct naming would be `router_attributes_to_request` maybe we can look on the form js parts how it is parsed out and do it here similar. We should implement such things then in `Selection` and `SingleSelection` to be consistent between them.

Let me know if you need anything.

dev-newvisibility:
I have to say that was my first thought (https://github.com/dev-newvisibility/sulu/tree/dev/selection-router-attributes-to-request) but then I found a [similar parameter](https://github.com/sulu/sulu/blob/6ebeaaa32dc0c9ba357232b8eeac044cf24d304c/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/SingleSelection.js#L310) in the SingleSelection and I decided to go the same way.

I'll try to implement your suggestions in the next few days.

alexander-schranz:
@dev-newvisibility Okay need deply check this option not 100% sure what it currently behave. Was thinking of `resource_store_properties_to_request` which already exists here: https://github.com/sulu/sulu/blob/6ebeaaa32dc0c9ba357232b8eeac044cf24d304c/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/SingleSelection.js#L49, https://github.com/sulu/sulu/blob/620462a0d5367cb74092e9d9007798037a87b47b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/Selection.js#L56

That does not match your usecase?

alexander-schranz:
If resource_store_properties_to_request doesn't work for you I'm fine to add this if you also put it into the `SingleSelection.js`.

dev-newvisibility:
> If resource_store_properties_to_request doesn't work for you I'm fine to add this if you also put it into the `SingleSelection.js`.

Unfortunately, for our use case, the resource store doesn't contain the necessary information in add's forms.


I've also added this option to SingleSelection.


> @dev-newvisibility Thx for giving this a try. I did not yet have a deep look at the code.
> 
> I think it would make sense to add something we already have on the FormBuilder `addRouterAttributesToFormRequest` https://github.com/sulu/sulu/blob/2.6.5/src/Sulu/Bundle/AdminBundle/Admin/View/FormViewBuilderInterface.php#L41.
> 
> So I think the correct naming would be `router_attributes_to_request` maybe we can look on the form js parts how it is parsed out and do it here similar. We should implement such things then in `Selection` and `SingleSelection` to be consistent between them.
> 
> Let me know if you need anything.

If it seems more consistent with Sulu, in this branch you can find a similar option implemented as "router_attributes_to_request": https://github.com/sulu/sulu/compare/2.6...dev-newvisibility:sulu:dev/selection-router-attributes-to-request

PR #7656, opened 2024-11-15
Remove integration of swiftmailer

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | #7653 
| License | MIT
| Documentation PR | sulu/sulu-docs# 

#### What's in this PR?

Remove integration of swiftmailer.

#### Why?

We will remove support for Symfony 5.4 as we require atleast Symfony 6.1 (we will go with 6.4) for the HttpKernel AbstractBundle class for new created bundles.

PR #7653, opened 2024-11-14
Add Article Bundle to monorepository

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? | yes
| BC breaks? | no 
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT
| Documentation PR | sulu/sulu-docs# 

#### What's in this PR?

Add Article Bundle to monorepository. This will merge article bundle into sulu/sulu current 3.0 branch (beeb08794280cbeee61c3562ed56999cac31d0bf (commit hash documented for later ignore article bundle commits in changelog)). 

> If you have to look at the core changes look in the `src` directory changes and ignore the `packages/article` which just contains all files form the article bundle.

#### Why?

This should make the maintainance of bundles easier.

PR #7657, opened 2024-11-15
Drop Support for Symfony 5.4, set min version of Symfony to ^6.4 || ^7.1 for using AbstractBundle class

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? |yes
| BC breaks? | no 
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | #7653, #7656 
| License | MIT

#### What's in this PR?

Set min version of Symfony to 6.

#### Why?

The merged article bundle in #7653 requires atleast Symfony 6.1 this will increaase the required Symfony version to 6.4.


composer.json
```diff
@@ -36,33 +36,32 @@
...
+        "friendsofsymfony/jsrouting-bundle": "^3.0",
+        "friendsofsymfony/rest-bundle": "^3.2",
         "gedmo/doctrine-extensions": "^3.0.4",
         "guzzlehttp/promises": "^1.0 || ^2.0",
         "handcraftedinthealps/goodby-csv": "^1.4",
         "handcraftedinthealps/rest-routing-bundle": "^1.0",
         "imagine/imagine": "^1.0",
         "jackalope/jackalope": "^1.4.5 || ^2.0",
-        "jms/serializer": "^3.16",
+        "jms/serializer": "^3.19",
```
alexander-schranz: I increased the version to the one which was installed when using `--prefer-lowest` e.g. doctrine/dbal 2 is not longer supported by our dependencies so we can increase it also in our own composer.json.

composer.json
```diff
@@ -36,33 +36,32 @@
...
-        "friendsofsymfony/http-cache": "^2.10 || ^3.0",
-        "friendsofsymfony/http-cache-bundle": "^2.10.1 || ^3.0",
-        "friendsofsymfony/jsrouting-bundle": "^2.7 || ^3.0",
-        "friendsofsymfony/rest-bundle": "^3.1",
-        "gedmo/doctrine-extensions": "^3.0.4",
+        "friendsofsymfony/http-cache": "^2.13 || ^3.0",
+        "friendsofsymfony/http-cache-bundle": "^2.12 || ^3.0",
+        "friendsofsymfony/jsrouting-bundle": "^3.0",
+        "friendsofsymfony/rest-bundle": "^3.2",
+        "gedmo/doctrine-extensions": "^3.8",
```
alexander-schranz, line 58: gedmo has dependency to doctrine persistence in older versions not defined and so runs into:

> Error: Call to undefined method Doctrine\Bundle\DoctrineBundle\Mapping\ClassMetadataFactory::getCacheDriver()

We need increase it this to atleast 3.5 or 3.6 but our lowest supported persistence versions already update it then to 3.8 so define ^3.8 here.

composer.json
```diff
@@ -72,97 +71,98 @@
...
+        "symfony/translation-contracts": "^2.5 || ^3.0",
+        "symfony/twig-bundle": "^6.4 || ^7.1",
+        "symfony/validator": "^6.4 || ^7.1",
+        "symfony/yaml": "^6.4 || ^7.1",
+        "toflar/psr6-symfony-http-cache-store": "^4.0",
         "twig/twig": "^2.13 || ^3.0",
         "webmozart/assert": "^1.9"
     },
     "require-dev": {
+        "coduo/php-matcher": "^6.0.14",
```
alexander-schranz, line 123: accidently added this to require instead of dev dependency move now to dev #7653

PR #7676, opened 2024-11-25
Remove check for content bundle package instead check for Bundle

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? |  yes
| BC breaks? | no
| Deprecations? | no  
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Remove check for content bundle package instead check for Bundle

#### Why?

Content Bundle was moved to core.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,7 +30,6 @@
         "ext-json": "*",
         "ext-simplexml": "*",
         "composer-runtime-api": "^2.0",
-        "composer/package-versions-deprecated": "^1.8",
         "composer/semver": "^3.4",
         "contao/imagine-svg": "^1.0",
         "doctrine/annotations": "^1.14 || ^2.0",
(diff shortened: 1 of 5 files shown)


composer.json
```diff
@@ -30,7 +30,6 @@
         "ext-json": "*",
         "ext-simplexml": "*",
         "composer-runtime-api": "^2.0",
-        "composer/package-versions-deprecated": "^1.8",
```
alexander-schranz, line 33: this seems not be used and our required packages seems migrated to the official composer api.

alexander-schranz:
@Prokyonn FYI

composer.json
```diff
@@ -30,7 +30,6 @@
...
         "composer/semver": "^3.4",
```
Prokyonn, line 33: I think we could also get rid of this dependency then @alexander-schranz

PR #7680, opened 2024-11-25
Replace ramsey uuid package with symfony uid package and v4 to v7 Uuid

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no 
| Deprecations? | no  
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Replace ramsey uuid package with symfony uid package.

#### Why?

As we already use most symfony packages it better to build ontop of the symfony uid package instead of another third party package. This make maintainance on our side easier as we need not check anothr package for PHP compatibility or issues with the Symfony Sulu core.


src/Sulu/Bundle/SecurityBundle/EventListener/AuhenticationFailureListener.php
```diff
@@ -46,10 +46,10 @@ public function onLoginFailure(AuthenticationFailureEvent $event)
 
             if ($this->passwordHasherFactory instanceof PasswordHasherFactoryInterface) {
```
alexander-schranz, line 47: this bridge can also be removed we should target it in an own pull request

packages/article/src/Domain/Model/Article.php
```diff
@@ -35,7 +35,7 @@ class Article implements ArticleInterface
     public function __construct(
         ?string $uuid = null
     ) {
-        $this->uuid = $uuid ?: Uuid::uuid4()->toString();
+        $this->uuid = $uuid ?: Uuid::v7()->__toString();
```
alexander-schranz, line 38: `->toString()` was later introduced so we need use `__toString()` for Symfony 6.4 compatibility.

alexander-schranz, line 38: Think `v7` makes more sense for all our cases as they are time based sorted and a good replacement for v4

Prokyonn (approved):

alexander-schranz:
Looks like we need replace ramsey/uuid also in the php-task. https://github.com/php-task/php-task/issues/50

/cc @mamazu if you want to have a look at this.

Issue #5701, opened 2020-12-11
Add Flysystem V2 Support


dieter-pollier:
| Q | A
| --- | ---
| Sulu Version | 2.2
| Browser Version | Browser name and version

# Actual Behavior

When I configure S3 as external media storage, I get the following error on the cli:
`Unrecognized option "s3" under "sulu_media.storages". Available option is "local".`

# Expected Behavior

Sulu should recognize the configuration

# Steps to Reproduce

I followed the docs: https://docs.sulu.io/en/2.2/cookbook/external-media-storage.html
* Installed the packages flysystem and flysystem-aws-s3-v3
* Added the configuration yaml-snippet

# Possible Solutions

If you have already ideas how to solve the issue, add them here.
(remove this section if not needed)

nnatter:
Hey, thanks for reporting. The `s3` option is only available if the `League\Flysystem\AwsS3v3\AwsS3Adapter` is found:
https://github.com/sulu/sulu/blob/0a30e7978638c8c0192768a468f324f34c0ed0ea/src/Sulu/Bundle/MediaBundle/DependencyInjection/Configuration.php#L210-L229

Does this class exist in your project?

dieter-pollier:
Thanks for your reply. It seems that the name of the class has been changed in version 2.x: https://github.com/thephpleague/flysystem-aws-s3-v3/tree/2.x.
I installed using the command: `composer require league/flysystem:^1.0.0 league/flysystem-aws-s3-v3:^1.0.0`, and now the config gets loaded.

nnatter:
Thanks for the information! I will create a PR for the documentation 🙂

alexander-schranz:
I will keep this issue open as I think we should find a way in future to support the new [flysystem version](https://flysystem.thephpleague.com/v2/docs/advanced/upgrade-to-2.0.0/). We currently only support v1. We maybe need at current state add a `conflict` for version 2 package.

nnatter:
Reminder: Revert https://github.com/sulu/sulu-docs/pull/604 when this is solved 🙂

ju1ius:
Hi, any news on this (or possible workaround)?
`league/flysystem` and `league/flysystem-aws-s3-v3` are now at version `^3.0`... 🤔

RafaelKr:
[...]

PR #7687, opened 2024-11-26
Upgrade flysystem to version 3 and switch to Flysystem Bundle to support more adapters 

mamazu:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? | yes
| BC breaks? | no 
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | #7694 #5701
| License | MIT
| Documentation PR | https://github.com/sulu/sulu-docs/pull/827

#### What's in this PR?

Upgrading flysystem to version 3.

#### Why?

Latest version of Flysystem and support more adapters by directly using the official Flysystem Bundle.


src/Sulu/Bundle/MediaBundle/Media/Storage/StorageInterface.php
```diff
@@ -14,6 +14,12 @@
 /**
  * Defines the operations of the StorageLayer.
  * The StorageLayer is a interface to centralized management of media store.
+ *
+ *  @phpstan-type StorageOptions array{
+ * "directory" ?: string,
+ * "segments" ?: int|string,
+ * "fileName" ?: string,
```
alexander-schranz: its `segement`:

```suggestion
 *     "directory" ?: string,
 *     "segment" ?: int|string,
 *     "fileName" ?: string,
```

src/Sulu/Bundle/MediaBundle/DependencyInjection/SuluMediaExtension.php
```diff
@@ -407,32 +410,34 @@ public function load(array $configs, ContainerBuilder $container)
...
+                        $flySystemAdapter,
+                        $config['storage']['segments'],
+                    ]
+                )
+            )
+            ->setPublic(true)
+        ;
 
-        $container->setAlias('sulu_media.storage', 'sulu_media.storage.' . $storage)->setPublic(true);
+        $container->setAlias('sulu_media.storage.default', 'sulu_media.storage')->setPublic(true);
```
alexander-schranz: There is not need for `sulu_media.storage.default` we can directly register it as `sulu_media.storage` and only has a alias to the `StorageInterface`

src/Sulu/Bundle/MediaBundle/DependencyInjection/SuluMediaExtension.php
```diff
@@ -407,32 +410,34 @@ public function load(array $configs, ContainerBuilder $container)
...
-                } else {
-                    // Resolve unused ENV Variables of other Adapter
-                    $container->resolveEnvPlaceholders($value, true);
-                }
-            }
-        }
-
-        $loader->load('services_storage_' . $storage . '.xml');
+        $service = $config['storage']['service'];
+        $flySystemService = $container->get($service, ContainerInterface::NULL_ON_INVALID_REFERENCE);
```
alexander-schranz: its required without flysystem configuration it should not generate the container

src/Sulu/Bundle/MediaBundle/DependencyInjection/SuluMediaExtension.php
```diff
@@ -407,32 +410,34 @@ public function load(array $configs, ContainerBuilder $container)
...
+        $flySystemAdapter = $container->get('flysystem.adapter.'.$service, ContainerInterface::NULL_ON_INVALID_REFERENCE);
+
+        $container
+            ->setDefinition(
+                'sulu_media.storage',
+                new Definition(
+                    FlysystemStorage::class,
+                    [
+                        $flySystemService,
+                        $flySystemAdapter,
```
alexander-schranz: Use just new Reference not container get. See https://github.com/sulu/sulu/blob/3.0/packages/article/src/Infrastructure/Symfony/HttpKernel/SuluArticleBundle.php#L112-L118 as example.

src/Sulu/Bundle/MediaBundle/DependencyInjection/SuluMediaExtension.php
mamazu: What do you mean, it should fail if the service isn't there?

src/Sulu/Bundle/MediaBundle/DependencyInjection/SuluMediaExtension.php
mamazu: Refactored.

src/Sulu/Bundle/MediaBundle/DependencyInjection/SuluMediaExtension.php
mamazu: Removed but should be added to the list of bc breaks.

config/packages/flysystem.yaml
```diff
@@ -0,0 +1,6 @@
+flysystem:
+    storages:
+        default.storage:
+            adapter: 'local'
+            options:
+                directory: '%kernel.project_dir%/var/uploads'
```
mamazu, line 6: The tests are failing now because this is a userland configuration. What should I do with that? Move it to a "prependExtension" of the media bundle or add this config to all bundles that have a dependency on the media bundle?

alexander-schranz: @mamazu register the flysystem bundle in the TestKernel and configure it in the test config also:

https://github.com/sulu/sulu/blob/d7a707680a263ec402e6ce71cb35fb554f45f215/src/Sulu/Bundle/TestBundle/Kernel/SuluTestKernel.php#L78

https://github.com/sulu/sulu/blob/2.6/src/Sulu/Bundle/TestBundle/Resources/app/config/sulu.yml

UPGRADE.md
```diff
@@ -24,6 +24,30 @@ Removed unused arguments:
...
+# config/packages/sulu_media.yaml
+sulu_media:
+    storage:
+        service: 'default.storage'
+        segment: 10
+```
+This will only create the service `sulu_media.storage` as the alias to `sulu_media.storage.default` has been removed.
+
+Other removed services:
+- `Sulu\\Bundle\\MediaBundle\\DependencyInjection\\S3ClientCompilerPass` (internal)
```
````suggestion
### Updating to flysystem 3

The Sulu media storage uses now the Flysystem Bundle which need to be configured:

```yaml
# config/packages/flysystem.yaml
flysystem:
    storages:
        default.storage:
            adapter: 'local'
            options:
                directory: '%kernel.project_dir%/var/uploads'
```

If you want use a different storage for Sulu you can configure it in:

```yaml
# config/packages/sulu_media.yaml
sulu_media:
    storage:
        service: 'default.storage' # this is default and not required to be configured.
```

This will only create the service `sulu_media.storage` as the alias to `sulu_media.storage.*` services has been removed.
````

Add `S3ClientCompilerPass`, `LocalStorage`, `S3Storage`, ... just to the removed classes, services list which already exists.

**Comment by mamazu** [comment]
```
1) Sulu\Bundle\LocationBundle\Tests\Unit\Geolocator\Service\NominatimGeolocatorTest::testGuzzleLocate with data set #0 ('10, Rue Alexandre Dumas, Paris', 2, ['10, Rue Alexandre Dumas, Ste-... Union', 'Rue Alexandre Dumas', '10', 'F-75011', 'Paris', 'FR', '2.3898894', '48.8529486'])
Error: Class "GuzzleHttp\Handler\MockHandler" not found
/home/runner/work/sulu/sulu/src/Sulu/Bundle/LocationBundle/Tests/Unit/Geolocator/Service/NominatimGeolocatorTest.php:54
```

This test failure seems unrelated.

File: src/Sulu/Bundle/MediaBundle/Media/Storage/FlysystemStorage.php [hunk 7]
```diff
@@ -133,22 +145,43 @@ protected function getFilePath(array $storageOptions): string
...
-            $this->filesystem->createDir($segmentPath);
+            $this->filesystem->createDirectory($segmentPath);
+        }
+    }
+
+    public function getPath(array $storageOptions): string
+    {
+        if ($this->adapter instanceof LocalFilesystemAdapter) {
+            $reflection = new \ReflectionClass($this->adapter);
+            $rootPath = $reflection->getProperty('rootPath')->getValue();
```
mamazu: Hack should be removed before going live.

alexander-schranz: I only see `rootLocation` param what param are we fetching here?

https://github.com/thephpleague/flysystem/blob/462d92253fc0bbae0acac6fbf1f64cab27eae5f5/src/Local/LocalFilesystemAdapter.php#L65

alexander-schranz: best would a compilerpass read the param from `LocalFilesystemAdapter` construct and set it here also as constructor parameter.

you get the defintion of the defined flysystem adapter get the argument and replace the argument of this service with the same one:

https://github.com/sulu/sulu/blob/d7a707680a263ec402e6ce71cb35fb554f45f215/src/Sulu/Bundle/MediaBundle/DependencyInjection/S3ClientCompilerPass.php#L32

Some examples: https://github.com/sulu/sulu/blob/d7a707680a263ec402e6ce71cb35fb554f45f215/src/Sulu/Bundle/AdminBundle/DependencyInjection/Compiler/ExposeResourceRoutesPass.php#L50

```php
$flysystemAdapterDefintion = $container->getDefinition($container->getParameter('sulu_media.flysystem_service'));

$flysystemAdapterDefintion->getClass(); // check if local

$storageDefintion = $container->getDefinition('sulu_media.storage');

$storageDefintion->replaceArgument(..., $flysystemAdapterDefintion->getAgument(...));
```

UPGRADE.md
```diff
@@ -24,6 +29,31 @@ Removed unused arguments:
...
+                directory: '%kernel.project_dir%/var/uploads'
+```
+
+If you want use a different storage for Sulu you can configure it in:
+
+```yaml
+# config/packages/sulu_media.yaml
+sulu_media:
+    storage:
+        service: 'default.storage' # this is default and not required to be configured.
```
```suggestion
        flysystem_service: 'default.storage' # this is default and not required to be configured.
```

would make this more clear.

src/Sulu/Bundle/MediaBundle/Tests/Unit/Media/Storage/FlysystemStorageTest.php
```diff
@@ -0,0 +1,239 @@
...
+    protected function setUp(): void
+    {
+        $this->flysystem = $this->prophesize(FilesystemOperator::class);
+        $this->flysystemAdapter = $this->prophesize(FilesystemAdapter::class);
+
+        $this->flysystemStorage = new FlysystemStorage(
+            $this->flysystem->reveal(),
+            $this->flysystemAdapter->reveal(),
+            100
+        );
```
alexander-schranz: Would not mock here anything just use the in memory adapter: https://github.com/thephpleague/flysystem-memory

mamazu:
The last error in the pipeline is we don't install `guzzlehttp` anymore. It's just a transitive dependency:

```
~/packages/sulu/sulu  ‹3.0› $ composer why guzzlehttp/guzzle
aws/aws-sdk-php                3.331.0 requires guzzlehttp/guzzle (^6.5.8 || ^7.4.5) 
google/auth                    v1.43.0 requires guzzlehttp/guzzle (^7.4.5)           
google/cloud-core              v1.60.0 requires guzzlehttp/guzzle (^6.5.8|^7.4.4)    
microsoft/azure-storage-common 1.5.2   requires guzzlehttp/guzzle (~6.0|^7.0)
```

What's the plan there? It's only there for BC layer. Should we explicitly require it or remove BC layer?

src/Sulu/Bundle/MediaBundle/Tests/Unit/Media/Storage/FlysystemStorageTest.php
mamazu: Done

alexander-schranz:
@mamazu remove the bc layer here: https://github.com/sulu/sulu/blob/15a676466de29d815d016b37dd7cb97fef1cc76d/src/Sulu/Bundle/LocationBundle/Geolocator/Service/NominatimGeolocator.php#L35C32-L44.

But we should do it in an own PR. And adopt the tests to user other client.

src/Sulu/Bundle/MediaBundle/DependencyInjection/SuluMediaExtension.php
```diff
@@ -407,32 +409,27 @@ public function load(array $configs, ContainerBuilder $container)
...
-        $loader->load('services_storage_' . $storage . '.xml');
+        $sorageServiceId = $config['storage']['flysystem_service'];
+        $adapterService = 'flysystem.adapter.' . $sorageServiceId;
+
+        $container->register('sulu_media.storage', FlysystemStorage::class)
+            ->setArguments([
+                new Reference($sorageServiceId),
+                new Reference($adapterService),
+                $config['storage']['segments'],
+                null,
```
alexander-schranz: add comment:
```suggestion
                null, // defined via compilerpass @see FlysystemCompilerPass
```

add also the `use ...\FlysystemCompilerPass` it can be clicked on the class name

src/Sulu/Bundle/MediaBundle/Media/Storage/StorageInterface.php
```diff
@@ -14,6 +14,12 @@
 /**
  * Defines the operations of the StorageLayer.
  * The StorageLayer is a interface to centralized management of media store.
+ *
+ *  @phpstan-type StorageOptions array{
+ *     "directory" ?: string,
+ *     "segment" ?: int|string,
+ *     "fileName" ?: string,
+ * }
```
alexander-schranz: I'm not sure to add this to the `interface` basically the storageinterface don't care what options to storage use and can be anything for custom storages. So would keep this here `array<string, string|null>`

mamazu: I think the best way would be to have an object that could be extended and not just a loose array, but for now your suggestion makes the most sense.

alexander-schranz: object would same problem as the typed array. So if we want not force how somebody store its media and what storageOptions they use we should as said keep the `array<string, string|null>`, this also avoids a bc break in the interface.

If I personally would rework the interface I would even more go with a simple `path` `string` instead of any object or array.

mamazu: Yeah I mean the Flysystem adapter also has a `PathNormalizer` which the user could customize if they wanted some fancy behavior.

mamazu: But it's reverted now.

src/Sulu/Bundle/MediaBundle/DependencyInjection/SuluMediaExtension.php
mamazu: Comment added, but we don't need a use statement because the class is in the same namespace.

UPGRADE.md
```diff
@@ -34,6 +39,31 @@ As part of the update of flysystem the support for the guzzle client package `gu
...
+
+```yaml
+# config/packages/sulu_media.yaml
+sulu_media:
+    storage:
+        flysystem_service: 'default.storage' # this is default and not required to be configured.
+```
+
+This will only create the service `sulu_media.storage` as the alias to `sulu_media.storage.*` services has been removed.
+
```
alexander-schranz, line 171: I would like to have maybe also example in the upgrade.md how to migrate the existing s3, google cloud storage and azure blob storage examples to Flysystem config:

### Sulu Media S3 to Flysystem S3

**Before*:

```
sulu_media:
    storage: s3
    storages:
        s3:
            key: 'your aws s3 key'
            secret: 'your aws s3 secret'
            bucket_name: 'your aws s3 bucket name'
            path_prefix: 'optional path prefix'
            region: 'eu-west-1'
```

**New**:

```
flysystem:
    storages:
        default.storage:
             # ...
```


### Sulu Google Cloud Config to Flysystem S3

**Before*:

```
sulu_media:
    storage: google_cloud
    storages:
        google_cloud:
            key_file_path: '/path/to/key.json'
            bucket_name: 'sulu-bucket'
            path_prefix: 'optional path prefix'
```

**New**:

```
flysystem:
    storages:
        default.storage:
             # ...
```



### Sulu Azure Config to Flysystem Config

**Before*:

```
sulu_media:
    storage: google_cloud
    storages:
        google_cloud:
            key_file_path: '/path/to/key.json'
            bucket_name: 'sulu-bucket'
            path_prefix: 'optional path prefix'
```

**New**:

```
flysystem:
    storages:
        default.storage:
             # ...
```


https://docs.sulu.io/en/2.6/cookbook/external-media-storage.html

UPGRADE.md
```diff
@@ -34,6 +39,31 @@ As part of the update of flysystem the support for the guzzle client package `gu
 
 The `GoogleGeolocator` and the `NominatimGeolocator` no longer support the Guzzle client and require a `Symfony\HttpClient` client instead.
 
+### Updating to flysystem 3
+
+The Sulu media storage uses now the Flysystem Bundle which need to be configured:
```
alexander-schranz, line 44: we should also mention if `superbalist/flysystem-google-storage` was used it need to be replaced by `league/flysystem-google-cloud-storage`.

And that the other flysystem adapters (s3, azure) and flysystem itself need update to Version 3 if used by the project.

UPGRADE.md
```diff
@@ -34,6 +39,123 @@ As part of the update of flysystem the support for the guzzle client package `gu
...
+        default.storage:
+            adapter: 'local'
+            options:
+                directory: '%kernel.project_dir%/var/uploads'
+```
+
+Here are some examples on how to migrate individual providers:
+
+<details>
+  <summary>Sulu Media S3 to Flysystem S3</summary>
```
alexander-schranz, line 59: Love this 👍

UPGRADE.md
```diff
@@ -34,6 +39,126 @@ As part of the update of flysystem the support for the guzzle client package `gu
...
+
+New:
+
+```yaml
+flysystem:
+    storages:
+        default.storage:
+            adapter: 'aws'
+            options:
+                client: 'aws_client_service' # The service ID of the Aws\S3\S3Client instance
```
alexander-schranz: Do they need to configure that in some kind of a way / create a service definition there self or does the `aws_client_service` already exist out of the box? 

If they need to create it there should be also documented that step for the uprade:

```yaml
services:
    aws_client_service
        class: Aws\S3\S3Client
```

UPGRADE.md
```diff
@@ -34,6 +39,126 @@ As part of the update of flysystem the support for the guzzle client package `gu
...
+```
+
+New:
+```yaml
+flysystem:
+    storages:
+        default.storage:
+            adapter: 'gcloud'
+            options:
+                client: 'gcloud_client_service' # The service ID of the Google\Cloud\Storage\StorageClient instance
```
alexander-schranz: same as above how this service is created?

UPGRADE.md
```diff
@@ -34,6 +39,126 @@ As part of the update of flysystem the support for the guzzle client package `gu
...
+sulu_media:
+    storage: google_cloud
+    storages:
+        google_cloud:
+            key_file_path: '/path/to/key.json'
+            bucket_name: 'sulu-bucket'
+            path_prefix: 'optional path prefix'
+```
+
+New:
```
alexander-schranz, line 117: empty lines before code blocks avoid breaking some specific markdown readers:

```suggestion

New:
```

UPGRADE.md
```diff
@@ -34,6 +39,126 @@ As part of the update of flysystem the support for the guzzle client package `gu
...
+    storage: google_cloud
+    storages:
+        google_cloud:
+            key_file_path: '/path/to/key.json'
+            bucket_name: 'sulu-bucket'
+            path_prefix: 'optional path prefix'
+```
+
+New
+```
```
````suggestion
New

```yaml
````

File: UPGRADE.md [hunk 19]
```diff
@@ -34,6 +39,126 @@ As part of the update of flysystem the support for the guzzle client package `gu
...
+flysystem:
+    storages:
+        default.storage:
+            adapter: 'azure'
+            options:
+                client: 'azure_client_service' # The service ID of the MicrosoftAzure\Storage\Blob\BlobRestProxy instance
+                container: 'container_name'
+                prefix: 'optional/path/prefix'
+```
+</details>
```
````suggestion
```

</details>
````

File: UPGRADE.md [hunk 20]
```diff
@@ -34,6 +39,126 @@ As part of the update of flysystem the support for the guzzle client package `gu
...
+    storages:
+        default.storage:
+            adapter: 'gcloud'
+            options:
+                client: 'gcloud_client_service' # The service ID of the Google\Cloud\Storage\StorageClient instance
+                bucket: 'bucket_name'
+                prefix: 'optional/path/prefix'
+```
+</details>
+<details>
```
````suggestion
```

</details>

<details>
````

alexander-schranz (changes requested):

UPGRADE.md
mamazu: Well that's a good question. 
Well first of all, if you don't install it explicitly it won't be there. (So by default the s3 bundle will no longer be installed.)

But it looks like the default service is not configured. However, I don't know how it was configured before so it might be that it is set up if you used to use the S3 integration. Same goes for the other integrations.

mamazu: Okay I've dug a little S3 is the only storage which Sulu used to set up with a compiler pass (now deleted). I've added the new service definition to the upgrade guide. As for the other storage solutions, the leauge bundles do not come with automatic configs so you have to set those services up yourself.

UPGRADE.md
mamazu: Yes, if you're already using this. You should already have this service set up. You just need to reference it here.

alexander-schranz (approved):

alexander-schranz:
@mamazu Thank you. This is such a great step forward for get ride of old dependencies, great work 👍

PR #7691, opened 2024-11-28
Using constructor property promotion in Content Bundle

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | #7548 
| License | MIT
| Documentation PR | -

#### What's in this PR?
Running constructor property promotion on the article bundle.

#### Why?
If it's part of the Sulu main bundle now, it has to play by the rules.

### DIFF
--- a/packages/content/src/Application/ContentWorkflow/Subscriber/UnpublishTransitionSubscriber.php
+++ b/packages/content/src/Application/ContentWorkflow/Subscriber/UnpublishTransitionSubscriber.php
@@ -31,22 +31,10 @@ use Symfony\Component\Workflow\Event\TransitionEvent;
  */
 class UnpublishTransitionSubscriber implements EventSubscriberInterface
 {
-    /**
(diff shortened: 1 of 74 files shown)


packages/content/src/Application/ContentWorkflow/Subscriber/UnpublishTransitionSubscriber.php
```diff
@@ -31,21 +31,15 @@
...
 {
-    /**
-     * @var DimensionContentRepositoryInterface
-     */
-    private $dimensionContentRepository;
-
     /**
      * @var EntityManagerInterface
      */
     protected $entityManager;
```
mamazu: This property was protected on an internal class that should not be extended. Now it's private.

mamazu:
Closing this as this has too many conflicts. This might make sense after the refactoring is done.

PR #7693, opened 2024-11-28
Replace piwik with matomo analytic templates

fnagel:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Related issues/PRs | https://github.com/sulu/sulu/discussions/7692
| License | MIT

#### What's in this PR?

Updated the Matomo templates.

#### Why?

Because the current ones are outdated.

### DIFF
--- a/src/Sulu/Bundle/WebsiteBundle/Resources/views/Analytics/matomo/body-open.html.twig
+++ b/src/Sulu/Bundle/WebsiteBundle/Resources/views/Analytics/matomo/body-open.html.twig
@@ -1,1 +1,1 @@
-<noscript><p><img src="{{ analytics.content.url }}/piwik.php?idsite={{ analytics.content.siteId }}" style="border:0;" alt="" /></p></noscript>
+<noscript><p><img referrerpolicy="no-referrer-when-downgrade" src="{{ analytics.content.url }}/matomo.php?idsite={{ analytics.content.siteId }}&amp;rec=1" style="border:0;" alt="" /></p></noscript>
--- a/src/Sulu/Bundle/WebsiteBundle/Resources/views/Analytics/matomo/head-close.html.twig
+++ b/src/Sulu/Bundle/WebsiteBundle/Resources/views/Analytics/matomo/head-close.html.twig
@@ -1,12 +1,12 @@
 <script>
-    var _paq = _paq || [];
+    var _paq = window._paq = window._paq || [];
     _paq.push(['trackPageView']);
     _paq.push(['enableLinkTracking']);
     (function() {
         var u="{{ analytics.content.url|trim('/') }}";
-        _paq.push(['setTrackerUrl', u+'/piwik.php']);
+        _paq.push(['setTrackerUrl', u+'matomo.php']);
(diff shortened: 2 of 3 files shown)


alexander-schranz:
@fnagel thx for the pull request can you add a short line to the `UPGRADE.md`

Something similar to:

```
### Piwik replaced with Matomo script

The script provided by Sulu for the piwik implementation has been updated to use mataomo path so the script is now pointing to matomo.php instead of the piwik.php file.
```

fnagel:
@alexander-schranz Changed the `UPGRADE.md`.

alexander-schranz (approved):

alexander-schranz:
@fnagel awesome. Thank you!

src/Sulu/Bundle/WebsiteBundle/Resources/views/Analytics/matomo/head-close.html.twig
```diff
@@ -1,12 +1,12 @@
 <script>
-    var _paq = _paq || [];
+    var _paq = window._paq = window._paq || [];
     _paq.push(['trackPageView']);
     _paq.push(['enableLinkTracking']);
     (function() {
         var u="{{ analytics.content.url|trim('/') }}";
-        _paq.push(['setTrackerUrl', u+'/piwik.php']);
+        _paq.push(['setTrackerUrl', u+'matomo.php']);
```
benr77, line 7: The forward slash has been removed here, which is causing invalid URLs to be constructed.

src/Sulu/Bundle/WebsiteBundle/Resources/views/Analytics/matomo/body-open.html.twig
```diff
@@ -1 +1 @@
-<noscript><p><img src="{{ analytics.content.url }}/piwik.php?idsite={{ analytics.content.siteId }}" style="border:0;" alt="" /></p></noscript>
+<noscript><p><img referrerpolicy="no-referrer-when-downgrade" src="{{ analytics.content.url }}/matomo.php?idsite={{ analytics.content.siteId }}&amp;rec=1" style="border:0;" alt="" /></p></noscript>
```
benr77, line 1: This also needs the `|trim('/')` filter to ensure there is not a duplicate forward slash.

src/Sulu/Bundle/WebsiteBundle/Resources/views/Analytics/matomo/head-close.html.twig
fnagel: You are right. I copied the latest code from the Matomo docs without checking what the u variable will contain. Thanks!

src/Sulu/Bundle/WebsiteBundle/Resources/views/Analytics/matomo/body-open.html.twig
fnagel: You're right. This seems to be an issue even before my changes.

PR #7694, opened 2024-11-30
Remove guzzle dependency from `LocationBundle`

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | yes
| Deprecations? | no (Removing deprecated code)
| Fixed tickets | -
| Related issues/PRs | #7687 
| License | MIT
| Documentation PR | -

#### What's in this PR?
Removing the guzzle dependency from the project.

#### Why?
Currently we still have a dependency on the guzzle http implementation for BC reasons. This merge request removes the option to use the guzzle http client in the `LocationBundle`.

The reason is that the guzzle http client isn't even a depnency of Sulu and only installed because of the old flysystem dependency. See https://github.com/sulu/sulu/pull/7687#issuecomment-2508161281


composer.json
```diff
@@ -50,7 +50,6 @@
         "friendsofsymfony/jsrouting-bundle": "^3.0",
         "friendsofsymfony/rest-bundle": "^3.2",
         "gedmo/doctrine-extensions": "^3.8",
-        "guzzlehttp/promises": "^1.0 || ^2.0",
```
alexander-schranz: the `promises` package need to stay will not be removed they are required that the http cache working performant with the symfony http client:

 - https://foshttpcachebundle.readthedocs.io/en/stable/overview.html#installation
 - https://symfony.com/doc/current/http_client.html#httplug

mamazu: Interesting. Very weird that this is not a dependency of either of those packages. Because we don't use any of the classes in the Sulu source code.

Readded.

alexander-schranz: @mamazu you can use any php http implementation with the FosHttpCache component.

https://github.com/FriendsOfSymfony/FOSHttpCache/blob/6151c80350a1c54e8723f927a50ee2f90fcd19f9/composer.json#L29

As Symfony based we choose the Symfony Http Client it provides the async: 

https://github.com/symfony/http-client/blob/955e43336aff03df1e8a8e17daefabb0127a313b/composer.json#L19

But its async implementation depends on guzzle promises ([documented here](https://symfony.com/doc/current/http_client.html#httplug)) which is not a direct dependency as you can use it without async still.

We may can discuss with @dbu if the FOSHttpCacheBundle as it is a symfony bundle may could require it as I think its uncommon that somebody using another client inside a symfony project?

mamazu: Yeah I think that would make sense. But that's probably a topic for another pull request.

UPGRADE.md
```diff
@@ -28,6 +28,10 @@ Removed unused arguments:
 
 The script provided by Sulu for the piwik implementation has been updated to use mataomo path so the script is now pointing to matomo.php instead of the piwik.php file.
 
+### Removing deprecated guzzle integration
+
+As part of the update of flysystem the package `guzzlehttp/guzzle` and `guzzlehttp/promise` have been removed. If you need it you need to manually require it.
```
alexander-schranz: The `GoogleGeolocator` and `NominatimGeolocator` not longer supporting the Guzzle Http Client instead a Symfony HttpClient instance need to give to its constructors now.

alexander-schranz (approved):

composer.json
dbu: FOSHttpCache [uses](https://github.com/FriendsOfSymfony/FOSHttpCache/blob/6151c80350a1c54e8723f927a50ee2f90fcd19f9/composer.json#L28) [php-http/discovery](https://docs.php-http.org/en/latest/discovery.html) which contains a composer plugin that requires a [suiteable http client](https://docs.php-http.org/en/latest/discovery.html#auto-installation) if none is already available in your project.

PR #7698, opened 2024-12-11
Regenerating phpstan baseline

mamazu:
Fixing phpstan for Sulu 2.6 by just updating the baseline.


alexander-schranz:
regenerating should always be done on lowest branch (currently 2.5).

PS: currently there is an upstream issue you might need https://github.com/sulu/sulu/pull/7699 to regenerate to baseline.

mamazu:
Oh yeah, well then I'll wait for the other thing to be fixed.

PR #7701, opened 2024-12-12
Splitting `Upgrade.md`

mamazu:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? | no 
| BC breaks? | no 
| Deprecations? | no  
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?
Splitting the upgrade guide.

#### Why?
Both the IDE and the Github diff view choke on big files.


UPGRADE-1.0.md
```diff
@@ -0,0 +1,693 @@
+## 1.0.0
```
alexander-schranz: I would create a `UPGRADE-1.md` including all from 1.6.24 -> 0 (including 1.6.24)

I would create a `UPGRADE-2.md` including all from 2.6.4 - 2.0.0-alpha1 (including the 2.0.0-alpha1 and 2.6.4) 

`UPGRADE.md` start with `3.0.0`

new -> old

mamazu: Yup that makes sense. Did that.

mamazu:
@alexander-schranz Rebased this onto 3.0 again. It would be great if this change could be merged soon because any changes to the `upgrade.md` file will cause conflicts.

alexander-schranz:
@mamazu to not lose the git history I did this in seperate commits so on the 2.6 branch the git blame works on UPGRADE-2.x.md and on 3.0 git blame on UPGRADE-3.x still works.

 - https://github.com/sulu/sulu/blame/2.6/UPGRADE-2.x.md
 - https://github.com/sulu/sulu/blame/3.0/UPGRADE-3.x.md

mamazu:
That's perfect.

Issue #7713, opened 2025-01-06
Preview not working with handler_id configured to session.handler.native_file


tstauer:
| Q | A
| --- | ---
| Sulu Version | 2.6.6
| PHP Version | 8.2.27
| DB Version | mysql  Ver 8.0.40 for Linux on aarch64 (MySQL Community Server - GPL)
| Browser Version | Firefox ESR 128.5.2esr (64-Bit)

# Actual Behavior
The preview does not work in the environment dev, instead a message "Warning: ini_set(): Session ini settings cannot be changed when a session is active" is shown.

# Expected Behavior

I would expect the preview to render.

# Steps to Reproduce

Configure the session like this:
config/packages/framework.yaml
```
framework:
    session:
        storage_factory_id: session.storage.factory.native
        handler_id: 'session.handler.native_file'
        save_path: '%env(STORAGE_PATH)%/sessions/'
        cookie_secure: auto
        cookie_samesite: lax
        name: SULUSESSID # This avoids conflicts with other applications running on the same domain
```

Then go to admin and try to preview a page:
<img width="1044" alt="Screenshot 2025-01-06 at 14 08 18" src="https://github.com/user-attachments/assets/09ca8bf7-4c4c-4f32-9406-b0582d3188c2" />

# Workaround
Set `handler_id` to `null` in the config.

alexander-schranz:
@tstauer can you create a small reproducer repository based on current 2.6 version of Sulu?

tstauer:
@alexander-schranz Thank you for your patience and great support.

I created a reproducer based on the sulu/skeleton branch "2.6".
https://github.com/tstauer/reproducer-7713-sulu

Made some changes on the docker configuration because I neither have a web server nor php installed locally.
Hope that helps.

ampaze:
I have the same issue. It is enough to set `save_path:` to trigger this. 
@alexander-schranz Please have another look.

Issue #7086, opened 2023-06-01
Symfony 6.3 Test Environment incompatibility `ini_set`  issue


alexander-schranz:
| Q | A
| --- | ---
| Sulu Version | 2.5 / 2.6
| Symfony Version | 6.3
| Related Issue | https://github.com/symfony/symfony/issues/49387
| Unrelated Issue | ~~other 6.3 Issue: https://github.com/sulu/SuluArticleBundle/issues/641~~
| PHP Version | 
| DB Version | 
| Browser Version | 

# Actual Behavior

The current test environment is not compatible with Symfony 6.3 and ends up in this error:

<img width="1251" alt="Bildschirmfoto 2023-06-01 um 11 53 45" src="https://github.com/sulu/sulu/assets/1698337/b6ab2276-162b-4376-b972-70559dea8c3e">

```
PHP Warning:  ini_set(): Session ini settings cannot be changed after headers have already been sent in /home/runner/work/SuluContentBundle/SuluContentBundle/vendor/symfony/http-foundation/Session/Storage/Handler/NativeFileSessionHandler.php on line 49
PHP Warning:  ini_set(): Session ini settings cannot be changed after headers have already been sent in /home/runner/work/SuluContentBundle/SuluContentBundle/vendor/symfony/http-foundation/Session/Storage/Handler/NativeFileSessionHandler.php on line 49
EEEE..........E..............................ESSSSSSESFSSSSSS.. 504 / 510 ( 98%)
......                                                          510 / 510 (100%)

Time: 00:08.519, Memory: 255.00 MB

There were 7 errors:

1) Sulu\Bundle\ContentBundle\Tests\Functional\Content\Infrastructure\Doctrine\DimensionContentQueryEnhancerTest::testGroupContentAdmin
ini_set(): Session ini settings cannot be changed after headers have already been sent
```

# Expected Behavior

It should not error.

# Steps to Reproduce

Run sulu/sulu tests on 6.3.

# Workaround

Stay on Symfony 6.2 currently with adding following to `composer.json` file:

```json
    "extra": {
        "symfony": {
            "require": "6.2.*"
        }
    }
```

# Background

The issue happens because of `gedmo` extensions. Calling: https://github.com/doctrine-extensions/DoctrineExtensions/blob/55e949979ae61f62378759bf50e0cea42767fba7/src/Tree/Hydrator/ORM/TreeObjectHydrator.php#L258 instance a service which tries todo something unexpectly.

PR #7714, opened 2025-01-06
Reset handler_id in preview, too

tstauer:
This fixes the message "Warning: ini_set(): Session ini settings cannot be changed when a session is active" in preview. The warning can come up when handler_id is set to 'session.handler.native_file' in framework.yaml for the session.

| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | #7713
| Related issues/PRs | #7086, #7101
| License | MIT
| Documentation PR | 

#### What's in this PR?

Resets the handler_id for the preview.

#### Why?

See #7713

### DIFF
--- a/src/Sulu/Bundle/PreviewBundle/Resources/config/config_preview.yml
+++ b/src/Sulu/Bundle/PreviewBundle/Resources/config/config_preview.yml
@@ -1,6 +1,7 @@
 framework:
     session:
         storage_factory_id: session.storage.factory.mock_file
+        handler_id: ~
     profiler:
         enabled: false
 


alexander-schranz:
Not sure yet about this as preview should still use what is defined in the application to match that state, else I think we run into other issues.

Based on the symfony documentation if you define a save_path you need define the handler_id to still use the native implementation:

```yaml
framework:
    session:
        storage_factory_id: session.storage.factory.native
        handler_id: null
        save_path: '%env(STORAGE_PATH)%/sessions/'
```

Does that work for you?

tstauer:
Hi Alex, thank you so much for your help.

Ah, that's part of the Symfony 7.2 documentation.
https://symfony.com/doc/current/reference/configuration/framework.html#handler-id
> If the framework.session.save_path option is set, then Symfony stores sessions using the native file session handler.

I'm checked for Symfony 6.4, where the configuration documentation doesn't mention that.
https://symfony.com/doc/6.4/reference/configuration/framework.html#handler-id

Found it in the change log of framework-bundle then.
https://github.com/symfony/framework-bundle/blob/v6.4.0/CHANGELOG.md
> Deprecate not setting either framework.session.handler_id or save_path config options; handler_id will default to null in 7.0 if save_path is not set and to session.handler.native_file otherwise
That is why I set `handler_id: 'session.handler.native_file' ` in the first place.

```
framework:
    session:
        storage_factory_id: session.storage.factory.native
        handler_id: null
        save_path: '%env(STORAGE_PATH)%/sessions/'
```
Works for me. However, if I don't set handler_id I get the ini_set message instead of the page.
I also just discovered my non-direct dependencies did upgrade to Symfony 7.2, so I have framework-bundle on 6.4* and http-foundation on 7.2.*.

Symfony also supports sessions in a database
https://symfony.com/doc/6.4/session.html#session-database-pdo 
I tried that and this also works:
```
framework:
    session:
        handler_id: Symfony\Component\HttpFoundation\Session\Storage\Handler\PdoSessionHandler
```

To sum that up I don't know what causes the issue in the first place. My impression is that https://github.com/symfony/symfony/issues/51600 is somehow related. That's why I created a pull request for the preview config.

alexander-schranz:
> I also just discovered my non-direct dependencies did upgrade to Symfony 7.2, so I have framework-bundle on 6.4* and http-foundation on 7.2.*.

I recommend you that you are using `symfony/flex` which will force a stable symfony compatibility via:

https://github.com/sulu/skeleton/blob/60d628b4c81dc7a71e93b490905e19fe1b36adbd/composer.json#L203-L208

You need allow that plugin:

https://github.com/sulu/skeleton/blob/60d628b4c81dc7a71e93b490905e19fe1b36adbd/composer.json#L195

It is enabled by default for any Symfony and so Sulu project.

If you are still on Symfony 6.4 you need `handler_id` defined

tstauer:
I'm now on Symfony 7 and back to a broken preview, even when I set handler_id to null, to ~ (tilde) or don't define handler_id at all. The preview only shows "Warning: ini_set(): Session ini settings cannot be changed when a session is active" again. :(

As soon as I set save_path the preview is broken.

Setting this does not help:
```
framework:
    session:
        storage_factory_id: session.storage.factory.native
        handler_id: null
        save_path: '%env(STORAGE_PATH)%/sessions/'
```

Setting the handler_id to null or ~ in config_preview.yaml does no longer fix the preview, either.

alexander-schranz:
@tstauer did you try the default settings? https://github.com/sulu/skeleton/blob/d616fe6b8f8766068a9ae5af84468aa6cd32f5e9/config/packages/framework.yaml#L9-L10

tstauer:
@alexander-schranz yes, the default settings work.
If I take the default settings and add `save_path` it's not working.

Unfortunately we could only use that when developing. The other environments are behind load balancers - we share cookies and sessions using a network mount to avoid requiring sticky sessions.

EDIT: I can set `save_path: ~` for dev and then I also get a preview. However, I suspect the warnings still occur on prod and are just logged.

alexander-schranz:
@tstauer would need to have a deeper look at this. You can still achieve what you need if you configure the `save_path` in your `php.ini` instead of symfony configs. Just define the `session.save_path` to your mounted directory.

PR #7720, opened 2025-01-08
Fix the lastmod sitemap property when lastModified is empty string

### LINKED ISSUE #7719: lastmod property not correct in sitemap
| Q | A
| --- | ---
| Sulu Version | 2.6.5
| PHP Version | 8.3

# Actual Behavior

The `lastmod` property of the Sulu page sitemap is always the current date.

# Expected Behavior

It should use the page's `lastModified` property, or the `changed` property as a fallback

# Steps to Reproduce

Render the sitemap

# Possible Solutions

It occurs at this line: https://github.com/sulu/sulu/blob/2.6/src/Sulu/Bundle/PageBundle/Sitemap/PagesSitemapProvider.php#L105

Apparently, `lastModified` is an empty string, not null or false, so it doesn't fallback to `changed`. Because of that it always creates a `DateTime` object based on `""`, so today.

I guess this should do the trick?

```
$changed = $contentPage['changed'];

if (!empty($contentPage['lastModified'])) {
    $changed = $contentPage['lastModified'];
}
```

eekes:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | fixes #7719
| Related issues/PRs | no
| License | MIT
| Documentation PR | no

#### What's in this PR?

Fixes the `lastmod` sitemap property

#### Why?

`lastmod` is always the current date if `lastModified` is empty. Because it's an empty string it doesn't fallback to `changed`

### DIFF
--- a/src/Sulu/Bundle/PageBundle/Sitemap/PagesSitemapProvider.php
+++ b/src/Sulu/Bundle/PageBundle/Sitemap/PagesSitemapProvider.php
@@ -102,7 +102,12 @@ class PagesSitemapProvider extends AbstractSitemapProvider
         string $host,
         string $scheme
     ) {
-        $changed = $contentPage['lastModified'] ?? $contentPage['changed'];
+        $changed = $contentPage['changed'];
+
+        if (!empty($contentPage['lastModified'])) {
+            $changed = $contentPage['lastModified'];
+        }
+
         if (\is_string($changed)) {
             $changed = new \DateTime($changed);
         }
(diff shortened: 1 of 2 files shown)


alexander-schranz:
@eekes thx, can you also add a test with lastmodified empty string? https://github.com/sulu/sulu/blob/e5cfa4d76104dabca227db25390b7b4aedd855e8/src/Sulu/Bundle/PageBundle/Tests/Unit/Sitemap/PagesSitemapProviderTest.php#L141

eekes:
@alexander-schranz Also done! ✅

alexander-schranz (approved):

alexander-schranz:
@eekes Thank you!

PR #7723, opened 2025-01-09
Fixing cache plugin

mamazu:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no 
| BC breaks? | no 
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?
Passing the required config to the fos-http-cache.

#### Why?
In Version 3.1.0 the friendsofsymfony/http-cache-bundle added a parameter to the `FOS\HttpCacheBundle\EventListener\CacheControlListener`. While this parameter has a default value, the bundle setting it up does not. 
See:
 https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/blob/7b737f770941bfb67e065d1be5975aa26a3ee2d7/src/DependencyInjection/FOSHttpCacheExtension.php#L51C1-L51C3

So we need to pass the value manually.

Maybe we could also update the bundle to 3.1.0 if they're doing BC breaks in minor versions now.

### DIFF
--- a/src/Sulu/Bundle/HttpCacheBundle/DependencyInjection/SuluHttpCacheExtension.php
+++ b/src/Sulu/Bundle/HttpCacheBundle/DependencyInjection/SuluHttpCacheExtension.php
@@ -34,6 +34,9 @@ class SuluHttpCacheExtension extends Extension implements PrependExtensionInterf
             'debug' => [
                 'enabled' => $config['debug']['enabled'],
             ],
+            'cache_control' => [
+                'ttl_header' => 'X-Reverse-Proxy-TTL',
+            ]
         ];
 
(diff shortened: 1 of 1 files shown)


alexander-schranz:
@mamazu this need to be fixed on the FOSHTTPCache side as it is a bc break there: https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/pull/639

alexander-schranz:
@mamazu FYI new version of FOSHttpCacheBundle (3.1.1) is released with a fix.

mamazu:
Should I make a pr updating to that version?

alexander-schranz:
@mamazu no not required older versions works fine and new version also.

Issue #7672, opened 2024-11-22
New Content Storage Roadmap (Sulu-3.0) 🚀 


alexander-schranz:
# When will Sulu 3.0 come

[Read more about on our Blog](https://sulu.io/blog/sulu-3-0-taking-a-bit-more-time-to-get-it-right)

# Problem description

Copy the not yet finished todo list out of https://github.com/sulu/sulu/pull/7611 and https://github.com/sulu/SuluContentBundle/pull/271 to keep tracking of it.

#### Selection Content Types

 - [x] ContentResolving https://github.com/sulu/sulu/pull/7611/commits/cff56651a99ec47ea552891c3c67fc392226107a
   - [x] Category
   - [x] Media 
 - [x] https://github.com/Prokyonn/sulu/pull/3
   - [x] Single Media
 - [x] https://github.com/Prokyonn/sulu/pull/4
   - [x] ImageMap
   - [x] Collection
   - [x] Single Collection
 - [x] https://github.com/Prokyonn/sulu/pull/1
   - [x] Account 
   - [x] Single Account
   - [x] Contact
   - [x] Single Contact
   - [x] Contact / Account
 - [x] https://github.com/Prokyonn/sulu/pull/2
    - [x] Tags
 - [x] Date https://github.com/sulu/sulu/pull/7729
 - [x] DateTime https://github.com/sulu/sulu/pull/7729
 - [x] Location (may not require own Resolver)
 - [x] SettingsResolver 
 - [x] ~~Snippet~~
 - [x] ~~Single Snippet~~
 - [x] ~~Page~~
 - [x] ~~Single Page~~

##### Inconsistencies

- [x] fix inconsistencies between ImageMapPropertyResolver and BlockProperty resolver regarding wrapping child objects in a ContenView or extracting it directly  #8092 
- [x] Create an additional interface for `MetadataAware` property resolvers, the `view` should only return scalar values, (e.g. metadata object must not be returned)    https://github.com/sulu/sulu/pull/8094
- [x] Blocks on `minOccurs="1" maxOccurs="1"` returns not a list: https://github.com/sulu/sulu/blob/418a50aaf9345e59c997b5ee2fdf4526bab72ec7/src/Sulu/Component/Content/Types/BlockContentType.php#L407-L409 Reported by @benr77 https://github.com/sulu/sulu/pull/7751  
[...]

Issue #7580, opened 2024-08-30
When moving a page the url of child pages is not reflected in `sulu_navigation_root_tree`


ampaze:
| Q | A
| --- | ---
| Sulu Version | 2.5.20
| PHP Version | 8.1

# Actual Behavior

When a page is moved to a different parent, the url of child pages is not correctly available in the frontend via `sulu_navigation_root_tree`. It is however correctly shown in the admin interface.

 For example:

![image](https://github.com/user-attachments/assets/62d3715c-5498-4160-8d0b-9c7fb366b80f)

![image](https://github.com/user-attachments/assets/273088f5-833d-4a11-9d66-7a6cde4882a8)

the parent url was correctly changed from `/wissen/tiere` to  `/wissen/portraits/tiere`, but the child page is still 
`/wissen/tiere/agutis` instead of `/wissen/portraits/tiere/agutis`.

I already tried to reindex all pages, but that does not change anything.

```bash
bin/console massive:search:reindex  --env=prod --provider=sulu_structure
bin/websiteconsole massive:search:reindex  --env=prod --provider=sulu_structure
```

I thought that maybe https://github.com/sulu/sulu/pull/7459 would have fixed this, but it does not. Seems like the navigation does not use the index.

# Expected Behavior

`sulu_navigation_root_tree` should always give back the correct current url of page.

# Steps to Reproduce

Move a page that has child pages to a different parent.

Issue #4990, opened 2020-01-20
[BUG] Generated URL of child page wrong if parent page name changes (rename, new translation, ...)


AndreasA:
| Q | A
| --- | ---
| Sulu Version | 2.0.1
| Browser Version | Chrome 79

# Actual Behavior

Creating the following page structure:
`Root page` --> `foo` --> `bar`

Now renaming (URL segment) `foo` to `my` results in the page `/foo/bar` being redirected to `/my/bar` which is correct.
However, all links generated for the page `bar` still generate `/foo/bar` which means the links on the page still point to the old URLs, even after cache clear.

Another issue with this (probably same reason) is that if a translation didn't exist and it was generated later for the original page `foo`, then the link for that localization is `/bar` and keeps being `/bar` as the parent page URL change was not registered.

Interestingly I think a new entry in the URL history is added when changing the parent page.

# Expected Behavior

For the scenario above the generated URLs should be `/my/bar` and not `/foo/bar`.
Meaning the current parent page should be taken into account (same regarding translations).

# Steps to Reproduce

See `Actual Behaviour`.

danrot:
Just checked, this is also happening on Sulu 1.6, therefore it should be fixed there.

alexander-schranz:
As written in https://github.com/sulu/sulu/issues/5302#issuecomment-631381253

The problems seems the urls are correctly updated on the default workspace and written there. But when the Parent Page get published. Only the parent node and url will be written to the live workspace. So until the childs are not republish they stay with the old url in the live workspace.

We can not automatically publish the child as there could be other changes there, but maybe we should update the urls when a parent page is publish. Problems I still see that the child pages in the index will still have the old url and there it would also need to be updated. 

If we go with we do not automatically update the child page urls in the live workspace we should at least look that the child pages are marked as draft so the content manager know that there is a change which can be published.

danrot:
> We can not automatically publish the child as there could be other changes there, but maybe we should update the urls when a parent page is publish. Problems I still see that the child pages in the index will still have the old url and there it would also need to be updated.

That sounds good to me, and I think that would also be the behavior the content manager expects.

PR #7726, opened 2025-01-10
Add new route implementation

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? |yes
| BC breaks? | yes
| Deprecations? | yes 
| Fixed tickets | fixes # 
| Related issues/PRs | see https://github.com/sulu/sulu/issues/7672
| License | MIT

#### What's in this PR?

Add new route implementation.

#### Why?

I did some experimenting to simplify the route handling in https://github.com/alexander-schranz/route-prototype. Instead of that the developer need manually care about updating things we will work with doctrine listeners which will update child urls automatically to avoid issues like:

 - https://github.com/sulu/sulu/issues/7580
 - https://github.com/sulu/sulu/issues/4990

Also we will not longer have a difference between unpublished or published URLs. Currently in Article Bundle I could create 2 articles with the url `/test` and `/test` and it did only error when try to publish one. This should not longer be the case as we already reserving and create a unique URL when saving the Entity and does not end in publishing errors specially when try using in combination with the AutomationBundle.

#### ToDo

 - [x] Create Entity
 - [x] Childs are update
 - [x] Grand childs are update
 - [x] Multi Localization support
 - [x] Multi Site support
 - [x] Support for non Site connected routes (article)
    - [x] Site A with `/test` and Site B `/test` and article parent Site A `/test/article` should not be updated when Site B `/test` was changed

#### ToDos in own Pull Requests

 - [x] History Handling https://github.com/sulu/sulu/pull/7830
 - [x] Add Multi Site tests https://github.com/sulu/sulu/pull/7831
 - [x] Add Multi Localization https://github.com/sulu/sulu/pull/7831
 - [x] Add CrossSite Site test (article (site: null) -> page (site: website)) https://github.com/sulu/sulu/pull/7834
 - [x] Add Test for url with same prefix are not effect on the same level https://github.com/sulu/sulu/pull/7836
 - [x] Tree Full Edit: https://github.com/sulu/sulu/pull/7835
 - [x] Route for AutoIncrement Entities: https://github.com/sulu/sulu/pull/7859
 - [x] Dynamic Router https://github.com/sulu/sulu/pull/7860
 - [ ] Locale Switcher https://github.com/sulu/sulu/pull/7874
 - [ ] Implement in Content Bundle https://github.com/sulu/sulu/pull/7749

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,7 +36,7 @@
         "doctrine/cache": "^1.12.1 || ^2.0",
         "doctrine/collections": "^1.6 || ^2.0",
         "doctrine/data-fixtures": "^1.4 || ^2.0",
-        "doctrine/dbal": "^3.3",
+        "doctrine/dbal": "^3.6",
         "doctrine/doctrine-bundle": "^2.6",
         "doctrine/doctrine-fixtures-bundle": "^3.4 || ^4.0",
         "doctrine/inflector": "^1.4.1 || ^2.0.1",
--- a/composer.json
+++ b/composer.json
@@ -211,6 +211,7 @@
             "Sulu\\Article\\": "packages/article/src",
             "Sulu\\Snippet\\": "packages/snippet/src",
             "Sulu\\Page\\": "packages/page/src",
+            "Sulu\\Route\\": "packages/route/src",
             "Sulu\\Content\\": "packages/content/src"
         },
         "exclude-from-classmap": [
@@ -225,6 +226,7 @@
             "Sulu\\Article\\Tests\\": "packages/article/tests",
             "Sulu\\Snippet\\Tests\\": "packages/snippet/tests",
             "Sulu\\Page\\Tests\\": "packages/page/tests",
+            "Sulu\\Route\\Tests\\": "packages/route/tests",
             "Sulu\\Content\\Tests\\": "packages/content/tests"
         }
     },
--- a/packages/route/config/doctrine/Route/Route.orm.xml
+++ b/packages/route/config/doctrine/Route/Route.orm.xml
@@ -1,0 +1,29 @@
+<?xml version="1.0" encoding="utf-8"?>
+<doctrine-mapping xmlns="http://doctrine-project.org/schemas/orm/doctrine-mapping"
+                  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                  xsi:schemaLocation="http://doctrine-project.org/schemas/orm/doctrine-mapping http://doctrine-project.org/schemas/orm/doctrine-mapping.xsd">
+    <entity
+        name="Sulu\Route\Domain\Model\Route"
+        table="ro_routes"
+    >
+        <unique-constraints>
+            <unique-constraint name="ro_routes_unique" fields="site, locale, slug" />
+            <unique-constraint name="ro_routes_resource_unique" fields="site, locale, resourceKey, resourceId" />
+        </unique-constraints>
+
+        <id name="id" type="integer" column="id">
+            <generator strategy="AUTO"/>
+        </id>
+
+        <field name="site" type="string" length="16" nullable="true"/>
+        <field name="locale" type="string" length="11"/>
+        <field name="slug" type="string"/>
+
+        <many-to-one field="parentRoute" target-entity="Sulu\Route\Domain\Model\Route">
+            <join-column name="parent_id" referenced-column-name="id" nullable="true" on-delete="SET NULL"/>
+        </many-to-one>
+
+        <field name="resourceKey" type="string" length="32"/>
+        <field name="resourceId" type="string" length="48"/>
+    </entity>
+</doctrine-mapping>
--- a/packages/route/config/doctrine/Route/Route.orm.xml
+++ b/packages/route/config/doctrine/Route/Route.orm.xml
@@ -17,7 +17,7 @@
 
         <field name="site" type="string" length="16" nullable="true"/>
         <field name="locale" type="string" length="11"/>
-        <field name="slug" type="string"/>
+        <field name="slug" type="string" length="164"/>
 
         <many-to-one field="parentRoute" target-entity="Sulu\Route\Domain\Model\Route">
             <join-column name="parent_id" referenced-column-name="id" nullable="true" on-delete="SET NULL"/>
--- a/packages/route/config/doctrine/Route/Route.orm.xml
+++ b/packages/route/config/doctrine/Route/Route.orm.xml
@@ -7,8 +7,8 @@
         table="ro_routes"
     >
         <unique-constraints>
-            <unique-constraint name="ro_routes_unique" fields="site, locale, slug" />
-            <unique-constraint name="ro_routes_resource_unique" fields="site, locale, resourceKey, resourceId" />
+            <unique-constraint name="ro_routes_unique" fields="site,locale,slug" />
+            <unique-constraint name="ro_routes_resource_unique" fields="site,locale,resourceKey,resourceId" />
         </unique-constraints>
 
         <id name="id" type="integer" column="id">
--- a/packages/route/config/doctrine/Route/Route.orm.xml
+++ b/packages/route/config/doctrine/Route/Route.orm.xml
@@ -8,16 +8,19 @@
     >
         <unique-constraints>
             <unique-constraint name="ro_routes_unique" fields="site,locale,slug" />
-            <unique-constraint name="ro_routes_resource_unique" fields="site,locale,resourceKey,resourceId" />
         </unique-constraints>
+
+        <indexes>
+            <index name="ro_routes_resource" fields="locale,resourceKey,resourceId"/>
+        </indexes>
 
         <id name="id" type="integer" column="id">
             <generator strategy="AUTO"/>
         </id>
 
-        <field name="site" type="string" length="16" nullable="true"/>
-        <field name="locale" type="string" length="11"/>
-        <field name="slug" type="string" length="164"/>
+        <field name="site" type="string" length="32" nullable="true"/>
+        <field name="locale" type="string" length="15"/>
+        <field name="slug" type="string" length="144"/>
 
         <many-to-one field="parentRoute" target-entity="Sulu\Route\Domain\Model\Route">
             <join-column name="parent_id" referenced-column-name="id" nullable="true" on-delete="SET NULL"/>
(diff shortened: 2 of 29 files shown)


composer.json
```diff
@@ -36,7 +36,7 @@
         "doctrine/cache": "^1.12.1 || ^2.0",
         "doctrine/collections": "^1.6 || ^2.0",
         "doctrine/data-fixtures": "^1.4",
-        "doctrine/dbal": "^3.3",
+        "doctrine/dbal": "^3.6",
```
alexander-schranz, line 39: > Deprecated `Connection::PARAM_*_ARRAY` constants: Use the corresponding constants on `ArrayParameterType` instead.

So I also replaced it in AccessControlRepository

The 3.6 was released 2 years ago: https://github.com/doctrine/dbal/releases/3.6.0 so we are fine to upgrade.

src/Sulu/Bundle/SecurityBundle/Entity/AccessControlRepository.php
```diff
@@ -87,7 +88,7 @@ public function findIdsWithGrantedPermissions(
             ->andWhere('accessControl.role IN (' . $systemRoleQueryBuilder->getDQL() . ')')
             ->setParameter('system', $system)
             ->setParameter('entityClass', $entityClass)
-            ->setParameter('entityIds', $entityIds, Connection::PARAM_STR_ARRAY)
+            ->setParameter('entityIds', $entityIds, ArrayParameterType::STRING)
```
alexander-schranz, line 90: DoctrineDBAL:

> Deprecated `Connection::PARAM_*_ARRAY` constants: Use the corresponding constants on `ArrayParameterType` instead.

packages/route/config/doctrine/Route/Route.orm.xml
```diff
@@ -0,0 +1,29 @@
...
+            <unique-constraint name="ro_routes_resource_unique" fields="site,locale,resourceKey,resourceId" />
+        </unique-constraints>
+
+        <id name="id" type="integer" column="id">
+            <generator strategy="AUTO"/>
+        </id>
+
+        <field name="site" type="string" length="16" nullable="true"/>
+        <field name="locale" type="string" length="11"/>
+        <field name="slug" type="string"/>
```
alexander-schranz: To support or versions of mysql site + locale + slug can not be longer together as `191` chars. Which means in this case 191 - 16 -11 = 164 chars for the slug.

There would be one workaround to support longer versions of URL by using a additional `slugHash` field based on widely supported `SHA1` algorithm. That way we can support longer Urls 164.

@benr77 already requested that the current `title` of 64 is to small for his usecases. If we increase that to 128 or 191. A longer parent title and a child page or 3 levels of navigation could already hit the 191 limit and we would require trim the url correctly. But that is more a theoretical issue. Which may already existed before.

alexander-schranz: Interesting the workaround with a `Hash` is also documented in MySQL docs itself but as a alternative for composite indexes so n our case it would include the site, locale, slug: https://dev.mysql.com/doc/refman/8.4/en/multiple-column-indexes.html

packages/route/src/Infrastructure/Doctrine/EventListener/RouteChangedUpdater.php
```diff
@@ -0,0 +1,124 @@
...
+
+            // update child and grand routes
+            $updateQueryBuilder = $connection->createQueryBuilder()
+                ->update($routesTableName, 'r')
+                ->set('slug', 'CONCAT(:newSlug' . $newSlugCast . ', SUBSTRING(slug, ' . (\strlen($oldSlug) + 1) . '))')
+                ->setParameter('newSlug', $newSlug, ParameterType::STRING)
+                ->where('parent_id IN (:parentIds)')
+                ->setParameter('parentIds', $parentIds, ArrayParameterType::INTEGER);
+
+            $updateQueryBuilder->executeStatement();
```
alexander-schranz: Not ideal handling of Postgres but works now as expected and is covers by tests.

Else the update query ends with follorwing error:

> Doctrine\DBAL\Exception\DriverException: An exception occurred while executing a query: SQLSTATE[42P18]: Indeterminate datatype: 7 ERROR:  could not determine data type of parameter $1

packages/route/config/doctrine/Route/Route.orm.xml
```diff
@@ -0,0 +1,29 @@
...
+            <unique-constraint name="ro_routes_unique" fields="site,locale,slug" />
+            <unique-constraint name="ro_routes_resource_unique" fields="site,locale,resourceKey,resourceId" />
+        </unique-constraints>
+
+        <id name="id" type="integer" column="id">
+            <generator strategy="AUTO"/>
+        </id>
+
+        <field name="site" type="string" length="16" nullable="true"/>
+        <field name="locale" type="string" length="11"/>
```
Prokyonn: I guess we should increase that also to `15`

packages/route/src/Domain/Model/Route.php
```diff
@@ -0,0 +1,86 @@
...
+
+    private ?string $locale;
+
+    private ?string $slug;
+
+    private ?Route $parentRoute;
+
+    private ?string $resourceKey;
+
+    private ?string $resourceId;
```
Prokyonn: Why are these all nullable?

packages/route/config/doctrine/Route/Route.orm.xml
alexander-schranz: changed the column length have to adopted the site and slug length to match into a index of 191. added a test case so we not accidently increase it.

packages/route/src/Domain/Model/Route.php
alexander-schranz: good catch, they should not.

packages/route/src/Domain/Model/Route.php
```diff
@@ -0,0 +1,88 @@
...
+    {
+        $this->slug = $slug;
+
+        return $this;
+    }
+
+    public function setParentRoute(?Route $parentRoute): void
+    {
+        $this->parentRoute = $parentRoute;
+    }
```
Prokyonn:
```suggestion
    public function setParentRoute(?Route $parentRoute): static
    {
        $this->parentRoute = $parentRoute;
        
        return $this;
    }
```

packages/route/src/Infrastructure/Doctrine/EventListener/RouteChangedUpdater.php
```diff
@@ -0,0 +1,126 @@
...
+use Doctrine\DBAL\ParameterType;
+use Doctrine\DBAL\Platforms\PostgreSQLPlatform;
+use Doctrine\ORM\Event\OnClearEventArgs;
+use Doctrine\ORM\Event\PostFlushEventArgs;
+use Doctrine\ORM\Event\PreUpdateEventArgs;
+use Sulu\Route\Domain\Model\Route;
+use Symfony\Contracts\Service\ResetInterface;
+
+/**
+ * @internal No BC promise are given for this class. Can be changed or removed at any time.
```
Prokyonn:
```suggestion
 * @internal No BC promises are given for this class. It may be changed or removed at any time.
```

packages/route/src/Infrastructure/Doctrine/Repository/RouteRepository.php
```diff
@@ -0,0 +1,96 @@
...
+
+use Doctrine\ORM\EntityManagerInterface;
+use Doctrine\ORM\EntityRepository;
+use Doctrine\ORM\Query;
+use Doctrine\ORM\QueryBuilder;
+use Sulu\Route\Domain\Model\Route;
+use Sulu\Route\Domain\Repository\RouteRepositoryInterface;
+
+/**
+ * @phpstan-import-type RouteFilter from \Sulu\Route\Domain\Repository\RouteRepositoryInterface
```
Prokyonn:
```suggestion
 * @phpstan-import-type RouteFilter from RouteRepositoryInterface
```

I think you can omit the full path

packages/route/src/Domain/Model/Route.php
```diff
@@ -0,0 +1,88 @@
...
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Route\Domain\Model;
+
+class Route
```
Prokyonn, line 17: Does it make sense to implement an interface to make it better extendable? 🤔

alexander-schranz: Would currently not yet support that. adding later interface isn't a bc break here. currently finalize the class.

alexander-schranz, line 17:
```suggestion
final class Route
```

packages/route/src/Domain/Model/Route.php
```diff
@@ -0,0 +1,90 @@
...
+    {
+        $this->slug = $slug;
+
+        return $this;
+    }
+
+    public function setParentRoute(?Route $parentRoute): static
+    {
+        $this->parentRoute = $parentRoute;
+
```
alexander-schranz:
```suggestion

```

packages/route/src/Domain/Model/Route.php
```diff
@@ -0,0 +1,89 @@
...
+    {
+        $this->slug = $slug;
+
+        return $this;
+    }
+
+    public function setParentRoute(?Route $parentRoute): static
+    {
+        $this->parentRoute = $parentRoute;
+        return $this;
```
alexander-schranz:
```suggestion
        $this->parentRoute = $parentRoute;

        return $this;
```

packages/route/src/Domain/Model/Route.php
```diff
@@ -0,0 +1,90 @@
...
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Route\Domain\Model;
+
+final class Route
```
alexander-schranz:
```suggestion
/**
 * @final
 */
class Route
```

Prokyonn (approved):

alexander-schranz:
8.4 error is unrelated to the PR. Looks like some issues in one of our dependencies maybe only a dev anomaly but we should test what is causing it symfony or the 2fa bundle.

PR #7727, opened 2025-01-10
Add rule to automatically mark classes internal

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?
A rule that marks all classes that implement a certain interface or extend a certain class as `@internal`

#### Why?
So that the developers don't have to think about that too hard and can php-cs-fixer do it's job.


src/CodeStyle/InternalClassFixer.php
```diff
@@ -0,0 +1,266 @@
...
+                if (\in_array($tokens[$i]->getContent(), $this->configuration['base_type'])) {
+                    return true;
+                }
+            }
+        }
+
+        return true;
+    }
+
+    // ============================ THE FOLLOWING METHODS ARE COPIES FROM PhpUnitFixer ===============================
```
mamazu: I wasn't really able to extract anything out of the class, so copying is the only way of extending it. But we could probably slim down the implementation a little and make it less generic.

src/CodeStyle/InternalClassFixer.php
```diff
@@ -0,0 +1,266 @@
...
+            [],
+        );
+    }
+
+    /**
+     * Determines if a class should be refactored.
+     * It should not be refactored if it does not implement or extend anything
+     * It should be refactored if one of the implemented Interfaces or extended classes are in the base_class list
+     *
+     * Limitations: Not checking namespaces, aliased imports
```
mamazu: It should be sufficient and if not we can still exclude files manually.

mamazu:
The big question is however, which classes are still missing that should be marked as internal? Currently it's just the `CompilerPassInterface`.

src/CodeStyle/InternalClassFixer.php
alexander-schranz, line 1: we should move this to `tests` directory this is nothing which is shipped with Sulu. And we handle things already in `tests/phpstan` we can also `tests/php-cs-fixers` we should add a test namespace to avoid `require` in the php cs fixer.

.php-cs-fixer.dist.php
```diff
@@ -1,5 +1,7 @@
 <?php
 
+require __DIR__ .'/vendor/autoload.php';
```
alexander-schranz, line 3: Think the autoloader is already called else PhpCsFixer would not already work?

mamazu: Nope, I tried removing it and it crashed.

mamazu:
I added a comment to the generated annotation. But that should be all.

mamazu:
Also: https://github.com/phpstan/phpstan/issues/1178

Then phpstan will also complain about you using them in a different context.

composer.json
```diff
@@ -222,7 +222,8 @@
     },
     "autoload-dev": {
         "psr-4": {
-            "App\\": "tests/"
+            "App\\": "tests/",
+            "PhpCsFixer\\": "tests/php-cs-fixer/"
```
alexander-schranz: would not introduce a new namespace lets us the `tests` namespace

tests/php-cs-fixer/InternalClassFixer.php
```diff
@@ -0,0 +1,272 @@
...
+/*
+ * This file is part of Sulu.
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer;
```
alexander-schranz:
```suggestion
namespace App\PhpCsFixer;
```

.php-cs-fixer.dist.php
```diff
@@ -14,8 +16,12 @@
     ->in(__DIR__);
 
 $config = new PhpCsFixer\Config();
-$config->setRiskyAllowed(true)
+$config
+    ->setRiskyAllowed(true)
     ->setParallelConfig(PhpCsFixer\Runner\Parallel\ParallelConfigFactory::detect())
+    ->registerCustomFixers([
+        new PhpCsFixer\InternalClassFixer(),
```
alexander-schranz:
```suggestion
        new App\PhpCsFixer\InternalClassFixer(),
```

PR #7730, opened 2025-01-14
Fix null deprecations for rtrim and json_decode

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Adds some checks to prevent deprecations:
`Deprecated: rtrim(): Passing null to parameter #1 ($string) of type string is deprecated`
`Deprecated: json_decode(): Passing null to parameter #1 ($json) of type string is deprecated`
`Deprecated: usort(): Returning bool from comparison function is deprecated, return an integer less than, equal to, or greater than zero`

### DIFF
--- a/src/Sulu/Bundle/PageBundle/Teaser/TeaserContentType.php
+++ b/src/Sulu/Bundle/PageBundle/Teaser/TeaserContentType.php
@@ -56,7 +56,7 @@ class TeaserContentType extends SimpleContentType implements PreResolvableConten
 
     protected function decodeValue($value)
     {
-        return \json_decode($value, true);
+        return $value !== null ? json_decode($value, true) : null;
     }
 
     protected function encodeValue($value)
--- a/src/Sulu/Bundle/PageBundle/Teaser/TeaserContentType.php
+++ b/src/Sulu/Bundle/PageBundle/Teaser/TeaserContentType.php
@@ -56,7 +56,7 @@ class TeaserContentType extends SimpleContentType implements PreResolvableConten
 
     protected function decodeValue($value)
     {
-        return $value !== null ? json_decode($value, true) : null;
+        return null !== $value ? \json_decode($value, true) : null;
     }
 
     protected function encodeValue($value)
--- a/src/Sulu/Component/Content/Types/ResourceLocator/Mapper/PhpcrMapper.php
+++ b/src/Sulu/Component/Content/Types/ResourceLocator/Mapper/PhpcrMapper.php
@@ -214,7 +214,7 @@ class PhpcrMapper implements ResourceLocatorMapperInterface
         \usort(
             $result,
             function(ResourceLocatorInformation $item1, ResourceLocatorInformation $item2) {
-                return $item1->getCreated() < $item2->getCreated();
+                return $item2->getCreated() <=> $item1->getCreated();
             }
         );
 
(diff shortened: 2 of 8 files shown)


src/Sulu/Component/Webspace/Manager/WebspaceManager.php
```diff
@@ -288,7 +291,7 @@ public function findUrlByResourceLocator(
 
         $this->portalUrlCache[$webspaceKey][$domain][$environment][$languageCode] = $portalUrl;
 
-        if (!$portalUrl) {
+        if (!$portalUrl || $resourceLocator === null) {
```
alexander-schranz: how it can be null if it overwritten above?

```
        if (null === $resourceLocator) {
            $resourceLocator = '/';
        }
```

Looks like a always false which may phpstan catches?

src/Sulu/Component/Content/Types/ResourceLocator/Mapper/PhpcrMapper.php
```diff
@@ -214,7 +214,7 @@ function($resourceLocator, NodeInterface $node) use (&$result) {
         \usort(
             $result,
             function(ResourceLocatorInformation $item1, ResourceLocatorInformation $item2) {
-                return $item1->getCreated() < $item2->getCreated();
+                return $item2->getCreated() <=> $item1->getCreated();
```
alexander-schranz, line 217: looks like something we should test?

src/Sulu/Bundle/PageBundle/Teaser/TeaserContentType.php
```diff
@@ -56,7 +56,7 @@ public function getDefaultParams(?PropertyInterface $property = null)
 
     protected function decodeValue($value)
     {
-        return \json_decode($value, true);
+        return $value !== null ? json_decode($value, true) : null;
```
alexander-schranz: Run:

```
composer config minimum-stability dev
composer update "symfony/*" "doctrine/*"
```

To run `composer lint` and `composer fix`

src/Sulu/Component/Content/Types/ResourceLocator/Mapper/PhpcrMapper.php
alexander-schranz: okay result didn't change just the return type?

src/Sulu/Bundle/PageBundle/Teaser/TeaserContentType.php
alexander-schranz:
```suggestion
        return $value !== null ? \json_decode($value, true) : null;
```

src/Sulu/Component/Content/Types/ResourceLocator/Mapper/PhpcrMapper.php
Prokyonn: yes only return type, it is still descending but with the spaceship operator I had to change the item2/item1 order and a test for this already exists PhpcrMapperTest::testLoadHistoryByContentUuid

src/Sulu/Component/Webspace/Manager/WebspaceManager.php
Prokyonn: Yes I first had the intention to catch it and return `null`, but the current implementation falls back to an empty string which will return the homepage, so I forgot to revert this part

alexander-schranz (approved):

alexander-schranz (approved):

PR #7733, opened 2025-01-16
Update phpstan and rector to version 2

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no 
| BC breaks? | no 
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Update phpstan and rector ^2 version.

#### Why?

Using the latest phpstan version and rector version to get better code analysis.


composer.json
```diff
@@ -154,8 +154,7 @@
         "symfony/runtime": "^5.4 || ^6.0 || ^7.0",
         "symfony/stopwatch": "^5.4 || ^6.0 || ^7.0",
         "symfony/var-dumper": "^5.4 || ^6.0 || ^7.0",
-        "symfony/web-profiler-bundle": "^5.4 || ^6.0 || ^7.0",
-        "thecodingmachine/phpstan-strict-rules": "^1.0"
```
alexander-schranz, line 158: this package seems unmaintained and not supports phpstan 2.0

.gitignore
```diff
@@ -40,6 +40,9 @@ phpunit.xml
 .phpunit.result.cache
 .phpunit.cache
 
+# phpstan
+phpstan.neon
+
```
alexander-schranz, line 45: we use the `phpstan.dist.neon` which is commited now for local overrides the phpstan.neon can be used.

phpstan-baseline.neon
alexander-schranz, line 1: There are a lot of new issues. We may need to check if some new rules of phpstan doesn't make sense for us. At current state I just let all enabled, think best is to test it when we merge this back to `3.0` branch and test it against the new bundles inside the `packages` directory to check what errors PHPStan would report there.

/cc @sulu/core-developer

Prokyonn (approved):

mamazu:
Most of the new errors are missing parameter types. This can be easily fixed by using a rector to fix the doctypes as real parameter types. Furthermore now every issue also has an `identifier` property. This blows up the baseline pretty good. But yes we still should try to reduce the size of the baseline.

PR #7737, opened 2025-01-17
Fix issue with trailing slash in Matomo analytics templates

### LINKED ISSUE #7732: Sulu 3.0 Matomo URLs are not correct
| Q | A
| --- | ---
| Sulu Version | 3.0
| PHP Version | 8.4

# Actual Behavior

The Matomo analytics feature has been updated to refer to `matomo.js` instead of the legacy `piwik.js`.

However, something appears to have changed regarding the trailing slash on the URL, and or the concatenation of the URL with the JS file.

# Expected Behavior

We get a URL formatted as `https://my-matomo-instance.com/matomo.js`.

# Steps to Reproduce

In the admin, set `https://my-matomo-instance.com/` (with a trailing slash).

In the front-end HTML, you get `var u="https://my-matomo-instance.com";` which has no trailing slash, which is then concatenated a couple of lines later using `g.src=u+'matomo.js';` which results in `https://my-matomo-instance.commatomo.js` which is not correct and has a missing `/` character.

Secondly, there is a fallback `` element, which gets a URL with a double forward slash e.g. `https://my-matomo-instance.com//matomo.js` which is also incorrect.

# Solution

Canonicalise the URL with a trailing slash when saving to the database, and then correct the template file to not add additional slashes.

fnagel:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #7732
| Related issues/PRs | #7692 #7693
| License | MIT

#### What's in this PR?

Fixed for missing trailing slashes.

#### Why?

In order to fix Matomo URLs.

#### Context

Follow-up to 84a6d7bf61e3e642064e91eb8fd9eaa50488ef67

### DIFF
--- a/src/Sulu/Bundle/WebsiteBundle/Resources/views/Analytics/matomo/body-open.html.twig
+++ b/src/Sulu/Bundle/WebsiteBundle/Resources/views/Analytics/matomo/body-open.html.twig
@@ -1,1 +1,1 @@
-<noscript><p><img referrerpolicy="no-referrer-when-downgrade" src="{{ analytics.content.url }}/matomo.php?idsite={{ analytics.content.siteId }}&amp;rec=1" style="border:0;" alt="" /></p></noscript>
+<noscript><p><img referrerpolicy="no-referrer-when-downgrade" src="{{ analytics.content.url|trim('/', 'right') }}/matomo.php?idsite={{ analytics.content.siteId }}&amp;rec=1" style="border:0;" alt="" /></p></noscript>
--- a/src/Sulu/Bundle/WebsiteBundle/Resources/views/Analytics/matomo/head-close.html.twig
+++ b/src/Sulu/Bundle/WebsiteBundle/Resources/views/Analytics/matomo/head-close.html.twig
@@ -3,10 +3,10 @@
     _paq.push(['trackPageView']);
     _paq.push(['enableLinkTracking']);
     (function() {
-        var u="{{ analytics.content.url|trim('/') }}";
-        _paq.push(['setTrackerUrl', u+'matomo.php']);
+        var u="{{ analytics.content.url|trim('/', 'right') }}";
+        _paq.push(['setTrackerUrl', u+'/matomo.php']);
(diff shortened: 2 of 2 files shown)


src/Sulu/Bundle/WebsiteBundle/Resources/views/Analytics/matomo/body-open.html.twig
```diff
@@ -1 +1 @@
-<noscript><p><img referrerpolicy="no-referrer-when-downgrade" src="{{ analytics.content.url }}/matomo.php?idsite={{ analytics.content.siteId }}&amp;rec=1" style="border:0;" alt="" /></p></noscript>
+<noscript><p><img referrerpolicy="no-referrer-when-downgrade" src="{{ analytics.content.url|trim('/') }}/matomo.php?idsite={{ analytics.content.siteId }}&amp;rec=1" style="border:0;" alt="" /></p></noscript>
```
alexander-schranz: why its more technical we should only [rtrim](https://twig.symfony.com/doc/1.x/filters/trim.html) url here only,  technical URLs are allowed to start with `//` which are ["Protocol-relative URLs":](https://en.wikipedia.org/wiki/URL#Protocol-relative_URLs)

```suggestion
<noscript><p><img referrerpolicy="no-referrer-when-downgrade" src="{{ analytics.content.url|trim('/', 'right') }}/matomo.php?idsite={{ analytics.content.siteId }}&amp;rec=1" style="border:0;" alt="" /></p></noscript>
```

fnagel: Sounds good to me. Changed the commit and updated the PR.

alexander-schranz (approved):

alexander-schranz:
@fnagel thank you!

benr77:
I've updated my 3.0 branch and this is now working perfectly. Thanks!

PR #7741, opened 2025-01-19
Running phpstan baseline update

mamazu:
Running `phpstan -b` acutally made the baseline smaller.


alexander-schranz:
I did not add things to the baseline yet because some parts need to be discussed. Whatever make phpstan less strict or change handling things inside the new code of `packages/` directory.

mamazu:
If you need me to help with typing I'm always up for it. :D

Issue #7547, opened 2024-07-26
Remove deprecated code for version Sulu version 3


mamazu:
| Q | A
| --- | ---
| Sulu Version | 3.0
| PHP Version | -
| DB Version | -
| Browser Version | -

For version 3 we should clean up the structure to actually remove stuff that's deprecated.

## Todos

- [x] Remove KernelBrowser (#7718)
- [x] Remove User Salt functionality (waits for #7743)
- [x] Remove Symfony BC layer (https://github.com/sulu/sulu/pull/7743)
- [x] `DeprecatedCompilerPass` should have been removed in 2.0
    - [x] #7721 (for removing Symfony 7.2 deprecations)
    - [x] #7424 
    - [x] #7405
    - [x] #7403 
- [x] `BreadcrumbItemInterface`
- [x] `TokenAlreadyRequestedException.php`
- [ ] Deprecated function in the `TabViewBuilderTrait`
- [x] `Sulu/Bundle/WebsiteBundle/Twig/Core/UtilTwigExtension::extract`
- [x] `Sulu/Bundle/WebsiteBundle/Twig/Meta/MetaTwigExtension`
- [x] `Sulu/Bundle/WebsiteBundle/Controller/WebsiteController::getRequest`
- [x]  `Sulu/Bundle/WebsiteBundle/Controller/RedirectController::redirectWebspaceAction`
- [ ] `Sulu\Bundle\MediaBundle\Media\ImageConverter\Transformation\CropTransformation`
- [x] `Container Parameter 'sulu_media.media.max_file_size'` (https://github.com/sulu/sulu/pull/7766)
- [ ] ContactBundle `Account::notes`
- [x] SecurityBundle `Permission::module`
- [ ] `Sulu/Bundle/ContactBundle/Controller/AbstractMediaController`
- [x] `ReturnTypeWillChange` add correct return type remove attribute (https://github.com/sulu/sulu/pull/8170)
- [ ] Migrate usage of `RequestParameterTrait` to `$request->get()` #7815 

### Js
- [x] securityContextStore.js
- [x] localizationStore.js `loadLocalizations`
- [ ] ToolbarActions have a lot of deprecated code in them
- [x] FullLoadingStrategy.js
- [x] webspaceStore.js (#7763)

mamazu:
This is probably more something for 4.0 or something as we don't want to have too many BC breaks in this version (as the main focus is getting rid of phpcr)

alexander-schranz:
[...]

PR #7743, opened 2025-01-20
Remove not longer required Symfony 5.4 backwards compatibility layer for Password Encoder and LogoutSuccessHandler

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | -
| Fixed tickets | -
| Related issues/PRs | https://github.com/sulu/sulu/pull/7657 #7547
| License | MIT
| Documentation PR | -

#### What's in this PR?

Removing old configurations of Symfony 5.4 since we don't support the version anyways.

#### Why?
In Sulu 3 the minimum version is 6.4.

Less things we need to port and maintain.


src/Sulu/Bundle/SecurityBundle/DependencyInjection/Compiler/AliasForSecurityEncoderCompilerPass.php
```diff
@@ -1,28 +0,0 @@
...
- * This source file is subject to the MIT license that is bundled
- * with this source code in the file LICENSE.
- */
-
-namespace Sulu\Bundle\SecurityBundle\DependencyInjection\Compiler;
-
-use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
-use Symfony\Component\DependencyInjection\ContainerBuilder;
-
-class AliasForSecurityEncoderCompilerPass implements CompilerPassInterface
```
mamazu: Technically the class is not deprecated, but the contents of the method is. Should I still remove it?

src/Sulu/Bundle/SecurityBundle/Resources/config/services.xml
```diff
@@ -39,6 +39,7 @@
     </parameters>
 
     <services>
+        <service id="sulu_security.encoder_factory" alias="security.password_hasher_factory" />
```
alexander-schranz: We could remove the `sulu_security.encoder_factory` and replace it with `security.password_hasher_factory` as the bc layer between encoder and hasher is not longer required.

alexander-schranz: And type it now as PasswordHasherFactoryInterface: https://github.com/sulu/sulu/blob/71d5cd1fd1ed3b4c1029e6e29a77ae07f7b758d4/src/Sulu/Bundle/SecurityBundle/Controller/ResettingController.php#L72 instead of the phpdoc

src/Sulu/Bundle/SecurityBundle/UserManager/UserManager.php
```diff
@@ -36,18 +36,14 @@
...
 {
     use RelationTrait;
 
-    /**
-     * @param PasswordHasherFactoryInterface|EncoderFactoryInterface|null $passwordHasherFactory
-     */
     public function __construct(
         protected ObjectManager $em,
-        private $passwordHasherFactory,
+        private ?PasswordHasherFactoryInterface $passwordHasherFactory,
```
alexander-schranz, line 44: It should not longer be the case that the hasher encoder can be null. So I think we can remove nullability here.

```suggestion
        private PasswordHasherFactoryInterface $passwordHasherFactory,
```

If it fails the test lets tackle it in an own PR updating the test kernel

mamazu:
This also means the entire functionality around the Salt can be removed. As the modern Symfony algorithms don't use that.

mamazu:
@alexander-schranz The UserManager changes seem to cause problems in the tests but in the dev container the service exists.

alexander-schranz:
lets revert that and keep the `on-invalid="null"` for the user manager

mamazu:
Done

src/Sulu/Bundle/SecurityBundle/User/UserProvider.php
```diff
@@ -32,7 +32,6 @@
  * configuration from the webspaces into account.
  *
  * @implements UserProviderInterface<UserInterface>
- * @implements PasswordUpgraderInterface<User>
```
alexander-schranz: this looks unexpected revert

alexander-schranz (approved):

alexander-schranz:
Thx @mamazu

PR #7749, opened 2025-01-24
Add new routing to new content integration

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? |yes
| BC breaks? | yes
| Deprecations? | yes 
| Fixed tickets | fixes # 
| Related issues/PRs | https://github.com/sulu/sulu/pull/7726, https://github.com/sulu/skeleton/pull/276
| License | MIT

#### What's in this PR?

Add new route implementation for the content bundle. This PR adds a few hacks and things we should revert before the stable release. Most things where added to keep compatibility to older bundles and routings.

#### Why?

Created the new route implementation in https://github.com/sulu/sulu/pull/7726

#### ToDo: Done in https://github.com/sulu/sulu/pull/7726

 - [x] Create Entity see https://github.com/sulu/sulu/pull/7726
 - [x] Childs are update see https://github.com/sulu/sulu/pull/7726
 - [x] Grand childs are update see https://github.com/sulu/sulu/pull/7726
 - [x] Multi Localization support see https://github.com/sulu/sulu/pull/7726
 - [x] Multi Site support see https://github.com/sulu/sulu/pull/7726
 - [x] Support for non Site connected routes (article)
    - [x] Site A with `/test` and Site B `/test` and article parent Site A `/test/article` 
 
#### Related Pull Requests

 - [x] History Handling https://github.com/sulu/sulu/pull/7830
 - [x] Add Multi Site tests https://github.com/sulu/sulu/pull/7831
 - [x] Add Multi Localization https://github.com/sulu/sulu/pull/7831
 - [x] Add CrossSite Site test (article (site: null) -> page (site: website)) https://github.com/sulu/sulu/pull/7834
 - [x] Add Test for url with same prefix are not effect on the same level https://github.com/sulu/sulu/pull/7836
 - [x] Tree Full Edit https://github.com/sulu/sulu/pull/7835
 - [x] Route for AutoIncrement Entities: https://github.com/sulu/sulu/pull/7859
 - [x] Dynamic Router https://github.com/sulu/sulu/pull/7860
 - [x] Locale Switcher https://github.com/sulu/sulu/pull/7874
 - [x] Implement in Content Bundle (THIS PR)

### DIFF
--- a/packages/content/config/services.xml
+++ b/packages/content/config/services.xml
@@ -101,7 +101,9 @@
             <argument type="service" id="sulu_content.content_structure_bridge_factory"/>
             <argument type="service" id="sulu_http_cache.cache_lifetime.resolver"/>
 
-            <tag name="sulu_route.defaults_provider"/>
+            <tag name="sulu_route.route_defaults_provider" resource-key="examples"/>
+            <tag name="sulu_route.route_defaults_provider" resource-key="articles"/>
+            <tag name="sulu_route.route_defaults_provider" resource-key="pages"/>
         </service>
 
         <!-- Structure Bridge Factory -->
--- a/packages/page/src/Infrastructure/Sulu/Route/WebspaceSiteRouteGenerator.php
+++ b/packages/page/src/Infrastructure/Sulu/Route/WebspaceSiteRouteGenerator.php
@@ -55,7 +55,7 @@ class WebspaceSiteRouteGenerator implements SiteRouteGeneratorInterface
         $url = $this->webspaceManager->findUrlByResourceLocator($slug, null, $locale, $site, $requestContext->getHost(), $requestContext->getScheme());
 
         if (null === $url) {
-            throw new \RuntimeException(\sprintf('No found for "%s" in locale "%s" and site "%s".', $slug, $locale, $site));
+            throw new \RuntimeException(\sprintf('No url found for "%s" in locale "%s" and site "%s".', $slug, $locale, $site));
         }
 
         return $url;
--- a/packages/page/src/Infrastructure/Sulu/Route/WebspaceSiteRouteGenerator.php
+++ b/packages/page/src/Infrastructure/Sulu/Route/WebspaceSiteRouteGenerator.php
@@ -47,7 +47,6 @@ class WebspaceSiteRouteGenerator implements SiteRouteGeneratorInterface
 
             $site = $currentRequest->attributes->get(RequestAttributeEnum::SITE->value);
 
-
             if (!\is_string($site)) {
                 throw new MissingRequestContextParameterException(RequestAttributeEnum::SITE->value);
             }
--- a/packages/page/src/Infrastructure/Sulu/Route/WebspaceSiteRouteGenerator.php
+++ b/packages/page/src/Infrastructure/Sulu/Route/WebspaceSiteRouteGenerator.php
@@ -1,0 +1,64 @@
+<?php
+
+/*
+ * This file is part of Sulu.
+ *
+ * (c) Sulu GmbH
+ *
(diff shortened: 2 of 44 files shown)


packages/content/config/services.xml
```diff
@@ -101,7 +101,9 @@
             <argument type="service" id="sulu_content.content_structure_bridge_factory"/>
             <argument type="service" id="sulu_http_cache.cache_lifetime.resolver"/>
 
-            <tag name="sulu_route.defaults_provider"/>
+            <tag name="sulu_route.route_defaults_provider" resource-key="examples"/>
```
Prokyonn, line 104: I guess this is only a temporary solution?

packages/page/src/Infrastructure/Sulu/Route/WebspaceSiteRouteGenerator.php
```diff
@@ -0,0 +1,63 @@
...
+
+            if (!\is_string($site)) {
+                throw new MissingRequestContextParameterException(RequestAttributeEnum::SITE->value);
+            }
+        }
+
+        $url = $this->webspaceManager->findUrlByResourceLocator($slug, null, $locale, $site, $requestContext->getHost(), $requestContext->getScheme());
+
+        if (null === $url) {
+            throw new \RuntimeException(\sprintf('No found for "%s" in locale "%s" and site "%s".', $slug, $locale, $site));
```
Prokyonn:
```suggestion
            throw new \RuntimeException(\sprintf('No url found for "%s" in locale "%s" and site "%s".', $slug, $locale, $site));
```

packages/route/src/Infrastructure/Doctrine/Repository/RouteRepository.php
```diff
@@ -70,7 +70,7 @@ protected function createQueryBuilder(array $filters): QueryBuilder
         if (\array_key_exists('site', $filters)) {
             $site = $filters['site'] ?? null;
             $queryBuilder->andWhere(
-                null === $site ? 'route.site IS NULL' : 'route.site = :site'
+                null === $site ? 'route.site IS NULL' : '(route.site = :site OR route.site IS NULL)'
```
Prokyonn, line 73: What is the use case for `site = null OR site = :site` ?

packages/content/config/services.xml
alexander-schranz: @Prokyonn yes, we need kind of find out which `doctrine` entities exist at compile / compiler pass where doctrine service not exists yet. I couldn't find a quick way for this and decided for the alphe with this workaround.

packages/route/src/Infrastructure/Doctrine/Repository/RouteRepository.php
alexander-schranz: article bundle. The request analyzer will tell us we are on site `sulu` when opening the page `https://sulu.io/blog/some-blog` but we should still get the article which is not site specific back.

alexander-schranz: @Prokyonn if its unclear we could go or transparency we can also go with `'siteOrNull' => $site`

Prokyonn: Ahh makes sense 👍  no I don't think we need `siteOrNull` I just haven't thought about articles

Prokyonn (approved):

alexander-schranz:
As there exists now both routing tables for the alpha part we need change the PHPCR MIgration bundle to use `ro_routes_next` table instead of `ro_routes` and migrate it to that. @Prokyonn

PR #7751, opened 2025-01-27
Fix block response for minOccurs=1 / maxOccurs=1

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | part of https://github.com/sulu/sulu/issues/7672
| License | MIT

#### What's in this PR?

Adjusts the response of the `BlockPropertyResolver` to return the single block instead of a list when `minOccurs` and `maxOccurs` are `1`.

#### Why?

The behaviour should not be changed between Sulu 2.6 and 3.0.

### DIFF
--- a/packages/content/src/Application/PropertyResolver/Resolver/BlockPropertyResolver.php
+++ b/packages/content/src/Application/PropertyResolver/Resolver/BlockPropertyResolver.php
@@ -94,6 +94,13 @@ class BlockPropertyResolver implements PropertyResolverInterface
             );
         }
 
+        $minOccurs = $metadata->getMinOccurs();
+        $maxOccurs = $metadata->getMaxOccurs();
+
+        if (1 === $minOccurs && 1 === $maxOccurs && \count($contentViews) > 0) {
+            $contentViews = $contentViews[0]->getContent();
+        }
+
         return ContentView::create($contentViews, [...$returnedParams]);
     }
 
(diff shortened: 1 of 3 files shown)


packages/content/src/Application/PropertyResolver/Resolver/BlockPropertyResolver.php
```diff
@@ -94,6 +94,13 @@ public function resolve(mixed $data, string $locale, array $params = []): Conten
             );
         }
 
+        $minOccurs = $metadata->getMinOccurs();
+        $maxOccurs = $metadata->getMaxOccurs();
+
+        if (1 === $minOccurs && 1 === $maxOccurs && \count($contentViews) > 0) {
+            $contentViews = $contentViews[0]->getContent();
+        }
+
```
alexander-schranz: @benr77 can you test if this ends in the expected result?

packages/content/src/Application/PropertyResolver/Resolver/BlockPropertyResolver.php
```diff
@@ -94,6 +94,13 @@ public function resolve(mixed $data, string $locale, array $params = []): Conten
             );
         }
 
+        $minOccurs = $metadata->getMinOccurs();
+        $maxOccurs = $metadata->getMaxOccurs();
+
+        if (1 === $minOccurs && 1 === $maxOccurs && \count($contentViews) > 0) {
+            $contentViews = $contentViews[0]->getContent();
```
alexander-schranz: @Prokyonn aren't we loosing the `view` now?

packages/content/src/Application/PropertyResolver/Resolver/BlockPropertyResolver.php
benr77: Yes, this is working correctly.

packages/content/src/Application/PropertyResolver/Resolver/BlockPropertyResolver.php
Prokyonn: You were right, fixed it.

alexander-schranz (approved):

benr77:
Confirmed this is working as expected :)

PR #7759, opened 2025-01-29
Remove deprecated `SecurityType`

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | yes
| Deprecations? | -
| Fixed tickets | -
| Related issues/PRs | #7547
| License | MIT
| Documentation PR | -

#### What's in this PR?
Removing the default fixture for Security Type entities.

#### Why?

We don't need to provide empty fixtures to the user. They can create fixtures on their own for the entities they want.


mamazu:
We could also remove the feature completely as it is currently not used anymore in the core.

alexander-schranz:
The whole `SecurityType` model/entity and its repository can be removed.

alexander-schranz:
we should not forgot to add a `UPGRADE.md` with a example mysql SQL to remove the tables / columns.

alexander-schranz (approved):

mamazu:
Now it should be ready

alexander-schranz:
Awesome thank you!

PR #7770, opened 2025-01-31
Mark deprecated twig functions

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | yes 
| Fixed tickets | -
| Related issues/PRs | #7769 
| License | MIT
| Documentation PR | -

#### What's in this PR?
Marking twig functions as deprecated.

#### Why?
We want to let the user know what functions are going to be removed in the next version.

### DIFF
--- a/src/Sulu/Bundle/WebsiteBundle/Twig/Core/UtilTwigExtension.php
+++ b/src/Sulu/Bundle/WebsiteBundle/Twig/Core/UtilTwigExtension.php
@@ -26,14 +26,22 @@ class UtilTwigExtension extends AbstractExtension
         return [
             new TwigFilter('sulu_util_multisort', 'Sulu\Component\Util\SortUtils::multisort'),
             new TwigFilter('sulu_util_filter', 'Sulu\Component\Util\ArrayUtils::filter'),
-            new TwigFilter('sulu_util_domain_info', [$this, 'extract']),
+            new TwigFilter(
+                'sulu_util_domain_info',
+                [$this, 'extract'],
+                ['deprecated' => true],
+            ),
         ];
     }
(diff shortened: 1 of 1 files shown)


src/Sulu/Bundle/WebsiteBundle/Twig/Core/UtilTwigExtension.php
```diff
@@ -26,14 +27,22 @@ public function getFilters()
         return [
             new TwigFilter('sulu_util_multisort', 'Sulu\Component\Util\SortUtils::multisort'),
             new TwigFilter('sulu_util_filter', 'Sulu\Component\Util\ArrayUtils::filter'),
-            new TwigFilter('sulu_util_domain_info', [$this, 'extract']),
+            new TwigFilter(
+                'sulu_util_domain_info',
+                [$this, 'extract'],
+                ['deprecation_info' => new DeprecatedCallableInfo(package: 'sulu/sulu', version: 2.3)],
```
alexander-schranz: not supported by our range of supported twig versions

mamazu: Then just deprecated => true?

alexander-schranz: if that works for all our supported twig versions

mamazu: Yes, that behaviour works in the old versions and (deprecated) in the new versions.

PR #7721, opened 2025-01-08
Use `tagged_iterator` instead of deprecated `tagged`

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no (Only applications below Symfony 4.4)
| Deprecations? | Removing ones
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?
Replacing the service type `tagged` with the `tagged_iterator`.

#### Why?
Less deprecations from Symfony:
```
Since symfony/dependency-injection 7.2: Type "tagged" is deprecated for tag <argument>, use "tagged_iterator" instead in "/home/mamazu/packages/sulu/sulu/packages/content/config/services.xml".
```

In Symfony the `tagged` type was introduced in 4.4 and Symfony converted to the new type in 5.0.

PR #7776, opened 2025-02-03
Adds missing return types & fixes deprecations

opctim:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? | no 
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | none
| Related issues/PRs | none
| License | MIT
| Documentation PR | none

#### What's in this PR?

A basic setup of sulu (out of the box) causes ~ 350 deprecation log messages to pop up in profiler. The sheer amount makes it hard to distiguish between user- and framework-caused deprecations.

#### Why?

This PR fixes lots of deprecations by using different (new) interfaces and adding non-bc return types to EventSubscribers, kernel extensions, compiler passes, etc. I structured my changes in commits for each type. 

Now, after my improvements, my application gives me ~ 140 (210 less) deprecation log messages.

#### Testing

Locally, besides producing less deprecation messages, no difference in behaviour has been observed. Everything tested worked as expected.

#### To Do

### DIFF
--- a/config/packages/doctrine.yaml
+++ b/config/packages/doctrine.yaml
@@ -9,6 +9,7 @@ doctrine:
         auto_generate_proxy_classes: true
         naming_strategy: doctrine.orm.naming_strategy.underscore_number_aware
         auto_mapping: true
+        enable_lazy_ghost_objects: true
 #        mappings:
 #            App:
 #                is_bundle: false
--- a/src/Sulu/Bundle/ActivityBundle/Application/Subscriber/DispatchSpecificDomainEventSubscriber.php
+++ b/src/Sulu/Bundle/ActivityBundle/Application/Subscriber/DispatchSpecificDomainEventSubscriber.php
@@ -22,7 +22,7 @@ class DispatchSpecificDomainEventSubscriber implements EventSubscriberInterface
     ) {
     }
 
-    public static function getSubscribedEvents()
+    public static function getSubscribedEvents(): array
     {
         return [
             DomainEvent::class => ['dispatchDomainEventWithSpecificEventName', 0],
--- a/src/Sulu/Bundle/ActivityBundle/Application/Subscriber/SetDomainEventUserSubscriber.php
+++ b/src/Sulu/Bundle/ActivityBundle/Application/Subscriber/SetDomainEventUserSubscriber.php
@@ -24,7 +24,7 @@ class SetDomainEventUserSubscriber implements EventSubscriberInterface
     ) {
     }
 
-    public static function getSubscribedEvents()
+    public static function getSubscribedEvents(): array
     {
         return [
             DomainEvent::class => ['setDomainEventUser', 256],
--- a/src/Sulu/Bundle/ActivityBundle/Application/Subscriber/StoreActivitySubscriber.php
+++ b/src/Sulu/Bundle/ActivityBundle/Application/Subscriber/StoreActivitySubscriber.php
@@ -22,7 +22,7 @@ class StoreActivitySubscriber implements EventSubscriberInterface
     ) {
     }
 
-    public static function getSubscribedEvents()
+    public static function getSubscribedEvents(): array
     {
         return [
             DomainEvent::class => ['storeActivity', -256],
--- a/src/Sulu/Bundle/AdminBundle/Command/DownloadLanguageCommand.php
+++ b/src/Sulu/Bundle/AdminBundle/Command/DownloadLanguageCommand.php
@@ -36,7 +36,7 @@ class DownloadLanguageCommand extends Command
         parent::__construct();
     }
 
-    protected function configure()
+    protected function configure(): void
     {
         $this->addArgument('languages', InputArgument::IS_ARRAY, 'The languages to download', $this->defaultLanguages)
             ->addOption(
--- a/src/Sulu/Bundle/AdminBundle/DependencyInjection/Compiler/AddAdminPass.php
+++ b/src/Sulu/Bundle/AdminBundle/DependencyInjection/Compiler/AddAdminPass.php
@@ -22,7 +22,7 @@ class AddAdminPass implements CompilerPassInterface
     public const ADMIN_POOL_DEFINITION_ID = 'sulu_admin.admin_pool';
     public const ADMIN_TAG = 'sulu.admin';
 
-    public function process(ContainerBuilder $container)
+    public function process(ContainerBuilder $container): void
     {
         if (!$container->hasDefinition(self::ADMIN_POOL_DEFINITION_ID)) {
             return;
--- a/src/Sulu/Bundle/AdminBundle/DependencyInjection/Compiler/AddMetadataProviderPass.php
+++ b/src/Sulu/Bundle/AdminBundle/DependencyInjection/Compiler/AddMetadataProviderPass.php
@@ -19,7 +19,7 @@ class AddMetadataProviderPass implements CompilerPassInterface
 {
     public const TAG = 'sulu_admin.metadata_provider';
 
-    public function process(ContainerBuilder $container)
+    public function process(ContainerBuilder $container): void
     {
         $pool = $container->getDefinition('sulu_admin.metadata_provider_registry');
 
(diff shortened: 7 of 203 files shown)


mamazu:
Welcome to the Sulu project and thank you for your contribution. Sulu does still have a lot of deprecation but in order to preserve backwards compatibility with the existing systems, adding type hints causes some issues. 

This pull request would probably be better targeted to 3.0 as we are doing massive code cleanups and BC breaks there. There you can also exclude the changes around the `DocumentManger` as this will be removed together with the phpCR integration.

Furthermore, some of the changes you proposed are already part of another PR like the `tagged` -> `tagged_iterator` service type (see here: #7721)

alexander-schranz:
@opctim Thanks for your effort. The only way to get the deprecations away on Symfony deprecation listener is to add a `@return` annotation via `phpdoc` to it where it say it is missing. Adding return types is a backwards compatibility break which we can not do in 2.6 version. See also @mamazu comment.

opctim:
@alexander-schranz no problem, thank you for the great CMS!

I might be missing something here, but how can adding return types be a BC if your minimum requirement for sulu 2.6 is PHP 8.2, as stated in your docs?

https://docs.sulu.io/en/2.6/upgrades/upgrade-2.x.html

alexander-schranz:
@opctim the most simpliest code is this one:

```php
<?php

class SuluClass {
    public function test(): void // added return type
    {}
}


class AppClass extends SuluClass {
    public function test() // existing extended class in projects or libs
    {}
}
```

Adding a return type will crash projects, libraries and bundles extending existing classes:

> Fatal error: Declaration of AppClass::test() must be compatible with SuluClass::test(): void

Only were previously already marked as `final` can safely add return types.

alexander-schranz:
For completness: In most cases when you are not tackle directly a Symfony upgrade or something like that. I suggest disabling the Deprecation logger like we already do in our default setup: https://github.com/sulu/skeleton/blob/d616fe6b8f8766068a9ae5af84468aa6cd32f5e9/config/packages/monolog.yaml#L58-L61

opctim:
Also, that would support symfony's initiative to add return types, as you support symfony up to 7.4:
https://symfony.com/blog/symfony-7-0-type-declarations

Additionally, function return types have been added in PHP 7.4, so I don't see where this would break compatibility.

An idea would be to revert the changes I made in [33e0e7f](https://github.com/sulu/sulu/pull/7776/commits/33e0e7f33736e2dfc64408799b649fe72d76dfb2) and [5b43631](https://github.com/sulu/sulu/pull/7776/commits/5b43631fcd1477d1717ec95eaa13624ebfdb7e1b).

Then, my changes would only cover symfony-internal type requirements, that would satisfy your demonstrated BC:

- EventSubscriber -> getSubscribedEvents(): array 
- Adding :void to Command configure()
- Adding :int to count() in the MediaRepository
- etc.

@alexander-schranz please take your time and have a look at the commits & what they cover. I am pretty sure this won't break existing implementations.

alexander-schranz:
> Also, that would support symfony's initiative to add return types, as you support symfony up to 7.4:
https://symfony.com/blog/symfony-7-0-type-declarations
> Additionally, function return types have been added in PHP 7.4, so I don't see where this would break compatibility.

Its still a break as shown in a example if we add `Native Return Types`. If we add a return type it will projects, libraries which did extend maybe existing classes. This is also why Symfony could only do that in Symfony 7.0 and not on 6.4 or any minor or patch release. The `vendor/bin/patch-type-declarations` is so nice for projects but not for Libraries which need to keep Backwards compatibility in mind.

But Symfony added a way to tell that you are aware of that deprecation by adding the PHPDoc:

```diff
+    /**
+     * @return void
+     */
     protected function configure()
```

This is no bc break and makes the deprecation listener happy by tell you are aware of that deprecation. Best would be have seperate Pull request for the things which are thrown as deprecation. Like tackling bundle classes by its own, Subscribers by its own and so on. Thinkgs mention by @mamazu don't need to be tackled as lot of that classes will be removed.

opctim:
@alexander-schranz I get what you mean and I'd go with that 100%, but I think you are missing something:

CompilerPasses & EventSubscriber:getSubscribedEvents() for example are not meant to be extended

alexander-schranz:
@opctim I know what you meant. But we missed to mark such classes and/or methods as final or internal and for BC reasons we should so keep it that way, as people overwrite many services which we might not know, so currently it is still possible todo that or extend such classes.

@mamazu is doing lot of work on 3.0, to mark more and more classes final or internal to keep such things easier in future and be more type strict. For 2.6 we only can go with the phpdocs. Which already helps to correctly native type it after it in 3.0.

opctim:
@alexander-schranz Ok, so as this seems like we won't be integrating my changes into 2.6, what are we going to do with this PR? 

Should we scrap it altogether or will you adopt my changes @mamazu ?

alexander-schranz:
@opctim think best would be have a direct look at which deprecations are thrown and add `/** @return ... */` annotation were required, to avoid the deprecation message. If you want to tackle that it would be great.

opctim:
@alexander-schranz I was talking about using my changes in 3.0

Adding PHPdoc to 2.6 means starting from scratch.

alexander-schranz:
@opctim for 3.0 make sense. But better to split it up into few small pull requests. e.g. Subscriber, Bundle/Extension, ... can you cherry pick the different commits out and rebase them to 3.0? And open own pr for each one?

All related to PHPCR, DocumentManager, Jackalope, Structures, can be ignored as that classes will be removed before the 3.0 release.

mamazu:
What I would suggest to be the next steps with this merge request:

> [Adds missing return types on Subscribers] [34f3280]

This is a great change and should be it's **own merge request targeting Sulu 3.0** with a note in the CHANGELOG.md that these methods now require a return types.

> [Adds missing return types on Commands] [2bd3146]

This is a great change and should be it's **own merge request targeting Sulu 3.0** with a note in the CHANGELOG.md that these methods now require a return types.

> [Fixes deprecation by explicitly setting doctrine.orm.enable_lazy_ghost_object] [03bdf13]

Same here. However I think this could also be a thing that every project has to try out for themselves if there are things breaking. I don't know if doing that in 3.0 is a good idea.

> [Adds missing return types on CompilerPasses] [d9dcede]

I'd wait with this one. We're trying to make the CompilerPasses final so that those changes will be easier. We're removing compiler passes and replacing them with `tagged_iterators` see (https://github.com/sulu/sulu/issues/7402)

> [Renames deprecated argument type "tagged" to "tagged_iterator"] [5ed8b8d]

That is probably a duplicate of my work and can be removed (see https://github.com/sulu/sulu/pull/7721)

> [Adds missing return types on Subscribers] [fa947f8]

This one could be merged with the first commit.

> [Fixes deprecations in kernel extensions] [6595f8a]

As @alexander-schranz said, I'm working on finalizing classes in there to lock down what people can do with that. When that's done this change also would be great in a seperate merge request targeting 3.0.

> [Adds missing return type in MediaRepository] [d9710e0]

This is a change that we could use in 2.6 but as phpdoc. This should already satisfy the static analyser and phpstorm to give you the correct type hints when you use it. In the future it's planned to do a release where we replace all phpdocs comments with real php type hints but that's something for maybe 4.0.

> [Adds missing return type DocumentManagerTest] [382b331]

This class will be removed in 3.0 so you might as well discard this part.

> [Adds missing return types to serializer handlers] [329ea39]
Geat change. Seperate merge request targeting 3.0 and an `UPGRADE.md` note.

> [Adds missing return types to DocumentManagerBundle\Session] [33e0e7f] 

This class will be removed in 3.0 so you might as well discard this part.

> [Adds missing return types to ExtensionContainer & ManagedExtensionCon…] [5b43631]

Same as above, I'd wait for the finalization of the extension classes for that, then separate PR targeting 3.0.

alexander-schranz:
@opctim Thank you, I cherry picked a punch of your changes to 3.0 ( https://github.com/sulu/sulu/pull/8302 ). As said for 2.6 we can not do such bc breaks, there we can only add `@return` annotations to silence the Symfony return type validator. Closing this one as your commits merged to 3.0.

PR #7777, opened 2025-02-03
Marking all Sulu `Extension`s as `final` classes.

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | yes
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?
Marking all Extension classes as `final` and their methods as `@internal`.

#### Why?
They should not be extended and only used by Symfony. We also don't want to mark them as `@internal` because the user still needs to reference them in the `bundles.php` file.

An example on how to properly done can be found in the ArticleBundle: https://github.com/sulu/sulu/blob/3.0/packages/article/src/Infrastructure/Symfony/HttpKernel/SuluArticleBundle.php#L56

### DIFF
--- a/src/Sulu/Bundle/ActivityBundle/SuluActivityBundle.php
+++ b/src/Sulu/Bundle/ActivityBundle/SuluActivityBundle.php
@@ -18,10 +18,13 @@ use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Extension\ExtensionInterface;
 use Symfony\Component\HttpKernel\Bundle\Bundle;
 
-class SuluActivityBundle extends Bundle
+final class SuluActivityBundle extends Bundle
 {
     use PersistenceBundleTrait;
 
+    /**
+     * @internal
+     */
     public function build(ContainerBuilder $container): void
     {
         $this->buildPersistence(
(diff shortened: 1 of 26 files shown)


src/Sulu/Bundle/ActivityBundle/SuluActivityBundle.php
```diff
@@ -18,10 +18,13 @@
 use Symfony\Component\DependencyInjection\Extension\ExtensionInterface;
 use Symfony\Component\HttpKernel\Bundle\Bundle;
 
-class SuluActivityBundle extends Bundle
+final class SuluActivityBundle extends Bundle
 {
     use PersistenceBundleTrait;
 
+    /**
+     * @internal
```
alexander-schranz: we should add the same comment as here to be more transparent: https://github.com/sulu/sulu/blob/c785900f346166d01304fbd6023ee4a106128550/packages/article/src/Infrastructure/Symfony/HttpKernel/SuluArticleBundle.php#L91

mamazu:
Done.

alexander-schranz (approved):

PR #7783, opened 2025-02-04
Fix some new phpstan issues in new packages

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no 
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT
| Documentation PR | sulu/sulu-docs# 

#### What's in this PR?

Fix some bew phpstan issues in new packages.

#### Why?

After upgrade to PHPStan 2.0 we got some new issues on the 3.0 branch we tackle them here before doing a baseline.

PR #7784, opened 2025-02-04
Fix deprecated null argument on ReflectionClass instantiation

stollr:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? | no 
| BC breaks? | no 
| Deprecations? | no 
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

This PR adds a check to the `PhpcrAccessControlProvider::supports` method which prevents from a deprecated call.

#### Why?

When Sulu's AccessControlManager tries to get the permissions for a SecurityCondition which does not have an objectType, it calls the PhpcrAccessControlProvider::supports method with a null value. This is in turn passed as argument to the ReflectionClass contructor resulting in a deprecation.

### DIFF
--- a/src/Sulu/Component/Security/Authorization/AccessControl/PhpcrAccessControlProvider.php
+++ b/src/Sulu/Component/Security/Authorization/AccessControl/PhpcrAccessControlProvider.php
@@ -73,6 +73,10 @@ class PhpcrAccessControlProvider implements AccessControlProviderInterface
 
     public function supports($type)
     {
+        if (null === $type) {
+            return false;
+        }
+
         try {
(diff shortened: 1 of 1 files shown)


alexander-schranz:
@stollr Thx for your effort. Actually the method should be never be called with `null`. Please check the changes in ~~https://github.com/sulu/sulu/pull/7783~~ (UPDATE: https://github.com/sulu/sulu/pull/7762) which should not longer call the supports method without a string value.

stollr:
Oh, I did not see that. I guess the changes have not been merged into 2.6 branch?

Issue #7792, opened 2025-02-05
Replace DateTime with DateTimeImmutable where possible


alexander-schranz:
# Problem description

Currently we are using in a lot of places still `DateTime` object. DateTime object is mutable. Which means I can call `$someObject->getCreated()->modify('+1 day')` to modify maybe accidently the created or changed date of a entity.

# Proposed solution

For safer code and avoid accidently changing datetimes from outside we should where possible migrate the Entities to `DateTimeImmutable`.

```xml
<field name="created" type="datetime" />
<field name="created" type="datetime_immutable" />
```

The `TimestampableInterface` and `TimestampableTrait` should also be replaced and its metadatalistener. Maybe we should check over the whole code base where it make sense.

Specific functions which require currently a DateTime may should use the DateTimeInterface like we adopted already here: https://github.com/sulu/sulu/pull/7702

PR #7795, opened 2025-02-06
Immutable datetimes on timestampable

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | yes
| Deprecations? | -
| Fixed tickets | -
| Related issues/PRs | #7792 
| License | MIT
| Documentation PR | /

#### What's in this PR?
Replacing the changed and updated fields to contain immutable datetime objects (see issue).

The interface is now strictly types to only accept `DatetimeImmutable` as input and return `DateTimeInterface` as output.

#### Why?
Technically the created at and updated add should only ever represent "now" and should never be modified.


UPGRADE.md
```diff
@@ -2,6 +2,16 @@
...
 
+### Using DateTimeImmutable for database columns
+
+The following columns now contain immutable datetime objects:
+
+All `created` and `updated` columns generated by the `TimestampableInterface`.
+
+> [!WARNING]  
+> You need to generate a migration for your own project, since this also effects custom entities in your project. For
+the Sulu base see migrations below.
```
alexander-schranz: we should add example SQL here showing what is changed in the core.

mamazu:
Looks like there are a lot of places that use the `TimestampableInterface` but not the trait. And with Api objects this will not be easily possbile.

alexander-schranz:
@mamazu which API entity use the Interface? Most shouldn't as they should be API specific and not extend that interface. The Entities itself should be able to remove current method in the interface as the timestampableinterface already defines the method signature.

mamazu:
The problem is not the interface its the usage of the trait. Like for example the `src/Sulu/Bundle/TagBundle/Entity/Tag.php` class does not use the `AuditableTrait` although the interface extends the `AuditableInterface` and it can't use the trait because the properties contain meta information in the form of attributes which the trait doesn't have.

wachterjohannes:
@alexander-schranz i have rebased the branch, changed the Interface to use \DateTimeImmutable and added a section to the UPGRADE

mamazu:
This (#7797) would be a pre-requite for this branch anyways.

src/Sulu/Bundle/CategoryBundle/Entity/CategoryInterface.php
```diff
@@ -222,35 +221,7 @@ public function setParent(?self $parent = null);
...
-    public function setCreator(?UserInterface $creator = null);
-
-    /**
-     * Set created.
-     * Note: This property is set automatically by the TimestampableSubscriber if not set manually.
-     *
-     * @return CategoryInterface
-     */
-    public function setCreated(\DateTime $created);
+    public function setChanged(\DateTimeImmutable $changed): self;
```
mamazu: Should this not be part of the AuditibleInterface?

wachterjohannes: @mamazu thanks and fixed

src/Sulu/Bundle/ContactBundle/Entity/Contact.php
```diff
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
```
alexander-schranz: do not declare this

src/Sulu/Bundle/MediaBundle/Entity/Collection.php
```diff
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
```
alexander-schranz: same not declare strict types aslong as the code style doesnt apply it

alexander-schranz (approved):

UPGRADE-3.x.md
```diff
@@ -647,3 +646,34 @@ In case of some errors on customized code, you can try to fix it and rerun the c
...
+
+### Using DateTimeImmutable for database columns
+
+The following columns now contain immutable datetime objects:
+
+All `created` and `changed` columns generated by the `TimestampableInterface`.
+
+> [!WARNING]  
+> You need to generate a migration for your own project, since this also effects custom entities in your project. For
+the Sulu base see migrations below.
```
mamazu, line 658: @wachterjohannes This doesn't only affect the migration in the database. This also means that people need to modify the trash handler for the entity. Maybe we could also handle this in a more general fashion.

PR #7796, opened 2025-02-06
Fix/media db deletion when physically missing

### LINKED ISSUE #7064: Media preview image deletion issue
| Q |
| --- | ---
| Sulu Version | 2.5.9
| Symfony Version | 6.2.9
| PHP Version | 8.2
| DB Version | MySQL 8.0.31

# Actual Behavior

I’ve got an issue with the preview image feature on the media. First, not an issue but a comment : when adding a preview image on a .odt document for instance, the preview Media is created with the same name and nothing to distinguish it from the odt Media in the Media list page.
This led me to the issue when I tried to delete the preview image instead of the original document, I've got this error :

Capture d’écran 2023-04-28 à 09 43 43

This one is ok, but while there is this exception, the media file is deleted from the filesystem !
So after that, of course, if you try to delete the preview image again :

Capture d’écran 2023-04-28 à 09 51 14

Deleting the original Media still works as it doesn’t cascade delete the preview image (that could be an issue too, or a feature... :clin_d'œil: ).

After that it becomes impossible to delete this Media from the BO.

steeven-th:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #7064
| Related issues/PRs |
| License | MIT

#### What's in this PR?

This PR fixes an issue where media deletion in the database would fail when the associated media file no longer existed physically. The change ensures that when a media file is missing from the filesystem, it can still be safely removed from the database without errors. Additionally, it logs a message in the activity to indicate that the media was deleted, but could not be moved to the Trash due to its absence from the filesystem.

#### Why?

The issue arises when attempting to delete a media entry from the database while its physical file is missing. Without this fix, the deletion process fails, potentially leaving orphaned entries in the database. This PR resolves that issue, ensures a smoother media management process, and provides clearer feedback through activity logging about why the media wasn’t moved to Trash.

sulu-delete-media-no-trash


steeven-th:
Closing this PR in favor of #8663, which was merged and provides the cleaned-up version of the same fix (handling both LocalStorage IOException and FlysystemStorage FileNotFoundException when the physical file is missing). Thanks!

PR #7805, opened 2025-02-12
Add error output to the phpcr cleanup command

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | yes
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?
* Adding error output to the `sulu:document:phpcr-cleanup -v` command so that the user can see what went wrong.
* Replacing the Confirmation question with a Symfony Confirmation question.
image
* Writing an error output of the single node clean command (right now it's just being caught and ignored)

#### Why?
* Depending on the kind of error maybe re-running the command might help. Or further investigations need to be done.
* Changing the default of the "do you want to do this" to yes:
    * makes it usable in scripts
    * you can now use the `-n` flag to skip the prompt
    * if you answer no, why did you call the command in the first place?

#### Example Usage

````
bin/console sulu:document:phpcr-cleanup
```
Same output as before. Just a list of ids.


````
bin/console sulu:document:phpcr-cleanup -v
```
Returns a list of ids and their error output. Maybe this should be redirected to a file then.

### DIFF
--- a/src/Sulu/Bundle/DocumentManagerBundle/Command/PHPCRCleanupCommand.php
+++ b/src/Sulu/Bundle/DocumentManagerBundle/Command/PHPCRCleanupCommand.php
@@ -183,6 +183,11 @@ class PHPCRCleanupCommand extends Command
                 if (0 !== $status) {
                     ++$stats['erroredNodes'];
                     $erroredUuids[$uuid] = $process->getErrorOutput();
+                    $this->logger->writeln(\sprintf(
+                        "# Error processing node '%s'\n\n%s\n",
+                        $uuid,
+                        $process->getErrorOutput(),
+                    ));
 
                     continue;
                 }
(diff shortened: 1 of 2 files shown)


src/Sulu/Bundle/DocumentManagerBundle/Command/PHPCRCleanupCommand.php
```diff
@@ -192,20 +182,20 @@ protected function execute(InputInterface $input, OutputInterface $output): int
                 }
                 if (0 !== $status) {
                     ++$stats['erroredNodes'];
-                    $erroredUuids[] = $uuid;
+                    $erroredUuids[$uuid] = $process->getErrorOutput();
```
wachterjohannes, line 185: @mamazu we should also write this to the debug markdown file! then it also would be possible to have that for lots of errored nodes

mamazu: Error output should now be also logged.

wachterjohannes (approved):

PR #7806, opened 2025-02-12
Requiring correct http methods on `AdminController`

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | technically yes
| Deprecations? | -
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?
The `metadata` and the `translations` endpoints never accept any data so calling them with anything other than a get would not make sense. Let's also enforce that.

#### Why?
This makes it clear to the caller of the API that this endpoint only returns data.

In the long run those should be moved to their own controller classes.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/config/routing.yml
+++ b/src/Sulu/Bundle/AdminBundle/Resources/config/routing.yml
@@ -10,13 +10,13 @@ sulu_admin.config:
 
 sulu_admin.metadata:
     path: /metadata/{type}/{key}
-    methods: 'GET'
+    methods: ['GET', 'OPTIONS']
     defaults:
         _controller: sulu_admin.admin_controller::metadataAction
 
 sulu_admin.translation:
     path: /translations
-    methods: 'GET'
+    methods: ['GET', 'OPTIONS']
     defaults:
         _controller: sulu_admin.admin_controller::translationsAction
 
--- a/src/Sulu/Bundle/AdminBundle/Resources/config/routing.yml
+++ b/src/Sulu/Bundle/AdminBundle/Resources/config/routing.yml
@@ -10,11 +10,13 @@ sulu_admin.config:
 
 sulu_admin.metadata:
     path: /metadata/{type}/{key}
+    methods: 'GET'
     defaults:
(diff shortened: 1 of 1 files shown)


alexander-schranz:
@mamazu do we need to keep `HEAD` or `OPTIONS` request in mind 🤔

mamazu:
I don't think the controller could handle them.

alexander-schranz:
@mamazu nothing about the handling its more about HEAD, OPTION request should behave the same like currently, they should not return a 406 or other error code. Not sure if GET includes HEAD and OPTION requests.

mamazu:
@alexander-schranz I just checked it out, the `HEAD` request still works but the option request doesn't. However if I revert it back to "ANY" the `OPTIONS` request returns the the same contents as the `GET` which I don't think is how it's supposed to work.

mamazu:
Did some research. The option request should just return the header "Allow: OPTIONS, GET, ..." which this controller doesn't do. So sending an option request to this endpoint would be nonsense anyways.

alexander-schranz:
As said its not about handling options request. More how this change could effect existing applications who maybe doing HEAD or OPTIONS request to our APIs.

mamazu:
Added `OPTION` as an allowed method again. `HEAD` requests work anyways (maybe some Symfony default handling)

alexander-schranz (approved):

src/Sulu/Bundle/AdminBundle/Controller/AdminController.php
```diff
@@ -173,6 +173,11 @@ public function configAction(): Response
 
     public function translationsAction(Request $request): Response
     {
+        if ($request->getMethod() == Request::METHOD_OPTIONS) {
+            $response = new Response();
+            $response->headers->set('Allow', 'GET, OPTIONS, HEAD');
+            return $response;
+        }
```
alexander-schranz:
```suggestion
```

Would not answer options request currently think that would be better handled in own bundle, via request listener.

mamazu:
Oops, that was code where I was just playing around. Removed it.

PR #7804, opened 2025-02-12
Add PageBundle with new Content Storage

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no yes
| New feature? | no yes
| BC breaks? | no yes
| Deprecations? | no yes 
| Fixed tickets | fixes # 
| Related issues/PRs | https://github.com/sulu/sulu/issues/7672
| License | MIT
| Documentation PR | sulu/sulu-docs# 

#### To Do

- [ ] Create a documentation PR
- [ ] Add breaking changes to UPGRADE.md

PR #7813, opened 2025-02-17
Add OrderPage handler

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Issues | part of https://github.com/sulu/sulu/issues/7672
| License | MIT

Depends on https://github.com/sulu/sulu/pull/7804

#### What's in this PR?

Adds the page order functionality to the Sulu 3.0 Page bundle

### DIFF
--- a/packages/page/src/Application/MessageHandler/OrderPageMessageHandler.php
+++ b/packages/page/src/Application/MessageHandler/OrderPageMessageHandler.php
@@ -29,27 +29,6 @@ final class OrderPageMessageHandler
 
     public function __invoke(OrderPageMessage $message): void
     {
-        $page = $this->pageRepository->getOneBy($message->getIdentifier());
-
-        $parent = $page->getParent();
-        $siblings = $this->pageRepository->getChildren($parent);
-        $currentPosition = \array_search($page, $siblings);
-
-        if (false === $currentPosition) {
-            throw new \RuntimeException('Node not found in sibling list');
-        }
-
-        $currentPosition = \intval($currentPosition);
-        $movementSteps = $currentPosition - \max(0, $message->getPosition() - 1);
-
-        if ($movementSteps > 0) {
-            $this->pageRepository->moveUp($page, $movementSteps);
-        } elseif ($movementSteps < 0) {
-            $this->pageRepository->moveDown($page, \abs($movementSteps));
-        }
-
-        if (true !== $this->pageRepository->verify()) {
-            $this->pageRepository->recover();
-        }
+        $this->pageRepository->reorderOne($message->getIdentifier(), $message->getPosition());
     }
 }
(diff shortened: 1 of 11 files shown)


packages/page/src/Domain/Repository/PageRepositoryInterface.php
```diff
@@ -145,4 +145,22 @@ public function countBy(array $filters = []): int;
...
+    public function moveUp(PageInterface $node, int $number): bool;
+
+    public function moveDown(PageInterface $node, int $number): bool;
+
+    /**
+     * @return bool|string[]
+     */
+    public function verify(): bool|array;
+
+    public function recover(): void;
```
alexander-schranz: We should not add this methods to our repository interface. We should add a single own reorder method with a signature matching our case.

If somebody requires the underlaying Nested Tree methods they can inject doctrine repository instead of our domain repository.

packages/page/src/Application/MessageHandler/OrderPageMessageHandler.php
```diff
@@ -0,0 +1,55 @@
...
+        }
+
+        $currentPosition = \intval($currentPosition);
+        $movementSteps = $currentPosition - \max(0, $message->getPosition() - 1);
+
+        if ($movementSteps > 0) {
+            $this->pageRepository->moveUp($page, $movementSteps);
+        } elseif ($movementSteps < 0) {
+            $this->pageRepository->moveDown($page, \abs($movementSteps));
+        }
```
alexander-schranz, line 49: this should be done inside our own reorder Repository method.

packages/page/src/Application/MessageHandler/OrderPageMessageHandler.php
```diff
@@ -0,0 +1,55 @@
...
+
+        if (true !== $this->pageRepository->verify()) {
+            $this->pageRepository->recover();
+        }
```
alexander-schranz: This should be done inside our own reorder Repository method.

packages/page/src/Domain/Repository/PageRepositoryInterface.php
mamazu: Maybe we could also split this up into multiple interfaces.

alexander-schranz (changes requested):
as discussed we should move this logic to the repository something like `reorderOneBy(array $filters, int $position)` to match our Domain.

alexander-schranz (approved):

packages/page/src/Domain/Repository/PageRepositoryInterface.php
alexander-schranz: our domain should stay by its own so we avoid global usage in the domain layer and interface inheritance. this make it easier to see with one check what the domain is what is available in the domain of that bundle.

PR #7817, opened 2025-02-21
Make the property resolver provider final and internal provide an interface for extendability

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Make the property resolver provider final, add a interface for extendability.

#### Why?

This service should not be extended only decorated, also we are free to change how the constructor is created as making it internal as so we are free to choose if tagged_iterator or tagged_locator is the best choice for us.

Overriding and changing the behaviour can still be done via decorating the service: https://symfony.com/doc/6.4/service_container/service_decoration.html

The new interface for decorating can also now be used then for creating `mocks`.

### DIFF
--- a/packages/content/src/Application/PropertyResolver/PropertyResolverProvider.php
+++ b/packages/content/src/Application/PropertyResolver/PropertyResolverProvider.php
@@ -16,7 +16,7 @@ namespace Sulu\Content\Application\PropertyResolver;
 use Sulu\Content\Application\PropertyResolver\Resolver\PropertyResolverInterface;
 
 /**
- * @internal The constructor of this class maybe change in future releases to add new features or improve performance.
+ * @internal The constructor of this class may change in future releases to add new features or improve performance.
  *           Use this service via the dependency injection container via the PropertyResolverProviderInterface only.
  *           It is still fine to use this class to create a new instance for mock-less unit tests,
  *           but no backwards compatibility promise can be given for the constructor.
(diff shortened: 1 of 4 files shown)


packages/content/src/Application/PropertyResolver/PropertyResolverProvider.php
```diff
@@ -15,7 +15,13 @@
 
 use Sulu\Content\Application\PropertyResolver\Resolver\PropertyResolverInterface;
 
-class PropertyResolverProvider
+/**
+ * @internal The constructor of this class maybe change in future releases to add new features or improve performance.
```
Prokyonn:
```suggestion
 * @internal The constructor of this class may change in future releases to add new features or improve performance.
```

packages/content/src/Application/PropertyResolver/PropertyResolverProvider.php
```diff
@@ -15,7 +15,13 @@
...
+ *           Use this service via the dependency injection container via the PropertyResolverProviderInterface only.
+ *
+ * If you need to override this service use service decoration via the provided interface: https://symfony.com/doc/6.4/service_container/service_decoration.html.
```
Prokyonn, line 24:
```suggestion
 * If you need to override this service use service decoration via the provided interface: https://symfony.com/doc/current/service_container/service_decoration.html.
```

I think we can use `current` as a version in the url

alexander-schranz: We should not use `current` as that can end up in Dead links or none supported version so I prefer current LTS (6.4) which is the documentation keep longest uptodate.

Prokyonn (approved):

PR #7819, opened 2025-02-24
Replace Sulu 3.0 packages PHPUnit Annotations with Attributes

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | #6988 
| License | MIT
| Documentation PR | -

#### What's in this PR?
Adding a rector rule that enforces the use of Attributes over Annotations.
Also adding the "packages" directory to the rector files.

#### Why?
Since the introduction of PhpUnit 10 we can use Attributes for everything. They're better supported by PHP than annotations.

#### Example Usage
See changes.

### DIFF
--- a/packages/article/tests/Functional/Integration/ArticleControllerTest.php
+++ b/packages/article/tests/Functional/Integration/ArticleControllerTest.php
@@ -19,9 +19,8 @@ use Symfony\Bundle\FrameworkBundle\KernelBrowser;
 
 /**
  * The integration test should have no impact on the coverage so we set it to coversNothing.
- *
- * @coversNothing
  */
+#[\PHPUnit\Framework\Attributes\CoversNothing]
 class ArticleControllerTest extends SuluTestCase
 {
     use AssertSnapshotTrait;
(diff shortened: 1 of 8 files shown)


packages/article/tests/Functional/Integration/ArticleControllerTest.php
```diff
@@ -19,9 +19,8 @@
 
 /**
  * The integration test should have no impact on the coverage so we set it to coversNothing.
- *
- * @coversNothing
  */
+#[\PHPUnit\Framework\Attributes\CoversNothing]
```
mamazu, line 23: Technically this does cover the controller, just not everything else.

rector.php
```diff
@@ -44,6 +46,7 @@
         // SymfonySetList::SYMFONY_CODE_QUALITY,
         // DoctrineSetList::DOCTRINE_CODE_QUALITY,
         // LevelSetList::UP_TO_PHP_80,
+        PHPUnitSetList::ANNOTATIONS_TO_ATTRIBUTES,
```
alexander-schranz, line 49: is PHPUnit 10 used on 2.6? then we probably should enable this also there?

alexander-schranz (approved):

PR #7822, opened 2025-02-24
Enforce phpunit attributes with rector rule

mamazu:
#7819 but for version 2.6


rector.php
```diff
@@ -41,7 +42,8 @@
         // SymfonySetList::SYMFONY_CODE_QUALITY,
         // DoctrineSetList::DOCTRINE_CODE_QUALITY,
         // LevelSetList::UP_TO_PHP_80,
+        PHPUnitSetList::ANNOTATIONS_TO_ATTRIBUTES,
     ])
     ->withRules([
-        StaticDataProviderClassMethodRector::class, // prepare for PHPUnit >= 10
+        StaticDataProviderClassMethodRector::class,
```
mamazu, line 48: We're using phpunit 10+ on this branch so not having this rule could introduce broken code.

PR #7826, opened 2025-02-26
Add global blocks to Sulu 3.0

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Issues | https://github.com/sulu/sulu/issues/7672
| License | MIT

#### What's in this PR?

Adds global blocks support for the new content structure for Sulu 3.0

### DIFF
--- a/packages/content/src/Application/PropertyResolver/Resolver/BlockPropertyResolver.php
+++ b/packages/content/src/Application/PropertyResolver/Resolver/BlockPropertyResolver.php
@@ -16,6 +16,8 @@ namespace Sulu\Content\Application\PropertyResolver\Resolver;
 use Psr\Log\LoggerInterface;
 use Sulu\Bundle\AdminBundle\Metadata\FormMetadata\FieldMetadata;
 use Sulu\Bundle\AdminBundle\Metadata\FormMetadata\FormMetadata;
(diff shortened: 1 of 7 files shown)


packages/content/src/Application/PropertyResolver/Resolver/BlockPropertyResolver.php
```diff
@@ -53,14 +56,17 @@ public function resolve(mixed $data, string $locale, array $params = []): Conten
...
+        $globalBlocksMetadata = $typedFormMetadata->getForms();
         $contentViews = [];
         foreach ($data as $key => $block) {
             if (!\is_array($block) || !isset($block['type']) || !\is_string($block['type'])) {
                 continue;
             }
 
             $type = $block['type'];
-            $formMetadata = $metadataTypes[$type] ?? null;
+            $formMetadata = $globalBlocksMetadata[$type] ?? $metadataTypes[$type] ?? null;
```
alexander-schranz: we prefer always the global block here now but what if a local block type was defined without a `ref` to a global block?

alexander-schranz (approved):

PR #7828, opened 2025-02-28
Fix Sitemap generation with multiple webspaces

### LINKED ISSUE #7756: Sitemap bug with multiple webspaces
| Q | A
| --- | ---
| Sulu Version | 2.6.6
| PHP Version | 8.3

# Actual Behavior

In case you have more than one webspace with a different host, the SitemapProviderPool will always return indices from the first loop result.

For example, suppose you have two webspaces:

- **A** (host: a.com) containing published pages and articles
- **B** (host: b.com) containing only published pages
When running the CLI command sulu:website:dump-sitemap, XML sitemap files will be generated for each webspace. The process correctly creates:

For **webspace A**: pages-1.xml, articles-1.xml
For **webspace B**: pages-1.xml
However, the issue occurs with the main sitemap.xml file. It is always duplicated from the first webspace in the loop.

For instance, if webspace A is processed first, then **webspace B**’s sitemap.xml **will incorrectly be copied from A**, but with the correct host (b.com). As a result, B’s sitemap.xml will reference b.com/pages-1.xml and b.com/articles-1.xml, even though articles-1.xml does not exist for webspace B.

# Expected Behavior

**Webspace A** (a.com): sitemap.xml should reference pages-1.xml and articles-1.xml.
**Webspace B** (b.com): sitemap.xml should reference only pages-1.xml.

# Steps to Reproduce

1. Define multiple webspaces with different hosts.

2. Ensure each webspace has different published entity types (e.g., one with pages and articles, another with only pages).

3. Run the command:

```php bin/console sulu:website:dump-sitemap```

4. Check the generated sitemap files for each webspace.

5. Observe that the sitemap.xml file for the second webspace is incorrect.

# Possible Solutions

In the https://github.com/sulu/sulu/blob/2.6/src/Sulu/Bundle/WebsiteBundle/Sitemap/SitemapProviderPool.php remove property `$index` and any related logic.

draconivis:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #7756
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

Fix the generation of the sitemaps in a multi-webspace setup by discerning between the indexes with a unique key

#### Why?

See #7756

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -31781,11 +31781,6 @@ parameters:
 			path: src/Sulu/Bundle/WebsiteBundle/Routing/RequestListener.php
 
 		-
-			message: "#^Property Sulu\\\\Bundle\\\\WebsiteBundle\\\\Sitemap\\\\Sitemap\\:\\:\\$lastmod \\(DateTime\\) does not accept DateTime\\|null\\.$#"
-			count: 1
-			path: src/Sulu/Bundle/WebsiteBundle/Sitemap/Sitemap.php
-
-		-
 			message: "#^Method Sulu\\\\Bundle\\\\WebsiteBundle\\\\Sitemap\\\\SitemapContentQueryBuilder\\:\\:buildSelect\\(\\) has no return type specified\\.$#"
 			count: 1
 			path: src/Sulu/Bundle/WebsiteBundle/Sitemap/SitemapContentQueryBuilder.php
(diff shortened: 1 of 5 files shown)


alexander-schranz:
pushed adjustment of the tests to avoid that we break this again. Also added the missing `kernel.reset`.

PR #7830, opened 2025-03-03
Add creating of history routes automatically for new route storage

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? | yes 
| BC breaks? | no 
| Deprecations? | no  
| Fixed tickets | fixes # 
| Related issues/PRs | ﻿https://github.com/sulu/sulu/pull/7726 part of https://github.com/sulu/sulu/issues/7672
| License | MIT

#### What's in this PR?

Add creating of history routes automatically for new route storage.

#### Why?

We should create history routes automatically.

### DIFF
--- a/packages/route/src/Domain/Model/Route.php
+++ b/packages/route/src/Domain/Model/Route.php
@@ -17,7 +17,7 @@ namespace Sulu\Route\Domain\Model;
 class Route
 {
     /** @internal */
-    public const HISTORY_RESOURCE_KEY = 'history';
+    public const HISTORY_RESOURCE_KEY = 'route_history';
 
     private ?int $id = null;
 
--- a/packages/route/src/Domain/Model/Route.php
+++ b/packages/route/src/Domain/Model/Route.php
@@ -16,6 +16,9 @@ namespace Sulu\Route\Domain\Model;
  */
 class Route
 {
+    /** @internal */
+    public const HISTORY_RESOURCE_KEY = 'history';
+
     private ?int $id = null;
 
     private ?string $site;
--- a/packages/route/src/Infrastructure/Doctrine/EventListener/RouteChangedUpdater.php
+++ b/packages/route/src/Infrastructure/Doctrine/EventListener/RouteChangedUpdater.php
@@ -17,6 +17,8 @@ use Doctrine\DBAL\Platforms\PostgreSQLPlatform;
 use Doctrine\ORM\Event\OnClearEventArgs;
 use Doctrine\ORM\Event\PostFlushEventArgs;
 use Doctrine\ORM\Event\PreUpdateEventArgs;
+use Doctrine\ORM\Mapping\ClassMetadata;
+use Doctrine\Persistence\ObjectManager;
 use Sulu\Route\Domain\Model\Route;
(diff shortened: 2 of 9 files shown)


packages/route/src/Domain/Model/Route.php
```diff
@@ -16,6 +16,9 @@
  */
 class Route
 {
+    /** @internal */
+    public const HISTORY_RESOURCE_KEY = 'history';
```
alexander-schranz: maybe `route_history` better match in case of a resource key. Think resource keys are snake case or?

src/Sulu/Bundle/ActivityBundle/Infrastructure/Doctrine/Repository/ActivityRepository.php
```diff
@@ -102,8 +102,11 @@ protected function getInsertQueryBuilder(ActivityInterface $activity): QueryBuil
         }
 
         // set value of id explicitly if class has a pre-insert identity-generator to be compatible with postgresql
-        if (!$classMetadata->idGenerator->isPostInsertGenerator()) {
-            $queryBuilder->setValue($classMetadata->getColumnName('id'), $classMetadata->idGenerator->generate($this->entityManager, $activity));
+        if (
+            null !== $classMetadata->idGenerator
+            && !$classMetadata->idGenerator->isPostInsertGenerator()
+        ) {
+            $queryBuilder->setValue($classMetadata->getColumnName('id'), $classMetadata->idGenerator->generateId($this->entityManager, $activity));
```
alexander-schranz, line 109: updated this to use the newer generator method and as the idGenerator can be null to check for that also. see also phpstan baseline update.

packages/route/src/Domain/Model/Route.php
Prokyonn: Yes, snake_case is used for resource keys.

I agree and would also go with something like `route_history`. It is not only more self-explanatory, but we also minimize the chance of colliding with resource keys of other custom entities.

packages/route/src/Infrastructure/Doctrine/EventListener/RouteChangedUpdater.php
```diff
@@ -14,19 +14,25 @@
...
+use Doctrine\ORM\Mapping\ClassMetadata;
 use Sulu\Route\Domain\Model\Route;
 use Symfony\Contracts\Service\ResetInterface;
 
 /**
  * @internal No BC promises are given for this class. It may be changed or removed at any time.
+ *
+ * This is kind of a complex mechanic here and "Test Driven Development" is recommended way to do any changes here.
+ * Reporting a bug inside the following logic please provide a failing test case. Easiest way in most use cases is
+ * to adopt the existing `RouteChangedUpdaterTest::provideRoutes` test data provider.
```
Prokyonn:
```suggestion
 * This is a complex mechanism, and "Test Driven Development" is the recommended way to implement any changes here.
 * When reporting a bug in the following logic, please provide a failing test case. The easiest way in most use cases 
 * is to adopt the existing `RouteChangedUpdaterTest::provideRoutes` test data provider.
```

PR #7835, opened 2025-03-05
Add handling for tree_full_edit to disconnect route from parent

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no 
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | part of https://github.com/sulu/sulu/pull/7726
| License | MIT

#### What's in this PR?

Add handling for tree_full_edit to disconnect route from parent.

As discussed with @chirimoya as soon as I edit more as the last part of a URL it is kind of a opt out from parent rout updates and so the route itself not longer has a `parentRoute` and is this so disconnected and lifes by itself.

Example:

```
 - /parent
     - /parent/child-a
```

Will be changed to

```
 - /parent
     - /child-a
```

The connection to the `parent` is removed and so it not longer gets updated.

#### Why?

Sulu supports tree_full_edit strategy to allow that a page can be edit fully and not only its leaf.

### DIFF
--- a/packages/route/src/Infrastructure/Doctrine/EventListener/RouteChangedUpdater.php
+++ b/packages/route/src/Infrastructure/Doctrine/EventListener/RouteChangedUpdater.php
@@ -98,6 +98,7 @@ class RouteChangedUpdater implements ResetInterface
                 ->andWhere(\is_string($site) ? 'parent.site = :site' : 'parent.site IS NULL')
                 ->andWhere('parent.locale = :locale')
                 ->andWhere('(parent.slug = :newSlug OR parent.slug LIKE :oldSlugSlash)') // direct child is using newSlug already updated as we are in PostFlush, grand child use oldSlugWithSlash as not yet updated
+                ->andWhere('(child.slug LIKE :oldSlugSlash)') // ignore disconnected child routes in case of full tree edit
                 ->setParameter('newSlug', $newSlug, ParameterType::STRING)
                 ->setParameter('oldSlugSlash', $oldSlug . '/%', ParameterType::STRING)
                 ->setParameter('locale', $locale, ParameterType::STRING);
@@ -115,7 +116,7 @@ class RouteChangedUpdater implements ResetInterface
              *     resource_id: string,
              * }> $childAndGrandChildResult
              */
-            $childAndGrandChildResult = $selectQueryBuilder->executeQuery()->fetchAllAssociative();
+            $childAndGrandChildResult = $selectQueryBuilder->executeQuery()->iterateAssociative();
             $parentIds = [];
             $childAndGrandChildHistoryRoutes = [];
             foreach ($childAndGrandChildResult as $childAndGrandChildRow) {
@@ -126,7 +127,7 @@ class RouteChangedUpdater implements ResetInterface
                     $locale,
                     $childAndGrandChildRow['slug'],
                     $childAndGrandChildRow['site'],
-                    null, // history never has parents ad they never will be updated
+                    null, // history never has parents as they never will be updated
                 );
             }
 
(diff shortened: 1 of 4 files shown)


packages/route/src/Domain/Model/Route.php
```diff
@@ -45,6 +45,13 @@ public function __construct(string $resourceKey, string $resourceId, string $loc
 
     public function setSlug(string $slug): static
     {
+        if ($this->parentRoute
+            && $this->slug !== $slug
+            && static::extractParentSlug($slug) !== static::extractParentSlug($this->slug)
+        ) {
+            $this->parentRoute = null;
```
alexander-schranz: lets discuss this deeper with @chirimoya /cc @Prokyonn .

Technically we don't need remove the relation, it only depends on what should happening if we do:

```
 - parent
    - parent/child
        - parent/child/grand-child -> parent/grand-child
```

->

```
 - parent -> changed
     - parent/child
        - parent/grand-child
```

what is the expected result:

**A.**

```
 - changed
    - changed/child
        - parent/grand-child
```

or 

**B.**

```
 - changed
    - changed/child
        - changed/grand-child
```

---

if we keep the relation and I later change 

```
 - parent
    - parent/child
        - parent/grand-child -> parent/child/grand-child
```

it will automatically opt in again into the parrent update without technical complexity to know again how to set the parentRoute relation. Else I'm not sure how we opt in again after a URL was once opt out. If behaviour `B.` is fine optin and optout are technically easier to implement. If we want behaviour `A.` optout is also easy, but optin I'm not sure yet.

packages/route/src/Infrastructure/Doctrine/EventListener/RouteChangedUpdater.php
```diff
@@ -115,7 +116,7 @@ public function postFlush(PostFlushEventArgs $args): void
              *     resource_id: string,
              * }> $childAndGrandChildResult
              */
-            $childAndGrandChildResult = $selectQueryBuilder->executeQuery()->fetchAllAssociative();
+            $childAndGrandChildResult = $selectQueryBuilder->executeQuery()->iterateAssociative();
```
alexander-schranz, line 119: should improve performance for bigger updates

packages/route/src/Infrastructure/Doctrine/EventListener/RouteChangedUpdater.php
```diff
@@ -144,25 +145,25 @@ public function postFlush(PostFlushEventArgs $args): void
...
+                }
+
+                // update child and grand routes
+                $updateQueryBuilder = $connection->createQueryBuilder()
+                    ->update($routesTableName, 'r')
+                    ->set('slug', 'CONCAT(:newSlug' . $newSlugCast . ', SUBSTRING(slug, ' . (\strlen($oldSlug) + 1) . '))')
+                    ->setParameter('newSlug', $newSlug, ParameterType::STRING)
+                    ->where('parent_id IN (:parentIds)')
+                    ->andWhere('slug LIKE :oldSlugSlash') // ignore disconnected child routes in case of full tree edit
+                    ->setParameter('oldSlugSlash', $oldSlug . '/%', ParameterType::STRING)
```
alexander-schranz, line 161: beside the indention only this lines where changed

packages/route/src/Domain/Model/Route.php
alexander-schranz: implemented now

packages/route/src/Infrastructure/Doctrine/EventListener/RouteChangedUpdater.php
```diff
@@ -98,6 +98,7 @@ public function postFlush(PostFlushEventArgs $args): void
                 ->andWhere(\is_string($site) ? 'parent.site = :site' : 'parent.site IS NULL')
                 ->andWhere('parent.locale = :locale')
                 ->andWhere('(parent.slug = :newSlug OR parent.slug LIKE :oldSlugSlash)') // direct child is using newSlug already updated as we are in PostFlush, grand child use oldSlugWithSlash as not yet updated
+                ->andWhere('(child.slug LIKE :oldSlugSlash)') // ignore disconnected child routes in case of full tree edit
```
alexander-schranz, line 101: this is the whole magic now we ignore the once which do not start the same way and are not a child

Prokyonn (approved):

PR #7839, opened 2025-03-07
Add navigation context to SuluPageBundle 3.0

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no yes
| BC breaks? | no
| Deprecations? | no
| Related issues/PRs | https://github.com/sulu/sulu/issues/7672
| License | MIT

#### What's in this PR?

Adds NavigationContexts to pages in the new SuluPageBundle 3.0


packages/page/config/doctrine/Page/PageDimensionContent.orm.xml
```diff
@@ -17,5 +17,17 @@
...
+
+        <one-to-many
+            field="navigationContexts"
+            target-entity="Sulu\Page\Domain\Model\PageDimensionContentNavigationContext"
+            mapped-by="pageDimensionContent"
+            orphan-removal="true"
+        >
+            <cascade>
+                <cascade-persist/>
+                <cascade-remove/>
```
alexander-schranz: we should never ever use `cascade-remove` just declare on the other side `on-delete="CASCADE"` to avoid unnecessary queries on deletion of entites.

alexander-schranz:
```suggestion
```

packages/page/config/doctrine/Page/PageDimensionContentNavigationContext.orm.xml
```diff
@@ -0,0 +1,20 @@
...
+    >
+        <id name="navigationContext" type="string" column="navigation_context" length="64" />
+        <id name="pageDimensionContent" association-key="true" />
+
+        <many-to-one field="pageDimensionContent" target-entity="Sulu\Page\Domain\Model\PageDimensionContent" inversed-by="navigationContexts">
+            <join-column name="page_dimension_content_id" referenced-column-name="id" on-delete="CASCADE" nullable="false" />
+        </many-to-one>
+
+        <indexes>
+            <index name="idx_navigationContext" columns="navigation_context"/>
```
alexander-schranz:
```suggestion
            <index name="idx_page_navigation_context" fields="navigationContext"/>
```

A `index` name is unique in postgres over the whole database. That is also why it is in doctrine internal prefixed with a table hash. we atleast should add `page` here. Also we used snack case for this.

packages/page/config/doctrine/Page/PageDimensionContentNavigationContext.orm.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
+<doctrine-mapping xmlns="http://doctrine-project.org/schemas/orm/doctrine-mapping"
+                  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                  xsi:schemaLocation="http://doctrine-project.org/schemas/orm/doctrine-mapping http://doctrine-project.org/schemas/orm/doctrine-mapping.xsd">
+    <entity
+        name="Sulu\Page\Domain\Model\PageDimensionContentNavigationContext"
+        table="pa_page_dimension_content_navigation_contexts"
+    >
+        <id name="navigationContext" type="string" column="navigation_context" length="64" />
+        <id name="pageDimensionContent" association-key="true" />
```
alexander-schranz: This kind of ids did make problems in the past in upgrades of doctrine as it is not a widely used feature.

So instead I would prefer we should go with a more common way to have not the need to fight against somebody eventually break this in doctrine orm again:

```suggestion
        <id name="id" type="integer" column="id">
            <generator strategy="AUTO"/>
        </id>

        <field name="navigationContext" type="string" column="name" length="64" />
```

packages/page/config/forms/content_settings_navigation_context.xml
```diff
@@ -0,0 +1,22 @@
...
+>
+    <key>content_settings_navigation_context</key>
+
+    <tag name="sulu_content.content_settings_form" instanceOf="Sulu\Page\Domain\Model\PageDimensionContentInterface" priority="100"/>
+
+    <properties>
+        <property
+            name="navigationContexts"
+            type="page_settings_navigation_select"
+            colspan="6"
```
alexander-schranz: Colspan looks strange as we don't know what the field after it is. Should we wrap this into a section `Page Settings`? 🤔

packages/page/config/forms/content_settings_navigation_context.xml
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" ?>
+<form xmlns="http://schemas.sulu.io/template/template"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://schemas.sulu.io/template/template http://schemas.sulu.io/template/form-1.0.xsd"
+>
+    <key>content_settings_navigation_context</key>
```
alexander-schranz: Depending if we want add a new section here we maybe rename the key 🤔 

```suggestion
    <key>content_settings_page_settings</key>
```

alexander-schranz (changes requested):

packages/page/config/forms/content_settings_navigation_context.xml
Prokyonn: Wrapped it into a section, but I still would like to use the colspan 6 to be consistent with the other settings like e.g. webspace

image

Prokyonn: Thinking about this... webspace should not even be there for pages 😂  we should tackle this in another PR

alexander-schranz: Yeah remove the interface and trait from pages :)

PR #7840, opened 2025-03-07
Add example to content controller about extending

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Add example to content controller about extending.

#### Why?

Would like that the code already show what best practices for extending something is. While most cases will be `decorating` a service here the controller is better with extending.

### DIFF
--- a/packages/content/src/UserInterface/Controller/Website/ContentController.php
+++ b/packages/content/src/UserInterface/Controller/Website/ContentController.php
@@ -23,8 +23,11 @@ use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\Exception\NotAcceptableHttpException;
 
 /**
- * It is common to extend this controller if you need to load additional data for a specific template. Inside the
- * `resolveSuluParameters` method you have access to all managed content and can extend it with additional parameters.
+ * This controller is commonly extended when additional data needs to be loaded for a specific template. 
+ * The `resolveSuluParameters` method provides access to all managed content, allowing you to extend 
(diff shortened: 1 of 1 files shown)


packages/content/src/UserInterface/Controller/Website/ContentController.php
```diff
@@ -23,8 +23,21 @@
 use Symfony\Component\HttpKernel\Exception\NotAcceptableHttpException;
 
 /**
- * TODO this controller will later replace the DefaultController of the WebsiteBundle
- *      and should be base for all content controllers.
+ * It is common to extend this controller if you need to load additional data for a specific template. Inside the
+ * `resolveSuluParameters` method you have access to all managed content and can extend it with additional parameters.
+ *
```
Prokyonn:
```suggestion
 * This controller is commonly extended when additional data needs to be loaded for a specific template. 
 * The `resolveSuluParameters` method provides access to all managed content, allowing you to extend 
 * it with additional parameters.
 *
```

packages/content/src/UserInterface/Controller/Website/ContentController.php
```diff
@@ -23,8 +23,21 @@
...
+ *      <code>
```
Prokyonn, line 32:
```suggestion
 * Example usage:
 *
 *      <code>
```

PR #7842, opened 2025-03-10
Fix mandatory condition for page tree route

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Adds property metadata to the PageTreeRouteContentType.

#### Why?

The mandatory condition of the `page_tree_route` content type did not work properly.


alexander-schranz:
Think we should backport this also to 2.5.

alexander-schranz (approved):

alexander-schranz (approved):

PR #7848, opened 2025-03-18
Create default history object in JS Router

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | #4709 
| License | MIT
| Documentation PR | -

#### What's in this PR?
Using the correct empty value for the `Router` constructor.

#### Why?

I tried updating to Typescript and noticed that these objects can not be empty. They need to be a history object. However we can't just add proper typing to the router, because the history package does not support flow.


mamazu:
We do never actually set this to anything other than the default. Maybe we could just default the constructor to have a default argument? Then all of the history code is local to the router.

PR #7849, opened 2025-03-19
Refactor content resolver to resolve recursively

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Related issues/PRs | #7672 
| License | MIT

#### What's in this PR?

Refactors the content resolver to resolve recursively and break circular resources after a specified depth

#### Why?

We have to resolve newly loaded resources (esp. ContentRichEntities) recursively to guarantee that everything is resolved properly.

### DIFF
--- a/packages/content/src/Application/ContentResolver/Resolver/ResolverInterface.php
+++ b/packages/content/src/Application/ContentResolver/Resolver/ResolverInterface.php
@@ -24,12 +24,5 @@ interface ResolverInterface
      *
      * @param DimensionContentInterface<T> $dimensionContent
      */
-    public function resolve(DimensionContentInterface $dimensionContent): ContentView;
-
-    /**
-     * @template T of ContentRichEntityInterface
-     *
-     * @param DimensionContentInterface<T> $dimensionContent
-     */
-    public function supports(DimensionContentInterface $dimensionContent): bool;
+    public function resolve(DimensionContentInterface $dimensionContent): ?ContentView;
 }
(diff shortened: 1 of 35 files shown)


packages/content/src/Application/ContentResolver/Resolver/ResolverInterface.php
```diff
@@ -25,4 +25,11 @@ interface ResolverInterface
      * @param DimensionContentInterface<T> $dimensionContent
      */
     public function resolve(DimensionContentInterface $dimensionContent): ContentView;
+
+    /**
+     * @template T of ContentRichEntityInterface
+     *
+     * @param DimensionContentInterface<T> $dimensionContent
+     */
+    public function supports(DimensionContentInterface $dimensionContent): bool;
```
alexander-schranz: Not really a fan of supports methods, as we don't use it in other cases in the content bundle and do a early return there. The supports methods ends up in a double check of the type because we always need todo checks inside the called method also.

What do you think about make the resolver nullable return type instead e.g.:

```php
public function resolve(DimensionContentInterface $dimensionContent): ?ContentView
```

Prokyonn: I was also not really sure, both options have some cons, e.g. null checking. But I don't have a strong opinion on this, so if you prefer it to be nullable, I can refactor it :)

alexander-schranz: as discussed we will go with `?ContentView`

alexander-schranz (approved):

PR #7850, opened 2025-03-19
Add compiler pass to decorate ResourceLoaders with an in memory cache

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no yes
| Deprecations? | no
| Related issues/PRs | #7672 
| License | MIT

#### What's in this PR?

Adds a CompilerPass that decorates all ResourceLoaders with an in-memory cache.

#### Why?

In some edge-cases we have to query the same resource twice, so we need a caching layer that only lives for the current request.


packages/content/src/Application/ResourceLoader/Loader/CachedResourceLoader.php
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Content\Application\ResourceLoader\Loader;
+
+use Symfony\Contracts\Service\ResetInterface;
+
+class CachedResourceLoader implements ResourceLoaderInterface, ResetInterface
```
alexander-schranz, line 19: Mark as `@internal` we maybe change the caching mechanic in future.

packages/content/src/Application/ResourceLoader/Loader/CachedResourceLoader.php
```diff
@@ -0,0 +1,63 @@
...
+
+        if (isset($this->cache[$cacheKey])) {
+            return $this->cache[$cacheKey];
+        }
+
+        $result = $this->decoratedResourceLoader->load($ids, $locale, $params);
+        $this->cache[$cacheKey] = $result;
+
+        return $result;
+    }
```
alexander-schranz: This looks unexpected we do NOT use here a memoize based cache mechanic here.

We need cache here by `id` (single) and merge it for next call. We can have first call with `[1, 2]` second with `[1, 3]` then the decorated is only called with `[3]`.:

```suggestion
    public function load(array $ids, ?string $locale, array $params = []): array
    {
        $notFoundIds = [];
        $cachedResult = [];
        foreach ($ids as $id) {
            $cacheKey = $id;
            if (!isset($this->cache[$cacheKey])) {
                 $notFoundIds = [];
                 
                 continue;
            }
            
            $cachedResult[$id] = $this->cache[$cacheKey];
        }
        
        $result = $this->decoratedResourceLoader->load($ids, $locale, $params);
        
        // cache new result
        // merge with cached result
        
        return $result;
    }
```

packages/content/tests/Unit/Content/Application/ResourceLoader/CachedResourceLoaderTest.php
```diff
@@ -0,0 +1,111 @@
...
+     */
+    private $decoratedResourceLoader;
+
+    private CachedResourceLoader $cachedResourceLoader;
+
+    protected function setUp(): void
+    {
+        $this->decoratedResourceLoader = $this->prophesize(ResourceLoaderInterface::class);
+        $this->cachedResourceLoader = new CachedResourceLoader(
+            $this->decoratedResourceLoader->reveal()
```
alexander-schranz: Would try to not mock instead create `anonymous` class implementation which `unset` an array and would throw a exception if it calls the same id again:

```php
    public static array $values = []; 

    public function load(array $ids, ?string $locale, array $params = []): array
    {
        $models = [];
        foreach ($ids as $id) {
            if (!isset(self::$values[$id])) {
                 throw new LogicException('The CachedResourceLoader should never call the same "id" multiple times. Id "%s" requested again');
            }
            
            $models = self::$values[$ids];
            unset(self::$values[$ids]); // unset to test not called again with same id
        }
        
        return $models;
    }
```

You can at start of every test refill it with:

`$this->decoratedResourceLoader::$values = [1 => ...]`

This way we not require any mocking.

packages/content/tests/Unit/Infrastructure/Symfony/HttpKernel/Compiler/ResourceLoaderCacheCompilerPassTest.php
```diff
@@ -0,0 +1,107 @@
...
+use PHPUnit\Framework\TestCase;
+use Prophecy\Argument;
+use Prophecy\PhpUnit\ProphecyTrait;
+use Sulu\Content\Application\ResourceLoader\Loader\CachedResourceLoader;
+use Sulu\Content\Infrastructure\Symfony\HttpKernel\Compiler\ResourceLoaderCacheCompilerPass;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Definition;
+use Symfony\Component\DependencyInjection\Reference;
+
+class ResourceLoaderCacheCompilerPassTest extends TestCase
```
alexander-schranz: we should not sue a mocked test here we should create a test using `AbstractCompilerPassTestCase` in this case.

alexander-schranz (changes requested):

packages/content/src/Infrastructure/Symfony/HttpKernel/Compiler/ResourceLoaderCacheCompilerPass.php
```diff
@@ -0,0 +1,39 @@
...
+        foreach ($resourceLoaders as $id => $tags) {
+            $container->register($id . '.cached', CachedResourceLoader::class)
+                ->setDecoratedService($id)  // This service decorates the original loader
+                ->setArguments([
+                    new Reference($id . '.cached.inner'),
+                ])
+                ->setPublic(false);
+
+            /** @var array<string, string> $tag */
+            foreach ($tags as $tag) {
```
Prokyonn: @Prokyonn verify if necessary

alexander-schranz (approved):

packages/content/src/Infrastructure/Symfony/HttpKernel/Compiler/ResourceLoaderCacheCompilerPass.php
Prokyonn: > All custom [service tags](https://symfony.com/doc/current/service_container/tags.html) from the decorated service are removed in the new service. Only certain built-in service tags defined by Symfony are retained: container.service_locator, container.service_subscriber, kernel.event_subscriber, kernel.event_listener, kernel.locale_aware, and kernel.reset.

https://symfony.com/doc/current/service_container/service_decoration.html

packages/content/src/Infrastructure/Symfony/HttpKernel/Compiler/ResourceLoaderCacheCompilerPass.php
```diff
@@ -0,0 +1,53 @@
...
+                ->setDecoratedService($id)
+                ->setArguments([
+                    new Reference($decoratorId . '.inner'),
+                ])
+                ->addTag('kernel.reset', ['method' => 'reset'])
+                ->setPublic(false);
+
+            $decoratorService = $container->getDefinition($decoratorId);
+
+            // Copy all tags from the original service to the decorator
```
alexander-schranz:
```suggestion
            // Copy all tags from the original service to the decorator
            // else tags get lost as documented in when decorating the service: https://symfony.com/doc/current/service_container/service_decoration.html
```

PR #7825, opened 2025-02-26
Add more details to exception messages in FormatManager

michaelzangerle:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

Improves the exception messages.

#### Why?

Generic exception messages are not helpful when debugging bugs. Examples before and after:

`Media was not found`  => `Media with id "%s" was not found.`
`Requested FileVersion for media was not found` =>  `Requested FileVersion "%s" for media with id "%s" was not found.`

PR #7853, opened 2025-03-20
Fix false unexception error logs in image proxy functionality

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no  
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

I propose to not log errors for ImageProxyExceptions which ends in a 404.

#### Why?

There are a few exception thrown but nothing really validates from my point of view as an error:

1. ImageProxyMediaNotFoundException (not found media) -> 404
2. ImageProxyMediaNotFoundException (not found file version) -> 404
3. ImageProxyMediaNotFoundException (slug not match) -> 404 ⚠️ introduced as security fix in but think still validates as not an error even some in the past did misused none matching URLs. Can eventually still make harder to debug for projects upgrade from previous versions now. /cc @chirimoya 
4. ImageProxyMediaNotFoundException (image format not found) -> 404
5. InvalidMimeTypeForPreviewException MimeType not supported -> 404
6. ImageProxyInvalidImageFormat MimeType not supported -> 404
7. ImageProxyInvalidImageFormat MimeType not supported -> 404

The only exception which should end in error is if the media entity exists but the original file not longer exists. The LocalStorage currently throws a `IOException` in that case, the flysystem sadly a ImageProxyMediaNotFoundException which is incorrect.

So this PR normalize that the adapters behave the same here, a `500 error` is fine as in that case as it really something wrong on the "Server" if original file is removed but entity still exists.

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/Controller/MediaStreamController.php
+++ b/src/Sulu/Bundle/MediaBundle/Controller/MediaStreamController.php
@@ -13,13 +13,13 @@ namespace Sulu\Bundle\MediaBundle\Controller;
 
 use Sulu\Bundle\MediaBundle\Admin\MediaAdmin;
 use Sulu\Bundle\MediaBundle\Entity\Collection;
-use Sulu\Bundle\MediaBundle\Entity\File;
 use Sulu\Bundle\MediaBundle\Entity\FileVersion;
 use Sulu\Bundle\MediaBundle\Entity\MediaInterface;
 use Sulu\Bundle\MediaBundle\Entity\MediaRepositoryInterface;
 use Sulu\Bundle\MediaBundle\Media\DispositionType\DispositionTypeResolver;
 use Sulu\Bundle\MediaBundle\Media\Exception\FileVersionNotFoundException;
-use Sulu\Bundle\MediaBundle\Media\Exception\ImageProxyException;
+use Sulu\Bundle\MediaBundle\Media\Exception\ImageProxyInvalidUrl;
+use Sulu\Bundle\MediaBundle\Media\Exception\ImageProxyUrlNotFoundException;
 use Sulu\Bundle\MediaBundle\Media\Exception\MediaException;
 use Sulu\Bundle\MediaBundle\Media\FormatCache\FormatCacheInterface;
 use Sulu\Bundle\MediaBundle\Media\FormatManager\FormatManagerInterface;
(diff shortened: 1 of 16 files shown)


src/Sulu/Bundle/MediaBundle/Media/FormatManager/FormatManager.php
```diff
@@ -161,8 +162,8 @@ public function returnImage($id, $formatKey, $fileName /*, int<1, max>|null $ver
                     $formatKey
                 );
             }
-        } catch (MediaException $e) {
-            $this->logger->error($e->getMessage(), ['exception' => $e]);
+        } catch (ImageProxyException $e) {
+            $this->logger->debug($e->getMessage(), ['exception' => $e]);
```
alexander-schranz, line 159: major change is this to be a `debug` now only

src/Sulu/Bundle/MediaBundle/Media/Exception/InvalidMimeTypeForPreviewException.php
```diff
@@ -11,7 +11,7 @@
 
 namespace Sulu\Bundle\MediaBundle\Media\Exception;
 
-class InvalidMimeTypeForPreviewException extends MediaException
+class InvalidMimeTypeForPreviewException extends ImageProxyException
```
alexander-schranz, line 14: This maybe should be renamed to InvalidMimeTypeImageProxyException in 3.0.

alexander-schranz:
Mentioning here @michaelzangerle because of https://github.com/sulu/sulu/pull/7825

src/Sulu/Bundle/MediaBundle/Controller/MediaStreamController.php
```diff
@@ -67,16 +66,16 @@ public function getImageAction(Request $request)
...
+        } catch (ImageProxyUrlNotFoundException $e) {
             throw new NotFoundHttpException('Image create error. Code: ' . $e->getCode(), $e);
         }
+
+        return $this->formatManager->returnImage(
+            $mediaProperties['id'],
+            $mediaProperties['format'],
+            $mediaProperties['fileName'],
+            $version,
+        );
```
alexander-schranz, line 79: moved outside as returnImage does catch its own exception and return always a response. So the above catch is only for `ImageProxyUrlNotFoundException` of the `analyzedMediaUrl` used.

Prokyonn (approved):

PR #7509, opened 2024-07-05
Add support for doctrine/orm 3

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT
| Documentation PR | sulu/sulu-docs# 

#### What's in this PR?

Add support for doctrine/orm 3.

Most of the changes cherry picked into https://github.com/sulu/sulu/pull/7857/files to prepare doctrine/orm 3 changes soonish.

#### Why?

A new version of orm was released and we should try to keep Sulu uptodate.

#### TODO

 - [x] Oro Extension Doctrine/Lexer 3 support: https://github.com/oroinc/doctrine-extensions/pull/108
 - [x] Data Fixtures 2 Support in Doctrine Bundle https://github.com/doctrine/DoctrineFixturesBundle/pull/457
 - [x] Recheck Partial as it is was readded to doctrine/orm ^3.3
 - [x] Oro Extension Doctrine/Orm 3 support: https://github.com/oroinc/doctrine-extensions/pull/97
 - [x] Oro Extension 3.0
    - [x] https://github.com/oroinc/doctrine-extensions/pull/112
    - [x] Release Missing: https://github.com/oroinc/doctrine-extensions/releases/tag/3.0
 - [x] Intersection Support for Prophecy: https://github.com/phpspec/prophecy/issues/535 or https://github.com/phpspec/prophecy/pull/637

PR #7857, opened 2025-03-25
Prepare some doctrine/orm 3 changes

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | parts of https://github.com/sulu/sulu/pull/7509
| License | MIT

#### What's in this PR?

Adjustments from https://github.com/sulu/sulu/pull/7509 without upgrading doctrine/orm 3.

#### Why?

Prepare support for doctrine/orm 3. Full support can not be added yet see: https://github.com/sulu/sulu/pull/7509

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -313,6 +313,12 @@ parameters:
 			path: src/Sulu/Bundle/ActivityBundle/Infrastructure/Doctrine/Repository/ActivityRepository.php
 
 		-
+			message: '#^Call to function method_exists\(\) with Doctrine\\ORM\\Id\\AbstractIdGenerator and ''generateId'' will always evaluate to true\.$#'
+			identifier: function.alreadyNarrowedType
+			count: 1
+			path: src/Sulu/Bundle/ActivityBundle/Infrastructure/Doctrine/Repository/ActivityRepository.php
+
+		-
 			message: '#^Parameter \#2 \$value of method Doctrine\\DBAL\\Query\\QueryBuilder\:\:setValue\(\) expects string, mixed given\.$#'
 			identifier: argument.type
 			count: 1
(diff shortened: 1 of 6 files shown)


src/Sulu/Component/SmartContent/Orm/DataProviderRepositoryTrait.php
```diff
@@ -339,16 +341,6 @@ private function appendRelationAnd(QueryBuilder $queryBuilder, $relation, $value
...
 
-    /**
-     * Creates a new QueryBuilder instance that is prepopulated for this entity name.
-     *
-     * @param string $alias
-     * @param string|null $indexBy
-     *
-     * @return QueryBuilder
-     */
-    abstract public function createQueryBuilder($alias, $indexBy = null);
```
alexander-schranz: adding the return type here would a major BC Break making lot of bundles fail. So removing the abstract and move it to a phpdoc is only way for Backards compatibility to orm 2 based bundles.

src/Sulu/Bundle/MediaBundle/Entity/MediaDataProviderRepository.php
```diff
@@ -238,7 +238,7 @@ protected function appendSortByJoins(QueryBuilder $queryBuilder, $alias, $locale
             ->setParameter('locale', $locale);
     }
 
-    public function createQueryBuilder($alias, $indexBy = null)
+    public function createQueryBuilder($alias, $indexBy = null): \Doctrine\ORM\QueryBuilder
```
Prokyonn, line 241: Return type should not be necessary here, because of the phpdoc and this will be a BC break

Prokyonn: I guess this should also be removed? 🤔 

```suggestion
    public function createQueryBuilder($alias, $indexBy = null)
```

Prokyonn (changes requested):

src/Sulu/Bundle/MediaBundle/Entity/MediaDataProviderRepository.php
alexander-schranz: This return type is required now. It can not be removed but I would think that this service is nobody `extends` from and would be fine with technical bc break here.

alexander-schranz:
rebased after we merged: https://github.com/sulu/sulu/pull/8184 and targetting the 3.0 branch.

Prokyonn (approved):

PR #7860, opened 2025-03-26
Add custom Symfony CMF Dynamic Router / Route Provider to new Route Bundle

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Implement the cmf dynamic router into the new route bundle. The implementation will be kept internal. The new router uses our own Route Model / Entity: https://github.com/sulu/sulu/blob/3.0/packages/route/config/doctrine/Route/Route.orm.xml 

#### Why?

The new route bundle should be the only bundle in core handling the CMS based routing.


packages/route/src/Infrastructure/SymfonyCmf/Routing/CmfRouteProvider.php
```diff
@@ -0,0 +1,60 @@
...
+ */
+class CmfRouteProvider implements RouteProviderInterface
+{
+    /**
+     * @param RouteCollectionForRequestLoaderInterface[] $routeCollectionForRequestLoaders
+     */
+    public function __construct(
+        private iterable $routeCollectionForRequestLoaders,
+    ) {
+    }
```
alexander-schranz: The whole idea is that this is our only connection point with the `Symfony\Cmf`. Marked it as `@internal` as this shouldn't be changed or used by projects. Also we are even free the replace it if in the future dynamic router is maybe part of symfony core or a new major version of it is released with changes.

With our own `RouteCollectionForRequestLoaderInterface` which just has `public function getRouteCollectionForRequest(Request $request): RouteCollection;` we have all about routing then in our own control. This removes the complex service definitions which we have here:

 - https://github.com/sulu/sulu/blob/b605ff31576d2f91dcd3b8596ebbe84350c9710c/src/Sulu/Bundle/RouteBundle/Resources/config/routing.xml#L11-L63
 - https://github.com/sulu/sulu/blob/b605ff31576d2f91dcd3b8596ebbe84350c9710c/src/Sulu/Bundle/CustomUrlBundle/Resources/config/routing.xml#L11-L108
 - https://github.com/sulu/SuluRedirectBundle/blob/6964ba5714cc1849e3243b6882a289fc845595f9/Resources/config/router.xml#L7-L51
 
So we are replacing the different `DynamicRouter` of 1.x and 2.x with a single `DynamicRouter`. Which config is a lot easier:

```php
            $builder->prependExtensionConfig(
                'cmf_routing',
                [
                    'chain' => [
                        'routers_by_id' => [
                            'router.default' => 100,
                            'cmf_routing.dynamic_router' => 20,
                        ],
                    ],
                    'dynamic' => [
                        'route_provider_service_id' => 'sulu_route.symfony_cmf_route_provider',
                        'enabled' => true,
                    ],
                ]
            );
```

```php
        if ($builder->hasExtension('cmf_routing')) {
            $services->set('sulu_route.symfony_cmf_route_provider')
                ->class(CmfRouteProvider::class)
                ->args([
                    tagged_iterator('sulu_route.route_collection_for_request_loader'),
                ]);
        }
```

I also added the `RouteDefaultsProviderInterface` which is very similar to our existing RouteBundle entity of 2.6 version. But simplified it that every resourceKey requires an own defaults provider.

packages/route/tests/Application/Kernel.php
```diff
@@ -41,13 +46,26 @@ public function registerBundles(): iterable
...
 
+        $hasDynamicRouting = false;
         foreach ($bundles as $key => $bundle) {
             // remove old route bundle to avoid conflicts
             if (DeprecatedSuluRouteBundle::class === $bundle::class
                 || SuluPreviewBundle::class === $bundle::class
+                || \str_contains($bundle::class, 'Sulu')
+                || \str_contains($bundle::class, 'Massive')
+                || \str_contains($bundle::class, 'PHPCR')
+                || \str_contains($bundle::class, 'SecurityBundle')
```
alexander-schranz, line 57: currently having lot of conflicts with Sulu and other bundles as that combinations not work properly so I removed all bundles which are not required to run the RouteBundle tests.

packages/route/src/Application/Routing/Matcher/RouteHistoryDefaultsProvider.php
```diff
@@ -0,0 +1,69 @@
...
+    public function getDefaults(Route $route): array
+    {
+        \assert($route->isHistory(), \sprintf('Route must be of type "%s", but "%s" given.', Route::HISTORY_RESOURCE_KEY, $route->getResourceKey()));
+
+        $resourceIdParts = \explode('::', $route->getResourceId(), 2);
+        $resourceKey = $resourceIdParts[0];
+        $resourceId = $resourceIdParts[1] ?? '';
+
+        if ('' === $resourceKey || '' === $resourceId) {
+            throw new \RuntimeException(\sprintf('The given history route "resourceId" need to contain resourceKey and resourceId separated by "::", but "%s" given.', $route->getResourceId()));
```
Prokyonn:
```suggestion
            throw new \RuntimeException(\sprintf('The given history route "resourceId" has to contain resourceKey and resourceId separated by "::", but "%s" given.', $route->getResourceId()));
```

packages/route/src/Application/Routing/Matcher/RouteHistoryDefaultsProvider.php
```diff
@@ -0,0 +1,69 @@
...
+        }
+
+        $targetRoute = $this->routeRepository->findOneBy([
+            'resourceKey' => $resourceKey,
+            'resourceId' => $resourceId,
+            'site' => $route->getSite(),
+        ]);
+
+        if (null === $targetRoute) {
+            throw new GoneHttpException(\sprintf('The target route with resourceKey "%s" and resourceId "%s" seems not longer exists.', $resourceKey, $resourceId));
```
Prokyonn:
```suggestion
            throw new GoneHttpException(\sprintf('The target route with resourceKey "%s" and resourceId "%s" no longer exists.', $resourceKey, $resourceId));
```

Prokyonn (approved):

packages/route/src/Infrastructure/SymfonyCmf/Routing/CmfRouteProvider.php
```diff
@@ -0,0 +1,58 @@
...
+
+    public function getRouteCollectionForRequest(Request $request): RouteCollection
+    {
+        foreach ($this->routeCollectionForRequestLoaders as $routeCollectionLoader) {
+            $routeCollection = $routeCollectionLoader->getRouteCollectionForRequest($request);
```
alexander-schranz, line 37: I'm little bit unsure about giving here the Request `$request` here. I see that the Symfony router is also using the Request or string pathInfo for the UrlMatcher: https://github.com/symfony/routing/blob/ee9a67edc6baa33e5fae662f94f91fd262930996/Matcher/UrlMatcher.php#L85

But from my point of view the `RequestContext` would maybe be a better option 🤔  to be more consistent between `Matcher` and `Generator`, maybe I miss understand the concept of the RequestContext, and we should stay with Request for matching and context for generating.

/cc @Prokyonn @dbu

dbu: as i understand, the matcher uses the Request as that holds a rich model of the whole request including headers. the header bag gives you precise access to headers, even parsing cache headers, or more relevant to routing the accept header etc.

packages/route/src/Application/Routing/Matcher/RouteCollectionForRequestRouteLoader.php
```diff
@@ -0,0 +1,73 @@
...
+        $site = $request->attributes->get(RequestAttributeEnum::SITE->value);
+        $slug = $request->attributes->get(RequestAttributeEnum::SLUG->value);
+
+        if ((null !== $site && !\is_string($site))
+            || !\is_string($slug)
+        ) {
+            return new RouteCollection();
+        }
+
+        $route = $this->routeRepository->findOneBy([
```
dbu, line 47: the general phylosophy of the dynamic router is that there could be several database entries potentially matching the request (the default entity of the cmf bundle allows for a parameter matching part in the url, and the usual constraints of symfony routes, so that the db could hold `/data/{id}` and `/data/special` and with the right rules the symfony matcher then picks the correct one.

or multiple routes with different requirements on the accept header.

if you only ever have one route per site & slug, fetching only one is fine, but if you are sure about that, i am not sure about doing a collection loader - you could also implement the RouterInterface directly and have it match the one route you found. (or do you expect that sometimes what you find with the slug is going to be not matched, because it has furhter constraints?)

packages/route/src/Application/Routing/Matcher/RouteCollectionForRequestRouteLoader.php
```diff
@@ -0,0 +1,73 @@
...
+            return new RouteCollection();
+        }
+
+        $routeDefaultsProvider = $this->routeDefaultsProviderLocator->get($route->getResourceKey());
+        \assert($routeDefaultsProvider instanceof RouteDefaultsProviderInterface, 'The RouteDefaultsProvider must implement RouteDefaultsProviderInterface but got: ' . \get_debug_type($routeDefaultsProvider));
+        $defaults = $routeDefaultsProvider->getDefaults($route);
+        $defaults['_sulu_route'] = $route;
+
+        $routeCollection = new RouteCollection();
+        $symfonyRoute = new SymfonyRoute(
```
dbu, line 63: again, just a note, not something i think you have to change: in the cmf bundle, we had our entity extend the symfony route model class. it needed a few hacks however, but as we potentially match many more than 1 single route, it seemed worthwile to not have to convert all the objects.

packages/route/src/Application/Routing/Matcher/RouteCollectionForRequestRouteLoader.php
alexander-schranz: Really thank you @dbu for taking a look at this,

I think we really have never the case that multiple routes are returned. If a "static route" is matched it should not call our Route Entity. If a Slug exists it will return that route entity, and go only to the next loader / provider if it didn't match.

That two or more routes are returned I really think never happens in our case as we only handling "direct" matches. The only thing is that multiple URLs can match the same route e.g. `/blog` and `/blog.rss` where we skipt the `.rss` in loading.

> (or do you expect that sometimes what you find with the slug is going to be not matched, because it has furhter constraints?)

Even it has further constraint it is route independent — so the controller may throw a 404 if now `/blog.rss` is requested but no `blog.rss.twig` exists. But that is a process totally outside of routing and don't need to be handled there.

To be sure the `ChainRouter` will only call the others if it could not find any match?

packages/route/src/Application/Routing/Matcher/RouteCollectionForRequestRouteLoader.php
alexander-schranz: Yeah we avoid `extend` from third party when it is not required. We burned in the past often our hands when third party classes added new methods or return types. So we go over composition over inheritance where possible :) But still thx for the suggestion.

packages/route/src/Application/Routing/Matcher/RouteCollectionForRequestRouteLoader.php
dbu: yes, the first router to match wins.

the routers can have a priority number to control the order (so you can have slugs overwrite symfony system routes, or protect system routes from being overwritten by slugs). 

i would tend to put the symfony system router first, to avoid database queries when a system route is matched, and to prevent somebody creating a page at `/login` or whatever, which could at least mess up things, or in the worst case post security risks.

packages/route/src/Application/Routing/Matcher/RouteCollectionForRequestRouteLoader.php
dbu: yep, makes a lot of sense to do it this way. the constructor certainly is covered by the symfony BC promise. random internal methods and properties can be tricky.

packages/route/src/Infrastructure/SymfonyCmf/Routing/CmfRouteProvider.php
dbu: i would look if you can make this method a custom RouterInterface implementation that either throws the route not found exception or returns the one route that matches the slug.

the dynamic router is built around the idea of finding several candidates and not reimplement the logic to decide which route candidate should win, with accept or parameters and regex etc.

packages/route/src/Application/Routing/Matcher/RouteCollectionForRequestRouteLoader.php
alexander-schranz: I see, thank you. Static Routes are prioritzed also in your case. We still have on our list to not allow creating dynamic patches which may match a static route for better UX, but from the priority view its not a problem.

packages/route/src/Infrastructure/SymfonyCmf/Routing/CmfRouteProvider.php
alexander-schranz: Okay, the Symfony `RouterInterface` just has the `getRouteCollection` without any parameters and so `Request` so to achieve same logic I would create own router then and inject there the requestStack to get currentRequest for the matching. It doesn't make any problems implementing that interface and return a empty collection in none match / cli context? Thought maybe it would cache the empty collection and will then not longer be called again.

https://github.com/symfony/routing/blob/ee9a67edc6baa33e5fae662f94f91fd262930996/RouterInterface.php#L32

it has a strange comment on it:

>      * WARNING: This method should never be used at runtime as it is SLOW.
>      *          You might use it in a cache warmer though.

packages/route/src/Application/Routing/Matcher/RouteCollectionForRequestRouteLoader.php
dbu: for that you could have a service that is triggered during save that does not have the chain router but only the symfony router and tries to match the slug and complains if it matches...

PR #7861, opened 2025-03-26
Fix invalid SuluPage 3.0 admin configuration

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| License | MIT

#### What's in this PR?

Fixes wrong views


Prokyonn:
cherry picked to https://github.com/sulu/sulu/pull/7869 as there we already have a compiler pass that removes the legacy page admin

PR #7867, opened 2025-03-27
Prepare Preview for new Route Bundle implemention

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? |  yes
| BC breaks? | yes
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Add a new `PreviewDefaultsProviderInterface` to decouple the `Preview` Bundle from the `RouteBundle`. Instead of `getObject` a new `getDefaults` is implemented. This should also in future make it easier to create a preview for any none routable entities.

```php

namespace Sulu\Bundle\PreviewBundle\Preview\Provider;

use Sulu\Bundle\PreviewBundle\Preview\PreviewContext;

interface PreviewDefaultsProviderInterface
{
    /**
     * Returns the Route Defaults to render the preview. Should at least return the '_controller' key.
     *
     * @example
     *
     * ```php
     *     return [
     *          'object' => $object,
     *           '_controller' => MyController::class . '::myAction',
     *           'view' => 'my-view',
     *     ];
     * ```
     *
     */
    public function getDefaults(PreviewContext $previewContext): array;

    /**
     * Called before rendering the preview — can be used to update the values of the defaults.
     *
     * @param array<string, mixed> $defaults
     * @param array<string, mixed> $data
     *
     */
    public function updateValues(PreviewContext $previewContext, array $defaults, array $data): array;

    /**
     * Called before rendering the preview — called example when template is switched.
     * Can be used to update the values of the defaults.
     *
     * @param array<string, mixed> $defaults
     * @param array<string, mixed> $context
     *
     */
    public function updateContext(PreviewContext $previewContext, array $defaults, array $context): array;

    public function getSecurityContext(PreviewContext $previewContext): ?string;
}
```

Syntaxed changed and serialize and deserialized is removed. Currently the Content Bundle does always load the whole entity from Database in serialize method thats why this methods are unnecessary. Still we later could introduced a `CachablePreviewDefaultsProviderInterface` with `serialize` and `deserialize` to improve performance if we want. But most performance improvements where done by not rendering the header and footer always, which is still done as the navigation queries did make most things slow. With future navigation cache maybe also that bottleneck goes away. 

#### Why?

Currently the RouteBundle is deeply integrated into the PreviewRenderer which make usage without RouteBundle hard. This also make it hard to have old and new RouteBundle side by side. So we will move step by step the preview to be independent from Route Bundle.

### DIFF
--- a/packages/content/tests/Unit/Content/Infrastructure/Sulu/Admin/ContentViewBuilderFactoryTest.php
+++ b/packages/content/tests/Unit/Content/Infrastructure/Sulu/Admin/ContentViewBuilderFactoryTest.php
@@ -24,6 +24,7 @@ use Sulu\Bundle\AdminBundle\Admin\View\ViewBuilderFactory;
 use Sulu\Bundle\PreviewBundle\Preview\Object\PreviewObjectProviderInterface;
 use Sulu\Bundle\PreviewBundle\Preview\Object\PreviewObjectProviderRegistry;
 use Sulu\Bundle\PreviewBundle\Preview\Object\PreviewObjectProviderRegistryInterface;
+use Sulu\Bundle\PreviewBundle\Preview\Provider\PreviewDefaultsProviderInterface;
 use Sulu\Component\Security\Authorization\PermissionTypes;
 use Sulu\Component\Security\Authorization\SecurityCheckerInterface;
 use Sulu\Content\Application\ContentAggregator\ContentAggregatorInterface;
@@ -74,7 +75,7 @@ class ContentViewBuilderFactoryTest extends TestCase
     }
 
     /**
-     * @param array<string, PreviewObjectProviderInterface> $providers
+     * @param array<string, PreviewObjectProviderInterface|PreviewDefaultsProviderInterface> $providers
      */
     protected function createPreviewObjectProviderRegistry(array $providers): PreviewObjectProviderRegistryInterface
     {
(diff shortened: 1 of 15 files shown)


packages/content/src/Infrastructure/Sulu/Preview/ContentObjectProvider.php
```diff
@@ -90,81 +90,57 @@ public function getObject($id, $locale)
...
+
+        if (!$object instanceof RoutableInterface) {
+            throw new \RuntimeException('Object without RoutableInterface need create there own PreviewDefaultsProvider');
+        }
+
+        return [
+            'object' => $object,
+            '_controller' => ContentController::class . '::indexAction', // TODO call RouteDefaultsProvider
+            'view' => 'base', // TODO should provided by RouteDefaultsProvider
+        ];
```
alexander-schranz: This changes are just a test we need implement this in: https://github.com/sulu/sulu/pull/7749 where the new routing will be implemented

packages/content/src/Infrastructure/Sulu/Preview/ContentObjectProvider.php
alexander-schranz, line 1: Revert this before merging this need implemented in https://github.com/sulu/sulu/pull/7749 else it will not be working.

PR #7868, opened 2025-03-27
Update settings resolver to correctly resolve localization urls

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| License | MIT

#### What's in this PR?

Localization resolving is refactored to work as expected.

#### Why?

The localizations were not resolved correctly


alexander-schranz (approved):

alexander-schranz:
We might will change this if its really part of Settings or another resolver. We need to have a look at this when we implement `properties` part for selections and navigation.

Issue #4084, opened 2018-07-27
Cache Navigation Twig extension between requests


alexander-schranz:
| Q | A
| --- | ---
| Bug? | no
| New Feature? | no
| Sulu Version | 1.6.21, 2.0.0-alpha1

#### Actual Behavior

Currently the twig navigation return value is not cached between requests.

#### Expected Behavior

The navigation as its a heavy process should be cached between requests.

#### Steps to Reproduce

Everytime a page is requested a query is send to the DB.

#### Possible Solutions

Use a doctrine cache to cache it between requests: https://gist.github.com/alexander-schranz/6763148880e60410ed77894ccb585802

Use `array` cache for develop and use `filesystem` for production as default.

choonge:
For our specific case, where we have 7 root menu nav contexts on every page, this would be even faster if we would do only 1 query and then store it in cache like:

```
$this->cache = [
    'main-nav-1' => [ .. ],
    'footer-nav-1' => [ .. ],
    ..
```

That way we only need to access the context as a sub-key of the cache array. However, I have been unable to make the ContentMapper return the navContext as part of the result array.

TomKeur:
I've enabled the navigation cache (from the gist) on a website with a lot of navigation, this is what happened in Blackfire (see screenshot)

![Screenshot from 2020-04-01 15-20-41](https://user-images.githubusercontent.com/1334822/78142941-d985db00-742d-11ea-8e7a-65d8f882bf47.png)

We went from: 2.37sec to: 458ms.
This is run on my local laptop, so I guess it will be even faster in production.

veewee:
In case you are looking for a solution right now:
We are currently bypassing the issue by wrapping the rendered navigation in twig cache blocks:

https://twig.symfony.com/doc/3.x/tags/cache.html

(cache invalidation can be done by listening to sulu's CacheClearEvent)

mamazu:
We are also currently looking into the performance of Sulu navigations and we have found that the slow part was the link preloading in the `PageLinkProvider`.

The problem is that our pages have a lot of properties and Sulu by default loads the entire content of the page to extract a few properties needed for link generation. If you directly query them in the database (see the sql below) you get a big performance improvement. 

Here are the values when trying it with a bunch of links in the dev mode (left is before, right is after)
![slow2](https://github.com/sulu/sulu/assets/14860264/aa885701-0c81-4295-bbfb-1dce8a80da19)

[...]

PR #7869, opened 2025-03-27
Add navigation twig extension for Sulu 3.0

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| License | MIT

#### What's in this PR?

Adds the first two methods for the NavigationTwigExtension

* `sulu_navigation_root_flat`
* `sulu_navigation_root_tree`


alexander-schranz:
As discussed we will move the logic of the navigation twig extension into a `NavigationRepository` which we can use inside the Twig Extension but later also for the Headless JSON Controller. We will add a `NavigationRepositoryInterface` for our different methods. This if we later have the `properties` and `normalizer` created we can decorate with a `CachedNavigationRepository` to solve: https://github.com/sulu/sulu/issues/4084

alexander-schranz:
PHPStan error unrelated

alexander-schranz (approved):

PR #7871, opened 2025-03-28
Mark most preview specific service as internal

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Mark most preview specific services as internal.

#### Why?

There are some deep connections between service in the preview bundle, as we should keep the `Public API` as small as possible to make the maintainance here easier I suggest to mark most of the Preview specific services and how it is rendered as `internal` this may make it easier in future to single kernel setup.

The only connection point for bundles is mostly the `PreviewObjectProviderInterface` / `PreviewDefaultsProviderInterface` all other services are not used by devs.


src/Sulu/Bundle/PreviewBundle/Preview/Object/PreviewObjectProviderRegistryInterface.php
```diff
@@ -11,6 +11,9 @@
 
 namespace Sulu\Bundle\PreviewBundle\Preview\Object;
 
+/**
+ * @internal No BC promises are given for this class. It may be changed or removed at any time.
+ */
 interface PreviewObjectProviderRegistryInterface
```
alexander-schranz, line 17: this will mostly be replaced by a service locator in one of @mamazu PRs.

src/Sulu/Bundle/PreviewBundle/Preview/PreviewInterface.php
```diff
@@ -14,7 +14,7 @@
 use Sulu\Bundle\PreviewBundle\Preview\Exception\ProviderNotFoundException;
 
 /**
- * Interface for preview.
+ * @internal No BC promises are given for this class. It may be changed or removed at any time.
  */
 interface PreviewInterface
```
alexander-schranz: we could remove the interfaces when we mark them as internal also

wachterjohannes: just remove it :) as this is already for 3.0

wachterjohannes (approved):

PR #7872, opened 2025-03-28
Fix UserProvider PHPStan Generic update for PasswordUpgraderInterface

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Remove generic definitions for PasswordUpgraderInterface as it was removed from Symfony core.

#### Why?

Not longer generic: https://github.com/symfony/symfony/pull/51283

### DIFF
--- a/src/Sulu/Bundle/SecurityBundle/User/UserProvider.php
+++ b/src/Sulu/Bundle/SecurityBundle/User/UserProvider.php
@@ -32,7 +32,6 @@ use Webmozart\Assert\Assert;
  * configuration from the webspaces into account.
  *
  * @implements UserProviderInterface<UserInterface>
- * @implements PasswordUpgraderInterface<User>
  */
 class UserProvider implements UserProviderInterface, PasswordUpgraderInterface
 {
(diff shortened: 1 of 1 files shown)


src/Sulu/Bundle/SecurityBundle/User/UserProvider.php
```diff
@@ -127,7 +126,7 @@ public function supportsClass($class): bool
 
     public function upgradePassword(PasswordAuthenticatedUserInterface $user, string $newHashedPassword): void
     {
-        if (!$user instanceof User) { // @phpstan-ignore-line we can not be 100% sure about this but generic forces us to check
```
alexander-schranz, line 130: Interesting is we didn't here either not trust the PHPStan generic and added a check already.

mamazu:
This is related to https://github.com/sulu/sulu/pull/7743

In more modern version Symfony doesn't need a salt and this means a different interface.

PR #7877, opened 2025-04-04
Add block visitors for new content storage

draconivis:
| Q | A
| --- | ---
| Bug fix? | no yes
| New feature? | no yes
| BC breaks? | no yes
| Deprecations? | no yes 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### To Do

### DIFF
--- a/packages/page/src/Infrastructure/Sulu/Content/PropertyResolver/BlockVisitor/SegmentBlockVisitor.php
+++ b/packages/page/src/Infrastructure/Sulu/Content/PropertyResolver/BlockVisitor/SegmentBlockVisitor.php
@@ -9,10 +9,10 @@
  * with this source code in the file LICENSE.
  */
 
-namespace Sulu\Component\Content\Types\Block;
+namespace Sulu\Page\Infrastructure\Sulu\Content\PropertyResolver\BlockVisitor;
 
-use Sulu\Component\Content\Compat\Block\BlockPropertyType;
 use Sulu\Component\Webspace\Analyzer\RequestAnalyzerInterface;
+use Sulu\Content\Application\PropertyResolver\BlockVisitor\BlockVisitorInterface;
 
 class SegmentBlockVisitor implements BlockVisitorInterface
 {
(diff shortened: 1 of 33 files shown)


packages/audienceTargeting/src/Infrastructure/Sulu/Content/PropertyResolver/BlockVisitor/TargetGroupBlockVisitor.php
alexander-schranz, line 1: this should be in src/Sulu/Bundle/Audienc....

alexander-schranz (approved):

alexander-schranz (changes requested):
We just need the new once which are used in the new block property resolver. The old aslong as the PHPCR content types are still there need to stay untouched (or atleast renamed to Deprecated if you have naming conflicts) as they are stilled used by the content types currently. We will remove them in an own PR when removing the content type interface.

packages/page/src/Infrastructure/Sulu/Content/PropertyResolver/BlockVisitor/SegmentBlockVisitor.php
```diff
@@ -9,20 +9,23 @@
  * with this source code in the file LICENSE.
  */
 
-namespace Sulu\Component\Content\Types\Block;
```
alexander-schranz: this should stay based on the old interface as used by the old block content type. So this are 2 seperated services. if you have service name conflicts prefix this old one with `sulu_page.deprecated_segement_block_visitor`

draconivis: i have kept both versions, just that i used `Old` instead of `Deprecated` for most of them. I can change that, but else i am not sure if i follow correctly on what the problem is, since old and new visitors are there

alexander-schranz: @draconivis looks like they are not longer injected into the block content type correctly because are not longer tagged. add the tag with the old prefix. PS: would only rename the class if they are in the same namespace else keep the old name

draconivis: @alexander-schranz i reverted some of the changes and added a tag to the deprecated visitors again.

alexander-schranz (approved):

PR #7888, opened 2025-04-10
Remove SnippetBundle and Snippet related classes

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? | no
| BC breaks? | no
| Deprecations? | no  
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Remove SnippetBundle and Snippet releated classes.

#### Why?

The old snippet bundle is replaced with the new storage.

#### To Do

- [x] Fix tests `vendor/bin/phpunit`
- [x] Fix Bundle tests `bin/runtests -i -a` https://github.com/sulu/sulu/pull/7901
- [ ] Fix registering of content types (do not longer validate type) :warning:  effects BC
- [ ] Fix registering of schema loader
- [x] Move Snippet JS to new Bundle https://github.com/sulu/sulu/pull/7891

Baseline adds some new error but all errors of classes which should be removed before the 3.0 release:

Bildschirmfoto 2025-04-17 um 11 43 07
s:


src/Sulu/Component/Content/Metadata/Loader/StructureXmlLoader.php
```diff
@@ -166,8 +166,13 @@ protected function parse($resource, \DOMXPath $xpath, $type)
             }
 
             $propertyType = $property->getType();
+
+            return true;
+
+            /*
+            // TODO maybe readd validation based on a JSON provided by admin build
```
alexander-schranz, line 173: Via the new pattern I think we are not longer able to validate at any point if a field type really exists, as we always have the fallback to a "default" property resolver which is used for text, text_line, and all other "simple" which not require any resolving mechanic.

Prokyonn:
@alexander-schranz I think we can merge this one, WDYT?

alexander-schranz:
@Prokyonn awesome 🎉

PR #7892, opened 2025-04-14
Add hint about jackrabbit

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Add hint about jackrabbit.

#### Why?

Make it clearer what for apache jackrabbit based projects the migration configuration is required and that jackrabbit is after the upgrade not longer needed. /cc @datenwerk-at

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -50,9 +50,9 @@ return [
 
 Configure the SuluPhpcrMigrationBundle in `config/packages/sulu_phpcr_migration.yaml`:
 
-> If you are currently using jackrabbit use the "jackrabbit://" based DSN string.
-> After the upgrade Apache Jackrabbit is not longer used by Sulu new content storage and can be
-> removed from your projects in most situation.
+> If you are currently using Jackrabbit, use the "jackrabbit://" based DSN string.
+> After the upgrade, Apache Jackrabbit is no longer used by Sulu’s new content storage and can be removed from
+> your projects in most situations.
 
 ```yaml
 sulu_phpcr_migration:
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -51,8 +51,8 @@ return [
 Configure the SuluPhpcrMigrationBundle in `config/packages/sulu_phpcr_migration.yaml`:
 
 > If you are currently using jackrabbit use the "jackrabbit://" based DSN string.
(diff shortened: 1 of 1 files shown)


UPGRADE.md
```diff
@@ -50,6 +50,10 @@ return [
 
 Configure the SuluPhpcrMigrationBundle in `config/packages/sulu_phpcr_migration.yaml`:
 
+> If you are currently using jackrabbit use the "jackrabbit://" based DSN string.
+> Apache Jackrabbit is after the upgrade not longer used by Sulu new content storage and can in
+> most situation uninstalled from your projects.
```
mamazu:
```suggestion
> After the upgrade Apache Jackrabbit is not longer used by Sulu new content storage and can be
> removed from your projects in most situation.
```

UPGRADE.md
```diff
@@ -50,6 +50,10 @@ return [
 
 Configure the SuluPhpcrMigrationBundle in `config/packages/sulu_phpcr_migration.yaml`:
 
+> If you are currently using jackrabbit use the "jackrabbit://" based DSN string.
+> After the upgrade Apache Jackrabbit is not longer used by Sulu new content storage and can be
+> removed from your projects in most situation.
```
```suggestion
> If you are currently using Jackrabbit, use the "jackrabbit://" based DSN string.
> After the upgrade, Apache Jackrabbit is no longer used by Sulu’s new content storage and can be removed from
> your projects in most situations.
```

Prokyonn (approved):

mamazu:
Would this note about how to configure the migration bundle also make sense in the migration bundle docs as well. I was trying out Sulu3 yesterday and was confused how this bundle should be set up (because there's no docs over there).

Prokyonn:
> Would this note about how to configure the migration bundle also make sense in the migration bundle docs as well. I was trying out Sulu3 yesterday and was confused how this bundle should be set up (because there's no docs over there).

@mamazu yes the documentation of the migration bundle should definitely be updated, I didn't have time yet to do it. But the update path with the configuration and the command should be part of the documentation.

alexander-schranz:
I think it make sense to copy the UPGRADE.md part into the README of the bundle.

PR #7893, opened 2025-04-16
Fix deprecated usage of execute method in doctrine query builder and statement

stollr:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

I have changed the call of `execute()` on Doctrine DBAL's QueryBuilder and Statement in favor of `executeStatement()` in all places that I could find. BC for DBAL v2 is kept.

#### Why?

The usage of the `execute` method in Doctrine DBAL's  [`QueryBuilder`](https://github.com/doctrine/dbal/blob/8b1e40a5027bda21296297a31d0f20bbde849513/src/Query/QueryBuilder.php#L375) and [`Statement`](https://github.com/doctrine/dbal/blob/8b1e40a5027bda21296297a31d0f20bbde849513/src/Statement.php#L170) classes was deprecated in v3.

### DIFF
--- a/src/Sulu/Bundle/CategoryBundle/Command/RecoverCommand.php
+++ b/src/Sulu/Bundle/CategoryBundle/Command/RecoverCommand.php
@@ -168,6 +168,12 @@ class RecoverCommand extends Command
                 WHERE ( c2.depth - 1 ) <> c1.depth';
 
         $statement = $this->entityManager->getConnection()->prepare($sql);
+
+        // DBAL 3 to 4 BC Layer
+        if (\method_exists($statement, 'executeStatement')) {
+            return $statement->executeStatement();
+        }
+
         $result = $statement->execute();
 
         if ($result) {
--- a/src/Sulu/Bundle/ContactBundle/Command/AccountRecoverCommand.php
+++ b/src/Sulu/Bundle/ContactBundle/Command/AccountRecoverCommand.php
@@ -167,6 +167,12 @@ class AccountRecoverCommand extends Command
                 WHERE ( c2.depth - 1 ) <> c1.depth';
 
         $statement = $this->entityManager->getConnection()->prepare($sql);
+
+        // DBAL 3 to 4 BC Layer
+        if (\method_exists($statement, 'executeStatement')) {
+            return $statement->executeStatement();
+        }
+
         $result = $statement->execute();
 
         if ($result) {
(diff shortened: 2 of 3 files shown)


src/Sulu/Bundle/CategoryBundle/Command/RecoverCommand.php
```diff
@@ -168,6 +168,12 @@ private function fixWrongDepthGap()
                 WHERE ( c2.depth - 1 ) <> c1.depth';
 
         $statement = $this->entityManager->getConnection()->prepare($sql);
+
+        // DBAL 3 to 4 BC Layer
+        if (\method_exists($statement, 'executeStatement')) {
+            return $statement->executeStatement();
```
alexander-schranz, line 174: is not the rowCount missing here?

src/Sulu/Bundle/ContactBundle/Command/AccountRecoverCommand.php
```diff
@@ -167,6 +167,12 @@ private function fixWrongDepthGap()
                 WHERE ( c2.depth - 1 ) <> c1.depth';
 
         $statement = $this->entityManager->getConnection()->prepare($sql);
+
+        // DBAL 3 to 4 BC Layer
+        if (\method_exists($statement, 'executeStatement')) {
+            return $statement->executeStatement();
```
alexander-schranz, line 173: does this include the rowCount result? see line 180

src/Sulu/Bundle/CategoryBundle/Command/RecoverCommand.php
stollr: `executeStatement` returns the number of the affected rows. So it is equivalent to the previous rowCount.

src/Sulu/Bundle/ContactBundle/Command/AccountRecoverCommand.php
stollr: Same as the other comment.

src/Sulu/Bundle/CategoryBundle/Command/RecoverCommand.php
stollr: Here's the link to [`executeStatement`](https://github.com/doctrine/dbal/blob/c6e996147013de567b8a57c0cc9a253a91cb36a9/src/Statement.php#L249)

alexander-schranz (approved):

PR #7900, opened 2025-04-16
Fix passing null to non-nullable argument in UserManager

stollr:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no 
| License | MIT

#### What's in this PR?

In the UserManager's `processEmail` method I have added a check for null-ness of the user's email before it is passed to PHP's `strcasecmp` function.

Additionally I added two unit test to test the UserManager's save method.

#### Why?

`strcasecmp` does not accept parameter values of type NULL. Passing NULL to non-nullable parameters of built-in functions was deprecated in PHP 8.1 (see [here](https://www.php.net/manual/en/migration81.deprecated.php)).

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -27502,7 +27502,7 @@ parameters:
 
 		-
 			message: "#^Call to an undefined method Sulu\\\\Component\\\\Security\\\\Authentication\\\\UserInterface\\:\\:getEmail\\(\\)\\.$#"
-			count: 1
+			count: 2
 			path: src/Sulu/Bundle/SecurityBundle/UserManager/UserManager.php
 
 		-
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -27502,7 +27502,7 @@ parameters:
 
 		-
 			message: "#^Call to an undefined method Sulu\\\\Component\\\\Security\\\\Authentication\\\\UserInterface\\:\\:getEmail\\(\\)\\.$#"
-			count: 2
+			count: 1
 			path: src/Sulu/Bundle/SecurityBundle/UserManager/UserManager.php
 
 		-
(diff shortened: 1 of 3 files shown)


src/Sulu/Bundle/SecurityBundle/UserManager/UserManager.php
```diff
@@ -685,7 +685,9 @@ private function processEmail(UserInterface $user, $email, $contact = null)
                 $user->setEmail($email);
             }
         } else {
-            if ($email && 0 !== \strcasecmp($email, $user->getEmail()) && !$this->isEmailUnique($email)) {
+            $emailChanged = $email && (null === $user->getEmail() || 0 !== \strcasecmp($email, $user->getEmail()));
+
+            if ($emailChanged && !$this->isEmailUnique($email)) {
```
alexander-schranz: Would avoid the duplicated email getEmail call and just go with `??`:

```suggestion
            $currentEmail = $user->getEmail() ?? '';
            if ($email && 0 !== \strcasecmp($email, $currentEmail) && !$this->isEmailUnique($email)) {
```

stollr: I have added the `$emailChanged` variable, because it was not clear to me what `$email && 0 !== \strcasecmp($email, $currentEmail)` means at first glance.

Second point: `$user->getEmail() ?? ''` works, but it is not very explicit.

alexander-schranz: Then would go this way:

```suggestion
            $currentEmail = $user->getEmail() ?? '';
            $hasEmailChanged = $email && 0 !== \strcasecmp($email, $currentEmail);

            if ($hasEmailChanged && !$this->isEmailUnique($email)) {
```

phpstan-baseline.neon
```diff
@@ -27627,7 +27627,7 @@ parameters:
 
 		-
 			message: "#^Call to an undefined method Sulu\\\\Component\\\\Security\\\\Authentication\\\\UserInterface\\:\\:getEmail\\(\\)\\.$#"
-			count: 1
+			count: 2
```
alexander-schranz:
```suggestion
			count: 1
```

stollr:
@alexander-schranz I have commited your suggestion

alexander-schranz:
@stollr thank you!

PR #7906, opened 2025-04-21
Ignore non existing sort columns

### LINKED ISSUE #7207: Ignore not existing columns in the ListBuilder

mamazu:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #7207
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?
The code changes proposed in the issue. :laughing:

Thanks @MrSrsen

#### Why?
Ignoring the sort query parameter when there is no column in the database with that names reduces the amount of errors.

This also helps with view switching as the filter settings when switching between views remain (even though the new view might not have those columns).

### DIFF
--- a/src/Sulu/Component/Rest/RestHelper.php
+++ b/src/Sulu/Component/Rest/RestHelper.php
@@ -76,7 +76,7 @@ class RestHelper implements RestHelperInterface
         }
 
         $sortBy = $this->listRestHelper->getSortColumn();
-        if (null != $sortBy) {
+        if (null != $sortBy && \array_key_exists($sortBy, $fieldDescriptors)) {
             $listBuilder->sort($fieldDescriptors[$sortBy], $this->listRestHelper->getSortOrder());
         }
     }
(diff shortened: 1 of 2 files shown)


alexander-schranz:
@mamazu please do not use `pull` use always `rebase` when update merge requests. Pull makes rebasing hard and can end in errors during squash merge.

mamazu:
@alexander-schranz Can you check the project settings. I think that should be a setting you can change there that it is updated with a rebase. But I can also recommit it without the merge.

mamazu:
The failing test seems really unrelated?

alexander-schranz:
@mamazu there seems no option to disable the normal Update you just need to use the `rebase` button instead of the Update Button:

Bildschirmfoto 2025-08-20 um 13 37 38

PR #7908, opened 2025-04-22
Move sulu core structure config to sulu admin config

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? |  yes
| BC breaks? | no 
| Deprecations? | yes 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Move sulu core structure config to sulu admin config.

#### Why?

We should get rid of structure specific config and make all kind of metadata based on Sulu Admin service nothinger else.

#### Example Usage

Current: https://docs.sulu.io/en/2.6/cookbook/override-dir-structure.html#overwrite-templates-configuration-files-path

Not sure yet. ~~Open the PR to discuss this Option `A`:~~

```yaml
sulu_admin:
    forms:
        - '%kernel.project_dir%/config/forms'
    lists:
        - '%kernel.project_dir%/config/lists'
    templates:
        default_type:
            page: default
        paths:
            page_skeleton: # we use a key already to make things overwrite able here
                path: '%kernel.project_dir%/config/templates/pages'
                type: page # not sure if we should go with pages now
```

alternative `B`:

```yaml
sulu_admin:
    forms:
        - '%kernel.project_dir%/config/forms'
    lists:
        - '%kernel.project_dir%/config/lists'
    templates:
        page:  # better use resource key here?
            default_type: default
            paths:
                page_skeleton: '%kernel.project_dir%/config/templates/pages'
```

/cc @Prokyonn 

This introduces a new parameter `sulu_admin.templates.configuration` which we can later use in the new loader:

```bash
bin/console debug:container --parameter sulu_admin.templates.configuration --format=json
```

#### ToDo

 - [x] Write parameter
 - [ ] Parse XMLs (upcoming Pull Request)


Prokyonn:
Personally I would go with variant `B`, as this is more readable and easier to extend because you immediately see, which parameters are dependant on each other, but it will be a bigger BC break and more difficult to migrate

alexander-schranz:
Think the migration is also for variant B kind of fine in both cases they need modify a few keys and at the end the container is only valid if they give a valid parts in it.

Migration for Variant B would look like this:

```diff
-sulu_core:
+sulu_admin:
-    content:
+    templates:
-         structure:
+         pages:
             paths:
-                 page_projectA:
-                     path: '%kernel.project_dir%/config/templates/pages/projectA'
-                     type: page
+                 page_projectA: '%kernel.project_dir%/config/templates/pages/projectA'
-                 page_projectB:
-                     path: '%kernel.project_dir%/config/templates/pages/projectB'
-                     type: page
+                 page_projectB: '%kernel.project_dir%/config/templates/pages/projectB'
```

src/Sulu/Bundle/AdminBundle/DependencyInjection/Configuration.php
```diff
@@ -76,6 +76,23 @@ public function getConfigTreeBuilder(): TreeBuilder
...
                 ->end()
+                ->arrayNode('templates')
+                    ->useAttributeAsKey('name')
+                    ->prototype('array')
+                        ->children()
+                            ->scalarNode('default_type')
+                                ->defaultValue(null)
+                                ->example('default')
+                            ->end()
+                            ->arrayNode('directories')
```
alexander-schranz, line 87: @Prokyonn did go with directories like we already did for `forms` and `lists`

packages/article/src/Infrastructure/Symfony/HttpKernel/SuluArticleBundle.php
```diff
@@ -274,6 +274,14 @@ public function prependExtension(ContainerConfigurator $container, ContainerBuil
                             // \dirname(__DIR__, 4) . '/config/forms',
                         ],
                     ],
+                    'templates' => [
+                        'articles' => [
```
alexander-schranz, line 278: did go with the resource key now

src/Sulu/Bundle/AdminBundle/DependencyInjection/SuluAdminExtension.php
```diff
@@ -144,16 +144,17 @@ public function load(array $configs, ContainerBuilder $container): void
         $configuration = $this->getConfiguration($configs, $container);
         $config = $this->processConfiguration($configuration, $configs);
 
-        $container->setParameter($this->getAlias() . '.name', $config['name']);
```
alexander-schranz, line 147: removed the `getAlias()` and used hardcoded value to make the parameters easier to find with searches

Prokyonn (approved):

src/Sulu/Bundle/AdminBundle/DependencyInjection/SuluAdminExtension.php
```diff
@@ -144,16 +144,17 @@ public function load(array $configs, ContainerBuilder $container): void
...
+        $container->setParameter('sulu_admin.email', $config['email']);
+        $container->setParameter('sulu_admin.user_data_service', $config['user_data_service']);
+        $container->setParameter('sulu_admin.resources', $config['resources']);
+        $container->setParameter('sulu_admin.collaboration_enabled', $config['collaboration']['enabled']);
+        $container->setParameter('sulu_admin.collaboration_interval', $config['collaboration']['interval']);
+        $container->setParameter('sulu_admin.collaboration_threshold', $config['collaboration']['threshold']);
+
+        $container->setParameter('sulu_admin.forms.directories', $config['forms']['directories'] ?? []);
+        $container->setParameter('sulu_admin.lists.directories', $config['lists']['directories'] ?? []);
+        $container->setParameter('sulu_admin.templates.configuration', $config['templates']);
```
Prokyonn, line 157: here is some inconsistency with the parameter names

`$config['collaboration']['enabled']` translates to `sulu_admin.collaboration_enabled`
but 
`$config['forms']['directories']` translates to `sulu_admin.forms.directories`

alexander-schranz: Really strange not sure why this is. I know that when 2.0 was build it was defined that service name should always be `<bundle>.<service>` and so no more as one dot should be inside them. Strange this wasn't done for the parameters here.

PR #7913, opened 2025-04-24
Fix detecting of content-types which are not available

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | 
| Related issues/PRs | 
| License | MIT
| Documentation PR |

#### What's in this PR?

This fixes the sulu:build command in the skeleton.

#### Why?

The audience taget group selection breaks the sulu:build if the bundle is not enabled.


Prokyonn (approved):

alexander-schranz:
As discussed the Page bundle should not register this. The audience targeting bundle should extend the excerpt if its registered. I probably would go with remove from the xml schema `on-invalid`. As with the new property resolver we not longer able to detect such things.

PR #7920, opened 2025-04-27
Grouping test output by Bundle in Github Actions

mamazu:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

image

#### Why?
With the amount of output the tests produce you have to scroll a lot when the first components fail. And the search isn't helping 'cause it only sometimes works.

This way you can just expand the section(s) that are broken.

### DIFF
--- a/.github/workflows/test-application.yaml
+++ b/.github/workflows/test-application.yaml
@@ -8,6 +8,9 @@ on:
             - '[0-9]+.[0-9]+'
     workflow_dispatch:
 
+env:
+    CI_GROUP_OUTPUT: true
+
 # automatically cancel previously started workflows when pushing a new commit to a branch
 concurrency:
     group: ${{ github.workflow }}-${{ github.ref }}
--- a/bin/runtests
+++ b/bin/runtests
@@ -352,11 +352,15 @@ if ($target = $input->getOption('target')) {
 }
 
 write_header('Bundle tests');
-
+$enableGrouping = getenv('CI_GROUP_OUTPUT') === 'true';
 foreach ($bundles as $bundle) {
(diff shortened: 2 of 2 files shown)


mamazu:
The output should only occur on CI, so we should add an Environment variable and check for it.

.github/workflows/test-application.yaml
```diff
@@ -8,6 +8,9 @@ on:
             - '[0-9]+.[0-9]+'
     workflow_dispatch:
 
+env:
+    IS_GITHUB_ACTION: 1
```
alexander-schranz: would more go with explicit name like `CI_GROUP_OUTPUT=1`

else we would not require to define an own variable as I think github would always set `GITHUB_ACTIONS=true`: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables 

I'm fine with both changes.

mamazu:
@alexander-schranz Looks weird with the true, but still works.

alexander-schranz (approved):

bin/runtests
```diff
@@ -351,9 +351,18 @@ if ($target = $input->getOption('target')) {
...
+    }
     run_bundle_tests($bundle);
+    if ($enableGrouping) {
+        echo "::endgroup::\n";
+    }
 }
+echo "\n";
 
 if (false === $input->getOption('no-component-tests') || $input->getOption('all')) {
     write_header('Components');
```
alexander-schranz, line 368: @mamazu we should put the components also into an own collapsed group, else its a little bit confusing currently.

PR #7921, opened 2025-04-28
Add some more hints to the 3.0 upgrade and move data migration down

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no 
| BC breaks? | no 
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Add some more hints to the 3.0 upgrade and move data migration down.

#### Why?

Some feedback from our Partner Tailr (@phpro) was that the datamigration was the early before the other changes which might end in a none working container. So we moved the whole data migration part ot the bottom, as all above could end in a invalid container. I also added `--no-scripts` to the early commands so it doesn't try to create the container.

Also added a hint about which SQL statements can not be auto generated. /cc @Prokyonn please recheck that parts.

### DIFF
--- a/UPGRADE-3.x.md
+++ b/UPGRADE-3.x.md
@@ -22,8 +22,8 @@ Ensure that the [Version202407111600](https://github.com/sulu/SuluArticleBundle/
 structure to the new one. After that you can remove the old article bundle from your code.
 
 ```shell
-composer remove sulu/article-bundle
-# composer remove elasticsearch/elasticsearch # can also be removed if not further required
+composer remove sulu/article-bundle --no-scripts
+# composer remove elasticsearch/elasticsearch --no-scripts # can also be removed if not further required
 ```
 
 Disable the old SuluArticleBundle `config/bundles.php`:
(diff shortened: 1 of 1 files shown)


UPGRADE-3.x.md
```diff
@@ -190,55 +253,142 @@ DELETE FROM se_user_settings WHERE settingsKey LIKE 'sulu_admin.list_store.snipp
...
+    storage:
+        flysystem_service: 'default.storage' # this is default and not required to be configured.
+```
+
+This will only create the service `sulu_media.storage` as the alias to `sulu_media.storage.*` services has been removed.
+
+### Preview Services changed
+
+Most of the PreviewBundle services are now internal and only the new `PreviewDefaultsProviderInterface` is
+now the way to communicate with the Bundle.
```
Prokyonn:
```suggestion
Most of the services in the PreviewBundle are now internal. The new `PreviewDefaultsProviderInterface` is
now the primary means of interacting with the bundle.
```

Prokyonn:
@alexander-schranz thinking about it.. should we also add to the pre update section the phpcr cleanup command? I think it would make sense to do a cleanup before the migration, this will speed up the migration and also not migrate unused properties 🤔

Prokyonn (approved):

PR #7922, opened 2025-04-28
Fix return empty array if teaser selection property resolver data not matching

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no 
| Deprecations? | no 
| Fixed tickets | fixes # 
| License | MIT

#### What's in this PR?

Return always empty array for teaser selection if unexpected data was given the the property resolver.

#### Why?

If I switch a block type or change property xmls without a migration a teaser selection should return a empty array instead of the eventually false raw data of another type.

Currently this code snippet can error:

```twig
{% for teaser in content.teasers %}
```

as it might return null, string or something else.

### DIFF
--- a/packages/content/src/Application/PropertyResolver/Resolver/TeaserSelectionPropertyResolver.php
+++ b/packages/content/src/Application/PropertyResolver/Resolver/TeaserSelectionPropertyResolver.php
@@ -32,7 +32,7 @@ class TeaserSelectionPropertyResolver implements PropertyResolverInterface
             || !\array_key_exists('items', $data)
             || !\is_array($data['items'])
         ) {
-            return ContentView::create($data, $returnedParams);
+            return ContentView::create([], $returnedParams);
         }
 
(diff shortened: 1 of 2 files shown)


packages/content/src/Application/PropertyResolver/Resolver/TeaserSelectionPropertyResolver.php
```diff
@@ -32,7 +32,7 @@ public function resolve(mixed $data, string $locale, array $params = []): Conten
             || !\array_key_exists('items', $data)
             || !\is_array($data['items'])
         ) {
-            return ContentView::create($data, $returnedParams);
```
Prokyonn, line 35: Can you also add a small test for this in the TeaserSelectionPropertyResolverTest ?

Prokyonn (approved):

packages/content/src/Application/PropertyResolver/Resolver/TeaserSelectionPropertyResolver.php
alexander-schranz: @Prokyonn good catch 👍  added

PR #7923, opened 2025-04-28
Replace old metadata loader in block and image map property resolvers

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no yes
| New feature? | no yes
| BC breaks? | no yes
| Deprecations? | no yes 
| Fixed tickets | fixes # 
| License | MIT

#### What's in this PR?

Replace old metadata loader in block and image map property resolvers.

#### Why?

The old loader may be removed we should use the service which we use already in the admin controller.

### DIFF
--- a/packages/content/src/Application/PropertyResolver/Resolver/BlockPropertyResolver.php
+++ b/packages/content/src/Application/PropertyResolver/Resolver/BlockPropertyResolver.php
@@ -16,8 +16,8 @@ namespace Sulu\Content\Application\PropertyResolver\Resolver;
 use Psr\Log\LoggerInterface;
 use Sulu\Bundle\AdminBundle\Metadata\FormMetadata\FieldMetadata;
 use Sulu\Bundle\AdminBundle\Metadata\FormMetadata\FormMetadata;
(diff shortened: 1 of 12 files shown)


packages/content/src/Application/PropertyResolver/Resolver/BlockPropertyResolver.php
```diff
@@ -56,8 +56,11 @@ public function resolve(mixed $data, string $locale, array $params = []): Conten
         \assert($metadata instanceof FieldMetadata, 'Metadata must be set to resolve blocks.');
         $metadataTypes = $metadata->getTypes();
 
-        /** @var TypedFormMetadata $typedFormMetadata */
-        $typedFormMetadata = $this->formMetadataLoader->getMetadata('block', $locale, []);
+        $typedFormMetadata = $this->metadataProviderRegistry->getMetadataProvider('form')
+            ->getMetadata('block', $locale, []);
+
+        \assert($typedFormMetadata instanceof TypedFormMetadata, 'Block form metadata not found for image map resolving.');
```
Prokyonn:
```suggestion
        \assert($typedFormMetadata instanceof TypedFormMetadata, 'Block form metadata not found for block resolving.');
```

Prokyonn (approved):

PR #7925, opened 2025-04-28
Remove reference handling of tags in properties parser

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no 
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Remove reference handling of tags in properties parser. And make the services internal.

#### Why?

The tags reference handling is really changed and valdiation and loading should be seperated, to be open for future changes we should also mark such service as internal as its loading should be in our full control.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/Parser/PropertiesXmlParser.php
+++ b/src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/Parser/PropertiesXmlParser.php
@@ -20,7 +20,7 @@ use Sulu\Component\Content\Metadata\SectionMetadata;
 use Symfony\Contracts\Translation\TranslatorInterface;
 
 /**
- * Parse properties structure from an XML file.
+ * @internal this class is not part of the public API and may be changed or removed without further notice
  */
 class PropertiesXmlParser
 {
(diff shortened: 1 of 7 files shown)


src/Sulu/Component/Content/Metadata/Loader/StructureXmlLoader.php
```diff
@@ -187,18 +182,6 @@ protected function parse($resource, \DOMXPath $xpath, $type)
             */
         });
 
-        // FIXME until excerpt-template is no page template anymore
-        // - https://github.com/sulu-io/sulu/issues/1220#issuecomment-110704259
-        if (!\array_key_exists('internal', $result) || !$result['internal']) {
-            if (isset($this->requiredTagNames[$type])) {
-                foreach ($this->requiredTagNames[$type] as $requiredTagName) {
-                    if (!\array_key_exists($requiredTagName, $tags)) {
-                        throw new RequiredTagNotFoundException($result['key'], $requiredTagName);
```
alexander-schranz, line 196: We should readd this but not inside the loader.

src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/Parser/PropertiesXmlParser.php
```diff
@@ -20,7 +20,7 @@
 use Symfony\Contracts\Translation\TranslatorInterface;
 
 /**
- * Parse properties structure from an XML file.
+ * @internal this class is not part of the public API and may be changed or removed without further notice
```
alexander-schranz, line 23: this should make it in future easier if we need again changes in the parser or loader as we don't longer need to keep bc in mind.

PR #7930, opened 2025-04-29
Make new metadata locale independent to contain all data

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Make new metadata not locale aware.

#### Why?

We want to remove the old metadata classes but to avoid that the metadata is loaded per locale again we need 

With the removal of the old metadata classes xml would be parsed per locale to avoid this we change the metadata to keep all admin locales in them.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/FormMetadata.php
+++ b/src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/FormMetadata.php
@@ -25,10 +25,10 @@ class FormMetadata extends AbstractMetadata
     private $name;
 
     /**
-     * @var string
+     * @var array<string, string>
      */
-    #[Exclude(if: "'admin_form_metadata_keys_only' in context.getAttribute('groups')")]
-    private $title;
+    #[Exclude]
+    private $titles = [];
 
     /**
      * @var ItemMetadata[]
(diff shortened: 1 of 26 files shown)


src/Sulu/Bundle/AdminBundle/Tests/Functional/Controller/fixtures/default.json
```diff
@@ -1,12 +1,9 @@
 {
     "name": "default",
-    "title": "Animals",
     "form": {
         "title": {
-            "label": "Animals",
             "disabledCondition": null,
             "visibleCondition": null,
-            "description": "",
```
alexander-schranz, line 9: small change here empty descriptions, title, label. infotests, placeholders ... are not longer returned.

src/Sulu/Bundle/AdminBundle/Tests/Functional/Controller/fixtures/default.json
```diff
@@ -1,12 +1,9 @@
 {
     "name": "default",
-    "title": "Animals",
     "form": {
```
alexander-schranz, line 3: we can not set the position of a property added via subscriber this are just moved down

src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/FormMetadata.php
```diff
@@ -80,14 +80,36 @@ public function getKey(): string
...
+    {
+        $this->titles[$locale] = $title;
+    }
+
+    /**
+     * @param array<string, string> $titles
+     */
+    public function setTitles(array $titles)
+    {
+        $this->titles = $titles;
```
Prokyonn, line 93: should we check here, that the key is always a string to prevent strange sideeffects later?

Prokyonn (approved):

Prokyonn:
As we are already touching all of these files, wouldn't it make sense to remove the php doc types and add them as real php types? @alexander-schranz

alexander-schranz:
@Prokyonn agree but would do that parts after the whole refactoring when we know which classes will stay.

src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/FormMetadata.php
alexander-schranz: as discussed the callers responsible to validate that it is a `array<string, string>` in our case mostly the xml parsers work.

PR #7928, opened 2025-04-28
WIP: Refactor properties xml loader to return new metadata

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT
| Documentation PR | sulu/sulu-docs# 

#### What's in this PR?

Refactor properties xml loader to return new metadata classes instead of old deprecated.

#### Why?

We want to remove the old metadata classes and services for 3.0 as they are build on top of some old phpcr content services.

PR #7934, opened 2025-04-30
Add template metadata to form metadata

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | fixes # 
| License | MIT

#### What's in this PR?

Add template metadata to form metadata.

#### Why?

We want to get rid of the structure for this we need to missing properties:

https://github.com/sulu/sulu/blob/33e8b72844761ffc7483a3a95649445c0f48c424/src/Sulu/Bundle/TestBundle/Resources/structures/pages/default.xml#L10-L12

to the new metadata. So in future we can use the template metadata instead of the structure or old metadata classes.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -2167,6 +2167,12 @@ parameters:
 			path: src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/FormMetadata.php
 
 		-
+			message: '#^Method Sulu\\Bundle\\AdminBundle\\Metadata\\FormMetadata\\FormMetadata\:\:setTemplate\(\) has no return type specified\.$#'
+			identifier: missingType.return
+			count: 1
+			path: src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/FormMetadata.php
+
+		-
 			message: '#^Method Sulu\\Bundle\\AdminBundle\\Metadata\\FormMetadata\\FormMetadata\:\:setTitle\(\) has no return type specified\.$#'
 			identifier: missingType.return
 			count: 1
--- a/src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/CacheLifetimeMetadata.php
+++ b/src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/CacheLifetimeMetadata.php
@@ -1,0 +1,45 @@
+<?php
+
+/*
+ * This file is part of Sulu.
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\AdminBundle\Metadata\FormMetadata;
+
+class CacheLifetimeMetadata
+{
+    private string $type;
+
+    private string $value;
+
+    public function __construct(string $type, string $value)
+    {
+        $this->type = $type;
+        $this->value = $value;
+    }
+
+    public function getType(): string
+    {
+        return $this->type;
+    }
+
+    public function setType(string $type): void
+    {
+        $this->type = $type;
+    }
+
+    public function getValue(): string
+    {
+        return $this->value;
+    }
+
+    public function setValue(string $value): void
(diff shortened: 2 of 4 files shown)


phpstan-baseline.neon
alexander-schranz, line 1: as discussed with @Prokyonn we will do a whole refactoring of the return types and metadata classes when removing all old classes

src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/CacheLifetimeMetadata.php
```diff
@@ -0,0 +1,45 @@
...
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\AdminBundle\Metadata\FormMetadata;
+
+class CacheLifetimeMetadata
+{
+    private string $type;
+
+    private string $value;
```
Prokyonn, line 18: Is string required due to the xml parsing or what is the reason for not using directly int?

src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/TemplateMetadata.php
```diff
@@ -0,0 +1,55 @@
...
+    {
+        return $this->view;
+    }
+
+    public function setView(?string $view): void
+    {
+        $this->view = $view;
+    }
+
+    public function getCacheLifetime(): ?CacheLifetimeMetadata
```
Prokyonn, line 46: 🤔 I’m not sure how I feel about this, on one hand, I would expect `getCacheLifetime` to return a scalar value directly, not an object. On the other hand, `getCacheLifetimeMetadata` is quite a long method name.

src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/CacheLifetimeMetadata.php
alexander-schranz: The resolver is reponsible to convert the string then into the correct integer value from metadata point it can be something like `* 3 * * *` (type="expression") or `2400` (type="seconds").

src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/TemplateMetadata.php
alexander-schranz: cachelifeime does have 2 types:

```xml
    <cacheLifetime type="expression">0 8 * * *</cacheLifetime>
```


```xml
    <cacheLifetime type="seconds">2400</cacheLifetime>
```

The [cachelifetimeresolver](https://github.com/sulu/sulu/blob/33e8b72844761ffc7483a3a95649445c0f48c424/src/Sulu/Bundle/HttpCacheBundle/CacheLifetime/CacheLifetimeResolver.php#L20) is responsible to convert the string value into a integer. So we can not return integer value here.

alexander-schranz: did choice `getCacheLifetime` instead of a postfix as it is also `getSchema` not `getSchemaMetadata` method, think the idea behind is that all inside that namespace is automatically Metadata (getForms, getItems, ...)

src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/CacheLifetimeMetadata.php
Prokyonn: 👍

src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/TemplateMetadata.php
Prokyonn: Yes that sounds reasonable 👍  thanks for the explanation

Prokyonn (approved):

PR #7936, opened 2025-05-02
Make backwards compatibility virtual parameters real

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | yes
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?
Making optional Parameters in 2.6 now required.

#### Why?
PHPstan does not like when the parameters aren't officially there.

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/Media/FormatManager/FormatManager.php
+++ b/src/Sulu/Bundle/MediaBundle/Media/FormatManager/FormatManager.php
@@ -87,8 +87,8 @@ class FormatManager implements FormatManagerInterface
                 throw new ImageProxyMediaNotFoundException(\sprintf('Media with id "%s" was not found.', $id));
             }
 
-            $latestFileVersion = $this->getLatestFileVersion($media);
-            $version ??= $latestFileVersion->getVersion(); // TODO remove this line in Sulu 3.0 currently bc layer when version is not given
+            $fileVersion = $this->getLatestFileVersion($media);
+            $version = null !== $version ? $version : $fileVersion->getVersion(); // TODO remove this line in Sulu 3.0 currently bc layer when version is not given
 
             /** @var File|null $file */
             $file = $media->getFiles()[0] ?? null;
@@ -104,15 +104,8 @@ class FormatManager implements FormatManagerInterface
                 throw new ImageProxyMediaNotFoundException(\sprintf('FileVersion "%s" for media with id "%s" was not found.', $requestedFileVersion->getVersion(), $id));
             }
 
-            // Generate a redirect for old versions
-            if ($latestFileVersion->getVersion() !== $requestedFileVersion->getVersion()) {
-                $formats = $this->getFormats(
-                    $id,
-                    $latestFileVersion->getName(),
-                    $latestFileVersion->getVersion(),
-                    $latestFileVersion->getSubVersion(),
-                    $latestFileVersion->getMimeType(),
-                );
+            if ($fileVersion->getVersion() !== $requestedFileVersion->getVersion()) {
+                $formats = $this->getFormats($id, $fileVersion->getName(), $fileVersion->getVersion(), $fileVersion->getSubVersion(), $fileVersion->getMimeType());
 
                 $formatUrl = $formats[$formatKey . '.' . $imageFormat] ?? null;
                 if (null === $formatUrl) {
@@ -122,10 +115,9 @@ class FormatManager implements FormatManagerInterface
                 return new RedirectResponse($formatUrl, 301);
             }
 
-            // Generate the current version
-            $supportedImageFormats = $this->converter->getSupportedOutputImageFormats($latestFileVersion->getMimeType());
+            $supportedImageFormats = $this->converter->getSupportedOutputImageFormats($fileVersion->getMimeType());
             if (empty($supportedImageFormats)) {
-                throw new InvalidMimeTypeForPreviewException($latestFileVersion->getMimeType() ?? '-null-');
+                throw new InvalidMimeTypeForPreviewException($fileVersion->getMimeType() ?? '-null-');
             }
 
             if (!\in_array($imageFormat, $supportedImageFormats)) {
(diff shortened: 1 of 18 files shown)


src/Sulu/Bundle/MediaBundle/Media/FormatManager/FormatManager.php
```diff
@@ -68,15 +68,8 @@ public function __construct(
         $this->logger = $logger ?: new NullLogger();
     }
 
-    public function returnImage($id, $formatKey, $fileName /*, int<1, max>|null $version = null */)
+    public function returnImage($id, $formatKey, $fileName, ?int $version)
```
mamazu: This parameter should not be nullable, but seeing that this would need a refactor of the `FormatManager` I've kept it nullable.

For Context: If the version is null then it will return the latest version (which is what we use in the `FormatCacheRegenerate`) If the version is not set to the current version, then it returns a redirect (very coupled with http). So maybe this method should be split up.

src/Sulu/Bundle/TagBundle/Tests/Unit/Search/TagsConverterTest.php
```diff
@@ -23,16 +23,6 @@ class TagsConverterTest extends TestCase
...
-    public function testConvert(): void
-    {
-        $tagManager = $this->prophesize(TagManager::class);
-        $tagsConverter = new TagsConverter($tagManager->reveal());
-
-        $tagManager->resolveTagNames(['Tag1', 'Tag2', 'Tag3'])->willReturn([1, 2, 3]);
-
-        $this->assertEquals([1, 2, 3], $tagsConverter->convert(['Tag1', 'Tag2', 'Tag3']));
-    }
-
```
mamazu: This behaviour has been removed. Since we always have to pass 2 parameters to the function and then the output has a completely different shape.

src/Sulu/Bundle/MediaBundle/Media/FormatManager/FormatManager.php
alexander-schranz: no its the behaviour we want. if no specific version is requested we returning the last one. Its fine this method is null.


```suggestion
    public function returnImage($id, $formatKey, $fileName, ?int $version = null)
```

src/Sulu/Bundle/MediaBundle/Media/FormatManager/FormatManager.php
```diff
@@ -94,8 +87,8 @@ public function returnImage($id, $formatKey, $fileName /*, int<1, max>|null $ver
                 throw new ImageProxyMediaNotFoundException(\sprintf('Media with id "%s" was not found.', $id));
             }
 
-            $fileVersion = $this->getLatestFileVersion($media);
-            $version = null !== $version ? $version : $fileVersion->getVersion(); // TODO remove this line in Sulu 3.0 currently bc layer when version is not given
+            $latestFileVersion = $this->getLatestFileVersion($media);
+            $version ??= $latestFileVersion->getVersion(); // TODO remove this line in Sulu 3.0 currently bc layer when version is not given
```
alexander-schranz: PS: refacotring methods in pull request which have a different target make reviews harder, we should do refactorings in own PRs and not part of such other PRs.

src/Sulu/Bundle/MediaBundle/Media/FormatManager/FormatManagerInterface.php
```diff
@@ -35,7 +35,7 @@ interface FormatManagerInterface
      *
      * @return Response
      */
-    public function returnImage($id, $formatKey, $imageFormat /*, int<1, max>|null $version = null */);
+    public function returnImage($id, $formatKey, $imageFormat, ?int $version);
```
alexander-schranz: same as above:


```suggestion
    public function returnImage($id, $formatKey, $imageFormat, ?int $version = null);
```

src/Sulu/Bundle/MediaBundle/Media/ImageConverter/ImagineImageConverter.php
```diff
@@ -93,9 +93,12 @@ public function getSupportedOutputImageFormats(?string $mimeType): array
...
     public function convert(FileVersion $fileVersion, $formatKey, $imageFormat)
     {
+        $rawResource = $this->storage->load($fileVersion->getStorageOptions());
+        /** @var string $mimeType */
+        $mimeType = \mime_content_type($rawResource);
         $imageResource = $this->mediaImageExtractor->extract(
-            $this->storage->load($fileVersion->getStorageOptions()),
-            $fileVersion->getMimeType()
+            $rawResource,
+            $fileVersion->getMimeType() ?? $mimeType,
```
alexander-schranz: this can really end in hard errors, resources here arenot able to call `mime_content_type` always as they are in a few scenerries streams. Also a FileVersion without a mimetype should never exist.

So this change should be reverted.

```suggestion
$imageResource = $this->mediaImageExtractor->extract(
            $this->storage->load($fileVersion->getStorageOptions()),
            $fileVersion->getMimeType(),
```

src/Sulu/Bundle/MediaBundle/Tests/Unit/Command/FormatCacheRegenerateCommandTest.php
```diff
@@ -36,13 +36,13 @@ public function setUp(): void
 
     public function testExecute(): void
     {
-        $this->formatManager->returnImage(1, '50x50', 'test.svg')
+        $this->formatManager->returnImage(1, '50x50', 'test.svg', null)
```
alexander-schranz: revert as yes this is optional and return current format.

alexander-schranz (changes requested):
@mamazu Thx for the PR, a few small change requests.

src/Sulu/Bundle/MediaBundle/Media/ImageConverter/ImagineImageConverter.php
mamazu: Reverted, but well the return type of the function is nullable. So I assumed that this is not 100% guaranteed.

alexander-schranz (approved):

mamazu:
Should I also add this error to the baseline?

alexander-schranz:
@mamazu did handle it the same way as in the FormatManager and added a comment that its known the resource can be none seekable. 

Thx for the changes!

PR #7940, opened 2025-05-02
Remove deprecated `WebspaceController`

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | yes
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | #7547
| License | MIT
| Documentation PR | -

#### What's in this PR?
Removing `WebspaceController`.

#### Why?
It's part of the `AdminController::config`


mamazu:
This also creates BC breaks in the analytics controller as the parent route doesn't exist anymore.

src/Sulu/Bundle/WebsiteBundle/Controller/AnalyticsController.php
```diff
@@ -99,13 +96,12 @@ public function getAction($webspace, $id)
...
      *
-     * @param string $webspace
-     *
      * @return Response
      */
-    public function postAction(Request $request, $webspace)
+    public function postAction(Request $request)
     {
         $data = $request->request->all();
+        $webspace = $request->attributes->getString('webspace');
```
alexander-schranz: doesnt make sense to change this to attributes->getString from my point of view. We should if possible keep the `, $webspace`.

mamazu: Yes but this will put it into the route path again which we don't want.

alexander-schranz: we should skip the `type: rest` and define route ourself.

alexander-schranz: the API urls should stay the same as before if possible

mamazu: Routing changes have been removed as this was handled in your PRs.

PR #7943, opened 2025-05-05
Add new properties parser which returns new metadata for sulu form xmls

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Add new properties parser which returns new metadata.

#### Example Usage

This is more internal service. External service should get the metadata via the `MetadataProviderRegistry` service instead of using the underlaying services.

#### To Do (Meta WIP)

- [x] Block `mapPropertyMetadata` similar to old and imagemap integration
- [x] Add `SchemaMetadataProvider` tests, see removed FormMetadataMapperTest.php Schema tests
- [x] Update content type Tests
- [x] Update other Tests
- [x] Fix `Call to a member function toJsonSchema()` on array
- [x] needs further validation and discussions if this can even be a float (i dont think as it comes from xml as string always now) https://github.com/sulu/sulu/pull/7943/files#r2086293858

#### To Do 2

- [x] Add test for `FieldMetadata::findOption`
- [x] Add test for `FormMetadata::findTag`
- [x] Add test for `FormMetadata::hasTag`

#### To Do 3 all test for `mapPropertyMetadata`

Check existing changs

- [x] Update test `ImageMapContentTypeTest` to new Metadata
- [x] Update test `MediaSelectionContentTypeTest` to new Metadata
- [x] Update test `SingleMediaSelectionTest::testMapPropertyMetadata` to new Metadata
- [x] Update test `PageTreeRouteContentTypeTest::testMapPropertyMetadata` to new Metadata
- [x] Update test `NumberTest::testMapPropertyMetadata` new Metadata
- [x] Update test maybe other `testMapPropertyMetadata` unit tests

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/FormMetadata/FormMetadataMapper.php
+++ b/src/Sulu/Bundle/AdminBundle/FormMetadata/FormMetadataMapper.php
@@ -26,10 +26,6 @@ use Sulu\Component\Content\Metadata\SectionMetadata as ContentSectionMetadata;
  */
 class FormMetadataMapper
 {
-    public function __construct()
-    {
-    }
-
     /**
      * @return ItemMetadata[]
      */
--- a/src/Sulu/Bundle/AdminBundle/FormMetadata/FormMetadataMapper.php
+++ b/src/Sulu/Bundle/AdminBundle/FormMetadata/FormMetadataMapper.php
@@ -17,7 +17,6 @@ use Sulu\Bundle\AdminBundle\Metadata\FormMetadata\ItemMetadata;
 use Sulu\Bundle\AdminBundle\Metadata\FormMetadata\OptionMetadata;
 use Sulu\Bundle\AdminBundle\Metadata\FormMetadata\SectionMetadata;
 use Sulu\Bundle\AdminBundle\Metadata\FormMetadata\TagMetadata;
-use Sulu\Bundle\AdminBundle\Metadata\SchemaMetadata\PropertyMetadataMapperRegistry;
 use Sulu\Component\Content\Metadata\BlockMetadata as ContentBlockMetadata;
 use Sulu\Component\Content\Metadata\PropertyMetadata as ContentPropertyMetadata;
 use Sulu\Component\Content\Metadata\SectionMetadata as ContentSectionMetadata;
@@ -27,7 +26,7 @@ use Sulu\Component\Content\Metadata\SectionMetadata as ContentSectionMetadata;
  */
 class FormMetadataMapper
 {
-    public function __construct(private PropertyMetadataMapperRegistry $propertyMetadataMapperRegistry)
+    public function __construct()
     {
     }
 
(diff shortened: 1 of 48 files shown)


src/Sulu/Bundle/AdminBundle/FormMetadata/FormMetadataMapper.php
```diff
@@ -60,14 +60,6 @@ public function mapChildren(array $children): array
...
     }
 
-    /**
-     * @param ContentItemMetadata[] $itemsMetadata
-     */
-    public function mapSchema(array $itemsMetadata): SchemaMetadata
-    {
-        return new SchemaMetadata($this->mapSchemaProperties($itemsMetadata));
-    }
-
```
alexander-schranz, line 70: This is the only method we may still require. As discussed with @Prokyonn we are moving this into an own SchemaMetadataProvider service which is build ontop of the new metadata and not the old one. We use that service then in the formxmlloader and in the future templatexmlloader (currently structurexmlloader)

src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/OptionMetadata.php
```diff
@@ -32,7 +32,7 @@ class OptionMetadata
     protected $type;
 
     /**
-     * @var string|int|OptionMetadata[]
+     * @var string|float|int|OptionMetadata[]
```
alexander-schranz: was adjusted by @draconivis in: https://github.com/alexander-schranz/sulu/pull/8#event-17631157226

we need to recheck if "new metadata" ever writes as a value `float` if not we need adjust the test that such data comes as string `'1.5'` and maybe need to parse it to a float value.

src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/Parser/PropertiesXmlParser.php
alexander-schranz, line 1: Most of this service is a copy of existing FormMetadataMapper and old structure loader logic, I did just a little bit refactoring on the xml parse and phpstan baselined the other parts.

src/Sulu/Bundle/AdminBundle/FormMetadata/FormMetadataMapper.php
```diff
@@ -35,7 +26,7 @@
  */
 class FormMetadataMapper
 {
-    public function __construct(private PropertyMetadataMapperRegistry $propertyMetadataMapperRegistry)
+    public function __construct()
```
Prokyonn: can be removed

Prokyonn (approved):

alexander-schranz:
Adopted the FormXmlLoader with blocks test to test also against global blocks as only the deprecated and the service whiche we want to remove included global blocks metadata.

PR #7948, opened 2025-05-09
Fix duplicated toolbar actions in PageAdmin

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Fix duplicated toolbar actions

#### Why?

Toolbar actions should not be duplicated.


alexander-schranz (approved):

alexander-schranz:
The content bundle toolbar actions could not be used as we have special webspace specific checks like:

```php
'visible_condition' => '(!_permissions || _permissions.edit) && __webspace.localizations|length > 1',
```

So we give own toolbar actions to the contentviewbuilder which are used then in this case.

PR #7950, opened 2025-05-13
Fix MimeTypeIndicator fallback in collapsed MediaSelection Block

### LINKED ISSUE #7638: Document thumbnail display bug when block is collapsed
| Q | A
| --- | ---
| Sulu Version | 2.6.5
| PHP Version | 8.3
| DB Version | MySQL 5.7
| Browser Version | -

# Actual Behavior

When using “Media selection” in a block, if the type is document, there is a display error when the block is folded. Instead of displaying the PDF icon, for example, it tries to display an image of the media.
Capture d’écran 2024-11-04 à 16 05 04

Capture d’écran 2024-11-04 à 16 05 17

# Expected Behavior

Displaying document type icons.

# Steps to Reproduce

Create a block and use media_selection with only one document type and add documents to the block.

# Possible Solutions

Retrieve and display the media type icon.

draconivis:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #7638
| Related issues/PRs | #7638
| License | MIT
| Documentation PR | -

#### What's in this PR?

adds a check if the media thumbnail is available and if not, uses the MimeTypeIndicator as a fallback

#### Why?

consistency with expanded Block and #7638

#### Example Usage

no usage
image

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/Resources/js/components/MimeTypeIndicator/MimeTypeIndicator.js
+++ b/src/Sulu/Bundle/MediaBundle/Resources/js/components/MimeTypeIndicator/MimeTypeIndicator.js
@@ -42,7 +42,10 @@ export default class MimeTypeIndicator extends React.PureComponent<Props> {
         }
 
         return (
-            <div className={mimeTypeIndicatorStyles.mimeTypeIndicator} style={mimeTypeStyles}>
+            <div
+                className={mimeTypeIndicatorStyles.mimeTypeIndicator}
+                style={mimeTypeStyles}
+            >
                 <Icon name={icon} />
             </div>
         );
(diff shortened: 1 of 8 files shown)


src/Sulu/Bundle/MediaBundle/Resources/js/components/MediaCard/tests/__snapshots__/MediaCard.test.js.snap
```diff
@@ -175,6 +175,7 @@ exports[`Render a MediaCard component with MimeTypeIndicator if an error appeare
   >
     <div
       class="mimeTypeIndicator"
+      data-testid="MimeTypeIndicator"
```
alexander-schranz: we are not using any data-testid in our tests we should remove this and use proper selections based on class instead

draconivis: The reason i added the testid is that there isn't a classname query in the testing-library and i didn't know what else to use.
And Backdrop also has a testid

Prokyonn:
not happy about the additional `fetch` but I guess we do not have any other options 🤔  

@alexander-schranz is this fine for you as well?

wachterjohannes (approved):

PR #7750, opened 2025-01-24
Remove deprecated `UserGroup` feature

mamazu:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? | no 
| BC breaks? |  yes
| Deprecations? | no  
| Fixed tickets | -
| Related issues/PRs | #7547
| License | MIT
| Documentation PR | -

#### What's in this PR?
Removing deprecated `UserGroup` and `Group` entities, associated controller, tests and orm files. 

#### Why?
This feature has been marked deprecated and announced removal in 3.0. So we don't want to disappoint people.

PR #7960, opened 2025-05-14
Fix upmerging conflict of non-existing property

mamazu:
When upmerging this pr #7900 together with this #7750 (removal of the feature). We're left with a broken state.


alexander-schranz:
Thx 👍

PR #7961, opened 2025-05-15
Change AuhenticationFailureListener to use new LoginFailureEvent

KotaruS:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

This change adds a check to the Sulu\Bundle\SecurityBundle\EventListener\AuhenticationFailureListener to make it work on Symfony 6.0+ with new Symfony event. The original event Symfony\Component\Security\Core\Event\AuthenticationFailureEvent has been replaced in Symfony 6.0 with Symfony\Component\Security\Http\Event\LoginFailureEvent.

#### Why?

Making sure the listener works across all sulu 2.5+ supported Symfony versions.


alexander-schranz (approved):

alexander-schranz:
Thx for the fix 👍 , can you add a test case for `AuhenticationFailureListener` to avoid a regression in future for this service. Simple unit test with `onLoginFailure($event)` should work.

src/Sulu/Bundle/SecurityBundle/EventListener/AuhenticationFailureListener.php
```diff
@@ -34,14 +36,15 @@ public function __construct(private $passwordHasherFactory, private UserReposito
     public static function getSubscribedEvents()
     {
         return [
-            AuthenticationFailureEvent::class => 'onLoginFailure',
+            class_exists(AuthenticationFailureEvent::class) ? AuthenticationFailureEvent::class : LoginFailureEvent::class => 'onLoginFailure',
```
alexander-schranz: does this depend on the `AuthenticationFailureEvent` or do we need listen to both events? as symfony 5.4 can be used with the old and new authenticator system?

KotaruS: You are right, it's not necessary. S is the case for `EncoderFactoryInterface` and `UsernameNotFoundException` replaced with `PasswordHasherFactoryInterface` and `UserNotFoundException` respectively. I cleaned it up.

mamazu:
There's a typo in the classname. `AuhenticationFailureListener` -> `AuthenticationFailureListener`

alexander-schranz:
for BC reason we keep the old class name.
for bc reasons we need support both encoders and hashers as Sulu 2.5 supports still Symfony 5.4.

Symfony 5.4 has a flag to use the old authenticator system depending on AuthenticationFailureEvent we also need to keep that way also for the people which didn't enable the new authenticator on 5.4.

In Sulu 3.0 branch we can then remove this bc layers.

alexander-schranz:
@KotaruS Thank you!

PR #7964, opened 2025-05-16
Add admin interface for managing snippet areas

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | yes
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?
Snippet Admin Controllers in the new structure.

#### Why?
Moving the classes to the new namespace, updating it to match the rest of Sulu 3.


packages/snippet/assets/js/views/SnippetAreas/stores/SnippetAreaStore.js
```diff
@@ -14,7 +14,7 @@ export default class SnippetAreaStore {
         this.webspaceKey = webspaceKey;
 
         ResourceRequester.getList('snippet_areas', {webspace: webspaceKey}).then(action((response) => {
-            this.snippetAreas = response._embedded.areas.reduce((snippetAreas, snippetArea) => {
+            this.snippetAreas = response._embedded.snippet_areas.reduce((snippetAreas, snippetArea) => {
```
mamazu, line 17: With the ListRepresentation the key in the json response changes.

packages/snippet/src/Infrastructure/Symfony/Normalizer/SnippetNormalizer.php
```diff
@@ -0,0 +1,53 @@
...
+     */
+    public function normalize(
+        mixed $data,
+        ?string $format = null,
+        array $context = [],
+    ): array {
+        /** @var array<mixed> $normalizedData */
+        $normalizedData = $this->objectNormalizer->normalize($data, $format, $context);
+
+        dd($this->snippetArea[$normalizedData['areaKey']]);
```
mamazu: Here we'd attach the title and the template for the snippet area as this doesn't need to be saved in the database.

alexander-schranz:
> Do we have a replacement for the MetadataFactory already? I'm currently putting most of the the snippet area info into the response via Normalizer. I'm not sure that's the correct way to do it.

@mamazu The admin should not require any metadata? Where do you require it? or where is it currently used?

mamazu:
@alexander-schranz For the title, template information and other things that the api used to return for every SnippetArea (see the Compiler Pass)

alexander-schranz:
You can keep that service at current state, we do currently the refactoring of the metadata and need to tackle the `getAreas` part in that also.

PS: if that is tackled we can refactor it to use the `MetadataProviderRegistry` service.

mamazu:
So I am using the `PaginatedRepresentation` for outputting. However this also adds pagination. Is there a non paginated representation class?

The CollectionRepresentation does not include the total number of entries.

mamazu:
@alexander-schranz I also need some guidance on how you want the SnippetAreaController handle the data loading. My approach is to load all snippet areas from the database (for the current webspace) and then fill the rest with the default snippet data. (See current implementation)

However my preferred solution would be to handle this all in a repository method. But I don't think that you can override the list builders repository method. So what do you suggest?

alexander-schranz:
@mamazu would at current state handle it in the controller.

mamazu:
@alexander-schranz In the SnippetAreaController we don't really need any customization points for the query (like we don't have pagination, filters etc). Would it be okay we just used a repository in here and dropped the whole `fieldDescriptor` and `listBuilderFactory` stuff?

alexander-schranz:
@mamazu else like any other list it should go over the field descriptors and listbuilder to create the query not the repository at current state we should be consistent over all new controllers and not add new patterns here. Enrich them with additional items, keys in the controller is still common and the controller responsability.

packages/snippet/src/Infrastructure/Symfony/CompilerPass/SnippetAreaCompilerPass.php
```diff
@@ -0,0 +1,123 @@
...
+        ]);
+        $structures[] = $structure;
+
+        $structure = new StructureMetadata();
+        $structure->setName('hotel');
+        $structure->setAreas([
+            ['key' => 'golf_hotel', 'template' => 'golf_hotel', 'cache-invalidation' => true, 'title' => ['Golf hotel']],
+            ['key' => 'sport_hotel', 'template' => 'sport_hotel', 'cache-invalidation' => true, 'title' => ['Sport hotel']],
+        ]);
+        $structures[] = $structure;
```
mamazu: TODO: Replace with the real data.

packages/snippet/src/UserInterface/Controller/Admin/SnippetAreaController.php
```diff
@@ -0,0 +1,129 @@
...
+        //$fieldDescriptors = $this->fieldDescriptorFactory->getFieldDescriptors(SnippetAreaInterface::RESOURCE_KEY);
+        //Assert::notNull($fieldDescriptors, 'Could not find field descriptors for resource key: ' . SnippetAreaInterface::RESOURCE_KEY);
+
+        /** @var DoctrineListBuilder $listBuilder */
+        //$listBuilder = $this->listBuilderFactory->create(SnippetAreaInterface::class);
+        //$listBuilder->setIdField($fieldDescriptors['id']); // We need to set this because it's the uuid doctrine column
+        //$listBuilder->setParameter('locale', $request->query->get('locale'));
+        //$listBuilder->setParameter('webspace', $request->query->get('webspace'));
+
+        //$this->restHelper->initializeListBuilder($listBuilder, $fieldDescriptors);
```
mamazu: // TODO: Currently I'm using the repository directly, as we need none of the functions of the list builder (pagination, filters, sorting etc)

mamazu (commented):
.

packages/snippet/config/routing_admin_api.yaml
```diff
@@ -33,3 +33,20 @@ sulu_snippet.post_snippet_trigger:
...
+sulu_snippet_area.get_snippet_areas:
+    path: /snippet-areas.{_format}
+    methods: GET
+    controller: sulu_snippet.admin_snippet_area_controller::cgetAction
+    format: json
+
+sulu_snippet_area.put_snippet_area:
+    path: /snippet-areas/{key}
+    methods: PUT
+    controller: sulu_snippet.admin_snippet_area_controller::putAction
```
Prokyonn, line 53:
```suggestion
sulu_snippet_area.put_snippet_area:
    path: /snippet-areas/{key}.{_format}
    methods: PUT
    controller: sulu_snippet.admin_snippet_area_controller::putAction
    format: json
```

packages/snippet/config/routing_admin_api.yaml
```diff
@@ -33,3 +33,20 @@ sulu_snippet.post_snippet_trigger:
...
+
+sulu_snippet_area.delete_snippet_area:
+    path: /snippet-areas/{key}
+    methods: PUT
+    controller: sulu_snippet.admin_snippet_area_controller::deleteAction
```
Prokyonn, line 59:
```suggestion
sulu_snippet_area.delete_snippet_area:
    path: /snippet-areas/{key}.{_format}
    methods: PUT
    controller: sulu_snippet.admin_snippet_area_controller::deleteAction
    format: json
```

packages/snippet/src/Application/Message/ModifySnippetAreaMessage.php
```diff
@@ -0,0 +1,53 @@
...
+
+class ModifySnippetAreaMessage
+{
+    private string $webspace;
+    private string $areaKey;
+
+    /**
+     * @var array{ uuid: string }
+     */
+    private array $snippet;
```
Prokyonn: personally I would call this `snippetIdentifier`

packages/snippet/src/Application/MessageHandler/RemoveSnippetAreaMessageHandler.php
```diff
@@ -0,0 +1,40 @@
...
+
+class RemoveSnippetAreaMessageHandler
+{
+    public function __construct(private EntityManagerInterface $entityManager)
+    {
+    }
+
+    public function __invoke(RemoveSnippetAreaMessage $message): void
+    {
+        $snippetAreaRepository = $this->entityManager->getRepository(SnippetArea::class);
```
Prokyonn: I would inject the repository or at least try to get it via the interface

packages/snippet/src/Domain/Model/SnippetArea.php
```diff
@@ -0,0 +1,81 @@
...
+namespace Sulu\Snippet\Domain\Model;
+
+use Symfony\Component\Uid\Uuid;
+
+class SnippetArea implements SnippetAreaInterface
+{
+    /**
+     * @var string
+     */
+    protected $uuid;
```
Prokyonn:
```suggestion
    protected string $uuid;
```

packages/snippet/src/Domain/Repository/SnippetAreaRepositoryInterface.php
```diff
@@ -0,0 +1,33 @@
...
+ * @see Sulu\Snippet\Infrastructure\Doctrine\Repository\SnippetAreaRepository
+ */
+interface SnippetAreaRepositoryInterface
+{
+    public function createNew(string $areaKey, string $webspaceKey, ?string $uuid = null): SnippetAreaInterface;
+
+    /**
+     * @return array<string, SnippetAreaInterface>
+     */
+    public function findByWebspace(string $webspaceKey): array;
```
Prokyonn: our new repositories should be consistent take a look at the PageRepositoryInterface, e.g. `findOneBy` / `findBy` / `getOneBy` / `getBy` with filters.

packages/snippet/src/Application/MessageHandler/RemoveSnippetAreaMessageHandler.php
```diff
@@ -0,0 +1,40 @@
...
+        $entityToDelete = $snippetAreaRepository->findOneBy([
+            'webspaceKey' => $message->getWebspaceKey(),
+            'areaKey' => $message->getAreaKey(),
+        ]);
+
+        if (null === $entityToDelete) {
+            return;
+        }
+
+        $this->entityManager->remove($entityToDelete);
```
Prokyonn: Add `remove` to the repository, then we do not even need the EntityManager here

packages/snippet/src/Application/MessageHandler/ModifySnippetAreaMessageHandler.php
```diff
@@ -0,0 +1,46 @@
...
+    public function __invoke(ModifySnippetAreaMessage $message): void
+    {
+        $webspaceKey = $message->getWebspace();
+        $areaKey = $message->getAreaKey();
+
+        $snippetArea = $this->snippetAreaRepository->findOneByWebspaceAndKey($webspaceKey, $areaKey);
+        if (null === $snippetArea) {
+            $snippetArea = $this->snippetAreaRepository->createNew($areaKey, $webspaceKey);
+
+            $this->entityManager->persist($snippetArea);
```
Prokyonn: Add `add` method to the repository, so that we do not need the EntityManager directly here

packages/snippet/src/Infrastructure/Doctrine/Repository/SnippetAreaRepository.php
```diff
@@ -0,0 +1,72 @@
...
+
+    /**
+     * @return array<string, SnippetAreaInterface>
+     */
+    public function findByWebspace(string $webspaceKey): array
+    {
+        $result = [];
+
+        /** @var array<SnippetAreaInterface> $queryResult */
+        $queryResult = $this->entityRepository->createQueryBuilder('o')
```
Prokyonn: implement a `createQueryBuilder` method similar to the other repositories `PageRepository` and do not use abbreviations, I think it is a lot more readable without them

packages/snippet/src/UserInterface/Controller/Admin/SnippetAreaController.php
Prokyonn: I would still use the listBuilder to have it more consistent and you never know, somebody may implement tons of snippet_area keys and therefore pagination may come in handy

packages/snippet/src/UserInterface/Controller/Admin/SnippetAreaController.php
```diff
@@ -0,0 +1,129 @@
...
+    {
+        $message = new ModifySnippetAreaMessage([
+            ...$request->attributes->all('_route_params'),
+            ...$request->request->all(),
+        ]);
+
+        /** @see Sulu\Snippet\Application\MessageHandler\SetSnippetAreaMessageHandler */
+        $this->handle(new Envelope($message, [new EnableFlushStamp()]));
+
+        return new Response(null, Response::HTTP_OK);
```
Prokyonn: We should still return the snippetArea values here, to have the API similar to the old one

packages/snippet/src/UserInterface/Controller/Admin/SnippetAreaController.php
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    public function deleteAction(Request $request): Response
+    {
+        $message = new RemoveSnippetAreaMessage($request->attributes->all());
+
+        /** @see Sulu\Snippet\Application\MessageHandler\RemoveSnippetAreaMessageHandler */
+        $this->handle(new Envelope($message, [new EnableFlushStamp()]));
+
+        return new Response(null, Response::HTTP_NO_CONTENT);
```
Prokyonn: We should still return the snippetArea values here, to have the API similar to the old one

Prokyonn (changes requested):

packages/snippet/src/Application/MessageHandler/RemoveSnippetAreaMessageHandler.php
mamazu: I think the cleanest way would be to add the `remove` method to the repository then we don't need the `EntityManager` here at all.

Edit: Done that. :laughing:

packages/snippet/src/UserInterface/Controller/Admin/SnippetAreaController.php
mamazu: I agree, but as far as I know the frontend implementation doesn't support that. (And it wasn't present in 2.6)

But I couldn't get the join to work with the list XML. Probably a skill issue. :D

packages/snippet/src/Domain/Repository/SnippetAreaRepositoryInterface.php
mamazu: What do you have in mind for this method? I don't see a problem with this.

packages/snippet/src/Application/Message/ModifySnippetAreaMessage.php
```diff
@@ -0,0 +1,53 @@
...
+
+    /**
+     * @var array{ uuid: string }
+     */
+    private array $snippetIdentifier;
+
+    /**
+     * @param array{webspace: string, key: string, snippet: array{uuid: string}} $requestData
+     */
+    public function __construct(array $requestData)
```
Prokyonn: rename to `$data` as the message is not only dispatched from a controller

I would prefer to have a separate parameter for the `snippetIdentifier`, would be nice if we can have all the new CRUD messages more or less consistent (e.g. article/page/snippet) WDYT @alex?

packages/snippet/src/Application/Message/RemoveSnippetAreaMessage.php
```diff
@@ -0,0 +1,39 @@
...
+
+class RemoveSnippetAreaMessage
+{
+    private string $webspaceKey;
+    private string $areaKey;
+
+    /**
+     * @param array<string, string> $requestData
+     */
+    public function __construct(array $requestData)
```
Prokyonn:
```suggestion
    public function __construct(array $data)
```

packages/snippet/src/Domain/Repository/SnippetAreaRepositoryInterface.php
Prokyonn: e.g. in the `PageRepository` we have a `findBy` method with predefined filters in the phpdoc, so e.g. `findByWebspace` would be `findBy($filters)` and the `$filter = ['webspaceKey' => 'sulu.io']`

This allows us to add more filters without breaking changes and the interface does not become unnecessary confusing/complex due to a large number of individual methods, which are mostly in a few use cases.

packages/snippet/src/Infrastructure/Doctrine/Repository/SnippetAreaRepository.php
```diff
@@ -0,0 +1,82 @@
...
+
+    /**
+     * @return array<string, SnippetAreaInterface>
+     */
+    public function findByWebspace(string $webspaceKey): array
+    {
+        $result = [];
+
+        /** @var array<SnippetAreaInterface> $queryResult */
+        $queryResult = $this->entityRepository->createQueryBuilder('area')
```
Prokyonn: Introduce a `createQueryBuilder` method in this repository similar to the [PageRepository::createQueryBuilder](https://github.com/sulu/sulu/blob/3.0/packages/page/src/Infrastructure/Doctrine/Repository/PageRepository.php#L270)

packages/snippet/src/Infrastructure/Symfony/HttpKernel/SuluSnippetBundle.php
```diff
@@ -168,6 +201,23 @@ public function loadExtension(array $config, ContainerConfigurator $container, C
...
+        $services->set('sulu_snippet.snippet_area_admin')
+            ->class(SnippetAreaAdmin::class)
+            ->args([
+                new Reference('sulu_admin.view_builder_factory'),
+                new Reference('sulu_security.security_checker'),
+            ])
+            ->tag('sulu.context', ['context' => 'admin'])
+            ->tag('sulu.admin');
+
+        $services->set('sulu_snippet.normalizer.snippet_area', SnippetAreaNormalizer::class)
```
Prokyonn:
```suggestion
        $services->set('sulu_snippet.snippet_area_normalizer', SnippetAreaNormalizer::class)
```

packages/snippet/translations/admin.de.json
```diff
@@ -20,7 +20,7 @@
     "sulu_snippet.last_modified_date": "Datum der Aktualisierung",
     "sulu_snippet.snippet_area": "Typ oder Bereich",
     "sulu_snippet.taxonomies": "Taxonomien",
-    "sulu_snippet.webspace_default_snippet": "{webspaceKey} Standard Schnipsel",
```
Prokyonn, line 23: Why did you remove the webspace?

Prokyonn (changes requested):

packages/snippet/src/Infrastructure/Sulu/Admin/SnippetAreaAdmin.php
```diff
@@ -0,0 +1,54 @@
...
+        $viewCollection->add(
+            $this->viewBuilderFactory
+                ->createViewBuilder('sulu_snippet.snippet_areas', '/snippet-areas', 'sulu_snippet.snippet_areas')
+                ->setOption('snippetEditView', SnippetAdmin::EDIT_TABS_VIEW)
+                ->setOption('tabTitle', 'sulu_snippet.webspace_default_snippets')
+                ->setOption('tabOrder', 3072)
+                ->setParent(PageAdmin::WEBSPACE_TABS_VIEW)
+                ->addRerenderAttribute('webspace')
+        );
+    }
```
Prokyonn, line 53: implementation for security method `getSecurityContexts()` missing

packages/snippet/translations/admin.de.json
mamazu: I didn't find that this translation is used anywhere.

packages/snippet/src/Domain/Repository/SnippetAreaRepositoryInterface.php
mamazu: Then I don't understand why we should need a custom repository in the first place. If we write the code that generic we can't type it properly anymore and we would loose some granularity in terms of joins. Like for example if we move the `Snippet` join into the create query builder this will slow down all queries that don't need a snippet and leads to over fetching.
In this case specifically we index the results by `areaKey`, this only makes sense if we have a webspace key as filter otherwise the results wouldn't be unique. And having a logic like "if webspaceKey in filters" return it indexed by area key and if no webpsaceKey filter exists return the bare list of `SnippetArea`s also sounds like a bad idea.

What I think would be reasonable would be to just forward the call to the native repository. But then again, why do we need the custom one. :D

packages/snippet/src/Infrastructure/Symfony/CompilerPass/SnippetAreaCompilerPass.php
```diff
@@ -0,0 +1,123 @@
...
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+/**
+ * @phpstan-type Entry array{key: string, template: string, title: array<string, string>, cache-invalidation: mixed}
+ */
+class SnippetAreaCompilerPass implements CompilerPassInterface
+{
+    public const SNIPPET_AREA_PARAM = 'sulu_snippet.areas';
+
+    public function process(ContainerBuilder $container): void
```
Prokyonn, line 38: As discussed with @alexander-schranz the compiler pass must be refactored. As we should not use any services, we decided to iterate over all snippet templates and read the `snippet-area` part directly in the compiler pass. Then set it as a parameter in the container.
If no snippet-area is defined, we should deregister the `SnippetAreaAdmin`, so that the empty tab is not visible in the admin UI

packages/snippet/src/Infrastructure/Sulu/Admin/SnippetAreaAdmin.php
mamazu: Yep, added.

packages/snippet/src/Infrastructure/Symfony/CompilerPass/SnippetAreaCompilerPass.php
mamazu: Is there already a reference implementation on how to read the snippet xml files in the 3.0 version or is it still the same as in 2.6?

packages/snippet/src/Infrastructure/Symfony/CompilerPass/SnippetAreaCompilerPass.php
```diff
@@ -11,113 +11,86 @@
...
-                }
-
-                $areas[$area['key']] = $area;
-            }
+            $metaData[$key] = $element;
         }
 
         \ksort($areas);
 
+        dump($areas);
```
alexander-schranz: 👀

mamazu: Yeah it's still work in progress. Me fighting xml. :crossed_swords:

packages/snippet/src/Infrastructure/Symfony/CompilerPass/SnippetAreaCompilerPass.php
mamazu: @alexander-schranz Okay, I have refactored the stuff to now read xml but this makes it harder to test the compiler pass. We don't have example xmls for that though. Do you have any better ideas for testing it?

Prokyonn: @mamazu we have already one snippet template in `packages/snippet/tests/Application/config/templates/snippets`, you can add multiple snippet areas there and also create another one to test the reading of multiple files.

mamazu: Okay, done. I don't know how to test the error case yet where two snippets define the same area. But other than that the test is good.

Now I think there is still the question of cache-invalidation I think.

mamazu:
@alexander-schranz @Prokyonn I've tried using the translator in the compiler pass, that doesn't work. I also can't use "TranslatableMessage" objects in the container parameters to defer translation to the Normalizer later. Any other ideas?

Prokyonn:
> @alexander-schranz @Prokyonn I've tried using the translator in the compiler pass, that doesn't work. I also can't use "TranslatableMessage" objects in the container parameters to defer translation to the Normalizer later. Any other ideas?

Hey @mamazu I've already fixed this on my local branch. You can use the translator, when your compiler pass has a low priority (e.g. -1024). As soon as my adjustments are finished and everything works again I'll push the update to this branch :)

packages/snippet/src/Infrastructure/Symfony/CompilerPass/SnippetAreaCompilerPass.php
```diff
@@ -0,0 +1,338 @@
...
+    /**
+     * @param array<string> $templateDirectories
+     *
+     * @return array<string, array{title: array<string, string>, cache-invalidation: bool, areaKey: string, template: string}>
+     */
+    private function loadAreasFromXmlFiles(array $templateDirectories): array
+    {
+        if (empty($templateDirectories)) {
+            return [];
+        }
```
mamazu: Can be removed, the function returns an empty array anyways if there are no files and ([] === $templateDirectories) would be better

mamazu:
I also dislike the heavy use of the empty function in this code. Seeing that phpstorm also isn't good at inferring types. This can make the code hard to read.

Prokyonn:
> I also dislike the heavy use of the empty function in this code. Seeing that phpstorm also isn't good at inferring types. This can make the code hard to read.

Yep refactored it a bit, although I'm still not really happy with this class 🙈

packages/snippet/tests/Functional/Integration/responses/snippet_area_cget.json
```diff
@@ -2,28 +2,25 @@
...
-                "template": "car",
-                "title": "Car",
-                "defaultUuid": null,
-                "defaultTitle": null,
-                "valid": true
+                "key": "hotel",
+                "snippetTitle": null,
+                "snippetUuid": null,
+                "templateKey": "snippet",
+                "title": "snippet_type.hotel"
```
mamazu: Is this supposed to be untranslated?

Prokyonn: The translation key was missing, good catch, thank you :)

alexander-schranz (approved):

alexander-schranz:
Thank you @mamazu @Prokyonn

PR #7965, opened 2025-05-19
Fix IcomoonProvider warning

martinlagler:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Related Issues | https://github.com/sulu/sulu/issues/8021
| Deprecations? | no
| License | MIT

#### What's in this PR?

Fix Undefined array key warning

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Icon/providers/IcomoonProvider.php
+++ b/src/Sulu/Bundle/AdminBundle/Icon/providers/IcomoonProvider.php
@@ -49,9 +49,9 @@ class IcomoonProvider implements IconProviderInterface
              * @var string $path
              */
             foreach ($icon->icon->paths as $index => $path) {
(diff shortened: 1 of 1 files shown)


src/Sulu/Bundle/AdminBundle/Icon/providers/IcomoonProvider.php
```diff
@@ -50,7 +50,7 @@ public function getIcons(string $path): array
              */
             foreach ($icon->icon->paths as $index => $path) {
                 /** @var \stdClass $attrs */
```
alexander-schranz:
```suggestion
                /** @var \stdClass|null $attrs */
```

alexander-schranz (changes requested):
typehint is wrong

PR #7966, opened 2025-05-19
Fixing pubic flysystem urls

### LINKED ISSUE #7957: Flysystem S3 redirect (Sulu 3)
| Q | A
| --- | ---
| Sulu Version | Sulu 3 alpha 2
| PHP Version | 8.3
| DB Version | MariaDB 10.9
| Browser Version | Chrome 136

# Actual Behavior

Opening a version of a file which is hosted on S3 throws an error "The file does not exist".

# Possible Solution
I think there is a missing redirection to the `public_url` that is defined in `flysystem.yaml`

mamazu:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #7957 
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?
Adding the call to the public url for remote services.

#### Why?
Otherwise generating the public url will no longer work.

Thanks to @hual7 for helping with S3 credentials.

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/Media/Storage/FlysystemStorage.php
+++ b/src/Sulu/Bundle/MediaBundle/Media/Storage/FlysystemStorage.php
@@ -170,8 +170,8 @@ class FlysystemStorage implements StorageInterface
 
     public function getPath(array $storageOptions): string
     {
-        if (null === $this->rootPath) {
-            return $this->getFilePath($storageOptions);
+        if ($this->getType($storageOptions) === StorageInterface::TYPE_REMOTE) {
+            return $this->filesystem->publicUrl($this->getFilePath($storageOptions));
         }
 
         return $this->rootPath . '/' . $this->getFilePath($storageOptions);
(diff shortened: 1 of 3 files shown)


alexander-schranz (approved):

src/Sulu/Bundle/MediaBundle/Media/Storage/FlysystemStorage.php
```diff
@@ -171,7 +171,7 @@ private function createDirectories(array $storageOptions): void
     public function getPath(array $storageOptions): string
     {
         if (null === $this->rootPath) {
-            return $this->getFilePath($storageOptions);
+            return $this->filesystem->publicUrl($this->getFilePath($storageOptions));
```
alexander-schranz, line 175: Local adapter need to return sti the path See 2.6

mamazu: Okay there are now three cases: remote, local, local with root path.

alexander-schranz (approved):

PR #7982, opened 2025-06-02
Add XmlTemplateFormMetadataLoader to replace old structure based metadata loading

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no  
| Fixed tickets | fixes # 
| Related issues/PRs | https://github.com/sulu/sulu/pull/7908
| License | MIT

#### What's in this PR?

Add new XmlTemplateFormMetadataLoader.

#### Why?

To replace old structure based metadata loading.

#### ToDo

 - [x] Settings Form Key handling
 - [x] Template Meta Title
 - [x] Block MinOccurs
 - [x] Tests

### DIFF
--- a/packages/article/src/Infrastructure/Symfony/HttpKernel/SuluArticleBundle.php
+++ b/packages/article/src/Infrastructure/Symfony/HttpKernel/SuluArticleBundle.php
@@ -275,7 +275,7 @@ final class SuluArticleBundle extends AbstractBundle
                         ],
                     ],
                     'templates' => [
-                        'articles' => [
+                        ArticleInterface::TEMPLATE_TYPE => [
                             'default_type' => null,
                             'directories' => [
                                 'app' => '%kernel.project_dir%/config/templates/articles',
--- a/src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/Visitor/BlockSettingsFormMetadataVisitor.php
+++ b/src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/Visitor/BlockSettingsFormMetadataVisitor.php
@@ -11,6 +11,7 @@
 
 namespace Sulu\Bundle\AdminBundle\Metadata\FormMetadata\Visitor;
 
+use Sulu\Bundle\AdminBundle\Metadata\FormMetadata\FieldMetadata;
 use Sulu\Bundle\AdminBundle\Metadata\FormMetadata\ItemMetadata;
 use Sulu\Bundle\AdminBundle\Metadata\FormMetadata\OptionMetadata;
 use Sulu\Bundle\AdminBundle\Metadata\FormMetadata\SectionMetadata;
@@ -38,6 +39,10 @@ class BlockSettingsFormMetadataVisitor implements TypedFormMetadataVisitorInterf
             if ($itemMetadata instanceof SectionMetadata) {
                 $this->enhanceBlockMetadata($itemMetadata->getItems());
 
+                continue;
+            }
+
+            if (!$itemMetadata instanceof FieldMetadata) {
                 continue;
             }
 
--- a/src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/Visitor/BlockSettingsFormMetadataVisitor.php
+++ b/src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/Visitor/BlockSettingsFormMetadataVisitor.php
@@ -1,0 +1,60 @@
+<?php
+
+/*
+ * This file is part of Sulu.
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\AdminBundle\Metadata\FormMetadata\Visitor;
+
+use Sulu\Bundle\AdminBundle\Metadata\FormMetadata\ItemMetadata;
+use Sulu\Bundle\AdminBundle\Metadata\FormMetadata\OptionMetadata;
+use Sulu\Bundle\AdminBundle\Metadata\FormMetadata\SectionMetadata;
+use Sulu\Bundle\AdminBundle\Metadata\FormMetadata\TypedFormMetadata;
+use Sulu\Bundle\AdminBundle\Metadata\FormMetadata\TypedFormMetadataVisitorInterface;
+
+/**
+ * @internal
+ */
+class BlockSettingsFormMetadataVisitor implements TypedFormMetadataVisitorInterface
+{
+    public function visitTypedFormMetadata(TypedFormMetadata $formMetadata, string $key, string $locale, array $metadataOptions = []): void
+    {
+        foreach ($formMetadata->getForms() as $formMetadata) {
+            $this->enhanceBlockMetadata($formMetadata->getItems());
+        }
+    }
+
+    /**
+     * @param ItemMetadata[] $itemsMetadata
+     */
+    private function enhanceBlockMetadata(array $itemsMetadata): void
+    {
+        foreach ($itemsMetadata as $itemMetadata) {
+            if ($itemMetadata instanceof SectionMetadata) {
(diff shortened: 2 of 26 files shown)


packages/article/src/Infrastructure/Symfony/HttpKernel/SuluArticleBundle.php
```diff
@@ -275,7 +275,7 @@ public function prependExtension(ContainerConfigurator $container, ContainerBuil
                         ],
                     ],
                     'templates' => [
-                        'articles' => [
+                        ArticleInterface::TEMPLATE_TYPE => [
```
alexander-schranz, line 278: I personally would remove the "template_type" currently singleton (page, snippet, article) and replace it with the resource key (pages, snippets, articles) but that should be discussed and tackled in an own Pull request.

src/Sulu/Bundle/CoreBundle/DependencyInjection/SuluCoreExtension.php
```diff
@@ -78,6 +78,20 @@ public function prepend(ContainerBuilder $container)
...
+                'templates' => [
+                    'block' => [ // TODO maybe blocks?
+                        'default_type' => null,
+                        'directories' => [
+                            'app' => '%kernel.project_dir%/config/templates/blocks',
+                        ],
+                    ],
+                ],
+            ]
+        );
```
alexander-schranz, line 93: we should move this in to the admin or content bundle not sure. Little bit splitted what the correct place is. /cc @Prokyonn, @chirimoya.

JS for global blocks is in the Admin Bundle. Metadata currently also. Template types definitions are in the specific bundles (articles, snippets, page).

Global Metadata service is currently in component would also required to move to the bundle which does the configuration. 

Block json metadata which also handles global blocks is currently part of the block content type currently inside components. The block property resolver will be part of content bundle, the json metadata maybe also but would not be required, currently property resolver interface and json schema interfaces are on the same class.

- https://github.com/sulu/sulu/blob/3.0/src/Sulu/Component/Content/Types/BlockContentType.php#L471-L505
- https://github.com/sulu/sulu/blob/3.0/src/Sulu/Component/Content/Types/Metadata/GlobalBlocksTypedFormMetadataVisitor.php
- https://github.com/sulu/sulu/blob/f77b61b1e5951ff7d855fa01640ca774e878d9b3/src/Sulu/Bundle/CoreBundle/DependencyInjection/SuluCoreExtension.php#L71-L74

src/Sulu/Component/Content/SmartContent/PageDataProvider.php
```diff
@@ -390,7 +390,7 @@ private function getTypes(): array
             $metadata = $this->formMetadataProvider->getMetadata('page', $user->getLocale(), []);
 
             foreach ($metadata->getForms() as $form) {
-                $types[] = ['type' => $form->getName(), 'title' => $form->getTitle($user->getLocale())];
+                $types[] = ['type' => $form->getKey(), 'title' => $form->getTitle($user->getLocale())];
```
alexander-schranz: I'm kind of confused about `name` and `key` in forms vs templates @Prokyonn lets discuss this.

src/Sulu/Bundle/AdminBundle/Resources/config/services.xml
```diff
@@ -209,7 +228,9 @@
             <argument>%sulu_core.locales%</argument>
             <argument>%sulu.cache_dir%/form-structures</argument>
             <argument>%kernel.debug%</argument>
+            <!--
             <tag name="sulu_admin.form_metadata_loader"/>
+            -->
```
alexander-schranz, line 240: Currently commented out but whole service should later be removed in upcoming PR with removing the structure metadata.

This PR responsibility is just replace all usage of metadata of the admin is not longer build on top of structure metadata instead using the new parsers of the admin bundle.

src/Sulu/Component/Content/SmartContent/PageDataProvider.php
alexander-schranz: tackled in https://github.com/sulu/sulu/pull/7984

src/Sulu/Bundle/AdminBundle/Tests/Functional/Controller/fixtures/default.json
```diff
@@ -270,7 +270,7 @@
                                     "required": false,
                                     "multilingual": true,
                                     "spaceAfter": null,
-                                    "minOccurs": null,
+                                    "minOccurs": 0,
```
alexander-schranz, line 273: TODO need to find out where this regression comes from.

src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/Visitor/BlockSettingsFormMetadataVisitor.php
```diff
@@ -0,0 +1,60 @@
...
+use Sulu\Bundle\AdminBundle\Metadata\FormMetadata\ItemMetadata;
+use Sulu\Bundle\AdminBundle\Metadata\FormMetadata\OptionMetadata;
+use Sulu\Bundle\AdminBundle\Metadata\FormMetadata\SectionMetadata;
+use Sulu\Bundle\AdminBundle\Metadata\FormMetadata\TypedFormMetadata;
+use Sulu\Bundle\AdminBundle\Metadata\FormMetadata\TypedFormMetadataVisitorInterface;
+
+/**
+ * @internal
+ */
+class BlockSettingsFormMetadataVisitor implements TypedFormMetadataVisitorInterface
```
alexander-schranz, line 24: copied from `StructureFormMetadataLoader` which will get removed in upcoming PR

src/Sulu/Bundle/PageBundle/DependencyInjection/SuluPageExtension.php
```diff
@@ -45,6 +46,13 @@ public function prepend(ContainerBuilder $container)
                             __DIR__ . '/../Resources/config/forms',
                         ],
                     ],
+                    'templates' => [
+                        PageInterface::TEMPLATE_TYPE => [
+                            'directories' => [
+                                'app' => '%kernel.project_dir%/config/templates/pages',
+                            ],
+                        ],
+                    ],
```
alexander-schranz, line 55: required currently also in the old page bundle as the new one is not registered yet in old other bundles because conflicts with the old one

src/Sulu/Bundle/AdminBundle/Tests/Functional/Controller/fixtures/default.json
alexander-schranz: Actually 0 is the correct value here: https://github.com/sulu/sulu/blob/2bb32422465bcda3b4cb361730a9703e5d1ec197/src/Sulu/Bundle/AdminBundle/Tests/Application/config/templates/pages/default.xml#L95

Shold we convert 0 to null or keep it as it is now? /cc @Prokyonn

Prokyonn: I think keeping it like this is more transparent

src/Sulu/Bundle/CoreBundle/DependencyInjection/SuluCoreExtension.php
Prokyonn: My gut feeling says contentBundle, but I cannot tell you a specific reason for it 😅, both cases are valid

Prokyonn (approved):

src/Sulu/Bundle/CoreBundle/DependencyInjection/SuluCoreExtension.php
alexander-schranz: okay lets tackle this in own PR my first indention was also content, thomas also, but discussion is will it be used outside of content bundle then it would be better to stay part of admin.

PR #4643, opened 2019-07-24
Support multiple form metadata loaders

markusmoosbrugger:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | 
| Related issues/PRs | https://github.com/sulu/SuluFormBundle/pull/192
| License | MIT
| Documentation PR | 

#### What's in this PR?

This PR adds the support for multiple form metadata loaders. The different FormMetadataLoader classes can be tagged in the associated **services.xml** files as **sulu_admin.form_metadata_loader** in order to get injected automatically in the **FormMetadataProvider**. The different FormMetadataLoader classes must implement the Interface **FormMetadataLoaderInterface**.

Furthermore this PR changes the loading and caching mechanism of form metadata. The **FormXmlLoader** returns now an instance of **LocalizedFormMetadataCollection**, which contains a **FormMetadata** object for each locale. Instead of the old metadata objects, the newer ones are used now. The old metadata classes are now marked as deprecated.

The **FormMetadataXmlLoader** implements the **FormMetadataLoaderInterface** and provides the functionality to load and cache the form metadata. This was done previously in the **FormMetadataProvider**. 
The **StructureLoader** is added as well in this PR and also implements the **FormMetadataLoaderInterface**. The StructureLoader is responsible for caching the structure metadata.

The class **FormMetadataMapper** handles the conversation between old and new metadata objects and is used in the **FormMetadataXmlLoader**,  **StructureLoader** and **FormXmlLoader**.

#### Why?
Enable to use a custom metadata loader provided from the form bundle (see PR https://github.com/sulu/SuluFormBundle/pull/192) . The functionality can also be useful for other bundles that want to provide their own metadata loader.

The modification to use the new metadata classes in the **FormXmlLoader** is necessary to avoid the usage of the old metadata objects in new loaders.

#### Example Usage

Add a service that implements the FormMetadataLoaderInterface. Afterwards tag this service in the services.xml file as **sulu_admin.form_metadata_loader**. 
Example:

~~~xml
<service id="sulu_form.metadata.form_field_type_provider" class="Sulu\Bundle\FormBundle\Metadata\FormTypeMetadataLoader" public="true">
            <argument type="service" id="sulu_form.dynamic.form_field_type_pool"/>
            <argument type="service" id="sulu_admin.properties_metadata.properties_xml_loader"/>
            <argument type="service" id="sulu_admin.form_metadata.form_xml_loader"/>
[...]

PR #7984, opened 2025-06-03
Replace form metadata name property with key property

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? | no
| BC breaks? | yes 
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Replace form metadata name with key.

#### Why?

Beside the `name` also a `key` was introduced in https://github.com/sulu/sulu/pull/4643 Pull Request. But seems like this are just two fields.

Currently it is very confusing as seen in a lot of tests where the setName was called with a value which make more sense for setTitle (e.g. `setName('Example Template')`). It is kind of intransparent here where name and where key is used and its not filled up in all cases correctly.

Both `<form>` and `<template>` xml both uses the `<key>` and as it we personally mostly talk about template or form keys I would go here with getKey and setKey methods over the names. The only case where `FormMetadata` is created by a `name` is inside block / image map every `<type name="text_block">` is also an own form metadata. Still I would for bc reasons not rename the type attribute name here to key, but internal I would set by the name attribute the form key.

#### Example Usage

```diff
- $formMetadata->getName();
+ $formMetadata->getKey();
```

#### To Do

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/FormMetadata.php
+++ b/src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/FormMetadata.php
@@ -70,7 +70,7 @@ class FormMetadata extends AbstractMetadata
     }
 
     /**
-     * @deprecated since 3.0, use setKey() instead
+     * @deprecated since 3.0, use setName() instead
      */
     #[VirtualProperty]
     #[SerializedName('name')]
@@ -119,7 +119,7 @@ class FormMetadata extends AbstractMetadata
 
         return \count($this->titles)
             ? $this->titles[\array_key_first($this->titles)]
-            : ($this->key ? \ucfirst($this->key) : '');
+            : ($this->key ? \ucfirst(\str_replace(['_', '-'], ' ', $this->key)) : '');
     }
 
     /**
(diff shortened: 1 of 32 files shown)


src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/FormMetadata.php
```diff
@@ -74,14 +69,15 @@ public function __construct()
...
-    {
-        $this->name = $name;
-    }
-
+    /**
+     * @deprecated since 3.0, use setName() instead
+     */
+    #[VirtualProperty]
+    #[SerializedName('name')]
+    #[Exclude(if: "'admin_form_metadata_keys_only' in context.getAttribute('groups')")]
```
alexander-schranz, line 77: Would like to remove this in a upcoming Pull Request. I removed all usage of it in PHP but for removing it from the API the JavaScript part need deeper testing and adjustments.

Added it to the todo list: https://github.com/sulu/sulu/issues/7672

src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/FormMetadata.php
```diff
@@ -123,7 +119,7 @@ public function getTitle(string $locale): string
 
         return \count($this->titles)
             ? $this->titles[\array_key_first($this->titles)]
-            : ($this->name ? \ucfirst($this->name) : '');
+            : ($this->key ? \ucfirst(\str_replace(['_', '-'], ' ', $this->key)) : '');
```
alexander-schranz, line 122: @Prokyonn this is a behaviour change while adopting the test I found `User_details` unexpected for the title and did add this for a more spelling name `User details`. Previously name was not fillup and so title was null for forms see snapshot tests of security and media bundle.

Prokyonn (approved):

PR #7989, opened 2025-06-05
Use new metadata for Preview ContentObjectProvider

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | yes
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | https://github.com/sulu/sulu/pull/7988
| License | MIT

#### What's in this PR?

Use new metadata for Preview ContentObjectProvider.

#### Why?

Get rid of the old structure metadata.

### DIFF
--- a/packages/content/tests/Application/ExampleTestBundle/Resources/config/services.yaml
+++ b/packages/content/tests/Application/ExampleTestBundle/Resources/config/services.yaml
@@ -83,7 +83,7 @@ services:
     example_test.example_preview_object_provider:
         class: Sulu\Content\Infrastructure\Sulu\Preview\ContentObjectProvider
         arguments:
-            - '@sulu_content.content_structure_bridge_factory'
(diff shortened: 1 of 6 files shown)


packages/content/tests/Application/ExampleTestBundle/Resources/config/services.yaml
```diff
@@ -83,7 +83,7 @@ services:
     example_test.example_preview_object_provider:
         class: Sulu\Content\Infrastructure\Sulu\Preview\ContentObjectProvider
```
alexander-schranz, line 84: I think I would rename this class from `ContentObjectProvider` to `ContentPreviewDefaultsProvider` but that should be an own PR. /cc @Prokyonn

Prokyonn (approved):

PR #7997, opened 2025-06-07
Fix null issue when structure type is not found

shochdoerfer:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no 
| License | MIT

#### What's in this PR?

While upgrading one of our Sulu projects to 2.6, I ran into an issue executing the `sulu:reference:refresh` command:
```
12:23:40 CRITICAL  [console] Error thrown while running command "sulu:reference:refresh". Message: "Call to a member function getProperties() on null" ["exception" => Error { …},"command" => "sulu:reference:refresh","message" => "Call to a member function getProperties() on null"]
```

To fix this, I added an explicit type check to the logic in `AbstractDocumentReferenceProvider::updateReferences()`.

Also, I updated the doc clock comment of the `\Sulu\Component\Content\Compat\StructureManagerInterface::getStructure()` method to indicate that null can also be returned, as the logic in the `\Sulu\Component\Content\Compat\StructureManager` class does it when an exception happens. I don't feel this counts as a BC break, but your mileage may vary.

### DIFF
--- a/src/Sulu/Bundle/DocumentManagerBundle/Reference/Provider/AbstractDocumentReferenceProvider.php
+++ b/src/Sulu/Bundle/DocumentManagerBundle/Reference/Provider/AbstractDocumentReferenceProvider.php
@@ -14,6 +14,7 @@ namespace Sulu\Bundle\DocumentManagerBundle\Reference\Provider;
 use Sulu\Bundle\ReferenceBundle\Application\Collector\ReferenceCollector;
 use Sulu\Bundle\ReferenceBundle\Domain\Repository\ReferenceRepositoryInterface;
 use Sulu\Bundle\ReferenceBundle\Infrastructure\Sulu\ContentType\ReferenceContentTypeInterface;
+use Sulu\Component\Content\Compat\StructureInterface;
 use Sulu\Component\Content\Compat\StructureManagerInterface;
 use Sulu\Component\Content\ContentTypeManagerInterface;
 use Sulu\Component\Content\Document\Behavior\ExtensionBehavior;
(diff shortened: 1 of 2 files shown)


shochdoerfer:
Looks like adding null as a return parameter is causing more issues than I expected. How should we cope with this situation?

src/Sulu/Component/Content/Compat/StructureManagerInterface.php
```diff
@@ -24,7 +24,7 @@ interface StructureManagerInterface
      * @param string $key
      * @param string $type
      *
-     * @return StructureInterface
+     * @return StructureInterface|null
```
alexander-schranz, line 27: I would not do that at current state as it indicates something wrong with the data which sure is a 500 error. We would require a deeper research if there is any valid case for `|null` but refresh reference is not.

src/Sulu/Bundle/DocumentManagerBundle/Reference/Provider/AbstractDocumentReferenceProvider.php
```diff
@@ -84,6 +85,9 @@ public function updateReferences($document, string $locale, string $context): vo
         }
 
         $templateStructure = $this->structureManager->getStructure($document->getStructureType(), $this->getStructureType());
+        if (!$templateStructure instanceof StructureInterface) {
+            return;
```
alexander-schranz, line 89: I think a exception would be better here as if a structure is not found there is sure a PHPCR migration missing for a removed template or changed template. Which should end in an error exit or 500 error code.

alexander-schranz (changes requested):
Sorry for the late response. I would more go with a exception in the refresher and keep the structuremanagerinterface unchanged currently.

PR #7999, opened 2025-06-07
Fix `SuluCollector` for profiler tab with new content structure

mamazu:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?
Updating collector in the Symfony profiler bar to use the new content entity.

#### Why?
Otherwise the profiler bar will just be empty.

### DIFF
--- a/src/Sulu/Bundle/WebsiteBundle/DataCollector/SuluCollector.php
+++ b/src/Sulu/Bundle/WebsiteBundle/DataCollector/SuluCollector.php
@@ -11,10 +11,10 @@
 
 namespace Sulu\Bundle\WebsiteBundle\DataCollector;
 
-use Sulu\Component\Content\Compat\StructureInterface;
 use Sulu\Component\Webspace\Analyzer\Attributes\RequestAttributes;
 use Sulu\Component\Webspace\Portal;
 use Sulu\Component\Webspace\Webspace;
+use Sulu\Page\Domain\Model\PageDimensionContent;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\DataCollector\DataCollector;
@@ -26,12 +26,12 @@ class SuluCollector extends DataCollector
     ) {
     }
 
-    public function data($key)
+    public function data(string|int $key): mixed
     {
         return $this->data[$key] ?? null;
     }
 
-    public function collect(Request $request, Response $response, ?\Throwable $exception = null)
+    public function collect(Request $request, Response $response, ?\Throwable $exception = null): void
     {
         if (!$request->attributes->has('_sulu')) {
             return;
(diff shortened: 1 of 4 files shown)


src/Sulu/Bundle/WebsiteBundle/Resources/views/Profiler/layout.html.twig
```diff
@@ -94,41 +92,19 @@
...
+            {{ self.table_row('dimensionClass', collector.data('structure').dimensionClass) }}
             {{ self.table_row('nodeState', collector.data('structure').nodeState) }}
+            {{ self.table_row('locale', collector.data('structure').locale) }}
             {{ self.table_row('published', collector.data('structure').published.format('Y-m-d H:i:s')) }}
-            {{ self.table_row('nodeState', collector.data('structure').nodeState) }}
             {{ self.table_row('navContexts', collector.data('structure').navContexts) }}
-            {{ self.table_row('shadowLocales', collector.data('structure').shadowLocales) }}
-            {{ self.table_row('contentLocales', collector.data('structure').contentLocales) }}
-            {{ self.table_row('shadowOn', collector.data('structure').shadowOn ? 'yes' : 'no') }}
-            {{ self.table_row('shadowBaseLanguage', collector.data('structure').shadowBaseLanguage) }}
```
mamazu, line 129: Is there already something like shadow locales in the new system?

src/Sulu/Bundle/WebsiteBundle/Resources/views/Profiler/layout.html.twig
```diff
@@ -94,41 +92,19 @@
...
-    </table>
     <h2>Structure</h2>
     {% if collector.data('structure') %}
     <table>
         <tbody>
             {{ self.table_row('id', collector.data('structure').id) }}
-            {{ self.table_row('class', collector.data('structure').objectClass) }}
-            {{ self.table_row('path', collector.data('structure').path) }}
-            {{ self.table_row('nodeType', collector.data('structure').nodeType) }}
-            {{ self.table_row('internal', collector.data('structure').internal ? 'yes' : 'no') }}
```
mamazu, line 121: Is there already something like "internal" in the new system?

mamazu:
@Prokyonn You're the expert on the new content structure. Can you have a look at this and tell me if this migration is correct?

src/Sulu/Bundle/WebsiteBundle/Resources/views/Profiler/layout.html.twig
Prokyonn: Yes we have shadows in the new system via the `ShadowInterface` and the `ShadowTrait`

The properties are `shadowLocales` and `shadowLocale`, `shadowLocales` is an `array<string, string>` which shows the mapping e.g. `de` => `en`

src/Sulu/Bundle/WebsiteBundle/Resources/views/Profiler/layout.html.twig
Prokyonn: No this is still missing 😞

src/Sulu/Bundle/WebsiteBundle/DataCollector/SuluCollector.php
```diff
@@ -76,33 +76,25 @@ public function collect(Request $request, Response $response, ?\Throwable $excep
...
 
         $structure = null;
-        if ($request->attributes->has('_route_params')) {
-            $params = $request->attributes->get('_route_params');
-            if (isset($params['structure'])) {
-                /** @var StructureInterface $structureObject */
-                $structureObject = $params['structure'];
+        if ($request->attributes->has('object')) {
+            $object = $request->attributes->get('object');
+            if ($object instanceof PageDimensionContent) {
```
Prokyonn: I think we should make it more flexible, so that it works also for custom content-bundle entities.

So we should check the interfaces individually, e.g.
instanceof WorkflowInterface --> getWorkflowPublished
instanceof ShadowInterface --> shadowLocales
instanceof WebspaceInterface --> webspace
instanceof TemplateInterface --> template
....

Prokyonn (changes requested):

src/Sulu/Bundle/WebsiteBundle/DataCollector/SuluCollector.php
mamazu: Yes makes sense. I did that. Can you check if it's modular enough?

src/Sulu/Bundle/WebsiteBundle/DataCollector/SuluCollector.php
```diff
@@ -76,39 +81,52 @@ public function collect(Request $request, Response $response, ?\Throwable $excep
...
-                    'changer' => $structureObject->getChanger(),
-                    'created' => $structureObject->getCreated(),
-                    'changed' => $structureObject->getChanged(),
-                ];
+        if ($request->attributes->has('object')) {
+            $object = $request->attributes->get('object');
+
+            $structure = [];
+            if ($object instanceof DimensionContentInterface) {
+                $parent = $object->getResource();
```
Prokyonn: I wouldn't call this `parent`, it is either the `resource` or the `contentRichEntity`, but parent may be confusing, because e.g. a page can really have a `parent`

src/Sulu/Bundle/WebsiteBundle/DataCollector/SuluCollector.php
```diff
@@ -76,39 +81,52 @@ public function collect(Request $request, Response $response, ?\Throwable $excep
...
+
+                if (\method_exists($parent, 'getUuid')) {
```
Prokyonn: the resource must implement the `ContentRichEntityInterface` therefore it must have a `getId()` method. In the long run it should be replaced with `getUuid` but we should use the existing interface method here

src/Sulu/Bundle/WebsiteBundle/DataCollector/SuluCollector.php
```diff
@@ -76,39 +81,52 @@ public function collect(Request $request, Response $response, ?\Throwable $excep
...
+
+            if ($object instanceof TemplateInterface) {
+                $structure['template'] = $object->getTemplateKey();
+            }
+
+            if ($object instanceof WorkflowInterface) {
+                $structure['published'] = $object->getWorkflowPublished();
+            }
+
+            if (\method_exists($object, 'getNavigationContexts')) {
```
Prokyonn: For more specific things like `navigationContexts` or `hasChildren` we can check if it is an instance of `PageDimensionContentInterface` or `PageInterface`

src/Sulu/Bundle/WebsiteBundle/Resources/views/Profiler/layout.html.twig
```diff
@@ -110,34 +123,6 @@
...
-            {{ self.table_row('nodeState', collector.data('structure').nodeState) }}
-            {{ self.table_row('published', collector.data('structure').published.format('Y-m-d H:i:s')) }}
-            {{ self.table_row('nodeState', collector.data('structure').nodeState) }}
-            {{ self.table_row('navContexts', collector.data('structure').navContexts) }}
-            {{ self.table_row('shadowLocales', collector.data('structure').shadowLocales) }}
-            {{ self.table_row('contentLocales', collector.data('structure').contentLocales) }}
```
Prokyonn, line 127: These are the `availableLocales` from the `DimensionContentInterface`

Prokyonn (changes requested):

Prokyonn (approved):

alexander-schranz (approved):

alexander-schranz:
@mamazu Thank you!

PR #8000, opened 2025-06-09
Add debug 404 image for image converter errors

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | replaces #3706
| License | MIT
| Documentation PR | -

#### What's in this PR?
Adding an error image in debug with the error code.

#### Why?
Makes it clear to the user that the image is missing but still retains the aspect ratio of the image that should be there.


mamazu:
This has some weird behavior though. When I return an image and give it a 404 response code Chrome shows the image despite the error code but Firefox still shows a broken image.

PR #8024, opened 2025-06-19
Using a clearer error message when asserting response codes

mamazu:
Before:
`HTTP status code 200 is not expected 400!`

After:
`Got status code 200 (Expected: 400)`

This way it's clearer what code we got and what we expected.

### DIFF
--- a/src/Sulu/Bundle/TestBundle/Testing/AssertHttpStatusCodeTrait.php
+++ b/src/Sulu/Bundle/TestBundle/Testing/AssertHttpStatusCodeTrait.php
@@ -53,7 +53,7 @@ trait AssertHttpStatusCodeTrait
                 $message = \implode(\PHP_EOL, \array_slice($message, 0, $debugLength));
 
                 $message = \sprintf(
-                    'HTTP status code %s is not expected %s!' . \PHP_EOL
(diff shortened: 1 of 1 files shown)


src/Sulu/Bundle/TestBundle/Testing/AssertHttpStatusCodeTrait.php
```diff
@@ -53,7 +53,7 @@ protected static function assertHttpStatusCode(int $code, $response, int $debugL
                 $message = \implode(\PHP_EOL, \array_slice($message, 0, $debugLength));
 
                 $message = \sprintf(
-                    'HTTP status code %s is not expected %s!' . \PHP_EOL
+                    'Got status code %s (Expected: %s)' . \PHP_EOL
```
alexander-schranz:
```suggestion
                    'Got HTTP status code %s, but expected %s!' . \PHP_EOL
```

/cc @Prokyonn

alexander-schranz:
If we do this change I would do it on the 2.x version already.

mamazu:
Rebased to 2.6

Prokyonn (approved):

Issue #8025, opened 2025-06-21
Pseudo localization not working


mamazu:
| Q | A
| --- | ---
| Sulu Version | 2.5+
| PHP Version | *
| DB Version | *
| Browser Version | *

# Actual Behavior
When enabling Symfonys pseudo localization feature to test if the UI to test if everything is translated, Sulu crashes in various places.

# Expected Behavior
Enabling the pseudo localization should work.

# Steps to Reproduce
Set this in your config:
```yaml
framework:
    translator:
        pseudo_localization: true

```

PR #8026, opened 2025-06-21
Fix the usage of Symfony's pseudo_localization feature

mamazu:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | #8025 
| License | MIT

#### What's in this PR?
Only calling the `setLocale` function on the translator if it does exist.

#### Why?
Otherwise it will crash.

## Context
This is just a stopgap solution for older versions. I will make another pull request to the 3.0 to use the `LocaleSwitcher` introduced into 6.1.

### DIFF
--- a/src/Sulu/Bundle/RouteBundle/Generator/SymfonyExpressionTokenProvider.php
+++ b/src/Sulu/Bundle/RouteBundle/Generator/SymfonyExpressionTokenProvider.php
@@ -13,7 +13,6 @@ namespace Sulu\Bundle\RouteBundle\Generator;
 
 use Symfony\Component\ExpressionLanguage\ExpressionFunction;
 use Symfony\Component\ExpressionLanguage\ExpressionLanguage;
-use Symfony\Contracts\Translation\LocaleAwareInterface;
 use Symfony\Contracts\Translation\TranslatorInterface;
 
 /**
@@ -22,7 +21,7 @@ use Symfony\Contracts\Translation\TranslatorInterface;
 class SymfonyExpressionTokenProvider implements TokenProviderInterface
 {
     /**
-     * @var TranslatorInterface&LocaleAwareInterface
+     * @var TranslatorInterface
      */
     private $translator;
 
@@ -33,15 +32,6 @@ class SymfonyExpressionTokenProvider implements TokenProviderInterface
 
     public function __construct(TranslatorInterface $translator)
     {
-        if (!$translator instanceof LocaleAwareInterface) {
-            throw new \LogicException(\sprintf(
-                'Expected "translator" in "%s" to be instance of "%s" but "%s" given.',
-                __CLASS__,
-                LocaleAwareInterface::class,
-                \get_class($translator)
-            ));
-        }
-
         $this->translator = $translator;
 
         $this->expressionLanguage = new ExpressionLanguage();
@@ -80,7 +70,7 @@ class SymfonyExpressionTokenProvider implements TokenProviderInterface
             }
 
             if ($entityLocale) {
-                $this->translator->setLocale($entityLocale);
+                $this->setLocale($entityLocale);
             }
 
             $result = $this->expressionLanguage->evaluate($name, [
@@ -93,6 +83,13 @@ class SymfonyExpressionTokenProvider implements TokenProviderInterface
         } catch (\Exception $e) {
             throw new CannotEvaluateTokenException($name, $entity, $e);
         } finally {
+            $this->setLocale($locale);
+        }
+    }
+
+    private function setLocale(string $locale): void
+    {
+        if (\method_exists($this->translator, 'setLocale')) {
             $this->translator->setLocale($locale);
         }
     }
--- a/src/Sulu/Bundle/RouteBundle/Generator/TranslatorWrapper.php
+++ b/src/Sulu/Bundle/RouteBundle/Generator/TranslatorWrapper.php
@@ -19,9 +19,6 @@ use Symfony\Contracts\Translation\TranslatorInterface;
  */
 class TranslatorWrapper implements TranslatorInterface, LocaleAwareInterface
 {
-    /**
-     * @param TranslatorInterface&LocaleAwareInterface $translator
-     */
     public function __construct(private TranslatorInterface $translator)
     {
     }
--- a/src/Sulu/Bundle/RouteBundle/Tests/Unit/TestRoutable.php
+++ b/src/Sulu/Bundle/RouteBundle/Tests/Unit/TestRoutable.php
@@ -1,0 +1,47 @@
+<?php
+
+/*
+ * This file is part of Sulu.
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\RouteBundle\Tests\Unit;
+
+use Sulu\Bundle\RouteBundle\Model\RoutableInterface;
+use Sulu\Bundle\RouteBundle\Model\RouteInterface;
+
+class TestRoutable implements RoutableInterface
+{
+    public ?string $name = null;
+
+    public function __construct(
+        private ?RouteInterface $route = null,
+        private int $id = 1,
+        private string $locale = 'de',
+    ) {
+    }
+
+    public function getId(): int
+    {
+        return $this->id;
+    }
+
+    public function getRoute(): ?RouteInterface
+    {
+        return $this->route;
+    }
+
+    public function setRoute(RouteInterface $route): void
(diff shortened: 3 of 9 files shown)


src/Sulu/Bundle/RouteBundle/Generator/SymfonyExpressionTokenProvider.php
```diff
@@ -93,7 +84,23 @@ public function provide($entity, $name/* , $options = [] */)
...
         }
     }
+
+    private function getLocale(): string
+    {
+        if ($this->translator instanceof LocaleAwareInterface) {
+            return $this->translator->getLocale();
+        }
+
+        return '';
```
alexander-schranz: This will not work in this supported range of Symfony versions correctly the getLocale may exist also when not have LocaleAwareInterface implementations.

I do not think there is any real world case where `$this->translator->getLocale()` will fail even the interfaces don't have that method in all symfony versions. So I think the getLocale method can be removed and the get change reverted. Symfony has added the getLocale as [phpdoc some time ago](https://github.com/symfony/translation-contracts/blob/7d0844cea8bd2631ef5c4d2f339cca396758926c/TranslatorInterface.php#L17) but LocaleAware handling was added later.

alexander-schranz (changes requested):

src/Sulu/Bundle/RouteBundle/Generator/SymfonyExpressionTokenProvider.php
mamazu: Yes, reverted and the other instance of the `LocaleAwareInterface` in the class has been replaced with a `method_exists` call.

src/Sulu/Bundle/RouteBundle/Generator/TranslatorWrapper.php
```diff
@@ -41,6 +38,11 @@ public function setLocale(string $locale): void
 
     public function getLocale(): string
     {
+        // This case should not exist, only with the pseudo locale provider, and it doesn't have a locale.
+        if (!\method_exists($this->translator, 'getLocale')) {
+            return 'en';
+        }
```
alexander-schranz: Don't get this. translatorinterface requires getLocale to exist how is this happening?

alexander-schranz: is there any real case where you find that the getLocale failed?

mamazu: I'll have to check but I found this bug through the article bundle where this was being used and crashed. I think this was also an older version of Symfony where the TranslatorInterface didn't have the getLocale method.

mamazu: Yeah in Symfony 5.4 the interface doesn't have the method so it will crash in the translation wrapper and the SymfonyExpressionTokenProvider (which I reverted)

alexander-schranz: which symfony and sulu version was used? if we add something like this we need to know when we can remove this. Keep in mind that `return 'en'` hardcoded can just end in other errors where `en` is not configured. So even not sure if we should merge it if this isn't easy to support.

mamazu: This is only a fix for Sulu 2.x. In Sulu 3 we are requiring Symfony version 6+ and there the interface has the method and we can remove it. Furthermore we can even use the LocalizationSwitcher. 

The fix there only affects the PseudolocationTranslator (which doesn't need a locale) so hardcoding anything is fine in my opinion.

alexander-schranz: Still same question. What version the error with getLocale appeared. Please provide what versions the `composer info` outputted for symfony, sulu and if releated to article bundle article bundle.

mamazu:
```
sulu/article-bundle                       2.6.0             Bundle for managing localized content-rich entities like blog-posts in the Sulu content management system
sulu/community-bundle                     3.0.x-dev 7a4e537 Sulu Bundle for building a community
sulu/sulu                                 2.5.19            Core framework that implements the functionality of the Sulu content management system
symfony/framework-bundle                  6.3.12            Provides a tight integration between Symfony components and the Symfony full-stack framework
```
But I just check the stack traces they're both in the sulu/sulu one in the website package and the other one in the routing bundle.
And now the tests also run into the same problem.

alexander-schranz: which class you get into the service? Which version of the symfony translation contracts is installed.

The 6.3.12 even has the getLocale inside the interface: https://github.com/symfony/symfony/blob/v6.3.12/src/Symfony/Contracts/Translation/TranslatorInterface.php

mamazu: Ah yes, you are correct. It's getting the Symfony\Component\Translation\PseudoLocalizationTranslator and while it does have a 'getLocale' function (at least in 6.3+). It doesn't have a setLocale. That's what's causing the crash. But for older versions it might still crash.

alexander-schranz: then I think we can revert this change and directly go with ->translator->getLocale() and only need to keep the setLocale check. I don't think that there is any version of the translator service which crashes calling the getter. So think setter check is only think which we need tackle here.

mamazu: And what do we do with the failing test case?

src/Sulu/Bundle/RouteBundle/Tests/Unit/TestRoutable.php
```diff
@@ -0,0 +1,47 @@
...
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\RouteBundle\Tests\Unit;
+
+use Sulu\Bundle\RouteBundle\Model\RoutableInterface;
+use Sulu\Bundle\RouteBundle\Model\RouteInterface;
+
+class TestRoutable implements RoutableInterface
```
alexander-schranz, line 17: don't see any requirement for this change inside this PR?

src/Sulu/Bundle/WebsiteBundle/EventListener/TranslatorListener.php
```diff
@@ -45,6 +45,10 @@ public function onKernelRequest(RequestEvent $event)
             return;
         }
 
+        if (!\method_exists($this->translator, 'setLocale')) {
+            return;
+        }
```
alexander-schranz, line 50: I hope this does not fail more. And only has effects on translation and not somewhere else.

src/Sulu/Bundle/RouteBundle/Tests/Unit/TestRoutable.php
mamazu: Well it was used in two tests and it made it easier to use a real class in the SymfonyExpressionProvider as well instead of the mock object.

src/Sulu/Bundle/WebsiteBundle/EventListener/TranslatorListener.php
mamazu: Yes this works. And for all "normal" translation cases this will never run.

alexander-schranz (approved):

src/Sulu/Bundle/WebsiteBundle/EventListener/TranslatorListener.php
alexander-schranz: did miss that, make sense

src/Sulu/Bundle/RouteBundle/Generator/TranslatorWrapper.php
mamazu: Do we just not run the test for Symfony 5.4?

alexander-schranz:
The failed test looks like issue of mocking. Can you try Translator instead of TranslatorInterface or if available on low level willImplement LocaleAwareInterface, as it can not mock `@method getLocale`

mamazu:
Test fixed by removing it. We're not going to fix the behaviour in Symfony 5.4 so there is no point in adding an additional test. If I were to use the `Translator` which always has the method, the test is the same as the one above it. And we can't also implement the LocaleAware interface as this does not exist in 5.4.

PR #8027, opened 2025-06-23
Bugfix: Segment redirect

### LINKED ISSUE #7695: Segments switch redirects to same page and switches back segment
| Q | A
| --- | ---
| Sulu Version | 2.6.4

# Actual Behavior

The twig variable "segments" on a page (only assigned to one segment), still points to itself on the other segment.
So if you use the link for the not assigned segment it will switch segment and redirects to itself, resulting in switching back segment, because this page only exists in the assigned segment.

# Expected Behavior

Like the language-chooser, only assigned segments should point to the selected segment of this page, all others should point to the startpage.

# Steps to Reproduce

Add segments to the webspace, create a page and only assign one segment. Check the urls in variable "segments".

# Possible Solutions

As Quickfix you can check for extenstion.excerpt.segments.
if excerpt segments is not empty and segmentKey not in excerpt segments them generate the switch url by yourself with:
```
path('sulu_website.segment_switch', { segment: segmentKey, url: sulu_content_root_path() })
```

I guess the same check could be used when creating the segments variable for the render.

draconivis:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #7695
| Related issues/PRs | #7695

#### What's in this PR?

If a page has a segment assigned, add a link to the homepage for the other segments instead of a switch link, just like the language switcher

#### Why?

#7695

### DIFF
--- a/src/Sulu/Bundle/WebsiteBundle/Resolver/ParameterResolver.php
+++ b/src/Sulu/Bundle/WebsiteBundle/Resolver/ParameterResolver.php
@@ -143,10 +143,14 @@ class ParameterResolver implements ParameterResolverInterface
         $segments = [];
         foreach ($webspace->getSegments() as $segment) {
             $segmentKey = $segment->getKey();
-            $segmentSwitchUrls[$segmentKey] = $this->segmentSwitchUrl . '?segment=' . $segmentKey . '&url=' . $url;
+            if (empty($segmentData = $structureData['extension']['excerpt']['segments']) || $segmentKey === $segmentData[$structureData['webspaceKey']]) {
+                $segmentUrl = $this->segmentSwitchUrl . '?segment=' . $segmentKey . '&url=' . $url;
+            } else {
+                $segmentUrl = $this->webspaceManager->findUrlByResourceLocator('/', null, $requestAnalyzer->getCurrentLocalization()->getLocale());
+            }
             $segments[$segmentKey] = [
                 'title' => $segment->getTitle($requestAnalyzer->getCurrentLocalization()->getLocale()),
-                'url' => $this->segmentSwitchUrl . '?segment=' . $segmentKey . '&url=' . $url,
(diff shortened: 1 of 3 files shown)


src/Sulu/Bundle/WebsiteBundle/Resolver/ParameterResolver.php
```diff
@@ -143,10 +143,14 @@ public function resolve(
         $segments = [];
         foreach ($webspace->getSegments() as $segment) {
             $segmentKey = $segment->getKey();
-            $segmentSwitchUrls[$segmentKey] = $this->segmentSwitchUrl . '?segment=' . $segmentKey . '&url=' . $url;
+            if (empty($segmentData = $structureData['extension']['excerpt']['segments']) || $segmentKey === $segmentData[$structureData['webspaceKey']]) {
```
Prokyonn: The new line in `ParameterResolver` reads `$structureData['extension']['excerpt']['segments']` and `[$structureData['webspaceKey']]` directly. When those keys are missing, PHP throws an "undefined array key" warning, and in dev that becomes a 500. You can already see it in this PR: the existing `testResolve` uses `'excerpt' => []`, and CI now shows `Warnings: 1` for WebsiteBundle.

Could you read it with `??`, once before the loop, like `ContentRouteProvider` does?

```php
$pageSegmentKey = $structureData['extension']['excerpt']['segments'][$webspace->getKey()] ?? null;
```

draconivis: fixed. the variable is now read before the loop. I decided to stick with reading the `webspaceKey` from `structureData` as i would've to adjust several tests if i were to use `$webspace->getKey()`

Prokyonn (approved):

PR #8030, opened 2025-06-25
Feature: List item action title added

havrankova:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?
List item action title added.

#### Why?
To display of list item action title when mouse hover.

#### Example Usage

getItemActionConfig(key, parameters) {
    const {disable_for_empty_selection: disableForEmptySelection = false} = this.options;
    const keyObject = key;

    return {
      type: 'button',
      label: translate('sulu_admin.order.generate_xml'),
      icon: 'su-document',
      title: translate('sulu_admin.order.generate_xml'),
      onClick: () => this.handleClick(keyObject),
    };
  }

#### To Do

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Table/ButtonCell.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Table/ButtonCell.js
@@ -10,6 +10,7 @@ type Props = {|
     onClick: ?(rowId: string | number, rowIndex: number) => void,
     rowId: string | number,
     rowIndex: number,
+    title?: string,
 |};
 
 export default class ButtonCell extends React.PureComponent<Props> {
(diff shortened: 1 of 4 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/components/Table/ButtonCell.js
```diff
@@ -29,11 +30,12 @@ export default class ButtonCell extends React.PureComponent<Props> {
...
             disabled,
             icon,
+            title,
         } = this.props;
 
         return (
             <Cell className={tableStyles.buttonCell}>
-                <button disabled={disabled} onClick={this.handleClick} type="button">
+                <button disabled={disabled} onClick={this.handleClick} type="button" title={title}>
                     <Icon name={icon} />
```
alexander-schranz, line 39: Title attributes should not be used as they effect accessibility in negative ways, we should go with a `aria-label` or better use our Tooltip component here if the title is set https://jsdocs.sulu.io/2.6/#tooltip

alexander-schranz (changes requested):
Thx for the pull request. The title attributes has more negative effects as positive once for accessibility and should not be used in our admin, aria-label should be used instead, but for visible effects we should go with the tooltip component here.

If you need any help let us know.

PR #8032, opened 2025-06-30
Add SmartContent support for Sulu 3.0

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | yes
| Deprecations? | no 
| Related issues/PRs | #7672 
| License | MIT

#### What's in this PR?

Adds support for SmartContents in the new content storage architecture for Sulu 3.0
Implements the DataProvider for
- pages
- articles
- media
- contacts
- accounts

#### Why?

Add SmartContent functionality for 3.0

#### To Do

- [ ] Add tests
- [x] Remove old SmartContent code
- [x] Add SnippetsDataProvider


packages/article/src/Infrastructure/Doctrine/Repository/ArticleRepository.php
```diff
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
```
wachterjohannes: no strict_types

packages/content/src/Application/PropertyResolver/Resolver/SmartContentPropertyResolver.php
```diff
@@ -0,0 +1,223 @@
...
+         *     tags_parameter: string,
+         *     categories_parameter: string,
+         *     website_tags_operator: 'AND'|'OR',
+         *     website_categories_operator: 'AND'|'OR',
+         *     exclude_duplicates: bool,
+         *     provider: string,
+         *     } $parameters
+         */
+        $parameters = \array_merge([
+            'provider' => 'pages', // TODO Should we use default provider for backwards compatibility?
```
wachterjohannes: not sure what you mean with that

packages/content/src/Application/SmartResolver/Resolver/SmartContentSmartResolver.php
```diff
@@ -0,0 +1,105 @@
...
+            ...$filters,
+            ...$sortBys,
+            'provider' => $provider,
+            'page' => $page,
+            'hasNextPage' => null !== $limit && ($total > ($limit * $page)),
+            'paginated' => null !== $limit,
+            'total' => $total,
+            'maxPage' => (null !== $limit) ? (int) \ceil($total / $limit) : null,
+            'limit' => $limit,
+            // categoryRoot
```
wachterjohannes: add a todo

wachterjohannes (changes requested):
@Prokyonn Code looks very good - remove all the strict-types but the rest is good

packages/content/src/Application/PropertyResolver/Resolver/SmartContentPropertyResolver.php
Prokyonn: in 2.6 the default "provider" is always `pages`

packages/content/src/Application/PropertyResolver/Resolver/SmartContentPropertyResolver.php
```diff
@@ -28,6 +30,8 @@ class SmartContentPropertyResolver implements PropertyResolverInterface
 {
     public function __construct(
         private RequestStack $requestStack,
+        private RequestAnalyzerInterface $requestAnalyzer,
+        private ?TargetGroupStoreInterface $targetGroupStore,
```
alexander-schranz: Lets discuss this tomorrow. I think the general smart content service should not have knowledge about page specific things (webspace, segments, ...) and the page provider should handle that.

packages/content/src/Application/PropertyResolver/Resolver/SmartContentPropertyResolver.php
```diff
@@ -0,0 +1,231 @@
...
+     *     includeSubFolders?: bool|null,
+     *     excludeDuplicates?: bool,
+     *     audienceTargeting?: bool
+     * } $data
+     * @param array<string, mixed> $params
+     */
+    public function resolve(mixed $data, string $locale, array $params = []): ContentView
+    {
+        if (!\is_array($data)) { // @phpstan-ignore function.alreadyNarrowedType
+            return ContentView::create($data, $params);
```
alexander-schranz: If the expected type is not correct, a property resolver should always return `null` for single values or empty list `[]` for expected multi values.

```suggestion
            return ContentView::create([], $params);
```

So the twig rendering does not crash.

Sadly the smart content is maybe hard or even impossible to narrow down from mixed to our expected array correctly.

packages/content/src/Application/PropertyResolver/Resolver/SmartContentPropertyResolver.php
```diff
@@ -0,0 +1,231 @@
...
+     *     presentAs?: string|null,
+     *     includeSubFolders?: bool|null,
+     *     excludeDuplicates?: bool,
+     *     audienceTargeting?: bool
+     * } $data
+     * @param array<string, mixed> $params
+     */
+    public function resolve(mixed $data, string $locale, array $params = []): ContentView
+    {
+        if (!\is_array($data)) { // @phpstan-ignore function.alreadyNarrowedType
```
alexander-schranz: We should catch also `[1,2,3]` or any other list like arrays like `[['type' => 'text', ...]` here as that indicate also as not valide smart content data.

Two possibilities:

Fast check:

```php
if (!\is_array($data) || \array_is_list($data)) {
```

If we want to keep `[]` as valid smart content we should go with:

```php
if (!\is_array($data) || ([] !== $data && \array_is_list($data)) {
```

packages/content/src/Application/PropertyResolver/Resolver/SmartContentPropertyResolver.php
alexander-schranz: as discussed we will create a:

```php
    /**
     * @var iterable<SmartContentFiltersVisitorInterface>
     */
    public function __construct(
        private RequestStack $requestStack,
        private iterable $smartContentFiltersVisitors,
```

packages/content/src/Application/PropertyResolver/Resolver/SmartContentPropertyResolver.php
```diff
@@ -0,0 +1,231 @@
...
+            'websiteTags' => \array_filter(\explode(',', $request->query->getString($parameters['tags_parameter']))),
+            'websiteTagOperator' => \strtoupper($parameters['website_tags_operator']),
+
+            // Types
+            'types' => $data['types'] ?? [],
+            'typesOperator' => 'OR',
+
+            // Other filters
+            'locale' => $parameters['locale'],
+            'webspaceKey' => $webspace?->getKey() ?? null,
```
alexander-schranz: As discussed the webspace Key is not required, the page smart content requires the dataSource which then automatically filters by specific webspace.

packages/article/src/Infrastructure/Sulu/Content/ArticleSmartContentProvider.php
```diff
@@ -0,0 +1,300 @@
+<?php
+
+declare(strict_types=1);
```
alexander-schranz: we should currently remove this as we not enforcing it.

packages/article/src/Infrastructure/Sulu/Content/ArticleSmartContentProvider.php
```diff
@@ -0,0 +1,300 @@
...
+        $queryBuilder->addSelect('filterDimensionContent.title');
+
+        /** @var OrderBy[]|null $queryParts */
+        $queryParts = $queryBuilder->getDQLPart('orderBy');
+        foreach ($queryParts ?? [] as $orderBy) {
+            foreach ($orderBy->getParts() as $order) {
+                [$column] = \explode(' ', $order);
+                $queryBuilder->addSelect($column);
+            }
+        }
```
alexander-schranz: we should add a comment why we do this:


```suggestion
        // distinct require all sorted columns to be also selected — so we are selecting them
        $queryParts = $queryBuilder->getDQLPart('orderBy');
        foreach ($queryParts ?? [] as $orderBy) {
            foreach ($orderBy->getParts() as $order) {
                [$column] = \explode(' ', $order);
                $queryBuilder->addSelect($column);
            }
        }
```

packages/content/src/Application/ContentResolver/Value/ResolvableInterface.php
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Content\Application\ContentResolver\Value;
+
+interface ResolvableInterface
```
alexander-schranz, line 17:
```suggestion
/**
 * @internal This class is intended for internal use only within the package/library. Modifying or depending on this class may result in unexpected behavior and is not supported.
 */
interface ResolvableInterface
```

packages/content/src/Infrastructure/Doctrine/JoinFilterTrait.php
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Content\Infrastructure\Doctrine;
+
+use Doctrine\ORM\QueryBuilder;
+
+trait JoinFilterTrait
```
alexander-schranz: Don't like to use a trait for this.

What you think about a `SmartContentQueryEnhancer` service instead.

packages/page/src/Infrastructure/Sulu/Content/PageSmartContentProvider.php
```diff
@@ -0,0 +1,358 @@
...
+        $queryBuilder->addSelect('filterDimensionContent.title');
+
+        /** @var OrderBy[]|null $queryParts */
+        $queryParts = $queryBuilder->getDQLPart('orderBy');
+        foreach ($queryParts ?? [] as $orderBy) {
+            foreach ($orderBy->getParts() as $order) {
+                [$column] = \explode(' ', $order);
+                $queryBuilder->addSelect($column);
+            }
+        }
```
alexander-schranz: We could use this part into the `SmartContentQueryEnhancer` service.

src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/OptionMetadata.php
```diff
@@ -32,7 +32,7 @@ class OptionMetadata
     protected $type;
 
     /**
-     * @var string|int|OptionMetadata[]
+     * @var bool|int|string|float|OptionMetadata[]|null|mixed
```
alexander-schranz: `float` is not supported by our parser and is kept as string:

https://github.com/sulu/sulu/blob/512639aebb408774d726e38be715c5f9e851c97e/src/Sulu/Bundle/AdminBundle/Metadata/XmlParserTrait.php#L26

https://github.com/sulu/sulu/blob/512639aebb408774d726e38be715c5f9e851c97e/src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/Parser/PropertiesXmlParser.php#L391-L392

```suggestion
     * @var bool|int|string|null|OptionMetadata[]
```


We should not support mixed here.

src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/OptionMetadata.php
```diff
@@ -81,15 +81,15 @@ public function getType()
     }
 
     /**
-     * @return int|string|OptionMetadata[]
+     * @return bool|int|string|float|OptionMetadata[]|mixed|null
```
alexander-schranz: same as above no float, no mixed

```suggestion
     * @return bool|int|string|null|OptionMetadata[]
```

src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/OptionMetadata.php
```diff
@@ -81,15 +81,15 @@ public function getType()
...
      */
     public function getValue()
     {
         return $this->value;
     }
 
     /**
-     * @param int|string|OptionMetadata[] $value
+     * @param bool|int|string|float|OptionMetadata[]|mixed $value
```
alexander-schranz:
```suggestion
     * @param bool|int|string|null|OptionMetadata[]
```

packages/content/src/Infrastructure/Doctrine/JoinFilterTrait.php
alexander-schranz: The question is if it is even part of the `AdminBundle`. Lets think about having Admin + Media Bundle enabled and want to use the smart content field type. The smart content item controller part of admin bundle should be able to resolve the ids without the requirement to the content bundle.

src/Sulu/Bundle/ContactBundle/Infrastructure/Sulu/Content/SmartContent/AccountSmartContentProvider.php
```diff
@@ -0,0 +1,176 @@
...
+
+use Doctrine\ORM\EntityManagerInterface;
+use Doctrine\ORM\QueryBuilder;
+use Sulu\Bundle\AdminBundle\SmartContent\Configuration\Builder;
+use Sulu\Bundle\AdminBundle\SmartContent\Configuration\ProviderConfigurationInterface;
+use Sulu\Bundle\AdminBundle\SmartContent\SmartContentProviderInterface;
+use Sulu\Bundle\ContactBundle\Admin\ContactAdmin;
+use Sulu\Bundle\ContactBundle\Entity\AccountInterface;
+use Sulu\Bundle\ContactBundle\Infrastructure\Sulu\Content\ResourceLoader\AccountResourceLoader;
+use Sulu\Content\Infrastructure\Doctrine\JoinFilterTrait;
```
alexander-schranz: See comment above, trait should be not used enhancer service instead which may live in the admin bundle not part of content to avoid connection to the content part.

packages/content/src/Application/Visitor/SmartContentFiltersVisitorInterface.php
```diff
@@ -0,0 +1,26 @@
...
+    /**
+     * This method is used to visit the data and apply any necessary filters to it, returning the modified data structure.
+     *
+     * @param array<string, mixed> $data
+     * @param array<string, mixed> $filters
+     * @param array<string, mixed> $parameters
+     *
+     * @return array<string, mixed>
+     */
+    public function visit(array $data, array $filters, array $parameters): array;
```
alexander-schranz, line 25: Not 100% sure if I would provide $data and $parameters here 🤔 .

alexander-schranz (changes requested):

packages/content/src/Infrastructure/Doctrine/JoinFilterTrait.php
Prokyonn: Added a service in the AdminBundle

packages/content/src/Application/Visitor/SmartContentFiltersVisitorInterface.php
Prokyonn: As discussed, we need $parameters in the `MediaSmartContentFiltersVisitor` and $data in the `AudienceTargetingSmartContentFiltersVisitor`

alexander-schranz:
Looks fine, we may should recheck where to add final and readonly keywords, but that own PRs added it to our content storage checklist.

PR #8035, opened 2025-07-04
Deprecate the LocalizationProviderCompilerPass

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | yes 
| Fixed tickets | -
| Related issues/PRs | #7402 
| License | MIT

#### What's in this PR?
Deprecating the `RegisterLocalizationProvidersPass` so that we can remove it in 3.0

#### Why?
Symfony already implements this behaviour which is much more convenient.

#### Example Usage
```xml
<service id="sulu.core.localization_manager" 
    class="Sulu\Component\Localization\Manager\LocalizationManager" 
    public="true"
>
    <argument type="tagged_iterator" tag="sulu.localization_provider" />
</service>
```

### DIFF
--- a/src/Sulu/Bundle/CoreBundle/DependencyInjection/Compiler/RegisterLocalizationProvidersPass.php
+++ b/src/Sulu/Bundle/CoreBundle/DependencyInjection/Compiler/RegisterLocalizationProvidersPass.php
@@ -16,9 +16,12 @@ use Symfony\Component\DependencyInjection\ContainerBuilder;
 
 /**
  * Adds tagged services to the LocalizationManager.
+ *
+ * @deprecated since 2.6 use tagged_iterators instead
  */
 class RegisterLocalizationProvidersPass implements CompilerPassInterface
 {
+    /** @deprecated since 2.6 use Sulu\Component\Localization\Provider\LocalizationProviderInterface::SERVICE_TAG */
     public const LOCALIZATION_PROVIDER_TAG = 'sulu.localization_provider';
 
     public function process(ContainerBuilder $container)
--- a/src/Sulu/Bundle/CoreBundle/DependencyInjection/Compiler/RegisterLocalizationProvidersPass.php
+++ b/src/Sulu/Bundle/CoreBundle/DependencyInjection/Compiler/RegisterLocalizationProvidersPass.php
@@ -21,7 +21,6 @@ use Symfony\Component\DependencyInjection\ContainerBuilder;
  */
 class RegisterLocalizationProvidersPass implements CompilerPassInterface
 {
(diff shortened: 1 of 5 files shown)


src/Sulu/Component/Localization/Manager/LocalizationManagerInterface.php
```diff
@@ -30,5 +30,8 @@ public function getLocalizations(): array;
      */
     public function getLocales(): array;
 
+    /**
+     * @deprecated since Use the constructor instead
+     */
```
mamazu, line 35: When this is deprecated and removed the `LocalizationManagerInterface` and the `LocalizationProviderInterface` are the same (just different method names). So I would suggest deprecating the entire interface and using the LocalizationProviderInterface everywhere.

mamazu (commented):
.

src/Sulu/Component/Localization/Provider/LocalizationProviderInterface.php
```diff
@@ -18,6 +18,8 @@
  */
 interface LocalizationProviderInterface
 {
+    public const SERVICE_TAG = 'sulu.localization_provider';
```
mamazu: Still needs a `registerForAutoconfiguration` with this tag.

src/Sulu/Bundle/CoreBundle/DependencyInjection/Compiler/RegisterLocalizationProvidersPass.php
```diff
@@ -16,9 +16,12 @@
 
 /**
  * Adds tagged services to the LocalizationManager.
+ *
+ * @deprecated since 2.6 use tagged_iterators instead
  */
 class RegisterLocalizationProvidersPass implements CompilerPassInterface
 {
+    /** @deprecated since 2.6 use Sulu\Component\Localization\Provider\LocalizationProviderInterface::SERVICE_TAG */
```
alexander-schranz:
```suggestion
```

src/Sulu/Component/Localization/Provider/LocalizationProviderInterface.php
alexander-schranz:
```suggestion
```

src/Sulu/Component/Localization/Provider/LocalizationProviderInterface.php
```diff
@@ -18,6 +18,7 @@
  */
 interface LocalizationProviderInterface
 {
+
```
alexander-schranz:
```suggestion
```

alexander-schranz (approved):

PR #8037, opened 2025-07-04
Deprecate FormatCacheClearerCompilerPass

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | yes
| Fixed tickets | -
| Related issues/PRs | #7402 
| License | MIT

#### What's in this PR?
Deprecating the `FormatCacheClearerCompilerPass` and auto registering all classes that implement the `CacheClearCompilerPass` interface.

#### Why?
Symfony already implements this feature. Why not use it?

#### Example Usage
```xml
<service id="sulu_media.format_cache_clearer" class="Sulu\Bundle\MediaBundle\Media\FormatCache\FormatCacheClearer">
    <argument type="tagged_iterator" tag="sulu_media.format_cache" index-by="alias" />
</service>
```

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/Media/FormatCache/FormatCacheClearer.php
+++ b/src/Sulu/Bundle/MediaBundle/Media/FormatCache/FormatCacheClearer.php
@@ -19,9 +19,17 @@ use Sulu\Bundle\MediaBundle\Media\Exception\CacheNotFoundException;
 class FormatCacheClearer implements FormatCacheClearerInterface
 {
     /**
-     * @var FormatCacheInterface[]
+     * @var array<string, FormatCacheInterface>
      */
     private $caches = [];
+
+    /**
+     * @param iterable<string, FormatCacheInterface> $caches
+     */
+    public function __construct(iterable $caches = [])
+    {
+        $this->caches = [...$caches];
+    }
 
     /**
      * Clear all or the given cache.
@@ -47,6 +55,8 @@ class FormatCacheClearer implements FormatCacheClearerInterface
 
     /**
      * Adds a cache to the aggregate.
+     *
+     * @deprecated use the constructor instead
      *
      * @param FormatCacheInterface $cache The cache
      * @param string $alias The cache alias
--- a/UPGRADE-2.x.md
+++ b/UPGRADE-2.x.md
@@ -7,6 +7,7 @@
 Deprecated compiler passes:
 
 * `Sulu\Bundle\SecurityBundle\DependencyInjection\Compiler\AccessControlProviderPass` -> `tagged_iterator`
+* `Sulu\Bundle\MediaBundle\DependencyInjection\FormatCacheClearerCompilerPass` (sulu_media.format_cache)
 
 ## 2.6.11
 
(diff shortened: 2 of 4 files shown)


src/Sulu/Bundle/MediaBundle/Media/FormatCache/FormatCacheClearerInterface.php
```diff
@@ -18,6 +18,8 @@
  */
 interface FormatCacheClearerInterface
 {
+    public const SERVICE_TAG = 'sulu_media.format_cache';
+
```
alexander-schranz: I don't think our interfaces should create dependency injection specific constants. So would remove this constant as it is unrelated to application logic.

src/Sulu/Bundle/MediaBundle/SuluMediaBundle.php
```diff
@@ -41,6 +42,10 @@ public function build(ContainerBuilder $container): void
             $container
         );
 
+        $container->registerForAutoconfiguration(FormatCacheClearerInterface::class)
+            ->addTag(FormatCacheClearerInterface::SERVICE_TAG)
+        ;
```
alexander-schranz: This is called in ourcases in the `Extension` not in `Bundle` classes see: https://github.com/search?q=repo%3Asulu%2Fsulu%20registerForAutoconfiguration&type=code

src/Sulu/Bundle/MediaBundle/Media/FormatCache/FormatCacheClearer.php
```diff
@@ -19,10 +19,18 @@
...
     private $caches = [];
 
+    /**
+     * @param iterable<string, FormatCacheInterface> $caches
+     */
+    public function __construct(iterable $caches = [])
+    {
+        $this->caches = [...$caches];
+    }
+
```
alexander-schranz, line 33: This is always an empty array on this branch or I'm missing something?

alexander-schranz (changes requested):

src/Sulu/Bundle/MediaBundle/Media/FormatCache/FormatCacheClearer.php
mamazu: Yes this is correct because otherwise it would have the cache clearer twice once from the constructor and once from the `add` calls. In 3.0 this will be non empty.

src/Sulu/Bundle/MediaBundle/Media/FormatCache/FormatCacheClearerInterface.php
mamazu: Removed.

src/Sulu/Bundle/MediaBundle/SuluMediaBundle.php
mamazu: Moved.

alexander-schranz:
Would deprecate the add method also, but only as phpdoc as it still called in core.

mamazu:
@alexander-schranz Done, should I also apply these changes to the other PRs that deprecate compiler passes?

alexander-schranz (approved):

alexander-schranz (approved):

PR #8038, opened 2025-07-06
Deprecate the `AddMetadataProviderPass`

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | yes
| Fixed tickets | -
| Related issues/PRs | #7402 
| License | MIT
| Documentation PR | -

#### What's in this PR?
Deprecating `AddMetadataProviderPass` and suggesting the tagged_locator instead.

#### Why?
- Symfony already has this feature.
- As far as I know the locator also only loads the services when they're actually used, so maybe also a performance thing?

#### Example Usage
```xml
<service id="Sulu\Bundle\AdminBundle\Metadata\MetadataProviderRegistry">
    <argument type="tagged_locator" tag="sulu_admin.metadata_provider" index-by="type" />
</service>
```

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Metadata/MetadataProviderInterface.php
+++ b/src/Sulu/Bundle/AdminBundle/Metadata/MetadataProviderInterface.php
@@ -13,7 +13,5 @@ namespace Sulu\Bundle\AdminBundle\Metadata;
 
 interface MetadataProviderInterface
 {
-    public const SERVICE_TAG = 'sulu_admin.metadata_provider';
-
     public function getMetadata(string $key, string $locale, array $metadataOptions): MetadataInterface;
 }
--- a/src/Sulu/Bundle/AdminBundle/Metadata/MetadataProviderInterface.php
+++ b/src/Sulu/Bundle/AdminBundle/Metadata/MetadataProviderInterface.php
@@ -13,5 +13,7 @@ namespace Sulu\Bundle\AdminBundle\Metadata;
 
 interface MetadataProviderInterface
 {
+    public const SERVICE_TAG = 'sulu_admin.metadata_provider';
+
     public function getMetadata(string $key, string $locale, array $metadataOptions): MetadataInterface;
 }
--- a/src/Sulu/Bundle/AdminBundle/Metadata/MetadataProviderRegistry.php
+++ b/src/Sulu/Bundle/AdminBundle/Metadata/MetadataProviderRegistry.php
@@ -11,14 +11,31 @@
 
 namespace Sulu\Bundle\AdminBundle\Metadata;
 
+use Psr\Container\ContainerInterface;
 use Sulu\Bundle\AdminBundle\Exception\MetadataProviderNotFoundException;
 
 class MetadataProviderRegistry
 {
     private $metadataProviders = [];
 
+    public function __construct(
+        private ?ContainerInterface $container = null,
+    ) {
+        if (null === $container) {
+            trigger_deprecation('sulu/sulu', '2.6', 'Not using the tagged_locator to pass MetadataProviders is deprecated');
+        }
+    }
+
     public function getMetadataProvider(string $type): MetadataProviderInterface
     {
+        if (null !== $this->container) {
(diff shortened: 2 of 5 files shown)


alexander-schranz:
We should never call `trigger_deprecation` without providing a way to have not triggered that deprecation. So instead we should just keep the PR simple add the `@deprecated` to the `add` method and refactor the whole code in 3.0

src/Sulu/Bundle/AdminBundle/Metadata/MetadataProviderInterface.php
```diff
@@ -13,5 +13,7 @@
 
 interface MetadataProviderInterface
 {
+    public const SERVICE_TAG = 'sulu_admin.metadata_provider';
+
```
alexander-schranz: not a fan of this constants would remove it, its are literals which only matter during service registrations and has nothing todo with the application or domain interfaces.

mamazu: Maybe moving it to the bundle instead? Because I still think they could be helpful. Especially when we're moving to a php based configuration for services.

src/Sulu/Bundle/AdminBundle/Metadata/MetadataProviderRegistry.php
```diff
@@ -11,23 +11,48 @@
...
 
+use Psr\Container\ContainerInterface;
 use Sulu\Bundle\AdminBundle\Exception\MetadataProviderNotFoundException;
 
 class MetadataProviderRegistry
 {
     private $metadataProviders = [];
 
+    /**
+     * @param ContainerInterface<MetadataProviderInterface> $container
```
alexander-schranz: this kind of ContainerInterface does not have a generic value / type.

src/Sulu/Bundle/AdminBundle/Metadata/MetadataProviderInterface.php
alexander-schranz: would not change it in 2.6 and in 3.0 just inline it without any constant.

src/Sulu/Bundle/AdminBundle/Metadata/MetadataProviderRegistry.php
```diff
@@ -11,23 +11,45 @@
...
 class MetadataProviderRegistry
 {
     private $metadataProviders = [];
 
+    public function __construct(
+        private ?ContainerInterface $container = null,
+    ) {
+        if (null === $container) {
+            trigger_deprecation('sulu/sulu', '2.6', 'Not using the tagged_locator to pass MetadataProviders is deprecated');
+        }
```
alexander-schranz: we should not trigger a deprecation here as the container is never set in this case by sulu core and would throw unnecessary not fixable deprecation.

src/Sulu/Bundle/AdminBundle/Metadata/MetadataProviderRegistry.php
```diff
@@ -11,23 +11,45 @@
...
         return $this->metadataProviders[$type];
     }
 
+    /**
+     * @deprecated since Sulu 2.6 use the constructor instead
+     */
     public function addMetadataProvider(string $type, MetadataProviderInterface $metadataProvider)
     {
+        trigger_deprecation('sulu/sulu', '2.6', 'Do not add metadata Providers manually. Use the constructor');
+
```
alexander-schranz: same this is still called. deprecation should only be triggered if they are not longer trigger by our own code.

alexander-schranz:
```suggestion
```

src/Sulu/Bundle/AdminBundle/Metadata/MetadataProviderRegistry.php
alexander-schranz:
```suggestion
```

alexander-schranz (approved):

PR #8042, opened 2025-07-08
Adding admin url in profiler bar

### LINKED ISSUE #8041: Link to the admin in the website profiler
# Problem description
When debugging a page generated by Sulu, it would be nice if you could also have a link in the profiler bar in the frontend like this:

Image

# Proposed solution

We would need to implement a solution that can handle content of various different bundles (News, Content, Article and Sulu pages). The best idea would be if the url generator would be extensible so that the bundles can generate their own routes.

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | yes
| Deprecations? | no
| Fixed tickets | fixes #8041 
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

### What's in this PR?
A way to generate a link to the admin panel for a given site.

### Why?
For debugging purposes it's nice to get the admin url for the current page to make it easier to find where to configured.

### Problems
* There is no way to generate the admin base url (I think that might be because of the different kernels admin/website)
* Support more formats out of the box? (like article bundle etc)
* Is the admin format configurable somewhere then we need a different approach.
* Only register the service in dev. Currently the only way I see is either with attributes + autowiring or php in the service container so no way in the xml.

### DIFF
--- a/src/Sulu/Bundle/WebsiteBundle/DataCollector/SuluCollector.php
+++ b/src/Sulu/Bundle/WebsiteBundle/DataCollector/SuluCollector.php
@@ -11,6 +11,7 @@
 
 namespace Sulu\Bundle\WebsiteBundle\DataCollector;
 
+use Sulu\Component\Content\Compat\Structure\PageBridge;
 use Sulu\Component\Content\Compat\StructureInterface;
 use Sulu\Component\Webspace\Analyzer\Attributes\RequestAttributes;
 use Sulu\Component\Webspace\Portal;
@@ -22,8 +23,10 @@ use Symfony\Component\HttpKernel\DataCollector\DataCollector;
 class SuluCollector extends DataCollector
 {
     public function __construct(
-        private string $kernelEnvironment = 'dev'
+        private string $kernelEnvironment = 'dev',
+        private string $adminUrl = '/admin',
     ) {
+        $this->adminUrl = \rtrim('//', $this->adminUrl);
     }
 
     public function data($key)
@@ -110,6 +113,8 @@ class SuluCollector extends DataCollector
             }
         }
         $this->data['structure'] = $structure;
+
+        $this->data['adminUrl'] = $this->getAdminUrlForRequest($request);
     }
 
     /**
@@ -123,6 +128,23 @@ class SuluCollector extends DataCollector
         foreach ($localizations as &$localization) {
             $localization = (string) $localization['language'] . ($localization['default'] ? ' (default)' : '');
         }
+    }
+
+    protected function getAdminUrlForRequest(Request $request): ?string
+    {
+        $locale = $request->getLocale();
+
+        /** @var RequestAttributes $requestAttributes */
+        $requestAttributes = $request->attributes->get('_sulu');
+        /** @var Webspace $webspace */
+        $webspace = $requestAttributes->getAttribute('webspace');
+
+        $object = $request->attributes->get('structure');
+        if (!$object instanceof PageBridge) {
+            return null;
+        }
+
+        return $this->adminUrl . '/#/webspaces/' . $webspace->getKey() . '/pages/' . $locale . '/' . $object->getUuid() . '/details';
     }
 
     public function getName()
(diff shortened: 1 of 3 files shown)


src/Sulu/Bundle/WebsiteBundle/DataCollector/AdminUrlGenerator.php
```diff
@@ -0,0 +1,33 @@
...
+ */
+class AdminUrlGenerator implements AdminUrlGeneratorInterface
+{
+    public function getAdminUrlForRequest(Request $request): ?string
+    {
+        $locale = $request->getLocale();
+        /** @var string $webspaceKey */
+        $webspaceKey = $request->attributes->get('_sulu')->getAttribute('webspace')->getKey();
+        /** @var string $pageId */
+        $pageId = $request->attributes->get('structure')?->getUuid();
```
Prokyonn: we should also check if this is really a page, otherwise the url will not work

Prokyonn (commented):
Looks good! When you fix the phpstan issues and resolve the other comment, we can merge this to 2.6 :)

> Only register the service in dev. Currently the only way I see is either with attributes + autowiring or php in the service container so no way in the xml.

@mamazu You could also create a separate services.xml file for the dev service and only register it in the SuluWebsiteExtension when the application is run in dev mode

src/Sulu/Bundle/WebsiteBundle/DataCollector/AdminUrlGenerator.php
```diff
@@ -0,0 +1,33 @@
...
+
+        return '/admin/#/webspaces/' . $webspaceKey . '/pages/' . $locale . '/' . $pageId . '/details';
```
Prokyonn: Hmm, we probably also need a way to get the prefix for the routes, as this can be configured differently per project 🤔 
e.g.

```yaml
sulu_tag_api:
    resource: "@SuluTagBundle/Resources/config/routing_api.yml"
    type: rest
    prefix: /sulu/admin/api

sulu_admin:
    resource: "@SuluAdminBundle/Resources/config/routing.yml"
    prefix: /sulu/admin

sulu_admin_api:
    resource: "@SuluAdminBundle/Resources/config/routing_api.yml"
    type: rest
    prefix: /sulu/admin/api
```
from the config/routes/sulu_admin.yaml

mamazu:
@Prokyonn ready for review. The remaining test failure is not related to the pr.

mamazu:
We're not going to make this configurable and inline it.

mamazu:
This PR now also contains a BC break, the Sulu collector is now only available in the "dev" environment. While before it was available in test and prod as well.

src/Sulu/Bundle/WebsiteBundle/DataCollector/SuluCollector.php
```diff
@@ -22,7 +22,8 @@
 class SuluCollector extends DataCollector
 {
     public function __construct(
-        private string $kernelEnvironment = 'dev'
+        private string $kernelEnvironment = 'dev',
+        private ?AdminUrlGeneratorInterface $adminUrlGenerator = null,
```
alexander-schranz: we should just inline this logic here and not introduce a new service for the profiler. We plan a new minor release where we introduce maybe a similar service, so we should if we want merge this into 2.6 currently just inline the logic into this Collector and not introduce a new class, interface and service for this.

alexander-schranz (changes requested):
see comment

src/Sulu/Bundle/WebsiteBundle/DataCollector/AdminUrlGenerator.php
mamazu: Can be configured using `$this->adminUrl`

mamazu:
Rebased and inlined. However this service is now still registered in dev and in prod. We might want to split this up in the new php config

Prokyonn:
This PR is put on hold, as we will introduce a new interface (probably in Sulu 3.1) that simplifies the creation of deep links for sulu admin entities. Afterwards we can decide if we backport the interface or make this feature only available in Sulu >= 3.1

PR #8044, opened 2025-07-10
Make link targets configurable

martinlagler:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Create an ExternLinkProvider and add targets with translations to the LinkConfiguration.

#### Why?

So the translations of the link targets can be overridden and also can be changed.

#### Example Usage

Inside the LinkProvider
```php
return LinkConfigurationBuilder::create()
            ->setTargets(['_blank' => 'sulu_admin.link_blank'])
            ...
```

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Link/overlays/ExternalLinkTypeOverlay.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Link/overlays/ExternalLinkTypeOverlay.js
@@ -170,7 +170,21 @@ class ExternalLinkTypeOverlay extends React.Component<LinkTypeOverlayProps> {
             target,
             title,
             href,
+            options,
         } = this.props;
+
+        if (!options || !options.targets) {
+            throw new Error('The ExternalLinkTypeOverlay needs atleast targets as options in order to work!');
+        }
+
+        const {
+            targets = {
+                '_blank': 'sulu_admin.link_target_blank',
+                '_self': 'sulu_admin.link_target_self',
+                '_parent': 'sulu_admin.link_target_parent',
+                '_top': 'sulu_admin.link_target_top',
+            },
+        } = options;
 
         return (
             <Dialog
@@ -197,10 +211,11 @@ class ExternalLinkTypeOverlay extends React.Component<LinkTypeOverlayProps> {
                     {this.protocol && this.protocol !== 'mailto:' && onTargetChange
                         && <Form.Field label={translate('sulu_admin.link_target')} required={true}>
                             <SingleSelect onChange={onTargetChange} value={target}>
-                                <SingleSelect.Option value="_blank">_blank</SingleSelect.Option>
-                                <SingleSelect.Option value="_self">_self</SingleSelect.Option>
-                                <SingleSelect.Option value="_parent">_parent</SingleSelect.Option>
-                                <SingleSelect.Option value="_top">_top</SingleSelect.Option>
+                                {Object.keys(targets).map((targetValue) => (
+                                    <SingleSelect.Option key={targetValue} value={targetValue}>
+                                        {translate(targets[targetValue])}
+                                    </SingleSelect.Option>
+                                ))}
                             </SingleSelect>
                         </Form.Field>
                     }
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Link/overlays/ExternalLinkTypeOverlay.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Link/overlays/ExternalLinkTypeOverlay.js
@@ -173,18 +173,7 @@ class ExternalLinkTypeOverlay extends React.Component<LinkTypeOverlayProps> {
             options,
         } = this.props;
 
-        if (!options || !options.targets) {
-            throw new Error('The ExternalLinkTypeOverlay needs atleast targets as options in order to work!');
-        }
-
-        const {
-            targets = {
-                '_blank': 'sulu_admin.link_target_blank',
-                '_self': 'sulu_admin.link_target_self',
-                '_parent': 'sulu_admin.link_target_parent',
-                '_top': 'sulu_admin.link_target_top',
-            },
-        } = options;
+        const targets = options?.targets || ['_blank', '_self', '_parent', '_top'];
 
         return (
             <Dialog
@@ -211,9 +200,9 @@ class ExternalLinkTypeOverlay extends React.Component<LinkTypeOverlayProps> {
                     {this.protocol && this.protocol !== 'mailto:' && onTargetChange
                         && <Form.Field label={translate('sulu_admin.link_target')} required={true}>
                             <SingleSelect onChange={onTargetChange} value={target}>
-                                {Object.keys(targets).map((targetValue) => (
+                                {targets.map((targetValue) => (
                                     <SingleSelect.Option key={targetValue} value={targetValue}>
-                                        {translate(targets[targetValue])}
+                                        {translate(`sulu_admin.link${targetValue}`)}
                                     </SingleSelect.Option>
                                 ))}
                             </SingleSelect>
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Link/tests/overlays/ExternalLinkTypeOverlay.test.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Link/tests/overlays/ExternalLinkTypeOverlay.test.js
@@ -22,7 +22,6 @@ test('Render overlay with an undefined URL', () => {
                 {
                     displayProperties: [],
                     resourceKey: '',
-                    targets: ['_blank', '_self', '_parent', '_top'],
                 }
             }
             rel={undefined}
@@ -32,19 +31,6 @@ test('Render overlay with an undefined URL', () => {
     );
 
     expect(externalLinkOverlay.find('Form').render()).toMatchSnapshot();
-});
-
-test('Render overlay without options', () => {
-    expect(() => shallow(
-        <ExternalLinkTypeOverlay
-            href={undefined}
-            onCancel={jest.fn()}
-            onConfirm={jest.fn()}
-            onHrefChange={jest.fn()}
-            open={true}
-            options={undefined}
-        />
-    )).toThrow('The ExternalLinkTypeOverlay needs atleast targets as options in order to work!');
 });
 
 test('Render overlay with mailto URL', () => {
@@ -62,12 +48,6 @@ test('Render overlay with mailto URL', () => {
                 {
                     displayProperties: [],
                     resourceKey: '',
-                    targets: {
-                        '_blank': 'sulu_admin.link_blank',
-                        '_self': 'sulu_admin.link_self',
-                        '_parent': 'sulu_admin.link_parent',
-                        '_top': 'sulu_admin.link_top',
-                    },
                 }
             }
             rel={undefined}
@@ -94,12 +74,6 @@ test('Render overlay with a URL', () => {
                 {
                     displayProperties: [],
                     resourceKey: '',
-                    targets: {
-                        '_blank': 'sulu_admin.link_blank',
-                        '_self': 'sulu_admin.link_self',
-                        '_parent': 'sulu_admin.link_parent',
-                        '_top': 'sulu_admin.link_top',
-                    },
                 }
             }
             rel={undefined}
@@ -129,12 +103,6 @@ test('Pass correct props to Dialog', () => {
                 {
                     displayProperties: [],
                     resourceKey: '',
-                    targets: {
-                        '_blank': 'sulu_admin.link_blank',
-                        '_self': 'sulu_admin.link_self',
-                        '_parent': 'sulu_admin.link_parent',
-                        '_top': 'sulu_admin.link_top',
-                    },
                 }
             }
             rel={undefined}
@@ -165,12 +133,6 @@ test('Display given URL with query parameters in href input', () => {
                 {
                     displayProperties: [],
                     resourceKey: '',
-                    targets: {
-                        '_blank': 'sulu_admin.link_blank',
-                        '_self': 'sulu_admin.link_self',
-                        '_parent': 'sulu_admin.link_parent',
-                        '_top': 'sulu_admin.link_top',
-                    },
                 }
             }
             target="_blank"
@@ -199,12 +161,6 @@ test('Do not call onHrefChange handler if input did not loose focus', () => {
                 {
                     displayProperties: [],
                     resourceKey: '',
-                    targets: {
-                        '_blank': 'sulu_admin.link_blank',
-                        '_self': 'sulu_admin.link_self',
-                        '_parent': 'sulu_admin.link_parent',
-                        '_top': 'sulu_admin.link_top',
-                    },
                 }
             }
             rel={undefined}
@@ -235,12 +191,6 @@ test('Fields should change immediately after protocol was changed', () => {
                 {
                     displayProperties: [],
                     resourceKey: '',
-                    targets: {
-                        '_blank': 'sulu_admin.link_blank',
-                        '_self': 'sulu_admin.link_self',
-                        '_parent': 'sulu_admin.link_parent',
-                        '_top': 'sulu_admin.link_top',
-                    },
                 }
             }
             rel={undefined}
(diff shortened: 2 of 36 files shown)


src/Sulu/Bundle/MarkupBundle/Markup/Link/LinkConfigurationBuilder.php
```diff
@@ -48,6 +48,11 @@ class LinkConfigurationBuilder
      */
     private $icon;
 
+    /**
+     * @var ?array
```
Prokyonn: add proper type e.g. `array<string, string>`

src/Sulu/Bundle/MarkupBundle/Markup/Link/LinkProviderPool.php
```diff
@@ -47,7 +47,10 @@ public function getConfiguration()
     {
         $configuration = [];
         foreach ($this->providers as $name => $provider) {
-            $configuration[$name] = $provider->getConfiguration();
+            /** @var ?LinkConfiguration|LinkConfigurationBuilder $providerConfiguration */
+            $providerConfiguration = $provider->getConfiguration();
+
+            $configuration[$name] = $providerConfiguration instanceof LinkConfigurationBuilder ? $providerConfiguration->getLinkConfiguration() : $providerConfiguration;
```
Prokyonn, line 61: Add a deprecation warning when the configuration is given instead of the Builder

src/Sulu/Bundle/MediaBundle/Media/ImageConverter/ImagineImageConverter.php
```diff
@@ -93,10 +93,20 @@ public function getSupportedOutputImageFormats(?string $mimeType): array
 
     public function convert(FileVersion $fileVersion, $formatKey, $imageFormat)
     {
-        $imageResource = $this->mediaImageExtractor->extract(
-            $this->storage->load($fileVersion->getStorageOptions()),
-            $fileVersion->getMimeType()
-        );
+        try {
```
Prokyonn: This should not be in this PR :D

src/Sulu/Bundle/PageBundle/Markup/Link/ExternalLinkProvider.php
```diff
@@ -0,0 +1,41 @@
...
+    public function __construct(
+        private TranslatorInterface $translator,
+    ) {
+    }
+
+    public function getConfiguration()
+    {
+        return LinkConfigurationBuilder::create()
+            ->setTitle($this->translator->trans('sulu_admin.external_link', [], 'admin'))
+            ->setResourceKey('')
```
Prokyonn: I think you don't even have to call these setters 🤔

Prokyonn (changes requested):
- Upgrade.md needs to be updated for the deprecation
- in the Link contenttype the if ($provider === 'external') can be removed as it should now work over the ExternalLinkProvider

src/Sulu/Bundle/PageBundle/Markup/Link/ExternalLinkProvider.php
martinlagler: Sadly this is required because otherwise these values would be null, but they are typed to string.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Link/overlays/ExternalLinkTypeOverlay.js
```diff
@@ -170,8 +170,17 @@ class ExternalLinkTypeOverlay extends React.Component<LinkTypeOverlayProps> {
             target,
             title,
             href,
+            options,
         } = this.props;
 
+        if (!options || !options.targets) {
+            throw new Error('The ExternalLinkTypeOverlay needs atleast targets as options in order to work!');
```
alexander-schranz, line 177: This is a BC break. If there are no targets given we should render all targets currently.

In 3.0 we can decide if no targets are given we should maybe just not render the target field here.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Link/overlays/ExternalLinkTypeOverlay.js
```diff
@@ -170,8 +170,17 @@ class ExternalLinkTypeOverlay extends React.Component<LinkTypeOverlayProps> {
...
+        }
+
+        const {
+            targets,
+        } = options;
```
alexander-schranz: This is a bc break we should keep the component as it was and prefill the values:

```suggestion
        const {
            targets = {
                '_blank': 'sulu_admin.link_target_blank',
                '_self': 'sulu_admin.link_target_self',
                '_parent': 'sulu_admin.link_target_parent',
                '_top': 'sulu_admin.link_target_top',
            },
        } = options;
```

src/Sulu/Bundle/PageBundle/Resources/translations/admin.de.json
```diff
@@ -89,5 +89,9 @@
     "sulu_activity.description.pages.preview_link_generated": "{userFullName} hat den öffentlichen Vorschaulink \"{resourceTitle}\" erstellt",
     "sulu_activity.description.pages.preview_link_revoked": "{userFullName} hat den öffentlichen Vorschaulink \"{resourceTitle}\" widerrufen",
     "sulu_reference.resource.pages": "Seite",
-    "sulu_reference.resource.snippets": "Schnipsel"
+    "sulu_reference.resource.snippets": "Schnipsel",
+    "sulu_admin.link_blank":  "_blank",
+    "sulu_admin.link_self": "_self",
+    "sulu_admin.link_parent": "_parent",
+    "sulu_admin.link_top": "_top"
```
alexander-schranz: would prefix them with target:


```suggestion
    "sulu_admin.link_target_blank":  "_blank",
    "sulu_admin.link_target_self": "_self",
    "sulu_admin.link_target_parent": "_parent",
    "sulu_admin.link_target_top": "_top"
```

src/Sulu/Bundle/PageBundle/Resources/translations/admin.en.json
```diff
@@ -89,5 +89,9 @@
     "sulu_activity.description.pages.preview_link_generated": "{userFullName} has generated a public preview link \"{resourceTitle}\"",
     "sulu_activity.description.pages.preview_link_revoked": "{userFullName} has revoked the public preview link \"{resourceTitle}\"",
     "sulu_reference.resource.pages": "Page",
-    "sulu_reference.resource.snippets": "Snippet"
+    "sulu_reference.resource.snippets": "Snippet",
+    "sulu_admin.link_blank":  "_blank",
+    "sulu_admin.link_self": "_self",
+    "sulu_admin.link_parent": "_parent",
+    "sulu_admin.link_top": "_top"
```
alexander-schranz: Would prefix them with target:

```suggestion
    "sulu_admin.link_target_blank":  "_blank",
    "sulu_admin.link_target_self": "_self",
    "sulu_admin.link_target_parent": "_parent",
    "sulu_admin.link_target_top": "_top"
```

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Link/tests/overlays/ExternalLinkTypeOverlay.test.js
```diff
@@ -18,6 +18,18 @@ test('Render overlay with an undefined URL', () => {
...
+            options={
+                {
+                    displayProperties: [],
+                    resourceKey: '',
+                    targets: {
+                        '_blank': 'sulu_admin.link_blank',
+                        '_self': 'sulu_admin.link_self',
+                        '_parent': 'sulu_admin.link_parent',
+                        '_top': 'sulu_admin.link_top',
+                    },
```
alexander-schranz: give the translation key into it feels like of strange for me. I don't think a developer should need to care about the translation at this point so maybe a list of targets would be better:


```suggestion
                    targets: ['_blank', '_self', '_parent', '_top'],
```

src/Sulu/Bundle/MarkupBundle/Markup/Link/LinkConfiguration.php
```diff
@@ -57,14 +57,29 @@ class LinkConfiguration
...
+    /**
+     * @var array<string, string>
+     */
+    #[Groups(['frontend'])]
+    private $targets = [
+        '_blank' => 'sulu_admin.link_blank',
+        '_self' => 'sulu_admin.link_self',
+        '_parent' => 'sulu_admin.link_parent',
+        '_top' => 'sulu_admin.link_top',
+    ];
```
alexander-schranz: This kind of feels wrong for me using translation keys inside this class.

```suggestion
    /**
     * @var array<string, string>
     */
    #[Groups(['frontend'])]
    private $targets = [
        '_blank',
        '_self',
        '_parent',
        '_top',
    ];
```

src/Sulu/Bundle/MarkupBundle/Markup/Link/LinkProviderPool.php
```diff
@@ -47,7 +47,14 @@ public function getConfiguration()
     {
         $configuration = [];
         foreach ($this->providers as $name => $provider) {
-            $configuration[$name] = $provider->getConfiguration();
+            /** @var LinkConfiguration|LinkConfigurationBuilder|null $providerConfiguration */
+            $providerConfiguration = $provider->getConfiguration();
+
+            if ($providerConfiguration instanceof LinkConfigurationBuilder) {
+                @trigger_deprecation('sulu/sulu', '2.6', 'The LinkProvider should return a LinkConfiguration and not a LinkConfigurationBuilder. The LinkConfigurationBuilder will not be supported in 3.0.');
```
alexander-schranz: not understanding this deprecation the LinkProvider are already typed to return always a LinkConfiguration here: https://github.com/sulu/sulu/blob/4a08016d7f1a58e612dd2e88c57e1cd13a1f77d3/src/Sulu/Bundle/MarkupBundle/Markup/Link/LinkProviderInterface.php#L24

you did add it to return also LinkConfigurationBuilder but directly deprecated it which feels wrong?

src/Sulu/Bundle/PageBundle/Markup/Link/ExternalLinkProvider.php
```diff
@@ -0,0 +1,55 @@
...
+
+namespace Sulu\Bundle\PageBundle\Markup\Link;
+
+use Ramsey\Uuid\Uuid;
+use Sulu\Bundle\MarkupBundle\Markup\Link\LinkConfigurationBuilder;
+use Sulu\Bundle\MarkupBundle\Markup\Link\LinkItem;
+use Sulu\Bundle\MarkupBundle\Markup\Link\LinkProviderInterface;
+use Symfony\Contracts\Translation\TranslatorInterface;
+
+class ExternalLinkProvider implements LinkProviderInterface
```
alexander-schranz: The builder looks kind of strange here nothing which we can give BC Support at current state so mark it as internal.

```suggestion
/**
 * @internal
 */
class ExternalLinkProvider implements LinkProviderInterface
```

src/Sulu/Bundle/PageBundle/Markup/Link/ExternalLinkProvider.php
```diff
@@ -0,0 +1,55 @@
...
+
+    public function preload(array $hrefs, $locale, $published = true)
+    {
+        if (0 === \count($hrefs)) {
+            return [];
+        }
+
+        $result = [];
+        foreach ($hrefs as $href) {
+            $result[] = new LinkItem(Uuid::uuid4()->toString(), $href, $href, true);
```
alexander-schranz: generate a uuid feels strange here, would just use the href as id.

Also not sure about the `title` that feels wrong from my point of view:


```suggestion
            $result[] = new LinkItem($href, '', $href, true);
```

src/Sulu/Bundle/PageBundle/Tests/Unit/Markup/Link/ExternalLinkProviderTest.php
```diff
@@ -0,0 +1,95 @@
...
+                [],
+                '',
+                '',
+                '',
+                [
+                    '_blank' => 'sulu_admin.link_blank',
+                    '_self' => 'sulu_admin.link_self',
+                    '_parent' => 'sulu_admin.link_parent',
+                    '_top' => 'sulu_admin.link_top',
+                ],
```
alexander-schranz: feels wrong to provide here translation keys:


```suggestion
                [
                    '_blank',
                    '_self',
                    '_parent',
                    '_top',
                ],
```

src/Sulu/Component/Content/Types/Link.php
```diff
@@ -94,10 +94,6 @@ public function getContentData(PropertyInterface $property): ?string
             return null;
         }
 
-        if (self::LINK_TYPE_EXTERNAL === $value['provider']) {
```
alexander-schranz, line 97: the constants seems not longer be used by us? So we could add `@deprecated`also here.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Link/types.js
```diff
@@ -28,6 +28,7 @@ export type LinkTypeOptions = {|
     listAdapter?: string,
     overlayTitle?: string,
     resourceKey: string,
+    targets?: { [key: string]: string },
```
alexander-schranz: would more go with `string[]`

alexander-schranz (changes requested):
Not sure if we really should make the translationto be part of the `builder` and `configuration`.

Currently a developer would in this case in the configuration and the using the js components also have to think about the translation keys.

So question is if we just go with `targets` `['_blank', '_self', '_parent', '_top']` and go with `'sulu_admin.link_target.' + key or something like that.

alexander-schranz:
I also think [MediaLinkTypeOverlay.js](https://github.com/sulu/sulu/blob/4a08016d7f1a58e612dd2e88c57e1cd13a1f77d3/src/Sulu/Bundle/MediaBundle/Resources/js/containers/Link/overlays/MediaLinkTypeOverlay.js#L66) is not targeted in this PR: 

https://github.com/search?q=repo%3Asulu%2Fsulu%20_blank&type=code

src/Sulu/Bundle/MarkupBundle/Markup/Link/LinkProviderPool.php
martinlagler: Oh I think i mixed this up, in the future it should return a LinkConfigurationBuilder

alexander-schranz: entry in UPGRADE.md missing about this change

alexander-schranz: also all existing link configuration providers need be changed to return the builder if not already done to avoid deprecation be thrown for core services.

/cc @Prokyonn changing return type is still kind of bc break, but we can poker here that nobody really did got into contact with this services.

Prokyonn: Yes it is a BC break from the phpdoc type perspective, but at the same time we added a BC layer to prevent anything from "breaking" and with 3.0 I would immediatley remove all of this to have it in a clean and consistent state.

This MR had some unforeseen side effects, I didn't anticipate that we more or less completely hard coded the external and media links 🙈 but it is also a good chance to get it consistent 😄

alexander-schranz: BC layer is for our service but if somebody did call the getConfiguration service and expected a LinkConfiguration but now gets a LinkConfigurationBuilder there service will fail. But as said we can poker here the nobody else did got into contact with this services yet and only created there own link providers which we still support.

src/Sulu/Bundle/PageBundle/Markup/Link/ExternalLinkProvider.php
```diff
@@ -0,0 +1,58 @@
...
+/*
+ * This file is part of Sulu.
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\PageBundle\Markup\Link;
```
alexander-schranz: ~~External links should also work when only have the admin bundle. So We should move this to the admin bundle or atleast the markup bundle.~~

src/Sulu/Bundle/PageBundle/Markup/Link/ExternalLinkProvider.php
Prokyonn: move to MarkupBundle

src/Sulu/Bundle/PageBundle/Markup/Link/ExternalLinkProvider.php
alexander-schranz: As discussed I would move to markup bundle as admin bundle should have no requirement to markup interfaces.

alexander-schranz (approved):

PR #8045, opened 2025-07-10
Fix category tree search

MarkusHolstein:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no

#### What's in this PR?

When searching for a category string, nearly always the expandedIds parameter is set in the query string to the controller.
As in a full tree view in the category administration, there is no $parentId in the request parameters, thus $categoriesByParentId[$parentId] delivers null, which leads to iterator_to_array(): Argument #1 ($iterator) must be of type Traversable|array, null given at src/Sulu/Component/Rest/ListBuilder/CollectionRepresentation.php:35, though categories were found, but not with parent "null"
This fixes this behaviour. and allows to search a full tree.

#### Why?

The server responds with a 500 Exception

#### Example Usage

Call url https://sulu.rocks/admin/api/categories?page=1&locale=en&expandedIds=3&limit=10&fields=name,id&search=Test3&flat=true


MarkusHolstein:
Hi @alexander-schranz,

I don't think, this is a feature ;-)
This fixes a bug in the existing and "live" code.

alexander-schranz:
Thx for the hint, does this also appear on 2.5?

MarkusHolstein:
I don't have a current installation of 2.5 right now. As the code of the controller is similar between both versions, I assume, this bug might also exist in 2.5 @alexander-schranz

alexander-schranz:
@MarkusHolstein can you change the target branch and rebase to 2.5:

```bash
# add sulu as remote if not already done
git fetch sulu

# rebase it from 2.6 to 2.5
git rebase -i sulu/2.6 --onto sulu/2.5

# push
git push <your-remote> <your-branch> --force
```

As it seems you did use directly a `2.6` branch instead of a new branch on your fork. It maybe better to create a new branch and PR ontop of 2.5. Recommend to create per Pull Request an own new branch:


```bash
git checkout -b bugfix/category-tree-search

# add sulu as remote if not already done
git fetch sulu

# rebase it from 2.6 to 2.5
git rebase -i sulu/2.6 --onto sulu/2.5

# push
git push <your-remote> bugfix/category-tree-search

# create new PR 
```

MarkusHolstein:
Hi @alexander-schranz,
as suggessted, I moved the PR to a branch based on 2.5.

Issue #8022, opened 2025-06-19
Broken moving and copy page


devdepart:
| Q | A
| --- | ---
| Sulu Version | "3.0.0-alpha3"
| PHP Version | 8.3.22
| DB Version |  PostgreSQL 17
| Browser Version |  Firefox 128.11.0esr

# Actual Behavior

When I try to move an existing page or create a copy of an existing page, the UI stops responding until the web page is updated. Once the web page is refreshed, there is no result in the form of moving or creating a copy of the landing page.

![Image](https://github.com/user-attachments/assets/2f5c9ecb-f63f-42d2-b73b-4a8b463c2ba6)
![Image](https://github.com/user-attachments/assets/2b0a325d-0529-46b7-b168-f4b1463c1de2)

# Expected Behavior

Copying (moving) an existing page

# Steps to Reproduce

I create a page
I highlight the page and select the menu item "Copy"
I select the target catalog
I click on the button

# Possible Solutions

alexander-schranz:
Thx for the issue, please use the issue template to provide all informations. We will look into it and update you as soon as tackled.

alexander-schranz:
This is implemented into the latest `3.0.*@dev` version. Thx for report.

PR #8046, opened 2025-07-11
Add page move/copy messages

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes https://github.com/sulu/sulu/issues/8022
| License | MIT

#### What's in this PR?

Adds copy page and move page in the tree view.

#### Why?

It was missing in Sulu 3.0


packages/page/src/Infrastructure/Doctrine/Repository/PageRepository.php
```diff
@@ -222,10 +221,18 @@ public function reorderOneBy(array $filters, int $position): void
...
             $this->entityRepository->moveDown($page, \abs($movementSteps));
         }
+    }
+
+    public function moveOneBy(array $sourceFilters, array $destinationFilters): void
+    {
+        $sourcePage = $this->getOneBy($sourceFilters);
+        $destinationPage = $this->getOneBy($destinationFilters);
 
-        if (true !== $this->entityRepository->verify()) {
```
Prokyonn, line 226: I removed this from the `reorderOneBy` method, because the tree is always invalid before flushing, so we always did call recover in a false positive way

packages/page/src/Application/MessageHandler/CopyPageMessageHandler.php
```diff
@@ -0,0 +1,71 @@
...
+    public function __construct(
+        private PageRepositoryInterface $pageRepository,
+        private ContentCopierInterface $contentCopier,
+        private LocalizationManagerInterface $localizationManager,
+    ) {
+    }
+
+    public function __invoke(CopyPageMessage $message): void
+    {
+        $page = $this->pageRepository->getOneBy($message->getIdentifier());
```
alexander-schranz: Would use speaking variables:

```suggestion
        $sourcePage = $this->pageRepository->getOneBy($message->getIdentifier());
```

packages/page/src/Application/Message/CopyPageMessage.php
```diff
@@ -0,0 +1,47 @@
...
+{
+    /**
+     * @param array{
+     *     uuid?: string
+     * } $identifier
+     * @param array{
+     *     uuid?: string
+     * } $destinationIdentifier
+     */
+    public function __construct(private array $identifier, private array $destinationIdentifier)
```
alexander-schranz: Not sure here kind of miss the `parent` word here.

```suggestion
    public function __construct(private array $identifier, private array $destinationParentIdentifier)
```

packages/page/src/Application/Message/CopyPageMessage.php
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+namespace Sulu\Page\Application\Message;
+
+class CopyPageMessage
+{
+    /**
+     * @param array{
+     *     uuid?: string
+     * } $identifier
```
alexander-schranz: not sure if we should go with:


```suggestion
     * @param array{
     *     uuid?: string
     * } $sourceIdentifier
```

packages/page/src/Application/Message/MovePageMessage.php
```diff
@@ -0,0 +1,49 @@
...
+     * @param array{
+     *     uuid?: string
+     * } $identifier
+     * @param array{
+     *     uuid?: string
+     * } $destinationIdentifier
+     */
+    public function __construct(
+        private array $identifier,
+        private array $destinationIdentifier,
```
alexander-schranz:
```suggestion
        private array $destinationParentIdentifier,
```

or:

alexander-schranz: Not sure about destination here, kind of missing `parent` in this variable names.

```suggestion
        private array $parentIdentifier,
```

packages/page/src/Application/MessageHandler/CopyPageMessageHandler.php
```diff
@@ -0,0 +1,71 @@
...
+        $targetParentPage = $this->pageRepository->getOneBy($message->getDestinationIdentifier());
+
+        $targetPage = $this->pageRepository->createNew();
+        $targetPage->setWebspaceKey($targetParentPage->getWebspaceKey());
+        $targetPage->setParent($targetParentPage);
+        $this->pageRepository->add($targetPage);
+
+        foreach ($this->localizationManager->getLocalizations() as $localization) {
```
alexander-schranz, line 47: Personally not like that we depend here on localizationManager instead of `getAvailableLocales` method of the dimension itself.

Maybe add a todo here currently.

packages/page/src/Application/MessageHandler/CopyPageMessageHandler.php
```diff
@@ -0,0 +1,71 @@
...
+final class CopyPageMessageHandler
+{
+    public function __construct(
+        private PageRepositoryInterface $pageRepository,
+        private ContentCopierInterface $contentCopier,
+        private LocalizationManagerInterface $localizationManager,
+    ) {
+    }
+
+    public function __invoke(CopyPageMessage $message): void
```
alexander-schranz: we should return the new `page` for easier unit testing.

packages/page/src/UserInterface/Controller/Admin/PageController.php
```diff
@@ -217,6 +219,20 @@ private function handleAction(Request $request, string $uuid): ?PageInterface //
...
+            /** @see Sulu\Page\Application\MessageHandler\MovePageMessageHandler */
+            /** @var null */
+            return $this->handle(new Envelope($message, [new EnableFlushStamp()]));
+        } elseif ('copy' == $action) {
+            $destinationUuid = $request->query->getString('destination');
+            $message = new CopyPageMessage(['uuid' => $uuid], ['uuid' => $destinationUuid]);
+
+            /** @see Sulu\Page\Application\MessageHandler\CopyPageMessageHandler */
+            /** @var null */
+            return $this->handle(new Envelope($message, [new EnableFlushStamp()]));
```
alexander-schranz: return null sounds strange here. Copy an article ends on the copied version: https://sulu.rocks/admin/#/articles/en/article/95c9cbc7-cb46-4ad2-9b4f-ceb7ecb4ae76/details
So think we need response something here.

packages/page/src/Application/Message/CopyPageMessage.php
alexander-schranz:
```suggestion
    public function __construct(
        private array $identifier,
        private array $destinationParentIdentifier,
        private ?string $destinationUuid = null,
    ) {
```

It should be possible to set the uuid from outside like we allow it in create message also: https://github.com/sulu/sulu/blob/3.0/packages/page/src/Application/Message/CreatePageMessage.php

alexander-schranz (changes requested):

packages/page/src/Application/MessageHandler/CopyPageMessageHandler.php
```diff
@@ -32,25 +33,26 @@ public function __construct(
...
 
-    public function __invoke(CopyPageMessage $message): void
+    public function __invoke(CopyPageMessage $message): PageInterface
     {
-        $page = $this->pageRepository->getOneBy($message->getIdentifier());
-        $targetParentPage = $this->pageRepository->getOneBy($message->getDestinationIdentifier());
+        $sourcePage = $this->pageRepository->getOneBy($message->getSourceIdentifier());
+        $targetParentPage = $this->pageRepository->getOneBy($message->getTargetParentIdentifier());
 
         $targetPage = $this->pageRepository->createNew();
```
alexander-schranz: @Prokyonn is not here the uuid missing:


```suggestion
        $targetPage = $this->pageRepository->createNew($message->getTargetUuid());
```

alexander-schranz (approved):

PR #8047, opened 2025-07-14
Deprecate the `findAllTags` function in `TagManager`

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | yes
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?
* Deprecating the `findAllTags` in favour of `findAll` default repository tag.
* Deprecate `findAll` on the TagManager

#### Why?
* Less methods to maintain and phpstan if taken seriously will tell you what kind of object the `findAll` returns.
* Don't proxy method calls through a big objects, it's not good for performance.

#### Context
Maybe having a twig method that returns all tags sounds like a bad performance hole waiting to happen. (People will most likely filter them in twig then, which is pretty inefficient.)

### DIFF
--- a/src/Sulu/Bundle/TagBundle/Tag/TagManager.php
+++ b/src/Sulu/Bundle/TagBundle/Tag/TagManager.php
@@ -43,6 +43,8 @@ class TagManager implements TagManagerInterface
     /**
      * Loads all the tags managed in this system.
      *
+     * @deprecated use the Repository findAll method
+     *
      * @return TagInterface[]
      */
     public function findAll()
--- a/UPGRADE-2.x.md
+++ b/UPGRADE-2.x.md
@@ -2,6 +2,11 @@
 
 ## 2.6.10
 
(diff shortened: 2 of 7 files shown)


src/Sulu/Bundle/TagBundle/Tag/TagManager.php
```diff
@@ -43,6 +43,8 @@ public function __construct(
     /**
      * Loads all the tags managed in this system.
      *
+     * @deprecated use the Repository findAll method
+     *
```
alexander-schranz, line 47: should be on the interface also

mamazu: Done

alexander-schranz:
@mamazu thank you!

PR #8055, opened 2025-07-21
Deprecate the sulu.context kernel parameter

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | yes
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Deprecate the `sulu.context` kernel parameter.

#### Why?

Why we did try to remove in the past more and more sulu.context based behaviour and it was a inofficial a long communicated that its service should not depend on it. Its time for official deprecate the parameter now. So we are save at time of 3.0 to remove it. Currently marking it at internal as we not yet completely removed it from internal usage, but this way we are free to remove it on 3.0 at any time.

### DIFF
--- a/UPGRADE-2.x.md
+++ b/UPGRADE-2.x.md
@@ -4,9 +4,9 @@
 
 ### Deprecate usage of `sulu.context` parameter
 
-The `sulu.context` parameter is deprecated and services should not longer depend on it.
+The `sulu.context` parameter is deprecated and services should no longer depend on it.
 
-If some services behave differently between an `admin` or a `website` request you should depend on
+If some services behave differently between an `admin` or a `website` request, you should depend on
 [Symfony Firewall Config](https://symfony.com/doc/7.4/security.html#fetching-the-firewall-configuration-for-a-request)
 instead.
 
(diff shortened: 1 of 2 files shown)


UPGRADE-2.x.md
```diff
@@ -1,5 +1,22 @@
 # Upgrade
 
+## 2.6.11
+
+### Deprecate usage of `sulu.context` parameter
+
+The `sulu.context` parameter is deprecated and services should not longer depend on it.
```
Prokyonn:
```suggestion
The `sulu.context` parameter is deprecated and services should no longer depend on it.
```

UPGRADE-2.x.md
```diff
@@ -1,5 +1,22 @@
...
+
+If some services behave differently between an `admin` or a `website` request you should depend on
```
Prokyonn:
```suggestion
If some services behave differently between an `admin` or a `website` request, you should depend on
```

UPGRADE-2.x.md
```diff
@@ -1,5 +1,22 @@
...
+[Symfony Firewall Config](https://symfony.com/doc/7.4/security.html#fetching-the-firewall-configuration-for-a-request)
+instead.
+
+```php
+$suluContext = $this->security?->getFirewallConfig($request)?->getName() === 'admin' ? 'admin' : 'website';
+```
+
+Still best practices is that your services do not depend on such states, instead of controllers or commands
+should be based on the request or command arguments give other parameters to your service to control its behavior.
```
Prokyonn:
```suggestion
Still, best practices is that your services do not depend on such states. Instead controllers or commands
should be based on the request or command arguments, giving other parameters to your service to control its behavior.
```

Prokyonn (approved):

PR #8054, opened 2025-07-18
Add Versioning to the ContentBundle

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Related issues/PRs | #7672 
| License | MIT
| Documentation PR | sulu/sulu-docs# 

#### What's in this PR?

Adds Versioning to content-bundle entities
- [x] pages
- [x] articles
- [x] snippets

#### To Do

- [ ] Create a documentation PR
- [ ] Add breaking changes to UPGRADE.md

PR #8060, opened 2025-07-22
Add timestamp support to DateTimeFieldTransformer

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| License | MIT

#### What's in this PR?

Adds support for timestamps in the DateTimeTransformer.

#### Why?

Timestamps do not need to be converted in the backend and we need it for the new versioning :)


alexander-schranz (approved):

alexander-schranz:
We discussed that we backport this to 2.6 as it will be the LTS version. We using it only in the new content storage on 3.0: https://github.com/sulu/sulu/pull/8054/files

PR #8093, opened 2025-08-18
Add ExampleSmartContentProvider and SmartContent tests

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

- Adds a SmartContentProvider implementation to the ExampleBundle
- Adds a bunch of tests for the SmartContent resolver.

#### Why?

The ExampleBundle should have implemented all the features to use them in the tests.


alexander-schranz (approved):

alexander-schranz:
Looks like there is an issue that the order by is not part of the distinct select?

PR #8094, opened 2025-08-18
Add PropertyResolverMetadataAwareInterface

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Related issues/PRs | #7672 
| License | MIT

#### What's in this PR?

Adds the PropertyResolverMetadataAwareInterface.

#### Why?

The BlockPropertyResolver and ImageMapPropertyResolver need the metadata to dynamically resolve the properties. PropertyResolver that do not need the Metadata should not have them in the "params" array, so that they will not be forwarded to the FE.

### DIFF
--- a/packages/content/src/Application/PropertyResolver/PropertyResolverProvider.php
+++ b/packages/content/src/Application/PropertyResolver/PropertyResolverProvider.php
@@ -14,7 +14,6 @@ declare(strict_types=1);
 namespace Sulu\Content\Application\PropertyResolver;
 
 use Sulu\Content\Application\PropertyResolver\Resolver\PropertyResolverInterface;
-use Sulu\Content\Application\PropertyResolver\Resolver\PropertyResolverMetadataAwareInterface;
 
 /**
  * @internal The constructor of this class may change in future releases to add new features or improve performance.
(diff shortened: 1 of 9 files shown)


packages/content/src/Application/PropertyResolver/PropertyResolverProvider.php
```diff
@@ -26,19 +27,19 @@
 final class PropertyResolverProvider implements PropertyResolverProviderInterface
 {
     /**
-     * @var PropertyResolverInterface[]
+     * @var array<PropertyResolverInterface|PropertyResolverMetadataAwareInterface>
```
alexander-schranz: Not happy about this. Let me think about it how we can avoid here Union Type.

Prokyonn: @alexander-schranz https://github.com/sulu/sulu/pull/8094/commits/0ea5d1704f65f78cab04ae764f5586bf47dd73be WDYT about this solution?

alexander-schranz: Like it 👍

PR #7995, opened 2025-06-06
Say goodbye to the old phpcr, document manager, page bundle services, cleanup 🫡

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? | no 
| BC breaks? | MANY say goodbye to phpcr, document manager they served us well 🫡
| Deprecations? | yes 
| Fixed tickets | fixes # 
| Related issues/PRs | https://github.com/sulu/skeleton/pull/295, up https://github.com/sulu/sulu/pull/8137, #7672
| License | MIT
| Documentation PR | sulu/sulu-docs# 

#### What's in this PR?

Remove step by step old services and find usages of them in unrelated cases.

#### Why?

Replacing PHPCR and so the document manager with the new content storage is target of 3.0. By already replace the old structure metadata usage in most places we can now step by step go to replace it.

#### ToDo

Following tests fail without PHPCR related services:

- [x] article
- [x] content
- [x] page
- [x] route
- [x] snippet
- [x] ActivityBundle
- [x] AdminBundle
- [x] AudienceTargetingBundle
- [x] CategoryBundle
- [x] ContactBundle
- [x] CoreBundle
- [x] HttpCacheBundle
- [x] LocationBundle
- [x] MarkupBundle
- [x] MediaBundle
- [x] PersistenceBundle
- [x] PreviewBundle
- [x] ReferenceBundle
- [x] ~~RouteBundle (will be removed)~~
- [x] SecurityBundle
- [x] TagBundle
- [x] TrashBundle
- [x] WebsiteBundle
- [x] Components

Other todos:

- [ ] Remove PHPCR dependencies from `composer.json`
- [ ] Remove unused component classes

PR #8099, opened 2025-08-18
Remove old Dynamic Route Services from WebsiteBundle

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | yes 
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | https://github.com/sulu/sulu/pull/7995
| License | MIT

#### What's in this PR?

Remove old Dynamic Route Services from WebsiteBundle.

#### Why?

The routing was rewritten from scratch and lives now directly inside the new `packages/route` bundle.


alexander-schranz:
think we need tackle this in the overall PR as it can not be removed without removing the whole old route bundle. closing so in favor of https://github.com/sulu/sulu/pull/7995

PR #8105, opened 2025-08-19
Move Sulu Version CompilerPass to AdminBundle to avoid depending on WebsiteBundle parameters

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | yes
| Deprecations? | no  
| Fixed tickets | fixes # 
| Related issues/PRs | https://github.com/sulu/sulu/pull/7995
| License | MIT

#### What's in this PR?

Move Sulu Version CompilerPass to Admin.

#### Why?

Avoid AdminBundle depend on WebsiteBundle parameter.

Previously the Admin Bundle was not registered under the WebsiteBundle context. Since 2.6 AdminBundle is required to be in both admin and website context and so this CompilerPass can now be moved to the Admin Bundle.


Prokyonn (approved):

src/Sulu/Bundle/AdminBundle/DependencyInjection/Compiler/SuluVersionPass.php
```diff
@@ -9,72 +9,81 @@
...
      * Read composer.lock file and return version of sulu.
-     *
-     * @param string $dir
-     *
-     * @return string
      */
-    private function getSuluVersion($dir)
+    private function getSuluVersion(string $dir): string
     {
-        $version = '_._._';
```
mamazu, line 40: Why don't we use the composer tools for that? `\Composer\InstalledVersions::getVersion('sulu/sulu')` or `\Composer\InstalledVersions::getPrettyVersion('sulu/sulu')` should already take care of that.

PR #8106, opened 2025-08-19
Use request context service instead of container params if no request is available

### LINKED ISSUE #7280: Webspace manager uses request context params instead of request context service
| Q | A
| --- | ---
| Sulu Version | 2.5
| PHP Version | doesn't matter
| DB Version | doesn't matter
| Browser Version | doesn't matter

# Actual Behavior

The webspace manager, which is configured in the core bundle, relies on the request context container parameters `router.request_context.host` and `router.request_context.scheme` (see [here](https://github.com/sulu/sulu/blob/e160828f94181a8eab539afba5ae7eca40dedba2/src/Sulu/Bundle/CoreBundle/Resources/config/webspace.xml#L44)) as fallback when no actual request object is available. This might happen during cache warming in a functional test.

But if the request context is configured by Symfony's `framework.router.default_uri` setting like it is described in their [docs](https://symfony.com/doc/current/routing.html#generating-urls-in-commands), the mentioned container parameters still uses the defaults (localhost and http). There is a issue which points this out https://github.com/symfony/symfony/issues/53919.

In my case this leads to wrong url generation in a functional test (kernel test case) when the cache is warmed.

# Expected Behavior

Sulu should use the request context values as fallback even if they are configured by the `default_uri` setting.

# Possible Solutions

This can be fixed by using the request context service instead of the container params.

stollr:
This change fixes this issue (described in #7280).

| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | yes
| Fixed tickets | fixes #7280
| License | MIT

#### What's in this PR?

Changed the way how the default host and scheme is fetched in the WebspaceManager and DumpSitemapCommand. Instead of directly accessing the container params `router.request_context.host` and `router.request_context.scheme` the information is fetched via the `RequestContext` service of Symfony.

#### Why?

The webspace manager, which is configured in the core bundle, relied on the request context container parameters `router.request_context.host` and `router.request_context.scheme` as fallback when no actual request object is available. This might happen during cache warming in a functional test.

But if the request context is configured by Symfony's `framework.router.default_uri` setting like it is described in their [docs](https://symfony.com/doc/current/routing.html#generating-urls-in-commands), the mentioned container parameters still uses the defaults (localhost and http). There is a issue which points this out https://github.com/symfony/symfony/issues/53919.

This could have lead to wrong url generation in a functional test (kernel test case) when the cache is warmed.

### DIFF
--- a/src/Sulu/Bundle/CoreBundle/Resources/config/webspace.xml
+++ b/src/Sulu/Bundle/CoreBundle/Resources/config/webspace.xml
@@ -41,8 +41,7 @@
                 <argument key="base_class">%sulu_core.webspace.base_class%</argument>
             </argument>
             <argument>%kernel.environment%</argument>
-            <argument>%router.request_context.host%</argument>
-            <argument>%router.request_context.scheme%</argument>
+            <argument type="service" id="router.request_context" />
             <argument type="service" id="sulu_admin.metadata_provider_registry" />
 
             <tag name="sulu.localization_provider"/>
--- a/src/Sulu/Bundle/WebsiteBundle/Command/DumpSitemapCommand.php
+++ b/src/Sulu/Bundle/WebsiteBundle/Command/DumpSitemapCommand.php
@@ -19,6 +19,7 @@ use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Filesystem\Filesystem;
+use Symfony\Component\Routing\RequestContext;
 
 #[AsCommand(name: 'sulu:website:dump-sitemap')]
 class DumpSitemapCommand extends Command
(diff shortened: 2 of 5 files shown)


mamazu:
Yes, I already ran into a similar issue when trying to move the Webspace config to Symfony config parsing. However I would suggest leaving the constuctor as is and just report that passing those arguments is deprecated. We can just have a function that sets the service into the class and have it fall back to the old implemenatation if it isn't there. Then we can clean it up in the 3.0 release with a small bc break.

stollr:
As far as I see there is no BC break for 2.x in my implementation. But adding a setter for the service will introduce one more BC break to 3.0.

alexander-schranz:
Think a none bc break here would be change the service definition to an expression:

```diff
-            <argument>%router.request_context.scheme%</argument>
-            <argument>%router.request_context.host%</argument>
+            <argument type="expression">service('router.request_context').getScheme()</argument>
+            <argument type="expression">service('router.request_context').getHost()</argument>
```

stollr:
@alexander-schranz The service definition file does not have a BC promise, does it?

In the end, the constructor already fetches the values from the request context:
```php
            $this->defaultHost = $requestContext->getHost();
            $this->defaultScheme = $requestContext->getScheme();
```
Imho providing those values as parameter in the service definition won't change anything.

alexander-schranz:
@stollr as @mamazu mention we should not change the constructor on 2.6.

stollr:
I have simplified the patch. It still changes the constructor a little bit but now, it only checks for difference between `router.request_context.host|schema` and `framework.router.default_uri`. This way users will get a proper deprecation message if they have to change something to prepare for 3.0.

What do you think?

alexander-schranz (approved):

alexander-schranz:
@stollr Thank you! Ported this to 3.0.

PR #8113, opened 2025-08-19
Add Validate command to validate js build version against admin version

### LINKED ISSUE #8078: Validate Admin JS Version via Command
# Problem description

We should add to the Admin Bundle additional to validate the JS Build is uptodate, as some devs generally forget todo JS updates after composer update.

# Proposed solution

Add a `sulu:admin:validate-build` to validate the JS Build. For this we need to admin package service (`assets._package_sulu_admin`) and get the path to the main.js file. Read the content of it and regex out the build version:

Image

Keep in mind the version can be anything composer provides so we need more as \d.\d.\d. So maybe something like: `(JavaScript\ build\ version:\ )([a-zA-Z0-9.\/@-]+)\\n\\n` but we need check dev builds on custom branches and so on.

The command should error (exit 1) if it does not match `sulu.version`.

This command we should then call to the `sulu/skeleton` `composer lint` part. This should target 2.6 branch.

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #8078 
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?
A command to validate the js admin build number against the backend.

#### Why?
see issue.

#### Example Usage
`bin/console sulu:admin:validate-build `

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Command/ValidateBuildCommand.php
+++ b/src/Sulu/Bundle/AdminBundle/Command/ValidateBuildCommand.php
@@ -21,11 +21,16 @@ use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Style\SymfonyStyle;
 use Webmozart\Assert\Assert;
 
+/**
+ * @internal no backwards compatibility promise is given for this class it can be removed, moved or changed at any time
+ *           it still save to add `sulu:admin:validate-build` to your CI and sciprts. The command and its arguments
+ *           are under the backards compatibility promise.
+ */
 #[AsCommand(
     name: 'sulu:admin:validate-build',
     description: 'Validate that the admin build version is the same version as the installed sulu package.',
 )]
-class ValidateBuildCommand extends Command
+final class ValidateBuildCommand extends Command
 {
     public function __construct(
         private Packages $assets,
--- a/src/Sulu/Bundle/AdminBundle/Command/ValidateBuildCommand.php
+++ b/src/Sulu/Bundle/AdminBundle/Command/ValidateBuildCommand.php
@@ -1,0 +1,94 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of Sulu.
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\AdminBundle\Command;
+
+use Symfony\Component\Asset\Packages;
+use Symfony\Component\Console\Attribute\AsCommand;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+use Webmozart\Assert\Assert;
+
+#[AsCommand(
+    name: 'sulu:admin:validate-build',
+    description: 'Validate that the admin build version is the same version as the installed sulu package.',
+)]
+class ValidateBuildCommand extends Command
+{
+    public function __construct(
+        private Packages $assets,
+        private string $projectDir,
+        private string $suluVersion,
+    ) {
+        parent::__construct();
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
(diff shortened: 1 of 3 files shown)


src/Sulu/Bundle/AdminBundle/Command/ValidateBuildCommand.php
```diff
@@ -0,0 +1,84 @@
...
+            return self::FAILURE;
+        } else {
+            $io->success('JS Build is up to date.');
+        }
+
+        $input->setInteractive(false);
+
+        return $this->getApplication()->doRun(new ArrayInput([
+            'command' => 'lint:container',
+        ]), $output);
```
alexander-schranz: This has nothing todo with validating the admin build and should be removed

```suggestion
        return 0;
```

src/Sulu/Bundle/AdminBundle/Command/ValidateBuildCommand.php
```diff
@@ -0,0 +1,84 @@
...
+                matches: $matches,
+            ),
+            'Invalid regex for finding the version'
+        );
+
+        if (0 == \count($matches)) {
+            $io->error('Could not find version.');
+        }
+
+        return $matches[1];
```
alexander-schranz: validate before return

src/Sulu/Bundle/AdminBundle/Command/ValidateBuildCommand.php
```diff
@@ -0,0 +1,84 @@
...
+    private function getJavascriptBuildVersion(SymfonyStyle $io): string
+    {
+        $filePath = $this->publicDir . '/public' . $this->assets->getUrl('main.js', 'sulu_admin');
+        $io->info('JS Admin build path: ' . $filePath);
+
+        $matches = [];
+        Assert::integer(
+            \preg_match(
+                pattern: '#JavaScript build version: ([_a-zA-Z0-9.\/@-]+)\\\\n\\\\n#',
+                subject: \file_get_contents($filePath),
```
alexander-schranz: validate before access

src/Sulu/Bundle/AdminBundle/Command/ValidateBuildCommand.php
```diff
@@ -0,0 +1,84 @@
...
+        $input->setInteractive(false);
+
+        return $this->getApplication()->doRun(new ArrayInput([
+            'command' => 'lint:container',
+        ]), $output);
+    }
+
+    private function getJavascriptBuildVersion(SymfonyStyle $io): string
+    {
+        $filePath = $this->publicDir . '/public' . $this->assets->getUrl('main.js', 'sulu_admin');
```
alexander-schranz: validate if exists before access

alexander-schranz (changes requested):

src/Sulu/Bundle/AdminBundle/Command/ValidateBuildCommand.php
mamazu: In the issue you said something about running composer lint. I don't know what this was about.

src/Sulu/Bundle/AdminBundle/Command/ValidateBuildCommand.php
mamazu: Is that enough?
```php
        $version = $matches[1] ?? null;
        if (!\is_string($version)) {
            return null;
        }

        return $version;
```

mamazu:
Should be fixed.

src/Sulu/Bundle/AdminBundle/Command/ValidateBuildCommand.php
alexander-schranz: in the skeleton we require then a PR which adds a new `lint-admin-build` command which is added to the. `composer lint` list here: https://github.com/sulu/skeleton/blob/590f5b77073a55a1bc6a31fdf85304cc8648a71c/composer.json#L137-L149

mamazu: Ah okay, so nothing to worry here but another PR for the skeleton when this is merged?

alexander-schranz:
@mamazu thank you!

PR #8123, opened 2025-08-28
Fix no search results on repeated search

### LINKED ISSUE #8058: Repeated search clears results + wildcard issue in MassiveSearchBundle
| Q | A
| --- | ---
| Sulu Version | 2.6.7 / 2.6.10
| PHP Version | 8.1.2
| DB Version | MySQL 5.7.42
| Browser Version | Firefox 140.0

# Actual Behavior

There are two issues noticed with the `massive/search-bundle`:

1. On the first search (triggered by pressing Enter or clicking the magnifier), correct results are returned. However, when pressing Enter again immediately or clicking the magnifier again, the results disappear.
2. Wildcard search with `*` does not work from the **left** side of the word. For example, searching for `Serv*` works fine and returns results (e.g. for the word "Service"). But searching for `*vice` returns nothing.

# Expected Behavior

1. Re-triggering the same search via Enter or clicking the magnifier should preserve the result list.
2. Left-side wildcard searches like `*vice` should return expected results if matches exist.

# Steps to Reproduce

1. Open the Admin UI.
2. Use the global search bar.
3. Type in a known word (e.g. `Service`) and press Enter — results are returned.
4. Press Enter again — results disappear.
5. Try searching using `Serv*` — results are returned.
6. Try searching using `*vice` — no results returned, even though the term exists.

# Possible Solutions

- Alex from the Sulu team mentioned this may be a bug in the Frontend Component.
- Regarding wildcards: Lucene might already be using internal wildcards which could confuse the tokenized search behavior. Maybe `*vice` is not valid due to the way terms are split during indexing.

draconivis:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #8058
| Related issues/PRs | #8058
| License | MIT
| Documentation PR | -

# What's in this PR?

Fixes the issue with the global search where it wouldn't display search results if you press enter immediately again after searching.

#### Why?

#8058

### DIFF
--- a/src/Sulu/Bundle/SearchBundle/Resources/js/containers/Search/stores/searchStore.js
+++ b/src/Sulu/Bundle/SearchBundle/Resources/js/containers/Search/stores/searchStore.js
@@ -39,7 +39,6 @@ class SearchStore {
     }
 
     @action search(query: ?string, index: ?string) {
-        this.resetResults();
         this.query = query;
         this.indexName = index;
     }
--- a/src/Sulu/Bundle/SearchBundle/Resources/js/containers/Search/stores/searchStore.js
+++ b/src/Sulu/Bundle/SearchBundle/Resources/js/containers/Search/stores/searchStore.js
@@ -39,12 +39,12 @@ class SearchStore {
     }
 
     @action search(query: ?string, index: ?string) {
+        this.resetResults();
         this.query = query;
         this.indexName = index;
     }
 
     @action resetResults() {
-        this.result.splice(0, this.result.length);
         this.page = 1;
         this.pages = undefined;
         this.total = undefined;
--- a/src/Sulu/Bundle/SearchBundle/Resources/js/containers/Search/tests/stores/SearchStore.test.js
+++ b/src/Sulu/Bundle/SearchBundle/Resources/js/containers/Search/tests/stores/SearchStore.test.js
@@ -40,8 +40,8 @@ test.each([
     });
 });
 
-test('Do not send search request when no search term is given and reset to empty array', () => {
+test('Do not send search request when no search term is given', () => {
     searchStore.search(undefined);
     expect(ResourceRequester.getList).not.toBeCalled();
-    expect(searchStore.result).toEqual([]);
+    expect(searchStore.result).toEqual([{ id: 1 }]);
 });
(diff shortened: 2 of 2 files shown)


src/Sulu/Bundle/SearchBundle/Resources/js/containers/Search/stores/searchStore.js
```diff
@@ -39,7 +39,6 @@ class SearchStore {
     }
 
     @action search(query: ?string, index: ?string) {
-        this.resetResults();
```
alexander-schranz: ths line was added when the paginiation was added here: https://github.com/sulu/sulu/commit/98f886eb5e2c4914a90a9f4513190c8f80ee851c

please recheck if the paginiation still works like expected.

maybe @zanturik knows why it was required.

zanturik: As far as I remember, I added it to reset the page number. If user searched something and selected page X, but on the new search there is no page X, he won't see anything. So, when new search is received - we cleared everything, including "current page number".

Not sure it's not done something else now.

alexander-schranz: @zanturik thx for the quick response! Make sense, we should check that this behaviour stays the same.

PeeTee12:
By removing the line in this PR, there is an issue with pagination, just like @zanturik mentioned in previous discussion. When you are on page X and select a different category where page X doesn't exist, you'll get an empty result.

I'd like to suggest removing line 48 instead.

`this.result.splice(0, this.result.length);`

I believe this line deletes the whole search result when a user presses Enter twice while using the search bar.
By removing this line, the result resets to page 1 when pressing Enter again and when I select a different search category the result also resets to page 1. The search result doesn't get deleted which is the desired behavior.

@draconivis Can you please check if this solution works as expected and doesn't break pagination?

draconivis:
@PeeTee12 thanks for the suggestion! it seems to be working correctly with the suggested change.

alexander-schranz:
@draconivis @PeeTee12 @zanturik Thank you all!

PR #8124, opened 2025-08-28
Add groups and add the full functionality to the article-bundle

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | yes type => group
| Fixed tickets | fixes parts of #7672 
| Related issues/PRs | # 
| License | MIT
| Documentation PR | none

#### What's in this PR?

This PR add groups to the template and use this as a replacement of article-types.

#### Why?

Missing feature in 3.0.

#### Example Usage

```xml
<?xml version="1.0" ?>

    <key>article</key>
    <group>blog</group>

    <view>articles/article</view>
    <controller>Sulu\Content\UserInterface\Controller\Website\ContentController::indexAction</controller>
    <cacheLifetime>604800</cacheLifetime>

    ...
```

#### To Do

- [x] Update Tests

### DIFF
--- a/packages/article/src/Infrastructure/Sulu/Admin/ArticleAdmin.php
+++ b/packages/article/src/Infrastructure/Sulu/Admin/ArticleAdmin.php
@@ -19,6 +19,8 @@ use Sulu\Bundle\AdminBundle\Admin\Navigation\NavigationItemCollection;
 use Sulu\Bundle\AdminBundle\Admin\View\ToolbarAction;
 use Sulu\Bundle\AdminBundle\Admin\View\ViewBuilderFactoryInterface;
 use Sulu\Bundle\AdminBundle\Admin\View\ViewCollection;
+use Sulu\Bundle\AdminBundle\Metadata\FormMetadata\FormGroup;
+use Sulu\Bundle\AdminBundle\Metadata\GroupProviderInterface;
 use Sulu\Component\Localization\Manager\LocalizationManagerInterface;
 use Sulu\Component\Security\Authorization\PermissionTypes;
 use Sulu\Component\Security\Authorization\SecurityCheckerInterface;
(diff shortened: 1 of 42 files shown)


packages/article/src/Infrastructure/Sulu/Content/ArticleSmartContentProvider.php
wachterjohannes, line 1: Add usage of group filter

packages/article/src/Infrastructure/Sulu/Admin/ArticleAdmin.php
```diff
@@ -49,111 +53,208 @@ public function __construct(
...
-            $navigationItem = new NavigationItem('sulu_article.articles');
-            $navigationItem->setPosition(20);
-            $navigationItem->setIcon('su-newspaper');
-            $navigationItem->setView(static::LIST_VIEW);
+        if (!$this->securityChecker->hasPermission(static::SECURITY_CONTEXT, PermissionTypes::EDIT)) {
+            return;
+        }
+
+        $hasArticleTypeWithEditPermissions = false;
+        if (1 === \count($this->groupProvider->getGroups())) {
```
Prokyonn, line 67: I guess this can be simplified with only the foreach loop without the if 🤔

src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/Loader/TemplateXmlLoader.php
```diff
@@ -66,6 +66,12 @@ protected function parse(string $resource, \DOMXPath $xpath, ?string $type): For
         \assert(\is_string($templateKey), 'Expected the template key of "' . $resource . '" to be defined.');
         $form->setKey($templateKey);
 
+        $templateGroup = $this->getValueFromXPath('/x:template/x:group', $xpath);
+        if (null !== $templateGroup) {
+            \assert(\is_string($templateGroup), 'Expected the template group to be "' . $resource . '" string or undefined.');
```
Prokyonn: This messages does not make any sense

src/Sulu/Bundle/AdminBundle/Metadata/GroupProvider.php
```diff
@@ -0,0 +1,49 @@
...
+            }
+            $groups[$group] = $groups[$group]->withTemplate($articleForm->getKey());
+        }
+
+        return $groups;
+    }
+
+    private function getGroupTitle(string $groupIdentifier): string
+    {
+        return \ucfirst($groupIdentifier);
```
Prokyonn, line 47: I think it would make sense to use a translation here, but that can be improved in another PR

packages/article/src/Infrastructure/Sulu/Admin/ArticleAdmin.php
wachterjohannes: no because for a single types we only check for the "main" article permissions

Prokyonn:
Looks good 👍 
@wachterjohannes should we add the "group"  in the contentResolving process? Not sure how this was in 2.6

image

Prokyonn (approved):

src/Sulu/Bundle/AdminBundle/Metadata/GroupProvider.php
wachterjohannes: Yeah we should do that in an additional PR

wachterjohannes:
@Prokyonn no in 2.6. we have no group in the twig - so i would not do that by now

src/Sulu/Bundle/AdminBundle/Metadata/GroupProvider.php
wachterjohannes: done here: https://github.com/sulu/sulu/pull/8255

PR #8130, opened 2025-09-02
Fix category tree search

MarkusHolstein:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes # 
| Related issues/PRs | Replaces #8045
| License | MIT

#### What's in this PR?

When searching for a category string, nearly always the expandedIds parameter is set in the query string to the controller.
As in a full tree view in the category administration, there is no $parentId in the request parameters, thus $categoriesByParentId[$parentId] delivers null, which leads to iterator_to_array(): Argument https://github.com/sulu/sulu/pull/1 ($iterator) must be of type Traversable|array, null given at src/Sulu/Component/Rest/ListBuilder/CollectionRepresentation.php:35, though categories were found, but not with parent "null"
This fixes this behaviour. and allows to search a full tree.

#### Why?

The server responds with a 500 Exception.
Replaces #8045 as suggested.

#### Example Usage

Call url https://sulu.rocks/admin/api/categories?page=1&locale=en&expandedIds=3&limit=10&fields=name,id&search=Test3&flat=true

### DIFF
--- a/src/Sulu/Bundle/CategoryBundle/Controller/CategoryController.php
+++ b/src/Sulu/Bundle/CategoryBundle/Controller/CategoryController.php
@@ -260,7 +260,8 @@ class CategoryController extends AbstractRestController implements ClassResource
             }
         }
 
-        if (!empty($expandedIds) && $request->get('search', false)) {
+        $search = $request->get('search');
+        if (!empty($expandedIds) && empty($search)) {
             $categoriesByParentId = [];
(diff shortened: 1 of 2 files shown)


alexander-schranz:
merged in https://github.com/sulu/sulu/pull/8274 Thx @MarkusHolstein

PR #8142, opened 2025-09-05
Remove old PathCleanup xml replace loader and compilerpass

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | yes
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | https://github.com/sulu/sulu/pull/8137, requires #8144
| License | MIT
| Documentation PR | sulu/sulu-docs# 

#### What's in this PR?

Remove old PathCleanup xml replace loader and compilerpass. 

#### Why?

For #8137

Added a Todo to our checklist to remove the old replacers files as soon as checked against the new sluggers. But 99% of the replacers are handled by symfony slugger now.

PR #8137, opened 2025-09-03
Say goodbye to the old phpcr, document manager, page bundle classes, cleanup 🫡 Part 2

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | #7995, #7672, next https://github.com/sulu/sulu/pull/8160
| License | MIT
| Documentation PR | sulu/sulu-docs# 

#### What's in this PR?

Remove DocumentManager, PHPCR based classes and services.

🫱  Added lines are only PHPStan baseline update. All other parts where handled in own pull requests linked below.

#### Why?

We removed the bundle in #7995.

PR #8144, opened 2025-09-05
Use new PathCleanup service in Media Bundle

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | yes
| Deprecations? | no yes 
| Fixed tickets | fixes # 
| Related issues/PRs | #8137, #8142
| License | MIT

#### What's in this PR?

The Media service depands still on the old PathCleanup services which should not be the case. This PR will migrate to the new one.

#### Why?

The CI failed in https://github.com/sulu/sulu/pull/8142


src/Sulu/Bundle/MediaBundle/Resources/config/services.xml
```diff
@@ -116,7 +116,7 @@
                  public="true">
             <argument type="service" id="sulu_media.disposition_type.resolver"/>
             <argument type="service" id="sulu.repository.media"/>
-            <argument type="service" id="sulu.content.path_cleaner"/>
+            <argument type="service" id="sulu_route.path_cleanup"/>
```
alexander-schranz, line 119: Not 100% happy that media bundle depends on a route bundle service but add it to our ToDo list to rediscuss this, alternative is slug it ourselve or use dependency inversion principle to have atleast an own interface on media bundle side which uses pathcleanup or fallbacks to an own slugger if route bundle is not installed (more theoretical case of clean architecture)

Prokyonn (approved):

PR #8162, opened 2025-09-08
Use flysystem in memory adapter for tests

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Use flysystem in memory adapter for tests.

#### Why?

Avoids that we need cleanup old files in test runs.

### DIFF
--- a/src/Sulu/Bundle/TestBundle/Resources/app/config/config.php
+++ b/src/Sulu/Bundle/TestBundle/Resources/app/config/config.php
@@ -38,4 +38,8 @@ return static function(PhpFileLoader $loader, ContainerBuilder $container) {
     } else {
         $loader->import('symfony-5-4.yml');
     }
+
+    if (\str_starts_with($container->getParameter('kernel.environment'), 'test')) {
+        $loader->import('flysystem_test.yml');
+    }
 };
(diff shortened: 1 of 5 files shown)


mamazu (approved):
True we don't need to depend on 3.29, 3.0 is sufficient good idea.

src/Sulu/Bundle/TestBundle/Resources/app/config/config.php
```diff
@@ -38,4 +38,8 @@
     } else {
         $loader->import('symfony-5-4.yml');
     }
+
+    if (\str_starts_with($container->getParameter('kernel.environment'), 'test')) {
+        $loader->import('flysystem_test.yml');
+    }
```
alexander-schranz, line 44: not using `when@test` in the test kernel as there are sometimg additional test envs like `test_with_something`. and we want all test envs to apply this config.

Prokyonn:
@alexander-schranz tests are failing

mamazu:
Probably because the download controller that just forwards the downloading via to the Webserver doesn't find the file anymore.

alexander-schranz:
the problem is that most test depend on creating things in admin context and read them in website context while between the memory storage get lost. currently will not spend more time on refactoring this tests so closing it until we may have single kernel setup enabled in test run.

the other issue is fixed here: https://github.com/sulu/sulu/pull/8172

PR #8169, opened 2025-09-09
Fix first route matching and loading route by site

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Get the first route match in the routes provider.

#### Why?

Technically there can exists a route on e.g. slug `/test` with and without a `site` defined.

### DIFF
--- a/packages/route/src/Application/ResourceLocator/ResourceLocatorGenerator.php
+++ b/packages/route/src/Application/ResourceLocator/ResourceLocatorGenerator.php
@@ -33,7 +33,6 @@ final readonly class ResourceLocatorGenerator implements ResourceLocatorGenerato
                 'resourceKey' => $request->parentResourceKey,
                 'resourceId' => $request->parentResourceId,
                 'locale' => $request->locale,
-                'site' => $request->site,
             ]);
 
             $parentPath = $parentRoute?->getSlug() ?: '/';
--- a/packages/route/src/Application/Routing/Matcher/RouteHistoryDefaultsProvider.php
+++ b/packages/route/src/Application/Routing/Matcher/RouteHistoryDefaultsProvider.php
@@ -45,7 +45,6 @@ final readonly class RouteHistoryDefaultsProvider implements RouteDefaultsProvid
         $targetRoute = $this->routeRepository->findOneBy([
             'resourceKey' => $resourceKey,
             'resourceId' => $resourceId,
-            'site' => $route->getSite(),
             'locale' => $route->getLocale(),
         ]);
 
--- a/packages/route/src/Domain/Repository/RouteRepositoryInterface.php
+++ b/packages/route/src/Domain/Repository/RouteRepositoryInterface.php
@@ -27,7 +27,6 @@ use Sulu\Route\Domain\Model\Route;
  *         resourceId: string,
  *     },
  * }
- *
  * @phpstan-type RouteSortBy array{
  *     site?: 'asc'|'desc',
  * }
--- a/packages/route/src/Domain/Repository/RouteRepositoryInterface.php
+++ b/packages/route/src/Domain/Repository/RouteRepositoryInterface.php
@@ -16,6 +16,7 @@ use Sulu\Route\Domain\Model\Route;
 /**
  * @phpstan-type RouteFilter array{
  *     site?: string|null,
+ *     siteOrNull?: string|null,
  *     locale?: string,
  *     locales?: string[],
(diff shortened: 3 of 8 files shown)


packages/route/src/Application/ResourceLocator/ResourceLocatorGenerator.php
```diff
@@ -33,7 +33,6 @@ public function generate(ResourceLocatorRequest $request): string
                 'resourceKey' => $request->parentResourceKey,
                 'resourceId' => $request->parentResourceId,
                 'locale' => $request->locale,
-                'site' => $request->site,
```
alexander-schranz, line 36: The site doesn't matter here, article is part of no site and the pageTreeRoute is part of a site. It should already be unique by resourceKey, resourceId and locale.

packages/route/src/Application/Routing/Matcher/RouteHistoryDefaultsProvider.php
```diff
@@ -45,7 +45,6 @@ public function getDefaults(Route $route): array
         $targetRoute = $this->routeRepository->findOneBy([
             'resourceKey' => $resourceKey,
             'resourceId' => $resourceId,
-            'site' => $route->getSite(),
```
alexander-schranz, line 48: same as above. even a page could have been moved from `webspace A` to `webspace B` which means we need target the new webspace here.

packages/route/src/Domain/Repository/RouteRepositoryInterface.php
```diff
@@ -16,6 +16,7 @@
 /**
  * @phpstan-type RouteFilter array{
  *     site?: string|null,
+ *     siteOrNull?: string|null,
```
alexander-schranz, line 19: make behaviour more transparent here?

packages/route/src/Infrastructure/Doctrine/Repository/RouteRepository.php
```diff
@@ -49,6 +50,18 @@ public function findOneBy(array $filters): ?Route
         return $queryBuilder->getQuery()->getOneOrNullResult(Query::HYDRATE_OBJECT);
     }
 
+    public function findFirstBy(array $filters, array $sortBys = []): ?Route
```
alexander-schranz, line 55: I did in previous project add `sortBys` to `findOneBy` but @chirimoya found it confusing why a `findOneBy` can have sorting and said it would be better to go with a `findFirstBy`.

packages/route/src/Domain/Repository/RouteRepositoryInterface.php
```diff
@@ -26,6 +27,10 @@
  *         resourceId: string,
  *     },
  * }
+ *
+ * @phpstan-type RouteSortBy array{
+ *     site?: 'asc'|'desc',
```
alexander-schranz, line 31: unsure about nullhandling in the different DBs.

Prokyonn (approved):

packages/route/src/Infrastructure/Doctrine/Repository/RouteRepository.php
Prokyonn: I think both versions make sense in a way, but `findFirstBy` is fine for me

packages/route/src/Infrastructure/Doctrine/Repository/RouteRepository.php
```diff
@@ -130,6 +157,29 @@ protected function createQueryBuilder(array $filters): QueryBuilder
...
+                            || $this->entityManager->getConnection()->getDatabasePlatform() instanceof OraclePlatform
+                        )
+                        && \array_key_exists('siteOrNull', $filters)
+                    ) => match ($order) {
+                        'asc' => 'desc',
+                        'desc' => 'asc',
+                        default => $order,
+                    },
+                    default => $order,
+                };
```
alexander-schranz: Sure not happy with this one but could not find a way to handle that constantly over all DBs even the documentation reference ASC NULLS FIRST, but it does not exist or work in the ORM mayby only DBAL where the query is a literal value given directly to the database.

The `testFindFirstBySiteOrNullSlugLocale` avoids a regression of this logic aslong as we test postgres in the CI.

PR #8171, opened 2025-09-09
Simplify Reference Path / extraction logic

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Adds the `path` directly to the `Reference` dto.
Adds the recursive extraction logic for References on the ContentView class instead of the ReferenceRefreshers

#### Why?

Simplified and less duplicated code.

### DIFF
--- a/packages/content/src/Application/ContentResolver/Value/ContentView.php
+++ b/packages/content/src/Application/ContentResolver/Value/ContentView.php
@@ -224,14 +224,12 @@ class ContentView
      * Recursively collect all references from this ContentView and nested ContentViews.
      * Each reference will have its path set to indicate where it was found.
      *
-     * @return array<Reference>
+     * @return iterable<Reference>
      */
-    public function getAllReferencesRecursively(string $basePath = ''): array
+    public function getAllReferencesRecursively(string $basePath = ''): iterable
     {
-        $allReferences = [];
-
         foreach ($this->references as $reference) {
-            $allReferences[] = new Reference(
+            yield new Reference(
                 $reference->getResourceId(),
                 $reference->getResourceKey(),
                 $basePath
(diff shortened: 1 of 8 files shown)


packages/content/src/Application/ContentResolver/Value/ContentView.php
```diff
@@ -217,4 +217,37 @@ public function setReferences(array $references): self
 
         return $this;
     }
+
+    /**
+     * Recursively collect all references from this ContentView and nested ContentViews.
+     * Each reference will have its path set to indicate where it was found.
+     *
+     * @return array<Reference>
+     */
```
alexander-schranz: We maybe at some state move it to a service so I would make it currently `@internal`.

```suggestion
    /**
     * @internal This method can be removed at any time no backwards compatibility promise is given for this.
     *
     * Recursively collect all references from this ContentView and nested ContentViews.
     * Each reference will have its path set to indicate where it was found.
     *
     * @return array<Reference>
     */
```

packages/content/src/Application/ContentResolver/Value/ContentView.php
```diff
@@ -217,4 +217,39 @@ public function setReferences(array $references): self
...
+
+    /**
+     * @internal This method can be removed at any time no backwards compatibility promise is given for this.
+     *
+     * Recursively collect all references from this ContentView and nested ContentViews.
+     * Each reference will have its path set to indicate where it was found.
+     *
+     * @return array<Reference>
+     */
+    public function getAllReferencesRecursively(string $basePath = ''): array
```
mamazu: Why not use a generator here?

alexander-schranz (approved):

packages/content/src/Application/ContentResolver/Value/ContentView.php
Prokyonn: Good idea :D

PR #8173, opened 2025-09-09
Move localizations handling to fix saving of Articles

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | #7672
| License | MIT

#### What's in this PR?

Temporary move localizations handling to content controller directly.

#### Why?

During SettingsResolver we don't know the current webspace, Resolving is also done during refresh of reference which is webspace unspecific so resolver should not access the webspace.

#### ToDo

 - [x] Test example with multiple available Languages

### DIFF
--- a/packages/content/src/UserInterface/Controller/Website/ContentController.php
+++ b/packages/content/src/UserInterface/Controller/Website/ContentController.php
@@ -15,8 +15,15 @@ namespace Sulu\Content\UserInterface\Controller\Website;
 
 use Sulu\Bundle\HttpCacheBundle\CacheLifetime\CacheLifetimeEnhancerInterface;
 use Sulu\Bundle\PreviewBundle\Preview\Preview;
+use Sulu\Component\Localization\Localization;
+use Sulu\Component\Webspace\Analyzer\Attributes\RequestAttributes;
+use Sulu\Component\Webspace\Manager\WebspaceManagerInterface;
+use Sulu\Component\Webspace\Webspace;
 use Sulu\Content\Application\ContentResolver\ContentResolverInterface;
 use Sulu\Content\Domain\Model\DimensionContentInterface;
+use Sulu\Content\Domain\Model\RoutableInterface;
+use Sulu\Route\Application\Routing\Generator\RouteGeneratorInterface;
+use Sulu\Route\Domain\Repository\RouteRepositoryInterface;
 use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
 use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\Request;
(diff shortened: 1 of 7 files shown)


Prokyonn:
@alexander-schranz In some tests there are `// TODO this should not be necessary` comments, all of them can be removed, when we do not depend on the site parameter anymore

packages/content/src/UserInterface/Controller/Website/ContentController.php
```diff
@@ -118,13 +130,94 @@ protected function renderSuluView(
...
 
+        $services['sulu_route.route_repository'] = RouteRepositoryInterface::class;
+        $services['sulu_route.route_generator'] = RouteGeneratorInterface::class;
+        $services['sulu_core.webspace.webspace_manager'] = WebspaceManagerInterface::class;
+
         return $services;
     }
+
+    /**
+     * TODO maybe we move this into a content website resolver service, depending on route localization switcher service.
```
alexander-schranz, line 160: ideal case would maybe something like this:

```php
// RouteBundle
interface LanguageSwitcherProvider {
    public function provide(string $resourceKey, string $resourceId, string $site, array $publishedLocales, $fallbackToStartpage = false): array;
}
```

As the route bundle don't know the other locales the service would require addtional locale provider as we fallback to startpage:

```php
// RouteBundle
interface SiteLocaleProviderInterface {
    public function provide(string $site): array
}

// RouteBundle Infrastructure Symfony HttpKernel
class KernelLocaleProvider implements SiteLocaleProviderInterface {
    public function provide(string $site): array
    {
        return $this->kernel->getEnabledLocales();
    }
}
```

And the page bundle decorates it via:

```php
// PageBundle decorates LocaleProvider
class WebspaceLocaleProvider implements SiteLocaleProviderInterface {
    public function provide(string $site): array
}
```

But for the next pre-release I currently would keep it in the content controller.

Prokyonn: I think it would be a good idea to link this comment in the TODO or create a new issue / add it to the existing roadmap issue

Prokyonn (approved):

packages/content/src/UserInterface/Controller/Website/ContentController.php
alexander-schranz: https://github.com/sulu/sulu/issues/8175

packages/content/src/UserInterface/Controller/Website/ContentController.php
```diff
@@ -118,13 +130,95 @@ protected function renderSuluView(
...
+            ];
+        }
+
+        $webspaceLocales = $this->container->get('sulu_core.webspace.webspace_manager')
+            ->getWebspaceCollection()
+            ->getWebspace($webspaceKey)
+            ?->getAllLocalizations() ?? [];
+
+        foreach ($webspaceLocales as $webspaceLocale) {
+            $locale = $webspaceLocale->getLocale(Localization::DASH);
```
benr77, line 206: @alexander-schranz Whilst testing the 3.0 branch, DASH here seems to have broken my entire site  - I switched it to UNDERSCORE and things started working again. Think this is a BC break as my DB is full of references to `en_gb` and not `en-gb` etc.

alexander-schranz: Can you create an issue for it? Need to check how the behaviour was on 2.6.

benr77: https://github.com/sulu/sulu/issues/8217

PR #8181, opened 2025-09-10
Make sure PathCleanup service behave the same as in 2.6

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? |  yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | #7672
| License | MIT

#### What's in this PR?

Make sure PathCleanup service behave the same as in 2.6.

#### Why?

In 2.x we had a custom replacers logic and a regex. Later we replaced the regex with AsciiSlugger, but never cleanup the replacers. We removed the replacer loader already in 3.0 but we want to keep some custom replacers for umlauts and `&` character.

This pull request readds some special replacers for `Ä`, `Ö`, `Ü` and `&` which are handled differently by AsciiSlugger in Symfony.

### DIFF
--- a/packages/route/tests/Unit/Application/ResourceLocator/PathCleanup/resources/replacers-26.xml
+++ b/packages/route/tests/Unit/Application/ResourceLocator/PathCleanup/resources/replacers-26.xml
@@ -1,0 +1,1220 @@
+<?xml version="1.0" encoding="utf-8"?>
+<replacers>
+    <item>
+        <column name="locale">default</column>
+        <column name="from"> </column>
+        <column name="to">-</column>
+    </item>
+    <item>
+        <column name="locale">default</column>
+        <column name="from">+</column>
+        <column name="to">-</column>
+    </item>
+    <item>
+        <column name="locale">default</column>
+        <column name="from">.</column>
+        <column name="to">-</column>
+    </item>
+    <item>
+        <column name="locale">default</column>
+        <column name="from">^</column>
+        <column name="to">-</column>
+    </item>
+    <item>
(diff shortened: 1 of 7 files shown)


src/Sulu/Bundle/MediaBundle/Tests/Functional/Controller/MediaControllerTest.php
```diff
@@ -451,7 +451,7 @@ public function testDownloadHeaderUmlauts(): void
 
         $this->assertHttpStatusCode(200, $this->client->getResponse());
         $this->assertEquals(
-            'inline; filename=wochentlich.jpeg; filename*=utf-8\'\'w%C3%B6chentlich.jpeg',
+            'inline; filename=woechentlich.jpeg; filename*=utf-8\'\'w%C3%B6chentlich.jpeg',
```
alexander-schranz, line 454: expected change as same again as on 2.6: https://github.com/sulu/sulu/blob/8089685a960484e22ea007ca7f42875792d93bfe/src/Sulu/Bundle/MediaBundle/Tests/Functional/Controller/MediaControllerTest.php#L446

packages/route/tests/Unit/Application/ResourceLocator/PathCleanup/resources/replacers-26.xml
alexander-schranz, line 1: Now this file only used for tests, copy of 2.6 file to test all our old slugs replacers.

Prokyonn (approved):

PR #8184, opened 2025-09-10
Prepare compatibility to doctrine/orm 3

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? | no 
| BC breaks? | yes
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Prepare compatibility to doctrine/orm 3.

#### Why?

The doctrine/orm 3 is not yet supported but some bc breaks will come to us which we need to tackle as soon as possible.

### DIFF
--- a/packages/article/config/doctrine/Article/Article.orm.xml
+++ b/packages/article/config/doctrine/Article/Article.orm.xml
@@ -6,7 +6,7 @@
         name="Sulu\Article\Domain\Model\Article"
         table="ar_articles"
     >
-        <id name="uuid" type="string" column="uuid">
+        <id name="uuid" type="string" column="uuid" length="36">
             <generator strategy="NONE"/>
         </id>
 
--- a/packages/content/src/Application/ContentMetadataInspector/ContentMetadataInspector.php
+++ b/packages/content/src/Application/ContentMetadataInspector/ContentMetadataInspector.php
@@ -13,7 +13,6 @@ declare(strict_types=1);
 
 namespace Sulu\Content\Application\ContentMetadataInspector;
 
-use Doctrine\Common\Util\ClassUtils;
 use Doctrine\ORM\EntityManagerInterface;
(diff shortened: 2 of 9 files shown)


packages/article/config/doctrine/Article/Article.orm.xml
```diff
@@ -6,7 +6,7 @@
         name="Sulu\Article\Domain\Model\Article"
         table="ar_articles"
     >
-        <id name="uuid" type="string" column="uuid">
+        <id name="uuid" type="string" column="uuid" length="36">
```
alexander-schranz, line 9: ORM 3 requires a length when string is used as primary key or join column

packages/content/src/Application/ContentMetadataInspector/ContentMetadataInspector.php
```diff
@@ -13,7 +13,6 @@
 
 namespace Sulu\Content\Application\ContentMetadataInspector;
 
-use Doctrine\Common\Util\ClassUtils;
```
alexander-schranz, line 16: removed and looks this behaviour currently not required on our versions

packages/content/src/Infrastructure/Doctrine/MetadataLoader.php
```diff
@@ -39,7 +39,7 @@ final class MetadataLoader
 {
     public function loadClassMetadata(LoadClassMetadataEventArgs $event): void
     {
-        /** @var ClassMetadataInfo<object> $metadata */
+        /** @var ClassMetadata<object> $metadata */
```
alexander-schranz, line 42: ClassMetadataInfo inherits from ClassMetadata, ClassMetadataInfo will be removed in newer doctrine versions

Prokyonn (approved):

PR #8190, opened 2025-09-16
Be stricter around required dependencies

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? |  yes
| New feature? | no
| BC breaks? | no 
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Define more required dependencies.

#### Why?

There should be no dependency on dependency be used.

Used https://github.com/shipmonk-rnd/composer-dependency-analyser to catch a few.


composer.json
```diff
@@ -25,10 +25,13 @@
         "docs": "https://docs.sulu.io/"
     },
     "require": {
-        "php": "^8.0 || ^8.1",
+        "php": "^8.0",
         "ext-dom": "*",
+        "ext-fileinfo": "*",
+        "ext-filter": "*",
         "ext-json": "*",
-        "ext-simplexml": "*",
```
alexander-schranz, line 31: didn't found any usage of simplexml in Sulu itself.

composer.json
```diff
@@ -85,12 +94,14 @@
...
         "symfony/framework-bundle": "^5.4 || ^6.0",
         "symfony/http-client": "^5.4 || ^6.0",
+        "symfony/http-client-contracts": "^2.4 || ^3.0",
         "symfony/http-foundation": "^5.4 || ^6.0",
         "symfony/http-kernel": "^5.4 || ^6.0",
         "symfony/intl": "^5.4 || ^6.0",
         "symfony/mailer": "^5.4 || ^6.0",
         "symfony/mime": "^5.4 || ^6.0",
         "symfony/options-resolver": "^5.4 || ^6.0",
+        "symfony/password-hasher": "^5.4 || ^6.0",
```
alexander-schranz, line 107: this changes a little bit the dependency tree as actually 2.5 already did install 7.3 password hasher which is kind of unexpected as we support symfony 7 only in Sulu 2.6

composer.json
```diff
@@ -133,7 +146,6 @@
         "phpstan/phpstan-symfony": "^1.0",
         "phpstan/phpstan-webmozart-assert": "^1.0",
         "phpunit/phpunit": "^9.5.20",
-        "psr/event-dispatcher": "^1.0",
```
alexander-schranz, line 136: is also used in src

composer.json
```diff
@@ -176,6 +188,7 @@
         "swiftmailer/swiftmailer": "< 6.1.3",
         "symfony/doctrine-bridge": "< 5.4.0",
         "symfony/error-handler": "< 5.4.0",
+        "symfony/polyfill-php81": "<1.23",
```
alexander-schranz, line 192: without it install polyfill 1.22 and upgrades symfony/form to 6.0 on lowest to keep the polyfill as lowest. to avoid that I added a conflict to <1.23

composer.json
```diff
@@ -25,10 +25,13 @@
         "docs": "https://docs.sulu.io/"
     },
     "require": {
-        "php": "^8.0 || ^8.1",
+        "php": "^8.0",
         "ext-dom": "*",
+        "ext-fileinfo": "*",
```
alexander-schranz, line 30: Required when media bundle is used. Think we are fine to require it generally so Sulu runs smoothly.

Prokyonn (approved):

Prokyonn (approved):

PR #8191, opened 2025-09-16
Add shipmonk/composer-dependency-analyser for lint our dependencies

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no  
| Fixed tickets | fixes # 
| Related issues/PRs | https://github.com/sulu/sulu/pull/8190
| License | MIT

#### What's in this PR?

Add shipmonk/composer-dependency-analyser for lint our dependencies

#### Why?

See https://github.com/sulu/sulu/pull/8190

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,7 +33,6 @@
         "ext-libxml": "*",
         "ext-mbstring": "*",
         "ext-simplexml": "*",
-        "composer/package-versions-deprecated": "^1.8",
         "contao/imagine-svg": "^1.0",
         "dantleech/phpcr-migrations-bundle": "^1.3",
         "doctrine/annotations": "^1.2 || ^2.0",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -33,7 +33,6 @@
         "ext-libxml": "*",
         "ext-mbstring": "*",
         "ext-simplexml": "*",
-        "composer/package-versions-deprecated": "^1.8",
```
alexander-schranz, line 36: not used by us was required for Sulu with PHP 7 support but not on 2.5 which min requirement is 8.0

alexander-schranz: we need to keep it on plugins as on `lowest` some dependency require it.

Prokyonn (approved):

PR #7718, opened 2025-01-08
Remove not longer required Sulu Kernel Browser

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no 
| BC breaks? | no
| Deprecations? | no  
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT
| Documentation PR | sulu/sulu-docs# 

#### What's in this PR?

Remove not longer required Sulu Kernel Browser.

#### Why?

We have a min requirement to Symfony 5.4 so this BC bridge can be removed.

PR #8198, opened 2025-09-16
Remove not longer required `SuluKernelBrowser`

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | yes
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | #7718 
| License | MIT
| Documentation PR | -

#### What's in this PR?
Remove not longer required Sulu Kernel Browser.

#### Why?
We have a min requirement to Symfony 5.4 so this BC bridge can be removed.


Prokyonn:
@mamazu I briefly discussed this with @wachterjohannes , as this is a BC break for all the tests and maybe even projects that used this, it is probably best to target this MR to the 3.0 branch

src/Sulu/Bundle/ActivityBundle/phpunit.xml.dist
```diff
@@ -22,6 +22,7 @@
     <php>
         <server name="APP_ENV" value="test" force="true"/>
         <env name="KERNEL_CLASS" value="Sulu\Bundle\ActivityBundle\Tests\Application\Kernel"/>
+          <env name="DATABASE_URL" value="mysql://root:ChangeMe@127.0.0.1:3306/sulu_test"/>
```
Prokyonn: Why is this necessary now?

mamazu: This was me trying to get the local setup working. I don't know how to do that otherwise

Prokyonn: What is the issue? It should work out of the box with the docker-compose from the repository. Otherwise you can adjust the config in the parameters.yml file.

Running a complete bundle you can run e.g. `php bin/runtests -C -i -t article` from the sulu/sulu directory, specific tests must be run from the bundle directly e.g. from the `packages/content` directory you can run `php ../../vendor/bin/phpunit tests/<path-to-the-test-file>.php`

mamazu: It gives me connection errors with mysql. Strangely this works in the newer bundles:
```
Caused by
PDOException: SQLSTATE[HY000] [1045] Access denied for user 'root'@'172.39.3.1' (using password: NO)
```
I couldn't find this IP anywhere in the code though.

src/Sulu/Bundle/ContactBundle/Tests/Functional/Controller/AccountControllerTest.php
```diff
@@ -129,13 +129,14 @@ function($account) {
...
-            '/api/accounts?flat=true',
-            [
-                'ids' => \implode(',', $ids),
-                'page' => 2,
-                'limit' => 10,
-            ]
-        );
+            '/api/accounts?' . \http_build_query(
+                [
+                    'flat' => true,
```
alexander-schranz:
```suggestion
                    'flat' => 'true',
```

src/Sulu/Bundle/ContactBundle/Tests/Functional/Controller/AccountControllerTest.php
```diff
@@ -97,10 +97,10 @@ function($account) {
 
         $this->client->jsonRequest(
             'GET',
-            '/api/accounts?flat=true',
-            [
+            '/api/accounts?' . \http_build_query([
+                'flat' => true,
```
alexander-schranz:
```suggestion
                'flat' => 'true',
```

alexander-schranz (changes requested):
Small change request, can you add a line to the UPGRADE.md with a example converting 'GET' requests to http_build_query?

mamazu:
Docs added. I'm not sure it's the best tone for it. But open to suggestions.

alexander-schranz:
@mamazu thank you!

PR #8218, opened 2025-09-22
Add test for `CollectionManager->delete`

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?
Adding a test that when deleting a collection it's meta data also gets deleted.

#### Why?
In preparation to moving this all to the database. Right now we have a cascade delete defined in the "CollectionMeta" that means we don't have to explicitly delete those but I want to add a test first so that we can be sure this works, before and after the refactoring.

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/Tests/Unit/Collection/CollectionManagerTest.php
+++ b/src/Sulu/Bundle/MediaBundle/Tests/Unit/Collection/CollectionManagerTest.php
@@ -12,8 +12,7 @@
 namespace Sulu\Bundle\MediaBundle\Tests\Unit\Collection;
 
 use Doctrine\Common\Collections\ArrayCollection;
-use Doctrine\ORM\EntityManager;
-use PHPUnit\Framework\TestCase;
+use Doctrine\ORM\EntityManagerInterface;
 use Prophecy\Argument;
 use Prophecy\PhpUnit\ProphecyTrait;
 use Prophecy\Prophecy\ObjectProphecy;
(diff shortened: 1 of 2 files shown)


alexander-schranz:
PS: this is already handled also on database level: https://github.com/sulu/sulu/blob/e648c1aa488d756553ef7af9c96ed752efac7b9b/src/Sulu/Bundle/MediaBundle/Resources/config/doctrine/CollectionMeta.orm.xml#L19

Okay to add test a unit test collectionmanager delete but testing the repo or meta is unnecessary as its just a Unit test and so nothing on collection meta need to be tested here.

src/Sulu/Bundle/MediaBundle/Tests/Unit/Collection/CollectionManagerTest.php
```diff
@@ -217,4 +217,20 @@ public function testSaveWithoutUserId(): void
...
+    public function testDeletingCollectionDeletesMeta()
+    {
+        $collectionEntity = $this->createEntity(1, 'de');
+
+        /** @var EntityManagerInterface $entityManager */
+        $entityManager = self::getContainer()->get(EntityManagerInterface::class);
+        $entityManager->persist($collectionEntity);
+        $entityManager->flush();
+
+        self::assertEquals(1, $entityManager->getRepository(CollectionMeta::class)->count());
```
alexander-schranz: this is a unit test this does test nothing.

mamazu:
I've converted the unit test to a kernel test and that's exactly what I want to test. I want to test that right now it removes the Meta Entity (currently there is some explicit code in the Collection manager for it):
```php
        foreach ($collectionEntity->getMeta() as $meta) {
            $this->em->remove($meta);
        }
```

And when this test is added we can savely remove this code and only rely on the database layer for consistency.

alexander-schranz:
Mostly this kind of code was added because we want to trigger a doctrine event to update external systems or search indexes on collection meta events. It would require some analyses when and why it was added.

mamazu:
That's the PR that added it and the comment on a similar kind of change sounds like a "better save than sorry approach". But I think adding a test would clear up this confusion. :D

https://github.com/sulu/sulu/pull/5939#discussion_r623840357

src/Sulu/Bundle/MediaBundle/Collection/Manager/CollectionManager.php
```diff
@@ -76,7 +76,7 @@ public function __construct(
         private MediaRepositoryInterface $mediaRepository,
         private FormatManagerInterface $formatManager,
         private UserRepositoryInterface $userRepository,
-        private EntityManager $em,
+        private EntityManagerInterface $em,
```
mamazu, line 79: Replacing the EntityManager implementation with the interface.

mamazu:
@alexander-schranz I think this should be good to go now the test actually makes sense now.

PR #8221, opened 2025-09-23
Use UNDERSCORE instead of DASH to maintain BC for locale strings

### LINKED ISSUE #8217: 3.0 - Switch to DASH from UNDERSCORE locale names breaks existing installs
| Q | A
| --- | ---
| Sulu Version | 3.0 latest
| PHP Version | 8.4
| DB Version | Postgres 18

I have a Sulu 2.6 app that is being upgraded to 3.0. 

https://github.com/sulu/sulu/pull/8173/files/3b36f8925f9bdc209bd62893cbebcce1d737e0ba#diff-4704286c93e96b0028f396e2400b8a160ba26bf6fc3e4455c111cfe0fd7b28d0R206

Whilst testing the 3.0 branch, DASH here seems to have broken my entire site - I switched it to UNDERSCORE and things started working again. I think this is a BC break as my DB is full of references to `en_gb` and not `en-gb` etc.

benr77:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #8217

#### What's in this PR?

Revert the use the previous format for locales stored in the database (en_gb) rather than the URL-friendly dashed version (en-gb).

#### Why?

Any database from a Sulu 2.6 install will use `en_gb` as the locale format.

This change in 3.0 breaks that and will require sweeping database migrations on huge sections of the DB data.

Therefore the simplest solution is to revert back to use UNDERSCORE as this will remove the need to carry out any extra database migrations when upgrading to 3.0.

### DIFF
--- a/packages/content/src/UserInterface/Controller/Website/ContentController.php
+++ b/packages/content/src/UserInterface/Controller/Website/ContentController.php
@@ -203,7 +203,7 @@ class ContentController extends AbstractController
             ?->getAllLocalizations() ?? [];
 
         foreach ($webspaceLocales as $webspaceLocale) {
-            $locale = $webspaceLocale->getLocale(Localization::DASH);
+            $locale = $webspaceLocale->getLocale(Localization::UNDERSCORE);
             if (\array_key_exists($locale, $localizations)) {
                 continue;
             }


Prokyonn (approved):

Prokyonn:
@benr77 Thank you!

PR #8224, opened 2025-09-24
Add search listener and search provider for accounts for SEAL

martinlagler:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Related issues/PRs | #7672
| License | MIT

#### What's in this PR?

Added ReindexProvider and IndexListener for Accounts for SEAL.

#### Why?

Because MassiveSearchBundle was removed

### DIFF
--- a/src/Sulu/Bundle/ContactBundle/Infrastructure/Sulu/Search/AccountReindexProvider.php
+++ b/src/Sulu/Bundle/ContactBundle/Infrastructure/Sulu/Search/AccountReindexProvider.php
@@ -1,0 +1,93 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of Sulu.
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\ContactBundle\Infrastructure\Sulu\Search;
+
+use CmsIg\Seal\Reindex\ReindexConfig;
+use CmsIg\Seal\Reindex\ReindexProviderInterface;
+use Doctrine\ORM\EntityManagerInterface;
+use Doctrine\ORM\EntityRepository;
+use Sulu\Bundle\ContactBundle\Entity\AccountInterface;
+
+/**
+ * @phpstan-type Account array{
+ *     id: int,
+ *     changed: \DateTimeImmutable,
+ *     created: \DateTimeImmutable,
+ *     name: string
+ * }
+ */
+class AccountReindexProvider implements ReindexProviderInterface
+{
+    /**
+     * @var EntityRepository<AccountInterface>
+     */
+    protected EntityRepository $accountRepository;
+
+    public function __construct(
+        EntityManagerInterface $entityManager,
+    ) {
+        $repository = $entityManager->getRepository(AccountInterface::class);
+
+        $this->accountRepository = $repository;
+    }
+
+    public function total(): ?int
+    {
+        return null;
+    }
+
+    public function provide(ReindexConfig $reindexConfig): \Generator
+    {
+        $accounts = $this->loadAccounts($reindexConfig->getIdentifiers());
+
+        /** @var Account $account */
+        foreach ($accounts as $account) {
+            yield [
+                'id' => AccountInterface::RESOURCE_KEY . '::' . ((string) $account['id']),
+                'resourceKey' => AccountInterface::RESOURCE_KEY,
+                'resourceId' => (string) $account['id'],
+                'changedAt' => $account['changed']->format('c'),
+                'createdAt' => $account['created']->format('c'),
+                'title' => $account['name'],
+            ];
+        }
+    }
+
+    /**
+     * @param string[] $identifiers
+     *
+     * @return iterable<Account>
+     */
+    private function loadAccounts(array $identifiers = []): iterable
+    {
+        $qb = $this->accountRepository->createQueryBuilder('account')
+            ->select('account.id')
+            ->addSelect('account.name')
+            ->addSelect('account.created')
+            ->addSelect('account.changed');
+
+        if (0 < \count($identifiers)) {
+            $qb->where('account.id IN (:ids)')
+                ->setParameter('ids', \array_map(fn ($identifier) => (int) \str_replace(AccountInterface::RESOURCE_KEY . '::', '', $identifier), $identifiers));
+        }
+
+        /** @var iterable<Account> */
+        return $qb->getQuery()->toIterable();
+    }
+
+    public static function getIndex(): string
+    {
+        return 'admin';
+    }
+}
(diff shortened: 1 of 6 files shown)


src/Sulu/Bundle/ContactBundle/Infrastructure/Sulu/Search/AccountReindexProvider.php
```diff
@@ -0,0 +1,81 @@
...
+
+namespace Sulu\Bundle\ContactBundle\Infrastructure\Sulu\Search;
+
+use CmsIg\Seal\Reindex\ReindexConfig;
+use CmsIg\Seal\Reindex\ReindexProviderInterface;
+use Doctrine\ORM\EntityManagerInterface;
+use Doctrine\ORM\EntityRepository;
+use Sulu\Bundle\ContactBundle\Entity\AccountInterface;
+
+class AccountReindexProvider implements ReindexProviderInterface
```
alexander-schranz, line 30:
```suggestion
/**
 * @internal This class is internal no backwards compatibility promise is given for this class
 *           use Symfony Dependency Injection to override or create your own ReindexProvider instead.
 */
final class AccountReindexProvider implements ReindexProviderInterface
```

src/Sulu/Bundle/ContactBundle/Infrastructure/Sulu/Search/AccountReindexProvider.php
```diff
@@ -0,0 +1,81 @@
...
+                'createdAt' => $account['created'] ? $account['created']->format('c') : null,
+                'title' => $account['name'],
+            ];
+        }
+    }
+
+    /**
+     * @param string[] $identifiers
+     */
+    private function loadAccounts(array $identifiers = []): \Generator
```
alexander-schranz: Define your return type:

```suggestion
     * @param string[] $identifiers
     *
     * @return \Generator<array{
     *     id: int,
     *     name: string,
     *     created: string,
     *     changed: string,
     * }>
     */
    private function loadAccounts(array $identifiers = []): \Generator
```

src/Sulu/Bundle/ContactBundle/Infrastructure/Sulu/Search/AccountReindexProvider.php
```diff
@@ -0,0 +1,81 @@
...
+            ->addSelect('account.name')
+            ->addSelect('account.created')
+            ->addSelect('account.changed');
+
+        if (0 < \count($identifiers)) {
+            $qb->where('account.id IN (:ids)')
+                ->setParameter('ids', \array_map(fn ($identifier) => (int) \str_replace(AccountInterface::RESOURCE_KEY . '::', '', $identifier), $identifiers));
+        }
+
+        return $qb->getQuery()->toIterable();
```
alexander-schranz, line 86: You can workaround the phpstan error this way:

```suggestion
        foreach ($qb->getQuery()->toIterable() as $row) {
            yield $row;
        }
```

alexander-schranz (changes requested):

alexander-schranz (approved):

src/Sulu/Bundle/ContactBundle/Tests/Functional/Infrastructure/Sulu/Search/AccountReindexProviderTest.php
```diff
@@ -0,0 +1,149 @@
...
+        $this->assertCount(2, $results);
+
+        $this->assertSame(AccountInterface::RESOURCE_KEY . '::' . $account1->getId(), $results[0]['id']);
+        $this->assertSame(AccountInterface::RESOURCE_KEY, $results[0]['resourceKey']);
+        $this->assertSame((string) $account1->getId(), $results[0]['resourceId']);
+        $this->assertSame('Test Account 1', $results[0]['title']);
+        $this->assertNotNull($results[0]['createdAt']);
+
+        $this->assertSame(AccountInterface::RESOURCE_KEY . '::' . $account2->getId(), $results[1]['id']);
+        $this->assertSame('Test Account 2', $results[1]['title']);
```
alexander-schranz: I would simplify the test:

Do not match every single property match against the whole results:

```suggestion
        $this->assertSame(
            [
                [
                     'id' => ...
                ],
            ],
            [...$result]
        );
```

src/Sulu/Bundle/ContactBundle/Tests/Functional/Infrastructure/Sulu/Search/AccountReindexProviderTest.php
```diff
@@ -0,0 +1,149 @@
...
+    {
+        $this->assertSame('admin', AccountReindexProvider::getIndex());
+    }
+
+    public function testTotal(): void
+    {
+        $this->assertNull($this->provider->total());
+    }
+
+    public function testProvideWithoutIdentifiers(): void
```
alexander-schranz:
```suggestion
    public function testProvideAll(): void
```

src/Sulu/Bundle/ContactBundle/Tests/Functional/Infrastructure/Sulu/Search/AccountReindexProviderTest.php
```diff
@@ -0,0 +1,149 @@
...
+
+        $this->assertCount(1, $results);
+    }
+
+    private function createAccount(string $name): Account
+    {
+        $account = new Account();
+        $account->setName($name);
+        $account->setCreated(new \DateTimeImmutable());
+        $account->setCreated(new \DateTimeImmutable());
```
alexander-schranz: The `setCreated` called twice should be `setChanged`.

Use two different dates which you match in the first test:

```suggestion
        $account->setCreated(new \DateTimeImmutable('2000-01-01 12:00:00'));
        $account->setChanged(new \DateTimeImmutable('2024-12-24 23:00:00'));
```

alexander-schranz (changes requested):

src/Sulu/Bundle/ContactBundle/Tests/Functional/Infrastructure/Sulu/Search/AccountReindexProviderTest.php
```diff
@@ -0,0 +1,149 @@
...
+        $this->entityManager->flush();
+
+        $config = ReindexConfig::create()
+            ->withIndex('admin')
+            ->withIdentifiers([]);
+
+        $results = \iterator_to_array($this->provider->provide($config));
+
+        $this->assertCount(1, $results);
+    }
```
alexander-schranz: Would skip this test empty = all. So think Sulu is not responsible to test that behaviour as it same as the first test.

We might should only test that the `ContactIndexListener` never dispatch a message without identifiers maybe 🤔

src/Sulu/Bundle/ContactBundle/Tests/Unit/Infrastructure/Sulu/Search/ContactIndexListenerTest.php
```diff
@@ -0,0 +1,93 @@
...
+
+    protected function setUp(): void
+    {
+        $this->messageBus = $this->prophesize(MessageBusInterface::class);
+        $this->listener = new ContactIndexListener($this->messageBus->reveal());
+    }
+
+    public function testOnAccountChangedWithAccountCreatedEvent(): void
+    {
+        $event = $this->prophesize(AccountCreatedEvent::class);
```
alexander-schranz: We should not mock this kind of objects.

alexander-schranz (changes requested):

PR #8230, opened 2025-09-25
Prepare extendability for SEO and Excerpt Content Tab

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Prepare extendability for SEO and Excerpt Content Tab. This is a draft introduce the concept.

#### Why?

In 2.6 the excerpt was easy to extend via XML we require same and should also make SEO tab extendable.

Adding new field to SEO and Excerpt should be as easy as creating just a new `content_excerpt.xml` with:

```xml
<?xml version="1.0" ?>
<form xmlns="http://schemas.sulu.io/template/template"
    xsi:schemaLocation="http://schemas.sulu.io/template/template http://schemas.sulu.io/template/form-1.0.xsd"
>
    <key>content_excerpt</key>

    <properties>
        <property name="excerpt/custom" type="text_line">
            <meta>
                <title>sulu_admin.custom</title>
            </meta>
        </property>
    </properties>
</form>
```

### DIFF
--- a/packages/content/config/forms/content_excerpt.xml
+++ b/packages/content/config/forms/content_excerpt.xml
@@ -6,43 +6,43 @@
     <key>content_excerpt</key>
 
     <properties>
-        <property name="excerptTitle" type="text_line">
+        <property name="excerpt/title" type="text_line">
             <meta>
                 <title>sulu_admin.title</title>
             </meta>
         </property>
 
-        <property name="excerptMore" type="text_line">
+        <property name="excerpt/more" type="text_line">
             <meta>
                 <title>sulu_page.text_for_more_link</title>
             </meta>
         </property>
 
-        <property name="excerptDescription" type="text_editor">
+        <property name="excerpt/description" type="text_editor">
             <meta>
                 <title>sulu_admin.description</title>
             </meta>
         </property>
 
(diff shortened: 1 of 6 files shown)


packages/content/src/Domain/Model/ExcerptInterface.php
alexander-schranz, line 1: fields which we do not query not require own column. specially als columns are limited in databases. so better to go with JSON fields for all which is not query able in seo or excerpt tab.

packages/content/config/forms/content_excerpt.xml
```diff
@@ -6,43 +6,43 @@
     <key>content_excerpt</key>
 
     <properties>
-        <property name="excerptTitle" type="text_line">
+        <property name="excerpt/title" type="text_line">
```
alexander-schranz, line 9: Maybe:

```
        <property name="ext/excerpt/title" type="text_line">
```

alexander-schranz:
replaced by #8382

PR #8232, opened 2025-09-26
Remove deprecated `AddMetadataPass`

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | yes
| Deprecations? | no
| Fixed tickets | fixes #8038 
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?
Removing deprecated MetadataProvider

#### Why?
🧹


mamazu:
Technically it would also make sense to remove the registry for only two types it doesn't make sense. It's only really used in the AdminController for the metadata Method and there we could just create an array of both services.

alexander-schranz (approved):

alexander-schranz:
Changes look fine but seems not to work yet?

mamazu:
Since this is not an interface it probably gets hung up on the implementation when trying to mock this object.

alexander-schranz:
you can remove the `readonly` you added currently to make the test pass, sure better would be instead of mock just create a instance of the registry, mocking in that case doesn't make a lot of sense. Less mocks less problems :D

mamazu:
As commented above. The tests would be easier if we can get rid of this entirely.

Prokyonn (approved):

Prokyonn:
@mamazu thank you :) 

as this is already working with the registry, I'll merge it like this. If you still want to remove the registry completely and try to inject the form/list providers explicitly, feel free to open a PR for this, so that it can be discussed further

PR #8236, opened 2025-09-26
Add block settings resolving

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Restores the settings in Sulu 3.0 as you knew it from 2.6
Additionally adds proper resolving of the settings form.

#### Why?

Missing feature in Sulu 3.0.

### DIFF
--- a/packages/page/config/forms/page_block_settings.xml
+++ b/packages/page/config/forms/page_block_settings.xml
@@ -1,0 +1,221 @@
+<?xml version="1.0" ?>
+<form xmlns="http://schemas.sulu.io/template/template"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://schemas.sulu.io/template/template http://schemas.sulu.io/template/form-1.0.xsd"
+>
+    <key>page_block_settings</key>
+
+    <properties>
+        <property name="hidden" type="checkbox">
+            <params>
+                <param name="description">
+                    <meta>
+                        <title>sulu_page.hide_block_description</title>
+                    </meta>
+                </param>
+                <param name="icon" value="su-hide" />
+                <param name="label">
+                    <meta>
+                        <title>sulu_page.hide_block_label</title>
+                    </meta>
+                </param>
+                <param name="skin" value="heading" />
+                <param name="type" value="toggler" />
+            </params>
+
+            <tag name="sulu.block_setting_icon" icon="su-hide" />
+        </property>
+
+        <section name="schedules">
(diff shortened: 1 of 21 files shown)


src/Sulu/Bundle/CoreBundle/Resources/config/request_analyzer.xml
```diff
@@ -58,7 +58,6 @@
         </service>
         <service id="sulu_core.request_processor.date_time"
                  class="Sulu\Component\Webspace\Analyzer\Attributes\DateTimeRequestProcessor">
-            <tag name="sulu.context" context="website"/>
```
Prokyonn, line 61: We probably have to also remove the context on other request analyzers, as we now resolve the content on "write" to collect the Reference-Entities

packages/page/config/forms/page_block_settings.xml
```diff
@@ -0,0 +1,221 @@
+<?xml version="1.0" ?>
+<form xmlns="http://schemas.sulu.io/template/template"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://schemas.sulu.io/template/template http://schemas.sulu.io/template/form-1.0.xsd"
+>
+    <key>page_block_settings</key>
```
alexander-schranz: would move this to the `content` bundle and name `content_block_settings` as the block visitors `hidden`, `schedules` also live there.

the segments part I would move to the `page` bundle and audience targeting part to the admin bundle.

theoretical it should be merged correctly if all bundles have the `content_block_settings.xml`.

alexander-schranz (approved):
Else looks good.

src/Sulu/Bundle/CoreBundle/Resources/config/request_analyzer.xml
alexander-schranz: 🤔  good point, but unsure if we should even filter out blocks during reference creation.

Prokyonn: Added it to the todo list

packages/page/config/forms/page_block_settings.xml
Prokyonn: Works with splitted files, but the order is not the same as we had before 🤔

PR #8240, opened 2025-09-29
Fix upgrade md and add a section for alpha/beta updates

Prokyonn:
| Q                  | A                |
  |--------------------|------------------|
  | Bug fix?           | yes              |
  | New feature?       | no               |
  | BC breaks?         | no               |
  | Deprecations?      | no               |
  | License            | MIT              |

  What's in this PR?

  This PR fixes schema inconsistencies in the Sulu 3.0 upgrade documentation and adds proper guidance for alpha/beta users. The changes include:

  1. Fixed CREATE TABLE statements: Updated all UUID columns from VARCHAR(255) to VARCHAR(36) to match actual Doctrine entity definitions
  2. Added alpha/beta upgrade section: Provides clear instructions for users upgrading from 3.0-alpha/beta releases who need schema optimization

  Why?

  Alpha/beta users were experiencing foreign key constraint errors when running doctrine:schema:update because:
  - Alpha/beta releases created tables with VARCHAR(255) for UUID columns
  - Final 3.0 entities expect VARCHAR(36) for UUID columns
  - MySQL prevents altering columns that participate in foreign key constraints
  - The existing upgrade documentation had inconsistent CREATE statements that didn't match the final entity definitions

  This caused the error: Cannot change column 'pageUuid': used in a foreign key constraint 'FK_209A42C0F099EEF3'


UPGRADE-3.x.md
```diff
@@ -1,5 +1,23 @@
 # Upgrade
 
+## Alpha/Beta Schema Update
```
Prokyonn, line 3: @sulu/core-developer WDYT should we add a section like this or just keep it documented in a new issue/discussion on github?

Prokyonn: As discussed, we will not merge this MR. The Upgrade file should not contain stuff for non final releases e.g. from alpha to beta. Otherwise we have to maintain this and it will get way too complicated.

PR #8242, opened 2025-09-30
Removing `MetadataProviderRegistry`

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | yes
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | https://github.com/sulu/sulu/pull/8038
| License | MIT

#### What's in this PR?
Removing the MetadataProviderRegistry

#### Why?
There is only one place where this is actually relevant and any places where it makes the code more complicated.


src/Sulu/Bundle/AdminBundle/Controller/AdminController.php
```diff
@@ -193,8 +193,7 @@ public function metadataAction(string $type, string $key, Request $request): Res
         $locale = $user->getLocale();
 
         $metadataOptions = $request->query->all();
-        $metadata = $this->metadataProviderRegistry->getMetadataProvider($type)
-            ->getMetadata($key, $locale, $metadataOptions);
+        $metadata = $this->metadataProviders->get($type)->getMetadata($key, $locale, $metadataOptions);
```
mamazu: This is the only usecase for keeping this class. But I think that the AdminController should be split up into multiple actions anyways, then we could just inject both providers and make a switch on them.

UPGRADE-3.x.md
```diff
@@ -504,6 +504,20 @@ now independent of the locale and contain all metadata for all locales this mean
...
+- `Sulu\Bundle\AdminBundle\Metadata\ListMetadata\ListMetadataProvider`
+            id="sulu_admin.list_metadata_provider"
+- `Sulu\Bundle\AdminBundle\Metadata\FormMetadata\FormMetadataProvider`
+            id="sulu_admin.form_metadata_provider"
+
+These are usually not used together nor are they extensible so the metadata registry has been removed. This also
+includes the following other classes:
+
+- WebspaceManager now takes FormMetadataProvider
+
```
mamazu, line 683: TODO: Update the docs when this feature gets accepted.

mamazu (commented):
@Prokyonn Not really merge ready but that would be my proposal.

alexander-schranz:
I don't think we will remove this service. It provides a single point to access the metadata which is from DX perspective better as seperate services. Even we break in future some metadata service they may not need to change the constructor which make upgrades via rector easier if just methods and params change.

mamazu:
I would disagree, for DX purposes there is definitely an overhead for testing. (See PR contents) Furthermore I would suggest if we keep it to replace the `get` function on the with `getList` and `getForm` because right now the code is sprinkled with the magic constants 'list' and 'form'.

Not to mention the possible performance impact this has when using in the property provider.

PR #8243, opened 2025-09-30
Replace `MediaType` entity with string column

### LINKED ISSUE #8223: Inlining the `MediaType` entity
# Current situation
In current version Sulu has a MediaType which is just a name and a description. There are the following usages:
* MediaTypeUpdateCommand
* MediaManager (updating the media to set the correct type)
* MediaTrashManager
* Tests that create this entity

Essentially this is a feature that is write only.

# Suggestion

I think there are two paths forward. If we want to keep the feature we could reduce it to a column with string type on the Media entity. 

## Motivation
* This would make queries simpler & faster (less joins)
* less setup for new projects (MediaTypes have their own fixtures)
* Drastically simplifying tests (since we don't need to set up those entities)
* Smaller database (we don't need to store a description of a type, we display nowhere)

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | yes
| Deprecations? | no
| Fixed tickets | closes #8223 
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?
Removing the `MediaType` entity and replacing it with the computed value from the TypeManager.

#### Todo
- [x] Write UPGRADE-3.md


mamazu:
The only problem I see is that trash data from the old version has a different shape than the new version. But one way of solving that would be if they try to restore an old config just return the TypeManager to redetermine the type again.

alexander-schranz (approved):

alexander-schranz (approved):

PR #8247, opened 2025-10-02
Custom urls routing

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no yes
| New feature? | no yes
| BC breaks? | no yes
| Deprecations? | no yes 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### To Do


packages/custom-url/src/Application/Routing/CustomUrlRouteCollectionLoader.php
```diff
@@ -0,0 +1,196 @@
...
+
+        return $this->handleContent($request, $customUrl, $collection);
+    }
+
+    private function buildUrlFromRequest(Request $request): string
+    {
+        $pathInfo = $request->getPathInfo();
+
+        // Remove format extension if present (e.g., .html, .json)
+        $position = \strrpos($pathInfo, '.');
```
mamazu: This would also break `/test.hallo/foo`

packages/custom-url/src/Domain/Model/CustomUrlRoute.php
```diff
@@ -48,4 +52,24 @@ public function getCustomUrl(): CustomUrlInterface
...
+
+    public function getTargetRoute(): ?CustomUrlRouteInterface
+    {
+        return $this->targetRoute;
+    }
+
+    public function setTargetRoute(?CustomUrlRouteInterface $targetRoute): void
+    {
+        $this->targetRoute = $targetRoute;
+    }
```
mamazu, line 74: Isn't this a property of the custom url already?

packages/custom-url/src/Infrastructure/Doctrine/EventListener/CustomUrlRouteHistoryListener.php
```diff
@@ -0,0 +1,137 @@
...
+ * Marks old routes as history and links them to new routes.
+ */
+class CustomUrlRouteHistoryListener
+{
+    /**
+     * @var array<string, array{oldPath: string, newPath: string, customUrl: CustomUrl}>
+     */
+    private array $pendingHistoryUpdates = [];
+
+    public function preUpdate(PreUpdateEventArgs $args): void
```
mamazu: This only needs to happen on CustomUrl update? Can't we use the domain event for that? Ohterwise this will run for any entity update which seems a little excessive.

packages/custom-url/src/Domain/Model/CustomUrlRoute.php
wachterjohannes: no we have to know which is the followup route to have the chain which is the newest route

packages/custom-url/src/Infrastructure/Doctrine/EventListener/CustomUrlRouteHistoryListener.php
wachterjohannes: i am currently refactoring this :)

alexander-schranz: the listener does nothing if no customurl is changed and it should do nothing if no url was changed.

You want todo the history url creation in the same transaction and not in an own. Also you want to trigger history url creation also when somebody directly modify the entity not using the our provided messages. Its to critical todo it on the domain event.

packages/custom-url/src/Infrastructure/Doctrine/EventListener/CustomUrlRouteHistoryListener.php
```diff
@@ -0,0 +1,137 @@
...
+use Doctrine\ORM\Event\PostFlushEventArgs;
+use Doctrine\ORM\Event\PreUpdateEventArgs;
+use Sulu\CustomUrl\Domain\Model\CustomUrl;
+use Sulu\CustomUrl\Domain\Model\CustomUrlRoute;
+
+/**
+ * Manages history routes when custom URL paths change.
+ * Marks old routes as history and links them to new routes.
+ */
+class CustomUrlRouteHistoryListener
```
alexander-schranz:
```suggestion
/**
 * @internal No BC promises are given for this class. It may be changed or removed at any time.
 *
 * Manages history routes when custom URL paths change.
 * Marks old routes as history and links them to new routes.
 */
class CustomUrlRouteHistoryListener
```

packages/custom-url/src/Infrastructure/Doctrine/EventListener/CustomUrlRouteHistoryListener.php
```diff
@@ -0,0 +1,137 @@
...
+{
+    /**
+     * @var array<string, array{oldPath: string, newPath: string, customUrl: CustomUrl}>
+     */
+    private array $pendingHistoryUpdates = [];
```
alexander-schranz: service requires reset interface

alexander-schranz: requires also onclear listener

alexander-schranz (changes requested):

packages/custom-url/src/Infrastructure/Doctrine/EventListener/CustomUrlRouteHistoryListener.php
wachterjohannes: have refactored this into the modify listener as this does not make sense to handle that in a doctrine listener

alexander-schranz: it does make sense if custom url get modified directly to still create route histories.

alexander-schranz (approved):

packages/custom-url/config/doctrine/CustomUrl/CustomUrlRoute.orm.xml
```diff
@@ -9,13 +9,23 @@
...
          </many-to-one>
 
+        <many-to-one field="targetRoute" target-entity="Sulu\CustomUrl\Domain\Model\CustomUrlRouteInterface">
+            <join-column name="target_route_uuid" referenced-column-name="uuid" nullable="true" on-delete="set null" />
+        </many-to-one>
+
         <indexes>
-            <index name="routes" columns="path" />
+            <index name="idx_route_path" columns="path" />
+            <index name="idx_route_history" columns="history" />
```
alexander-schranz: Index name need to be unique over the DB and shoould so have table name as a prefix.

```suggestion
            <index name="custom_url_route_path_idx" fields="path" />
            <index name="custom_url_route_history_idx" fields="history" />
```

similar to: https://github.com/sulu/sulu/blob/4ef044f0223ee0c814f295f87fb6e7ff5076a80c/packages/route/config/doctrine/Route/Route.orm.xml#L14

packages/custom-url/config/doctrine/CustomUrl/CustomUrlRoute.orm.xml
```diff
@@ -9,13 +9,23 @@
...
+                <option name="default">0</option>
+            </options>
+        </field>
 
         <many-to-one field="customUrl" target-entity="Sulu\CustomUrl\Domain\Model\CustomUrlInterface" inversed-by="routes">
             <join-column name="customUrl" referenced-column-name="uuid" nullable="false" on-delete="cascade" />
          </many-to-one>
 
+        <many-to-one field="targetRoute" target-entity="Sulu\CustomUrl\Domain\Model\CustomUrlRouteInterface">
+            <join-column name="target_route_uuid" referenced-column-name="uuid" nullable="true" on-delete="set null" />
```
alexander-schranz: why not `CASCADE` here if the target is removed?

alexander-schranz (changes requested):

packages/custom-url/config/doctrine/CustomUrl/CustomUrlRoute.orm.xml
wachterjohannes: absolutly right :)

PR #8258, opened 2025-10-13
Improve performance of `MediaTypeUpdateCommand`

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?
Replacing the heavy entity loading in the UpdateMediaTypeCommand with a lighter array that updates a lot of elements at once.

#### Why?
When you have a lot of media this command can take a long time to run. From testing on a project with 9k media entries and 15k file versions the time the update took went from 5 seconds to under a second.


mamazu:
This will require some kind of backwards compatibility break if we want to move it to the repository. Otherwise since the command is the only place where we use it, we might as well move it to there.

alexander-schranz (approved):

PR #8262, opened 2025-10-16
Add implementation for block ids

### LINKED ISSUE #7846: Add Block IDs for later reference
# Problem description

Inside content management we want reference block via a id.

Requirement is:

 - they need to be unique for the current page
 - we can use them as reference for anchor navigation
 - we can use them as reference for future planned shadow structure feature

# Proposed solution

Maybe a short hash algorithm can be used if it not require general uniquness:

```php
php -r 'echo hash("xxh32", time());'
```

outputs:

```php
a561af03
```

A PHP endpoint can also be used to generate such ids:

We can discus it but for me using someones JS library for create hashes or uuids which we don't know well I would avoid. As such libraries are preferred target of supply chain attacks.

Technically id generator need be enabled via a new property like we do for block settings also:

```xml
<param name="block_id_generator" value="true" /> 
```

We can also go with `value="xxh32"` later if we want support uuid based generators which are longer and really unique over all.

For pages, articles, snippet that param is automatically set like we do for the settings param, reference: https://github.com/sulu/sulu/blob/f4ae26ada81d882d1505ab5476c2b9fe087b6437/src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/StructureFormMetadataLoader.php#L193-L198

The submitted block should look like this:

```php
[
     {
           "_id": "<generator_value>",
           "title": "Other field"
     },
     {
           "_id": "<generator_value>",
           "title": "Other field"
     },
]
```

We can NOT! set the id as key JSON Specification don't forces keeping of the order of object and databases return object keys different order as they where given to it. And blocks are kind of list / array and so the array need to stay and is a less bc break.

Even we should "migrate" and generate the key in the migration from 2.6 to 3.0: https://github.com/sulu/SuluPHPCRMigrationBundle the frontend should still generate a new id if no id is give.

# Complications

 - [ ] The new `_id` need to be excluded from all copy and paste functionality

# Open Questions

 - [ ] A. Is uniqueness on object level enough (shorter id could be used)?
 - [ ] B. Is a PHP side generation of the id a accept solution (avoid JS dependencies)

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #7846
| Related issues/PRs | none
| License | MIT
| Documentation PR | none

#### What's in this PR?

This PR add the new block ids.

#### Why?

This block ids can be used later for different features.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Controller/BlockIdController.php
+++ b/src/Sulu/Bundle/AdminBundle/Controller/BlockIdController.php
@@ -33,14 +33,7 @@ class BlockIdController
     {
         $length = $request->query->getInt('length', 1);
 
-        // Generate single ID for backwards compatibility
-        if (1 === $length) {
-            $id = $this->blockIdGenerator->generateId();
-
-            return new JsonResponse(['id' => $id]);
-        }
-
-        // Generate multiple IDs for batch operations
+        // Generate IDs and always return array format
         $blockIds = [];
         for ($i = 0; $i < $length; ++$i) {
             $blockIds[] = ['id' => $this->blockIdGenerator->generateId()];
--- a/src/Sulu/Bundle/AdminBundle/Controller/BlockIdController.php
+++ b/src/Sulu/Bundle/AdminBundle/Controller/BlockIdController.php
@@ -15,6 +15,7 @@ namespace Sulu\Bundle\AdminBundle\Controller;
 
 use Sulu\Bundle\AdminBundle\Application\BlockIdGenerator\BlockIdGeneratorInterface;
 use Symfony\Component\HttpFoundation\JsonResponse;
+use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 
 /**
@@ -28,10 +29,23 @@ class BlockIdController
     ) {
     }
 
-    public function generateAction(): Response
+    public function generateAction(Request $request): Response
     {
-        $id = $this->blockIdGenerator->generateId();
+        $length = $request->query->getInt('length', 1);
 
-        return new JsonResponse(['id' => $id]);
+        // Generate single ID for backwards compatibility
+        if (1 === $length) {
+            $id = $this->blockIdGenerator->generateId();
+
+            return new JsonResponse(['id' => $id]);
+        }
+
+        // Generate multiple IDs for batch operations
+        $blockIds = [];
+        for ($i = 0; $i < $length; ++$i) {
+            $blockIds[] = ['id' => $this->blockIdGenerator->generateId()];
+        }
+
+        return new JsonResponse(['_embedded' => ['blockIds' => $blockIds]]);
     }
 }
--- a/src/Sulu/Bundle/AdminBundle/Controller/BlockIdController.php
+++ b/src/Sulu/Bundle/AdminBundle/Controller/BlockIdController.php
@@ -1,0 +1,37 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of Sulu.
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\AdminBundle\Controller;
+
+use Sulu\Bundle\AdminBundle\Application\BlockIdGenerator\BlockIdGeneratorInterface;
+use Symfony\Component\HttpFoundation\JsonResponse;
+use Symfony\Component\HttpFoundation\Response;
+
+/**
+ * Controller for generating unique block IDs.
+ * Used by the frontend to generate IDs when creating or duplicating blocks.
+ */
+class BlockIdController
+{
+    public function __construct(
+        private readonly BlockIdGeneratorInterface $blockIdGenerator,
+    ) {
+    }
+
+    public function generateAction(): Response
+    {
+        $id = $this->blockIdGenerator->generateId();
+
+        return new JsonResponse(['id' => $id]);
+    }
+}
--- a/src/Sulu/Bundle/AdminBundle/Resources/config/routing_api.yaml
+++ b/src/Sulu/Bundle/AdminBundle/Resources/config/routing_api.yaml
@@ -49,3 +49,5 @@ sulu_admin.post_block_ids:
     format: json
     requirements:
         _format: json|csv
+    options:
+        expose: true
--- a/src/Sulu/Bundle/AdminBundle/Resources/config/routing_api.yaml
+++ b/src/Sulu/Bundle/AdminBundle/Resources/config/routing_api.yaml
@@ -40,3 +40,12 @@ sulu_admin.get_items:
     format: json
     requirements:
         _format: json|csv
+
+# Block IDs
+sulu_admin.post_block_ids:
+    path: '/block-ids.{_format}'
+    methods: GET
+    controller: 'sulu_admin.block_id_controller::generateAction'
+    format: json
+    requirements:
+        _format: json|csv
--- a/src/Sulu/Bundle/AdminBundle/Resources/config/routing_api.yaml
+++ b/src/Sulu/Bundle/AdminBundle/Resources/config/routing_api.yaml
@@ -44,7 +44,7 @@ sulu_admin.get_items:
 # Block IDs
 sulu_admin.post_block_ids:
     path: '/block-ids.{_format}'
-    methods: GET
+    methods: POST
     controller: 'sulu_admin.block_id_controller::generateAction'
     format: json
     requirements:
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/BlockCollection.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/BlockCollection.js
@@ -18,6 +18,7 @@ type Props<T: string, U: {_id?: string, type: T, ...}> = {|
     defaultType: T,
     disabled: boolean,
     generateBlockId?: () => Promise<string>,
+    generateBlockIds?: (count: number) => Promise<Array<string>>,
     icons?: Array<Array<string>>,
     maxOccurs?: ?number,
     minOccurs?: ?number,
@@ -158,7 +159,7 @@ class BlockCollection<T: string, U: {_id?: string, type: T, ...}> extends React.
     };
 
     @action handlePasteBlocks = async(insertionIndex: number) => {
-        const {generateBlockId, onChange, onDisplaySnackbar, value} = this.props;
+        const {generateBlockId, generateBlockIds, onChange, onDisplaySnackbar, value} = this.props;
 
         if (this.hasMaximumReached) {
             throw new Error('The maximum amount of blocks has already been reached!');
@@ -178,6 +179,16 @@ class BlockCollection<T: string, U: {_id?: string, type: T, ...}> extends React.
             insertionIndex, 0, ...this.pasteableBlocks.map(() => ++BlockCollection.idCounter)
         );
 
+        // Count how many blocks need new IDs (blocks without _id are from copy operations)
+        const blocksNeedingIds = this.pasteableBlocks.filter((block) => !block._id).length;
+
+        // Generate IDs in batch if we have the batch function and multiple blocks need IDs
+        let generatedIds = [];
+        if (generateBlockIds && blocksNeedingIds > 1) {
+            generatedIds = await generateBlockIds(blocksNeedingIds);
+        }
+
+        let idIndex = 0;
         const newElements = await Promise.all(
             this.pasteableBlocks.map(async(block) => {
                 // paste block with default type if type of block in clipboard is not known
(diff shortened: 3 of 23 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/services/blockIdGenerator/blockIdGenerator.js
alexander-schranz, line 1: services which have only a single instance/create there instance in the export (export default `new` ...) are using lowercase filenames to indicate you get a object not a class:

`src/Sulu/Bundle/AdminBundle/Resources/js/services/blockIdGenerator/blockIdGenerator.js`

More JS like is exporting function instead of classes.

alexander-schranz (changes requested):
didn't do a review yet just a hint here:

src/Sulu/Bundle/AdminBundle/Resources/config/routing_api.yaml
```diff
@@ -40,3 +40,14 @@ sulu_admin.get_items:
     format: json
     requirements:
         _format: json|csv
+
+# Block IDs
+sulu_admin.post_block_ids:
+    path: '/block-ids.{_format}'
+    methods: GET
```
alexander-schranz: We are generating something, so this should not be a GET as a GET is Idempotent. Only `POST` and `PATCH` requests should be used for generating things.

```suggestion
    methods: POST
```

src/Sulu/Bundle/AdminBundle/Controller/BlockIdController.php
```diff
@@ -0,0 +1,37 @@
...
+    public function __construct(
+        private readonly BlockIdGeneratorInterface $blockIdGenerator,
+    ) {
+    }
+
+    public function generateAction(): Response
+    {
+        $id = $this->blockIdGenerator->generateId();
+
+        return new JsonResponse(['id' => $id]);
```
alexander-schranz: In case of copy blocks we do now a request per block? Should we keep in mind todo a single request here if possible with `?length=...`

```suggestion
        $blockIds = [];
        $request->query->getInt('length', 1)
        
        for ($i = 0; $i<=$length;++$i) {
            $blockIds = ['id' => $this->blockIdGenerator->generateId()]; // maybe already return it in this format make it easier to merge it together locally
        }

        return new JsonResponse(['_embedded' => ['blockIds' => $blockIds]); // not sure about this in a post Action format
```

If you think we should keep it as a single id endpoint I'm fine with it.

src/Sulu/Bundle/AdminBundle/Resources/js/services/blockIdGenerator/blockIdGenerator.js
```diff
@@ -0,0 +1,23 @@
...
+import Requester from '../Requester';
+
+/**
+ * Generate a unique block ID from the backend API.
+ *
+ * @returns {Promise<string>} The generated block ID
+ */
+function generateBlockId(): Promise<string> {
+    const url = symfonyRouting.generate('sulu_admin.post_block_ids');
+    return Requester.get(url)
```
alexander-schranz:
```suggestion
    return Requester.post(url)
```

alexander-schranz (changes requested):

src/Sulu/Bundle/AdminBundle/Controller/BlockIdController.php
```diff
@@ -28,10 +29,23 @@ public function __construct(
...
 
-    public function generateAction(): Response
+    public function generateAction(Request $request): Response
     {
-        $id = $this->blockIdGenerator->generateId();
+        $length = $request->query->getInt('length', 1);
 
-        return new JsonResponse(['id' => $id]);
+        // Generate single ID for backwards compatibility
+        if (1 === $length) {
```
alexander-schranz: Would not do that, it would make the API just unnecessary complex, to have 2 different format depend on query string.

Would always return a list.

alexander-schranz (changes requested):

wachterjohannes:
@alexander-schranz done

src/Sulu/Bundle/AdminBundle/Controller/BlockIdController.php
wachterjohannes: you are absolutly right :)

src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/BlockCollection.js
```diff
@@ -128,8 +129,8 @@ class BlockCollection<T: string, U: {type: T}> extends React.Component<Props<T,
         return indexes;
     }
 
-    @action handleAddBlock = (insertionIndex: number) => {
-        const {defaultType, onChange, value} = this.props;
+    @action handleAddBlock = async(insertionIndex: number) => {
```
alexander-schranz, line 158: isnt that a keyword? then should be a space here not?

```suggestion
    @action handleAddBlock = async (insertionIndex: number) => {
```

src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/BlockCollection.js
```diff
@@ -142,13 +143,23 @@ class BlockCollection<T: string, U: {type: T}> extends React.Component<Props<T,
 
             const elementsBefore = value.slice(0, insertionIndex);
             const elementsAfter = value.slice(insertionIndex);
+
+            // $FlowFixMe: Dynamic property assignment
+            const newBlock: any = {type: defaultType};
```
alexander-schranz: we should not use `any` type it or use `Object`.

src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/BlockCollection.js
```diff
@@ -142,13 +143,23 @@ class BlockCollection<T: string, U: {type: T}> extends React.Component<Props<T,
...
+
+            // Generate block ID if function is provided
```
alexander-schranz: AI comment? 

```suggestion
```

src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/BlockCollection.js
```diff
@@ -142,13 +143,23 @@ class BlockCollection<T: string, U: {type: T}> extends React.Component<Props<T,
...
+            if (generateBlockIds) {
+                const ids = await generateBlockIds(1);
+                newBlock._id = ids[0];
+            }
+
             // $FlowFixMe
-            onChange([...elementsBefore, {type: defaultType}, ...elementsAfter]);
+            onChange([...elementsBefore, newBlock, ...elementsAfter]);
```
alexander-schranz: Would avoid the `ids[0]` access go with soemthing like:

```ts
const newBlocks: Object[] = [{type: defaultType}];
if (generateBlockIds) {
    const ids = await generateBlockIds(1);
    ids.forEach((blockId, key) => {
        newBlocks[key] = {...blockId, ...newBlocks[key]};
    });
}

onChange([...elementsBefore, ...newBlocks, ...elementsAfter]);
```

src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/BlockCollection.js
```diff
@@ -301,7 +343,14 @@ class BlockCollection<T: string, U: {type: T}> extends React.Component<Props<T,
         const blocks = [];
 
         indexes.forEach(( index) => {
-            blocks.push({...toJS(value[index])});
+            const block = {...toJS(value[index])};
+
+            // Remove _id from copied blocks so new IDs are generated on paste
+            if (generateBlockIds && block._id) {
+                delete block._id;
+            }
```
alexander-schranz: what about nested blocks 🤔 ?

src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/BlockCollection.js
```diff
@@ -324,9 +373,22 @@ class BlockCollection<T: string, U: {type: T}> extends React.Component<Props<T,
...
+        const blocks = [];
+
+        indexes.forEach(( index) => {
+            const block = {...toJS(value[index])};
+
+            // Preserve _id for cut blocks (don't delete it like in copy)
+            blocks.push(block);
+        });
 
-        this.copyBlocks(indexes, false);
```
alexander-schranz, line 329: don't understand why we need to change this? why we can not `copyBlocks`.

src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/BlockCollection.js
```diff
@@ -457,7 +519,10 @@ class BlockCollection<T: string, U: {type: T}> extends React.Component<Props<T,
                     className={blockCollectionStyles.addButton}
                     disabled={disabled || this.hasMaximumReached}
                     icon="su-plus"
-                    onClick={this.handleAddBlock}
+                    // eslint-disable-next-line react/jsx-no-bind
+                    onClick={(value) => {
+                        this.handleAddBlock(value);
+                    }}
```
alexander-schranz: we should instead define the method this binding correctly instead of eslint ignore here something

src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/BlockCollection.js
```diff
@@ -468,7 +533,10 @@ class BlockCollection<T: string, U: {type: T}> extends React.Component<Props<T,
                         className={blockCollectionStyles.addButton}
                         disabled={disabled || this.hasMaximumReached}
                         icon="su-copy"
-                        onClick={this.handlePasteBlocks}
+                        // eslint-disable-next-line react/jsx-no-bind
+                        onClick={(value) => {
+                            this.handlePasteBlocks(value);
+                        }}
```
alexander-schranz: we should instead define the method this binding correctly instead of eslint ignore here something

src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/BlockCollection.js
```diff
@@ -269,8 +304,15 @@ class BlockCollection<T: string, U: {type: T}> extends React.Component<Props<T,
...
-            newValue = [...elementsBefore, {...toJS(newValue[index])}, ...elementsAfter];
-        });
+            const duplicatedBlock = {...toJS(newValue[index])};
+
+            // Use pre-generated ID
+            if (generateBlockIds && generatedIds.length > count) {
+                duplicatedBlock._id = generatedIds[count];
+            }
+
+            newValue = [...elementsBefore, duplicatedBlock, ...elementsAfter];
```
alexander-schranz: what about nested blocks? should we first search for all appeareance of `_id` in nested object / arrays?

src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/BlockCollection.js
```diff
@@ -168,14 +179,31 @@ class BlockCollection<T: string, U: {type: T}> extends React.Component<Props<T,
...
+        // Count how many blocks need new IDs (blocks without _id are from copy operations)
+        const blocksNeedingIds = this.pasteableBlocks.filter((block) => !block._id).length;
+
+        // Generate IDs in batch if needed
+        let generatedIds = [];
+        if (generateBlockIds && blocksNeedingIds > 0) {
+            generatedIds = await generateBlockIds(blocksNeedingIds);
+        }
+
+        let idIndex = 0;
```
alexander-schranz: not required map provides you an index:

```js
const newElements = this.pasteableBlocks.map((block, blockIndex) => {
```

src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/BlockCollection.js
```diff
@@ -168,14 +179,31 @@ class BlockCollection<T: string, U: {type: T}> extends React.Component<Props<T,
...
             // paste block with default type if type of block in clipboard is not known
-            if (!this.props.types?.[block.type]) {
-                return {...block, type: this.props.defaultType};
+            const newBlock = !this.props.types?.[block.type]
+                ? {...block, type: this.props.defaultType}
+                : {...block};
+
+            // Generate new ID only if block doesn't have one (copy operation)
+            // Cut operation preserves the _id
+            if (generateBlockIds && !newBlock._id) {
```
alexander-schranz: paste always should replace it. I even not only handle thing sin the `paste` part not in copy, cut part?

alexander-schranz (changes requested):

alexander-schranz (approved):

alexander-schranz:
@wachterjohannes can you check the merge conflict.

Want to merge it for the next release, still we should have look at the comments.

src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/BlockCollection.js
wachterjohannes: the current version allow to keep the ID for the cut and replace it for paste - as cut would keep it unique - but i could change that

wachterjohannes:
@alexander-schranz i created a follow UP pr with the pregeneration of block-ids https://github.com/sulu/sulu/pull/8288

PR #7570, opened 2024-08-23
Custom urls without phpcr (Admin Panel)

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | closes #7566 
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?
Replacing the custom url implementation of phpcr with some doctrine entities.

#### Why?
phpcr should be removed in the next major version and we can really simplify the logic here.

#### Todo

- [x] Upgrade SQL statements

#### Admin (Prio 1)
- [x] Path cleanup (eg. * will be replaced with -)
- [x] Don't add url to history if it's already in there
- [x] Fix serialization
- [x] Performance Tests
- [x] Fix and write new tests

PR #8263, opened 2025-10-16
Fix phpstan errors with custom urls

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | -
| Fixed tickets | -
| Related issues/PRs | #7570
| License | MIT

#### What's in this PR?
Fixing type errors from custom url admin PR.

#### Why?
We want the code to work.

### DIFF
--- a/packages/custom-url/src/Infrastructure/Sulu/Trash/CustomUrlTrashItemHandler.php
+++ b/packages/custom-url/src/Infrastructure/Sulu/Trash/CustomUrlTrashItemHandler.php
@@ -32,19 +32,22 @@ use Webmozart\Assert\Assert;
 
 /** @phpstan-type TrashRestoreData array{
  *     title: string,
- *     creator: UserInterface,
- *     created: string,
- *     changer: string,
- *     changed: string,
+ *     published: bool,
  *     baseDomain: string,
  *     domainParts: array<string>,
- *     canonical: bool,
+ *     targetLocale: string,
+ *     targetDocument: null|string,
  *     redirect: bool,
+ *     canonical: bool,
  *     noFollow: bool,
  *     noIndex: bool,
  *     targetUuid: string,
  *     targetLocale: string,
  *     webspaceKey: string,
+ *     creator: null|string,
+ *     created: string,
+ *     changer: null|string,
+ *     changed: string,
  * }
  */
 final class CustomUrlTrashItemHandler implements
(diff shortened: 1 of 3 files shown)


packages/custom-url/src/Infrastructure/Sulu/Trash/CustomUrlTrashItemHandler.php
```diff
@@ -68,6 +71,7 @@ public function store(object $customUrl, array $options = []): TrashItemInterfac
     {
         Assert::isInstanceOf($customUrl, CustomUrlInterface::class);
 
+        /** @var TrashRestoreData $data */
         $data = [
```
mamazu, line 75: Just a nice to know thing: If you typehint a literal array like that you can very easily check that the definition is correct, because it will complain if the type does not match the data.

packages/custom-url/src/Infrastructure/Sulu/Trash/CustomUrlTrashItemHandler.php
```diff
@@ -112,12 +116,12 @@ public function restore(TrashItemInterface $trashItem, array $restoreFormData =
             $customUrlMapper->mapCustomUrlData($customUrl, $data);
         }
 
-        if (isset($data['creator']) && null !== $data['creator']) {
```
mamazu, line 115: Isset already includes a null check. In my opinion an `=== null` would be enough here but isset makes sure that this exists.

packages/custom-url/tests/Unit/Infrastructure/Sulu/Trash/CustomUrlTrashItemHandlerTest.php
```diff
@@ -78,7 +78,6 @@ public function testStoreAndRestore(): void
         static::assertSame($originalCustomUrl->getUuid(), $restoredCustomUrl->getUuid());
         static::assertSame('sulu_io', $restoredCustomUrl->getWebspace());
         static::assertSame('2025-04-20T00:00:00+00:00', $restoredCustomUrl->getCreated()->format('c'));
-        static::assertNotNull($restoredCustomUrl->getChanged());
```
mamazu, line 81: This can not be null, either it's a Datetime or it crashes inside the method

alexander-schranz (approved):

alexander-schranz:
@mamazu Thank you!

PR #8265, opened 2025-10-16
Replace resource_locator by route field type by replace modeResolver with a mode schemaOptions

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | yes (remove old `resource_locator` in favor of `route` and some more code changes)
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs |  #7672, requires: #8264,
| License | MIT

#### What's in this PR?

Add handling of route mode via metadata visitor instead on the javascript side.

#### Why?

Keep the webspace config also in Sulu 3.0 in mind where always `route` is used.

This reduces the requirement of JavaScript based mode resolver and handle the mode via a metadata visitor.

```diff
diff --git a/packages/route/assets/js/index.js b/packages/route/assets/js/index.js
index cb78e4f8f9..f8b19a8d59 100644
--- a/packages/route/assets/js/index.js
+++ b/packages/route/assets/js/index.js
@@ -2,7 +2,6 @@
 import {fieldRegistry} from 'sulu-admin-bundle/containers';
 import initializer from 'sulu-admin-bundle/services/initializer';
 import {ResourceLocator} from './containers';
-import type {FieldTypeProps} from 'sulu-admin-bundle/containers/Form/types';

 initializer.addUpdateConfigHook('sulu_route', (config: Object, initialized: boolean) => {
     if (initialized) {
@@ -14,17 +13,7 @@ initializer.addUpdateConfigHook('sulu_route', (config: Object, initialized: bool
         ResourceLocator,
         {
             historyResourceKey: 'routes',
-            modeResolver: (props: FieldTypeProps<?string>) => {
-                const {
-                    schemaOptions: {
-                        mode: {
-                            value: mode = 'full',
-                        } = {},
-                    },
-                } = props;
-
-                return Promise.resolve(mode);
-            },
+            defaultMode: 'tree_leaf_edit',
             resourceStorePropertiesToRequest: { // maybe move to schemaOptions and prepend via MetadataListener
                 parentUuid: 'parentId',
                 parentId: 'parentId',
```

A metadata visitor sets a new param to define the mode correctly:

```diff
        <property name="url" type="route" mandatory="true">
            <meta>
                <title lang="de">Adresse</title>
                <title lang="en">Resourcelocator</title>
            </meta>

+            <!-- No changes for enduser: the following is set by a metadata visitor via the webspace config:
+            <params>
+                 <param name="mode" value="tree_leaf_edit" />
+            </params>
+            -->

            <tag name="sulu.rlp"/>
        </property>
```

### DIFF
--- a/UPGRADE-3.x.md
+++ b/UPGRADE-3.x.md
@@ -585,7 +585,23 @@ renamed to `url` in your templates and use the `route` always.
                 <title lang="de">Adresse</title>
             </meta>
 
-            <tag name="sulu_article.article_route"/>
+-            <tag name="sulu_article.article_route"/>
++            <tag name="sulu.rlp"/>
+        </property>
+```
+
+To make things easier between pages and articles the `resource_locator` field type is now replaced by the `route` 
+field type, so all routable entities use the same field type.:
+
+```diff
+-        <property name="url" type="resource_locator">
++        <property name="url" type="route">
+            <meta>
+                <title lang="en">Resourcelocator</title>
+                <title lang="de">Adresse</title>
+            </meta>
+            
+            <tag name="sulu.rlp"/>
         </property>
 ```
 
--- a/UPGRADE-3.x.md
+++ b/UPGRADE-3.x.md
@@ -605,6 +605,9 @@ field type, so all routable entities use the same field type.:
         </property>
 ```
 
+The `ResourceLocator` JS React Component now uses `tree_leaf_edit` and `tree_full_edit` instead of `leaf` and `full`,
+to be consistent with the PHP naming.
+
 ### Upgrade the Controller references
 
 The controller in the page/article templates have to be adjusted use the new controller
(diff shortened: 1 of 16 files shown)


UPGRADE-3.x.md
```diff
@@ -585,10 +585,29 @@ renamed to `url` in your templates and use the `route` always.
...
++            <tag name="sulu.rlp"/>
         </property>
 ```
 
+To make things easier between pages and articles the `resource_locator` field type is now replaced by the `route` 
+field type, so all routable entities use the same field type.:
+
+```diff
+-        <property name="url" type="resource_locator">
++        <property name="url" type="route">
```
chirimoya, line 598: Should we keep the property name `url`? I think `slug` would be more accurate.

But we can rediscuss it. PS: It would require some refactoring in the route mapping as that requires currently `url` be set.

chirimoya (approved):

UPGRADE-3.x.md
alexander-schranz: For `name` I would go with `url` as it make more sense todo `<a href="{{ page.url }}>` and not `slug`. As it was planned that we sulu_content_path during content resolving and not longer require it in twig to be called.

PR #8267, opened 2025-10-17
Improve typing for `DoctrineFieldDescriptor`s

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?
Improving typehints for the `DoctrineListBuilderFactory` and updating all related types.

#### Why?
We want to reduce the size of the phpstan config.

:warning: This also adds two new entries to the phpstan config, but they are inside of deprecated code and should vanish when upmerging.

### DIFF
--- a/src/Sulu/Component/Rest/ListBuilder/Doctrine/DoctrineListBuilderFactoryInterface.php
+++ b/src/Sulu/Component/Rest/ListBuilder/Doctrine/DoctrineListBuilderFactoryInterface.php
@@ -11,17 +11,17 @@
 
 namespace Sulu\Component\Rest\ListBuilder\Doctrine;
 
-use Sulu\Component\Rest\ListBuilder\ListBuilderInterface;
-
 /**
- * Defines the interface for the Factory of the DoctrineListBuilde.
+ * Defines the interface for the Factory of the DoctrineListBuilder.
  */
 interface DoctrineListBuilderFactoryInterface
 {
     /**
-     * @param string $entityName
+     * Creates a new DoctrineListBuilder for the given entity name and returns it.
      *
-     * @return ListBuilderInterface
+     * @param class-string $entityName
+     *
+     * @return DoctrineListBuilder
      */
     public function create($entityName);
 }
--- a/src/Sulu/Component/Rest/ListBuilder/ListBuilderInterface.php
+++ b/src/Sulu/Component/Rest/ListBuilder/ListBuilderInterface.php
@@ -206,7 +206,7 @@ interface ListBuilderInterface
      * Defines a constraint for the rows to return.
      *
      * @param FieldDescriptorInterface $fieldDescriptor The FieldDescriptor which is checked
-     * @param string $value The value the FieldDescriptor should have
+     * @param int|string|float|bool|null $value The value the FieldDescriptor should have
      * @param string $comparator The comparator use to compare the values
      *
      * @return ListBuilderInterface
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -3651,11 +3651,6 @@ parameters:
 			path: src/Sulu/Bundle/CategoryBundle/Content/Types/CategorySelection.php
 
 		-
-			message: "#^Call to an undefined method Sulu\\\\Component\\\\Rest\\\\ListBuilder\\\\ListBuilderInterface\\:\\:setParameter\\(\\)\\.$#"
-			count: 1
-			path: src/Sulu/Bundle/CategoryBundle/Controller/CategoryController.php
-
-		-
 			message: "#^Left side of && is always true\\.$#"
 			count: 1
 			path: src/Sulu/Bundle/CategoryBundle/Controller/CategoryController.php
@@ -3947,11 +3942,6 @@ parameters:
 
 		-
 			message: "#^Parameter \\#2 \\$string of function explode expects string, mixed given\\.$#"
-			count: 1
-			path: src/Sulu/Bundle/CategoryBundle/Controller/KeywordController.php
-
-		-
-			message: "#^Parameter \\#2 \\$value of method Sulu\\\\Component\\\\Rest\\\\ListBuilder\\\\AbstractListBuilder\\:\\:where\\(\\) expects string, Sulu\\\\Bundle\\\\CategoryBundle\\\\Entity\\\\CategoryTranslationInterface given\\.$#"
 			count: 1
 			path: src/Sulu/Bundle/CategoryBundle/Controller/KeywordController.php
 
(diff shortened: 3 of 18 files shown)


src/Sulu/Component/Rest/ListBuilder/ListBuilderInterface.php
```diff
@@ -206,7 +206,7 @@ public function setPermissionCheck(UserInterface $user, $permission);
      * Defines a constraint for the rows to return.
      *
      * @param FieldDescriptorInterface $fieldDescriptor The FieldDescriptor which is checked
-     * @param string $value The value the FieldDescriptor should have
+     * @param mixed $value The value the FieldDescriptor should have
```
alexander-schranz: I think we should support only scalar int|float|string here nothing more. If IN is supported maybe array of them.

src/Sulu/Component/Rest/ListBuilder/Doctrine/DoctrineListBuilderFactoryInterface.php
```diff
@@ -11,17 +11,17 @@
...
- * Defines the interface for the Factory of the DoctrineListBuilde.
+ * Defines the interface for the Factory of the DoctrineListBuilder.
  */
 interface DoctrineListBuilderFactoryInterface
 {
     /**
-     * @param string $entityName
+     * Creates a new DoctrineListBuilder for the given entity name and returns it.
+     *
+     * @param class-string $entityName
```
alexander-schranz, line 22: Is that true? As entityName in XML definition can be any valid join or select alias. For joins on the same table multiple times even you can not use the same entityName again and need use a custom string alias. But not sure how the join is created here and if it constructed same as a list.

We should make sure that the join entity name is marked as `string` and not `class-string` can you check that.

alexander-schranz (changes requested):

src/Sulu/Component/Rest/ListBuilder/Doctrine/DoctrineListBuilderFactoryInterface.php
mamazu: I think this has to do with the `securedEntityName` which is also a class-name and defaults to the entityName. So just to be sure this should probably be a class name.

mamazu: Furthermore some functions called `getEntityClassName` also are incorrectly typed with just `string` but one step at a time.

alexander-schranz: Okay then it should be fine for this methods for all metadata and whatever other calls constructed by the metadata is valid to be `string` as field names can be any kind of alias string. The naming there is just very confusing and maybe in future will be aligned better.

alexander-schranz:
This seems add a bunch of new PHPStan errors in 2.6 to the baseline:

```
 ------ ---------------------------------------------------------------------------------------------------------------------------------------------
  Line   src/Sulu/Bundle/ContactBundle/Controller/AbstractMediaController.php
 ------ ---------------------------------------------------------------------------------------------------------------------------------------------
  257    Parameter #2 $value of method Sulu\Component\Rest\ListBuilder\AbstractListBuilder::where() expects bool|float|int|string|null, mixed given.
         🪪  argument.type
 ------ ---------------------------------------------------------------------------------------------------------------------------------------------

 ------ --------------------------------------------------------------------------------------------------------------------------------
  Line   src/Sulu/Bundle/ContactBundle/Controller/AccountController.php
 ------ --------------------------------------------------------------------------------------------------------------------------------
  149    Parameter #1 $fieldDescriptor of method Sulu\Component\Rest\ListBuilder\Doctrine\DoctrineListBuilder::addSelectField() expects
         Sulu\Component\Rest\ListBuilder\FieldDescriptorInterface, mixed given.
         🪪  argument.type
  150    Parameter #1 $idField of method Sulu\Component\Rest\ListBuilder\Doctrine\DoctrineListBuilder::setIdField() expects
         Sulu\Component\Rest\ListBuilder\Doctrine\FieldDescriptor\DoctrineFieldDescriptorInterface, mixed given.
         🪪  argument.type
  151    Parameter #1 $fieldDescriptor of method Sulu\Component\Rest\ListBuilder\AbstractListBuilder::where() expects
         Sulu\Component\Rest\ListBuilder\FieldDescriptorInterface, mixed given.
         🪪  argument.type
  152    Parameter #1 $fieldDescriptor of method Sulu\Component\Rest\ListBuilder\AbstractListBuilder::sort() expects
         Sulu\Component\Rest\ListBuilder\FieldDescriptorInterface, mixed given.
         🪪  argument.type
  152    Parameter #2 $order of method Sulu\Component\Rest\ListBuilder\AbstractListBuilder::sort() expects string, mixed given.
         🪪  argument.type
  202    Parameter #1 $fieldDescriptor of method Sulu\Component\Rest\ListBuilder\AbstractListBuilder::where() expects
         Sulu\Component\Rest\ListBuilder\FieldDescriptorInterface, mixed given.
         🪪  argument.type
 ------ --------------------------------------------------------------------------------------------------------------------------------

 ------ -------------------------------------------------------------------------------------------------------------------
  Line   src/Sulu/Bundle/ContactBundle/Controller/ContactController.php
 ------ -------------------------------------------------------------------------------------------------------------------
  256    Parameter #1 $fieldDescriptor of method Sulu\Component\Rest\ListBuilder\AbstractListBuilder::addGroupBy() expects
         Sulu\Component\Rest\ListBuilder\FieldDescriptorInterface, mixed given.
         🪪  argument.type
  262    Parameter #1 $fieldDescriptor of method Sulu\Component\Rest\ListBuilder\AbstractListBuilder::where() expects
         Sulu\Component\Rest\ListBuilder\FieldDescriptorInterface, mixed given.
         🪪  argument.type
 ------ -------------------------------------------------------------------------------------------------------------------

 ------ --------------------------------------------------------------------------------------------------------------------------------------------
  Line   src/Sulu/Bundle/MediaBundle/Controller/MediaController.php
 ------ --------------------------------------------------------------------------------------------------------------------------------------------
  278    Parameter #1 $fieldDescriptor of method Sulu\Component\Rest\ListBuilder\Doctrine\DoctrineListBuilder::addPermissionCheckField() expects
         Sulu\Component\Rest\ListBuilder\Doctrine\FieldDescriptor\DoctrineFieldDescriptor, Sulu\Component\Rest\ListBuilder\FieldDescriptorInterface
         given.
         🪪  argument.type
  280    Parameter #1 $user of method Sulu\Component\Rest\ListBuilder\Doctrine\DoctrineListBuilder::setPermissionCheck() expects
         Sulu\Component\Security\Authentication\UserInterface, mixed given.
         🪪  argument.type
 ------ --------------------------------------------------------------------------------------------------------------------------------------------

 ------ --------------------------------------------------------------------------------------------------------------
  Line   src/Sulu/Bundle/SecurityBundle/Controller/RoleController.php
 ------ --------------------------------------------------------------------------------------------------------------
  105    Parameter #1 $fieldDescriptor of method Sulu\Component\Rest\ListBuilder\AbstractListBuilder::where() expects
         Sulu\Component\Rest\ListBuilder\FieldDescriptorInterface, mixed given.
         🪪  argument.type
 ------ --------------------------------------------------------------------------------------------------------------

 ------ ----------------------------------------------------------------------------------------------------------------------------------------------
  Line   src/Sulu/Component/Rest/ListBuilder/Filter/TextFilterType.php
 ------ ----------------------------------------------------------------------------------------------------------------------------------------------
  36     Parameter #2 $value of method Sulu\Component\Rest\ListBuilder\ListBuilderInterface::where() expects bool|float|int|string|null, mixed given.
         🪪  argument.type
 ------ ----------------------------------------------------------------------------------------------------------------------------------------------
```

PR #8268, opened 2025-10-17
Fix flacky FormMetadataProviderTest

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? |  yes
| New feature? | no 
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Fix flacky FormMetadataProviderTest which make the pipeline file from time to time.

#### Why?

Was introduced in https://github.com/sulu/sulu/pull/8124

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Tests/Functional/Metadata/FormMetadata/FormMetadataProviderTest.php
+++ b/src/Sulu/Bundle/AdminBundle/Tests/Functional/Metadata/FormMetadata/FormMetadataProviderTest.php
@@ -116,8 +116,9 @@ class FormMetadataProviderTest extends KernelTestCase
             ['tags' => ['test2' => ['test' => 'test-value2']]]
         );
         $this->assertInstanceOf(TypedFormMetadata::class, $typedForm);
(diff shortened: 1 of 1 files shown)


src/Sulu/Bundle/AdminBundle/Tests/Functional/Metadata/FormMetadata/FormMetadataProviderTest.php
```diff
@@ -116,34 +116,38 @@ public function testGetMetadataTagFiltered(): void
             ['tags' => ['test2' => ['test' => 'test-value2']]]
         );
         $this->assertInstanceOf(TypedFormMetadata::class, $typedForm);
-        $this->assertCount(1, $typedForm->getForms());
```
alexander-schranz, line 119: not required to check for count if I already do a `assertSame` on it. It provides more infos what is missing as just the count assertion.

The PHPUnit output is in such cases better and more informative:

Before with assertCount:

```
Expected count was 2 got 1.
```

After with assertSame:

```
[
     'default',
+    'overview',
]
```

PR #8271, opened 2025-10-20
Add search listener and search provider for snippets for SEAL

martinlagler:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no 
| Deprecations? | no
| Related issues/PRs | #7672 
| License | MIT

#### What's in this PR?

Added ReindexProvider and IndexListener for Snippets.


packages/snippet/src/Domain/Event/SnippetRemovedEvent.php
```diff
@@ -21,11 +21,13 @@ class SnippetRemovedEvent extends DomainEvent
 {
     /**
      * @param array<string, mixed> $context
+     * @param string[]|null $allLocales
      */
     public function __construct(
         private string $snippetId,
         private ?string $snippetTitle,
         private array $context = [],
+        private ?array $allLocales,
```
martinlagler: @alexander-schranz How should we handle this ? Last param optional...

alexander-schranz: do we have other Events which have `locales` or `allLocales` already?

martinlagler: Yes, but there is no mandatory param before...
https://github.com/sulu/sulu/blob/3.0/src/Sulu/Bundle/CategoryBundle/Domain/Event/CategoryRemovedEvent.php#L27

alexander-schranz: Is the `context` even filled up with some value? Do other events have a `context` parameter in the constructor?

martinlagler: @alexander-schranz  No it's not filled up [see](https://github.com/sulu/sulu/blob/3.0/packages/snippet/src/Application/MessageHandler/RemoveSnippetMessageHandler.php#L51) 

Also the ArticleRemovedEvent has a `context` but it's also not filled up [see](https://github.com/sulu/sulu/blob/3.0/packages/article/src/Domain/Event/ArticleRemovedEvent.php#L28)

I guess I can remove it ?

alexander-schranz: I'm thinking of if we should save it inside context and do: 

```php
public function getAllLocales(): ?array
{
    return $this->context['locales'] ?? null;
}
```

@wachterjohannes what do you think we should handle this?

martinlagler: We can do that, but then i would also change it for the [CategoryRemovedEvent](https://github.com/sulu/sulu/blob/3.0/src/Sulu/Bundle/CategoryBundle/Domain/Event/CategoryRemovedEvent.php#L27)

wachterjohannes: yes put it into the context

alexander-schranz (approved):

PR #8276, opened 2025-10-22
Remove deprecated object format for entity selections

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | yes
| Deprecations? | -
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?
Removing the option to configure the deprecated object format.

#### Why?
We don't seem to use this parameter anymore so might as well remove it.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/AutoCompletePopover/Suggestion.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/AutoCompletePopover/Suggestion.js
@@ -49,13 +49,13 @@ export default class Suggestion extends React.PureComponent<Props> {
             <span>
                 {splittedText.map((splitText, index) => {
                     return (
-                        <div key={'suggestion.'+index}>
+                        <>
                             {splitText}
                             {highlightedWords && highlightedWords[index]
                                 ? <strong>{highlightedWords[index]}</strong>
                                 : null
                             }
-                        </div>
+                        </>
                     );
                 })}
             </span>
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/AutoCompletePopover/Suggestion.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/AutoCompletePopover/Suggestion.js
@@ -49,13 +49,13 @@ export default class Suggestion extends React.PureComponent<Props> {
             <span>
                 {splittedText.map((splitText, index) => {
                     return (
-                        <>
+                        <div key={'suggestion.'+index}>
                             {splitText}
                             {highlightedWords && highlightedWords[index]
                                 ? <strong>{highlightedWords[index]}</strong>
                                 : null
                             }
-                        </>
+                        </div>
                     );
                 })}
             </span>
--- a/src/Sulu/Bundle/ContactBundle/Api/Contact.php
+++ b/src/Sulu/Bundle/ContactBundle/Api/Contact.php
@@ -845,13 +845,26 @@ class Contact extends ApiWrapper
 
     /**
      * Returns main account.
+     *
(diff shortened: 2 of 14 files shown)


alexander-schranz (changes requested):
Did you test if saving of the changed form fields still work? 

For me it looks like the Services would be required to change where they get now the `id` previously the Form did send to the API:

`$data['mainContact']['id']` and now the id is in `$data['mainContact']`.

Also the API need to change it response as the value is not longer `{"mainContact": {"id": 1}}` instead it is `{"mainContact": 1}`?

If I understand the flag correctly?

src/Sulu/Bundle/ContactBundle/Api/Contact.php
```diff
@@ -851,12 +851,7 @@ public function getGender()
     #[Groups(['fullContact'])]
     public function getAccount()
     {
-        $mainAccount = $this->entity->getMainAccount();
-        if (!\is_null($mainAccount)) {
-            return new Account($mainAccount, $this->locale);
-        }
-
-        return;
+        return $this->entity->getMainAccount()?->getId();
```
alexander-schranz: this method should still return the account.

Create a new `getAccountId` method for the admin controller serialization.

alexander-schranz (changes requested):

src/Sulu/Bundle/AdminBundle/Resources/js/components/AutoCompletePopover/Suggestion.js
```diff
@@ -49,13 +49,13 @@ export default class Suggestion extends React.PureComponent<Props> {
             <span>
                 {splittedText.map((splitText, index) => {
                     return (
-                        <>
+                        <div key={'suggestion.'+index}>
```
alexander-schranz: that can not be changed to div this should stay as a fragment its not allowed to be a tag here.

mamazu: Yeah but I don't think fragments support keys?

mamazu:
This api is a little hard to test since any structurally incorrect data is ignored and just set to null.

mamazu:
I have no idea why the tests fail, any ideas?

mamazu:
Okay, so the state of this PR is: the frontend works. You can save and load from the API in the two cases that changed. However the test still fail. I might have forgotten to change some of the places where the data is mapped but since there is no validation it doesn't tell me which places need to be changed as well.

src/Sulu/Bundle/AdminBundle/Resources/js/components/AutoCompletePopover/Suggestion.js
alexander-schranz: If that change is required we would require it on 2.5. I think `<React.Fragment key="">` is the way to have keys on fragments.

alexander-schranz (commented):
Think one is the `mainAccount` here:

https://github.com/sulu/sulu/blob/57b6e1a756662a46a5cdea2151d2b872361c3d41/src/Sulu/Bundle/ContactBundle/Controller/AccountController.php#L585

But looks like there is a issue with the position or is that working in the UI?

src/Sulu/Bundle/AdminBundle/Resources/js/components/AutoCompletePopover/Suggestion.js
mamazu: Split into #8298

alexander-schranz (approved):

alexander-schranz:
@mamazu awesome, thank you!

Issue #5214, opened 2020-04-14
Mini text editor for 


alexander-schranz:
# Problem description

It would be great to have a mini text editor which does only contains inline html elements. The usecase is mostly for titles or labels which contains just some bold, italics words or link:

Example:

```twig
<label>{{ content.textEditor|raw }}</label>
```

Block tags like `p`, `ul`, `table` are not allowed inside a `p` tag.

# Proposed solution

To think bigger it would be great define different configs and use them:

```xml
<property name="line" type="text_editor">
      <param>
            <param name="config" value="mini_editor"/>
      </param>
</property>
```

The ckeditor registry could look like this:

```
ckeditorConfigRegistry.add(..., 'mini_editor'); // to avoid bc break this second parameter is optional and fallback to `default`
```

As I think ckeditor does not support avoiding the `p` tag I would go with removing it in the content type as I think we don't want to implement such logic in the js:

```
<property name="line" type="text_editor">
      <param>
            <param name="config" value="mini_editor"/>
            <param name="strip_p_tags" value="true"/>
      </param>
</property>
```

The `strip_p_tags` will replace `</p><p>` with `<br/><br/>` and remove `<p>` and `</p>` in the content type.

danrot:
You've already made the same suggestion as your proposed solution in https://github.com/sulu/sulu/issues/4898#issuecomment-554394189, and I haven't changed my mind about my following comment. I think this quickly gets out of hand, resulting in many different configs, and it is not clear from looking at it what it really does.

However, I can see that limiting the number HTML tags makes sense. And I could imagine that CKEditor supports that somehow. The first thing that comes to my mind is to [remove some plugins using the configuration](https://ckeditor.com/docs/ckeditor5/latest/builds/guides/integration/configuration.html#removing-features). This example shows that this is at least possible for the Heading, and stuff like lists are also implemented as plugins. So I could imagine that we could deactivate everything here already, without the need of implementing something like this in the serverside code.

danrot:
Yesterday another solution came to my mind: What about using `XIncludes` for this? Then we would have a solution that works for other content types as well, not only for the text editor.

alexander-schranz:
@danrot Must say xincludes is for me like saying ignore the problem and tell the developers you need implement it yourself always and you will have no possibility to effect third party bundles editors.

danrot:
Ok, I've added this to the wrong issue... But the initial problem here could be solved by a `inline_only` param as well.

[...]

PR #8279, opened 2025-10-23
Add possibility to remove p tags from text_editor

martinlagler:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | yes
| Deprecations? | no
| Related issues/PRs | #5214
| License | MIT

#### What's in this PR?

Add possibility to script p tags from the `text_editor`. 

#### Why?

To fix FormBundle issue and add possibility to use text_editor without p tags.

#### Example Usage

```xml
    <property name="title" type="text_editor" mandatory="true">
        <meta>
            <title>sulu_form.title</title>
        </meta>

        <params>
            <param name="enter_mode" value="br"/>
        </params>
    </property>
```

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/CKEditor5/CKEditor5.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/CKEditor5/CKEditor5.js
@@ -71,8 +71,8 @@ export default class CKEditor5 extends React.Component<Props> {
                 value,
                 disabled,
                 options: {
-                    strip_p_tags: {
-                        value: stripPTags = false,
+                    enter_mode: {
+                        value: enterModeValue = 'p',
                     } = {},
                 } = {},
             } = this.props;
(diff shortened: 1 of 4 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/containers/CKEditor5/tests/utils.test.js
```diff
@@ -0,0 +1,82 @@
+// @flow
+import {removePTags, addPTags} from '../utils';
+
+test('Test remove p tags', () => {
+    const html = '<p>This is a paragraph.</p>';
+    const expected = '<!--p-->This is a paragraph.<!--/p-->';
+    expect(removePTags(html)).toBe(expected);
```
alexander-schranz: any way we can detect such basic example so the expected result is just `This is a paragraph.`.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/CKEditor5/CKEditor5.js
```diff
@@ -75,13 +87,25 @@ export default class CKEditor5 extends React.Component<Props> {
...
         }
     }
 
     componentDidMount() {
-        const {formats, locale} = this.props;
+        const {
+            formats,
+            locale,
+            options: {
+                strip_p_tags: {
```
alexander-schranz: would go with `enter_mode` and currently we have 2 `'p'|'br'`. 

Like we did for Sulu 1.6 and CKEditor 4: https://github.com/sulu/SuluFormBundle/blob/f72a61ce4f3724a58424cf78ed61fd7bf857759c/Resources/views/forms/fields/default-field.html.twig#L97

alexander-schranz (changes requested):

PR #8249, opened 2025-10-03
Add Internal and External Link possibility to Pages and Content Bundle

Prokyonn:
| Q | A
  | --- | ---
  | Bug fix? | no
  | New feature? | yes
  | BC breaks? | no
  | Deprecations? | no
  | Fixed tickets | -
  | Related issues/PRs | #7672 
  | License | MIT
  | Documentation PR | sulu/sulu-docs# 

<details><summary>Old Description</summary>

  #### What's in this PR?

  Adds page type/behavior functionality to the page package, allowing pages to be configured as:
  - **Content pages**: Standard pages with content (default behavior)
  - **Internal link pages**: Pages that redirect to another page internally
  - **External link pages**: Pages that redirect to an external URL

  **Implementation includes:**

  - `ContentBehavior` domain model, interface, and trait
  - `ContentBehaviorDataMapper` for mapping behavior data
  - `ContentBehaviorMerger` for merging behavior across dimensions
  - `ContentAggregationEnhancer` interface and `PageContentBehaviorEnhancer` implementation for resolving internal/external link targets
  - Comprehensive test coverage for all new functionality

</details>

Add a `LinkInterface` and `LinkTrait` used by the PageDimension to Link to other pages in SmartContent, Navigation. Effects the Sitemap as such pages are not listed there.

  #### Why?

  Missing feature that exists in Sulu 2.6

PR #8284, opened 2025-10-27
Fix media smart content filters default behaviour

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no 
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Return current filters instead of unset them.

#### Why?

@Prokyonn stumbled over this in https://github.com/sulu/sulu/pull/8249. As it's not related directly to #8249 merging this as an own PR.


alexander-schranz:
@Prokyonn FYI I also added a test class as was missing test for the whole class.

martinlagler (approved):

PR #8288, opened 2025-10-28
Block ids pregenerate

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | #7672 
| Related issues/PRs | https://github.com/sulu/sulu/pull/8262
| License | MIT
| Documentation PR | none

#### What's in this PR?

This PR add the pregeneration of block-ids if the server respond blocks with missing id.

#### Why?

When blocks are created e.g. in fixtures the user could forget the id and so the UI adds them later.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/BlockCollection.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/BlockCollection.js
@@ -98,13 +98,13 @@ class BlockCollection<T: string, U: {_id?: string, type: T, ...}> extends React.
     }
 
     componentDidUpdate(prevProps: Props<T, U>) {
-        const {value} = this.props;
+        const {generateBlockIds, value} = this.props;
 
         // Only call ensureBlockIds if:
-        // 1. The value reference changed (not the one we just set)
-        // 2. There are actually blocks without IDs
-        // 3. We're not currently generating IDs
-        if (prevProps.value !== value && !this.isGeneratingIds) {
+        // 1. generateBlockIds function is provided
+        // 2. The value reference changed
+        // 3. There are actually blocks without IDs
+        if (generateBlockIds && prevProps.value !== value) {
             const hasBlocksWithoutIds = value && value.some((block) => !block._id);
             if (hasBlocksWithoutIds) {
                 this.ensureBlockIds();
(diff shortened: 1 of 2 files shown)


alexander-schranz:
@wachterjohannes rebase this one so we can review it.

wachterjohannes:
> @wachterjohannes rebase this one so we can review it.

done :)

src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/BlockCollection.js
```diff
@@ -91,13 +92,72 @@ class BlockCollection<T: string, U: {_id?: string, type: T, ...}> extends React.
...
+    }
+
+    componentDidUpdate(prevProps: Props<T, U>) {
+        const {value} = this.props;
+
+        // Only call ensureBlockIds if:
+        // 1. The value reference changed (not the one we just set)
+        // 2. There are actually blocks without IDs
+        // 3. We're not currently generating IDs
+        if (prevProps.value !== value && !this.isGeneratingIds) {
```
alexander-schranz: if the param is not defined we also don't need to call `this.ensureBlockIds();`

src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/BlockCollection.js
```diff
@@ -50,6 +50,7 @@ class BlockCollection<T: string, U: {_id?: string, type: T, ...}> extends React.
     @observable expandedBlocks: Array<boolean> = [];
     @observable selectedBlocks: Array<boolean> = [];
     @observable mode: BlockMode = 'sortable';
+    @observable isGeneratingIds: boolean = false;
```
alexander-schranz, line 53: This feels strange. Is that really required? If we add, paste ablock it should already have a block id and so hasBlocksWithoutIds should always be false.

And if the store is updated by the backend which can happen and not provide ids it should generate ids. SO I would remove this state as I think it would make issues if store is edited from outside.

alexander-schranz (changes requested):

src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/BlockCollection.js
alexander-schranz: @wachterjohannes did you miss this comment?

wachterjohannes: @alexander-schranz no i wanted to avoid that the generateIds run multiple times at the same time see: https://github.com/sulu/sulu/pull/8288/files#diff-67f069e752b8a5f4447294d36596f4596b5237278dd9c3de93d5ab79e297538dR124

PR #8297, opened 2025-10-30
Rename admin search providers and listeners

martinlagler:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Related issues/PRs | #7672
| License | MIT

#### What's in this PR?

Rename Search providers and listeners to separate admin and website search.
This PR is build on top of #8294

### DIFF
--- a/packages/article/src/Infrastructure/Symfony/HttpKernel/SuluArticleBundle.php
+++ b/packages/article/src/Infrastructure/Symfony/HttpKernel/SuluArticleBundle.php
@@ -425,8 +425,8 @@ final class SuluArticleBundle extends AbstractBundle
             ])
             ->tag('sulu_route.route_defaults_provider', ['resource_key' => 'articles']);
 
-        $services->set('sulu_article.article_index_listener')
-            ->class('Sulu\Article\Infrastructure\Sulu\Search\ArticleIndexListener')
+        $services->set('sulu_article.admin_article_index_listener')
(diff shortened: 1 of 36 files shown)


packages/article/src/Infrastructure/Symfony/HttpKernel/SuluArticleBundle.php
```diff
@@ -425,8 +425,8 @@ public function loadExtension(array $config, ContainerConfigurator $container, C
             ])
             ->tag('sulu_route.route_defaults_provider', ['resource_key' => 'articles']);
 
-        $services->set('sulu_article.article_index_listener')
-            ->class('Sulu\Article\Infrastructure\Sulu\Search\ArticleIndexListener')
+        $services->set('sulu_article.admin_article_index_listener')
+            ->class('Sulu\Article\Infrastructure\Sulu\Search\AdminArticleIndexListener')
```
mamazu, line 429: If you're already replacing it maybe using class strings like "AdminArticleIndexListener::class" Then a rename would have caught that.

alexander-schranz (approved):

packages/article/src/Infrastructure/Symfony/HttpKernel/SuluArticleBundle.php
alexander-schranz: @martinlagler make sense, to avoid also conflicts here.

PR #8299, opened 2025-11-03
Fix handling of null in Resolvers

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | none
| Related issues/PRs | none
| License | MIT
| Documentation PR | none

#### What's in this PR?

This PR improves type safety in property resolvers and pagination handling:

1. **Nullable resolver parameters**: Added null checks and nullable type hints to `LinkPropertyResolver` and `TeaserSelectionPropertyResolver` to handle cases where external
links or unresolved teasers are passed to resolvers.

2. **Integer type casting**: Added explicit `(int)` cast for pagination limit in `PageSmartContentProvider` to ensure type safety.

#### Why?

**Property Resolvers:**
External links are not passed as `LinkItem` objects (as noted in existing comments), which could cause type errors when the resolver closure expects a non-null `LinkItem`.
Similarly, teaser selections may contain items that don't resolve to valid `Teaser` resources. Without null guards, these scenarios would cause runtime type errors.

**Pagination:**
The `addPagination` method expects an integer parameter, but `$filters['limit']` from request data could potentially be a string or mixed type, leading to type
inconsistencies.

### DIFF
--- a/packages/content/src/Domain/Model/LinkInterface.php
+++ b/packages/content/src/Domain/Model/LinkInterface.php
@@ -15,8 +15,6 @@ namespace Sulu\Content\Domain\Model;
 
 interface LinkInterface
 {
-    public function getLinkProvider(): ?string;
-
     /**
      * @return array<string, mixed>|null
      */
--- a/packages/content/src/Domain/Model/LinkInterface.php
+++ b/packages/content/src/Domain/Model/LinkInterface.php
@@ -15,6 +15,8 @@ namespace Sulu\Content\Domain\Model;
 
 interface LinkInterface
 {
+    public function getLinkProvider(): ?string;
+
     /**
      * @return array<string, mixed>|null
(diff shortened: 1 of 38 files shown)


Prokyonn:
@wachterjohannes please retest the changes in this MR for your project

packages/content/src/Domain/Model/LinkInterface.php
```diff
@@ -15,6 +15,8 @@
 
 interface LinkInterface
 {
+    public function getLinkProvider(): ?string;
```
alexander-schranz: would not add this as its only for query and the getLinkData contains the provider if we need access it.

alexander-schranz (approved):

PR #8301, opened 2025-11-03
Deprecate `ImageTransformationCompilerPass`

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | yes
| Fixed tickets | -
| Related issues/PRs | #7402
| License | MIT
| Documentation PR | =

#### What's in this PR?
Deprecating the compilerpass for the image transformations and replacing it with a tagged_locator.

#### Why?
* Less code
* Lazily intitalizing services for transformations

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/Media/ImageConverter/TransformationPool.php
+++ b/src/Sulu/Bundle/MediaBundle/Media/ImageConverter/TransformationPool.php
@@ -24,7 +24,7 @@ class TransformationPool implements TransformationPoolInterface
      */
     private $transformations = [];
 
-    public function __construct(private ContainerInterface $container)
+    public function __construct(private ?ContainerInterface $container = null)
     {
     }
 
--- a/src/Sulu/Bundle/MediaBundle/Media/ImageConverter/TransformationPool.php
+++ b/src/Sulu/Bundle/MediaBundle/Media/ImageConverter/TransformationPool.php
@@ -48,7 +48,10 @@ class TransformationPool implements TransformationPoolInterface
     public function get($name)
     {
         try {
-            return $this->container->get($name);
+            /** @var TransformationInterface $service */
+            $service = $this->container->get($name);
+
+            return $service;
         } catch (\Throwable) {
             // do nothing and try the old logic.
         }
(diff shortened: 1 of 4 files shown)


src/Sulu/Bundle/MediaBundle/Resources/config/services.xml
```diff
@@ -220,7 +220,9 @@
         <service id="sulu_media.image.scaler" class="%sulu_media.image.scaler.class%" />
         <service id="sulu_media.image.cropper" class="%sulu_media.image.cropper.class%" />
 
-        <service id="sulu_media.image.transformation_pool" class="%sulu_media.image.transformation_pool.class%" />
+        <service id="sulu_media.image.transformation_pool" class="%sulu_media.image.transformation_pool.class%">
+            <argument type="tagged_locator" tag="sulu_media.image.transformation" index-by="alias" />
+        </service>
```
alexander-schranz: No this should only be done on 3.0, on 2.6 the compiler pass still takes care of it and we don't want double added.

```suggestion
        <service id="sulu_media.image.transformation_pool" class="%sulu_media.image.transformation_pool.class%" />
```

src/Sulu/Bundle/MediaBundle/Media/ImageConverter/TransformationPool.php
```diff
@@ -23,8 +24,14 @@ class TransformationPool implements TransformationPoolInterface
      */
     private $transformations = [];
 
+    public function __construct(private ContainerInterface $container)
+    {
+    }
+
```
alexander-schranz: Needs be nullable for bc reasons:

```suggestion
    public function __construct(private ?ContainerInterface $container = null)
    {
    }

```

src/Sulu/Bundle/MediaBundle/Media/ImageConverter/TransformationPool.php
```diff
@@ -40,6 +47,15 @@ public function add(TransformationInterface $transformation, $alias)
      */
     public function get($name)
     {
+        try {
+            /** @var TransformationInterface $service */
+            $service = $this->container->get($name);
```
alexander-schranz: Not sure how we handled this previously?

But container can be null.

alexander-schranz (changes requested):

alexander-schranz (approved):

Issue #8253, opened 2025-10-05
Filter not working in single_snippet_selection


manuxi:
| Q | A
| --- | ---
| Sulu Version | 3.0.x
| PHP Version | 8.3.3
| DB Version | 5.7.32
| Browser Version | Chrome latest

# Actual Behavior

I use several snippets, one with key: hero

```xml
<property name="snippet_hero" type="single_snippet_selection"
    <meta>
        <title lang="en">Snippet (Hero)</title>
    </meta>
    <params>
        <param name="types" value="hero"/> <!-- types not working -->
        <param name="default" value="hero"/>
        <param name="loadExcerpt" value="true"/>
    </params>
</property>
```

Does not restrict the snippet selection to heros, I'm seeing all snippets.

# Expected Behavior

I do only see snippets of filtered type in selection

# Steps to Reproduce

Just some snippets and the xml above

alexander-schranz:
Thx for the report. Seems like the `types` filter is not keep in mind in the new Endpoint, we need to tackle that in the cgetAction of the new SnippetController.

PR #8305, opened 2025-11-03
Fix template filtering in selections

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Fixed tickets | fixes https://github.com/sulu/sulu/issues/8253
| License | MIT

#### What's in this PR?

- Adds support for filtering snippets with the `types` queryparameter (filters for the template)
- Adds support for filtering articles for their `groups` with the `types` query parameter

#### Why?

Missing functionality in Sulu 3.0.

### DIFF
--- a/packages/snippet/src/UserInterface/Controller/Admin/SnippetController.php
+++ b/packages/snippet/src/UserInterface/Controller/Admin/SnippetController.php
@@ -83,6 +83,26 @@ final class SnippetController
         $listBuilder->setParameter('locale', $request->query->get('locale'));
         $this->restHelper->initializeListBuilder($listBuilder, $fieldDescriptors);
 
+        $typesParam = $request->query->get('types');
+        $types = \array_filter(\explode(',', \is_string($typesParam) ? $typesParam : ''));
+
+        if (0 !== \count($types)) {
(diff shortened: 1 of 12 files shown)


packages/snippet/src/UserInterface/Controller/Admin/SnippetController.php
```diff
@@ -83,6 +83,15 @@ public function cgetAction(Request $request): Response
         $listBuilder->setParameter('locale', $request->query->get('locale'));
         $this->restHelper->initializeListBuilder($listBuilder, $fieldDescriptors);
 
+        $typesParam = $request->query->get('types');
```
Prokyonn, line 86: This would be equivalent with the 2.6 controller. For articles we have `templates` as a parameter and for pages we do not have any filtering documented for 2.6. 

@alexander-schranz How should we proceed here, I think it would make sense to make it consistent across these three entities.

alexander-schranz: As I remember `types` is the only parameter directly supported by the selection field types.

Can not find `templates` in articles content types of 2.6 be supported out of the box or configured. Looks they use types, but yes the article types in that case.

alexander-schranz (approved):

PR #8307, opened 2025-11-04
Add Admin and Website permissions for Sulu 3.0

draconivis:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | 
| Related issues/PRs | 
| License | MIT
| Documentation PR | 

#### What's in this PR?

Adds back the permissions tab and ensure the permissions are applied correctly

#### Why?

currently not working as in 2.6

### DIFF
--- a/packages/content/src/Application/ContentNormalizer/Normalizer/SecuredEntityNormalizer.php
+++ b/packages/content/src/Application/ContentNormalizer/Normalizer/SecuredEntityNormalizer.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace Sulu\Content\Application\ContentNormalizer\Normalizer;
 
+use Sulu\Component\Security\Authentication\UserInterface;
 use Sulu\Component\Security\Authorization\AccessControl\AccessControlManagerInterface;
 use Sulu\Component\Security\Authorization\AccessControl\SecuredEntityInterface;
 use Sulu\Content\Domain\Model\DimensionContentInterface;
@@ -48,14 +49,17 @@ class SecuredEntityNormalizer implements NormalizerInterface
 
         $allPermissions = $this->accessControlManager->getPermissions(
             \get_class($securedEntity),
-            $securedEntity->getId()
+            (string) $securedEntity->getId()
         );
+
+        $user = $this->tokenStorage?->getToken()?->getUser();
+        $suluUser = $user instanceof UserInterface ? $user : null;
 
         $permissions = $this->accessControlManager->getUserPermissionByArray(
             null,
             $securedEntity->getSecurityContext(),
             $allPermissions,
-            $this->tokenStorage?->getToken()?->getUser()
+            $suluUser
         );
 
         $normalizedData['_permissions'] = $permissions;
--- a/packages/content/src/Application/ContentNormalizer/Normalizer/SecuredEntityNormalizer.php
+++ b/packages/content/src/Application/ContentNormalizer/Normalizer/SecuredEntityNormalizer.php
@@ -1,0 +1,71 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of Sulu.
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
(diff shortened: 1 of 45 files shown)


packages/page/src/Infrastructure/Sulu/Admin/PageAdmin.php
```diff
@@ -139,7 +139,7 @@ public function configureViews(ViewCollection $viewCollection): void
                     new ToolbarAction(
                         'sulu_admin.copy_locale',
                         [
-                            'visible_condition' => '(!_permissions || _permissions.edit) && __webspace.localizations|length > 1',
+                            'visible_condition' => '(_permissions && _permissions.edit) && __webspace.localizations|length > 1',
```
Prokyonn: What is the reason for this change? The toolbar should be visible either when no permissions are enabled or when permissions are enabled and the user has edit permissions

packages/page/src/Domain/Model/Page.php
```diff
@@ -14,14 +14,15 @@
...
+use Sulu\Component\Security\Authorization\AccessControl\SecuredEntityInterface;
 use Sulu\Content\Domain\Model\ContentRichEntityTrait;
 use Sulu\Content\Domain\Model\DimensionContentInterface;
 use Symfony\Component\Uid\Uuid;
 
 /**
  * @experimental
  */
-class Page implements PageInterface
+class Page implements PageInterface, SecuredEntityInterface
```
Prokyonn: Interface should be added to the `PageInterface` directly

packages/page/src/Infrastructure/Symfony/HttpKernel/SuluPageBundle.php
```diff
@@ -568,6 +570,8 @@ public function prependExtension(ContainerConfigurator $container, ContainerBuil
                                 'list' => 'sulu_page.get_pages',
                                 'detail' => 'sulu_page.get_page',
                             ],
+                            'security_class' => Page::class,
```
Prokyonn, line 585: Would this also work to use the PageInterface? Not sure how this works when e.g. the page is extended by another custom implementation @alexander-schranz

packages/page/src/UserInterface/Controller/Admin/PageController.php
```diff
@@ -245,7 +253,7 @@ private function getData(Request $request): array
         );
     }
 
-    private function getLocale(Request $request): string
+    public function getLocale(Request $request): string
```
Prokyonn, line 256: Public? 🤔

packages/page/src/Infrastructure/Symfony/HttpKernel/SuluPageBundle.php
alexander-schranz: Would prefer change this to resource key, resource id. As we already need migrate the access control.

packages/page/src/UserInterface/Controller/Admin/PageController.php
alexander-schranz: Think security interface requires this as roles are added to a user per locale.

packages/content/src/Application/ContentNormalizer/Normalizer/SecuredEntityNormalizer.php
```diff
@@ -0,0 +1,75 @@
...
+use Sulu\Component\Security\Authorization\AccessControl\AccessControlManagerInterface;
+use Sulu\Component\Security\Authorization\AccessControl\SecuredEntityInterface;
+use Sulu\Content\Domain\Model\DimensionContentInterface;
+use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface;
+
+/**
+ * Adds permission information to entities implementing SecuredEntityInterface.
+ * This normalizer mimics the behavior of SecuredEntitySubscriber for Symfony Serializer.
+ */
+class SecuredEntityNormalizer implements NormalizerInterface
```
alexander-schranz, line 26: This service is not related to the Content Domain.

Not sure if this is more a Infrastructure Security Sulu Service or a Sulu Content service in Sulu Security.

Would currently mark it as `@internal` so we are still able to move whereever we want it.

Fine in upcoming PR.

PR #8311, opened 2025-11-04
Remove deprecated_object_format (part 2)

mamazu:
Found another instance of the deprecated_object_format processing in the backend

refs #8276


mamazu:
Why is postgres sorting things differently.

PR #8322, opened 2025-11-07
Select with mandatory fields

manuxi:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no yes
| Deprecations? | no
| License | MIT

#### Why?


manuxi:
Clean PR: #8502

PR #8324, opened 2025-11-10
Support Default param in SingleSnippetSelection

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Related issues/PRs | #7672 
| License | MIT

#### What's in this PR?

- adds support for the "default" param in the snippet template.

#### Why?

- this was already supported in Sulu 2.6

#### Example Usage

```twig
        <property name="header" type="single_snippet_selection">
            <meta>
                <title lang="en">Header-Snippet</title>
            </meta>

            <params>
                <param name="default" value="header-area"/>
            </params>
        </property>```
```

### DIFF
--- a/packages/snippet/src/Domain/Repository/SnippetAreaRepositoryInterface.php
+++ b/packages/snippet/src/Domain/Repository/SnippetAreaRepositoryInterface.php
@@ -53,6 +53,16 @@ interface SnippetAreaRepositoryInterface
      *     uuids?: string[],
      *     webspaceKey?: string,
      *     areaKey?: string,
+     * } $filters
+     */
+    public function findSnippetUuidBy(array $filters): ?string;
+
+    /**
+     * @param array{
+     *     uuid?: string,
+     *     uuids?: string[],
+     *     webspaceKey?: string,
+     *     areaKey?: string,
      *     page?: int,
      *     limit?: int,
      * } $filters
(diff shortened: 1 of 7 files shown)


packages/snippet/src/Domain/Repository/SnippetAreaRepositoryInterface.php
```diff
@@ -47,6 +47,16 @@ public function getOneBy(array $filters): SnippetAreaInterface;
...
 
+    /**
+     * @param array{
+     *     uuid?: string,
+     *     uuids?: string[],
+     *     webspaceKey?: string,
+     *     areaKey?: string,
+     * } $filters
+     */
+    public function findSnippetUuidBy(array $filters): ?string;
```
alexander-schranz: avoid the `Snippet` prefix we are already in the Snippet Repository.


```suggestion
    public function findOneUuidBy(array $filters): ?string;
```

packages/snippet/src/Infrastructure/Doctrine/Repository/SnippetAreaRepository.php
```diff
@@ -76,6 +76,20 @@ public function findOneBy(array $filters): ?SnippetAreaInterface
         return $snippetArea;
     }
 
+    public function findSnippetUuidBy(array $filters): ?string
```
alexander-schranz:
```suggestion
    public function findOneUuidBy(array $filters): ?string
```

alexander-schranz (changes requested):

alexander-schranz:
@Prokyonn we require it also for `snippet_selection`: https://docs.sulu.io/en/2.6/reference/content-types/snippet_selection.html

PR #8325, opened 2025-11-10
Hide SnippetArea when no area is defined and adjust translations

Prokyonn:
| Q | A
  | --- | ---
  | Bug fix? | yes
  | New feature? | yes
  | BC breaks? | no
  | Deprecations? | no
  | License | MIT

  #### What's in this PR?

  This PR improves the snippet area functionality by:

  1. **Conditionally hiding the "Default Snippets" tab** when no snippet areas are configured in the project
  2. **Fixing translation inconsistencies** in both English and German translation files

  #### Why?

  **Problem 1: Empty Tab Visibility**
  The "Default Snippets" tab was always visible in webspace settings, even when no snippet areas were defined in any `snippet.xml` file.

  **Problem 2: Translation Issues**
  Multiple translation inconsistencies existed:
  - English file contained German strings (e.g., "Typ oder Bereich", "Taxonomien")
  - German file mixed English and German (e.g., "Snippet auswählen" instead of "Schnipsel auswählen")
  - Wrong translation key used (`pages.restored` instead of `snippets.restored`)


packages/snippet/src/Infrastructure/Sulu/Admin/SnippetAreaAdmin.php
```diff
@@ -41,6 +45,10 @@ public function configureViews(ViewCollection $viewCollection): void
             return;
         }
 
+        if (empty($this->snippetAreas)) {
+            return;
+        }
```
alexander-schranz: I would even unregister the service in the compiler pass. What do you think?

alexander-schranz (commented):
Can you add also a section to the `UPGRADE.md` that areas now need to be defined?

alexander-schranz (approved):

PR #8330, opened 2025-11-11
Deal with absent default meta

matthiasnoback:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

This change fixes the functionality for uploading a new file version in the Media bundle.

#### Why?

Technically (and in our database in practice too!), the default file meta is not necessarily available for every media file. If you upload a new version, the current code in `__clone` expects to find a default meta and calls `getLocale()` on it. It should deal with the fact that the return value of `getDefaultMeta()` may be `null`.

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/Api/Media.php
+++ b/src/Sulu/Bundle/MediaBundle/Api/Media.php
@@ -940,7 +940,7 @@ class Media extends ApiWrapper
     /**
      * @param bool $create
      *
-     * @return FileVersionMeta
+     * @return ($create is true ? FileVersionMeta : FileVersionMeta|null)
      */
     private function getMeta($create = false)
     {
--- a/src/Sulu/Bundle/MediaBundle/Entity/FileVersion.php
+++ b/src/Sulu/Bundle/MediaBundle/Entity/FileVersion.php
@@ -533,7 +533,7 @@ class FileVersion implements AuditableInterface
     /**
      * Get defaultMeta.
      *
-     * @return FileVersionMeta
+     * @return FileVersionMeta|null
      */
     public function getDefaultMeta()
     {
--- a/src/Sulu/Bundle/MediaBundle/Entity/FileVersion.php
+++ b/src/Sulu/Bundle/MediaBundle/Entity/FileVersion.php
@@ -549,7 +549,7 @@ class FileVersion implements AuditableInterface
             $this->id = null;
             /** @var FileVersionMeta[] $newMetaList */
             $newMetaList = [];
-            $defaultMetaLocale = $this->getDefaultMeta()->getLocale();
+            $defaultMetaLocale = $this->getDefaultMeta()?->getLocale();
             /** @var FileVersionContentLanguage[] $newContentLanguageList */
             $newContentLanguageList = [];
             /** @var FileVersionPublishLanguage[] $newPublishLanguageList */
--- a/src/Sulu/Bundle/MediaBundle/Tests/Functional/Trash/MediaTrashItemHandlerTest.php
+++ b/src/Sulu/Bundle/MediaBundle/Tests/Functional/Trash/MediaTrashItemHandlerTest.php
@@ -260,7 +260,7 @@ class MediaTrashItemHandlerTest extends SuluTestCase
         static::assertSame([], $restoredFile1Version1->getProperties());
         static::assertSame(101, $restoredFile1Version1->getFocusPointX());
         static::assertSame(202, $restoredFile1Version1->getFocusPointY());
-        static::assertSame('de', $restoredFile1Version1->getDefaultMeta()->getLocale());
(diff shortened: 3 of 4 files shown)


alexander-schranz:
@matthiasnoback Hello 👋 ,

> Technically (and in our database in practice too!), the default file meta is not necessarily available for every media file.

That would end in other errors in some cases also if the media would not have the expected default meta, so my question is how that media did end up in that strange unexpected state? How were the media entities created in your case, was the media manager service used or was it a manual import where the default meta was missed? How did the medias ended up in that strange state without default meta?

I'm thinking if we should log here an error or throw a better exception as the media seems to be in broken unexpected state from Sulu perspective and currently not see how in the core a media could endup in that broken state.

matthiasnoback:
Thanks for your reply! I understand. I have looked up the code that I used to import an existing collection of files into the CMS, and indeed, I used `MediaManagerInterface::save()`. That seemed like the most logical way to do it. However, I understand that if you do it from the regular controller, then it would automatically have a default meta.
I haven't run into other issues, not having a default meta is no problem so far. Do you suggest fixing the data in the database, adding a default meta?

matthiasnoback:
Hmm, I just realized we do have another issue that could be related: deleting media sometimes doesn't work, deleting the stored file, but not the Media object.

alexander-schranz:
>  I have looked up the code that I used to import an existing collection of files into the CMS, and indeed, I used MediaManagerInterface::save()

The MediaManager sure sounds like the correct way here, do you have a short code example which produces a media without a default meta? Does the media have in the UI also no title in your case? Maybe we need here better handling of the `'title'` / default meta creation inside the Manager.

matthiasnoback:
I used this code:

```php
$uploadedFile = new UploadedFile($tempFilePath, $fileName);
$media = $this->mediaManager->save($uploadedFile, [
    'locale' => 'nl',
    'collection' => $collectionId,
], 1);
```

Looking at `MediaManager::createMedia()` it sets the default meta `$fileVersion->setDefaultMeta($fileVersion->getMeta()->first() ?: null);`, which assumes at least one meta has been provided, but this only happens if there is a `'title'` key in the array, because `setTitle()` calls `getMeta()` with `true`, so it would be created.

alexander-schranz:
I see I think we should move this logic to the media manager: https://github.com/sulu/sulu/blob/972e74f602aefe9f4bafdcbb6b999554eb77da4d/src/Sulu/Bundle/MediaBundle/Controller/AbstractMediaController.php#L46-L59 if its a new media (without id) and no `'title'` is given. Else we endup in such import scripts with a lot of unnamed medias, specially if it are documents instead of images this would be end in a UX issue as documents has no previews.

Unsure how we handle it with existing messed up data. Was the `new version` / `__clone` currently the only case where you did run into such issues? I think we atleast need check smart content media if sorted by title.

matthiasnoback:
There is also an issue with deleting media that has no default meta. The file is removed from storage, but moving it to the trash doesn't work, because somewhere it also calls `getDefaultMeta()->getLocale()`. That's it, as far as I know. I think I thought it wasn't necessary to have a default meta, since its return type is nullable. Anyway, tackling it at the start is the best option I'd say, when you call `save()` without a `"title"` key should not be possible.

alexander-schranz:
@matthiasnoback make sense, do you want to add them also.

We should add here a check: https://github.com/sulu/sulu/blob/120c5efbbc7770a6206fb11910cc2ec2d236f32a/src/Sulu/Bundle/MediaBundle/Trash/MediaTrashItemHandler.php#L117

And here https://github.com/sulu/sulu/blob/120c5efbbc7770a6206fb11910cc2ec2d236f32a/src/Sulu/Bundle/MediaBundle/Trash/MediaTrashItemHandler.php#L265-L267

```
if ($metaData['locale'] === $fileVersionData['defaultMetaLocale']
    || !isset($fileVersionData['defaultMetaLocale'])
) {
```

alexander-schranz:
@matthiasnoback I adjusted it so you not longer should run into 500 errors. But for UX reasons I still recommend you to create a defaultMeta always so the UI doesn't look empty. The https://github.com/sulu/sulu/pull/8348 should handle that now.

Thx for the report and PR!

matthiasnoback:
Hi @alexander-schranz ! Sorry for the delay, I was busy with some Fortran things. Thanks for making these changes, looks great. And thanks (to everyone there :)) for Sulu. It has been very useful. If you like, I can share some detailed feedback, and examples of generic tooling I made to reduce code duplication in Admin classes and Admin controllers. You can send your email address to me in a private message.

matthiasnoback:
Oh, wait, it's on your GitHub profile hehe.

PR #8335, opened 2025-11-12
Add test to profile queries on content resolving

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| License | MIT

#### What's in this PR?

Adds a test to assert queries for nested entities during the resolve process.

#### Why?

To prevent regressions and performance issues that are not detected.

### DIFF
--- a/packages/content/tests/Functional/Integration/ExampleProfilerWebsiteTest.php
+++ b/packages/content/tests/Functional/Integration/ExampleProfilerWebsiteTest.php
@@ -1,0 +1,165 @@
+<?php
+
+/*
+ * This file is part of Sulu.
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Content\Tests\Functional\Integration;
+
+use Doctrine\Bundle\DoctrineBundle\DataCollector\DoctrineDataCollector;
+use Sulu\Bundle\TestBundle\Testing\SuluTestCase;
+use Sulu\Content\Tests\Functional\Traits\CreateCategoryTrait;
+use Sulu\Content\Tests\Functional\Traits\CreateMediaTrait;
+use Sulu\Content\Tests\Functional\Traits\CreateTagTrait;
+use Sulu\Content\Tests\Traits\CreateExampleTrait;
+use Symfony\Bundle\FrameworkBundle\KernelBrowser;
+
+class ExampleProfilerWebsiteTest extends SuluTestCase
+{
+    use CreateCategoryTrait;
+    use CreateExampleTrait;
+    use CreateMediaTrait;
+    use CreateTagTrait;
+
+    private KernelBrowser $client;
+
+    protected function setUp(): void
+    {
+        self::purgeDatabase();
+    }
+
+    public function testResolveNestedResolvableResources(): void
+    {
+        $collection1 = self::createCollection(['title' => 'collection-1', 'locale' => 'en']);
+        $mediaType = self::createMediaType(['name' => 'Image', 'description' => 'This is an image']);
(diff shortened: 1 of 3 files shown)


packages/content/tests/Functional/Integration/ExampleProfilerWebsiteTest.php
```diff
@@ -0,0 +1,186 @@
...
+
+        $queriesData = $dbCollector->getQueries();
+        $this->assertArrayHasKey('default', $queriesData);
+
+        /** @var list<array{sql: string}> $queries */
+        $queries = $queriesData['default'];
+
+        // Assert total query count - this is the primary efficiency assertion
+        // Ensures no N+1 query problems when loading nested resolvable resources
+        self::assertCount(10, $queries, 'Total query count should remain at 10 to ensure no N+1 queries');
```
alexander-schranz: remove total check this makes failing test just hard to know whats going wrong.

packages/content/tests/Functional/Integration/ExampleProfilerWebsiteTest.php
```diff
@@ -0,0 +1,186 @@
...
+        // Media should appear only once despite multiple references across nested examples
+        self::assertSame(1, $tableCounts['me_media'] ?? 0, 'Media table should be queried only once (batch loading)');
+        self::assertSame(1, $tableCounts['me_files'] ?? 0, 'Media files should be joined only once');
+        self::assertSame(1, $tableCounts['me_file_versions'] ?? 0, 'Media file versions should be joined only once');
+        self::assertSame(1, $tableCounts['me_collections'] ?? 0, 'Media collections should be joined only once');
+        self::assertSame(1, $tableCounts['me_media_types'] ?? 0, 'Media types should be joined only once');
+
+        // Analytics queried once at the end
+        self::assertSame(1, $tableCounts['we_analytics'] ?? 0, 'Analytics should be queried once');
+        self::assertSame(1, $tableCounts['we_domains'] ?? 0, 'Analytics domains should be joined once');
```
alexander-schranz: Combine it into a single `assertSame` call:

```
$this->assertSame([
    // ...
    'ro_routes' => 2,
    // Analytics queried once at the end
    'we_analytics' => 1,
    'we_domains' => 1,
],  $tableCounts);`

File: packages/content/tests/Functional/Integration/ExampleProfilerWebsiteTest.php [hunk 3]
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    /**
+     * Extracts all table names from a SQL query (FROM and JOIN clauses).
+     *
+     * @return list<string>
+     */
+    private function extractTablesFromQuery(string $sql): array
+    {
+        \preg_match_all('/(?:FROM|JOIN)\s+([a-z_]+)\s+[a-z0-9_]/i', $sql, $matches);
```
alexander-schranz: why `JOIN` also?

alexander-schranz: If we combine from and join I would not do any counts and do just:

```php
$this->asserSame([
    // ..
    'test_example_dimension_contents.ta_tags.ca_categories',
    'ro_routes',
    'we_analytics.we_domains',
], $queryTableWithJoins);
```

alexander-schranz (changes requested):

packages/content/tests/Functional/Integration/ExampleProfilerWebsiteTest.php
Prokyonn: I thought it would make sense to also see the effect when there are suddenly more joins, but we can also remove it if you don't think it has any value

Prokyonn: not sure if I understood your assertion correctly, but it should now be implemented like this 😬

alexander-schranz (approved):

PR #8341, opened 2025-11-13
Fix links for Articles and Pages in the Admin Search

martinlagler:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Related issues/PRs | #7672
| License | MIT

#### What's in this PR?

Fix links for Articles and Pages in the Admin Search.


packages/search/src/Infrastructure/Symfony/HttpKernel/SuluSearchBundle.php
```diff
@@ -55,6 +55,11 @@ public function configure(DefinitionConfigurator $definition): void
                                     ->isRequired()
                                     ->cannotBeEmpty()
                                 ->end()
+                                ->arrayNode('resultToRouteName')
```
martinlagler, line 58: @alexander-schranz Can you please take a look at this config 😅 
It works, but I think it should be better formatted...

packages/search/assets/js/containers/Search/Search.js
```diff
@@ -65,20 +66,23 @@ class Search extends React.Component<Props> {
...
-        const resultId = result.resourceId || null;
-        const resultLocale = result.locale || null;
-        const resultParams = {
-            id: resultId,
-            locale: resultLocale,
-        };
+
+        let route = routeName;
+
+        if (resultToRouteName) {
```
martinlagler, line 75: Should I remove this if and add an empty `resultToRouteName` to the tests ?

packages/search/src/Infrastructure/Symfony/HttpKernel/SuluSearchBundle.php
alexander-schranz: confused by the two `end` 🤔 need to check

packages/search/assets/js/containers/Search/Search.js
alexander-schranz: fine with it as it maybe is not set.

alexander-schranz:
@martinlagler can you check the flow error think new param is missing in the definition.

packages/search/src/Infrastructure/Symfony/HttpKernel/SuluSearchBundle.php
```diff
@@ -55,6 +55,11 @@ public function configure(DefinitionConfigurator $definition): void
...
+                                    ->useAttributeAsKey('name')
+                                    ->scalarPrototype()
+                                    ->end()
```
alexander-schranz:
```suggestion
                                    ->scalarPrototype()->end()
```

from the docs: https://symfony.com/doc/current/components/config/definition.html#adding-node-definitions-to-the-tree

PR #8343, opened 2025-11-13
Use snapshot version of composer when test against dev versions

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no 
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Use snapshot version of composer when test against dev versions.

#### Why?

We test `@dev` against our dependencies but not composer itself. This way we can give composer early feedback and avoid issues in stable releases like:  https://github.com/composer/composer/discussions/12603

### DIFF
--- a/.github/workflows/test-application.yaml
+++ b/.github/workflows/test-application.yaml
@@ -191,7 +191,7 @@ jobs:
                       phpcr-transport: doctrinedbal
                       dependency-versions: 'highest'
                       php-extensions: 'ctype, iconv, mysql, imagick'
-                      tools: 'composer:v2'
+                      tools: 'composer:snapshot'
                       composer-stability: 'dev'
                       env:
                           SYMFONY_DEPRECATIONS_HELPER: weak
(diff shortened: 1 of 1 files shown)


.github/workflows/test-application.yaml
```diff
@@ -241,8 +241,12 @@ jobs:
               if: ${{ matrix.composer-stability }}
               run: composer config minimum-stability ${{ matrix.composer-stability }}
 
+            - name: Use composer snapshot version
+              if: ${{ matrix.composer-stability == 'dev' }}
+              run: composer self-update --snapshot
```
Seldaek: You can also do this as `tools: composer:snapshot` in shivammathur/setup-php.. so I think you can set that in line 194 instead, it'd be more efficient probably :)

alexander-schranz: @Seldaek didn't know about that. Awesome thx!

Prokyonn (approved):

PR #8345, opened 2025-11-14
Add support for DescendantProviderInterface for pages

Prokyonn:
| Q | A
  | --- | ---
  | Bug fix? | no
  | New feature? | yes
  | BC breaks? | no
  | Deprecations? | no
  | Related issues/PRs | #7672 
  | License | MIT

  #### What's in this PR?

  This PR implements `DescendantProviderInterface` support for pages, adding:
  - Permission checks for all descendant pages when deleting a parent page
  - Warning dialog when attempting to delete pages with children (requires `force=true` to proceed)
  - `findDescendantIdsById()` method in PageRepository for querying page hierarchies

  #### Why?

Missing feature from Sulu 3.0

### DIFF
--- a/packages/page/src/Application/MessageHandler/RemovePageMessageHandler.php
+++ b/packages/page/src/Application/MessageHandler/RemovePageMessageHandler.php
@@ -12,22 +12,15 @@
 namespace Sulu\Page\Application\MessageHandler;
 
 use Sulu\Bundle\ActivityBundle\Application\Collector\DomainEventCollectorInterface;
-use Sulu\Bundle\SecurityBundle\System\SystemStoreInterface;
 use Sulu\Bundle\TrashBundle\Application\TrashManager\TrashManagerInterface;
-use Sulu\Component\Rest\Exception\InsufficientDescendantPermissionsException;
-use Sulu\Component\Security\Authentication\UserInterface;
-use Sulu\Component\Security\Authorization\AccessControl\AccessControlRepositoryInterface;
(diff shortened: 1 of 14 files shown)


packages/page/src/Application/MessageHandler/RemovePageMessageHandler.php
```diff
@@ -28,9 +37,16 @@
...
+    /**
+     * @param array<string, int> $permissions
+     */
     public function __construct(
         private PageRepositoryInterface $pageRepository,
         private DomainEventCollectorInterface $domainEventCollector,
+        private AccessControlRepositoryInterface $accessControlRepository,
+        private SystemStoreInterface $systemStore,
+        private ?Security $security,
+        private array $permissions,
```
alexander-schranz: Unsure handle this in the application layer as its unrelated to page domain, as security is more related to UserInterface or Infrastracture Layer 🤔 .

Prokyonn: > Unsure handle this in the application layer as its unrelated to page domain, as security is more related to UserInterface or Infrastracture Layer 🤔 .

Yeah, I wasn’t completely sure about it either. It also feels a bit odd in the controller.

alexander-schranz: as discussed we move to request listener

alexander-schranz (approved):

PR #8356, opened 2025-11-16
Remove deprecated preview cache

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | maybe
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?
Removing the deprecated `PreviewCache` class.

#### Why?
This class was only there for Backwards compatibility with two cache systems, doctrine's cache and Symfony's cache. Since doctrine cache is deprecated we can just assume a Symfony cache here.


src/Sulu/Bundle/PreviewBundle/Preview/Preview.php
```diff
@@ -83,15 +73,15 @@ public function stop(string $token): void
             return;
         }
 
-        $this->cache->delete($token);
+        $this->cache->deleteItem($token);
```
alexander-schranz, line 76: interesting did the CacheItemPoolInterface supported both delete and deleteItem? or is it maybe a bug which we need tackle in 2.6? Same for contains / hasItem?

Same for the other changes?

mamazu: The method name `delete` only exists on Sulu's `PreviewCache class which then abstracted away the multiple different cache interfaces we support.

alexander-schranz (approved):

alexander-schranz:
Can `doctrine/cache` be removed from the `composer.json` now?

mamazu:
Sadly not. There are still other parts of the code that depend on this. But I will create some more PRs to tackle those later.

alexander-schranz:
@mamazu Thank you!

PR #8357, opened 2025-11-17
Add support for url reference types in new route generator

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? |  yes
| BC breaks? | no 
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Add support for route reference types in Url Generators. https://github.com/symfony/symfony/blob/4d91dfc4501c0b2e21caa0ea4eb8a8dacd618872/src/Symfony/Component/Routing/Generator/UrlGeneratorInterface.php#L34-L55

#### Why?

Usage to clear the http cache via URL.

### DIFF
--- a/packages/route/src/Application/Routing/Generator/RouteGeneratorInterface.php
+++ b/packages/route/src/Application/Routing/Generator/RouteGeneratorInterface.php
@@ -12,11 +12,12 @@
 namespace Sulu\Route\Application\Routing\Generator;
 
 use Sulu\Route\Domain\Exception\MissingRequestContextParameterException;
+use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 
 interface RouteGeneratorInterface
 {
     /**
      * @throws MissingRequestContextParameterException
      */
-    public function generate(string $slug, ?string $locale = null, ?string $site = null): string;
+    public function generate(string $slug, ?string $locale = null, ?string $site = null, int $referenceType = UrlGeneratorInterface::ABSOLUTE_PATH): string;
 }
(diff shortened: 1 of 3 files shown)


packages/route/src/Application/Routing/Generator/RouteGeneratorInterface.php
```diff
@@ -12,11 +12,12 @@
...
 use Sulu\Route\Domain\Exception\MissingRequestContextParameterException;
+use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 
 interface RouteGeneratorInterface
 {
     /**
      * @throws MissingRequestContextParameterException
      */
-    public function generate(string $slug, ?string $locale = null, ?string $site = null): string;
+    public function generate(string $slug, ?string $locale = null, ?string $site = null, int $referenceType = UrlGeneratorInterface::ABSOLUTE_PATH): string;
```
alexander-schranz, line 22: Unsure about have to reference a `Symfony Constant` in our own Domain / Application Interface.

Alternative is we go with `bool $relative = true` / `bool $absolute = false`.
`bool $relative = true` would with `(bool) UrlGeneratorInterface::ABSOLUTE_URL`.

@Prokyonn what do you think?

Prokyonn: My gut feeling would be to go with the Symfony constant 🤔 
To be more decoupled we could also introduce our own constant/enum that references the symfony constant.

On the other hand, using the interface would implicitly say that we also must support all 4 types.

Prokyonn (approved):

packages/route/src/Application/Routing/Generator/RouteGeneratorInterface.php
alexander-schranz: as discussed we will go with the symfony constants.

PR #8359, opened 2025-11-17
Use absolute urls for cache invalidation

Prokyonn:
| Q | A
  | --- | ---
  | Bug fix? | yes
  | New feature? | no
  | BC breaks? | no
  | Deprecations? | no
  | License | MIT

  #### What's in this PR?

  This PR fixes cache invalidation failures when using Varnish by resolving Route slugs to full URLs
  with hosts before invalidating cache.

  #### Why?

Sulu 3.0's cache invalidation throws `FOS\HttpCache\Exception\MissingHostException`

  To Do
-  [ ] Test manually with Varnish configuration

### DIFF
--- a/packages/article/src/Infrastructure/Sulu/HttpCache/EventSubscriber/ArticleCacheInvalidationSubscriber.php
+++ b/packages/article/src/Infrastructure/Sulu/HttpCache/EventSubscriber/ArticleCacheInvalidationSubscriber.php
@@ -18,6 +18,8 @@ use Sulu\Article\Domain\Event\ArticleWorkflowTransitionAppliedEvent;
 use Sulu\Article\Domain\Model\ArticleDimensionContentInterface;
 use Sulu\Article\Domain\Model\ArticleInterface;
 use Sulu\Bundle\HttpCacheBundle\Cache\CacheManagerInterface;
+use Sulu\Component\Webspace\Manager\WebspaceManagerInterface;
+use Sulu\Component\Webspace\Webspace;
 use Sulu\Content\Application\ContentAggregator\ContentAggregatorInterface;
 use Sulu\Content\Domain\Exception\ContentNotFoundException;
 use Sulu\Content\Domain\Model\WorkflowInterface;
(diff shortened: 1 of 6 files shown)


packages/article/src/Infrastructure/Sulu/HttpCache/EventSubscriber/ArticleCacheInvalidationSubscriber.php
```diff
@@ -88,7 +90,16 @@ private function invalidateArticlePaths(string $uuid, ?string $locale): void
         ]);
 
         foreach ($routes as $route) {
-            $this->cacheManager->invalidatePath($route->getSlug());
+            $urls = $this->webspaceManager->findUrlsByResourceLocator(
```
alexander-schranz: we should go over the `RouteGeneratorInterface` here. Requires: https://github.com/sulu/sulu/pull/8357

packages/article/src/Infrastructure/Sulu/HttpCache/EventSubscriber/ArticleCacheInvalidationSubscriber.php
```diff
@@ -88,7 +91,14 @@ private function invalidateArticlePaths(string $uuid, ?string $locale): void
         ]);
 
         foreach ($routes as $route) {
-            $this->cacheManager->invalidatePath($route->getSlug());
+            $url = $this->routeGenerator->generate(
+                $route->getSlug(),
+                $route->getLocale(),
+                $route->getSite(),
```
alexander-schranz: `site` is empty for articles. This needs keep the configured `mainWebspace` and `additionalWebspace` in mind.

alexander-schranz: As discussed with @chirimoya. we should do `invalidatePath` only if there is not tag based caching active.

In case of tag based caching disabled we clear all webspaces which have the current locale of that route activated. So we need inject the webspace manager here to get all webspaces.

So something like:

```php
// article was maybe previously in other webspace active so clear all webspaces with that locale
foreach ($webspace as $webspace) {
    $url = $this->routeGenerator->generate(
        $route->getSlug(),
        $route->getLocale(),
        $webspace->getKey(),
     );
}
```

Prokyonn: @alexander-schranz thinking about the routes, should we also invalidate history urls?

alexander-schranz: @Prokyonn Would currently not do that inside this service as URL changes are not longer part of the "publishing" process. Lets discuss that tomorrow have something in mind.

alexander-schranz (approved):

PR #8361, opened 2025-11-17
Fix duplicated unlocalized dimension on trash restore

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no yes
| New feature? | no yes
| BC breaks? | no yes
| Deprecations? | no yes 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### To Do

### DIFF
--- a/packages/article/tests/Functional/Integration/responses/article_post_restore.json
+++ b/packages/article/tests/Functional/Integration/responses/article_post_restore.json
@@ -26,7 +26,7 @@
         "Tag 4"
     ],
     "excerptTitle": "Excerpt Title 2",
-    "ghostLocale": "de",
+    "ghostLocale": "en",
     "id": "@string@",
     "image": null,
     "lastModified": "2022-05-08T00:00:00+00:00",
--- a/packages/content/src/Application/DimensionContentCollectionFactory/DimensionContentCollectionFactory.php
+++ b/packages/content/src/Application/DimensionContentCollectionFactory/DimensionContentCollectionFactory.php
@@ -14,44 +14,49 @@ declare(strict_types=1);
 namespace Sulu\Content\Application\DimensionContentCollectionFactory;
 
 use Sulu\Content\Application\ContentDataMapper\ContentDataMapperInterface;
+use Sulu\Content\Application\ContentMetadataInspector\ContentMetadataInspectorInterface;
 use Sulu\Content\Domain\Factory\DimensionContentCollectionFactoryInterface;
 use Sulu\Content\Domain\Model\ContentRichEntityInterface;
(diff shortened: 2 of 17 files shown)


packages/content/src/Application/DimensionContentCollectionFactory/DimensionContentCollectionFactory.php
alexander-schranz: do we really need this still?

Prokyonn: In most cases probably not 🤔 should we have it as a fallback (e.g. \count($page->getDimensionContent()) === 0) or would you completely remove it?

Prokyonn: But we would always overfetch, as we do not filter by versions, when we completely rely on the `$contentRichEntity->getDimensionContent()` method

alexander-schranz: Yeah think we need rafactor that part more as now we load all versions always if somebody trigger this methods without a lazy collection.

packages/content/src/Domain/Model/DimensionContentCollection.php
```diff
@@ -48,23 +47,21 @@ class DimensionContentCollection implements \IteratorAggregate, DimensionContent
...
      * DimensionContentCollection constructor.
      *
-     * @param array<int, T> $dimensionContents
+     * @param array<int, T>|Collection<int, T> $dimensionContents
      * @param mixed[] $dimensionAttributes
      * @param class-string<T> $dimensionContentClass
      */
     public function __construct(
-        array $dimensionContents,
+        array|Collection $dimensionContents,
```
Prokyonn, line 55: @alexander-schranz should we allow only Collection here ?

alexander-schranz: I think its okay to remove it.

packages/article/tests/Functional/Integration/responses/article_post_restore.json
```diff
@@ -26,7 +26,7 @@
         "Tag 4"
     ],
     "excerptTitle": "Excerpt Title 2",
-    "ghostLocale": "de",
+    "ghostLocale": "en",
```
alexander-schranz, line 29: expected?

Prokyonn: yes, we remove `en` and then restore again `en` so the `ghostLocale` should be same as `locale` as we do not need a fallback

alexander-schranz (approved):

PR #8362, opened 2025-11-17
Add some more type improvements for Admin and some more classes (part 2)

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | #7558 
| License | MIT
| Documentation PR | -

#### What's in this PR?
Upmerging #7558 

#### Why?
The classes have been moved and got lost when upmerging.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/FormMetadata.php
+++ b/src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/FormMetadata.php
@@ -23,32 +23,23 @@ class FormMetadata extends AbstractMetadata
      * @var array<string, string>
      */
     #[Exclude]
-    private $titles = [];
+    private array $titles = [];
 
     /**
      * @var ItemMetadata[]
      */
     #[SerializedName('form')]
     #[Exclude(if: "'admin_form_metadata_keys_only' in context.getAttribute('groups')")]
-    private $items = [];
+    private array $items = [];
 
-    /**
-     * @var SchemaMetadata
-     */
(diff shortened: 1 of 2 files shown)


src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/FormMetadata.php
mamazu, line 1: There's also a deprecation in this file that doesn't make sense in line 78
```php
    /**
     * @deprecated since 3.0, use setName() instead
     */
    #[VirtualProperty]
    #[SerializedName('name')]
    #[Exclude(if: "'admin_form_metadata_keys_only' in context.getAttribute('groups')")]
    public function getName(): string
    {
        return $this->key;
    }
```

alexander-schranz: should be `use getKey() instead` frontend still is build ontop of .name instead .key

src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/FormMetadata.php
```diff
@@ -148,6 +156,8 @@ public function getItems(): array
 
     /**
      * @param ItemMetadata[] $items
+     *
+     * @return void
```
alexander-schranz: we can add native return types here in 3.0

alexander-schranz:
@mamazu thank you!

PR #8364, opened 2025-11-18
Add 404 response on missing entity

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| License | MIT

#### What's in this PR?

Adds a 404 response from the EntityControllers to prevent frontend field-types to crash.

#### Why?

Frontend field types like selections did completely crash when a 500 error was returned. E.g. when a used parent-page was removed.

### DIFF
--- a/packages/article/src/UserInterface/Controller/Admin/ArticleController.php
+++ b/packages/article/src/UserInterface/Controller/Admin/ArticleController.php
@@ -18,9 +18,11 @@ use Sulu\Article\Application\Message\ModifyArticleMessage;
 use Sulu\Article\Application\Message\RemoveArticleMessage;
 use Sulu\Article\Application\Message\RemoveArticleTranslationMessage;
 use Sulu\Article\Application\Message\RestoreArticleVersionMessage;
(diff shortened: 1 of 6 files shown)


packages/article/src/UserInterface/Controller/Admin/ArticleController.php
```diff
@@ -169,18 +170,22 @@ public function getAction(Request $request, string $id): Response // TODO route
...
                 ],
-                $dimensionAttributes,
-            ),
-            [
-                ArticleRepositoryInterface::GROUP_SELECT_ARTICLE_ADMIN => true,
-            ],
-        );
+            );
+        } catch (ArticleNotFoundException) {
+            return new Response('', 404);
```
Prokyonn: Needs to be adjusted (see Sulu 2.6 EntityNotFoundException)

alexander-schranz (approved):

alexander-schranz:
A test would be nice :)

PR #8365, opened 2025-11-18
Set author on entity creation

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| License | MIT

#### What's in this PR?

Adds the author/authoredDate on creation.

#### Why?

This was the default behaviour of Sulu 2.6


packages/article/src/Application/MessageHandler/CreateArticleMessageHandler.php
```diff
@@ -30,13 +32,28 @@ public function __construct(
         private ArticleRepositoryInterface $articleRepository,
         /** @var iterable<ArticleMapperInterface> */
         private iterable $articleMappers,
-        private DomainEventCollectorInterface $domainEventCollector
+        private DomainEventCollectorInterface $domainEventCollector,
+        private ?TokenStorageInterface $tokenStorage = null,
```
alexander-schranz, line 36: not sure about this, but let keep it to fix the issue.

Prokyonn: What would be the alternative way to get the logged in user ?

PR #8368, opened 2025-11-18
Fix store/restore translation for page/article/snippet

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no 
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Add store and restore of single translations.

#### Why?

Restore of single translation not working.

#### To Do


packages/page/src/Infrastructure/Sulu/Trash/PageTrashItemHandler.php
alexander-schranz, line 224: If early return we can remove the else:

```suggestion

            return $page;
        }

        $this->domainEventCollector->collect(new PageRestoredEvent(
            $page,
            $pageTitle,
            $context,
            $restoreData,
        ));
```

alexander-schranz (changes requested):

PR #8369, opened 2025-11-19
Fix vanishing many to many relations

Prokyonn:
| Q | A
  | --- | ---
  | Bug fix? | yes
  | New feature? | no
  | BC breaks? | no
  | Deprecations? | no
  | License | MIT

  #### What's in this PR?

  This PR fixes a critical bug in the dimension content architecture where excerpt tags, categories,
  and audience target groups were being lost from the original draft (version=0) dimension content
  after publishing a snippet, article, or page.

  #### Why?

  When publishing content (snippets, articles, pages), the system creates multiple dimension content
  records:
  1. Original draft (version=0) - should keep its excerpt data
  2. Versioned draft (version=timestamp) - copy of the data
  3. Live content (version=0) - published copy

  However, the **original draft was losing its excerpt tags, categories, and target groups** in the
  database after publish.

  **Root Cause:**

  During the publish workflow, when copying data to new dimension contents:
  1. The same Tag/Category entity instances are shared across multiple dimension contents (from
  TagFactory)
  2. When setExcerptTags() is called on a new dimension content, clear() removes ALL associations from
   Doctrine's Unit of Work
  3. This inadvertently deletes join table entries for the original draft dimension content

### DIFF
--- a/packages/content/src/Domain/Model/ExcerptTrait.php
+++ b/packages/content/src/Domain/Model/ExcerptTrait.php
@@ -125,10 +125,17 @@ trait ExcerptTrait
     public function setExcerptCategories(array $excerptCategories): void
     {
         $this->initializeCategories();
-        $this->excerptCategories->clear();
 
-        foreach ($excerptCategories as $excerptCategory) {
-            $this->excerptCategories->add($excerptCategory);
+        foreach ($this->excerptCategories as $existingCategory) {
+            if (!\in_array($existingCategory, $excerptCategories, true)) {
+                $this->excerptCategories->removeElement($existingCategory);
+            }
+        }
+
+        foreach ($excerptCategories as $newCategory) {
+            if (!$this->excerptCategories->contains($newCategory)) {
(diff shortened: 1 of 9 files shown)


alexander-schranz (approved):

packages/content/src/Domain/Model/ExcerptTrait.php
```diff
@@ -125,10 +125,17 @@ public function getExcerptCategories(): array
     public function setExcerptCategories(array $excerptCategories): void
     {
         $this->initializeCategories();
-        $this->excerptCategories->clear();
```
alexander-schranz, line 128: We should introduce a PHPStan rule for this clear makes in our setUp always problems 🙈 As second flush in our domain event listener (which is on postFlush event) has the unit of work of previous flush not completely cleaned up.

alexander-schranz:
@Prokyonn merge as soon as you are fine with the tests. I updated the snapshots.

PR #8372, opened 2025-11-19
Disallow call of doctrine collection clear method

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no 
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | https://github.com/sulu/sulu/pull/8369
| License | MIT

#### What's in this PR?

Disallow call of doctrine collection clear method

#### Why?

Calling `clear` actually triggers a `DELETE ..` query which removes all related entities, as in postFlush the UnitOfWork is not 100% cleanup it ends up in retrigger the delete query even if the values has changed. So basically in a Sulu Setup we should not allow calling the clear method of a already persisted collection.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -157,6 +157,7 @@
         "scheb/2fa-totp": "^6.0 || ^7.0",
         "scheb/2fa-trusted-device": "^6.0 || ^7.0",
         "shipmonk/composer-dependency-analyser": "^1.8",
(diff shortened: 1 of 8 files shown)


Prokyonn (approved):

alexander-schranz:
Before merge into 3.0 we should merge: https://github.com/sulu/sulu/pull/8369

PR #8385, opened 2025-11-20
Remove doctrine cache in favor of Symfony cache

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | yes
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?
Removing the last remains of the doctrine cache integration

#### Why?
The bundle has been replaced by the Symfony version.

### DIFF
--- a/src/Sulu/Bundle/ContactBundle/Tests/Unit/ContactTwigExtensionTest.php
+++ b/src/Sulu/Bundle/ContactBundle/Tests/Unit/ContactTwigExtensionTest.php
@@ -11,8 +11,6 @@
 
 namespace Sulu\Bundle\ContactBundle\Tests\Unit;
 
-use Doctrine\Common\Cache\Cache;
-use Doctrine\Common\Cache\Psr6\DoctrineProvider;
 use PHPUnit\Framework\TestCase;
 use Prophecy\PhpUnit\ProphecyTrait;
 use Prophecy\Prophecy\ObjectProphecy;
@@ -25,15 +23,7 @@ class ContactTwigExtensionTest extends TestCase
 {
     use ProphecyTrait;
 
-    /**
-     * @var ContactTwigExtension
-     */
-    private $extension;
-
-    /**
-     * @var Cache
-     */
-    private $cache;
+    private ContactTwigExtension $extension;
 
     /**
      * @var ObjectProphecy<ContactRepository>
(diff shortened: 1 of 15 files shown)


src/Sulu/Bundle/ContactBundle/Twig/ContactTwigExtension.php
```diff
@@ -34,28 +35,21 @@ public function __construct(
...
     /**
-     * @param int $id id to resolve
+     * @param string $id id to resolve
      *
      * @return ContactInterface|null
+     *
+     * @throws InvalidArgumentException
      */
-    public function resolveContactFunction($id)
+    public function resolveContactFunction(string $id)
```
mamazu: This function is now strictly typed otherwise it would fail when getting the cache.

alexander-schranz: did `null` previously also fail?

mamazu: Yes, there is also a string constraint on the id. Since this class doesn't have strict_types any int should be fine as well.

alexander-schranz: but think `int` is correct and what is actually given into it?

Does the cache require string then we should maybe cast it for the cache id

```php
        return $this->cache->get((string) $id, function() use ($id) {
            return $this->contactRepository->find($id);
        });
```

src/Sulu/Bundle/ContactBundle/Tests/Unit/ContactTwigExtensionTest.php
```diff
@@ -61,16 +56,16 @@ public function testResolveContactFunction(): void
         $this->contactRepository->find(1)->willReturn($contact1);
         $this->contactRepository->find(2)->willReturn($contact2);
 
-        $contact = $this->extension->resolveContactFunction(1);
```
alexander-schranz, line 64: this does make more sense as mostly I would call it with a int not with a string.

src/Sulu/Bundle/ContactBundle/Twig/ContactTwigExtension.php
alexander-schranz: looks like you adopted this for UserTwigExtension but not yet for the ContactTwigExtension.

alexander-schranz (approved):

alexander-schranz:
@mamazu Thank you!

PR #8388, opened 2025-11-21
Add WebsiteReindexProviderVisitor to extract indexable content for article/page

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

- Adds a visitor interface to the WebsitePageReindexProvider and WebsiteArticleReindexProvider
- Adds a visitor that extracts the indexable data from the content for both entities

#### Why?

For website search the content should also be indexed.


packages/page/src/Infrastructure/Sulu/Search/WebsitePageReindexProvider.php
alexander-schranz: thought we returned here a Page and could give the Page Entity to the visitor 🤔 but seems like false marked.

Prokyonn: Yes unfortunately it is only the Page `phpstan-type`, I guess this was implemented without hydration to improve performance

packages/article/src/Infrastructure/Sulu/Search/Visitor/WebsiteArticleReindexContentVisitor.php
```diff
@@ -0,0 +1,225 @@
...
+    private function collectSearchableFields(array $items, array &$fields, string $prefix = ''): void
+    {
+        foreach ($items as $item) {
+            if (!($item instanceof FieldMetadata)) {
+                continue;
+            }
+
+            $hasSearchTag = false;
+            foreach ($item->getTags() as $tag) {
+                if ('sulu.search.field' === $tag->getName()) {
```
alexander-schranz: Can we make a constant for this here.

packages/page/src/Infrastructure/Sulu/Search/WebsitePageReindexProvider.php
Prokyonn: but we could also add a visitor to the query builder, than it would still be extendable also for custom adjustments on the page entity

packages/page/src/Infrastructure/Sulu/Search/Visitor/WebsitePageReindexProviderVisitorInterface.php
```diff
@@ -0,0 +1,31 @@
...
+{
+    /**
+     * Visit and manipulate the reindex data for a page dimension content.
+     *
+     * @param array<string, mixed> $result The raw query result containing page data
+     * @param array<string, mixed> $data
+     *
+     * @return array<string, mixed>
+     */
+    public function visit(array $result, array $data): array;
```
alexander-schranz:
```suggestion
    public function enhanceDocument(array $queryResult, array $document): array;
    
    public function enhanceQuery(QueryBuilder $queryBuilder): void;
```

packages/page/src/Infrastructure/Sulu/Search/Visitor/WebsitePageReindexProviderVisitorInterface.php
```diff
@@ -0,0 +1,31 @@
...
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Page\Infrastructure\Sulu\Search\Visitor;
+
+/**
+ * Interface for visitors that can process and manipulate reindex data
+ * for website page dimension content during search indexing.
+ */
+interface WebsitePageReindexProviderVisitorInterface
```
alexander-schranz:
```suggestion
interface WebsitePageReindexProviderEnhancerInterface
```

alexander-schranz (commented):
Did discuss with Thomas that maybe this make sense.

What do you think about?

packages/page/src/Infrastructure/Sulu/Search/Visitor/WebsitePageReindexProviderVisitorInterface.php
Prokyonn: Sounds good 👍

alexander-schranz (approved):

alexander-schranz:
@Prokyonn looks good 👍 Rebase required.

PR #8389, opened 2025-11-21
Update data types for new bundles

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? | no
| BC breaks? | yes
| Deprecations? | no yes 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Adds native php types instead of phpdoc types for entities in the refactored bundles.

#### Why?

Better type safety.

### DIFF
--- a/packages/custom-url/src/Domain/Model/CustomUrlRoute.php
+++ b/packages/custom-url/src/Domain/Model/CustomUrlRoute.php
@@ -33,7 +33,7 @@ class CustomUrlRoute implements CustomUrlRouteInterface
         $this->uuid = Uuid::v7()->__toString();
     }
 
-    public function getUuid(): ?string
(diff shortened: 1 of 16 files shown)


packages/custom-url/src/Domain/Model/CustomUrlRoute.php
```diff
@@ -33,7 +33,7 @@ public function __construct(
         $this->uuid = Uuid::v7()->__toString();
     }
 
-    public function getUuid(): ?string
+    public function getUuid(): string
```
alexander-schranz, line 36: good catch 👍

alexander-schranz (approved):

PR #8391, opened 2025-11-21
Add security context to admin search

martinlagler:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Related issues/PRs | #7672
| License | MIT

#### What's in this PR?

Add security context to admin search.


packages/article/src/Infrastructure/Sulu/Search/AdminArticleReindexProvider.php
```diff
@@ -96,6 +96,7 @@ public function provide(ReindexConfig $reindexConfig): \Generator
                 'metadata' => [
                     'group' => $groupIdentifier,
                 ],
+                'securityContext' => 'sulu.article.articles_' . ($groupIdentifier ?? 'default'),
```
alexander-schranz: Can we use the `ArticleAdmin::getArticleSecurityContext()` here

alexander-schranz (changes requested):
Else looks good 👍 

Any way we can also adopt the SearchControllerTest by give the test user a role with a security contexts?

martinlagler:
@alexander-schranz There is no Test for the Admin search controller 🙈

PR #8393, opened 2025-11-21
Add PHP 8.5 to CI

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Check Support for PHP 8.5

#### Example Usage

PHP 🐘 got a new version which we should support.

### DIFF
--- a/src/Sulu/Component/SmartContent/ContentType.php
+++ b/src/Sulu/Component/SmartContent/ContentType.php
@@ -326,14 +326,20 @@ class ContentType extends ComplexContentType implements ContentTypeExportInterfa
             return 1;
         }
 
-        $page = (int) $this->requestStack->getCurrentRequest()->get($pageParameter, 1);
+        $page = $this->requestStack->getCurrentRequest()->get($pageParameter, 1);
+
+        if (!\is_numeric($page)) {
+            return 1;
+        }
+
+        if ($page > \PHP_INT_MAX) { // needs to be done before cast e.g.: ?page=99999999999999999999
+            return \PHP_INT_MAX;
+        }
+
+        $page = (int) $page;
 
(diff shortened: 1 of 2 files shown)


src/Sulu/Component/SmartContent/ContentType.php
```diff
@@ -326,16 +326,22 @@ private function getCurrentPage($pageParameter)
...
             return \PHP_INT_MAX;
         }
 
+        $page = (int) $page;
+
+        if ($page <= 1) {
+            $page = 1;
+        }
+
         return $page;
```
alexander-schranz, line 345: This failed: https://github.com/sulu/sulu/blob/c1f3719473593faa3b2f2dfb5566ce3cd3886871/src/Sulu/Component/SmartContent/Tests/Unit/ContentTypeTest.php#L473 

With:

>  The float-string "99999999999999999999" is not representable as an int, cast occurred

alexander-schranz: Cast was introduced in https://github.com/sulu/sulu/pull/7881 by @martinlagler but we should cast it later after the max check.

mamazu: How does Symfony handle this with the `getInt` function on the request query bag?

alexander-schranz: > Parameter value "page" is invalid and flag "FILTER_NULL_ON_FAILURE" was not set.

Prokyonn (approved):

PR #8395, opened 2025-11-23
Remove optional constructor arguments for BC

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs |-
| License | MIT
| Documentation PR | -

#### What's in this PR?
They should have been removed long ago, but let's clean it up now.

#### Why?
This should be set from the container, there should be no reason why it wouldn't be.

### DIFF
--- a/src/Sulu/Bundle/SecurityBundle/Tests/Functional/Controller/UserControllerTest.php
+++ b/src/Sulu/Bundle/SecurityBundle/Tests/Functional/Controller/UserControllerTest.php
@@ -318,9 +318,7 @@ class UserControllerTest extends SuluTestCase
                 'email' => 'manager@test.com',
                 'password' => 'verysecurepassword',
                 'locale' => 'en',
-                'contact' => [
-                    'id' => $this->contact1->getId(),
-                ],
+                'contact' => $this->contact1->getId(),
             ]
         );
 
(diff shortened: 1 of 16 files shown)


src/Sulu/Bundle/SecurityBundle/UserManager/UserManager.php
```diff
@@ -476,7 +468,7 @@ private function addUserRole(UserInterface $user, $userRoleData)
     /**
      * Returns the contact with the given id.
      *
-     * @param int $id
+     * @param int|string|null $id
```
alexander-schranz: This should be kept as it is, unless what the callers currently does in core. To project outside it should only be called officially with int.

```suggestion
     * @param int $id
```

alexander-schranz (changes requested):

src/Sulu/Bundle/SecurityBundle/Tests/Functional/Controller/UserControllerTest.php
```diff
@@ -318,9 +318,7 @@ public function testPostWithEntireContactObject(): void
                 'email' => 'manager@test.com',
                 'password' => 'verysecurepassword',
                 'locale' => 'en',
-                'contact' => [
-                    'id' => $this->contact1->getId(),
-                ],
+                'contact' => $this->contact1->getId(),
```
alexander-schranz: the current user controller uses still `contact => ['id' => 1]` so this change I think would first require changing how the users controller serialize in general, before adopting any tests:

Image

mamazu:
I've reverted the whole contact situation. We can tackle that seperately.

PR #8406, opened 2025-11-25
Add native types to CategoryBundle

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | yes
| Deprecations? | no
| Related issues/PRs | depends on https://github.com/sulu/sulu/pull/8405
| License | MIT

#### What's in this PR?

#### Why?

Better developer experience.


src/Sulu/Bundle/CategoryBundle/Entity/Category.php
```diff
@@ -82,125 +61,125 @@ public function __construct()
...
     }
 
-    public function getTranslations()
+    public function getTranslations(): Collection
     {
         return $this->translations;
     }
 
-    public function findTranslationByLocale($locale)
+    public function findTranslationByLocale(?string $locale): CategoryTranslationInterface|false
```
alexander-schranz: @Prokyonn should we convert this to null via `->first() ?: null` so null safe operator `?->` can used on it?

src/Sulu/Bundle/CategoryBundle/Entity/Category.php
```diff
@@ -209,19 +188,19 @@ function(CategoryTranslationInterface $translation) use ($locale) {
...
-    public function addChild(CategoryInterface $child)
+    public function addChild(CategoryInterface $child): static
     {
         $this->children[] = $child;
 
         return $this;
     }
 
-    public function removeChild(CategoryInterface $child)
+    public function removeChild(CategoryInterface $child): void
```
alexander-schranz: would `return $this` (`: static`) in all methods which currently may are `void`

src/Sulu/Bundle/CategoryBundle/Entity/CategoryMeta.php
```diff
@@ -16,90 +16,72 @@
  */
 class CategoryMeta implements CategoryMetaInterface
```
alexander-schranz: what does this entity 🤔  ? is it editable by the UI?

src/Sulu/Bundle/TagBundle/Entity/Tag.php
```diff
@@ -90,7 +84,7 @@ public function setChanged(\DateTimeImmutable $changed): self
         return $this;
     }
 
-    public function __toString()
+    public function __toString(): string
     {
         return (string) $this->getName();
```
mamazu, line 89: Since getName is typed to a string, we don't need to cast here.

alexander-schranz: we need set atleast a default for `$this->name` as `__toString()` is always called by Doctrine if exception happening and then instead of showing that exception it shows null to string error: https://github.com/sulu/sulu/pull/7478

mamazu: Then maybe something like `return $this->name ?? 'unnamed';`?

alexander-schranz: would keep as before empty string, doctrine fallbacks to a object hash I think then like when no __toString exists which is better in the trace.

alexander-schranz:
@Prokyonn did discuss with @chirimoya that we can remove the unused `CategoryMeta` and `metas` from the category entity: https://github.com/sulu/sulu/blob/3.0/src/Sulu/Bundle/CategoryBundle/Entity/CategoryMeta.php

src/Sulu/Bundle/CategoryBundle/Category/KeywordManager.php
```diff
@@ -37,7 +37,7 @@ public function __construct(
     ) {
     }
 
-    public function save(KeywordInterface $keyword, CategoryInterface $category, $force = null)
+    public function save(KeywordInterface $keyword, CategoryInterface $category, $force = null, bool $isPersisted = false)
```
Prokyonn: @wachterjohannes remove isPersisted again (we use `isNew` from the entity directly)

src/Sulu/Bundle/CategoryBundle/Category/KeywordManager.php
```diff
@@ -48,7 +48,7 @@ public function save(KeywordInterface $keyword, CategoryInterface $category, $fo
             throw new KeywordIsMultipleReferencedException($keyword);
         }
 
-        if (null !== $keyword->getId()
+        if ($isPersisted
```
Prokyonn: replace with $keyword->isNew

src/Sulu/Bundle/CategoryBundle/Category/KeywordManager.php
```diff
@@ -63,7 +63,7 @@ public function save(KeywordInterface $keyword, CategoryInterface $category, $fo
 
         /** @var CategoryTranslationInterface $categoryTranslation */
         foreach ($keyword->getCategoryTranslations() as $categoryTranslation) {
-            $event = $keyword->getId()
+            $event = $isPersisted
```
Prokyonn: replace with $keyword-isNew

src/Sulu/Bundle/CategoryBundle/Category/KeywordManagerInterface.php
```diff
@@ -32,7 +32,7 @@ interface KeywordManagerInterface
      *
      * @return KeywordInterface
      */
-    public function save(KeywordInterface $keyword, CategoryInterface $category, $force = null);
+    public function save(KeywordInterface $keyword, CategoryInterface $category, $force = null, bool $isPersisted = false);
```
Prokyonn: @wachterjohannes remove isPersisted again (we use `isNew` from the entity directly)

wachterjohannes (approved):

PR #8410, opened 2025-11-25
Add native types to ContactBundle entities and interfaces

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | yes
| Deprecations? | no
| License | MIT

#### What's in this PR?

#### Why?

Better developer experience.

### DIFF
--- a/src/Sulu/Bundle/ContactBundle/Controller/PositionController.php
+++ b/src/Sulu/Bundle/ContactBundle/Controller/PositionController.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 /*
  * This file is part of Sulu.
  *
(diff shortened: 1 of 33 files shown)


src/Sulu/Bundle/ContactBundle/Controller/PositionController.php
```diff
@@ -177,6 +177,7 @@ public function cdeleteAction(Request $request)
                 }
 
                 $positionId = $position->getId();
+                \assert(null !== $positionId);
```
wachterjohannes: no asserts in controller

src/Sulu/Bundle/ContactBundle/Controller/PositionController.php
```diff
@@ -215,6 +216,7 @@ public function deleteAction($id)
...
```
wachterjohannes: no asserts in controller

wachterjohannes (changes requested):

PR #8411, opened 2025-11-25
Add native types to SecurityBundle entities and interfaces

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | yes
| Deprecations? | no
| License | MIT

#### What's in this PR?

#### Why?

Better developer experience.

### DIFF
--- a/src/Sulu/Bundle/SecurityBundle/Entity/User.php
+++ b/src/Sulu/Bundle/SecurityBundle/Entity/User.php
@@ -51,7 +51,7 @@ class User extends ApiEntity implements UserInterface, EquatableInterface, Audit
 
     #[Expose]
     #[Groups(['frontend', 'fullUser', 'profile'])]
-    protected string $locale = '';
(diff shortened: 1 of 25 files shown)


src/Sulu/Bundle/SecurityBundle/Entity/User.php
```diff
@@ -34,113 +35,58 @@ class User extends ApiEntity implements UserInterface, EquatableInterface, Audit
...
-    protected $password;
+    protected string $password = '';
 
-    /**
-     * @var string
-     */
     #[Expose]
     #[Groups(['frontend', 'fullUser', 'profile'])]
-    protected $locale;
+    protected string $locale = '';
```
alexander-schranz:
```suggestion
    protected string $locale = 'en';
```

alexander-schranz (approved):

PR #8426, opened 2025-12-01
Select SocialMediaProfiles during initial select

MarkusHolstein:
| Q | A
| --- | ---
| Bug fix? | somewhat
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This allows to save account entities in for example a redis cache, because SocialMediaProfiles are loaded eager and thus can be serialized.

#### Why?

We cache account entites for quick access as part of a relation in snippets. When those snippets are stored "as is" in redis, somehow because of Doctrine-magic only those elements are persisted, that are already loaded in the object. Because for the socialMediaProfiles were joined but not selected, the persisting of the object leads to incomplete (but functional) objects. So, when loading the snippet with the account from redis, the socialMediaProfiles are gone, because the intrinsic loading is "broken".
Since there were also JOINs for these items anyway, there shouldn't be a perceivable performance issue.
For consistencies sake, I also added the select to the findAccountByIdAndDelete method.

Becuase it is only Doctrine-magic, I did not add any tests, because, if the tests keep running, the entities can be loaded.


alexander-schranz:
I'm unsure if we should fetch here more if most not require it and we doing already a lot of overfetching in the accounts. What do you think @chirimoya.

@MarkusHolstein current workaround for you can be before caching you loop over the social profiles or other lazy collections so they are loaded by doctrine and also cached inside redis.

PR #8427, opened 2025-12-01
Clean up of database on account deletion

MarkusHolstein:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Cleans up the database, when deleting an account.

#### Why?

When deleting an account, all it's relations where kept in the database. Though not accessible via GUI, they polluted the database. The ORM does not seem to reflect deletion-cascading, so I implemented this manually. Though, the entries in the relation tables like `co_account_$ENTITY` are deleted.
Additionally there was a bug, when the addresses were tried to be removed. The addresses were also kept, because of it checked, if the address had ANY relation, which would include the relation to the account itself.

### DIFF
--- a/src/Sulu/Bundle/ContactBundle/Controller/AccountController.php
+++ b/src/Sulu/Bundle/ContactBundle/Controller/AccountController.php
@@ -770,17 +770,6 @@ class AccountController extends AbstractRestController implements ClassResourceI
                 $this->entityManager->remove($address);
             }
 
-            $bankAccounts = $account->getBankAccounts();
-            foreach ($bankAccounts as $bankAccount) {
-                if (\count($bankAccount->getContacts()) > 0) {
-                    continue;
-                }
-                if ($bankAccount->getAccounts()->toArray() !== [$account]) {
-                    continue;
-                }
-                $this->entityManager->remove($bankAccount);
-            }
-
             $flatRelations = \array_merge(
                 $account->getUrls()->toArray(),
                 $account->getEmails()->toArray(),
@@ -788,15 +777,17 @@ class AccountController extends AbstractRestController implements ClassResourceI
                 $account->getFaxes()->toArray(),
                 $account->getNotes()->toArray(),
                 $account->getSocialMediaProfiles()->toArray(),
+                $account->getBankAccounts()->toArray()
             );
 
-            foreach ($flatRelations as $relation) {
-                $this->entityManager->remove($relation);
-            }
-
-            $phones = $account->getPhones();
-            foreach ($phones as $phone) {
-                $this->entityManager->remove($phone);
+            foreach ($flatRelations as $relationEntity) {
+                if (\count($relationEntity->getContacts()) > 0) {
+                    continue;
+                }
+                if ($relationEntity->getAccounts()->toArray() !== [$account]) {
+                    continue;
+                }
+                $this->entityManager->remove($relationEntity);
             }
 
             // Remove related contacts if removeContacts is true.
--- a/src/Sulu/Bundle/ContactBundle/Controller/AccountController.php
+++ b/src/Sulu/Bundle/ContactBundle/Controller/AccountController.php
@@ -24,6 +24,8 @@ use Sulu\Bundle\ContactBundle\Domain\Event\AccountContactRemovedEvent;
 use Sulu\Bundle\ContactBundle\Domain\Event\AccountCreatedEvent;
 use Sulu\Bundle\ContactBundle\Domain\Event\AccountModifiedEvent;
 use Sulu\Bundle\ContactBundle\Domain\Event\AccountRemovedEvent;
+use Sulu\Bundle\ContactBundle\Entity\Account;
+use Sulu\Bundle\ContactBundle\Entity\AccountAddress;
 use Sulu\Bundle\ContactBundle\Entity\AccountContact as AccountContactEntity;
 use Sulu\Bundle\ContactBundle\Entity\AccountInterface;
 use Sulu\Bundle\ContactBundle\Entity\AccountRepositoryInterface;
@@ -739,6 +741,7 @@ class AccountController extends AbstractRestController implements ClassResourceI
         }
 
         $delete = function($id) use ($request) {
+            /** @var ?Account $account */
             $account = $this->accountRepository->findAccountByIdAndDelete($id);
 
             if (!$account) {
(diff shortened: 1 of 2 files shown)


src/Sulu/Bundle/ContactBundle/Controller/AccountController.php
```diff
@@ -752,9 +755,39 @@ public function deleteAction($id, Request $request)
...
+
+            $flatRelations = \array_merge(
+                $account->getUrls()->toArray(),
+                $account->getEmails()->toArray(),
+                $account->getPhones()->toArray(),
+                $account->getFaxes()->toArray(),
+                $account->getNotes()->toArray(),
+                $account->getSocialMediaProfiles()->toArray(),
+                $account->getBankAccounts()->toArray()
+            );
```
mamazu, line 781: Why don't we use the cascade function of the database? That way it would optimize also the RAM usage since we don't need to load the entities that we want to remove.

MarkusHolstein: I always try to stay in the "mood" of the class. At this point it seemed to be intentional to delete those relations via PHP and not cascading (as the address entity and backaccount entity are already implemented that way, and the orm already has those cascadings for the relation table, but not for the entity itself). Additionally, the current implementation would allow for "hardcore third normalisation", so users could have implemented adding one of those entities to releate to multiple account and contact entities. So I thought it would be best and non-breaking, to implement it in PHP.

alexander-schranz: Maybe we can work with some kind of orphan removal. The database schema is a little bit strange and thats why cascade not working as its a typical CRM based schema very traditional where all could be shared, even its on the core side not used that way project may have adopted it to use that mechanic.

But @mamazu is right that we run here maybe into performance issues we we fetch all. The previous one with `isEmpty` did just do a count query, which sure didn't work as expected as removal was not yet flushed.

maybe some orphan removal would work here to let such things to handle by doctrine. Does somebody want try to dive into that rabbit hole? Else I will put it on our list :).

mamazu: Okay I've played around with this PR. Adding orphan removal to the property doesn't change also delete this. From the documentation it sounds like orphan removal is for when you modify a collection to remove items that those items are then also deleted from the DB.

However adding `<cascade-remove />` to the relation does the job. I don't know how it is with re-used entities but I think this should not be the case anyways. (see diff below)

One thing that I don't understand though is, why isn't this handled by the database because all join columns have the `on-delete="CASCADE"` property?

<details>
<summary>Diff that works with minimal code changes</summary>

```diff
diff --git a/src/Sulu/Bundle/ContactBundle/Controller/AccountController.php b/src/Sulu/Bundle/ContactBundle/Controller/AccountController.php
index 05209f8069..458119f008 100644
--- a/src/Sulu/Bundle/ContactBundle/Controller/AccountController.php
+++ b/src/Sulu/Bundle/ContactBundle/Controller/AccountController.php
@@ -770,25 +770,25 @@ class AccountController extends AbstractRestController implements ClassResourceI
                 $this->entityManager->remove($address);
             }
 
-            $flatRelations = \array_merge(
-                $account->getUrls()->toArray(),
-                $account->getEmails()->toArray(),
-                $account->getPhones()->toArray(),
-                $account->getFaxes()->toArray(),
-                $account->getNotes()->toArray(),
-                $account->getSocialMediaProfiles()->toArray(),
-                $account->getBankAccounts()->toArray()
-            );
+            //$flatRelations = \array_merge(
+                //$account->getUrls()->toArray(),
+                //$account->getEmails()->toArray(),
+                //$account->getPhones()->toArray(),
+                //$account->getFaxes()->toArray(),
+                //$account->getNotes()->toArray(),
+                //$account->getSocialMediaProfiles()->toArray(),
+                //$account->getBankAccounts()->toArray()
+            //);
 
-            foreach ($flatRelations as $relationEntity) {
-                if (\count($relationEntity->getContacts()) > 0) {
-                    continue;
-                }
-                if ($relationEntity->getAccounts()->toArray() !== [$account]) {
-                    continue;
-                }
-                $this->entityManager->remove($relationEntity);
-            }
+            //foreach ($flatRelations as $relationEntity) {
+                //if (\count($relationEntity->getContacts()) > 0) {
+                    //continue;
+                //}
+                //if ($relationEntity->getAccounts()->toArray() !== [$account]) {
+                    //continue;
+                //}
+                //$this->entityManager->remove($relationEntity);
+            //}
 
             // Remove related contacts if removeContacts is true.
             if (null !== $request->get('removeContacts')
diff --git a/src/Sulu/Bundle/ContactBundle/Resources/config/doctrine/Account.orm.xml b/src/Sulu/Bundle/ContactBundle/Resources/config/doctrine/Account.orm.xml
index c18f8a41be..c254d7dbc8 100644
--- a/src/Sulu/Bundle/ContactBundle/Resources/config/doctrine/Account.orm.xml
+++ b/src/Sulu/Bundle/ContactBundle/Resources/config/doctrine/Account.orm.xml
@@ -102,6 +102,7 @@
         <many-to-many field="urls" target-entity="Sulu\Bundle\ContactBundle\Entity\Url" inversed-by="accounts">
             <cascade>
                 <cascade-persist/>
+                <cascade-remove />
             </cascade>
             <join-table name="co_account_urls">
                 <join-columns>
@@ -115,6 +116,7 @@
         <many-to-many field="phones" target-entity="Sulu\Bundle\ContactBundle\Entity\Phone" inversed-by="accounts">
             <cascade>
                 <cascade-persist/>
+                <cascade-remove />
             </cascade>
             <join-table name="co_account_phones">
                 <join-columns>
@@ -128,6 +130,7 @@
         <many-to-many field="emails" target-entity="Sulu\Bundle\ContactBundle\Entity\Email" inversed-by="accounts">
             <cascade>
                 <cascade-persist/>
+                <cascade-remove />
             </cascade>
             <join-table name="co_account_emails">
                 <join-columns>
@@ -141,6 +144,7 @@
         <many-to-many field="notes" target-entity="Sulu\Bundle\ContactBundle\Entity\Note" inversed-by="accounts">
             <cascade>
                 <cascade-persist/>
+                <cascade-remove />
             </cascade>
             <join-table name="co_account_notes">
                 <join-columns>
@@ -154,6 +158,7 @@
         <many-to-many field="faxes" target-entity="Sulu\Bundle\ContactBundle\Entity\Fax" inversed-by="accounts">
             <cascade>
                 <cascade-persist/>
+                <cascade-remove />
             </cascade>
             <join-table name="co_account_faxes">
                 <join-columns>
@@ -169,6 +174,7 @@
                       inversed-by="accounts">
             <cascade>
                 <cascade-persist/>
+                <cascade-remove />
             </cascade>
             <join-table name="co_account_social_media_profiles">
                 <join-columns>
@@ -182,6 +188,7 @@
         <many-to-many field="bankAccounts" target-entity="Sulu\Bundle\ContactBundle\Entity\BankAccount" inversed-by="accounts">
             <cascade>
                 <cascade-persist/>
+                <cascade-remove />
             </cascade>
             <join-table name="co_account_bank_accounts">
                 <join-columns>
diff --git a/src/Sulu/Bundle/ContactBundle/Tests/Functional/Controller/AccountControllerTest.php b/src/Sulu/Bundle/ContactBundle/Tests/Functional/Controller/AccountControllerTest.php
index 4f1836d8ee..39e930bf20 100644
--- a/src/Sulu/Bundle/ContactBundle/Tests/Functional/Controller/AccountControllerTest.php
+++ b/src/Sulu/Bundle/ContactBundle/Tests/Functional/Controller/AccountControllerTest.php
@@ -1477,14 +1477,14 @@ class AccountControllerTest extends SuluTestCase
         $trashItem = $this->trashItemRepository->findOneBy(['resourceKey' => 'accounts', 'resourceId' => $accountId]);
         $this->assertNotNull($trashItem);
 
-        self::assertNull($this->em->find(Url::class, $url->getId()));
-        self::assertNull($this->em->find(Email::class, $email->getId()));
-        self::assertNull($this->em->find(Phone::class, $phone->getId()));
-        self::assertNull($this->em->find(Fax::class, $fax->getId()));
-        self::assertNull($this->em->find(Address::class, $address->getId()));
-        self::assertNull($this->em->find(Note::class, $note->getId()));
-        self::assertNull($this->em->find(SocialMediaProfile::class, $socialMediaProfile->getId()));
-        self::assertNull($this->em->find(BankAccount::class, $bankAccount->getId()));
+        self::assertNull($this->em->find(Url::class, $url->getId()), 'Expecting urls to be cleaned up');
+        self::assertNull($this->em->find(Email::class, $email->getId()), 'Expecting emails to be cleaned up');
+        self::assertNull($this->em->find(Phone::class, $phone->getId()), 'Expecting phones to be cleaned up');
+        self::assertNull($this->em->find(Fax::class, $fax->getId()), 'Expecting faxes to be cleaned up');
+        self::assertNull($this->em->find(Address::class, $address->getId()), 'Expecting addresses to be cleaned up');
+        self::assertNull($this->em->find(Note::class, $note->getId()), 'Expecting notes to be cleaned up');
+        self::assertNull($this->em->find(SocialMediaProfile::class, $socialMediaProfile->getId()), 'Expecting socialMediaProfile to be cleaned up');
+        self::assertNull($this->em->find(BankAccount::class, $bankAccount->getId()), 'Expecting bank accounts to be cleaned up');
     }
 
     public function testDeleteParentById(): void
diff --git a/src/Sulu/Bundle/ContactBundle/phpunit.xml.dist b/src/Sulu/Bundle/ContactBundle/phpunit.xml.dist
index d37886add3..58b29ce056 100644
--- a/src/Sulu/Bundle/ContactBundle/phpunit.xml.dist
+++ b/src/Sulu/Bundle/ContactBundle/phpunit.xml.dist
@@ -22,6 +22,7 @@
 
     <php>
         <server name="APP_ENV" value="test" force="true"/>
         <env name="KERNEL_CLASS" value="Sulu\Bundle\ContactBundle\Tests\Application\Kernel"/>
         <env name="SYMFONY_DEPRECATIONS_HELPER" value="weak"/>
         <ini name="date.timezone" value="UTC"/>
```
</details>

PR #8428, opened 2025-12-02
Add RouteSchema support

Prokyonn:
| Q                  | A               |
  |--------------------|-----------------|
  | Bug fix?           | no              |
  | New feature?       | yes             |
  | BC breaks?         | no              |
  | Deprecations?      | no              |
  | Fixed tickets      | fixes #         |
  | Related issues/PRs | #               |
  | License            | MIT             |
  | Documentation PR   | sulu/sulu-docs# |

  What's in this PR?

  - Adds RouteSchemaEvaluator to process custom route schemas using Symfony ExpressionLanguage
  - Supports expressions like /{object.title}, /{object['title']}, /{translator.trans('key', [], 'admin')}

  Why?

Adds the missing feature from 2.6

### DIFF
--- a/packages/route/src/Application/ResourceLocator/ArrayAccessObject.php
+++ b/packages/route/src/Application/ResourceLocator/ArrayAccessObject.php
@@ -17,6 +17,8 @@ namespace Sulu\Route\Application\ResourceLocator;
  * A wrapper class that allows both property access (object.title) and
  * array access (object['title']) syntax in ExpressionLanguage expressions.
  *
+ * @internal no backwards compatibility promises are given for this class it can be removed at any time
+ *
  * @implements \ArrayAccess<string, mixed>
  */
 final class ArrayAccessObject implements \ArrayAccess
--- a/packages/route/src/Application/ResourceLocator/ArrayAccessObject.php
+++ b/packages/route/src/Application/ResourceLocator/ArrayAccessObject.php
@@ -1,0 +1,69 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of Sulu.
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Route\Application\ResourceLocator;
+
+/**
+ * A wrapper class that allows both property access (object.title) and
+ * array access (object['title']) syntax in ExpressionLanguage expressions.
+ *
+ * @implements \ArrayAccess<string, mixed>
+ */
+final class ArrayAccessObject implements \ArrayAccess
+{
+    /**
+     * @param array<string, mixed> $data
+     */
+    public function __construct(
+        private array $data,
+    ) {
+    }
+
+    public function __get(string $name): mixed
+    {
+        return $this->data[$name] ?? null;
+    }
+
+    public function __isset(string $name): bool
+    {
+        return isset($this->data[$name]);
+    }
+
+    public function offsetExists(mixed $offset): bool
+    {
+        return isset($this->data[$offset]);
+    }
+
+    public function offsetGet(mixed $offset): mixed
+    {
+        return $this->data[$offset] ?? null;
+    }
+
+    public function offsetSet(mixed $offset, mixed $value): void
+    {
+        throw new \LogicException('ArrayAccessObject is read-only.');
+    }
+
+    public function offsetUnset(mixed $offset): void
+    {
+        throw new \LogicException('ArrayAccessObject is read-only.');
+    }
+
+    /**
+     * @return array<string, mixed>
+     */
+    public function getArrayCopy(): array
+    {
+        return $this->data;
+    }
+}
--- a/packages/route/src/Application/ResourceLocator/ResourceLocatorGenerator.php
+++ b/packages/route/src/Application/ResourceLocator/ResourceLocatorGenerator.php
@@ -22,48 +22,22 @@ final readonly class ResourceLocatorGenerator implements ResourceLocatorGenerato
     public function __construct(
         private RouteRepositoryInterface $routeRepository,
         private PathCleanupInterface $pathCleanup,
-        private ?RouteSchemaProcessor $routeSchemaProcessor = null,
+        private ?RouteSchemaEvaluator $routeSchemaProcessor = null,
     ) {
     }
 
     public function generate(ResourceLocatorRequest $request): string
     {
-        $parentPath = '/';
-        if ($request->parentResourceId) {
-            $parentRoute = $this->routeRepository->findOneBy([
-                'resourceKey' => $request->parentResourceKey,
-                'resourceId' => $request->parentResourceId,
-                'locale' => $request->locale,
-            ]);
-
-            $parentPath = $parentRoute?->getSlug() ?: '/';
-        }
-
-        $parentPath = \rtrim($parentPath, '/');
+        $parentPath = $this->resolveParentPath($request);
 
         if (null !== $request->routeSchema && null !== $this->routeSchemaProcessor) {
             $path = $this->routeSchemaProcessor->process($request);
-
-            $uniquePath = $this->createUnique(
-                $parentPath . $path,
-                $request->locale,
-                $request->webspace,
-                $request->resourceKey,
-                $request->resourceId,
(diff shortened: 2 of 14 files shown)


packages/route/src/Application/ResourceLocator/ArrayAccessObject.php
```diff
@@ -0,0 +1,69 @@
...
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Route\Application\ResourceLocator;
+
+/**
+ * A wrapper class that allows both property access (object.title) and
+ * array access (object['title']) syntax in ExpressionLanguage expressions.
+ *
+ * @implements \ArrayAccess<string, mixed>
```
alexander-schranz, line 22: Its only used to make the expression language "easier" we should keep it internal.

```suggestion
 * A wrapper class that allows both property access (object.title) and
 * array access (object['title']) syntax in ExpressionLanguage expressions.
 *
 * @internal No backwards compatibility promises are given for this class it can be removed at any time.
 *
 * @implements \ArrayAccess<string, mixed>
```

packages/route/src/Application/ResourceLocator/ResourceLocatorGenerator.php
```diff
@@ -22,6 +22,7 @@
     public function __construct(
         private RouteRepositoryInterface $routeRepository,
         private PathCleanupInterface $pathCleanup,
+        private ?RouteSchemaProcessor $routeSchemaProcessor = null,
```
alexander-schranz: fine to force this as a bugfix:


```suggestion
        private RouteSchemaProcessorInterface $routeSchemaProcessor,
```

packages/route/src/Application/ResourceLocator/ResourceLocatorGenerator.php
```diff
@@ -38,13 +39,29 @@ public function generate(ResourceLocatorRequest $request): string
...
+
+            if ($request->relative) {
+                return \substr($uniquePath, \strlen($parentPath));
+            }
 
-        // TODO routeSchema
+            return $uniquePath;
+        }
+
+        $parts = \array_map(fn ($part) => $this->pathCleanup->cleanup($part, $request->locale), $request->parts);
```
alexander-schranz: We need do this for the parts of RouteSchemaProcessor also.

packages/route/src/Application/ResourceLocator/ResourceLocatorGenerator.php
```diff
@@ -38,13 +39,29 @@ public function generate(ResourceLocatorRequest $request): string
             $parentPath = $parentRoute?->getSlug() ?: '/';
         }
 
-        $parts = \array_map(fn ($part) => $this->pathCleanup->cleanup($part, $request->locale), $request->parts);
+        $parentPath = \rtrim($parentPath, '/');
 
-        $path = '/' . \implode('-', $parts);
+        if (null !== $request->routeSchema && null !== $this->routeSchemaProcessor) {
+            $path = $this->routeSchemaProcessor->process($request);
```
alexander-schranz: Not sure about given the `ResourceLocatorRequest` to it but fine with it but would maybe introduce a:

```
$parts = \array_map(fn ($part) => $this->pathCleanup->cleanup($part, $request->locale), $request->parts);
$request->withParts($parts);
```

even I'm not a fan normally with with methods.

packages/route/src/Application/ResourceLocator/ResourceLocatorGenerator.php
```diff
@@ -38,13 +39,29 @@ public function generate(ResourceLocatorRequest $request): string
...
+                $parentPath . $path,
+                $request->locale,
+                $request->webspace,
+                $request->resourceKey,
+                $request->resourceId,
+            );
+
+            if ($request->relative) {
+                return \substr($uniquePath, \strlen($parentPath));
+            }
```
alexander-schranz: avoid this duplication:


```suggestion
```

packages/route/src/Application/ResourceLocator/ResourceLocatorGenerator.php
```diff
@@ -38,13 +39,29 @@ public function generate(ResourceLocatorRequest $request): string
...
 
-        // TODO routeSchema
+            return $uniquePath;
```
alexander-schranz: Avoid this duplication:

```suggestion
```

packages/route/src/Application/ResourceLocator/ResourceLocatorGenerator.php
```diff
@@ -38,13 +39,29 @@ public function generate(ResourceLocatorRequest $request): string
             $parentPath = $parentRoute?->getSlug() ?: '/';
         }
 
-        $parts = \array_map(fn ($part) => $this->pathCleanup->cleanup($part, $request->locale), $request->parts);
+        $parentPath = \rtrim($parentPath, '/');
 
-        $path = '/' . \implode('-', $parts);
+        if (null !== $request->routeSchema && null !== $this->routeSchemaProcessor) {
```
alexander-schranz: fine to call the routeSchemaProcessor always with a default schema:

```
$request->routeSchema ?? 'implode("-", object)';
```

packages/route/src/Application/ResourceLocator/RouteSchemaEvaluator.php
```diff
@@ -0,0 +1,109 @@
...
+
+    public function __construct(
+        private TranslatorInterface $translator,
+        private PathCleanupInterface $pathCleanup,
+    ) {
+        $this->expressionLanguage = new ExpressionLanguage();
+        $this->registerFunctions();
+    }
+
+    public function process(ResourceLocatorRequest $request): string
```
alexander-schranz: Lets create a interface for it but make this implementation internal as we may in new versions add more to the constructor.

alexander-schranz:
```suggestion
    /**
     * @throws InvalidRouteSchemaException
     */
    public function process(ResourceLocatorRequest $request): string
```

packages/route/src/Application/ResourceLocator/ResourceLocatorGenerator.php
alexander-schranz: but maybe not necessary if we just handle it always in the routeSchemaProcess method

packages/route/src/Application/ResourceLocator/RouteSchemaEvaluator.php
```diff
@@ -0,0 +1,109 @@
...
+            try {
+                $value = $this->expressionLanguage->evaluate($expression, $context);
+            } catch (SyntaxError $e) {
+                throw new InvalidRouteSchemaException(
+                    \sprintf('Invalid expression "%s" in route schema: %s', $expression, $e->getMessage()),
+                    $e
+                );
+            }
+            /** @var string|int|float $value */
+            $replacements[$token] = $this->pathCleanup->cleanup((string) $value, $request->locale);
```
alexander-schranz, line 64: Any way that we can do cleanup in the parent service and not require todo it in this and the parent.

So this returns uncleanup version and cleaning up is done in the parent service.

packages/route/src/Application/ResourceLocator/ResourceLocatorGenerator.php
```diff
@@ -38,13 +39,29 @@ public function generate(ResourceLocatorRequest $request): string
...
+                return \substr($uniquePath, \strlen($parentPath));
+            }
 
-        // TODO routeSchema
+            return $uniquePath;
+        }
+
+        $parts = \array_map(fn ($part) => $this->pathCleanup->cleanup($part, $request->locale), $request->parts);
+
+        $path = '/' . \implode('-', $parts);
```
alexander-schranz: we could create a default schema which does this, so the process is always same see above.

alexander-schranz (changes requested):

packages/route/src/Application/ResourceLocator/RouteSchemaEvaluator.php
alexander-schranz: Or only call in this service service, atleast that cleanup is done at one place not multiple.

alexander-schranz (approved):

Issue #7823, opened 2025-02-25
Sort Collection by name in Media Library


alexander-schranz:
# Problem description

For the content manager the collection are random sorted in media selections, media move and media library itself. This is very confusing as seen in the screenshots:

![Image](https://github.com/user-attachments/assets/d8e9dc79-15a8-4ff7-8bd8-792aaa01c3ec)

![Image](https://github.com/user-attachments/assets/5c81c688-ad89-4c2f-9113-c71aaeb55f70)

# Proposed solution

We should sort by the name / title like we do for categories already correctly:

![Image](https://github.com/user-attachments/assets/27cec8ae-9364-413f-b7ee-92e81d8149e6)

# Open Question

 - [ ] What about System Collection should we still put them ontop?

alexander-schranz:
@chirimoya what should we do about the system collection, should they be at the top still?

chirimoya:
@alexander-schranz Shouldn't they rather come at the end?

alexander-schranz:
@chirimoya currently system collection are listed first. For the selection I would also prefer it at the bottom. For the media library itself I'm not sure as then system collection would be always on the last page?

chirimoya:
Normally, you don’t use any system collection actively at all. So yes, the last page is totally fine in my opinion.

MarkusHolstein:
I think, this can be closed?

PR #8432, opened 2025-12-04
Add Sort collections by title per default

MarkusHolstein:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Related issues/PRs | #7823
| License | MIT

#### What's in this PR?

Media collections are sorted alphabetically by default. Additionally system collections are sorted to the bottom as suggested in #7823 

#### What's more

Just to point this out:
Because of subselects, which are implicitly done because of the pagination I think, the amount of parameters were not correct, because in the subselect, the OrderBy was copied, but the corresponding parameter not.


alexander-schranz:
Note to me, need to test it if collection only exist in one language and other not.

Prokyonn:
@alexander-schranz Should the `cgetAction` also be sorted? 
As the Admin UI does not send any sort parameters, therefore we get the default insert order:

`/admin/api/collections?locale=en&selectedIds=&fields=title,objectCount,id&flat=true`
image

MarkusHolstein:
Sorry, I oversaw that. I am going to fix this.

Prokyonn (approved):

alexander-schranz (approved):

PR #8437, opened 2025-12-05
Fix query string parameters when using media as smart content data provider

### LINKED ISSUE #8433: Unable to filter media smart content using mimetype and/or type query string parameters
| Q | A
| --- | ---
| Sulu Version | 2.6, 3.0
| PHP Version | 8.2
| DB Version | MariaDB 10.11.13
| Browser Version | N/A

# Actual Behavior

When using smart content with `media` as the data provider, the `mimetype` and `type` query string parameters trigger a Doctrine `QueryException`:

> Too many parameters: the query defines 4 parameters and you bound X

Where X is 5 or 6 depending on if you use one or both query string parameters.

In `DataProviderRepositoryTrait::findByFiltersIds()`, [`append()` is called](https://github.com/sulu/sulu/blob/6c7eb6e953c0bf2644a9250edbcd2d842d95a562/src/Sulu/Component/SmartContent/Orm/DataProviderRepositoryTrait.php#L131-L139) for the data source. With the `media` data source, this adds `WHERE` conditions for `mimeType` and `type` if the associated query string parameters are set.

But then in `MediaDataProviderRepository::appendDatasource()`, instead of using `andWhere()`, `where()` is used, [which wipes out any other conditions](https://github.com/sulu/sulu/blob/6c7eb6e953c0bf2644a9250edbcd2d842d95a562/src/Sulu/Bundle/MediaBundle/Entity/MediaDataProviderRepository.php#L208-L215) set before this point.

This means that anything that gets set in the `append()` function is added to the list of parameters for the query, but since the query is missing the associated `WHERE` statements, an exception is triggered due to the number of parameters not matching.

I haven't tested this with the 3.0 branch yet. But the [equivalent code in that branch](https://github.com/sulu/sulu/blob/289e51279a851ed105f35803ec8476dead2c7371/src/Sulu/Bundle/MediaBundle/Infrastructure/Sulu/Content/SmartContent/MediaSmartContentProvider.php#L266-L274) also uses `where()` and not `andWhere()`.

# Expected Behavior

It should be possible to filter `media` smart content results by `mimetype` and `type`, as [described in the documentation](https://docs.sulu.io/en/2.6/reference/content-types/smart_content.html#media), without triggering a critical error.

# Steps to Reproduce

Append `?mimetype=someMimeType` or `?type=someType` to a page URL that uses smart content with `media` as the data provider.

# Possible Solutions

Replace

```php
->where('collection.lft BETWEEN parentCollection.lft AND parentCollection.rgt');
```

with

```php
->andWhere('collection.lft BETWEEN parentCollection.lft AND parentCollection.rgt');
```

in [`MediaDataProviderRepository`](https://github.com/sulu/sulu/blob/6c7eb6e953c0bf2644a9250edbcd2d842d95a562/src/Sulu/Bundle/MediaBundle/Entity/MediaDataProviderRepository.php#L215). Assuming this is also an issue in 3.0, the fix should be the same there.

NeuralClone:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #8433
| Related issues/PRs | --
| License | MIT
| Documentation PR | --

#### What's in this PR?

This PR fixes a Doctrine `QueryException` when using `media` as the smart content data provider that occurs when attempting to filter the results with the `mimetype` and `type` query string parameters. See #8433 for more details.

#### Why?

To allow the filtering of smart content with the [documented query string parameters](https://docs.sulu.io/en/2.6/reference/content-types/smart_content.html#media).

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/Entity/MediaDataProviderRepository.php
+++ b/src/Sulu/Bundle/MediaBundle/Entity/MediaDataProviderRepository.php
@@ -212,7 +212,7 @@ class MediaDataProviderRepository implements DataProviderRepositoryInterface
                     Join::WITH,
                     'parentCollection.id = :collectionId'
                 )
-                ->where('collection.lft BETWEEN parentCollection.lft AND parentCollection.rgt');
+                ->andWhere('collection.lft BETWEEN parentCollection.lft AND parentCollection.rgt');
         }
 
         return ['collectionId' => $datasource];
--- a/src/Sulu/Bundle/MediaBundle/Tests/Functional/Entity/MediaDataProviderRepositoryTest.php
+++ b/src/Sulu/Bundle/MediaBundle/Tests/Functional/Entity/MediaDataProviderRepositoryTest.php
@@ -541,6 +541,66 @@ class MediaDataProviderRepositoryTest extends SuluTestCase
                 [],
                 [],
             ],
+            // datasource sub folder with options mimetype/type
+            [
+                ['dataSource' => 1, 'includeSubFolders' => true],
+                null,
+                0,
+                null,
+                \array_slice(self::$mediaData, 0, 2),
+                [],
+                ['mimetype' => 'image/jpg', 'type' => 'image'],
+            ],
+            // datasource no sub folder with options mimetype/type
+            [
+                ['dataSource' => 1, 'includeSubFolders' => false],
+                null,
+                0,
+                null,
+                \array_slice(self::$mediaData, 0, 1),
+                [],
+                ['mimetype' => 'image/jpg', 'type' => 'image'],
+            ],
+            // datasource sub folder with option mimetype
+            [
+                ['dataSource' => 1, 'includeSubFolders' => true],
+                null,
+                0,
+                null,
+                \array_slice(self::$mediaData, 0, 2),
+                [],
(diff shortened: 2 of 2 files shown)


alexander-schranz:
@NeuralClone thx for the pull request. Can we add a test case here to test this case also: https://github.com/sulu/sulu/blob/138e038908197021e6fbb90ea6b3921a1d5b86b8/src/Sulu/Bundle/MediaBundle/Tests/Functional/Entity/MediaDataProviderRepositoryTest.php#L534

NeuralClone:
@alexander-schranz Done! I think the test cases I added cover everything. But if not, I'd be happy to make adjustments.

alexander-schranz (approved):

alexander-schranz:
@NeuralClone thank you!

alexander-schranz:
@NeuralClone do you want to tackle the issue also on the 3.0 branch: https://github.com/sulu/sulu/blob/5959a6e018e8eda93dbc5a80c95a30c720a76466/src/Sulu/Bundle/MediaBundle/Infrastructure/Sulu/Content/SmartContent/MediaSmartContentProvider.php#L273

NeuralClone:
@alexander-schranz You're welcome! And sure, I'd be happy to.

PR #8438, opened 2025-12-06
Remove php-cr "node_order" build target in massive_build

mamazu:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?
Remove the dependency of the "maintain" build job on the "node_order" job.

#### Why?
I've run `bin/console sulu:build` without any argument it lists all jobs and the "maintain" job still has a dependency on "node_order" from the phpcr implementation.

Maybe we should also consider validating this kind of errors in the massive bundle at compile time.

## Alternative solution
Remove the "maintain" build target. This might break running "sulu:build maintain"

### DIFF
--- a/src/Sulu/Bundle/CoreBundle/DependencyInjection/SuluCoreExtension.php
+++ b/src/Sulu/Bundle/CoreBundle/DependencyInjection/SuluCoreExtension.php
@@ -198,8 +198,6 @@ class SuluCoreExtension extends Extension implements PrependExtensionInterface
                             'security' => [],
                         ],
                     ],
-                    'maintain' => [
-                    ],
                 ],
             ];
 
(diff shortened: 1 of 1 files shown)


src/Sulu/Bundle/CoreBundle/DependencyInjection/SuluCoreExtension.php
```diff
@@ -199,9 +199,6 @@ public function prepend(ContainerBuilder $container): void
                         ],
                     ],
                     'maintain' => [
-                        'dependencies' => [
-                            'node_order' => [],
-                        ],
                     ],
```
mamazu: Maybe we can even remove the entire job? But that might be a little BC break.

alexander-schranz: fine with remove the entire job if it currently does nothing.

mamazu: Removed, but is there a way to prevent this in the future? Like adding a test to load the config?

alexander-schranz (approved):

PR #8439, opened 2025-12-06
Fix the autoloader warnings and moving namespaces

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?
Adding excluded namespaces for autoloading and fixed some namespaces of test directories and the autoloader when dumped does not throw any warnings:

image

#### Why?
When moving from the old directory structure to the new one some of the namespaces were left out which generates a lot of warnings if you try to optimize the autoloader:

image

Some of the tests when moved from the old bundle structure to the new one haven't updated their namespaces.

## Additional context

We could also add a check to the pipeline to prevent this: `composer dump-autoload -o --strict-psr`


alexander-schranz (approved):

alexander-schranz:
@mamazu Thank you!

Issue #7585, opened 2024-09-06
Add more Debug Commands


TheCadien:
# Problem description

In Sulu there are some functions that are facilitated by a debug statement on the command line, as is also common in the Symfony framework. 
I think we should follow suit more often

Like feature : https://github.com/sulu/sulu/issues/7267

# Proposed solution

Here is a list of ideas

- [ ] `bin/console sulu:debug:routes` 
Currently, only all routes stored in the Symfony framework are displayed via “debug:router”. 
An additional command for debugging the Sulu routing would make sense .

- [ ] `bin/console sulu:debug:webspace`
Displays the list of all currently active webspaces with the option of displaying them individually and their configuration

mamazu:
- I'm an experienced Symfony Dev so `debug:routes` sounds kinda excessive. But if you find a way to add information to the route the normal router couldn't provide why not. 
- `debug:webspaces` Yes, especially since there is some processing going on so it's not just a copy of the xml contents. However, if we ever move to symfony config with the webspaces this might become deprecated.

## More ideas
Sulu has many optional features for a resource like Trashing, tracking references etc. Maybe it would be a good idea to have a command like `sulu:debug:resource` to debug what features are available for a given resource something like:

```
bin/console sulu:debug:resource event

Admin Config: ✔ (maybe show where in the tree it is)
List view: ✔ (config/list/event.xml)
Preview: ✔(Example\\PreviewProvider\\EventPreview)
Trashing: ✔ (Example\\Trash\\EventTrashHandler)
Security: ✔ (maybe show the roles it has)
```

I don't know how easy this would be to implement but I would definitely be using it a lot because I always mess those things up.

alexander-schranz:
Routes I think is maybe something less used in content management system. But the admin bundle could provide commands for:

 - `bin/console sulu:admin:debug-webspace` (not sure if we move webspaces to symfony config)
 - `bin/console sulu:admin:debug-resource` (why not sure as `bin/console debug:config sulu_admin resources` does already give a lot of information
 - already existing is `bin/console sulu:admin:debug-view`
 - `bin/console sulu:admin:debug-navigation` could may helpful

PS: There will be no `sulu:debug` namespace as the commands will live in the admin bundle and not a seperate debug bundle. Our command conventation is `sulu:<bundle>:<command-name>`.

About `debug:routes` that command would better live in https://github.com/symfony-cmf/routing-bundle, so it works out of the box with every dynamic router and isn't Sulu specific.

PR #8440, opened 2025-12-08
Add `sulu:content:debug` command

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | #7585
| License | MIT
| Documentation PR | -

#### What's in this PR?
A command "sulu:debug:content" that lists all content resolvers, property resolvers and resource loaders that are registered in Sulu. This should help when trying to create new content and debug why it might not work.

image

#### Why?
As part of the FormBundle upgrade process this was very helpful to see what I configured incorrectly. (This command could also highlight errors (missing loaders etc. in the future)

### DIFF
--- a/packages/content/src/Infrastructure/Symfony/Command/DebugContentCommand.php
+++ b/packages/content/src/Infrastructure/Symfony/Command/DebugContentCommand.php
@@ -23,7 +23,11 @@ use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Style\SymfonyStyle;
 
-#[AsCommand('sulu:debug:content', description: 'Debugging the content services to see if they are correctly registered')]
+/**
+ * @internal the command can be used and console command call is under bc promise but no bc promise is given for this PHP class itself
+ * as it may will have additional dependencies in future
+ */
+#[AsCommand('sulu:content:debug', description: 'Debugging the content services to see if they are correctly registered')]
 final class DebugContentCommand extends Command
 {
     /**
--- a/packages/content/src/Infrastructure/Symfony/Command/DebugContentCommand.php
+++ b/packages/content/src/Infrastructure/Symfony/Command/DebugContentCommand.php
@@ -14,6 +14,7 @@ declare(strict_types=1);
 namespace Sulu\Content\Infrastructure\Symfony\Command;
 
 use Sulu\Content\Application\ContentResolver\Resolver\ResolverInterface;
+use Sulu\Content\Application\PropertyResolver\Resolver\PropertyResolverInterface;
 use Sulu\Content\Application\ResourceLoader\Loader\CachedResourceLoader;
 use Sulu\Content\Application\ResourceLoader\Loader\ResourceLoaderInterface;
 use Symfony\Component\Console\Attribute\AsCommand;
@@ -21,7 +22,6 @@ use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Style\SymfonyStyle;
-use Sulu\Content\Application\PropertyResolver\Resolver\PropertyResolverInterface;
 
 #[AsCommand('sulu:debug:content', description: 'Debugging the content services to see if they are correctly registered')]
 final class DebugContentCommand extends Command
--- a/packages/content/src/Infrastructure/Symfony/Command/DebugContentCommand.php
+++ b/packages/content/src/Infrastructure/Symfony/Command/DebugContentCommand.php
@@ -21,17 +21,19 @@ use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Style\SymfonyStyle;
+use Sulu\Content\Application\PropertyResolver\Resolver\PropertyResolverInterface;
 
 #[AsCommand('sulu:debug:content', description: 'Debugging the content services to see if they are correctly registered')]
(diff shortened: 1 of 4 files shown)


TheCadien:
I like that ! See reference in https://github.com/sulu/sulu/issues/7585

TheCadien (approved):

packages/content/src/Infrastructure/Symfony/Command/DebugContentCommand.php
```diff
@@ -0,0 +1,104 @@
...
+use Sulu\Content\Application\ResourceLoader\Loader\CachedResourceLoader;
+use Sulu\Content\Application\ResourceLoader\Loader\ResourceLoaderInterface;
+use Symfony\Component\Console\Attribute\AsCommand;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+
+#[AsCommand('sulu:debug:content', description: 'Debugging the content services to see if they are correctly registered')]
+final class DebugContentCommand extends Command
```
alexander-schranz: We prefix all new commands with the bundle name so if this lives in SuluContentbundle it should be `sulu:content:debug` naming convention is `<namespace>:<bundle>:<commandname>`.

While we provide BC for the command name arguments, options we not provide any bc for the command class itself so we mark the command class as `@internal` as no other need. This adds as easier possibility to add more debug informations in future:

```suggestion
/**
 * @internal The command can be used and console command call is under bc promise but no bc promise is given for this PHP class itself
 * as it may will have additional dependencies in future.
 */
#[AsCommand('sulu:content:debug', description: 'Debugging the content services to see if they are correctly registered')]
final class DebugContentCommand extends Command
```

packages/content/src/Infrastructure/Symfony/Command/DebugContentCommand.php
```diff
@@ -0,0 +1,104 @@
...
+            $cached = false;
+            if ($loader instanceof CachedResourceLoader) {
+                $loader = $loader->getInnerClass();
+                $cached = true;
+            }
+
+            $tableData[] = [
+                $key,
+                $loader,
+                $cached ? 'x' : '',
```
alexander-schranz: @Prokyonn isn't every loader cached? Does this output make sense? Else we should remove the cached column.

mamazu: Yes as far as I know, every loader is decorated with a cached version in a compiler pass.

packages/content/src/Infrastructure/Symfony/Command/DebugContentCommand.php
mamazu: Sounds good. Should I also create a pr for all the other commands to add aliases to them so that we can name them the same way?

packages/content/src/Infrastructure/Symfony/Command/DebugContentCommand.php
alexander-schranz: Then would remove the cached column as that is just something internal.

packages/content/src/Infrastructure/Symfony/Command/DebugContentCommand.php
alexander-schranz: for existing we currently keep the old names.

packages/content/src/Infrastructure/Symfony/Command/DebugContentCommand.php
```diff
@@ -0,0 +1,108 @@
...
+            ];
+        }
+
+        $io->table(['Content Type', 'Loader Class', 'Cached'], $tableData);
```
Prokyonn: we shouldn't use the name "Content Type" anymore 😬

packages/content/src/Infrastructure/Symfony/Command/DebugContentCommand.php
Prokyonn: yes all resource loader are decorated with a cache

packages/content/src/Infrastructure/Symfony/Command/DebugContentCommand.php
```diff
@@ -0,0 +1,108 @@
...
+        foreach ($resolvers as $type => $resolver) {
+            $tableData[] = [$type, $resolver::class];
+        }
+        $io->table(['Type', 'Resolver Class'], $tableData);
+    }
+
+    /**
+     * @param iterable<string, PropertyResolverInterface> $resolvers
+     */
+    private function printPropertyResolver(SymfonyStyle $io, iterable $resolvers): void
```
Prokyonn, line 98: Not sure if this will be misleading, as not all property types have an individual propertyResolver, some field types like `text_area` or `text_line` use the DefaultPropertyResolver 🤔 

At least when we document this command, we should add a hint for this

packages/content/src/Infrastructure/Symfony/Command/DebugContentCommand.php
mamazu: What other name would you suggest?

packages/content/src/Infrastructure/Symfony/Command/DebugContentCommand.php
mamazu: Added it, see screenshot in the PR description.

packages/content/src/Infrastructure/Symfony/Command/DebugContentCommand.php
mamazu: I've renamed it just to "Type" to fit in with the rest.

draconivis (approved):
LGTM

Issue #4682, opened 2019-08-20
Adding new property with default_value does not work for exist pages


alexander-schranz:
| Q | A
| --- | ---
| Bug? | yes
| New Feature? | no
| Sulu Version | 2.0.0-RC1
| Browser Version | Browser name and version

#### Actual Behavior

If I add a new single_select with default_value the default_value will not be selected for pages which are added as the response of the loading of the page will be an empty string.

#### Expected Behavior

The content type should return the default value or maybe not value? Not sure. 

#### Steps to Reproduce

1. Create a new page
2. Add a new single select to it:

```
                <property name="overviewType" type="single_select">
                    <meta>
                        <title lang="en">Type</title>
                    </meta>

                    <params>
                        <param name="default_value" value="pages"/>

                        <param name="values" type="collection">
                            <param name="pages">
                                <meta>
                                    <title lang="en">Pages</title>
                                </meta>
                            </param>

                            <param name="articles">
                                <meta>
                                    <title lang="en">Articles</title>
                                </meta>
                            </param>
                        </param>
                    </params>
                </property>

                <property name="someproperty" type="text_line" visibleCondition="overviewType == 'pages'" />
```

3. Open the created page

The default_value is not selected and the property is not shown.

danrot:
Not sure about this one either... What if it is possible to select a value of undefined? Then it should not just set this value automatically :thinking:

danrot:
We could probably find out by checking if the property previously existed on the node in PHPCR. If the property is already there and has a value of `null`, then we know it was an explicit `null`, not just because it was not filled yet.

PR #8446, opened 2025-12-09
Update block type on nested block switching

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| License | MIT

#### What's in this PR?

Loops always through blocks to update default block types.

#### Why?

When you use nested blocks and two block types, both have a `blocks` property with the same name, but different block-types as an inner block, the request crashes, because the FE does not override the block-type to the new default block-type of the new `blocks` property.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
@@ -65,7 +65,7 @@ class FieldBlocks extends React.Component<FieldTypeProps<Array<BlockEntry>>> {
 
         let newValue = toJS(value);
 
-        if (newValue && types !== oldTypes) {
+        if (newValue) {
             if (!defaultType) {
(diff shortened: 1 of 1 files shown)


src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
```diff
@@ -65,7 +62,7 @@ class FieldBlocks extends React.Component<FieldTypeProps<Array<BlockEntry>>> {
```
Prokyonn, line 65: @alexander-schranz for nested blocks `types !== oldTypes` is always false, because apparently the `types` were already updated to the new ones. Do you have any idea how we can still check that without always iterating over all blocks, as this could be decrease performance.

alexander-schranz:
is this fixed with https://github.com/sulu/sulu/pull/8662? also is https://github.com/sulu/sulu/issues/4682 also resolved?

PR #8447, opened 2025-12-10
Fix BC break in PropertyXMLParser

mamazu:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | https://github.com/sulu/SuluFormBundle/pull/415
| License | MIT

#### What's in this PR?
Returning the name with the parsed content of the XML

#### Why?
This is how it was in 2.6 and the form bundle needs this otherwise building the config produces a white page in the admin (probably because of the changed json config)

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/Parser/PropertiesXmlParser.php
+++ b/src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/Parser/PropertiesXmlParser.php
@@ -34,7 +34,7 @@ class PropertiesXmlParser
     }
 
     /**
-     * @return array<FieldMetadata|SectionMetadata>
+     * @return array<string, FieldMetadata|SectionMetadata>
      */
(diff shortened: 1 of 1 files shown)


mamazu:
@Prokyonn You might need this for forms v3

alexander-schranz:
Not sure of the change, would avoid it if possible as the normal loader works It sounds more something missing on the Form Bundle side which wasn't keep in mind in the merging mechanic.

mamazu:
Well the alternative would be to traverse the config and re-extract the property names from the property again, this sounds like the least amount of overhead to do it here.

Found it the old code used multiple `addItem` calls instead of `setItems`. Still not a fan of it. But it works.

mamazu:
Oh yeah 2.6 also didn't have that either. Well then I'll have to look into the form bundle again.

PR #8451, opened 2025-12-11
Add support for Doctrine DBAL 4 and Doctrine Persistence 4

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no yes 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT
| Documentation PR | sulu/sulu-docs# 

#### What's in this PR?

Add support for Doctrine DBAL 4 and Doctrine Persistence 4.

#### Why?

ORM 3 supports new DBAL and Persistence component versin.

Issue #8216, opened 2025-09-22
Symfony 8 Support


alexander-schranz:
# Problem description

We should keep here track what is required to support Symfony 8 and maybe already tackle changes.

# Proposed solution

Link below related required PR to be merged for Symfony 8 version:

For all which want to help bring this forward please first tackle one of the below dependencies before tackling things in Sulu as before the dependencies are not tackled we can no test Symfony 8:

## Dependencies

 - [x] SEAL: https://github.com/PHP-CMSIG/search
    - [x] Issue: https://github.com/PHP-CMSIG/search/issues/633
    - [x] Release Tag: [0.12.8](https://github.com/PHP-CMSIG/search/releases/tag/0.12.8)
 - [x] Sulu Messenger: https://github.com/sulu/messenger
    - [x] Pull Request: https://github.com/sulu/messenger/pull/15
    - [x] Release Tag: https://github.com/sulu/messenger/releases/tag/0.2.10
 - [x] Handcraftedinthealps Code Coverage Checker: https://github.com/handcraftedinthealps/code-coverage-checker/
    - [x] Pull Request https://github.com/handcraftedinthealps/code-coverage-checker/pull/10
    - [x] Release Tag: https://github.com/handcraftedinthealps/code-coverage-checker/releases/tag/0.2.8
 - [x] FOSRest: https://github.com/FriendsOfSymfony/FOSRestBundle
    - [x] Pull Request: https://github.com/FriendsOfSymfony/FOSRestBundle/pull/2427
    - [x] Pull Request: https://github.com/FriendsOfSymfony/FOSRestBundle/pull/2429
    - [x] Release Tag: https://github.com/FriendsOfSymfony/FOSRestBundle/releases/tag/3.9.0
 - [x] JMS Serializer: https://github.com/schmittjoh/serializer/
    - [x] Pull Request: https://github.com/schmittjoh/serializer/pull/1615
    - [x] Release Tag https://github.com/schmittjoh/serializer/releases/tag/3.32.6
 - [x] JMS Serializer Bundle: https://github.com/schmittjoh/JMSSerializerBundle
    - [x] Pull Request: https://github.com/schmittjoh/JMSSerializerBundle/pull/959
    - [x] Release Tag: https://github.com/schmittjoh/JMSSerializerBundle/releases/tag/5.5.2
 - [x] Flysystem Bundle: 
    - [x] Pull Request: https://github.com/thephpleague/flysystem-bundle/pull/188
    - [x] Release Tag
[...]

PR #8452, opened 2025-12-11
Upgrade Doctrine Bundle to Version 3

alexander-schranz:
Built on top of: https://github.com/sulu/sulu/pull/8451

| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | https://github.com/sulu/sulu/pull/8451, https://github.com/sulu/sulu/issues/8216
| License | MIT

#### What's in this PR?

Update Doctrine Bundle to Version 3 (only support for >=PHP 8.4 https://github.com/doctrine/DoctrineBundle/blob/3.1.0/composer.json#L32

#### Why?

ORM 3, DBAL 4 and Persistence 4 is now supported.

### DIFF
--- a/config/packages/doctrine.yaml
+++ b/config/packages/doctrine.yaml
@@ -6,7 +6,6 @@ doctrine:
         # either here or in the DATABASE_URL env var (see .env file)
         #server_version: '8.0.27'
     orm:
-        auto_generate_proxy_classes: true
         naming_strategy: doctrine.orm.naming_strategy.underscore_number_aware
         auto_mapping: false
 #        mappings:
@@ -26,7 +25,6 @@ when@test:
 when@prod: &prod
     doctrine:
         orm:
-            auto_generate_proxy_classes: false
             query_cache_driver:
                 type: pool
                 pool: doctrine.system_cache_pool
--- a/src/Sulu/Bundle/TestBundle/Resources/app/config/sulu.yml
+++ b/src/Sulu/Bundle/TestBundle/Resources/app/config/sulu.yml
@@ -16,9 +16,8 @@ framework:
 
 doctrine:
     orm:
-        auto_generate_proxy_classes: "%kernel.debug%"
         naming_strategy: doctrine.orm.naming_strategy.underscore
-        auto_mapping: true
+        auto_mapping: false
     dbal:
         url: '%env(resolve:DATABASE_URL)%'
         dbname_suffix: '_test%env(default::TEST_TOKEN)%'
--- a/src/Sulu/Bundle/TestBundle/Resources/app/config/sulu.yml
+++ b/src/Sulu/Bundle/TestBundle/Resources/app/config/sulu.yml
@@ -17,7 +17,6 @@ framework:
 doctrine:
     orm:
         naming_strategy: doctrine.orm.naming_strategy.underscore
-        auto_mapping: false
     dbal:
         url: '%env(resolve:DATABASE_URL)%'
         dbname_suffix: '_test%env(default::TEST_TOKEN)%'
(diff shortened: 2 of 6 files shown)


config/packages/doctrine.yaml
```diff
@@ -6,9 +6,8 @@ doctrine:
         # either here or in the DATABASE_URL env var (see .env file)
         #server_version: '8.0.27'
     orm:
-        auto_generate_proxy_classes: true
         naming_strategy: doctrine.orm.naming_strategy.underscore_number_aware
-        auto_mapping: true
+        auto_mapping: false
```
alexander-schranz, line 10: no other value as `false` is now accepted now on doctrine bundle version 3.

config/packages/doctrine.yaml
```diff
@@ -26,7 +25,6 @@ when@test:
 when@prod: &prod
     doctrine:
         orm:
-            auto_generate_proxy_classes: false
```
alexander-schranz, line 29: Unsure about this as you normally want this on `false` on production und only `true` on dev. With lazy object in PHP 8.4 this is not longer required but for PHP 8.2 and 8.3 we would require them and not even sure since which bundle version 8.4 would use native objects or if flag is required there.

config/packages/doctrine.yaml
alexander-schranz: Like in the new bundles we need also in the old once define where the mapping files are. Else we run into:

```
The class 'Sulu\Bundle\TagBundle\Entity\Tag' was not found in the chain con  
  figured namespaces Sulu\Snippet\Domain\Model, Sulu\Article\Domain\Model, Su  
  lu\Page\Domain\Model, Sulu\Bundle\ReferenceBundle\Domain\Model, Sulu\Bundle  
  \TrashBundle\Domain\Model, Sulu\Bundle\PreviewBundle\Domain\Model, Sulu\Cus  
  tomUrl\Domain\Model, Sulu\Route\Domain\Model, Sulu\Bundle\ActivityBundle\Do  
  main\Model, Gedmo\Tree\Entity            
```

See https://github.com/sulu/sulu/blob/46862495e915bfa03d49fa9d6810e4523bce6e15/packages/route/src/Infrastructure/Symfony/HttpKernel/SuluRouteBundle.php#L234-L252

Think this would be best as an own Pull Request to add prepend doctrine extension of the Resource/config/doctirine directory to every bundle.

@mamazu if you want to tackle that let me know.

src/Sulu/Bundle/TestBundle/Resources/app/config/sulu.yml
```diff
@@ -16,9 +16,8 @@ framework:
 
 doctrine:
     orm:
-        auto_generate_proxy_classes: "%kernel.debug%"
         naming_strategy: doctrine.orm.naming_strategy.underscore
-        auto_mapping: true
+        auto_mapping: false
```
alexander-schranz: same here we should keep true if doctrine bundle 2 is set

src/Sulu/Bundle/TestBundle/Resources/app/config/sulu.yml
```diff
@@ -16,9 +16,8 @@ framework:
 
 doctrine:
     orm:
-        auto_generate_proxy_classes: "%kernel.debug%"
```
alexander-schranz, line 19: we should keep this if doctrine/doctrine-bundle 2 is installed, so move this to a PHP based config.

Prokyonn (approved):

Issue #8455, opened 2025-12-14
Sulu 3.0 single_snippet_selection Resolution in Deeply Nested Blocks


dfsoeten:
| Q | A
| --- | ---
| Sulu Version | 3.0.0
| PHP Version | 8.4.15
| DB Version | 8.0.44
| Browser Version | 144.0.2

# Actual Behavior

The `single_snippet_selection` `<property>` does not resolve (properly) in deeply nested blocks (~3).

<img width="708" height="448" alt="Image" src="https://github.com/user-attachments/assets/1174a4b2-a7ee-417d-872f-a4f594c3bf7b" />

# Expected Behavior

The `single_snippet_selection` should resolve regardless of nesting level.

# Steps to Reproduce

See reproducable repo [here](https://github.com/dfsoeten/sulu-3.0-snippet-nesting-bug/blob/master/config/templates/pages/default.xml#L46).

*or*

Add the following XML to a page template:

```xml
<block name="nest1">
    <types>
        <type name="nest1">
            <meta>
                <title lang="en">Nest 1</title>
            </meta>
            <properties>
                <block name="nest2">
                    <types>
                        <type name="nest2">
                            <meta>
                                <title lang="en">Nest 2</title>
                            </meta>
                            <properties>
                                <block name="nest3">
                                    <types>
                                        <type name="nested_single_snippet_selection">
                                            <meta>
                                                <title lang="en">Nest 3</title>
                                            </meta>

                                            <properties>
                                                <property name="snippet" type="single_snippet_selection">
                                                    <meta>
                                                        <title lang="en">Snippet</title>
                                                    </meta>

                                                    <params>
                                                        <param name="types" value="default"/>
                                                    </params>
                                                </property>
                                            </properties>
                                        </type>
                                    </types>
                                </block>
                            </properties>
                        </type>
                    </types>
                </block>
            </properties>
        </type>
    </types>
</block>
```

# Possible Solutions

n/a

mamazu:
In the ContentResolverConfig there is a max_depth:
https://github.com/sulu/sulu/blob/f45de8f7168b6e40d2748e22e2b1d4247fea6ddd/packages/content/config/resolvers.xml#L36-L47

Maybe increasing this would help, but you're right this should be configurable.

alexander-schranz:
> Maybe increasing this would help, but you're right this should be configurable.

[...]

PR #8456, opened 2025-12-14
Make max resolver depth configurable

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | #8455 
| License | MIT
| Documentation PR | -

#### What's in this PR?
Adding a configuration parameter for the content resolver max depth.

#### Why?
It should not be hardcoded. And there was already a todo there.

#### Example Usage
```yaml
# config/packages/sulu_content.yaml
sulu_content:
    content_resolver:
        max_depth: 10
```

### DIFF
--- a/packages/content/config/resolvers.xml
+++ b/packages/content/config/resolvers.xml
@@ -40,8 +40,7 @@
             <argument type="service" id="sulu_content.resolvable_resource_replacer" />
             <argument type="service" id="sulu_content.content_view_data_normalizer" />
             <argument type="service" id="sulu_content.content_aggregator" />
-            <!--  TODO          <argument>%sulu_content.resource_resolver.max_depth%</argument>-->
-            <argument>5</argument>
+            <argument>%sulu_content.resource_resolver.max_depth%</argument>
             <argument type="service" id="sulu_content.content_enhancer"/>
         </service>
         <service id="Sulu\Content\Application\ContentResolver\ContentResolverInterface" alias="sulu_content.content_resolver"/>
--- a/packages/content/config/resolvers.xml
+++ b/packages/content/config/resolvers.xml
@@ -40,7 +40,7 @@
             <argument type="service" id="sulu_content.resolvable_resource_replacer" />
             <argument type="service" id="sulu_content.content_view_data_normalizer" />
             <argument type="service" id="sulu_content.content_aggregator" />
-            <argument>0</argument> <!-- See config for: sulu_content.content_resolver.max_depth -->
+            <argument /> <!-- See via SuluContentBundle config for: sulu_content.content_resolver.max_depth -->
             <argument type="service" id="sulu_content.content_enhancer"/>
         </service>
         <service id="Sulu\Content\Application\ContentResolver\ContentResolverInterface" alias="sulu_content.content_resolver"/>
--- a/packages/content/config/resolvers.xml
+++ b/packages/content/config/resolvers.xml
@@ -40,7 +40,7 @@
             <argument type="service" id="sulu_content.resolvable_resource_replacer" />
             <argument type="service" id="sulu_content.content_view_data_normalizer" />
             <argument type="service" id="sulu_content.content_aggregator" />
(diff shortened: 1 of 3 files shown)


packages/content/config/resolvers.xml
```diff
@@ -40,8 +40,7 @@
             <argument type="service" id="sulu_content.resolvable_resource_replacer" />
             <argument type="service" id="sulu_content.content_view_data_normalizer" />
             <argument type="service" id="sulu_content.content_aggregator" />
-            <!--  TODO          <argument>%sulu_content.resource_resolver.max_depth%</argument>-->
-            <argument>5</argument>
+            <argument>%sulu_content.resource_resolver.max_depth%</argument>
```
Prokyonn: I'm not really happy with the name, maybe something like `sulu_content.resolvable_resource.max_depth` although this would break our naming convention that is usually `<bundle>_<name>`

WDYT @alexander-schranz

packages/content/src/Infrastructure/Symfony/HttpKernel/SuluContentBundle.php
```diff
@@ -38,7 +38,16 @@ final class SuluContentBundle extends AbstractBundle
      */
     public function configure(DefinitionConfigurator $definition): void
     {
-        $definition->rootNode();
+        $definition->rootNode() // @phpstan-ignore-line
+            ->children()
+                ->arrayNode('resource_resolver')
```
Prokyonn: the example in the PR description does not match with this config

packages/content/config/resolvers.xml
alexander-schranz: Symfony mostly goes today without any parameters, maybe we should do the same.

After loading the `resolvers.xml`:

```php
$definition = $container->getDefinition('sulu_content.content_resolver');
$definition->setArgument('$argumentName', $config[...
```

This allows the container builder inline it directly.

packages/content/src/Infrastructure/Symfony/HttpKernel/SuluContentBundle.php
alexander-schranz: @Prokyonn we should maybe reflect the service name here.

```suggestion
                ->arrayNode('content_resolver')
```

Prokyonn: Sounds good for me 👍

mamazu: Fixed

packages/content/config/resolvers.xml
mamazu: Agreed and changed. Maybe this can be simplified when the entire config is in php

packages/content/config/resolvers.xml
```diff
@@ -40,8 +40,7 @@
             <argument type="service" id="sulu_content.resolvable_resource_replacer" />
             <argument type="service" id="sulu_content.content_view_data_normalizer" />
             <argument type="service" id="sulu_content.content_aggregator" />
-            <!--  TODO          <argument>%sulu_content.resource_resolver.max_depth%</argument>-->
-            <argument>5</argument>
+            <argument>0</argument> <!-- See config for: sulu_content.content_resolver.max_depth -->
```
alexander-schranz: Does this also work, so we have no 0 here as its kind of misleading:

```suggestion
            <argument><!-- See via SuluContentBundle config for: sulu_content.content_resolver.max_depth --></argument> 
```

Prokyonn: @mamazu can you please verify this

mamazu: Removed the unintuitive default value.

Prokyonn:
looks good from my side, WDYT @alexander-schranz ?

Prokyonn (approved):

PR #8459, opened 2025-12-16
Add a default collection builder for creating test collections in dev

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?
Adding a builder to create one default media collection.

#### Why?
When setting up sulu for testing it would be nice to also have a default collection since you can't upload anything without one.

#### Example Usage
`bin/console sulu:build default_collection`

### DIFF
--- a/src/Sulu/Component/Media/SystemCollections/DefaultCollectionBuilder.php
+++ b/src/Sulu/Component/Media/SystemCollections/DefaultCollectionBuilder.php
@@ -1,0 +1,74 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of Sulu.
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Component\Media\SystemCollections;
+
+use Massive\Bundle\BuildBundle\Build\BuilderContext;
+use Massive\Bundle\BuildBundle\Build\BuilderInterface;
+use Sulu\Bundle\MediaBundle\Api\Collection;
+use Sulu\Bundle\MediaBundle\Collection\Manager\CollectionManagerInterface;
+use Sulu\Component\Webspace\Manager\WebspaceManagerInterface;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+readonly class DefaultCollectionBuilder implements BuilderInterface
+{
+    private ?InputInterface $input;
+    private ?OutputInterface $output;
+
+    public function __construct(
+        private CollectionManagerInterface $collectionManager,
+        private WebspaceManagerInterface $webspaceManager,
+    ) {
+    }
+
+    public function getName(): string
+    {
+        return 'default_collection';
+    }
+
+    public function getDependencies(): array
+    {
+        return ['database', 'fixtures'];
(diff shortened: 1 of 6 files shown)


src/Sulu/Component/Media/SystemCollections/DefaultCollectionBuilder.php
```diff
@@ -0,0 +1,78 @@
...
+            $this->output?->writeln('Updating test collection for ' . $localization);
+
+            $existingId = $this->collectionManager->getByKey('test_collection', $locale)?->getId();
+            if (null !== $existingId && $this->input?->getOption('destroy')) {
+                $this->collectionManager->delete($existingId);
+                $existingId = null;
+            }
+
+            $this->collectionManager->save([
+                'title' => 'Test Collection',
```
Prokyonn: I wouldn't name it Test Collection, probably something like `Default Collection`

@sulu/core-developer WDYT?

alexander-schranz: Would not use "collection" postfix here. Im not a fan of test or default would go with "Content" / "Inhalt". But @chirimoya should decide, also even if he want it to have such collection or not.

chirimoya: When creating a collection for testing, I usually name it `Images`.

mamazu: The only thing I would like to add is, this collection is only created in dev for faster testing. Maybe the name should also reflect that. (not that people wonder why it's not there in prod)

chirimoya: From my experience, a lot of projects get initialised with `bin/adminconsole sulu:build dev`. And I don’t want a test collection to exist in production.

mamazu: Okay, updated the name.

mamazu:
Renamed to "Images"

alexander-schranz:
I stumbled today over that in some previous releases we already added the `Add collection` button to the media selection overlay. So its not that an upload a media is blocked by something.

I'm unsure to add a default collection as its better force content manager by begin starting its own structure. Instead of forcing some general image, documents, ... structure, which they may think need to follow then or are afraid to remove that kind of collection. 🤔 

@sulu/core-developer what do you think?

Bildschirmfoto 2026-01-16 um 13 39 14

Maybe in future we could add some "Install test data" to the skeleton when we may introduce a Installer or something like that?

wachterjohannes:
@alexander-schranz @mamazu i would also avoid adding this by default.  the only option to which i see now is to add a target dev_data (or a similar name) which would add this and maybe other in the future

mamazu:
So the current implemenation already adds a build target to the `sulu:build dev` command.

I still think that this is a valuable Addition, as with a clean Sulu installation. When you click on media and then try to upload a media, you can not select any destination because the only existing collection is a system collection and there is also no way of creating a new one from here:
image

mamazu:
Closed, we don't need this. Projects should have this in their own fixture set.

PR #8466, opened 2025-12-18
Use ClockInterface in TimestampableProvider

### LINKED ISSUE #8115: Inject and use Clock service into TimestampableSubscriber
# Problem description

The `TimestampableSubscriber` should use the Clock Service to make setting created and changed in tests easier.

https://github.com/sulu/sulu/blob/3.0/src/Sulu/Component/Persistence/EventSubscriber/ORM/TimestampableSubscriber.php

# Proposed solution

Require clock psr and symfony clock component and inject the clock service. I'm not sure if I would use the ClockTrait it doesn't feel good so more would go with directly inject it via Constructor?

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | yes
| Fixed tickets | fixes #8115 
| Related issues/PRs | -
| License | MIT

#### What's in this PR?
Using the ClockInterface for the timestampable subscriber.

#### Why?
Making the tests better.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -70,6 +70,7 @@
         "symfony-cmf/routing-bundle": "^3.0",
         "symfony/asset": "^6.4 || ^7.1",
         "symfony/cache": "^6.4 || ^7.1",
+        "symfony/clock": "^6.4 || ^7.1",
         "symfony/config": "^6.4 || ^7.1",
         "symfony/console": "^6.4 || ^7.1",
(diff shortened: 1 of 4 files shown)


mamazu:
The clock returns a DateTimePoint which the serializer handles different from a DateTimeImmutable. What is the best course of action, change the serializer or fix the tests?

PR #8471, opened 2025-12-21
Prevent Media from being moved multiple times

### LINKED ISSUE #8052: Sulu moves media multiple times on subsequent move execution via Frontend
| Q | A
| --- | ---
| Sulu Version | 2.6.9
| PHP Version | 8.3.6
| Browser Version | Chrome 138.0.7204.101

# Actual Behavior

If a user selects and moves an image via the media interface and then selects an image again and moves it to another folder, the first image is also moved to the last selected folder, unless the user has changed folders in between.

# Expected Behavior

The first image should stay in the first selected target directory and not also be moved into the second target directory.

# Steps to Reproduce

- Have 2 images ("image1", "image2") in "sample collection".
- Have two collections "collection 1", "collection 2"
- Select "image1" and select "move selected" and choose "collection 1" as target
- Select "image2" and select "move selected" and choose "collection 2" as target

Both images are now in "collection 2"

# Possible Solutions

I don't know :-(

MarkusHolstein:
#8052

| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #8052
| Related issues/PRs | 
| License | MIT

#### What's in this PR?

Fixes #8052 .

#### Why?

Fixes a problem in media management, when moving multiple images on subsequent calls. Image selection was not cleared after moving them to a new collection. So moving another image to another collection resulted in image1  being moved to the same collection as the last moved image.

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/Resources/js/views/MediaOverview/MediaOverview.js
+++ b/src/Sulu/Bundle/MediaBundle/Resources/js/views/MediaOverview/MediaOverview.js
@@ -189,6 +189,7 @@ class MediaOverview extends React.Component<ViewProps> {
             this.collectionListStore.reload();
             this.showMediaMoveOverlay = false;
             this.mediaMoving = false;
+            this.mediaListStore.clearSelection();
         }));
     };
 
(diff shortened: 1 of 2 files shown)


alexander-schranz:
@MarkusHolstein thank you!

PR #8473, opened 2025-12-22
Add proper return types for the RequestAnalyzerInterface

Prokyonn:
| Q                  | A               |
  |--------------------|-----------------|
  | Bug fix?           | no              |
  | New feature?       | no              |
  | BC breaks?         | yes             |
  | Deprecations?      | no              |
  | Fixed tickets      | fixes #         |
  | Related issues/PRs | #               |
  | License            | MIT             |
  | Documentation PR   | sulu/sulu-docs# |

  What's in this PR?

  Updates RequestAnalyzerInterface methods to return nullable types, as the request analyzer is not guaranteed to have a request context available (e.g., in console commands).

  Why?

  The RequestAnalyzer retrieves values from the current HTTP request. When code is executed outside of a request context (CLI commands, etc.), there is no request available and these methods return null. Previously, the interface declared non-nullable return types which could lead to runtime errors when code assumed these values would always be present.

  This change makes the nullability explicit in the interface, forcing consumers to handle the case where no request context exists.

### DIFF
--- a/src/Sulu/Bundle/WebsiteBundle/Locale/PortalDefaultLocaleProvider.php
+++ b/src/Sulu/Bundle/WebsiteBundle/Locale/PortalDefaultLocaleProvider.php
@@ -30,12 +30,6 @@ class PortalDefaultLocaleProvider implements DefaultLocaleProviderInterface
 
     public function getDefaultLocale()
     {
-        $portal = $this->requestAnalyzer->getPortal();
-
-        if (null === $portal) {
-            return null;
-        }
-
-        return $portal->getDefaultLocalization();
+        return $this->requestAnalyzer->getPortal()?->getDefaultLocalization();
     }
 }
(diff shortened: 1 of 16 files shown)


src/Sulu/Bundle/WebsiteBundle/Locale/PortalDefaultLocaleProvider.php
```diff
@@ -30,6 +30,12 @@ public function __construct(RequestAnalyzerInterface $requestAnalyzer)
...
     public function getDefaultLocale()
     {
-        return $this->requestAnalyzer->getPortal()->getDefaultLocalization();
+        $portal = $this->requestAnalyzer->getPortal();
+
+        if (null === $portal) {
+            return null;
+        }
+
+        return $portal->getDefaultLocalization();
```
mamazu: Why not `$this->requestAnlayser->getPortal()?->getDefaultLocalization()`?

alexander-schranz (approved):

PR #8477, opened 2025-12-28
fix: Resolve snippets with STAGE_LIVE to fix preview ContentNotFoundException

### LINKED ISSUE #8476: [Bug] Preview fails with ContentNotFoundException when page uses snippet_selection
## Description

When previewing a page that contains a `snippet_selection` property, the preview fails with a `ContentNotFoundException`. This happens because `ContentResolver` re-aggregates snippets using the parent's stage (`draft` in preview mode), but snippets only have `live` dimension content available.

## Sulu Version
- Sulu 3.0.1
- Symfony 7.2

## Steps to Reproduce

1. Create a snippet template (e.g., `beer.xml`)
2. Create a page template with a `snippet_selection` property:
```xml
<property name="beers" type="snippet_selection">
    <params>
        <param name="types" value="beer"/>
    </params>
</property>
```
3. Create and **publish** a snippet
4. Create a page, select the published snippet in the `snippet_selection` field
5. Observe the preview panel - it shows an error

## Expected Behavior

The preview should display the page with the resolved snippet content.

## Actual Behavior

Preview fails with:
```
ContentNotFoundException: Could not load content with id "019b6573-9aa4-7188-899a-f0d2c31084f7" and attributes: {"locale":"en","stage":"draft"}
```

## Root Cause Analysis

The issue is in `ContentResolver.php` at lines 75-80:

```php
if ($resource instanceof ContentRichEntityInterface) {
    $childContent = $this->contentAggregator->aggregate($resource, [
        'locale' => $locale,
        'stage' => $stage,  // BUG: Uses parent's stage (draft in preview)
    ]);
```

**Flow:**
1. `SnippetResourceLoader` correctly loads snippets with `STAGE_LIVE`
2. It returns `Snippet` entities (which implement `ContentRichEntityInterface`)
3. `ContentResolver` sees these are `ContentRichEntityInterface` instances
4. It re-aggregates them using the **parent's stage** (`draft` in preview mode)
5. `ContentAggregator` can't find dimension content with `stage=draft` for the snippet
6. `ContentNotFoundException` is thrown

**Why snippets don't have draft content:**
Snippets are typically only published (live), not kept in draft state like pages being edited. When a user selects a snippet in a page, they expect to see the published version.

## Suggested Fix

Snippets should always be resolved with `STAGE_LIVE`, regardless of the parent's stage. Possible solutions:

### Option 1: Check resource type in ContentResolver
```php
if ($resource instanceof ContentRichEntityInterface) {
    $childStage = $resource instanceof SnippetInterface 
        ? DimensionContentInterface::STAGE_LIVE 
        : $stage;
    
    $childContent = $this->contentAggregator->aggregate($resource, [
        'locale' => $locale,
        'stage' => $childStage,
    ]);
```

### Option 2: Have SnippetResourceLoader return resolved DimensionContent
Instead of returning the `Snippet` entity, return the already-aggregated `DimensionContent` with `STAGE_LIVE`. This bypasses the `ContentRichEntityInterface` check.

## Workaround

We implemented a custom `SnippetResourceLoader` that fully resolves snippets with `STAGE_LIVE` and returns the resolved data structure, bypassing the re-aggregation:

```php
class PreviewSafeSnippetResourceLoader implements ResourceLoaderInterface
{
    public function load(array $ids, ?string $locale, array $params = []): array
    {
        // Load snippets with STAGE_LIVE
        $snippets = $this->snippetRepository->findBy([
            'uuids' => $ids,
            'locale' => $locale,
            'stage' => DimensionContentInterface::STAGE_LIVE,
        ], [], [SnippetRepositoryInterface::GROUP_SELECT_SNIPPET_WEBSITE => true]);

        $contentResolver = $this->container->get('sulu_content.content_resolver');

        $mappedResult = [];
        foreach ($snippets as $snippet) {
            $dimensionContent = $this->contentAggregator->aggregate($snippet, [
                'locale' => $locale,
                'stage' => DimensionContentInterface::STAGE_LIVE,
            ]);
            
            // Return fully resolved data instead of entity
            $mappedResult[$snippet->getId()] = $contentResolver->resolve($dimensionContent);
        }

        return $mappedResult;
    }
}
```

## Related

This may be related to similar issues with referenced content in preview mode.

james2001:
## Summary

This PR fixes a bug where pages using `snippet_selection` fail to render in the admin preview with a `ContentNotFoundException`.

**Root cause:** When previewing a page, `ContentResolver` re-aggregates any `ContentRichEntityInterface` objects (including snippets) using the parent's stage parameter. In preview mode, this stage is `STAGE_DRAFT`, but snippets typically only have `STAGE_LIVE` content, causing the aggregation to fail.

**Solution:** `SnippetResourceLoader` now fully resolves snippets with `STAGE_LIVE` and returns the resolved data structure (with `content`, `view`, `resource`, `extension` keys) instead of the raw `Snippet` entity. This bypasses the problematic re-aggregation in `ContentResolver`.

## Changes

- `SnippetResourceLoader`: Now uses `ContentAggregator` and `ContentResolver` to fully resolve snippets before returning them
- `SuluSnippetBundle`: Added `ContentAggregator` and `ContentResolver` as dependencies for the resource loader service

## Related Issue

Fixes #8476

## Test Plan

1. Create a snippet (e.g., a "Beer" snippet)
2. Publish the snippet
3. Create a page with a `snippet_selection` field referencing the snippet
4. Open the page in admin preview mode
5. ✅ The page should render correctly with snippet data displayed
6. ✅ No `ContentNotFoundException` should occur

### DIFF
--- a/packages/snippet/src/Infrastructure/Sulu/Content/ResourceLoader/SnippetResourceLoader.php
+++ b/packages/snippet/src/Infrastructure/Sulu/Content/ResourceLoader/SnippetResourceLoader.php
@@ -13,6 +13,8 @@ declare(strict_types=1);
 
 namespace Sulu\Snippet\Infrastructure\Sulu\Content\ResourceLoader;
 
+use Sulu\Content\Application\ContentAggregator\ContentAggregatorInterface;
+use Sulu\Content\Application\ContentResolver\ContentResolverInterface;
 use Sulu\Content\Application\ResourceLoader\Loader\ResourceLoaderInterface;
 use Sulu\Content\Domain\Model\DimensionContentInterface;
 use Sulu\Snippet\Domain\Repository\SnippetRepositoryInterface;
@@ -28,6 +30,8 @@ class SnippetResourceLoader implements ResourceLoaderInterface
 
     public function __construct(
         private SnippetRepositoryInterface $snippetRepository,
+        private ContentAggregatorInterface $contentAggregator,
+        private ContentResolverInterface $contentResolver,
     ) {
     }
 
@@ -52,7 +56,17 @@ class SnippetResourceLoader implements ResourceLoaderInterface
 
         $mappedResult = [];
         foreach ($result as $snippet) {
-            $mappedResult[$snippet->getId()] = $snippet;
+            // Aggregate the snippet with STAGE_LIVE to get the DimensionContent
+            $dimensionContent = $this->contentAggregator->aggregate($snippet, [
+                'locale' => $locale,
+                'stage' => DimensionContentInterface::STAGE_LIVE,
+            ]);
+
+            // Fully resolve the snippet content and return the resolved data structure.
+            // This ensures snippets are always resolved with STAGE_LIVE, regardless of
+            // the parent's stage (e.g., draft in preview mode), preventing
+            // ContentNotFoundException when previewing pages with snippet_selection.
+            $mappedResult[$snippet->getId()] = $this->contentResolver->resolve($dimensionContent);
         }
 
         return $mappedResult;
--- a/packages/snippet/src/Infrastructure/Symfony/HttpKernel/SuluSnippetBundle.php
+++ b/packages/snippet/src/Infrastructure/Symfony/HttpKernel/SuluSnippetBundle.php
@@ -337,6 +337,8 @@ final class SuluSnippetBundle extends AbstractBundle
             ->class(SnippetResourceLoader::class)
             ->args([
                 new Reference('sulu_snippet.snippet_repository'),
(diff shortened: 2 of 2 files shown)


Prokyonn:
Hey @james2001 thanks for your PR.

The ResourceLoaders are not responsible for resolving the content. These should be simple services that only load the entity from the database. Especially because they do not have all the necessary informations, e.g. when only specific properties should be loaded.

The issue is that in the ContentResolver we did try to fetch nested resources with `stage="stage"` instead of `stage="live"`.

The issue should be properly fixed in https://github.com/sulu/sulu/pull/8489

PR #8481, opened 2025-12-30
Update UPGRADE of ro_routes table query to include webspace

zangai:
Updated the 'ro_routes' table query to replace 'site' with 'webspace'. This adheres to the change in the Route model as done in: https://github.com/sulu/sulu/pull/8418

| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

This PR changes the create query for the new ro_routes table to use the `webspace` field instead of the `site` field. As this change was done after the current version of the UPGRADE-3 doc.

#### Why?

The old query causes trouble during the migration.

### DIFF
--- a/UPGRADE-3.x.md
+++ b/UPGRADE-3.x.md
@@ -305,7 +305,7 @@ ALTER TABLE ro_routes RENAME TO ro_routes_old;
 And create a new `ro_routes` table with the following structure:
 
 ```sql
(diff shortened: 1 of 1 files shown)


alexander-schranz:
@zangai thank you!

PR #8487, opened 2026-01-07
Fix Saving zero value on SimpleContentType

draconivis:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

Add strict comparison in the SimpleContentType so one can save values like 0.

#### Why?

When saving a SingleSelect block where the parameter is 0, the value isn't persisted and thus it falls back to the default value on reload.

#### Example Usage

```xml
<properties>
    <property name="size" type="single_select">
        <meta>
            <title>Size</title>
        </meta>

        <params>
            <param name="default_value" value="25"/>

                <param name="0"><meta><title lang="en">0%</title></meta></param>
                <param name="25"><meta><title lang="en">25%</title></meta></param>
                <param name="100"><meta><title lang="en">100%</title></meta></param>
            </param>
        </params>
    </property>
</properties>
```

### DIFF
--- a/src/Sulu/Bundle/PageBundle/Tests/Unit/Content/Types/SingleSelectTest.php
+++ b/src/Sulu/Bundle/PageBundle/Tests/Unit/Content/Types/SingleSelectTest.php
@@ -1,0 +1,125 @@
+<?php
+
+/*
+ * This file is part of Sulu.
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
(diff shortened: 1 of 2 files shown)


alexander-schranz:
@draconivis can you add a test case for this?

alexander-schranz:
@draconivis can you also check if we require anything in 3.0? Mostly I would think there it should just work as we don't have an own write process and save the data as it come from the frontend and have no specific resolver.

draconivis:
@alexander-schranz i added the property to a template in 3.0 and it is working there, so no adjustments needed.

PR #8493, opened 2026-01-08
Add default properties for smart content resolving

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Related issues/PRs | https://github.com/sulu/sulu/pull/8492
| License | MIT

#### What's in this PR?

Adds default properties for content resolving for the page / article smart content.

#### Why?

In Sulu 2.6 this was already the case and this prevents from unknowingly loading a lot of unnecessary data like nested entities that will otherwise be resolved.

### DIFF
--- a/packages/content/src/Application/SmartResolver/Resolver/SmartContentSmartResolver.php
+++ b/packages/content/src/Application/SmartResolver/Resolver/SmartContentSmartResolver.php
@@ -144,11 +144,11 @@ class SmartContentSmartResolver implements SmartResolverInterface
         ];
 
         $configuration = $smartContentProvider->getConfiguration();
-        $configProperties = $configuration->getProperties();
+        $defaultProperties = $configuration->getDefaultProperties();
         /** @var array<string, string> $passedProperties */
         $passedProperties = $params['properties'] ?? [];
-        $properties = null !== $configProperties
-            ? \array_replace($configProperties, $passedProperties)
+        $properties = null !== $defaultProperties
+            ? \array_replace($defaultProperties, $passedProperties)
             : ($passedProperties ?: null);
 
         return ContentView::createResolvables(
--- a/packages/content/src/Application/SmartResolver/Resolver/SmartContentSmartResolver.php
+++ b/packages/content/src/Application/SmartResolver/Resolver/SmartContentSmartResolver.php
@@ -143,12 +143,25 @@ class SmartContentSmartResolver implements SmartResolverInterface
             'excluded' => [],
         ];
 
+        $configuration = $smartContentProvider->getConfiguration();
+        $configProperties = $configuration->getProperties();
+        /** @var array<string, string>|null $passedProperties */
+        $passedProperties = $params['properties'] ?? null;
+        $passedProperties = \is_array($passedProperties) ? $passedProperties : null;
+
+        $properties = match (true) {
+            null === $configProperties && null === $passedProperties => null,
+            null === $configProperties => $passedProperties,
+            null === $passedProperties => $configProperties,
+            default => \array_merge($configProperties, $passedProperties),
+        };
+
         return ContentView::createResolvables(
             ids: \array_map(static fn (array $item) => $item['id'], $result),
             resourceLoaderKey: $smartContentProvider->getResourceLoaderKey(),
             view: $view,
             metadata: [
-                'properties' => $params['properties'] ?? null,
+                'properties' => $properties,
             ]
         );
     }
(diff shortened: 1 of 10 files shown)


src/Sulu/Bundle/AdminBundle/SmartContent/Configuration/BuilderInterface.php
```diff
@@ -108,6 +108,15 @@ public function enableSorting(array $sorting);
...
     public function enableView(string $view, array $resultToView);
 
+    /**
+     * Sets properties to be included when resolving smart content items.
+     *
+     * @param array<string, string> $properties
+     *
+     * @return BuilderInterface
+     */
+    public function enableProperties(array $properties);
```
Prokyonn: Would `enableDefaultProperties` be better?

src/Sulu/Bundle/AdminBundle/SmartContent/Configuration/ProviderConfiguration.php
```diff
@@ -239,4 +244,25 @@ public function setResultToView(?array $resultToView): void
...
+        return $this->properties;
+    }
+
+    /**
+     * @param array<string, string> $properties
+     */
+    public function setProperties(array $properties): void
+    {
+        $this->properties = $properties;
+    }
```
alexander-schranz, line 262: Would allow null:

```suggestion
    /**
     * @param array<string, string>|null $properties
     */
    public function setProperties(?array $properties): void
    {
        $this->properties = $properties;
    }
```

src/Sulu/Bundle/AdminBundle/SmartContent/Configuration/ProviderConfiguration.php
```diff
@@ -239,4 +244,25 @@ public function setResultToView(?array $resultToView): void
     {
         $this->resultToView = $resultToView;
     }
+
+    public function hasProperties(): bool
+    {
+        return null !== $this->properties && [] !== $this->properties;
+    }
```
alexander-schranz: Not sure if we should add has here.

packages/content/src/Application/SmartResolver/Resolver/SmartContentSmartResolver.php
```diff
@@ -143,12 +143,25 @@ public function resolve(SmartResolvable $resolvable, ?string $locale = null): Co
...
+        /** @var array<string, string>|null $passedProperties */
+        $passedProperties = $params['properties'] ?? null;
+        $passedProperties = \is_array($passedProperties) ? $passedProperties : null;
+
+        $properties = match (true) {
+            null === $configProperties && null === $passedProperties => null,
+            null === $configProperties => $passedProperties,
+            null === $passedProperties => $configProperties,
+            default => \array_merge($configProperties, $passedProperties),
+        };
```
alexander-schranz: The match looks complex, why not just merge it directly?

```suggestion
        $properties = null;
        $configProperties = $configuration->getProperties();
        /** @var array<string, string>$passedProperties */
        $passedProperties = $params['properties'] ?? [];
        if (null !== $configProperties) { // even nothing is passed if enabled we can do:
            $properties = \array_replace(
                $configProperties,
                $passedProperties,
            );
        }
```

alexander-schranz (changes requested):

packages/content/src/Application/SmartResolver/Resolver/SmartContentSmartResolver.php
alexander-schranz: Eventually go with `$properties = [];` to resolve nothing as default to avoid overfetch in smart content providers.

Edgecase would be snippet where we want to resolve all by default. Complicated topic need think about it, lets discuss tomorrow.

src/Sulu/Bundle/AdminBundle/SmartContent/Configuration/BuilderInterface.php
alexander-schranz: from interface point it enables the properties features and you would give defaultProperties to it, so maybe: ?

```
public function enableProperties(array $defaultProperties);
```

src/Sulu/Bundle/AdminBundle/SmartContent/Configuration/ProviderConfiguration.php
```diff
@@ -239,4 +244,20 @@ public function setResultToView(?array $resultToView): void
...
+        return $this->properties;
+    }
+
+    /**
+     * @param array<string, string>|null $properties
+     */
+    public function setProperties(?array $properties): void
+    {
+        $this->properties = $properties;
+    }
```
alexander-schranz, line 262: should we then rename this cases to defaultProperties param and method name?

alexander-schranz (approved):

alexander-schranz (approved):

PR #8494, opened 2026-01-08
Fix previously set where conditions being overridden in MediaSmartContentProvider

### LINKED ISSUE #8433: Unable to filter media smart content using mimetype and/or type query string parameters

NeuralClone:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #8433
| Related issues/PRs | #8437
| License | MIT
| Documentation PR | --

#### What's in this PR?

This PR fixes the bug described in #8433 on the 3.0 branch.

#### Why?

To allow the filtering of media smart content with the [documented query string parameters](https://docs.sulu.io/en/3.0/reference/property-types/smart_content.html#media) and prevent previously set `where` conditions from being overridden.

#### Tests

I wasn't quite sure how to handle tests for this. To trigger the bug, the `includeSubFolders` filter has to be set. The existing tests for MediaSmartContentProvider don't seem to be setup with that in mind. But I admit that my experience with Sulu 3.0's changes is quite limited.

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/Infrastructure/Sulu/Content/SmartContent/MediaSmartContentProvider.php
+++ b/src/Sulu/Bundle/MediaBundle/Infrastructure/Sulu/Content/SmartContent/MediaSmartContentProvider.php
@@ -270,7 +270,7 @@ readonly class MediaSmartContentProvider implements SmartContentProviderInterfac
                         Join::WITH,
                         'parentCollection.id = :collectionId',
                     )
-                    ->where('collection.lft BETWEEN parentCollection.lft AND parentCollection.rgt')
+                    ->andWhere('collection.lft BETWEEN parentCollection.lft AND parentCollection.rgt')
                     ->setParameter('collectionId', $filters['dataSource']);
             }
         }


alexander-schranz:
Merging this to get it as part of the release. Still if you have some time would be nice to have a test also for 3.0 for this issue. Curious why https://github.com/sulu/sulu/blob/1f433fe19e21cdb5a9572e95ccb6af5c82668601/src/Sulu/Bundle/MediaBundle/Tests/Functional/Infrastructure/Sulu/Content/SmartContent/MediaSmartContentProviderTest.php#L747-L748 didn't catch that and ended in the same error as you had.

NeuralClone:
I was thinking it might take a while to look into this and that I'd have to wait until I had more time in a few weeks. But the test you linked to helped a ton.

I can reliably trigger the error by reverting my changes and adding these filters to the test you linked to:

```php
[
    // ...
    'includeSubFolders' => true,
    'mimetype' => 'image/svg+xml',
]
```

This adds the `mimetype` filter to the query here: 
https://github.com/sulu/sulu/blob/e8b863a6f26064b16b3a49251890601d6db7c72d/src/Sulu/Bundle/MediaBundle/Infrastructure/Sulu/Content/SmartContent/MediaSmartContentProvider.php#L255-L259

The `includeSubFolders` filter then forces the else statement to be executed:

https://github.com/sulu/sulu/blob/e8b863a6f26064b16b3a49251890601d6db7c72d/src/Sulu/Bundle/MediaBundle/Infrastructure/Sulu/Content/SmartContent/MediaSmartContentProvider.php#L262-L275

That else block then overrides the previously set `where` and causes the same `QueryException` to be thrown that I was originally getting. 

I'll see if I can add a test to address this in the next few days. Thank you for linking to that specific test. It saved me some time!

PR #8501, opened 2026-01-09
Fix Template overrides fail due to missing copy logic

### LINKED ISSUE #8500: Template overrides fail due to missing copy logic
| Q | A
| --- | ---
| Sulu Version | 3.0.1
| PHP Version | 8.3.3
| DB Version | MySQL 5.6
| Browser Version | N/A (Backend issue)

# Actual Behavior

When overriding a custom template (e.g., `event.xml`) in a project's `config/templates/` directory, the `FormMetadata::merge()` method triggers two issues regarding the `template` property (which holds `<view>`, `<controller>`, and `<cacheLifetime>` configuration).

**1. Initialization Error**
Accessing the metadata for the overridden template (e.g., via a custom `RouteDefaultsProvider`) throws the following error:
```
Typed property Sulu\Bundle\AdminBundle\Metadata\FormMetadata\FormMetadata::$template must not be accessed before initialization

```

This happens because the `$template` property is declared as `private ?TemplateMetadata $template;` but not initialized by default.

**2. Data Loss during Merge**
Even if the uninitialized property issue is resolved (e.g., by initializing it with `null`), a secondary logical error persists. The system subsequently throws the following exception:

```
No template metadata found for template "event"
```

**Root Cause:**
The `FormMetadata::merge()` method creates a new `FormMetadata` instance but fails to transfer the `$template` property (which holds the `TemplateMetadata` object) from the source objects to the merged result. Consequently, the merged metadata object loses all information regarding `<view>`, `<controller>`, and `<cacheLifetime>`.

# Expected Behavior

When two templates with the same key are merged (bundle template + project override), the `<view>`, `<controller>`, and `<cacheLifetime>` configuration should be preserved in the merged `FormMetadata` object.

Template overrides should work seamlessly, allowing projects to customize templates while retaining routing metadata.

# Steps to Reproduce

1. Create a custom bundle with a content type (e.g., `event`) that registers templates via `sulu_admin.templates` configuration
2. Add `<view>`, `<controller>`, and `<cacheLifetime>` tags to the bundle's template XML
3. Override the template in the project's `config/templates/` directory (same `<key>`)
4. Implement a `RouteDefaultsProvider` that calls `$formMetadata->getTemplate()`
5. Access a published entity's route on the website

**Result:** `Error: Typed property ... must not be accessed before initialization`

# Possible Solutions

In `Sulu\Bundle\AdminBundle\Metadata\FormMetadata\FormMetadata`:

1. Initialize the property with `null`:
```php
// Before
private ?TemplateMetadata $template;

// After
private ?TemplateMetadata $template = null;
```

2. Copy the template in the `merge()` method:
```php
public function merge(self $otherForm): self
{
    // ... existing logic ...

    // Copy template metadata (view, controller, cacheLifetime)
    try {
        if ($otherForm->getTemplate()) {
            $mergedForm->setTemplate($otherForm->getTemplate());
        } elseif ($this->getTemplate()) {
            $mergedForm->setTemplate($this->getTemplate());
        }
    } catch (\Error) {
        // Neither has template initialized - ignore
    }

    return $mergedForm;
}
```

This ensures that template routing configuration is preserved when templates are merged during the metadata loading process.

manuxi:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no 
| Deprecations? | no
| License | MIT

#### Why?

Fixes #8500


src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/FormMetadata.php
```diff
@@ -262,6 +262,12 @@ public function merge(self $otherForm): FormMetadata
         $mergedForm->setResources(\array_merge($this->getResources(), $otherForm->getResources()));
         $mergedForm->setSchema($this->getSchema()->merge($otherForm->getSchema()));
 
+        if ($otherForm->getTemplate()) {
+            $mergedForm->setTemplate($otherForm->getTemplate());
+        } elseif ($this->getTemplate()) {
+            $mergedForm->setTemplate($this->getTemplate());
+        }
+
```
mamazu: I would recommend a shorter version.
```suggestion
        $mergedFormTemplate = $otherForm->getTemplate() ?? $this->getTemplate();
        if ($mergedFormTemplate) {
            $mergedForm->setTemplate($mergedFormTemplate);
        }

```

src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/FormMetadata.php
```diff
@@ -262,6 +262,11 @@ public function merge(self $otherForm): FormMetadata
         $mergedForm->setResources(\array_merge($this->getResources(), $otherForm->getResources()));
         $mergedForm->setSchema($this->getSchema()->merge($otherForm->getSchema()));
 
+        $mergedFormTemplate = $otherForm->getTemplate() ?? $this->getTemplate();
+        if ($mergedFormTemplate) {
+            $mergedForm->setTemplate($mergedFormTemplate);
+        }
```
alexander-schranz: The `TemplateMetadata` should have its own merge method which creates the new instance, something like:

```suggestion
        $formTemplate = $this->getTemplate();
        $otherFormTemplate = $otherForm->getTemplate();
        $mergedFormTemplate = null;
        
        if ($formTemplate instanceof ... || $otherFormTemplate instanceOf ...) {
                $formTemplate ??= new TemplateMetadata();
                $otherFormTemplate ??= new TemplateMetadata();
                
                $mergedFormTemplate = $formTemplate->merge($otherFormTemplate)
        }

        $mergedForm->setTemplate($mergedFormTemplate);
```

As we always need to to create a new instance of TemplateMetadata not reuse existing object here.

src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/FormMetadata.php
```diff
@@ -36,7 +36,7 @@ class FormMetadata extends AbstractMetadata
     private SchemaMetadata $schema;
 
     #[Exclude()]
-    private ?TemplateMetadata $template;
+    private ?TemplateMetadata $template = null;
```
alexander-schranz: Revert this. It should not longer be a problem if correctly handle in merge. If not defined it would mean again that there is another bug and better run then again in the unitialize issue instead of returning false metadata.


```suggestion
    private ?TemplateMetadata $template;
```

alexander-schranz (changes requested):
Thx for the pull request some small changes:

src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/FormMetadata.php
manuxi: Lint seems to fail with #message: "Typed property Sulu\Bundle\AdminBundle\Metadata\FormMetadata\FormMetadata::$template must not be accessed before initialization"

alexander-schranz (approved):

src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/FormMetadata.php
alexander-schranz: will have a look at it else changes look fine.

src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/TemplateMetadata.php
```diff
@@ -49,4 +49,13 @@ public function setCacheLifetime(?CacheLifetimeMetadata $cacheLifetime): void
     {
         $this->cacheLifetime = $cacheLifetime;
     }
+
+    public function merge(self $other): self
+    {
+        return new self(
+            $other->getController() ?: $this->getController(),
+            $other->getView() ?: $this->getView(),
+            $other->getCacheLifetime() ?? $this->getCacheLifetime()
```
alexander-schranz: One question why is this not `??` is there an error in the parser returning empty string even if the tag does not exist?

alexander-schranz:
@manuxi Can you check if the last changes works for you?

src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/TemplateMetadata.php
manuxi: I use ?: for controller and view because empty strings should be considered ‘not set’.

alexander-schranz: Would not do that. If somebody don't want override it they should just not write the tag inside there template. Adopted to check only for null here.

alexander-schranz (approved):

manuxi (commented):
Looks good to me

src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/TemplateMetadata.php
```diff
@@ -49,4 +49,19 @@ public function setCacheLifetime(?CacheLifetimeMetadata $cacheLifetime): void
     {
         $this->cacheLifetime = $cacheLifetime;
     }
+
+    public function merge(self $other): self
+    {
+        $mergedCacheLifetime = match (true) {
+            null !== $other->cacheLifetime => clone ($other->cacheLifetime),
+            null !== $this->cacheLifetime => clone ($this->cacheLifetime),
```
alexander-schranz, line 57: @Prokyonn beside the other metadata the cachelifetime is a single self containing xml tag and so there is can not be a `merge` method as there is nothing to merge.

So I currently using `clone` here. Alternative would be that we create a `$this->cacheLifetime->copy()` method. What do you think for transparency?

Prokyonn: Do you have any other use cases in mind, where we would need to copy/clone the `cacheLifetime` ? 

if we only use it here, I would go with the current solution, but if there are more cases, we should introduce a `copy` method.

alexander-schranz: Don't think so. Form Template Metadata is only here merged. And so only here cachelifetime copied.

alexander-schranz:
@manuxi Thank you!

PR #8502, opened 2026-01-10
Fix mandatory check for select fields

manuxi:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no yes
| Deprecations? | no
| License | MIT

#### Why?

Whith the fix its working now:

image


alexander-schranz (commented):
Did a few adjustments.

 - Moved the metadata mapper to the content type as that is how we did it for 2.6
 - Removed Stubs, we only do mocks if required (and with prophecy) for all classes which can easily be created we create them ourselves
 - Simplify check of JSON Schema, its easier to debug a whole assertSame as just every small value of the array

Still thank you here for your work on this 👍

alexander-schranz:
@manuxi thx merged to `2.6@dev` and `3.0@dev`

PR #8503, opened 2026-01-12
Add mimetype and partial combined filter tests for MediaSmartContentProvider

NeuralClone:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | --
| Related issues/PRs | #8433, #8494
| License | MIT
| Documentation PR | --

#### What's in this PR?

Adds a `mimetype` filtering test and a test to complement the fix in #8494 for `MediaSmartContentProvider`.

#### Why?

To make sure #8433 doesn't reoccur in the 3.0 branch.

### DIFF
--- a/packages/content/tests/Functional/Traits/CreateMediaTrait.php
+++ b/packages/content/tests/Functional/Traits/CreateMediaTrait.php
@@ -76,7 +76,8 @@ trait CreateMediaTrait
      * @param array{
      *     title?: string,
      *     description?: string,
-     *     type ?: string,
+     *     type?: string,
+     *     mimetype?: string,
      *     locale?: string,
      * } $data
      */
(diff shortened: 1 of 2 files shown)


alexander-schranz:
@Prokyonn can you do the review of this as you most familiar with the test class here and maybe idea how it should be implemented.

Prokyonn (approved):

Prokyonn:
@NeuralClone Thank you for the pull request! Looks good from my side.

As far as I see the reason why our other tests did not trigger the bug was because we never had the combination of mimeType + includeSubFolders right?

https://github.com/sulu/sulu/pull/8503/changes#diff-2618a03c1dfde344d4656e4ae4dc1543a5a7273b0bf264940992f17cda9e071cR853-R854

NeuralClone:
@Prokyonn Yes, along with the dataSource filter also being needed, that is correct. The same combination of filters was also missing in the 2.6 test cases.

And you're welcome!

PR #8504, opened 2026-01-12
Fix list pagination with joined fields

### LINKED ISSUE #8467: Admin list view filters based on joins have wrong result limits per page
| Q | A
| --- | ---
| Sulu Version | 2.6
| PHP Version | 8.2
| DB Version | MariaDB
| Browser Version | Any

# Actual Behavior

Filters based on joins in admin list view not working correctly. Less results than expected are on a single page.  

# Expected Behavior

The list view should use the correct limit.

# Steps to Reproduce

Use Sulu Rocks demo. Add 5 more contacts. Add two tags to all contacts: tag1 and tag2. Than filter for tag1 and tag2 and now you have paginate over two pages. This is unnecessary. Sorry I dont when demo will be reset.

https://sulu.rocks/admin/#/contacts?filter.tagId%5B0%5D=1&filter.tagId%5B1%5D=2

No filter: no

With 2 tags in filter, pagination appears without need: filter

# Possible Solutions

I think the subquery should be set to disctinct, anywhere. Maybe in this function: https://github.com/sulu/sulu/blob/2.6/src/Sulu/Component/Rest/ListBuilder/Doctrine/DoctrineListBuilder.php#L357

Because of joins, ids are duplicated in the subquery result.

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #8467
| Related issues/PRs |
| License | MIT
| Documentation PR | N/A (internal bug fix)

#### What's in this PR?

This PR fixes a pagination bug in `DoctrineListBuilder` where filtering by joined fields (e.g., tags) returns incorrect result counts per page.

**Changes:**
- Added `hasJoins()` method to detect when JOINs are present in filter fields
- Extended `createSubQueryBuilder()` to accept an `applyDistinct` parameter
- Modified `findIdsByGivenCriteria()` to automatically apply DISTINCT when JOINs are detected
- Modified `count()` to use `COUNT(DISTINCT id)` when JOINs are present

The fix automatically applies DISTINCT to ID subqueries when:
1. User explicitly called `distinct(true)`, OR
2. JOINs are detected in the query

#### Why?

When filtering admin list views by joined fields (especially one-to-many relationships like tags), the ID subquery used for pagination returns duplicate IDs because each entity appears once per joined row.

**Problem scenario:**
- Request: 10 items per page (limit 10)
- Entity has 3 matching tags → appears 3 times in JOIN result
- Subquery returns: `[1, 1, 1, 2, 2, 3, 3, 3, 4, 5]` (10 rows with duplicates)
- LIMIT 10 is applied to this duplicated result
- After de-duplication: `[1, 2, 3, 4, 5]` (only 5 unique IDs)
- Result: Page shows 5 items instead of 10

The fix ensures DISTINCT is applied **before** LIMIT, returning the correct number of unique entities.

### DIFF
--- a/src/Sulu/Component/Rest/ListBuilder/Doctrine/DoctrineListBuilder.php
+++ b/src/Sulu/Component/Rest/ListBuilder/Doctrine/DoctrineListBuilder.php
@@ -196,7 +196,6 @@ class DoctrineListBuilder extends AbstractListBuilder
     {
         $applyDistinct = $this->distinct || $this->hasJoins();
 
-        // Use COUNT(DISTINCT ...) when distinct is needed to avoid counting duplicates from JOINs
         $countExpression = $applyDistinct
             ? 'COUNT(DISTINCT ' . $this->idField->getSelect() . ')'
             : 'COUNT(' . $this->idField->getSelect() . ')';
@@ -363,8 +362,6 @@ class DoctrineListBuilder extends AbstractListBuilder
      */
     protected function findIdsByGivenCriteria()
     {
-        // Apply DISTINCT if user explicitly called distinct(true) OR JOINs are present
-        // to prevent duplicate IDs when filtering by joined fields
         $applyDistinct = $this->distinct || $this->hasJoins();
 
         $subQueryBuilder = $this->createSubQueryBuilder(
--- a/src/Sulu/Component/Rest/ListBuilder/Doctrine/DoctrineListBuilder.php
+++ b/src/Sulu/Component/Rest/ListBuilder/Doctrine/DoctrineListBuilder.php
@@ -194,7 +194,14 @@ class DoctrineListBuilder extends AbstractListBuilder
 
     public function count()
     {
-        $subQueryBuilder = $this->createSubQueryBuilder('COUNT( ' . $this->idField->getSelect() . ')', false);
+        $applyDistinct = $this->distinct || $this->hasJoins();
+
+        // Use COUNT(DISTINCT ...) when distinct is needed to avoid counting duplicates from JOINs
+        $countExpression = $applyDistinct
+            ? 'COUNT(DISTINCT ' . $this->idField->getSelect() . ')'
+            : 'COUNT(' . $this->idField->getSelect() . ')';
+
+        $subQueryBuilder = $this->createSubQueryBuilder($countExpression, false);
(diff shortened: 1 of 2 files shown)


alexander-schranz:
Is a 2.6 issue or?

wachterjohannes:
@alexander-schranz yes - wrong base branch in github! but ignore it for now - its only a first Prototype! i have to refactor and test it properly

alexander-schranz (approved):

src/Sulu/Component/Rest/ListBuilder/Doctrine/DoctrineListBuilder.php
```diff
@@ -194,7 +194,14 @@ public function addPermissionCheckField(DoctrineFieldDescriptor $fieldDescriptor
 
     public function count()
     {
-        $subQueryBuilder = $this->createSubQueryBuilder('COUNT( ' . $this->idField->getSelect() . ')', false);
+        $applyDistinct = $this->distinct || $this->hasJoins();
+
+        // Use COUNT(DISTINCT ...) when distinct is needed to avoid counting duplicates from JOINs
```
alexander-schranz:
```suggestion
```

src/Sulu/Component/Rest/ListBuilder/Doctrine/DoctrineListBuilder.php
```diff
@@ -356,7 +363,15 @@ protected function createNonGroupQueryBuilder(QueryBuilder $queryBuilder)
      */
     protected function findIdsByGivenCriteria()
     {
-        $subQueryBuilder = $this->createSubQueryBuilder($this->getSelectAs($this->idField));
+        // Apply DISTINCT if user explicitly called distinct(true) OR JOINs are present
+        // to prevent duplicate IDs when filtering by joined fields
```
alexander-schranz:
```suggestion
```

alexander-schranz (changes requested):
unnecesarry comments just describe that

wachterjohannes:
@alexander-schranz removed them

alexander-schranz (approved):

PR #8516, opened 2026-01-14
Respect requested ids over flat parameter

MarkusHolstein:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | not really
| Deprecations? | no
| Fixed tickets |
| Related issues/PRs |
| License | MIT

#### What's in this PR?

Allows to display attached categories in `list_overlay`-view.

#### Why?

When requesting categories via list_overlay configuration, the requested looks like this:
`admin/api/categories?locale=de&ids=16846,16884&page=1&flat=true`.
Because the order of the request checks, the flat parameter took precedence over the requested ids and so the "flat"-tree was handled. But since, when requesting ids the result should be always flat anyway, I reordered the the if branches.

#### Steps to reproduce

- Configure categories in settings tab to be used as list_overlay
- Open the category list and attach some categories to the contentpage
- Save and Refresh the page.
- The categories are not shown anymore

### DIFF
--- a/src/Sulu/Bundle/CategoryBundle/Controller/CategoryController.php
+++ b/src/Sulu/Bundle/CategoryBundle/Controller/CategoryController.php
@@ -87,7 +87,12 @@ class CategoryController extends AbstractRestController implements ClassResource
             $parentId = null;
         }
 
-        if ('true' == $request->get('flat')) {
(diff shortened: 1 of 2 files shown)


MarkusHolstein:
Clsoed, because I could not switch target Branch

Issue #6529, opened 2022-04-07
category filter shows nothing after refresh


datenwerk-at:
| Q | A
| --- | ---
| Sulu Version | 2.4.2
| PHP Version | 7.4.21
| DB Version | MySQL 8.0.28

# Actual Behavior
If a category filter is applied for e.g. "Contacts" => "People" and the category is a child, then category is empty after reloading the page.

# Expected Behavior
After page reload the applied category filter is shown.

# Steps to Reproduce
- create category "Parent"
- create category "Child" which is a child of "Parent
- go to "Contacts" => "People" choose filter "Categories" and apply the filter for "Child"
- the list is filtered with "Child" and "Child" is shown as filter
- now reload the page, e.g. with CMD+R, the category name is gone

alexander-schranz:
Sorry for the late response.

I did have a look at the Categories API. And it seems currently it does not support to return a specific categories by list of `ids`. Currently the API seems only be optimized for category Tree View. Which from the given ids can will always return all parent categories and all parent siblings category.

And in the specific auto complete case that would be the API needed to handle completely different that it returns only the specific ones and not contain any parent and siblings like it is required for the Tree.

So to handle this issue I think there is a major refactoring required about how the Categories API should work in specific `Views`. The most expected I think would be:

 - `/admin/api/categories?ids=5,6&locale=en&flat=true` returns a flat list of the given ids
 - `/admin/api/categories?ids=5,6&locale=en&flat=true&includeParents=true&includeSiblings=true` returns the data required by the Tree List View, also this should be used what currently `expandedIds` is also used by user settings to keep the same categories open when go back to the tree few as the API should not care why the parents and siblings are required (selection, already expanded, ...) it should just flag what it should additional load.
 
There is currently historical chaos as we have currenty different ids parameter which I think all should be just `ids`:

 - `ids` filter currently only working for root categories
 - `selectedIds` loads the given ids including its parants, siblings, and parent siblings
 - `expandedIds` work same as selectedIds
 
So I think here some cleanup is required. I'm not sure yet how we can solve this issue without adding more chaos to this very old API but not breaking something currently working for this API.

draconivis:
this was fixed with #8517

PR #8517, opened 2026-01-14
Fix list_overlay in category_selection not working as expected (#8517)

MarkusHolstein:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | not really
| Deprecations? | no
| Fixed tickets |
| Related issues/PRs |
| License | MIT

#### What's in this PR?

Allows to display attached categories in `list_overlay`-view.

#### Why?

When requesting categories via list_overlay configuration, the requested looks like this:
`admin/api/categories?locale=de&ids=16846,16884&page=1&flat=true`.
Because the order of the request checks, the flat parameter took precedence over the requested ids and so the "flat"-tree was handled. But since, when requesting ids the result should be always flat anyway, I reordered the the if branches.

#### Steps to reproduce

- Configure categories in settings tab to be used as list_overlay
- Open the category list and attach some categories to the contentpage
- Save and Refresh the page.
- The categories are not shown anymore


src/Sulu/Bundle/CategoryBundle/Controller/CategoryController.php
```diff
@@ -87,7 +87,12 @@ public function cgetAction(Request $request)
             $parentId = null;
         }
 
-        if ('true' == $request->get('flat')) {
+        if ($request->query->has('ids')) {
+            $ids = \array_filter(\explode(',', $request->query->get('ids')));
+            $entities = $this->categoryManager->findByIds($ids);
+            $categories = $this->categoryManager->getApiObjects($entities, $locale);
```
alexander-schranz: This looks unexpected the list requests of a Sulu UI should never work with Category Entities / API Objects. As Entity and Api Objects are doing a lot of Overfetching which should not done for lists.

It should always work in a list with `flat` data and the listBuilder of it.

So we would need to know why the listBuilder does not work for you as expected in this case? Why it ignores the `ids` filter?

Or did I understand it correctly that it returns all categories instead of filter by specific ids?

alexander-schranz: PS: If we would have cleanup the controller by remove all unnecessary API parts. Which was first planned for 3.0 only the `flat=true` part would exist and the other elses and none flat responses would have been removed from list controllers.

MarkusHolstein: I did not change that behaviour, all I did was moving the `$request->query->has('ids')` above the `'true' == $request->get('flat')` if branch. The behaviour inside the branch is untouched.

alexander-schranz: Now `categoryManager` is called for even a `flat=true` exists. Which should not happening. So whatever the issue is it need to be handled as a flat request via the listBuilder. So it returns the categories in a preformant way not that it uses any of categoryManager based responses.

~~Can you provide a more detail what the real issue is the description is quiet technical but not really describing what the issue for the enduser appears.~~

alexander-schranz: Is the error maybe here: 

```
$expandedIds = \array_filter(\explode(',', $request->get('expandedIds', $request->get('selectedIds', ''))));
```

the `selectedIds` sounds like it should be `ids` or is `selectedIds` somewhere used?

MarkusHolstein: Yes, I just thought about that and will test.
The selectedIds are used, when opening the overlay, but are always empty.

MarkusHolstein: Ok, when commenting out the ids-branch and juming into the flat=true-branch, and replacing:
`$expandedIds = \array_filter(\explode(',', $request->get('expandedIds', $request->get('selectedIds', ''))));`
With
`$expandedIds = \array_filter(\explode(',', $request->get('expandedIds', $request->get('selectedIds', $request->query->get('ids', '')))));`
(and removing a bug with a not set parentId in line 286)
The result looks as expected.

So should I implement it this way? Is this the right order: expandedIds, selectedIds, ids, or should I reorder them somehow?

alexander-schranz: Can you commit it. I need do some deeper testing how it effects other cases. Little bit confused why we didn't do `?flat=true&ids=1,2,3&expanded=true` instead of expandedIds param.

MarkusHolstein: I committed it. For now without tests

alexander-schranz:
@Prokyonn this issue may also related to: https://github.com/sulu/sulu/issues/6529

Prokyonn:
@MarkusHolstein I tested this Branch locally and found a regression in the Category listview. Unfortunately I cannot push to your branch, but can you please commit the following diff and also verify that the given issue from this PR is still solved.

```diff
diff --git a/src/Sulu/Bundle/CategoryBundle/Controller/CategoryController.php b/src/Sulu/Bundle/CategoryBundle/Controller/CategoryController.php
index 2a6fb1ceab..2697da6fac 100644
--- a/src/Sulu/Bundle/CategoryBundle/Controller/CategoryController.php
+++ b/src/Sulu/Bundle/CategoryBundle/Controller/CategoryController.php
@@ -193,11 +193,11 @@ class CategoryController extends AbstractRestController implements SecuredContro

     protected function getListRepresentation(
         Request $request,
-                $locale,
-                $parentId = null,
-                $expandedIds = [],
-                $expandSelf = false,
-                $includeRoot = false,
+        $locale,
+        $parentId = null,
+        $expandedIds = [],
+        $expandSelf = false,
+        $includeRoot = false,
         bool $defaultSort = true
     ) {
         $listBuilder = $this->initializeListBuilder($locale, $defaultSort);
@@ -256,7 +256,11 @@ class CategoryController extends AbstractRestController implements SecuredContro
             }
         }

-        if (!empty($expandedIds) && !$search && !empty($parentId)) {
+        $usesIdsFallback = $request->query->has('ids')
+            && !$request->query->has('expandedIds')
+            && !$request->query->has('selectedIds');
+
+        if (!empty($expandedIds) && !$search && !$usesIdsFallback) {
             $categoriesByParentId = [];
             foreach ($categories as &$category) {
                 $categoryParentId = $category['parent'];
@@ -276,7 +280,7 @@ class CategoryController extends AbstractRestController implements SecuredContro
                 ];
             }

-            $categories = $categoriesByParentId[$parentId];
+            $categories = $categoriesByParentId[$parentId] ?? [];
         }

         if ($includeRoot && !$parentId) {
```

Prokyonn:
> @Prokyonn this issue may also related to: #6529

yes, #6529 is also fixed with this PR and the given diff from https://github.com/sulu/sulu/pull/8517#issuecomment-4170069102

src/Sulu/Bundle/CategoryBundle/Controller/CategoryController.php
```diff
@@ -36,15 +35,10 @@
 /**
  * Makes categories available through a REST API.
  */
-class CategoryController extends AbstractRestController implements ClassResourceInterface, SecuredControllerInterface
+class CategoryController extends AbstractRestController implements SecuredControllerInterface
```
Prokyonn: revert this, not part of this PR scope

src/Sulu/Bundle/CategoryBundle/Controller/CategoryController.php
```diff
@@ -36,15 +35,10 @@
...
 {
     use RequestParametersTrait;
 
-    /**
-     * @deprecated Use the CategoryInterface::RESOURCE_KEY constant instead
-     */
-    protected static $entityKey = 'categories';
```
Prokyonn, line 46: revert this, not part of this PR scope

Prokyonn (changes requested):

Prokyonn (approved):

Prokyonn:
@MarkusHolstein Thank you!

PR #8522, opened 2026-01-14
Add form metadata to container as resources to trigger recompile

### LINKED ISSUE #8488: Hot Reload for Template changes
# Problem description

Add Hot Reload for template changes under development. For better performance we not longer access the filesystem for metadata https://github.com/sulu/sulu/pull/8457/changes, under development we should still do that and check if we need reload the metadata for better developer experience.

Importantunder `stage`, `prod` ENV nothing more as following should happen: https://github.com/sulu/sulu/blob/ca98d4777c522454101c8394659ddd062e3eaf02/src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/XmlFormMetadataLoader.php#L63-L67 no additional cache or file lookups. Only under debug / dev additonal hot reload should be added.

# Proposed solution

Maybe add a ConfigCache but only when debug is true.

Not 100% sure maybe have a compilerpass with `$container->addResource` can also make sure that symfony will rebuild the container or warmup?

mamazu:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #8488 
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?
Adding form metadata xml files to the container so that it re-compiles on form metadata changes.

#### Why?
After various optimizations the hotreload broke.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/DependencyInjection/Compiler/FormMetadataCachePass.php
+++ b/src/Sulu/Bundle/AdminBundle/DependencyInjection/Compiler/FormMetadataCachePass.php
@@ -17,10 +17,15 @@ use Symfony\Component\Config\Resource\DirectoryResource;
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 
+/**
+ * @internal no backwards compatibility promise, only for internal use.
+ */
 class FormMetadataCachePass implements CompilerPassInterface
 {
     public function process(ContainerBuilder $container): void
     {
+        $kernelProjectDir = $container->getParameter('kernel.project_dir');
+
         foreach ($container->getParameter('sulu_admin.forms.directories') as $directory) {
             $this->addDirectory($directory, $container);
         }
@@ -41,9 +46,27 @@ class FormMetadataCachePass implements CompilerPassInterface
     private function addDirectory(string $directory, ContainerBuilder $container): void
     {
         // Resolving container parameters
-        $directory = \preg_replace_callback('#%([^%]+)%#', fn ($match) => $container->getParameter($match[1]), $directory);
-        if (\file_exists($directory)) {
-            $container->addResource(new DirectoryResource($directory, '/\.xml$/'));
+        $directory = $container->resolveEnvPlaceholders(
+            \preg_replace_callback(
+                '#%([^%]+)%#',
+                static function(array $match) use ($container): string {
+                    /** @var string $param */
+                    $param = $container->getParameter($match[1]);
+
+                    return $param;
+                },
+                $directory
+            )
+        );
+
+        if (!\is_string($directory)) {
+            return;
         }
+
+        if (!\file_exists($directory) || !\is_dir($directory)) {
+            return;
+        }
+
+        $container->addResource(new DirectoryResource($directory, '/\.xml$/'));
     }
 }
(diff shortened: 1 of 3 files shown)


alexander-schranz:
Should we exclude directories which are part of `vendor`? What is best practice here?

src/Sulu/Bundle/AdminBundle/SuluAdminBundle.php
```diff
@@ -30,5 +31,7 @@ public function build(ContainerBuilder $container): void
         if ($container->hasExtension('fos_js_routing')) {
             $container->addCompilerPass(new ExposeResourceRoutesPass());
         }
+
+        $container->addCompilerPass(new FormMetadataCachePass());
```
alexander-schranz: Should we add a lower priortiy here if somebody uses compiler passes to extend the dirs?

Also not sure why the CI failing. As I thought the values are always defined: 

https://github.com/sulu/sulu/blob/c2c1de660c91144921b679ca5c60bb6531bfb64f/src/Sulu/Bundle/AdminBundle/DependencyInjection/SuluAdminExtension.php#L170-L172
    
  

. The AdminBundle may not always registered but then the compiler pass also should not be called 🤔

src/Sulu/Bundle/AdminBundle/DependencyInjection/Compiler/FormMetadataCachePass.php
```diff
@@ -22,11 +22,28 @@ class FormMetadataCachePass implements CompilerPassInterface
...
+                $this->addDirectory($directory, $container);
+            }
         }
-        foreach ($container->getParameter('sulu_admin.list.directories') as $directory) {
+    }
+
+    private function addDirectory(string $directory, ContainerBuilder $container): void
+    {
+        // Resolving container parameters
+        $directory = \preg_replace_callback('#%([^%]+)%#', fn ($match) => $container->getParameter($match[1]), $directory);
```
alexander-schranz: what is resolved here? Can you show an example?

src/Sulu/Bundle/AdminBundle/DependencyInjection/Compiler/FormMetadataCachePass.php
```diff
@@ -22,11 +22,28 @@ class FormMetadataCachePass implements CompilerPassInterface
...
+        if (\file_exists($directory)) {
```
alexander-schranz: should we do `\is_dir` here?

src/Sulu/Bundle/AdminBundle/DependencyInjection/Compiler/FormMetadataCachePass.php
alexander-schranz: if env vars we maybe should go with:

```
$directory = $container->resolveEnvPlaceholders($directory, true);
```

alexander-schranz (approved):

alexander-schranz:
I decided to add resources only if we are isDebug (dev,test) to not add any file system checks on stage or prod.

The vendor check is not required as Symfony handles that already here:

https://github.com/symfony/symfony/blob/dcd11486c0c12ea84866e1b7dae53ec68297a456/src/Symfony/Component/DependencyInjection/ContainerBuilder.php#L273

alexander-schranz:
@mamazu thank you!

PR #8530, opened 2026-01-15
Disable automapping and manually specify mapping

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | depends on https://github.com/sulu/SuluContentBundle/pull/276
| License | MIT
| Documentation PR | -

#### What's in this PR?
Adding manual entity mapping to all Sulu bundles.

#### Why?
We don't want to rely on auto_mapping as future versions of doctrine will conflict with this.

### DIFF
--- a/config/packages/doctrine.yaml
+++ b/config/packages/doctrine.yaml
@@ -27,7 +27,6 @@ when@prod: &prod
     doctrine:
         orm:
             auto_generate_proxy_classes: false
-            auto_mapping: true
             query_cache_driver:
                 type: pool
                 pool: doctrine.system_cache_pool
--- a/config/packages/doctrine.yaml
+++ b/config/packages/doctrine.yaml
@@ -8,7 +8,7 @@ doctrine:
     orm:
         auto_generate_proxy_classes: true
         naming_strategy: doctrine.orm.naming_strategy.underscore_number_aware
-        auto_mapping: true
+        auto_mapping: false
 #        mappings:
 #            App:
 #                is_bundle: false
--- a/config/packages/doctrine.yaml
+++ b/config/packages/doctrine.yaml
@@ -27,6 +27,7 @@ when@prod: &prod
     doctrine:
         orm:
             auto_generate_proxy_classes: false
+            auto_mapping: true
             query_cache_driver:
                 type: pool
                 pool: doctrine.system_cache_pool
--- a/phpstan.dist.neon
+++ b/phpstan.dist.neon
@@ -17,6 +17,8 @@ parameters:
     parallel:
         processTimeout: 300.0
     reportUnmatchedIgnoredErrors: false
+    ignoreErrors:
+        - message: '#^Property .*\\Entity\\.*\:\:\$id is never written, only read\.$#'
     excludePaths:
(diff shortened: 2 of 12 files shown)


config/packages/doctrine.yaml
```diff
@@ -8,7 +8,7 @@ doctrine:
     orm:
         auto_generate_proxy_classes: true
         naming_strategy: doctrine.orm.naming_strategy.underscore_number_aware
-        auto_mapping: true
+        auto_mapping: false
```
mamazu, line 11: This will disable automapping for all downstream projects as well right? I've explicitly enabled auto_mapping in prod again for a nicer fallback but, I'm not sure we want this. Otherwise I don't know how to test this.

alexander-schranz: no `sulu/sulu` is only for the CI and us to develop Sulu. Only `sulu/skeleton` changes would have effects on projects. And only if they would change it manually during update as Sulu would not overwrite it in a composer update.

config/packages/doctrine.yaml
```diff
@@ -27,6 +27,7 @@ when@prod: &prod
     doctrine:
         orm:
             auto_generate_proxy_classes: false
+            auto_mapping: true
```
alexander-schranz: Would remove this, aslong all works as expected with false all is fine.


```suggestion
```

config/packages/doctrine.yaml
alexander-schranz: Only thing which effects third party bundles is the test bundle config here: https://github.com/sulu/sulu/blob/9e6f05a2c150b560f46fab722d670b3140c56c17/src/Sulu/Bundle/TestBundle/Resources/app/config/sulu.yml#L16

This we are not allowed to change.

mamazu:
And done

alexander-schranz (approved):

mamazu:
@alexander-schranz Baseline and type issues fixed.

phpstan.dist.neon
```diff
@@ -17,6 +17,8 @@ parameters:
     parallel:
         processTimeout: 300.0
     reportUnmatchedIgnoredErrors: false
+    ignoreErrors:
+        - message: '#^Property .*\\Entity\\.*\:\:\$id is never written, only read\.$#'
```
alexander-schranz: no global ignores, should just be baselined

```suggestion
```

alexander-schranz:
@mamazu can you check third party core bundles like Form, Redirect, Community Bundle, Automation (including PHP Task Bundle) which also currently use auto mapping. Target there also 2.x branch.

mamazu:
- [ ] FormBundle https://github.com/sulu/SuluFormBundle/pull/419
- [ ] RedirectBundle https://github.com/sulu/SuluRedirectBundle/pull/104
- [ ] CommmunityBundle https://github.com/sulu/SuluCommunityBundle/pull/186
- [ ] AutomationBundle https://github.com/sulu/SuluAutomationBundle/pull/88
- [ ] PHP Task Bundle github.com/php-task/TaskBundle/pull/59

PR #8555, opened 2026-01-19
Migrate TrashBundle to service php definition

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | 
| Related issues/PRs | #8216 
| License | MIT
| Documentation PR | -

#### What's in this PR?
Migrating the xml config to php

#### Why?
Symfony 8 support


alexander-schranz (approved):

composer-dependency-analyser.php
```diff
@@ -14,6 +14,8 @@
 
 $config = new Configuration();
 
+require __DIR__ . '/vendor/symfony/dependency-injection/Loader/Configurator/ContainerConfigurator.php'; // see https://github.com/shipmonk-rnd/composer-dependency-analyser/issues/147#issuecomment-2202156380
+
```
alexander-schranz, line 18: I pushed this, as it is needed workaround for PHP service definitions we already use this on the 3.0 version: https://github.com/sulu/sulu/blob/3.0.5/composer-dependency-analyser.php#L15

mamazu: This only happens with functions if you use the naked `TaggedIteratorArgument` class just like we use `new Reference` instead of the `service` function then there are no problems. I don't know. If we want to use Symfony helper functions or not.

PR #8563, opened 2026-01-21
Fix to take loadExcerpt and shadowLocale into account for caching on snippet resolving

MarkusHolstein:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | 
| Related issues/PRs |
| License | MIT
| Documentation PR |

#### What's in this PR?

Allows to resolve Snippets with extension and without extensions on subsequent calls to ->resolve

#### Why?

When loading a Snippet via the Twig-function sulu_snippet_load_by_area, the snippet does not contain the extensions. 
If you load the same Snippet via a custom twig function on a later call, and you want to resolve the same Snippet, the resovler always returns the cached version, which does not contain the extension, even if the $loadExcerpt parameter is set.

#### Example Usage

$snippet1 = $snippetresolver->resolve('123-123-123-123', 'demo', 'de');
$snippet2 = $snippetresolver->resolve('123-123-123-123', 'demo', 'de', null, true);

$snippet2 does not contain the extensions.

This also is true for the other way round.

### DIFF
--- a/src/Sulu/Bundle/SnippetBundle/Snippet/SnippetResolver.php
+++ b/src/Sulu/Bundle/SnippetBundle/Snippet/SnippetResolver.php
@@ -53,7 +53,7 @@ class SnippetResolver implements SnippetResolverInterface, ResetInterface
     {
         $snippets = [];
         foreach ($uuids as $uuid) {
-            $cacheKey = \sprintf('%s|%s', $locale, $uuid);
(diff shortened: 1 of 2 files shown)


alexander-schranz (approved):

alexander-schranz:
@MarkusHolstein thank you!

PR #8568, opened 2026-01-23
Fixing phpstan/php-cs-fixer after dependency update

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?
Fixing phpstan after dependency update

#### Why?
Pipeline should be green and it is only used with strings so we might as well restrict it.

### DIFF
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -47,7 +47,11 @@ $config->setRiskyAllowed(true)
         'fully_qualified_strict_types' => false,
         'new_with_parentheses' => true,
         'trailing_comma_in_multiline' => ['after_heredoc' => true, 'elements' => ['array_destructuring', 'arrays', 'match']],
+        'no_useless_else' => false,
+        // Disabling this because php-cs-fixer doesn't know that it implicitly returns null
+        'no_useless_return' => false
     ])
     ->setFinder($finder);
 
 return $config;
+
--- a/src/Sulu/Bundle/MediaBundle/Media/Manager/MediaManager.php
+++ b/src/Sulu/Bundle/MediaBundle/Media/Manager/MediaManager.php
@@ -390,7 +390,7 @@ class MediaManager implements MediaManagerInterface
      */
     private function buildData($uploadedFile, $data, $user)
     {
-        if (!($uploadedFile instanceof UploadedFile)) {
+        if (!$uploadedFile instanceof UploadedFile) {
             throw new InvalidFileException('Given uploaded file is not of instance UploadedFile');
         }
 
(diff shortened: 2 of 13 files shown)


Prokyonn (approved):

Prokyonn:
@mamazu do we also need this for the 2.6 branch?

mamazu:
Maybe I don't know if the 2.6 pipeline is also broken. But I can rebase it to 2.6

Prokyonn:
> Maybe I don't know if the 2.6 pipeline is also broken. But I can rebase it to 2.6

@mamazu yes can you please rebase to 2.6 and there are now also php-cs-fixer adjustments required, can you maybe add them also 😇

mamazu:
Running php-cs-fixer will however cause conflicts when upmerging. :D

src/Sulu/Bundle/ContactBundle/Contact/AbstractContactManager.php
```diff
@@ -1265,71 +1257,70 @@ protected function createAddress($addressData, &$isMain = null)
             throw new EntityIdAlreadySetException(self::$addressEntityName, $addressData['id']);
         } elseif (!$addressType) {
             throw new EntityNotFoundException(self::$addressTypeEntityName, $addressData['addressType']);
-        } else {
-            $address = new Address();
```
Prokyonn, line 1269: @mamazu I think we should disable the "useless-else" rule, as this will make debugging more difficult and also change the "blame" user. For 3.0 we can maybe enable it

src/Sulu/Bundle/ContactBundle/Api/Account.php
```diff
@@ -202,8 +202,6 @@ public function getParent()
         if ($account) {
             return new self($account, $this->locale);
         }
-
-        return;
```
alexander-schranz, line 206: looks unexpected removal should be more a `return null;`

src/Sulu/Bundle/ContactBundle/Contact/AbstractContactManager.php
```diff
@@ -752,13 +752,12 @@ protected function addEmail($contact, $emailData)
             throw new EntityIdAlreadySetException(self::$emailEntityName, $emailData['id']);
         } elseif (!$emailType) {
             throw new EntityNotFoundException(self::$emailTypeEntityName, $emailData['emailType']);
-        } else {
```
alexander-schranz, line 755: see current no advantages doing this. Would disable currently this PHP CS Rule.

alexander-schranz (changes requested):
For better Git Blame I would remove the useless-else rule currently.

The removal of `return;` looks false and that lines should be probably be changed to `return null;`

Examples:

src/Sulu/Bundle/ContactBundle/Contact/AbstractContactManager.php
mamazu: Agreed. Espcially not when trying to fix phpstan errors. :D

alexander-schranz (approved):

Prokyonn (approved):

Prokyonn:
@mamazu thank you!

Issue #8569, opened 2026-01-25
Add template priority/ordering support


enzobrigati:
# Problem description

Templates are displayed in alphabetical order in the admin template selector. There is no way to customize this order.  

When building a site with multiple templates (homepage, portal, collection, article, default...), it would be helpful to display the most commonly used templates at the top of the selector.                                                                                                           
                                                                                                                                                                                                                                                                                                          
The only workaround is to prefix template filenames with numbers (01_homepage.xml, 02_portal.xml), but this also affects the template key which is not ideal.

# Proposed solution                                                                                                                                                                                                                                                                                   

Add a <priority> element in the template's `<meta>` section:
                                                                                                                                                                                                                                                                                                    
```xml
<meta>                                                                                                                                                                                                                                                                                                  
      <title lang="en">Homepage</title>                                                                                                                                                                                                                                                                   
      <priority>100</priority>                                                                                                                                                                                                                                                                            
  </meta>
```


Higher priority values would appear first in the admin template selector. Templates without priority would default to 0.

PR #8572, opened 2026-01-26
Add priority ordering support for templates

enzobrigati:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets |
| Related issues/PRs | #8569
| License | MIT

#### What's in this PR?

This PR adds a `<priority>` element to the template's `<meta>` section, allowing developers to customize the display order of templates in the admin template selector.

**Changes include:**
- XSD schema update to allow `<priority>` in template meta
- `FormMetadata` class: added `priority` property with getter/setter
- `MetaXmlParser`: parsing of `<priority>` element
- `TemplateXmlLoader`: transfer priority value to FormMetadata
- Frontend: `metadataStore.js` includes priority in type transformation
- Frontend: `TypeToolbarAction.js` sorts by priority first, then by title
- Unit tests for PHP and JS

#### Why?

Templates are currently displayed in alphabetical order in the admin template selector. There is no way to customize this order.

When building a site with multiple templates (homepage, portal, collection, article, default...), it is helpful to display the most commonly used templates at the top of the selector.

The only workaround was to prefix template filenames with numbers (01_homepage.xml, 02_portal.xml), but this also affects the template key which is not ideal.

#### Example Usage

```xml

    <key>homepage</key>

    <meta>
        <title lang="en">Homepage</title>
        <title lang="de">Startseite</title>
        <priority>100</priority>
    </meta>

    <!-- ... -->
</template>
```

- Higher priority values appear first in the admin template selector
- Templates without `<priority>` default to 0
- Templates with the same priority are sorted alphabetically by title

#### To Do

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -1803,7 +1803,7 @@ parameters:
 		-
 			message: '#^Parameter \#3 \$context of method Sulu\\Bundle\\AdminBundle\\Metadata\\FormMetadata\\Parser\\MetaXmlParser\:\:getValueFromXPath\(\) expects DOMNode\|null, DOMNameSpaceNode\|DOMNode given\.$#'
 			identifier: argument.type
-			count: 1
+			count: 2
 			path: src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/Parser/MetaXmlParser.php
 
 		-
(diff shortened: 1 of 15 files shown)


alexander-schranz:
@enzobrigati Thank you for you pull request. 

There are currently some discussions going on in the core team how we want to handle this. A priority is maybe some very personal view on such objects and for different devs and different contenet manager there are different priorities. Also content manager and developer change over the time and the system stay.

So we are currently in discussions if we want to sort this kind of things always Alphabetically, as that something which make searching for the human most easiest.

I will keep you uptodate and let the pull request open until there is a decision made.

alexander-schranz:
PS: @enzobrigati if you require a quick solution for your project. Templates can have tags and you could use tags `<tag name="app.order" order="1"/>` and a custom [`TypedFormMetadataVisitorInterface`](https://github.com/sulu/sulu/blob/3.0/src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/TypedFormMetadataVisitorInterface.php) to change the order of forms.

enzobrigati:
Hi @alexander-schranz , thanks for the feedback and the workaround suggestion! :)

I understand the concern about priority being "personal", but I'd like to clarify the use case:                                                                                                                                                                                                           

In some projects priority is a project architecture decision, not a user preference. The developer knows which templates are most commonly used in their specific project and sets the order accordingly.                                                                                                                                                                                                                                                  

The proposed solution is opt-in and non-breaking:
- Templates without <priority> default to 0 and remain sorted alphabetically
- Only projects that need custom ordering use it

That said, I completely understand if the team decides to keep things simple with alphabetical sorting. I'll use the tag workaround in the meantime.

  Thanks for keeping the PR open!

PR #8576, opened 2026-01-27
No useless return (the fixed version)

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | refs #8568 
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?
Removing useless returns and updating the useful ones.

#### Why?
Implicit php behaviour is never a good thing, and running php-cs shouldn't make phpstan unhappy.

Updating the documentation to also include null in those cases where it is used.

### DIFF
--- a/src/Sulu/Bundle/ContactBundle/Api/Account.php
+++ b/src/Sulu/Bundle/ContactBundle/Api/Account.php
@@ -191,7 +191,7 @@ class Account extends ApiWrapper
     /**
      * Get parent.
      *
-     * @return AccountInterface
+     * @return self|null
      */
     #[VirtualProperty]
     #[SerializedName('parent')]
@@ -203,7 +203,7 @@ class Account extends ApiWrapper
             return new self($account, $this->locale);
         }
 
-        return;
+        return null;
     }
 
     /**
@@ -951,8 +951,6 @@ class Account extends ApiWrapper
                 }
             }
         }
-
-        return;
     }
 
     /**
@@ -1003,8 +1001,6 @@ class Account extends ApiWrapper
                 'thumbnails' => $this->logo->getFormats(),
             ];
         }
-
-        return;
     }
 
     /**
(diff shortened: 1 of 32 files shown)


src/Sulu/Bundle/ContactBundle/Api/Account.php
```diff
@@ -951,8 +951,6 @@ public function getMainAddress()
                 }
             }
         }
-
-        return;
     }
```
alexander-schranz, line 956: should also be `return null;` here as not return would be void. Which may technical end in null but type specific correct is do explict return for such cases.

src/Sulu/Bundle/ContactBundle/Api/Account.php
```diff
@@ -1003,8 +1006,6 @@ public function getLogo()
                 'thumbnails' => $this->logo->getFormats(),
             ];
         }
-
-        return;
```
alexander-schranz, line 1007: still some missingthink basically every return; here should be return null; :)

removing the return or return; would only make sense for `: void` return types.

src/Sulu/Bundle/MediaBundle/Api/Collection.php
```diff
@@ -283,7 +283,7 @@ public function getPreview()
     }
 
     /**
-     * @param array $preview
+     * @param array|null $preview
```
alexander-schranz: looks unexpected for me would not allow this currently, even there are some cases called maybe with null.

src/Sulu/Bundle/MediaBundle/Collection/Manager/CollectionManager.php
```diff
@@ -687,7 +687,7 @@ protected function getPreview($id, $locale)
             }
         }
 
-        return;
+        return null;
```
alexander-schranz: would `return [];`

src/Sulu/Bundle/ContactBundle/Api/Account.php
mamazu: Does the code still exist, I can't find it.

alexander-schranz (approved):

PR #8577, opened 2026-01-27
Fix linting issues after dependency updates

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| License | MIT

#### What's in this PR?

Fixes linting issues from dependency updates (phpstan / php-cs-fixer)

#### Why?

Pipeline should be green :D


mamazu (approved):
Nice, converting the protected properties into private for final classes 👍

alexander-schranz (approved):

PR #8578, opened 2026-01-27
Fix upgrade 3.0 queries

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | yes

#### What's in this PR?

Fixes errors in the Upgrade file.

### DIFF
--- a/UPGRADE-3.x.md
+++ b/UPGRADE-3.x.md
@@ -618,10 +618,10 @@ DROP TABLE se_user_groups;
 
 ### Migrate User entity columns
 
-The `apiKey` column type was changed to `guid` and the `idContacts` join column is now nullable:
+The `idContacts` join column is now nullable to be consistent with the entity property.
 
 ```sql
-ALTER TABLE se_users CHANGE apiKey apiKey CHAR(36) DEFAULT NULL, CHANGE idContacts idContacts INT DEFAULT NULL;
+ALTER TABLE se_users CHANGE idContacts idContacts INT DEFAULT NULL;
 ```
 
 ### Migrate on TimestampableInterface depend entities to DateTimeImmutable
--- a/UPGRADE-3.x.md
+++ b/UPGRADE-3.x.md
@@ -540,7 +540,7 @@ DROP TABLE me_file_version_content_languages;
 DROP TABLE me_file_version_publish_languages;
 ```
 
-Also the media types have been moved to the `me_media` table itself:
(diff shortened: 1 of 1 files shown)


UPGRADE-3.x.md
```diff
@@ -1033,28 +1053,6 @@ ALTER TABLE se_permissions DROP module;
...
--- Creating a new index
-CREATE INDEX IDX_A694E5728CDE5729 ON me_media (type);
-
--- Removing the old data
-DROP TABLE me_media_types;
-DROP INDEX IDX_A694E57284671716 ON me_media;
-ALTER TABLE me_media DROP idMediaTypes;
-```
-
-If you have queries to the `MediaType` entity completely remove that.
```
alexander-schranz, line 1056: What happened to this?

Prokyonn: This is already part of the "MediaBundle" section

Prokyonn: https://github.com/sulu/sulu/pull/8578/changes#diff-c698cbb1f2afce655c27e32ebfc8d9f37af04b8164a1d8d609da71d2858c6697R546-R558

UPGRADE-3.x.md
```diff
@@ -607,6 +616,14 @@ DROP TABLE se_groups;
 DROP TABLE se_user_groups;
 ```
 
+### Migrate User entity columns
+
+The `apiKey` column type was changed to `guid` and the `idContacts` join column is now nullable:
+
+```sql
+ALTER TABLE se_users CHANGE apiKey apiKey CHAR(36) DEFAULT NULL, CHANGE idContacts idContacts INT DEFAULT NULL;
```
Prokyonn: I'm not really sure about the apiKey (guid), has this been changed in a dbal version to CHAR(36)? I don't remember that we modified this in sulu 3.0

alexander-schranz: @Prokyonn we may should replace `guid` with `string` even on 2.6?

UPGRADE-3.x.md
alexander-schranz: to make sure if somebody search for missing class `MediaType` they find it in the upgrade file:

```suggestion
Also the `MediaType` entity has been removed the value has been moved to the `me_media` table itself as a string column:
```

UPGRADE-3.x.md
Prokyonn: Done in https://github.com/sulu/sulu/pull/8579

wachterjohannes (approved):

alexander-schranz (approved):

PR #8581, opened 2026-01-28
Allow to load excerpt via sulu_snippet_load_by_area twig function

MarkusHolstein:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| License | MIT
| Documentation PR | sulu/sulu-docs#883

#### What's in this PR?

This PR allows to load a default snippet with "sulu_snippet_load_by_area" optionally with its excerpt information.

#### Why?

When loading a default snippet via the new sulu_snippet_load_by_area method, it is currently not possible to load the extension data. To get those data, one would currently need to load the default snippet first, and then issue another load via sulu_snippet_load, because the loadExcerpt parameter is present in this method. 
So for consistency and easier use, I added this parameter.

#### Example Usage

```
{% set snippet = sulu_snippet_load_by_area('footer_social_media_links', null, null, true); %}
```

#### To Do

### DIFF
--- a/src/Sulu/Bundle/SnippetBundle/Tests/Unit/Twig/SnippetAreaTwigExtensionTest.php
+++ b/src/Sulu/Bundle/SnippetBundle/Tests/Unit/Twig/SnippetAreaTwigExtensionTest.php
@@ -247,7 +247,7 @@ class SnippetAreaTwigExtensionTest extends TestCase
             $this->snippetAreaReferenceStore
         );
 
-        $this->assertEquals(
+        $this->assertSame(
             ['title' => 'Test Snippet EN'],
             $twigExtension->loadByArea('test', 'demo_io', null, true)
         );
--- a/src/Sulu/Bundle/SnippetBundle/Tests/Unit/Twig/SnippetAreaTwigExtensionTest.php
+++ b/src/Sulu/Bundle/SnippetBundle/Tests/Unit/Twig/SnippetAreaTwigExtensionTest.php
@@ -222,7 +222,7 @@ class SnippetAreaTwigExtensionTest extends TestCase
         );
     }
 
-    public function testLoadByAreWithLoadExcerpt(): void
+    public function testLoadByAreaWithLoadExcerpt(): void
     {
         $this->defaultSnippetManager = $this->prophesize(DefaultSnippetManagerInterface::class);
         $this->requestAnalyzer = $this->prophesize(RequestAnalyzerInterface::class);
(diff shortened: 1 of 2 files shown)


src/Sulu/Bundle/SnippetBundle/Tests/Unit/Twig/SnippetAreaTwigExtensionTest.php
```diff
@@ -221,4 +221,35 @@ public function testLoadByAreaOtherWebspace(): void
             $twigExtension->loadByArea('test', 'demo_io', 'en')
         );
     }
+
+    public function testLoadByAreWithLoadExcerpt(): void
```
Prokyonn:
```suggestion
    public function testLoadByAreaWithLoadExcerpt(): void
```

MarkusHolstein: Fixed

Prokyonn (approved):

alexander-schranz (approved):

src/Sulu/Bundle/SnippetBundle/Twig/SnippetAreaTwigExtension.php
```diff
@@ -46,10 +46,11 @@ public function getFunctions()
      * @param string $area
      * @param string $webspaceKey
      * @param string $locale
+     * @param bool $loadExcerpt
      *
      * @return array
      */
-    public function loadByArea($area, $webspaceKey = null, $locale = null)
+    public function loadByArea($area, $webspaceKey = null, $locale = null, $loadExcerpt = false)
     {
```
alexander-schranz: None final classes can not add new params without add a bc layer.

```suggestion
    public function loadByArea($area, $webspaceKey = null, $locale = null/*, $loadExcerpt = false */)
{
         $loadExcerpt = func_num_args() > 3 ? func_get_args()[3] : false;
```

alexander-schranz (changes requested):
Class is not final so we need add a bc layer to add such feature.

src/Sulu/Bundle/SnippetBundle/Tests/Unit/Twig/SnippetAreaTwigExtensionTest.php
```diff
@@ -221,4 +221,35 @@ public function testLoadByAreaOtherWebspace(): void
...
+        $this->snippetResolver->resolve(['1234'], 'demo_io', 'de', null, true)->shouldBeCalled()->willReturn([['title' => 'Test Snippet EN']]);
+
+        $twigExtension = new SnippetAreaTwigExtension(
+            $this->defaultSnippetManager->reveal(),
+            $this->requestAnalyzer->reveal(),
+            $this->snippetResolver->reveal(),
+            $this->snippetAreaReferenceStore
+        );
+
+        $this->assertEquals(
```
alexander-schranz:
```suggestion
        $this->assertSame(
```

alexander-schranz (approved):

alexander-schranz:
@MarkusHolstein Thank you!

PR #8583, opened 2026-01-29
Allow to remove types from FieldMetaData

MarkusHolstein:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes
| Related issues/PRs | 
| License | MIT

#### What's in this PR?

This PR allows to remove a FormMetaData type item from a FieldMetaData item.

#### Why?

As discussed, we are modifying the global blocks list depending on role permissions, so we can hide custom modules from some users. But since there were no setters, we need to use reflection to update the containing types inside of a FieldMetaData item. With this PR, we could remove the reflection part from our code.

#### Example Usage

I added unit tests to test the new method.

```php
[...]
    public function testRemoveTypeRemovesFromTypes(): void
    {
        $formMetadata1 = $this->setupFormMetadata('dummy1');
        $formMetadata2 = $this->setupFormMetadata('dummy2');

        $metaData = new FieldMetadata('dummy');
        $metaData->addType($formMetadata1);
        $metaData->addType($formMetadata2);

        $metaData->removeType('dummy1');

        $this->assertSame([
            'dummy2' => $formMetadata2,
        ], $metaData->getTypes());
[...]
```

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/FieldMetadata.php
+++ b/src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/FieldMetadata.php
@@ -98,9 +98,6 @@ class FieldMetadata extends ItemMetadata
 
     public function removeType(string $type): void
     {
-        if (!isset($this->types[$type])) {
-            return;
-        }
         unset($this->types[$type]);
     }
 
--- a/src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/FieldMetadata.php
+++ b/src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/FieldMetadata.php
@@ -96,6 +96,14 @@ class FieldMetadata extends ItemMetadata
         $this->types[$type->getName()] = $type;
     }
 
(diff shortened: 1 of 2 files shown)


src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/FieldMetadata.php
```diff
@@ -96,6 +96,14 @@ public function addType(FormMetadata $type): void
         $this->types[$type->getName()] = $type;
     }
 
+    public function removeType(string $type): void
+    {
+        if (!isset($this->types[$type])) {
+            return;
+        }
+        unset($this->types[$type]);
```
alexander-schranz: unset not requires a isset check before:


```suggestion
        unset($this->types[$type]);
```

alexander-schranz (changes requested):
Thx for the pull request think we can simplify the removeType part.

alexander-schranz:
@MarkusHolstein Thank you!

PR #8586, opened 2026-01-29
Fix apply query and anchor to internal page links also

ymc-thzi:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | - 
| Related issues/PRs | - 
| License | MIT
| Documentation PR | - 

#### What's in this PR?

This PR fixes a bug where query parameters and anchors configured in custom page links are not appended to internal page link URLs.

**Changes:**
- Extracted query/anchor appending logic into a new `appendQueryAndAnchor()` method
- Applied this method to both `resolvePage()` (internal page links) and `resolveLink()` (external links)
- This ensures consistent behavior across all link types

#### Why?

When configuring a custom link on a page with query parameters and/or anchors:
- ✅ External links correctly get query/anchor appended  
- ❌ Internal page links ignore query/anchor settings

This creates inconsistent behavior and makes it impossible to use query parameters or anchors with internal page links (e.g., in navigation menus).

The root cause is that the `PageLinkDimensionContentEnhancer` class had the query/anchor appending logic in `resolveLink()` (for external links) but not in `resolvePage()` (for internal links).

#### Example Usage

**Before this fix:**
```php
// Page with custom internal link to another page
// Link settings: query = "category=2", anchor = "test"
// Result: /target-page (query and anchor are lost ❌)
```

**After this fix:**
```php
// Same configuration as above
// Result: /target-page?category=2#test (query and anchor preserved ✅)
```

#### To Do

- [x] Bug fix does not require documentation changes
- [x] No breaking changes to add to UPGRADE.md

### DIFF
--- a/packages/page/src/Infrastructure/Sulu/Content/ContentResolver/PageLinkDimensionContentEnhancer.php
+++ b/packages/page/src/Infrastructure/Sulu/Content/ContentResolver/PageLinkDimensionContentEnhancer.php
@@ -103,7 +103,7 @@ class PageLinkDimensionContentEnhancer implements DimensionContentEnhancerInterf
         $targetDimensionContent = $this->contentAggregator->aggregate($page, $dimensionAttributes);
 
         $url = $targetDimensionContent->getTemplateData()['url'] ?? null;
-        if (\is_string($url)) {
+        if (\is_string($url) && null !== $linkData) {
             $url = $this->appendQueryAndAnchor($url, $linkData);
         }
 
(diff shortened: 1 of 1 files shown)


alexander-schranz (approved):
Fine for me currently. @Prokyonn We at some time should check how we can maybe make that link resolving not be splitted up to much as the query and anchor part seems be in [different places](https://github.com/sulu/sulu/blob/07b6542cce8dd05a958c884ccd9c3db6c929579a/packages/content/src/Application/PropertyResolver/Resolver/LinkPropertyResolver.php#L47-L53) done again.

Prokyonn (approved):

PR #8591, opened 2026-01-31
Fix Preview not including unlocalised values (#8574)

### LINKED ISSUE #8574: Non-multilingual fields lose their value in template preview
| Q | A
| --- | ---
| Sulu Version | 3.0.3
| PHP Version | 8.3
| DB Version |  mariadb:10.11
| Browser Version | Any

# Actual Behavior

Fields defined with `multilingual="false"` are not displayed correctly in the template preview.
After changing the value of any field and reloading the template preview, the value of the non-multilingual field becomes `null`, even though it was previously set.

# Expected Behavior

Fields with `multilingual="false"` should correctly display their values in the template preview, even after refreshing the preview.

# Steps to Reproduce

1. Create a template with a field defined as `multilingual="false"`.
2. Dump this field in a Twig template.
3. Assign a value to this field in the Sulu admin.
4. Verify that the field has the correct value in the template preview after the initial preview load.
5. Change the value of any field and wait for the preview to reload.
6. Observe that the non-multilingual field value becomes `null`.

# Possible Solutions

The issue seems to be caused by the order of operations in `TemplateDataMapper`.

Current order:
```php
$unlocalizedDimensionContent->setTemplateData($unlocalizedData);
$localizedDimensionContent->setTemplateKey($template);
$localizedDimensionContent->setTemplateData($localizedData);
```

In experimental SuluContentBundle, the order was:
```php
$localizedDimensionContent->setTemplateKey($template);
$localizedDimensionContent->setTemplateData($localizedData);
$unlocalizedDimensionContent->setTemplateData($unlocalizedData);
```

After restoring the previous order (from SuluContentBundle), the value is displayed correctly in the template preview.

jsiefer:
PreviewDimensionContentCollection must return two contents, localized and unlocalized that then later get merged back together.

| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no 
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #8574
| Related issues/PRs | 
| License | MIT
| Documentation PR | 

#### What's in this PR?

The _PreviewDimensionContentCollection_ returns the same content instance for localized and unlocalized, which results in overwriting each other inside the _ContentDataMapper_.
I think _PreviewDimensionContentCollection_ must return separate instances that then later can be merged back together. 

#### Why?

Currently non-localized fields are not available in the preview. 

#### Caveats

The _ContentObjectProvider_ requires a new dependency for the _ContentMerger_, _SuluPageBundle_ and _SuluArticleBundle_ seem to initialize them, not sure if we can get around it. Perhaps this requires more documentation.

#### To Do

----

I am not really familiar with the code base, please let me know if there is anything else I can provide or should look up.

### DIFF
--- a/packages/content/src/Infrastructure/Sulu/Preview/PreviewDimensionContentCollection.php
+++ b/packages/content/src/Infrastructure/Sulu/Preview/PreviewDimensionContentCollection.php
@@ -13,7 +13,6 @@ declare(strict_types=1);
 
 namespace Sulu\Content\Infrastructure\Sulu\Preview;
 
-use Doctrine\Common\Collections\ArrayCollection;
 use Sulu\Content\Domain\Model\DimensionContentCollectionInterface;
 use Sulu\Content\Domain\Model\DimensionContentInterface;
 
@@ -23,10 +22,18 @@ use Sulu\Content\Domain\Model\DimensionContentInterface;
  * @template-covariant T of DimensionContentInterface
  *
  * @implements DimensionContentCollectionInterface<T>
- * @implements \IteratorAggregate<T>
  */
-class PreviewDimensionContentCollection implements \IteratorAggregate, DimensionContentCollectionInterface
+class PreviewDimensionContentCollection implements DimensionContentCollectionInterface
 {
+    /**
+     * @var T
+     */
+    private readonly DimensionContentInterface $unlocalizedDimensionContent;
+    /**
+     * @var T
+     */
+    private readonly DimensionContentInterface $localizedDimensionContent;
+
     /**
      * @param T $previewDimensionContent
      */
@@ -34,6 +41,8 @@ class PreviewDimensionContentCollection implements \IteratorAggregate, Dimension
         private DimensionContentInterface $previewDimensionContent,
         private string $previewLocale
     ) {
+        $this->localizedDimensionContent = clone $previewDimensionContent;
+        $this->unlocalizedDimensionContent = clone $previewDimensionContent;
     }
 
     public function getDimensionContentClass(): string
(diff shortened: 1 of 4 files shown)


packages/content/src/Infrastructure/Sulu/Preview/PreviewDimensionContentCollection.php
```diff
@@ -23,17 +22,27 @@
...
+
     /**
      * @param T $previewDimensionContent
      */
     public function __construct(
         private DimensionContentInterface $previewDimensionContent,
         private string $previewLocale
     ) {
+        $this->localizedDimensionContent = clone $previewDimensionContent;
+        $this->unlocalizedDimensionContent = clone $previewDimensionContent;
```
alexander-schranz, line 45: unsure if we should clone this object as we are not in control which dimension content object the people create 🤔 @Prokyonn think we need find another solution for this problem.

draconivis:
@jsiefer Thank you! we decided to go with the `ContentAggregator` instead of the `ContentMerger` to prevent BC breaks and replace the `PreviewDimensionContentCollection` but your changes were the right direction.

alexander-schranz (approved):

packages/content/tests/Functional/Infrastructure/Sulu/Preview/ContentObjectProviderTest.php
```diff
@@ -0,0 +1,88 @@
...
+
+    protected function setUp(): void
+    {
+        /** @var ContentObjectProvider<ExampleDimensionContent, Example> $contentObjectProvider */
+        $contentObjectProvider = self::getContainer()->get('example_test.example_preview_object_provider');
+        $this->contentObjectProvider = $contentObjectProvider;
+    }
+
+    /**
+     * Regression test for #8574: the preview must map the localized and unlocalized dimension
```
Prokyonn: trim comment, should not really be necessary as the test name should already explain that

Prokyonn (approved):

PR #8593, opened 2026-02-02
Fix phpstan issues after dependency upadte

Prokyonn:
| Q | A
  | --- | ---
  | Bug fix? | no
  | New feature? | no
  | BC breaks? | no
  | License | MIT

  #### What's in this PR?

  - Added PHPDoc return type annotations to methods with implicit nullable return types (`Contact::getId()`, `CustomUrlController::getLocale()`, `CustomUrlRouteController::getLocale()`)
  - Fixed type coercion in `ContactManager` by casting `$contactId` to `int` when constructing `ContactRemovedEvent`
  - Updated functional tests to explicitly cast `Contact::getId()` returns to `int` when used in string interpolation

  #### Why?

  PHPStan's stricter analysis after dependency updates flagged implicit type issues. Methods returning `int|null` without explicit return type annotations cause type inference problems, and passing potentially
  nullable values where integers are expected triggers errors. These changes satisfy PHPStan's type checking requirements without changing runtime behavior.

### DIFF
--- a/src/Sulu/Bundle/ContactBundle/Contact/ContactManager.php
+++ b/src/Sulu/Bundle/ContactBundle/Contact/ContactManager.php
@@ -193,7 +193,7 @@ class ContactManager extends AbstractContactManager implements DataProviderRepos
             $this->em->remove($contact);
 
             $this->domainEventCollector->collect(
-                new ContactRemovedEvent($contactId, $contactFullName)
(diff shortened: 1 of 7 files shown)


Prokyonn:
already done in https://github.com/sulu/sulu/pull/8592

PR #8596, opened 2026-02-03
Fix the ORM 3 integration for extended Entities with relations

### LINKED ISSUE #8483: Extending Account does not work anymore since 3.0.1
| Q | A
| --- | ---
| Sulu Version | 3.0.1
| PHP Version | 8.3.3
| DB Version | mysql

# Actual Behavior

Using my bundle which enritches the account data does not work anymore since 3.0.1.
With 3.0.0 my bundle worked as intended. 
With 3.0.1 no accounts are showing up and I'm getting in the Webconsole
[Semantical Error] line 0, col 620 near 'Sulu_Bundle_ContactBundle_Entity_AccountAddress': Error: Class Manuxi\SuluAdditionalAccountDataBundle\Entity\Account has no association named accountAddresses

Tried to remove the Attributes and using an orm.xml without luck.

# Expected Behavior

Accounts show up and have desired properties.

# Steps to Reproduce

Extending the Account as I did:
[...]
use Sulu\Bundle\ContactBundle\Entity\Account as SuluAccount;

#[ORM\Entity]
#[ORM\Table(name: 'co_accounts')]
class Account extends SuluAccount
{
    #[ORM\Column(type: Types::STRING, length: 255, nullable: true)]
    private ?string $regCourt;
[...]

(See https://github.com/manuxi/SuluAdditionalAccountDataBundle/tree/3.x)

# Possible Solutions

Don't know.

mamazu:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no 
| Deprecations? | no
| Fixed tickets | fixes #8483
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?
Fix for the copying of association mapping.

#### Why?
In doctrine 3 the metadata are now an object so checking if the $value variable is an array will return false. We should also check if it's an AssociationMapping class. 

### BC
I don't know if the `AssociationMapping` class exists in doctrine 2 but since it should be lazily evaluated, for doctrine 2 the first part should be true and it should not come to the second part.

### DIFF
--- a/src/Sulu/Component/Persistence/EventSubscriber/ORM/MetadataSubscriber.php
+++ b/src/Sulu/Component/Persistence/EventSubscriber/ORM/MetadataSubscriber.php
@@ -100,13 +100,16 @@ class MetadataSubscriber
 
             // map relations
             foreach ($parentMetadata->getAssociationMappings() as $key => $value) {
-                // can be changed to $value->type and $value->sourceEntity if min version is doctrine/orm 3+
+                // The following code is littered with phpstan-ignore-next-line because of doctrine/orm 2/3 compat
+                // doctrine/orm 3: changed to $value->type and $value->sourceEntity (AssociationMapping object)
+                // doctrine/orm 2: changed to $value['type'] and $value['sourceEntity'] (array/ArrayAccess object)
+
                 // @phpstan-ignore-next-line argument.type
                 if ($this->hasRelation($value['type'])) {
                     $value['sourceEntity'] = $metadata->getName();
-                    // can be removed if min version is doctrine/orm 3+
-                    // @phpstan-ignore-next-line function.impossibleType
+                    // @phpstan-ignore-next-line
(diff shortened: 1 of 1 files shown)


alexander-schranz:
I think we can ignore the phpstan-error. But I think that:

> can be removed if min version is doctrine/orm 3+

Is false and misleading as it seems be requried for ORM 3 still. So think we should remove the comment or change it that the `is_array` check can be removed then?

mamazu:
Well that depends, if we want to support both Doctrine 2 and 3 then we need both. In Doctrine 2 it's an array in and in 3 it's an object.

Alternatively we could just remove the entire if check as this should always be that.

alexander-schranz (approved):

alexander-schranz:
Awesome. @mamazu thank you!

manuxi:
Great work! Thank you, @mamazu!

PR #8597, opened 2026-02-03
Use system cache for Icons

### LINKED ISSUE #8585: Icon cache pool should be system cache pool not a app cache pool
| Q | A
| --- | ---
| Sulu Version | 2.6, 3.0
| PHP Version | 
| DB Version | 
| Browser Version | 

# Actual Behavior

Currently the icon cache pool of the admin is cached in the `var/share` but it don't require to be a shared cache pool as it is once generated and can be a system cache which is faster and also `var/share` caches are not cleared via `cache:clear` but think system caches are?

https://github.com/sulu/sulu/blob/07b6542cce8dd05a958c884ccd9c3db6c929579a/src/Sulu/Bundle/AdminBundle/DependencyInjection/SuluAdminExtension.php#L60

# Expected Behavior

Move to system cache.

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #8585 
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?
Using the system cache.

#### Why?
We don't need to clear the icon cache every time the cache:clear command is called.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/DependencyInjection/SuluAdminExtension.php
+++ b/src/Sulu/Bundle/AdminBundle/DependencyInjection/SuluAdminExtension.php
@@ -58,7 +58,7 @@ class SuluAdminExtension extends Extension implements PrependExtensionInterface
                                 'adapter' => 'cache.app',
                             ],
                             'sulu_admin.icon_cache' => [
-                                'adapter' => 'cache.app',
+                                'adapter' => 'cache.system',
                             ],
                         ],
                     ],


alexander-schranz:
Can you rebase this to 2.6?

mamazu:
Yes.

alexander-schranz (approved):

alexander-schranz:
@mamazu thank you!

PR #8598, opened 2026-02-03
Replace all Request->get calls in csv handler with Request->query->get for Symfony 8 compatibility

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | yes
| Fixed tickets | -
| Related issues/PRs | #8216 
| License | MIT
| Documentation PR | -

#### What's in this PR?
* Replacing `$request->get` with `$request->query->getString` in the CSV handler.
* Updating the `convertValue` function to have stricter types now (no BC break, cause private)
* Refactoring the tests
    * Extracting the instanciation of the CsvHandler into the setup method
    * Moving the header checks into it's own function
    * Replacing mocking the Request with an actual request.

#### Why?
It's deprecated in Symfony 8

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -7941,25 +7941,13 @@ parameters:
 		-
 			message: '#^Binary operation "\." between ''/api/categories/'' and mixed results in an error\.$#'
 			identifier: binaryOp.invalid
-			count: 27
-			path: src/Sulu/Bundle/CategoryBundle/Tests/Functional/Controller/CategoryControllerTest.php
-
-		-
-			message: '#^Binary operation "\." between ''/api/categories…'' and mixed results in an error\.$#'
-			identifier: binaryOp.invalid
-			count: 16
+			count: 1
 			path: src/Sulu/Bundle/CategoryBundle/Tests/Functional/Controller/CategoryControllerTest.php
 
 		-
 			message: '#^Binary operation "\." between mixed and ''\.'' results in an error\.$#'
 			identifier: binaryOp.invalid
 			count: 1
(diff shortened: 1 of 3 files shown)


src/Sulu/Component/Rest/RequestParametersTrait.php
```diff
@@ -17,6 +17,8 @@
 
 /**
  * handles request parameters.
+ *
+ * @deprecated
```
mamazu: @alexander-schranz  This will no longer work with Symfony 8 anways. I don't know what the policy is here.

mamazu:
Or should this be rebased down to 2.6?

Prokyonn:
Looks good to me 

> Or should this be rebased down to 2.6?

I think it would make sense @alexander-schranz WDYT?

src/Sulu/Component/Rest/RequestParametersTrait.php
alexander-schranz: we should target this in an Own PR with Upgrade Guide.

```suggestion
```

I think best way would be a !method_exists($request, 'get') and throw a BadMethodCallException  https://www.php.net/manual/en/class.badmethodcallexception.php

alexander-schranz:
> Looks good to me
> 
> > Or should this be rebased down to 2.6?
> 
> I think it would make sense @alexander-schranz WDYT?

Fine for me. @mamazu do you want to rebase it and target the 2.6 branch?

mamazu:
Rebased.

alexander-schranz (approved):

mamazu:
Looks like the type-save functions on the input bag don't exist in Symfony 5.4 yet. I've added a BC layer, we should remove this in 3.0 though.

PR #8599, opened 2026-02-03
Fix AccessControlQueryEnhancer for multiple roles

### LINKED ISSUE #8499: No Access to Collections if one ANY users role has insufficient permissions
| Q | A
| --- | ---
| Sulu Version | 2.6
| PHP Version | 8.3

# Actual Behavior

I have enabled permissions on collections. As an admin user, I have full access to all collections. When a user has only "Add" rights on a collection, they cannot see it, which is expected behavior.

However, if an admin user is additionally assigned the editor role, a permissions conflict occurs: the user unexpectedly loses the ability to view the collection. It seems that the more restrictive rights from the editor role override the admin rights, which should not happen.

# Expected Behavior

When adding the editors role to an admin user, the editor permissions should add up to the admin users admin permissions

# Steps to Reproduce

- As a user with only role Admin:
- Create a media collection
- Add "View"-Permissions for admin role
- Add "Add"-Permissions for editor role
- The collection is visible
- Add role editor to the user
- The media collection is hidden

# Possible Solutions

There seems to be a problem with the AccessControlQueryEnhancer.
If I "extract" the plain SQL query it looks somehow like this:

```
SELECT collection.id AS id_0 FROM 
	me_collections collection 
	INNER JOIN se_access_controls acl ON (acl.entityClass = ? AND acl.entityIdInteger = collection.id) 
	INNER JOIN se_roles roles ON acl.idRoles = roles.id 
	WHERE 
		((acl.permissions & ?) <> ? AND acl.permissions IS NOT NULL) 
		AND roles.id IN (?)
```

The resulting ids are used as blacklist for the original query.
But if you have two roles with one role not matching the bitmask (like 64 for view requested, but having a role having only 32 as permission) this one entry is found and results in this id beeing blacklisted.

So the solution would be to wrap that query in another SELECT, so the result would look a bit like this:

```
SELECT collection.id
FROM me_collections collection
WHERE NOT EXISTS (
    SELECT 1
    FROM se_access_controls acl
    INNER JOIN se_roles roles ON acl.idRoles = roles.id
    WHERE acl.entityClass = ?
      AND acl.entityIdInteger = collection.id
      AND roles.id IN (?)
      AND (acl.permissions & ?) = ?
)
```
But since this changes the current behaviour, I am unsure what to do.

# Additional

This would not have been a big issue by itself, as even the explanation sounds weird, but there seems to be another bug in the DoctrineAccessControlProvider, when removing permissions from a collection, the permission value is saved as 0-permission, which led us to loose permissions unexpectedly. But I have to track this down and will open another issue for this.

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | no yes
| New feature? | no yes
| BC breaks? | no yes
| Deprecations? | no yes 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?
Changes permission logic from blacklist (NOT IN) to whitelist (EXISTS) approach to correctly handle users with multiple roles.

Problem:
Users with multiple roles (e.g., Admin + Editor) would lose access to entities if ANY role lacked the required permission, even when another role granted full access. This occurred because the old logic excluded entities where any role failed the permission check.

Solution:
- Replaced NOT IN blacklist with EXISTS whitelist subquery
- Changed BIT_AND check from <> to = for correct permission matching
- Implements OR logic: user has access if ANY role grants permission
- Added handling for users with no roles (deny access)

Fixes #8499 

#### To Do

### DIFF
--- a/UPGRADE-2.x.md
+++ b/UPGRADE-2.x.md
@@ -15,6 +15,12 @@ The type of the `apiKey` in the `se_users` table has been changed to `string`.
 ALTER TABLE se_users CHANGE apiKey apiKey VARCHAR(128) DEFAULT NULL;
 ```
 
+An index on the `idRoles` column has been added to the `se_access_controls` table to improve query performance for permission checks with multiple roles.
+
+```sql
+CREATE INDEX IDX_C526DC5238C751C4 ON se_access_controls (idRoles);
+```
+
 ## 2.6.16
 
 * Deprecated \Sulu\Bundle\MediaBundle\Controller\AbstractMediaController::getTitleFromUpload -> MediaManager::getTitleFromUpload
--- a/UPGRADE-2.x.md
+++ b/UPGRADE-2.x.md
@@ -17,6 +17,8 @@ ALTER TABLE se_users CHANGE apiKey apiKey VARCHAR(128) DEFAULT NULL;
 
 An index on the `idRoles` column has been added to the `se_access_controls` table to improve query performance for permission checks with multiple roles.
 
+The access control query logic for doctrine entities with multiple roles has been fixed. Access is now granted if any assigned role grants the requested permission for an entity, instead of incorrectly denying access when another assigned role has no permission for the same entity.
+
 ```sql
 CREATE INDEX IDX_C526DC5238C751C4 ON se_access_controls (idRoles);
 ```
--- a/src/Sulu/Bundle/SecurityBundle/AccessControl/AccessControlQueryEnhancer.php
+++ b/src/Sulu/Bundle/SecurityBundle/AccessControl/AccessControlQueryEnhancer.php
@@ -69,12 +69,11 @@ class AccessControlQueryEnhancer implements AccessControlQueryEnhancerInterface
     }
 
     /**
-     * Following function uses an own query to load the restricted (not accessible ids). This is faster as embedding the
-     * query as subquery. Also loading the "accessible" ids would be more performance intense because sulu has more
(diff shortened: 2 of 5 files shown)


alexander-schranz (changes requested):
Do not mock query builder, that tests are to hard to keep maintained. Better create a functional test and use the profiler datacollector to get the query and match against it.

alexander-schranz:
What about the indexes you said? Also was this changes testet against the previous project which introduced the old logic? Are there some benchmarks before and after?

MarkusHolstein:
Not to be annoying, but are there any news here? ;-/

wachterjohannes:
@MarkusHolstein i will put this on my prio list this week

UPGRADE-2.x.md
```diff
@@ -15,6 +8,12 @@ The type of the `apiKey` in the `se_users` table has been changed to `string`.
 ALTER TABLE se_users CHANGE apiKey apiKey VARCHAR(128) DEFAULT NULL;
 ```
 
+An index on the `idRoles` column has been added to the `se_access_controls` table to improve query performance for permission checks with multiple roles.
+
+```sql
+CREATE INDEX IDX_C526DC5238C751C4 ON se_access_controls (idRoles);
+```
+
```
Prokyonn, line 25: Add a comment that the logic changed about validating if a user has access or not depending on his roles

Prokyonn (approved):

PR #8600, opened 2026-02-04
Fix block field defaults issue

### LINKED ISSUE #8512: Adding a block omits default values for selects
| Q | A
| --- | ---
| Sulu Version | 3.0.2
| PHP Version | 8.3.3
| DB Version | MySQL

# Actual Behavior

When adding a block none of the selects use the defined default values. Changing the block then uses them again.

https://github.com/user-attachments/assets/f0c71c39-35cc-4223-92a0-8f9d08f2289f

# Expected Behavior

When adding a block the default values are set for selets.

# Steps to Reproduce

Add a block with a select.

# Possible Solutions

n.a.

wachterjohannes:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #8512
| Related issues/PRs | none
| License | MIT
| Documentation PR | none

#### What's in this PR?

**Root Cause:**
MobX observable arrays (expandedBlocks, selectedBlocks, generatedBlockIds) were updated BEFORE async block-id generation, creating an inconsistent state that prevented field components from properly initializing their defaults via onChange in their constructors.

**Solution:**
1. Move MobX observable updates to AFTER async operations (using runInAction)
2. Make fillArrays async and generate block IDs before calling onChange
3. Remove getBlockDefaults workaround - field components now handle defaults via constructor logic (consistent with root form fields)

This ensures atomic state updates and allows field components to properly apply their default values when blocks are created.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/BlockCollection.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/BlockCollection/BlockCollection.js
@@ -1,6 +1,6 @@
 // @flow
 import React from 'react';
-import {action, observable, toJS, reaction, computed} from 'mobx';
+import {action, observable, toJS, reaction, computed, runInAction} from 'mobx';
 import {observer} from 'mobx-react';
 import classNames from 'classnames';
 import {arrayMove, translate, clipboard} from '../../utils';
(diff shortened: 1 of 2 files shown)


wachterjohannes:
@manuxi could you test this if it works in your case?

manuxi:
Did test your pr but unfortunately it didn't fix the bug described in #8512

wachterjohannes:
@manuxi i tried it again and for me it seems to work - have you rebuild the JS when checking out my branch?

manuxi:
Sure. Did it 2 times in different ways.

alexander-schranz:
Did test it and it did work for me the same way as in 2.6. On `2.6` already single_select fields are required to have a mandatory on them if a defautl value is choosen. See https://github.com/sulu/sulu/issues/8512#issuecomment-3755690250

If we want to change that behaviour we would require to change it on 2.6, as this is a 3.0 fix all is fine.

PR #8602, opened 2026-02-04
Improve update build command fallback to latest 2.6 files

alexander-schranz:
| Q | A
| --- | ---
| Bug fix? | yes 
| New feature? | no 
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

Improve update build command fallback to latest 2.6 files, currently it did use 2.x where the files not exist ending it to remove all files instead of updating them.

#### Why?

Improve update build command fallback to latest 2.6 files. Also currently the `2.x` would just remove all files instead of throw an error on not found file.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Command/UpdateBuildCommand.php
+++ b/src/Sulu/Bundle/AdminBundle/Command/UpdateBuildCommand.php
@@ -17,13 +17,14 @@ use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Style\SymfonyStyle;
 use Symfony\Component\Filesystem\Filesystem;
-use Symfony\Component\HttpClient\Exception\ClientException;
 use Symfony\Component\Process\Process;
 use Symfony\Contracts\HttpClient\HttpClientInterface;
 
 #[AsCommand(name: 'sulu:admin:update-build', description: 'Updates the administration application JavaScript build by downloading the official build from the sulu/skeleton repository or building the assets via npm.')]
 class UpdateBuildCommand extends Command
 {
+    private const DEFAULT_DEV_VERSION = '2.6';
+
     public const EXIT_CODE_ABORTED_MANUAL_BUILD = 1;
     public const EXIT_CODE_COULD_NOT_INSTALL_NPM_PACKAGES = 2;
     public const EXIT_CODE_COULD_NOT_BUILD_ADMIN_ASSETS = 3;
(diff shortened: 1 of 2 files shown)


src/Sulu/Bundle/AdminBundle/Command/UpdateBuildCommand.php
```diff
@@ -17,13 +17,14 @@
...
 use Symfony\Component\Console\Style\SymfonyStyle;
 use Symfony\Component\Filesystem\Filesystem;
-use Symfony\Component\HttpClient\Exception\ClientException;
 use Symfony\Component\Process\Process;
 use Symfony\Contracts\HttpClient\HttpClientInterface;
 
 #[AsCommand(name: 'sulu:admin:update-build', description: 'Updates the administration application JavaScript build by downloading the official build from the sulu/skeleton repository or building the assets via npm.')]
 class UpdateBuildCommand extends Command
 {
+    private const DEFAULT_DEV_VERSION = '2.6';
```
alexander-schranz, line 26: should be change to 3.0 on 3.0 branch. Not sure how we could make sure to keep this updated?

@Prokyonn

src/Sulu/Bundle/AdminBundle/Command/UpdateBuildCommand.php
```diff
@@ -257,13 +258,9 @@ private function getRemoteFile(string $remoteRepository, string $path): string
...
         $path = \str_replace(\DIRECTORY_SEPARATOR, '/', $path);
 
-        try {
-            $response = $this->httpClient->request('GET', $remoteRepository . $path);
+        $response = $this->httpClient->request('GET', $remoteRepository . $path);
 
-            return $response->getContent();
-        } catch (ClientException $e) {
-            return '';
-        }
```
alexander-schranz, line 266: not catch this exception the remote files should always exist else it would be kind of a bug. So 404 should thrown here really out and not being catched.

src/Sulu/Bundle/AdminBundle/Command/UpdateBuildCommand.php
Prokyonn: Can we somehow get it from composer? Or otherwise we could add a script that verifies this in the Ci

Prokyonn (approved):

src/Sulu/Bundle/AdminBundle/Command/UpdateBuildCommand.php
alexander-schranz: Think a CI task which verifies this against by target branch would be awesome.

PR #8607, opened 2026-02-06
Add dynamic view resolving with access to the loaded entity

### LINKED ISSUE #8478: Sulu 3.0 - Snippets do not have their "template" type in the view data
In Sulu 2.6, we were able to loop over snippets defined inside our sidebar snippet, and use their `template` value to construct a Twig template path - e.g.

```twig
{% for index, snippet in content.sidebar.snippets %}
    <div class="snippet-container">
        {{ include('includes/snippets/' ~ view.sidebar.snippets[index].template ~ '.html.twig', {
            data: snippet,
            view: view.sidebar.snippets[index],
        }, ignore_missing: true) }}
    </div>
{% endfor %}
```

However, now in Sulu 3.0 the `view.sidebar.snippets` property does not exist, and there is no way to get the snippet template type string as we previous did using `view.sidebar.snippets[index].template` etc.

Is there a new way to do this now? Or is it possible to add this to 3.0?
Thanks

Prokyonn:
| Q                  | A
| ---                | ---
| Bug fix?           | yes
| New feature?       | yes
| BC breaks?         | yes (twig template view shape — see `UPGRADE-3.x.md`)
| Deprecations?      | no
| Fixed tickets      | fixes #8478
| Related issues/PRs | —
| License            | MIT
| Documentation PR   | sulu/sulu-docs#

#### What's in this PR?

This change introduces a generic mechanism so that any property type or resource loader can bubble view-side metadata up to the parent's view. The benefit applies uniformly to all selection variants (single & multi for `snippet`, `page`, `article`) and to any property resolver that emits view data on a `ResolvableResource` (e.g. `link`, smart content, or custom resolvers).

##### 1. New `ResourceLoaderContentViewInterface`

ResourceLoaders can now contribute per-resource view and content metadata via `resolveContentView(mixed $source): ContentView`. `PageResourceLoader`, `ArticleResourceLoader`, `SnippetResourceLoader`, `ExampleResourceLoader` (test bundle) and `CachedResourceLoader` (delegates) implement it.

The returned ContentView's `view` payload (e.g. `uuid`, `template`, `webspaceKey`, `parent`) is merged into the parent view at the position where the resource was referenced. The `content` payload (e.g. `authored`, `lastModified`) is merged into the parent content. Custom resource loaders can opt in by implementing the interface.

##### 2. `viewEnhancements` flow in `ResolvableResourceReplacer`

The replacer now walks the content tree tracking the path as `list<int|string>` and emits a `viewEnhancements` map alongside the replaced content. `ContentResolver` then merges those enhancements back into the view via `PropertyAccessor`. This works for any `ResolvableResource` regardless of which property resolver produced it:

- single-resource hits flat-merge into `view.<field>` (e.g. a `link` field's `{provider, href, title}` payload now reaches the view as expected)
- multi-selection items become a flat numeric list at `view.<field>` (matches the Sulu 2.6 shape, eliminating the old `ids`/`types`-mixed-into-numeric-array bug)
- multi-selection resolvers that need to coexist with resolver-level view metadata can opt into wrapping by passing a non-empty `view: [...]` and `itemsPropertyName: 'items'` (or any key); items then live under that key alongside the resolver-level data

`ResolvableResource` gains `$itemsPropertyName` to opt into the wrapped behavior.

##### 3. Selection resolvers simplified

`SnippetSelectionPropertyResolver`, `PageSelectionPropertyResolver`, `ArticleSelectionPropertyResolver` (and their single variants) no longer build view arrays manually. They pass `view: []` and let the ResourceLoader contribute the per-resource view metadata.

##### 4. Internal path representation unified

`ResolvableResourceReplacer`, `ContentResolver`, and `ContentViewDataNormalizer` all use `list<int|string>` for content-tree paths internally, converting to PropertyAccessor's bracket-string only at the API boundary. This replaces several fragile string operations (`substr_replace`, `strpos`, `substr_count('][')`, regex stripping) with plain array primitives.

#### Why?

Previously, property resolvers that emit a `ResolvableResource` had no way to surface the resolved value's view-side metadata back into the parent's view. Each resolver had to construct view arrays manually (e.g. `view: ['ids' => $ids, ...$params]`) without access to the loaded entity, and view payloads on nested resolvables (link metadata, sub-blocks, …) had nowhere to land.

That manifested as several concrete reported issues:

- **Single selections lost their inner view.** For example, `single_snippet_selection`'s view was stuck at `{id, types}` — templates reading `view.<snippetField>.<nestedKey>` (e.g. link metadata, sub-blocks) saw nothing. The same applied to `single_page_selection` and `single_article_selection`.
- **Multi-selections mixed metadata with items.** `snippet_selection`, `page_selection`, and `article_selection` returned views like `{ids, types, ...params, 0, 1, …}`, so `{% for x in view.<field> %}` yielded unexpected entries.
- **Nested link metadata was unreachable.** A `link` property nested inside a snippet/page/article had its title and provider flattened away on the data side and never bubbled up on the view side, despite `LinkPropertyResolver` already producing the right view payload.

The new architecture gives any property resolver and any resource loader a uniform way to contribute to the resolved view, makes single/multi variants behave symmetrically, and restores the flat-list shape that Sulu 2.6 emitted for multi selections so existing template logic ports cleanly.

#### BC breaks

PHP API: none. All concrete services touched are `@internal` (or `@final` in the case of `ContentResolver`); the `ContentResolverInterface` public contract is unchanged.

Twig template view shape: yes — multi-selection view changes from `{ids, types, 0, 1, ...}` to a flat numeric list `[v0, v1, ...]`; single-selection view exposes additional resolved-entity metadata. Migration paths are documented in `UPGRADE-3.x.md`.

#### Example Usage

Implement `ResourceLoaderContentViewInterface` on a custom loader to contribute view/content metadata:

```php
use Sulu\Content\Application\ContentResolver\Value\ContentView;
use Sulu\Content\Application\ResourceLoader\Loader\ResourceLoaderContentViewInterface;

class CustomResourceLoader implements ResourceLoaderContentViewInterface
{
    public function resolveContentView(mixed $source): ContentView
    {
        return ContentView::create(
            ['authored' => $source->getAuthored()?->format('c')],
            ['uuid' => $source->getUuid(), 'template' => $source->getTemplateKey()],
        );
    }

    // ... load() and getKey() as before
}
```

For multi-selection-style resolvers, items default to a flat numeric list. If you also need to expose resolver-level metadata alongside the items, pass a non-empty `view` and an `itemsPropertyName`:

```php
ContentView::createResolvablesWithReferences(
    ids: $ids,
    resourceLoaderKey: 'page',
    resourceKey: 'pages',
    view: ['presentAs' => 'grid'],   // resolver-level metadata
    itemsPropertyName: 'items',      // items will live under view.<field>.items
);
```

#### To Do

- Create a documentation PR

### DIFF
--- a/packages/article/src/Infrastructure/Sulu/Content/ResourceLoader/ArticleResourceLoader.php
+++ b/packages/article/src/Infrastructure/Sulu/Content/ResourceLoader/ArticleResourceLoader.php
@@ -13,8 +13,11 @@ declare(strict_types=1);
 
 namespace Sulu\Article\Infrastructure\Sulu\Content\ResourceLoader;
 
+use Sulu\Article\Domain\Model\ArticleDimensionContentInterface;
 use Sulu\Article\Domain\Repository\ArticleRepositoryInterface;
-use Sulu\Content\Application\ResourceLoader\Loader\ResourceLoaderInterface;
+use Sulu\Content\Application\ContentResolver\Value\ContentView;
+use Sulu\Content\Application\ResourceLoader\Loader\ResourceLoaderContentViewInterface;
+use Sulu\Content\Domain\Model\AuthorInterface;
 use Sulu\Content\Domain\Model\DimensionContentInterface;
 
 /**
@@ -22,7 +25,7 @@ use Sulu\Content\Domain\Model\DimensionContentInterface;
  *
  * @final
  */
-class ArticleResourceLoader implements ResourceLoaderInterface
+class ArticleResourceLoader implements ResourceLoaderContentViewInterface
 {
     public const RESOURCE_LOADER_KEY = 'article';
 
@@ -56,6 +59,30 @@ class ArticleResourceLoader implements ResourceLoaderInterface
         }
 
         return $mappedResult;
+    }
+
+    public function resolveContentView(mixed $source): ContentView
+    {
+        $view = [];
+        $content = [];
+
+        if ($source instanceof ArticleDimensionContentInterface) {
+            $view = [
+                'uuid' => $source->getResource()->getUuid(),
+                'template' => $source->getTemplateKey(),
+                'mainWebspace' => $source->getMainWebspace(),
+                'additionalWebspaces' => $source->getAdditionalWebspaces(),
+            ];
+        }
+
+        if ($source instanceof AuthorInterface) {
+            $content = [
+                'authored' => $source->getAuthored()?->format('c'),
+                'lastModified' => $source->getLastModified()?->format('c'),
+            ];
+        }
+
+        return ContentView::create($content, $view);
     }
 
     public static function getKey(): string
--- a/packages/article/src/Infrastructure/Sulu/Content/ResourceLoader/ArticleResourceLoader.php
+++ b/packages/article/src/Infrastructure/Sulu/Content/ResourceLoader/ArticleResourceLoader.php
@@ -16,7 +16,7 @@ namespace Sulu\Article\Infrastructure\Sulu\Content\ResourceLoader;
 use Sulu\Article\Domain\Model\ArticleDimensionContentInterface;
 use Sulu\Article\Domain\Repository\ArticleRepositoryInterface;
 use Sulu\Content\Application\ContentResolver\Value\ContentView;
-use Sulu\Content\Application\ResourceLoader\Loader\ResourceLoaderContentViewInterface;
+use Sulu\Content\Application\ResourceLoader\Loader\ResourceLoaderContentViewEnhancementInterface;
 use Sulu\Content\Domain\Model\AuthorInterface;
 use Sulu\Content\Domain\Model\DimensionContentInterface;
 
@@ -25,7 +25,7 @@ use Sulu\Content\Domain\Model\DimensionContentInterface;
  *
  * @final
  */
-class ArticleResourceLoader implements ResourceLoaderContentViewInterface
+class ArticleResourceLoader implements ResourceLoaderContentViewEnhancementInterface
 {
     public const RESOURCE_LOADER_KEY = 'article';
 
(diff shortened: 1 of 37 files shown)


packages/content/src/Application/ContentResolver/ContentResolver.php
```diff
@@ -118,28 +126,51 @@ public function resolve(DimensionContentInterface $dimensionContent, ?array $pro
...
 
+        $finalContent = $replacerResult['content'];
+        $viewEnhancements = $replacerResult['viewEnhancements'];
+
+        foreach ($viewEnhancements as $path => $items) {
+            $existing = $this->propertyAccessor->getValue($resolvedContent['view'], $path);
+            $existingView = \is_array($existing) ? $existing : [];
+            if ([] === $existingView) {
+                $this->propertyAccessor->setValue($resolvedContent['view'], $path, $items);
+            } elseif (1 === \count($items) && \is_array($items[0])) {
```
Prokyonn: TODO: This branch will also be used for multi `selections` with only 1 selected item, which will lead to inconsistencies of the data for 1 or more items selected

packages/content/src/Application/ContentResolver/ContentResolver.php
```diff
@@ -118,28 +126,51 @@ public function resolve(DimensionContentInterface $dimensionContent, ?array $pro
...
+                $this->propertyAccessor->setValue($resolvedContent['view'], $path, \array_merge($existingView, $items[0]));
+            } else {
+                $existingView['items'] = $items;
```
Prokyonn: Not sure what the best way is to aggregate or nest the values per item, in the teasers we did already use `items` therefore I did here go with the same

packages/article/src/Infrastructure/Sulu/Content/ResourceLoader/ArticleResourceLoader.php
```diff
@@ -58,6 +61,30 @@ public function load(array $ids, ?string $locale, array $params = []): array
         return $mappedResult;
     }
 
+    public function resolveContentView(mixed $source): ContentView
```
wachterjohannes:
```suggestion
    public function enhanceContentView(mixed $source, ContentView $contentView): ContentView
```

or similar name

wachterjohannes (changes requested):

packages/content/src/Application/ContentResolver/ContentResolver.php
```diff
@@ -119,6 +134,29 @@ public function resolve(DimensionContentInterface $dimensionContent, ?array $pro
...
+                                $entityContentView = $resourceLoader->resolveContentView($childContent);
+                                $entityContentData = $entityContentView->getContent();
+
+                                if (\is_array($entityContentData) && [] !== $entityContentData) {
+                                    /** @var array<string, mixed> $existingContent */
+                                    $existingContent = $resolvedValue['content'];
+                                    $resolvedValue['content'] = \array_merge($existingContent, $entityContentData);
+                                }
+
+                                $contentView = ContentView::create([], $entityContentView->getView());
```
wachterjohannes:
```suggestion
                                $contentView = ContentView::create($resolvedValue, $entityContentView->getView());
```

packages/article/src/Infrastructure/Sulu/Content/ResourceLoader/ArticleResourceLoader.php
Prokyonn: renamed it to `resolveContentViewEnhancement` as this is only an enhancement that will be used to enhance the resolved data later on

packages/content/src/Application/ContentResolver/ContentResolver.php
Prokyonn: the [] is correct, as we did already enhance the content, but the viewEnhancement must be deferred until later, so we only have to delegate the viewEnhancements and contentEnhancement can be empty

wachterjohannes (approved):

PR #8616, opened 2026-02-10
Fix publishedState state in ArticleController cgetAction

martinlagler:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Change listBuilder order in ArticleController, because otherwise the initializeListBuilder function will overwrite all addSelectField calls.

#### Why?

Because the article_selection always shows `unpublished` publishedState.


Prokyonn:
@martinlagler looks good 👍  can you please also add additionally a published article in the `ArticleControllerTest` so that we also have tested the `publishedState: true` case

alexander-schranz (approved):

Prokyonn (approved):

PR #8618, opened 2026-02-10
Add array check for multiselection to avoid errors after content type…

martinlagler:
… change

| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Add array check for multi selection.

#### Why?

To avoid errors when the content type changed and the old data still exists.


alexander-schranz:
Currently the concept in all field types is error as it indicates missing migrations. We have currently I think no field type which ignores invalid data. We would need deeper discuss this handling.

alexander-schranz:
As discussed we should instead work with a error boundary to catch exception and provide a button to reset the data.

Think somewhere in the Form Renderer Field renderer here: https://github.com/sulu/sulu/blob/2cb43760e9492db4dcbb5368467771a602ea75b4/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/Field.js#L223

React Docs:
 - https://legacy.reactjs.org/docs/error-boundaries.html
 - https://react.dev/reference/react/Component#catching-rendering-errors-with-an-error-boundary

PR #8621, opened 2026-02-10
Rename Filename in MediaBundle

hual7:
| Q | A
| --- | ---
| New feature? | yes

#### What's in this PR?

Adds the possibility to set a locale file name in MediaBundle and overwrites the filename for output. This would be very helpful for defining SEO file names in Sulu.

#### Why?

It's annoying to upload a new file just to change the file name. Furthermore, the file name is not locale specific.

### DIFF
--- a/src/Sulu/Bundle/MediaBundle/Api/Media.php
+++ b/src/Sulu/Bundle/MediaBundle/Api/Media.php
@@ -343,6 +343,36 @@ class Media extends ApiWrapper
     }
 
     /**
+     * @param string|null $seoFilename
+     *
+     * @return $this
+     */
+    public function setSeoFilename($seoFilename)
+    {
+        $this->getMeta(true)->setSeoFilename($seoFilename);
+
+        return $this;
+    }
+
+    /**
+     * Returns SEO filename for media.
+     *
+     * @return string|null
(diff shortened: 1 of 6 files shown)


alexander-schranz:
This sadly not as an easy task as it first look like. We would require a full history of filenames and not can just overwrite the filename else redirects which is SEO required don't work as expected. That is why currently there the only way its can be "renamed" is uploading a new file version as that handles all edge cases correctly with a clear history.

But I personally like the idea, and maybe we can plan something for 3.1 to provide something a new feature like this. We need discuss how the data structure for such history should be (create a new FileVersion vs. FileVersionNameHistory).

From UI perspective it should be in my opinion more a action and overlay where also a hint is shown that it's a heavy process triggered here as it invalidates image formats caches.

hual7:
Without renaming the original filename it should be easy to integrate and only generating a new cache file if the locale "seo-filename" is set. Both files will be valid, without deleting the original. If the locale seo filename is overwritten, a new file is written, without deleting the history of cached files.

Bildschirmfoto 2026-02-10 um 21 19 55

 
Yes, a history would be great, when the file cache needs to be regenerated and/or redirects must be generated.

PR #8625, opened 2026-02-11
Add AccessControl to NavigationRepository

### LINKED ISSUE #8565: Permissions are not checked for navigation and page listings
| Q | A |
| --- | --- |
| Sulu Version | 3.0.3 |
| PHP Version | 8.3 |
| DB Version | MySQL 8 |
| Browser Version | Chrome 142 |

## Actual Behavior

In **Sulu 3.0.3**, pages that are protected by security permissions are still visible to **anonymous users**, even though they should not be.

The webspace is configured with security enabled:

```xml
    <system>intranet</system>
</security>
```

According to the documentation:

> Sulu will automatically check in this webspace if the current visitor is allowed to see the current page and only show pages in the navigation and smart content listings if the visitor has sufficient permissions.

However, when accessing the website as an anonymous user:

Navigation menus still display pages that require specific roles

- page_selection properties return pages that should not be accessible
- smart_content properties also include restricted pages
- Direct access to these pages is correctly denied, but they should not be visible at all

# Expected Behavior

When permission-check="true" is enabled on a webspace:

Pages requiring specific roles should not appear:

  - in navigation menus
  - in page_selection results
  - in smart_content results
- Anonymous users should only see pages they are allowed to access
- Permission checks should be applied consistently across all page listings

# Steps to Reproduce

1. Configure a webspace with security enabled:

```xml
    <system>intranet</system>
</security>
```

2. Initialize security roles:

```shell
bin/console sulu:security:init
```

3. Create a page that requires a specific role in the intranet security system.
4. Ensure the page is not accessible to the anonymous role.
5. Access the website as an anonymous user.
6. Observe:

  - the navigation menu
  - a page_selection field
  - a smart_content field

7. Notice that restricted pages are still visible.

# Additional Information

- The anonymous role exists and is active.
- Access control works correctly when accessing restricted pages directly.
- The issue affects all page listing mechanisms (navigation, page selection, smart content).
- This suggests that permission checks are not consistently applied when resolving page collections.

Prokyonn:
| Q | A
  | --- | ---
  | Bug fix? | yes
  | New feature? | no
  | BC breaks? | no
  | Deprecations? | yes 
  | Fixed tickets | fixes #8565
  | Related issues/PRs | #
  | License | MIT

  #### What's in this PR?

  - Added ACL query filtering to `NavigationRepository` so that pages denied to the anonymous role (or current user) are excluded from website navigation, matching the behavior already present in `PageRepository`
  - Introduced `AccessControlQueryEnhancer` integration in `NavigationRepository::createQueryBuilder()`, respecting webspace security configuration
  - Breadcrumb queries skip access control (`skipAccessControl: true`) so ancestors always appear regardless of permissions
  - Added deprecation warning when `$webspaceManager`, `$accessControlQueryEnhancer`, or `$permissions` constructor arguments are not provided
  - Added comprehensive permission tests covering flat/tree navigation, mixed permissions, children filtering, breadcrumb behavior, and pages without explicit ACL entries

  #### Why?

  When a webspace has `permission-check="true"` enabled, pages with denied permissions still appeared in website navigation menus. The `NavigationRepository` was not applying any ACL filtering, even though
  `PageRepository` already does. This PR adds the.       missing ACL integration to `NavigationRepository` so permission checks are consistently applied across all page listing mechanisms.

### DIFF
--- a/packages/page/src/Infrastructure/Doctrine/Repository/NavigationRepository.php
+++ b/packages/page/src/Infrastructure/Doctrine/Repository/NavigationRepository.php
@@ -68,6 +68,10 @@ class NavigationRepository implements NavigationRepositoryInterface
         private ?Security $security = null,
         private ?array $permissions = null,
     ) {
+        if (null === $webspaceManager || null === $accessControlQueryEnhancer || null === $permissions) {
+            @trigger_deprecation('sulu/sulu', '3.0', 'Not passing the "$webspaceManager", "$accessControlQueryEnhancer" and "$permissions" arguments to "%s" is deprecated and they will be required in a future version.', __METHOD__);
+        }
+
         $repository = $entityManager->getRepository(PageInterface::class);
         Assert::isInstanceOf($repository, NestedTreeRepository::class);
 
@@ -238,13 +242,13 @@ class NavigationRepository implements NavigationRepositoryInterface
      */
     private function fetchNestedSetValues(string $uuid, array $fields): ?array
     {
-        $qb = $this->entityRepository->createQueryBuilder('page');
-        $qb->select(...\array_map(fn ($f) => "page.{$f}", $fields))
+        $queryBuilder = $this->entityRepository->createQueryBuilder('page');
+        $queryBuilder->select(...\array_map(fn ($f) => "page.{$f}", $fields))
             ->where('page.uuid = :uuid')
             ->setParameter('uuid', $uuid);
 
         /** @var array<string, int|string>|null $result */
-        $result = $qb->getQuery()->getOneOrNullResult();
+        $result = $queryBuilder->getQuery()->getOneOrNullResult();
 
         return $result;
     }
@@ -537,7 +541,7 @@ class NavigationRepository implements NavigationRepositoryInterface
         if ($permission) {
             $this->accessControlQueryEnhancer->enhance(
                 $queryBuilder,
-                $user instanceof UserInterface ? $user : null,
+                $user,
                 $permission,
                 Page::class,
                 $alias,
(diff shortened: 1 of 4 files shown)


packages/page/src/Infrastructure/Doctrine/Repository/NavigationRepository.php
```diff
@@ -48,13 +54,24 @@ class NavigationRepository implements NavigationRepositoryInterface
...
     public function __construct(
         EntityManagerInterface $entityManager,
         private DimensionContentQueryEnhancer $dimensionContentQueryEnhancer,
         private ContentAggregatorInterface $contentAggregator,
         private ContentResolverInterface $contentResolver,
         private bool $audienceTargetingEnabled = false,
+        private ?WebspaceManagerInterface $webspaceManager = null,
+        private ?AccessControlQueryEnhancer $accessControlQueryEnhancer = null,
+        private ?Security $security = null,
+        private ?array $permissions = null,
```
alexander-schranz: I think we should make the implementation `@internal` and `final` and give only Backwards compatibility promise on the Interface, to make fixing bugs, add features and improve performance easier. 

```suggestion
        private WebspaceManagerInterface $webspaceManager,
        private AccessControlQueryEnhancer $accessControlQueryEnhancer,
        private ?Security $security,
        private array $permissions,
        private bool $audienceTargetingEnabled = false,
```

Its a too critical part in my opinion of the System to have not full control of it.

packages/page/src/Infrastructure/Doctrine/Repository/NavigationRepository.php
```diff
@@ -493,9 +512,44 @@ private function createQueryBuilder(array $filters): QueryBuilder
...
+        string $webspaceKey,
+        string $alias
+    ): void {
+        if (!$this->webspaceManager || !$this->accessControlQueryEnhancer) {
+            return;
+        }
+
+        $webspace = $this->webspaceManager->findWebspaceByKey($webspaceKey);
+        /** @var UserInterface|null $user */
+        $user = $webspace && $webspace->hasWebsiteSecurity() && $this->security ? $this->security->getUser() : null;
```
alexander-schranz: We should not expect `UserInterface` on the website part there can be different firewalls with different user providers: 


```suggestion
        $user = $webspace && $webspace->hasWebsiteSecurity() && $this->security?->getUser();
        
        if (! $user instanceof UserInterface) {
            $user = null;
        }
```

alexander-schranz (changes requested):

alexander-schranz (approved):

PR #8627, opened 2026-02-12
Improve LinkProvider performance and generate localized urls

### LINKED ISSUE #8611: Sulu 3.0 Page and Article Link Providers not resolved
| Q | A
| --- | ---
| Sulu Version | 3.0.3
| PHP Version | 8.4.17 (unrelated)
| DB Version | MySQL 8.4 (unrelated)
| Browser Version | Firefox (unrelated)

# Actual Behavior

Sulu 3.0 Page and Article Link Providers not resolved e.g. Page with `/en` URL get not resolved to `/en/test`. 

# Expected Behavior

Resolve the URL of Article  and PageLinkProviders.

# Additional Performance Improvement

Do not fetch any entities in LinkProviders and fetch just what is required to build the URL. Create an own query instead.

Prokyonn:
| Q | A
  | --- | ---
  | Bug fix? | yes
  | New feature? | no
  | BC breaks? | no
  | Deprecations? | no
  | Fixed tickets | fixes #8611
  | Related issues/PRs | #
  | License | MIT

  #### What's in this PR?

  - Replaced entity hydration in `PageLinkProvider` and `ArticleLinkProvider` with direct DQL queries that fetch only the columns needed (uuid, title, slug, webspaceKey)
  - Resolved URLs via `WebspaceManagerInterface::findUrlByResourceLocator()` instead of relying on `templateData['url']` which was never populated
  - Fixed `PageLinkDimensionContentEnhancer` to resolve internal link URLs from routes instead of reading empty template data
  - Added webspace-aware URL resolution for articles (mainWebspace / additionalWebspace / requestWebspace fallback)
  - Extracted shared `LinkUrlTrait` for query string and anchor appending
  - Added comprehensive unit tests for both providers including edge cases (draft stage, null slug, null webspace, unresolvable URLs)

  #### Why?

  In Sulu 3.0, `PageLinkProvider`, `ArticleLinkProvider` did not take the webspace localization into account on the generated urls.

  The fix replaces the heavy ContentManager/ContentAggregator abstraction layer with lean DQL queries and uses `findUrlByResourceLocator()` to build full URLs from route slugs.

### DIFF
--- a/packages/article/src/Infrastructure/Sulu/Content/ArticleLinkProvider.php
+++ b/packages/article/src/Infrastructure/Sulu/Content/ArticleLinkProvider.php
@@ -13,26 +13,28 @@ declare(strict_types=1);
 
 namespace Sulu\Article\Infrastructure\Sulu\Content;
 
+use Doctrine\ORM\EntityManagerInterface;
+use Sulu\Article\Domain\Model\ArticleDimensionContent;
 use Sulu\Article\Domain\Model\ArticleInterface;
-use Sulu\Article\Domain\Repository\ArticleRepositoryInterface;
 use Sulu\Bundle\HttpCacheBundle\ReferenceStore\ReferenceStoreInterface;
 use Sulu\Bundle\MarkupBundle\Markup\Link\LinkConfigurationBuilder;
 use Sulu\Bundle\MarkupBundle\Markup\Link\LinkItem;
 use Sulu\Bundle\MarkupBundle\Markup\Link\LinkProviderInterface;
-use Sulu\Content\Application\ContentManager\ContentManagerInterface;
+use Sulu\Component\Webspace\Analyzer\RequestAnalyzerInterface;
+use Sulu\Component\Webspace\Manager\WebspaceManagerInterface;
 use Sulu\Content\Domain\Model\DimensionContentInterface;
-use Sulu\Content\Infrastructure\Doctrine\DimensionContentQueryEnhancer;
 use Symfony\Contracts\Translation\TranslatorInterface;
 
 /**
- * @interal This class is an integration to the SuluMarkupBundle and can be changed any time.
- *          Use Symfony dependency injection service decoration and change the behaviour if you need.
+ * @internal This class is an integration to the SuluMarkupBundle and can be changed any time.
+ *           Use Symfony dependency injection service decoration and change the behaviour if you need.
  */
 final class ArticleLinkProvider implements LinkProviderInterface
 {
     public function __construct(
-        private readonly ContentManagerInterface $contentManager,
-        private readonly ArticleRepositoryInterface $articleRepository,
+        private readonly EntityManagerInterface $entityManager,
+        private readonly WebspaceManagerInterface $webspaceManager,
+        private readonly RequestAnalyzerInterface $requestAnalyzer,
         private readonly ReferenceStoreInterface $referenceStore,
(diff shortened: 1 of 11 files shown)


packages/page/src/Infrastructure/Sulu/Content/ContentResolver/PageLinkDimensionContentEnhancer.php
```diff
@@ -26,10 +29,13 @@
  */
 class PageLinkDimensionContentEnhancer implements DimensionContentEnhancerInterface
 {
+    use LinkUrlTrait;
+
     public function __construct(
         private PageRepositoryInterface $pageRepository,
         private ContentAggregatorInterface $contentAggregator,
         private LinkProviderPoolInterface $linkProviderPool,
+        private WebspaceManagerInterface $webspaceManager,
```
alexander-schranz: Would use the `RouteGeneratorInterface` not longer the webspaceManager for generating the URLs.

packages/article/src/Infrastructure/Sulu/Content/ArticleLinkProvider.php
```diff
@@ -13,26 +13,28 @@
...
+ * @internal This class is an integration to the SuluMarkupBundle and can be changed any time.
+ *           Use Symfony dependency injection service decoration and change the behaviour if you need.
  */
 final class ArticleLinkProvider implements LinkProviderInterface
 {
     public function __construct(
-        private readonly ContentManagerInterface $contentManager,
-        private readonly ArticleRepositoryInterface $articleRepository,
+        private readonly EntityManagerInterface $entityManager,
+        private readonly WebspaceManagerInterface $webspaceManager,
```
alexander-schranz: would use the `RouteGeneratorInterface` instead of the webspace manager. Also there you can create a functional fake instead of mock it: https://github.com/sulu/sulu/blob/2cb43760e9492db4dcbb5368467771a602ea75b4/packages/route/tests/Unit/Application/Routing/Generator/RouteGeneratorTest.php#L32-L68

alexander-schranz (approved):

packages/page/tests/Functional/Infrastructure/Sulu/Content/PageSmartContentResolverTest.php
```diff
@@ -162,7 +167,7 @@ public function testResolveSmartContentWithLinks(): void
 
         // Test internal link - resolves to target page's URL with custom title
         $this->assertArrayHasKey('Internal Link Page', $pagesByTitle);
-        $this->assertSame('/target-page', $pagesByTitle['Internal Link Page']['url']);
+        $this->assertSame('http://sulu.io/en/target-page', $pagesByTitle['Internal Link Page']['url']);
```
Prokyonn, line 170: @alexander-schranz hmm an `internal` link will now be rendered with the full url.

default content pages will still only return the unlocalized path.

alexander-schranz: Sure? Isn't this not a test only case where rouer context is localhost. The `routeGenerator` normally return absolut path not absolute url like Symfony route generator.

Prokyonn: yes you are right, this is because the portal / router context is not matching 👍

PR #8628, opened 2026-02-12
Fix smart content provider (Page webspace permissions, article webspace filter)

martinlagler:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| License | MIT

#### What's in this PR?

Check for webspace permissions in the page smart_content provider.
Only load articles with the current webspace as mainWebspace or additionalWebspace in the smart_content provider.

#### Why?

To avoid permission issues and have the same logic as in 2.6.x.

### DIFF
--- a/packages/article/tests/Functional/Infrastructure/Sulu/Content/ArticleSmartContentProviderTest.php
+++ b/packages/article/tests/Functional/Infrastructure/Sulu/Content/ArticleSmartContentProviderTest.php
@@ -16,6 +16,7 @@ namespace Sulu\Article\Tests\Functional\Infrastructure\Sulu\Content;
 use PHPUnit\Framework\Attributes\DataProvider;
 use Sulu\Article\Application\Message\ApplyWorkflowTransitionArticleMessage;
 use Sulu\Article\Application\Message\CreateArticleMessage;
+use Sulu\Article\Application\Message\ModifyArticleMessage;
 use Sulu\Article\Domain\Model\ArticleInterface;
 use Sulu\Bundle\AdminBundle\SmartContent\SmartContentProviderInterface;
 use Sulu\Bundle\CategoryBundle\Entity\CategoryInterface;
@@ -41,6 +42,8 @@ use Symfony\Component\Messenger\Stamp\HandledStamp;
  *     excerptTags?: string[],
  *     author?: int|null,
  *     authored?: string|null,
+ *     mainWebspace?: string,
+ *     additionalWebspace?: string[],
  * }
  *
  * @phpstan-import-type SmartContentBaseFilters from SmartContentProviderInterface
@@ -128,12 +131,35 @@ class ArticleSmartContentProviderTest extends SuluTestCase
             'template' => 'blog',
         ]);
 
+        self::modifyArticle(
+            self::$articles['tech2']->getUuid(),
+            [
+                'title' => 'Cloud Computing',
+                'excerptCategories' => [self::$categories['tech']->getId(), self::$categories['business']->getId()],
+                'excerptTags' => [self::$tags['cloud']],
+                'authored' => '2023-02-20T14:30:00+00:00',
+                'template' => 'blog',
+                'mainWebspaceKey' => 'blog',
+            ],
+        );
+
         self::$articles['sports1'] = self::createArticle([
             'title' => 'Football Season',
             'excerptCategories' => [self::$categories['sports']->getId()],
             'excerptTags' => [self::$tags['football']],
             'authored' => '2023-03-10T09:15:00+00:00',
         ]);
+
+        self::modifyArticle(
+            self::$articles['sports1']->getUuid(),
+            [
+                'title' => 'Football Season',
+                'excerptCategories' => [self::$categories['sports']->getId()],
+                'excerptTags' => [self::$tags['football']],
+                'authored' => '2023-03-10T09:15:00+00:00',
+                'additionalWebspace' => ['blog'],
+            ],
+        );
 
         self::$articles['sports2'] = self::createArticle([
             'title' => 'Tennis Championship',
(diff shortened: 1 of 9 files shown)


src/Sulu/Bundle/AdminBundle/SmartContent/SmartContentProviderInterface.php
```diff
@@ -33,6 +33,7 @@
  *      offset: int,
  *      includeSubFolders: bool,
  *      excludeDuplicates: bool,
+ *      webspaceKey?: string|null,
```
martinlagler, line 36: Is it a problem if the webspaceKey param is optional here ?

alexander-schranz: would always give it to the smart content so providers don't require to check if the key exists, we do the same for the other nullables like dataSource and limit, ... .

```suggestion
 *      webspaceKey: string|null,
```

alexander-schranz: but unsure if we should inject the request analyzer here or how currently pages handles `segmentKey` and `targetGroupid`, ...

How is the website handling this? The changes seems effect only the admin, and so unsure if it should be just a requestAnalyzer->getWebspace instead.

packages/page/src/UserInterface/Controller/Admin/PageController.php
```diff
@@ -314,7 +321,7 @@ private function handleAction(Request $request, string $uuid): ?PageInterface //
             /** @var PageInterface|null */
             return $this->handle(new Envelope($message, [new EnableFlushStamp()]));
         } elseif ('restore' === $action) {
-            $version = \intval($request->query->get('version'));
+            $version = (int) $request->query->get('version');
```
alexander-schranz:
```suggestion
            $version = $request->query->getInt('version');
```

alexander-schranz (commented):
How did 2.6 handle the filtering by webspace Key in smart content. Was there the request analyzer used inside the providers or what was the case there?

martinlagler:
@alexander-schranz 
For the admin a parameter was used [see](https://github.com/sulu/sulu/blob/2.6/src/Sulu/Bundle/PageBundle/Controller/SmartContentItemController.php#L74)
and for the website the request analyzer was used [see](https://github.com/sulu/sulu/blob/2.6/src/Sulu/Component/SmartContent/ContentType.php#L227)

src/Sulu/Bundle/AdminBundle/SmartContent/SmartContentProviderInterface.php
Prokyonn: @alexander-schranz Why would you add it directly to the interface?

Currently the webspace is added to the filters in the same way as segmentKey / targetGroupId with the SmartContentFiltersVisitor pattern. 
- [PageSmartContentFiltersVisitor](https://github.com/sulu/sulu/blob/3.0/packages/page/src/Infrastructure/Sulu/Content/Visitor/PageSmartContentFiltersVisitor.php)
- [SegmentSmartContentFiltersVisitor](https://github.com/sulu/sulu/blob/3.0/packages/page/src/Infrastructure/Sulu/Content/Visitor/SegmentSmartContentFiltersVisitor.php)
- [AudienceTargetingSmartContentFiltersVisitor](https://github.com/sulu/sulu/blob/3.0/src/Sulu/Bundle/AudienceTargetingBundle/Visitor/AudienceTargetingSmartContentFiltersVisitor.php)
- [MediaSmartContentFiltersVisitor](https://github.com/sulu/sulu/blob/3.0/src/Sulu/Bundle/MediaBundle/Infrastructure/Sulu/Content/Visitor/MediaSmartContentFiltersVisitor.php)

I don't think that every SmartContentProvider must know the webspaceKey 🤔

alexander-schranz: @Prokyonn missed that part as it was not inside this pull request, then I'm fine with it being `?...`.

alexander-schranz: @Prokyonn should we rename `PageSmartContentFiltersVisitor` to `WebspaceSmartContentFiltersVisitor`?

Prokyonn: > @Prokyonn should we rename PageSmartContentFiltersVisitor to WebspaceSmartContentFiltersVisitor?

Yes we can do that 👍

packages/article/tests/Functional/Infrastructure/Sulu/Content/ArticleSmartContentProviderTest.php
```diff
@@ -128,13 +131,36 @@ public static function setUpBeforeClass(): void
             'template' => 'blog',
         ]);
 
+        self::modifyArticle(
```
Prokyonn, line 134: Why not updating the initial createArticle method call?

packages/article/tests/Functional/Infrastructure/Sulu/Content/ArticleSmartContentProviderTest.php
```diff
@@ -128,13 +131,36 @@ public static function setUpBeforeClass(): void
...
+
         self::$articles['sports1'] = self::createArticle([
             'title' => 'Football Season',
             'excerptCategories' => [self::$categories['sports']->getId()],
             'excerptTags' => [self::$tags['football']],
             'authored' => '2023-03-10T09:15:00+00:00',
         ]);
 
+        self::modifyArticle(
+            self::$articles['sports1']->getUuid(),
```
Prokyonn, line 154: Why not updating the initial createArticle method call?

packages/article/tests/Functional/Infrastructure/Sulu/Content/ArticleSmartContentProviderTest.php
```diff
@@ -703,6 +729,45 @@ private static function createArticle(
         return $article;
     }
 
+    /**
+     * @param ArticleData $data
+     */
+    private static function modifyArticle(
```
Prokyonn, line 735: tbh I don't see why we need this method, as it only edits articles that could be created with this data initially 🤔

src/Sulu/Bundle/AdminBundle/SmartContent/SmartContentProviderInterface.php
alexander-schranz: done

packages/page/tests/Functional/Integration/PageControllerPermissionsTest.php
```diff
@@ -0,0 +1,200 @@
...
+{
+    use AssertSnapshotTrait;
+    use CreatePageWithPermissionsTrait;
+
+    /**
+     * @var KernelBrowser
+     */
+    protected $client;
+
+    protected function setUp(): void
```
Prokyonn, line 45: can be removed

packages/page/tests/Functional/Integration/PageControllerPermissionsTest.php
```diff
@@ -0,0 +1,200 @@
...
+    {
+    }
+
+    private function createHomepage(string $uuid, string $webspaceKey): PageInterface
```
Prokyonn, line 49: Why not using the CreatePage trait?

packages/page/tests/Functional/Integration/PageControllerPermissionsTest.php
```diff
@@ -0,0 +1,200 @@
...
+        self::getEntityManager()->persist($homepage);
+        self::getEntityManager()->flush();
+
+        return $homepage;
+    }
+
+    /**
+     * @param array<string, mixed> $data
+     */
+    private function createPage(
```
Prokyonn, line 65: Why not using the CreatePage trait?

alexander-schranz (approved):

Prokyonn (approved):
We can fix the test setup comments in a followup :)

packages/article/tests/Functional/Infrastructure/Sulu/Content/ArticleSmartContentProviderTest.php
martinlagler: I think it's not possible to set mainWebspaceKey/additionalWebspace on create.

PR #8634, opened 2026-02-12
Add case-insensitive LIKE search for PostgreSQL

marek-kava:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?

Wrap LIKE queries with LOWER() on both column and search parameter to ensure case-insensitive search across all supported databases.      

Changed files:             
  - `AbstractDoctrineFieldDescriptor::getSearch()`
  - `DoctrineCaseFieldDescriptor::getSearch()`    
  - `ListQueryBuilder` - dynamic LIKE search builder
  - `ContactRepository` - firstName/lastName search
  - `MediaRepository` - media title search
  - `CollectionRepository` - collection title search

#### Why?

LIKE operator in PostgreSQL is case-sensitive by default. Searching for "Opel" won't find "opel" in admin lists. MySQL handles this implicitly via its default utf8_general_ci collation, so the issue only affects PostgreSQL users.

LOWER() LIKE LOWER() is preferred over ILIKE (PostgreSQL-specific) and COLLATE (requires schema migration). Since LIKE '%...%' already skips indexes, LOWER() has no performance impact.

### DIFF
--- a/src/Sulu/Bundle/ContactBundle/Entity/ContactRepository.php
+++ b/src/Sulu/Bundle/ContactBundle/Entity/ContactRepository.php
@@ -424,7 +424,7 @@ class ContactRepository extends EntityRepository implements DataProviderReposito
             ->where(\sprintf('contact.id NOT IN (%s)', $subQueryBuilder->getDQL()));
 
         if ($search) {
-            $queryBuilder->andWhere('contact.firstName LIKE :firstNameSearch or contact.lastName LIKE :lastNameSearch');
+            $queryBuilder->andWhere('LOWER(contact.firstName) LIKE LOWER(:firstNameSearch) or LOWER(contact.lastName) LIKE LOWER(:lastNameSearch)');
             $queryBuilder->setParameter('firstNameSearch', '%' . $search . '%');
             $queryBuilder->setParameter('lastNameSearch', '%' . $search . '%');
         }
--- a/src/Sulu/Bundle/MediaBundle/Entity/CollectionRepository.php
+++ b/src/Sulu/Bundle/MediaBundle/Entity/CollectionRepository.php
@@ -211,7 +211,7 @@ class CollectionRepository extends NestedTreeRepository implements CollectionRep
                 $qb->setParameter('depth', \intval($depth));
             }
             if (null !== $search) {
-                $qb->andWhere('collectionMeta.title LIKE :search');
+                $qb->andWhere('LOWER(collectionMeta.title) LIKE LOWER(:search)');
                 $qb->setParameter('search', '%' . $search . '%');
             }
             if (null !== $offset) {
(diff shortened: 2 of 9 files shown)


alexander-schranz:
Unsure about the effects on performance. Just know that `lower` would not use an Index, but unsure as searches are using `%...%` and my old knowledge was that only direct match and `...%` uses indexes and `%...%` not. How much this effects the performance?

/cc @wachterjohannes @Prokyonn

wachterjohannes:
> Unsure about the effects on performance. Just know that `lower` would not use an Index, but unsure as searches are using `%...%` and my old knowledge was that only direct match and `...%` uses indexes and `%...%` not. How much this effects the performance?
> 
> /cc @wachterjohannes @Prokyonn

@alexander-schranz yeah %...% is already a bad thing so i would say adding LOWER does not add any complexity - so i asked claude about that and this was the answer:

```
LIKE '%...%' in PostgreSQL — no B-tree index either way. Your old knowledge is correct: B-tree indexes only help with LIKE 'prefix%' patterns. With a leading wildcard, it's a sequential scan regardless, so LOWER() doesn't "lose" you an index.
The cost of LOWER() per row is negligible. It's a simple function call during the sequential scan. On tables up to hundreds of thousands of rows, the difference is essentially unmeasurable compared to the I/O cost of the scan itself.
Better alternatives in PostgreSQL:

ILIKE — PostgreSQL's built-in case-insensitive LIKE. WHERE column ILIKE '%term%' does the same as WHERE LOWER(column) LIKE LOWER('%term%') but is cleaner. Performance is virtually identical though — still a sequential scan with %...%.
pg_trgm extension + GIN index — this is the real answer if performance matters. Trigram indexes actually can accelerate %...% searches:

sql   CREATE EXTENSION pg_trgm;
   CREATE INDEX idx_column_trgm ON table USING GIN (column gin_trgm_ops);
This makes LIKE '%term%', ILIKE '%term%', and even LOWER(column) LIKE '%term%' use the index. Massive speedup on large tables.

Full-text search (tsvector/tsquery) — for natural language search rather than substring matching.

Bottom line: For your %...% queries, LOWER() vs no LOWER() has virtually zero performance impact since both result in a sequential scan. If the table grows large enough that the scan becomes a problem, pg_trgm with a GIN index is the proper solution — and it supports case-insensitive patterns too.
```

so i would say it make sense to add it as the UX increases BUT the cost of performance should be negligible

Prokyonn (approved):

alexander-schranz:
Is the same true for MariaDB and MySQL? @wachterjohannes

alexander-schranz (approved):

wachterjohannes:
@alexander-schranz yes - but if we use a _ci collation (utf8mb4_unicode_ci for example) it would not be neccesary

wachterjohannes (approved):

marek-kava:
@chirimoya hello, is there a chance these changes will be approved?

chirimoya (approved):

Issue #8637, opened 2026-02-13
Add x-default hreflang tag to SEO extension template


rogoit:
| Q | A                                                                                                                                  
  | --- | ---                                           
  | Sulu Version | 2.6                                                                                                                     
  | PHP Version | 8.3                                                                                                                    
  | DB Version | MySQL 8.0                                                                                                                 
  | Browser Version | N/A                                                                                                                  

  # Actual Behavior

  Sulu renders `hreflang="de"` and `hreflang="en"` but never adds `hreflang="x-default"`. Every multi-language Sulu site is missing this.
  SEO tools flag it on every single page.

  # Expected Behavior

  An additional `<link rel="alternate" href="..." hreflang="x-default">` pointing to the default locale should be rendered. Google
  recommends this for all multi-language sites:
  https://developers.google.com/search/docs/specialty/international/localized-versions#xdefault

  # Steps to Reproduce

  1. Set up two localizations in webspace config
  2. Publish any page
  3. View source — no `x-default` hreflang

  # Possible Solutions

  Add after the hreflang loop in `src/Sulu/Bundle/WebsiteBundle/Resources/views/Extension/seo.html.twig`:

  ```twig
  {%- if localizations[defaultLocale|default('de')]|default() -%}
      <link rel="alternate" href="{{ localizations[defaultLocale|default('de')].url }}" hreflang="x-default">
  {%- endif -%}

  Happy to provide a PR.

alexander-schranz:
Thank you for your effort,

The `x-default` is on most SEO websites / suggestions defined as a fallback "Locale", but thats widely spreaded misinformation and not correct. Its a very special own URL where the user can choose the Language or is redirect by its accept-languages. You can read about country selector part in google and yandex docs:

 - https://developers.google.com/search/docs/specialty/international/localized-versions?hl=de#xdefault
 - https://yandex.com/support/webmaster/en/yandex-indexing/locale-pages.html#auto

A example of a correct implementation of `x-default` is https://www.booking.com/ homepage:
The x-default: Which is there `https://www.booking.com/index.html` which redirects you by your accept language to the correct page `https://www.booking.com/index.de.html`. Subpages has on booking.com no x-default.

Still doesn't mean that subpages can not also have a x-default in case of google developer docs it is so there is the x-default `https://developers.google.com/search/docs/specialty/international/localized-versions` and you get redirected to your language `https://developers.google.com/search/docs/specialty/international/localized-versions?hl=de`.

[...]

PR #8638, opened 2026-02-13
Add hreflang x-default to seo template (#8637)

rogoit:
Hope it helps.


alexander-schranz:
Thx for the effort. But as written the x-default is not correct here and also this would break existing projects with there own x-default definition. More details in the the issue: https://github.com/sulu/sulu/issues/8637#issuecomment-3901857033

Issue #5054, opened 2020-02-11
Medias Area adding a new Folder Form improvement


marcostastny:
# Problem description

Currently the Folder Form does not set Focus after showing the form.
Currently pressing Enter in the Folder Form does nothing.

# Proposed solution

From a user experience point of view it would be nice if the focus would be set automatically and if pressing enter would submit the form.

nnatter:
Also, the Form Overlay should close itself when pressing `esc` in my opinion 😇

danrot:
@nnatter I just tested, the overlay already closes itself when pressing `esc` :thinking: 

@marcostastny The question is if the suggestions you make should only apply to this very specific form. I think setting the focus on the first element would make sense in any form (actually I am asking myself right now why we haven't thought about that in the first place)
And regarding pressing enter: I would say that should not automatically save in e.g. the page form, but in the media form it would make sense. Maybe it makes sense for overlays in general? The question to me is if I want e.g. to confirm a delete dialog with a press on enter as well (would also be known from operating systems, since they also handle it this way) or if we are too afraid that people will start deleting stuff on accident.

nnatter:
@danrot The overlay does not close itself if one of the input fields is active/focused when `esc` is pressed. 🙂 

Regarding to focusing the first field automatically: I am in for that. 
Not sure about saving forms on enter. This might be confusing, especially if the form includes a `text_editor` or something like that.

danrot:
@nnatter Then the first `esc` press blurs the current field, and the second closes the overlay (at least for me on Linux...)

The enter part would be about the overlays and dialogs, not really about the form. As I've written before it would IMO not make any sense on e.g. the page form.

nnatter:
> Then the first esc press blurs the current field, and the second closes the overlay (at least for me on Linux...)

Does not work in Safari and Chrome on macOS Catalina for me..

marcostastny:
@danrot @nnatter From my testing i can confirm, esc closes the overlay when the focus is not already set. But pressing esc does not remove the focus from the input, so it will not close if the focus is already set. At least for me on Catalina and Chrome.

rogoit:
Safari? I only use chrome many others firefox i think ;)

PR #8639, opened 2026-02-14
Add autofocus param to text_line / Input field type

### LINKED ISSUE #8051: Autofocus for media collection add dialog
# Problem description

When adding a new media collection, it would be good usability, especially for power users, to autofocus the title field, as this is the first field, where a user needs to enter data.
Additionally, it would be great, if the input type="text" field has the focus, the form would be submitted on pressing enter.

# Proposed solution

Allow using the keyboard to fill and submit an new collection

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #8051, https://github.com/sulu/sulu/issues/5054
| Related issues/PRs | -
| License | MIT
| Documentation PR | https://github.com/sulu/sulu-docs/pull/896

#### What's in this PR?

Add possibility to autofocus a Input.

#### Why?

This makes keyboard editing easier in Sulu.


src/Sulu/Bundle/AdminBundle/FormMetadata/FormMetadataMapper.php
```diff
@@ -127,6 +127,7 @@ private function mapProperty(ContentPropertyMetadata $property, string $locale):
         $field->setType($property->getType());
         $field->setColSpan($property->getColSpan());
         $field->setRequired($property->isRequired());
+        $field->setAutoFocus($property->isAutoFocus());
```
alexander-schranz: this should just be a `<param name="autofocus" value="true" />` no own field metadata thing.

So on PHP side there are no changes required :)

You can access it the same was then as the `headline` `true` here: https://github.com/sulu/sulu/blob/aeb2a0b5fef566fc24b216184f2012b7f7d1f4c0/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/Input.js#L29-L31

On the frontend params are in the schemaOptions.

mamazu:
@alexander-schranz This is indeed much easier than I thought and now it's even configurable. :tada: Where would the documentation for this feature be, if I wanted to add that as well.

alexander-schranz:
@mamazu params are documented on the specific field type configs e.g.: https://docs.sulu.io/3.x/reference/property-types/text_line.html

alexander-schranz:
@mamazu can you add a similar test like here for the autofocus?: https://github.com/sulu/sulu/blob/aeb2a0b5fef566fc24b216184f2012b7f7d1f4c0/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/tests/fields/Input.test.js#L51-L68

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/Input.js
```diff
@@ -29,6 +29,9 @@ export default class Input extends React.Component<FieldTypeProps<?string>> {
                 headline: {
                     value: headline,
                 } = {},
+                autoFocus: {
```
alexander-schranz: all schemaOptions params currently seems to be snake_case here. Only the react params are camelCase. Should we go with `autofocus` think `auto_focus` is strange. /cc @Prokyonn

alexander-schranz (changes requested):

src/Sulu/Bundle/CustomUrlBundle/Resources/config/forms/custom_url_details.xml
```diff
@@ -12,6 +12,7 @@
             </meta>
             <params>
                 <param name="headline" value="true"/>
+                <param name="autoFocus" value="true"/>
```
alexander-schranz: see comment above:


```suggestion
                <param name="autofocus" value="true"/>
```

src/Sulu/Bundle/MediaBundle/Resources/config/forms/collection_details.xml
```diff
@@ -12,6 +12,7 @@
             </meta>
             <params>
                 <param name="headline" value="true"/>
+                <param name="autoFocus" value="true"/>
```
alexander-schranz: see comment above:


```suggestion
                <param name="autofocus" value="true"/>
```

alexander-schranz (changes requested):

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/Input.js
mamazu: I've renamed the property but the JS variable still stays like `autoFocus`?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/Input.js
```diff
@@ -76,6 +79,7 @@ export default class Input extends React.Component<FieldTypeProps<?string>> {
 
         return (
             <InputComponent
+                autoFocus={!!autoFocus}
```
alexander-schranz, line 82: our schemaOption variable is lower case so we need adjust here it also.

```suggestion
                autoFocus={!!autofocus}
```

alexander-schranz (changes requested):

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/Input.js
alexander-schranz: yes thats correct internal all variables are camelCase. And looks like react uses `autoFocus` so we stay there the same.

alexander-schranz (approved):

src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/fields/Input.js
```diff
@@ -76,6 +79,7 @@ export default class Input extends React.Component<FieldTypeProps<?string>> {
 
         return (
             <InputComponent
+                autoFocus={!!autofocus}
```
alexander-schranz: my failure:


```suggestion
                autoFocus={!!autoFocus}
```

alexander-schranz:
CI is related to: https://github.com/sulu/sulu/pull/8646

Prokyonn (approved):

PR #8644, opened 2026-02-16
Fix multi locale queries

### LINKED ISSUE #8623: Sulu 3.0: Retrieve sulu_snippet_load_by_area data from another locale
| Q | A
| --- | ---
| Sulu Version | 3.0.3
| PHP Version | 8.4

# Actual Behavior
When retrieving data via `{% set settings = sulu_snippet_load_by_area('webspace_settings', locale: 'de') %}` on EN page, I get the following error.

`An exception has been thrown during the rendering of a template ("Sulu\Bundle\AdminBundle\Metadata\FormMetadata\FormMetadataProvider::getMetadata(): Argument #2 ($locale) must be of type string, null given, called in /Users/sulu/sites/sulu-3/vendor/sulu/sulu/packages/content/src/Application/ContentResolver/Resolver/TemplateResolver.php on line 43").
`
# Expected Behavior
Get snippet data from locale DE or another locale that is not EN.

Prokyonn:
| Q | A
  | --- | ---
  | Bug fix? | yes
  | New feature? | no
  | BC breaks? | no
  | Deprecations? | no
  | Fixed tickets | fixes #8623
  | Related issues/PRs | #
  | License | MIT

  #### What's in this PR?

  - Added `createQuery()` method to `DimensionContentQueryEnhancer` that conditionally sets `Query::HINT_REFRESH` when the QueryBuilder joins `dimensionContent`
  - Updated all content repositories (Article, Page, Snippet, Navigation) to use `createQuery()` instead of `$queryBuilder->getQuery()`
  - Added unit tests for conditional hint behavior and a functional regression test loading the same entity with different locales

  #### Why?

  When the same content entity is loaded multiple times with different dimension attributes (e.g. first EN then DE locale) in a single request, Doctrine's identity map returns the cached entity from the first
  query. This means the `dimensionContents` collection stays stale — the second query returns EN dimension content instead of DE. Setting `HINT_REFRESH` forces Doctrine to re-hydrate entities from query results,
  ensuring each locale query returns the correct dimension content.

  #### To Do

### DIFF
--- a/packages/content/src/Infrastructure/Doctrine/DimensionContentQueryEnhancer.php
+++ b/packages/content/src/Infrastructure/Doctrine/DimensionContentQueryEnhancer.php
@@ -15,6 +15,7 @@ namespace Sulu\Content\Infrastructure\Doctrine;
 
 use Doctrine\Common\Collections\Criteria;
 use Doctrine\DBAL\ArrayParameterType;
+use Doctrine\ORM\Query;
 use Doctrine\ORM\Query\Expr\Join;
 use Doctrine\ORM\QueryBuilder;
 use Sulu\Content\Domain\Model\DimensionContentInterface;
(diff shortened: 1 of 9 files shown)


packages/content/src/Infrastructure/Doctrine/DimensionContentQueryEnhancer.php
```diff
@@ -67,6 +68,28 @@ class DimensionContentQueryEnhancer
...
+     *
+     * @return Query<mixed, mixed>
+     */
+    public function createQuery(QueryBuilder $queryBuilder): Query
+    {
+        $query = $queryBuilder->getQuery();
+
+        if (\in_array('dimensionContent', $queryBuilder->getAllAliases(), true)) {
+            $query->setHint(Query::HINT_REFRESH, true);
+        }
```
alexander-schranz, line 88: Would not do this in the Enhancer, specially because avoid such checks as for the alias. So would do it directly in the specific repositories.

Did also talk with @chirimoya about the case, its also in discussion if we disable for GET requests the UnitOfWork in generally 🤔

Prokyonn: Yes this is just a poc, I wasn't really sure where to put it either, but if possible I would like a more centralized way, so that you do not have to worry about it in every repository 🤔

PR #8645, opened 2026-02-17
Adding interface language in <html> tag

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no
| Deprecations? | no 
| Fixed tickets | -
| Related issues/PRs | #6444 
| License | MIT

#### What's in this PR?
Adding interface locale to the html lag tag.

#### Why?
Better accessibility.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/views/Admin/main.html.twig
+++ b/src/Sulu/Bundle/AdminBundle/Resources/views/Admin/main.html.twig
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="{{ app.request.locale|split('_')[0] }}">
+<html lang="{{ app.request.locale|replace({'_': '-'}) }}">
 <head>
     {%- block meta ~%}
         <meta charset="utf-8">
--- a/src/Sulu/Bundle/AdminBundle/Resources/views/Admin/main.html.twig
+++ b/src/Sulu/Bundle/AdminBundle/Resources/views/Admin/main.html.twig
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html>
+<html lang="{{ app.request.locale|split('_')[0] }}">
 <head>
     {%- block meta ~%}
         <meta charset="utf-8">


src/Sulu/Bundle/AdminBundle/Resources/views/Admin/main.html.twig
```diff
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html>
+<html lang="{{ app.request.locale|split('_')[0] }}">
```
alexander-schranz: is there already a mechanic im place (JS) which update the `lang` based on the users system locale?

mamazu: grafik

That's the html on sulu.rocks for the admin panel. It's not being set with js it seems.

Prokyonn:
```suggestion
<html lang="{{ app.request.locale|replace({'_': '-'}) }}">
```

I think sulu does not force to use `_` as a separator, so e.g. `en-US` would also be correctly parsed, WDYT?

mamazu: Yes that's better.

Prokyonn (approved):

src/Sulu/Bundle/AdminBundle/Resources/views/Admin/main.html.twig
alexander-schranz: we would require that the JS update it as the UI language depends on the user locale. so the user initializer may need change it after login.

mamazu: This definitely makes sense. But I think that the initial load (before any js has been run) should also have a default locale. We can still update the locale with js later too.

alexander-schranz: prefer no locale until the JS is tackled over a false locale ;)

mamazu: Fair enough, then I'll close this here and someone who knows js can tackle this issue.

PR #8647, opened 2026-02-19
Add FormMetadataProvider InMemory cache

Prokyonn:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? | yes
| BC breaks? | no 
| Deprecations? | no 
| Fixed tickets | fixes # 
| Related issues/PRs | # 
| License | MIT

#### What's in this PR?

This PR implements a very simple InMemory Cache for the FormMetadata to proof the value of it. 

If we decide to go with this approach, the implementation has to be adjusted properly with a CachedFormMetadata decorator and also implementing the ResetInterface.

#### Why?

FormMetadata are fetched multiple times and also all the visitors are executed especially in navigation requests multiple times. This could be a easy improvements.

### DIFF
--- a/packages/article/src/Infrastructure/Sulu/Search/Visitor/WebsiteArticleReindexContentEnhancer.php
+++ b/packages/article/src/Infrastructure/Sulu/Search/Visitor/WebsiteArticleReindexContentEnhancer.php
@@ -16,10 +16,10 @@ namespace Sulu\Article\Infrastructure\Sulu\Search\Visitor;
 use Doctrine\ORM\QueryBuilder;
 use Sulu\Bundle\AdminBundle\Metadata\FormMetadata\FieldMetadata;
 use Sulu\Bundle\AdminBundle\Metadata\FormMetadata\FormMetadata;
-use Sulu\Bundle\AdminBundle\Metadata\FormMetadata\FormMetadataProvider;
 use Sulu\Bundle\AdminBundle\Metadata\FormMetadata\ItemMetadata;
 use Sulu\Bundle\AdminBundle\Metadata\FormMetadata\TagMetadata;
 use Sulu\Bundle\AdminBundle\Metadata\FormMetadata\TypedFormMetadata;
+use Sulu\Bundle\AdminBundle\Metadata\MetadataProviderInterface;
 
 /**
  * @internal if you need to override this service, create a new service based on the WebsiteArticleReindexProviderEnhancerInterface
@@ -41,7 +41,7 @@ class WebsiteArticleReindexContentEnhancer implements WebsiteArticleReindexProvi
     ];
 
     public function __construct(
-        private FormMetadataProvider $formMetadataProvider,
+        private MetadataProviderInterface $formMetadataProvider,
     ) {
     }
 
@@ -90,7 +90,7 @@ class WebsiteArticleReindexContentEnhancer implements WebsiteArticleReindexProvi
     private function getTemplateFormMetadata(string $templateKey, string $locale): ?FormMetadata
     {
         /** @var TypedFormMetadata $formMetadata */
-        $formMetadata = $this->formMetadataProvider->getMetadata('article', $locale);
+        $formMetadata = $this->formMetadataProvider->getMetadata('article', $locale, []);
         /** @var array<string, FormMetadata> $forms */
         $forms = $formMetadata->getForms();
 
--- a/src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/CachedFormMetadataProvider.php
+++ b/src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/CachedFormMetadataProvider.php
@@ -18,7 +18,7 @@ use Symfony\Contracts\Service\ResetInterface;
 /**
  * @internal this class should not be extended or initialized by any application outside of sulu
  */
-class CachedFormMetadataProvider implements MetadataProviderInterface, ResetInterface
+final class CachedFormMetadataProvider implements MetadataProviderInterface, ResetInterface
 {
     /**
      * @var array<string, MetadataInterface>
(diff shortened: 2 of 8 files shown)


alexander-schranz:
As reported this seems to drop specially for website have lot of global blocks the ttfb from 900-800 to <300. As discussed I would move the caching logic into an own `CachedFormMetadataProvider`, and use instead of serialize the whole metadata (as it may contains in some cases specially in admin also objects) to use spl_obj_hash loop and use a faster `xxh128` algorithm over md5 which we already use in our http caches via psr6: https://github.com/Toflar/psr6-symfony-http-cache-store/blob/9d4c4b49df283c0f8cca7e5c5f6a798c6debd215/src/Psr6Store.php#L103

PS: ResetInterface and kernel.reset tag should also be added to the new `CachedFormMetadataProvider` which decorates existing implementation.

packages/article/src/Infrastructure/Sulu/Search/Visitor/WebsiteArticleReindexContentEnhancer.php
```diff
@@ -41,7 +41,7 @@ class WebsiteArticleReindexContentEnhancer implements WebsiteArticleReindexProvi
     ];
 
     public function __construct(
-        private FormMetadataProvider $formMetadataProvider,
+        private MetadataProviderInterface $formMetadataProvider,
```
alexander-schranz, line 44: Would like to change this later to the MetadataProviderRegistry but thats unrelated to this Pull Request.

src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/CachedFormMetadataProvider.php
```diff
@@ -0,0 +1,89 @@
...
+            'key' => $key,
+            'locale' => $locale,
+            'metadataOptions' => $metadataOptions,
+        ]);
+
+        if (isset($this->cache[$cacheKey])) {
+            return $this->cache[$cacheKey];
+        }
+
+        $metadata = $this->inner->getMetadata($key, $locale, $metadataOptions);
```
alexander-schranz: Would go for decorations with `...$args` operator if we may add in future some other optional params. Like we did in the past all args are forwarded: 

https://github.com/sulu/sulu/blob/378994c4a173ba70153216f8bbfb64c3c3017d8d/src/Sulu/Bundle/MediaBundle/Media/ImageConverter/MediaImageExtractor.php#L35

What do you think?

```suggestion
    public function getMetadata(...$args): MetadataInterface
    {
        $cacheKey = $this->buildCacheKey($args);

        if (isset($this->cache[$cacheKey])) {
            return $this->cache[$cacheKey];
        }

        $metadata = $this->inner->getMetadata(...$args);
```

src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/CachedFormMetadataProvider.php
```diff
@@ -0,0 +1,89 @@
...
+
+        if ($metadata->isCacheable()) {
+            $this->cache[$cacheKey] = $metadata;
+        }
+
```
alexander-schranz: For in memory cache I would not keep isCacheable in mind as that's currently used for "response" headers only.

```suggestion
        $this->cache[$cacheKey] = $metadata;

```

src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/CachedFormMetadataProvider.php
```diff
@@ -0,0 +1,89 @@
...
+            return \implode(',', $parts);
+        }
+
+        if (\is_object($value)) {
+            return \spl_object_hash($value);
+        }
+
+        if (!\is_scalar($value) && null !== $value) {
+            return '';
+        }
```
alexander-schranz: This seems only a theoretical state? Then I would just use `assert` which is compiled away on prod envs:


```suggestion
       \assert(\is_scalar($value) || $value === null, 'Expected a scalar value here got: ' . \get_debug_type($value));
```

src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/CachedFormMetadataProvider.php
```diff
@@ -0,0 +1,89 @@
...
+
+        return (string) $value;
```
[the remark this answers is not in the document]
alexander-schranz: don't see requirement for this. what do you see?

src/Sulu/Bundle/AdminBundle/Resources/config/services.xml
```diff
@@ -52,15 +52,23 @@
         </service>
 
         <service
-            id="sulu_admin.form_metadata_provider"
+            id="sulu_admin.form_metadata_provider.inner"
```
alexander-schranz: decorating inside the same .xml does not work?

src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/CachedFormMetadataProvider.php
```diff
@@ -0,0 +1,89 @@
...
+namespace Sulu\Bundle\AdminBundle\Metadata\FormMetadata;
+
+use Sulu\Bundle\AdminBundle\Metadata\MetadataInterface;
+use Sulu\Bundle\AdminBundle\Metadata\MetadataProviderInterface;
+use Symfony\Contracts\Service\ResetInterface;
+
+/**
+ * @internal this class should not be extended or initialized by any application outside of sulu
+ */
+class CachedFormMetadataProvider implements MetadataProviderInterface, ResetInterface
```
alexander-schranz:
```suggestion
final class CachedFormMetadataProvider implements MetadataProviderInterface, ResetInterface
```

src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/CachedFormMetadataProvider.php
```diff
@@ -0,0 +1,89 @@
...
+     */
+    private function buildCacheKey(array $input): string
+    {
+        return \hash('xxh128', $this->normalizeValue($input));
+    }
+
+    private function normalizeValue(mixed $value): string
+    {
+        if (\is_array($value)) {
+            \ksort($value);
```
alexander-schranz: I'm unsure if this does really bring something in our case? As I think calls are always the same and we not require any ksort? what do you think?

alexander-schranz (changes requested):

src/Sulu/Bundle/AdminBundle/Metadata/FormMetadata/CachedFormMetadataProvider.php
alexander-schranz: but we should add a test that this flag has no effects and still cached.

src/Sulu/Bundle/AdminBundle/Tests/Unit/Metadata/FormMetadata/CachedFormMetadataProviderTest.php
```diff
@@ -0,0 +1,187 @@
...
+        $cached = new CachedFormMetadataProvider($inner);
+
+        $result1 = $cached->getMetadata('page', 'en', ['a' => '1', 'b' => '2']);
+        $result2 = $cached->getMetadata('page', 'en', ['b' => '2', 'a' => '1']);
+
+        $this->assertSame($metadata, $result1);
+        $this->assertSame($metadata, $result2);
+        $this->assertSame(1, $callCount);
+    }
+
```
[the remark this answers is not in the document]
alexander-schranz: as we removed ksort we can close this one.

alexander-schranz (approved):

alexander-schranz:
This looks good to go we reduced in the profile of @dfsoeten the calls to the FormMetadataProvider from 342 to 6 and so visitors aren't process again every time which also includes expression handlers.

PR #8653, opened 2026-02-24
Fix copy locale from detail-view

Prokyonn:
| Q | A
  | --- | ---
  | Bug fix? | yes
  | New feature? | no
  | BC breaks? | no
  | Deprecations? | no
  | Fixed tickets | fixes #
  | Related issues/PRs | #
  | License | MIT

  #### What's in this PR?

  - Fixed copy locale action in the page detail-view when no explicit `src` parameter is provided
  - Added fallback to use the `locale` query parameter as source locale when `src` is absent
  - Support multiple comma-separated destination locales in the `dest` query parameter
  - Added functional test covering the `locale`-as-src fallback scenario

  #### Why?

  When triggering the `copy_locale` action from the detail-view, the `src` query parameter was not being sent, only `locale` was available. This caused the copy to fail because the handler received an empty source
   locale. The fix falls back to `locale` when `src` is absent, aligning the detail-view trigger with the list-view trigger.

### DIFF
--- a/packages/page/src/UserInterface/Controller/Admin/PageController.php
+++ b/packages/page/src/UserInterface/Controller/Admin/PageController.php
@@ -286,15 +286,23 @@ final class PageController implements SecuredControllerInterface, SecuredObjectC
         }
 
         if ('copy_locale' === $action) {
-            $message = new CopyLocalePageMessage(
-                ['uuid' => $uuid],
-                (string) $request->query->get('src'),
-                (string) $request->query->get('dest'),
-            );
+            $srcLocale = (string) ($request->query->get('src') ?: $request->query->get('locale'));
+            $destLocales = \array_filter(\array_map('trim', \explode(',', (string) $request->query->get('dest'))));
 
-            /** @see \Sulu\Page\Application\MessageHandler\CopyLocalePageMessageHandler */
-            /** @var PageInterface|null */
-            return $this->handle(new Envelope($message, [new EnableFlushStamp()]));
+            $result = null;
+            foreach ($destLocales as $destLocale) {
+                $message = new CopyLocalePageMessage(
+                    ['uuid' => $uuid],
+                    $srcLocale,
+                    $destLocale,
+                );
+
+                /** @see \Sulu\Page\Application\MessageHandler\CopyLocalePageMessageHandler */
+                /** @var PageInterface|null */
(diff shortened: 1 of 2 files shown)


packages/page/src/UserInterface/Controller/Admin/PageController.php
```diff
@@ -286,15 +286,23 @@ private function handleAction(Request $request, string $uuid): ?PageInterface //
...
+
+            $result = null;
+            foreach ($destLocales as $destLocale) {
+                $message = new CopyLocalePageMessage(
+                    ['uuid' => $uuid],
+                    $srcLocale,
+                    $destLocale,
+                );
+
+                /** @see \Sulu\Page\Application\MessageHandler\CopyLocalePageMessageHandler */
```
wachterjohannes, line 300: no FCQN

wachterjohannes (changes requested):
ELSE LGTM

packages/page/src/UserInterface/Controller/Admin/PageController.php
Prokyonn: So far we did always use the FCQN with @see, I think that otherwise Phpstorm cannot navigate directly to the class

wachterjohannes (approved):

packages/page/src/UserInterface/Controller/Admin/PageController.php
wachterjohannes: @Prokyonn then all good 😂

packages/page/src/UserInterface/Controller/Admin/PageController.php
```diff
@@ -286,15 +286,23 @@ private function handleAction(Request $request, string $uuid): ?PageInterface //
         }
 
         if ('copy_locale' === $action) {
-            $message = new CopyLocalePageMessage(
-                ['uuid' => $uuid],
-                (string) $request->query->get('src'),
-                (string) $request->query->get('dest'),
-            );
+            $srcLocale = (string) ($request->query->get('src') ?: $request->query->get('locale'));
+            $destLocales = \array_filter(\array_map('trim', \explode(',', (string) $request->query->get('dest'))));
```
mamazu, line 290: `$request->query->getString('dest')` would replace the string cast.

PR #8654, opened 2026-02-24
Fix crop images issue

### LINKED ISSUE #6527: Image crop is ignored when bounds exceed image
# Actual Behavior

When cropping an image you can drag the bounding box.
There's an issue that allows the box to be dragged slightly' under' the image. If this happens, the selected cropping area seems to be ignore and Sulu just uses the full selection.

It's hard to explain, but I hope the screenshots below illustrate the issue.

I'm also not entirely sure if it's related to this issue: https://github.com/sulu/sulu/issues/6481

# Expected Behavior

You should not be able to drag the bounding box outside the image.

# Steps to Reproduce

Take https://sulu.rocks/admin/#/media/de/32/details for example.

This works fine:
image
Result:
image

When I drag the box all the way to the bottom, you notice that the white 'selection border' goes slightly outside the image (this is not possible with any of the other sides): 
image
You can save it like this, and Sulu remembers the selected area (it's still configured the same when reopening the cropping tool), but in the frontend somehow the selection gets ignored and it uses the 'automatic' crop as if you did not select an area:
image

# Possbile solution

I'm not sure if the issue is actually that you can drag beyond the image bounds or that the code that generates the actual cropped image should just 'limit' to the image bounds.

Prokyonn:
| Q | A
  | --- | ---
  | Bug fix? | yes
  | New feature? | no
  | BC breaks? | no
  | Deprecations? | no
  | Fixed tickets | fixes https://github.com/sulu/sulu/issues/6527 
  | Related issues/PRs | #
  | License | MIT

  #### What's in this PR?

  - Normalized crop selection coordinates in `ImageRectangleSelection` to clamp values
  within natural image bounds, preventing out-of-bounds crop regions
  - Fixed `naturalDataToScaled` and `scaledDataToNatural` conversions to use edge-based
  arithmetic (left+width → right, top+height → bottom) instead of converting
  width/height independently, avoiding rounding drift
  - Added `normalizeCropParameters` to `ImagineImageConverter` that clamps crop
  coordinates to image dimensions and returns `null` for invalid crops
  - Converted floating-point crop selections to integer pixel coordinates in
  `CropOverlay` using floor/ceil on edges to preserve the full selected area
  - Added unit tests for all three layers (JS component, JS overlay, PHP converter)

  #### Why?

  When cropping images, floating-point rounding errors during the scaled-to-natural
  coordinate conversion could produce crop regions that extend beyond the actual image
  boundaries. This caused Imagine to throw exceptions when attempting to crop outside
  the image dimensions. The fix normalizes coordinates at every layer: the JS selection
  component, the crop overlay, and the PHP image converter.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/ImageRectangleSelection/ImageRectangleSelection.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/ImageRectangleSelection/ImageRectangleSelection.js
@@ -23,6 +23,16 @@ class ImageRectangleSelection extends React.Component<Props> {
     image: Image;
     @observable imageLoaded = false;
 
+    normalizeNaturalSelection(data: SelectionData): SelectionData {
+        const {naturalHeight, naturalWidth} = this.image;
+        const width = Math.min(Math.max(data.width, 0), naturalWidth);
+        const height = Math.min(Math.max(data.height, 0), naturalHeight);
+        const left = Math.min(Math.max(data.left, 0), naturalWidth - width);
+        const top = Math.min(Math.max(data.top, 0), naturalHeight - height);
+
+        return {height, left, top, width};
+    }
+
     naturalHorizontalToScaled = (h: number) => {
         return Math.max(h * this.scaledImageWidth / this.image.naturalWidth, 0);
     };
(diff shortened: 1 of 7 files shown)


Prokyonn:
replaced by https://github.com/sulu/sulu/pull/8694

PR #8655, opened 2026-02-24
Improve phpcr cleanup command

Prokyonn:
| Q | A
  | --- | ---
  | Bug fix? | no
  | New feature? | yes
  | BC breaks? | no
  | Deprecations? | no
  | Fixed tickets | fixes #
  | Related issues/PRs | #
  | License | MIT

  #### What's in this PR?

  - **Batch processing**: The single-node subprocess now accepts multiple UUIDs, reducing process spawn overhead. A new `--batch-size` option (default 50) controls how many nodes each subprocess handles.
  - **Stale locale cleanup**: Detects locales on PHPCR nodes that no longer exist in the webspace configuration and removes their localized properties from both default and live sessions.
  - **Invalid shadow reference cleanup**: Resets shadow-on/shadow-base properties when the shadow base locale no longer exists in the webspace.
  - **Orphaned webspace tree removal**: Detects and removes entire `/cmf/<key>` trees for webspaces that are no longer configured.
  - **Stale route locale removal**: Removes route locale directories (e.g. `/cmf/<webspace>/routes/<locale>`) that are no longer in the webspace configuration.
  - **Unit tests**: Added `PHPCRCleanupCommandTest` and `PHPCRCleanupSingleNodeCommandTest` covering the new detection and cleanup methods.

  #### Why?

  The PHPCR cleanup command previously spawned one subprocess per node, which is very slow on large repositories. Batching nodes into subprocesses significantly reduces overhead. Additionally, when webspaces or
  locales are removed from configuration, their PHPCR data was left behind — stale locale properties, orphaned webspace trees, invalid shadow references, and unused route directories would accumulate over time,
  causing confusion and wasted storage.

### DIFF
--- a/src/Sulu/Bundle/DocumentManagerBundle/Command/PHPCRCleanupSingleNodeCommand.php
+++ b/src/Sulu/Bundle/DocumentManagerBundle/Command/PHPCRCleanupSingleNodeCommand.php
@@ -19,6 +19,7 @@ use PHPCR\SessionInterface;
 use Sulu\Bundle\HttpCacheBundle\EventSubscriber\InvalidationSubscriber;
 use Sulu\Component\Content\Document\Behavior\WorkflowStageBehavior;
 use Sulu\Component\Content\Document\Subscriber\PHPCR\CleanupNode;
+use Sulu\Component\Content\Document\Subscriber\ShadowLocaleSubscriber;
 use Sulu\Component\Content\Document\WorkflowStage;
 use Sulu\Component\Content\Metadata\Factory\StructureMetadataFactoryInterface;
 use Sulu\Component\DocumentManager\Behavior\Mapping\UuidBehavior;
@@ -65,6 +66,11 @@ class PHPCRCleanupSingleNodeCommand extends Command
      * @var array<string, string>
      */
     private array $aliasMapping = [];
+
+    /**
+     * @var array<string, array<string, mixed>>|null
+     */
+    private ?array $cachedLocalesByWebspaces = null;
 
     /**
      * @var array<string, OptionsResolver>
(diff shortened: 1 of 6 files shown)


wachterjohannes (approved):

src/Sulu/Bundle/DocumentManagerBundle/Command/PHPCRCleanupSingleNodeCommand.php
```diff
@@ -108,28 +111,138 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
+
+        if ($totalStats['nodesProcessed'] > 0) {
+            return self::SUCCESS;
+        }
+
+        if ($totalStats['nodesErrored'] > 0) {
+            return self::FAILURE;
+        }
 
+        return self::IGNORED;
```
mamazu: This is a little misleading. Now that you batch process the nodes, the entire batch gets marked as "successfully processed" if only one of the nodes actually is. Maybe We should switch the errored node check up to at least give errors a higher priority.

Prokyonn: Yes I think there are definitely some areas to improve, this is just a POC, but in a project cleanup I could reduce the duration from ~1,5h to ~25 minutes so far :D

mamazu: Interesting. If you really need speed, the previous version of this command just worked with raw PHPCR. Parsing the content manually and removing properties like this. This would be the best for raw speed, but this wasn't accepted at this will not take any custom extensions into account. (Just as an FYI, this might be the way to go for a big project)

Prokyonn: Yes we need both, it must work with custom extensions, but still should be done in a reasonable time.

wachterjohannes (approved):

PR #8659, opened 2026-02-25
Add check for invalid http_cache true configuration

### LINKED ISSUE #8640: Throw exception when framework.http_cache is enabled
| Q | A
| --- | ---
| Sulu Version | all (2.6.21 and 3.0.4)
| PHP Version | 8.4.12 (unrelated)
| DB Version | MySQL 8.4 (unrelated)
| Browser Version | Firefox 147

# Actual Behavior

Sulu is not compatible that the http_cache is enabled on framework Symfony level via:

```yaml
when@prod:
    framework:
        http_cache: true
```

This make the cache misbehave and not work as expected.

# Expected Behavior

We should validate this misconfiguration and throw a error inside a SuluHttpCache CompilerPass. As HttpCache is in our case handled in the `index.php` file before Framework is booted.

draconivis:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #8640
| Related issues/PRs | #8640
| License | MIT
| Documentation PR | -

#### What's in this PR?

Adds a Compiler Pass that checks if the framework.http_cache is set to true.

#### Why?

#8640 

#### Example Usage

```yaml
# framework.yaml
framework:
	http_cache: true
```
With `http_cache: true` the Compiler Pass will throw an exception depending on if the service is public as it always exists: https://github.com/symfony/symfony/blob/0fcf3560679ef39e85d7b196ba3d2e14513d43c6/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php#L934-L936

### DIFF
--- a/src/Sulu/Bundle/HttpCacheBundle/DependencyInjection/SuluHttpCacheCompilerPass.php
+++ b/src/Sulu/Bundle/HttpCacheBundle/DependencyInjection/SuluHttpCacheCompilerPass.php
@@ -1,0 +1,25 @@
+<?php
+
+/*
+ * This file is part of Sulu.
+ *
+ * (c) Sulu GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\HttpCacheBundle\DependencyInjection;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+class SuluHttpCacheCompilerPass implements CompilerPassInterface
+{
+    public function process(ContainerBuilder $container): void
+    {
+        if ($container->hasDefinition('http_cache') && $container->getDefinition('http_cache')->isPublic()) {
+            throw new \InvalidArgumentException('Enabling the Symfony Http Cache is not compatible with Sulu! Cache is handled in index.php!');
+        }
+    }
+}
--- a/src/Sulu/Bundle/HttpCacheBundle/DependencyInjection/SuluHttpCacheCompilerPass.php
+++ b/src/Sulu/Bundle/HttpCacheBundle/DependencyInjection/SuluHttpCacheCompilerPass.php
@@ -14,6 +14,9 @@ namespace Sulu\Bundle\HttpCacheBundle\DependencyInjection;
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 
+/**
+ * @internal no backwards compatibility promise is given for this class it can be removed at any time
+ */
 class SuluHttpCacheCompilerPass implements CompilerPassInterface
 {
     public function process(ContainerBuilder $container): void
--- a/src/Sulu/Bundle/HttpCacheBundle/SuluHttpCacheBundle.php
+++ b/src/Sulu/Bundle/HttpCacheBundle/SuluHttpCacheBundle.php
@@ -11,8 +11,15 @@
 
 namespace Sulu\Bundle\HttpCacheBundle;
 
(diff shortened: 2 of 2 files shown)


alexander-schranz (approved):

src/Sulu/Bundle/HttpCacheBundle/DependencyInjection/SuluHttpCacheCompilerPass.php
```diff
@@ -0,0 +1,25 @@
...
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Sulu\Bundle\HttpCacheBundle\DependencyInjection;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+class SuluHttpCacheCompilerPass implements CompilerPassInterface
```
alexander-schranz, line 20:
```suggestion
/**
 * @internal no backwards compatibility promise is given for this class it can be removed at any time
 */
class SuluHttpCacheCompilerPass implements CompilerPassInterface
```

src/Sulu/Bundle/HttpCacheBundle/DependencyInjection/SuluHttpCacheCompilerPass.php
```diff
@@ -0,0 +1,28 @@
...
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+/**
+ * @internal no backwards compatibility promise is given for this class it can be removed at any time
+ */
+class SuluHttpCacheCompilerPass implements CompilerPassInterface
+{
+    public function process(ContainerBuilder $container): void
+    {
+        if ($container->hasDefinition('http_cache') && $container->getDefinition('http_cache')->isPublic()) {
```
alexander-schranz, line 24: @draconivis one question why did we check for `isPublic` here?

draconivis: because the service is always defined, but depending on whether the value is `true` or `false`, the service is public or not

alexander-schranz: Thx, I see for later reference: https://github.com/symfony/symfony/blob/0fcf3560679ef39e85d7b196ba3d2e14513d43c6/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php#L934-L936

fnagel:
@alexander-schranz After the latest dependency update, I'm running into this warning.  Can you help me out here? 

Does Sulu work with the Symfony HTTP Cache or not? [This](https://docs.sulu.io/2.x/bundles/http_cache.html#symfony-http-cache) reads as it does work with it, but I guess I need to disable `http_cache` in the config anyway as the usage of the Symfony HTTP Cache is [controlled in the index.php](https://github.com/sulu/sulu/blob/2.6/public/index.php#L58), right?

alexander-schranz:
@fnagel Yes it works with Symfony Http Cache but not with the Framework integration of the Http Cache. It handles http caching in the index.php before any Kernel is even booted, specially because of its very special two Kernel Setup Sulu uses.

fnagel:
@alexander-schranz Thanks for the response, highly appreciated!
 
So, just to be sure: disabling the Symfony config is fine and should have no side effects on caching as long as I make sure the index.php has the caching enabled, right?

Wishing you a wonderful Easter holiday!

alexander-schranz:
@fnagel yes you can always orientate by current configs in [sulu/skeleton](https://github.com/sulu/skeleton) if you are unsure about what are the Sulu defaults. Thx wish you the same :).

PR #8662, opened 2026-02-27
Fix single_select default_value not applied inside blocks

steeven-th:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | 
| Related issues/PRs | #4682, #6115
| License | MIT
| Documentation PR | 

#### What's in this PR?

Fixes `single_select` `default_value` not being applied when the field is used inside a block. The select displays "PLEASE CHOOSE" instead of the configured default value. The same `single_select` with `default_value` works correctly at page level (outside blocks).

Two issues were identified and fixed:

1. **`FieldRenderer.handleChange` drops the `ChangeContext` parameter**
   The handler signature only accepted `(name, value)`, silently discarding the third `context` argument passed by `Field.js`. This `context` carries `{isDefaultValue: true}`, which is needed by the form store to apply defaults without marking the form as dirty (introduced in #6115). The fix propagates `context` through `FieldRenderer` → `FieldBlocks.handleBlockChange` → `onChange`.

2. **`SingleSelect` calls `onChange` in the `constructor`**
   Calling `onChange` during construction is a React anti-pattern. At that point, the component tree is not fully mounted, and MobX observable updates triggered through the `FieldBlocks` chain are not properly propagated. Moving the default value logic to `componentDidMount` ensures the component is fully mounted and the observable chain works correctly.

Both fixes are required for a complete solution: `componentDidMount` ensures proper timing for MobX observable propagation in blocks, and context propagation ensures the form is not incorrectly marked as dirty when defaults are applied.

#### Why?

When adding a block containing a `single_select` with `default_value`, the default is never applied. Users see "PLEASE CHOOSE" and must manually select a value, even though the XML template specifies a default.

**Steps to reproduce:**
1. Create a page template with a block containing a `single_select` with `default_value`
2. Create a page using this template
3. Click "Add block"
4. The `single_select` inside the block shows "PLEASE CHOOSE" instead of the default value
5. A `single_select` with the same configuration at page level (outside blocks) works correctly

#### Example Usage

```xml
<block name="content" default-type="text">
    <types>
        <type name="text">
            <properties>
                <property name="alignment" type="single_select">
                    <params>
                        <param name="default_value" value="left"/>
                            <param name="left"><meta><title lang="en">Left</title></meta></param>
                            <param name="center"><meta><title lang="en">Center</title></meta></param>
                        </param>
                    </params>
                </property>
            </properties>
        </type>
    </types>
</block>
```

#### To Do

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
@@ -14,7 +14,7 @@ import conditionDataProviderRegistry from '../Form/registries/conditionDataProvi
 import {getDifference} from '../../utils/DifferenceCalculator';
 import blockPreviewTransformerRegistry from './registries/blockPreviewTransformerRegistry';
 import FieldRenderer from './FieldRenderer';
-import type {BlockError, FieldTypeProps, FormStoreInterface} from '../Form/types';
+import type {BlockError, ChangeContext, FieldTypeProps, FormStoreInterface} from '../Form/types';
 import type {BlockEntry} from './types';
 import type {Message} from '../../components/BlockCollection/types';
 
@@ -267,7 +267,7 @@ class FieldBlocks extends React.Component<FieldTypeProps<Array<BlockEntry>>> {
         this.value = value;
     };
 
-    handleBlockChange = (index: number, name: string, value: Object, context?: Object) => {
+    handleBlockChange = (index: number, name: string, value: Object, context?: ChangeContext) => {
         const {onChange} = this.props;
         const oldValues = this.value;
 
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
@@ -267,7 +267,7 @@ class FieldBlocks extends React.Component<FieldTypeProps<Array<BlockEntry>>> {
         this.value = value;
     };
 
-    handleBlockChange = (index: number, name: string, value: Object) => {
+    handleBlockChange = (index: number, name: string, value: Object, context?: Object) => {
         const {onChange} = this.props;
         const oldValues = this.value;
 
@@ -280,7 +280,7 @@ class FieldBlocks extends React.Component<FieldTypeProps<Array<BlockEntry>>> {
 
         this.setValue(newValues);
 
-        onChange(newValues);
+        onChange(newValues, context);
     };
 
     handleBlocksChange = (value: Object) => {
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
@@ -56,7 +56,12 @@ class FieldBlocks extends React.Component<FieldTypeProps<Array<BlockEntry>>> {
         const {types: oldTypes} = prevProps;
 
         if (!equals(toJS(prevProps.value), toJS(value))){
-            this.setValue(value);
+            // Only sync from props if no local changes were made since the last render.
+            // This prevents stale echoed values from overwriting local changes
+            // (e.g., default values applied by field components during mount).
+            if (!this.value || equals(toJS(this.value), toJS(prevProps.value))) {
+                this.setValue(value);
+            }
         }
 
         if (!types || !oldTypes) {
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldRenderer.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldRenderer.js
@@ -11,7 +11,7 @@ type Props = {|
     errors?: ErrorCollection,
     formInspector: FormInspector,
     index: number,
-    onChange: (index: number, name: string, value: *) => void,
+    onChange: (index: number, name: string, value: *, context?: Object) => void,
     onFieldFinish: ?() => void,
     onSuccess: ?() => void,
     router: ?Router,
@@ -26,9 +26,9 @@ export default class FieldRenderer extends React.Component<Props> {
         showAllErrors: false,
     };
 
-    handleChange = (name: string, value: *) => {
+    handleChange = (name: string, value: *, context?: Object) => {
         const {index, onChange} = this.props;
-        onChange(index, name, value);
+        onChange(index, name, value, context);
     };
 
     render() {
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldRenderer.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldRenderer.js
@@ -2,7 +2,7 @@
 import React from 'react';
 import Router from '../../services/Router';
 import {FormInspector, Renderer} from '../Form';
-import type {ErrorCollection} from '../Form/types';
+import type {ChangeContext, ErrorCollection} from '../Form/types';
 import type {Schema} from '../Form';
 
 type Props = {|
@@ -11,7 +11,7 @@ type Props = {|
     errors?: ErrorCollection,
     formInspector: FormInspector,
     index: number,
-    onChange: (index: number, name: string, value: *, context?: Object) => void,
+    onChange: (index: number, name: string, value: *, context?: ChangeContext) => void,
     onFieldFinish: ?() => void,
     onSuccess: ?() => void,
     router: ?Router,
@@ -26,7 +26,7 @@ export default class FieldRenderer extends React.Component<Props> {
         showAllErrors: false,
     };
 
-    handleChange = (name: string, value: *, context?: Object) => {
+    handleChange = (name: string, value: *, context?: ChangeContext) => {
         const {index, onChange} = this.props;
         onChange(index, name, value, context);
     };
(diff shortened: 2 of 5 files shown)


manuxi:
Sounds like it's solving the issue I've described here? https://github.com/sulu/sulu/issues/8512
Unfortunately it's not solved although it's closed.

alexander-schranz:
@manuxi the 3.0 the problems are fixed and works like in 2.6. Currently if you want to have a default_value keep in mind for `single_select` you need also set the single_select to `mandatory`.

We need discuss if we want to break that behaviour or not. So at current state I recommend go with:

```xml
<property name="list4" type="single_select" mandatory="true"><!-- mandatory -->
    <meta>
        <title lang="en">Mandotory Single Select</title>
    </meta>

    <params>
        <param name="default_value" value="option2"/>

            <param name="option1">
                <meta>
                    <title lang="en">Option 1</title>
                </meta>
            </param>

            <param name="option2">
                <meta>
                    <title lang="en">Option 2</title>
                </meta>
            </param>
        </param>
    </params>
</property>
```

steeven-th:
Thanks for the feedback @alexander-schranz!

I understand the `mandatory="true"` workaround, and it does make the default value appear. However, I think the underlying issue is still worth addressing:

1. **`default_value` and `mandatory` are conceptually independent.** A non-mandatory select can still have a sensible default pre-selected. Requiring `mandatory` for `default_value` to work feels like an implementation side-effect rather than intentional design. For example, a "text alignment" select with `default_value="left"` shouldn't need to be mandatory — the user should simply be able to clear it if needed.

2. **The technical bug is real and broader than `single_select`.** The `ChangeContext` (carrying `{isDefaultValue: true}`) is silently dropped in `FieldRenderer.handleChange` because the handler signature only accepts `(name, value)` and discards the third argument. This affects **any field type inside a block** that relies on `ChangeContext` propagation — not just `single_select`. Fixing this propagation seems like a safe improvement regardless of the `mandatory` discussion.

Happy to adapt the PR if you'd prefer a different scope, or close it if you consider the current behavior intentional. Just let me know!

alexander-schranz:
@steeven-th was just show a workaround for manuxi issue. We will have a look at the pull request changes asap. One additional question we need check why it currently only effects the single_select and why the default works for the select type. Did you find out what the select does differently that it did work there without any changes?

steeven-th:
Thanks for the question! I looked into this and found that **the `Select` (multi-select) field has the exact same issue inside blocks** — default values are not applied either.

Both components use the same pattern to apply defaults:
- `SingleSelect.js` (line 28): `onChange(defaultValue, {isDefaultValue: true})` in the constructor
- `Select.js` (line 40): `onChange(defaultValues, {isDefaultValue: true})` in the constructor

The code is identical, so the bug affects both equally.

I also tested with `mandatory="true"` on both `single_select` and `select` inside a block, and **it does not change the behavior** — the default values are still not applied.

This confirms that the root cause is the `ChangeContext` being dropped in `FieldRenderer.handleChange`, which affects any field type inside a block that relies on context propagation for default values.

manuxi:
> Currently if you want to have a default_value keep in mind for `single_select` you need also set the single_select to `mandatory`

Sorry to not have pointed that out (as you can see the asterisks in the video): all those fields are mandatory.

Template (this is included):
```xml
<?xml version="1.0" encoding="utf-8"?>
    <properties>
        <property name="class" type="single_select" mandatory="true" colspan="4">
            <meta>
                <title lang="en">Block Format</title>
                <title lang="de">Formatierung des Blocks</title>
            </meta>
            <params>
                <param name="default_value" value="block-white"/>

                    <param name="block-white">
                        <meta>
                            <title lang="en">Background Color: White</title>
                            <title lang="de">Hintergrundfarbe: Weiß</title>
                        </meta>
                    </param>
                    <param name="block-white-90">
                        <meta>
                            <title lang="en">Background Color: White (10% Transparency)</title>
                            <title lang="de">Hintergrundfarbe: Weiß (10% Transparenz)</title>
                        </meta>
                    </param>
                    <!-- more -->
                </param>
            </params>
        </property>
    </properties>
</template>
```

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldRenderer.js
```diff
@@ -11,7 +11,7 @@ type Props = {|
     errors?: ErrorCollection,
     formInspector: FormInspector,
     index: number,
-    onChange: (index: number, name: string, value: *) => void,
+    onChange: (index: number, name: string, value: *, context?: Object) => void,
```
Prokyonn: Could we use here the `ChangeContext` instead of `Object`? (applies also for all other added context parameters)

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/tests/FieldRenderer.test.js
```diff
@@ -118,7 +118,32 @@ test('Should call onChange callback with correct index', () => {
 
     formRenderer.find(Renderer).prop('onChange')('test', 'value');
 
-    expect(changeSpy).toBeCalledWith(2, 'test', 'value');
+    expect(changeSpy).toBeCalledWith(2, 'test', 'value', undefined);
+});
+
+test('Should pass context through onChange callback', () => {
```
Prokyonn, line 124: Not sure if this is easily possible, but can we add a test to verify, that the ResourceStore is not dirty after applying the default?

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldRenderer.js
steeven-th: Done! Replaced all `Object` types with `ChangeContext` in `FieldRenderer.js` (Props type + `handleChange` signature) and `FieldBlocks.js` (`handleBlockChange` signature).

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/tests/FieldRenderer.test.js
steeven-th: The existing test `"Should set correct default values for multiple single_select in blocks"` already verifies that `onChange` is called with `{isDefaultValue: true}` context when defaults are applied inside blocks.

The dirty flag itself is managed by `ResourceFormStore.change()`, which uses `resourceStore.set()` instead of `resourceStore.change()` when `isDefaultValue` is true — this is already covered by the `ResourceFormStore` tests (`"Should call set method of ResourceStore for default data"`).

A full integration test verifying `dirty === false` after default application inside blocks would require not mocking `ResourceFormStore`, which would significantly increase the test scope. Since the context propagation and the dirty flag logic are both independently tested, I think the current coverage is sufficient. Let me know if you'd prefer a different approach!

Prokyonn: Sounds good, thanks for the detailed explanation :)

Prokyonn (approved):

Prokyonn:
Looks good from my side, tested it locally and works as expected 👍 

@alexander-schranz do you also want to take a quick look?

Prokyonn:
> Thanks for the question! I looked into this and found that **the `Select` (multi-select) field has the exact same issue inside blocks** — default values are not applied either.
> 
> Both components use the same pattern to apply defaults:
> 
> * `SingleSelect.js` (line 28): `onChange(defaultValue, {isDefaultValue: true})` in the constructor
> * `Select.js` (line 40): `onChange(defaultValues, {isDefaultValue: true})` in the constructor
> 
> The code is identical, so the bug affects both equally.
> 
> I also tested with `mandatory="true"` on both `single_select` and `select` inside a block, and **it does not change the behavior** — the default values are still not applied.
> 
> This confirms that the root cause is the `ChangeContext` being dropped in `FieldRenderer.handleChange`, which affects any field type inside a block that relies on context propagation for default values.

@steeven-th Thinking about this again... shouldn't we also adjust the Select.js then to also work? Can you verify this please. I thought when I tested it on your branch, that already both worked, but then the changes in the SingleSelect would be not necessary 🤔

steeven-th:
@Prokyonn You're right! The `ChangeContext` propagation fix in `FieldRenderer`/`FieldBlocks` is what actually solves the issue for both `SingleSelect` and `Select` — since the root cause was the context being dropped in the handler chain.

The `constructor` → `componentDidMount` change in `SingleSelect` was an extra precaution, but if both field types already work with just the propagation fix, it's not strictly necessary.

I can revert the `SingleSelect.js` change to keep the PR minimal (only `FieldRenderer` + `FieldBlocks` context propagation) if you'd prefer. Let me know!

Prokyonn:
> @Prokyonn You're right! The `ChangeContext` propagation fix in `FieldRenderer`/`FieldBlocks` is what actually solves the issue for both `SingleSelect` and `Select` — since the root cause was the context being dropped in the handler chain.
> 
> The `constructor` → `componentDidMount` change in `SingleSelect` was an extra precaution, but if both field types already work with just the propagation fix, it's not strictly necessary.
> 
> I can revert the `SingleSelect.js` change to keep the PR minimal (only `FieldRenderer` + `FieldBlocks` context propagation) if you'd prefer. Let me know!

@steeven-th Thank you for the confirmation! I’d prefer to keep the PR minimal and revert the SingleSelect. The most important thing is to have both files consistent, since they’re more or less doing the same thing. :)

src/Sulu/Bundle/AdminBundle/Resources/js/containers/FieldBlocks/FieldBlocks.js
```diff
@@ -56,7 +56,12 @@ class FieldBlocks extends React.Component<FieldTypeProps<Array<BlockEntry>>> {
         const {types: oldTypes} = prevProps;
 
         if (!equals(toJS(prevProps.value), toJS(value))){
-            this.setValue(value);
+            // Only sync from props if no local changes were made since the last render.
+            // This prevents stale echoed values from overwriting local changes
+            // (e.g., default values applied by field components during mount).
+            if (!this.value || equals(toJS(this.value), toJS(prevProps.value))) {
+                this.setValue(value);
```
Prokyonn, line 63: Without this change the default values did not work for me in multiple nested levels of blocks

alexander-schranz (approved):

alexander-schranz:
Thx @steeven-th @Prokyonn

PR #8663, opened 2026-02-27
Fix media deletion when physical file is missing from storage

steeven-th:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | 
| Related issues/PRs | 
| License | MIT
| Documentation PR | 

#### What's in this PR?

When a media file is referenced in the database but physically missing from storage (disk, S3, etc.), Sulu crashes during deletion because the trash handler tries to `move()` a non-existent file.

This PR adds a **force-delete mechanism** that detects the missing file and allows the user to delete the database record directly, skipping the trash step.

**Backend changes:**
- `MediaManager::delete()` accepts a new `bool $force = false` parameter
- `MediaTrashItemHandler::store()` wraps the `storage->move()` call in a try/catch to detect `FileNotFoundException` from Flysystem and re-throws as `SuluFileNotFoundException`
- `MediaController::deleteAction()` accepts a `?force=true` query parameter and catches `FileNotFoundException`
- New `MediaRemovedNoTrashEvent` for activity logging when force-deleting
- New `MediaNotFoundException` REST exception with HTTP 410 Gone status (code 1107)
- `FlattenExceptionNormalizer` supports the new exception interface

**Frontend changes:**
- **Form view** (`DeleteToolbarAction.js`): On 410 response, shows a confirmation dialog asking the user to force-delete. Cancel closes the dialog, Confirm retries with `?force=true`.
- **List view** (`List.js`): On 410 response, shows an informational dialog explaining that force-delete is only available for single items (not batch). Both OK and Cancel close the dialog.

**No BC breaks:** The `StorageInterface::move()` signature is unchanged.

#### Why?

In production environments, media files can go missing from storage for various reasons (failed migrations, manual cleanup, storage failures, etc.). When this happens, the corresponding database entries become impossible to delete through the admin UI, as the trash system crashes trying to move the missing file.

This fix provides a graceful degradation path: the admin is informed that the physical file is missing and can choose to remove the database record anyway.

#### Example Usage

```
# Force-delete a media whose file is missing
DELETE /api/media/42?force=true
```

In the admin UI, the user simply clicks "Delete" on the media. If the file is missing, a dialog appears asking whether to force-delete (Form view) or informing that individual deletion is required (List view).

#### To Do

#### Compatibility notes for 3.0

The following files have diverged significantly in `3.0` and may require manual adaptation when merging:
- `FlysystemStorage.php` — Flysystem v3 API (`rename()` → `move()`)
- `LocalStorage.php` — removed in 3.0
- `DeleteToolbarAction.js` — refactored (observable `deleteLocale`, simplified `delete()`)
- `MediaManager.php` — strict typing, `DateTimeImmutable`

The core logic (try/catch in `MediaTrashItemHandler`, new exception classes, frontend dialog handling) should merge cleanly.


Prokyonn:
@steeven-th Thanks for the PR!
 We discussed this internally and decided that adding an additional dialog for this case isn’t necessary. Could you please refactor the code to silently catch the exception from the  TrashManager  and remove the entity?

If it’s easily possible, we should also add extra data to the Activity Log entry so that it’s clear the media has been completely removed and can’t be restored.

This situation typically only occurs during development or when something is removed from the  var/uploads  directory, so we don’t think there’s a need to add extra dialogs here.

steeven-th:
@Prokyonn Thanks for the feedback! I've refactored the PR as requested:

**What changed:**
- Removed the frontend force-delete dialog entirely (no more 410 handling, no extra dialogs)
- Removed the `force` query parameter and backend force-delete logic
- Instead, `MediaManager::delete()` now wraps `TrashManager::store()` in a **try/catch** — if trash storage fails (e.g. physical file missing), deletion continues silently
- The activity log differentiates between the two cases:
  - `MediaRemovedEvent` → normal deletion with trash backup
  - `MediaRemovedNoTrashEvent` → deletion without trash (activity log shows *"has removed the media without moving it to the trash"*)
- Cleaned up all related REST exception classes (`MediaNotFoundException`, `MediaNotFoundExceptionInterface`, `EXCEPTION_CODE_MEDIA_NOT_FOUND`)

**Net result:** 63 additions, 299 deletions — much simpler approach.

All existing tests pass, and two new tests cover both scenarios (trash success + trash failure).

src/Sulu/Bundle/MediaBundle/Controller/MediaController.php
```diff
@@ -358,13 +358,11 @@ public function putAction($id, Request $request)
     /**
      * Delete a media with the given id.
      *
-     * @param int $id
-     *
      * @return Response
      */
-    public function deleteAction($id)
+    public function deleteAction(int $id)
```
Prokyonn: due to our BC constraints, we cannot add types in minor upgrades, please revert it

src/Sulu/Bundle/MediaBundle/Resources/translations/admin.en.json
```diff
@@ -93,6 +93,7 @@
     "sulu_activity.description.media.modified": "{userFullName} has changed the media \"{resourceTitle}\"",
     "sulu_activity.description.media.moved": "{userFullName} has moved the media \"{resourceTitle}\" from \"{context_previousCollectionTitle}\" to \"{context_newCollectionTitle}\"",
     "sulu_activity.description.media.removed": "{userFullName} has removed the media \"{resourceTitle}\"",
+    "sulu_activity.description.media.removed_no_trash": "{userFullName} has removed the media \"{resourceTitle}\" without moving it to the trash",
```
Prokyonn: For me this sounds like the user intentionally force removed it without putting it in the trash. But in reality the file was broken/missing and there was no other option. 

Any idea how we could rephrase that?
WDYT? @alexander-schranz @chirimoya

src/Sulu/Bundle/MediaBundle/Controller/MediaController.php
steeven-th: Good catch, reverted - the `int` type hints on `deleteAction($id)` and the closure have been removed.

src/Sulu/Bundle/MediaBundle/Resources/translations/admin.en.json
steeven-th: You're right, the current wording is misleading. How about something that makes it clear this was due to a missing file rather than an intentional choice?

**Suggestion:**
> `{userFullName} has removed the media "{resourceTitle}" (the file could not be moved to the trash because it was missing from storage)`

German:
> `{userFullName} hat das Medium "{resourceTitle}" gelöscht (die Datei konnte nicht in den Papierkorb verschoben werden, da sie im Speicher nicht gefunden wurde)`

Happy to adjust the wording once you and @alexander-schranz / @chirimoya agree on the phrasing.

Prokyonn: Your suggestion sounds good for me 👍 

ping @alexander-schranz @chirimoya  😬

src/Sulu/Bundle/MediaBundle/Media/Manager/MediaManager.php
```diff
@@ -651,10 +654,20 @@ public function delete($id, $checkSecurity = false)
...
+        $trashStored = false;
         if (null !== $this->trashManager) {
-            $this->trashManager->store(MediaInterface::RESOURCE_KEY, $mediaEntity);
+            try {
+                $this->trashManager->store(MediaInterface::RESOURCE_KEY, $mediaEntity);
+                $trashStored = true;
+            } catch (\Exception $e) {
+                // Trash storage failed (e.g. physical file missing), continue with deletion
+            }
         }
```
Prokyonn:
```suggestion
        $trashDisabledOrStored = true;
if (null !== $this->trashManager) {
            try {
                $this->trashManager->store(MediaInterface::RESOURCE_KEY, $mediaEntity);
            } catch (\Exception $e) {
                    $trashDisabledOrStored = false;
                // Trash storage failed (e.g. physical file missing), continue with deletion
            }
}
```

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/List.js
```diff
@@ -15,7 +15,8 @@ import SingleListOverlay from '../SingleListOverlay';
 import {translate} from '../../utils';
 import DeleteReferencedResourceDialog from '../DeleteReferencedResourceDialog';
 import DeleteDependantResourcesDialog from '../DeleteDependantResourcesDialog';
-import {ERROR_CODE_DEPENDANT_RESOURCES_FOUND, ERROR_CODE_REFERENCING_RESOURCES_FOUND} from '../../constants';
+import {ERROR_CODE_DEPENDANT_RESOURCES_FOUND,
+    ERROR_CODE_REFERENCING_RESOURCES_FOUND} from '../../constants';
```
alexander-schranz:
```suggestion
import {ERROR_CODE_DEPENDANT_RESOURCES_FOUND, ERROR_CODE_REFERENCING_RESOURCES_FOUND} from '../../constants';
```

src/Sulu/Bundle/AdminBundle/Resources/js/views/Form/toolbarActions/DeleteToolbarAction.js
```diff
@@ -19,7 +19,6 @@ export default class DeleteToolbarAction extends AbstractFormToolbarAction {
     @observable showDialog: boolean = false;
     @observable referencingResourcesData: ?ReferencingResourcesData = undefined;
     @observable dependantResourcesData: ?DependantResourcesData = undefined;
```
alexander-schranz, line 21:
```suggestion
    @observable dependantResourcesData: ?DependantResourcesData = undefined;

```

alexander-schranz (commented):
think we can revert all JS changes now:

PS: can you also update the Pull request description to reflect the new implementation.

src/Sulu/Bundle/MediaBundle/Media/Manager/MediaManager.php
Prokyonn: With this change the new activity is only used when the trash is enabled and storing is not possible. In the case of a disabled trash, it does not make sense to dispatch the new event :)

Prokyonn (changes requested):

src/Sulu/Bundle/MediaBundle/Media/Manager/MediaManager.php
steeven-th: Applied your suggestion — renamed to `$trashDisabledOrStored` so the `MediaRemovedNoTrashEvent` is only dispatched when trash is enabled but storing fails, not when trash is disabled.

src/Sulu/Bundle/AdminBundle/Resources/js/containers/List/List.js
steeven-th: Fixed, reverted to single-line import.

src/Sulu/Bundle/AdminBundle/Resources/js/views/Form/toolbarActions/DeleteToolbarAction.js
steeven-th: Fixed, restored the missing blank line.

src/Sulu/Bundle/MediaBundle/Resources/translations/admin.en.json
Prokyonn: Can you update the translation, rest looks good 👍 Thank you!

src/Sulu/Bundle/MediaBundle/Trash/MediaTrashItemHandler.php
```diff
@@ -94,6 +94,7 @@ public function store(object $media, array $options = []): TrashItemInterface
             foreach ($file->getFileVersions() as $fileVersion) {
                 // move original file into trash directory
                 $trashStorageOptions = \array_merge($fileVersion->getStorageOptions(), ['directory' => 'trash']);
+
```
Prokyonn:
```suggestion
```

src/Sulu/Bundle/MediaBundle/Resources/translations/admin.en.json
steeven-th: Updated both translations:

**EN:** `{userFullName} has removed the media "{resourceTitle}" (the file could not be moved to the trash because it was missing from storage)`

**DE:** `{userFullName} hat das Medium "{resourceTitle}" gelöscht (die Datei konnte nicht in den Papierkorb verschoben werden, da sie im Speicher nicht gefunden wurde)`

src/Sulu/Bundle/MediaBundle/Trash/MediaTrashItemHandler.php
steeven-th: Removed the extra blank line.

Prokyonn (approved):

Prokyonn:
Thank you @steeven-th

PR #8665, opened 2026-03-02
Add support for nested template properties

### LINKED ISSUE #5063: Use Sections for the Output
A feature suggestion that I would like to discuss.

In Sulu XML Templates it is currently possible to put the properties in a "Section".
https://docs.sulu.io/en/2.0/book/templates.html#sections

However, these sections are only used for separation in the GUI of the admin backend.
In the output of the page, for example through the Document Manager. This section is not used.
The output is a flat hirachia.

For Example
`<section name="organizationalDetails">
        

        <properties>
            <property name="startDate" type="date">
                <meta>
                    <title lang="en">Start Date</title>
                </meta>
            </property>
            <property name="endDate" type="date">
                <meta>
                    <title lang="en">End Date</title>
                </meta>
            </property>
        </properties>
    </section>`

that xml would be an JSON Output like this

`
...
startDate : "17.02.2020"
endDate : "17.02.2030"
...
`

I would now discuss. Wouldn't it bring advantages, especially with very large pages.
Use the sections to get more structure in the output

forexample
`
...
organizationalDetails : {
startDate : "17.02.2020"
endDate : "17.02.2030"
}
...
`

Prokyonn:
| Q | A
  | --- | ---
  | Bug fix? | no
  | New feature? | yes
  | BC breaks? | no
  | Deprecations? | no
  | Fixed tickets | fixes #5063
  | Related issues/PRs | #
  | License | MIT

  #### What's in this PR?

  - Adds support for using slashes in template property names to enable nested property structures (e.g. `nestedProperty/value`)
  - The slash-separated property name is resolved to its root key when mapping and resolving template data, so the API uses the root name (`nestedProperty`) rather than the full path (`nestedProperty/value`)
  - Two one-line changes in `TemplateDataMapper` and `MetadataResolver` to extract the root property name via `explode('/', $name, 2)[0]`
  - Includes unit tests for both the data mapper and metadata resolver, a functional integration test for page CRUD, and a test template fixture

  #### Why?

  Currently, Sulu templates produce a flat hierarchy for all properties regardless of how they are organized. As discussed in #5063, allowing slashes in property names enables structured/nested output in the API,
  giving developers a way to logically group related properties (e.g. `organizationalDetails/startDate`) while the API resolves them to a nested structure.

  #### Example Usage

  In a template XML:
  ```xml
  <property name="nestedProperty/value" type="text_line">
      <meta>
          <title lang="en">Nested Property</title>
      </meta>
  </property>

  The API will accept and return the data using the root key:
  {
      "nestedProperty": "Nested Value"
  }

### DIFF
--- a/packages/content/src/Application/ContentDataMapper/DataMapper/TemplateDataMapper.php
+++ b/packages/content/src/Application/ContentDataMapper/DataMapper/TemplateDataMapper.php
@@ -98,10 +98,11 @@ class TemplateDataMapper implements DataMapperInterface
         $localizedData = [];
         foreach ($metadata->getFlatFieldMetadata() as $property) {
             $name = $property->getName();
+            $name = \explode('/', $name, 2)[0];
 
             $isMultilingual = $property->isMultilingual();
 
-            $value = $isMultilingual ? $defaultLocalizedData[$name] ?? null : $defaultLocalizedData[$name] ?? null;
+            $value = $isMultilingual ? $defaultLocalizedData[$name] ?? null : $unlocalizedData[$name] ?? null;
             if (\array_key_exists($name, $data)) { // values not explicitly given need to stay untouched for e.g. for shadow pages urls
                 $hasAnyValue = true;
                 $value = $data[$name];
--- a/packages/content/src/Application/MetadataResolver/MetadataResolver.php
+++ b/packages/content/src/Application/MetadataResolver/MetadataResolver.php
@@ -35,13 +35,13 @@ class MetadataResolver
      * @param ItemMetadata[] $items
      * @param mixed[] $data
      *
-     * @return ContentView[]
+     * @return array<string, ContentView>
      */
     public function resolveItems(array $items, array $data, string $locale): array
     {
         $contentViews = [];
         foreach ($items as $key => $item) {
-            $name = $key;
+            $name = (string) $key;
             $type = $item->getType();
             if ($item instanceof SectionMetadata) {
                 $contentViews = \array_merge(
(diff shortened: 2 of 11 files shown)


packages/content/tests/Unit/Content/Application/ContentDataMapper/DataMapper/TemplateDataMapperTest.php
```diff
@@ -167,6 +167,35 @@ public function testMapDataPreview(): void
         $this->assertSame(['title' => 'Test Localized'], $localizedDimensionContent->getTemplateData());
     }
 
+    public function testMapNestedPropertyData(): void
+    {
+        $data = [
+            'template' => 'template-key',
+            'unlocalizedField' => 'Test Unlocalized',
+            'title' => 'Test Localized',
+            'nestedProperty' => 'Nested Value',
```
alexander-schranz: Would expect here:

```suggestion
            'nestedProperty' => ['value' => 'Nested Value'],
```

packages/content/src/Application/MetadataResolver/MetadataResolver.php
```diff
@@ -49,6 +49,7 @@ public function resolveItems(array $items, array $data, string $locale): array
                     $this->resolveItems($item->getItems(), $data, $locale)
                 );
             } else {
+                $name = \explode('/', $name, 2)[0];
                 $params = $item instanceof FieldMetadata ? $this->serializeFieldMetadataOptions($item) : [];
                 $contentViews[$name] = $this->resolveProperty($type, $data[$name] ?? null, $locale, $item, $params);
```
alexander-schranz: This seems wrong from my point of view and not handling that there can be multiple props:

```xml
        <property name="nestedProperty/value_1" type="text_line">
            <meta>
                <title lang="en">Nested Property 1</title>
            </meta>
        </property>

        <property name="nestedProperty/value_2" type="text_line">
            <meta>
                <title lang="en">Nested Property 2</title>
            </meta>
        </property>
```

packages/content/src/Application/ContentDataMapper/DataMapper/TemplateDataMapper.php
```diff
@@ -97,24 +97,82 @@ private function getTemplateData(
...
         $localizedData = [];
         foreach ($metadata->getFlatFieldMetadata() as $property) {
-            $name = $property->getName();
+            $path = $property->getName();
 
             $isMultilingual = $property->isMultilingual();
 
-            $value = $isMultilingual ? $defaultLocalizedData[$name] ?? null : $defaultLocalizedData[$name] ?? null;
-            if (\array_key_exists($name, $data)) { // values not explicitly given need to stay untouched for e.g. for shadow pages urls
+            [, $value] = $this->readValueByPath($defaultLocalizedData, $path);
```
alexander-schranz: Lets discuss this tomorrow as previous said I don't think we need support complexitiy in the TemplateDataMapper as it only supports root level parts.

So the first part of `/` would had been here enough.

packages/content/src/Application/ContentDataMapper/DataMapper/TemplateDataMapper.php
```diff
@@ -98,10 +98,11 @@ private function getTemplateData(
         $localizedData = [];
         foreach ($metadata->getFlatFieldMetadata() as $property) {
             $name = $property->getName();
+            $name = \explode('/', $name, 2)[0];
 
             $isMultilingual = $property->isMultilingual();
 
-            $value = $isMultilingual ? $defaultLocalizedData[$name] ?? null : $defaultLocalizedData[$name] ?? null;
+            $value = $isMultilingual ? $defaultLocalizedData[$name] ?? null : $unlocalizedData[$name] ?? null;
```
Prokyonn, line 105: Is this maybe connected to https://github.com/sulu/sulu/issues/8574 ?

Prokyonn: This was introduced with shadow functionality, so we have to properly test this (2 templates with unlocalized data)

alexander-schranz: what did you find out here?

alexander-schranz (approved):

packages/content/src/Application/ContentDataMapper/DataMapper/TemplateDataMapper.php
Prokyonn: I found some potential other issues regarding shadows, I'll open a follow up PR about this

Issue #7652, opened 2024-11-14
Template page - 2 sections with colspan=“6” appear offset


steeven-th:
| Q | A
| --- | ---
| Sulu Version | 2.6.5
| PHP Version | 8.3
| DB Version | MySQL 5.7
| Browser Version | Chromium Engine Version 130.0.6723.117

# Actual Behavior
Some side-by-side elements are offset in height.

**_Exemple 1 :_**
```xml
<section name="autopromo">
    <meta>
        <title>Autopromo</title>
    </meta>
    <properties>
        <block name="autopromo" default-type="autopromo" minOccurs="0">
            <params>
                <param name="add_button_text">
                    <meta>
                        <title>Ajouter</title>
                    </meta>
                </param>
            </params>

            <types>
                <type name="autopromo">
                    <properties>
                        <property name="image" type="media_selection" maxOccurs="1" colspan="6">
                            <meta>
                                <title>Image</title>
                            </meta>
                        </property>

                        <property name="color" type="website_settings_colors_picker" colspan="6">
                            <meta>
                                <title>Couleur de la languette</title>
                            </meta>
                        </property>

                        <property name="title" type="text_line" colspan="6">
                            <meta>
                                <title>Titre</title>
                            </meta>
                        </property>

                        <property name="sub_title" type="text_line" colspan="6">
                            <meta>
                                <title>Sous titre</title>
                            </meta>
                        </property>
                    </properties>
                </type>
            </types>
        </block>
    </properties>
</section>
```
![image](https://github.com/user-attachments/assets/dd90c026-c139-4255-a909-d9179eeb6f87)

**_Exemple 2 :_**
```xml
<section name="content_three">
    <meta>
        <title>app.content</title>
    </meta>

    <properties>
        <property name="content_three.title" type="text_line" mandatory="true">
            <meta>
                <title>app.title</title>
            </meta>
        </property>

        <property name="content_three.description" type="text_area">
            <meta>
                <title>app.description</title>
            </meta>
        </property>

        <section name="content_three.button.left" colspan="6">
            <meta>
                <title>app.left</title>
            </meta>
            <properties>
                <property name="content_three.button.left.title" type="text_line" mandatory="true" colspan="6">
                    <meta>
                        <title>app.button_title</title>
                    </meta>
                </property>

                <property name="content_three.button.left.link" type="link" mandatory="true" colspan="6">
                    <meta>
                        <title>app.button_link</title>
                    </meta>

                    <params>
[...]

PR #8669, opened 2026-03-04
Fix form grid misalignment when info_text is present on some fields

steeven-th:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | 
| Related issues/PRs | 
| License | MIT
| Documentation PR | 

#### What's in this PR?

- Always render the `description-label` div in the `Field` component, even when no `description` (`info_text`) is provided.
- This ensures consistent field heights across a row, preventing layout misalignment.

#### Why?

When fields on the same row use `colspan` to share a 12-column line, and some fields have `<info_text>` while others don't, the next row becomes visually misaligned. Fields "staircase" instead of aligning on a clean horizontal line.

**Root cause:** The grid uses `float: left` for layout. The `description-label` div (which has `min-height: 20px` in CSS) was only rendered when a description was provided. Fields with `info_text` were taller than those without, causing subsequent floated items to latch onto different vertical positions.

**Fix:** Always render the `description-label` div (even when empty), matching the existing pattern already used by the `error-label` div which is always rendered regardless of whether an error exists. The existing `min-height: 20px` CSS rule ensures consistent heights.

#### Reproduction

Create a form XML with two visual rows where only some fields have `<info_text>`. The key is having `colspan` fields that don't add up to 12 on each "visual row" concept — the float layout wraps them, and height differences cause staircase alignment on the next row.

```xml
<!-- Row 1: 3 + 2 + 2 + 2 + 3 = 12 columns, only some fields have info_text -->
<property name="field_a" type="single_select" colspan="3">
    <meta>
        <title>Font Family</title>
    </meta>
    <params>
            <param name="heading"><meta><title>Heading</title></meta></param>
            <param name="body"><meta><title>Body</title></meta></param>
            <param name="accent"><meta><title>Accent</title></meta></param>
        </param>
    </params>
</property>
<property name="field_b" type="single_select" colspan="2">
    <meta>
        <title>Font Weight</title>
    </meta>
    <params>
            <param name="400"><meta><title>400 (Regular)</title></meta></param>
            <param name="700"><meta><title>700 (Bold)</title></meta></param>
        </param>
    </params>
</property>
<property name="field_c" type="text_line" colspan="2">
    <meta>
        <title>Font Size</title>
        <info_text>Use a valid CSS value (e.g. 1.5rem, 16px)</info_text>
    </meta>
</property>
<property name="field_d" type="single_select" colspan="2">
    <meta>
        <title>Font Style</title>
    </meta>
    <params>
            <param name="normal"><meta><title>Normal</title></meta></param>
            <param name="italic"><meta><title>Italic</title></meta></param>
        </param>
    </params>
</property>
<property name="field_e" type="text_line" colspan="3">
    <meta>
        <title>Line Height</title>
        <info_text>Use a unitless value (e.g. 1.5) or CSS value</info_text>
    </meta>
</property>

<!-- Row 2: same structure but WITHOUT info_text — these fields staircase -->
<property name="field_f" type="single_select" colspan="3">
    <meta>
        <title>Font Family</title>
    </meta>
    <params>
            <param name="heading"><meta><title>Heading</title></meta></param>
            <param name="body"><meta><title>Body</title></meta></param>
            <param name="accent"><meta><title>Accent</title></meta></param>
        </param>
    </params>
</property>
<property name="field_g" type="single_select" colspan="2">
    <meta><title>Font Weight</title></meta>
    <params>
            <param name="400"><meta><title>400 (Regular)</title></meta></param>
            <param name="700"><meta><title>700 (Bold)</title></meta></param>
        </param>
    </params>
</property>
<property name="field_h" type="text_line" colspan="2">
    <meta><title>Font Size</title></meta>
</property>
<property name="field_i" type="single_select" colspan="2">
    <meta><title>Font Style</title></meta>
    <params>
            <param name="normal"><meta><title>Normal</title></meta></param>
            <param name="italic"><meta><title>Italic</title></meta></param>
        </param>
    </params>
</property>
<property name="field_j" type="text_line" colspan="3">
    <meta><title>Line Height</title></meta>
</property>
```

**Expected:** Row 2 fields align on a clean horizontal line.
**Actual:** Row 2 fields "staircase" — fields below shorter ones (without `info_text`) start higher than fields below taller ones (with `info_text`).

Capture d'écran 2026-03-04 à 13 28 35

#### Long-term recommendation

This fix is minimal and appropriate for `2.6`. However, the underlying issue stems from the `float: left` layout used by the `Grid` component (`item.scss`). A more robust long-term solution for `3.0` would be to replace the float-based grid with CSS Grid:

```scss
.grid {
    display: grid;
    grid-template-columns: repeat(12, 1fr);
    column-gap: 20px;
}

.colSpan-3 { grid-column: span 3; }
// etc.
```

CSS Grid automatically creates equal-height rows, which would eliminate this entire class of layout issues without needing to ensure every field has the same internal DOM structure.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Form/Field.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Form/Field.js
@@ -140,11 +140,9 @@ class Field<T: string | number> extends React.Component<Props<T>> {
                         </label>
                     }
                     {children}
-                    {description &&
-                        <div className={fieldStyles.descriptionLabel}>
-                            {description}
-                        </div>
-                    }
+                    <div className={fieldStyles.descriptionLabel}>
+                        {description}
+                    </div>
                     <div className={fieldStyles.errorLabel}>
                         {error}
                     </div>
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Form/tests/__snapshots__/Field.test.js.snap
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Form/tests/__snapshots__/Field.test.js.snap
@@ -18,6 +18,9 @@ exports[`Display a field with a required label 1`] = `
         Test
       </div>
       <div
+        class="descriptionLabel"
+      />
+      <div
         class="errorLabel"
       />
     </div>
@@ -42,6 +45,9 @@ exports[`Display a field with an error 1`] = `
         Test
       </div>
       <div
+        class="descriptionLabel"
+      />
+      <div
         class="errorLabel"
       >
         Error! Help!
@@ -62,6 +68,9 @@ exports[`Display a field with colSpan and after space 1`] = `
       <div>
         Test
       </div>
+      <div
+        class="descriptionLabel"
+      />
       <div
         class="errorLabel"
       />
@@ -111,6 +120,9 @@ exports[`Display a field with label 1`] = `
         Test
       </p>
       <div
+        class="descriptionLabel"
+      />
+      <div
         class="errorLabel"
       />
     </div>
@@ -150,6 +162,9 @@ exports[`Display a field with label and type 1`] = `
         Test
       </p>
       <div
+        class="descriptionLabel"
+      />
+      <div
         class="errorLabel"
       />
     </div>
@@ -168,6 +183,9 @@ exports[`Display a field without label 1`] = `
       <p>
         Test
       </p>
+      <div
+        class="descriptionLabel"
+      />
       <div
         class="errorLabel"
       />
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/components/Form/tests/__snapshots__/Form.test.js.snap
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/components/Form/tests/__snapshots__/Form.test.js.snap
@@ -18,6 +18,9 @@ exports[`Render a form 1`] = `
         </label>
         Test 1
         <div
+          class="descriptionLabel"
+        />
+        <div
           class="errorLabel"
         />
       </div>
@@ -34,6 +37,9 @@ exports[`Render a form 1`] = `
           Test1
         </label>
         Test 2
+        <div
+          class="descriptionLabel"
+        />
         <div
           class="errorLabel"
         />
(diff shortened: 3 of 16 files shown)


alexander-schranz:
A floating grid is currently the expected behaviour as there are cases where big boxes are used on the left side (3) and two small boxes on the right side (9) being float to it the left one.

```
[ A 3 ] [ B 9 ]
[ A 3 ] [ C 9 ]
[ A 3 ]
```

We can currently not change the behaviour as that would break existing UIs inside different projects. I recommend for your case wrap a properties in a `<section>` tag without a meta that is a way to escape the floating grid.

steeven-th:
Thanks for the quick feedback @alexander-schranz, and for explaining the floating grid rationale — the multi-row block use case makes total sense!

I tried the `<section>` workaround, but it introduces a visible horizontal separator between each section, which isn't ideal when you just want clean rows without visual separation (e.g. repeating the same field pattern for H1, H2, H3, etc.).

Also, I'd like to point out that this fix doesn't change the float behavior at all — it only ensures the `description-label` div is always rendered (even when empty), which is already the pattern used by `error-label` (always rendered regardless of whether there's an error). The float layout continues to work exactly as before, including the multi-row block scenario you described.

The only thing that changes is that fields without `info_text` now have the same height as fields with `info_text`, preventing the staircase effect on the next wrapped line.

Would you be open to reconsidering? Happy to discuss further or adjust the approach if needed.

Capture d’écran 2026-03-04 à 16 30 53

alexander-schranz:
Thats kind of interesting thought there is no line if there is no title. But seems like there is some special handling done to get rid of the line inside the contact details.

Think if we should add some config params for sections to control there more 🤔

steeven-th:
Thanks for considering it! Yes, I tested it on my side — I added a screenshot in my previous comment showing the visual separator added by a `<section>` without a title.

A config param for sections would definitely be a great long-term solution. In the meantime, could this fix (always rendering the `description-label` div) be merged as an immediate solution? It's a one-line change, risk-free, and consistent with the existing `error-label` pattern.

Both approaches aren't mutually exclusive — the config param for sections would bring more flexibility, while this fix simply corrects an inconsistency in the Field component rendering.

mamazu:
refs #7652

PR #8670, opened 2026-03-06
Add autofocus to headlines in Overlays

mamazu:
| Q | A
| --- | ---
| Bug fix? | no 
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | #8051 
| License | MIT
| Documentation PR | -

#### What's in this PR?
Adding autofocus to more forms (mostly all the headline properties)

#### Why?
We've introduced the autofocus property in #8639 but it wasn't really used in the Sulu project. Configuring most resources to use this improves the accessibility, usability for power users and directs the user to the form field that has to be filled first (usually the name of the resource)

### DIFF
--- a/src/Sulu/Bundle/AudienceTargetingBundle/Resources/config/forms/target_group_details.xml
+++ b/src/Sulu/Bundle/AudienceTargetingBundle/Resources/config/forms/target_group_details.xml
@@ -12,7 +12,6 @@
             </meta>
             <params>
                 <param name="headline" value="true"/>
-                <param name="autofocus" value="true"/>
             </params>
         </property>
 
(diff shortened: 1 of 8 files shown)


alexander-schranz:
I'm not sure if thats a good idea. Because forms are used in add and edit the same. And most case for `edits` I don't want get autofocused while navigating through the systems.

So personally would only use autofocus for overlays which do not have multiple tabs.

/cc @sulu/core-developer

mamazu:
I noticed that as well. The reason I don't think this is relevant is, if you're currently using the mouse anyways then it doesn't matter where the focus is.
But I agree if it only would apply for create views this would be better.

For the tags this definitely makes sense as there is only one field (but the UI is still not completely keyboard compatible)

wachterjohannes:
@alexander-schranz i see it the same as @mamazu we should autofocus which makes it easier to navigate thrue the form when opening it

alexander-schranz:
@wachterjohannes as said yes for overlays, no for most forms specially for pages, articles, which triggers additional JS code like url generation on title first unfocus.

mamazu:
Yes updated the PR to only include it in the analytics pop up and the tag form. The tag form is only one field so focusing anything else wouldn't make much sense.

alexander-schranz (approved):

PR #8658, opened 2026-02-25
Improve list performance with additional indices

Prokyonn:
| Q | A
  | --- | ---
  | Bug fix? | no
  | New feature? | no
  | BC breaks? | no
  | Deprecations? | no
  | Fixed tickets | fixes #
  | Related issues/PRs | #8643
  | License | MIT
  | Documentation PR | sulu/sulu-docs#

  #### What's in this PR?

  - Added composite index on `stage`, `version`, `locale` for dimension content entities
  - Added `resource_lookup` index on the resource foreign key + `stage`, `version`, `locale`, `ghostLocale` columns
  - Added `resource_template_lookup` index for entities implementing `TemplateInterface`, covering the resource foreign key + `stage`, `version`, `locale`, `templateKey` columns
  - Introduced a `getResourceForeignKeyColumn` helper that dynamically resolves the resource foreign key column from the `dimensionContents` association mapping

  #### Why?

  Queries on dimension content tables frequently filter by resource, stage, version, and locale. Without proper composite indices, these lookups result in full table scans or suboptimal index usage, degrading
  performance as content grows. The new indices cover the most common query patterns for content resolution and template-based lookups.

PR #8678, opened 2026-03-13
Improve ghost locale handling

Prokyonn:
| Q | A
  | --- | ---
  | Bug fix? | no
  | New feature? | no
  | BC breaks? | no
  | Deprecations? | no
  | Fixed tickets | fixes #
  | Related issues/PRs | #8658
  | License | MIT

  #### What's in this PR?

  - Introduces ghost locale exclusion in `FieldDescriptorFactory` — when `excludeGhosts` is true or only a single locale is configured, ghost-related field descriptors are stripped and `case-property` fields are
  simplified to their primary branch
  - Converts duplicate `property` + `ghostProperty` pairs to `case-property` in articles, snippets, and pages list XML configs (e.g. `templateKey`)
  - Adds `DoctrineCaseFieldDescriptor::getCase1FieldDescriptor()` / `getCase2FieldDescriptor()` to expose individual branches
  - Expands `DoctrineListBuilder::createInExpression()` for case fields into an optimized OR pattern (`(a IS NOT NULL AND a IN (:v)) OR (a IS NULL AND b IN (:v))`) that allows index usage on both branches
  - Simplifies controller code in `ArticleController`, `SnippetController`, and `PageController` — removes manual ghost OR-expression building in favor of `$listBuilder->in()` on the case field
  - Adds unit and functional tests for the new behavior

  #### Why?

  Ghost locale handling in lists required every controller to manually build complex OR expressions combining `dimensionContentTemplateKey` and `ghostDimensionContentTemplateKey` fields. This was duplicated across
  articles, snippets, and pages, was error-prone, and prevented the database from using indexes efficiently on the CASE expression.

  By moving the ghost/non-ghost logic into `FieldDescriptorFactory` and teaching `DoctrineListBuilder` how to decompose case fields into indexable expressions, controllers become simpler and list queries perform
  better — especially when filtering or searching (where ghost results are excluded entirely, eliminating unnecessary joins).

  #### To Do

### DIFF
--- a/packages/page/src/UserInterface/Controller/Admin/PageController.php
+++ b/packages/page/src/UserInterface/Controller/Admin/PageController.php
@@ -87,8 +87,8 @@ final class PageController implements SecuredControllerInterface, SecuredObjectC
         $locale = $request->query->get('locale');
         $parentId = $request->query->get('parentId');
         $webspaceKey = $request->query->get('webspace');
-        $excludeGhosts = $request->query->getBoolean('exclude-ghosts', false);
-        $hasSearchOrFilter = $request->query->has('search')
+        $excludeGhostPages = $request->query->getBoolean('exclude-ghosts', false);
+        $excludeGhosts = $request->query->has('search')
             || !empty($request->query->all('filter'));
         $excludeShadows = $request->query->getBoolean('exclude-shadows', false);
         $expandedIds = \array_filter(\explode(',', (string) $request->query->get('expandedIds')));
@@ -100,7 +100,7 @@ final class PageController implements SecuredControllerInterface, SecuredObjectC
             $filters['webspaceKey'] = $webspaceKey;
         }
 
-        if ($excludeGhosts) {
+        if ($excludeGhostPages) {
             $filters['ghostLocale'] = null;
         }
 
@@ -122,7 +122,7 @@ final class PageController implements SecuredControllerInterface, SecuredObjectC
             includedFields: $includedFields,
             listKey: 'pages',
             filterByParentId: empty($ids),
-            excludeGhosts: $hasSearchOrFilter,
+            excludeGhosts: $excludeGhosts,
         );
 
         return new JsonResponse($this->normalizer->normalize(
--- a/packages/page/src/UserInterface/Controller/Admin/PageController.php
+++ b/packages/page/src/UserInterface/Controller/Admin/PageController.php
@@ -88,6 +88,8 @@ final class PageController implements SecuredControllerInterface, SecuredObjectC
         $parentId = $request->query->get('parentId');
         $webspaceKey = $request->query->get('webspace');
         $excludeGhosts = $request->query->getBoolean('exclude-ghosts', false);
+        $hasSearchOrFilter = $request->query->has('search')
+            || !empty($request->query->all('filter'));
         $excludeShadows = $request->query->getBoolean('exclude-shadows', false);
         $expandedIds = \array_filter(\explode(',', (string) $request->query->get('expandedIds')));
         $ids = $request->query->get('ids');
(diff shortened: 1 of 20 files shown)


packages/page/src/UserInterface/Controller/Admin/PageController.php
```diff
@@ -88,6 +88,8 @@ public function cgetAction(Request $request): Response
         $parentId = $request->query->get('parentId');
         $webspaceKey = $request->query->get('webspace');
         $excludeGhosts = $request->query->getBoolean('exclude-ghosts', false);
+        $hasSearchOrFilter = $request->query->has('search')
```
wachterjohannes: https://github.com/sulu/sulu/pull/8678/changes#diff-5de0e5740816284a2543e3ab6cdbe35c1e9eeb458d3b1f9c19ca3fcf32583804R89-R91

is called `$excludeGhosts` in the other controller

packages/snippet/src/UserInterface/Controller/Admin/SnippetController.php
```diff
@@ -75,23 +75,31 @@ public function cgetAction(Request $request): Response
         // TODO this should be SnippetRepository::findFlatBy / ::countFlatBy methods
         //      but first we would need to avoid that the restHelper requires the request.
         //
+        $excludeGhosts = $request->query->has('search')
```
wachterjohannes: https://github.com/sulu/sulu/pull/8678/changes#diff-5de0e5740816284a2543e3ab6cdbe35c1e9eeb458d3b1f9c19ca3fcf32583804R89-R91

is called $excludeGhosts in the other controller

wachterjohannes (changes requested):

src/Sulu/Component/Rest/ListBuilder/Metadata/FieldDescriptorFactory.php
```diff
@@ -128,7 +134,75 @@ public function getFieldDescriptors(string $listKey): ?array
             return null;
         }
 
-        return \unserialize(\file_get_contents($configCache->getPath()));
+        /** @var array<string, FieldDescriptorInterface> $fieldDescriptors */
+        $fieldDescriptors = \unserialize(\file_get_contents($configCache->getPath()));
+
+        if ($excludeGhosts || $this->isSingleLocale()) {
+            return $this->excludeGhostFieldDescriptors($fieldDescriptors);
```
alexander-schranz: This feels too wrong for me here. I don't think that field descriptors should know anything about locales or ghost. For me this should be handled in controllers.

src/Sulu/Component/Rest/ListBuilder/Metadata/FieldDescriptorFactoryInterface.php
```diff
@@ -19,7 +19,7 @@
 interface FieldDescriptorFactoryInterface
 {
     /**
-     * @return FieldDescriptorInterface[]
+     * @return FieldDescriptorInterface[]|null
      */
-    public function getFieldDescriptors(string $listKey): ?array;
+    public function getFieldDescriptors(string $listKey, bool $excludeGhosts = false): ?array;
```
alexander-schranz: field descriptors should not know about Ghost they have nothing todo with content or locales, they are just list providers.

alexander-schranz (changes requested):
I don't think that this is a good idea to handle anything of this inside field descriptors. Fiel descriptors should not be have any knowledge about ghost or locales.

alexander-schranz:
as discussed we will try with `remove...` method instead of add knowledge about ghost to field descriptors.

alexander-schranz (approved):

wachterjohannes (approved):

PR #8679, opened 2026-03-13
Fixing the time mockery

### LINKED ISSUE #8584: CollaborationRepository tests are flaky
| Q | A
| --- | ---
| Sulu Version | >=2.6
| PHP Version | PHP8.3
| DB Version |
| Browser Version |

# Actual Behavior

When running the CollaborationRepositoryTest, the tests sometimes fail and sometimes not.
It seems that the problem is in this line here:
[        $this->assertEquals($started + 10, $collaboration1->getChanged());](https://github.com/sulu/sulu/blob/07b6542cce8dd05a958c884ccd9c3db6c929579a/src/Sulu/Bundle/AdminBundle/Tests/Unit/Entity/CollaborationRepositoryTest.php#L246)

Sometimes the getChanged() result is one tick higher.

Though the `ClockMock::withClockMock(true);` setting is set, it should never be applied, since to my understanding the time and sleep-methods could never be mocked or interchanged, since the documentation states in https://symfony.com/doc/current/components/phpunit_bridge.html#clock-mocking that " "fully qualified function calls" (e.g \time()) cannot be mocked."

# Expected Behavior

Tests should always run successfully.

# Possible Solutions

Besides refactoring and introducing other dependencies for time handling, an easy fix would be something like 
`$this->assertTrue($collaboration1->getChanged() >= $started + 10);`
or
`$this->assertTrue($collaboration1->getChanged() >= $started + 10 && $collaboration1->getChanged() <= $started + 12);`

mamazu:
| Q | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| BC breaks? | no
| Deprecations? | no
| Fixed tickets | fixes #8584 
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?
Replacing the real clock with a mock clock. This speeds up the tests as there's no real waiting.

This however also requires that the functions aren't fully qualified as the Mock clock uses the php function resolution.
-> In tests: Looking into the current namespace for sleep function, finds mock uses mock
-> In prod: Looks into current namespace. Finds nothing, looks into the global namespace, finds real waiting implementation, uses wall clock.

#### Why?
Flakey tests are annoying.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Entity/Collaboration.php
+++ b/src/Sulu/Bundle/AdminBundle/Entity/Collaboration.php
@@ -91,7 +91,6 @@ class Collaboration
      */
     public function updateTime()
     {
-        $this->changed = \time();
-        dump(\date('Y-m-d H:i:s', $this->changed));
+        $this->changed = 'time'();
     }
 }
--- a/src/Sulu/Bundle/AdminBundle/Entity/Collaboration.php
+++ b/src/Sulu/Bundle/AdminBundle/Entity/Collaboration.php
@@ -42,8 +42,8 @@ class Collaboration
         private $resourceKey,
         private $id,
     ) {
-        $this->started = time();
-        $this->changed = time();
+        $this->started = \time();
+        $this->changed = \time();
     }
 
     /**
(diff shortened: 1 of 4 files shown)


mamazu:
To be honest, the easiest version would probably be to just add a setter for the `changed` property and then use a mocked clock in the `CollaborationRepository`.

PR #8680, opened 2026-03-16
Make admin config consistant with other modules

mamazu:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | no
| BC breaks? | no (making a service lazy is not a bc break)
| Deprecations? | no
| Fixed tickets | -
| Related issues/PRs | -
| License | MIT
| Documentation PR | -

#### What's in this PR?
* Making the admin config consistent with all other bundles.
* Adding phpstan error messages because of unused properties

#### Why?
* Less surprises when trying to figure out how the admin works
* Less special cases in the `AdminController` and `NavigationRegistry`
* Moving properties out of the `AdminController` in the future.


src/Sulu/Bundle/AdminBundle/Admin/SuluAdmin.php
```diff
@@ -0,0 +1,98 @@
...
+use Sulu\Bundle\AdminBundle\FieldType\FieldTypeOptionRegistryInterface;
+use Sulu\Bundle\AdminBundle\SmartContent\SmartContentProviderInterface;
+use Sulu\Bundle\ContactBundle\Contact\ContactManagerInterface;
+use Sulu\Bundle\MarkupBundle\Markup\Link\LinkProviderPoolInterface;
+use Sulu\Bundle\SecurityBundle\Entity\User;
+use Sulu\Component\Localization\Manager\LocalizationManagerInterface;
+use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface;
+use Webmozart\Assert\Assert;
+
+class SuluAdmin extends Admin
```
mamazu: This class should probably be final so that nobody messes with it

src/Sulu/Bundle/AdminBundle/Admin/SuluAdmin.php
```diff
@@ -0,0 +1,104 @@
...
+use Sulu\Bundle\ContactBundle\Contact\ContactManagerInterface;
+use Sulu\Bundle\ContactBundle\Entity\ContactAddress;
+use Sulu\Bundle\ContactBundle\Entity\ContactInterface;
+use Sulu\Bundle\MarkupBundle\Markup\Link\LinkProviderPoolInterface;
+use Sulu\Bundle\SecurityBundle\Entity\User;
+use Sulu\Component\Localization\Manager\LocalizationManagerInterface;
+use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface;
+use Webmozart\Assert\Assert;
+
+final class SuluAdmin extends Admin
```
mamazu, line 36: Add `@internal`

alexander-schranz, line 36: as discussed we should make this service `@internal` as it will maybe requires new constructor parameters if new features are added to Sulu Admin and we not provide any BC for this class:


```suggestion
/**
 * @internal no backwards compatibility promise is given for this method it could be removed or changed at any time
 *           create your own service class extending `Admin` instead
 */
final class SuluAdmin extends Admin
```

src/Sulu/Bundle/AdminBundle/Controller/AdminController.php
```diff
@@ -53,22 +52,22 @@ public function __construct(
...
         private TranslatorBagInterface $translatorBag,
         private MetadataProviderRegistry $metadataProviderRegistry,
-        private ViewRegistry $viewRegistry,
-        private NavigationRegistry $navigationRegistry,
-        private FieldTypeOptionRegistryInterface $fieldTypeOptionRegistry,
-        private ContactManagerInterface $contactManager,
-        private iterable $smartContentProviders,
-        private LinkProviderPoolInterface $linkProviderPool,
-        private LocalizationManagerInterface $localizationManager,
+        private ViewRegistry $viewRegistry, // deprecated
```
alexander-schranz: This are private variables we could already set them to `nullable`:

```suggestion
        private ?ViewRegistry $viewRegistry, // deprecated
```

and inject `null` here for this services which we not longer use here.

We just need to keep the constructor Parameters for backwards compatibility.

src/Sulu/Bundle/AdminBundle/Admin/SuluAdmin.php
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    public function getConfigKey(): string
+    {
+        return 'sulu_admin';
+    }
+
+    public static function getPriority(): int
+    {
+        return \PHP_INT_MAX - 1;
```
alexander-schranz: Would not use `PHP_INT_MAX` maybe more a `10_000` if we want be high.

As in general it can happen somebody doing SuluAdmin::getPriority() + 1 to hook in front of it, and another then SuluAdmin::getPriority() + 1 if some build on top of each other priorities. Why `- 1` is more common we I would avoid PHP_INT_MAX for any priority.

As PHP_INT_MAX can also make problems if the container is build in x64 system and run on x32 system. So the builded container should avoid usage of PHP_INT_MAX variable. As this method result is dumped into the container build we need avoid it here.

alexander-schranz (changes requested):

src/Sulu/Bundle/AdminBundle/Resources/config/services.php
```diff
@@ -100,30 +101,49 @@
...
+            '%sulu_admin.collaboration_interval%',
+            '%sulu_admin.collaboration_enabled%',
+        ])
+        ->tag('sulu.admin')
+        ->tag('sulu.context', ['context' => 'admin'])
+    ;
+
     $services->set('sulu_admin.admin_pool', AdminPool::class)
         ->public()
+        ->lazy()
```
alexander-schranz, line 146: @mamazu can you provide the stack trace or the circular loop error infos why we had to add lazy here? So we have atleast it here as a comment of this PR why this was added.

mamazu: SuluAdmin -> ViewRegistry -> AdminPool -> list of all admins which includes SuluAdmin

At least that's how I understood it. And since it's running with an iterator, this error will not trigger the Symfony circular loop detecting and just crashes with out of memory.

mamazu:
@alexander-schranz This should also be ready to merge now.

PR #8689, opened 2026-03-20
Fix partial JSON field merges in form stores

Prokyonn:
| Q | A
  | --- | ---
  | Bug fix? | yes
  | New feature? | no
  | BC breaks? | no
  | Deprecations? | no
  | Fixed tickets | fixes #
  | Related issues/PRs | #
  | License | MIT

  #### What's in this PR?

  - Replaced shallow spread merge with a `deepMerge` function in
  `AbstractFormStore.addMissingSchemaProperties()`
  - Schema-defined nested properties (e.g., `excerpt/title`,
  `seo/description`) are now correctly initialized even when the data
  contains a partial nested object
  - Added comprehensive tests covering partial, empty, fully populated,
  null, and deeply nested data scenarios

  #### Why?

  When `addMissingSchemaProperties()` merged schema defaults with existing
   data using a shallow spread (`{...schemaFields, ...this.data}`), any
  partial nested object in the data (e.g., `{excerpt: {title: 'value'}}`)
  would completely overwrite the schema-generated defaults for that key.
  This caused sibling properties like `excerpt/description` or
  `excerpt/keywords` to be lost, breaking MobX observability for those
  fields and preventing form components from rendering or reacting to
  changes on them.

  The `deepMerge` approach recursively merges nested plain objects while
  still allowing data values (including `null` and arrays) to take
  precedence over schema defaults.

### DIFF
--- a/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/AbstractFormStore.js
+++ b/src/Sulu/Bundle/AdminBundle/Resources/js/containers/Form/stores/AbstractFormStore.js
@@ -7,6 +7,28 @@ import type {Schema, SchemaEntry} from '../types';
 
 export const SECTION_TYPE = 'section';
 
+function deepMerge(base: Object, override: Object): Object {
+    const result = {...base};
+    for (const key of Object.keys(override)) {
+        const overrideVal = override[key];
+        const baseVal = result[key];
+
+        if (
+            overrideVal !== null
+            && typeof overrideVal === 'object'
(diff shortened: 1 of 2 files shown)


alexander-schranz:
@Prokyonn will have a look at this later, we should still modify the https://github.com/sulu/SuluPHPCRMigrationBundle that the excerpt looks the same like when the page is created via UI.

alexander-schranz (approved):

alexander-schranz:
@Prokyonn should we target this for 2.6?

